PR #1309, opened 2022-03-10
Shipping rates UI changes: Validation for offer free shipping and minimum order amount

### LINKED ISSUE #1289: Shipping rates UI changes: Validation for offer free shipping and minimum order amount
## User story

- Follow-up to PR https://github.com/woocommerce/google-listings-and-ads/pull/1266#pullrequestreview-895997852:

> The form can continue to save even if the "Minimum order to qualify for free shipping" field is not filled in for any country. Is this the expected result?

ecgan:

Closes #1289.

In this PR, we perform validation on the "offer free shipping" value.

- If there are non-free shipping rates, users are required to choose an offer free shipping value. If users didn't choose an option (the value is `undefined`), we add the `errors.offer_free_shipping` property.
- If the offer free shipping value is yes (`true`), and there are no minimum order amount specified, we will also add the `errors.offer_free_shipping` property.

When there are `errors.offer_free_shipping`, users won't be able to click on the "Save changes" or "Continue" button to proceed.

Technical note: I have also changed the `checkErrors` function signature. I removed the `shippingRates` parameter, since now we can use the same shipping rates value from the `values` object retrieved from Form. In the future when we move the shipping times value into Form, we should also do the same to the `checkErrors`'s `shippingTimes` parameter.

### Screenshots:

📹 Demo video with my voice annotation: 

https://user-images.githubusercontent.com/417342/157692321-f38586e1-0e98-4f58-b125-a577021c2b55.mov

You can test this in either Setup MC or Edit Free Listings.

1. Go to shipping rate section. Complete all the fields in the page. The "Save changes" / "Continue" button should be enabled.
2. Put 0 value for all shipping rates for all countries. Offer free shipping card should not be visible. The "Save changes" / "Continue" button should be enabled.
3. Change one shipping rate to non-free. Offer free shipping card should be visible with unchecked options. The "Save changes" / "Continue" button should be disabled.
4. Check "No" for offer free shipping. The "Save changes" / "Continue" button should be enabled.
5. Check "Yes" for offer free shipping, but do not enter any minimum order amount yet. The "Save changes" / "Continue" button should be disabled.
6. Enter at least one minimum order amount. The "Save changes" / "Continue" button should be enabled.

### Changelog entry

> Add - Validation for offer free shipping.

### DIFF
--- a/js/src/components/free-listings/configure-product-listings/checkErrors.js
+++ b/js/src/components/free-listings/configure-product-listings/checkErrors.js
@@ -12,12 +12,7 @@ const validShippingRateSet = new Set( [ 'automatic', 'flat', 'manual' ] );
 const validShippingTimeSet = new Set( [ 'flat', 'manual' ] );
 const validTaxRateSet = new Set( [ 'destination', 'manual' ] );
 
-const checkErrors = (
-	values,
-	shippingRates,
-	shippingTimes,
-	finalCountryCodes
-) => {
+const checkErrors = ( values, shippingTimes, finalCountryCodes ) => {
 	const errors = {};
 
 	/**
@@ -55,8 +50,8 @@ const checkErrors = (
 
 	if (
 		values.shipping_rate === 'flat' &&
-		( shippingRates.length < finalCountryCodes.length ||
-			shippingRates.some( ( el ) => el.rate < 0 ) )
+		( values.shipping_country_rates.length < finalCountryCodes.length ||
+			values.shipping_country_rates.some( ( el ) => el.rate < 0 ) )
 	) {
 		errors.shipping_rate = __(
 			'Please specify shipping rates for all the countries. And the estimated shipping rate cannot be less than 0.',
--- a/js/src/components/free-listings/configure-product-listings/checkErrors.js
+++ b/js/src/components/free-listings/configure-product-listings/checkErrors.js
@@ -26,6 +26,20 @@ const checkErrors = ( values, shippingTimes, finalCountryCodes ) => {
 	}
 
 	if (
+		values.shipping_rate === 'flat' &&
+		( values.shipping_country_rates.length < finalCountryCodes.length ||
+			values.shipping_country_rates.some( ( el ) => el.rate < 0 ) )
+	) {
+		errors.shipping_rate = __(
+			'Please specify shipping rates for all the countries. And the estimated shipping rate cannot be less than 0.',
+			'google-listings-and-ads'
+		);
+	}
+
+	/**
+	 * Check offer free shipping.
+	 */
+	if (
 		values.offer_free_shipping === undefined &&
 		values.shipping_country_rates.some( isNonFreeFlatShippingRate )
 	) {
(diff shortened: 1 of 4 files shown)


js/src/components/free-listings/configure-product-listings/checkErrors.js
```diff
@@ -27,15 +27,41 @@ const checkErrors = (
...
+		);
+	}
+
+	if (
+		values.offer_free_shipping === true &&
+		values.shipping_country_rates.every(
+			( shippingRate ) =>
+				shippingRate.options.free_shipping_threshold === undefined
+		)
+	) {
```
eason9487: 🚧 

It's unable to save when selecting `automatic` or `manual` shipping. And it would be nicer to have test cases to cover this.

Kapture 2022-03-11 at 14 08 56

js/src/components/free-listings/configure-product-listings/checkErrors.js
```diff
@@ -27,15 +27,41 @@ const checkErrors = (
...
 
+	/**
+	 * Check offer free shipping.
+	 */
+	if (
+		values.offer_free_shipping === undefined &&
+		values.shipping_country_rates.some( isNonFreeFlatShippingRate )
+	) {
+		errors.offer_free_shipping = __(
+			'Please select an option for offering free shipping.',
```
eason9487: Not sure if this meaning is a bit ambiguous. I'd like to confirm would it be better to describe with something like "Please select an option for whether to offer free shipping."?

ecgan: How is it ambiguous? 🤔 

Anyway, I have changed it based on your suggestion: [`dc8a6d9` (#1309)](https://github.com/woocommerce/google-listings-and-ads/pull/1309/commits/dc8a6d939cb542b8cf3e99f2c3c470dc19a2450c). I think that works. 😄

js/src/components/free-listings/configure-product-listings/checkErrors.js
ecgan: 🤦‍♂️ Forgot about the settings up there. Thanks @eason9487. I have fixed it in [`d290236` (#1309)](https://github.com/woocommerce/google-listings-and-ads/pull/1309/commits/d290236cc85afdbec3e9d69da5aaabe648b92579), with tests in it.

js/src/components/free-listings/configure-product-listings/checkErrors.js
eason9487: Thanks!

> How is it ambiguous?

The message is used to prompt the merchant for choosing the Yes/No option for whether to offer free shipping. "Please select an option for offering free shipping." looks like asking the merchant to offer free shipping, but my English ability can't confirm if I got the meaning wrong. 🤦‍♂️  So I wanted to double-confirm it. 🙏

eason9487 (approved):
Tested locally and works well.

js/src/components/free-listings/configure-product-listings/checkErrors.js
ecgan: Ooh now I get what you mean, and I think you have a point there, and your suggestion is clearer in that sense. Thanks for the explanation. 😄

PR #1310, opened 2022-03-10
Refactor SiteVerification and add unit tests

### LINKED ISSUE #1227: Separate unit testing for SiteVerification class
Followup issue for #1226

> The functionality for verifying a site URL was very complex to unit test as part of the account setup steps. For this reason it was moved into the SiteVerification class, it is a step in setting up the account but it's better placed within it's own class (much like the PhoneVerification class) so it can be reused by the SiteVerificationController (if those API endpoints are still used). It will need separate unit testing which can be handled in a separate PR.

In addition to creating the unit testing, we need to recheck the purpose of the SiteVerificationController. If it's still being used/needed then it can reuse some of the functionality to verify the site from the SiteVerification class instead of having the logic contained within the controller class.

mikkamp:

The purpose of this PR was to add unit testing for the site verification. Which is a followup from the verify_site function being moved out of the account controller. Unit tests are created for both SiteVerification and SiteVerificationMeta.

In addition to this the SiteVerificationController was removed. It hasn't been used for quite a while but at the time we decided to leave it in since it facilitated testing the verification step individually. At this point I think it's better we don't leave untested code which hasn't been refactored like the rest of SiteVerification (we can always revert if we need it again).

Closes #1227 

The main part is to confirm that the created unit tests work correctly. However some manual testing of the site verification is warranted since there were some refactors.

1. Disconnect Merchant Account (from Connection Test page)
2. Go through the onboarding setup again to reconnect the Merchant Account (need a public URL for the site verification)
3. Complete at least the onboarding step to connect all the accounts (the rest are optional)
4. Confirm the site verification details are saved in the `wp_options` table with the name `gla_site_verification`:
```
a:2:{
  s:8:"verified"; s:3:"yes";
  s:8:"meta_tag"; s:54:"<meta name="google-site-verification" content="abc" />";
}
```
5. Check the source code of the home page, the following meta tag should be found in the header:
```
<!-- Google site verification - Google Listings & Ads -->
<meta name="google-site-verification" content="abc" />
```

### Changelog entry
* Add - Unit testing for the site verification step.


jorgemd24 (approved):
Thanks @mikkamp, LGTM 👍 

✅  I have followed the instructions and I got the expected results.
✅  The tests are covering the main functions of SiteVerification & SiteVerificationMeta

PR #1312, opened 2022-03-10
Combine duplicate account issues per country

### LINKED ISSUE #1198: Duplicated account issues on the Product Feed page
### Describe the bug:

Appears 8 duplicated account issues on the Product Feed page.

I also check the API via [the "Try this API" function on Google website](https://developers.google.com/shopping-content/reference/rest/v2.1/accountstatuses/get). And it results in 8 **almost duplicated** account issues. The different parts are the `country` and `destination` properties.

https://github.com/woocommerce/google-listings-and-ads/blob/fe1bbf4d49b6a7d53a38403ae17bc6a807087fd8/tests/proxy/mocks/mc/content/accountstatuses.json#L6-L78

### Steps to reproduce:

💡  Since I didn't go further to survey the reasons why Google API returns such content, and this issue might not be reproducible for every MC account, I use the API response I got as the mock data.

1. Checkout `mock/1198-reproduce-duplicated-mc-account-issues` branch.
2. Follow [the proxy instructions](https://github.com/woocommerce/google-listings-and-ads/blob/fe1bbf4d49b6a7d53a38403ae17bc6a807087fd8/tests/proxy/proxy.md) to start the local proxy server.
3. Delete the transient of MC issues `wp transient delete gla_mc_statuses`.
4. Go to the Product Feed page.

### Expected behavior:

Not sure what's the expected behavior. But at least, it should not show totally the same and duplicated account issues in UI. Probably we could either show only one issue for almost the same issues or show their differences in UI.

- [x] Delete `mock/1198-reproduce-duplicated-mc-account-issues` branch after resolving this issue.

### Actual behavior:

image

image

### Additional details:

Looks like it has some relation to the target audience countries.

A few days ago, I finished onboarding setup with 'US', 'JP', 'AU' and 'TW' as target audience countries. From the raw response data, the `country` properties are also these four countries.

mikkamp:

This PR resolves the duplicate account issues by applying two changes:

1. Merge account issues with identical text and store each country in an array, which populates the applicable_countries column (note that this column is not visible in the UI)
2. Pass the destinations parameter when requesting the account status so we only get issues with `destintation = Shopping`

Closes #1198

*Note: While creating this PR I also noticed there were duplicates for variable products, since that's outside of the scope of this PR I logged it in #1316*

1. Test with a merchant account that has multiple account issues  
2. Go to Connection Test > More Merchant Center > Clear Status Cache and clear it
3. Go to Marketing > Google Listings & Ads > Product Feed and confirm the issue list doesn't have lots of duplicate account issues 
4. Check the database table `wp_gla_merchant_issues` and confirm any account issues has a list of countries which apply to the issue 

### Changelog entry
* Fix - Combine duplicate account issues per country.

### DIFF
--- a/src/API/Google/Merchant.php
+++ b/src/API/Google/Merchant.php
@@ -203,7 +203,13 @@ class Merchant implements OptionsAwareInterface {
 		$id = $id ?: $this->options->get_merchant_id();
 
 		try {
-			$mc_account_status = $this->service->accountstatuses->get( $id, $id );
+			$mc_account_status = $this->service->accountstatuses->get(
+				$id,
+				$id,
+				[
+					'destinations' => 'Shopping',
+				]
+			);
 		} catch ( GoogleException $e ) {
 			do_action( 'woocommerce_gla_mc_client_exception', $e, __METHOD__ );
 			throw new Exception( __( 'Unable to retrieve Merchant Center account status.', 'google-listings-and-ads' ), $e->getCode() );
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -372,19 +372,40 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 		$account_issues = [];
 		$created_at     = $this->cache_created_time->format( 'Y-m-d H:i:s' );
 		foreach ( $merchant->get_accountstatus()->getAccountLevelIssues() as $issue ) {
-			$account_issues[] = [
-				'product_id' => 0,
-				'product'    => __( 'All products', 'google-listings-and-ads' ),
-				'code'       => $issue->getId(),
-				'issue'      => $issue->getTitle(),
-				'action'     => __( 'Read more about this account issue', 'google-listings-and-ads' ),
-				'action_url' => $issue->getDocumentation(),
-				'created_at' => $created_at,
-				'type'       => self::TYPE_ACCOUNT,
-				'severity'   => $issue->getSeverity(),
-				'source'     => 'mc',
-			];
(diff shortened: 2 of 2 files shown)


src/API/Google/Merchant.php
```diff
@@ -203,7 +203,13 @@ public function get_accountstatus( int $id = 0 ): AccountStatus {
 		$id = $id ?: $this->options->get_merchant_id();
 
 		try {
-			$mc_account_status = $this->service->accountstatuses->get( $id, $id );
+			$mc_account_status = $this->service->accountstatuses->get(
+				$id,
+				$id,
+				[
+					'destinations' => 'Shopping',
```
ianlin, line 210: If I understand correctly from the [document](https://developers.google.com/shopping-content/guides/accountstatuses) It says `accountstatuses->get` will respond product issues with only `destination: Shopping` by default. But it looks great if adding it in the code as it's clearer that we'd only want `Shopping` destination and we don't need to check the document in order to know this infomation.

ianlin (approved):
LGTM. Tested locally using @eason9487's `mock/1198-reproduce-duplicated-mc-account-issues` branch and the solution in this PR, the duplicate issues become only one now.

src/API/Google/Merchant.php
mikkamp: I agree, I mentioned it a bit more [here](https://github.com/woocommerce/google-listings-and-ads/issues/1198#issuecomment-1064099488). We were initially expecting the default to work, but that seems to no longer give the same results so we have to force the type.

mikkamp:
Thanks for the review, greatly appreciated.

PR #1307, opened 2022-03-09
Shipping rates UI changes: Adjust the logic of prompting for unsaved shipping rates

eason9487:
### Changes proposed in this Pull Request:

Closes #1290 

- Add a new function `hasUnsavedShippingRates` for comparing the saved shipping rates to the one editing via UI.
- Apply `hasUnsavedShippingRates` to check if there are any unsaved shipping rates.

📝  As the implementation is still possible to be adjusted, test cases of the new function will be supplemented with another PR after this one gets merged.

### Screenshots:

💡  To avoid scrolling page all the time, I temporarily removed Hero during the recording

https://user-images.githubusercontent.com/17420811/157431092-ce41d6bc-f1d9-4843-819a-9f80c7a024d5.mp4

### Detailed test instructions:

1. Go to step 2 of the free listings edit page.
2. Make changes to the shipping rates and free shipping threshold.
3. Refresh the current webpage or use other ways to navigate away. There should be a prompt popped-up to ask if you sure want to leave with unsaved changes.
   image
4. Revert the changes from step 2.
5. Refresh the current webpage or use other ways to navigate away. It should be able to refresh/leave page directly without prompting.
6. Repeat steps 2-5 with different changes.

PR #1267, opened 2022-02-22
Shipping rates UI changes: Make Setup MC support country-level free shipping and minimum order amount

ecgan:
### Changes proposed in this Pull Request:

- Part of a PR refactor from the huge PR https://github.com/woocommerce/google-listings-and-ads/pull/1218.
- Depends on PR:
  - https://github.com/woocommerce/google-listings-and-ads/pull/1266
  - https://github.com/woocommerce/google-listings-and-ads/pull/1262
    - https://github.com/woocommerce/google-listings-and-ads/pull/1202
    - https://github.com/woocommerce/google-listings-and-ads/pull/1249
- Partly addresses the tech debt: https://github.com/woocommerce/google-listings-and-ads/issues/288.

Figma: 2bb94-pb

This PR contains changes to make the Edit Free Listings support country-level free shipping and minimum order amount setup. It uses the `ShippingRateSection` shared component created from #1266.

#### Prior to this PR

In **Setup MC**, the shipping rate code has the following file structure:

image

There are a number of "smart components" where the components have the logics the make API calls. For example, see `AddRateModal`: https://github.com/woocommerce/google-listings-and-ads/blob/a6963bc069d58bb0ebf361ae41f9a1633de0b4da/js/src/setup-mc/setup-stepper/setup-free-listings/shipping-rate/shipping-rate-setup/add-rate-button/add-rate-modal/index.js#L45-L55

Upon clicking "Save" button, it will make API call to upsert shipping rate, and update wp.data store, and the UI would update accordingly.

One reason for the above implementation is that we want to have an "auto-save" feature for the Setup MC flow. If users were to proceed through the Setup MC flow and quit halfway, they can come back to it later and continue from where they left off.

#### In this PR

We make use of the `ShippingRateSection` shared component, so all the interactions within ShippingRateSection is the same as the Edit Free Listings flow. Shipping rates data are retrieved and stored in the parent form state (with the form key `shipping_country_rates`), and are propagated up and down to the components via `value` and `onChange` props.

The auto-save feature is now done in `FormContent` using `useAutoSaveShippingRatesEffect`, just like `useAutoSaveSettingsEffect`.

### Screenshots:

Shipping Rates section, with the new "offer free shipping" card and "minimum order" card. This UI and the code is the same for both Setup MC flow and Edit Free Listings flow.

image

📹  Demo video with my voice annotation:

https://user-images.githubusercontent.com/417342/154854800-d3d9a1a7-d93e-4ebf-b807-520e4ea5fb11.mov

[...]

PR #1262, opened 2022-02-22
Shipping rates UI changes: wp.data

ecgan:
### Changes proposed in this Pull Request:

- Part of a PR refactor from the huge PR https://github.com/woocommerce/google-listings-and-ads/pull/1218.
- Depends on changes from the following PRs: 
  - https://github.com/woocommerce/google-listings-and-ads/pull/1202
  - https://github.com/woocommerce/google-listings-and-ads/pull/1249

This PR contains wp.data changes to make things work with the new shipping rates API changes. The `upsertShippingRates` and `deleteShippingRates` are removed and replaced with `saveShippingRates` that is able to handle both within the function.

In subsequent PRs, shipping rates in the UI will be stored in the form state, and `saveShippingRates` will be called with the complete shipping rates array from the form state.

### Detailed test instructions:

This is wp.data API changes, there isn't a UI way to test it in this PR branch.

You can test this PR using browser console like this:

```js
wp.data.resolveSelect( 'wc/gla' ).getShippingRates().then( ( rates ) => {
  // Mutate `rates` here if want to test rate updating or deletion
  wp.data.dispatch( 'wc/gla' ).saveShippingRates( rates );
} )
```

Alternatively, you can check out PR #1218 which contains the same wp.data changes, and test the Setup MC or Edit Free Listings flows, and you would be able to observe the API calls from there.

### Changelog entry

>

PR #1313, opened 2022-03-10
Shipping rates UI changes: Change `saveShippingRates` into custom hook

### LINKED ISSUE #1287: Shipping rates UI changes: Change `saveShippingRates` into custom hook
## User story

This is a follow-up to https://github.com/woocommerce/google-listings-and-ads/pull/1262#discussion_r817832807:

> 1. In wp.data, we don't need `saveShippingRates`, we want `upsertShippingRates` and `deleteShippingRates` actions.
> 2. We come up with a custom hook like the following, which consumes the above `upsertShippingRates` and `deleteShippingRates` actions:
> 
> ```js
> // an example:
> const useSaveShippingRate = () => {
> 	const saveShippingRate = useCallback( () => {
> 		// do the deleteShippingRates and upsertShippingRates here.
> 	}, [])
> 
> 	return saveShippingRate
> }
> ```
> 
> 3. Change the code in Edit Free Listings and Setup MC to use the above hook.

In addition, we might want to consider doing the following too, based on https://github.com/woocommerce/google-listings-and-ads/pull/1262#discussion_r817858184:

> > Moreover, they also cause redundant API calls and database operations on the server-side internally. Especially when many countries are selected, the auto-save feature causes more redundant calls and loads to server.
> 
> We can address this within the `saveShippingRates (newShippingRates)` function (action or custom hook):
> 
> 1. Delete the old shipping rates, like what we already have.
> 2. For the new shipping rates, we loop through them and compare with the shipping rate object found from wp.data store. We should do a deep comparison here, to cater for `shippingRate.options.free_shipping_threshold` etc. If they are not equal (i.e. they have been changed), we add the new shipping rate to an array pending to be upserted.
> 3. With the array of shipping rates to be upserted, we call upsertShippingRates, and get back the result.
> 4. Merge the result with the unchanged shipping rates and put them into wp.data store.
> 
> With this approach, we will only call upsertShippingRate for those shipping rates that are newly created or have been changed, and the consumer would still call the same API `saveShippingRates (newShippingRates)`, which is nice. I didn't implement this because I thought it was somewhat of an early optimization, however we can certainly do that later.

Another suggestion is mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1262#discussion_r818004417:

> I was just looking the code again and I think this function should be just `deleteShippingRates( deleteIds = [])` and the logic to get the deleteIds could go in something like `const deleteIds = getDeleteIds( newRates, oldRates)`.
> Like this deleteShippingRates is agnostic about how to get the deleteIds and it could be reused easily and the testing should be only on `getDeleteIds` that is the part doing the logic part.

Related to another suggestions in https://github.com/woocommerce/google-listings-and-ads/pull/1267#discussion_r820908402 and https://github.com/woocommerce/google-listings-and-ads/pull/1267#discussion_r820912825:

> I can look into removing the try/catch block for `saveShippingRates` (refactor it with `deleteShippingRates` and `upsertShippingRates`) in actions.js in #1287, so that when there is an error in those shipping rates operations, the error would bubble up and can be handled by the components. The error will be caught by the catch block we are discussing here.

cc @eason9487 and @jorgemd24 (code reviewers in the above PR).

ecgan:

Closes #1287.

In this PR, we changed the `saveShippingRates` wp.data actions into custom hook. This addresses the points noted in issue #1287:

1. We add `upsertShippingRates` and `deleteShippingRates` wp.data actions, and use `saveShippingRates` custom hook that wraps around them.
2. `upsertShippingRates` will only be called for changed or newly added shipping rates.
3. We have a `getDeleteIds` function to encapsulate the logic of how to get the IDs to be deleted.
4. No try/catch block in `upsertShippingRates` and `deleteShippingRates` wp.data actions and in `saveShippingRates` custom hook, so that the error can propagate up and be caught by the components to display the necessary error message based on the context.

### Screenshots:

📹 Demo video with my voice:

https://user-images.githubusercontent.com/417342/157846888-e14921e8-2ec3-4ccb-84da-3af1059e9d06.mov

You can test this in Setup MC or Edit Free Listings.

1. Open browser dev tools to observe network requests. It is recommended for you to have the Redux Dev Tools to observe wp.data store changes.
2. Go to shipping rate section. Add, edit, or delete shipping rates.
    - Note: to delete shipping rate, you may have to modify the target audience countries.
3. Click on "Continue" / "Save changes" button. In the network panel, there should be a delete shipping rate request for deleted shipping rates, and upsert shipping rate request for added or edited shipping rates. When the requests are done, it should be reflected in the wp.data store.
4. Without reloading the page, repeat step 2 and 3 above for a few more times. This is to test the comparison between the shipping rates from form data and wp.data store.
5. Reload the page. The shipping rates should be the same as before the reload.


js/src/edit-free-campaign/index.js
```diff
@@ -148,10 +149,11 @@ export default function EditFreeCampaign() {
 	// Check what've changed to show prompt, and send requests only to save changed things.
 	const didAudienceChanged = ! isEqual( targetAudience, savedTargetAudience );
 	const didSettingsChanged = ! isEqual( settings, savedSettings );
-	const didRatesChanged = hasUnsavedShippingRates(
+	const unsavedShippingRates = getUnsavedShippingRates(
 		shippingRates,
 		savedShippingRates
 	);
+	const didRatesChanged = unsavedShippingRates.length > 0;
```
eason9487: 🚧 
`differenceWith` won't tell whether any items in the first array don't exist in the second array. So It's unable to get a correct result of `didRatesChanged` when shipping rates are deleted.

```js
differenceWith( [ 1, 2, 3 ], [ 1, 2, 3 ] );
// > []

differenceWith( [ 1 ], [ 1, 2, 3 ] );
// > []
```

Kapture 2022-03-14 at 11 32 22

js/src/setup-mc/setup-stepper/setup-free-listings/index.js
```diff
@@ -41,7 +42,8 @@ const SetupFreeListings = ( props ) => {
 	const {
 		data: finalCountryCodesData,
 	} = useTargetAudienceFinalCountryCodes();
-	const { saveSettings, saveShippingRates } = useAppDispatch();
+	const { saveSettings } = useAppDispatch();
+	const { saveShippingRates } = useSaveShippingRates();
```
eason9487, line 46: 💅 

> A further thought is, if that API error would be thrown to this component, then perhaps we could move `setSaving( false )` into `catch` block. The continue button could be set back to enabled only if errors occurred, otherwise, keeping it showing the loading state before bringing merchant to the next step should be fine.

The above is quoted from [my comment in PR 1267](https://github.com/woocommerce/google-listings-and-ads/pull/1267#discussion_r821328150). Sorry for bothering this. As result of the error could be thrown out now, I'd like to mention it again in case of it was just missed.

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
```diff
@@ -1,12 +1,12 @@
...
 import useIsEqualRefValue from '.~/hooks/useIsEqualRefValue';
 import useDebouncedCallbackEffect from '.~/hooks/useDebouncedCallbackEffect';
+import useSaveShippingRates from '.~/hooks/useSaveShippingRates';
 
 const useAutoSaveShippingRatesEffect = ( shippingRates ) => {
-	const { saveShippingRates } = useAppDispatch();
+	const { saveShippingRates } = useSaveShippingRates();
 	const shippingRatesRefValue = useIsEqualRefValue( shippingRates );
 
 	useDebouncedCallbackEffect( shippingRatesRefValue, saveShippingRates );
```
eason9487: 🚧 
The errors could be thrown out from `upsertShippingRates` and `deleteShippingRates` to `saveShippingRates`, but errors are unable to be handled after `saveShippingRates` is wrapped debounce. It may cause an uncaught error when using `useAutoSaveShippingRatesEffect`.

Kapture 2022-03-14 at 12 31 12

The root cause can be simplified as the `bar` function as follow

```js
// Catchable error
async function foo() {
	throw new Error();
}

// Uncatchable error - Simulate the debounce callback
async function bar() {
	setTimeout( () => {
		throw new Error();
	} );
}

( async function () {
	try {
		await foo();
	} catch ( e ) {
		console.log( e );
	}

	try {
		await bar();
	} catch ( e ) {
		console.log( e );
	}
} )();
```

js/src/edit-free-campaign/getUnsavedShippingRates.js
```diff
@@ -18,15 +18,14 @@ import { isEqual, differenceWith, at } from 'lodash';
...
+ * @param  {Array<ShippingRate>} newShippingRates The new shipping rates to be compared. This can be coming from Form values.
+ * @param  {Array<ShippingRate>} oldShippingRates The old shipping rates to be compared against. This should be previously received from API.
+ * @return {Array<ShippingRate>} Unsaved shipping rates from newShippingRates.
  */
-export default function hasUnsavedShippingRates( rates, savedRates ) {
-	if ( rates.length !== savedRates.length ) {
-		return true;
-	}
-
+export default function getUnsavedShippingRates(
```
eason9487: ❓ 
The original `hasUnsavedShippingRates` also contains the logic to determine whether any deletion has occurred, but the `getUnsavedShippingRates` function only returns the changed shipping rates but does not include those deleted ones. Would it be better to use a name like `getDiffShippingRates` or something like that?

Further, instead of changing the original `hasUnsavedShippingRates` interface, I'd suggest extracting the following parts into a new function. When writing tests, `if ( rates.length ! == savedRates.length )` and `diffRates.length > 0` could be included in test cases together, rather than externalizing these logic to *edit-free-campaign/index.js*.

https://github.com/woocommerce/google-listings-and-ads/blob/34cacd305135b79ba0fda599a8f20db170404eaf/js/src/edit-free-campaign/hasUnsavedShippingRates.js#L30-L40

js/src/data/reducer.js
```diff
@@ -124,6 +124,35 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
 			return setIn( state, 'mc.shipping.rates', action.shippingRates );
 		}
 
+		case TYPES.UPSERT_SHIPPING_RATES: {
```
eason9487, line 127: 💅 I think maybe it would be better to add back the related reducer tests that were deleted in #1262

js/src/data/actions.js
```diff
@@ -75,30 +73,7 @@ export function* fetchShippingRates() {
...
-			method: 'DELETE',
-			data: {
-				ids: deleteIds,
-			},
-		} );
-	}
-}
-
-function* upsertShippingRates( shippingRates ) {
+export function* upsertShippingRates( shippingRates ) {
```
jorgemd24, line 82: 💅

It would be nice to add a JSDoc describing the argument `shippingRates` that is receving this function.

js/src/data/actions.js
```diff
@@ -107,57 +82,32 @@ function* upsertShippingRates( shippingRates ) {
...
+	};
+}
 
-		/**
-		 * There are new shipping rates, we call API to upsert them.
-		 */
-		const upsertedShippingRates = yield upsertShippingRates(
-			newShippingRates
-		);
+export function* deleteShippingRates( ids ) {
```
jorgemd24, line 110: 💅

It would be nice to indicate that ids is an array of ids.

jorgemd24 (commented):
Thanks, @ecgan 👍  I have left a couple of comments about some functions. In general is working well, except for the things that @eason9487 mentioned before.

js/src/edit-free-campaign/getUnsavedShippingRates.js
ecgan: > The original `hasUnsavedShippingRates` also contains the logic to determine whether _any deletion has occurred_, but the `getUnsavedShippingRates` function only returns the changed shipping rates but _does not include those deleted ones_.

🤦 My bad. When I saw the `if ( rates.length ! == savedRates.length )`, somehow I thought it is used to just short-circuit and quickly return a result without having to run `differenceWith` computation. I'll think of something to make things work.

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
ecgan: Thanks @eason9487. For this auto-save feature, I think we do not need to present a UI prompt to notify users on error, so we can just catch the error and do nothing. I'll work on it.

js/src/edit-free-campaign/index.js
ecgan: Thanks. That was my mistake from changing the `hasUnsavedShippingRates` to `getUnsavedShippingRates`, sorry about that. I have reverted the changes there in [`549f26f` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/549f26f133191617876642d032fe6afb96d017ca). 🙏

js/src/setup-mc/setup-stepper/setup-free-listings/index.js
ecgan: Actually I think that is fine. In case for some reasons the `onContinue` doesn't go to the next page, having the button displayed there with loading state might look like a bug.

Anyway, I'm fine with moving it into the catch block too (because `onContinue` does go to the next page), so I have moved it there in [`43513b9` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/43513b9dd139c8aecd30bffc01f0a6caf2c10bad).

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
ecgan: This is fixed in [`bd89578` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/bd8957890f9c22e2db4452f660bdef60dbafeed1).

js/src/edit-free-campaign/getUnsavedShippingRates.js
ecgan: I have come up with a new function called `getDifferentShippingRates` in [`5f8649b` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/5f8649b99da1ca81795ff1205857e02f8726e6b8). This should now be fixed.

js/src/data/reducer.js
ecgan: Oh yes! Done in [`c5b9818` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/c5b9818e4a3f7ee919c20ff15e60efac7a8beab0).

js/src/data/actions.js
ecgan: Done in [`367c7c9` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/367c7c92a512b98a6211e8ea529df4e9830bfc48).

js/src/data/actions.js
ecgan: Done in [`367c7c9` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/367c7c92a512b98a6211e8ea529df4e9830bfc48).

js/src/edit-free-campaign/hasUnsavedShippingRates.test.js
```diff
@@ -0,0 +1,258 @@
+/**
+ * Internal dependencies
+ */
+import hasUnsavedShippingRates from './hasUnsavedShippingRates';
+
+describe( 'hasUnsavedShippingRates', () => {
```
eason9487, line 14: 💅 

I was going to make up the test for #1307 before, and after comparing some test cases that are missed here, one of the missing cases are partially covered in getDifferentShippingRates.test.js though.

I attach what I wrote as a reference, and the missing parts are marked with a code comment `// Could be added`.

```js
/**
 * External dependencies
 */
import { cloneDeep, set } from 'lodash';

/**
 * Internal dependencies
 */
import { SHIPPING_RATE_METHOD } from '.~/constants';
import hasUnsavedShippingRates from './hasUnsavedShippingRates';

const method = SHIPPING_RATE_METHOD.FLAT_RATE;

describe( 'hasUnsavedShippingRates', () => {
	let savedRates;
	let rates;

	beforeEach( () => {
		savedRates = [
			{
				id: '1',
				country: 'US',
				rate: 12.34,
				currency: 'USD',
				options: { free_shipping_threshold: 1000 },
				method,
			},
			{
				id: '2',
				country: 'JP',
				rate: 4900,
				currency: 'JPY',
				options: { free_shipping_threshold: 2000 },
				method,
			},
		];
		rates = cloneDeep( savedRates );
	} );

	it( 'when both shipping rates are the same, should return false', () => {
		expect( hasUnsavedShippingRates( rates, savedRates ) ).toBe( false );
	} );

	// Could be added
	it( 'when the `id` of shipping rates are different, should still return false', () => {
		delete rates[ 0 ].id;
		delete savedRates[ 1 ].id;

		expect( hasUnsavedShippingRates( rates, savedRates ) ).toBe( false );
	} );

	// Could be added
	it( 'when the `options.free_shipping_threshold` property of one of the shipping rates does not exist and the other is `undefined`, should still return false', () => {
		rates[ 0 ].options.free_shipping_threshold = undefined;
		savedRates[ 0 ].options = {};

		expect( hasUnsavedShippingRates( rates, savedRates ) ).toBe( false );
	} );

	// Could be added
	it( 'the comparison is not sequential, two same sets of shipping rates should return false', () => {
		rates.reverse();

		expect( hasUnsavedShippingRates( rates, savedRates ) ).toBe( false );
	} );

	it( 'when there are deleted shipping rates, should return true', () => {
		rates.pop();

		expect( hasUnsavedShippingRates( rates, savedRates ) ).toBe( true );
	} );

	it( 'when there are new shipping rates, should return true', () => {
		rates.push( {} );

		expect( hasUnsavedShippingRates( rates, savedRates ) ).toBe( true );
	} );

	// Could be added.
	// (Test each property in an atomical way.)
	it.each( [
		[ 'currency', 'AUD' ],
		[ 'rate', 50.88 ],
		[ 'country', 'AU' ],
		[ 'options', {} ],
		[ 'options.free_shipping_threshold', undefined ],
		[ 'options.free_shipping_threshold', 5000 ],
		[ 'method', 'bumpy_rate' ],
	] )(
		'when the property `%s` is edited, for example to %s, should return false',
		( path, value ) => {
			set( rates[ 0 ], path, value );

			expect( hasUnsavedShippingRates( rates, savedRates ) ).toBe( true );
		}
	);
} );
```

js/src/edit-free-campaign/hasUnsavedShippingRates.js
```diff
@@ -11,6 +11,14 @@ import { isEqual, differenceWith, at } from 'lodash';
  * This function compares the key fields of two shipping rate arrays
  * and returns whether there are any unsaved shipping changes.
  *
+ * There are unsaved shipping changes when there are:
+ *
+ * - deleted shipping rates.
+ * - newly added shipping rates.
+ * - edited shiping rates.
```
eason9487, line 18: Thanks for adding more detailed descriptions! 💯

eason9487 (commented):
🚧  or 📜 
Found another issue that already existed before this project. When there is a failed API call, the notification at the bottom still shows "Your changes to your Free Listings have been saved and will be synced to your Google Merchant Center account."

https://github.com/woocommerce/google-listings-and-ads/blob/154565b1739ba5a9982caaab60447a05503d00ba/js/src/edit-free-campaign/index.js#L188-L208

https://user-images.githubusercontent.com/17420811/158565390-9cf4b921-0adf-4eda-b37e-a6ec65f972a2.mp4


Maybe this could be fixed in the meantime, or I could open a follow-up issue for it. (I don't really remember, maybe there is already a related issue).

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
eason9487: > This is fixed in [`bd89578` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/bd8957890f9c22e2db4452f660bdef60dbafeed1).

Sorry, I'm not sure if I was missing something, but looks like this issue may not have been fixed?

https://user-images.githubusercontent.com/17420811/158566092-9d49e7e0-5cc3-4293-9eb6-b0572bb36118.mp4

ecgan: D'oh, it wasn't you who is missing something, it was me missing async await. 🤦‍♂️ Fixed in [`a69a8fd` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/a69a8fd30c011e82fa006e2ba56658bae996a93f).

Here's a demo video of the fix:

https://user-images.githubusercontent.com/417342/158599930-9335521e-26fe-47dc-9328-4bc23e6f49bc.mp4

ecgan:
@eason9487 ,

> Found another issue that already existed before this project. When there is a failed API call, the notification at the bottom still shows "Your changes to your Free Listings have been saved and will be synced to your Google Merchant Center account."
> ...
> Maybe this could be fixed in the meantime, or I could open a follow-up issue for it. (I don't really remember, maybe there is already a related issue).

Yeah it is an existing issue, I noticed it when I was testing it too, thought of fixing it later (too many things for this shipping rate project already). I have created issue https://github.com/woocommerce/google-listings-and-ads/issues/1331 for this, with my thoughts on how to fix it in the issue. 🙏

js/src/edit-free-campaign/hasUnsavedShippingRates.test.js
ecgan: Your tests there look so good, I ended up using all of them and replacing mine in [`6a9e5f8` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/6a9e5f8eac3ef1d652cfabd0d20ca14aa6bf1065). 😆 👍 

(I have my own opinions in testing, but I don't hold them as strong as my other opinions 😆 , and yours are good in here. 😄)

PS: the `(Test each property in an atomical way.)` with `it.each` is nice. 🙂 👍 

Thank you! 🙏

js/src/edit-free-campaign/hasUnsavedShippingRates.test.js
```diff
@@ -0,0 +1,93 @@
...
+	it.each( [
+		[ 'currency', 'AUD' ],
+		[ 'rate', 50.88 ],
+		[ 'country', 'AU' ],
+		[ 'options', {} ],
+		[ 'options.free_shipping_threshold', undefined ],
+		[ 'options.free_shipping_threshold', 5000 ],
+		[ 'method', 'bumpy_rate' ],
+	] )(
+		'when the property `%s` is edited, for example to %s, should return false',
```
eason9487: 🙏  Sorry, I wrote the wrong expected result here. 🤦‍♂️  
```suggestion
		'when the property `%s` is edited, for example to %s, should return true',
```

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
```diff
@@ -1,15 +1,37 @@
...
+				await saveShippingRates( value );
+			} catch ( error ) {}
+		},
+		[ saveShippingRates ]
+	);
+
+	useDebouncedCallbackEffect(
+		shippingRatesRefValue,
+		saveShippingRatesCallback
+	);
```
eason9487: ❔ 
Due to the async call of debounce, it requires wrapping an additional layer of function to catch the error. Considering the problem that the debounce processing was trying to solve already doesn't exist, and the [consistency reason](https://github.com/woocommerce/google-listings-and-ads/pull/1267#discussion_r820441914) I didn't understand, it presents a more unnecessary complexity now.

If it doesn't offer any benefit to maintain consistency with the debounce processing of `useAutoSaveSettingsEffect`, I would still recommend simplifying it.

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
```diff
@@ -1,15 +1,37 @@
...
+	 * and want it to fail silently in the background.
+	 */
+	const saveShippingRatesCallback = useCallback(
+		async ( value ) => {
+			try {
+				await saveShippingRates( value );
+			} catch ( error ) {}
+		},
+		[ saveShippingRates ]
+	);
```
eason9487: 💅 
The callback reference changes don't affect the use of `useDebouncedCallback`. The `useCallback` could be omitted.

eason9487 (approved):
Thanks for the improvement and adding test cases! LGTM. Left some non-blocking comments for minor things.

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
ecgan: Ahh right, I forgot that is already handled by [`useDebouncedCallback`](https://github.com/xnimorz/use-debounce#debounced-callbacks). Removed in [`e9f512b` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/e9f512bcb99d7baa61ab1760504b8919d4059229). 👍

js/src/edit-free-campaign/hasUnsavedShippingRates.test.js
ecgan: I was too excited and didn't notice that either 😆 Fixed in [`6624e83` (#1313)](https://github.com/woocommerce/google-listings-and-ads/pull/1313/commits/6624e836019c23039a9bf49fac54f095323ca6b7).

js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveShippingRatesEffect.js
ecgan: I have created issue https://github.com/woocommerce/google-listings-and-ads/issues/1335 to simplify things even further, by replacing the auto effects with Form `onChange`. We can address the debounce processing there, and by having things together inside `onChange` handler (instead of separate auto effects), I think it should make things easier to understand too. 🙂

PR #1314, opened 2022-03-11
Tree Select Component - Add Partially checked feature

puntope:

This PR implements a new feature in Tree Select Component for having a Partially Checked Group. 

A partially checked group is a group (node with children) in which any of their children is selected. 

If a node is not checked but is partially checked. Then it shows a different UI like this.

Screenshot 2022-03-11 at 13 18 37

### Screenshots:

https://user-images.githubusercontent.com/5908855/157838414-0c67d971-ac35-4841-8716-d9985c359072.mov

1. Check out this branch and go to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fcomponent-test`
2. In the selector you can see how Spain is selected by default
3. Open the selector and You will see how All Countries and Europe groups are partially selected.
4. If you check Europe, it turns to checked and selects all the options
5. If you deselect an item on that group. For example Italy. It turns to partially Checked
6. If you go to North America, United States and Select NY it will partially check United States, North America and All Countries.

### Additional details:

- We use Array.protoType.some() in order to check if any of the children are checked or partially checked.
- In case a node is partially checked we apply a className `is-partially-checked`
- When a node is not checked and has that class, we apply some styles on it

### DIFF
--- a/js/src/components/tree-select-control/options.js
+++ b/js/src/components/tree-select-control/options.js
@@ -40,19 +40,18 @@ const Options = ( {
 	};
 
 	/**
-	 * Verifies if an option is partially checked.
-	 * An option is partially checked if any of their children is checked or partially checked
+	 * Verifies if an option has some children checked.
 	 *
-	 * @param {Option} parent the Option to verify if its partially checked
-	 * @return {boolean} True if it's partially Checked, false otherwise
+	 * @param {Option} parent the Option to verify
+	 * @return {boolean} True if any at least one of the children is checked, false otherwsie
 	 */
-	const isPartiallyChecked = ( parent ) => {
+	const hasSomeChildrenChecked = ( parent ) => {
 		if ( ! parent.children?.length ) {
 			return false;
 		}
 
 		return parent.children.some(
-			( child ) => isChecked( child ) || isPartiallyChecked( child )
+			( child ) => isChecked( child ) || hasSomeChildrenChecked( child )
 		);
 	};
 
@@ -79,10 +78,9 @@ const Options = ( {
 
 	return options.map( ( option ) => {
 		const isRoot = option.value === '';
-
 		const isExpanded = isRoot || nodesExpanded.includes( option.value );
-
 		const hasChildren = !! option.children?.length;
+		const optionIsChecked = isChecked( option );
 
 		return (
 			<div
@@ -114,12 +112,13 @@ const Options = ( {
 					<CheckboxControl
 						className={ classnames(
 							'woocommerce-tree-select-control__option',
-							isPartiallyChecked( option ) &&
+							! optionIsChecked &&
+								hasSomeChildrenChecked( option ) &&
 								'is-partially-checked'
 						) }
 						value={ option.value }
 						label={ option.label }
-						checked={ isChecked( option ) }
+						checked={ optionIsChecked }
 						onChange={ ( checked ) => {
 							onChange( checked, option );
 						} }
--- a/js/src/components/tree-select-control/options.js
+++ b/js/src/components/tree-select-control/options.js
@@ -27,6 +27,36 @@ const Options = ( {
 	onNodesExpandedChange = () => {},
 } ) => {
 	/**
+	 * Verifies if an option is checked.
+	 * An option is checked if their value is selected or all of their children are selected
+	 *
+	 * @param {Option} option The option to verify if is checked
+	 * @return {boolean} True if checked, false otherwise
+	 */
+	const isChecked = ( option ) => {
+		return (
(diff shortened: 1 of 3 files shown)


js/src/components/tree-select-control/options.js
```diff
@@ -86,13 +112,14 @@ const Options = ( {
 					) }
 
 					<CheckboxControl
-						className={ 'woocommerce-tree-select-control__option' }
+						className={ classnames(
+							'woocommerce-tree-select-control__option',
+							isPartiallyChecked( option ) &&
```
jorgemd24: ❓

It seems that the class `is-partially-checked` is still showing once the children are 'full' checked. Imagine that someone wants to give a special style using that class, the style will still show even when all the children are selected.

js/src/components/tree-select-control/options.js
```diff
@@ -26,6 +26,36 @@ const Options = ( {
...
+	 *
+	 * @param {Option} parent the Option to verify if its partially checked
+	 * @return {boolean} True if it's partially Checked, false otherwise
+	 */
+	const isPartiallyChecked = ( parent ) => {
+		if ( ! parent.children?.length ) {
+			return false;
+		}
+
+		return parent.children.some(
```
jorgemd24, line 53: ❓ 

I think the use of `Array.protoType.some()` in here is not that convenient, imagine that all children are selected, this function will still return true even when it is not partially checked.

js/src/components/tree-select-control/options.test.js
```diff
@@ -51,4 +51,36 @@ describe( 'TreeSelectControl - Options Component', () => {
...
 		} );
 	} );
+
+	it( 'Partially selects groups', () => {
+		const { queryByText } = render(
+			<Options options={ options } value={ [ 'ES' ] } />
+		);
+
+		const partiallyCheckedOption = queryByText( 'Europe' );
+		const unCheckedOption = queryByText( 'North America' );
```
jorgemd24, line 61: 💅

It would be nice to check that the rest options are unchecked (North America, Asia, etc) and the only one that has to be checked is 'Europe'.  Maybe it could be done using the `getAllBy` query.

jorgemd24 (commented):
I have tested and it is working as expected however I left some comments regards the `isPartiallyChecked` function and about the test.

Something that I found and is not related to this PR, is that when I click on the selector or the `label` for the selector, it collapses and then it expands I think, the user would expect some toggling. If it is showing it should hide and if it is hidden it should show. 


https://user-images.githubusercontent.com/2488994/157893437-92b28f4f-309a-49ba-981b-1fa6861377d3.mp4

Thanks!  👍

js/src/components/tree-select-control/options.js
puntope: We are handling that in CSS (checking it with `:not(:checked)`) however I added the check in JS in this commit https://github.com/woocommerce/google-listings-and-ads/pull/1314/commits/e7e9081c9335c419846d61b23c9312ea28d5c769

js/src/components/tree-select-control/options.js
puntope: Same as the previous comment, my way of thinking was that a Full Checked Option is also partially checked. 

We just prioritize the Full Checked over the Partially Checked. This is mainly to avoid iterating to every node again that could be expensive in there are tons of nodes.

Since we applied this commit https://github.com/woocommerce/google-listings-and-ads/commit/e7e9081c9335c419846d61b23c9312ea28d5c769 with an extra validation to see if the option is checked, I think we don't need a full refactor of the function for now.

js/src/components/tree-select-control/options.test.js
puntope: This I'll handle in a future PR at the end of the Project polishing tests. 

Thanks 👍

puntope:
> Something that I found and is not related to this PR, is that when I click on the selector or the `label` for the selector, it collapses and then it expands I think, the user would expect some toggling. If it is showing it should hide and if it is hidden it should show.

Yes, I'm aware of that. It's related to `useFocusOutside` experimental component. I need to go deeper into it at some point (in other PR :p ) 

Let me know if now this one is ready to go :)

js/src/components/tree-select-control/options.js
jorgemd24: Yeah the extra validation helped, thanks!

jorgemd24 (approved):
Thanks, @puntope for the changes. It fixed the issue with the class `is-partially-checked`.
 
This PR is ready to go as it is working as expected and the issue with the class has been resolved ✅

PR #1281, opened 2022-03-03
Tree Select Components - Expand/collapse 

puntope:
### Changes proposed in this Pull Request:

Adding functionality for expanding and collapsing groups of nodes in the tree. 

Thread info: pcTzPl-fg-p2

### Screenshots:

https://user-images.githubusercontent.com/5908855/156535643-e6331260-db12-4ee6-a000-2303a86fba58.mov

### Detailed test instructions:

1. Checkout the branch
2. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fcomponent-test` 
3. You can see component TreeSlectControl with the label `Select Country` and the placeholder `Select`
4. Clicking on "Select" expands the TreeSlector and shows text cursor instead of the placeholder  
5. Tree Shows  the groups collapsed 
6. Clicking on one group expands the group but not the subgroups
7. Clicking again close the group. 

### Additional details:

- Implemented expand/collapse logic
- Updated Aria Roles to use Tree instead of Listbox
- Instead of `list.js` file with `List` and `Options` component inside it. We renamed it to `Options` and located the wrapper directly in the main component file. 
- Updated CSS [according to this comment](https://github.com/woocommerce/google-listings-and-ads/pull/1243#issuecomment-1054538974). Please notice in Figma the boxes are "aligned" as in the previous version. Maybe good to discuss this with the Design team. 
- Refactor Tests dividing them into multiple files  
- It's pending an extra PR in the future with alternative ways to handle option changes, search, expand/collapse etc...

PR #1315, opened 2022-03-11
Polishing Expand Collapse

puntope:

Follow up for -> https://github.com/woocommerce/google-listings-and-ads/pull/1281

This PR includes bug/requests in previous PRs.

- Optional Key support 
- Set ROOT_VALUE as {} instead of '' allowing to have values as ''
- Fix tests in Collapse/Expand
- Expand node on selection ([more info](pdkevu-52-p2))

1. Checkout the branch
3. Clicking on a continent expand the group
4. You can have empty values in the Options

### DIFF
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -17,6 +17,7 @@ import useIsEqualRefValue from '.~/hooks/useIsEqualRefValue';
 import Control from './control';
 import Options from './options';
 import './index.scss';
+import { ROOT_VALUE } from '.~/components/tree-select-control/constants';
 
 /**
  * The Option type Object. This is how we send the options to the selector.
@@ -24,7 +25,8 @@ import './index.scss';
  * @typedef {Object} Option
  * @property {string} value The value for the option
  * @property {string} label The label for the option
- * @property {Option[]} [children] The children Option objects.
+ * @property {Option[]} [children] The children Option objects
+ * @property {string} [key] Optional unique key for the Option. It will fallback to the value property if not defined
  *
  * Example of Options data structure:
  *   [
@@ -83,8 +85,14 @@ const TreeSelectControl = ( {
 	const [ nodesExpanded, setNodesExpanded ] = useState( [] );
 
 	const treeOptions = useIsEqualRefValue(
-		selectAllLabel
-			? [ { label: selectAllLabel, value: '', children: options } ]
+		selectAllLabel !== false
+			? [
+					{
+						label: selectAllLabel,
+						value: ROOT_VALUE,
+						children: options,
+					},
+			  ]
 			: options
 	);
 
@@ -168,6 +176,10 @@ const TreeSelectControl = ( {
 	 */
 	const handleParentChange = ( checked, option ) => {
 		const newValue = [ ...value ];
+
+		if ( checked && ! nodesExpanded.includes( option.value ) ) {
+			setNodesExpanded( [ ...nodesExpanded, option.value ] );
+		}
 
 		function loadChildren( parent ) {
 			if ( ! parent.children ) {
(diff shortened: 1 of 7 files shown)


js/src/components/tree-select-control/index.js
```diff
@@ -169,6 +177,10 @@ const TreeSelectControl = ( {
 	const handleParentChange = ( checked, option ) => {
 		const newValue = [ ...value ];
 
+		if ( checked && ! nodesExpanded.includes( option.value ) ) {
+			setNodesExpanded( [ ...nodesExpanded, option.value ] );
+		}
```
eason9487, line 182: ❔  I'm not quite sure it only needs to expand the children of the first layer or children of all layers in the sub-tree. Maybe needs to confirm with the design team.

js/src/components/tree-select-control/tags.js
```diff
@@ -37,7 +37,7 @@ const Tags = ( { tags, disabled, onChange = () => {} } ) => {
 	return (
 		<div className="woocommerce-tree-select-control__tags">
 			{ tags.map( ( item, i ) => {
-				if ( ! item.label || ! item.id ) {
+				if ( ! item.label ) {
```
eason9487, line 40: ❔ 
I don't quite catch what the original logic was intended to deal with, so this comment may be pointless. But if the logic was removed only to allow the option with `value: ''`, it could also consider replacing `! item.id` with `! item.hasOwnProperty( 'id' )`.

js/src/components/tree-select-control/options.test.js
```diff
@@ -18,37 +18,46 @@ const options = [
...
+		option.children.forEach( ( child ) => {
+			const childItem = queryByText( child.label );
+			expect( childItem ).toBeFalsy();
+		} );
+
+		fireEvent.click( optionItem );
+
+		option.children.forEach( ( child ) => {
+			const childItem = queryByText( child.label );
+			expect( childItem ).toBeTruthy();
```
eason9487, line 66: 💅  
Could be separated into another test case to make it more atomic, or maybe add a code comment for indicating this part is testing when the parent is checked, the children should be expanded.

eason9487 (approved):
Testing well and LGTM. Left two ❔  comments mainly because I'm not quite sure if there are any concerns. Approved in advance to avoid blocking this PR if no adjustment is needed.

js/src/components/tree-select-control/index.js
puntope: I will

js/src/components/tree-select-control/tags.js
puntope: This is to show the label with value as '' yes. 

The original logic was because I never thought we wanted to support '' as a value.

puntope:
@eason9487 

Sorry, I got a problem using `{}` as the default root value.

Since in search functionality I'm filtering the options I do a copy at some point of the tree, so the root option changes and the logic we built for checking if an element is root is not working anymore.

So I changed the name for "__WC_TREE_SELECT_COMPONENT_ROOT__" for now. I think it is safe.

Could you confirm is still approved?

js/src/components/tree-select-control/tags.js
eason9487: Got it, thanks!

eason9487 (approved):
> Since in search functionality I'm filtering the options I do a copy at some point of the tree, so the root option changes and the logic we built for checking if an element is root is not working anymore.
> 
> So I changed the name for "**WC_TREE_SELECT_COMPONENT_ROOT**" for now. I think it is safe.
> 
> Could you confirm is still approved?

Uncertain about the context of the copy, but the collision rate of this string is extremely small, so I think it's fine to avoid getting stuck in subsequent development.

PR #1266, opened 2022-02-22
Shipping rates UI changes: Make Edit Free Listings support country-level free shipping and minimum order amount

ecgan:
### Changes proposed in this Pull Request:

- Part of a PR refactor from the huge PR https://github.com/woocommerce/google-listings-and-ads/pull/1218.
- Depends on PR:
  - https://github.com/woocommerce/google-listings-and-ads/pull/1262
    - https://github.com/woocommerce/google-listings-and-ads/pull/1202
    - https://github.com/woocommerce/google-listings-and-ads/pull/1249
- Partly addresses the tech debt: https://github.com/woocommerce/google-listings-and-ads/issues/288.

This PR contains code refactoring on the shipping rate section and make the Edit Free Listings support country-level free shipping and minimum order amount setup.

#### Prior to this PR

In **Edit Free Listings**, the shipping rate code has the following file structure:

image

They are mostly "dumb components" with no API calls and no knowledge of wp.data store. Values are stored in the parent Form component, and there is no auto-save feature like in Setup MC.

#### In this PR

Shipping rate section is now a shared component called `ShippingRateSection`. The following is the files structure:

image

Here's a visualization of the component composition structure: 

image

All the components highlighted above are dumb components; they do not have any API calls in them. Shipping rates data are retrieved and stored in the parent form state (with the form key `shipping_country_rates`), and are propagated up and down to the components via `value` and `onChange` props.

`FlatShippingRatesInputCards` will only show up when users select flat shipping rate option. It contains `EstimatedShippingRatesCard` and `FreeShippingCards`. 

`EstimatedShippingRatesCard` is from [`ShippingCountriesForm`](https://github.com/woocommerce/google-listings-and-ads/blob/a6963bc069d58bb0ebf361ae41f9a1633de0b4da/js/src/components/free-listings/configure-product-listings/shipping-rate/shipping-rate-setup/countries-form.js), refactored and simplified. The values in this card depends on the `shipping_country_rates` form value.

`FreeShippingCards` is a wrapper around `OfferFreeShippingCard` and `MinimumOrderCard`. It will only be displayed when there is a shipping rate bigger than 0. It contains `useState` with _initial value_ depending on the `shipping_country_rates` form value; subsequent value depends on users selection of "Yes" / "No". When "Yes" is selected, it will render `MinimumOrderCard`.

`MinimumOrderCard` is a new component. It depends on the `shipping_country_rates` form value.

### Screenshots:

[...]

PR #1317, opened 2022-03-11
Fix navigate away prompt showing up when we change `offer_free_shipping` back and forth

### LINKED ISSUE #1318: Changing `offer_free_shipping` value back and forth causes navigate away prompt to show up
### Describe the bug:

In Edit Free Listings, when we change the offer free shipping value back and forth, the navigate away prompt would show up. This is because the `offer_free_shipping` value is being grouped as settings, and settings is considered changed.

📹 Demo video of the issue, with my voice:

https://user-images.githubusercontent.com/417342/157856288-1edcfc8e-4b73-4319-b74b-e8b916a46d45.mp4

### Steps to reproduce:

1. Go to Edit Free Listings. 
2. Change the offer free shipping value, and then change it back to the original value. 
    - Note: if you change from "yes" to "no", your minimum order amount would have been cleared, and you would have to key in the same minimum order amount again when you change back to "yes".
3. Navigate away by reloading the page. The navigate away prompt would show up even when all the values are the same and there are no visible changes to the users.

### Expected behavior:

There should be no prompt, because all the values are the same and there are no visible changes to the users.

### Actual behavior:

The navigate away prompt would show up.

ecgan:

Closes #1318.

In Edit Free Listings, when we change the offer free shipping value back and forth, the navigate away prompt would show up. This is because the `offer_free_shipping` value is being grouped as settings, and settings is considered changed.

In this PR, we use a new function `getSettings` here to extract the settings data from the form values object, leaving `offer_free_shipping` out, so changing the offer free shipping value would not be considered as settings changed.

### Screenshots:

📹 Demo video with my voice:

https://user-images.githubusercontent.com/417342/157855292-b31c5f46-dbc8-401a-ad15-6a77ea9883e8.mov

1. Go to Edit Free Listings. 
2. Change the offer free shipping value, and then change it back to the original value. 
    - Note: if you change from "yes" to "no", your minimum order amount would have been cleared, and you would have to key in the same minimum order amount again when you change back to "yes".
4. Navigate away by reloading the page. Since all the values are the same and there are no changes, there should be no prompt for you, and you should be able to navigate away.

### Changelog entry

>

### DIFF
--- a/js/src/edit-free-campaign/setup-free-listings/index.js
+++ b/js/src/edit-free-campaign/setup-free-listings/index.js
@@ -19,6 +19,19 @@ import FormContent from './form-content';
  * @typedef {import('.~/data/actions').CountryCode} CountryCode
  */
 
+const getSettings = ( values ) => {
+	return {
+		shipping_rate: values.shipping_rate,
+		shipping_time: values.shipping_time,
+		tax_rate: values.tax_rate,
+		website_live: values.website_live,
+		checkout_process_secure: values.checkout_process_secure,
+		payment_methods_visible: values.payment_methods_visible,
+		refund_tos_visible: values.refund_tos_visible,
+		contact_info_visible: values.contact_info_visible,
+	};
+};
+
 /**
  * Setup step to configure free listings.
  *
@@ -78,7 +91,6 @@ const SetupFreeListings = ( {
 		const {
 			shipping_country_rates: newShippingRates,
 			shipping_country_times: newShippingTimes,
-			...newSettings
 		} = newVals;
 
 		switch ( change.name ) {
@@ -89,7 +101,7 @@ const SetupFreeListings = ( {
 				onShippingTimesChange( newShippingTimes );
 				break;
 			default:
-				onSettingsChange( change, newSettings );
+				onSettingsChange( change, getSettings( newVals ) );
 		}
 	};
 
--- a/js/src/edit-free-campaign/setup-free-listings/index.js
+++ b/js/src/edit-free-campaign/setup-free-listings/index.js
@@ -20,6 +20,17 @@ import FormContent from './form-content';
  * @typedef {import('.~/data/actions').CountryCode} CountryCode
  */
 
+const settingsFieldNames = [
+	'shipping_rate',
+	'shipping_time',
+	'tax_rate',
+	'website_live',
+	'checkout_process_secure',
+	'payment_methods_visible',
+	'refund_tos_visible',
+	'contact_info_visible',
+];
+
 /**
  * Get settings object from Form values.
  *
@@ -34,16 +45,7 @@ import FormContent from './form-content';
  * @return {Object} Settings object.
  */
 const getSettings = ( values ) => {
-	return pick( values, [
-		'shipping_rate',
-		'shipping_time',
-		'tax_rate',
-		'website_live',
-		'checkout_process_secure',
-		'payment_methods_visible',
-		'refund_tos_visible',
-		'contact_info_visible',
-	] );
+	return pick( values, settingsFieldNames );
 };
 
 /**
@@ -101,15 +103,12 @@ const SetupFreeListings = ( {
 	};
 
 	const handleChange = ( change, values ) => {
-		switch ( change.name ) {
-			case 'shipping_country_rates':
-				onShippingRatesChange( values.shipping_country_rates );
-				break;
-			case 'shipping_country_times':
-				onShippingTimesChange( values.shipping_country_times );
-				break;
-			default:
-				onSettingsChange( change, getSettings( values ) );
+		if ( change.name === 'shipping_country_rates' ) {
+			onShippingRatesChange( values.shipping_country_rates );
+		} else if ( change.name === 'shipping_country_times' ) {
+			onShippingTimesChange( values.shipping_country_times );
+		} else if ( settingsFieldNames.includes( change.name ) ) {
+			onSettingsChange( change, getSettings( values ) );
 		}
 	};
 
--- a/js/src/edit-free-campaign/setup-free-listings/index.js
+++ b/js/src/edit-free-campaign/setup-free-listings/index.js
@@ -92,22 +92,16 @@ const SetupFreeListings = ( {
 		setSaving( false );
 	};
 
-	const handleChange = ( change, newVals ) => {
-		// Un-glue form data.
-		const {
-			shipping_country_rates: newShippingRates,
-			shipping_country_times: newShippingTimes,
-		} = newVals;
-
+	const handleChange = ( change, values ) => {
 		switch ( change.name ) {
 			case 'shipping_country_rates':
-				onShippingRatesChange( newShippingRates );
+				onShippingRatesChange( values.shipping_country_rates );
 				break;
 			case 'shipping_country_times':
-				onShippingTimesChange( newShippingTimes );
+				onShippingTimesChange( values.shipping_country_times );
 				break;
 			default:
-				onSettingsChange( change, getSettings( newVals ) );
+				onSettingsChange( change, getSettings( values ) );
 		}
 	};
 
--- a/js/src/edit-free-campaign/setup-free-listings/index.js
+++ b/js/src/edit-free-campaign/setup-free-listings/index.js
@@ -40,9 +40,11 @@ const settingsFieldNames = [
  * Get settings object from Form values.
  *
  * This method is used to pick out form fields that are specific to settings.
- * If we are adding a new settings field, it should be added into `settingsFieldNames`.
+ * If we are adding a new settings field that will be saved via the settings API,
+ * it should be added into `settingsFieldNames`.
  *
- * If a new field is added into the form that is not related to settings (e.g. `offer_free_shipping`),
+ * If a new field is added into the form that is not related to settings (e.g. `offer_free_shipping`)
+ * and will NOT be saved via settings API,
  * we do not need to add the field into `settingsFieldNames`,
  * and things should continue to work as expected (e.g. the navigate away prompt).
  *
--- a/js/src/edit-free-campaign/setup-free-listings/index.js
+++ b/js/src/edit-free-campaign/setup-free-listings/index.js
@@ -3,6 +3,7 @@
  */
 import { Form } from '@woocommerce/components';
 import { useState } from '@wordpress/element';
+import { pick } from 'lodash';
 
 /**
  * Internal dependencies
@@ -26,16 +27,16 @@ import FormContent from './form-content';
  * @return {Object} Settings object.
  */
 const getSettings = ( values ) => {
-	return {
-		shipping_rate: values.shipping_rate,
-		shipping_time: values.shipping_time,
-		tax_rate: values.tax_rate,
-		website_live: values.website_live,
-		checkout_process_secure: values.checkout_process_secure,
-		payment_methods_visible: values.payment_methods_visible,
-		refund_tos_visible: values.refund_tos_visible,
-		contact_info_visible: values.contact_info_visible,
-	};
+	return pick( values, [
+		'shipping_rate',
+		'shipping_time',
+		'tax_rate',
+		'website_live',
+		'checkout_process_secure',
+		'payment_methods_visible',
+		'refund_tos_visible',
+		'contact_info_visible',
+	] );
 };
 
 /**
--- a/js/src/edit-free-campaign/setup-free-listings/index.js
+++ b/js/src/edit-free-campaign/setup-free-listings/index.js
@@ -23,6 +23,13 @@ import FormContent from './form-content';
 /**
  * Get settings object from Form values.
  *
+ * This method is used to intentionally pick out form fields that are specific to settings.
+ * If we are adding a new settings field, it should be added into this function.
+ *
+ * If a new field is added into the form that is not related to settings (i.e. `offer_free_shipping`),
+ * we do not need to add the field into the function,
+ * and things should continue to work as expected (e.g. the navigate away prompt).
+ *
  * @param {Object} values Form values.
  * @return {Object} Settings object.
  */
@@ -111,6 +118,7 @@ const SetupFreeListings = ( {
 			<Hero />
 			<Form
 				initialValues={ {
+					// These are the fields for settings.
 					shipping_rate: settings.shipping_rate,
 					shipping_time: settings.shipping_time,
 					tax_rate: settings.tax_rate,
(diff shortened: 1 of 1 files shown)


js/src/edit-free-campaign/setup-free-listings/index.js
```diff
@@ -19,6 +19,19 @@ import FormContent from './form-content';
...
+		shipping_rate: values.shipping_rate,
+		shipping_time: values.shipping_time,
+		tax_rate: values.tax_rate,
+		website_live: values.website_live,
+		checkout_process_secure: values.checkout_process_secure,
+		payment_methods_visible: values.payment_methods_visible,
+		refund_tos_visible: values.refund_tos_visible,
+		contact_info_visible: values.contact_info_visible,
+	};
+};
```
eason9487: ❓
Sorry, I can't see the purpose of this handling. There is already having the processing of un-glue form data. Looks like it could be addressed by omitting `offer_free_shipping` with that process.

```diff
@@ -70,6 +70,7 @@ const SetupFreeListings = ( {
                const {
                        shipping_country_rates: newShippingRates,
                        shipping_country_times: newShippingTimes,
+                       offer_free_shipping, // eslint-disable-line camelcase, no-unused-vars
                        ...newSettings
                } = newVals;
```

ecgan: Yea, I thought about the approach you mentioned. The problem with it is that if someone were to add another field to the Form, it would cause similar issue like #1318. So I use this approach - picking out specific Settings fields - to ensure issue like #1318 won't happen again.

eason9487: > The problem with it is that if someone were to add another field to the Form, it would cause similar issue like #1318. So I use this approach - picking out specific Settings fields - to ensure issue like #1318 won't happen again.

Unfortunately, this approach might not achieve this purpose. Regardless of the approach, a conceptually similar problem exists and there is no guarantee that it won't happen again.

- If a new field is added that will be saved in the API, it will still be missed in `getSettings`.
- If a new field is added that will not be saved in the API, it will also be missed when un-glue form data.

Since neither approach can prevent it, considering that the un-glue form data is an existing process and fewer fields need to be filtered out than picked out, would it be better to use the un-glue form data to process it incidentally?

ecgan: Note: @eason9487 and I had a related conversation in our internal slack channel, see: p1647313955981019-slack-C01E9LB4X61

---

> * If a new field is added that will be saved in the API, it will still be missed in `getSettings`.

Yes, there will be chances of issues happening. However, because it is more explicit with `getSettings`, it is easier to notice and fix issues related to it, e.g. "I added a new settings field, something doesn't work, I look at getSettings, it is _relatively obvious_ that I forgot to add it in there."

> * If a new field is added that will not be saved in the API, it will also be missed when un-glue form data.

Yes. Issue #1318 is an example of that, and using `getSettings` would help with that.

> considering that the un-glue form data is an existing process and fewer fields need to be filtered out than picked out, would it be better to use the un-glue form data to process it incidentally?

I still think that specifically picking out the wanted fields is better than taking out the unwanted fields and leaving the wanted fields in the rest object, i.e.:

```js
// Approach 1: get the wanted fields via getSettings.
const settings = getSettings( values );

// Approach 2: leave the wanted fields in rest settings object.
const { unwanted1, ...settings } = values;
```

With Approach 1, if someone were to add a new non-settings form field, `settings` will still be the same object shape; with Approach 2, `settings` will contain the new field, which may be accidental and may not be desired, and this is the thing I would like to avoid. 

---

I'm feeling like this is the time where we got to "agree to disagree". 😅

eason9487: Thanks for the explanation.

---

I'd like to clarify my point was in advocating maintainability and readability when discussed in this PR comment and that internal slack thread.

### In terms of maintainability

Both methods have conceptually similar issues. Hope this table could make the point clearer:
| Method<br>╲<br>New field to be added | Pick out needed<br>(`getSettings`)                                                                                                              | Pick out unneeded<br>(un-glue data in `handleChange`)                                                                                              |
|--------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
| will be saved via API                | Must be added as a field to be removed, or the unsaved checking will be incorrect and the settings to be called to API will miss the new field. | No adjustment required                                                                                                                             |
| won't be saved via API               | No adjustment required                                                                                                                          | Must be added as a field to be removed, or the unsaved checking will be incorrect and the settings to be called to API will have an invalid field. |

No matter which approaches, the new field is always automatically included on one side and adjustment-needed/missed on the other side. In this consideration, there is no difference in higher or lower chance when no one can predict which new field will be added in the future. So I thought there is no difference in the pros and cons between these two methods.

### In terms of readability

Although I understood what the problem is that `getSettings` is trying to solve at this point at the first time review, I didn't see the purpose of adding additional `getSettings` that is equivalent (at least in my opinion) to the opposite of its problem when there was already the un-glue processing in place. So I started with a ❓  to clarify if I was missing something.

After some discussion, I understood your reason is that `getSettings` is clearer and more explicit, and has a higher chance of being noticed. With the original `getSettings` code, I agree that subjectively it had a higher chance of being noticed, but for me it's because the function looks strange and confusing, so it's easy to notice, not because it's an explicit function. In addition, unless the explicit factor is simply and only the visual size of the code block, the glue and un-glue processings are not less explicit in comparison. At least they appear in pairs with their code comments.

Also, in the slack thread, I described additional confusion in reading the original `getSettings` code, as well as spending some time thinking about questions that seemed relevant but were out of the scope of this PR.

### My summary

In summary, readability was my main concern in the original `getSettings`. And I don't think it's convincing to infer which method has a higher or lower chance based on the developer's individual experience of discovering/encountering this problem in the past. Because if I were to speak from personal experience about the chance of discovering this problem, as a code reviewer role, I had noticed it in almost all of several recent PRs that have relevant changes. But I would not infer which approach has a ~80% chance of being noticed and therefore conclude that which has a higher chance.

I believe we could all agree that describing the purpose for `getSettings` and how to add a new field can more clearly simplify these contexts and their implied problems (table above).

---

Thanks for the code comments and adjustments!

In this part, it's still based on the readability and maintainability perspectives.

### Readability

After further adjustments, the purpose and context are clearer. Thanks! I will comment additionally could be more clearly described in a few sentences.

### Maintainability

❔

A new question arises.

The needed settings in `initialValues` below is the same as the result of `getSettings( settings )`. Can `getSettings` also be used to change it to the following?
```js
<Form
    initialValues={ {
        ...getSettings( settings ),
        // ... omitted below
    } }
>
```

Or is it risky to do so? If it's not appropriate to do so, whether it would be easy to misunderstand that it could be done in future maintenance?

### Further maintainability

📜  or ignore this section.

In my opinion, the root cause of these concerns and problems is mixing the UI state with \<Form\> values. This is exactly why I advocated that \<Form\> `initialValues` should be as simple as possible in https://github.com/woocommerce/google-listings-and-ads/pull/1266#discussion_r820377015.

Borrowing a \<Form\> value as a state might not be a good idea:
- The state will also appear in the arguments of `onChange` and `onSubmit` callbacks. The mix of form values and states in \<Form\> callbacks could be confusing when reading the code, further reducing readability and increasing complexity when maintaining it.
- The state will trigger \<Form\>'s `onChange` callback when calling `setValue`, which makes callback to be triggered when there is no "real" form values changed. So, additional processing was added:
   https://github.com/woocommerce/google-listings-and-ads/blob/67c33332901f24e7fd97dafa987271801d46608e/js/src/edit-free-campaign/setup-free-listings/index.js#L115-L117

Move `offer_free_shipping` out of `initialValues` could avoid these side effects and separate the responsibility of data/state handling across a complex form UI.

js/src/edit-free-campaign/setup-free-listings/index.js
```diff
@@ -19,6 +20,39 @@ import FormContent from './form-content';
...
+	'payment_methods_visible',
+	'refund_tos_visible',
+	'contact_info_visible',
+];
+
+/**
+ * Get settings object from Form values.
+ *
+ * This method is used to pick out form fields that are specific to settings.
+ * If we are adding a new settings field, it should be added into `settingsFieldNames`.
```
eason9487: 💅 
Would it be clearer to point out the **new settings field** as "a new settings field that will be saved via API"?

js/src/edit-free-campaign/setup-free-listings/index.js
```diff
@@ -19,6 +20,39 @@ import FormContent from './form-content';
...
+ *
+ * If a new field is added into the form that is not related to settings (e.g. `offer_free_shipping`),
```
eason9487: 💅
Would it be clearer to point out the **new field** is a new field that **will NOT be saved via API**?

eason9487 (approved):
Tested locally and works well. Left two 💅  comments and a new ❔ concern in the **Maintainability** section of https://github.com/woocommerce/google-listings-and-ads/pull/1317#discussion_r827607006. And none of them is a blocker for this PR to address the primary issue. Approved in advance.

js/src/edit-free-campaign/setup-free-listings/index.js
ecgan: Done in [`ce81aad` (#1317)](https://github.com/woocommerce/google-listings-and-ads/pull/1317/commits/ce81aad3b5a988027d3aa422ed1632ae1955c8e4).

js/src/edit-free-campaign/setup-free-listings/index.js
ecgan: Done in [`ce81aad` (#1317)](https://github.com/woocommerce/google-listings-and-ads/pull/1317/commits/ce81aad3b5a988027d3aa422ed1632ae1955c8e4).

js/src/edit-free-campaign/setup-free-listings/index.js
ecgan: Thanks for the detailed explanation, @eason9487 ! 🙂 👍 

Here's my opinions on your questions above:

> The needed settings in `initialValues` below is the same as the result of `getSettings( settings )`. Can `getSettings` also be used to change it to the following?

Technically yes, it can be used in that way, and devs could have also done it with `...settings` like below:

```jsx
<Form
    initialValues={ {
        ...getSettings( settings ),
        // ... some other fields.
		...settings
    } }
>
```

However, personally, I don't favor that, for few reasons:

- I prefer to see the long complete written list of fields inside the `initialValues` props. Whenever I need to recall the field name, I just need to find the form and look at its `initialValues`. It is a readability point for me.
- I also think it is risky and prone to error, see example below:

```jsx
<Form
    initialValues={ {
        ...getSettings( settings ),
        // ... some other fields.
		tax_rate: 'abc123' // this tax_rate overwrites tax_rate from getSettings, which may not be desired.
    } }
>
```

The above problem wouldn't happen when we have the complete list of fields in initialValues props, like how we are having it now in our code base. 

> If it's not appropriate to do so, whether it would be easy to misunderstand that it could be done in future maintenance?

Unfortunately I don't have an idea to prevent that, other than the code comments. 😅 Also, this problem isn't really specific to `getSettings`, as devs could have also done it with `...settings` in initialValues.


> In my opinion, the root cause of these concerns and problems is mixing the UI state with `<Form>` values.
> ...
> Move `offer_free_shipping` out of `initialValues` could avoid these side effects and separate the responsibility of data/state handling across a complex form UI.

Thanks for your opinion! Again, this is where we have different opinions. 😆 🙏  

I do think that it makes sense to put `offer_free_shipping` into the form, for the following reasons:

- It has a related UI with the "yes" / "no" radio options.
- It participates in form validation, and when validation fails (e.g. users did not select "yes" / "no" when it is required), we set `errors.offer_free_shipping` for it. See PR https://github.com/woocommerce/google-listings-and-ads/pull/1309.
- Related to the above `errors.offer_free_shipping`, if / when we want to show validation error messages to the users, we can get the error messages for all fields from the `errors` object in a coherent manner. 
    - If we were to do this with `useState`, then it would look a bit weird, i.e. for all other fields, we get the error messages from `errors` object, but for `offer_free_shipping` specifically, we have to get it using another way.
- When users change the radio options from "yes" to "no", we need to clear the shipping rates' free shipping threshold. There are interdependencies between them, and for me, and it is natural / common / make sense to have interdependencies between form fields, and not so much for form fields and useState.

On the other hand, if we choose to use `useState` or other mechanism to store `offer_free_shipping` value instead of using form, I think it would actually be more complicated. 

> Borrowing a `<Form>` value as a state might not be a good idea

Thanks for the points. With my above reasons mentioned, I think the points you raised are fine / expected / acceptable "trade-offs". I think it is expected and not an issue, for me. 💅 😄 

---

I have one question to ask, to get to know your opinion better and to be sure: 

> In my opinion, the root cause of these concerns and problems is mixing the UI state with `<Form>` values.

In your opinion, does it mean that all the Form values should always have a relevant API? So, fields that are purely UI state and not backed by API should not be in Form values? 🤔 

I'm asking in that way because we are talking about `offer_free_shipping` field, which is purely UI side and not backed by API.

Curious to know your thoughts here.

eason9487: Thanks for sharing your thought!

---

> I do think that it makes sense to put `offer_free_shipping` into the form, for the following reasons: [...]

I did understand the convenience of having the UI state associated with the form values. We can easily use form's operation APIs that are accessible in several places at development time.

However, I would like to emphasize that it also brings side effects that make state and event management less robust and requires checking for any new control handling overflows or pollutions during ongoing maintenance, and then adding more patches. Even though sometimes we can make those patches seem to be wroten well, sometimes they are workarounds.

Because forms are inherently complex, as applications continue to iteratively improve and become easier for merchants to use, things might get trickier and trickier. Especially this happens with/in forms.

> On the other hand, if we choose to use `useState` or other mechanisms to store `offer_free_shipping` value instead of using form, I think it would actually be more complicated.

One of the primary influences on the overall robustness of architecture is how data and states are managed. The use of `useState` or other state management mechanisms has been something that the React ecosystem, and even other frameworks, are constantly working on. It is true that these strategies usually bring complexity, but these complexities bring relatively robust management practices via reducing side effects.

> In your opinion, does it mean that all the Form values should always have a relevant API?

No, I don't mean that all form values need to be saved with API.

> So, fields that are purely UI state and not backed by API should not be in Form values?

It depends on how a \<Form\> component is designed, with Woo's one, It has some nice designs in other parts, but I wouldn't recommend doing so if it's possible.

PR #1319, opened 2022-03-11
1292: Fix fatal error when the plugin is activated before WooCommerce.

### LINKED ISSUE #1292: Activating Google Listings & Ads (before WooCommerce) led to a fatal error.
### Describe the bug:

Fatal error caused when Google Listings and Ads was activated before WooCommerce.

https://user-images.githubusercontent.com/92185881/156963422-3d32890b-5502-412d-8693-031d80ebe663.mov

The short version above where Google Listings & Ads was activated without WooCommerce.

https://user-images.githubusercontent.com/92185881/156963625-d8b51365-2eb3-4756-837b-56e368eef59d.mp4

The full video showed the error and how to correctly activate both Google Listings and Ads and WooCommerce.

### Steps to reproduce:

1. Installed both Google Listings & Ads and WooCommerce (both not activated yet).
2. Activate Google Listings & Ads first. Results in a fatal error.
3. Activate WooCommerce first, and then Google Listings & Ads. All good!

### Expected behavior:

A helpful error message such as "WooCommerce is not yet installed or activated. Would you like to start your WooCommerce store?" instead of a fatal error.

### Actual behavior:

A fatal error message.

### Additional details:

I encountered this while troubleshooting a website by deactivating all plugins and activating the Automattic + WooCommerce plugins in bulk. This possibly led to the plugins being activated in alphabetical order, which meant Google Listing & Ads, before WooCommerce, that led to the fatal error message.

rawdreeg:

* Add requirement check in the plugins_loaded hook.
* Made ActivatedEvents implement Activateable interfaces. 
* Provide DELAYED_ACTIVATE option flag.
* Set DELAYED_ACTIVATE flag when PluginFactory::instance()->activate() has failed. 
* If DELAYED_ACTIVATE is set, run activate() on plugin_loaded and remove the DELAYED_ACTIVATE flag

Closes #1292.

### Screenshots:

WC not installed

Screenshot 2022-03-17 at 16 31 07

1. Installed both Google Listings & Ads and WooCommerce (both not activated yet).
2. Activate Google Listings & Ads first. You should see a notice stating that WC needs to be activated first.
3. Activate WooCommerce first, and then Google Listings & Ads. All good!

### Changelog entry

> Fix - Fatal error when the plugin is activated before WooCommerce.


src/Internal/Requirements/RequirementValidator.php
```diff
@@ -39,6 +39,14 @@ public static function instance(): RequirementValidator {
 	 * @param RuntimeException $e
 	 */
 	protected function add_admin_notice( RuntimeException $e ) {
+		// Check if the plugin is active. If not, display error message and terminate.
+		if ( ! is_plugin_active( 'woocommerce-google-listings-and-ads/woocommerce-google-listings-and-ads.php' ) ) {
```
mikkamp: The plugin and folder don't have the `woocommerce-` prefix. This means that we will always die with the error and never show the admin notice. There are certain cases when we already activated the plugin we want to show the admin notice.

There are also cases where it gets installed in a different folder. Often this is when a version gets added in the folder name. We can probably have a look how that's done in other extensions but otherwise something along the lines of setting a constant in the main plugin file like:
```php
define( 'WC_GLA_INSTALLED_DIR', basename( __DIR__ ) );
```

And then using that here:
```suggestion
		if ( ! is_plugin_active( trailingslashit( WC_GLA_INSTALLED_DIR ) . 'google-listings-and-ads.php' ) ) {
```

You could probably also pull in PluginHelper since that's a trait, and then add a helper function to get the active plugin string (get_main_filename is already there).

mikkamp: Another point about `is_plugin_active`, that function is only available in certain circumstances (when plugin is activating or after admin_init hook). So if I have the plugin already activated and then I simulate one of the scenarios where it's missing a requirement it gives a fatal error that the function is undefined.

Maybe something like this is needed:
```php
if ( ! function_exists( 'is_plugin_active' ) ) {
	require_once ABSPATH . '/wp-admin/includes/plugin.php';
}

**Review by mikkamp — COMMENTED** [review]
Thanks for the changes, I had some trouble testing this as outlined in the comments.

Also I tried testing with our plugin activated, and then deactivating WooCommerce. If that happens the notice won't be shown because our plugin initializes on the hook `woocommerce_loaded` so if that hook is not called we never load the PluginValidator class. We might want to refactor that although we'd have to set some internal state so we can prevent it from registering services if we call it from a different location.

File: src/Internal/Requirements/RequirementValidator.php [hunk 2]
```diff
@@ -39,6 +39,14 @@ public static function instance(): RequirementValidator {
...
+			die(
+				esc_html( $e->getMessage() )
```
mikkamp: For wp-cli we might want to add a PHP_EOL here otherwise when I activate on the command line it doesn't jump to the next line before showing the prompt.

src/Internal/Requirements/RequirementValidator.php
rawdreeg: Thanks @mikkamp, I have added this check.

src/Internal/Requirements/RequirementValidator.php
rawdreeg: Thanks, @mikkamp, I have appended PHP_EOL to the string.

src/Internal/Requirements/RequirementValidator.php
rawdreeg: Thanks for picking this up. I was being completely daft on this. I have used the PluginHelper trait.

rawdreeg:
> Thanks for the changes, I had some trouble testing this as outlined in the comments.
> 
> Also I tried testing with our plugin activated, and then deactivating WooCommerce. If that happens the notice won't be shown because our plugin initializes on the hook `woocommerce_loaded` so if that hook is not called we never load the PluginValidator class. We might want to refactor that although we'd have to set some internal state so we can prevent it from registering services if we call it from a different location.

@mikkamp Regarding the scenario where WooCommerce is deactivated while our extension is active, I thought about this a little:

Could we use this action hook:

https://developer.wordpress.org/reference/hooks/deactivated_plugin/

and call `PluginFactory::instance()->deactivate(); `on that action. Obviously, the deactivate will need to be refactored a bit. What do you think?

R

mikkamp:
> Could we use this action hook:
> https://developer.wordpress.org/reference/hooks/deactivated_plugin/
> and call PluginFactory::instance()->deactivate(); on that action. Obviously, the deactivate will need to be refactored a bit. What do you think?

We could run some code on that hook. Although that function states it's called right after a plugin is deactivated. So if we call `PluginFactory::instance()->deactivate()` the `registered_services` array would be empty, so it wouldn't deactivate any individual service. Probably better to register a deactivation hook like [here](https://github.com/woocommerce/pinterest-for-woocommerce/pull/373/commits/3517d2e0b5b10b782367693ec312bac2593fe33d).
One thing to keep in mind though is that if we toggle WC off and then on we would need the activation hook as well, which I don't think is very useful, because all that time our plugin remained active so we didn't need to reset everything.

rawdreeg:
> We could run some code on that hook. Although that function states it's called right after a plugin is deactivated. So if we call `PluginFactory::instance()->deactivate()` the `registered_services` array would be empty, so it wouldn't deactivate any individual service. Probably better to register a deactivation hook like [here](https://github.com/woocommerce/pinterest-for-woocommerce/pull/373/commits/3517d2e0b5b10b782367693ec312bac2593fe33d). One thing to keep in mind though is that if we toggle WC off and then on we would need the activation hook as well, which I don't think is very useful, because all that time our plugin remained active so we didn't need to reset everything.

@mikkamp I was able to test that registere_services is definitely picked up when WC is deactivated. I just deactivate our plugin when WC is deactivated. One thing I haven't been able to do is to throw some sort of notice that our plugin has been disabled because WC is not enabled. Other than that this works as intended. Any suggestion on how I can show some notice on deactivation?

google-listings-and-ads.php
```diff
@@ -67,6 +67,34 @@ function () {
...
+/**
+ * Remove our plugin if WooCommerce is deactivated.
+ *
+ * @param string $plugin
+ *   Path to the plugin file relative to the plugins directory.
+ * @param bool   $network_activation
+ *  Whether the plugin is being activated for all sites in the network
+ */
+function detect_wc_deactivation( string $plugin, bool $network_activation ): void {
+	if ( $plugin === 'woocommerce/woocommerce.php' ) {
```
mikkamp: I think it would be best to use `plugin_basename( WC_PLUGIN_FILE )` if it's defined. Although I have mixed feelings about forcing our extension to deactivate when WooCommerce is deactivated. Maybe we can discuss that more with the team.

Also if we decide to go ahead with this I'd suggest to move the logic for deactivating into the GoogleListingsAndAdsPlugin class.

If we need anything in the main plugin file, I'd suggest to at most add a deactivation hook for WC, something like:
```php
// Register deactivation hook for WooCommerce
if ( defined( 'WC_PLUGIN_FILE' ) {
	register_deactivation_hook(
		WC_PLUGIN_FILE,
		function ( $network_wide ) {
			PluginFactory::instance()->woocommerce_deactivated( $network_wide );
		}
	);
}
```

mikkamp (commented):
Thanks for the changes, I'm no longer able to reproduce the fatal.

> One thing I haven't been able to do is to throw some sort of notice that our plugin has been disabled because WC is not enabled. Other than that this works as intended. Any suggestion on how I can show some notice on deactivation?

I'm not a 100% on whether we want to force deactivate, because that's kind of hidden from the user and I'm not aware of a solution to show a notice because the code in the extension no longer runs.

If we would decide to leave the extension active I think we should notify them that it doesn't do anything because WC isn't active. For that we would need to do something along the lines of:
- Run `PluginValidator::validate` on `plugins_loaded` instead of `woocommerce_loaded`
- Since it's a static call we can store a static result of validate within the class
- In `maybe_register_services` call `PluginValidator::is_validated`

google-listings-and-ads.php
rawdreeg: I have updated the PR with the following:

- Add requirement check in the plugins_loaded hook.
- Made ActivatedEvents implement Activateable, Deactivateable interfaces. Run maybe_track_activation_source() on activation.
- Provide TRACKED_ACTIVATION_SOURCE option flag.
- Set TRACKED_ACTIVATION_SOURCE flag when maybe_track_activation_source().
- Run maybe_track_activation_source() on register to run maybe_track_activation_source() if it was not run on Activate for any reasons

google-listings-and-ads.php
```diff
@@ -37,11 +37,6 @@
 	return;
 }
 
-// Validate the required versions of everything our plugin depends on.
-if ( ! VersionValidator::instance()->validate() ) {
-	return;
-}
-
```
mikkamp: Why was this removed? The VersionValidator is mainly used for confirming the minimum PHP version is met, which we want to do as early as possible. I don't see that check being done anywhere else after the changes.

src/Infrastructure/GoogleListingsAndAdsPlugin.php
```diff
@@ -88,6 +94,11 @@ public function register(): void {
 			self::SERVICE_REGISTRATION_HOOK,
 			function() {
 				$this->maybe_register_services();
+
+				// Check if activation source is set.
+				if ( ! $this->options->get( OptionsInterface::TRACKED_ACTIVATION_SOURCE ) ) {
+					$this->container->get( ActivatedEvents::class )->maybe_track_activation_source();
+				}
```
mikkamp: Could you clarify why this is done here?

Based on the discussions, I thought our intention was to store an option `PLUGIN_ACTIVATED`, which would stay false until the activate function has run with the requirement check passing. So I'd be expecting to see something like this here:
```php
if ( ! $this->options->get( OptionsInterface::PLUGIN_ACTIVATED ) ) {
	$this->activate();
}
```

And then within the activate function we check something like `PluginValidator::validated()` and have it run the activation and update the option if the conditions are met. Something like that would be forcing the activation code to be run (albeit delayed) after dependencies have been met.

src/Tracking/Events/ActivatedEvents.php
```diff
@@ -74,5 +85,24 @@ public function maybe_track_activation_source(): void {
...
+
+	}
+
+	/**
+	 * Activate the service.
+	 *
+	 * @return void
+	 */
+	public function activate(): void {
+		$this->maybe_track_activation_source();
```
mikkamp, line 88: Moving this to activate is a good thing because we can ensure that all the dependencies which are required for this class are available. However if we look at the code of `maybe_track_activation_source` it relies on server variables which are set during the call to any functions hooked into `register_activation_hook`. If we are activating the plugin at a later stage those variables won't be set so it can't detect how the plugin was activated. Note, that this is only for tracking purposes, so it won't break anything, but it will mean we will be tracking activation source incorrectly for any site that activated itself later.

src/Tracking/Events/ActivatedEvents.php
```diff
@@ -50,6 +58,9 @@ public function register(): void {}
 	 * Track when the extension is activated from a source.
 	 */
 	public function maybe_track_activation_source(): void {
+		// Update db flag.
+		$this->options->update( OptionsInterface::TRACKED_ACTIVATION_SOURCE, true );
```
mikkamp: If our intention is to have an option keep track of whether the `activate` code was run, then I don't think this option is needed, since we can confirm that the main `activate` function is called only once.

src/Infrastructure/GoogleListingsAndAdsPlugin.php
```diff
@@ -40,6 +46,7 @@ final class GoogleListingsAndAdsPlugin implements Plugin {
 	 */
 	public function __construct( ContainerInterface $container ) {
 		$this->container = $container;
+		$this->set_options_object( new Options() );
```
mikkamp: Since this is the main class the OptionsAware interface and trait don't really work or make sense. Also we don't generally instantiate new classes if they are provided by the container, we let the container handle whether it should return a new or a shared instance. If options is needed in this class then I'd suggest to create a new private member and then set it from the container.

```suggestion
		$this->options   = $container->get( OptionsInterface::class );
```

mikkamp (commented):
Thanks for the changes, I've left some comments as I thought our intention was to have an option to confirm that the activate code was run only once (albeit delayed) and not specifically for tracking the activation source.

google-listings-and-ads.php
rawdreeg: I replaced this with:

```
		// Check requirements.
		if ( ! PluginValidator::validate() ) {
			return;
		}
```

in the plugins_loaded hook

src/Infrastructure/GoogleListingsAndAdsPlugin.php
rawdreeg: Ah. Thanks for this. I will correct

src/Tracking/Events/ActivatedEvents.php
rawdreeg: I am looking at https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/Events/ActivatedEvents.php#:~:text=maybe_track_activation_source , it seems we’re only tracking on install (plugin-install.php), so even currently, we don't track later stage activation at all.

google-listings-and-ads.php
rawdreeg: added this check back. thanks

src/Infrastructure/GoogleListingsAndAdsPlugin.php
rawdreeg: Thanks, as per our previous discussion. I implemented the PLUGIN_ACTIVATED flag. I am checking the flag in the init hook to avoid issues with wpdb:$action_scheduler that only get checked on init.

src/Infrastructure/GoogleListingsAndAdsPlugin.php
rawdreeg: turns out, I get anything from the container at the constructor level. I decided not to initialize an `options` property.

rawdreeg:
@mikkamp I updated this PR with the following changes: 

* Add requirement check in the plugins_loaded hook.
* Made ActivatedEvents implement Activateable interfaces. 
* Provide PLUGIN_ACTIVATED option flag.
* Set PLUGIN_ACTIVATED flag when PluginFactory::instance()->activate() has executed successfully. 
* Delay the activate() execution on init if it hasn't been run previously due to missing dependencies

Thanks,

src/Infrastructure/GoogleListingsAndAdsPlugin.php
```diff
@@ -101,6 +110,13 @@ function() {
 				if ( $this->container->has( JobInitializer::class ) ) {
 					$this->container->get( JobInitializer::class )->register();
 				}
+
+				// Check if plugin requirements are valid and run activation.
+				if ( PluginValidator::validate()
+					&& ! $this->container->get( OptionsInterface::class )->get( OptionsInterface::PLUGIN_ACTIVATED ) ) {
+					$this->activate();
+				}
```
mikkamp: Can we have a static in the PluginValidator class that stores the validated value with a helper function to check it? Otherwise we are running it a second time, I think it's better to minimise the amount of code we run if it's potentially going to be each page load. `PluginValidator::validate()` is actually called 3 times on each request where it doesn't fail, so it would be nice to ensure it only gets called the first time on plugins_loaded priority 1.

I was also considering having it hook into admin_init like the example [here](https://developer.wordpress.org/reference/functions/register_activation_hook/#process-flow). Although I think that would limit the activation to be run from WP-CLI.

Minor: There isn't any fixed standard for WP, but the way it's indented makes it confusing at first glance whether line 117 is part of the if condition or the code block. Would it be better to put the closing on the next line like [here](https://www.php-fig.org/psr/psr-12/#51-if-elseif-else)?

src/Infrastructure/GoogleListingsAndAdsPlugin.php
```diff
@@ -50,15 +50,21 @@ public function __construct( ContainerInterface $container ) {
 	public function activate(): void {
 		$this->maybe_register_services();
 
+		// Do not run activate if service registration failed.
+		if ( empty( $this->registered_services ) ) {
+			return;
+		}
```
mikkamp: If we have a static value to return in PluginValidator, can we check that here instead of an empty set of registered services? That will make it a little more self documenting to explain why we don't want to activate. Also can we move it to before `maybe_register_services` since it doesn't make sense to register services if we know it's going to fail.

src/Infrastructure/GoogleListingsAndAdsPlugin.php
```diff
@@ -88,6 +96,7 @@ public function register(): void {
 			self::SERVICE_REGISTRATION_HOOK,
 			function() {
 				$this->maybe_register_services();
+
```
mikkamp: Minor: If we aren't adding any code, no need to add a newline.

mikkamp (commented):
Thanks for all the changes it's getting there.

Testing the code a bit more I noticed that for plugins that have already been installed and update to this newer code. They will have activated but the option PLUGIN_ACTIVATED will not be true. Which means it will run the activate code a second time, in theory that's safe because all the code in the service activate functions can be repeated. However I don't think we should be introducing a pattern where the activation functions can be called twice.

I would prefer to follow the the example [here](https://developer.wordpress.org/reference/functions/register_activation_hook/#process-flow), so that would mean:

1. In the activation hook if PluginValidator fails set option DELAYED_ACTIVATE to true
2. On init if PluginValidator succeeds and DELAYED_ACTIVATE is true
   - Run activate
   - Delete DELAYED_ACTIVATE

src/Tracking/Events/ActivatedEvents.php
rawdreeg: Thanks. This has been addressed in the newer commits

src/Infrastructure/GoogleListingsAndAdsPlugin.php
rawdreeg: Thanks, I have updated the check to check for PluginValidator::$is_validated. I realized the activation hook runs before plugin_loaded so PluginValidate::validate() is not called at all. I kept the call in  maybe_register_services() to be able to update PluginValidator::$is_validated value

src/Infrastructure/GoogleListingsAndAdsPlugin.php
rawdreeg: Should be corrected.

src/Infrastructure/GoogleListingsAndAdsPlugin.php
rawdreeg: I have updated this. And fixed the indentation (Sorry, I carry that from Symfony and Drupal -but will learn ).

Note: I realized the activation hook runs before plugin_loaded so PluginValidate::validate() is not called at all. I kept the call in maybe_register_services()

rawdreeg:
@mikkamp made the following changes in my update:

- Provide DELAYED_ACTIVATE option flag.
- Set DELAYED_ACTIVATE flag when PluginFactory::instance()->activate() has failed.
- If DELAYED_ACTIVATE is set, run activate() on plugin_loaded and remove the DELAYED_ACTIVATE flag

I also addressed the last round of comments.

Thanks

src/Infrastructure/GoogleListingsAndAdsPlugin.php
```diff
@@ -50,6 +50,15 @@ public function __construct( ContainerInterface $container ) {
 	public function activate(): void {
 		$this->maybe_register_services();
 
+		// Do not run activate if service registration failed.
```
mikkamp: Can we clarify here that it's because a Plugin requirement is not valid or the PluginValidator failed?

I also don't really like that we have to call maybe_register_services first to detect if it's been validated, I think we should break out of the activate function as early as possible.

Maybe we can refactor PluginValidator a bit, something like:
```php
private static $is_validated = null;

public static function validate(): bool {
	if ( null !== self::$is_validated ) {
		return self::$is_validated;
	}

	self::$is_validated = true;
	...Rest of validation code here
```

That way we can always call `validate` and it will ensure internally that it's only run once. Otherwise the way it is now it will still run the validation twice in certain circumstances. It's not so nice when that results in a double notice:
image

After that refactor it should in theory be safe to remove the validation check from `maybe_register_services` but we might want to leave it in place, because there could still be another plugin which calls `woogle_get_container()`.

mikkamp (commented):
Thanks for the changes, I'm getting the delayed activation to work nicely and it doesn't trigger for sites where I've already activated.

I just left a comment about a small refactor to prevent a double notice. After that it should be pretty much good to go.

src/Infrastructure/GoogleListingsAndAdsPlugin.php
rawdreeg: Thanks, @mikkamp, nice catch. I implemented the suggested refactor. And I decided to leave the validation check from maybe_register_services just to be safe.

mikkamp: Thanks that gets rid of the double notice. I think it would be nicer to make `$is_activated` private/protected and then refactor activate slightly (I think it's important to break out early):

```php
	public function activate(): void {
		// Delay activation if a required plugin is missing or an incompatible plugin is active.
		if ( ! PluginValidator::validate() ) {
			// Using update_option because we cannot access the option service
			// when the services have not been registered.
			update_option( 'gla_' . OptionsInterface::DELAYED_ACTIVATE, true );
			return;
		}

		$this->maybe_register_services();
		...
```

rawdreeg: Good catch. It makes things more consistent as well. I have implemented the refactor.

mikkamp (approved):
Thanks that looks cleaner. I haven't found any other way to break the code so it looks good to go.

The only thing I noticed that we can still change is in the activate function we can move the `$this->maybe_register_services()` line below the block where it does the validation. If the validation fails we don't need those services so it's cleaner to put the exit blocks at the top of a function.

rawdreeg:
> The only thing I noticed that we can still change is in the activate function we can move the `$this->maybe_register_services()` line below the block where it does the validation. If the validation fails we don't need those services so it's cleaner to put the exit blocks at the top of a function.

I have moved the  `$this->maybe_register_services()` below the requirement check block. Makes a lot more sense. Thanks for all the help (and patience) on this one. I feel more comfortable around the codebase as well.

PR #1321, opened 2022-03-11
Tree Select Component - Search functionality

puntope:

This PR implements the Search Functionality in the Tree Select Component. 

### Screenshots:

https://user-images.githubusercontent.com/5908855/158163460-047069e6-f9a9-4462-8d9d-74284b5fab5a.mov

1.  Checkout this PR 
2. Go to `wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fcomponent-test` 
3. Click on the search input and type some text
4. The search function should trigger after 3 characters are typed
5. It only shows options matching with the text typed
6. The labels of the matching options have the label highlighted with the match
7. It expands all the nodes
8. Removing to less than 3 characters doesn't apply any filter

### DIFF
--- a/js/src/components/tree-select-control/control.js
+++ b/js/src/components/tree-select-control/control.js
@@ -21,7 +21,7 @@ import Tags from './tags';
  * @param {boolean} props.disabled True if the component is disabled
  * @param {Function} props.onFocus On Focus Callback
  * @param {Function} props.onTagsChange Callback when the Tags change
- * @param {Function} props.onSearch Callback when the Input value changes
+ * @param {Function} props.onInputChange Callback when the Input value changes
  * @return {JSX.Element} The rendered component
  */
 const Control = ( {
@@ -32,7 +32,7 @@ const Control = ( {
 	disabled,
 	onFocus = () => {},
 	onTagsChange = () => {},
-	onSearch = () => {},
+	onInputChange = () => {},
 } ) => {
 	const inputRef = useRef();
 	const hasTags = tags.length > 0;
@@ -78,7 +78,7 @@ const Control = ( {
 					aria-expanded={ isExpanded }
 					disabled={ disabled }
 					onFocus={ onFocus }
-					onChange={ onSearch }
+					onChange={ onInputChange }
 				/>
 			</div>
 		</div>
--- a/js/src/components/tree-select-control/control.js
+++ b/js/src/components/tree-select-control/control.js
@@ -21,6 +21,7 @@ import Tags from './tags';
  * @param {boolean} props.disabled True if the component is disabled
  * @param {Function} props.onFocus On Focus Callback
  * @param {Function} props.onTagsChange Callback when the Tags change
+ * @param {Function} props.onSearch Callback when the Input value changes
  * @return {JSX.Element} The rendered component
  */
 const Control = ( {
@@ -31,6 +32,7 @@ const Control = ( {
 	disabled,
 	onFocus = () => {},
 	onTagsChange = () => {},
+	onSearch = () => {},
 } ) => {
 	const inputRef = useRef();
 	const hasTags = tags.length > 0;
@@ -76,6 +78,7 @@ const Control = ( {
 					aria-expanded={ isExpanded }
 					disabled={ disabled }
 					onFocus={ onFocus }
+					onChange={ onSearch }
 				/>
 			</div>
 		</div>
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { escapeRegExp, cloneDeep } from 'lodash';
+import { cloneDeep } from 'lodash';
 import { __ } from '@wordpress/i18n';
 import { useEffect, useMemo, useState, useRef } from '@wordpress/element';
 import classnames from 'classnames';
@@ -87,6 +87,7 @@ const TreeSelectControl = ( {
 	const [ filter, setFilter ] = useState( '' );
 	const [ filteredOptions, setFilteredOptions ] = useState( [] );
 
+	// We will save in a REF previous search filter queries to avoid re-query the tree and save performance
 	const searchFiltersCache = useRef( {} );
 
 	const treeOptions = useIsEqualRefValue(
@@ -105,14 +106,11 @@ const TreeSelectControl = ( {
 		setTreeVisible( false );
 	} );
 
+	useEffect( () => {
+		queryOptions();
+	}, [ treeOptions, filter ] );
+
 	const hasChildren = ( option ) => option.children?.length;
-
-	const getSearchFilterCache = () =>
-		searchFiltersCache.current && searchFiltersCache.current[ filter ];
-
-	const setSearchFilterCache = ( newValue ) => {
-		searchFiltersCache.current[ filter ] = newValue;
-	};
 
 	/**
 	 * Optimizes the performance for getting the tags info
@@ -136,6 +134,12 @@ const TreeSelectControl = ( {
 
 		return repository;
 	}, [ treeOptions ] );
+
+	const getSearchFilterCache = () => searchFiltersCache.current[ filter ];
+
+	const setSearchFilterCache = ( newValue ) => {
+		searchFiltersCache.current[ filter ] = newValue;
+	};
 
 	/**
 	 * Get formatted Tags from the selected values.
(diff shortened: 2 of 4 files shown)


js/src/components/tree-select-control/control.js
```diff
@@ -31,6 +32,7 @@ const Control = ( {
 	disabled,
 	onFocus = () => {},
 	onTagsChange = () => {},
+	onSearch = () => {},
```
eason9487: ❔ 
I'm curious why it was named `onSearch` rather than `onChange`?

💅 
And the default callback `() => {}` seems to be unnecessary because it's ok to set `onChange` prop as `undefined` onto a \<input\>.

js/src/components/tree-select-control/index.js
```diff
@@ -100,6 +106,12 @@ const TreeSelectControl = ( {
 		setTreeVisible( false );
 	} );
 
+	useEffect( () => {
+		queryOptions();
+	}, [ treeOptions, filter ] );
```
eason9487: ❓ Suggest avoiding the `react-hooks/exhaustive-deps` warning.

image

js/src/components/tree-select-control/index.js
```diff
@@ -217,6 +235,93 @@ const TreeSelectControl = ( {
...
+	/**
+	 * Perform the search query filter in the Tree options
+	 *
+	 * 1. Check if the search query is already cached and return it if so.
+	 * 2. Deep Copy the tree. Since we are going to modify his children and labels recursively.
+	 * 3. In case of filter, we apply the filter option function to the tree.
+	 *
+	 */
+	const queryOptions = () => {
+		const cachedFilter = getSearchFilterCache( filter );
```
The naming of `cachedFilter` looks like it's a filter but actually it's a cached result of filtered options. I'd suggest renaming this variable.

js/src/components/tree-select-control/index.js
```diff
@@ -217,6 +235,93 @@ const TreeSelectControl = ( {
...
+			return;
+		}
+
+		let filteredTreeOptions = cloneDeep( treeOptions );
+
+		if ( filter ) {
+			filteredTreeOptions = filteredTreeOptions.filter( filterOption );
+		}
+
+		setSearchFilterCache( filteredTreeOptions );
```
eason9487: 🚧 
When `options` or `selectAllLabel` props of \<TreeSelectControl\> change, the cache processing does not take it into account, resulting in incorrect results.

Kapture 2022-03-15 at 14 57 09

---

💅 
If `getSearchFilterCache` and `setSearchFilterCache` functions were not expected to be reused later, probably they could be omitted and then operate the cache inside this function.

https://github.com/woocommerce/google-listings-and-ads/blob/be2d6bfaffdac070dbaf657912d1d992f28eab24/js/src/components/tree-select-control/index.js#L138-L142

js/src/components/tree-select-control/index.js
```diff
@@ -217,6 +235,93 @@ const TreeSelectControl = ( {
...
+	 */
+	const filterOption = ( option ) => {
+		if ( hasChildren( option ) ) {
+			option.children = option.children.filter( filterOption );
+			return !! option.children.length;
+		}
+
+		const match = option.label
+			.toLowerCase()
+			.indexOf( filter.toLowerCase() );
```
eason9487: 💅 
The case insensitive handling of search string could be moved to `handleOnSearch` to increase the cache hit rate.

js/src/components/tree-select-control/index.js
```diff
@@ -217,6 +235,93 @@ const TreeSelectControl = ( {
...
+
+		if ( cachedFilter ) {
+			setFilteredOptions( cachedFilter );
+			return;
+		}
+
+		let filteredTreeOptions = cloneDeep( treeOptions );
+
+		if ( filter ) {
+			filteredTreeOptions = filteredTreeOptions.filter( filterOption );
```
eason9487: 💅 
This processing could also be done by recursive array `reduce`, which eliminates the need for `cloneDeep`.

js/src/components/tree-select-control/index.js
```diff
@@ -217,6 +235,93 @@ const TreeSelectControl = ( {
...
+			<span>
+				<span>{ optionLabel.substring( 0, matchPosition ) }</span>
+				<strong>
+					{ optionLabel.substring(
+						matchPosition,
+						matchPosition + filter.length
+					) }
+				</strong>
+				<span>
+					{ optionLabel.substring( matchPosition + filter.length ) }
```
eason9487: 💅  How about giving a variable for `matchPosition + filter.length`?

js/src/components/tree-select-control/index.js
```diff
@@ -217,6 +235,93 @@ const TreeSelectControl = ( {
...
+
+	/**
+	 * Applies the strong tag in the Option label characters matching the filter
+	 *
+	 * @param {string} optionLabel The label to highlight
+	 * @param {number} matchPosition The position when the highlight starts based on the filter match
+	 * @return {JSX.Element|*} A highlighted label
+	 */
+	const highlightOptionLabel = ( optionLabel, matchPosition ) => {
+		if ( matchPosition < 0 || ! filter?.length ) return optionLabel;
```
eason9487: ❔ 
Wondering are there any cases the `filter` will be undefined or null? If not, the optional chaining operator seems to be unnecessary here. Or even this check could be removed entirely.

js/src/components/tree-select-control/options.js
```diff
@@ -17,16 +17,18 @@ import { ROOT_VALUE } from './constants';
...
- * @param {Option[]} params.options List of options to be rendered
- * @param {string[]} params.value List of selected values
- * @param {string[]} params.nodesExpanded List of expanded nodes.
- * @param {Function} params.onChange Callback when an option changes
- * @param {Function} params.onNodesExpandedChange Callback when a node is expanded/collapsed
+ * @param {Object} props Component parameters
+ * @param {Option[]} props.options List of options to be rendered
+ * @param {string[]} props.value List of selected values
+ * @param {string[]} props.nodesExpanded List of expanded nodes.
+ * @param {string} props.filter The filter applied to the Tree Options
```
eason9487: 💅  Would it be better to indicate it's an optional prop with a default value?

js/src/components/tree-select-control/options.js
```diff
@@ -83,9 +90,10 @@ const Options = ( {
 
 	return options.map( ( option ) => {
 		const isRoot = option.value === ROOT_VALUE;
-		const isExpanded = isRoot || nodesExpanded.includes( option.value );
 		const hasChildren = !! option.children?.length;
 		const optionIsChecked = isChecked( option );
+		const isExpanded =
+			filter.length || isRoot || nodesExpanded.includes( option.value );
```
eason9487: 🚧 
Here needs typecasting for `filter.length`. Or, maybe change the `filter` prop to `isSearching` or `isFiltering`. Otherwise, it will set an invalid value to `aria-expanded` attribute.

js/src/components/tree-select-control/index.js
```diff
@@ -217,6 +235,93 @@ const TreeSelectControl = ( {
...
+	 *
+	 * This function mutates the option
+	 *
+	 * @param {Option} option The option to filter
+	 * @return {boolean} True if it pass the filter, false otherwise
+	 */
+	const filterOption = ( option ) => {
+		if ( hasChildren( option ) ) {
+			option.children = option.children.filter( filterOption );
+			return !! option.children.length;
```
By the design in Figma, parent options will still be displayed even if there are no matching children. Has the requirement changed?

image

js/src/components/tree-select-control/control.js
puntope: > I'm curious why it was named onSearch rather than onChange?

Since the control can trigger Tags Change and Input Value Change I thought it was more descriptive `onSearch`

js/src/components/tree-select-control/index.js
puntope: I have huge doubts about this as well. 

My main wonder is. If we show the continents, what happens if the user expand them? I guess if so, we should remove the chevron. 

I'll ask around in Component P2

puntope: Another question (maybe for the future) is if the tree is multilevel, and we don't hide the parents. A parent without results but for example with 4-5 levels would be expandable several times to show nothing at the end.

js/src/components/tree-select-control/options.js
puntope: Insane catch! Thanks! Fixed

js/src/components/tree-select-control/options.js
puntope: Updated. Please notice I changed also the name to `{boolean} isFiltered ` following other suggestions

js/src/components/tree-select-control/index.js
puntope: Fixed in last commits

js/src/components/tree-select-control/index.js
puntope: Done in Last commits

js/src/components/tree-select-control/index.js
puntope: I'll apply that in a future PR together with other pending minor changes.

js/src/components/tree-select-control/index.js
puntope: Fixed in last commits

js/src/components/tree-select-control/index.js
puntope: Great you discovered this! Fixed in last commits

js/src/components/tree-select-control/index.js
puntope: Fixed in last commits

puntope:
Howdy @eason9487 

I pushed a few commits implementing some of your suggestions. 

- Fix Hooks dependencies
- Fix cache refresh on Options update. 
- Fix casting in filters/expand  check 
- Update some naming/logic 

Pending work I'd do in future PR

- Show Parent nodes without results (if needed)
- Use recursive reducer instead of verbose function with cloneDeep

Thanks  🙏

js/src/components/tree-select-control/control.js
eason9487: Got it.

In terms of the internal relationship between these components, I was thinking the `onSearch` of  \<Control\> seems to be much close to a `onChange` callback from the perspective of \<TreeSelectControl\>. Because it only forwards the callback to the `onChange` of \<input\> without doing anything related to search.

In addition, the relevant JSDoc describes the *change* behavior rather than *search*:
- https://github.com/woocommerce/google-listings-and-ads/blob/f222e095f1c3aa94d496283990b22f3621a31767/js/src/components/tree-select-control/control.js#L24
- https://github.com/woocommerce/google-listings-and-ads/blob/f222e095f1c3aa94d496283990b22f3621a31767/js/src/components/tree-select-control/index.js#L299

So, perhaps it could be considered using the `onChange` instead.

js/src/components/tree-select-control/index.js
```diff
@@ -100,6 +112,67 @@ const TreeSelectControl = ( {
 		setTreeVisible( false );
 	} );
 
+	/**
+	 * Perform the search query filter in the Tree options
+	 *
+	 * 1. Check if the search query is already cached and return it if so.
+	 * 2. Deep Copy the tree. Since we are going to modify his children and labels recursively.
```
eason9487: 💅 `his` looks like a typo?

```suggestion
	 * 2. Deep Copy the tree. Since we are going to modify its children and labels recursively.
```

js/src/components/tree-select-control/index.js
```diff
@@ -100,6 +112,67 @@ const TreeSelectControl = ( {
...
+			filteredTreeOptions = filteredTreeOptions.filter( filterOption );
+		}
+
+		filteredOptionsCache.current[ filter ] = filteredTreeOptions;
+		setFilteredOptions( filteredTreeOptions );
+	}, [ treeOptions, filter ] );
+
+	useEffect( () => {
+		queryOptions();
+	}, [ queryOptions ] );
```
eason9487: 💅 
The `useCallback` at [line 125](https://github.com/woocommerce/google-listings-and-ads/blob/f222e095f1c3aa94d496283990b22f3621a31767/js/src/components/tree-select-control/index.js#L125)
```js
const queryOptions = useCallback( () => {
```
could be replaced with
```js
useEffect( () => {
```

Thus, the `useEffect` at lines 170-172 could be omitted. But if the descriptive of purpose is better in this way now, then it's good to keep it as-is.

eason9487 (approved):
Thanks for the adjustments! Tested locally and works well. LGTM. Left some comments for minor things and they are not blockers for this PR.

puntope:
Thanks, @eason9487 I applied your last suggestions ❤️

PR #1325, opened 2022-03-14
Update copy for Free and Enhanced Listings merge

### LINKED ISSUE #1296: Update copy for Free Listings Merge
At the end of March, Free and Enhanced Listings will be merged together (enhanced listings will not exist) - they will simply be free listings.

As a result, a number of content areas within the extension will need to be checked and updated to reflect the change.

* Get Started / Landing Page
* Product Feed Overview (and tooltip/pop-up content)

image-9

- [ ] Update related copy in Woo Docs

Internal project thread reference with a spreadsheet version of the changes can be found at:

pcTzPl-nO-p2

eason9487:

Closes #1296 

- Remove all "Partially Active" wordings or sentences.
- Remove "standard" or "enhanced" listings wordings.

#### 📷  Issues to resolve table

The new screenshot after adopting the suggestion in https://github.com/woocommerce/google-listings-and-ads/pull/1325#issuecomment-1135936945.

The ~new~ old screenshot after resolving code conflicts and rebasing in https://github.com/woocommerce/google-listings-and-ads/pull/1325#issuecomment-1135753523.

The original screenshot when opening PR.

1. Go to the Product Feed page.
2. Check if the copies are the same as new copies in the spreadsheet of pcTzPl-nO-p2.
3. Global search `partially`, `standard`, and, `enhanced` in the codebase to check if there are related copies that are still using these wordings.

### Changelog entry

> Tweak - Update copy for Free and Enhanced Listings merge.

### DIFF
--- a/js/src/product-feed/product-statistics/index.js
+++ b/js/src/product-feed/product-statistics/index.js
@@ -58,7 +58,7 @@ const ProductStatistics = () => {
 							<SummaryNumber
 								key="active"
 								label={ __(
-									'Active / Partially Active',
+									'Active',
 									'google-listings-and-ads'
 								) }
 								value={ data.statistics.active }
--- a/js/src/product-feed/product-statistics/product-status-help-popover/index.js
+++ b/js/src/product-feed/product-statistics/product-status-help-popover/index.js
@@ -41,7 +41,7 @@ const ProductStatusHelpPopover = () => {
 			<p>
 				{ createInterpolateElement(
 					__(
-						'After submission, Google assigns each product a status: <strong>Active, Partially Active, Expiring, Pending, or Disapproved.</strong>',
+						'After submission, Google assigns each product a status: <strong>Active, Expiring, Pending, or Disapproved.</strong>',
 						'google-listings-and-ads'
 					),
 					map
@@ -50,16 +50,7 @@ const ProductStatusHelpPopover = () => {
 			<p>
 				{ createInterpolateElement(
 					__(
-						'<strong>‘Active’ products</strong> are fully approved and eligible to appear in standard and enhanced listings on Google.',
-						'google-listings-and-ads'
-					),
-					map
-				) }
-			</p>
-			<p>
-				{ createInterpolateElement(
-					__(
-						'<strong>‘Partially active’ products</strong> are fully approved and eligible to appear in standard listings only.',
+						'<strong>‘Active’ products</strong> are fully approved and eligible to appear in free listings on Google.',
 						'google-listings-and-ads'
 					),
 					map
--- a/js/src/product-feed/product-statistics/status-box/feed-status.js
+++ b/js/src/product-feed/product-statistics/status-box/feed-status.js
@@ -53,7 +53,7 @@ function FeedStatus() {
 			label={
 				<span className="gla-success">
 					{ __(
-						'Standard free listings setup completed',
+						'Free listings setup completed',
 						'google-listings-and-ads'
 					) }
 				</span>
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -98,7 +98,7 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 
 	/**
 	 * Get the Product Statistics (updating caches if necessary). This is the
-	 * number of product IDs with each status (active and partially active are combined).
+	 * number of product IDs with each status (approved and partially approved are combined as active).
 	 *
 	 * @param bool $force_refresh Force refresh of all product status data.
 	 *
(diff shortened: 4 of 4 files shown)


src/API/Site/Controllers/MerchantCenter/IssuesController.php
```diff
@@ -239,10 +239,10 @@ protected function get_schema_title(): string {
 	 */
 	protected function maybe_override_issue_values( array $issue ): array {
 		$is_account_issue = MerchantStatuses::TYPE_ACCOUNT === $issue['type'];
-		if ( $is_account_issue && "Account isn't eligible for enhanced free listings" === $issue['issue'] ) {
-			$issue['issue']      = 'Show products on additional surfaces across Google through enhanced free listings';
+		if ( $is_account_issue && 'merchant_quality_low' === $issue['code'] ) {
```
eason9487: Change to compare with the issue `code`. Ref: https://developers.google.com/shopping-content/guides/account-issues#merchant_quality_low

src/API/Site/Controllers/MerchantCenter/IssuesController.php
```diff
@@ -239,10 +239,21 @@ protected function get_schema_title(): string {
...
+		 * Ref: https://developers.google.com/shopping-content/guides/account-issues#merchant_quality_low
+		 *
+		 * Issue string "Account isn't eligible for free listings" for matching
+		 * the updated copy after Free and Enhanced Listings merge.
+		 *
+		 * TODO: Remove the condition of matching the $issue['issue']
+		 *       if its issue code is the same as 'merchant_quality_low'
+		 *       after Free and Enhanced Listings merge.
+		 */
+		if ( $is_account_issue && ( 'merchant_quality_low' === $issue['code'] || "Account isn't eligible for free listings" === $issue['issue'] ) ) {
```
ecgan: ❔ I've read the code comment, Google docs reference link and P2, but still wondering if we really need the `"Account isn't eligible for free listings" === $issue['issue']` part. I removed them manually in my testing and it still seems to work good. 

I'm thinking that we should have checked based on `'merchant_quality_low' === $issue['code']` in the first place, and checking based on `$issue['issue']` (whether it is "enhanced free listings" in the old code or "free listings" in the new code) isn't necessary. 

The TODO code comment says "Remove the condition of matching the $issue['issue'] if its issue code is the same as 'merchant_quality_low' after Free and Enhanced Listings merge." Will it be not same? I guess we don't know the answer and we are just trying to be a bit defensive here? If that's the case, I think that's fine.

ecgan (approved):
Tested and compared the UI with the new copies in the spreadsheet, everything looks good. Left a minor ❔ comment below on the PHP conditional check on `$issue`. LGTM 👍

src/API/Site/Controllers/MerchantCenter/IssuesController.php
eason9487: > Will it be not same? I guess we don't know the answer and we are just trying to be a bit defensive here? If that's the case, I think that's fine.

Yes. So far we only know that the content of `$issue['issue']` will become "Account isn't eligible for free listings", but it is not certain if they will use the same `$issue['code']`, so a compatible approach is taken here. After Google API response has been updated, the condition needs to be left here will be clear.

ecgan:
I have added `status: blocked` label in this PR, as we are pending confirmation from Google side.

eason9487:
📝  Related PR #1437

puntope:
> 📝 Related PR https://github.com/woocommerce/google-listings-and-ads/pull/1437

Yeah, I wanted to move maybe_override_issue to before database insertion instead to avoid inconsistencies when hiding warning issues. 

But it can be adjusted easily after merge :)

eason9487:
Since there are code conflicts and it had more than 1600 commits behind `develop`, the conflict was resolved together with rebasing onto the `develop`. And the visual result of the **Issues to resolve table** is different from before, the new one is below.

puntope:
> Since there are code conflicts and it had more than 1600 commits behind develop, the conflict was resolved together with rebasing onto the develop. And the visual result of the Issues to resolve table is different from before, the new one is below.

Right now we use the issue details (if any) to populate the action. Maybe we could add a bit more of information in action field instead. Or just remove it. (If action field is null, the link goes directly to the docs without modal)

eason9487:
@puntope, thanks for the nice suggestions! 💯 

I adopted the second one in https://github.com/woocommerce/google-listings-and-ads/pull/1325/commits/58ceb42f752941a1facb3bea6d992669fa192716 since the `action` returned from Google API is `null` already. The latest screenshot is below and is also added to PR description.

eason9487:
Removed the `status: blocked` label since the Google side has launched the Free and Enhanced Listings merge.

Issue #1047, opened 2021-10-13
Enhance CI jobs in the GitHub Actions


eason9487:
## User story

As a developer, it'd be great

- if the CI could run jobs as fast as possible
- if the CI could run only the jobs that really need to be run
- if the CI could display colorful logs

## Technical

Some possible action items

- [x] Cancel the previous jobs for pull requests that have not been completed - #1224
    - Could be done with the workflow syntax `concurrency`
    - https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency
    - https://github.com/WordPress/wordpress-develop/blob/5.8.1/.github/workflows/coding-standards.yml#L37-L42
- [x] Remove `node` and `npm` requirements during the **PHP Unit Tests** job - #1140
    - [Prepare node](https://github.com/woocommerce/google-listings-and-ads/blob/42c13e5f3989e59c68efd2b59d1b0a7bf24c6b07/.github/workflows/php-unit-tests.yml#L18-L21)
    - [Prepare node](https://github.com/woocommerce/google-listings-and-ads/blob/42c13e5f3989e59c68efd2b59d1b0a7bf24c6b07/.github/workflows/php-unit-tests.yml#L44-L45)
    - [Build client bundles](https://github.com/woocommerce/google-listings-and-ads/blob/42c13e5f3989e59c68efd2b59d1b0a7bf24c6b07/.github/workflows/php-unit-tests.yml#L60-L61)
    - ~~❓  Not sure if the client files within `js/build/` are required when running PHP unit tests~~
- [x] Evaluate `npm ci` caching and skipping, as discussed in [this thread](https://github.com/woocommerce/google-listings-and-ads/pull/1133/#discussion_r761753593) - #1220
- [x] Only run **JavaScript and CSS Linting** job when the change files are ~~under the `js/src` folder~~ including any js/scss files for pull requests. - #1326
- [x] Only run **PHP Coding Standards** job when the change files are ~~under the `src` folders~~ including any PHP files for pull requests. - #1326
- [x] Adjust **PHP Coding Standards** job to cover [all listed PHP files](https://github.com/woocommerce/google-listings-and-ads/blob/1.11.1/phpcs.xml.dist#L115-L117) by `./*`. The details can be found in [this comment](https://github.com/woocommerce/google-listings-and-ads/issues/1047#issuecomment-1065090293). - #1320
- [x] Clean up Travis CI related setup - #1056
[...]

PR #1326, opened 2022-03-14
Run partial CI jobs only for JS/SCSS/PHP file changes

eason9487:

Parts of #1047 

- Only run **JavaScript and CSS Linting** and **JavaScript Unit Tests** jobs when the change files are including `.js` or `.scss` files in pull requests, or in the commits on the primary branches.
- Only run **PHP Coding Standards** and **PHP Unit Tests** jobs when the change files are including `.php` files in pull requests, or in the commits on the primary branches.

Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

Check the three commits triggering jobs below, they should trigger more jobs as more files are changed.

💡  If change the order to commit PHP file changes first, it should expect only the PHP-related jobs to be added to trigger. But I think it should be ok to provide one of the verification flow in this PR.

### DIFF
--- a/.github/workflows/js-css-linting.yml
+++ b/.github/workflows/js-css-linting.yml
@@ -5,7 +5,17 @@ on:
     branches:
       - trunk
       - develop
+    paths:
+      - "**.js"
+      - "**.scss"
+      - ./.github/workflows/js-css-linting.yml
+      - ./.github/actions/prepare-node/action.yml
   pull_request:
+    paths:
+      - "**.js"
+      - "**.scss"
+      - ./.github/workflows/js-css-linting.yml
+      - ./.github/actions/prepare-node/action.yml
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/js-unit-tests.yml
+++ b/.github/workflows/js-unit-tests.yml
@@ -5,7 +5,15 @@ on:
     branches:
       - trunk
       - develop
+    paths:
+      - "**.js"
+      - ./.github/workflows/js-unit-tests.yml
+      - ./.github/actions/prepare-node/action.yml
   pull_request:
+    paths:
+      - "**.js"
+      - ./.github/workflows/js-unit-tests.yml
+      - ./.github/actions/prepare-node/action.yml
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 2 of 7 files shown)


puntope (commented):
Thanks for this PR. It would be much more efficient now with the paths. 

❓  I'm wondering what would happen if we are loading some kind of .svg file in the frontend and for example, we change the name of that .svg. Then this test will not run and we could potentially miss the error.

❓ A bit of the same happens with .json files in PHP. What if someone changes some mock.json file or `.js` file in the PHP test folder? 

Wondering if it'd better to add as well everything in ´/js/src´ for JS jobs and everything in `/src` and `/tests` for PHP jobs

eason9487:
@puntope, thanks for the review comments.

> ❓ I'm wondering what would happen if we are loading some kind of .svg file in the frontend and for example, we change the name of that .svg. Then this test will not run and we could potentially miss the error.

If the name of an SVG file is changed, usually the js files importing that SVG file would be changed accordingly. So in this case, I assumed the js related jobs would be triggered in most cases.

If the only change is to the content of an SVG file, then the js related jobs will not be triggered. But it's not a concern since the `.svg` files are mocked when running unit tests. They should not affect the results of unit tests.

https://github.com/woocommerce/google-listings-and-ads/blob/1fbbde745aef5670290e08da462da000eb7836da/jest.config.js#L7-L8

> ❓ A bit of the same happens with .json files in PHP. What if someone changes some mock.json file or `.js` file in the PHP test folder?

Sorry, I'm not 100% sure of your concern. I assumed you meant the js/json files under the `/tests/mocks` and `/tests/proxy` folders. If yes, these files are not used in PHP unit tests.

Apart from the two mocking js files under `/tests/data`, I don't find any json or js files are associated with PHP unit tests. And it should be ok to not include those mocking js files.

---

But maybe I should add composer.json, package.json, and their lock files to the `paths` of job configs accordingly. 🤔

eason9487:
Additional changes:
- Added composer.json and composer.lock to the `paths` of **PHP Unit Tests** job.
- Added package.json and package-lock.json to the `paths` of **JavaScript Unit Tests** job.

puntope (approved):
✅  LGTM

After discussion of some points that could affect this PR I think we are good to go.

Issue #1248, opened 2022-02-16
Multi-country targeting for Google Ads campaign


eason9487:
### Objective 

Enable merchants to target multiple countries in a single Google Ads campaign, reducing management overhead of campaign/budget and increasing the ability of campaign algorithm to scale.

### Overview

Change the functionality of campaign creation and editing to make it be able to specify multiple countries in a single Google Ads campaign.

![image](https://user-images.githubusercontent.com/17420811/154234977-da9b289f-6bb2-4393-a2c0-f5172b891da1.png)

### The current state of the system

Currently, it only allows merchants to select one country when creating a campaign. It also results in a very inefficient campaign setup experience, especially if the store ships to all 22 EU countries or worldwide.

![image](https://user-images.githubusercontent.com/17420811/154233583-e518082e-87ed-4643-b86b-fb9fc53effd4.png)

## Terminology

💡  Scope: this project and all related issues/pull requests

- The "old campaign(s)" refers to the existing campaigns created **before** releasing this feature.
- The "new campaign(s)" refers to the campaigns created **after** this feature is finished.

## Technical

💡  We will create a shared branch `feature/1248-multi-country-targeting-campaign`. All relevant developments should be branched from it and merged back to it. And open a PR to merge the shared branch to `develop` once the overall feature gets ready to release.

💡  The old campaigns will still be presented in a single-selected country UI when editing. The rest of relevant pages will be presented on a multi-country basis.

### Frontend

The UI changes should be applied to campaign creation and edit pages. All frontend routings will stay with the current.

#### Phase 1

- 📌  **Move the free ads credit box to step 1 of ads account setup.**
    - Implemented in #1253
    - Before
    ![image](https://user-images.githubusercontent.com/17420811/154430320-b99d4db2-ed72-4c57-88bb-5f07cf3e9c62.png)
    - After
    ![image](https://user-images.githubusercontent.com/17420811/154430442-0273dd81-55dc-4bfb-a455-eb1f226ec56a.png)
    - This box will only be displayed in step 1 of ads setup flow. It won't be shown on the single-page campaign creation/edit page. (Confirmed with Jill)
    - :warning: The copies have been changed after #1217 and the updated copies should be the latest. So the copies in the box will be different between implementation and visual design.
- 📌  **Separate edit UI for both old and new campaigns.**
    - Implemented in #1268
    - Edit an old campaign
[...]

PR #1327, opened 2022-03-14
Update create campaigns API with multiple targeted locations

ianlin:

Part of #1248, this PR extends the `POST ads/campaigns` API so the client can provide `targeted_locations` in the request payload to target a campaign in multiple countries.

Some changes that are worth mentioning are:

- Previously the `POST ads/campaigns` API accepts a `country` parameter in the request payload and set it in [Campaign.ShoppingSetting](https://developers.google.com/google-ads/api/reference/rpc/v9/Campaign.ShoppingSetting)'s `sales_country` field. The new API remove the `country` parameter in the request payload and set the **_the country of store's address_** as the country of sale. Ref: `pcTzPl-bo-p2#comment-502`.
- When creating campaign criterion for a campaign in `src/API/Google/AdsCampaignCriterion.php`, I originally added `resource_name` in the create operation like what `AdsGroup.php` did with a unique temporary ID, but the Google Ads API would return an error: `"INCONSISTENT_FIELD_VALUES": "The field's contents don't match another field that represents the same data."`. The error disappeared after removing `resource_name` for a campaign criterion and looks like there were no problems for creating multiple campaigns with multiple campaign criteria. I'm unsure why this happened but found that the [campaign criterion API doesn't have `resource_name` field](https://developers.google.com/google-ads/api/fields/v9/campaign_criterion
) while `AdsGroup` or `CampaignBudget` has `resource_name`.

#### Create a new campaign

Screenshot 2022-03-14 at 17 53 46

#### The newly created campaign is included in `GET ads/campaigns` API

Screenshot 2022-03-14 at 17 58 24

#### The newly created campaign is returned in `GET ads/campaigns/:campaign-id` API

Screenshot 2022-03-14 at 17 58 58

#### The newly created campaign is displayed in Google Ads dashboard

Screenshot 2022-03-14 at 17 57 28

1. Create a campaign by requesting to `POST ads/campaigns` using Postman like the screenshots in the `Screenshots` section. Remember to add your `wp-cookie` and `x-wp-nonce` in the request headers. 
    - To get the value of `wp-cookie` and `x-wp-nonce`, go to `Marketing -> Google Listings & Ads`, open Chrome dev tool's Network tab, clicks any requests to WP, e.g. `wc/gla/ads/campaigns`, you can get those values in the request headers.
2. Request to `GET ads/campaigns` and the newly created campaign should be included in the response data.
3. Get the campaign ID from the above response, and request to `GET ads/campaigns/:campaign-id`, the newly created campaign should be returned.
4. Go to Google Ads dashboard and check if the newly campaign exists with correct targeted locations.

### Changelog entry

> Update - Extend create campaigns API to accept multiple targeted locations


src/API/Google/AdsCampaignCriterion.php
```diff
@@ -0,0 +1,65 @@
...
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Google
+ */
+class AdsCampaignCriterion {
+
+	use ApiExceptionTrait;
+
+	/**
+	 * Returns a set of operations to create multiple campaign criteria.
+	 *
+	 * @param int    $campaign_id            Campaign ID.
```
mikkamp: This parameter is passed down to create_operation, but it's not used there. We only need the campaign resource name to link it to the campaign. So we can remove this parameter from both functions.

src/API/Google/AdsCampaignCriterion.php
```diff
@@ -0,0 +1,65 @@
...
+	 * @param int    $location_id            Targeted location ID.
+	 *
+	 * @return MutateOperation
+	 */
+	protected function create_operation( int $campaign_id, string $campaign_resource_name, int $location_id ): MutateOperation {
+		$campaign_criterion = new CampaignCriterion(
+			[
+				'campaign' => $campaign_resource_name,
+				'negative' => false,
+				'status'   => CampaignCriterionStatus::number( 'enabled' ),
```
mikkamp: A class for mapping the status is only useful if we are going to be using the status in our API responses, where we map them to lowercase strings. Which is the case for Campaigns where we can enable/pause them. However if we are just going to set the status internally then we can just use the Google class directly (so no need to create the class CampaignCriterionStatus), you can see an example of this when an ad group is created in `ad_group_create_operation`.

src/API/Google/AdsCampaign.php
```diff
@@ -176,21 +185,35 @@ public function get_campaign( int $id ): array {
 	 */
 	public function create_campaign( array $params ): array {
 		try {
+			$base_country = $this->container->get( WC::class )->get_base_country();
+			$location_ids = array_map(
+				function ( $country_code ) {
+					return $this->google_helper->find_country_id_by_code( $country_code );
```
mikkamp, line 192: We might need some extra validation here as I can produce a fatal by setting the targeted countries to `KP`. It's a valid country code so it passes the controller validation. But it's not on our supported countries list so it returns a null here.

We could consider adding some stricter validation in the controller, maybe we can swap out `$this->get_country_code_validate_callback()` for a stricter version as we need both a valid country code and a supported country code.

tests/Unit/API/Google/AdsCampaignTest.php
```diff
@@ -300,6 +312,45 @@ public function test_create_campaign_exception() {
 		}
 	}
 
+	public function test_create_campaign_exception_invalid_location_id() {
```
mikkamp, line 340: Missing a test either here or in the control, for a valid country code but not on our supported list. Ex: `KP`

Also I'm not sure if this test would have been better placed in AdsCampaignCriterionTest.php. What this test is really doing is passing `$location_ids  = [ null ]` after the mapping. But since we have a mocked criterion class that's not a problem. If we do that in the actual class it calls `create_operation` with null which gives a fatal because it expects an int.

mikkamp (commented):
The endpoint is working as expected when I pass the right values. Although as mentioned in the comments I still found a case where a fatal occurs, which wasn't caught by the tests.

src/API/Google/AdsCampaignCriterion.php
ianlin: Nice catch, thanks! It was originally used for generating campaign criterion's resource name, forgot to remove it after I change to not adding a resource name for a campaign criterion. Updated in 0f661f6.

src/API/Google/AdsCampaignCriterion.php
ianlin: Learnt, thanks! Updated in 50bd307.

tests/Unit/API/Google/AdsCampaignTest.php
ianlin: > Missing a test either here or in the control, for a valid country code but not on our supported list. Ex: KP

I added the test of unsupported countries in the controller via e790283, thanks!

> Also I'm not sure if this test would have been better placed in AdsCampaignCriterionTest.php. 

This test is the case when we send invalid location IDs to Google Ads API via `mutate` method so I generated a mutate mock exception to mimic the situation. So I think it would be good to place it in `AdsCampaignTest`.

src/API/Google/AdsCampaign.php
ianlin: Thanks for catching this critical error! I updated in e790283 but not 100% satisfy with the solution, because there will be an unused `$this->google_helper` inside `CampaignController`. Would like to know if there's any better solution when a trait needs to call a service's method?

ianlin:
Hi @mikkamp, I updated the codes based on your suggestions and is ready for another review, thanks!

src/API/Google/AdsCampaign.php
mikkamp: I agree that it's probably not the best pattern to use. GoogleHelper used to be a trait as well, and for a similar reason it got converted to a Service. Also the way you are doing it now means if any controller class has google_helper it will do a stricter validation, otherwise not. Most likely the stricter check is useful for other controllers as well, but we'd have to check that in each controller.
My suggestion was more along the lines of swapping out `get_country_code_validate_callback` to a local function within the controller which uses both the function in the trait and the google_helper function.

Another option if we want to make the validation reusable would be to add a GoogleHelperAwareInterface (and related trait) like is done for the ISO service. I'd still create a separate function like `get_supported_country_code_validate_callback` so the controller can decide which one to use (and we remain compatible with what we used to have).

tests/Unit/API/Google/AdsCampaignTest.php
mikkamp: Thanks makes sense, I agree it belongs more to the situation where we call the mutate. I just found it a shame that the tests didn't catch the error. Which made me wonder who has the responsibility of validating that location_ids is an array of only int's. And whether we should be enforcing that internally, or if we are content to rely on the validation in the controller.

src/API/Google/AdsCampaign.php
mikkamp: Another thought was, since there are no other controllers that enforce only supported countries then we could decide to follow that same pattern. The UI already has a list of supported countries to work with, so generally it won't send an API request with unsupported countries.

Although we would still need to fix AdsCampaign to make sure it filters out unsupported countries (or null values when converting to ID which is equal to an unsupported country). Just calling `array_filter` after mapping might be sufficient enough. That would be a much simpler approach. Yes we would have a less strict controller, but as long as it's not possible to produce a fatal it is a similar pattern we have for other controllers.

I would lean towards this solution and then create a followup issue to make some of our controllers stricter to enforce supported countries.

mikkamp (commented):
Thanks @ianlin for the changes, I'm no longer able to reproduce the fatal. 

The more I think about it, I'm leaning towards making the controller less strict since the UI already has a set of supported countries it limits to. I think we can go for the simpler option and create a follow up issue so we don't block this PR further, see [comment](https://github.com/woocommerce/google-listings-and-ads/pull/1327#discussion_r827788235)
Might just need some slight refactoring of the tests to make sure we correctly filter out unsupported values.

tests/Unit/API/Google/AdsCampaignCriterionTest.php
```diff
@@ -0,0 +1,62 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Google;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\AdsCampaignCriterion;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Tools\HelperTrait\GoogleAdsClientTrait;
+use Google\Ads\GoogleAds\V9\Enums\CampaignCriterionStatusEnum\CampaignCriterionStatus;
+use PHPUnit\Framework\MockObject\MockObject;
```
mikkamp: Just a small note: This test isn't mocking anything so this import isn't used.

src/API/Google/AdsCampaign.php
ianlin: I finished implementing `GoogleHelperAwareInterface` via 0ffa8dd before seeing your last comment.. 😅 I think if you think the `GoogleHelperAwareInterface` implementation is OK we can go with this solution, but I'll still create a followup issue to make sure other controllers to enforce supported countries if they need to.

Also, I added an `array_filter` in `AdsCampaign.php` via 1f139d6 to make sure we don't pass location IDs array with null value to `AdsCampaignCriterion`'s `create_operation` method.

Thanks for all the thorough thoughts! 🙏

tests/Unit/API/Google/AdsCampaignTest.php
ianlin: Thanks a lot! I added an array_filter in AdsCampaign.php via https://github.com/woocommerce/google-listings-and-ads/commit/1f139d64b917c5e2088e46864a3f9a820a8f94e5 to make sure we don't pass location IDs array with null value to AdsCampaignCriterion's create_operation method.

tests/Unit/API/Google/AdsCampaignCriterionTest.php
ianlin: Thanks! Added in 2df522e.

ianlin:
@mikkamp Thanks for all the useful suggestions, really useful and learnt a lot as always. 😄 The PR is ready for another round of review.

I've updated the codes based on your suggestions, but I implemented `GoogleHelperAwareInterface` rather than the simpler version as I saw your last comment after I finished implementing it. Please check this https://github.com/woocommerce/google-listings-and-ads/pull/1327#discussion_r828058503 for more info.

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
```diff
@@ -147,14 +156,55 @@ public function test_create_campaign_invalid_country_code() {
...
 		$this->assertEquals( 'rest_invalid_param', $response->get_data()['code'] );
-		$this->assertEquals( 'Invalid parameter(s): country', $response->get_data()['message'] );
+		$this->assertEquals( 'Invalid parameter(s): targeted_locations', $response->get_data()['message'] );
+		$this->assertEquals( 400, $response->get_status() );
+	}
+
+	public function test_create_campaign_unsupported_country() {
+		$this->country_supported = false;
+		$this->google_helper->method( 'is_country_supported' )
+			->willReturnCallback( function () { return $this->country_supported; } );
```
mikkamp: I believe the intention was to return the value of the property in the callback, so we only need to set the callback once in setup.

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
```diff
@@ -147,14 +156,55 @@ public function test_create_campaign_invalid_country_code() {
...
+		];
+
+		$response = $this->do_request( self::ROUTE_CAMPAIGNS, 'POST', $campaign_data );
+
+		$this->assertEquals( 'rest_invalid_param', $response->get_data()['code'] );
+		$this->assertEquals( 'Invalid parameter(s): targeted_locations', $response->get_data()['message'] );
 		$this->assertEquals( 400, $response->get_status() );
 	}
 
+	public function test_create_campaign_empty_targeted_locations() {
```
mikkamp, line 178: From my understanding of the design we should always be passing at least one targeted country when we create a new campaign, but maybe I'm seeing that wrong. Which means this test should fail with a validation error.

Normally we could use `'minItems' => 1` in the schema, but since we use a custom validation function that would have to be evaluated within the validation function. The default validation gets the schema from the request, [see](https://developer.wordpress.org/reference/functions/rest_validate_request_arg/). But it might make sense to always require at least one country if it's an array, so we might not need to access the schema.

src/API/Google/AdsCampaign.php
```diff
@@ -176,21 +185,42 @@ public function get_campaign( int $id ): array {
...
+				},
+				$params['targeted_locations']
+			);
+
+			$location_ids = array_filter(
+				$location_ids,
+				function ( $location_id ) {
+					return isset( $location_id );
+				}
+			);
```
mikkamp: For `array_filter` it mentions:
> If no callback is supplied, all empty entries of array will be removed.

```suggestion
			$location_ids = array_filter( $location_ids );
```
We'd be filtering out more than just null but that shouldn't be harmful.

mikkamp (approved):
Thanks for the changes, the solution looks good (so a followup issue for other controllers sounds good).

I get the expected test results, although I'm a bit curious about whether targeted locations can ever be empty, but besides that it looks pretty much good to go.

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
ianlin: You're right, updated in 73b1c41, thanks!

src/API/Google/AdsCampaign.php
ianlin: Nice, updated in be14290, thanks!

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
ianlin: I think it makes sense, added empty check in the country code validation function via c9ea09e. Also tested on other controllers that use the same validator and seems fine. But should keep an eye on them if it causes problem.

PR #1328, opened 2022-03-15
Add/more tags

puntope:

This PR adds the "X More Tags / Show less" functionality in the Tree Select Control Component

More info -> pcTzPl-fg-p2#comment-725

In brief, we want a parameter to set the maximum number of tags visible in the input, together with a button saying "x more tags" in case there are more tags than that maximum. When we click on that button, all the tags are shown. 

### Screenshots:

https://user-images.githubusercontent.com/5908855/158345132-4fb35b28-f9e9-4e1e-9bb7-73e130e04cea.mov

Notice the default limit is set as 5 tags.

1. Checkout this branch
3. By default you can see the only Tag selected "Spain" without any button
4. Add countries until 5. Still no button
5. Add one country more and see "+ 1 More"
6. Add another country and see "+ 2 More"
7.  Click on "+ 2 More" and see the 2 tags and "Show less"
8.  Click on "Show less" and you will see again 5 tags and "+ 2 More"

### DIFF
--- a/js/src/components/tree-select-control/tags.js
+++ b/js/src/components/tree-select-control/tags.js
@@ -3,6 +3,8 @@
  */
 import { __, sprintf } from '@wordpress/i18n';
 import { Tag } from '@woocommerce/components';
+import { useState } from '@wordpress/element';
+import { Button } from '@wordpress/components';
 
 /**
  * A list of tags to display selected items.
@@ -11,8 +13,19 @@ import { Tag } from '@woocommerce/components';
  * @param {Object[]} props.tags The tags
  * @param {Function} props.onChange The method called when a tag is removed
  * @param {boolean} props.disabled True if the plugin is disabled
+ * @param {number} props.maxVisibleTags The maximum number of tags to show. 0 or less than 0 evaluates to "Show All".
  */
-const Tags = ( { tags, disabled, onChange = () => {} } ) => {
+const Tags = ( {
+	tags = [],
+	disabled,
+	maxVisibleTags = 0,
+	onChange = () => {},
+} ) => {
+	const [ showAll, setShowAll ] = useState( false );
+	const maxTags = Math.max( 0, maxVisibleTags );
+	const shouldShowAll = showAll || ! maxTags;
+	const visibleTags = shouldShowAll ? tags : tags.slice( 0, maxTags );
+
 	if ( ! tags.length ) {
 		return null;
 	}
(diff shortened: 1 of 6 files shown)


js/src/components/tree-select-control/tags.js
```diff
@@ -11,8 +13,19 @@ import { Tag } from '@woocommerce/components';
  * @param {Object[]} props.tags The tags
  * @param {Function} props.onChange The method called when a tag is removed
  * @param {boolean} props.disabled True if the plugin is disabled
+ * @param {number} props.maxVisibleTags The maximum number of tags to show. 0 or less than 0 evaluates to "Show All".
  */
-const Tags = ( { tags, disabled, onChange = () => {} } ) => {
+const Tags = ( {
+	tags = [],
```
eason9487, line 19: 💅  It would be nicer to update its JSDoc along with the change of adding `= []`.

js/src/components/tree-select-control/tags.js
```diff
@@ -11,8 +13,19 @@ import { Tag } from '@woocommerce/components';
  * @param {Object[]} props.tags The tags
  * @param {Function} props.onChange The method called when a tag is removed
  * @param {boolean} props.disabled True if the plugin is disabled
+ * @param {number} props.maxVisibleTags The maximum number of tags to show. 0 or less than 0 evaluates to "Show All".
```
eason9487: 💅  Would it be better to indicate it's an optional prop with a default value?

js/src/components/tree-select-control/tags.js
```diff
@@ -57,6 +70,23 @@ const Tags = ( { tags, disabled, onChange = () => {} } ) => {
...
+					isTertiary
+					onClick={ () => {
+						setShowAll( ! showAll );
+					} }
+				>
+					{ showAll
+						? __( 'Show less', 'google-listing-and-ads' )
+						: sprintf(
+								// translators: %d: The number of extra tags to show
+								__( '+ %d More', 'google-listing-and-ads' ),
```
eason9487: 💅 
"More" seems to be lowercase.

eason9487 (approved):
Tested locally and works well. LGTM. Left some comments for minor things that might make this functionality be better.

---

💅 
It would be nicer to avoid a slight UI jumping when the "+1 more" / "show less" button appears and disappears.

Kapture 2022-03-15 at 18 45 00

---
💅 
When \<TreeSelectControl\> disabled, the "+1 more" / "show less" button is still available, which is expected behavior, but its style looks like disabled.

puntope:
❤️  Thanks for the review. I updated the code with your suggestions.

PR #1277, opened 2022-03-01
Add gTag info in the Product View page

songchen2022:
### Changes proposed in this Pull Request:

Add gTag info to the product view page
Closes # .

Add the product info in the gTag in product view event. This event measure when a user visits a product page.

### Screenshots:

Screen Shot 2022-03-01 at 9 03 08 AM

### Detailed test instructions:

1. Connect Merchant Center and Ads accounts.
2. Visit the public-facing store.
3.  View the page source code and confirm:
The gTag is preceded by the new HTML comment:
gtag("event", "view_item", {...})

Issue #148, opened 2021-01-26
Add complete setup UI / Modals


findingsimple:
Placeholder

PR #1329, opened 2022-03-15
Add Woo gTag remarketing and conversion signals 

songchen2022:
Add gTag info in home/product/cart/purchase/add_to_cart page

Closes # .

Add remarketing and conversion signals in the gTag in the following pages (home, product, add to cart action, cart, purchase) 
More details: https://docs.google.com/document/d/1UouhR5Hj-9gc5jPhxWPLX8-nXhoHG5jMeiva7iQrhy8/edit#

### Screenshots:

Screen Shot 2022-05-23 at 1 23 53 PM

Screen Shot 2022-05-24 at 1 10 15 PM

Screen Shot 2022-05-25 at 11 30 35 AM

Screen Shot 2022-05-24 at 1 09 25 PM

Screen Shot 2022-05-25 at 11 31 49 AM

1.  Connect Merchant Center and Ads accounts.
2.  Visit the public-facing store.
3.  View the page source code and confirm:

- In the each page
The gTag is preceded by the new HTML comment:
gtag("event", "page_view", {...})

- In the product page
The gTag is preceded by the new HTML comment:
gtag("event", "view_item", {...})	

- In the cart page
The gTag is preceded by the new HTML comment:
gtag("event", "page_view", {...})

- In the purchase page
The gTag is preceded by the new HTML comment:
gtag("event", "purchase", {...})


src/Google/GlobalSiteTag.php
```diff
@@ -66,18 +67,57 @@ public function register(): void {
...
 		add_action(
 			'wp_head',
-			function() use ( $ads_conversion_id, $ads_conversion_label ) {
+			function () use ( $ads_conversion_id, $ads_conversion_label ) {
 				$this->maybe_display_event_snippet( $ads_conversion_id, $ads_conversion_label );
 			},
 			1000000
 		);
+
+		add_action(
```
mikkamp, line 99: I think we should use different hooks for each event, see [here](https://github.com/woocommerce/google-listings-and-ads/pull/1277#discussion_r838272800). However if they are going to all be using the same hook we can just hook into it once and call all the snippets at once, and use the default priority.

src/Google/GlobalSiteTag.php
```diff
@@ -104,14 +144,13 @@ function( $gtag_snippet ) use ( $ads_conversion_id ) {
...
 
-
 	/**
 	 * Display the JavaScript code to load the Global Site Tag framework.
 	 *
 	 * @param string $ads_conversion_id Google Ads account conversion ID.
 	 */
 	protected function display_global_site_tag( string $ads_conversion_id ) {
-		// phpcs:disable WordPress.WP.EnqueuedResources.NonEnqueuedScript
+        // phpcs:disable WordPress.WP.EnqueuedResources.NonEnqueuedScript
```
mikkamp: This should be left indented with tabs, and the matching enable should also be indented with tabs.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
 			esc_js( $order->get_total() ),
 			esc_js( $order->get_currency() ),
 			esc_js( $order->get_id() ),
 		);
 	}
 
+	/**
+	 * Display the JavaScript code to track the product view page.
+	 */
+	public function display_view_item_event_snippet(): void {
```
mikkamp: We use a hook with an anonymous function to call these helper functions. That means that they can all be protected/private. Making them public indicates that they could be called from outside this class.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+                    "ecomm_pagetype": "purchase",
+                    "send_to": "GLA",
+                    transaction_id": "%s",
+                    "currency": "%s",
+                    "country": "%s",
+                    "value": "%s",
+                    "new_customer": "%s",
+                    "tax": "%s",
+                    "shipping": "%s",
+                    "delivery_posatal_code": "%s",
```
mikkamp: Looks like this is a small typo and should be `delivery_postal_code`. Is there any documentation which shows the fields, most of these don't seem to be included in the [documentation here](https://developers.google.com/analytics/devguides/collection/ga4/reference/events#purchase)

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			$language = 'English';
+		}
+		printf(
+			'<script>
+            gtag("event", "purchase",
+				{
+                    "developer_id.%s": "true",
+                    "ecomm_pagetype": "purchase",
+                    "send_to": "GLA",
+                    transaction_id": "%s",
```
mikkamp: Looks like this is missing a starting quote.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			);
+		}
+		$value = WC()->cart->total;
+		printf(
+			'<script>
+            gtag("event", "page_view",
+				{"send_to": "GLA",
+				"ecomm_pagetype": "cart",
+				"value": "%s",
+				 items: [' . esc_html( $item_info ) . ']}); </script>',
```
mikkamp: If this is already escaped JS then I don't think we should escape it again using `esc_html`.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+				}',
+				esc_js( $id ),
+				esc_js( $price ),
+				esc_js( $name ),
+				esc_js( $quantity )
+			);
+		}
+		$value = WC()->cart->total;
+		printf(
+			'<script>
```
mikkamp: We repeat printing script tags quite often. Since we now support a minimum of WordPress 5.7, why don't we use the function [wp_print_inline_script_tag](https://developer.wordpress.org/reference/functions/wp_print_inline_script_tag/) for this.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+				esc_js( $quantity ),
+			);
+
+		}
+
+		$is_new_customer = false;
+		if ( $order->get_user_id() ) {
+			$total_orders = wc_get_customer_order_count( $order->get_user_id() );
+		} else {
+			$total_orders = WC_Order_Export_Data_Extractor::get_customer_order_count_by_email( $order->get_billing_email() );
```
mikkamp: This is not a function that is available in WooCommerce, it seems to be from the plugin `woo-order-export-lite`.

Probably better to use `wc_get_orders` here setting the [customer](https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query#customer) and making sure to return only ids (otherwise it will load all the order objects).

I think we should extract this to a separate function which returns a boolean `$is_new_customer` so we can clearly define what we are doing within the function.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+				"value": "%s",
+				 items: [' . esc_html( $item_info ) . ']}); </script>',
+			esc_js( $value ),
+		);
+	}
+
+	/**
+	 * Display the JavaScript code to track the purchase page.
+	 */
+	public function display_purchase_page_snippet(): void {
```
mikkamp: Can we combine this with the `maybe_display_event_snippet` and give the function a name that reflects what it's tracking. Combining them means we don't have to run the same conditional checks and load the order twice.

Also the function for tracking the conversion has a safeguard to ensure it's only output once. If we refresh the order received page it won't repeat the conversion event, but this purchase event will be repeated every time the page is releoaded.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			$product_name = $item->get_name();
+			$quantity     = $item->get_quantity();
+			$price        = $item->get_subtotal();
+			$item_info    = $item_info . sprintf(
+				'{
+                    "id": "gla_%s",
+                    "price": %s,
+                    "google_business_vertical": "retail",
+                    "name":"%s",
+                    "quanitity":"%s",
```
mikkamp: Looks like a typo `quantity`

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+
+	/**
+	 * Display the JavaScript code to track the add to cart button.
+	 *
+	 * @param string $message Add to cart messages.
+	 * @param array  $products Product ID list.
+	 */
+	public function custom_action_add_to_cart( $message, $products ) {
+		// Only display this tag info after click the add to cart button .
+		$product = wc_get_product( array_key_first( $products ) );
```
mikkamp: If this filter supports multiple items, why do we only get the first item? The add_to_cart event supports an array of items.

src/Google/GlobalSiteTag.php
```diff
@@ -66,18 +67,57 @@ public function register(): void {
...
+		add_action(
+			'wp_body_open',
+			function () {
+				$this->display_purchase_page_snippet();
+			},
+			1000005
+		);
+
+		add_filter(
+			'wc_add_to_cart_message_html',
```
mikkamp: I'm not sure if this is the most reliable way to track adding an item to the cart. Our other tracking solutions detect this quite differently. This is also not compatible with the WooCommerce all product / cart blocks. I think we need to do some deeper investigation in how to detect this reliably.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+					esc_js( $product->get_name() ),
+					esc_js( join( '& ', $product->get_categories() ) ),
+				);
+			},
+			1000005
+		);
+
+		do_action(
+			'wp_body_open'
+		);
```
mikkamp: This is not the right place to call this action, as this causes the action to be called twice, and any function that hooks into this action will be called again here. If we want this JS to be output after this filter is called, it's probably better to hook it to `wp_footer`. Although we are probably going to need to adjust this anyways to make it compatible with blocks.

mikkamp (commented):
Thanks for submitting the initial code. Since this PR contains all the tracking code, including the code in PR #1277, the comments I left there apply for this PR as well.

As for the failing unit tests, you might want to rebase this PR off of the latest develop as those have been fixed there.

src/Google/GlobalSiteTag.php
```diff
@@ -156,14 +195,241 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+                    "tax": "%s",
+                    "shipping": "%s",
+                    "delivery_posatal_code": "%s",
+                    "aw_feed_country": "%s",   
+                    "aw_feed_language": "%s",                 
+                    items: [' . esc_html( $item_info ) . ']}); </script>',
+			esc_js( self::DEVELOPER_ID ),
+			esc_js( $order->get_id() ),
+			esc_js( $order->get_currency() ),
+			esc_js( WC()->countries->get_base_country() ),
```
mikkamp: Like the WP proxy class, we also have a WC proxy for getting the country.

src/Google/GlobalSiteTag.php
mikkamp: Also besides the potential of this generating a "heavy" database query to scan all orders, is there any particular need/benefit for tracking whether they are a first time customer?
This isn't data that pertains to the current purchase, but data about historic purchases.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for pointing this out. All these fields follow the [requirement](https://docs.google.com/presentation/d/1QRYvGK-ObTK30CcQVRSM6m0NnshFrkWPG3toxFISypI/edit?usp=sharing&resourcekey=0-3frjhBOKobpte7zH6gX4oQ) and [Design doc](https://docs.google.com/document/d/1UouhR5Hj-9gc5jPhxWPLX8-nXhoHG5jMeiva7iQrhy8/edit#)

src/Google/GlobalSiteTag.php
songchen2022: It makes sense. Thanks. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Great idea! We don't need all the data information of the order. We only want to count the number of the previous orders to check if this is first time customer. Returning ids should be enough.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: In my understanding, customer could add single product or multiple same products with clicking the "add to cart" button on one product page of current template. We could get all the required data from one single product. Therefore I only read the first product info here. Please correct me if I'm wrong.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: I couldn't get idea about this comment. Could I get more hint about this suggestion? Thanks.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions.  However, I think it probably couldn't be combined into one function. `maybe_display_event_snippet` is hooked with `wp_head`, but `display_purchase_page_snippet` is hooked with `wp_body_open`. Please correct me if I'm wrong.

mikkamp: Yes that's correct, they are hooked with two different hooks, but there is no reason they can't use the same one.

In fact I would actually suggest to hook into `woocommerce_before_thankyou` which will already pass the `order_id` so we don't need to retrieve it from the query. I also noticed another flaw, if we look at the WooCommerce thank you (order received) page, it also checks the following condition:

```php
$order->has_status( 'failed' )
```

In both the purchase and conversion events we don't check this condition. So if the payment was denied the order will still be tracked as a conversion/purchase.

src/Google/GlobalSiteTag.php
mikkamp: In most cases that is correct. A [grouped product](https://github.com/woocommerce/woocommerce/blob/6.3.1/plugins/woocommerce/includes/class-wc-form-handler.php#L838) is an exception though, it will pass an array of products to that function.

Although I'm not sure how relevant this is if the solution is not compatible with blocks. Have you done some testing to see if this code is triggered when using the All Products block?

src/Google/GlobalSiteTag.php
mikkamp: In the constructor of GlobalSiteTag we pass an instance of [`WP`](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.1/src/Proxies/WP.php#L23), we also have a proxy class for [`WC`](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.1/src/Proxies/WC.php#L22)

So if we pass the class in the constructor then we can get the base country like is done in [other classes](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.1/src/MerchantCenter/TargetAudience.php#L77)
```php
$shop_country = $this->wc->get_base_country();
```

Also if we implement the different hook for the conversion as suggested [here](https://github.com/woocommerce/google-listings-and-ads/pull/1329#discussion_r842434349), then we no longer use the WP proxy class, so we can replace it with WC.

songchen2022:
#1404

src/Google/GlobalSiteTag.php
songchen2022: Thanks for the suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated to "woocommerce_add_to_cart_button_html". Not sure if this is compatible with the Woocommerce all product/cart product. If it is not, could I get some of your suggestions? Thanks.

src/Google/GlobalSiteTag.php
songchen2022: It makes sense. Thanks. Updated.

src/Google/GlobalSiteTag.php
mikkamp: The problem with using a hook like `woocommerce_add_to_cart_button_html` is that it's going to be called every time an add to cart button is output on the page. So for a shop page that could be multiple times:
image

Also I'm not quite sure if `woocommerce_add_to_cart_button_html` is even used, looking at the template files I see  `woocommerce_before_add_to_cart_button` but I don't see the button html being filtered.

In the function `custom_action_add_to_cart` the event is output directly to the page without any condition so all the events will be triggered as soon as the page is finished loading and not only when the button is clicked. As explained [here](https://support.google.com/analytics/answer/11147304) it should be attached to some kind of event listener.

In our Google Analytics integration extension we do this through jQuery as can be seen [here](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/1.5.8/includes/class-wc-google-gtag-js.php#L312-L316). I don't think we need to necessarily rely on jQuery and can use vanilla JavaScript, however we do need to take into consideration that the Add to Cart button can be added both on single pages, archive pages, and through various shortcodes and blocks.

The blocks are the most problematic because they have their own [tracking implementation](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/extensions/google-analytics/index.ts) which is done through JavaScript and doesn't get triggered through PHP hooks.
I don't have a solution at the moment how we can support those, that can be a project in itself and is going to require a deep dive into coming up with a compatible solution for all plugins to work in harmony.

src/Google/GlobalSiteTag.php
```diff
@@ -156,12 +192,222 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+						esc_js( $product->get_name() ),
+						esc_js( join( '& ', $this->product_helper->get_categories( $product ) ) ),
+					);
+					wp_print_inline_script_tag( $add_to_cart_gtag );
+				}
+			);
+
+			do_action(
+				'wp_footer'
+			);
```
mikkamp: We shouldn't be doing any `do_action` call here. The code we are wanting to output on the page is hooked into `wp_footer` through the calls to `add_action` but we want to let the theme handle calling `do_action( 'wp_footer' );`. So we should completely remove this code here.

src/Google/GlobalSiteTag.php
```diff
@@ -172,4 +418,21 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+	 */
+	private static function is_first_time_customer( $customer_email ): bool {
+		$query = new \WC_Order_Query(
+			[
+				'return' => 'ids',
+			]
+		);
+		$query->set( 'customer', $customer_email );
+		$orders = $query->get_orders();
+		return count( $orders ) === 1 ? 'true' : 'false';
```
mikkamp: Since we are only interested in knowing if there is at least 1 order with the matching email. We can limit the query to only one result, so it can stop scanning the DB table after finding the first match. For large sites searching customer orders can potentially be slow.

src/Google/GlobalSiteTag.php
```diff
@@ -133,18 +173,14 @@ function gtag(){dataLayer.push(arguments);}
 	 *
 	 * @param string $ads_conversion_id Google Ads account conversion ID.
 	 * @param string $ads_conversion_label Google Ads conversion label.
+	 * @param int    $order_id The order id.
 	 */
-	public function maybe_display_event_snippet( string $ads_conversion_id, string $ads_conversion_label ): void {
+	public function maybe_display_event_snippet( string $ads_conversion_id, string $ads_conversion_label, int $order_id ): void {
```
mikkamp: Since we've added multiple event snippets I don't think this function name/purpose is very clear anymore. Can we rename this function to describe which event snippet(s) it's displaying.

Also since we now have a group name 'GLA' is it still relevant to pass the conversion ID and label for tracking? Or can we just send the event to the same group as the other events?

src/Google/GlobalSiteTag.php
```diff
@@ -156,12 +192,222 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
 		$order->save_meta_data();
 
 		printf(
-			'<script>gtag("event", "conversion", {"send_to": "%s","value": "%s","currency": "%s","transaction_id": "%s"});</script>',
+			'<script>
```
mikkamp: Should we also use `wp_print_inline_script_tag` here?

src/Google/GlobalSiteTag.php
songchen2022: Done.

src/Google/GlobalSiteTag.php
songchen2022: Great suggestion! I updated it to 2 here.

src/Google/GlobalSiteTag.php
songchen2022: Updated the name. 'GLA' group is relevant to passing the conversion ID but I'm not sure if it is same as $ads_conversion_id/$ads_conversion_label. Could I know how to make sure if they are the same group?

src/Google/GlobalSiteTag.php
songchen2022: Done.

src/Google/GlobalSiteTag.php
songchen2022: That makes sense. Could I keep using ''wc_add_to_cart_message_html" here and add "TODO" for further investigation?

mikkamp: I think that would be the best approach for now.

src/Google/GlobalSiteTag.php
mikkamp: In [this document](https://support.google.com/google-ads/answer/9028614?hl=en#zippy=%2Cinstall-with-global-site-tag) it seems to indicate that the purchase event should also be sent with `conversion_id/label`. Which is why I asked if they should all have a matching `send_to`. I wasn't able to find any clear details in the docs though. Is there someone from the Google Ads team that could clarify this?

songchen2022: Thanks for pointing this out. After checked [this doc](https://docs.google.com/presentation/d/1JVJ7wsA22nwB-YWOJGAqxcouIDsv7D9cPj6UjXK0eBI/edit?userstoinvite=chennyx%40google.com&resourcekey=0-A2hcqh8CwxpRGQGQLPVk6A#slide=id.g10e1beef46c_0_373), it should use $ads_conversion_id/$ads_conversion_label here as you said. Updated.

songchen2022:
All the comments have been resolved. Could you please help to review it again? Thanks!

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +191,223 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+		);
+		wp_print_inline_script_tag( $view_item_gtag );
+	}
+
+	/**
+	 * Display the JavaScript code to track all pages.
+	 */
+	private function display_page_view_event_snippet(): void {
+		$page_view_gtag =
+		sprintf(
```
mikkamp: Normally we would place this on one line (the indentation for the next lines is fine).

```suggestion
		$page_view_gtag = sprintf(
```

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +191,223 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			'gtag(
+                    "event", "page_view", {
+                    "send_to": "GLA",
+                    "developer_id.%s": "true",});',
+			esc_js( self::DEVELOPER_ID )
+		);
+		wp_print_inline_script_tag( $page_view_gtag );
```
mikkamp: I was testing this page view event and when I view the frontend page I see a double request being sent to an URL like (the query parameters show it's a page_view event):

```
https://www.google.com/pagead/1p-user-list/405088140/?random=1651651983812&cv=9&fst=1651651200000&num=1&bg=ffffff&guid=ON&u_h=1200&u_w=1920&u_ah=1173&u_aw=1920&u_cd=24&u_his=4&u_tz=60&u_java=false&u_nplug=5&u_nmime=2&gtm=2oa520&sendb=1&data=event%3Dpage_view&frm=0&url=https%3A%2F%2Fwoo-google.test%2F&ref=https%3A%2F%2Fwoo-google.test%2Fshop%2F&tiba=woo-google%20%E2%80%93%20Just%20another%20WordPress%20site&async=1&fmt=3&is_vtc=1&random=569364553&resp=GooglemKTybQhCsO&rmt_tld=0&ipr=y
```

On the homepage I actually see 3 of them being sent. Based on the documentation [here](https://support.google.com/analytics/answer/11403294?hl=en) it seems that the `page_view` event is automatically sent unless `send_page_view` is set to false. Is it preferred to have them automatically sent or use a snippet?

When I use it combination with the Google Analytics Integration extension I can see that `send_page_view` is set to false, but for a different send to property so it still seems to be sent multiple times to the group GLA.

src/Google/GlobalSiteTag.php
```diff
@@ -66,18 +83,42 @@ public function register(): void {
...
+			function () {
+				$this->display_view_item_event_snippet();
+			}
+		);
+
+		add_action(
+			'wp_body_open',
+			function () {
+				$this->display_page_view_event_snippet();
+				$this->display_cart_page_snippet();
```
mikkamp: When I view the cart page both of these snippets get added. Since they are both for the event `page_view`, should only one be displayed?

Or should we have a conditional check within `display_page_view_event_snippet` and check `is_cart` to determine if we should call `display_cart_page_snippet`?

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +191,223 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+                    "developer_id.%s": "true",
+                    "ecomm_pagetype": "product",
+                    "value": "%s",
+                    items:[{
+                    "id": "gla_%s",
+                    "price": %s,
+                    "google_business_vertical": "retail",
+                    "name":"%s",
+                    "category":"%s",
+                    }]});',
```
mikkamp: When I view the data which gets sent in the URL it shows:
```
data: event=view_item;ecomm_pagetype=product;google_business_vertical=retail;id=gla_1001
```

I don't see it passing along the item name and category there, based on the documentation [here](https://developers.google.com/analytics/devguides/collection/ga4/reference/events#view_item) it shows that they should be `item_name` and `item_category`. In the [PRD](https://docs.google.com/document/d/1UouhR5Hj-9gc5jPhxWPLX8-nXhoHG5jMeiva7iQrhy8/) it refers to both `name` and `item_name` so it's not clear which one it should be.

Also with the mixed indentation and usage of quotes it makes it a little harder to read. For example `items` doesn't have double quotes around it, while the rest do. Adding the double quotes makes it look like it's in JSON format, since it's JS the double quotes aren't required (which is how it's in most of the examples).

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +191,223 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			esc_js( self::DEVELOPER_ID ),
+			esc_js( (string) $product->get_price() ),
+			esc_js( $product->get_id() ),
+			esc_js( (string) $product->get_price() ),
```
mikkamp: This field we are type casting to a string (similar to the value field), but the get_price function already returns a string, so the typecast doesn't change anything. Unlike the value field we aren't enclosing it in double quotes though, so that seems like we are expecting it to be a float or double. Should it be converted/cast to one of types instead?
The documentation [here](https://developers.google.com/analytics/devguides/collection/ga4/reference/events#view_item) suggest both fields should be a number and not a string.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +191,223 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			esc_js( $product->get_name() ),
+			esc_js( join( '& ', $this->product_helper->get_categories( $product ) ) ),
```
mikkamp: With multiple categories this shows up like:

```
"category":"one&amp; two",
```

Should that be using `' & '` as separator instead? There is another location where this same separator is being used.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +191,223 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+				esc_js( $quantity )
+			);
+		}
+		$value          = WC()->cart->total;
+		$page_view_gtag = sprintf(
+			'gtag("event", "page_view",
+				{"send_to": "GLA",
+				"ecomm_pagetype": "cart",
+				"value": "%s",
+				 items: [' . $item_info . ']});',
```
mikkamp: Since we are using sprintf here we can just replace this with a `%s` and add `$item_info` to the list of variables. I guess it's a matter of personal preference but I'd suggest to avoid mixing different methods to replace variables.

Personally I find this the clearest so you don't have to look which %s matches which variable, but sprintf is perfectly fine as well:

```php
		$page_view_gtag = "
			gtag(
				'event',
				'page_view',
				{
					send_to: 'GLA',
					ecomm_pagetype: 'cart',
					value: {$value},
					items: [
						{$item_info}
					]
				}
			);
		";
```

src/Google/GlobalSiteTag.php
```diff
@@ -172,4 +418,22 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+	private static function is_first_time_customer( $customer_email ): bool {
+		$query = new \WC_Order_Query(
+			[
+				'limit'  => 2,
+				'return' => 'ids',
+			]
+		);
+		$query->set( 'customer', $customer_email );
+		$orders = $query->get_orders();
+		return count( $orders ) === 1 ? 'true' : 'false';
```
mikkamp: The return type for this function is `bool` so best to return it as `true` or `false` without quotes.

src/Google/GlobalSiteTag.php
```diff
@@ -66,18 +83,42 @@ public function register(): void {
...
+		add_action(
+			'wp_body_open',
+			function () {
+				$this->display_page_view_event_snippet();
+				$this->display_cart_page_snippet();
+			}
+		);
+
+		add_filter(
+			'wc_add_to_cart_message_html',
```
mikkamp: This hook works when we add a product to the cart from a single product page. Both when we have redirect to cart enable and when it's disabled. But when we have redirect to cart disabled and Ajax add to cart enabled, we don't get this notification, so the add to cart event is never tracked.

image

This is the default configuration for WooCommerce so there are quite a few stores who will have it set like this, so I think we should consider adding the tracking based on when the button is clicked, like is done in the [Google Analytics Integration](https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/1.5.8/includes/class-wc-google-analytics-js.php#L509-L515) (although that would need to be converted from jQuery to regular JavaScript).

mikkamp (commented):
Thanks for all the changes, I left some additional comments as I went through the code.

I was testing the snippets in the browser and can view the events getting sent to a URL like `https://googleads.g.doubleclick.net/pagead/viewthroughconversion...` but I'm not seeing all the fields being passed in this request. Is there a way I can view reports about these events in my Google Ads account (to confirm it's arriving as expected)?

mikkamp:
Another question I wanted to ask about the implementation.

We currently track the event `add_to_cart`, but what about `remove_from_cart` and `change_cart_quantity`. The first event doesn't seem complete without tracking the latter two, as it could give an inaccurate picture of what was added to the cart.

src/Google/GlobalSiteTag.php
songchen2022: Great suggestions. I will check with the team if we should show cart_view event or should we add a conditional check within display_page_view_event_snippet.

src/Google/GlobalSiteTag.php
songchen2022: Thanks so much for getting back to me. I saw you're working on this [RP](https://github.com/woocommerce/google-listings-and-ads/pull/1488) Will this PR help track the "add to cart" action on single product page and multiple product page with "add to cart" buttons?

songchen2022:
> Another question I wanted to ask about the implementation.
> 
> We currently track the event `add_to_cart`, but what about `remove_from_cart` and `change_cart_quantity`. The first event doesn't seem complete without tracking the latter two, as it could give an inaccurate picture of what was added to the cart.

--> This is a good question. For current design, we only track "add to cart" action in the product page now. Because I think we could get the information of changing cart. For `remove_from_cart`,  I checked it in the test store which can only be done in the cart page. After updating the quantity to zero, we have to click the "update cart" button which will refresh the cart page. The gtag will be injected after refreshing. For 'change_cart_quantity', this can be done in the product page and cart page. The action will be similar with 'remove_from_cart' in the cart page. In the product page, 'change_cart_quantity' could be done with clicking the "add to cart" button. The gtag will be injected by this action. Therefore, I think we still could get the accurate picture of the cart. Please correct me if I'm wrong.

src/Google/GlobalSiteTag.php
mikkamp: Right now the intention of that PR is to only work with the Gutenberg blocks. So it only works with a block version of an archive page which is created using the All Products block. It doesn't do anything for traditional archive pages, or single product pages.
However it might be a good idea to include that there as well. Since it loads a JS file which can react to specific events, it might be better to look for add_to_cart buttons on those pages and then track the event. Which is probably a better solution then injecting a script for each add_to_cart button. Let me know if you think that's a better approach and I can work on adding that to the PR.

What I do need clarified though, is why there is a difference between using `name` or `item_name` fields in the docs and PRD. It doesn't look like the item details get tracked correctly so I'll need some clarification on what fields need to get sent in the event.

src/Google/GlobalSiteTag.php
songchen2022: Done. Thanks for your suggestions!

src/Google/GlobalSiteTag.php
songchen2022: Done.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Added a conditional check within display_page_view_event_snippet.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your reply. It will be better to add tracking to each "add to cart" button. If it's fine to add the track event on the single product page, I will delete this gtag part here later.

 And I've updated the name in the list of items to item_name.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions. I set send_page_view to false now.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for pointing this out. I've update the `name` to `item_name` and `category` to `item_category`. And updated the indentation. Could I know if there is any indentation standard I could follow?

songchen2022:
> Thanks for all the changes, I left some additional comments as I went through the code.
> 
> I was testing the snippets in the browser and can view the events getting sent to a URL like `https://googleads.g.doubleclick.net/pagead/viewthroughconversion...` but I'm not seeing all the fields being passed in this request. Is there a way I can view reports about these events in my Google Ads account (to confirm it's arriving as expected)?

> Caio replied this question in the Questions Doc[https://docs.google.com/document/d/1nCBTDaTyr6YsckayH5J0vxQoeWYHiuKRs4mgCTjD7BI/edit#](url)

songchen2022:
Thanks for reviewing. I have replied all the comments, could you please help take a look again? Thanks!

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +198,215 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			}
+			$value          = WC()->cart->total;
+			$page_view_gtag = sprintf(
+				'gtag("event", "page_view", {
+				send_to: "GLA",
+				ecomm_pagetype: "cart",
+				value: "%s",
+				items: [%s]});',
+				esc_js( $value ),
+				esc_js( $item_info ),
```
mikkamp: The data in `$item_info` has already been escaped. Escaping it twice is causing newline characters to show up as `\n` which is causing console errors.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +198,215 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+
+		// Track the purchase page
+		$language = $this->wp->get_locale();
+		if ( 'en_US' === $language ) {
+			$language = 'English';
+		}
+		$purchase_page_gtag =
+		sprintf(
+			'gtag("event", "purchase", {
+			developer_id.%s: "true",
```
mikkamp: A parameter name that uses other characters like a `.` should be enclosed in quotes otherwise this generates a syntax error (several cases in the code where this parameter name is not enclosed in quotes).

Although my question is, if we need to send it with each event (it's not included in the purchase event). According to the documentation [here](https://developers.google.com/tag-platform/gtagjs/reference#set) it mentions that if it's set it will be sent for all subsequent events, so it seems unnecessary to include it again in each event.

src/Google/GlobalSiteTag.php
mikkamp: Thanks for the response. Although this doesn't seem to be fully working yet, since add to cart buttons can show up in many locations we need to make sure they are all tracked. The settings in WooCommerce > Settings > Products > Add to cart behaviour, influence when and how the add to cart message is shown.

These are the various scenarios:
- (Working) Archive pages that have been created with blocks (handled by PR #1488)
- (Working) Single product page add to cart button (page is refreshed so message is loaded)
- (Not working) Single product page upsells and related products (added to cart through ajax so message is not shown)
- (Not working) Archive pages that have been created as a regular shop page without blocks (added to cart through ajax so message is not shown)
image

If we enable the option "Redirect to the cart page after successful addition" then the add to cart message is shown however it hooks the snippet into `wp_footer` which is never called because of the immediate page redirect to the cart.

So again I think we should switch to JS code which hooks into the add_to_cart buttons and triggers when the buttons are clicked. Since there are so many different types of add_to_cart buttons across various pages, I suggest this is done in a JS file instead of outputting a snippet for each button, which would be a similar approach to what's being done in PR #1488 (so I suggest moving the add_to_cart handling to that PR).

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +198,215 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+				$id = $cart_item['product_id'];
+
+				// gets the product object
+				$product = $cart_item['data'];
+				$name    = $product->get_name();
+				$price   = $product->get_price();
+				// gets the cart item quantity
+				$quantity = $cart_item['quantity'];
+
+				$item_info = $item_info . sprintf(
```
mikkamp: If we want to concat as a string we can use `$item_info .= sprintf...`

Although in this case this concatenation fails when we have multiple items in the cart page. Because it adds the items in array like this:
```
[ { ... }{ ... } ]
```
Those objects should be separated by comma's if there are multiple items. My suggestion in cases like this is to make `$item_info` an array, and then use `join` to put them together, which will add the separator character as needed.

Example:
```
$item_info = [];
foreach ( WC()->cart->get_cart() as $cart_item ) {
	$item_info[] = sprintf( '{ ... }' );
}

$page_view_gtag = sprintf(
	'...items: [ %s ],...',
	join( ',', $item_info )
);
```

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +198,215 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+						esc_js( $product->get_id() ),
+						esc_js( (string) $product->get_price() ),
+						esc_js( $product->get_name() ),
+						esc_js( join( ' & ', $this->product_helper->get_categories( $product ) ) ),
+					);
+					wp_print_inline_script_tag( $add_to_cart_gtag );
+				}
+			);
+
+			return $message;
```
mikkamp: The return is done within the for loop, which means it will only track the first item added to cart.

src/Proxies/WP.php
```diff
@@ -32,7 +32,7 @@ class WP {
 	 */
 	public function __construct() {
 		global $wp;
-		$this->wp =& $wp;
+		$this->wp = & $wp;
```
mikkamp: Any particular reason this needs to be changed? It's assigning by reference so the preference is to write it as `=&`.
Was it maybe a global search and replace when adding a space before & for this line which changed it?
`esc_js( join( ' & ', $this->product_helper->get_categories( $product ) ) )`

mikkamp (commented):
Thanks for all the changes, I've added some additional comments.

Regarding the item fields we still need to wait on further clarification, because based on the documentation for [item parameters](https://support.google.com/google-ads/answer/7305793?hl=en#zippy=%2Cretail) it's showing only `id` and `google_business_vertical` are supported.

Also regarding the add_to_cart tracking I'm still thinking we should move that to #1488 to be able to track all the various add_to_cart buttons correctly.

findingsimple:
Hi @songchen2022 thanks for the work so far 👏 

I noticed we're picking up some basic console and syntax errors when reviewing your code (examples [here](https://github.com/woocommerce/google-listings-and-ads/pull/1329#discussion_r877947403) and [here](https://github.com/woocommerce/google-listings-and-ads/pull/1329#discussion_r877954525))

Before asking us to review could I please ask you to make sure you've tested your changes (made sure events are working etc,) and made sure you're catching these sorts of issues. 

It'll save time for everyone in the long run, I assume you've probably got better tooling/access to be able to test/view/catch events in Ads on your end so hopefully, it's not a difficult task. For us, we have to check everything as a normal user would so anything you can do before we get to it is a win I think.

src/Google/GlobalSiteTag.php
songchen2022: Moving the add_to_cart handling to your [PR](https://github.com/woocommerce/google-listings-and-ads/pull/148) sounds good to me. I'll remove this "add to cart" tracking part.

src/Google/GlobalSiteTag.php
songchen2022: Good catch. Updated.

src/Proxies/WP.php
songchen2022: After I run `./vendor/bin/phpcs`, it reminder me to fix this format by run PHPCBF. I did that which caused this error. Fixed.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

songchen2022:
> 

Thanks for your suggestions! Will follow that.

songchen2022:
Hi Michael, thanks for reviewing. I've replied the comments. Could you please help take a look again? Thanks!

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			items: [%s]});',
+			esc_js( "{$ads_conversion_id}/{$ads_conversion_label}" ),
+			esc_js( $order->get_id() ),
+			esc_js( $order->get_currency() ),
+			esc_js( $this->wc->get_base_country() ),
+			esc_js( $order->get_total() ),
+			esc_js( $is_new_customer ),
+			esc_js( $order->get_cart_tax() ),
+			esc_js( $order->get_total_shipping() ),
+			esc_js( $order->get_billing_postcode() ),
```
mikkamp, line 258: The billing postcode was added as a parameter here but there wasn't any %s added for it to replace, so now there is one parameter too many.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			delivery_postal_code: "%s",
+			aw_feed_country: "%s",
+			aw_feed_language: "%s",
+			items: [%s]});',
+			esc_js( "{$ads_conversion_id}/{$ads_conversion_label}" ),
+			esc_js( $order->get_id() ),
+			esc_js( $order->get_currency() ),
+			esc_js( $this->wc->get_base_country() ),
+			esc_js( $order->get_total() ),
+			esc_js( $is_new_customer ),
```
mikkamp: For reusability of the helper function I think it was much nicer for it to return a boolean. Having the values returned as strings is how we want to represent it in the snippet but not the result we want to return. I'd suggest to go back to the function `is_first_time_customer` returning a boolean and then replace this line with the output we want:

```suggestion
			$is_new_customer ? 'true' : 'false',
```

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+	}
+
+	/**
+	 * Display the JavaScript code to track all pages.
+	 */
+	private function display_page_view_event_snippet(): void {
+		$page_view_gtag = sprintf(
+			'gtag("event", "page_view", {
+			send_to: "GLA",});',
+		);
```
mikkamp: We don't need to use sprintf here since it's just a regular string, also the trailing comma after send_to looks out of place.
This code can also be within the if block since we don't need to define it with this value if we are on the cart page. Personally I'd also return within the if block so we don't need the else and can prevent an additional indent. Something like this:

```php
		if ( ! is_cart() ) {
			// Display the JavaScript code to track page views.
			wp_print_inline_script_tag(
				'gtag( "event", "page_view", { send_to: "GLA" } );'
			);
			return;
		}

		// Display the JavaScript code to track the cart page view.
		...
```

src/Google/GlobalSiteTag.php
```diff
@@ -172,4 +357,22 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
 	public static function is_needed(): bool {
 		return true;
 	}
+
+	/**
+	 * Check if it is the new customer order.
```
mikkamp: I think this function could benefit from a better description explaining why we are looking for 2 orders and not just one. Maybe something like:

```
* Check if the customer has previous orders.
* Called after order creation (check for older orders including the order which was just created).
```

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			items:[{
+				id: "gla_%s",
+				price: %s,
+				google_business_vertical: "retail",
+				name: "%s",
+				category: "%s",
+			}]});',
+			esc_js( (float) $product->get_price() ),
+			esc_js( $product->get_id() ),
+			esc_js( (float) $product->get_price() ),
```
mikkamp: In both these get_price calls we need to use the function `wc_get_price_to_display( $product )`, it would already return a float so no need to type cast it. Also if we know it's going to return only a float then esc_js doesn't really do anything.

The reason we need to use this, is because if I have products prices set excluding tax, but display in product pages including tax then we track the price without tax which isn't the same as we are displaying on the product pages.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			$item_info = [];
+
+			foreach ( WC()->cart->get_cart() as $cart_item ) {
+				// gets the product id
+				$id = $cart_item['product_id'];
+
+				// gets the product object
+				$product = $cart_item['data'];
+				$name    = $product->get_name();
+				$price   = $product->get_price();
```
mikkamp: This should also be the price depending on how it's displayed on cart pages.
```suggestion
				$price   = WC()->cart->display_prices_including_tax() ? wc_get_price_including_tax( $product ) : wc_get_price_excluding_tax( $product );
```

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+				// gets the cart item quantity
+				$quantity = $cart_item['quantity'];
+
+				$item_info[] = sprintf(
+					'{
+					id: "gla_%s",
+					price: %s,
+					google_business_vertical: "retail",
+					name:"%s",
+					quantity:"%s",
```
mikkamp: We are tracking price as a float, should quantity also be a regular int instead of a string?

Same for the purchase snippet.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+					esc_js( $name ),
+					esc_js( $quantity )
+				);
+			}
+			$value          = WC()->cart->total;
+			$page_view_gtag = sprintf(
+				'gtag("event", "page_view", {
+				send_to: "GLA",
+				ecomm_pagetype: "cart",
+				value: "%s",
```
mikkamp: Per item we track the price as a regular float, but here we are turning the total cart value into a string. We don't seem to have the full documentation on all the parameter, but this [document](https://support.google.com/google-ads/answer/7305793?hl=en#zippy=%2Cretail) is showing the retail example with the value set as float.

Same for the conversion and purchase snippets.

mikkamp (commented):
Thanks for implementing all those changes, it's looking good.

It's still not clear on the full documentation for some of the items parameters as I don't see any price/quantity info getting passed, but I added a few minor comments based on what we currently know.
I also added some comments about using the correct helper function to track the correct price depending on tax settings. You can test the various scenarios by setting up taxes in WooCommerce > Settings > Tax and add some basic tax rates, then change how taxes are displayed on shop or cart pages.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: I see. It makes sense. Thanks! Updated.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for the suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your suggestions. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated to float. Thanks!

src/Google/GlobalSiteTag.php
songchen2022: Thanks for your detailed explanation.Updated.

songchen2022:
> Thanks for implementing all those changes, it's looking good.
> 
> It's still not clear on the full documentation for some of the items parameters as I don't see any price/quantity info getting passed, but I added a few minor comments based on what we currently know. I also added some comments about using the correct helper function to track the correct price depending on tax settings. You can test the various scenarios by setting up taxes in WooCommerce > Settings > Tax and add some basic tax rates, then change how taxes are displayed on shop or cart pages.

Caio is working on the documentation. Will let you know if we have any updates. Thanks for your explanation about tracking the correct price.

songchen2022:
Hi Michael, thanks for reviewing and the suggestions. I've replied the comments. Could you please help take a look again? Thanks!

src/Google/GlobalSiteTag.php
```diff
@@ -13,14 +14,18 @@
...
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
 
 /**
  * Main class for Global Site Tag.
  */
 class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareInterface {
 
+
+
```
mikkamp: Minor: Just one empty line is enough here.

src/Google/GlobalSiteTag.php
```diff
@@ -91,10 +124,10 @@ public function activate_global_site_tag( string $ads_conversion_id ) {
 		if ( $this->gtag_js->is_adding_framework() ) {
 			add_filter(
 				'woocommerce_gtag_snippet',
-				function( $gtag_snippet ) use ( $ads_conversion_id ) {
+				function ( $gtag_snippet ) use ( $ads_conversion_id ) {
 					return preg_replace(
 						'~(\s)</script>~',
-						"\tgtag('config', '" . $ads_conversion_id . "');\n$1</script>",
+						"\tgtag('config', '" . $ads_conversion_id . "', { 'groups': 'GLA', 'send_page_view': 'false' });\n$1</script>",
```
mikkamp: According to the [docs](https://support.google.com/analytics/answer/11403294?hl=en) the `send_page_view` should be set to just a boolean false.

src/Google/GlobalSiteTag.php
```diff
@@ -104,47 +137,48 @@ function( $gtag_snippet ) use ( $ads_conversion_id ) {
...
+		dataLayer.push(arguments);
+	  }
+	  gtag('js', new Date());
+	  gtag('set', 'developer_id.<?php echo esc_js( self::DEVELOPER_ID ); ?>', true);
 
-	gtag('config','<?php echo esc_js( $ads_conversion_id ); ?>');
-</script>
+	  gtag('config', '<?php echo esc_js( $ads_conversion_id ); ?>', {
+		'groups': 'GLA',
+		'send_page_view': 'false'
```
mikkamp: Same here, just a boolean should be sufficient.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+		sprintf(
+			'gtag("event", "purchase", {
+			ecomm_pagetype: "purchase",
+			send_to: "%s",
+			transaction_id: "%s",
+			currency: "%s",
+			country: "%s",
+			value: %f,
+			new_customer: %s,
+			tax: "%s",
```
mikkamp: Tax is also a monetary amount like value.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+	/**
+	 * Display the JavaScript code to track all pages.
+	 */
+	private function display_page_view_event_snippet(): void {
+		if ( ! is_cart() ) {
+			wp_print_inline_script_tag(
+				'gtag("event", "page_view", {send_to: "GLA"});'
+			);
+			return;
+		} else {
```
mikkamp: Since the return was added in the if block above. We no longer need an else block. All the code below can be moved outside of the if block so we can indent it with one tab less.

src/Google/GlobalSiteTag.php
```diff
@@ -155,13 +189,164 @@ public function maybe_display_event_snippet( string $ads_conversion_id, string $
...
+			foreach ( WC()->cart->get_cart() as $cart_item ) {
+				// gets the product id
+				$id = $cart_item['product_id'];
+
+				// gets the product object
+				$product = $cart_item['data'];
+				$name    = $product->get_name();
+				$price   = WC()->cart->display_prices_including_tax() ?
+				wc_get_price_including_tax( $product ) : wc_get_price_excluding_tax( $product );
+				$product->get_price();
```
mikkamp: It's probably best to keep this all on one line. There are no strict styling rules if it's on multiple lines, but it would be clearer if the part on the following lines is indented (if you decide not to put it on one line).
Also it looks like the `$product->get_price` line is left over code, it doesn't do anything.

mikkamp (commented):
Thanks for those changes. I've left some additional comments, mostly minor.

It looks pretty much ready to go, once these comments have been addressed. I've changed the base branch to feature/gtag-events so we can release all the GTag PR's together when they are ready. This means we don't necessarily need to wait to merge till the exact variables are clear and can fix them in a follow up PR once the documentation is ready.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

src/Google/GlobalSiteTag.php
songchen2022: Done.

src/Google/GlobalSiteTag.php
songchen2022: Thanks for pointing this out. I found the return value of get_cart_tax() of the cart is string and the return value of get_cart_tax()  of the order is float. I mixed them. Updated.

src/Google/GlobalSiteTag.php
songchen2022: Updated.

songchen2022:
Thanks so much for reviewing! I've replied the comments. Please let me know if anything I need to do.

mikkamp (approved):
Thanks, those changes look good. I'm going to go ahead and merge this to the feature branch, so I can rebase PR #1488 as a followup from this code.

Once we have the final information about the item parameters we can create a followup PR to adjust any of the values if necessary.

songchen2022:
Thanks so much for all your help! Please let me know if anything I could do.

PR #1330, opened 2022-03-16
Tree Select Component - Basic Accessibility 

puntope:

This PR implements the next accessibility features:

- ESC to close the Tree
- TAB to go to the next Component 
- BACKSPACE to remove tags (if no search text is present) 

### Screenshots:

https://user-images.githubusercontent.com/5908855/158558092-df710b46-5a85-465f-a39d-7205e3ff14f6.mov

1. Checkout this branch
3. Enter in the Selector and see how it expands
4. Press Tab and see how it goes to the next selector
5. Go back to the selector and see how it opens again
6. Press Escape and see how the tree collapses
7. Type some text. 
8. Press Backspace and see how the text is deleted
9. When there is no text, press again Backspace and see how it removes the last tag

### DIFF
--- a/js/src/components/tree-select-control/control.js
+++ b/js/src/components/tree-select-control/control.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import classnames from 'classnames';
-import { forwardRef, useRef } from '@wordpress/element';
+import { forwardRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
--- a/js/src/components/tree-select-control/control.js
+++ b/js/src/components/tree-select-control/control.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import classnames from 'classnames';
-import { forwardRef } from '@wordpress/element';
+import { useRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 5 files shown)


js/src/components/tree-select-control/control.js
```diff
@@ -25,67 +25,71 @@ import Tags from './tags';
...
+		const hasTags = tags.length > 0;
+		const showPlaceholder = ! hasTags && ! isExpanded;
 
-	return (
-		/*
+		return (
+			/*
 		ESLint Disable reason
 		https://github.com/woocommerce/woocommerce-admin/blob/main/packages/components/src/select-control/control.js#L200
 		*/
```
eason9487: 💅  The indentation of code comment is not aligned.
```suggestion
			/**
			 * ESLint Disable reason
			 * https://github.com/woocommerce/woocommerce-admin/blob/main/packages/components/src/select-control/control.js#L200
			 */
```

js/src/components/tree-select-control/index.js
```diff
@@ -296,8 +315,13 @@ const TreeSelectControl = ( {
 	};
 
 	return (
+		// eslint-disable-next-line jsx-a11y/no-static-element-interactions
 		<div
 			{ ...focusOutside }
+			onKeyDown={ onKeyDown }
+			onClick={ () => {
+				setTreeVisible( true );
+			} }
```
eason9487, line 324: ❔ 
The dropdown menu is already shown by triggering the internal `onClick` of \<Control\> and then the `onFocus` of \<Control\>. I can't figure out what the cases will be dealt with here. It seems to be working fine after deleting it?

js/src/components/tree-select-control/index.js
```diff
@@ -163,6 +168,20 @@ const TreeSelectControl = ( {
 		setFilteredOptions( filteredTreeOptions );
 	}, [ treeOptions, filter ] );
 
+	const onKeyDown = ( event ) => {
+		if ( inputRef.current.value ) return;
```
I'm not sure if the \<input\> instance (`inputRef.current`) will continue to be used in subsequent accessibility development?

If not, it should be possible to set state from `handleOnInputChange` by adding a new `useState` and getting the same value here. Instead of using `inputRef.current.value` to get it.

js/src/components/tree-select-control/control.js
```diff
@@ -25,67 +25,71 @@ import Tags from './tags';
...
+			disabled,
+			maxVisibleTags,
+			onFocus = () => {},
+			onTagsChange = () => {},
+			onInputChange = () => {},
+		},
+		inputRef
+	) => {
+		const hasTags = tags.length > 0;
+		const showPlaceholder = ! hasTags && ! isExpanded;
```
eason9487: ❓ 

Although \<Control\> is an internal component, given the design of a component and the fact that \<Control\> itself also operates inputRef, e.g. `inputRef.current.focus();`. I'd recommended to add `useImperativeHandle` so that the internal `inputRef` operation can have fallback tolerance if the `ref` prop is not set from the consumer side.

Another advantage is that it can decide which interfaces to be exposed, rather than exposing the \<input\> instance.

💡  It will need to import `useRef` and `useImperativeHandle` on the top if took the suggestion directly.

```suggestion
		ref
	) => {
		const inputRef = useRef();
		const hasTags = tags.length > 0;
		const showPlaceholder = ! hasTags && ! isExpanded;

		useImperativeHandle(
			ref,
			() => ( {
				get value() {
					return inputRef.current.value;
				},
			} ),
			[]
		);
```

And if it's easier to expose the \<input\> instance, it could also use this method to do the fallback handling.
```suggestion
		ref
	) => {
		const inputRef = useRef();
		const hasTags = tags.length > 0;
		const showPlaceholder = ! hasTags && ! isExpanded;

		useImperativeHandle( ref, () => inputRef.current, [] );
```

Ref: https://reactjs.org/docs/hooks-reference.html#useimperativehandle

eason9487 (commented):
💡  Two comments are associated and interlinked with the forwarding/getting of \<Control\>'s reference. Please check them together.
- https://github.com/woocommerce/google-listings-and-ads/pull/1330#discussion_r827916991
- https://github.com/woocommerce/google-listings-and-ads/pull/1330#discussion_r827931014

js/src/components/tree-select-control/index.js
puntope: Why did we add the onClick? 

This is because of the Escape feature. When you click escape, the Tree collapses but the input still has the focus. Hence, if you click on it, no onFocus event is triggered anymore and the tree doesn't expand.

Why we didn't remove the onFocus? 

On the other side, if we remove the onFocus, when the user navigates over the elements with Tab and the Tree Select gets the focus, it will not open since onFocus is not there anymore. 

Having both is the only way I found so far, open to fresh ideas tho...

puntope:
Thanks for the tips @eason9487  I changed the forward strategy finally for a state-based strategy. So I guess we don't need to implement the imperative handler (at least not yet).

js/src/components/tree-select-control/index.js
eason9487: Thanks for the clarification! I thought about it, probably both event handlers are necessary and cannot omit either one.

eason9487 (approved):
Testing well. LGTM.

PR #1334, opened 2022-03-17
Update delete campaigns API to remove its campaign criteria

ianlin:

Part of #1248, this PR updates the `DELETE ads/campaigns/:campaign-id` API so it will remove the campaign's criterion as well.

Unlike campaign budget, the removed locations do not exist anymore, even though I used the same `setRemove` operation method. From the Google Ads dashboard, the removed campaign's location is empty either, so I assume this is an expected behaviour.

#### 1. Create a campaign with targeted locations

delete-criterion-1-create

#### 2. Get the newly created campaign

The status should be `enabled`.

delete-criterion-1-get-1

#### 3. Delete the campaign

delete-criterion-1-delete

#### 4. Delete again

The API should respond the error.

delete-criterion-1-delete-2

#### 5. Get the same campaign again

The status should be `removed`.

delete-criterion-1-get-2

1. Create a campaign by requesting to `POST ads/campaigns` using Postman like the screenshots in the `Screenshots` section. Remember to add your `wp-cookie` and `x-wp-nonce` in the request headers. 
    - To get the value of `wp-cookie` and `x-wp-nonce`, go to `Marketing -> Google Listings & Ads`, open Chrome dev tool's Network tab, clicks any requests to WP, e.g. `wc/gla/ads/campaigns`, you can get those values in the request headers.
2. Get the newly created campaign by requesting to `GET ads/campaigns/:campaign-id`, the campaign's status should be `enabled`.
3. Delete the campaign by requesting `DELETE ads/campaigns/:campaign-id`, the API call should be successful.
4. Delete the campaign again, the API should respond an error: `This campaign has already been deleted`.
5. Get the campaign again by requesting to `GET ads/campaigns/:campaign-id`, the campaign's status should be `removed`.
6. Go to Google Ads dashboard and check the campaign is removed and the locations are empty.

### Changelog entry

> Update - Extend delete campaign API to delete its campaign criteria

### DIFF
--- a/src/API/Google/AdsCampaignCriterion.php
+++ b/src/API/Google/AdsCampaignCriterion.php
@@ -67,7 +67,6 @@ class AdsCampaignCriterion implements OptionsAwareInterface {
 		$results = ( new AdsCampaignCriterionQuery() )
 			->set_client( $this->client, $this->options->get_ads_id() )
 			->where( 'campaign_criterion.campaign', $campaign_resource_name )
-			->where( 'campaign_criterion.negative', 'false', '=' )
 			->where( 'campaign_criterion.status', 'REMOVED', '!=' )
 			->where( 'campaign_criterion.location.geo_target_constant', '', 'IS NOT NULL' )
 			->get_results();
--- a/src/API/Google/AdsCampaignCriterion.php
+++ b/src/API/Google/AdsCampaignCriterion.php
@@ -3,6 +3,10 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query\AdsCampaignCriterionQuery;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\Ads\GoogleAdsClient;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Google\Ads\GoogleAds\Util\V9\ResourceNames;
 use Google\Ads\GoogleAds\V9\Common\LocationInfo;
 use Google\Ads\GoogleAds\V9\Enums\CampaignCriterionStatusEnum\CampaignCriterionStatus;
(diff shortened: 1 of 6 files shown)


src/API/Google/AdsCampaignCriterion.php
```diff
@@ -35,6 +55,35 @@ function ( $location_id ) use ( $campaign_resource_name ) {
...
+	 * @return array
+	 */
+	public function delete_operations( string $campaign_resource_name ): array {
+		$operations = [];
+
+		$results = ( new AdsCampaignCriterionQuery() )
+			->set_client( $this->client, $this->options->get_ads_id() )
+			->where( 'campaign_criterion.campaign', $campaign_resource_name )
+			->where( 'campaign_criterion.negative', 'false', '=' )
+			->where( 'campaign_criterion.status', 'REMOVED', '!=' )
```
mikkamp, line 70: I'm not able to see the status either in the Google dashboard, should we also remove it here?

src/API/Google/AdsCampaignCriterion.php
```diff
@@ -35,6 +55,35 @@ function ( $location_id ) use ( $campaign_resource_name ) {
...
+	 * @param string $campaign_resource_name
+	 * @return array
+	 */
+	public function delete_operations( string $campaign_resource_name ): array {
+		$operations = [];
+
+		$results = ( new AdsCampaignCriterionQuery() )
+			->set_client( $this->client, $this->options->get_ads_id() )
+			->where( 'campaign_criterion.campaign', $campaign_resource_name )
+			->where( 'campaign_criterion.negative', 'false', '=' )
```
mikkamp: When we want to delete the campaign completely, wouldn't it make sense to remove all location criteria, regardless if they are negative or not?

mikkamp (commented):
Thanks for the changes, it's working as expected. I just left a few small comments about things I noticed.

Although I was thinking, should we really be removing these locations? When I view a removed campaign I can still see how it was set up and which removed budget and removed ad group it used. But I can no longer see what locations it was configured for, since we clean those up.
Maybe we should ask Google what they would recommend.

src/API/Google/AdsCampaignCriterion.php
ianlin: Makes sense, thanks for the suggestion. Updated in 5a8c28f.

ianlin:
Hey @mikkamp, thanks for the review! I'm also curious why the locations criteria were really removed and what would be the best way to implement it. I asked in the Google Q&A doc per your suggestion then.

ianlin:
Hey @mikkamp I've created a post pb0Spc-2pt-p2 to discuss whether we should only set the REMOVED status to the campaign resource and leave everything else alone.

ianlin:
Close this PR as we've concluded we'd only remove the parent campaign (set to `REMOVED` status) based on the thorough test by @mikkamp. Ref: pb0Spc-2pt-p2#comment-5182.

A followup issue to remove other delete operations is #1362.

PR #1336, opened 2022-03-17
Tree Select Control - Keyboard Navigation

puntope:

Implements a11y Keyboard navigation in the Tree Select Control

ArrowUP -> Move up in the tree
ArrowDown -> Moves down in the tree
ArrowLeft -> Collapses a group
ArrowRight -> Expands a group
Space -> Selects the node
ESC -> Closes the tree
ENTER -> Open the tree
Delete -> Deletes the search letter or the las tag
Tab -> Moves out to the next component

### Screenshots:

https://user-images.githubusercontent.com/5908855/158868822-300e90d5-0aa5-4736-9cb2-5e3f0408b8e1.mov

1. Checkout this PR
2. Go to  `wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fcomponent-test`
3. Click on the tree select component
4. Navigate with arrow as indicated at the top of this description 

### Additional details:

This is a first working approximation to the feature. Besides all the code I consider necessary to explain the next points

- We don't use anymore CheckboxControl, now we created the Checkbox manually. This is because somehow CheckboxControl doesn't allow ref. 
- We load the refs in the options when we build the option repository. This repository will be the main source of fast access to know which one is the parent and get the ref based on key/value
- We added a data-idx in the options to act as a reference of the position of that ode inside a group.
- When we select the option or expand it, it gets the focus and we can navigate from that index.
- Navigation is disabled when searching.


js/src/components/tree-select-control/index.js
```diff
@@ -116,6 +126,102 @@ const TreeSelectControl = ( {
...
+		}
+		return child || getSibling( parent );
+	};
+
+	/**
+	 * Gets an option from the repository
+	 *
+	 * @param {Option} option The option to get from the Repository
+	 */
+	const getOption = useCallback(
```
eason9487: 💅 
How about renaming this function to indicate that it gets the associated inner option from the repository?

js/src/components/tree-select-control/index.js
```diff
@@ -32,6 +42,7 @@ import {
  * @property {string} label The label for the option
  * @property {Option[]} [children] The children Option objects
  * @property {string} [key] Optional unique key for the Option. It will fallback to the value property if not defined
+ * @property {Object} [ref] React Ref associated to the option. This is for system purposes only.
```
eason9487: 💅 
To separate the `options` prop and their associated inner options in the repository, we could use the other two `@typedef` and extend the `Option` type. In this way, the internal property `ref` could be hidden from component users.

```js
/**
 * @typedef {Object} BaseInnerOption
 * @property {Option|undefined} parent The parent option of this inner option.
 * @property {Object} ref React Ref associated to this inner option.
 *
 * @typedef {Option & BaseInnerOption} InnerOption
 */
```

In addition, we could use `InnerOption` type for those functions that require/return `InnerOption` type option, for example the `getOption` at below:

```js
const getOption = useCallback(
	/**
	 * Gets an option from the repository
	 *
	 * @param {Option} option The option to get inner option from the repository.
	 * @return {InnerOption|undefined} The inner option from repository.
	 */
	( option ) => {
		if ( ! option ) return;
		return optionsRepository[ option.key ?? option.value ];
	},
	[ optionsRepository ]
);
```

💡  The whole JSDoc of `getOption` is moved inside the `useCallback` so that the editor can show the info correctly.

js/src/components/tree-select-control/index.js
```diff
@@ -163,52 +271,107 @@ const TreeSelectControl = ( {
...
+			setTreeVisible( true );
 			event.preventDefault();
 		}
 
-		if ( ESCAPE === event.key ) {
-			setTreeVisible( false );
+		if ( inputControlValue ) return;
+
+		if ( ARROW_UP === event.key ) {
+			const parent = getParent( focused.el );
```
eason9487: 🚧 
When the focus is on \<Control\>, an exception error occurs while pressing the up key.

Kapture 2022-03-18 at 14 00 22

js/src/components/tree-select-control/index.js
```diff
@@ -163,52 +271,107 @@ const TreeSelectControl = ( {
...
+					: filteredOptions[ filteredOptions.length - 1 ];
+
+				setFocused( {
+					el,
+					parent: getParent( el ),
+					idx: getIndex( el ),
+				} );
+			} else if ( child ) {
+				const el = nodesExpanded.includes( child.value )
+					? getLastChild( child )
```
eason9487: 🚧  or ❓ 

When moving up, if parent nodes are crossed and a deeper layer is existing and expanded, the deeper child node will be skipped.

Kapture 2022-03-18 at 13 54 48

js/src/components/tree-select-control/index.js
```diff
@@ -163,52 +271,107 @@ const TreeSelectControl = ( {
...
+				setFocused( {
+					el: getOption( child ),
+					parent: getParent( child ),
+					idx: 0,
+				} );
+			} else {
+				const newIndex = getIndex( focused.el ) + 1;
+				const parent = getParent( focused.el );
+				const child = parent?.children[ newIndex ];
+				const sibling = ! child && getSibling( parent );
```
The namings seem to be imprecise. 
- The actual node of `child` variable looks more like "the sibling of the current node"?
- And the actual node of `sibling` variable looks like "the next sibling node of the current node's nested parents (if existing)"?

js/src/components/tree-select-control/index.js
```diff
@@ -163,52 +271,107 @@ const TreeSelectControl = ( {
 
 		let filteredTreeOptions = cloneDeep( treeOptions );
 
-		if ( filter ) {
-			filteredTreeOptions = filteredTreeOptions.filter( filterOption );
-		}
-
```
eason9487, line 169: ❔ 
Remove this checking will make the HTML structure of all `label` become the following when not searching. Perhaps making up the removed checking to another place would be nicer.

```html
<label class="components-checkbox-control__label">
  <span>
    <span></span>
    <strong></strong>
    <span>Spain</span>
  </span>
</label>
```

js/src/components/tree-select-control/options.js
```diff
@@ -81,17 +84,39 @@ const Options = ( {
...
+	 * @param {Event} event The KeyDown event
+	 * @param {Option} option The option where the event happened
+	 * @param {boolean} isExpanded True if the node is expanded, false otherwise
+	 */
+	const handleKeyDown = ( event, option, isExpanded ) => {
+		if ( event.key === ARROW_RIGHT && ! isExpanded ) {
+			toggleExpanded( option );
+		} else if ( event.key === ARROW_LEFT && isExpanded ) {
+			toggleExpanded( option );
+		} else if ( event.key === SPACE ) {
```
eason9487, line 126: ❔ According to the spec, looks like the toggling is also supported by the Enter key.

js/src/components/tree-select-control/checkbox.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * External dependencies
+ */
+import { Icon, check } from '@wordpress/icons';
+
```
eason9487, line 5: 💅  Add the unresolved `Option` type, or the `InnerOption` type if took this [suggestion](https://github.com/woocommerce/google-listings-and-ads/pull/1336#discussion_r829697056).

```js
/**
 * @typedef {import('./').Option} Option
 */
```

js/src/components/tree-select-control/index.js
```diff
@@ -163,52 +271,107 @@ const TreeSelectControl = ( {
...
-		if ( BACKSPACE === event.key ) {
-			onChange( value.slice( 0, -1 ) );
+		if ( ENTER === event.key ) {
+			setTreeVisible( true );
 			event.preventDefault();
 		}
 
-		if ( ESCAPE === event.key ) {
-			setTreeVisible( false );
+		if ( inputControlValue ) return;
```
eason9487, line 341: > Navigation is disabled when searching.

Will this be added in the subsequent PR or be a future roadmap?

js/src/components/tree-select-control/index.js
```diff
@@ -163,52 +271,107 @@ const TreeSelectControl = ( {
...
+
+		if ( ARROW_UP === event.key ) {
+			const parent = getParent( focused.el );
+			const child = parent?.children[ focused.idx - 1 ];
+
+			if ( ! parent ) {
+				const el = root
+					? getLastChild( root )
```
eason9487: 🚧  or ❓ 
Same as https://github.com/woocommerce/google-listings-and-ads/pull/1336#discussion_r829717689.

eason9487:
🚧

Found that the expanding behavior is inconsistent when unchecking all checked options with the mouse and keyboard.

https://user-images.githubusercontent.com/17420811/158959024-8b5c68da-2c85-44d8-9814-cae53e941149.mp4

puntope:
Howdy @eason9487 , huge thanks for all the bugs and improvements you found and suggested 

I applied several fixes in the last commits:

- Fixes Node Expand inconsistency
- Fixes jumps  bug in navigation
- Fix exception when clicking arrows on the search control
- Improve naming and JSDocs
- Refactoring / Ploshing logic in Arrow Navigation Functions 

Also while I was solving the bugs mentioned I found another set of bugs when there is no Root node. Now they are fixed as well.

js/src/components/tree-select-control/options.js
puntope: Not sure how accurate are the specs with the real needs. Since we are selecting with SPACE the node and we don't have any Key for expanding the tree itself when the user closes it with ESC I thought it will be more useful ENTER for that.

js/src/components/tree-select-control/index.js
puntope: Yes, I'm working already on that (in future PR). However, we will need to iterate manually over the tree since we don't know which nodes are present or not. But I don't think it will be problematic since after the search we should have fewer nodes.

js/src/components/tree-select-control/options.js
eason9487: This makes sense, thanks for the explanation!

js/src/components/tree-select-control/index.js
```diff
@@ -116,6 +133,144 @@ const TreeSelectControl = ( {
...
+			  );
+	};
+
+	/**
+	 * Get the next option available in the tree
+	 *
+	 * The function attempts to get the next option in the current level
+	 * if there is no more options, it tries to get the next option on the upper level.
+	 *
+	 * @param {Option} option TThe reference option to get the next element
```
eason9487: A typo
```suggestion
	 * @param {Option} option The reference option to get the next element
```

js/src/components/tree-select-control/options.js
```diff
@@ -81,17 +85,53 @@ const Options = ( {
...
+			onChange( event.target.checked, option );
```
eason9487, line 127: 💅  or ❓ 
I just noticed that when pressing Space key on a focused checkbox, the default behavior is already the toggling of checked and unchecked states* in the checkbox itself. (*The native state of <input>, not the state of React.) And I tested Chrome, FireFox, and Safari on macOS, they can work correctly after removing lines 126-127.

And removing it could also make the `onChange` callback of \<Options\> be triggered correctly because the `event.target.checked` state got from here is the opposite. The final result of state is still correct due to the following `onChange` event of \<Checkbox\> being triggered after it.

https://github.com/woocommerce/google-listings-and-ads/blob/db75784d1b83d67a1d38367fffe2c89373cd898b/js/src/components/tree-select-control/options.js#L178-L180

For example, when a \<Checkbox\> is **unchecked**, the [`handleOptionsChange` callback](https://github.com/woocommerce/google-listings-and-ads/blob/db75784d1b83d67a1d38367fffe2c89373cd898b/js/src/components/tree-select-control/index.js#L388) which is passed as `onChange` prop of \<Options\> will receive two events with different `checked` as following after pressing Space.

1. Frist event. The `checked` value is **false** called from [the `onChange` in `handleKeyDown`](https://github.com/woocommerce/google-listings-and-ads/blob/db75784d1b83d67a1d38367fffe2c89373cd898b/js/src/components/tree-select-control/options.js#L127). The input is still unchecked when triggering the `onKeyDown`
2. Second event. The `checked` value is **true** called from [the `onChange` in \<input\>'s `onChange`](https://github.com/woocommerce/google-listings-and-ads/blob/db75784d1b83d67a1d38367fffe2c89373cd898b/js/src/components/tree-select-control/options.js#L180). This state change is triggered by the \<input\> itself.

eason9487 (approved):
Thanks for the adjustments! Testing well. Left two new comments that could be tweaked but they were not blocking this PR.

ecgan:
Just want to drop a note to say I happened to notice there were 44 files changed in this PR, because there were changes from `develop` branch that were merged into `add/a11y-navigation` branch but not in `feature/tree-select-control-component` branch. 

image

I have fixed the issue by merging `develop` branch into `feature/tree-select-control-component` branch and then merging `feature/tree-select-control-component` into `add/a11y-navigation`. Now we have only 6 files changed in this PR. 🙂

js/src/components/tree-select-control/options.js
puntope: Thanks @eason9487  - To be implemented in next PR within this feature https://github.com/woocommerce/google-listings-and-ads/pull/1336#discussion_r829755331

puntope:
> I have fixed the issue by merging `develop` branch into `feature/tree-select-control-component` branch and then merging `feature/tree-select-control-component` into `add/a11y-navigation`. Now we have only 6 files changed in this PR. 🙂

Thanks @ecgan 👍 

I guess then this comment p1647928212038389-slack-C01E9LB4X61 is not needed anymore right?

ecgan:
> I guess then this comment p1647928212038389-slack-C01E9LB4X61 is not needed anymore right?

Yup 😄

PR #1226, opened 2022-02-01
Refactor MCAccountController and add unit tests

mikkamp:
### Changes proposed in this Pull Request:

This PR is a replacement for #973 and is created based on the feedback in the review of the unit tests for the Merchant AccountController, see here. The issue #388 mentions the same strategy.

In this PR I refactored the MCAccountController and moved all the account handling to a separate MCAccountService class. This simplified the unit testing as we can test each class individually.
One unit test was created for the MCAccountController (which is based on the RESTControllerUnitTest). Another unit test was created for the MCAccountService class.

The `disconnect` and `get_connected_status` function were moved from MerchantCenterService class to the MCAccountService class since they are directly related to the account (at the moment we still don't have unit tests for this class so no tests needed to be adjusted there).

The functionality for verifying a site URL was very complex to unit test as part of the account setup steps. For this reason it was moved into the SiteVerification class, it is a step in setting up the account but it's better placed within it's own class (much like the PhoneVerification class) so it can be reused by the SiteVerificationController (if those API endpoints are still used). It will need separate unit testing which can be handled in a separate PR.

Resolves part of #388

### Detailed test instructions:

The main testing is to confirm that all the added unit tests pass.

However since we also refactored the MerchantCenter account functions we need to confirm that they all continue to function the same. Following are all the affected requests and their expected outcome.

Enable the Connection Test page through a [code snippet](https://github.com/woocommerce/google-listings-and-ads/wiki/Hooks-and-Filters)

#### Merchant disconnect
- Go to Connection Test page 
- Click MC Disconnect
- Confirm that we get a success response

#### Create a new Merchant account
- Make sure we are using a public URL (ngrok or localtunnel)
- Go to Marketing > Google Listings & Ads > Set up free listings in Google (might need to refresh dashboard if we just disconnected)
- Connect WP and Google accounts (if not already connected)
- Click on "Or, create a new Merchant Center account"
- Confirm merchant account is connected
image

#### Link existing Merchant account
- From previous test, click on "Or, connect to a different Google Merchant Center account"
- Select a previously created account (that's also a subaccount)
- Confirm we see the message to reclaim the URL
image
- Reclaim URL and confirm we are once again connected
image

#### Link existing standalone Merchant account
- From previous test, click on "Or, connect to a different Google Merchant Center account"
- Select a previously created account (that's a standalone account, not created through our extension)
- Confirm we see the error message that we are unable to claim the URL
image

#### Switch URL for a previously created site
[...]

PR #1208, opened 2022-01-21
Refactor AdsAccountController and add unit tests

mikkamp:
### Changes proposed in this Pull Request:

This PR is a replacement for #977 and is created based on the feedback in the review of the unit tests for the Merchant AccountController, see [here](https://github.com/woocommerce/google-listings-and-ads/pull/973#discussion_r697233209). The issue #388 mentions the same strategy.

So in this PR I refactored the AdsAccountController and moved all the account handling to a separate AdsAccountService class. This simplified the unit testing as we can test each class individually.
One unit test was created for the AdsAccountController (which is based on the RESTControllerUnitTest). Another unit test was created for the AdsAccountService class.

The disconnect function was already in the AdsService class but I didn't want all the account functions to be mixed in with the regular functions in this class. There isn't a lot in the AdsService class, but if we copy this same pattern for the MerchantCenterService then it would be better to keep the account functions in a separate service since they are only used when connecting / disconnecting.

Resolves part of #388

### Detailed test instructions:
The main testing is to confirm that all the added unit tests pass.

However since we also refactored the Ads account functions we need to confirm that they all continue to function the same. Following are all the affected REST API requests and their expected responses. Each one can be tested with the help of Postman (or with the Connection Test page which builds the request internally).

#### Disconnect Ads account

```
DELETE https://domain.test/wp-json/wc/gla/ads/connection
Response:
{
	"status": "success",
	"message": "Successfully disconnected."
}
```

#### Create new account

```
POST https://domain.test/wp-json/wc/gla/ads/accounts
Body: <empty>
Response Code: 428
Response:
{
	"message": "Billing setup must be completed.",
	"billing_url": "https://ads.google.com/nav/startacceptinvite?ivid=123&ocid=456",
	"billing_status": "unknown"
}
```

When testing creating a new account, we need to visit the `billing_url` link and click the accept button (or view the email and click the link to accept the account). Entering the billing account info can be skipped while testing.

#### Check billing status

```
GET https://domain.test/wp-json/wc/gla/ads/billing-status
Response:
{
	"status": "pending",
	"billing_url": "https://ads.google.com/nav/startacceptinvite?ivid=123&ocid=456"
}
```

This endpoint will return status `approved`, once the billing details have been completed, or for testing we can use a [code snippet](https://github.com/woocommerce/google-listings-and-ads/wiki/Hack-for-working-locally-with-APIs#use-ads-account-without-setting-billing-data) to force it to return approved.

#### Connect an existing account

[...]

PR #1340, opened 2022-03-18
Refactor/middleware requests

### LINKED ISSUE #388: Refactor Merchant and Ads Account controllers
Based on this comment: https://github.com/woocommerce/google-listings-and-ads/pull/367#issuecomment-808127870

Both the Merchant and Ads Account controllers need to reduce the amount of functionality they encapsulate. Since the introduction of the multistep onboarding we allowed too much to creep in. Now that we have a Service class for both Merchant and Ads, we should move some of the functionality there.

> The API/Google/Merchant is more of a proxy (wrapper) class but I like to think of MerchantCenterService as more of a helper class that handles our business logic related to managing the merchant center.

> The AccountController is handling too much logic and it's always good to have thin controllers (and not so thin services).

mikkamp:

This PR refactors the Proxy class and renames it to Middleware. In most of our classes we already renamed it to Middleware using an alias.

The following steps were taken in the Proxy.php class to allow for easier testing:
- `disconnect_ads_account` is no longer used
- Ads currency functions moved to API\Google\Ads.php 
- `get_connected_ads_account` moved to Ads\AccountService.php
- `get_ads_account_ids` can be moved to API\Google\Ads.php WCS doesn't handle this request directly like the other requests in Proxy.php
- `get_merchant_ids` is not used as it was renamed to `get_merchant_accounts`
- `get_merchant_id` is no longer needed as we can get it from `options->get_merchant_id`
- Move update ads functions to Ads class
- Move update merchant functions to Merchant class
- Rename proxy class to middleware
- Pass Ads class to BudgetController since the currency functions got moved from Middleware

Unit tests were added for the classes:
- Ads
- Middleware

Closes #388

The important part is to confirm that the unit tests pass and cover the functionality within each class.

In addition to the unit tests we can also test the refactored functionality through the onboarding UI, this will need to be done by testing all the steps which send a manager request to the WCS.

1. Get list of Merchant account 
2. Link an existing Merchant account
3. Create a new Merchant account
4. Link an existing Ads account
5. Create a new Ads account
6. Claim website override

Detailed steps for reproducing these can be found in #1208 and #1226
For the last step we need to verify the site after it's already been verified on another site, see instructions "Switch URL for a previously created site" in #1226 

### Changelog entry

* Update - Refactor the Middleware class and add unit testing.


mikkamp:
This PR needs a fix for merge conflicts, but I'll wait till the review is completed.

src/API/Google/Merchant.php
```diff
@@ -314,4 +314,15 @@ public function has_access( string $email ): bool {
...
+
+	/**
+	 * Update the Merchant Center ID to use for requests.
+	 *
+	 * @param int $id  Merchant ID number.
+	 *
+	 * @return bool
+	 */
+	public function update_merchant_id( int $id ): bool {
+		return $this->options->update( OptionsInterface::MERCHANT_ID, $id );
```
ianlin, line 327: 🚧  Needs to import `OptionsInterface` from the correct path otherwise a PHP fatal would occur:

```
PHP Fatal error:  Uncaught Error: Class 'Automattic\WooCommerce\GoogleListingsAndAds\API\Google\OptionsInterface' not found in /var/www/html/wp-content/plugins/google-listings-and-ads/src/API/Google/Merchant.php:326
```

Would be great to add a unit test to catch this error, or maybe a linter rule that can capture this?

ianlin (commented):
High quality PR, commits are small and tidy, very easy to review such a big change. 👍 

Overall is good, tested step by step from the test instructions,
just need to import `OptionsInterface` from the correct path in `src/API/Google/Merchant.php`. I'm wondering if there's a linter rule that can capture it as from my Node.js experience there was a similar rule for it but not sure for PHP.

src/API/Google/Merchant.php
mikkamp: Nice catch. As long as the unit tests cover that line of code, it will produce an error. Seems like I forgot to add the test for that function though. I already had the test for `update_ads_id`, so I just copied that into MerchantTest

mikkamp:
Thanks for the review @ianlin. I've added the missing unit test and confirmed it caught the missing import. I've also merged the branch with develop and had to modify one more test since we get the ads currency from the Ads class instead of Middleware now. It's ready for another round of review.

ianlin (approved):
Tested again and everything works as expected and the fatal no longer exists. Good to go!

PR #1341, opened 2022-03-18
Get suggested shipping rates for provinces/states and postal codes from WC zones

### LINKED ISSUE #1230: Shipping Integration - Map rates for provinces/states and postal codes
As part of #1228

This issue is dependent on #1229.

At the end of milestone one, we currently have an integration with WooCommerce shipping zones. However, it only works at a country level. Meaning that the shipping methods are all grouped together and synchronized for each country. Even if they are defined at a more granular level such as a state or zip code.

Each WooCommerce shipping zone can have multiple regions along with multiple postcodes associated with it. The regions can be defined as continents, countries, or states and provinces of each country. The postcodes can also be defined either as ranges (90301-90305), wildcards (9003*), or one number (90210).

Even though WooCommerce supports the state/province level for many countries (see [states.php](https://github.com/woocommerce/woocommerce/blob/9d08ac8ac3723c897e08ac2ff8a6d950f0e26b5b/plugins/woocommerce/i18n/states.php)), Merchant Center only supports state-level granularity for a handful of countries. The list of supported countries along with their subregions should be provided by the helper methods implemented in #1229.

The `ShippingZone` class handles the processing of WooCommerce shipping zones and groups their methods and rates by country.

Modify the `parse_shipping_zones` method so that it stores the list of methods for each supported region separately. This means that we first need to check if the zone’s country is one of the ones supported by Merchant Center on a more granular level and then instead of grouping the shipping methods by country, we group them by the state, province, etc.

The postcodes stored in each zone should also be taken into account. If a zone has postcodes associated with it then all of its methods only apply to those postcodes, even if its region is on the country level.

The `get_shipping_rates_for_country` method should be modified to return the shipping rates for each region down to the postcode granularity. The returned rates can still be grouped by country. A new property can be added to the returned response containing the rates for each sub-region within that country (if supported by Merchant Center).

For more details on the mapping check #1228.

nima-karimi:

Closes #1230

This PR introduces new services to parse the WooCommerce shipping zones and return the shipping rates for each country and its subdivisions (state, provides, postcodes, etc.).

The `ShippingZone` service functionality is divided into three other classes that handle parsing the locations, methods, and grouping them together.

Instead of mapping the rates as multidimensional arrays, several value classes are introduced to hold `Location`, different types of `ShippingRate`, and a combination of the two: `LocationRate`. This allows us to have a consistent verifiable format when dealing with the suggested shipping rates throughout the codebase.

The `ShippingZone` service has now a new function that returns the rates grouped by countries. This is used in the suggestions API since we currently only suggest rates on a country level.

The previous method `get_shipping_rates_for_country` is now refactored to return the shipping rates for all subdivisions of a country as well.

All shipping rate suggestions are returned as a `LocationRate` object.

I've also removed the `currency` property from suggested shipping rates since they all always have the same currency as WooCommerce. The `currency` for the rates is now directly fetched from WooCommerce in suggestions API and Googe settings sync service.

The new functionality to return the shipping rates for subdivisions isn't being used anywhere in the code so it's not possible to test it without manually calling the `ShippingZone::get_shipping_rates_for_country` method somewhere and inspecting its output.

1. Define multiple shipping zones each containing various regions, postcodes, shipping methods, and rates for shipping classes.
2. Make a GET request to the `mc/shipping/rates/suggestions` API with the country_codes parameter set to an array of country codes from your defined shipping zones and confirm the rates are returned the same as you defined in your shipping zone settings. Note that the rates will be grouped at a country level.
3. Send a `POST` request to `wc/gla/mc/settings` and include the following data as JSON body: `{ "shipping_rate":"automatic" }`
4. Send a `POST` request to `wc/gla/mc/settings/sync`
5. Confirm that it's successful.
6. Check the shipping rates in your MC account and confirm that they match what you defined in Woo shipping zones. Again, the syncing functionality is still working on a country level so you won't see any state-specific rates synced.

To test that the code also returns rates on a subdivision level, you can call the `ShippingZone::get_shipping_rates_for_country` method anywhere in the code (for example, in the shipping rates suggestions controller) and inspect its output via xDebug or simply `var_dump`. It will return the rates as a flat array but each will have its `location` property set correctly.

### Additional details:

A side effect of these changes is that the `GET mc/shipping/rates/suggestions` API now returns the suggestions differently. Previously, if a `free_shipping` rate with a `free_shipping_threshold` was among the suggestions, this API would return it as a `flat_rate` with a `rate` of `0`. With this PR, the `free_shipping` suggestion will be returned as-is.

The other important change is that the `free_shipping_threshold` is now only expected to be set as an `option` for the `free_shipping` method. A `flat_rate` method no longer should have a `free_shipping_threshold` option.

For example, if the following shipping zones were created in WC settings:
image

This would be the response of `GET mc/shipping/rates/suggestions`:
```JSON
[
    {
        "country": "TR",
        "method": "free_shipping",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "US",
        "method": "flat_rate",
        "currency": "USD",
        "rate": 20,
        "options": {}
    },
    {
        "country": "US",
        "method": "free_shipping",
        "currency": "USD",
        "rate": 0,
        "options": {
            "free_shipping_threshold": 4.99
        }
    },
    {
        "country": "ES",
        "method": "free_shipping",
        "currency": "USD",
        "rate": 0,
        "options": {
            "free_shipping_threshold": 100
        }
    }
]
```

Note the suggested shipping rates for US. Previously the "free_shipping_threshold" would be merged with the "flat_rate" method but now it's on a separate "free_shipping" rate.

The `POST mc/shipping/rates` API is also affected in the same way. This means that the `free_shipping_threshold` option is always expected to be set on a `free_shipping` rate and Free Shipping should be defined as a rate with a type equal to `free_shipping`. A flat_rate with a cost of `0` would still work but it cannot have the `free_shipping_threshold` option set on it.

For instance, to set a flat rate of 20 with conditional free shipping of orders above 4.99 for the US, the following rates should be sent to the `POST mc/shipping/rates/batch` API:

```JSON
{
  "rates": [
    {
        "country": "US",
        "method": "flat_rate",
        "currency": "USD",
        "rate": 20,
        "options": {}
    },
    {
        "country": "US",
        "method": "free_shipping",
        "currency": "USD",
        "rate": 0,
        "options": {
            "free_shipping_threshold": 4.99
        }
    }
  ]
}
```

### Changelog entry

> Add - Get shipping rates suggestions for provinces/states and postal codes

### DIFF
--- a/src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php
+++ b/src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php
@@ -8,6 +8,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\ShippingRat
 use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
 use Automattic\WooCommerce\GoogleListingsAndAds\Internal\Interfaces\ISO3166AwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
 use Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ShippingZone;
 use WP_REST_Request as Request;
 
@@ -37,14 +38,21 @@ class ShippingRateSuggestionsController extends BaseController implements ISO316
 	protected $shipping_zone;
 
 	/**
+	 * @var WC
+	 */
+	protected $wc;
+
+	/**
 	 * ShippingRateSuggestionsController constructor.
 	 *
 	 * @param RESTServer   $server
 	 * @param ShippingZone $shipping_zone
+	 * @param WC           $wc
 	 */
-	public function __construct( RESTServer $server, ShippingZone $shipping_zone ) {
+	public function __construct( RESTServer $server, ShippingZone $shipping_zone, WC $wc ) {
 		parent::__construct( $server );
 		$this->shipping_zone = $shipping_zone;
+		$this->wc            = $wc;
 	}
 
 	/**
@@ -89,16 +97,22 @@ class ShippingRateSuggestionsController extends BaseController implements ISO316
 			$country_codes = $request->get_param( 'country_codes' );
 			$rates_output  = [];
 			foreach ( $country_codes as $country_code ) {
-				$suggestions = $this->shipping_zone->get_shipping_rates_for_country( $country_code );
+				$suggestions = $this->shipping_zone->get_shipping_rates_grouped_by_country( $country_code );
 
 				// Prepare the output.
 				$suggestions = array_map(
-					function ( $rate ) use ( $request ) {
-						// Remove the shipping class rates from the response because they are not needed.
-						if ( isset( $rate['options']['shipping_class_rates'] ) ) {
-							unset( $rate['options']['shipping_class_rates'] );
+					function ( $location_rate ) use ( $request ) {
+						$item = $location_rate->jsonSerialize();
+
+						// Add the store currency to each rate.
+						$item['currency'] = $this->wc->get_woocommerce_currency();
+
+						// Set the rate to 0 if it is not set.
+						if ( ! isset( $item['rate'] ) ) {
+							$item['rate'] = 0;
 						}
-						$response = $this->prepare_item_for_response( $rate, $request );
+
+						$response = $this->prepare_item_for_response( $item, $request );
 
 						return $this->prepare_response_for_collection( $response );
 					},
--- a/src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php
+++ b/src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php
@@ -109,7 +109,7 @@ class ShippingRateSuggestionsController extends BaseController implements ISO316
 
 						// Set the rate to 0 if it is not set.
 						if ( ! isset( $item['rate'] ) ) {
-							$item['rate'] = 0;
+							$item['rate'] = 0.0;
 						}
 
 						$response = $this->prepare_item_for_response( $item, $request );
--- a/src/Shipping/ShippingZone.php
+++ b/src/Shipping/ShippingZone.php
@@ -87,13 +87,12 @@ class ShippingZone implements Service {
 	public function get_shipping_rates_for_country( string $country_code ): array {
 		$this->parse_shipping_zones();
 
-		$location_rates = $this->location_rates[ $country_code ];
-		if ( empty( $location_rates ) ) {
+		if ( empty( $this->location_rates[ $country_code ] ) ) {
 			return [];
 		}
 
 		// Process the rates for each country subdivision separately.
-		$location_rates = array_map( [ $this->rates_processor, 'process' ], $location_rates );
+		$location_rates = array_map( [ $this->rates_processor, 'process' ], $this->location_rates[ $country_code ] );
 
 		// Convert the string array keys to integers.
 		$country_rates = array_values( $location_rates );
@@ -117,13 +116,12 @@ class ShippingZone implements Service {
 	public function get_shipping_rates_grouped_by_country( string $country_code ): array {
 		$this->parse_shipping_zones();
 
-		$location_rates = $this->location_rates[ $country_code ];
-		if ( empty( $location_rates ) ) {
+		if ( empty( $this->location_rates[ $country_code ] ) ) {
 			return [];
 		}
 
 		// Convert the string array keys to integers.
-		$country_rates = array_values( $location_rates );
+		$country_rates = array_values( $this->location_rates[ $country_code ] );
 
 		// Flatten and merge the country shipping rates.
 		$country_rates = array_merge( [], ...$country_rates );
--- a/src/Shipping/ShippingZone.php
+++ b/src/Shipping/ShippingZone.php
@@ -3,11 +3,8 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Shipping;
 
-use Automattic\WooCommerce\GoogleListingsAndAds\Google\GoogleHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
-use WC_Shipping_Method;
-use WC_Shipping_Zone;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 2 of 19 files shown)


src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php
```diff
@@ -89,16 +97,22 @@ protected function get_suggestions_callback(): callable {
...
-							unset( $rate['options']['shipping_class_rates'] );
+					function ( $location_rate ) use ( $request ) {
+						$item = $location_rate->jsonSerialize();
+
+						// Add the store currency to each rate.
+						$item['currency'] = $this->wc->get_woocommerce_currency();
+
+						// Set the rate to 0 if it is not set.
+						if ( ! isset( $item['rate'] ) ) {
+							$item['rate'] = 0;
```
mikkamp: In most locations I'm seeing rate being treated as a float, here I'm seeing rate get set to 0 which isn't a problem if it's later handled as a float. But in `Settings:update_shipping_rates` there is a strict comparison `0 !== $rate` how can I confirm that hasn't been translated to a float?

mikkamp (commented):
Thanks for these changes, initially looking at the code everything seems to be intact, and the included unit tests run without issues.

I'm not quite sure about the manual tests though, as it seems most of it still returns it in the old format. For the suggestion controller you mentioned:
> Note that the rates will be grouped at a country level.

What kind of grouping should I expect to see? If I have shipping for multiple states I'm still only getting one for the country and I get flat rate and free shipping as separate rates, with no grouping:
```json
	{
		"country": "US",
		"method": "flat_rate",
		"currency": "EUR",
		"rate": 18,
		"options": {}
	},
	{
		"country": "US",
		"method": "free_shipping",
		"currency": "EUR",
		"rate": 0,
		"options": {
			"free_shipping_threshold": 10
		}
	}
```

I also tested `get_shipping_rates_for_country` but for some postcodes I'm getting the literal wildcards I entered:
```php
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\Location)#17153 (3) {
      ["country":protected]=>
      string(2) "US"
      ["state":protected]=>
      NULL
      ["postcodes":protected]=>
      array(3) {
        [1]=>
        string(13) "99501...99950"
        [2]=>
        string(5) "1234*"
        [3]=>
        string(5) "23456"
      }
```
Is that expected, are those supported in the same way by Google?
If we are supporting those, them I'm missing that in the unit tests.

Testing with a flat rate using a calculation `5 + (10 * [qty])` gets ignored completely, I assume that's the correct behaviour?

Let me know if I should be testing any of that differently, otherwise I think it should be pretty much good to go.

src/Shipping/ShippingZone.php
```diff
@@ -59,423 +72,112 @@ public function __construct( WC $wc, GoogleHelper $google_helper ) {
...
-				'options'  => [],
-			];
-			if ( isset( $methods[0]['options']['min_amount'] ) ) {
-				$rate['options']['free_shipping_threshold'] = $methods[0]['options']['min_amount'];
-			}
+	public function get_shipping_rates_grouped_by_country( string $country_code ): array {
+		$this->parse_shipping_zones();
 
-			return [ $rate ];
+		$location_rates = $this->location_rates[ $country_code ];
```
mikkamp: When requesting the suggestions from the API I'm getting a notice on this line:
```
PHP Notice:  Undefined index: NL in wp-content/plugins/google-listings-and-ads/src/Shipping/ShippingZone.php on line 120
```

NL has a shipping zone and a rate, but it's skipped because it's `10 * [qty]`.

nima-karimi: This should be fixed by https://github.com/woocommerce/google-listings-and-ads/pull/1341/commits/21ad55f085c0d60ff1e0a357d12a3fdc833411fe

src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php
nima-karimi: I've added two commits to fix these. This rate should be set to a float. And the comparison is now changed to "> 0" so that would avoid any type checking issues. In general, the rate is always returned as a float but since we're passing them around as arrays it's not possible to type check them everywhere. I aim to change this in the next PR by using the value objects introduced in this PR.

nima-karimi:
> What kind of grouping should I expect to see? If I have shipping for multiple states I'm still only getting one for the country and I get flat rate and free shipping as separate rates, with no grouping:

What I meant by grouping was that if there are multiple shipping rates of the **same method** for different states of a country we group them together and present them as one. So for example, if you had two free shipping options and two flat rate options for different states of the US, they will be grouped together using their **method** and you get only one "flat_rate" shipping and one "free_shipping".

> I also tested get_shipping_rates_for_country but for some postcodes I'm getting the literal wildcards I entered:
> Is that expected, are those supported in the same way by Google?

Yes, these are expected and are supported by Google. I don't think we need to test them separately because we don't have any logic applied to them. We just get the postcodes from WC and pass them to Google. And that part is tested in unit tests.

> Testing with a flat rate using a calculation 5 + (10 * [qty]) gets ignored completely, I assume that's the correct behaviour?

Yes, we can't sync any shipping methods that include a calculation or variable expressions because Google does not support these. Since we are guesstimating the shipping rates, I decided it'd be best to ignore these completely.


Thanks for the review, @mikkamp 🙏 This is ready for another go.

mikkamp (approved):
Thanks for the clarifications, I'm getting the results I'd expect from the function `get_shipping_rates_for_country`:

```php
array(4) {
  [0]=>
  object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\LocationRate)#17096 (2) {
    ["location":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\Location)#17101 (3) {
      ["country":protected]=>
      string(2) "US"
      ["state":protected]=>
      string(2) "CA"
      ["postcodes":protected]=>
      array(0) {
      }
    }
    ["shipping_rate":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ShippingRateFlat)#17099 (2) {
      ["rate":protected]=>
      float(18)
      ["shipping_class_rates":protected]=>
      array(0) {
      }
    }
  }
  [1]=>
  object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\LocationRate)#17158 (2) {
    ["location":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\Location)#17140 (3) {
      ["country":protected]=>
      string(2) "US"
      ["state":protected]=>
      string(2) "FL"
      ["postcodes":protected]=>
      array(0) {
      }
    }
    ["shipping_rate":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ShippingRateFlat)#17156 (2) {
      ["rate":protected]=>
      float(12)
      ["shipping_class_rates":protected]=>
      array(0) {
      }
    }
  }
  [2]=>
  object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\LocationRate)#17159 (2) {
    ["location":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\Location)#17140 (3) {
      ["country":protected]=>
      string(2) "US"
      ["state":protected]=>
      string(2) "FL"
      ["postcodes":protected]=>
      array(0) {
      }
    }
    ["shipping_rate":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ShippingRateFree)#17157 (1) {
      ["threshold":protected]=>
      float(10)
    }
  }
  [3]=>
  object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\LocationRate)#17106 (2) {
    ["location":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\Location)#17153 (3) {
      ["country":protected]=>
      string(2) "US"
      ["state":protected]=>
      NULL
      ["postcodes":protected]=>
      array(3) {
        [1]=>
        string(13) "99501...99950"
        [2]=>
        string(5) "1234*"
        [3]=>
        string(5) "23456"
      }
    }
    ["shipping_rate":protected]=>
    object(Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ShippingRateFlat)#17152 (2) {
      ["rate":protected]=>
      float(10)
      ["shipping_class_rates":protected]=>
      array(2) {
        [0]=>
        array(2) {
          ["class"]=>
          string(5) "heavy"
          ["rate"]=>
          float(30)
        }
        [1]=>
        array(2) {
          ["class"]=>
          string(5) "light"
          ["rate"]=>
          float(18)
        }
      }
    }
  }
}
```

So it looks good to go, I think the PR is just missing a changelog entry to make the change tracking easier at release.

PR #1343, opened 2022-03-21
Fix the `woocommerce-admin` dependency for PHP unit tests

eason9487:

The `Automattic\WooCommerce\Admin\Install` namespace has been changed after [woocommerce-admin](https://github.com/woocommerce/woocommerce-admin) was merged into [woocommerce](https://github.com/woocommerce/woocommerce). This PR updates the namespace to fix the install processing of PHP unit tests.

Ref:
- https://github.com/woocommerce/woocommerce/blob/c4628b966be16218a99d093cf3672ad79a4f235e/plugins/woocommerce/src/Internal/Admin/Install.php#L6
- https://github.com/woocommerce/woocommerce-admin/pull/8317/files#diff-fe4f6d042955f294a9435960add587d284b232384dc9c5bee164316d501564a3R6

### Screenshots:

image

Go to the [Checks tab](https://github.com/woocommerce/google-listings-and-ads/pull/1343/checks) of this PR to see if all test cases in each PHP Unit Tests jobs were executed and passed.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -80,8 +80,8 @@ function install_woocommerce() {
 	WC_Install::install();
 
 	// Initialize the WC API extensions.
-	\Automattic\WooCommerce\Admin\Install::create_tables();
-	\Automattic\WooCommerce\Admin\Install::create_events();
+	\Automattic\WooCommerce\Internal\Admin\Install::create_tables();
+	\Automattic\WooCommerce\Internal\Admin\Install::create_events();
 
 	// Reload capabilities after install, see https://core.trac.wordpress.org/ticket/28374.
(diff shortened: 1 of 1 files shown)


message-dimke:
Fix works well!

As per @eason9487 from Slack: 

> Currently, since the test environment always uses the latest trunk version when woocommerce does not exist, backward compatibility is only needed during the conversion period for the present. However, if we will add the functionality to specify the WC version later, the compatibility process will be needed.

Still need not to forget when adjusting `./bin/install-wp-tests.sh` to add some checks for backward compatibility with versions below `WC 6.4.0-dev`

message-dimke (approved):

PR #1344, opened 2022-03-21
Returns continents data in get countries API if continents=true parameter is provided

ianlin:

Part of #1356 (original #1248), this PR updates the `GET mc/countries` API so it can return continents data when needed.

- Added a new query parameter `continents` for the API, if `continents=true` the API will include the continents data in the response.
- Updated the response format by adding a new first level `countries` and `continents` in the response JSON object, please check #1248 for the response example.

📔  Note: since the API response format has been changed, it requires the changes in the front end in order to make it work in the UI.

#### Get countries with the query parameter `continents` set to true or 1

mc-countries-continent

mc-countries-continent-1

#### Get countries without the query parameter `continents`

mc-countries-no-continent

#### Get countries with invalid query parameter `continents`

mc-countries-invalid-continent

Since the API response format has been changed, it requires the changes in the front end in order to make it testable through the UI. So using Postman for testing at this stage instead.

1. Request to `GET mc/countries`, the response data should only contain `countries` data.
2. Request to `GET mc/countries?continents=true` or `GET mc/countries?continents=1`, the response data should contain `countries` and `continents` data.
3. Request to `GET mc/countries?continents=` or `GET mc/countries?continents=xxx` or any other value, the 400 bad request with invalid parameter(s) error should be returned.

### Changelog entry

> Update - Respond continents data in get countries API if continents=true parameter is provided


src/API/Site/Controllers/MerchantCenter/SupportedCountriesController.php
```diff
@@ -70,17 +72,19 @@ public function register_routes(): void {
...
 	protected function get_countries_callback(): callable {
-		return function() {
-			return $this->get_supported_countries();
+		return function( Request $request ) {
+			return $this->get_supported_countries( $request );
 		};
 	}
 
 	/**
 	 * Get the array of supported countries.
```
mikkamp: We might want to update this description since it's no longer getting an array of countries.

src/API/Site/Controllers/MerchantCenter/SupportedCountriesController.php
```diff
@@ -70,17 +72,19 @@ public function register_routes(): void {
...
 	 *
+	 * @param Request $request
+	 *
 	 * @return array
 	 */
-	protected function get_supported_countries(): array {
+	protected function get_supported_countries( Request $request ): array {
```
mikkamp: In line with the previous comment, would it be wiser/cleaner to leave this function to only return a list of supported countries, and a separate function which returns the continents data?
That way the `get_countries` callback can determine what the response object will look like and whether it needs to include continents as well. If we are going to put it all in one function then I'd suggest to pass a boolean `$include_continents` to this function instead of the full request.

mikkamp (commented):
Thanks the changes work as expected.

Although the first question that came to mind was, if the intention of this controller is to return a list of supported countries, why do we return unsupported countries in the continent list? Are we leaving the filtering up to the UI? Should that be something we do in the backend instead? It wasn't quite clear how this API is to be consumed, but it might be something we should clarify.

I just left a few small comments, but it should be pretty much good to go, if we want to be reusing this controller for both countries and continents.

ianlin:
Hi @mikkamp, thanks for the good suggestions as always. I agree with you it's not consistent if we return a list of supported countries along with unsupported countries in the continent list. Here are some of the ideas @eason9487 and I discussed before:

1. Rather than returns everything in `GET mc/countries` API, add a new API `GET mc/continents` to get the list of continents
    ```js
    // GET mc/countries
    {
        "TW": {
            "name": "Taiwan",
            "currency": "TWD",
        }
    }

    // GET mc/continents
    {
        "AS": {
            "name": "Asia",
            "countries": [
                "JP",
                "TW",
                "CN", // unsupported country
            ]
        }
    }
    ```
    - Pros: It's clear API design-wise.
    - Cons 1: UI needs to call two APIs in order to get all the results, not an optimised solution performance-wise.
    - Cons 2: UI needs to implement the filtering logic in order to show the results.
2. Returns a list of supported countries with all countries in the continent list in `GET mc/countries` API (current implementation)
    ```js
    // GET mc/countries?continents=true
    {
        "countries": {
            "TW": {
                "name": "Taiwan",
                "currency": "TWD",
            }
        },
        "continents": {
            "AS": {
                "name": "Asia",
                "countries": [
                    "JP",
                    "TW",
                    "CN", // unsupported country
                ]
            }
        }
    }
    ```
    - Pros: Get everything in one API request.
    - Cons 1: Response data is not consistent. There are unsupported countries in the continent list.
    - Cons 2: UI needs to implement the filtering logic in order to show the results.
3. Returns a list of supported countries, and add a `continent` field under each country object. E.g.
    ```js
    // GET mc/countries?continents=true
    {
        "TW": {
            "name": "Taiwan",
            "currency": "TWD",
            "continent": "Asia"
        }
    }
    ```
    - Pros: It's clear API design-wise, and only return supported countries.
    - Cons 1: There would be multiple duplicate continent data.
    - Cons 2: UI needs to implement the filtering logic in order to show the results.


However, looking at the design (eQ9O4m2flzrcAiDMXkOW0m-fi-4141%3A86260) again, it seems the format of continents in example 1. and 2. without unsupported countries can fulfil the needs of UI 🤔. So what about we just create a new API `GET mc/continents` and returns the continent list with supported countries only? Which means the UI doesn't need to call the `GET mc/countries` API, and also might not need to have filtering logic in order to display the result. The API response example:

```js
// GET mc/continents
{
    "AS": {
        "name": "Asia",
        "countries": [
            "JP",
            "TW",
            // Only supported countries
        ]
    }
}
```

ianlin:
Just a side note that this PR can be release in a later version that includes TreeSelectController, so we have time to figure out the best solution for it. 😃 Ref: p1647932509127199/1647931181.455169-slack-CK365S85V

mikkamp:
If it only needs to get either a list of supported countries or a list of continents (including supported countries), then I'd say the cleaner approach would be to create a new endpoint (we can always deprecate/remove the supported countries endpoint later if needed). Does this new endpoint need each country name / currency as well?

If it does need both lists separated in one request then I'd suggest to go for your option 2 but just filter each set of countries. We already have the set of supported countries, so it wouldn't take much to filter the list in each continent (with an `array_intersect`).

ianlin:
You're right @mikkamp, the UI still needs each country name / currency so the option 2 with only supported countries is better. I will update it tomorrow, thanks!

mikkamp:
Here is another [helper function](https://github.com/woocommerce/google-listings-and-ads/pull/1341/files#diff-e5e54e70e3677013fec896bd6a8490847bc64b414e5f573c5945804d85733b95R120) for shipping. Maybe a shared function could be put in GoogleHelper.

ianlin:
Hey @mikkamp thanks for all the suggestions, I've refactor the code in 9f6ddc6 to make the continents field only contains supported country.

> Here is another [helper function](https://github.com/woocommerce/google-listings-and-ads/pull/1341/files#diff-e5e54e70e3677013fec896bd6a8490847bc64b414e5f573c5945804d85733b95R120) for shipping. Maybe a shared function could be put in GoogleHelper.

I've added a TODO comment in the code and a followup [issue](https://github.com/woocommerce/google-listings-and-ads/issues/1351) as well.

Lastly, the PHP unit tests failed because there are issues when we run tests using WC trunk branch. In this [commit](https://github.com/woocommerce/woocommerce/commit/3dd60322d5490d7d86ad33e7d732f276df73a12d#diff-85dbbe55dbabeeef038b8db070b9859f9fd7a69135d89cdd2844e2814b31d605R77) it update the import path of feature-config.php from `/includes/feature-config.php` to `/includes/react-admin/feature-config.php` but in this [PR](https://github.com/woocommerce/woocommerce/pull/32304) it removes the file `/includes/react-admin/feature-config.php`.

I tried to use the stable tag `6.3.1` to make sure the tests pass locally, ref: p1648024673352789/1647844210.846629-slack-CK365S85V

mikkamp (approved):
Thanks, that looks good. There is a fix for the unit tests available [here](https://github.com/woocommerce/google-listings-and-ads/pull/1352), but they pass locally so we can ignore them for this PR.

The only thing I noticed is that the unit test doesn't cover removing a continent when there are no supported countries, but that's probably fine to leave as is, if we are moving it to GoogleHelper, as it can be tested there.

ianlin:
Thanks for reviewing @mikkamp. I've merged the #1352 for unit tests so the tests passed.

> unit test doesn't cover removing a continent when there are no supported countries

Thanks! Added a note in #1351.

Issue #1333, opened 2022-03-17
Bump wp-scripts to 22.1.0 and webpack to v5, and enhance webpack related usages


eason9487:
## User story

As a developer, it'd be great

- if there were more productive development tools.
- if we could reduce potential security concerns from dependent packages.
- If we could clear some temporary programs that can be resolved along with the dependent package upgrades.

### Describe the solution you'd like 

- Upgrade `webpack` version to v5.x, which is used by `@wordpress/scripts` since v18.0.1.
- Introduce **Fast Refresh** for React, which is supported in `webpack` v5 via `@wordpress/scripts` v20.0.1+. So that the frontend development experience should be improved.
   - About the Fast Refresh: https://github.com/pmmmwh/react-refresh-webpack-plugin
   - Development only
- Clean up temporary solutions that are no longer needed.

## Technical

Shared branch: `tweak/1333-wp-scripts-webpack`

### Stage 1 - Upgrade dependencies

- [x] Upgrade `@wordpress/scripts` version to 22.1.0 and other packages that depend on it.
- [x] Upgrade webpack config version to v5.x.
- [x] Resolve breaking changes.
- [x] Resolve linting errors.

### Stage 2 - Fast Refresh

- [x] Introduce **Fast Refresh** for React.

### Stage 3 - Clean up

- [x] SVG assets resolving and exporting.
- [x] Clean up unneeded and deduplicate npm packages.
- [x] Revert 0167ec4fc531e89df0b607aec5f7e5957a7f09fe after getting approval and before merging.

Anything else?

eason9487:
📝  `@woocommerce/eslint-plugin` 2.0.0 was released a few days ago. Probably we could upgrade `eslint` and related packages as well.

PR #1345, opened 2022-03-21
Add the Fast Refresh for React development

eason9487:

It's the stage 2 of #1333

- Introduce **Fast Refresh** for React

📌  Before starting the code review, there may be one thing we have to decide on first. I'd like to know your thoughts about them. Please have a look at the next section.

###  Additional details:

This PR contains four implementations at different levels, which are inversely proportional between convenience and degree of monkey patch. 🙈 

1. **Level 1** - The minimum requirements to enable Fast Refresh
   - [Complete code diff](https://github.com/woocommerce/google-listings-and-ads/pull/1345/commits/69dbd3a3eb9d3ab5f1c991593af0101b8ddce52b)
   - It requires the enabled `SCRIPT_DEBUG` flag and activated Gutenberg, which replaces all `@wordpress/*` packages with the versions inside Gutenberg.
   - That is, the `@wordpress/*` packages used through DEWP will be changed to newer versions when developing. Usually, we want to stay with the lowest supported version to avoid misuse of features that don't exist yet.
   - Another problem is that when the development environment uses HTTPS, there is an error that occurs every time the page is reloaded. Fast Refresh works fine, but the error is slightly distracting.
      
2. **Level 2** - Prevent the first connection from failing if the development environment uses HTTPS
   - Compare to level 1, there is an [additional change](https://github.com/woocommerce/google-listings-and-ads/pull/1345/commits/28bc5c130ce8e907acfad33772961ac4c5246937) to replace the plugin.
3. **Level 3** - Enable it without the need for Gutenberg
   - [Complete code diff](https://github.com/woocommerce/google-listings-and-ads/pull/1345/files/f5c6622c6512b4d989edeb9c9ec652bfb6884c5f)
   - It's still compatible with the activated Gutenberg
   - It still has to switch the `SCRIPT_DEBUG` flag manually
   - It avoids the major concern of level 1.
4. **Level 4** - Enable it without the need for enabling the `SCRIPT_DEBUG` flag
   - Compare to level 3, it requires [another monkey patch](https://github.com/woocommerce/google-listings-and-ads/pull/1345/commits/8dfbeb2a6dced0096243cb9ab5c89bdb83f4499d) to use the development version of React and ReactDOM.
   - It's still compatible with the enabled `SCRIPT_DEBUG` flag
   - This is the most convenient way to use it but has the most code injections.
   - Lower maintainability in the future updates from `@wordpress/scripts`

#### 📹 Automatically update changed modules after rebuilding js and scss files

https://user-images.githubusercontent.com/17420811/159233409-eed70a4d-8fcb-42f2-ade5-5bd9d1ac4791.mp4

#### 📹 It works well with the state of React Hooks

https://user-images.githubusercontent.com/17420811/159679953-4df3bded-382f-4760-a36e-72e33b5c92f6.mp4

:warning: However, it is important to note that hooks that have a deps array such as `useEffect` may be triggered after Fast Refresh updating modules.

#### Fast Refresh
1. `npm run start:hot`
1. Switching all combinations of the enabled/disabled `SCRIPT_DEBUG` flag and the activated/deactivated Gutenberg. Check if the Fast Refresh works properly as in the above video.

#### The original watch mode
1. `npm start`
1. It should still work correctly as before.

### Changelog entry

> Add - Fast Refresh for React development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,7 @@ Now you can build the files using one of these commands:
 -   `npm run build` : Build a production version
 -   `npm run dev` : Build a development version
 -   `npm run start` : Build a development version, watch files for changes
--   `npm run start:hot` : Build a development version in Fast Refresh mode, watch files for changes. To use this mode, it requires
-    1. The [`SCRIPT_DEBUG`](https://wordpress.org/support/article/debugging-in-wordpress/#script_debug) flag enabled.
+-   `npm run start:hot` : Build a development version in Fast Refresh mode, watch files for changes.
 
 ## Helper Scripts
 
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ Now you can build the files using one of these commands:
 -   `npm run build` : Build a production version
 -   `npm run dev` : Build a development version
 -   `npm run start` : Build a development version, watch files for changes
+-   `npm run start:hot` : Build a development version in Fast Refresh mode, watch files for changes. To use this mode, it requires
+    1. The [`SCRIPT_DEBUG`](https://wordpress.org/support/article/debugging-in-wordpress/#script_debug) flag enabled.
+    1. The [Gutenberg](https://wordpress.org/plugins/gutenberg/) plugin activated.
 
 ## Helper Scripts
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ Now you can build the files using one of these commands:
 -   `npm run start` : Build a development version, watch files for changes
 -   `npm run start:hot` : Build a development version in Fast Refresh mode, watch files for changes. To use this mode, it requires
     1. The [`SCRIPT_DEBUG`](https://wordpress.org/support/article/debugging-in-wordpress/#script_debug) flag enabled.
-    1. The [Gutenberg](https://wordpress.org/plugins/gutenberg/) plugin activated.
 
 ## Helper Scripts
 
--- a/src/Admin/Admin.php
+++ b/src/Admin/Admin.php
@@ -86,10 +86,11 @@ class Admin implements Service, Registerable, Conditional {
 		);
 
 		add_action(
-			'admin_enqueue_scripts',
-			function() {
-				$this->enqueue_webpack_runtime_for_dev();
-			}
+			'wp_default_scripts',
+			function( $scripts ) {
+				$this->inject_fast_refresh_for_dev( $scripts );
+			},
+			20
 		);
 	}
 
@@ -243,24 +244,48 @@ class Admin implements Service, Registerable, Conditional {
 	 *
 	 * The runtime.js file is created when the front-end is developed in Fast Refresh mode
 	 * and must be loaded together to enable the mode.
+	 *
+	 * When Gutenberg is not installed or not activated, the react dependency will not have
+	 * the 'wp-react-refresh-entry' handle, so here injects the Fast Refresh scripts we built.
+	 *
+	 * @param WP_Scripts $scripts WP_Scripts instance.
 	 */
-	private function enqueue_webpack_runtime_for_dev() {
+	private function inject_fast_refresh_for_dev( $scripts ) {
 		if ( ! ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ) {
 			return;
 		}
 
-		$file_path = "{$this->get_root_dir()}/js/build/runtime.js";
-
-		if ( ! file_exists( $file_path ) ) {
+		if ( ! file_exists( "{$this->get_root_dir()}/js/build/runtime.js" ) ) {
 			return;
 		}
 
-		wp_enqueue_script(
+		$react_script = $scripts->query( 'react', 'registered' );
+
+		if ( ! $react_script ) {
+			return;
+		}
+
+		$plugin_url = $this->get_plugin_url();
+
+		$scripts->add(
 			'gla-webpack-rumtime',
-			"{$this->get_plugin_url()}/js/build/runtime.js",
-			[],
-			(string) filemtime( $file_path ),
-			false
+			"{$plugin_url}/js/build/runtime.js",
+			[]
 		);
+		$react_script->deps[] = 'gla-webpack-rumtime';
+
+		if ( ! in_array( 'wp-react-refresh-entry', $react_script->deps, true ) ) {
+			$scripts->add(
+				'wp-react-refresh-runtime',
+				"{$plugin_url}/js/build-dev/react-refresh-runtime.js",
+				[]
+			);
+			$scripts->add(
+				'wp-react-refresh-entry',
+				"{$plugin_url}/js/build-dev/react-refresh-entry.js",
+				[ 'wp-react-refresh-runtime' ]
+			);
+			$react_script->deps[] = 'wp-react-refresh-entry';
+		}
 	}
 }
--- a/src/Admin/Admin.php
+++ b/src/Admin/Admin.php
@@ -84,6 +84,13 @@ class Admin implements Service, Registerable, Conditional {
 				return $this->add_plugin_links( $links );
 			}
 		);
+
+		add_action(
+			'admin_enqueue_scripts',
+			function() {
+				$this->enqueue_webpack_runtime_for_dev();
+			}
+		);
 	}
 
 	/**
@@ -229,5 +236,31 @@ class Admin implements Service, Registerable, Conditional {
 	 */
 	protected function enableReports(): bool {
 		return apply_filters( 'woocommerce_gla_enable_reports', true );
+	}
+
+	/**
+	 * This method is ONLY used during development.
+	 *
+	 * The runtime.js file is created when the front-end is developed in Fast Refresh mode
+	 * and must be loaded together to enable the mode.
+	 */
+	private function enqueue_webpack_runtime_for_dev() {
+		if ( ! ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) ) {
+			return;
+		}
+
+		$file_path = "{$this->get_root_dir()}/js/build/runtime.js";
+
+		if ( ! file_exists( $file_path ) ) {
+			return;
+		}
+
+		wp_enqueue_script(
+			'gla-webpack-rumtime',
+			"{$this->get_plugin_url()}/js/build/runtime.js",
+			[],
+			(string) filemtime( $file_path ),
+			false
+		);
 	}
 }
--- a/src/Admin/Admin.php
+++ b/src/Admin/Admin.php
@@ -248,13 +248,12 @@ class Admin implements Service, Registerable, Conditional {
 	 * When Gutenberg is not installed or not activated, the react dependency will not have
 	 * the 'wp-react-refresh-entry' handle, so here injects the Fast Refresh scripts we built.
 	 *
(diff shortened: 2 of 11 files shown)


puntope:
> 📌 Before starting the code review, there may be one thing we have to decide on first. I'd like to know your thoughts about them. Please have a look at the next section.

I will do 4. My preference is to pay the tax of adding a bit of code in exchange of a smooth development experience. Besides that I don't think they are huge changes in the code that make me consider bad to go for solution 4. 

The only doubts I have is with SCRIPT_DEBUG. Feels safer for me to use it even tho we are doing all the logic only if `runtime.js` is loaded and supposedly that happens only in fastRefresh mode, so my opinion is still 4.

README.md
```diff
@@ -60,6 +60,7 @@ Now you can build the files using one of these commands:
 -   `npm run build` : Build a production version
 -   `npm run dev` : Build a development version
 -   `npm run start` : Build a development version, watch files for changes
+-   `npm run start:hot` : Build a development version in Fast Refresh mode, watch files for changes.
```
puntope, line 64: 💅  Another option to consider could be `npm run watch`

webpack.config.js
```diff
@@ -1,8 +1,13 @@
 const webpack = require( 'webpack' );
 const defaultConfig = require( '@wordpress/scripts/config/webpack.config' );
+const { hasArgInCLI } = require( '@wordpress/scripts/utils' );
```
puntope, line 3: ❓ Maybe dumb question. Not a super expert in Webpack. But why we are loading all this logic in `webpack.config.js` if we are using `webpack-development.config.js`?

webpack.config.js
```diff
@@ -94,12 +99,15 @@ const webpackConfig = {
...
+				'ReactRefreshPlugin',
 			];
 			return ! filteredPlugins.includes( plugin.constructor.name );
 		} ),
 		new DependencyExtractionWebpackPlugin( {
 			injectPolyfill: true,
-			externalizedReport: '../../.externalized.json',
+			externalizedReport: hasReactFastRefresh
+				? false
+				: '../../.externalized.json',
```
puntope: 💅 

Potentially replaceable by 

`externalizedReport: ! hasReactFastRefresh && '../../.externalized.json',`

webpack.config.js
```diff
@@ -117,7 +125,15 @@ const webpackConfig = {
...
-	],
+		/**
+		 * If the development environment uses HTTPS,
+		 * it will fail when first connecting to webpack dev server,
+		 * so turn off the `overlay` option here.
+		 * Ref: https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/v0.5.4/docs/TROUBLESHOOTING.md#component-not-updating-with-bundle-splitting-techniques
+		 */
+		hasReactFastRefresh &&
+			new ReactRefreshWebpackPlugin( { overlay: false } ),
+	].filter( Boolean ),
```
puntope: 💅 

I'd add a comment somewhere clarifying this filter(Boolean) is for removing falsy items because maybe `ReactRefreshWebpackPlugin` is not loaded.

 Notice other option could be to use spread. 
 
 ```
requestToHandle,
} ),
/**
		 * Automatic polyfills for native node.js modules were removed from webpack v5.
		 * And `@wordpress/components` below version 16.x depends on `@wordpress/compose`,
		 * which directly accesses `process.env`.
		 * Ref: https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4012.0.8/packages/compose/src/hooks/use-reduced-motion/index.js#L21
		 *
		 * So the fallback is required here.
		 * It may be possible to remove this fallback
		 * when `@wordpress/components` is upgraded above 17.0.0+,
		 * or when it is removed from the `requestToExternal` function above.
		 */
...(hasReactFastRefresh ? [] : [new ReactRefreshWebpackPlugin( { overlay: false } ) ])
 ],
 ``` 

Or even more reusable

```
requestToHandle,
} ),
...reactFastRefreshPlugins
]
```

```
const reactFastRefreshPlugins = () => {
    if (! hasReactFastRefresh) return [];
    
    // Enabled plugins with hasReactFastRefresh go here
    return [
        new ReactRefreshWebpackPlugin( { overlay: false } ),
    ]
}
```

src/Admin/Admin.php
```diff
@@ -230,4 +238,59 @@ public function get_view( string $view, array $context_variables = [] ): string
...
+	 * When Gutenberg is not installed or not activated, the react dependency will not have
+	 * the 'wp-react-refresh-entry' handle, so here injects the Fast Refresh scripts we built.
+	 *
+	 * The Fast Refresh also needs the development version of React and ReactDOM.
+	 * They will be replaced if the SCRIPT_DEBUG flag is not enabled.
+	 *
+	 * @param WP_Scripts $scripts WP_Scripts instance.
+	 */
+	private function inject_fast_refresh_for_dev( $scripts ) {
+		if ( ! file_exists( "{$this->get_root_dir()}/js/build/runtime.js" ) ) {
```
puntope, line 257: ❓  Why here we use `{$this->get_root_dir()}/js/build/runtime.js`. but in line 277 we use `{$plugin_url}/js/build/runtime.js` it's a different file? or just two ways to get the same thing?

src/Admin/Admin.php
```diff
@@ -230,4 +238,59 @@ public function get_view( string $view, array $context_variables = [] ): string
...
+			return;
+		}
+
+		$react_script = $scripts->query( 'react', 'registered' );
+
+		if ( ! $react_script ) {
```
puntope, line 263: ❓  Shouldn't be also react-dom presence a condition to abort?

src/Admin/Admin.php
```diff
@@ -84,6 +84,14 @@ function( $links ) {
 				return $this->add_plugin_links( $links );
 			}
 		);
+
+		add_action(
+			'wp_default_scripts',
+			function( $scripts ) {
+				$this->inject_fast_refresh_for_dev( $scripts );
+			},
+			20
```
puntope, line 93: ❓  Why 20?

puntope:
❓ 

When I run it using https://wp16.test/ (my host name) it works like a charm.

However when I run `npm run start:hot` and enter into NGROK url is not working.

puntope (commented):
Thanks for this implementation 👌 

Some stuff to discuss before approval ❓/ 💅

README.md
eason9487: `npm run start` is also running on watch mode. I think it would be better to have the meaning of fast refresh if want to use another script name.

webpack.config.js
eason9487: Sorry, I do not quite understand this question. Could you provide more context?

src/Admin/Admin.php
eason9487: The `file_exists` checks whether a file exists via the file path in the filesystem, and the enqueue script function will let the browser load the script file via the server URL. So they are the same file but need different URIs.

src/Admin/Admin.php
eason9487: Assumed they are always used together in the webpage scope, so I think it should be ok to only check one of them. Ref: https://github.com/WordPress/wordpress-develop/blob/5.6.8/src/wp-includes/script-loader.php#L83-L84

src/Admin/Admin.php
eason9487: Gutenberg uses the default priority (10) to register the vendor scripts of Fast Refresh, so here uses a lower priority to ensure the action hook will be called after Gutenberg. Thus, it's able to check if the vendor scripts of Fast Refresh are already registered by Gutenberg. Ref: https://github.com/WordPress/gutenberg/blob/%40wordpress/scripts%4022.1.0/lib/client-assets.php#L203-L221

webpack.config.js
eason9487: Changed in https://github.com/woocommerce/google-listings-and-ads/pull/1345/commits/28ec1ca3e9b345021675849ec61f763fccbbb15c.

webpack.config.js
eason9487: I took another way in https://github.com/woocommerce/google-listings-and-ads/pull/1345/commits/7421728b46738229945c010ee5ba12a891c43e5f. Please have a look.

eason9487:

> When I run it using https://wp16.test/ (my host name) it works like a charm.
> However when I run `npm run start:hot` and enter into NGROK url is not working.

When the Fast Refresh is enabled, it will create a WebSocket connection for communicating event messages, and it's connected by the localhost hostname. So it cannot be used through ngrok. The original **livereload** also uses the same mechanism and has the same situation.

image

Even though the hostname can be changed by patching the [webpack dev server config](https://github.com/WordPress/gutenberg/blob/%40wordpress/scripts%4022.1.0/packages/scripts/config/webpack.config.js#L268-L274), it still needs to forward the incoming WebSocket connection with 8887 port to your localhost:8887 upon ngrok. Probably it would have much more extra tuning to make it work, I guess.

I would suggest only using Fast Fresh when developing locally.

webpack.config.js
puntope: Yes, so, we have `webpack.config.js` in which we have all our project config. There was some code added there in this PR. 

Then I see in package.json we use `"start:hot": "npm run dev -- --config ./webpack-development.config.js && npm start -- --hot --allowed-hosts all",`

We call `./webpack-development.config.js` however we still loaded code in `webpack.config.js`. 

My doubt is why we are loading new code regarding fast refresh in `webpack.config.js` if we are actually loading `./webpack-development.config.js` for that.

webpack.config.js
puntope: Looking good, thanks

puntope (approved):
✅  LGTM

- I tested changing SCSS, and JS in an out useEffect contexts and the page automatically refreshed. 
- I tested the plain `npm run start` and works as before
- I could run lint:js, lint:css and test:js successfully

webpack.config.js
eason9487: The React Fast Refresh is supposed to be bundled with vendor libraries together and initialized/loaded **before** React. But upon the WP architecture, vendor libraries are bundled separately and maybe loaded via DEWP, so the React Fast Refresh needs to be enabled in an [externalizing way](https://github.com/pmmmwh/react-refresh-webpack-plugin/blob/main/docs/TROUBLESHOOTING.md#externalising-react).

It's the same In GLA, so the `webpack-development.config.js` is for building the React Fast Refresh code externally and also mixing the building flow of DEWP, and then using the js dependency and loading mechanism of WP to ensure they will be loaded before React.

While it is possible to create separate js files by adding those `entry` to `webpack.config.js`, sharing the `plugins` of webpack config will make the React Fast Refresh files unable to use different DEWP initial settings.

Exporting two webpack configs as an array from `webpack.config.js` will cause the built React Fast Refresh js files to be deleted by [another webpack plugin](https://github.com/WordPress/gutenberg/blob/%40wordpress/scripts%4022.1.0/packages/scripts/config/webpack.config.js#L226-L231), `CleanWebpackPlugin`, after saving GLA source code changes because the plugins are not shared between the two webpack configs. The root cause of the unexpected deletion with `CleanWebpackPlugin` can be found in https://github.com/woocommerce/pinterest-for-woocommerce/issues/330.

About the `ReactRefreshWebpackPlugin`, it needs to work with the app codes, so it's placed in the `webpack.config.js`.

PR #1349, opened 2022-03-22
Tree Select Component - Keyboard navigation with Search 

puntope:

- This PR implements the Keyboard navigation also when the search filter is applied. 
- This PR improves the SPACE key down using the default handler based on @eason9487 [suggestion](https://github.com/woocommerce/google-listings-and-ads/pull/1336#discussion_r831787504) 

### Screenshots:

https://user-images.githubusercontent.com/5908855/159465552-2776dc1b-43dd-4781-886b-becd93cda689.mov

1.   Checkout this branch
2.  Go to `wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fcomponent-test` 
3.  Check that the normal keyboard navigation is working  (navigate up and down over the elements, start from the search control and also in a specific element)
4. Insert some search and check that it's working the same
5.  Check that the SPACE selector is working and selects the Nodes
6. Check with and without root node

### Aditional details

The work was mainly to update `getPreviousOption` and `getNextOption` in order to check if actually, the found option has its ref available.  We established an extra pointer in order to iterate over the next/prev nodes if the found node is not visible.

### DIFF
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -229,32 +229,70 @@ const TreeSelectControl = ( {
 	 * if there is no more options, it tries to get the next option on the upper level.
 	 *
 	 * @param {Option} option TThe reference option to get the next element
+	 * @param increaseLevel
+	 * @param position
+	 * @param offset
+	 * @param checkExpanded
 	 * @return {Object} The next available option in the tree
 	 */
-	const getNextOption = ( option ) => {
+	const getNextOption = ( option, offset = 0, checkExpanded = true ) => {
+		// if no option is provided means we are focused in the search control.
+		// Hence the element to focus is the first
 		if ( ! option ) {
 			return getOptionFromRepository( filteredOptions[ 0 ] );
 		}
 
+		// if the option is expanded (or is the root)...
 		if (
-			nodesExpanded.includes( focused.value ) ||
-			focused.value === ROOT_VALUE
+			checkExpanded &&
+			( nodesExpanded.includes( option.value ) ||
+				option.value === ROOT_VALUE ||
+				( hasChildren( option ) && !! filter ) )
 		) {
-			return getOptionFromRepository( focused.children[ 0 ] );
+			// if there are still elements to check in the children
+			if ( offset < option.children.length ) {
+				const nextEl = getOptionFromRepository(
+					option.children[ offset ]
+				);
+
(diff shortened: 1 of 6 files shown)


js/src/components/tree-select-control/options.js
```diff
@@ -123,8 +109,6 @@ const Options = ( {
 			toggleExpanded( option );
 		} else if ( event.key === ARROW_LEFT && isExpanded ) {
 			toggleExpanded( option );
-		} else if ( event.key === SPACE ) {
-			onChange( event.target.checked, option );
```
eason9487, line 127: 💅 
Its code comment could be removed as well.
https://github.com/woocommerce/google-listings-and-ads/blob/539ed3d954ec042dec1ef6f5e4975510cb123957/js/src/components/tree-select-control/options.js#L101

js/src/components/tree-select-control/index.js
```diff
@@ -338,19 +426,20 @@ const TreeSelectControl = ( {
...
 
 		if ( ARROW_DOWN === event.key ) {
+			if ( ! filteredOptions.length ) return;
 			setFocused( getNextOption( focused ) );
 			event.preventDefault();
 		}
 
 		if ( BACKSPACE === event.key ) {
+			if ( inputControlValue ) return;
 			onChange( value.slice( 0, -1 ) );
```
If the focus is on a checkbox, the behavior of pressing the Backspace key is different when searching and non-searching.

Kapture 2022-03-23 at 15 49 45

eason9487 (commented):
💅 
It isn't a problem introduced by this PR, but perhaps it can be fixed by the way.
Kapture 2022-03-23 at 15 43 29

js/src/components/tree-select-control/index.js
puntope: > ❓ If the focus is on a checkbox, the behavior of pressing the Backspace key is different when searching and non-searching.
> 
> Kapture 2022-03-23 at 15 49 45

Yes, this is because when is there a search text the behaviour should be to delete text instead of deleting the tags. However, I removed the keyDown for that in the Parent component and move it to the Control component.

puntope:
Howdy @eason9487  Thanks for the review! I updated it with your suggestions. Thanks

eason9487 (approved):
Thanks for the adjustments! The keyboard navigation functionality works correctly with all shapes of trees I can think of. 👍

PR #1352, opened 2022-03-23
Fix unit test to work with the latest stable version of WooCommerce

mikkamp:

WooCommerce Admin is being moved into the WooCommerce monorepo repository. Since we were basing our unit tests on the trunk branch this was causing some failures.

This PR changes the install script to support a WC version being passed to the script, by default this will be `latest` which fetches the latest stable version from WP.org

In addition to that the WP test matrix was modified to test latest and L-2.

1. Setup the test dependencies as instructed in the [readme ](https://github.com/woocommerce/google-listings-and-ads/blob/develop/README.md#install-test-dependencies): `./bin/install-wp-tests.sh gla_test root root localhost`
2. Check that the latest version is installed in `/tmp/wordpress/wp-content/plugins/woocommerce` (there should be a generated file `version-6.3.1`)
3. Run the unit tests `vendor/bin/phpunit` and confirm there are no errors or warnings
4. Setup the dependencies using a newer version of WC (6.4) `./bin/install-wp-tests.sh gla_test root root localhost latest release/6.4`
5. Check that the specified version is installed in `/tmp/wordpress/wp-content/plugins/woocommerce` (there should be a generated file `version-release-6.4`)
6. Run the unit tests `vendor/bin/phpunit` and confirm there are no errors or warnings

*Note: If we pass the WC_VERSION as trunk the unit tests will still fail, but I think we can wait to support any of those changes until it's stable.*

### Changelog entry
* Fix - Unit tests use latest stable version of WooCommerce

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       matrix:
         php: [7.3, 7.4]
-        wp-version: [5.6, 5.7, 5.8]
+        wp-version: [5.7, 5.8, latest]
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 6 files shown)


budzanowski (approved):
Very nice!

PR #1295, opened 2022-03-07
Add type to Performance Card in order to differentiate between Paid a…

puntope:
### Changes proposed in this Pull Request:

Closes #1285 

We want to show different messages depending on Free or Paid Campaigns. Right now we are showing same message for both.

### Screenshots:

Screenshot 2022-03-07 at 18 27 26

Screenshot 2022-03-07 at 18 31 08

### Detailed test instructions:

1. You should have Paid Campaign set-up done. 
2. Go to the code on this branch in `js/src/dashboard/summary-section/index.js` 
3. In lines 66-67 and lines 34-35 set `data` as `false` and `loaded` as `true` to simulate no data error.
4. Go to the dashboard and see messages like the screenshots. Paid one shows text regarding google Ads, as well as link. Free one shows text regarding Google Merchant Center as well as the Link to Google Merchant centre dashboard. 
5. Roll back your test changes and see how the Paid and free summary renders as always. 

### Additional details:

- Created a new component CampaignNoData. It renders the message and button depending on the campaign type.  

### Changelog entry

> FIx - Show right link and message in Paid Campaigns report when there is no data available

PR #1355, opened 2022-03-23
Refactor Dashboard's `SummarySection`

tomalec:

There are three distinct changes in this PR:
1. [Remove `PerformanceCard`'s `numberOfItems`](https://github.com/woocommerce/google-listings-and-ads/commit/d25ba94c9b15373dc846f89457972f99e79023b2) unused property, YAGNI.
2. [Make SummarySection use SummaryCard directly,](https://github.com/woocommerce/google-listings-and-ads/commit/0f525544aa697f6b03fa17b8b7454c4a53f27fb6)  to separate responsibilities and reduce repetition.
3. [Make PerformanceCard render no date message given explicitely,](https://github.com/woocommerce/google-listings-and-ads/commit/e42dc562dfb9000c1fad5d7928ca0d94a10224ad) instead of hiding it inside `PerformanceCardNoData`.

   Keep the content for free vs. paid performance cards all together in `js/src/dashboard/summary-section/index.js`, making other components more content-agnostic.

    Addresses https://github.com/woocommerce/google-listings-and-ads/pull/1295#issuecomment-1062832949.

I think the first two are less opinionated, so if you don't like my refactoring proposal from 3. I'm happy with merging just 1. & 2.

The principle behind 3. is to keep logic and the content distinction between the free and paid cards in one place - `js/src/dashboard/summary-section/index.js` which defines `Free-` and `PaidPerformanceCard`. Then keep `PerformanceCard` pretty dumb component that only composes visually the data & content given, uniformly for both.

### Screenshots:
n/a visual effect should be exactly the same as before

1. Open `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard`
4. The page should behave as before
5. Review and run JS tests `npm run test:js`

### Changelog entry


js/src/dashboard/summary-section/performance-card.test.js
```diff
@@ -9,49 +9,25 @@ import { render } from '@testing-library/react';
 import PerformanceCard from '.~/dashboard/summary-section/performance-card';
```
puntope, line 9: 💅 

This one I guess We could delete it, since we are covering this in `summary-section.test.js`

puntope (approved):
✅  LGTM !

Thanks

js/src/dashboard/summary-section/performance-card.test.js
tomalec: To some extent yes. But `summary-section.test.js` tests that the specific texts are rendered, regardless of the component and implementation underneath. While this tests the `PerformanceCard` and its capabilities.

Currently, we have 1:1 relation between them. But as time flies we could use `PerformanceCard` somewhere else as well. Then for example change `SummarySection` so it will render the link directly, and boom we no longer have `PerformanceCard` covered by tests.

So it's unit vs. integration testing.

PR #1357, opened 2022-03-24
Allow setting free shipping rates (WIP)

### LINKED ISSUE #1332: Setup MC: batch saving shipping rate suggestions throws error when suggestions contain non-"flat-rate" method
### Describe the bug:

When users have a non-"flat-rate" method in their WooCommerce store settings, and they go to Setup MC Step 3 shipping rates section, they would see an error message popup saying "There was an error trying to save shipping rates."

When we look into the API calls, in shipping rates suggestions API (`GET /mc/shipping/rates/suggestions`), it returns the shipping rate with "free_shipping" method. Then, when we try to save the suggestions by calling upsert shipping rates API (`POST /mc/shipping/rates/batch`), it returns an error saying `rates: "rates[28][method] is not flat_rate."`.

Note that this issue does not happen when we have a shipping zone with multiple methods containing flat_rate.

Shipping zone with only free shipping method:

image

Shipping rates suggestions contain the shipping rate with free shipping method:

image

Trying to save the suggestions returns error:

image

📹 Demo video with my voice:

https://user-images.githubusercontent.com/417342/158656859-f8a166df-8496-4b9c-86b1-0674e2f4f4d0.mp4

### Steps to reproduce:

1. Go to WooCommerce shipping settings and setup a country with only free shipping.
    - Note: if the country has both free shipping and flat rate, then this issue won't happen. 
2. Open dev tools to observe network requests. 
3. Go to Setup MC Step 3. If you have existing shipping rates setup, you can delete all the shipping rates and reload the page.
4. Look into the shipping rates suggestions API response. It would contain the shipping rate with free shipping.
5. Look into the saving shipping rates API response. It returns error saying the shipping rate method is not flat_rate.

### Expected behavior:

If saving shipping rate can only work with flat_rate method, then the suggestions API should return only "flat_rate" shipping rates, without "free_shipping" shipping rates.

Ultimately, there shouldn't be error when we try to save the shipping rates suggestions.

### Actual behavior:

There is an error when we try to save the shipping rates suggestions.

### Additional details:

This issue happens in the following branches:

- `feature/refactor-shipping-rates-api` branch (d9a3d6d4421270e8d36dcf0086e2b5f49a84bdc0) 
- `feature/shipping-rates-ui/main` branch (0b6fc1589f312f6f395db917a299ef4189f4ac2d).

nima-karimi:

This is a work-in-progress because even though the backend changes are complete, we need to confirm that the UI works with the API changes. @ecgan, can you please check if the API changes will work with the current UI?

Closes #1332.

This PR changes the shipping rates API schema to allow setting free shipping rates. Previously, the API assumed that a free shipping rate would be sent as a `flat_rate` with a cost of 0 and if it had a minimum order threshold, that would've also been sent as part of the flat_rate method options. However, since this might get confusing and not similar to how WooCommerce stores shipping rates, I've decided to change this logic.

Now, if the user checks the "I offer free shipping for orders over a certain price" option in UI, another "free_shipping" rate must be sent to the API that includes the "free_shipping_threshold" set by the user and a rate of 0.

The `free_shipping_threshold` option should only be sent for a "free_shipping" rate and NOT a "flat_rate".

The suggestions API is also modified to include a "free_shipping" rate when a free_shipping method with a minimum order amount exists in the WooCommerce zones for the country. Previously, if there was a free shipping method for the country, the suggestions API would still return a "flat_rate" method but set the "free_shipping_threshold" option for it.

The DB migration is also changed so that it creates new "free_shipping" methods for all countries if the user had previously set the "offers_free_shipping" and "free_shipping_threshold" globally.

**Test the DB migration**:

1. Change the global MC settings by modifying the `gla_merchant_center` option in `wp_options` and set `offers_free_shipping` set to 1 and `free_shipping_threshold` set to any value.
1. Update the `WC_GLA_VERSION` constant to a higher version than the current one.
2. Go to the `Migration20211228T1640692399::get_applicable_version` method and modify the `x.x.x` version to be equal to or greater than the one you set above.
3. Open any page in your website admin
4. Confirm that the `wp_gla_shipping_rates` table is updated and includes the options and method columns. The method column must be set by default to `flat_rate` but the table should also include rows with the `free_shipping` method that includes a `free_shipping_threshold` in the `options` column.

**Test APIs**

1. Define multiple shipping zones each containing various regions and shipping methods. Make sure you have at least one free shipping method with a minimum order amount option.
2. Make a `GET` request to the `mc/shipping/rates/suggestions` API with a `country_codes[]` parameter and confirm the rate that is returned is the same that you defined in your shipping zone settings.
3. Confirm that the returned shipping rates suggestions include a "free_shipping" rate with the correct "free_shipping_threshold"
4. Make a POST request to the mc/shipping/rates containing the following JSON data
```JSON
{
  "country_code":"DE",
  "rates":[
    {
      "method":"flat_rate",
      "currency":"USD",
      "rate":50
    },
    {
      "method":"free_shipping",
      "currency":"USD",
      "rate":0,
      "options":{
        "free_shipping_threshold":"100"
      }
    }
  ]
}
```
5. Make a GET requests to the `mc/shipping/rates` APIs and confirm the shipping rate you set above is returned correctly.
6. Make a POST request to the `wc/gla/mc/settings/sync` API
7. Confirm that the shipping rates you have just set are reflected in your Merchant Center account (https://merchants.google.com/mc/shipping/services). Check the minimum order value as well.

### Changelog entry

>

### DIFF
--- a/src/API/Google/Settings.php
+++ b/src/API/Google/Settings.php
@@ -86,20 +86,14 @@ class Settings {
 			if ( isset( $options['free_shipping_threshold'] ) ) {
 				$minimum_order_value = (float) $options['free_shipping_threshold'];
 
-				if ( 0 !== $rate ) {
-					// Add a conditional free-shipping service if the current rate is not free.
-					$services[] = $this->create_conditional_free_shipping_service( $country, $currency, $minimum_order_value );
-				} else {
-					// Set the minimum order value if the current rate is free.
-					$service->setMinimumOrderValue(
-						new Price(
-							[
-								'value'    => $minimum_order_value,
-								'currency' => $currency,
-							]
-						)
-					);
-				}
+				$service->setMinimumOrderValue(
+					new Price(
+						[
+							'value'    => $minimum_order_value,
+							'currency' => $currency,
+						]
+					)
+				);
 			}
 
 			$services[] = $service;
@@ -346,31 +340,6 @@ class Settings {
 		if ( array_key_exists( $country, $times ) ) {
 			$service->setDeliveryTime( $this->create_time_object( intval( $times[ $country ] ) ) );
 		}
-
-		return $service;
-	}
-
-	/**
-	 * Create a free shipping service.
-	 *
-	 * @param string $country
-	 * @param string $currency
-	 * @param float  $minimum_order_value
-	 *
-	 * @return Service
-	 */
-	protected function create_conditional_free_shipping_service( string $country, string $currency, float $minimum_order_value ): Service {
-		$service = $this->create_shipping_service( $country, ShippingZone::METHOD_FREE, $currency, 0 );
-
-		// Set the minimum order value to be eligible for free shipping.
-		$service->setMinimumOrderValue(
-			new Price(
-				[
-					'value'    => $minimum_order_value,
-					'currency' => $currency,
-				]
-			)
-		);
 
 		return $service;
 	}
--- a/src/API/Site/Controllers/ShippingRateSchemaTrait.php
+++ b/src/API/Site/Controllers/ShippingRateSchemaTrait.php
@@ -42,6 +42,7 @@ trait ShippingRateSchemaTrait {
 				'description'       => __( 'The shipping method.', 'google-listings-and-ads' ),
 				'enum'              => [
 					ShippingZone::METHOD_FLAT_RATE,
+					ShippingZone::METHOD_FREE,
 				],
 				'context'           => [ 'view', 'edit' ],
 				'validate_callback' => 'rest_validate_request_arg',
--- a/src/DB/Migration/Migration20211228T1640692399.php
+++ b/src/DB/Migration/Migration20211228T1640692399.php
@@ -63,10 +63,31 @@ class Migration20211228T1640692399 extends AbstractMigration {
 			$this->wpdb->query( "ALTER TABLE `{$this->wpdb->_escape( $this->shipping_rate_table->get_name() )}` ALTER COLUMN `method` DROP DEFAULT" );
 
 			$mc_settings = $this->options->get( OptionsInterface::MERCHANT_CENTER );
-			if ( isset( $mc_settings['offers_free_shipping'] ) && false !== boolval( $mc_settings['offers_free_shipping'] ) && isset( $mc_settings['free_shipping_threshold'] ) ) {
-				// Move the free shipping threshold from the options to the shipping rate table.
-				$options_json = json_encode( [ 'free_shipping_threshold' => (float) $mc_settings['free_shipping_threshold'] ] );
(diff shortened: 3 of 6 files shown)


ecgan:
@nima-karimi , on the surface, it does seem to help fix the the issue #1332. I also think it may be good to make it similar to how WooCommerce stores shipping rates. 

However, there are quite a number of API changes here, and the UI we have doesn't work as per user expectation and the API expectation stated in the PR description here.

For an example, we have two groups of "Free shipping for all orders" below, because in the UI side we also group by shipping methods, and the groups correspond to "free shipping" and "flat rate". It may be confusing for the users here.

image

> Now, if the user checks the "I offer free shipping for orders over a certain price" option in UI, **_another "free_shipping" rate_** must be sent to the API that includes the "free_shipping_threshold" set by the user and a rate of 0.

This is another part that needs to be worked on in the UI side.

> The `free_shipping_threshold` option should only be sent for a "free_shipping" rate and NOT a "flat_rate".

In the current UI, we are sending "flat_rate" with `free_shipping_threshold` option, and it seems to still work as expected. Work is needed in the UI side to match the API expectation. 

---

> However, since this might get confusing and not similar to **_how WooCommerce stores shipping rates_**, I've decided to change this logic.

Again, I think it may be good to make it similar to WooCommerce settings. However, to achieve that, I think we may also need to get design input quite a bit too.

Currently in our GLA UI, the "free shipping" and "flat rate" shipping settings is mixed in the shipping rate section UI. To make it similar with WC settings, I think the shipping rate section should reflect the WC settings structure too, e.g. display countries, and within the countries, specify flat rate and/or free shipping. (Currently we have flat rate and free shipping in one card, and then free shipping's threshold amount in another card, which doesn't really reflect WC settings structure.)

In other words, I think it requires quite a bit of time to make it similar to WC Settings. 😅

nima-karimi:
Thanks for the comments, @ecgan. Given that we're close to releasing time, it's best to steer away from drastic changes and try to resolve this with a simpler solution.

I think regardless of how UI displays this information, it should always send a `flat_rate` to the shipping rates API because that's how the API is currently designed to work.

Since the UI already recognizes a `flat_rate` shipping method with a rate of `0` as a `free_shipping` method, it makes sense for it to simply send the same `flat_rate` shipping to the API instead of a `free_shipping` method. That way https://github.com/woocommerce/google-listings-and-ads/issues/1332 would be resolved because the only issue there seems to be API not recognizing a `free_shipping` rate.

Another possible solution would be for the API to recognize a `free_shipping` rate, which is what this PR is implementing; however, to simply add the `free_shipping` rate to the list of acceptable shipping methods would be conflicting with how we parse and store the shipping rates at the moment. This cascades to a bunch of other changes.

I'm closing this PR as we discussed (p1647456507645559-slack-C01E9LB4X61).

Issue #1356, opened 2022-03-24
Replace the selector of targeting audiences with TreeSelectControl component


eason9487:
# Context

It's a follow-up of #1248 - Multi-country targeting for Google Ads campaign

💡  Looks like #112 could be closed after finishing this task.

## Objective 

Replace the current selector component of targeting audiences with `<TreeSelectControl>` component. Make it easier for merchants to select countries.

![image](https://user-images.githubusercontent.com/17420811/154234977-da9b289f-6bb2-4393-a2c0-f5172b891da1.png)

### The targeting audience selectors to be replaced:

- Step 2 in the onboarding setup
- Step 1 in the free listings edit
- Step 2 in the paid campaign setup
- Campaign creation page
- Campaign edit page

## The current state of the system

Currently, it shows a flat country list for merchants to select audiences. It also results in a very inefficient setup experience, especially if the store ships to all 22 EU countries or worldwide.

![image](https://user-images.githubusercontent.com/17420811/159849343-82a9808d-13e2-45e4-9858-b8a6a9469f3a.png)

# Terminology

💡  Scope: this project and all related issues/pull requests

- The "old campaign(s)" refers to the existing campaigns created **before** releasing this feature.
- The "new campaign(s)" refers to the campaigns created **after** this feature is finished.

# Technical

💡  We will create a shared branch `feature/tree-select-audience`. All relevant developments should be branched from it and merged back to it. And open a PR to merge the shared branch to `develop` once the overall feature gets ready to release.

💡  The old campaigns will still be presented in a single-selected country UI when editing. The rest of relevant pages will be presented on a multi-country basis.

## Frontend

The UI changes should be applied to campaign creation and edit pages. All frontend routings will stay with the current.

### Phase 1

Completed. Please refer to phase 1 of frontend section in #1248 for more details.

### Phase 2

Some tasks have been completed. Please refer to phase 2 of frontend section in #1248 for more details.

📌 **Integrate with changed `mc/countries` API**
- Adjust data access to make the original country data still compatible for use in UI
- Get continent data from API

📌 **Replace the single country selection component with the dedicated component**
- Implement multi-country selection component by `<TreeSelectControl>` component.
- Apply the country and continent data as tree structure options.

### Phase 3

📌 **Refactor the countries selectors**

[...]

PR #1358, opened 2022-03-24
Integrate frontend data store with the changed `mc/countries` API

eason9487:

This PR implemented the **Integrate with changed mc/countries API** part in #1356

- Rename the selector, resolver and reducer of the original `getCountries` in wp-data, and adjust them to get Merchant Center supported countries and continents from API. The relevant hook `useGetCountries` was renamed as well.
   - The reason for renaming is that it returns the countries supported by the Merchant Center, not the countries supported by WC or all countries.
- Replace all `getCountries` and `useGetCountries` uses with `useMCCountries`.

### Screenshots:

image

image

1. Go to step 1 in the free listings edit page.
2. Check if the countries in the audience selectors are the same as before. It should have 93 countries.


js/src/hooks/useTargetAudienceFinalCountryCodes.js
```diff
@@ -30,16 +31,15 @@ import { STORE_KEY } from '.~/data';
  *
  */
 const useTargetAudienceFinalCountryCodes = () => {
-	return useSelect( ( select ) => {
-		const { getTargetAudience, getCountries, isResolving } = select(
-			STORE_KEY
-		);
+	const {
+		data: supportedCountries,
+		isResolving: countriesLoading,
```
eason9487, line 36: 📝  The reason for still using `isResolving` is to keep the original state of its interface consistent.

js/src/data/resolvers.js
```diff
@@ -128,8 +127,25 @@ getGoogleMCContactInformation.shouldInvalidate = ( action ) => {
 	return action.type === TYPES.VERIFIED_MC_PHONE_NUMBER;
 };
 
-export function* getCountries() {
-	yield fetchCountries();
+export function* getMCCountriesAndContinents() {
+	try {
+		const query = { continents: true };
```
jorgemd24, line 132: ❔

What was the reason to move the API request logic from action.js and putting it all in the resolvers.js? Is there any convention/standard for these cases?

js/src/data/selectors.js
```diff
@@ -88,9 +88,17 @@ export const getGoogleMCPhoneNumber = createRegistrySelector(
 	}
 );
 
-export const getCountries = ( state ) => {
-	return state.mc.countries;
-};
+export const getMCCountriesAndContinents = createSelector(
```
jorgemd24, line 91: 👏

Nice using memoized values here. 🙂

jorgemd24 (approved):
Thanks @eason9487, LGTM 👍 

✅  All countries are displayed.
✅  No errors are shown

js/src/data/resolvers.js
eason9487: There might not be a standard or convention, I was interpreting it from the `@wordpress/data` (wp-data) implementation design.

Actions are mainly the mechanism for making changes to states in the store (of wp-data), for example, a create/update/delete call corresponding to the API. And selectors are the primary mechanism for retrieving states from the store. Consider that most states are obtained from APIs, so wp-data uses resolvers as the side effect handlers of the same name selectors. Therefore, selectors and resolvers correspond more to read calls to the API.

Moreover, the UI often has multiple places requesting the same state in parallel, wp-data internally handles the same selector calls, allows the concurrent calls to issue only one API request, and notifies observers when the store is updated. In addition, its interface also includes the `isResolving`/`hasFinishedResolution` status of states.

If uses actions to retrieve data from API, it will lose the benefits of selector and resolver mechanisms. Also, from an external use point of view, the exposed state operation interfaces are selectors and actions when using wp-data. Moving this to resolvers.js could make the interface clearer and avoid misuse.

jorgemd24: Thanks, @eason9487 for your detailed explanation 😃

PR #1359, opened 2022-03-24
Add blurb to inform users to connect Google Ads child account and not a parent MCC account

### LINKED ISSUE #1346: Add blurb to encourage users to connect child Ads account and not a parent MCC

> Currently, while our plug-in automatically prevents merchants from linking an MCA (multi-client account in merchant center), it currently does not necessarily prevent users from linking an MCC (manager account in Google Ads). We know that
> (1) campaign creation does not happen and (2) the users are not able to access insightful reporting after linking a manager account

#### How to reproduce the problem

Create and connect an an MCC account that has child accounts

Initial quick solution is to add a "blurb" to the UI to encourage users to only connect child accounts.

We will likely create follow up project/issues to address the underlying problem at a later point.

## Technical

TBC - Initial assumption is this should hopefully be mostly a cosmetic change.

## Figma link

TBC - cc @j111q - to update once we have confirm copy/designs

## Event tracking

Ensure any links in blurb are firing a tracks event to help us gauge interactions.

ecgan:

Closes #1346.

This PR adds a "blurb" or message to the UI to encourage users to only connect child Google Ads accounts, not manager account.

The message will only be displayed when there are more than one Google Ads account. Clicking on the "Learn more" link will log a track event: 

```
wcadmin_gla_documentation_link_click {context: 'setup-ads-connect-account', link_id: 'connect-sub-account', href: 'https://support.google.com/google-ads/answer/6139186'}
```

Other technical notes:

- I have changed the `AdsAccountSelectControl` so that it receives `accounts` prop that is passed down from `NonConnected` component and doesn't need to rely on `useExistingGoogleAdsAccounts` hook.
- `AdsAccountSelectControl` is only used in `ConnectAds` component. I have moved it from the shared components folder into the local folder together with `ConnectAds`.

### Screenshots:

When there is only one Google Ads account:

image

When there are more than one Google Ads account: 

image

Clicking on the "Learn more" link will log a track event: 

image

Note: to simulate a single Google Ads account or multiple Google Ads accounts, you can modify the following lines:

https://github.com/woocommerce/google-listings-and-ads/blob/e06e6c313d6529b739100cc3fd62534963c4e014/js/src/data/reducer.js#L241-L243

```js
		// For single account:
		case TYPES.RECEIVE_ACCOUNTS_GOOGLE_ADS_EXISTING: {
			return setIn( state, 'mc.accounts.existing_ads', [ 11111 ] );
		}

		// For multiple accounts:
		case TYPES.RECEIVE_ACCOUNTS_GOOGLE_ADS_EXISTING: {
			return setIn( state, 'mc.accounts.existing_ads', [ 11111, 22222 ] );
		}
```

1. Disconnect your Google Ads account.
2. Go to Setup Ads flow. You should see the screenshots above.
3. Click on the "Learn more" link. You should see track event being logged.

### Changelog entry

> Tweak - Add message to advise users to only connect Google Ads child account, not manager account.


ecgan:
@j111q , please see the screenshots in the PR description above. It is a bit different from what we have in your design, specifically the text message goes over / above the "Connect" button (instead of having the same width as the drop down select control and having some empty space on the right). I make it that way to be consistent with the other cards, e.g. Switch URL Card. I hope that looks good.

The design:

image

The implemented UI:

image

The design for Switch URL Card:

image

j111q:
Yup that's fine. Let me just update the Figma so it's consistent and I don't forget. Thanks for heads up Gan! 🙏 🚀

js/src/setup-ads/ads-stepper/setup-accounts/google-ads-account-card/connect-ads/index.scss
```diff
@@ -2,4 +2,12 @@
 	.app-select-control {
 		flex-grow: 1;
 	}
+
+	.wcdl-subsection-body {
+		margin-bottom: 8px;
+
+		a {
+			text-decoration: none;
```
tomalec: :question: Do we really need to remove this underline? In all the other flows we do have underlined, and I think it'd be nice to have a conventional and consistent experience.

ecgan: Ahh you're right, for example the Learn more link in the Reclaim URL Card below:

image

I'll make the changes to make it consistent then. 🙂

ecgan: Done in [`64f64ae` (#1359)](https://github.com/woocommerce/google-listings-and-ads/pull/1359/commits/64f64ae0570616fe539dfc2915c2dd6dfbae2a6c).

image

tomalec (approved):
Tested locally, reviewed the code, LGTM.

PR #1360, opened 2022-03-24
Additional data points for tracker snapshot.

### LINKED ISSUE #1278: Additional data points for tracker snapshot
In https://github.com/woocommerce/google-listings-and-ads/pull/1167 we introduced some changes to inbox notifications.

This included some additional conditions
* has_account_issues
* has_at_least_one_synced_product
* is_setup_started (ads)

The marketing team would like to use these data points/conditions for some marketing campaigns and as such would like this information exposed in tracker snapshot data https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/TrackerSnapshot.php#L76-L95

has_account_issue: true/false
has_at_least_one_synced_product: true/false
is_ads_setup_started: true/false

For consideration/input?

* I'm not sure if there are performance concerns as a result of introducing these - please advise during development

rawdreeg:

- Added additional data points for tracker snapshot.
- I modified the scope of has_at_least_one_synced_product() and has_account_issues method() to be able to pull data. 

Closes #1278.

Note: I had doubt about performance of [has_at_least_one_synced_product() ](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Notes/AbstractSetupCampaign.php#:~:text=protected%20function%20has_at_least_one_synced_product()%3A%20bool%20%7B), but I can't immediately see an issue here. get_product_statistics() has a a type of caching mechanism.

1. Add a code snippet to temporarily log the tracker data:
```
add_action(

	'admin_init',
	function() {
		$tracks = apply_filters( 'woocommerce_tracker_data', [] );
		error_log( json_encode( $tracks, JSON_PRETTY_PRINT ) );
	}
);
```

2. Refresh a page to confirm that the data is logged
3. Check the log file to confirm that the data is shown as expected
```
{
    "extensions": {
        "gla": {
            "settings": {
                "version": "1.11.1",
                "db_version": "1.11.1",
                "tos_accepted": "yes",
                "google_connected": "yes",
                "mc_setup": "yes",
                "ads_setup": "no",
                "target_audience": [
                    "ZA"
                ],
                "shipping_rate": "flat",
                "shipping_time": "flat",
                "offers_free_shipping": "no",
                "free_shipping_threshold": "",
                "tax_rate": "manual",
                "has_account_issue": "yes",
                "has_at_least_one_synced_product": "no",
                "ads_setup_started": "no"
            }
        }
    }
}
```

### Changelog entry

- Add  - Additional data points for tracker snapshot


src/Notes/AbstractSetupCampaign.php
```diff
@@ -109,7 +109,7 @@ public function should_be_added(): bool {
 	 * @since 1.11.0
 	 * @return bool
 	 */
-	protected function has_account_issues(): bool {
+	public function has_account_issues(): bool {
```
mikkamp: I don't think it's the right solution to make these public. When we want to get details about the Merchant Center, we should be getting it from the MerchantCenterService class (it has lots of very similar functions already).

I think it would be better to do the following:
- Move these two helper functions to MerchantCenterService
- Pass MerchantCenterService instead of MerchantStatuses to AbstractSetupCampaign
- Call the helper functions in both the Note and the Tracker snapshot

src/Tracking/TrackerSnapshot.php
```diff
@@ -77,18 +79,25 @@ protected function get_settings(): array {
 		/** @var TargetAudience $target_audience */
 		$target_audience = $this->container->get( TargetAudience::class );
 		$mc_settings     = $this->options->get( OptionsInterface::MERCHANT_CENTER );
+		/** @var AdsService $ads_service */
+		$ads_service = $this->container->get( AdsService::class );
+		/** @var SetupCampaignNote $campaign_setup */
+		$campaign_setup = $this->container->get( SetupCampaignNote::class );
```
mikkamp: As mentioned in the previous comment we want to get data from the AdsService and the MerchantCenterService, not directly from a Note which has a different function.

src/Tracking/TrackerSnapshot.php
```diff
@@ -77,18 +79,25 @@ protected function get_settings(): array {
...
+			'google_connected'                => $this->get_boolean_value( OptionsInterface::GOOGLE_CONNECTED ),
+			'mc_setup'                        => $this->get_boolean_value( OptionsInterface::MC_SETUP_COMPLETED_AT ),
+			'ads_setup'                       => $this->get_boolean_value( OptionsInterface::ADS_SETUP_COMPLETED_AT ),
+			'target_audience'                 => $target_audience->get_target_countries(),
+			'shipping_rate'                   => $mc_settings['shipping_rate'] ?? '',
+			'shipping_time'                   => $mc_settings['shipping_time'] ?? '',
+			'tax_rate'                        => $mc_settings['tax_rate'] ?? '',
+			'has_account_issue'               => $campaign_setup->has_account_issues() ? 'yes' : 'no',
+			'has_at_least_one_synced_product' => $campaign_setup->has_at_least_one_synced_product() ? 'yes' : 'no',
+			'is_ads_setup_started'            => $ads_service->is_setup_started() ? 'yes' : 'no',
```
mikkamp: We already have `ads_setup` so to me it would have been more logical to name it `ads_setup_started`. But not a big deal if it's requested to be named like that.

mikkamp (commented):
Thanks for the changes they are working as expected.

Regarding performance issues, tracker snapshots are run once per day through Action Scheduler, so we should be pretty safe to call these.

Just left some comments about getting the data from a service class instead of from a Note.

src/Notes/AbstractSetupCampaign.php
rawdreeg: Thanks, @mikkamp! I have implemented your suggestions and updated the PR

src/Tracking/TrackerSnapshot.php
rawdreeg: I pretty much copied this key value from the issue ticket. Not sure if the name is set in stone. I will check with @jconroy

findingsimple: `ads_setup_started` is fine 👍

rawdreeg: Sweet! updated!

mikkamp (approved):
Thanks for the refactor. The code looks good, and I get the results I expect when testing.

PR #1361, opened 2022-03-24
Refactoring product filters to fix filter parameter type.

### LINKED ISSUE #1259: Fix ProductFilter to return products and remove `$return_ids` parameter
Not sure `$return_ids` is needed here. Method is only used at one single place and returns ids. Filters also filter products, so, maybe, should always return products and then, if you need ids, map to get ids outside of filter. This is still going to be another loop, but same time it is not going to raise the order of complexity.
Besides there seems to be a bug in this class with `filter_sync_ready_products` using `$return_ids` when `woocommerce_gla_get_sync_ready_products_filter` filter expects Product objects, not ids.

# Make sure tests capture this first and then start fixing.

https://github.com/woocommerce/google-listings-and-ads/blob/3b2dd688aeaff4cbddce20112e6c159d6e5b3eb4/src/Product/ProductFilter.php#L66

_Originally posted by @message-dimke in https://github.com/woocommerce/google-listings-and-ads/pull/1238#r810368533_

message-dimke:
A task was to fix filter parameter data type. I also refactored the code to remove `$return_ids` seconds parameter from product filter functions as the one which is not needed. Added product filter unit tests and adjusted the other code which was using filters to make sure it is still consistent with the changes made. Checked those changes to be covered with existing unit tests to make sure we have it tested and no changes passed through unnoticed.

Closes #1259.

Recently a bug was found with feeding `woocommerce_gla_get_sync_ready_products_filter ` filter with array of product IDs when it was made to process `WC_Product` type of records instead.

1. Since GitHub does run tests itself, nothing to check here;

### Changelog entry

> Fix - bug with filter parameter data type.

### DIFF
--- a/src/Product/ProductRepository.php
+++ b/src/Product/ProductRepository.php
@@ -148,7 +148,7 @@ class ProductRepository implements Service {
 	public function find_sync_ready_products( array $args = [], int $limit = - 1, int $offset = 0 ): FilteredProductList {
 		$results = $this->find( $this->get_sync_ready_products_query_args( $args ), $limit, $offset );
 
-		return $this->product_filter->filter_sync_ready_products( $results, false );
+		return $this->product_filter->filter_sync_ready_products( $results );
 	}
 
 	/**
@@ -164,7 +164,7 @@ class ProductRepository implements Service {
 	 */
 	public function find_delete_product_ids( array $ids, int $limit = - 1, int $offset = 0 ): array {
 		$results = $this->find_by_ids( $ids, $limit, $offset );
-		return $this->product_filter->filter_products_for_delete( $results, true );
+		return $this->product_filter->filter_products_for_delete( $results )->get_product_ids();
 	}
 
 	/**
@@ -179,7 +179,7 @@ class ProductRepository implements Service {
 	public function find_sync_ready_product_ids( array $args = [], int $limit = - 1, int $offset = 0 ): FilteredProductList {
 		$results = $this->find( $this->get_sync_ready_products_query_args( $args ), $limit, $offset );
 
-		return $this->product_filter->filter_sync_ready_products( $results, true );
+		return $this->product_filter->filter_sync_ready_products( $results )->get_product_ids();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/Product/ProductRepository.php
```diff
@@ -174,12 +174,12 @@ public function find_delete_product_ids( array $ids, int $limit = - 1, int $offs
...
 	 * @param int   $offset Amount to offset product results.
 	 *
-	 * @return FilteredProductList List of WooCommerce product IDs after filtering.
+	 * @return FilteredProductList List of WooCommerce products after filtering.
 	 */
 	public function find_sync_ready_product_ids( array $args = [], int $limit = - 1, int $offset = 0 ): FilteredProductList {
 		$results = $this->find( $this->get_sync_ready_products_query_args( $args ), $limit, $offset );
 
-		return $this->product_filter->filter_sync_ready_products( $results, true );
+		return $this->product_filter->filter_sync_ready_products( $results );
```
jorgemd24, line 182: 💅 

The comment of this function could be updated, so it does not mention that we are returning IDs. 

`Find and return an array of WooCommerce product IDs ready to be submitted to Google Merchant Center.`

Also the name of this function is confusing as `find_sync_ready_product_ids` sounds like will return `productIds` instead of `FilteredProductList`

tests/Unit/Product/ProductFilterTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+		$this->product_helper->expects( $this->exactly( 3 ) )
+			->method( 'is_sync_ready' )
+			->withConsecutive( [ $product_a ], [ $product_b ], [ $product_c ] )
+			->willReturnOnConsecutiveCalls( false, true, false );
+
+		$this->product_helper->expects( $this->once() )
+			->method( 'is_sync_failed_recently' )
+			->with( $product_b )
+			->willReturn( false );
```
jorgemd24, line 119: ❓

I think it could be convenient to add a test covering when the product is `is_sync_ready = true` and  `is_sync_failed_recently = true` so we can cover all the scenarios.

jorgemd24 (approved):
Thanks, @message-dimke for fixing this and adding the tests! It looks good, I have left a couple of comments related to the tests and the function name.

PR #1363, opened 2022-03-25
Only return `flat_rate` suggestions in the API

nima-karimi:

Related to #1332.

This PR modifies the suggestion service class to only return `flat_rate` shipping rates as suggestions. Previously, if a country had only one `free_shipping` rate it would return it as is, but now it maps them to a `flat_rate` method.

This change is required because the shipping rates API (`POST mc/shipping/rates`) only accepts `flat_rate` methods and it would be confusing to return `free_shipping` rates even though we don't accept them in our other APIs.

1. Define a shipping zone that contains only one free shipping method.
2. Make a `GET` request to the `mc/shipping/rates/suggestions` API with a `country_codes` parameter and confirm the free shipping rate is returned as a `flat_rate` with a rate of `0`

### Changelog entry

> Tweak - Return only `flat_rate` shipping methods in the suggestions API


ecgan:
> Previously, if a country had only one `free_shipping` rate it would return it as is, but now _it maps them to a `flat_rate` method_.

I was testing this PR with the UI. It does seem to fix the original issue #1332 - I have a Vietnam shipping zone with Free Shipping method only, and the suggestions API returns "VN" with `flat_rate` method and `0` rate, and we can call save shipping rate API successfully.

image

image

However, I think I found another issue with a different scenario. 😅 

See screenshot below. I have a Sri Lanka shipping zone with only Free Shipping method, _and_ with minimum order amount 50, and the suggestions API returns "LK" with `flat_rate` method and `0` rate and `50` free shipping threshold. 

image

image

Because the rate is 0, it is considered as "free shipping" as a whole, and minimum order amount cannot be entered by the users nor be displayed to the users. (This is actually a [rule](https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=4656%3A112688) stated in Figma. 😅 ) So the `50` free shipping threshold from the suggestions API is being ignored here. 

image

So for this scenario, there is a mismatch in user's expectation: 

- In WC Settings, users only want to offer to Sri Lanka free shipping with minimum order amount 50. 
- in GLA shipping rate UI, Sri Lanka has free shipping, without minimum order amount.

nima-karimi:
Hmm. This sounds like an issue in the front-end though. We decided to keep the front-end as simple as possible and a rule like this is a bit complex. I mean why would anyone only offer a free-shipping rate with a min order amount without any alternatives? 😅 I don't think there are use cases for this and even if there are, we can't do anything about it while keeping the front-end design as-is.

ecgan:
> I mean why would anyone only offer a free-shipping rate with a min order amount without any alternatives? 😅

Yeah, I agree too. My above scenario is like "I will do free shipping to you in Sri Lanka if you order with minimum amount 50. If your order amount is less than 50, I won't ship to you _at all_", which kinda doesn't make sense. 😅  However, _technically_, users can set that up in WC Settings. 

Side note: I also feel this is related to my comment in the other PR https://github.com/woocommerce/google-listings-and-ads/pull/1357#issuecomment-1077836461, where we are mixing WooCommerce's "free shipping" and "flat rate" in our API and UI, which is causing the difficulties we have here. 🤔 

cc @j111q , as this relates to front-end design. 

---

> I don't think there are use cases for this and even if there are, we can't do anything about it while keeping the front-end design as-is.

I'm thinking, with the above "Sri Lanka" scenario issue, even though it is unlikely, the impact could be high to the users, we might want to put a hold on this PR first while figuring out a way forward (e.g. change the front-end design, so that API and UI can work nicely on this together)?

Meanwhile, for issue #1332 , I can come up with a quick fix in the UI side by filtering out shipping rates suggestions with `free_shipping` method. Let me know if that sounds good, or if you like to address it differently in API side? 🤔

nima-karimi:
> I'm thinking, with the above "Sri Lanka" scenario issue, even though it is unlikely, the impact could be high to the users, we might want to put a hold on this PR first while figuring out a way forward (e.g. change the front-end design, so that API and UI can work nicely on this together)?

But even without this PR, wouldn't the result be the same? I mean if a country has only a free shipping option with a minimum order amount, the suggestions API still returns the same rate albeit with a "free_shipping" method. How does the UI display it in this case?

I'm okay with doing it in the UI if it has fewer side effects because I've changed this logic in my other PR for milestone 2: https://github.com/woocommerce/google-listings-and-ads/pull/1341

ecgan:
> But even without this PR, wouldn't the result be the same? I mean if a country has only a free shipping option with a minimum order amount, the suggestions API still returns the same rate albeit with a "free_shipping" method. How does the UI display it in this case?

With the same WC Settings that I have, using the `feature/refactor-shipping-rates-api` branch without this PR, yes the suggestions API returns the shipping rates with "free_shipping" method. 

image

Then, we try to save the suggestions by calling batch save shipping rates API, and it throws error (the red color "batch" request in the screenshot above).

Because there is an error in saving the suggestions, we show a notice to advise the users to set up the shipping rates manually.

image
 
On the UI side, we do not have any code that works with "free_shipping" method. We only have code that works with "flat_rate" method.

> I'm okay with doing it in the UI if it has fewer side effects because I've changed this logic in my other PR for milestone 2: #1341

Okay gotcha, I'll come up with a PR for it. 🙂 👍

nima-karimi:
Closing this PR since the changes are no longer relevant and applied in a different way in https://github.com/woocommerce/google-listings-and-ads/pull/1422

PR #1365, opened 2022-03-25
Return an empty array when no campaigns exist for GET ads/campaigns API

### LINKED ISSUE #1364: GET ads/campaigns API throws error when there are no campaigns
### Describe the bug:

This is a bug introduced in #1280. When there are no any campaigns in an account, `GET ads/campaigns` will throw an error:

```
{
    "message": "Error retrieving campaigns: Error in campaign.id IN (''): invalid number .",
    "errors": {
        "BAD_NUMBER": "Error in campaign.id IN (''): invalid number .",
        "INVALID_ARGUMENT": "Request contains an invalid argument."
    }
}
```

Because when no campaigns exist it's calling a query:

```
SELECT campaign.id,campaign_criterion.location.geo_target_constant
FROM campaign_criterion
WHERE campaign.id IN ('') // This causes the problem
AND campaign_criterion.negative = 'false'
AND campaign_criterion.status != 'REMOVED'
AND campaign_criterion.location.geo_target_constant IS NOT NULL;
```

We need to add more checks in `combine_campaigns_and_campaign_criterion_results()` method of `src/API/Google/AdsCampaign.php`.

### Steps to reproduce:

1. Remove any existing campaigns.
2. Request to the API `GET ads/campaigns` by Postman.
3. The API throws an error as in the above description.

### Expected behavior:

The API should return an empty array when there are no campaigns exist.

### Actual behavior:

The API throws an error.

ianlin:

This PR fixes a bug that when no campaigns exist the `GET ads/campaigns` API will throw an error. It should return an empty array instead.

Closes #1364.

### Screenshots:

Screenshot 2022-03-25 at 17 57 23

1. Remove all campaigns in your account.
2. Request to the API `GET ads/campaigns`.
3. The API should return an empty array.

### Additional details:

I was struggling to add a unit test for this case as this is an error from the Google API. Might be more suitable to have the integration test for it.

### Changelog entry

> Fix - Returns an empty array when no campaigns exist.

### DIFF
--- a/tests/Unit/API/Google/AdsCampaignTest.php
+++ b/tests/Unit/API/Google/AdsCampaignTest.php
@@ -70,6 +70,11 @@ class AdsCampaignTest extends UnitTest {
 
 	public function test_get_campaigns_empty_list() {
 		$this->generate_ads_campaign_query_mock( [], [] );
+
+		// Method search() only being called by AdsCampaignQuery
+		$this->service_client->expects( $this->once() )
+			->method( 'search' );
+
 		$this->assertEquals( [], $this->campaign->get_campaigns() );
 	}
 
@@ -109,6 +114,10 @@ class AdsCampaignTest extends UnitTest {
 		];
 
 		$this->generate_ads_campaign_query_mock( $campaigns_data, $campaign_criterion_data );
+
+		$this->service_client->expects( $this->exactly( 2 ) )
+			->method( 'search' );
+
 		$this->assertEquals( $campaigns_data, $this->campaign->get_campaigns() );
 	}
 
--- a/tests/Unit/API/Google/AdsCampaignTest.php
+++ b/tests/Unit/API/Google/AdsCampaignTest.php
@@ -69,12 +69,7 @@ class AdsCampaignTest extends UnitTest {
 	}
 
 	public function test_get_campaigns_empty_list() {
-		$this->generate_ads_campaign_query_mock( [], [] );
-
-		// Method search() only being called by AdsCampaignQuery
-		$this->service_client->expects( $this->once() )
-			->method( 'search' );
-
+		$this->generate_ads_campaign_query_mock_with_no_campaigns();
 		$this->assertEquals( [], $this->campaign->get_campaigns() );
 	}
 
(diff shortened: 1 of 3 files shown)


tests/Unit/API/Google/AdsCampaignTest.php
```diff
@@ -109,6 +114,10 @@ public function test_get_campaigns() {
 		];
 
 		$this->generate_ads_campaign_query_mock( $campaigns_data, $campaign_criterion_data );
+
+		$this->service_client->expects( $this->exactly( 2 ) )
+			->method( 'search' );
```
mikkamp: `service_client` is kind of part of the HelperTrait, so I think it would be better to move this expectation within the `generate_ads_campaign_query_mock` since that's the function that returns two results. Otherwise it looks a bit confusing what is expected in this test, and why it's not expected in the tests `test_get_campaigns_with_nonexist_location_id` and `test_get_campaigns_with_invalid_location_id`.

tests/Unit/API/Google/AdsCampaignTest.php
```diff
@@ -70,6 +70,11 @@ public function setUp() {
 
 	public function test_get_campaigns_empty_list() {
 		$this->generate_ads_campaign_query_mock( [], [] );
+
+		// Method search() only being called by AdsCampaignQuery
+		$this->service_client->expects( $this->once() )
+			->method( 'search' );
```
mikkamp: Since `service_client` is part of HelperTrait I think it would be cleaner to create a separate helper function `generate_ads_campaign_query_mock_with_no_campaigns`. A comment can be added there that it only expects one of the queries to run.

mikkamp (approved):
Thanks for the changes, it's working as expected now.
Just left a small suggestion for making the tests a little cleaner.

tests/Unit/API/Google/AdsCampaignTest.php
mikkamp: Also maybe we can update the doc block for `generate_ads_campaign_query_mock` since it now generates mocked responses for two queries.

ianlin: Thanks, good suggestions! Updated in 40fafee.

tests/Unit/API/Google/AdsCampaignTest.php
ianlin: Thanks, also updated in 40fafee.

PR #1366, opened 2022-03-25
Remove delete operations for campaign budget and ads group

### LINKED ISSUE #1362: Remove delete operations for budgets and ad groups
## User story

When a campaign is removed (being set to REMOVED status), I'd like the associated budget, ad group, ad group ad, and product group to be ineligible without needing to set REMOVED status to them. Only set REMOVED status to the campaign.

## Technical

Ref: pb0Spc-2pt-p2

Originally when a campaign is removed (set to `REMOVED`) we also set its associated resources (budget, ad group, ad group ad, product group) to `REMOVED` as well. Since it is just setting `REMOVED` status so the history of those resources would be preserved.

In #1334, we were trying to set `REMOVED` to campaign criterion when deleting a campaign, however we found out the history of the campaign criterion would not be preserved so we asked Google about it and they suggested when deleting a campaign we can only set `REMOVED` to the campaign and let the rest of the resources being untouched.

Thanks @mikkamp for doing the complete test on it, we conclude that it's safe to just set `REMOVED` to the campaign and remove other delete operations. Those associated resources' status would still be `ENABLED` (except for campaign budget) but in the Google Ads dashboard they were ineligible since their parent campaign is removed so we don't have to worry about anyone editing/changing those campaign resources. In our UI we only allow editing of (non removed) campaigns so they will never be queried or show up.

This is the opposite change to #390 but the results would be the same - the deleted campaign's associated resources are becoming ineligible.

## Acceptance criteria

When a campaign is deleted (set to `REMOVED`), its associated resources should be ineligible as well in Google Ads dashboard.

ianlin:

Closes #1362. This PR remove delete operations of campaign budget and ads group. For detailed information please see #1362.

1. Create a campaign.
2. Remove the newly created campaign.
3. From the GLA dashboard, this campaign should not appear.
4. From the Google Ads dashboard, this campaign's status should be `Removed` and the budget amount is still visible.
    Screenshot 2022-03-25 at 18 14 10
5. From the Google Ads dashboard, this campaign's `ad group` should still be visible and its status should be `Not eligible`.
    Screenshot 2022-03-25 at 18 13 44
6. From the Google Ads dashboard, this campaign's `ad` should still be visible and its status should be `Not eligible`.
    Screenshot 2022-03-25 at 18 14 34
7. From the Google Ads dashboard, this campaign's `location` is still visible.
    Screenshot 2022-03-25 at 18 18 42

### Changelog entry

> Update - Remove delete operations for campaign budget and ads group.


mikkamp (commented):
Thanks, it's working as expected when I delete a campaign.

I think we can take this one step further and also clear up the no longer used query code, (we don't query or edit the ad group and it's resources). So we can remove the following:

- AdsGroupQuery.php
- AdsGroupAdQuery.php
- AdsListingGroupQuery.php
- GoogleAdsClientTrait::generate_ads_group_query_mock
- GoogleAdsClientTrait::generate_ad_group_ad_resource_name
- GoogleAdsClientTrait::generate_listing_group_resource_name

ianlin:
@mikkamp Updated the code in 6c4526d and also merge develop branch to resolve conflicts.

mikkamp (approved):
Thanks for removing the additional unused code. It's looking good and it's still functioning as expected.

PR #1367, opened 2022-03-25
Filter out shipping rates suggestions that are not flat_rate method

### LINKED ISSUE #1332: Setup MC: batch saving shipping rate suggestions throws error when suggestions contain non-"flat-rate" method

ecgan:

Closes #1332.

In this PR, we filter out shipping rates suggestions that are not "flat_rate" method, so that subsequently we are able to save the shipping rates suggestions successfully, addressing issue #1332.

### Screenshots:

When we get shipping rates suggestions, it might return shipping rates with "free_shipping" method: (total 30 in the screenshot below)

image

The PR will filter out those that are not "flat_rate" method, and call save shipping rates API: (total 27 in the screenshot below, all of them are "flat_rate")

image

1. Go to WC Setting and setup some shipping zones with mix of different methods.
    - You may refer to issue #1332 and https://github.com/woocommerce/google-listings-and-ads/pull/1363#issuecomment-1078779535 for some examples. 
    - There should be some shipping zones containing only one method, and that one method is a free shipping method.
2. Open your browser dev tool to observe network requests. 
3. Go to Setup MC, proceed to shipping rate section. 
4. There should be a suggestions API call. Look into the response. There should be some non flat rate method.
5. If this is your first time visiting Setup MC shipping rates, you would not have any existing shipping rates data, and Setup MC will automatically save the suggestions by calling batch shipping rate API. Look into the API request. It should not contain free shipping / non flat rate methods. 
    - If you already have existing shipping rates data, you can delete all of them and do a page reload to trigger the automatic suggestions saving.

### DIFF
--- a/js/src/setup-mc/setup-stepper/setup-free-listings/useShippingRatesSuggestions.js
+++ b/js/src/setup-mc/setup-stepper/setup-free-listings/useShippingRatesSuggestions.js
@@ -9,6 +9,7 @@ import { addQueryArgs } from '@wordpress/url';
 import useTargetAudienceFinalCountryCodes from '.~/hooks/useTargetAudienceFinalCountryCodes';
 import useApiFetchEffect from '.~/hooks/useApiFetchEffect';
 import { API_NAMESPACE } from '.~/data/constants';
+import { SHIPPING_RATE_METHOD } from '.~/constants';
 
 /**
  * @typedef {Object} ShippingRatesSuggestionsResult
@@ -21,6 +22,10 @@ import { API_NAMESPACE } from '.~/data/constants';
  *
  * This depends on the `useTargetAudienceFinalCountryCodes` hook,
  * i.e. the target audience countres specified in Setup MC Step 2.
+ *
+ * This will only return shipping rates suggestions with FLAT_RATE method.
+ * Other methods (e.g. free shipping) are filtered out because
+ * they are not well supported in the API and UI yet.
  *
  * @return {ShippingRatesSuggestionsResult} Result object with `loading` and `data`.
  */
@@ -45,9 +50,19 @@ const useShippingRatesSuggestions = () => {
 		}
 	);
 
+	/**
+	 * Shipping rate suggestions data with only FLAT_RATE method.
+	 *
+	 * Other methods (e.g. free shipping) are filtered out because
+	 * they are not well supported in the API and UI yet.
+	 */
+	const data = dataSuggestions?.filter(
+		( el ) => el.method === SHIPPING_RATE_METHOD.FLAT_RATE
+	);
+
 	return {
 		loading: loadingFinalCountryCodes || loadingSuggestions,
-		data: dataSuggestions,
+		data,
 	};
 };
 


jorgemd24 (approved):
Thanks, @ecgan, LGTM 👍 

✅  I have tested with and without this PR and I see that is filtering only flat rates and the API error is not shown.

I just see that when the user has free shipping for a specific country, for example, France:
- The free shipping will not be auto-fill because we are filtering only flat rates.
- The user needs to add this manually and it will be updated in the API as `method: flat_rate` and `rate:0`

So can we not handle this conversion from free-shipping to flat rate (rate:0) automatically? Maybe something to consider in the future?

ecgan:
@jorgemd24 , thank you. 🙂 

> So can we not handle this conversion from free-shipping to flat rate (rate:0) automatically? Maybe something to consider in the future?

Yeah @nima-karimi and I have been having some discussions about this, in Slack threads and in PR comments like https://github.com/woocommerce/google-listings-and-ads/pull/1357#issuecomment-1077836461 and https://github.com/woocommerce/google-listings-and-ads/pull/1363#issuecomment-1078779535. There are quite some complications and difficulties.

For example, in the "France" scenario you mentioned with "free shipping" method without minimum order amount, that is pretty straightforward and it seems makes sense to auto-convert it to `method: flat_rate` and `rate:0`.

However, in https://github.com/woocommerce/google-listings-and-ads/pull/1363#issuecomment-1078779535, I brought up a "Sri Lanka" scenario where there is only one method - "free shipping" - with minimum order amount 50. There is no other shipping method, no "flat rate" method. It should be an unlikely case, because users usually wouldn't setup only one "free shipping" method with minimum order amount, but it is technically possible and can happen. Our code isn't ready to handle such scenario yet. 

That's why in this PR, we are filtering out those "free shipping" method, and I think we can work on them in the next iterations. 

---

Putting down some notes that I thought about "free shipping":

- If there is a country with "free shipping" method with "minimum order amount 50", and it has no flat rate method, then do not return that as a suggestion, and there would be no issue.
- If the country has "free shipping" method with "minimum order amount 50", and has a "flat rate" method with rate 20, then the suggestion shipping rate would be `{ method: "flat_rate", rate : 20, options: { free_shipping_threshold: 50 }}`.
- If the country has "free shipping" method without minimum order amount, then the suggestion shipping rate would be `{ method: "flat_rate", rate : 0, options: { } }`.

PR #1369, opened 2022-03-28
Automatically generate Tracking docs from JSDoc

tomalec:

1. Automatically gather tracking docs and generate a MD document.
2. Format docs for PHP-triggered in the same convention as JS ones.
3. I also changed the event names to contain `gla_` prefix to follow the Track conversions guideline: 
   > They should be hard coded so they are easily searchable in the codebase.

See https://github.com/woocommerce/google-listings-and-ads/pull/1401 for more docs moved

1. Review the https://github.com/woocommerce/google-listings-and-ads/blob/tweak/jsdoc-gen/src/Tracking/README.md
2. Make some change in any `@event` `@fires` or document a new one in JSDoc.
3. `npm i && npm run doc:tracking`

For easier code review, please refer to the [commits list](https://github.com/woocommerce/google-listings-and-ads/pull/1369/commits):
- Copy `woo-tracking-jsdoc` from Pinterest https://github.com/woocommerce/google-listings-and-ads/pull/1369/commits/d7e16328017e8ea63e0bd46364b4c90217536070
- Configure and adjust for GLA https://github.com/woocommerce/google-listings-and-ads/pull/1369/files/d7e16328017e8ea63e0bd46364b4c90217536070..c4e406b05e124d0ee07a6e6805ec170442e4629d
- Fix in `woo-tracking-docs` https://github.com/woocommerce/google-listings-and-ads/pull/1369/commits/e9805a3c8a82fa926a0523bd5690a5cbe071c571

### Additional details:

1. :scroll: :nail_care:  The number of `devDependencies` is pretty big, maybe we can create a Ventures plugin that wraps all atomic ones related to our coding standards into a single one
	- :heavy_plus_sign:  It will be easier to share and maintain that list across multiple extensions. See https://github.com/woocommerce/pinterest-for-woocommerce/pull/434
	- :heavy_minus_sign: It is abstracting away the config, making it a tad harder to inspect.
2. :scroll: :grey_question:  I wonder about the purpose of a few functions in `js/src/utils/recordEvent.js` do we really need those that are just 1:1 proxies? (//cc @ecgan as the first author)  like: 
   ```js
   export const recordFilterEvent = ( data ) => {
	   recordEvent( 'gla_filter', data );
   };
   ```
   - I think having them here leads to obsolete code, like `recordPreLaunchChecklistCompleteEvent` which is still there, the event is still in docs even though, it's not executed anywhere
   - I think, it's also against the principle behind: "They should be hardcoded so they are easily searchable in the codebase." - so that we should have the event name close to the trigger point in the code, not far behind proxies, and libraries.
   - I believe they blur the picture of the actual usage, like: `export const recordSetupAdsEvent = ( target, trigger = 'click' ) => {` may give an impression, that there are many emitters of that event, and that `trigger` may have multiple values. Where in fact, there is [only one emitter, and one possible combination](https://github.com/woocommerce/google-listings-and-ads/blob/0e80cf865ac1b0535c28deee10c4fd89ee679b40/js/src/setup-ads/top-bar/index.js#L19) of values which is `'back', 'click'`
3. :scroll: Emitters that are default exports (`export default class Foo`), are reported as `exports`, this is due to [jsdoc bug](https://github.com/jsdoc/jsdoc/issues/1132). We can use CKEditor's workaround for it https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/jsdoc-plugins/lib/export-fixer/export-fixer.js.  But that's expected to work in jsdoc https://github.com/jsdoc/jsdoc/blob/3025520e156f072bb6de6939193213353dc0c810/packages/jsdoc/test/fixtures/exportdefaultclass.js, so I'd rather try to PR there. Also, event with just `exports` it's an improvement over what we had previously, so we could fix that later.

### Changelog entry

> Tweak - Automatically generate Tracking events docs from JSDoc.


ecgan:
@tomalec ,

> 2\. I wonder about the purpose of a few functions in `js/src/utils/recordEvent.js` do we really need those that are just 1:1 proxies? (//cc @ecgan as the first author)  like:

My memory tells me that idea (`recordFilterEvent` or `record*Event`) came from you 😂 I dug around and found your comment in my PR in Nov 2020 https://github.com/woocommerce/google-listings-and-ads/pull/70#discussion_r529575316, which subsequently leads to change from `recordEvent` to `record*Event` in [`e7a601e` (#70)](https://github.com/woocommerce/google-listings-and-ads/pull/70/commits/e7a601e12351c1fe04c02e04f187bb2f0f6a6bfa). I guess it then went out of control from there to `recordEvent.js`. 😆 

I agree with the bullet points stated in your PR description here, I think it's good to move away from those `record*Event`, especially when they are only used in / called from one place, they should be put inside that one place instead of in `recordEvent.js`, e.g. `recordTableSortEvent` is only used in `AppTableCard`, we don't need `recordTableSortEvent` in `recordEvent.js`.

tomalec:
> My memory tells me that idea (recordFilterEvent or record*Event) came from you joy I dug around and found your comment in my PR in Nov 2020 https://github.com/woocommerce/google-listings-and-ads/pull/70#discussion_r529575316, which subsequently leads to change from recordEvent to record*Event in [e7a601e (#70)](https://github.com/woocommerce/google-listings-and-ads/pull/70/commits/e7a601e12351c1fe04c02e04f187bb2f0f6a6bfa).

:see_no_evil: Then it was a nice miscommunication and went out of control from step 1 :laughing: My point (not stated clear enough) was to provide a more informative function name than `handleToggle` - that states what it actually does. Not to make a convention of wrappers for `recordEvent`

tomalec:
I split (and force pushed) this PR to two 
 1. (this) Automatically gather tracking docs and generate an MD document.
 2. Moving and formatting docs around https://github.com/woocommerce/google-listings-and-ads/pull/1401

.jsdocrc.json
```diff
@@ -0,0 +1,20 @@
...
+  ],
+  "typescript": {
+    "moduleRoot": "js/src"
+  },
+  "templates": {
+    "woo-tracking-jsdoc": {
+      "path": "src/Tracking/README.md"
+    }
+  }
+}
```
puntope: 💅  Empty line missed. Unsure if this is our code standard (to have the empty line) If so, it would be good to have it in the linter. I can happily investigate that.

js/src/components/help-icon-button.js
```diff
@@ -16,7 +16,7 @@ import AppIconButton from '.~/components/app-icon-button';
  *
  * @param {Object} props Props
  * @param {string} props.eventContext Context to be used in `gla_help_click` track event.
- * @return {import(".~/components/app-icon-button")} The button.
+ * @return {import(".~/components/app-icon-button").default} The button.
```
puntope, line 19: ❓  What's the relation of this within the PR?

js/src/reports/utils.js
```diff
@@ -27,9 +27,9 @@ export function getIdsFromQuery( queryString = '' ) {
 }
 
 /**
- * @typedef {import(".").IntervalsData} IntervalsData
- * @typedef {import('.').PerformanceData} PerformanceData
- * @typedef {import(".").TotalsData} TotalsData
+ * @typedef {import("./index").IntervalsData} IntervalsData
```
puntope, line 30: ❓  What's the relation of this within the PR?

package.json
```diff
@@ -61,6 +62,10 @@
 		"eslint-plugin-import": "^2.25.4",
 		"file-loader": "^6.2.0",
 		"h2o2": "^8.2.0",
+		"jsdoc": "^3.6.10",
+		"jsdoc-advanced-types-plugin": "github:tomalec/jsdoc-advanced-types-plugin#add/return-support",
```
puntope, line 66: 💅  ❓  I guess this fork should be potentially added as well to the tool repo?

package.json
```diff
@@ -61,6 +62,10 @@
...
+		"jsdoc-plugin-intersection": "^1.0.4",
+		"jsdoc-plugin-typescript": "^2.0.6",
```
puntope, line 68: ❓  Why do we need TS support?

src/Tracking/README.md
```diff
@@ -284,3 +284,54 @@ $this->conditionally_share_with_tags( Loaded::class );
...
+#### Emitters
+- [`CustomerEffortScorePrompt`](js/src/components/customer-effort-score-prompt/index.js#L49) whenever the CES snackbar (notice) is open
+
+### [`gla_edit_mc_store_address`](js/src/components/contact-information/store-address-card.js#L124)
+"Edit MC store address" Tracking event
+#### Properties
+|   |   |   |
+|---|---|---|
+`path` | `string` | A page from which the link was clicked.
+`subpath` | `string\|undefined` | A subpage from which the link was clicked.
```
puntope, line 320: 💅 

In my MD reader, I the `|` breaks the table like:

Screenshot 2022-04-06 at 10 50 30

After [some investigation](https://www.markdownguide.org/extended-syntax/#escaping-pipe-characters-in-tables) seems like the right way to do this is with HTML character code (`&#124;`). However, I tried it locally and didn't work inside the backticks, so I guess our best try is to do something like

 \`string\`\&# 124;\`undefined\`  

That would render it like `string`&#124;`undefined`

src/Tracking/README.md
```diff
@@ -284,3 +284,54 @@ $this->conditionally_share_with_tags( Loaded::class );
...
+#### Emitters
+- [`StoreAddressCardPreview`](js/src/components/contact-information/store-address-card.js#L144) Whenever "Edit" is clicked.
+
+### [`gla_edit_wc_store_address`](js/src/components/contact-information/store-address-card.js#L23)
+"Edit WC store address" Tracking event
+#### Properties
+|   |   |   |
```
puntope, line 327: 💅 / ❓ 

The table shows an empty header. I will ad headers (like name, type, description) or remove that header. 

Screenshot 2022-04-06 at 10 50 30

woo-tracking-jsdoc/publish.js
```diff
@@ -0,0 +1,131 @@
...
+					);
+					const description = property.description.replace(
+						/\|/g,
+						'\\|'
+					);
+					mdResult += `\`${
+						property.name
+					}\` | \`${ type }\` | ${ description.replace(
+						/\s*\n\s*/g,
+						' <br> '
```
puntope, line 87: ❓ 
Why do we need to replace `/n` with `<br>`?

puntope (commented):
Thanks for implementing this ❤️ 

Left some comments. Only a few of them are blockers.

Thanks :)

puntope:
> 📜 Emitters that are default exports (export default class Foo), are reported as exports, this is due to https://github.com/jsdoc/jsdoc/issues/1132. We can use CKEditor's workaround for it https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/jsdoc-plugins/lib/export-fixer/export-fixer.js. But that's expected to work in jsdoc https://github.com/jsdoc/jsdoc/blob/3025520e156f072bb6de6939193213353dc0c810/packages/jsdoc/test/fixtures/exportdefaultclass.js, so I'd rather try to PR there. Also, event with just exports it's an improvement over what we had previously, so we could fix that later.

That would be cool 🚀  One thing that could be added to that is fix some warnings in the events

Screenshot 2022-04-06 at 10 39 41

src/Tracking/README.md
puntope: Seems like it's a problem with my IDE reader since I see Github is rendering properly the table. 

So this is now 💅  instead of  🚧  and could be ignored

package.json
tomalec: Deffinitely

js/src/components/help-icon-button.js
tomalec: Otherwise, it's not parsable by jsdoc. I guess it's the effect of us using "almost" ES modules ;)

js/src/reports/utils.js
tomalec: Same as for https://github.com/woocommerce/google-listings-and-ads/pull/1369#discussion_r843554736 jsdoc needs a more explicit import name than `.`

package.json
tomalec: Because we use TS notation in our code comments, like  imported types `{import('data/actions').ShippingRate}`

src/Tracking/README.md
tomalec: Thanks for the investigation. Now, I'm not sure which way to go. 
There is no official single spec for Markdown tables. The original one https://daringfireball.net/projects/markdown/syntax#backslash does not mention tables.
On one hand `foo \| bar` is easier to write and read while reading the code and making writing int the easiest is the goal of this PR. Also, the GitHub flavored MD specifies `\|` as an escape https://github.github.com/gfm/#example-200

On the other hand, this clearly doesn't make reading easy for you and other devs using your IDE. :confused:

src/Tracking/README.md
tomalec: The header row is required to construct the table. I just thought that having "name, type, description" repeated everywhere even for single property, could blur the picture

woo-tracking-jsdoc/publish.js
tomalec: To support multiline descriptions, otherwise, a line break would "finish" the markdown table.

tomalec:
> One thing that could be added to that is fix some warnings in the events

Thanks, I'll remove `@type` ones. It was a leftover from my experiments in differentiating tracking events from any other. `@ignore` is tricky, as there are no `@ignore` tags in our code, and we have a lot of `index.js` files :) Will see what I can do.

src/Tracking/README.md
tomalec: :scroll: 
Also, as it changes the package used in Pinterest, maybe we can postpone that change until we have a separate place to host shared tool?

tomalec:
@puntope I believe I addressed all your comments. Would you mind another round?

src/Tracking/README.md
puntope: I don't really mind to be honest,  I can read directly in the code / md  and since now is working I guess its better to give Github more priority than IDE. So my suggestion is to keep it like you did 👌

src/Tracking/README.md
puntope: :til: (row as required prop)

puntope (approved):
Thanks for all the clarifications. This PR is good to go ✅

PR #1370, opened 2022-03-28
Feature/tree select control component

puntope:

This PR implements the Full Tree Select Component 

Check the component thread and features here: 
pdkevu-kA-p2

Go to [Story Book page](https://woocommerce.github.io/google-listings-and-ads/?path=/story/tree-select-control--base)

**Test Control**

1. Click on the Input opens the tree
2. Clicking outside closes the tree
3. It shows the label defined in the `label` param
4. It shows the defined placeholder in the `placeholder` param
5. Searching (more than 3 characters) filters the tree 
6. Selecting options in the tree load tags in the select control
7. Removing the search input also removes the filter applied
8. In the disabled state (with `disabled` prop as true) you cannot edit the search input, remove tags or open the tree
9. With the focus on the input and with some text on it, clicking on Backspace removes the last letter of the text

**Test Tags**

1.  Selecting options create tags in the input
2. Selecting a group loads all the tags of that group 
3. Be aware of the param `maxVisibleTags` if there are more tags than the max number, it will show a button "X More"
4. If you click on "x More" it shows all the tags. 
5. With the focus on the input and without any text on it, clicking on Backspace removes tags

**Test Tree** 

1. The `options` param defines the tree structure 
2. Clicking on a group selects all the sub-options of that group
3. Clicking again unselects all the options of the group
4. Clicking on a single option checks that option and also partially checks the parent.
5. Clicking on the chevron in a parent expands the subtree.  
6. Removing tags also deselect the option.
7. You can navigate over the three with the keyboard.


ecgan:
@puntope , there is a merge conflict in your PR here.

Also, I am thinking, since this PR only contains other smaller PRs that have been reviewed, maybe you want to consider merging it into `develop` branch without needing review? Or you would like someone to test it as per the test instructions first before merging it?

src/Menu/ComponentTest.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
tomalec: Is this file still needed, given we have Storybook?

storybook/main.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * External dependencies
+ **/
+const path = require( 'path' );
+// eslint-disable-next-line import/no-extraneous-dependencies
```
tomalec, line 5: :question: I think if we use this package, we should add it to our devDeps. Or is it skipped as it is used only on deployed site?

puntope: Howdy @tomalec  It generates a static files uploaded to GH Pages. SO I think we can go ahead without adding deps. Besides that it's going to be moved finally to external repo

src/Menu/ComponentTest.php
puntope: Thanks for catching this @tomalec. I removed the entry

PR #1371, opened 2022-03-28
PMax - Remove only campaign and not asset group

mikkamp:

Based on the behaviour change covered in #1366 we only need to delete the main campaign. For the PMax campaigns this was refactored to support removing all the assets for either a SSC or a PMax campaigns. This PR removes that code so only the main campaign gets deleted.

1. When deleting an older SSC we should only see the campaign as removed, see test instructions in #1366 to see what it should look like in the dashboard
2. When deleting a PMax campaign we should see the campaign as removed, but the asset group as enabled
image
3. The location data should also be viewable for the removed campaign
image

### Changelog entry
* Update - Remove only the campaign and not the asset group when deleting a campaign.


ianlin (approved):
Thanks for the changes, it works as expected and I can see the campaign with `performance_max` type being created and removed from the Google dashboard. When the PMax campaign is removed Its asset group is still enabled and its location is still visible.

PR #1373, opened 2022-03-29
Integrate Storybook

puntope:

This PR adds Story Book to the project.

### Screenshots:

Screenshot 2022-03-29 at 10 25 03

1. Checkout this branch
2. Run `npm i` to install all the new dependencies
3. Run `npm run storybook` in order to create a local instance of storybook
4. Inside the StoryBookpage, check the Tree Select Component
5. Run `npm run storybook:build`
6. See how it creates a static storybook instance inside `.storybook/dist`
7. Go to `https://woocommerce.github.io/google-listings-and-ads/?path=/story/tree-select-control--base` 
8. You can see the StoryBook public page
9. Edit `js/src/components/tree-select-control/stories/index.js` (for example in line 79, change ES to FR
10. Run `npm run storybook:deploy` to publish a new version of the public StoryBook
11. Go to `https://woocommerce.github.io/google-listings-and-ads/?path=/story/tree-select-control--base`  and see the updated changes 

### Additional details:

- `/.storybook/main.js` has the main config for the storybook
- `/.storybook/preview.js` loads the general styles
- `/.storybook/styles.scss` WP/WC specific styles

You can create stories inside `js/src` folder, just adding the folder stories inside a component.

- Removed `color-studio` dependency and component test page since they are not anymore necessary.
- Tweak some style variables


tomalec:
Oh my! Thank you so much for setting it up! I was looking forward to having our own storybook/components library gallery, fiddler & docs for so long :)

Can we do something about descriptions in docs, they look corrupted:
image

Also, it seems we now have mixed tabs and spaces in `package.json`, and running `npm install` makes `git diff` dirty by fixing those.

tomalec:
Ouch, seems that Storybook ignores JSDoc for quite a time https://github.com/storybookjs/storybook/issues/9766

In my opinion, it's pretty, bad, because WooCommerce repo shows that keeping docs up to date is problematic. In my opinion, it was problematic there, as the component documentation was separated from the component code. That's why I like JSDoc - you document the code where you write it, so it's easier to keep them in sync and reduces the context switch cost for the author.

puntope:
> Ouch, seems that Storybook ignores JSDoc for quite a time https://github.com/storybookjs/storybook/issues/9766

I'm checking out 'storybook-addon-react-docgen' to see if that could help us.

puntope:
@tomalec 

I fixed the package.json format issue. 

Also, I investigated `addons-docs` and `storybook-addon-react-docgen` as an attempt to have a proper doc using our current JSDOC work. However, the only improvement I could see is to hide the ugly JSODC stringified.... :( 

 As a suggestion I would work on that in a separate PR since it seems to require more work and investigation

package.json
```diff
@@ -33,14 +33,25 @@
...
 	"devDependencies": {
 		"@hapi/hapi": "^20.1.3",
+	  	"@babel/core": "^7.17.8",
+		"@storybook/addon-actions": "^6.4.19",
+		"@storybook/addon-docs": "^6.5.0-alpha.51",
+		"@storybook/addon-essentials": "^6.4.19",
+		"@storybook/addon-interactions": "^6.4.19",
+		"@storybook/addon-links": "^6.4.19",
+		"@storybook/react": "^6.4.19",
+		"@storybook/testing-library": "0.0.9",
```
tomalec, line 54: :nail_care: :scroll: 
Initially, I got this doubt in my [own PR](https://github.com/woocommerce/google-listings-and-ads/pull/1369).
This is quite a list of dependencies. I'm afraid as time goes, we may start to lose control of which one is needed for what, some of them may remain here even if they won't be needed any longer, and may introduce some version compact problems. Also, if we are going to introduce storybook for our other repos, it would be nice to have a unified setup, so we will not have to rethink all those tangled cross dependencies wiring every time.

In general, from the plugin's perspective, we would like to "install storybook" here, and "install jsdoc for tracks" in mine.

Maybe we could wrap those dev ex tools into something like: `@ventures/storybook' and `@ventures/jsdoc-for-tracks`, and install & import it as a one thing.

tomalec (approved):
Tested locally, reviewed the code LGTM.

:scroll: However, I think it would be nice to clean up the "Docs" tab in some future tab.
 - it does not get the docs (list of all params, their descriptions, etc.) from JSDoc 
 - it contains some distracting bits - stringified params in the description mentioned above, params with rather irrelevant controls like: 
 
image

Issue #775, opened 2021-06-10
Handle Jetpack/WPCOM account disconnect and prompt user to reconnect


findingsimple:
### Describe the bug:

If a merchant has GLA setup (and Jetpack Installed) and then they remove Jetpack (and disconnect the connection) the GLA UI will stop working (and I assume some of the underlying functionality - without diving in very deep) - failed requests to WCS.

We need a way to catch the disconnect and prompt a user to reconnect - especially when viewing the UI.

Related to https://github.com/woocommerce/google-listings-and-ads/issues/486

I think one of the ideas was to have a global/central way to handle account connectivity and prompt users to reconnect any accounts that become disconnected.

tomalec:
Here is an example of a merchant experiencing such a problem 4203273-zd-woothemes

It seems there is no error message displayed in the UI even suggesting the Jetpack is an issue but,
- Dashboard and Reports pages throw errors
- Paid Performance Card is forever loading
	
	![image](https://user-images.githubusercontent.com/17435/128548552-9ee3f253-5f28-4dac-8ef1-bec664defdb0.png)

- Settings are forever loading
	
	![image](https://user-images.githubusercontent.com/17435/128548412-e0518811-6574-4d2f-b455-6f5fb660a872.png)

- Product Feed redirect to an empty `/wp-admin/admin.php?page=wc-admin&subpath=%2Freconnect-accounts&path=%2Fgoogle%2Fsettings`
	
	![image](https://user-images.githubusercontent.com/17435/128548466-0c4c3e00-1580-4c02-92dd-6192f9bae06f.png)

mikkamp:
We have the design specs ready how this should be handled in the UI: pcTzPl-mN-p2

However before that's implemented the initial stage would be to detect we've been disconnected in the backend, so we can flag to the UI that a reconnect is needed.

We have a similar "global" connection status for the Google account. In the backend we keep track of this using the GOOGLE_CONNECTED option, whenever we receive a 401 error from a WCS request (proxied or direct), we [mark the option as false](https://github.com/woocommerce/google-listings-and-ads/blob/1.11.1/src/Internal/DependencyManagement/GoogleServiceProvider.php#L196-L198).

In the UI this handled with the following logic:
> If any API endpoint starts with '/wc/gla' and it responds with HTTP status code 401, the frontend side will forcibly redirect the webpage to the Google accounts reconnection page.

[...]

Issue #195, opened 2021-02-04
Reorganize front-end files, to have a dedicated pages folder


tomalec:
_This is an issue to continue a discussion on pages folder started at https://github.com/woocommerce/google-listings-and-ads/pull/192#issuecomment-773084286_

@eason9487 [wrote](https://github.com/woocommerce/google-listings-and-ads/pull/192#issuecomment-773084286):
> Since we did not create a dedicated folder for pages, we can still have the alias for the js/src directory. And add a `pages` alias if we adjust the file structure for putting these pages together in the future.

@ecgan [wrote](https://github.com/woocommerce/google-listings-and-ads/pull/192#issuecomment-773108174):
> Yeah I agree that we should have a dedicated folder for pages, it would be easier to tell which ones are the entry point / components at the highest level. 😄

-----

Personally, I totally upvote this idea as well.

I think we can have a folder for pages that matches the navigation, like:

```
/js/src/
├── components/
├── css/
├── hooks/
├── pages/
│   ├── getting-started
│   ├── dashboard
│   ├── reports
│   │   ├── programs
│   │   └── products
│   ├── product-feed
│   └── settings
├── utils/
├── tasks/
└── wcdl/
```

It would help to:
- identify highest-level components and a relationship between individual components
- make folders smaller

## Doubts

- Should we put `setup-*` pages inside "Getting started", as they do not have their own menu entry after the plugin is set up?
- Shouldn't we have `hooks` per-component, or move them to utils?

## Additional comments

I think it would be good to move all the files with one PR, instead of adapting gradually.

//cc @haszari maybe you would like to comment on that as well.

eason9487:
@puntope wrote in https://github.com/woocommerce/google-listings-and-ads/pull/1374#issuecomment-1084429913:

> 💅 Not related to this PR but after seeing your nice work in refactoring the cards to the components I was wondering if it would be even nicer to organize the components folder.
> 
> * Cards -> All the cards
> * Icons -> Icons
> * Controls -> Inputs / user interaction elements like inputs, selects, etc
> * Rest in the root...

I think it's also a good thought for this follow-up issue.

PR #1374, opened 2022-03-29
Re-connecting WPCOM Account - Adjust the linked accounts section in settings

eason9487:

It's a part of frontend in #775

- Move `<WPComAccountCard>`, `<GoogleMCAccountCard>`, and `<GoogleAdsAccountCard>` to the shared component folder.
   - 💡  Moved a lot of files, adjusted several files, and internalized WPCOM appearance to `<AccountCard>`. This is to give a little better consistency to their composition of components across all similar account cards.
- Change the linked accounts section in settings to use Card UI to present each account.

#### 📷  Not yet connected to a Google Ads account
image

#### 📷  Has a connected Google Ads account
image

1. Go to the Settings page.
2. Check if the UI of **Linked Accounts** section is the same as the design in Figma.
3. Go to the onboarding setup and Google ads account setup pages.
4. Check if all account cards are displayed as the same as before.

### Additional details:

Figma link: eQ9O4m2flzrcAiDMXkOW0m-fi-6257%3A116720

### Changelog entry

> Tweak - Adjust the UI of the linked accounts section on the Settings page.


js/src/components/account-card/index.js
```diff
@@ -38,11 +40,24 @@ const googleLogo = (
...
 const appearanceDict = {
 	[ APPEARANCE.EMPTY ]: {},
+	[ APPEARANCE.WPCOM ]: {
+		icon: wpLogo,
+		title: 'WordPress.com',
+	},
 	[ APPEARANCE.GOOGLE ]: {
 		icon: googleLogo,
-		title: __( 'Google account', 'google-listings-and-ads' ),
+		title: __( 'Google', 'google-listings-and-ads' ),
```
eason9487, line 60: 📝 
I checked several recent designs on Figma. All Google account cards no longer have "account" in the title, so I removed it here.

js/src/components/google-mc-account-card/connected-card.js
```diff
@@ -18,8 +18,15 @@ import useDispatchCoreNotices from '.~/hooks/useDispatchCoreNotices';
...
 import { useAppDispatch } from '.~/data';
 
-const ConnectedCard = ( props ) => {
-	const { googleMCAccount } = props;
+/**
+ * Renders a Google Merchant Center account card UI with connected account information.
+ * It also provides a switch button that lets user connect with another account.
+ *
+ * @param {Object} props React props.
+ * @param {{ id: number }} props.googleMCAccount A data payload object contains user's Google Merchant Center account ID.
```
puntope:
```suggestion
 * @param {{ id: number }} props.googleMCAccount A data payload object containing the user's Google Merchant Center account ID.
```

js/src/components/google-ads-account-card/connected.js
```diff
@@ -0,0 +1,27 @@
+/**
```
puntope: 💅 / ❓  Why in `js/src/components/wpcom-account-card/index.js` this is named `connected-wpcom-account-card.js` whereas here is just `connected.js`?

puntope:
💅  Not related to this PR but after seeing your nice work in refactoring the cards to the components I was wondering if it would be even nicer to organize the components folder. 

- Cards -> All the cards
- Icons -> Icons
- Controls -> Inputs / user interaction elements like inputs, selects, etc 
- Rest in the root...

puntope:
❓ / 🚧  Unsure if this is an issue or it should be like this, but Figma looks different.

- Figma: No underline. (I guess it underlines on hover) 
- PR: Underlined, with kinda outline effect on hover


https://user-images.githubusercontent.com/5908855/161042505-fbebc108-a38b-42e5-9a7b-85e11326c813.mov

puntope:
💅 / ❓ /  🚧  

Probably a heritage from previous implementations but the padding for the footer area in the Ads Cards is smaller in the Implementation than the Figma

Screenshot 2022-03-31 at 13 18 14

I guess in `js/src/wcdl/section/card/footer/index.scss` instead of `--large-gap / 2 ` would be more accurate `--large-gap / 1.5`

puntope (commented):
✅  Overall looking good @eason9487 

Left a few comments 👌

js/src/components/google-mc-account-card/connected-card.js
eason9487: Is there a grammar problem with this?

js/src/components/google-ads-account-card/connected.js
eason9487: I originally intended to minimize the number of renaming files, so I used their original names. Considering that there are two exposed components, which may not match when globally searching for the codebase, I changed their internal names to the same one by https://github.com/woocommerce/google-listings-and-ads/pull/1374/commits/c0f7fe8fbfc6a681816e8af945c49fab6514aede.

eason9487:

> ❓ / 🚧 Unsure if this is an issue or it should be like this, but Figma looks different.
> 
> * Figma: No underline. (I guess it underlines on hover)
> * PR: Underlined, with kinda outline effect on hover
Kapture 2022-03-31 at 20 41 12

I cannot reproduce this problem. GLA should have a higher priority than `@wordpress/components` for button style overriding as the following SCSS styles. Probably it needs your help with checking the override relationship between them from DevTool.
- https://github.com/woocommerce/google-listings-and-ads/blob/af2cb919d1127f75313008ac7bdeee4fae1dcca8/js/src/css/shared/_gutenberg-components.scss#L2
- https://github.com/woocommerce/google-listings-and-ads/blob/af2cb919d1127f75313008ac7bdeee4fae1dcca8/js/src/css/shared/_gutenberg-components.scss#L29-L31

> 💅 / ❓ / 🚧
> 
> Probably a heritage from previous implementations but the padding for the footer area in the Ads Cards is smaller in the Implementation than the Figma

Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/1374/commits/af2cb919d1127f75313008ac7bdeee4fae1dcca8.

> 💅 Not related to this PR but after seeing your nice work in refactoring the cards to the components I was wondering if it would be even nicer to organize the components folder.

Sounds good! I added your comment to related issue #195.

js/src/components/google-mc-account-card/connected-card.js
puntope: - props.googleMCAccount - A data payload object contains user's Google Merchant Center account ID. 

First is referring to googleMCAccount , but then Is saying That a Data Payload object, in general, contains Google Merchant Center Account ID. 

Could be like "props.fishEater - A person eats fish" 

- props.googleMCAccount -  A data payload object containing the user's Google Merchant Center account ID.

It is clear that we are actually referring to googleMCAccount being the payload object containing Google Merchant Center account ID. Also "A data payload that contains the user's Google Merchant Center account ID." works fine. 

Could be like "props.fishEater - A person that eats fish",  "props.fishEater - A person eating fish"  

It was definitely a super minor correction and I understood perfectly the intent of the property. But I as a non-native English speaker also love when someone points out my very usual English mistakes. Definitely I'm not the best to talk about English 🙈

puntope:
> I cannot reproduce this problem. GLA should have a higher priority than `@wordpress/components` for button style overriding as the following SCSS styles. Probably it needs your help with checking the override relationship between them from DevTool.
> 
> *

Crazy, it comes from Automatewoo

Screenshot 2022-03-31 at 15 52 46

Screenshot 2022-03-31 at 15 55 04

puntope:
> Probably a heritage from previous implementations but the padding for the footer area in the Ads Cards is smaller in the Implementation than the Figma

> Adjusted in https://github.com/woocommerce/google-listings-and-ads/commit/af2cb919d1127f75313008ac7bdeee4fae1dcca8.

I was meant the padding top/bottom that makes the footer are higher. Sorry I maybe was not clear or explaining myself in a confusing way.

Screenshot 2022-03-31 at 15 58 00

I guess in js/src/wcdl/section/card/footer/index.scss instead of `--large-gap / 2` would be more accurate to use --> `--large-gap / 1.5`

puntope:
> Crazy, it comes from Automatewoo

As a suggestion, or potential solution maybe we can wrap the shared styles inside `.gla-something` to get priority in the cascade

js/src/components/google-mc-account-card/connected-card.js
eason9487: Thanks for the explanation! I fixed them in https://github.com/woocommerce/google-listings-and-ads/pull/1374/commits/cb541963947610af17db28c699ad8799b15fd383.

eason9487:
@puntope
 
> I was meant the padding top/bottom that makes the footer are higher. Sorry I maybe was not clear or explaining myself in a confusing way.

Oh sorry, it's my fault. I only looked at the screenshot yesterday and started fixing it. 🤦‍♂️ 
 
> I guess in js/src/wcdl/section/card/footer/index.scss instead of `--large-gap / 2` would be more accurate to use --> `--large-gap / 1.5`

The link style button has padding in Figma, and it has more legacy style issues between DEWP, `@wordpress/components`, and GLA. 



Therefore, I tried another way to fix them in \<AccountCard\> scope by https://github.com/woocommerce/google-listings-and-ads/pull/1374/commits/27489370b78e110ce8b9e714fce342a0dda07b8b.

With [another style fixing](https://github.com/woocommerce/google-listings-and-ads/blob/e928c8c2cc35206457be11f122bf916eda820df5/js/src/components/account-card/index.scss#L1-L2) in #1385, the card layout should be much closer to visual design. After all, there is still a 3px gap from the border style, but this will need to further override the default card style, so I guess the difference is probably tolerable.

---

> > Crazy, it comes from Automatewoo
> 
> As a suggestion, or potential solution maybe we can wrap the shared styles inside `.gla-something` to get priority in the cascade

That's a good point! Adding another `.gla-something` can set a higher priority, and this also makes sure GLA won't override other styles that are out of GLA pages. Since this's a bit out of this PR scope, I created a follow-up issue #1391 for it.

puntope (approved):
✅  Thanks for the changes and for creative ing the follow-up PR this is good to go

PR #1384, opened 2022-03-30
Adjusting unit tests to work with PHP 8 and PHPUnit 9

### LINKED ISSUE #850: Support running unit tests on PHP 8
This is split from #828

Due to some version conflicts, we currently do not support running PHPUnit tests on PHP 8. We can use a method similar to WooCommerce core to add support for it:
https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/tests/README.md#running-tests-in-php-8

message-dimke:
Add support for running unit tests with PHP8 and PHPUnit9.

Closes #850 .

- Updating unit tests code to comply with PHPUnit9 codebase;
- Removing some WooCommerce dependencies we are not using from unit tests bootstrap;
- Updating GitHub workflow to run tests with PHP8 and PHPUnit9;

### Screenshots:

Adjusting_unit_tests_to_work_with_PHP_8_and_PHPUnit_9_·_woocommerce_google-listings-and-ads_7f91819

In order to make sure everything is working, go to https://github.com/woocommerce/google-listings-and-ads/pull/1384/checks and check if all the checks pass.


mikkamp (commented):
Thanks for jumping through all the hoops and sticking with it till the finish line, I've been wanting to get our unit tests running on PHP 8 for a long time.

This [article](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/) mentions a pro tip:
> If you want your CI build to fail when PHPUnit encounters PHP native deprecation notices, make sure to add convertDeprecationsToExceptions="true" to your PHPUnit configuration file as the default value for this setting has been changed to false in PHPUnit 9.5.10/8.5.21.

Can we add that to phpunit.xml.dist? It doesn't detect any additional errors, but it's helpful so we don't miss any future deprecations.

Can we also get the README.md updated with how to test locally, by running the line to update phpunit (can't commit those changes). Right now it still mentions we don't support unit testing on PHP 8.

I also see a warning still showing up for the tests on PHP8:
```
1) Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Google\MerchantTest::test_get_products_multiple_pages
The at() matcher has been deprecated. It will be removed in PHPUnit 10. Please refactor your test to not rely on the order in which methods are invoked.
```

I think we can replace that with using `exactly` and then `withConsecutive` to test different parameters like is done [here](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.0/tests/Unit/Jobs/CleanupSyncedProductsTest.php#L74-L79)

mikkamp:
Also noticed that newer versions of PHPUnit create the file `.phpunit.result.cache`. It's recommended to add this file to `.gitignore` so it doesn't accidentally get committed. I had to delete it after testing this PR locally with PHPUnit 9.

mikkamp (approved):
Thanks for all the changes. It looks good to go. It just needs a conflict resolved in one of the test files.

PR #1386, opened 2022-03-31
Only sync shipping rates if shipping time is set to `flat`

### LINKED ISSUE #1380: Shipping time is required if we are syncing shipping rates
### Describe the bug:
Consider the following scenario; A user has stated that they will enter their shipping time settings in Google but they will enter their shipping rates manually or want them to be synced from WooCommerce shipping zones.

image

When we are submitting shipping rates to Merchant Center, the [deliveryTime](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#deliverytime) is a required value. We expect this value to be set in the database when we sync the Merchant Center settings and if it's not in the database, syncing the shipping rates will fail.

### Steps to reproduce:

1. Choose the options mentioned above for the shipping time and shipping rate settings for your free listing
2. Save the settings so that they can be submitted to Google
3. Confirm that the sync fails due to a missing required value (deliveryTime) error.

### Expected behavior:
Shipping rates can't be synced without a deliveryTime; therefore, either the shipping time value must become required if the user has chosen the automatic or manual option for their shipping rates, or we need to fetch the shipping time from their Merchant Center settings if they have already set it up there.

### Actual behavior:
Syncing shipping rates fails if no estimated shipping time has been entered.

### Additional details:
A quick fix would be to make the shipping time input required on the UI if the user has chosen to manually enter their shipping rates or to sync them from their WooCommerce settings.

I don't think Google allows setting a global deliveryTime without assigning it to specific shipping services. This means that if the user chooses the "complex" option for the delivery time they also need to enter their shipping rates in Merchant Center. We might need to change the design on the shipping step so that these two are tied together.

image

cc: @j111q @ecgan

nima-karimi:

Resolves a part of #1380

This PR adds a condition before syncing the shipping rates to check whether the shipping time value is set for all countries and also the shipping_time option is set to "flat" (and NOT "manual"). This confirms that the shipping rates are only synced when the users have set a value for shipping time and prevent sending the rates stored in DB even if the user has chosen the "manual"/complex option in their settings.

1. Edit your free listing and change the shipping rate to "flat" mode and enter a new value
2. Change the shipping time to "manual" mode (manual = "My shipping settings are complex. I will enter my shipping times manually in Google Merchant Center.")
3. Save the settings and confirm that the shipping rates are NOT synced
4. Edit your free listing and add a new country to your target countries.
5. Change the shipping time to "flat" mode but do NOT enter a value for the new countries you added in the previous step
6. Save the settings and confirm that the shipping rates are NOT synced and an error is returned

### Changelog entry

> Fix - Do not sync shipping rates if the shipping time setting is set to complex

### DIFF
--- a/src/API/Google/Settings.php
+++ b/src/API/Google/Settings.php
@@ -5,6 +5,7 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\ShippingRateQuery as RateQuery;
 use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\ShippingTimeQuery as TimeQuery;
+use Automattic\WooCommerce\GoogleListingsAndAds\Exception\InvalidValue;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\TargetAudience;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
@@ -58,19 +59,24 @@ class Settings {
 			$rates = $this->get_shipping_rates_from_database();
 		}
 
-		$this->update_shipping_rates( $rates );
+		$times = $this->get_times();
+
+		$this->update_shipping_rates( $rates, $times );
 	}
 
 	/**
 	 * Submit the given shipping rates to Google.
 	 *
 	 * @param array $rates A multidimensional array of shipping rates and their properties.
+	 * @param array $times An array of estimated shipping times (in days) set for shipping countries.
 	 *
 	 * @see ShippingZone::get_shipping_rates_for_country() for the format of the $rates array.
 	 *
+	 * @throws InvalidValue If no shipping time is found for a country.
+	 *
 	 * @since 1.12.0
 	 */
-	public function update_shipping_rates( array $rates ) {
+	public function update_shipping_rates( array $rates, array $times ) {
 		$settings = new ShippingSettings();
 		$settings->setAccountId( $this->get_account_id() );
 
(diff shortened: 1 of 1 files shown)


src/API/Google/Settings.php
```diff
@@ -342,10 +355,7 @@ protected function create_shipping_service( string $country, string $method, str
 		$rate_groups[] = $this->create_rate_group_object( $currency, $rate, $method );
 		$service->setRateGroups( $rate_groups );
 
-		$times = $this->get_times();
-		if ( array_key_exists( $country, $times ) ) {
-			$service->setDeliveryTime( $this->create_time_object( intval( $times[ $country ] ) ) );
-		}
```
jorgemd24, line 348: ❓ 

Could we not throw the error here? 

```
if ( array_key_exists( $country, $times ) ) {
			$service->setDeliveryTime( $this->create_time_object( intval( $times[ $country ] ) ) );
}
else{
       throw new InvalidValue( 'No estimated shipping time found for country: ' . $country );
}
```

And the rest of the logic could stay as it was.

jorgemd24 (approved):
Thanks @nima-karimi It is working as expected. See a comment that I left about the code.

✅  It is not syncing to Google when the shipping time is manual.
✅  I can complete the onboarding without errors.
✅  I have deleted manually one of the shipping times and I get the error: `No estimated shipping time found for country`

Something that I notice is that even when we are not synced to Google, we still sent the same message:

`"message": "Successfully synchronized settings with Google."` and the front-end is showing a similar message: 

image

src/API/Google/Settings.php
nima-karimi: We could but then there might be cases where we don't sync anything to Google and still return a successful response. In theory, if this exception is thrown it means there's a logic error in our code because in https://github.com/woocommerce/google-listings-and-ads/issues/1380 we will sync the rates only if a shipping time is set. Let me know what you think.

nima-karimi:
Thanks for the review, @jorgemd24 🙏 

> Something that I notice is that even when we are not synced to Google, we still sent the same message:
> "message": "Successfully synchronized settings with Google." and the front-end is showing a similar message:

We are still syncing other user settings to MC, just not the shipping rates. So I think this message is correct.

src/API/Google/Settings.php
jorgemd24: > We could but then there might be cases where we don't sync anything to Google and still return a successful response

Could you elaborate on this a bit more? As far as I understand, having the current implementation of this PR or throwing the exception in `create_shipping_service` would have the same behaviour. If one of the rates does not have a time, it will throw an exception and it is not going to sync any rate.

nima-karimi: Hmm. I'm not sure I understand. Here we only set the delivery time. The actual exception is thrown in the `update_shipping_rates`. Do you mean we should move that validation to here?

jorgemd24: Yes, when I was doing the review I was wondering what was the reason to do the validation in `update_shipping_rates` and not in `create_shipping_service`. If we did it in `create_shipping_service` would not be necessary to add the argument `delivery_days` to the methods: `create_shipping_service` and `create_conditional_free_shipping_service` and all the logic could stay the same as it was before.

nima-karimi: My main reason for refactoring those methods was that the `get_times` method was being called for every country and now it can be called only once. It was unnecessary to query the DB on every for loop where it could've been done just once.

jorgemd24: Aaaah! Thanks for the explanation, now it makes sense. My point of view was more from "throwing the exception if there is no time set" but I did not realize that the issue with `get_times` was there before this PR.

Issue #314, opened 2021-03-10
Investigate if we can reduce the Google Client Library bundle size (and verdor directory size in general)


jconroy:
Currently, the Google Ads client library weighs in at ~96 MB with support for 4 different versions of the API

This is an issue to investigate ways we can minimize this - presumable we can drop the older version support which looks like adds up to ~48mb

findingsimple:
Just noting that I believe we're still seeing issues as a result of size of the extension e.g. "Auto install fails, manual upload also fails."

mikkamp:
Linking a [branch](https://github.com/woocommerce/google-listings-and-ads/tree/cleanup-google-ads-services) that I created a while back. What that does is it removes unused Services in the Google Ads library, which results in about 10MB of files less. Now that we completed all the unit testing for classes that use the Ads library I feel more comfortable pushing that across the line. Are we adding a higher priority to this issue?

findingsimple:
> Are we adding a higher priority to this issue?

I'd like to yep

mikkamp:
We've implemented several steps to reduce the library sizes:
- Script for cleaning up vendor files #516
- Script for removing unused Google Ads Services #1495

We've consistently stayed around the 7MB mark for the final release zip. Which is an acceptable plugin size. Closing this issue for now but it can be reopened if we come across additional issues where the size becomes a problem.

PR #1388, opened 2022-03-31
Add unit test for AdsConversionAction

mikkamp:

This PR adds a unit test for the AdsConversionAction class.

The error codes were changed to map from RPC codes to HTTP error codes (to fall in line with other classes), although the conversion action is part of the Ads account connection controller so it never returns these response codes directly to the UI.

This PR is in preparation for the [branch](https://github.com/woocommerce/google-listings-and-ads/tree/cleanup-google-ads-services) which cleans up unused code from the Ads library (removes @ 10MB of unused code). Before removing parts from the library we want to make sure all the Ads library code we are using is covered with unit tests, so we can be confident there are no failures.

Facilitates additional steps for #314 

Confirm that the unit tests pass and cover all the code in the AdsConversionAction class.

### Changelog entry
* Add - Unit test for AdsConversionAction.


src/API/Google/AdsConversionAction.php
```diff
@@ -133,14 +136,17 @@ public function get_conversion_action( $resource_name ): array {
...
 			do_action( 'woocommerce_gla_ads_client_exception', $e, __METHOD__ );
 			$message = $e->getMessage();
+			$code    = $e->getCode();
+
 			if ( $e instanceof ApiException ) {
 				$message = $e->getBasicMessage();
+				$code    = $this->map_grpc_code_to_http_status_code( $e );
 			}
 
 			throw new Exception(
```
jorgemd24, line 146: At the moment we are testing only if the Exception is an instance of ApiExeption, we could test the case when the Exception is not an instance of ApiExeption to cover all the possibilities.

jorgemd24 (approved):
Thanks, @mikkamp, I have checked the tests and I see that we are testing the main functions used by the Ads Library in the AdsConversionAction class.

I left a small comment about the tests. 

✅  All tests passed.

src/API/Google/AdsConversionAction.php
mikkamp: Thanks, added in ab5b1f5

PR #1390, opened 2022-03-31
Filter Ads accounts

### LINKED ISSUE #317: Filter ads accounts
Currently, we list all existing ads account to a user when connecting, including Manager and test accounts.

I'd need to double-check with @JamesAntisdel but I think I recall briefly discussing that we should filter the list.

Screen Shot 2021-03-10 at 6 57 05 pm

(if possible)

mikkamp:

This PR filters Ads accounts to exclude any test / manager accounts. This is done by sending a separate request to retrieve the account details using a search request. The calls can't be batched because each one needs to send a different account ID. Since we retrieve account details we now also have access to the account name which is included in the response. This changes the responses from an array of integers to an array of account details.

Note: We use search requests instead of GET requests because the latter are limited to a 1000 per day.
Request the account currency was also changed in this PR (although that gets cached so isn't repeated).

When testing I did notice that it's a bit slower and will increase the amount of requests that are sent (one for all account IDs and a request per account). However this is only needed when onboarding/connecting an Ads account, so I think it's still acceptable.

The issue mentions checking if we have ADMIN access to the account which would allow us to link it. In order to do so we would need to expand the query to retrieve data from both `customer` and `customer_user_access`. I haven't included this for now as it could potentially slow down the requests even more.

Closes #317

#### Get accounts
`GET https://domain.test/wp-json/wc/gla/ads/accounts`

Response:
```json
[
	{
		"id": 1234567890,
		"name": "Main Ads Account"
	},
	{
		"id": 2345678901,
		"name": "Secondary Ads Account"
	}
]
```

If you have access to any test or manager accounts, they should no longer show up in the list.

### TODO:
- [x] UI still expects a list of account ID's, so it will need to be adjusted to support the new account details.

### Changelog entry
* Add - Filter Ads accounts to exclude manager and test accounts.
* Add - Return account names when retrieving the list of existing accounts.

### DIFF
--- a/js/src/components/google-ads-account-card/connect-ads/ads-account-select-control.js
+++ b/js/src/components/google-ads-account-card/connect-ads/ads-account-select-control.js
@@ -1,13 +1,12 @@
 /**
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
 import AppSelectControl from '.~/components/app-select-control';
-import toAccountText from '.~/utils/toAccountText';
 
 const AdsAccountSelectControl = ( props ) => {
 	const { accounts, ...rest } = props;
@@ -18,8 +17,13 @@ const AdsAccountSelectControl = ( props ) => {
 			label: __( 'Select one', 'google-listings-and-ads' ),
 		},
 		...accounts.map( ( acc ) => ( {
-			value: acc,
-			label: toAccountText( acc ),
+			value: acc.id,
+			label: sprintf(
+				// translators: 1: account name, 2: account ID.
+				__( '%1$s (%2$s)', 'google-listings-and-ads' ),
+				acc.name,
+				acc.id
+			),
 		} ) ),
 	];
 
--- a/src/API/Google/Ads.php
+++ b/src/API/Google/Ads.php
@@ -16,6 +16,7 @@ use Google\Ads\GoogleAds\Util\V9\ResourceNames;
 use Google\Ads\GoogleAds\V9\Enums\AccessRoleEnum\AccessRole;
 use Google\Ads\GoogleAds\V9\Enums\MerchantCenterLinkStatusEnum\MerchantCenterLinkStatus;
 use Google\Ads\GoogleAds\V9\Resources\MerchantCenterLink;
+use Google\Ads\GoogleAds\V9\Services\CustomerServiceClient;
 use Google\Ads\GoogleAds\V9\Services\MerchantCenterLinkOperation;
 use Google\ApiCore\ApiException;
 use Google\ApiCore\ValidationException;
@@ -40,6 +41,13 @@ class Ads implements OptionsAwareInterface {
 	protected $client;
 
 	/**
+	 * Client for fetching accounts details.
+	 *
+	 * @var CustomerServiceClient
+	 */
+	protected $customer_service;
+
+	/**
 	 * Ads constructor.
 	 *
 	 * @param GoogleAdsClient $client
@@ -49,22 +57,25 @@ class Ads implements OptionsAwareInterface {
 	}
 
 	/**
-	 * Get Ads IDs associated with the connected Google account.
+	 * Get Ads accounts associated with the connected Google account.
 	 *
-	 * @return int[]
+	 * @return array
 	 * @throws ExceptionWithResponseData When an ApiException is caught.
 	 */
-	public function get_ads_account_ids(): array {
+	public function get_ads_accounts(): array {
 		try {
-			$customers = $this->client->getCustomerServiceClient()->listAccessibleCustomers();
-			$ids       = [];
+			$customers = $this->get_customer_service()->listAccessibleCustomers();
+			$return    = [];
 
 			foreach ( $customers->getResourceNames() as $name ) {
-				/** @var string $name */
-				$ids[] = $this->parse_ads_id( $name );
+				$account = $this->get_account_details( $name );
+
+				if ( $account ) {
+					$return[] = $account;
+				}
 			}
 
-			return $ids;
+			return $return;
 		} catch ( ApiException $e ) {
 			do_action( 'woocommerce_gla_ads_client_exception', $e, __METHOD__ );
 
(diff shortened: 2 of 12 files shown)


jorgemd24:
This is a follow-up of pcTzPl-EY-p2#comment-1319

I did a quick testing and I saw that the manager and test accounts are filtered (without checking the access role). If we go ahead with this PR I can do the review and adjust the UI so it is able to work with the new response, a JS dev can review this part.

As far as I know, there is only one place where it displays the existing account ads. See image:

image

In this PR, the accounts are displayed with the following format `accountID : accountName` in the connection test page. Should we follow the same format for the `Select an existing account` section?

https://github.com/woocommerce/google-listings-and-ads/blob/02f4aa3e30d0458f96eb9f696d66c728df02052a/src/ConnectionTest.php#L914

image

cc: @jconroy, @mikkamp, @j111q

mikkamp:
> In this PR, the accounts are displayed with the following format accountID : accountName in the connection test page. Should we follow the same format for the Select an existing account section?

For selecting the Merchant Center account we use the [following format](https://github.com/woocommerce/google-listings-and-ads/blob/1.13.6/js/src/components/merchant-center-select-control/index.js#L19-L20):

```
Account name ・ https://domain.com (123456789)
```

For the Ads account we only have the name and ID, but it would seem more consistent to go with something like:
```
Account name (123456789)
```

src/API/Google/Ads.php
```diff
@@ -49,22 +51,25 @@ public function __construct( GoogleAdsClient $client ) {
...
-	 * @return int[]
+	 * @return array
 	 * @throws ExceptionWithResponseData When an ApiException is caught.
 	 */
-	public function get_ads_account_ids(): array {
+	public function get_ads_accounts(): array {
 		try {
 			$customers = $this->client->getCustomerServiceClient()->listAccessibleCustomers();
-			$ids       = [];
+			$return    = [];
```
jorgemd24: 💅

What do you think if we rename the variable $return for $accounts?

src/API/Google/Ads.php
```diff
@@ -16,6 +17,7 @@
 use Google\Ads\GoogleAds\V9\Enums\AccessRoleEnum\AccessRole;
 use Google\Ads\GoogleAds\V9\Enums\MerchantCenterLinkStatusEnum\MerchantCenterLinkStatus;
 use Google\Ads\GoogleAds\V9\Resources\MerchantCenterLink;
+use Google\Ads\GoogleAds\V9\Services\CustomerServiceClient;
```
jorgemd24: It seems that `CustomerServiceClient` is not used, we could remove it.

src/API/Google/Ads.php
```diff
@@ -153,10 +158,10 @@ public function has_access( string $email ): bool {
 		$ads_id = $this->options->get_ads_id();
```
jorgemd24, line 157: Not introduced in this PR but maybe is a good moment to update it. The description in this function seems incorrect `Check if we have access to the merchant account.`, I think it should be `Check if we have access to the ads account.`

tests/Unit/API/Site/Controllers/Ads/AccountControllerTest.php
```diff
@@ -68,29 +73,29 @@ public function setUp(): void {
...
+			->willReturn( self::TEST_ACCOUNTS );
 
 		$response = $this->do_request( self::ROUTE_ACCOUNTS, 'GET' );
 
-		$this->assertEquals( self::TEST_ACCOUNT_IDS, $response->get_data() );
+		$this->assertEquals( self::TEST_ACCOUNTS, $response->get_data() );
 		$this->assertEquals( 200, $response->get_status() );
 	}
 
 	public function test_get_accounts_with_no_ids() {
```
jorgemd24: We could update no_ids with no_accounts, but maybe it sounds too repetive

jorgemd24 (commented):
Thanks, @mikkamp, I tested it locally and it works as expected 👍 

Something that I noticed is that the schema response for the endpoint `ads/accounts` does not match with the actual response:

https://github.com/woocommerce/google-listings-and-ads/blob/29ea3fbad09ec7edb5d9464b6ef803b9a965b181/src/API/Site/Controllers/Ads/AccountController.php#L170-L186
Actual response: 

```
	[
	   {
		"id": 1234567890,
		"name": "Main Ads Account"
	      }
	]
```

js/src/components/google-ads-account-card/connect-ads/ads-account-select-control.js
```diff
@@ -1,13 +1,12 @@
...
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
 import AppSelectControl from '.~/components/app-select-control';
-import toAccountText from '.~/utils/toAccountText';
```
mikkamp, line 10: As far as I can see the function toAccountText is no longer used after this removal, should we remove that from the utils directory as well?

mikkamp:
> Something that I noticed is that the schema response for the endpoint ads/accounts does not match with the actual response

This is one of those cases where the controller has multiple schema responses. The one that's setup is the response for creating/linking an ads account. When we list accounts the response is not passed through the function `prepare_item_for_response`. Ideally we should discuss these scenarios a bit more in pb0Spc-2pd-p2.

mikkamp:
Thanks for the review @jorgemd24
I've address those points you mentioned, it's ready for another round. In regards to the JS code you added, it's working fine and looks good to me. But feel free to get a second opinion if you think it's necessary.

jorgemd24 (approved):
Thanks, @mikkamp for the adjustments, it is looking good. I am not sure if you missed this comment https://github.com/woocommerce/google-listings-and-ads/pull/1390#discussion_r909514468 or do you prefer to name the variable `$return` rather than `$accounts`?

>This is one of those cases where the controller has multiple schema responses. The one that's setup is the response for creating/linking an ads account. When we list accounts the response is not passed through the function prepare_item_for_response. Ideally we should discuss these scenarios a bit more in pb0Spc-2pd-p

Yes, I think adds confusion for the consumer as the schema is not telling us how our data is structured. 

>In regards to the JS code you added, it's working fine and looks good to me. But feel free to get a second opinion if you think it's necessary.

Thanks, I think it is fine, basically, I am following the same code as [MerchantCenterSelectControl](https://github.com/woocommerce/google-listings-and-ads/blob/1.13.6/js/src/components/merchant-center-select-control/index.js#L19-L20)

mikkamp:
Thanks for the additional review.

> I am not sure if you missed this comment https://github.com/woocommerce/google-listings-and-ads/pull/1390#discussion_r909514468 or do you prefer to name the variable $return rather than $accounts?

I read it but missed it when I went to apply all the fixes. It was copied from some older code and we use that pattern in a few other places in the code, but I agree that it's much clearer to specify what the variable contains.

Everything looks good to go then, I'll merge it once the tests pass. Thanks again.

PR #1394, opened 2022-04-01
Add unit test for AdsReport

mikkamp:

This PR adds a unit test for the AdsReport class.

This PR is in preparation for the [branch](https://github.com/woocommerce/google-listings-and-ads/tree/cleanup-google-ads-services) which cleans up unused code from the Ads library (removes @ 10MB of unused code). Before removing parts from the library we want to make sure all the Ads library code we are using is covered with unit tests, so we can be confident there are no failures.

Facilitates additional steps for #314 

Confirm that the unit tests pass and cover all the code in the AdsReport class.

### Changelog entry
* Add - Unit test for AdsReport.


rawdreeg (approved):
LGTM!

PR #1407, opened 2022-04-05
Fix Fatal error when settings are saved - Migration Class

### LINKED ISSUE #1396: Fatal error when settings are saved with an invalid value
### Describe the bug:
We save the `gla_merchant_center` options as an array of values for the Merchant Center settings. However if this is not stored in the database as an array (serialized data) it causes a fatal, preventing the Merchant onboarding steps from being completed.

```
2022-03-30T07:47:41+00:00 CRITICAL Uncaught TypeError: Return value of Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings::get_settings() must be of the type array, string returned in wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php:545
Stack trace:
#0 wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php(123): Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings->get_settings()
#1 wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php(51): Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings->should_sync_shipping()
#2 wp-content/plugins/google-listings-and-ads/src/API/Site/Controllers/MerchantCenter/SettingsSyncController.php(67): Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings->sync_shipping()
#3 wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php on line 545
```

In two of the cases I looked at the `gla_merchant_center` and it appeared to have been saved with the value `1` (could have also been a true value). However I'm not able to track down how the option got saved with this value, in the code there is only one location where we update the value: https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/API/Site/Controllers/MerchantCenter/SettingsController.php#L73
At first glance I wasn't able to find any way to end up with the value of 1 after sending a request to this endpoint. So I'm assuming there was something external which changed this data.

### Steps to reproduce:
1. Save the option `gla_merchant_center` with the value of 1 
2. Try to complete the onboarding steps
3. See the failure at the last step of onboarding
4. Confirm the error is present in the fatal error log

### Expected behavior:
For the merchant settings value to be overwritten during onboarding so we can proceed to syncing the settings without any fatal errors.

### Actual behavior:
Fatal error occurs when settings are being synced to the Merchant Center.

### Additional details:

The error was reported in the following cases:
https://wordpress.org/support/topic/error-when-completing-the-setup/
4907372-zen
4909627-zen
4906550-zen

In two of the cases the log files showed errors with the phone verification endpoint:

```
2022-03-30T09:31:57+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Merchant::request_phone_verification
{
	"error": {
		"code": 500,
		"message": "Internal error",
		"errors": [
			{
				"message": "Internal error",
				"domain": "global",
				"reason": "backendError"
			}
		],
		"status": "INTERNAL",
		"details": [
			{
				"@type": "type.googleapis.com/google.rpc.ErrorInfo",
				"reason": "internalError",
				"domain": "global"
			}
		]
	}
}
```

```
2022-03-31T07:13:26+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Merchant::request_phone_verification {
  "error": {
    "code": 429,
    "message": "Retries quota exceeded.",
    "errors": [
      {
        "message": "Retries quota exceeded.",
        "domain": "global",
        "reason": "rateLimitExceeded"
      }
    ],
    "status": "RESOURCE_EXHAUSTED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "quotaExceeded",
        "domain": "usageLimits"
      }
    ]
  }
}
```

However taking a closer look at the timestamps, this seems to have happened after the first Fatal error with the settings occurred. So I'm assuming this was a side effect of retrying the onboarding (including phone verification step) several times.

In addition to adjusting the `get_settings` function to always return an array. We might also want to refactor [this line](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/API/Google/Settings.php#L173) to make use of the `get_settings` helper function.

jorgemd24:

Closes #1396.

The new migration class is updating `gla_merchant_center` options:

https://github.com/woocommerce/google-listings-and-ads/blob/3d5ea83a3d6bdb6da60b5d2901fc45d26344ccbf/src/DB/Migration/Migration20211228T1640692399.php#L77

 however if the option `gla_merchant_center` does not exist is used as default value `null` and then is being updated as an empty string `gla_merchant_center` causing the error.

https://github.com/woocommerce/google-listings-and-ads/blob/3d5ea83a3d6bdb6da60b5d2901fc45d26344ccbf/src/DB/Migration/Migration20211228T1640692399.php#L65

In this PR we are only migrating the shipping settings if it has been stored as an array to prevent any errors. In addition to this there are several other locations where we expect the merchant settings to be an array.

When we update the merchant settings if it's loaded as a string, we still index it to update the values. Which causes some odd behaviour because it uses string indexing instead of indexing the array. The result is that the option `gla_merchant_center` is updated with either one letter, or if we set a boolean a value of 1 (which was what was happening in issue #1396).

If we still leave the merchant settings as a string the next error that's encountered is when we try to sync it to Google. This PR also resolves that issue, by enforcing the settings to be read as a string.

#### Succeed through onboarding when starting from scratch

1. Delete all options related to the GLA plugin `DELETE FROM wordpress.wp_options WHERE option_name LIKE '%gla_%' `to simulate a fresh installing.
2. After completing Step 3 of the onboarding process: `Configure your product listings` check that `gla_merchant_center` option is a serialised object and not a string.
3. Complete the onboarding process.
4. No errors should be displayed and the option `gla_merchant_center ` should be a serialised object, not a string.

#### Trigger DB migration with invalid settings

1. Disconnect Merchant center through the connection test page
2. Set the DB option `gla_db_version` to a value older than `1.12.2`
3. Set the DB option `gla_merchant_center` to the value `1`
4. Refresh an admin page to trigger the migration
5. Confirm there are no fatal errors

Before the PR this would have failed with:
```
PHP Fatal error:  Uncaught Error: Cannot unset string offsets in wp-content/plugins/google-listings-and-ads/src/DB/Migration/Migration20211228T1640692399.php:75
Stack trace:
#0 wp-content/plugins/google-listings-and-ads/src/DB/Migration/Migrator.php(47): Automattic\WooCommerce\GoogleListingsAndAds\DB\Migration\Migration20211228T1640692399->apply()
#1 wp-content/plugins/google-listings-and-ads/src/DB/Installer.php(53): Automattic\WooCommerce\GoogleListingsAndAds\DB\Migration\Migrator->migrate()
#2 wp-content/plugins/google-listings-and-ads/src/Installer.php(103): Automattic\WooCommerce\GoogleListingsAndAds\DB\Installer->install()
#3 wp-content/plugins/google-listings-and-ads/src/Installer.php(83): Automattic\WooCommerce\GoogleListingsAndAds\Installer->install()
#4 wp-content/plugins/google-listings-and-ads/src/Installer.php(66): Automattic\WooCommerce in wp-content/plugins/google-listings-and-ads/src/DB/Migration/Migration20211228T1640692399.php on line 75
```

#### Override invalid merchant settings

Since some merchants failed with previous versions we released, we need to ensure the plugin can recover and save valid settings.

1. Set the DB option `gla_merchant_center` to an empty string
2. Send a request to update the merchant settings

`POST https://domain.test/wp-json/wc/gla/mc/settings`
Body:
```json
{
	"shipping_rate": "automatic",
	"shipping_time": "flat",
	"tax_rate": "destination",
}
```

3. Confirm that the data in the response is the same as the one we sent
4. Check the DB option `gla_merchant_center` and ensure it's saved as a serialized array

Previously this would have used string indexing and the response would have returned the data as just the letter `d`, the option would also be saved with this character. If the last option we passed was a boolean it would have been a 1.

#### Sync settings to Google

The original issue some merchants were experiencing was that it was unable to sync to Google and giving a fatal error that get_settings was not returning an array.

1. Set the DB option `gla_merchant_center` with the value `1`
2. Send a request to sync the settings:

`POST https://domain.test/wp-json/wc/gla/mc/settings/sync`
*with an empty body*

3. Confirm that there are no fatal errors

Before the PR this would have failed with:
```
PHP Fatal error:  Uncaught TypeError: Return value of Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings::get_settings() must be of the type array, string returned in wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php:556
Stack trace:
#0 wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php(135): Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings->get_settings()
#1 wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php(52): Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings->should_sync_shipping()
#2 wp-content/plugins/google-listings-and-ads/src/API/Site/Controllers/MerchantCenter/SettingsSyncController.php(67): Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Settings->sync_shipping()
#3 wp-includes/rest-api/class-wp-rest-server.php(1141): Automattic\WooCommerce\GoogleListingsAndAd in wp-content/plugins/google-listings-and-ads/src/API/Google/Settings.php on line 556
```

### Changelog entry

* Fix - Prevent fatal errors when migrating or syncing merchant settings.

### DIFF
--- a/src/API/Google/Settings.php
+++ b/src/API/Google/Settings.php
@@ -132,7 +132,9 @@ class Settings {
 	 * @return bool
 	 */
 	protected function should_sync_shipping(): bool {
-		return in_array( $this->get_settings()['shipping_rate'], [ 'flat', 'automatic' ], true ) && 'flat' === $this->get_settings()['shipping_time'];
+		$shipping_rate = $this->get_settings()['shipping_rate'] ?? '';
+		$shipping_time = $this->get_settings()['shipping_time'] ?? '';
+		return in_array( $shipping_rate, [ 'flat', 'automatic' ], true ) && 'flat' === $shipping_time;
 	}
 
 	/**
@@ -143,7 +145,7 @@ class Settings {
 	 * @since 1.12.0
 	 */
 	protected function should_get_shipping_rates_from_woocommerce(): bool {
-		return 'automatic' === $this->get_settings()['shipping_rate'];
+		return 'automatic' === ( $this->get_settings()['shipping_rate'] ?? '' );
 	}
 
 	/**
--- a/src/API/Google/Settings.php
+++ b/src/API/Google/Settings.php
@@ -170,7 +170,7 @@ class Settings {
 			return false;
 		}
 
-		return 'destination' === ( $this->get_options_object()->get( OptionsInterface::MERCHANT_CENTER )['tax_rate'] ?? 'destination' );
+		return 'destination' === ( $this->get_settings()['tax_rate'] ?? 'destination' );
 	}
 
 	/**
@@ -553,6 +553,7 @@ class Settings {
 	 * @return array
 	 */
 	protected function get_settings(): array {
-		return $this->get_options_object()->get( OptionsInterface::MERCHANT_CENTER );
+		$settings = $this->get_options_object()->get( OptionsInterface::MERCHANT_CENTER );
+		return is_array( $settings ) ? $settings : [];
 	}
 }
--- a/src/API/Site/Controllers/MerchantCenter/SettingsController.php
+++ b/src/API/Site/Controllers/MerchantCenter/SettingsController.php
@@ -66,6 +66,10 @@ class SettingsController extends BaseOptionsController {
 		return function( Request $request ) {
 			$schema  = $this->get_schema_properties();
 			$options = $this->options->get( OptionsInterface::MERCHANT_CENTER, [] );
+			if ( ! is_array( $options ) ) {
+				$options = [];
+			}
+
 			foreach ( $schema as $key => $property ) {
 				$options[ $key ] = $request->get_param( $key ) ?? $options[ $key ] ?? $property['default'] ?? null;
 			}
--- a/src/DB/Migration/Migration20211228T1640692399.php
+++ b/src/DB/Migration/Migration20211228T1640692399.php
@@ -62,7 +62,11 @@ class Migration20211228T1640692399 extends AbstractMigration {
 			// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 			$this->wpdb->query( "ALTER TABLE `{$this->wpdb->_escape( $this->shipping_rate_table->get_name() )}` ALTER COLUMN `method` DROP DEFAULT" );
 
-			$mc_settings = $this->options->get( OptionsInterface::MERCHANT_CENTER, [] );
+			$mc_settings = $this->options->get( OptionsInterface::MERCHANT_CENTER );
+			if ( ! is_array( $mc_settings ) ) {
+				return;
+			}
+
 			if ( isset( $mc_settings['offers_free_shipping'] ) && false !== boolval( $mc_settings['offers_free_shipping'] ) && isset( $mc_settings['free_shipping_threshold'] ) ) {
 				// Move the free shipping threshold from the options to the shipping rate table.
 				$options_json = json_encode( [ 'free_shipping_threshold' => (float) $mc_settings['free_shipping_threshold'] ] );
--- a/src/DB/Migration/Migration20211228T1640692399.php
+++ b/src/DB/Migration/Migration20211228T1640692399.php
@@ -62,7 +62,7 @@ class Migration20211228T1640692399 extends AbstractMigration {
 			// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 			$this->wpdb->query( "ALTER TABLE `{$this->wpdb->_escape( $this->shipping_rate_table->get_name() )}` ALTER COLUMN `method` DROP DEFAULT" );
 
-			$mc_settings = $this->options->get( OptionsInterface::MERCHANT_CENTER );
+			$mc_settings = $this->options->get( OptionsInterface::MERCHANT_CENTER, [] );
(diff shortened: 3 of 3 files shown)


rawdreeg (dismissed):
Thanks for fixing this @jorgemd24 

The issue is solved by this PR.

LGTM!

mikkamp:
I did some testing with this fix. While it does resolve the issue if we start from scratch. For users that have run into the fatal error and the option `gla_merchant_center` is already set as an empty string, it will continue to produce the fatal.

That's because the default `[]` is only passed when the option doesn't exist, but in those cases it would exist so it doesn't resort to the default and it remains an empty string.

rawdreeg (approved):
Thanks, @mikkamp, this addresses all the issues with empty and non-array values saved in the option. 

LGTM!

mikkamp:
Just an additional update after some more testing by @ecgan 

It appears that for new installs the DB migration function is run, and has been running for users that update to 1.12.0, because we forgot to update the tag from 'x.x.x'. So this caused the `gla_merchant_settings` to be updated as an empty string which brought the issues were it would fail during onboarding.

See additional information in p1649226751245569-slack-CK365S85V

divydovy:
Another case 4925334-zen

PR #1408, opened 2022-04-06
Remove `record*Event` utils

tomalec:

As per discussion in https://github.com/woocommerce/google-listings-and-ads/pull/1369#issuecomment-1088714972
This PR removes many `record*Event` utils in favor of using `recordEvent` directly.

`record*Event` function are often 1:1 proxies, like: 
   ```js
   export const recordFilterEvent = ( data ) => {
	   recordEvent( 'gla_filter', data );
   };
   ```
   - I think having them here leads to obsolete code, like `recordPreLaunchChecklistCompleteEvent` which is still there, the event is still in docs even though, it's not executed anywhere
   - I think, it's also against the principle behind: "They should be hardcoded so they are easily searchable in the codebase." - so that we should have the event name close to the trigger point in the code, not far behind proxies, and libraries.
   - I believe they blur the picture of the actual usage, like: `export const recordSetupAdsEvent = ( target, trigger = 'click' ) => {` may give an impression, that there are many emitters of that event, and that `trigger` may have multiple values. Where in fact, there is [only one emitter, and one possible combination](https://github.com/woocommerce/google-listings-and-ads/blob/0e80cf865ac1b0535c28deee10c4fd89ee679b40/js/src/setup-ads/top-bar/index.js#L19) of values which is `'back', 'click'`

After the changes in this PR, `.~/utils/recordEvent.js` contains only one export of a util that actually has some logic, and a bunch of JSDocs for track events that are fired from multiple places.

The changes consist mostly of moving code around, but we could potentially affect a lot of flows (not covered by tests).

0. Enable track events logging in browser's console `localStorage.setItem( 'debug', 'wc-admin:*' );` 
 
##### `recordTableSortEvent` & `recordColumnToggleEvent`

1.  Go to:
	1. `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard`
	2. `/wp-admin/admin.php?page=wc-admin&reportKey=programs&path=%2Fgoogle%2Freports`
	3. `/wp-admin/admin.php?page=wc-admin&reportKey=products&path=%2Fgoogle%2Freports`
4. Toggle some columns (⋮ Kebab icon),  `wcadmin_gla_table_header_toggle { column, report, status }` should be fired
5. Sort table

1. Go to `https://wp582wc590.test/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed`
2. Sort and toggle columns. No event should be fired, but component should work without an error.

##### `recordTooltipViewedEvent`

1. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed`
2. Click on "Overview (?)" tooltip.
3. `wcadmin_gla_tooltip_viewed {id: 'product-status'} ` should be fired

##### `recordDatepickerUpdateEvent` & `recordFilterEvent`

1. Go to 
   - `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Freports`
   - `/wp-admin/admin.php?page=wc-admin&reportKey=products&path=%2Fgoogle%2Freports`
3. Change something in datepicker, and click "Update"
4. `wcadmin_gla_datepicker_update {report: 'reports-programs', period, compare} ` should be fired
5. Change something in products/programs filter
6. `'gla_filter', { report, filter, [filterVariations]}` should be fired

##### `recordChartTabClickEvent`

1.  Go to:
	1. `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard`
	2. `/wp-admin/admin.php?page=wc-admin&reportKey=programs&path=%2Fgoogle%2Freports`
	3. `/wp-admin/admin.php?page=wc-admin&reportKey=products&path=%2Fgoogle%2Freports`
2. Click on any performance summary card
3. `wcadmin_gla_chart_tab_click {report, context}` should be fired.

##### `recordLaunchPaidCampaignClickEvent`

1. Go to
 - setup ads `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads`
 - create paid campaign
3. Finish the creation flow 
4. `'gla_launch_paid_campaign_button_click', { audiences, budget }` should be fired

##### `recordSetupAdsEvent`
1. Go to Ads setup `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads`
   1. Click on back `<` button, `'gla_setup_ads', { target: 'back', trigger: 'click' }` should be fired
   2. 
##### `recordSetupMCEvent`

1. Go to MC setup `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-mc`
   1. Click on back `<` button, `'gla_setup_mc', { target: 'back', trigger: 'click' }` should be fired
   3. Continue through steps, `'gla_setup_mc', { target, trigger: 'click' }`  should be fired for `'step{1-3}_continue'`

### Changelog entry

> Tweak - Refactor, remove `record*Event` utils.


tomalec:
When doing record*Event I noticed that `gla_setup_mc` is rather confusing. It's properties are described as:
> - `target` Button ID
> - `trigger` Action (e.g. `click`)

I see a few problems here:
1. `trigger` is always set to `click` - do we really need such a property, that always has a static value?
2. `target` property contains an ID of the button that **triggered** the action, not the target of the merchant's flow. (This may relate to DOM `event.target`, but this is a technical implementation detail, and I'm not sure if that matters for people analyzing data)

So when a merchant clicks a "Continue" button on step 1 and goes to step two, we fire `gla_setup_mc { trigger: 'click', target: 'step1'}`, whereas when I'd be to describe in plain English, what an event just happened I'd say:
> An event in `MC setup` just happened: the merchant `trigger`ed it by clicking/tapping/activating the "Continue" button on `step 1` and is being `target`ted to `step 2`

So, `gla_setup_mc { trigger: 'step1-continue', target 'step2' }`

js/src/product-feed/product-feed-table-card/index.js
```diff
@@ -17,12 +17,13 @@ import {
...
 	TablePlaceholder,
 } from '@woocommerce/components';
+import { recordEvent } from '@woocommerce/tracks';
 import classnames from 'classnames';
 
 /**
  * Internal dependencies
  */
-import recordEvent, { recordTablePageEvent } from '.~/utils/recordEvent';
+import { recordTablePageEvent } from '.~/utils/recordEvent';
```
puntope, line 26: ❓  Why we are maintaining this `recordTablePageEvent` in the utils file ?

puntope (approved):
✅  LGTM

I was testing manually the events mentioned in the description as well as the code.

❓  Only one question regarding why we are actually maintaining `recordTablePageEvent` in the utils file.

puntope:
> When doing record*Event I noticed that gla_setup_mc is rather confusing.

Makes sense your suggestions for me. However maybe it's good to check with Jason.. unsure if it is has some background decision for using those properties.

js/src/product-feed/product-feed-table-card/index.js
tomalec: It is used in two files `js/src/product-feed/issues-table-card/index.js`, `js/src/product-feed/product-feed-table-card/index.js`, so we need to share it somewhere, so I left it where it was to reduce the number of changes.

Do you have a better idea? Maybe somewhere in `js/src/product-feed/`, but then we would need a dedicated file anyway and I'm not sure if it's worth scattering those functions across more and more files.

tomalec:
@jconroy if you happen to have some time, I'd be happy to know your opinion on https://github.com/woocommerce/google-listings-and-ads/pull/1408#issuecomment-1090434772

PR #1409, opened 2022-04-06
Edit Free Listings - Remove corresponding shipping rate and shipping time when target audience is changed

### LINKED ISSUE #1304: Shipping rates UI changes: Incorrectly saved shipping settings when editing free listings
### Describe the bug:

- This issue is duplicated in:
	- https://github.com/woocommerce/google-listings-and-ads/issues/1387 by @nima-karimi .
	- https://github.com/woocommerce/google-listings-and-ads/issues/1402 by @ecgan .

When editing free listings, the final saved shipping settings (shipping rates and shipping times) could be incorrect if we change the target audience.

https://user-images.githubusercontent.com/17420811/157383414-12ba6a77-c099-4389-8612-21635229c5aa.mp4

In issue https://github.com/woocommerce/google-listings-and-ads/issues/1387, @nima-karimi mentioned:

> When editing a free listing, the shipping time country selector input will not get updated based on the countries selected at the previous step in the onboarding wizard. It seems to retain the values previously entered by the user (and stored in the database).
> 
> This leads to another issue where if a user changes the target countries but keeps the number of countries selected by this input to the same, the validator passes and allows the user to save their shipping settings.

In issue https://github.com/woocommerce/google-listings-and-ads/issues/1402, @ecgan mentioned:

> (Note that this was discussed / introduced in https://github.com/woocommerce/google-listings-and-ads/pull/444#issue-853572386.)
> 
> To fix this issue, the easiest way would be to use the similar `useAutoClearShippingEffect` approach in Edit Free Listings. Alternatively, we can check the deleted country and removed them from shipping rates and shipping times accordingly.

### Steps to reproduce:

1. Open browser's DevTool and navigate to the Network tab.
1. In the onboarding setup, select **United States** and 'Japan' as target audience countries and finish the setup.
1. Go to edit free listings.
1. Remove one of target audience countries at step 1. For example, remove **Japan**.
1. Continue to step 2, the removed country still appears in all shipping settings and can be edited.
1. Click the "Save changes" button. Refresh the current webpage after successful saving.
1. Check the Network tab of DevTool. The shipping settings were saved incorrectly/inconsistently:
   - The `countries` property in the `mc/target_audience` response should have only `'US'` (**United States**), which is correct.
   - The `mc/shipping/rates` response still has the shipping rate entity for `'JP'` (**Japan**).
   - The `mc/shipping/times` response still has the shipping time entity for `'JP'` (**Japan**).

### Expected behavior:

After successful saving shipping settings in the free listings edit page, the saved data should be consistent across these APIs.
- **GET** `mc/target_audience`
- **GET** `mc/shipping/rates`
- **GET** `mc/shipping/times`

### Actual behavior:

The shipping settings could be saved incorrectly/inconsistently.

### LINKED ISSUE #1402: Edit free listings - change shipping rates and shipping times when audience countries are changed
### Describe the bug:

In Setup MC, when users go through shipping settings, go back to change / delete audience countries and go to shipping settings again, the shipping settings are cleared and users have to set them up again. This is done by using the `useAutoClearShippingEffect`. This ensures that a country that is not a target audience country will not have the corresponding shipping settings.

In Edit Free LIstings however, the logic is different. When users go back to change / delete audience countries and go to shipping settings again, the shipping settings are **not** cleared. As a result, a country that has been deleted in target audience would still show up in shipping rate and shipping times, and users are still able to save the settings successfully. This should be deemed as a bug. (Note that this was discussed / introduced in https://github.com/woocommerce/google-listings-and-ads/pull/444#issue-853572386.)

To fix this issue, the easiest way would be to use the similar `useAutoClearShippingEffect` approach in Edit Free Listings. Alternatively, we can check the deleted country and removed them from shipping rates and shipping times accordingly.

### Steps to reproduce:

1. Go to Edit Free Listings.
2. Remove an existing country in target audience.
3. Go to shipping settings.

### Expected behavior:

The removed country should not be in shipping rates and shipping times.

### Actual behavior:

The removed country is still in shipping rates and shipping times.

ecgan:

Closes #1304 .

In this PR, when users remove a target audience country in Edit Free Listings, the corresponding shipping rate and shipping time will also be removed.

### Screenshots:

📹  Demo video with my voice demonstrating changing selected countries:

https://user-images.githubusercontent.com/417342/162034256-0aca8494-5719-47cc-9c38-4b755527e837.mp4

📹  Demo video with my voice demonstrating changing target audience location from 'all' to 'selected':

https://user-images.githubusercontent.com/417342/162438798-9b8d85de-7daa-4493-a7bc-851c1dac9a87.mov

Changing individual selected countries:

1. Go to Edit Free Listings. Without making any changes, look at both Step 1 and Step 2 (shipping rates and shipping times), and determine the country that you want to remove.
2. Go to Step 1. Remove the country from the target audience list.
3. Go to Step 2. The country that has been removed from the target audience should not be in the shipping rates and shipping times.
4. Save changes. Reload the page. Make sure the settings are saved and loaded correctly.

Changing target audience location:

1. Go to Edit Free Listings. In Step 1, "selected countries only" should be selected. Take note of the selected countries.
2. Change the option to "All countries".
3. Go to Step 2. You would need to add in more rates and times for more countries. Add some countries in.
4. Go back to Step 1, change the option back to "selected countries only". The selected countries should be same as the countries you took note of earlier.
5. Go to Step 2. The rates and times you added should have been removed, because they are no longer applicable.
6. Save changes. Reload the page. Make sure the settings are saved and loaded correctly.

### Additional details:

With this PR, the countries in shipping rates and shipping times should always be the same as (or a subset of) the target audience countries. We can then simplify some of the related code such as this one in `EditRateModal`:

https://github.com/woocommerce/google-listings-and-ads/blob/0ad357a9e51c315bdea95a50cd38713042eded75/js/src/components/shipping-rate-section/estimated-shipping-rates-card/edit-rate-modal.js#L33-L36

### Changelog entry

> Fix - Remove corresponding shipping rate and shipping time when target audience is changed.


puntope (approved):
✅  LGFM

I was doing some testing. All good 

However ,  is not related to this PR, but I found strange when the Save Setting button is just disabled but you can kinda click on it. (Adding again a country after deleting it) 

Also it doesn't show any error or indication to the user of whats going on, so that could be frustrating for the user. I knew because I watched the video but otherwise I would find hard to know that I need to add again the country shipping cost and time...  

Unsure if this was reported somewhere, but if not, I think this should be evaluated as a new issue. (I can created that if so)

ecgan:
@puntope , thanks for the review. I realized there was an issue, where if we change the target audience option from 'all' to 'selected', the shipping rates and shipping times may be incorrect. I have pushed another commit [`3cd09cb` (#1409)](https://github.com/woocommerce/google-listings-and-ads/pull/1409/commits/3cd09cb8e4f9c0d568174ad3c85d94f3cb266b53) (just a few lines of changes) to fix this. I have also updated the PR description with another video and test instructions. Could you give this another quick round of review please? 🙏 

> found strange when the Save Setting button is just disabled but you can kinda click on it.

Ooh interesting. I don't know why, but I think it might be intended design behavior from `@wordpress/components`, as I see it behaving the same in https://wordpress.github.io/gutenberg/?path=/story/components-button--disabled. You can click on the disabled button and there is a flashing styling. 🤷 

> Also it doesn't show any error or indication to the user of whats going on, so that could be frustrating for the user. I knew because I watched the video but otherwise I would find hard to know that I need to add again the country shipping cost and time

Yeah, we have been having it that way, there wasn't a design or plan for it. Agree that it really isn't ideal. I'll dig around and open an issue for it. (Update: I've created issue https://github.com/woocommerce/google-listings-and-ads/issues/1415 for this.)

puntope:
> @puntope , thanks for the review. I realized there was an issue, where if we change the target audience option from 'all' to 'selected', the shipping rates and shipping times may be incorrect. I have pushed another commit [3cd09cb (#1409)](https://github.com/woocommerce/google-listings-and-ads/pull/1409/commits/3cd09cb8e4f9c0d568174ad3c85d94f3cb266b53) (just a few lines of changes) to fix this. I have also updated the PR description with another video and test instructions. Could you give this another quick round of review please? 🙏

Mmm gotcha, Good you catch that. I'll test again

puntope:
I did the recommended Happy Path as well as some weirdos tests and I saw this.

https://user-images.githubusercontent.com/5908855/162446491-cb1b5cd0-876f-46d2-8a7d-bb75730a3338.mov

When I save with all countries selected and the option "My shipping settings are complex. I will enter my shipping rates and times manually in Google Merchant Center." and then switch back to "My shipping settings are simple. I can manually estimate flat shipping rates" then All countries appear with the last value selected in Shipping times.

Whereas if I change to all countries and I go directly to  "My shipping settings are simple. I can manually estimate flat shipping rates" then I need to add a new time to continue.

ecgan:
@puntope , thanks for weirdos tests! 😃 👍 

I looked into the code and found out that _`deleteShippingTimes` was never called in the Edit Free Listings flow_. What this means is that in your video, the big bunch of shipping times with "3 days" value are coming from the backend API / wp.data store, it just appears to be deleted in the front-end but not deleted in the database. If you try to delete the shipping times directly by clicking on the delete button in edit time modal, and then save changes, and then reload the page, you will see the shipping times still show up. I have created issue https://github.com/woocommerce/google-listings-and-ads/issues/1416 (with my demo video inside) for this issue.

With the discovery of this issue, I'm going to put this PR on-hold first, and revisit this PR after we get the above issue resolved.

js/src/edit-free-campaign/index.js
```diff
@@ -178,6 +178,40 @@ export default function EditFreeCampaign() {
...
+		updateTargetAudience( newTargetAudience );
+
+		/**
+		 * If users removed a target audience country,
+		 * or changed the target audience location from 'all' to 'selected',
+		 * we remove the country from shipping rates and shipping times.
+		 */
+		if (
+			newTargetAudience.countries.length <
+				targetAudience.countries.length ||
```
eason9487: ❓ 

In step 1, if remove a country and add the removed one back, which means the selected countries are the same as the saved ones after these operations, the unsaved data prompt will be shown.

https://user-images.githubusercontent.com/17420811/163166983-50c8b38d-8756-48f6-8f32-662c6859674b.mp4

ecgan: Thanks for catching that! 👍 

Now I'm thinking, instead of putting the code under `handleChooseAudienceChange`, we can put it inside `handleChooseAudienceContinue` (when users click on the "Continue" button) and `handleStepClick` (when users go from Step 1 to Step 2). This means that if users remove a country and add them back and then leave, the unsaved data prompt should not appear. I'll work on the change.

ecgan: @eason9487 , turns out the issue in your video happens in the `develop` branch too. The problem is that even though in the UI the form fields look the same and there are "no changes" to the users, the underlying data `targetAudience.countries` and `savedTargetAudience.countries` have the same country codes but in different orders, and hence does not satisfy the `isEqual` check on target audience data.

https://user-images.githubusercontent.com/417342/163433622-3900e98f-a045-4332-90ab-b9a60e2286e0.mp4

I'm thinking we can still proceed with the changes I proposed above, and open another issue for dealing with the problem in the video separately. Sounds good?

eason9487: > turns out the issue in your video happens in the `develop` branch too.

Thanks for the finding! I didn't notice that. 😂 

> Now I'm thinking, instead of putting the code under `handleChooseAudienceChange`, we can put it inside `handleChooseAudienceContinue` (when users click on the "Continue" button) and `handleStepClick` (when users go from Step 1 to Step 2).

> I'm thinking we can still proceed with the changes I proposed above, and open another issue for dealing with the problem in the video separately. Sounds good?

It sounds good to open another issue for the incorrect comparison of `targetAudience.countries`.

However, I'm not so sure would it be a good idea to move the code to `handleChooseAudienceContinue` and `handleStepClick`. The shipping rates and times could be restored by using `savedShippingRates` and `savedShippingTimes` in the `handleChooseAudienceChange` instead. And, with the new comparison function `getDifferentShippingTimes` added in #1420, the `didTimesChanged` could be changed to be compared in the same way as shipping rates. In this way, it could more appropriately determine if there is unsaved shipping data.

ecgan: > The shipping rates and times could be restored by using `savedShippingRates` and `savedShippingTimes` in the `handleChooseAudienceChange` instead.

Interesting. So I guess you are saying when we remove a country and add the removed one back, we try to find the shipping rate and shipping time for that country from `savedShippingRates` and `savedShippingTimes` and restore it back, right?

I thought of one scenario which might end up being funny / buggy:

1. Let's say we have a country "JP" with shipping rate 12 and shipping time 12. (These values are stored in `savedShippingRates` and `savedShippingTimes`.)
2. We go to Step 2, change the values to shipping rate 10 and shipping time 10. (These values are stored in the "form" / `useState`.)
3. We go to Step 1, remove the country JP. (Here, form useState would no longer have the shipping rate 10 and shipping time 10 for JP, because we remove them in `handleChooseAudienceChange`.)
4. We add the country JP back in.
    - Here, we should restore back the shipping rate and shipping time value for JP. What should they be: 12 or 10? 🤔 I'm thinking it should be the latest value 10, but we already lose that value here.
 	
> And, with the new comparison function `getDifferentShippingTimes` added in #1420, the `didTimesChanged` could be changed to be compared in the same way as shipping rates. In this way, it could more appropriately determine if there is unsaved shipping data.

Yup, that makes sense. I'll look into that. 🙂 👍

eason9487: > I thought of one scenario which might end up being funny / buggy: [...]

That's a very detailed consideration! 👍  I didn't think of this scenario. Maybe moving the filtering of shipping rates/times to the render layer could handle all the form operation combinations. For example:

```js
const finalCountries = getFinalCountries( targetAudience );
const [ formShippingRates, formShippingTimes ] = useMemo( () => {
	const countries = finalCountries || [];
	return [
		shippingRates.filter( ( { country } ) =>
			countries.includes( country )
		),
		shippingTimes.filter( ( { countryCode } ) =>
			countries.includes( countryCode )
		),
	];
}, [ finalCountries, shippingRates, shippingTimes ] );

// ... omitted

const loadedShippingRates = ! hfrShippingRates ? null : formShippingRates;
const loadedShippingTimes = ! hfrShippingTimes ? null : formShippingTimes;

// ... omitted

const didRatesChanged = hasUnsavedShippingRates(
	formShippingRates,
	savedShippingRates
);
```

ecgan: > Maybe moving the filtering of shipping rates/times to the render layer could handle all the form operation combinations. For example: [...]

I think that is going to make things more complicated. 😅 With that approach, we will have the following states:

- `savedShippingRates`: data from backend API.
- `shippingRates`: this will then act like a "cached" / "staging" area to store shipping rates. When users modify shipping rates values in Step 2, this will be updated.
- `formShippingRates`: we get this by filtering `shippingRates` against selected countries in Step 1. 

Then, here's another scenario which might end up looking buggy:

1. Go to Edit Free Listings.
2. In Step 1, users add a new country "SG".
3. In Step 2, users add shipping rate 22 and shipping time 22 for "SG". (`shippingRates` is updated here.)
4. Go back to Step 1, and remove country "SG". 
5. Go to Step 2, "SG" is not visible in shipping rates and shipping times. Users would think it has been removed. (However, "SG" is still in `shippingRates`. It is not visible in UI because we are using `formShippingRate`.)
6. Go back to Step 1, add back the country "SG".
7. Go to Step 2, users would see shipping rate 22 and shipping time 22 for "SG", which looks like a bug.

---

After some thinking from the above scenarios, I think an important thing to point out is about _the switching between Step 1 and Step 2_. When users switch from one step to another, it acts like a confirmation that "I am done with this step, you can apply the changes and show me the updated things in the next step". What this means is that I think it makes sense to put the state update inside `handleChooseAudienceContinue` (when users click on the "Continue" button) and `handleStepClick` (when users go from Step 1 to Step 2) (as per my earlier comment https://github.com/woocommerce/google-listings-and-ads/pull/1409#discussion_r850596474).

I'll give this a try and we see how it goes. 😄

ecgan: > It sounds good to open another issue for the incorrect comparison of `targetAudience.countries`.

👍 I've created issue https://github.com/woocommerce/google-listings-and-ads/issues/1431 for this.

ecgan: > I'll give this a try and we see how it goes. 😄

I made the changes in [`cf9a72f` (#1409)](https://github.com/woocommerce/google-listings-and-ads/pull/1409/commits/cf9a72ffec704e9abf2509358fb64227ceb5d22b), by doing the state update inside `handleChooseAudienceContinue` (when users click on the "Continue" button) and `handleStepClick` (when users go from Step 1 to Step 2).

ecgan:
I've merged PR #1420 into this PR, and added two commits [`cf9a72f` (#1409)](https://github.com/woocommerce/google-listings-and-ads/pull/1409/commits/cf9a72ffec704e9abf2509358fb64227ceb5d22b) and [`8e0ae93` (#1409)](https://github.com/woocommerce/google-listings-and-ads/pull/1409/commits/8e0ae939a647850e74babe1cb4d1497dee1b9be3) based on discussions from https://github.com/woocommerce/google-listings-and-ads/pull/1409#discussion_r851390835. 

This PR is ready for another round of review.

js/src/edit-free-campaign/index.js
eason9487: > I think that is going to make things more complicated.

If the requirement asked for handling that scenario, then the complexity from the code perspective was not a blocker for implementation.

> Then, here's another scenario which might end up looking buggy:
> 
> [...]
> 6. Go back to Step 1, add back the country "SG".
> 7. Go to Step 2, users would see shipping rate 22 and shipping time 22 for "SG", which looks like a bug.

If the form value is entered by the user, then it could also be interpreted as convenient handling, which let the user don't need to re-enter the value again when they try to add back a country.

> After some thinking from the above scenarios, I think an important thing to point out is about _the switching between Step 1 and Step 2_. When users switch from one step to another, it acts like a confirmation that "I am done with this step, you can apply the changes and show me the updated things in the next step".

Thanks for the explanation! This looks reasonable as well.

eason9487 (commented):
The shipping times with the same values but in a different order will be considered as changed data. Is this issue expected to be handled in this PR?
https://github.com/woocommerce/google-listings-and-ads/blob/8e0ae939a647850e74babe1cb4d1497dee1b9be3/js/src/edit-free-campaign/index.js#L131

ecgan:
> The shipping times with the same values but in a different order will be considered as changed data. Is this issue expected to be handled in this PR?

@eason9487 , I'll handle it as a separate PR for issue https://github.com/woocommerce/google-listings-and-ads/issues/1431 (the issue is created as per https://github.com/woocommerce/google-listings-and-ads/pull/1409#discussion_r851369454).

eason9487:
> I'll handle it as a separate PR for issue #1431 (the issue is created as per [#1409 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/1409#discussion_r851369454)).

Got it. Since #1431 and https://github.com/woocommerce/google-listings-and-ads/pull/1409#discussion_r851369454 don't mentioned the issue of comparing shipping time data, I thought the fix was missed in this PR.

eason9487 (approved):
LGTM.

ecgan:
> Got it. Since #1431 and [#1409 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/1409#discussion_r851369454) don't mentioned the issue of comparing **_shipping time_** data, I thought the fix was missed in this PR.

Oops sorry, I just realized you were talking about **_shipping times_**, and the issue https://github.com/woocommerce/google-listings-and-ads/issues/1431 I linked above is about _target audience selected countries_, I mistakenly thought we were talking about that. 🤦 

I have created issue https://github.com/woocommerce/google-listings-and-ads/issues/1433 to handle the shipping times issue.

ecgan:
I have removed the `status: blocked` label, it is no longer blocked since https://github.com/woocommerce/google-listings-and-ads/pull/1409#issuecomment-1100280042, after merging PR https://github.com/woocommerce/google-listings-and-ads/pull/1420 into this PR.

PR #1410, opened 2022-04-06
Code refactor: minimum order modals

ecgan:

In this code refactor PR, we are making some code changes to the minimum order modals (add modal and edit modal):

- Rename "Modal" to "FormModal", to more accurately indicate that the component contains a modal wrapped in a Form with form state in it.
- Rename `value` prop to `initialValues` prop. We shouldn't use `value` because it's not really a controlled component and we can't really dynamically control the value shown in the modal by modify `value`. `initialValues` will be passed directly into the form's `initialValues` props.
- Rename `onChange` prop to `onSubmit` prop. Same reason as above, we shouldn't use `onChange` because it's not really a controlled component and we don't really get the value while the users are changing them in the modal. We only get the value when users submit the form.
- Add JSDoc for the component and its props.

The functionality of the modals remain the same.

The minimum order modals will be used as the reference to modify the add rate modal and edit rate modal later.

### Screenshots:

You can access the add modal and edit modal by clicking on the add button and edit button:

image

1. Go to Setup MC or Edit Free Listings, shipping rate section. Make sure you have some shipping rates bigger than 0. Select "yes" for offer free shipping.
2. In the mInimum order card, click on edit button, modify some values, and click on update button. It should work as per existing behavior.
3. In the mInimum order card, click on edit button, and click on delete button. It should work as per existing behavior.
4. Click on the add button, enter some values, and click on add. It should work as per existing behavior.


js/src/components/shipping-rate-section/minimum-order-card/add-minimum-order-form-modal.js
```diff
@@ -14,8 +14,29 @@ import AppInputPriceControl from '.~/components/app-input-price-control/index.js
...
+ * @param {function()} props.onRequestClose Callback to close the modal.
+ * @param {function(Object)} props.onSubmit Callback when the form is submitted, with the form value.
+ */
+const AddMinimumOrderFormModal = ( props ) => {
+	const {
+		countryOptions,
+		initialValues,
+		onSubmit = noop,
+		onRequestClose = noop,
+	} = props;
```
tomalec: :nail_care: Just a personal preference, not tp have to think: "Is `props` going to be used somewhere below as well?"
```suggestion
const AddMinimumOrderFormModal = ( {
	countryOptions,
	initialValues,
	onSubmit = noop,
	onRequestClose = noop,
} ) => {
```

However, then we'd need 
```
		'jsdoc/check-param-names': [
			'warn',
			{ disableExtraPropertyReporting: false },
		],
```
to avoid complaints about documented non-destructed `props.initialValues`. (Which I find rather strange, as it didn't complain about just `props`)

js/src/components/shipping-rate-section/minimum-order-card/edit-minimum-order-form-modal.js
```diff
@@ -4,6 +4,7 @@
 import { Button } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { Form } from '@woocommerce/components';
+import { noop } from 'lodash';
```
tomalec: :nail_care:/:question:
I always have this recurring doubt, do we need to import a full blown library (with all dependency costs) for something as simple as `const noop = () => {};`

js/src/components/shipping-rate-section/minimum-order-card/edit-minimum-order-form-modal.js
```diff
@@ -13,12 +14,33 @@ import AppInputPriceControl from '.~/components/app-input-price-control';
...
+ * @param {function()} props.onRequestClose Callback to close the modal.
+ * @param {function(Object)} props.onSubmit Callback when the form is submitted, with the form value.
+ */
+const EditMinimumOrderFormModal = ( props ) => {
+	const {
+		countryOptions,
+		initialValues,
+		onRequestClose = noop,
+		onSubmit = noop,
+	} = props;
```
tomalec: :nail_care: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1410/files#r845530511

tomalec (approved):
tested locally, reviewed the code, LGTM :)

js/src/components/shipping-rate-section/minimum-order-card/add-minimum-order-form-modal.js
ecgan: > not tp have to think: "Is `props` going to be used somewhere below as well?"

Interesting, I never thought about that, I think that sounds good. 🙂 

Am I doing this correctly? I'm still seeing ESLint JSDoc complaining: 🤔 

image

I end up using another way, by using a new JSDoc typedef `MinimumOrderGroup`, in [`11ca55c` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/11ca55c460539722dc72b4abb502f12570d55910).

js/src/components/shipping-rate-section/minimum-order-card/edit-minimum-order-form-modal.js
ecgan: Same fix in [`11ca55c` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/11ca55c460539722dc72b4abb502f12570d55910).

js/src/components/shipping-rate-section/minimum-order-card/edit-minimum-order-form-modal.js
ecgan: I'm feeling like we don't need to worry about it that much because we are loading one shared copy of lodash via DEWP anyway. And I also think using `import { noop } from 'lodash'` is better than having to define `const noop = () => {};` in many files.

ecgan:
@tomalec , thanks for the review. I'm going to leave this PR on-hold for a bit. The idea is that I would like shipping rates modal and minimum order modals to be similar or consistent. As I work on the shipping rates modals, I might make some changes and further tweak these minimum order modals. If there are more changes, I'll request review from you again. 😄 🙏 

One thing that I have been thinking in this PR is about the delete minimum order operation in the Edit Minimum Order Modal, whether we should use `onSubmit`, or have another props `onDelete` because delete operation doesn't really require "form submission". I'm also thinking to have a `handleDelete` handler in the `MinimumOrderCard` component (in addition to the current `handleAddSubmit` and `handleEditChange`) so that we have add, edit and delete operation code in one place (currently the delete operation is being handled inside `handleEditChange`). I'm leaning towards making these changes. The downside to this is that we end up with more code. If you have comments on this, it would be good to hear from you.

I'll look into shipping rates work first and see if I want to proceed with making the changes above to keep shipping rate and minimum order consistent. Keeping this PR open for the mean time.

js/src/components/shipping-rate-section/minimum-order-card/edit-minimum-order-form-modal.js
tomalec: Given the problems, we had with DEWP ;) it would be nice to have fewer reasons to use it ;)

tomalec:
> One thing that I have been thinking in this PR is about the delete minimum order operation in the Edit Minimum Order Modal, whether we should use `onSubmit`, or have another props `onDelete` because delete operation doesn't really require "form submission".

Do you mean that the lines 
https://github.com/woocommerce/google-listings-and-ads/pull/1410/files#diff-0ec19ffcf6fc4e15e362791d741d5f07318625dd8d10fe0c685bab99479d1c61R42-R45

so `EditMinimumOrderFormModal~handleDeleteClick` will cal explicitly `props.onDelete()` instead of `props.onSubmit( {countries: [], currency: undefined, threshold: undefined } )`?

If I understand it correctly, then I really like the idea the more explicit the communication between components => the easier it's to understand when reading in future, the better :)

ecgan:
@tomalec , yes! 🙂 Alright, gotcha. I'll look into it later then. Thank you!

ecgan:
@tomalec and @eason9487 , I am done making changes in this PR. This PR is pretty much similar to https://github.com/woocommerce/google-listings-and-ads/pull/1424. Appreciate your review on this. 🙂 🙏 

One thing I would like to point out is that while I was working on the changes (together with the experience in PR https://github.com/woocommerce/google-listings-and-ads/pull/1424), I had a realization that there are duplicated code between `MinimumOrderInputControl`, `ShippingRateInputControl` and `CountriesTimeInput` (the input control for shipping time). All of them has a label, an edit button, and a price/number input, and their layout / CSS is similar. I am thinking to refactor this into a general reusable component like `EditNumberInputControl` (the "Edit" naming is to signify the Edit button), with an `onEditClick` prop. This component should not have any business logic in it, and I think we can replace/remove the three individual custom *InputControl components above by using this new component. I will come out with a separate issue and PR on this.

eason9487:
> @eason9487 , I am done making changes in this PR. This PR is pretty much similar to #1424. Appreciate your review on this.

Thanks for notifying me! However, after a quick look, my conclusions and doubts are the same as #1424, and plus two additional concerns were seen during the quick view:
1. Suggest avoiding using CSS names that easily name collisions.
   https://github.com/woocommerce/google-listings-and-ads/commit/d2fa14037ef96fa2a23ee2e15d8feb00a98f6fd0#diff-f0f988203d80fa2fafac5aae546b8e8dc9945a737ff379431ee0766d201ab374R43
1. The unused `Button` could be removed.
   https://github.com/woocommerce/google-listings-and-ads/commit/68f5b429189dbac9506765b981f7f4bc2645183e#diff-c991bb60d0003297c4da783fc55e661633ca16c5297173c1b156c8d1beae6019R4

I'm afraid I wouldn't be a suitable reviewer to review the new changes in this PR. Sorry, I took the liberty of presuming that your conclusion in this PR might be the same as #1424. Under this premise and in order to avoid the same discussion and then still enter the "agree to disagree" situation again, I tend not to participate in this round of code review. 🙏

ecgan:
@eason9487 ,

> 1\. Suggest avoiding using CSS names that easily name collisions.

I was thinking of making it same as shipping rate side, and we can remove that in the input control refactor later. I have changed it in [`1e98707` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/1e987076319d0d366e7179f2c7470906ebff053e).

> 2\. The unused `Button` could be removed.

Hmm my editor (VS Code) ESLint is not showing error for me. I think it is because of the typedef Button import (https://github.com/woocommerce/google-listings-and-ads/commit/68f5b429189dbac9506765b981f7f4bc2645183e#diff-c991bb60d0003297c4da783fc55e661633ca16c5297173c1b156c8d1beae6019R19), possibly causing conflict in ESLint. I have removed it in [`ed37e40` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/ed37e40bd6d5bf00ab0ad67db5a9f95c02c45039). 

Thanks for catching the above. 🙂 🙏 

> my conclusions and doubts are the same as #1424, [...]
> I took the liberty of presuming that your conclusion in this PR might be the same as #1424. Under this premise and in order to avoid the same discussion and then still enter the "agree to disagree" situation again, I tend not to participate in this round of code review. 🙏

Yeah I was already having the expectation that your review comments on this PR would be the same as #1424. 😅 I think that is fine, don't worry about it, the more important thing is we are getting things moving. 

Thanks again for the quick review and your reply on this.

eason9487:
Thanks for resolving the code conflicts to the replacements of TreeSelectControl in 67fb06a9bc00169c12201fa9720d41b4ea4517bf. 🙏  There are two `AppCountrySelect` in the corresponding JSDoc could be updated to `SupportedCountrySelect`:
1. https://github.com/woocommerce/google-listings-and-ads/blob/67fb06a9bc00169c12201fa9720d41b4ea4517bf/js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/add-minimum-order-form-modal.js#L26
2. https://github.com/woocommerce/google-listings-and-ads/blob/67fb06a9bc00169c12201fa9720d41b4ea4517bf/js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/edit-minimum-order-form-modal.js#L26

ecgan:
> There are two `AppCountrySelect` in the corresponding JSDoc could be updated to `SupportedCountrySelect`:

Thanks @eason9487 ! Updated in [`3c8edae` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/3c8edaed1176f026a862e1b8677443b70e1ad76a). 😃 👍

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.js
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Get handlers for shipping rate minimum order.
+ *
+ * @param {Object} props
+ * @param {Array<ShippingRate>} props.value Shipping rates.
+ * @param {(newValue: Array<ShippingRate>) => void} props.onChange Callback called with new data when minimum order for shipping rates are changed.
+ * @return {getMinimumOrderHandlersResult} Handlers.
+ */
+const getMinimumOrderHandlers = ( { value, onChange } ) => {
```
tomalec: Similarly to https://github.com/woocommerce/google-listings-and-ads/pull/1424#discussion_r860270678:

It's somewhat alerting to me that `getHandlers` returns one `handler` and two `get*Handlers`. `getHandler` that returns `get*Handler` isn't it `getGetHandler` ;) - that sounds quite redundant

Also, what is the benefit of having one function (`getHandlers`) that returns a set of three disjointed functions? They do not need access to each other or share any logic, do they?

Maybe we can just `export { getAddHandler, getChangeHandler, getDeleteHandler }` directly? Then we will save on constructing & destructing, and avoid the naming problem mentioned above. That will also make them more atomic and easier to test and maintain independently.

tomalec: :nail_care: Plus, I'd make it at least 
```suggestion
const getMinimumOrderHandlers = ( value, onChange ) => {
```

As we call it as `getMinimumOrderHandlers( { value, onChange } );` anyway, so constructing and deconstructing only blurs the picture for me

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.test.js
```diff
@@ -0,0 +1,290 @@
...
+
+		mockOnChange = jest.fn();
+		handlers = getMinimumOrderHandlers( {
+			value,
+			onChange: mockOnChange,
+		} );
+	} );
+
+	describe( 'handleAddSubmit', () => {
+		it( 'returns value with the newly added group', () => {
```
tomalec: :construction: Same as in https://github.com/woocommerce/google-listings-and-ads/pull/1424#discussion_r860274770

```suggestion
		it( 'Calls the `onChange` callback with the value containing the newly added group', () => {
```

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.test.js
```diff
@@ -0,0 +1,290 @@
...
+					options: {
+						free_shipping_threshold: 50,
+					},
+				},
+			] );
+		} );
+	} );
+
+	describe( 'getChangeHandler', () => {
+		it( 'returns value updated based on changed group threshold', () => {
```
tomalec: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1424#discussion_r860277610

:construction: (here and below)
"When called with a value, returns a function, which when called with a new value, calls `onChange` callback with a changed...."

:nail_care: :grey_question:
Just writing the above gives me a smell that maybe this state keeping logic may be a tad overcomplicated

Probably will get simplified and more obvious if the handlers would be separate functions
https://github.com/woocommerce/google-listings-and-ads/pull/1410/files#r860293079

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/add-minimum-order-form-modal.js
```diff
@@ -0,0 +1,66 @@
...
+ * @param {Array<CountryCode>} props.countryOptions Array of country codes options, to be used as options in SupportedCountrySelect.
+ * @param {MinimumOrderGroup} props.initialValues Initial values for the form.
+ * @param {(values: MinimumOrderGroup) => void} props.onSubmit Callback when the form is submitted, with the form value.
+ * @param {() => void} props.onRequestClose Callback to close the modal.
+ */
+const AddMinimumOrderFormModal = ( {
+	countryOptions,
+	initialValues,
+	onSubmit,
+	onRequestClose = noop,
```
tomalec: As in https://github.com/woocommerce/google-listings-and-ads/pull/1424#discussion_r860287002 I gout cough with https://github.com/woocommerce/google-listings-and-ads/issues/1439

:nail_care: YAGNI, we don't need a default.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/edit-minimum-order-form-modal.js
```diff
@@ -0,0 +1,84 @@
...
+ * @param {(values: MinimumOrderGroup) => void} props.onSubmit Callback when the form is submitted, with the form value.
+ * @param {() => void} props.onDelete Callback when delete button is clicked.
+ * @param {() => void} props.onRequestClose Callback to close the modal.
+ */
+const EditMinimumOrderFormModal = ( {
+	countryOptions,
+	initialValues,
+	onSubmit,
+	onRequestClose = noop,
+	onDelete = noop,
```
tomalec: :nail_care: YAGNI

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/minimum-order-form-modal.js
```diff
@@ -11,57 +10,48 @@ import { noop } from 'lodash';
...
+ * @param {MinimumOrderGroup} props.initialValues Initial values for the form.
+ * @param {(values: MinimumOrderGroup) => void} props.onSubmit Callback when the form is submitted, with the form value.
+ * @param {() => void} props.onRequestClose Callback to close the modal.
+ */
+const MinimumOrderFormModal = ( {
+	countryOptions,
+	renderButtons = noop,
+	initialValues,
+	onSubmit,
+	onRequestClose = noop,
```
tomalec: :nail_care: YAGNI
```suggestion
	renderButtons,
	initialValues,
	onSubmit,
	onRequestClose,
```

tomalec (changes requested):
I left some comments. The :construction: ones are in test descriptions. Even though test descriptions are not a showstoppers, but IMHO they visualize a bit overly complicated structure of `get*Handlers` which if changed could have an impact, on actually crucial flow.

I didn't test the app behavior yet, will do that tomorrow.

Also, I share a number of concerns mentioned by @eason9487 in https://github.com/woocommerce/google-listings-and-ads/pull/1424 but overall, I think this PR may not solve the problems mentioned there, but puts us one step forward making the code better than before. 

Hopefully, we could address the concerns of both of you eventually after further improvements.

tomalec:
Another, kind of nonspecific confusion I got here is that the flow of the `onChange` callbacks is hard to follow for me. I'll look deeper tomorrow if there is anything I would improve or at least point to.

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.js
ecgan: > It's somewhat alerting to me that `getHandlers` returns one `handler` and two `get*Handlers`. `getHandler` that returns `get*Handler` isn't it `getGetHandler` ;) - that sounds quite redundant

Ahh right, I'll think of a better naming. If you have a suggestion, do share. 😄 

> Also, what is the benefit of having one function (`getHandlers`) that returns a set of three disjointed functions? They do not need access to each other or share any logic, do they?

They don't need access to each other. I was thinking that the three functions are related ("add", "edit" and "delete" operations; part of CRUD operations), and because they all take in `value` and `onChange`, I thought it make sense to group them together. The benefit is that on the caller side, they just need to have one call to `getMinimumOrderHandlers` to get all three of them (instead of three separate calls).

To be more specific, I mean: 

```js
// current way - one call to get all three:
const { getA, getB, getC } = callGetABC( { value, onChange } );

// alternative way 1:
const getA = callGetA( { value, onChange } );
const getB = callGetB( { value, onChange } );
const getC = callGetB( { value, onChange } );

// alternative way 2:
const [ getA, getB, getC ] = [
	callGetA( { value, onChange } ),
	callGetB( { value, onChange } ),
	callGetB( { value, onChange } ),
];
```

> Maybe we can just `export { getAddHandler, getChangeHandler, getDeleteHandler }` directly? Then we will save on constructing & destructing, and avoid the naming problem mentioned above. That will also make them more atomic and easier to test and maintain independently.

That is a good idea! I'll look into it. 😄 👍 

> 💅 Plus, I'd make it at least [...]

There is no change in your suggested change...? 😅 I'm guessing you mean changing it to `getMinimumOrderHandlers ( value, onChange )` (accept two parameters instead of one object parameter)? I was thinking because they were both coming from the react props, it makes sense to keep them in object props structure. WDYT? Open to changes here.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-input-control.js
```diff
@@ -41,23 +44,7 @@ const MinimumOrderInputControl = ( props ) => {
...
-						}
-						modal={
-							<EditMinimumOrderModal
-								countryOptions={ countryOptions }
-								value={ value }
-								onChange={ handleEditChange }
-							/>
-						}
-					/>
+					{ labelButton }
```
tomalec: Personally, I'd keep the label button implementation here. It will simplify the use in js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js and make `MinimumOrderInputControl` self-sufficient component with `onDelete` and `onChange`.

tomalec: See https://github.com/woocommerce/google-listings-and-ads/compare/feature/refactor-minimum-order-modal...feature/refactor-minimum-order-modal-labelButton

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/add-minimum-order-form-modal.js
ecgan: Let's have a quick discussion on this. It is a practice that I learned from past experience, and if it is not a good practice, I should really change it from now. 😄 

The reason I put in `noop` or `() => {}` as default value for this functions is to prevent runtime error and prevent blowing up the whole page. When someone missed to specify the function and it breaks, it might just break one small part of the UI, and users may still be able to continue with their main flow without being impacted.

This is the closest article I found on the internet related to this topic - https://www.digitalocean.com/community/tutorials/how-to-customize-react-components-with-props:

> It’s good practice to always add a default value to prevent runtime errors if a prop is not required.

Curious to hear your thoughts on this. 🙏

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-input-control.js
ecgan: I thought about a better implementation, by having a generic reusable `EditNumberInputControl` (Number InputControl with Edit button, with `onEditClick` callback) that can be used in shipping rate, shipping time and minimum order, so we don't need to have `MinimumOrderInputControl` etc. See issue https://github.com/woocommerce/google-listings-and-ads/issues/1444. 🙂

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.js
tomalec: > There is no change in your suggested change...? sweat_smile I'm guessing you mean changing it to `getMinimumOrderHandlers ( value, onChange )` (accept two parameters instead of one object parameter)? 

Yes, I fixed the comment above.

>  I was thinking because they were both coming from the react props, it makes sense to keep them in object props structure. WDYT? Open to changes here.

I think in most of the cases data come from React props, and I still don't see an actual benefit in https://github.com/woocommerce/google-listings-and-ads/pull/1410/files#diff-19741a9dbe0895f99f4b04a6a88e21d3a8445917bea2f62ccf3f2610acff92d8R36 of packing them to another props-like object, different form the original props. To me it introduces even more confusion: another data type to process in my brain, and think "is that format used expected by some other code as well?" "How those default values affect the flow, what arethe cases when that are different from the original props, so we don't do `getMinimumOrderHandlers( props )`?"

tomalec:
Tested locally, and from functional point of view it works good.

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.js
tomalec: I agree that 
```js
const { getA, getB, getC } = callGetABC( value, onChange );
// is shorter than 
const getA = callGetA( value, onChange );
const getB = callGetB( value, onChange );
const getC = callGetB( value, onChange );
```
but, personally, I don't think it's worth the complication of having:
>  a function, that given some args returns a function, that given another args returns a function, that is a handler, that given yet another args translates them and calls a callback given to the very first function.

https://github.com/woocommerce/google-listings-and-ads/compare/feature/refactor-minimum-order-modal-labelButton...feature/refactor-minimum-order-modal-handlers
(:warning: it's base on branch from https://github.com/woocommerce/google-listings-and-ads/pull/1410/files/3c8edaed1176f026a862e1b8677443b70e1ad76a#r860841409)

Here is another proposal, to use just good, old, native `bind` to avoid the problem of a `getGetGetChangeHandler` and describing it.

--- 

Alternatively, we can just move the handlers' definitions back to `MinimumOrderCard` and we will make all those problems not applicable.
Then we could test the handlers' behavior via the component (which would test more of a component's behavior than implementation detail).

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-input-control.js
tomalec: I see.

But, I wouldn't squeeze #1444 here, and rather solve it by a separate PR. As I may argue, that what we have here is the best solution for 1444. For example, at least the `Edit` button could be a part of such `*InputControl`, and only the modal would be provided by a consumer.

After my commit, the component is closer to what it was before, leaving the clean field to address #1444 separately.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/add-minimum-order-form-modal.js
tomalec: Yep, blowing up the whole page is definitely not something we want. And I also hate when something crashes in our plugin, and we end up with a blank page of doom.

But to me, this seems to be a systematic/platform problem (that's another reason I like native Web Platform - if your custom element throws, everything around works intact).

In the article, you linked there is: "**If a prop isn’t required**" and I think in our app those callbacks are required.

I see this as a hard trade:
In case of a bug having an explicit, easy to debug error, with a risk to blow UX vs having just impaired but hopefully semi-functional UI but silent, hard to track bug.


Also, we don't set defaults for other props, why? We do not protect our component against not providing or providing wrong values for other args. Like, what if `countryOptions` is malformed/missing?

----

Speaking of the practice itself, I think it's a good one if applied to reusable published components. When we do not know/control how they will be used. But when this "someone" who may miss is also us - we are the authors and consumers - we control everything. Then preparing for scenarios that we know, are not there is a wasted effort.

To protect against failures, personally, I'd add more tests and make sure the platform we use can isolate crashes gracefully as Web does.

So I'm fine with keeping it here for sake of future usage. I just thought :nail_care: it may never happen that we would really have such scenario, and it's not worth anticipating everything.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/minimum-order-form-modal.js
tomalec: In regards of https://github.com/woocommerce/google-listings-and-ads/pull/1410/files/3c8edaed1176f026a862e1b8677443b70e1ad76a#r860844885 here we do not use `onRequestClose` ourselves, but only forward it, so we may also let the actual consumer of that prop (`AppModal`) decide on defaults.

ecgan: Ahh yes, agree, generally I don't use `noop` when we just forward the function and do not call the function, think I missed this when I was changing things. Removed in [`a908f74` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/a908f740845ec32fe075dc8b4bcd99992206e295).

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.js
ecgan: > I think in most of the cases data come from React props, and I still don't see an actual benefit in [#1410 (files)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/files#diff-19741a9dbe0895f99f4b04a6a88e21d3a8445917bea2f62ccf3f2610acff92d8R36) of packing them to another props-like object, different form the original props. [...]

Thanks, I think that sounds fair. I have made the changes accordingly in this commit: [`4c71bd9` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/4c71bd931906fd376367f98c3b003b3c79627516).

I thought of a way to get around the hard problem of naming things 😂 , e.g. by using `onSubmit={ getMinimumOrderChangeHandler( value, onChange )( group ) }`. There is no "intermediary" function naming.

From the above, I also thought about changing the function signature to `getMinimumOrderChangeHandler( value, onChange, oldGroup, newGroup )`, similar to your code example. However, I don't quite like the way it looks. Maybe it's because I'm already used to the nested curried functions (like `() => () => {}`) where each "level" represent different things. With `getMinimumOrderChangeHandler( value, onChange )( oldGroup )( newGroup )`, the first level `value` and `onChange` relates to shipping rates that we received from props, the second level represents the rendered "oldGroup", and the third level represents the callback generated by the user action.

Let me know if you think it is okay. 🙏 

> Here is another proposal, to use just good, old, native `bind` to avoid the problem of a `getGetGetChangeHandler` and describing it.

💅 Personal preference: every time I see `.bind(null,...)`, I just feel like it is an eye sore and I don't like it. 😂  

---

> Alternatively, we can just move the handlers' definitions back to `MinimumOrderCard` and we will make all those problems not applicable.

I think that is a step backward. I like breaking things smaller (like what you have suggested for the handlers), and making the handlers outside the component allows us to test the handlers individually, which is good. 🙂 

> Then we could test the handlers' behavior via the component (which would test more of a component's behavior than implementation detail).

We can have handler testing and also component testing. I think they are / can be complementary to each other, and not necessarily one replacing another.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-input-control.js
ecgan: > But, I wouldn't squeeze #1444 here, and rather solve it by a separate PR.

Yes, I plan to solve it in a separate PR, not in this one. 🙂 

> After my commit, the component is closer to what it was before, leaving the clean field to address https://github.com/woocommerce/google-listings-and-ads/issues/1444 separately.

Yeah, I understand your point. However, I'm also thinking that it is good for this PR now to be similar to the shipping rate side of things, so when I work on #1444 later, because shipping rate and minimum order are similar, it is also somewhat easy to make the code change, and easy for reviewer. 

I think it is fine for us to leave this concern about label button here, because the concern will go away when we address #1444 (hopefully soon 🙂).

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-form-modals/add-minimum-order-form-modal.js
ecgan: > In the article, you linked there is: "If a prop isn’t required" and I think in our app those callbacks are required.

Yeah. If there is a linting that will throw error saying we didn't provide a value for required props, that will be really good. (I used to have that in my past experience, I think that was a React + TypeScript project.)

> Speaking of the practice itself, I think it's a good one if applied to _reusable published components_. _When we do not know/control how they will be used_. But when this "someone" who may miss is also us - we are the authors and consumers - we control everything. Then preparing for scenarios that we know, are not there is a wasted effort.

Ahh, I think this is a very good point, and I think now I know why or how I got that practice from. 🙂 Thanks. I'll proceed to remove the `noop`. 

> To protect against failures, personally, I'd add more tests and make sure the platform we use can isolate crashes gracefully as Web does.

I was just thinking to say adding the defensive `noop` code can be complementary to adding more tests, both can exist together.

ecgan: I have removed the `noop` usage in [`e452bed` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/e452bedf64da09c9f73a9a77de7ae3e7077f5188). 🙂

ecgan:
@tomalec , thanks, I have made changes and addressed your comments, including the test descriptions. This is now ready for another round of review. 🙂 🙏

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
```diff
@@ -14,85 +18,79 @@ import AppButtonModalTrigger from '.~/components/app-button-modal-trigger';
...
+						countryOptions={ countryOptions }
+						initialValues={ group }
+						onSubmit={ getMinimumOrderChangeHandler(
+							value,
+							onChange
+						)( group ) }
+						onDelete={ getMinimumOrderDeleteHandler(
+							value,
+							onChange
+						)( group ) }
```
tomalec: At this point, if we are going to provide all those arguments here, why not simply?:
```suggestion
						onSubmit={ getMinimumOrderChangeHandler(
							value,
							onChange,
							group ) }
						onDelete={ getMinimumOrderDeleteHandler(
							value,
							onChange,
							group ) }
```

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.js
tomalec: >  I also thought about changing the function signature to `getMinimumOrderChangeHandler( value, onChange, oldGroup, newGroup )`, similar to your code example. However, I don't quite like the way it looks. Maybe it's because I'm already used to the nested curried functions (like `() => () => {}`) where each "level" represent different things. 

Wow, really? :) Personally, I don't like the way `getMinimumOrderChangeHandler( value, onChange )( oldGroup )( newGroup )` looks because I need to head scratch and read the docs to know:
- "What is the function that is the result of `getMinimumOrderChangeHandler( value, onChange )`?"
-  "What is that function meant for?"
-  "What is its signature?"
-  "What are its arguments?"
- "Is it meant to be used alone?"
- and the same story for `getMinimumOrderChangeHandler( value, onChange )( oldGroup )`

VSCode hint looks pretty complicated, and requires a handwritten description to make more sense:
image

While with `handleChange( value, onChange, oldGroup, newGroup )`:
- The IDE hint is straightforward
- There are no functions to name or ask for
- There is less code to implement those
- There are less function calls to be executed in run-time
- The stack trace is shorter and easier to reason, as there is one named function instead of a bunch of anonymous ones.

image

> With `getMinimumOrderChangeHandler( value, onChange )( oldGroup )( newGroup )`, the first level `value` and `onChange` relates to shipping rates that we received from props, the second level represents the rendered "oldGroup", and the third level represents the callback generated by the user action.

We have plenty of cases when the function argument values come from many sources, and I think it's not the reason to create more functions, just because someone who calls a function, may fetch arguments from different contexts.

-----

We never call `getMinimumOrderChangeHandler( value, onChange )` without the third argument, so I still don't see a reason to create another function and separate it there

-----

> Personal preference: every time I see `.bind(null,...)`, I just feel like it is an eye sore and I don't like it.

But this is a native function, part of the language for ages. Created specifically for that, so we don't have to "polyfill" it every time manually.

```js
onChange={ handleChange.bind( null, value, onChange, groups[ 0 ] ) }
onChange={ ( newGroup ) => { handleChange( value, onChange, groups[ 0 ], newGroup ) }
onChange={ getMinimumOrderChangeHandler(
							value,
							onChange
						)( group[ 0 ] ) }
```

For me, the first is the easiest to read for me (plus the shortest and cheapest), and the latter is the most complicated to read, write, test, and execute.

And My personal preference: 
My eyes are sore of reading `() => () => () => {}` or  `foo( bar )( baz )( biz )` as it feels like [Brainfuck](https://en.wikipedia.org/wiki/Brainfuck), [Brackets](http://p-nand-q.com/programming/languages/brackets.html), or some other esoteric language :joy: . While `foo.bind( context, bar, baz )` explicitly "Binds to foo the context and bar baz arguments"

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-input-control.js
tomalec: > I plan to solve it in a separate PR, not in this one. 

But the change is here. So you are partially including the #1444 fix here.

And in my opinion, this change does not add value and does not simplify the minimum order modal, but the opposite - it makes the logic more dislocated and adds more unnecessary code.

> However, I'm also thinking that it is good for this PR now to be similar to the shipping rate side of things, so when I work on #1444 later, because shipping rate and minimum order are similar, it is also somewhat easy to make the code change, and easy for reviewer.
> 
> I think it is fine for us to leave this concern about label button here, because the concern will go away when we address #1444 (hopefully soon ).


Speaking of a future review of #1444, to be the situation look like, we are here introducing more bloat and more concenrns in the name of "refactoring" to be refactored in the future. It does not sound like a courtesy for the future reviewer, but adding more work there and maintaining questionable implementation in the meantime.

As a #1444 reviewer, I'd rather see the older (pre-shipping rate https://github.com/woocommerce/google-listings-and-ads/pull/1424) implementation in the diff, to be able to evaluate that the #1444 fix delivers something better than what we had. And not just something better than post-#1424, but potentially worse than pre-#1424.

For me "leaving the concern of a button" would actually mean not changing the implementation of `labelButton` in this PR. Then approaching #1444 fixing without a bias of already made solution

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
ecgan: Thanks @tomalec , I've read your reasonings in https://github.com/woocommerce/google-listings-and-ads/pull/1410#discussion_r861957968, and I think you have some good and valid points there 🙂 👍  I'll make the changes like what you suggested here (i.e. `onSubmit={ getMinimumOrderChangeHandler( value, onChange, group ) }`).

ecgan: I have made the changes in [`5d03559` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/5d0355925e568082f0d32a3d522429527b463f22) and [`c6f4f8f` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/c6f4f8fd5bc63e507cb758fa2860eafa293eb7a0).

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-input-control.js
ecgan: > So you are partially including the #1444 fix here.

Well, yes, it is kind of a partial fix.

> this change does not add value and does not simplify the minimum order modal, but the opposite - _it makes the logic more dislocated and adds more unnecessary code_.
> [...]
> Speaking of a future review of #1444, to be the situation look like, we are here _introducing more bloat and more concenrns_ in the name of "refactoring" to be refactored in the future. It does not sound like a courtesy for the future reviewer, but adding more work there and maintaining questionable implementation in the meantime.

I don't quite agree with the _above emphasized statement_ (maybe I don't see the way you see it?). I think it is a partial fix towards a better approach, in the following ways: 

- We move the modal out of the `MinimumOrderInputControl`, and put the modal in the parent `MinimumOrderCard`.
- Because of the above point, `MinimumOrderInputControl` does not need `countryOptions` and `onDelete` props.

Even if we don't do the above now in this PR, when we work on issue #1444, we shall do the above anyway, replacing `MinimumOrderInputControl` with `EditNumberInputControl` with `onEditClick` prop. So, we are doing "part of #1444" early in this PR. I don't see how the above "introduce more bloat and concerns". 😅 

> As a #1444 reviewer, I'd rather see the older (pre-shipping rate #1424) implementation in the diff, _to be able to evaluate that the #1444 fix delivers something better than what we had. And not just something better than post-#1424, but potentially worse than pre-#1424._
> 
> For me "leaving the concern of a button" would actually mean not changing the implementation of `labelButton` in this PR. Then _approaching #1444 fixing without a bias of already made solution_

Thanks, this is an interesting perspective, and I think this is fair 🙂 👍  I have made the changes in [`193cb19` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/193cb19768843bfbfc13b8bba2dca856a8e00c86). 🙏

js/src/components/shipping-rate-section/minimum-order-card/getMinimumOrderHandlers.js
ecgan: Thanka @tomalec , for your detailed explanations and bullet points! You shared some good perspectives there. I gave more thoughts about it and decided to change them and simplify them to `onChange={ getMinimumOrderChangeHandler( value, onChange, group ) }`, as commented in https://github.com/woocommerce/google-listings-and-ads/pull/1410#discussion_r862767458. 🙂 👍

ecgan:
@tomalec , thank you, I have made changes based on your feedback. This is ready for another round. 🙂 🙏

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
```diff
@@ -14,75 +18,27 @@ import AppButtonModalTrigger from '.~/components/app-button-modal-trigger';
...
-			if ( newGroup.countries.includes( newShippingRate.country ) ) {
-				newShippingRate.options.free_shipping_threshold =
-					newGroup.threshold;
-			}
-
-			return newShippingRate;
-		} );
-
-		onChange( newValue );
-	};
```
tomalec, line 83: What is the actual benefit of extracting those two handlers outside?

Personally, I don't see this as a degradation. We turned two simple functions, 60lines of fairly readable code, into 4 separate files  (7 incl. tests) ~ 40lines each, that exports function, to generate a function, to generate a function, that's complicated to describe in test, name, and use.

If this is only, to add tests for those handlers, then I'd say we should test it as the `MinimumOrderCard` behavior (https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/4ebd22391627539c9c7182008417ff18bd4a1056). Otherwise, we test the very implementation detail, then affect the code itself to do such tests.

I'll try to write an alternative proposal.

tomalec (changes requested):

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
tomalec: Also, those function does not have much "event handling" functionality. All they do is translate API to API, then forward the result to an actual handler. So they are rather API translating proxy/glue-code, than "handlers". 

If we treat them as such, it would be easier to use, name, and test them.

See https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/f682f9e2fdb7b20c88804535fe4d80f5891a9f47 - It makes overall code about smaller, I think it's easier to write and read tests for those functions, and to me 

```jsx
onDelete={ () =>
    onChange( deleteGroup( value, group ) )
}
```
Is more explicit and informative than
```jsx
onDelete={ getMinimumOrderDeleteHandler(
	value,
	onChange,
	group
) }
```

tomalec: Also, this translation from `MinimumOrderGroup` to `ShippingRate` and back, seems like a wasted effort to me. `MinimumOrderGroup` type is created and used on the frontend, so we could translate back and work, writing additional functions like the ones above when we could work on the original data structure in all components.

Therefore, I'd rather avoid increasing the about of found around that translation, separate just the translation to the side, then eventually remove the need for it.

tomalec: While doing https://github.com/woocommerce/google-listings-and-ads/commit/f682f9e2fdb7b20c88804535fe4d80f5891a9f47, I realized that add and delete handlers are actually the change handlers - respectively without `oldGroup`, and without `newGroup`. So I'd roll them to one function

tomalec: See https://github.com/woocommerce/google-listings-and-ads/pull/1499 for full proposal, feel free to merge here any subset of its commits.

ecgan: @tomalec , thanks for the review and for your PR #1499, the PR looks good and I like it! 😃 👍 I left my review on your PR in https://github.com/woocommerce/google-listings-and-ads/pull/1499#pullrequestreview-974077665, and I think we can have both PRs merged into the `develop` branch, let me know how you like to proceed. 🙂

tomalec (approved):
Reviewed the code, and tested locally.
Given we are going to merge https://github.com/woocommerce/google-listings-and-ads/pull/1499 here or shortly after, LGTM

PR #1412, opened 2022-04-07
Adding Last Line Rule on ESLINT 

puntope:

This PR adds a new rule in our ESLINT config in order to show the Linter error when no new line is added at the end of the file.

Context:

Besides the fact we have the rule in the `.editorconfig`  we're still encountering some files with this issue in several PRs which relies on the developer eyes to identify the lack of the new line.

This PR solves that by triggering the ESLINT error. 

1. Checkout this PR
2. Go to any JS file and remove the last empty line *
3. Run `npm run lint:js`
4. Error should be shown
5. Add again the empty line
6. Error should be fixed

*In case you editor adds automatically the new line (like my case) you can go to `.editorconfig` and set `insert_final_newline = true` as `false`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,5 +17,6 @@ module.exports = {
 				additionalHooks: 'useSelect',
 			},
 		],
+		'eol-last': 2,
 	},
 };


tomalec:
I'm not sure if that's needed. Even on `develop` I git an error from ``prettier rule set:
```
$ npm run lint:js -- js/src/index.js 


/home/tomalec/repos/google-listings-and-ads/js/src/index.js
  131:3  error  Insert `⏎`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

I think the reason why `.jsdocrc.json` was not caught in another PR is that its filename starts with `.` so it's ignored:

```
$ npm run lint:js -- .jsdocrc.json 

> google-listings-and-ads@1.12.4 lint:js /home/tomalec/repos/google-listings-and-ads
> wp-scripts lint-js ".jsdocrc.json"


/home/tomalec/repos/google-listings-and-ads/.jsdocrc.json
  0:0  warning  File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override

✖ 1 problem (0 errors, 1 warning)
```

puntope:
> I'm not sure if that's needed.

Good catch then :) Closing this since it is not necessary

PR #1413, opened 2022-04-08
Integrate Proxied Request Review Status 

puntope:

Related to GLA Request Review Project 

After finishing the API integration in WCS (already in vipgo) the integration of the middleware as well as the logic fr merging the responses was needed.

This PR adds the next to the feature:

- Proxy request in the Middleware
- Adapt RequestReviewController route method
- Add a new Helper RequestReviewStatuses with the logic to merge the different regionStatues of the 2 programs (free and ads) 
- Updated tests

For now. The logic for merging the response works like this:
We check if any region status is elegible for review, if so, we set `reviewAllowed` as true.
If any region status has Disapproved or Warning state, we add the issues in the `issues` field  
For the account `status` it's still some discussion but for now is build with a priority system, having the worst status as the max priority.

APPROVED
ONBOARDING
PENDING_REVIEW
UNDER_REVIEW
WARING
DISAPPROVED

So for example, if some region is DISAPPROVED the status of the response will be DISAPPROVED doesn't matter if others are in UNDER_REVIEW because DISAPPROVED as more priority. 

1. Checkout this branch
2. Go to GLA Product Feed 
3. Check the network panel and you should see a request to `/wc/gla/mc/review`
4. In the response you should see the fields  `issues`, `status` and `reviewAllowed`. The values will depend actually on your GLA account.


src/API/Google/Middleware.php
```diff
@@ -492,7 +492,7 @@ protected function new_account_name(): string {
 	}
 
 	/**
-	 * Generate a default account name based on the date.
+	 * Generate a default account name based on the date./var/www/html/wp-content/plugins/google-listings-and-ads/src/Options/OptionsInterface.php
```
rawdreeg: `/var/www/html/wp-content/plugins/google-listings-and-ads/src/Options/OptionsInterface.php`

I think this was included by mistake. Could you remove it, please?

src/Google/RequestReviewStatuses.php
```diff
@@ -0,0 +1,79 @@
...
+					}
+
+					$issues = array_merge( $issues, $region_status['reviewIssues'] ?? [] );
+				}
+
+				if ( $status === $this::DISAPPROVED ) {
+					continue;
+				}
+
+				if ( $region_status['eligibilityStatus'] === $this::UNDER_REVIEW ) {
```
rawdreeg: In cases where the $status variable is checked after an assignment wouldn't be better to immediately `continue` after the assignment?

```
if ( $region_status['eligibilityStatus'] === $this::UNDER_REVIEW ) {
	$status = $this::UNDER_REVIEW;
}

if ( $status === $this::UNDER_REVIEW ) {
	continue;
}
```
could be:

```
if ( $region_status['eligibilityStatus'] === $this::UNDER_REVIEW ) {
	$status = $this::UNDER_REVIEW;
	continue;
}
```

rawdreeg (commented):
Thanks, @puntope.

I have tested the changes, and they work as expected. And the test looks good too.

I added a couple of comments for your consideration. Let me know if you have any questions.

R

src/Google/RequestReviewStatuses.php
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	public function get_statuses_from_response( array $response ) {
+		$issues         = [];
+		$review_allowed = false;
+		$status         = 'APPROVED';
+
+		foreach ( $response as $program_type ) {
+			foreach ( $program_type['data']['regionStatuses'] as $region_status ) {
+
+				if ( $region_status['reviewEligibilityStatus'] === $this::ELIGIBLE ) {
```
jorgemd24: Just a quick comment, shouldn't the `constant` - `$this::ELIGIBLE` be invoked with `self` instead of `$this`? The main reason is that usually `$this` refers to the current object and in this case, we are referring to the current class.

src/API/Google/Middleware.php
puntope: Crazy, thanks for catching it.

puntope: Fixed here: https://github.com/woocommerce/google-listings-and-ads/pull/1413/commits/5ac7f63d1cd4c4682fccfbbbc4bbc081db509f33

src/Google/RequestReviewStatuses.php
puntope: Not sure about that @rawdreeg 

**Right now:**

First iteration -> region US:  eligibilityStatus =  UNDER_REVIEW

```
if ( $region_status['eligibilityStatus'] === $this::UNDER_REVIEW ) {
	$status = $this::UNDER_REVIEW; // -> Reached --> $status = $this::UNDER_REVIEW
}

if ( $status === $this::UNDER_REVIEW ) {
	continue; // --> Reached 
}
```

Second iteration -> region MX:  eligibilityStatus =  PENDING_REVIEW

```
if ( $region_status['eligibilityStatus'] === $this::UNDER_REVIEW ) {
	$status = $this::UNDER_REVIEW; // -> NOT  Reached
}

if ( $status === $this::UNDER_REVIEW ) {
	continue; // --> Reached 
}

if ( $region_status['eligibilityStatus'] === $this::PENDING_REVIEW ) {
	$status = $this::PENDING_REVIEW; // NOT REACHED
}

```
Hence the result is correct because UNDER_REVIEW as priority over PENDING_REVIEW

**Proposed change:**

First iteration -> region US:  eligibilityStatus =  UNDER_REVIEW
```
if ( $region_status['eligibilityStatus'] === $this::UNDER_REVIEW ) {
	$status = $this::UNDER_REVIEW; // -> Reached --> $status = $this::UNDER_REVIEW
	continue; // --> Reached 
}

```

Second iteration -> region MX:  eligibilityStatus =  PENDING_REVIEW

```
if ( $region_status['eligibilityStatus'] === $this::UNDER_REVIEW ) {
	$status = $this::UNDER_REVIEW; // -> NOT  Reached
	continue; // --> NOT Reached 
}

if ( $region_status['eligibilityStatus'] === $this::PENDING_REVIEW ) {
	$status = $this::PENDING_REVIEW; // REACHED -> STATUS -> PENDING_REVIEW
}
```
Hence the result is not correct because UNDER_REVIEW as priority over PENDING_REVIEW and we evaluated status to PENDING_REVIEW

src/Google/RequestReviewStatuses.php
puntope: Thanks for the Catch @jorgemd24 

You're right. PHP self refers to the class members, but not for any particular object. 

Fixed here: https://github.com/woocommerce/google-listings-and-ads/pull/1413/commits/4310f5e545790ba43ed19d74b18d4e77edd8c7b6

puntope:
@jorgemd24 @rawdreeg  

This is ready for a new round 👌

src/Google/RequestReviewStatuses.php
rawdreeg: Thanks for clarifying.  I initially missed that  `$status ` needed to be checked in subsequent iterations.

rawdreeg (approved):
Thanks for looking at the feedback. This works as described!

PR #1414, opened 2022-04-08
Redirect notification to reconnect page

mikkamp:

This PR changes the redirect link from the settings page directly to the reconnect-wpcom-account subpath.

In addition to this the Jetpack account controller was changed to test both if the manager is active as well as having a valid token. If this is not done then a scenario where Jetpack is connected but the token is invalid will still return connected as true, which prevents us from viewing the reconnect page.

1. Disconnect Jetpack using the Connection Test page
2. Send a request to check the Google account status: `GET https://domain.test/wp-json/wc/gla/google/connected`
3. Check WooCommerce > Home > Inbox and confirm we see a notification to reconnect our WordPress.com account
4. Check the link and confirm it redirects to `admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings&subpath=%2Freconnnect-wpcom-account`
5. Click on the link and see the reconnect page
image

#### Test connected with invalid token

1. Modify the DB option `jetpack_private_options` and change `blog_token` to `fail_token`. Send a request to check the status:
`GET https://domain.test/wp-json/wc/gla/jetpack/connected`
2. Confirm that `active` returns `no`
3. Go to the settings page and confirm we now redirect to the reconnect page.

### Changelog entry
* Tweak - Redirect notification to WP.com reconnect page.

### DIFF
--- a/src/API/Site/Controllers/Jetpack/AccountController.php
+++ b/src/API/Site/Controllers/Jetpack/AccountController.php
@@ -198,8 +198,7 @@ class AccountController extends BaseOptionsController {
 			return false;
 		}
 
-		$token = $this->manager->get_tokens()->get_access_token( false, false, false );
-		return $token && ! is_wp_error( $token );
+		return false !== $this->manager->get_tokens()->get_access_token();
 	}
 
 	/**
--- a/src/API/Site/Controllers/Jetpack/AccountController.php
+++ b/src/API/Site/Controllers/Jetpack/AccountController.php
@@ -189,11 +189,17 @@ class AccountController extends BaseOptionsController {
 
 	/**
 	 * Determine whether Jetpack is connected.
+	 * Check if manager is active and we have a valid token.
 	 *
 	 * @return bool
 	 */
 	protected function is_jetpack_connected(): bool {
-		return $this->manager->is_active();
+		if ( ! $this->manager->is_active() ) {
+			return false;
+		}
+
+		$token = $this->manager->get_tokens()->get_access_token( false, false, false );
+		return $token && ! is_wp_error( $token );
 	}
 
 	/**
--- a/src/Notes/ReconnectWordPress.php
+++ b/src/Notes/ReconnectWordPress.php
@@ -65,7 +65,7 @@ class ReconnectWordPress extends AbstractNote {
 		$note->add_action(
 			'reconnect-wordpress',
 			__( 'Go to Google Listings & Ads', 'google-listings-and-ads' ),
-			$this->get_settings_url()
+			add_query_arg( 'subpath', '/reconnect-wpcom-account', $this->get_settings_url() )
 		);
 
 		return $note;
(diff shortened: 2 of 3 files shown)


src/API/Site/Controllers/Jetpack/AccountController.php
```diff
@@ -189,11 +189,17 @@ protected function get_connected_callback(): callable {
...
 	 * @return bool
 	 */
 	protected function is_jetpack_connected(): bool {
-		return $this->manager->is_active();
+		if ( ! $this->manager->is_active() ) {
+			return false;
+		}
+
+		$token = $this->manager->get_tokens()->get_access_token( false, false, false );
+		return $token && ! is_wp_error( $token );
```
puntope: ❓ / 💅  

Why not just `return $this->manager->get_tokens()->get_access_token() !== false`

puntope (approved):
For me looks good ✅ 🆗 👍  thanks for implementing it. 

What i tested?

- Disconnected Jetpack and checked the API response ✅ 

I called `/wp-json/wc/gla/google/connected` and got

```
{
    "message": "Please reconnect your Jetpack account.",
    "status": 401,
    "code": "JETPACK_DISCONNECTED"
}
```
I also called `wp-json/wc/gla/jetpack/connected` and got

```
{
    "active": "no",
    "owner": "no",
    "displayName": "",
    "email": ""
}

```

The Inbox notification was there with the right link. ✅ 

Then I reconnected Jetpack again and after that I changed the `blog_token` in option `jetpack_private_options` to `fail_token` 

Got the expected response in `/wp-json/wc/gla/jetpack/connected`  ✅ 

```
{
    "active": "no",
    "owner": "no",
    [...]
}
```
Then I reconnected again. And worked. The 'fail_token'  option was still 'fail_token' and the `/wp-json/wc/gla/jetpack/connected` response was active=yes. ✅ 

I changed it back to `blog-token` and the response was active=no again .✅ 

❓ 

- One thing I noticed was that when the token is not right we don't show the inbox message 
- Also Another question left in the code regarding `is_jetpack_connected` function

src/API/Site/Controllers/Jetpack/AccountController.php
mikkamp: Nice spot, I didn't look at the function close enough to realize that with the default parameters it will suppress the errors.

mikkamp:
Thanks for the review.

> * One thing I noticed was that when the token is not right we don't show the inbox message

We only detect that the token is not valid when we use it in a request to the WCS. Looking at your steps it seemed the request you sent after invalidating the token was only to `/wp-json/wc/gla/jetpack/connected`. If you send one to `/wp-json/wc/gla/google/connected` which gets details from the WCS about the connection, it should recreate the notice (if it was properly removed after reconnecting).

> * Also Another question left in the code regarding is_jetpack_connected function

Thanks for spotting that, I've changed it to your suggestion and modified the tests.

PR #1418, opened 2022-04-11
Re-connecting WPCOM Account - Disable menu links during the re-connecting process

eason9487:

It's a part of frontend in #775

- Make the GLA menu links of WC Navigation appear disabled.
- When entering the account reconnecting process, lock the available link to the corresponding reconnecting page.
- Don't display the classic navigation tabs during the reconnecting process.
   - The visual design doesn't include how the navigation links should be looked like with the classic navigation. And this change is awaiting confirmation from the design team. If it needs more adjustments, I will create another PR for it.

### Screenshots:

https://user-images.githubusercontent.com/17420811/162701688-638cdd88-b76f-4fe5-989d-75ad47607aba.mp4

💡  If you are using a local site for testing, please set the site to be accessible from the public network first.

1. Enable the new WooCommerce navigation on the WC Settings page: `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`
   - Maybe it needs to install the **WooCommerce Admin** plugin first to have that feature option.
   - image
1. Go to the Connection Test page `/wp-admin/admin.php?page=connection-test-admin-page`.
1. Click on the "Disconnect Jetpack" button.
1. Go to a GLA page. It should make the GLA menu links of WC Navigation disabled except for the Settings link.
1. Go to another admin page that is rendered by React. For example, the **Marketing > Overview** page.
1. When clicking on the GLA Settings link for going back, it should redirect to the WPCOM account reconnecting page.
1. After the reconnecting process is finished and back to the GLA Settings page, other page links should be enabled.

### Additional details:

Before any 401 error occurs, the GLA menu links of WC Navigation will still be enabled since it only starts locking the re-connecting link as the available link once it enters the re-connecting page.

### DIFF
--- a/js/src/settings/reconnectionLock.js
+++ b/js/src/settings/reconnectionLock.js
@@ -29,7 +29,7 @@ let unlockCallback = null;
  * and uses the current link as the locked link. When clicking on the Settings link,
  * it will redirect to the locked link.
  */
-function lockInReconnection() {
+export function lockInReconnection() {
 	if ( unlockCallback ) {
 		return;
 	}
@@ -86,11 +86,9 @@ function lockInReconnection() {
  * Enables the GLA menu links of WC Navigation that are locked by
  * the `lockInReconnection` function.
  */
-function unlockFromReconnection() {
+export function unlockFromReconnection() {
 	if ( unlockCallback ) {
 		unlockCallback();
 		unlockCallback = null;
 	}
 }
-
-export { lockInReconnection, unlockFromReconnection };
--- a/js/src/settings/reconnectionLock.js
+++ b/js/src/settings/reconnectionLock.js
@@ -1,0 +1,96 @@
+/**
+ * External dependencies
+ */
+import { getHistory, getNewPath } from '@woocommerce/navigation';
+import { getQueryArgs } from '@wordpress/url';
+import { isEqual } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import { pagePaths } from '.~/utils/urls';
+import isWCNavigationEnabled from '.~/utils/isWCNavigationEnabled';
+import './reconnectionLock.scss';
+
+/*
+ * When required accounts were disconnected, the available link is
+ * the corresponding reconnecting subpage under the Settings page.
+ *
+ * But there is no reliable way to disable the menu link of WC Navigation,
+ * so the given workaround is using a pair of singleton locking and unlocking
+ * functions to disable the GLA menu links globally, and lock the current link
+ * as the available link when redirecting.
+ */
+
+let unlockCallback = null;
+
+/**
+ * Disables the GLA menu links of WC Navigation except for the Settings link
+ * and uses the current link as the locked link. When clicking on the Settings link,
+ * it will redirect to the locked link.
+ */
+function lockInReconnection() {
+	if ( unlockCallback ) {
+		return;
+	}
+
+	const lockedQuery = getQueryArgs( document.location.search );
+	const handleClick = ( e ) => {
+		const { target } = e;
+
+		if ( target.tagName !== 'A' ) {
+			return;
+		}
+
+		const paths = [
+			pagePaths.dashboard,
+			pagePaths.reports,
+			pagePaths.productFeed,
+			pagePaths.settings,
+		];
+
+		const targetQuery = getQueryArgs( target.getAttribute( 'href' ) );
+		if ( paths.includes( targetQuery.path ) ) {
+			e.stopPropagation();
+			e.preventDefault();
+
+			/**
+			 * With WC Navigation, the available entry link on menu is the GLA Settings;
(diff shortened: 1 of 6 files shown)


js/src/utils/urls.js
```diff
@@ -8,9 +8,15 @@ import { getNewPath } from '@woocommerce/navigation';
...
-const settingsPath = '/google/settings';
+export const pagePaths = {
+	getStarted: '/google/start',
+	setupMC: '/google/setup-mc',
+	setupAds: '/google/setup-ads',
+	dashboard: '/google/dashboard',
+	reports: '/google/reports',
+	productFeed: '/google/product-feed',
+	settings: '/google/settings',
+};
```
puntope, line 19: ❓ / 💅  

Maybe not for this PR (or also we want this to be explicit) but since all the paths has google as prefix and are just the `kebab-case` transformation of the key we could consider to have a getGLAPagePath function.

js/src/settings/reconnectionLock.scss
```diff
@@ -0,0 +1,24 @@
+// The workaround for making the GLA menu links of WC Navigation appear as disabled
```
puntope, line 1: ❓ /  💅 

Regarding last-child. Would not be more accurate to use the href, or setup some kind of disable class on the items? Just in case someday we change Settings position in the menu.

js/src/settings/reconnectionLock.js
```diff
@@ -0,0 +1,96 @@
+/**
```
puntope, line 1: Nice workaround. 

💅  I don't really see the unlock/lock Reconnection as a good naming IMHO. Seems for me more declarative "disableMenuItems" or "enableMenuItems" at the end is what we are doing right?

puntope (commented):
I tested the Happy path and works, as well as entering manually in the URL also triggers the reconnect and disables the links.

However, there is a case when I go back to Home and then click back in google Listing and Ads, I still see the links active. 


https://user-images.githubusercontent.com/5908855/162761283-02933f06-3034-4776-8ca4-31363cb17c3e.mov

js/src/utils/urls.js
eason9487: Being explicit might be better for this since they have the same strings on the PHP side.

js/src/settings/reconnectionLock.js
eason9487: It doesn't only disable the menu links but also locks a link as the available one and tries to redirect the browser to that one when needed. The primary purpose is a bit closer to locking the re-connecting page as the only available page.

js/src/settings/reconnectionLock.scss
eason9487: > or setup some kind of disable class on the items?

The menu is rendered by React as well. Adding CSS class or setting `disabled` attribute could still be cleared when re-rendering.

> Regarding last-child. Would not be more accurate to use the href

I considered it, but since the menu is controlled by WC core, there are chances that it will be unreliable either way. Such as, some query strings in `href` are escaped, and their orders are not guaranteed. But on second thought, it's still a little better maintenance than the last-child method. I adjusted it in https://github.com/woocommerce/google-listings-and-ads/pull/1418/commits/aa8260028e76ca97611c4bd44e18ecec760e29c2.

eason9487:

> I tested the Happy path and works, as well as entering manually in the URL also triggers the reconnect and disables the links.
> 
> However, there is a case when I go back to Home and then click back in google Listing and Ads, I still see the links active.

This is the case mentioned in the Additional details section of this PR. We cannot detect whether any accounts are disconnected before the user enters GLA admin pages. And eventually, the user will still be brought to the re-connecting page after viewing any GLA admin pages, then triggering the lock mechanism.

js/src/settings/reconnectionLock.scss
puntope: Thanks for the change. 

💅  I was thinking more about `&:not([href*="path=%2Fgoogle%2Fsettings"])`  WDTY?

puntope (approved):
✅  LGTM - All of the potential issues were clarified. Thanks :) 

💅  Still some optional non blocking suggestion about the css part.

js/src/settings/reconnectionLock.scss
eason9487: > I was thinking more about `&:not([href*="path=%2Fgoogle%2Fsettings"])` WDTY?

From the perspective of reliability, when using the `href` attribute to select the menu links but if the CSS selector fails to match the original targets due to an update in the future:
- Include-list (the current selector) has a higher chance of mismatch, so the targeting GLA menu links could be displayed as enabled.
- Exclude-list (the `&:not([href*="path=%2Fgoogle%2Fsettings"])` selector) has a higher chance of matching the menu link of GLA Settings and leads to the Settings link being displayed as disabled.

After comparing this, I chose to use the include-list approach to reduce the possibility of the GLA Settings link being displayed as disabled.

PR #1420, opened 2022-04-11
Edit Free Listings: `saveShippingTimes` to delete and upsert shipping times

### LINKED ISSUE #1416: Edit Free Listings: Shipping times reappear after saving changes and reloading the page
### Describe the bug:

In Edit Free Listings, after we delete a shipping time and click on the "save changes" button, it appears to be deleted on the front-end, but when we reload the page, the shipping times still show up. They are actually not deleted at all.

At the code level, if we look for `deleteShippingTimes`, it is only called from Setup MC flow, but not in Edit Free Listings flow.

Note: 

- This issue is part of https://github.com/woocommerce/google-listings-and-ads/issues/1304. 
- This issue deals with the missing `deleteShippingTimes` to call delete shipping time API.

📹  Demo video with my voice:

https://user-images.githubusercontent.com/417342/162479661-04a5c01a-249c-42d6-a5c6-74b43167a96a.mov

### Steps to reproduce:

Pre-condition: you should already have Free Listings setup with the "selected countries only" option selected.

1. Go to Edit Free Listings.
2. In Step 1, "selected countries only" should have been previously selected. Remove a country from the dropdown. Take note of the countries selected there.
3. In Step 2, go to shipping rates section, and remove those unneeded countries so that you would only have shipping rates for the countries you selected above. This is so that you can click on the "Save changes" button below.
4. Go to shipping times section, and remove those unneeded countries so that you would only have shipping times for the countries you selected above.
5. Click on "Save changes" button.
6. Reload the page and look at the shipping times.

### Expected behavior:

Deleted shipping times should not show up.

### Actual behavior:

Deleted shipping times still show up and they are not deleted.

### Additional details:

I looked into older versions and it seems like this issue has been around since the beginning. https://github.com/woocommerce/google-listings-and-ads/releases/tag/1.0.0 has this issue.

ecgan:

Closes https://github.com/woocommerce/google-listings-and-ads/issues/1416.

This PR fixes #1416 by using a new hook called `useSaveShippingTimes`. It is similar with the `useSaveShippingRates` hook introduced in PR https://github.com/woocommerce/google-listings-and-ads/pull/1313.

- `saveShippingTimes` is a function that that wraps around the existing `deleteShippingTimes` and `upsertShippingTimes` wp.data actions.
- `deleteShippingTimes` will be called for shipping times that have been deleted.
- `upsertShippingTimes` will be called for changed or newly added shipping times.
- There will not be API calls for shipping times that are not changed.

### Screenshots:

📹  Demo video with my voice:

https://user-images.githubusercontent.com/417342/162801404-f8ada1d8-b051-4b1c-9a31-bd396c216a96.mov

1. Go to Edit Free Listings.
2. In target audience step, delete an existing country, and add another country.
3. In shipping step, make sure that the shipping rates and shipping times for the deleted country has been removed. You may have to remove them manually.
4. Add shipping rate and shipping time for the newly added country.
5. Click on "Save changes" button.
6. Reload the page.
7. The shipping rates and shipping times setup should be the same as before the reload.

### Changelog entry

> Fix - Call API to delete shipping times in Edit Free Listings when users remove the shipping times.

### DIFF
--- a/js/src/edit-free-campaign/index.js
+++ b/js/src/edit-free-campaign/index.js
@@ -26,6 +26,7 @@ import useDispatchCoreNotices from '.~/hooks/useDispatchCoreNotices';
 import HelpIconButton from '.~/components/help-icon-button';
 import hasUnsavedShippingRates from './hasUnsavedShippingRates';
 import useSaveShippingRates from '.~/hooks/useSaveShippingRates';
+import useSaveShippingTimes from '.~/hooks/useSaveShippingTimes';
 
 /**
  * Function use to allow the user to navigate between form steps without the prompt.
@@ -50,33 +51,6 @@ function isNotOurStep( location ) {
  */
 
 /**
- * Due to the lack of a single API for updating shipping data altogether,
- * we need to send upserts separately for each price/time.
- * Deletes are not needed, as we validate the form not to miss any available country.
- *
- * @param {(groupedCountrySetting: ShippingRate) => Promise} batchUpsertAction
- * @param {Array<ShippingRate>} newData
- * @param {Function} getGroupKey A callback function to ask for a unique key for each shipping group.
- */
-function saveShippingData( batchUpsertAction, newData, getGroupKey ) {
-	const groupMap = new Map();
-	newData.forEach( ( item ) => {
-		const key = getGroupKey( item );
-		const { countryCode, ...rest } = item;
-		let group = groupMap.get( key );
-
-		if ( ! group ) {
-			group = { ...rest, countryCodes: [] };
-			groupMap.set( key, group );
-		}
-
-		group.countryCodes.push( countryCode );
-	} );
-
-	return Array.from( groupMap.values() ).map( batchUpsertAction );
-}
-
-/**
  * Page Component to edit free campaigns.
  * Provides two steps:
  *  - Choose your audience
@@ -93,12 +67,9 @@ export default function EditFreeCampaign() {
 	} = useTargetAudienceFinalCountryCodes();
 
 	const { settings: savedSettings } = useSettings();
-	const {
-		saveTargetAudience,
-		saveSettings,
-		upsertShippingTimes,
-	} = useAppDispatch();
+	const { saveTargetAudience, saveSettings } = useAppDispatch();
 	const { saveShippingRates } = useSaveShippingRates();
+	const { saveShippingTimes } = useSaveShippingTimes();
 
 	const [ targetAudience, updateTargetAudience ] = useState(
 		savedTargetAudience
@@ -223,11 +194,7 @@ export default function EditFreeCampaign() {
 				saveTargetAudience( targetAudience ),
 				saveSettings( settings ),
 				saveShippingRates( shippingRates ),
-				...saveShippingData(
-					upsertShippingTimes,
-					shippingTimes,
-					( item ) => item.time
-				),
+				saveShippingTimes( shippingTimes ),
 			] );
 
 			// Sync data once our changes are saved, even partially succesfully.
--- a/js/src/edit-free-campaign/index.js
+++ b/js/src/edit-free-campaign/index.js
@@ -46,11 +46,6 @@ function isNotOurStep( location ) {
 }
 
 /**
- * @typedef {import('.~/data/actions').ShippingRate} ShippingRate
- * @typedef {import('.~/data/actions').CountryCode} CountryCode
- */
-
-/**
  * Page Component to edit free campaigns.
  * Provides two steps:
  *  - Choose your audience
--- a/js/src/hooks/useSaveShippingTimes.js
+++ b/js/src/hooks/useSaveShippingTimes.js
@@ -1,0 +1,99 @@
+/**
+ * External dependencies
+ */
+import { useCallback } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { useAppDispatch } from '.~/data';
+import useShippingTimes from './useShippingTimes';
+import getDeletedShippingTimes from '.~/utils/getDeletedShippingTimes';
+import getDifferentShippingTimes from '.~/utils/getDifferentShippingTimes';
+
+/**
+ * @typedef { import(".~/data/actions").ShippingTime } ShippingTime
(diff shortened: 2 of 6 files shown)


js/src/edit-free-campaign/index.js
```diff
@@ -49,33 +50,6 @@ function isNotOurStep( location ) {
  * @typedef {import('.~/data/actions').CountryCode} CountryCode
  */
 
-/**
- * Due to the lack of a single API for updating shipping data altogether,
- * we need to send upserts separately for each price/time.
- * Deletes are not needed, as we validate the form not to miss any available country.
- *
- * @param {(groupedCountrySetting: ShippingRate) => Promise} batchUpsertAction
```
eason9487, line 57: 💅 
Since there is no `ShippingRate` and `CountryCode` reference in this file JSDoc, their `@typedef` could be removed as well.

js/src/hooks/useSaveShippingTimes.js
```diff
@@ -0,0 +1,99 @@
...
+const getShippingTimesGroups = ( shippingTimes ) => {
+	const timeGroupMap = new Map();
+	shippingTimes.forEach( ( { countryCode, time } ) => {
+		const group = timeGroupMap.get( time ) || { countryCodes: [], time };
+		group.countryCodes.push( countryCode );
+		timeGroupMap.set( time, group );
+	} );
+
+	return Array.from( timeGroupMap.values() );
+};
```
eason9487, line 54: ❓
Apart from the property name `countryCodes`, the `getShippingTimesGroups` function is completely the same as the two existing functions:

- [getCountriesTimeArray](https://github.com/woocommerce/google-listings-and-ads/blob/be3ad280d25d7177e725f7a2556c1c4281f7f0a6/js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/getCountriesTimeArray.js)
- [getCountriesTimeArray](https://github.com/woocommerce/google-listings-and-ads/blob/be3ad280d25d7177e725f7a2556c1c4281f7f0a6/js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/getCountriesTimeArray.js)

Would you think they could be reused? Or even further reduced to a single one?

eason9487 (commented):
The functions in `useSaveShippingTimes` hook work well if only test their implementations. However, the scope of this PR and the correction target seems a bit confusing when viewing from the shipping edit form.

❓ 
It seems that if this PR was based on PR #1409, it would make the target issues to be fixed look clearer. The current merge target is `develop`, which makes this PR look like it doesn't fully fix the problem in #1416.

While I can probably understand the overall context of this issue, it might be a bit confusing for other reviewers. For examples:
1. If delete the shipping time directly in step 2, it cannot be saved, which means that the issue description and reproduction steps in #1416 are probably incomplete.
2. If just delete the audience country in step 1 and then save it directly in step 2, the shipping data stored in APIs are still inconsistent (consistency in the shipping APIs data is the expected result of #1304).

js/src/edit-free-campaign/index.js
ecgan: Good catch. Removed in [`7545d59` (#1420)](https://github.com/woocommerce/google-listings-and-ads/pull/1420/commits/7545d592be0b67a96e7ea3dcaa466cce676aada4).

js/src/hooks/useSaveShippingTimes.js
ecgan: That's a good idea actually, to reduce to a single one. I forgot about the `getCountriesTimeArray` functions, and I created the `getShippingTimesGroups` based on the deleted `saveShippingData` (https://github.com/woocommerce/google-listings-and-ads/pull/1420/files#diff-5297a79a23cac9dd2adeb1b6289484c6a7ebe09f096a488d19fd7ac68213fd1dL61-L77). 😂 

However, I'm not very keen in doing it in this PR. If I were to do it now, it means the Setup MC area would be impacted as well, and make this PR more complicated. Also, `getShippingTimesGroups` has a slightly different structure in its returned data `{ countryCodes, time }`, while `getCountriesTimeArray` has `{ countries, time }`. 

I'm thinking to have a code refactor on shipping time section to make it similar with shipping rate section (hopefully soon). This means that eventually we will have one copy of `getCountriesTimeArray` instead of two copies. I can think of how to consolidate `getShippingTimesGroups` and `getCountriesTimeArray` at that time (say, changing `getCountriesTimeArray` to return `{ countryCodes, time }`, since upsert shipping time works with that structure), and come out with tests for it.

I'll come out with a follow-up issue on this.

ecgan: I've created issue https://github.com/woocommerce/google-listings-and-ads/issues/1426 for this.

ecgan:
> It seems that if this PR was based on PR #1409, it would make the target issues to be fixed look clearer. The current merge target is `develop`, which makes this PR look like it doesn't fully fix the problem in #1416.
> ...
> 1\. If delete the shipping time directly in step 2, it cannot be saved, which means that the issue description and reproduction steps in #1416 are probably incomplete.

Thanks for pointing that out (and with your further explanation in [slack thread](p1649847139065299-slack-C01E9LB4X61)). I have updated the issue description and reproduction steps in #1416. My intention is for #1416 to deal with the missing delete shipping time API call, and for it to be independent and unrelated to form data interaction between Edit Free Listings Step 1 and Step 2. 

The intention for this PR #1420 is to purely fix the missing delete shipping time API call in #1416.

> 2\. If just delete the audience country in step 1 and then save it directly in step 2, the shipping data stored in APIs are still inconsistent (consistency in the shipping APIs data is the expected result of #1304).

Yeah, so turns out issue #1304 consists of two smaller sub-issues #1402 (form data problem between Edit Free Listings Step 1 and Step 2) and #1416 (missing delete shipping time API call problem). 

I have rebased this PR #1420 on top of PR #1409. When this PR #1420 is merged into PR #1409, issue #1416 should be closed. Then when PR #1409 is merged, issue #1304 should be closed.

eason9487 (approved):
Thanks for the PR update! Tested locally and works well. LGTM.

PR #1421, opened 2022-04-13
Replace all country selectors with TreeSelectControl component

eason9487:

This PR implemented the **Replace the single country selection component with the dedicated component** and **Refactor the countries selectors** parts in #1356

- Change \<SupportedCountrySelect\> to use \<TreeSelectControl\>
- Change \<AudienceCountrySelect\> to use \<SupportedCountrySelect\>
- Replace all \<AppCountrySelect\> with \<SupportedCountrySelect\> in the shipping rate/time settings

📌  The modal viewport covers most parts of \<TreeSelectControl\> within pop-up modals of shipping. This was the same when using the former selector but it's getting worse after replacement. I'll check this UI problem with the design team. If further adjustment is needed, I will open a separate PR for it.

#### 📷  Create new campaign
image

#### 📷  Edit a campaign
image

#### 📷  Add shipping rates
image

#### 📷  Edit shipping rates
image

#### 📷  Add free threshold of shipping rates
image

#### 📷  Edit free threshold of shipping rates
image

#### 📷  Add shipping times
image

#### 📷  Edit shipping times
image

#### Campaign creation and edit

1. The country selector should be the \<TreeSelectControl\> component.
1. Finish the campaign creation and go to edit the just created campaign.
1. The countries listed on the \<TreeSelectControl\> should be the same when creating and cannot be edited.

#### Shipping settings

1. Go to step 2 of the onboarding setup.
1. The country selector for audience should be the \<TreeSelectControl\> component.
1. Go to the next step for setting up shipping.
1. Check every country selectors of all shipping rates, times, and thresholds. All selectors should be the \<TreeSelectControl\> component.
1. After changing the selecting countries, the selected ones should be reflected on the relevant UI and the subsequent setup processing.
1. Go to step 1 of editing the free listings and repeat steps 2-5. 

### Additional details:

Originally, I wanted to develop phases 2 and 3 separately to reduce code conflicts with Shipping Integration. However, considering that the shipping rate part of Shipping Integration has been merged and the testing scope of both phases is almost the same. In order to reduce the effort of code reviews and testing, I decided to finish them in the same PR.


puntope:
❓  

One thing I just noticed:

- Free Campaign contains for example 3 countries (Italy,France, Spain) ✅ 
- Programs table shows Italy + 2 more. ✅ 
- Create Paid Campaign with 2 countries (example Italy, France)  ✅ 
- Programs table for paid campaign shows  Italy + 1 more ✅ 
- Edit Free Campaign and remove Italy. ✅ 
- Programs table shows France + 1 more. ✅ 
- Programs table for paid campaign shows  Italy + 1 more ❓ 
- Editing the campaign only shows France.

This behaviour get even stranger when I remove all the countries selected in Paid Campaign (in this case France). The Program table still shows  Italy + 1 more but if I click in Edit Campaign it shows an empty selector. I'm able to save tho... but stills show Italy + 1 more. If I add again the countries inside the Free Campaign. And go back to the Edit paid campaign it shows them again. 

As far as I understand would be clearer if we still show the countries in the Paid Campaign even tho are deleted from Free Campaign. 

Video clarifying the procedure:

https://user-images.githubusercontent.com/5908855/163539479-163ed5bc-dc2c-441d-8ba7-86292a32ab25.mov

puntope:
> 📌 The modal viewport covers most parts of <TreeSelectControl> within pop-up modals of shipping. This was the same when using the former selector but it's getting worse after replacement. I'll check this UI problem with the design team. If further adjustment is needed, I will open a separate PR for it.

Probably this is being discussed in some PR but I want to give my personal opinion about one UI idea I have to solve that. 

Maybe we can, instead of opening the modal, render the Selector when the user click on edit just right there between the label and the input for the shipping days/cost.

eason9487:
> This behaviour get even stranger when I remove all the countries selected in Paid Campaign (in this case France). The Program table still shows Italy + 1 more but if I click in Edit Campaign it shows an empty selector. I'm able to save tho... but stills show Italy + 1 more. If I add again the countries inside the Free Campaign. And go back to the Edit paid campaign it shows them again.
> 
> As far as I understand would be clearer if we still show the countries in the Paid Campaign even tho are deleted from Free Campaign.

@puntope, thanks for the nice catch! I added a new prop for appending additional countries when using \<AudienceCountrySelect\> on the campaign edit page. The code diff can be viewed in https://github.com/woocommerce/google-listings-and-ads/pull/1421/commits/933895c948839cf8282ecc7f39b3e56e1c6edb2f.

Edit: Added another commit for tweaking the JSDoc https://github.com/woocommerce/google-listings-and-ads/pull/1421/commits/d2eeb6d685057a67cd9a0a2c52f86a736cd401fa. 🙏

js/src/components/supported-country-select/supported-country-select.js
```diff
@@ -0,0 +1,46 @@
+/**
```
puntope, line 1: ❓  Whats the reason of using 

`js/src/components/supported-country-select/index.js`  for exporting `js/src/components/supported-country-select/supported-country-select.js` and not just directly use `js/src/components/supported-country-select/index.js` or `js/src/components/supported-country-select/supported-country-select.js` ?

Maybe I missed some discussion for the convention?

puntope (approved):
✅  LGTM 

Thanks for the implementation and further changes 🙏

js/src/components/supported-country-select/supported-country-select.js
eason9487: If I remember right, this was advocated by @tomalec. I got this convention (or preference?) during one of code reviews a few months ago, but I couldn't find that comment. The most relevant post I can find is p6q8Tx-1Uk-p2. I also lean toward this method, which brings more benefits when using the editor to find files, and when moving files via git.

ecgan: > not just directly use `js/src/components/supported-country-select/index.js`...

We have this in our code base, and we end up having a lot of code file with `index.js`, and it makes searching and navigating the files in the editors cumbersome, e.g. you will have a lot of "index.js" files in your tabs; when you do cmd+P in VS code and search for "index.js" a lot of results show up, etc. Personally it is not a good experience for me.

> or `js/src/components/supported-country-select/supported-country-select.js` ?

If we do this without the index.js, then when we import the component, the import statement would be longer, like:

```js
import SupportedCountrySelect from '.~/components/supported-country-select/supported-country-select';
```

By having `js/src/components/supported-country-select/index.js` for exporting `js/src/components/supported-country-select/supported-country-select.js`, then we would have a shorter import statement: 

```js
import SupportedCountrySelect from '.~/components/supported-country-select';
```

@eason9487 gave a good P2 post there in his comment. Another related good one is p90Yrv-2d4-p2.

eason9487: > when you do cmd+P in VS code and search for "index.js" a lot of results show up, etc. Personally it is not a good experience for me.

Same. It's inconvenient in Sublime Text too.

puntope: I missed that post. 

> By having js/src/components/supported-country-select/index.js for exporting js/src/components/supported-country-select/supported-country-select.js, then we would have a shorter import statement:

Is the name long really important? I mean with IDE autocompletion I don't get why is bad to have long names. 

My thoughts are that if we are using `index.js` we still have the same problem right?

If you search for `supported-country-select` in this case. It will show you as well the index since the directory matching the search query as well.

puntope: BTW. I agree with not using `index.js` as a name. I guess was also discussed in Google Dev Tools meeting.

ecgan: > Is the name long really important? I mean with IDE autocompletion I don't get why is bad to have long names.

I think it is not about "bad to have long names", it's just that it is "nicer to have shorter import statement". 😆 

Another thing about re-export in `index.js` in a folder is that it tells which files are the main things exported out from the folder, and which files are not meant to be used outside the folder. For example, in https://github.com/woocommerce/google-listings-and-ads/tree/develop/js/src/components/shipping-rate-section/estimated-shipping-rates-card, there are a number of files. By looking at `index.js`, it serves as a hint to us that only `EstimatedShippingRatesCard` are meant to be used outside the folder, while other files are meant to be used within the folder.

tomalec: > If you search for `supported-country-select` in this case. It will show you as well the index since the directory matching the search query as well.

Yes, you will get `index.js` as a result as well, but `supported-country-select.js` has a higher match :) Plus that's file name that is used/emphasized in many other places: search results, git status, debugger, etc.
image

> Another thing about re-export in `index.js` in a folder is that it tells which files are the main things exported out from the folder, and which files are not meant to be used outside the folder […]

:+1: 
In my mental model, thanks to that approach folder structure may improve modular DevX. If a folder contains a module/component of any kind (as opposed to a directory of for example `utils/`) any file or folder inside is considered an implementation detail, and only things exported explicitly in `index.js` are meant/recommended to be used outside. For me that goes in pair with mapping `./my-module`: `./my-module.js`, `./my-module/index.js`

puntope: > By looking at `index.js`, it serves as a hint to us that only `EstimatedShippingRatesCard` are meant to be used outside the folder, while other files are meant to be used within the folder.
> and only things exported explicitly in index.js are meant/recommended to be used outside.

I get the point, but still feels annoying for me the need of create `index.js` file for exporting the real entry point (estimated-shipping-rates-card.js), 2 files instead of 1 per each component created. 

In case is needed to add multiple exports then yes, would be good to use that `index.js`, besides that I liked more when exportable components are in separated and isolated folders. 

As an example, if I have a component in the directory FormField  with the component <FormField> that uses internally a <Label> component and that <Label> could be used "outside" IMO <Label> should be created outside FormField folder, hence no real need for that `index.js`.

PR #1422, opened 2022-04-13
Changes to the shipping rates and suggestions API

### LINKED ISSUE #1395: Better shipping rates suggestions
## User story

This is related to https://github.com/woocommerce/google-listings-and-ads/pull/1341, #1382, #1383.

I was looking into the shipping rates suggestions changes, and I noticed there are a bit of inconsistencies (or, a gap in expectations) in the suggestions data returned. I'm hoping that via this issue, we can have a standardized expectation that can work across the back-end API and front-end UI.

In my WooCommerce shipping zones settings, I have the following setup:

image

The zone names indicate the setting inside the zones: 

- "flat 0" means flat rate with cost 0.
- "flat 25" means flat rate with cost 25.
- "free shipping 0" means free shipping with no minimum amount. 
- "free shipping 50" means free shipping with minimum order amount 50.

With the above shipping zones setup, one would think there would be 12 shipping rates suggestions.

This is what I got when I call the suggestions API:

image

All the countries are there in the response, but there are only 9 suggestions.

With the current shipping rate suggestion data structure, and the current shipping rate UI, it can become complicated and confusing.

**I would like to suggest a "standardized" shipping rate data and UI structure.**

Refer to the following annotated screenshot:

image
 
In short, the idea is to convert WC shipping zone rates into "flat_rate" shipping rates in GLA as much as possible, with "free_shipping" complementing the "flat_rate". Note that this means that the shipping method in WC may not match with the shipping rate method in GLA, e.g. "free shipping" in WC would be converted to "flat rate" with amount 0 in GLA.

Here's the proposed suggestions API response based on the above WC shipping zones settings, with actual response from the above screenshot, and my proposed expected response: (I use ❓ to indicate "inconsistency" / "unknown" questions.)

- Indonesia (flat 0)
    - Actual: `{country: "ID", method: "flat_rate", currency: "USD", rate: 0, options: {}}`
    - Expected: `{country: "ID", method: "flat_rate", currency: "USD", rate: 0, options: {}}`
    - 🟢  Match.

- Myanmar (flat 25)
	- Actual: `{country: "MM", method: "flat_rate", currency: "USD", rate: 25, options: {}}`
	- Expected: `{country: "MM", method: "flat_rate", currency: "USD", rate: 25, options: {}}`
	- 🟢  Match.

- Malaysia (free shipping 0)
	- Actual: `{country: "MY", method: "free_shipping", currency: "USD", rate: 0, options: {}}`
	- Expected: `{country: "MY", method: "flat_rate", currency: "USD", rate: 0, options: {}}`
	- 🔴  Not match - should convert it to flat_rate.

- Vietnam (free shipping 50)
	- Actual `{country: "VN", method: "free_shipping", currency: "USD", rate: 0, options: {free_shipping_threshold: 50} }`
	- Expected: Do not return this suggestion.
	- 🔴  Not match - Vietnam has free shipping with minimum order amount 50, but does not have a flat rate. We should not return this suggestion. Previously I mentioned this in https://github.com/woocommerce/google-listings-and-ads/pull/1363#issuecomment-1078779535 and https://github.com/woocommerce/google-listings-and-ads/pull/1367#issuecomment-1079087068.

- Philippines (flat 0, free shipping 0)
	- Actual: `{country: "PH", method: "free_shipping", currency: "USD", rate: 0, options: {}}`
		- ❓ Note that there are two WC shipping zone methods, but there are only one returned. Why is that?
	- Expected: `{country: "PH", method: "flat_rate", currency: "USD", rate: 0, options: {}}`
	- 🔴  Not match - should use / convert to flat_rate.

- Singapore (flat 0, free shipping 50)
	- Actual: `{country: "SG", method: "flat_rate", currency: "USD", rate: 0, options: {}}` and `{country: "SG", method: "free_shipping", currency: "USD", rate: 0, options: {free_shipping_threshold: 50} }`
		- ❓ Why do we have two suggestions here?
	- Expected (assuming we are returning cheapest shipping rate, hence discarding the "free shipping 50"): `{country: "SG", method: "flat_rate", currency: "USD", rate: 0, options: {}}`
	- 🔴  Not match - the actual response would not be able to work with the current UI, because when we have flat rate 0, it does not make sense to enter minimum order amount 50 in the UI. 

- Thailand (flat 25, free shipping 0)
	- Actual: `{country: "TH", method: "free_shipping", currency: "USD", rate: 0, options: {}}`
	- Expected (assuming we are returning cheapest shipping rate, hence discarding the "flat 25"): `{country: "TH", method: "flat_rate", currency: "USD", rate: 0, options: {}}`
	- 🔴  Not match - should convert it to flat_rate.

- Sri Lanka (flat 25, free shipping 50)
	- Actual: `{country: "LK", method: "flat_rate", currency: "USD", rate: 25, options: {}}` and `{country: "LK", method: "free_shipping", currency: "USD", rate: 0, options: {free_shipping_threshold: 50} }`
	- Expected: `{country: "LK", method: "flat_rate", currency: "USD", rate: 25, options: {}}` and `{country: "LK", method: "free_shipping", currency: "USD", rate: 0, options: {free_shipping_threshold: 50} }`
	- 🟢  Match.

With the above expected API response, we would be able to better display the data in the UI. It would be clearer and consistent that "Estimated shipping rates" card UI works with "flat_rate", and "Minimum order" card UI works with "free_shipping", and this consistency applies to loading and saving shipping rates too (not just shipping rates suggestions).

To implement the above proposal in the shipping rates suggestions API.

To implement the above proposal in the front-end UI, by doing a conversion process after getting the suggestions response. However, this isn't really ideal.

## Acceptance criteria

The API response is as above proposal.

nima-karimi:

Closes #1395.

This PR implements the suggestion changes offered by @ecgan in #1395. For more context please check that issue.

Based on the code introduced in #1341, the suggestions API returned different values that were not expected by the front-end code. This led to some discrepancies between the data structures defined in the UI and the data returned by the API. @ecgan and I discussed this and we concluded that it would be best if the data structures were the same. Given that the suggestions API is only used by our own UI code, I've made some changes to make the API responses aligned with how the UI expects them.

With this PR, only the `flat_rate` shipping method is accepted by the shipping rates APIs. The `free_shipping` method is no longer accepted and to indicate free shipping, a rate of 0 should be set for the shipping method `flat_rate`.

When a `free_shipping_threshold` value is set on a `flat_rate` method, it means that if the minimum order amount hits that value, the shipping rate will be free regardless of the `rate` value.

Moreover, if a country has only a conditional free shipping rate (with a minimum order amount/free_shipping_threshold) and no other rates available, then the conditional free rate will be ignored and no suggestions will be returned for that country.

To parse and format the shipping suggestions, a new ShippingSuggestionService is added with unit tests.

@ecgan I think it would be best if you could test this PR and give a green light since we've discussed this before.

1. Go over the scenarios mentioned in #1395 and confirm the results are as expected.

### Changelog entry

>


src/Shipping/LocationRatesProcessor.php
```diff
@@ -56,21 +76,8 @@ public function process( array $location_rates ): array {
...
-		} elseif ( $new_rate instanceof ShippingRateFree &&
-				   $existing_rate instanceof ShippingRateFree &&
-				   (float) $new_rate->get_threshold() > (float) $existing_rate->get_threshold()
-		) {
-			$replace = true;
-		}
-
-		return $replace;
+		return $new_rate->get_rate() > $existing_rate->get_rate() ||
+			   (float) $new_rate->get_min_order_amount() > (float) $existing_rate->get_min_order_amount();
```
jorgemd24, line 59: ❓

I think this should be:

```
		return $new_rate->get_rate() < $existing_rate->get_rate() ||
			   (float) $new_rate->get_min_order_amount() < (float) $existing_rate->get_min_order_amount();
```

Otherwise it will return the most expensive rate.

src/Shipping/ShippingSuggestionService.php
```diff
@@ -0,0 +1,96 @@
...
+			$serialized = $location_rate->jsonSerialize();
+
+			// Check if there is a conditional free shipping rate (with minimum order amount).
+			// We will set the minimum order amount as the free shipping threshold for other rates.
+			$shipping_rate = $location_rate->get_shipping_rate();
+			if ( $shipping_rate->is_free() && $shipping_rate->has_min_order_amount() ) {
+				$free_threshold = $shipping_rate->get_min_order_amount();
+
+				// Ignore the conditional free rate if there are other rates.
+				if ( count( $location_rates ) > 1 ) {
```
jorgemd24, line 70: ❔

I am not really sure why this is necessary, for example if we have a free shipping with a min. order amount and no other rates, the rate is skipped anyway. 

I think we could remove `if ( count( $location_rates ) > 1 )` but I do not know if I am missing something?

tests/Unit/Shipping/LocationRatesProcessorTest.php
```diff
@@ -22,69 +21,88 @@ class LocationRatesProcessorTest extends UnitTest {
...
+			new LocationRate( $location, new ShippingRate( 200 ) ),
 		];
-		$processed = $this->rates_processor->process( $location_rates );
+		$processed      = $this->rates_processor->process( $location_rates );
 		$this->assertEquals( 1, count( $processed ) );
-		/** @var ShippingRateFlat $flat_rate */
-		$flat_rate = $processed[0]->get_shipping_rate();
-		$this->assertInstanceOf( ShippingRateFlat::class, $flat_rate );
-		$this->assertEquals( 200, $flat_rate->get_rate() );
+		$this->assertEquals( 200, $processed[0]->get_shipping_rate()->get_rate() );
```
jorgemd24, line 30: ❓

Should not return the cheapest value? 50?

tests/Unit/Shipping/LocationRatesProcessorTest.php
```diff
@@ -22,69 +21,88 @@ class LocationRatesProcessorTest extends UnitTest {
...
-		$this->assertEquals( 1, count( $processed ) );
-		/** @var ShippingRateFree $free_rate */
-		$free_rate = $processed[0]->get_shipping_rate();
-		$this->assertInstanceOf( ShippingRateFree::class, $free_rate );
-		$this->assertEquals( 200, $free_rate->get_threshold() );
+		$processed      = $this->rates_processor->process( $location_rates );
+		$this->assertEquals( 2, count( $processed ) );
+		foreach ( $processed as $location_rate ) {
+			if ( $location_rate->get_shipping_rate()->is_free() ) {
+				$this->assertEquals( 200, $location_rate->get_shipping_rate()->get_min_order_amount() );
```
jorgemd24, line 55: ❓

Should not return 50 as the minum order amount?

jorgemd24 (commented):
Thanks, @nima-karimi 👍  I have left some comments about the code, in general, is working as expected, except that is returning the most expensive rate instead of the cheapest.

ecgan:
@nima-karimi ,

> @ecgan I think it would be best if you could test this PR and give a green light since we've discussed this before.
> 
> 1. Go over the scenarios mentioned in [Better shipping rates suggestions #1395](https://github.com/woocommerce/google-listings-and-ads/issues/1395) and confirm the results are as expected.

I tested this out using this WC shipping zones setting as per #1395, and I can confirm this works perfect and matches the expectations in #1395! 💯  

When I visit the Setup MC shipping rate section, and selects the "Simple" flat shipping rate options, this is the UI that I get, which is all good 👍 :

image

This is the shipping rate suggestions API response data that I got:

```json
[
  {
    "country": "ID",
    "method": "flat_rate",
    "currency": "USD",
    "rate": 0,
    "options": {}
  },
  {
    "country": "MM",
    "method": "flat_rate",
    "currency": "USD",
    "rate": 25,
    "options": {}
  },
  {
    "country": "MY",
    "method": "flat_rate",
    "currency": "USD",
    "rate": 0,
    "options": {}
  },
  {
    "country": "PH",
    "method": "flat_rate",
    "currency": "USD",
    "rate": 0,
    "options": {}
  },
  {
    "country": "SG",
    "method": "flat_rate",
    "currency": "USD",
    "rate": 0,
    "options": {}
  },
  {
    "country": "TH",
    "method": "flat_rate",
    "currency": "USD",
    "rate": 0,
    "options": {}
  },
  {
    "country": "LK",
    "method": "flat_rate",
    "currency": "USD",
    "rate": 25,
    "options": { "free_shipping_threshold": 50 }
  }
]

```

---

One side question: 

> With this PR, only the `flat_rate` shipping method is accepted by the shipping rates APIs. The `free_shipping` method is no longer accepted and to indicate free shipping, a rate of 0 should be set for the shipping method `flat_rate`.

Will there be any method other than `flat_rate` in the future? If there is only going to be one `flat_rate` method, maybe we can consider removing `method` field from shipping rates in future PR? 🤔  

Note: if we were to remove the `method` field, UI front-end code change is required.

src/Shipping/LocationRatesProcessor.php
nima-karimi: This is intentional and specified in the requirements of this feature. The reasoning is that since we are grouping and estimating the rates it would be safer just to assume the most expensive rate for the group rather than applying the cheapest rate. For example, if a country-wide rate is $50 but a state-wide rate is $20, it would make more sense to apply the $50 rate to the whole country rather than the $20 rate.

tests/Unit/Shipping/LocationRatesProcessorTest.php
nima-karimi: No it's intentionally returning the most expensive rate as I mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1422#discussion_r851109599

tests/Unit/Shipping/LocationRatesProcessorTest.php
nima-karimi: see https://github.com/woocommerce/google-listings-and-ads/pull/1422#discussion_r851109599

The same rule applies to the minimum order amount because we wouldn't want to set a low minimum order amount for a group if it only applies to a state or postcode within a country, etc.

src/Shipping/ShippingSuggestionService.php
nima-karimi: > I am not really sure why this is necessary, for example if we have a free shipping with a min. order amount and no other rates, the rate is skipped anyway.

This is for the inverse condition, meaning that we have conditional free shipping rates AND other rates. What this part does is that it gets the `min_order_amount` value from a conditional free shipping rate and adds it as a `free_shipping_threshold` to all the other rates. When we add that value to the other rates, we need to remove the free shipping rate from the list because it's already implied by that `free_shipping_threshold` value in other rates.

This is just a data format change so that it can be returned the way UI expects it.

Instead of returning this data:

```JSON
    {
        "country": "US",
        "method": "flat_rate",
        "currency": "USD",
        "rate": 20
    }
    {
        "country": "US",
        "method": "flat_rate",
        "currency": "USD",
        "rate": 0,
        "min_order_amount": 100
    }
```

We would return the following in the API response:

```JSON
    {
        "country": "US",
        "method": "flat_rate",
        "currency": "USD",
        "rate": 20,
        "options": {
            "free_shipping_threshold": 100
        }
    }
```

nima-karimi: > I think we could remove if ( count( $location_rates ) > 1 ) but I do not know if I am missing something?

I rather keep this condition because if we remove it then the code here would depend on the conditions defined in other classes with no way to check if the logic is as expected. This condition helps make sure the conditional free rate is always ignored when its parameters are applied to the other rates.

src/Shipping/LocationRatesProcessor.php
jorgemd24: Thanks, @nima-karimi for the clarification, I got confused because [in this comment](https://github.com/woocommerce/google-listings-and-ads/issues/1395#issue-1190146411) we mentioned we would expect to return the cheapest shipping rate:

>Thailand (flat 25, free shipping 0)
Actual: {country: "TH", method: "free_shipping", currency: "USD", rate: 0, options: {}}
Expected (assuming **we are returning cheapest shipping rate**, hence discarding the "flat 25"): {country: "TH", method: "flat_rate", currency: "USD", rate: 0, options: {}}
🔴 Not match - should convert it to flat_rate.

I thought that we would follow the same idea, to suggest the cheapest and not the most expensive. 

On the other hand, I see the point that it is safer to show the most expensive although maybe it is not the best suggestion.

nima-karimi:
> Will there be any method other than flat_rate in the future? If there is only going to be one flat_rate method, maybe we can consider removing method field from shipping rates in future PR? 🤔

I thought about doing this but we do have the pick up method that might be introduced in the front-end. I'm not sure if we are ever going to introduce options for the user to set a `pickup` method in the UI. If we don't have any intentions of doing that in the near future then we can remove it completely.

src/Shipping/ShippingSuggestionService.php
jorgemd24: Thanks @nima-karimi that makes sense 👍

src/Shipping/LocationRatesProcessor.php
nima-karimi: Oh yeah, that's only the case if there is a free shipping method involved. Although I see the contradictions here. Given the same reasoning, one could argue that we should not return the free shipping but the most expensive flat-rate shipping for that group.

I think it would work best if we return the most expensive rate just to be on the safer side and don't surprise the user with a low rate on their Google profile. I can remove the condition that sets all rates to free if there is a free_shipping rate involved. @jorgemd24 @ecgan Do you think this would be preferable? Using the same Thailand example shared above, we would return the $25 rate instead of the $0

jorgemd24: In my opinion, the issue is because it is difficult to make the balance between "safe" and a good suggestion only based on the rates that are set up in WC.

It sounds right that for example if Thailand has free shipping it should return a rate of $0 although, as you mentioned before, it is contradictory to the requirements of this feature. If we think from the "safer side", the free shipping could never apply if there is other flat rate > $0.

When there is more than one flat rate > $0, it is impossible to know which one is the right one so I suppose it is good to go with the most expensive, although this can go against the user's expectations. 

We need just to make sure that everyone understands that we are not returning the cheapest shipping rate, we return free shipping if exists otherwise we return the most expensive.

nima-karimi: Yes, it's a trade-off. And we're merely guessing the rates to make the onboarding smoother and also telling them to always check them in their Merchant Center settings after the onboarding is done. I think to be consistent it would be best to always return the most expensive rate regardless of whether they have a free-shipping rate or not. I've made some changes in https://github.com/woocommerce/google-listings-and-ads/pull/1422/commits/4f79a601e47dd46fbb9634a4b9297269214414af to always return the most expensive rate.

nima-karimi:
@ecgan I've removed the `method` from all shipping rate-related APIs and also the DB tables. You can create another PR off of this one to apply the necessary changes to the UI code.

I don't think we'll be adding the pickup method to the UI anytime soon since it clutters the current UI and goes against the phrase "My shipping settings are simple".

nima-karimi:
@jorgemd24 @ecgan Thanks for all the comments. I think it would be best to do another round of testing given that I've changed some logic in the last commits to this PR. To test the UI, @ecgan needs to first apply the changes.

ecgan:
@nima-karimi , 

> @ecgan I've removed the `method` from all shipping rate-related APIs and also the DB tables. You can create another PR off of this one to apply the necessary changes to the UI code.

Okay, I'll work on that.

> I think it would be best to do another round of testing given that I've changed some logic in the last commits to this PR.

I tested the latest code change here with the same WC shipping zones settings. Below is the suggestions API response:

```json
[
    {
        "country": "ID",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "MM",
        "currency": "USD",
        "rate": 25,
        "options": {}
    },
    {
        "country": "MY",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "PH",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "SG",
        "currency": "USD",
        "rate": 0,
        "options": {
            "free_shipping_threshold": 50
        }
    },
    {
        "country": "TH",
        "currency": "USD",
        "rate": 25,
        "options": {}
    },
    {
        "country": "LK",
        "currency": "USD",
        "rate": 25,
        "options": {
            "free_shipping_threshold": 50
        }
    }
]
```

Other than the removal of `method`, there are two countries that are different from the previous API response in https://github.com/woocommerce/google-listings-and-ads/pull/1422#issuecomment-1099495644:

- For `SG`, we have rate 0 and free_shipping_threshold 50. That doesn't look right? 🤔 
- For `TH`, the rate is now 25 instead of the previous 0. This is expected since we are now suggesting the most expensive rate, as per https://github.com/woocommerce/google-listings-and-ads/pull/1422#discussion_r851197518.

nima-karimi:
Thanks a lot for testing this again, @ecgan 🙏 I've made some changes in https://github.com/woocommerce/google-listings-and-ads/pull/1422/commits/10a74e75f34d22cfe242867d97d0b899eef23af8 that should resolve the issue with the `SG` case. Let me know how it works out.

ecgan:
@nima-karimi , thank you. I checked out your code and here's the suggestions API response:

```json
[
    {
        "country": "ID",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "MM",
        "currency": "USD",
        "rate": 25,
        "options": {}
    },
    {
        "country": "MY",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "PH",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "SG",
        "currency": "USD",
        "rate": 0,
        "options": {}
    },
    {
        "country": "TH",
        "currency": "USD",
        "rate": 25,
        "options": {}
    },
    {
        "country": "LK",
        "currency": "USD",
        "rate": 25,
        "options": {
            "free_shipping_threshold": 50
        }
    }
]
```

For the "SG" case, now it has rate 0 without the free_shipping_threshold option. Seems all good now. 💯 🙂

jorgemd24 (commented):
Thanks @nima-karimi for the changes, I had a look at the Google MC and I saw that the name of the rate was "false" and it is because the following warning:  `WARNING sprintf(): Too few arguments on line number 10`

image

https://github.com/woocommerce/google-listings-and-ads/blob/10a74e75f34d22cfe242867d97d0b899eef23af8/src/API/Google/Settings.php#L300-L305

It should be ` 'Flat rate - %1$s %2$s'`:

I also notice that when there is a country for example Italy with a flat rate: of $5 and  free shipping with a minimum order is not syncing properly.

I get two flat rates in the MC - One flat rate of $5 and a flat rate of $0 and I can't see the minimum order restriction.

image

image


image

nima-karimi:
Thanks, @jorgemd24 

> It should be 'Flat rate - %1$s %2$s':

I fixed this in https://github.com/woocommerce/google-listings-and-ads/pull/1422/commits/720c0c00b2a4826c5363a821919419af66b0b1f3

> I also notice that when there is a country for example Italy with a flat rate: of $5 and free shipping with a minimum order is not syncing properly.

Does this happen when you choose the "Recommended" option or the manual option to sync the settings? Can you try checking out the branch for the https://github.com/woocommerce/google-listings-and-ads/pull/1455 PR and see if this works there? I've made some changes in that PR that will essentially overwrite a bunch of codes in here once it's merged.

jorgemd24:
Thanks @nima-karimi, 

>Does this happen when you choose the "Recommended" option or the manual option to sync the settings? 

Yeah, I chose the "Recommended" option.

>Can you try checking out the branch for the https://github.com/woocommerce/google-listings-and-ads/pull/1455 PR and see if this works there? 

I have tried this scenario:

Spain Flat Rate: $15
Italy: Flat Rate of $5 and free shipping minimum order of $10.

I got the following in the MC:

https://user-images.githubusercontent.com/2488994/166671873-ab7a39b1-640f-48eb-83e0-6304cc92331c.mov

- I see three rates and all of them appear with the name "All Products" and the Shipping rate name as empty, is this intended? 
- The rate for Spain is OK
- The rate for Italy shows two rates one of $5 and the other one of $0, with a minimum order of $10. So it looks good.

Let's wait until the other PR is reviewed so we can check that everything is working fine here.

nima-karimi:
> I see three rates and all of them appear with the name "All Products" and the Shipping rate name as empty, is this intended?

Yeah, I intentionally removed the rate name because it was optional and I didn't see any benefits to adding it.

nima-karimi:
@jorgemd24 👋 https://github.com/woocommerce/google-listings-and-ads/pull/1455 is now approved and ready to merge. I rather get this PR approved first before merging the other one because it's a rather large PR and it will further complicate the review of this one 😅 Let me know what you think. Would it be okay for you if I merge https://github.com/woocommerce/google-listings-and-ads/pull/1455 into this one so you can test the functionality?

jorgemd24 (approved):
Hey @nima-karimi!

> I rather get this PR approved first before merging the other one because it's a rather large PR and it will further complicate the review of this one 😅

Yeah, you are right, anyway I tested that is returning the correct shipping rates for different methods, and I can confirm that it is returning the most expensive rate and the property method has been removed.

The problems syncing to the MC are not relevant anymore as the method name is not used (before it was syncing as `false`) and using PR #1455 I see that is syncing the rates properly.

PR #1423, opened 2022-04-13
Apply migrations by the order of their applicable versions

nima-karimi:

To avoid unintended consequences, it's best if we sort the database migrations by their applicable versions before applying them to the DB. Currently, they are applied in the order that they're defined in the container.

1. Define two new DB migration classes. You can use the same structure as `MigrationVersion141::class` but instead of altering a table, simply add a new row to one.
2. Set the value in `get_applicable_version` for each of them to consecutive versions one after another. Both should be larger than the current one defined in `WC_GLA_VERSION` 
3. Add the migration classes to the container in `DBServiceProvider::class` right before the other migrations. For the purpose of this test, the migration with a lower applicable version should be placed after the one with a larger applicable version.
4. Change the version defined in `WC_GLA_VERSION` to something higher than the versions defined in the migrations.
5. Refresh your admin page and confirm that the migrations have been run in the order of their applicable versions (e.g. by checking the new rows are inserted in the expected order).

### Changelog entry

> Tweak - Run DB migrations in order of their applicable versions.

### DIFF
--- a/src/DB/Migration/Migrator.php
+++ b/src/DB/Migration/Migrator.php
@@ -28,6 +28,14 @@ class Migrator implements Service {
 	 */
 	public function __construct( array $migrations ) {
 		$this->migrations = $migrations;
+
+		// Sort migrations by version.
+		uasort(
+			$this->migrations,
+			function ( MigrationInterface $migration_a, MigrationInterface $migration_b ) {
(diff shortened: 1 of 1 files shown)


jorgemd24 (approved):
Thanks, LGTM 👍 

✅  I have tested changing the order of the migrations classes and I see that is ordered depending on the `get_applicable_version`.

PR #1265, opened 2022-02-22
Shipping rates UI changes: Better SelectControl label

ecgan:
### Changes proposed in this Pull Request:

- Part of a PR refactor from the huge PR https://github.com/woocommerce/google-listings-and-ads/pull/1218.

In this PR, I have simplified the way to specify label for the SelectControl. Previously, we had to manually craft a div and CSS to make the UI look nice. Now, we just need to make use of the `label` props. The impacted area in this PR is the EditRateModal in Edit Free Listings flow.

### Screenshots:

SelectControl label in EditRateModal in Edit Free Listings flow:

image

### Detailed test instructions:

1. Go to Edit Free Listings.
2. In a shipping rate row, click on the Edit button.
3. Verify that the UI looks like the above screenshot.

### Changelog entry

>

PR #1424, opened 2022-04-13
Refactor and simplify `EstimatedShippingRatesCard`

ecgan:

- This PR depends on PR https://github.com/woocommerce/google-listings-and-ads/pull/1409.
- This PR shares some code structure similarity / consistency with https://github.com/woocommerce/google-listings-and-ads/pull/1410.

In this PR, we do some simplification and refactor for codes within the folder `estimated-shipping-rates-card`:

- Rename AddRateModal and EditRateModal to AddRateFormModal and EditRateFormModal.
- Make the code structure more consistent and similar to codes in minimum order card.
- Remove unused / unneeded code, as a follow up to PR https://github.com/woocommerce/google-listings-and-ads/pull/1409 (shipping rates' countries are now always a subset of audience countries).
- Change the naming of `price` to `rate`, to make things consistent.
- Change `AggregatedShippingRate` typedef to `ShippingRateGroup`.
- Move `groupShippingRatesByMethodCurrencyRate` from the utils folder in the local component folder. `groupShippingRatesByMethodCurrencyRate` works with shipping rate group, and shipping rate groups all work within the the EstimatedShippingRatesCard component hierarchy.
- Better JSDoc and typedefs.

Things should still work as is before and after this PR. You can test the following in Setup MC or Edit Free Listings.

1. Delete all the shipping rates, so that we will see the initial `ShippingRateInputControl` with empty input box.
2. For the initial `ShippingRateInputControl` with empty input box, enter a rate value inline. It should work as expected.
3. For the initial `ShippingRateInputControl` with empty input box, click on edit button, enter countries and rate value in the modal, and click the update button. You should see `ShippingRateInputControl` with the rate value.
4. For the initial `ShippingRateInputControl` with empty input box, click on edit button, enter countries and rate value in the modal, and click the delete button. You should see initial `ShippingRateInputControl` with empty input box.
5. For a `ShippingRateInputControl`, you should be able to change the value inline.
6. For a `ShippingRateInputControl`, you should be able to click on the edit button and change the values in the modal. Click on the update button, and the shipping rates should be displayed correctly on the page.
7. For a `ShippingRateInputControl`, you should be able to click on the edit button, and click on the delete button in the modal. The shipping rates should be removed from the page.
8. If there are remaining audience countries with no shipping rate yet, you should see an "Add another rate" button. Click on it, and you should see the add rate modal with the remaining countries filled in. Enter a rate value and click on the add button. The shipping rate should show up on the page.
9. If all the countries already have a corresponding shipping rate, you should not see the "Add another rate" button.


js/src/components/shipping-rate-section/estimated-shipping-rates-card/add-rate-form-modal.js
```diff
@@ -4,27 +4,36 @@
...
+ */
+
 /**
  * Form to add a new rate for selected country(-ies).
  *
  * @param {Object} props
- * @param {Array<CountryCode>} props.countries A list of country codes to choose from.
- * @param {Function} props.onRequestClose
- * @param {function(AggregatedShippingRate): void} props.onSubmit Called with submitted value.
+ * @param {Array<CountryCode>} props.countryOptions Array of country codes options, to be used as options in AppCountrySelect.
```
eason9487: 💅  Probably the first **options** wording is not needed because they are not yet the data structures of option.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/edit-rate-form-modal.js
```diff
@@ -50,8 +54,8 @@ const EditRateModal = ( {
 			);
 		}
 
-		if ( values.price < 0 ) {
-			errors.price = __(
+		if ( values.rate < 0 ) {
+			errors.rate = __(
```
eason9487: Thanks for changing them to be consistent! 💯

js/src/components/shipping-rate-section/estimated-shipping-rates-card/typedefs.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @typedef { import(".~/data/actions").CountryCode } CountryCode
+ */
+
+/**
+ * @typedef {Object} ShippingRateGroup
+ * @property {Array<CountryCode>} countries Array of selected country codes.
+ * @property {string} method Shipping method, e.g. "flat_rate".
+ * @property {string} currency Currency.
+ * @property {number} rate Rate value.
```
eason9487: 💅  The type of `rate` could be null.
https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/estimated-shipping-rates-card.js#L157

js/src/components/shipping-rate-section/estimated-shipping-rates-card/estimated-shipping-rates-card.js
```diff
@@ -28,93 +34,196 @@ const defaultShippingRate = {
...
+	 *
+	 * Shipping rate group will be converted into shipping rates, and propagate up via `onChange`.
+	 *
+	 * @param {ShippingRateGroup} newGroup Shipping rate group.
+	 */
+	const handleAddSubmit = ( { countries, method, currency, rate } ) => {
+		const newShippingRates = countries.map( ( country ) => ( {
 			...defaultShippingRate,
 			country,
+			method,
```
eason9487: ❔ 
I checked the \<ShippingRateInputControl\>, \<AddRateFormModal\>, and \<EditRateFormModal\> and found they all don't change the `method` value, and its value is already set by `...defaultShippingRate`. Not sure why here needs to override it by the newGroup's value again?

js/src/components/shipping-rate-section/estimated-shipping-rates-card/estimated-shipping-rates-card.js
```diff
@@ -28,93 +34,196 @@ const defaultShippingRate = {
...
+			newGroup.countries.forEach( ( country ) => {
+				const existingIndex = newValue.findIndex(
+					( shippingRate ) => shippingRate.country === country
+				);
+				const oldShippingRate = newValue[ existingIndex ];
+				const newShippingRate = {
+					...defaultShippingRate,
+					...oldShippingRate,
+					country,
+					method: newGroup.method,
```
eason9487: ❔  it's similar to the above question of overriding the `method` value.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/groupShippingRatesByMethodCurrencyRate.test.js
```diff
@@ -1,7 +1,7 @@
 /**
  * Internal dependencies
  */
-import groupShippingRatesByMethodPriceCurrency from './groupShippingRatesByMethodPriceCurrency';
+import groupShippingRatesByMethodCurrencyRate from './groupShippingRatesByMethodCurrencyRate';
 
 describe( 'groupShippingRatesByMethodPriceCurrency', () => {
```
eason9487: 💅  The corresponding renaming is missed in the description.

eason9487 (commented):
Thanks for the adjustments. The relevant codes look nicer. Especially the unifying of the property name for `rate`. I searched the codebase and found there are some leftover `price` that could be updated:
- https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/edit-free-campaign/index.js#L54
- https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/groupShippingRatesByMethodCurrencyRate.test.js#L7

And the `price` in these two seems incorrect:
- https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/components/shipping-rate-section/minimum-order-card/add-minimum-order-modal.js#L31
- https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/components/shipping-rate-section/minimum-order-card/edit-minimum-order-modal.js#L39

---

❓ or 📜 

Another thought is about the \<AddRateFormModal\>, \<EditRateFormModal\>, and the naming of this PR changed.

### Deduplicate the code

Considering that most code of \<AddRateFormModal\> and \<EditRateFormModal\> is duplicated, it may be possible to further merge them into one. Thus, it is even possible to move the shipping rate modal out of the \<ShippingRateInputControl\> and make the \<EstimatedShippingRatesCard\> contain only one modal instance.

In this way, the callback processing of shipping rate modal could be shared to a greater degree and have better maintainability.

### Naming of components

For example, the new name `AddRateFormModal` conveys that it contains Form and Modal, while the naming of `ShippingRateInputControl` does not convey that it contains Form and Modal. But I don't think it's a good idea to have something like **ShippingRateInputControlFormModal**.

Moreover, the implementation of `ShippingRateInputControl` is a bit unintuitive in terms of how the combination of components. Almost every couple of weeks, I re-realized that the edit rate modal is wrapped in the `ShippingRateInputControl`.

Moving the `EditRateFormModal` out might lead to better refactoring results, and the naming of the components could be more consistent within the scope under the **shipping-rate-section** folder.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/add-rate-form-modal.js
ecgan: Edited in [`4119e51` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/4119e51a7364d392371ac5f25ce8cd6456072caa).

js/src/components/shipping-rate-section/estimated-shipping-rates-card/typedefs.js
ecgan: Updated in [`f332ef1` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/f332ef13cf18055fbe5d7186ead6991f8000127d). I've changed it to `undefined` instead of `null`, to make it consistent with minimum order side (where the minimum order's `threshold` can be `undefined`).

js/src/components/shipping-rate-section/estimated-shipping-rates-card/groupShippingRatesByMethodCurrencyRate.test.js
ecgan: Fixed in [`6bd1246` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/6bd124682c580f8fdbd4c7c125fe856c052cc2fe). 👍

js/src/components/shipping-rate-section/estimated-shipping-rates-card/estimated-shipping-rates-card.js
ecgan: That's because I'm treating the components like a "black box", without looking into its implementation. We pass down the `method` value via props and get the value back via event handlers. The components _might_ change the value, and we should use that. (We only know that the components don't change the `method` value because we look into its implementation - a "white box".)

Side note: We probably don't need to be too concerned on this, because the `method` property shall be removed from the API side (see https://github.com/woocommerce/google-listings-and-ads/pull/1422#issuecomment-1100022747) and we would remove it from our UI code here soon.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/estimated-shipping-rates-card.js
ecgan: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1424#discussion_r852127304.

ecgan:
> I searched the codebase and found there are some leftover `price` that could be updated: [...]

Thanks! I was searching for "price" within the "estimated-shipping-rates-card" folder only. 😄 

> * https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/edit-free-campaign/index.js#L54

This function has already been removed in my previous PR (see https://github.com/woocommerce/google-listings-and-ads/pull/1420/files#diff-5297a79a23cac9dd2adeb1b6289484c6a7ebe09f096a488d19fd7ac68213fd1dL54), it is no longer there in `develop` branch.

> * https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/groupShippingRatesByMethodCurrencyRate.test.js#L7

This has been updated in [`6bd1246` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/6bd124682c580f8fdbd4c7c125fe856c052cc2fe), as per https://github.com/woocommerce/google-listings-and-ads/pull/1424#discussion_r852115456.

> And the `price` in these two seems incorrect:
> 
> * https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/components/shipping-rate-section/minimum-order-card/add-minimum-order-modal.js#L31
> * https://github.com/woocommerce/google-listings-and-ads/blob/792961ad21545995fe6cd1620a2fb65df55c811b/js/src/components/shipping-rate-section/minimum-order-card/edit-minimum-order-modal.js#L39

Yeah, I think that was copy-paste-fail 🤦 They should be `errors.threshold` instead. I have fixed them in the minimum order modal PR in [`c4dccb2` (#1410)](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/c4dccb28ddd49c184d22ba107c3073bc13790d07).

---

> Considering that most code of <AddRateFormModal> and <EditRateFormModal> is duplicated, it may be possible to further merge them into one. Thus, it is even possible to move the shipping rate modal out of the <ShippingRateInputControl> and make the <EstimatedShippingRatesCard> contain only one modal instance.

Hmm yeah, I'll look into this.

> the naming of `ShippingRateInputControl` does not convey that it contains Form and Modal. But I don't think it's a good idea to have something like **ShippingRateInputControlFormModal**.
> ...
> the implementation of `ShippingRateInputControl` is a bit unintuitive in terms of how the combination of components.

Yeah, agree. I thought about the same thing too. The "problem" is that we have the "edit" button being part of the input control's label content. I'll look into how to make it better.

> Almost every couple of weeks, I re-realized that the edit rate modal is wrapped in the `ShippingRateInputControl`.

That is an interesting pain point to hear. Thanks for sharing that.

> Moving the `EditRateFormModal` out might lead to better refactoring results, and the naming of the components could be more consistent within the scope under the **shipping-rate-section** folder.

Yeah. I'll look into that. 👍

ecgan:
@eason9487 , I have made a few more changes based on your comments. See summaries below. Have a look and let me know what you think.

> ### Deduplicate the code

I came up with two ways.

In my first commit [`e3fc6a9` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/e3fc6a9173dd4e3cb4f6336b970aaa5bc6979a47), I came up with a `RateFormModal` that contains `Form` and `Modal` component in it. In a way, it feels familiar, simpler to read, well integrated (at least for me); however, the props `additionalButtons` and `submitButtonChildren` feel kinda hacky for me.

In my second commit [`fe564f6` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/fe564f6a55a7f32154b46a4953e16885f5eb0c80), I changed the `RateFormModal` into two separate `RateForm` and `RateModal` components. It is more modular without the hacky props.

> Moving the `EditRateFormModal` out might lead to better refactoring results, and the naming of the components could be more consistent within the scope under the **shipping-rate-section** folder.

In [`9b7ac66` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/9b7ac6667ae182a23557e3f343c3bd75dcb9a370), the `EditRateFormModal` is now imported and used inside `EstimatedShippingRatesCard`. One benefit from this refactor is that `ShippingRateInputControl` now only requires an `editRateFormModal` prop, without needing to have `countryOptions` and `onDelete` props.

eason9487:
> I have made a few more changes based on your comments. See summaries below. Have a look and let me know what you think. [...]

Thanks for the changes and the extraction of reusable code! 🙏  Some parts of code look better now. However, the ways of extracting \<RateForm\> and \<RateModal\> components bring much higher coupling relations between them, \<AddRateFormModal\>, \<EditRateFormModal\>, and \<ShippingRateInputControl\> than before.

In my opinion, now the \<RateForm\> always has to compose \<RateModal\> in a specific way, so \<AddRateFormModal\> and \<EditRateFormModal\> need to know how to compose form and modal under their high coupling rules. In addition, \<ShippingRateInputControl\> still owns the visible state of the model, and it has to be a modal, even though the modal has been changed to be passed in via prop. Because they are coupling on the \<AppButtonModalTrigger\>.

Therefore, I think it doesn't make them more modular. Comparing the changes, I'd prefer the revision 6bd124682c580f8fdbd4c7c125fe856c052cc2fe because the high coupling parts were at least encapsulated together inside the modals.

---

My thought based on the implementation of the two modals was relatively simple and highly similar, so it only requires adding two `if` conditions to merge them and to further simplify some code on the consumer side. I drafted a demo revision for the concept. The code diff can be viewed in [this link](https://github.com/woocommerce/google-listings-and-ads/compare/6bd124682c580f8fdbd4c7c125fe856c052cc2fe...aa2be71d401d8773db6a03e0563cd00e06315808). The demo also includes the extraction of edit modal, the shared instance of merged modal, and the simplification of some reusable code.

💡  Please note that the naming and related docs were not modified properly and correspondingly due to a demo purpose.

ecgan:
@eason9487 , thanks for the comments, and the [linked code diff](https://github.com/woocommerce/google-listings-and-ads/compare/6bd124682c580f8fdbd4c7c125fe856c052cc2fe...aa2be71d401d8773db6a03e0563cd00e06315808)! 🙏  That is really helpful to understand your idea.

> the <RateForm> always has to compose <RateModal> in a specific way, so <AddRateFormModal> and <EditRateFormModal> need to know how to compose form and modal under their high coupling rules

I made a change in [`3941fba` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/3941fbacedc63b018230ff9e2401ecf0fb917f0f), so that `RateForm` and `RateModal` can be used independently without being aware of each other. I think that would address your comment above, though eventually I have removed the components in subsequent changes.

---

I've added a few more changes in this PR. Now, we have a `RateFormModal` that encapsulates the `Form` and `Modal`, with `renderButtons` prop. `renderButtons` will be passed `formProps` that can be used in rendering the buttons in modal footer. This simplifies some repetitive code between `AddRateFormModal` and `EditRateFormModal`. `RateForm` and `RateModal` has been removed.

IMO, I like this `RateFormModal` better than the previous ways I came up in https://github.com/woocommerce/google-listings-and-ads/pull/1424#issuecomment-1101638027.

---

> My thought based on the implementation of the two modals was relatively simple and highly similar, so it only requires adding two `if` conditions to merge them and to further simplify some code on the consumer side. I drafted a demo revision for the concept. The code diff can be viewed in [this link](https://github.com/woocommerce/google-listings-and-ads/compare/6bd124682c580f8fdbd4c7c125fe856c052cc2fe...aa2be71d401d8773db6a03e0563cd00e06315808). The demo also includes the extraction of edit modal, the shared instance of merged modal, and the simplification of some reusable code.
>
> 💡 Please note that the naming and related docs were not modified properly and correspondingly due to a demo purpose.

I looked into your code, and here's my subjective opinions from my past experience. (I have actually used such code in the past.) 💅  😅  

1\. My first opinion is about the having one modal `EditRateFormModal` for both add and edit operations, differentiated by an `isEdit` prop (or it can be something like a `mode` prop). We would need to have the `if` checks on the prop, and do conditional operation and rendering, which IMO is less ideal.

On the consumption side, when we want an edit modal, we pass `isEdit: true` and `onDelete` props. When we want an add modal, it does not need the `isEdit` and `onDelete` props. IMO, the difference like this can bring confusion and friction at times, and might cause errors if we are not careful (e.g. accidentally use add modal when we want edit modal).

Having an `isEdit` prop with just true/false to control edit/add is probably not a big deal. However, my bigger concern is about the pattern it brings. Imagine if we have a `mode` prop with values "add" / "edit", and sometime in the future someone wants to add "view" or "duplicate" mode, in the name of reusability, it would be tempting for one to add those modes and add more `if` checks into the component, making the component more and more bloated. 

With the latest code we have here, with `AddRateFormModal` and `EditRateFormModal` using a base `RateFormModal`, the code is more straightforward. There are no `if` conditional operation and rendering. `RateFormModal` is used by both modals, and the difference between both modals is in the buttons with the `renderButtons` props. 

On the consumption side, it is also easy to see when and where we are using `AddRateFormModal` and `EditRateFormModal`, it is clear with less confusion and less possible mistakes. We would know `EditRateFormModal` with `onDelete` prop would support delete operation, while `AddRateFormModal` does not.

If someone wants to add "view" or other operations, they can come up with a `ViewRateFormModal` component, and either reuse the base `RateFormModal` or have a totally different implementation without reusing `RateFormModal`. `RateFormModal` would stay lean. On the consumption side, it is easy to understand when we see the usage of `ViewRateFormModal` too.

2\. My second opinion is about the usage of `useState`, setting it with modalProps via `onEditClick`. I think I understand your idea of having `onEditClick`, so that `ShippingRateInputControl` does not have coupling / does not need to know about `EditRateFormModal` (I thought about this too). However, the thing is, within `ShippingRateInputControl` label, we do have an Edit button, and the whole purpose of the edit button is to bring up the `EditRateFormModal` anyway. In other words, I don't see a big difference or advantage of having `onEditClick` over `editButtonModal` props, when `onEditClick` is ultimately used to just display the modal anyway. In fact, by having `editButtonModal` props, it makes it clear that `ShippingRateInputControl` depends on a modal (`EditRateFormModal`), and we have the the explanation for it in the component JSDoc comment.

IMO, to have the least coupling, the edit button should be moved out from `ShippingRateInputControl`, but I think the effort would be too much at this moment.

With `editButtonModal` prop, we do not need to have `useState`, and it make things slightly more straightforward and one less thing to think or worry about.

---

In the end, I think I am happy with the current code. Your comments did help a lot in it. Do you think they are acceptable? Are we okay to "agree to disagree"? 😂

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validate.test.js
```diff
@@ -0,0 +1,43 @@
...
+ * Internal dependencies
+ */
+import validate from './validate';
+
+const validValues = {
+	countries: [ 'US' ],
+	rate: 0,
+};
+
+describe( 'validate', () => {
```
eason9487, line 11: 💅  or 🚧 
Although the #1435 issue isn't introduced by this PR, perhaps the missed invalid value check of shipping rates could be solved here first, given its test cases to be created in this PR.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/edit-rate-form-modal.js
```diff
@@ -0,0 +1,80 @@
...
+import { noop } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import RateFormModal from './rate-form-modal.js';
+
+/**
+ * @typedef { import(".~/data/actions").CountryCode } CountryCode
+ * @typedef { import("./typedefs.js").ShippingRateGroup } ShippingRateGroup
```
eason9487: 💅 
The `./typedefs.js` path is incorrect.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/rate-form-modal.js
```diff
@@ -0,0 +1,86 @@
...
+import validate from './validate.js';
+import AppModal from '.~/components/app-modal';
+import AppInputPriceControl from '.~/components/app-input-price-control/index.js';
+import VerticalGapLayout from '.~/components/vertical-gap-layout';
+import AppCountrySelect from '.~/components/app-country-select';
+
+/**
+ * @typedef { import("@wordpress/components").Button } Button
+ * @typedef { import(".~/data/actions").CountryCode } CountryCode
+ * @typedef { import("./typedefs.js").ShippingRateGroup } ShippingRateGroup
```
eason9487: 💅 
The `./typedefs.js` path is incorrect.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validate.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * @typedef { import("./typedefs.js").ShippingRateGroup } ShippingRateGroup
```
eason9487: 💅 
The `./typedefs.js` path is incorrect.

eason9487:
Thanks for addressing some coupling concerns!

> 1. My first opinion is about the having one modal `EditRateFormModal` for both add and edit operations, differentiated by an `isEdit` prop (or it can be something like a `mode` prop). We would need to have the `if` checks on the prop, and do conditional operation and rendering, which IMO is less ideal.
> 
> On the consumption side, when we want an edit modal, we pass `isEdit: true` and `onDelete` props. When we want an add modal, it does not need the `isEdit` and `onDelete` props. IMO, the difference like this can bring confusion and friction at times, and might cause errors if we are not careful (e.g. accidentally use add modal when we want edit modal).

Sorry, I couldn't imagine how a `isEdit` prop or a `mode` prop brings confusion and friction. The props of React compnents or the attributes of DOM elements extensively to be adapted highly for similar or reusable components/elements, especially that controling prop can be more described via JSDoc, and devs don't even need to open the file if their code editor could show the JSDoc by a popup tip. And how much confusion and friction it brings when opening the file and seeing the simple `if ( isEdit ) { buttons.unshift( { <button onClick={ onDelete } [...] /> } ) }`?

Or, do the following components bring confusion and fraction?
```js
const view = (
  <>
    <input type="radio" />
    <input type="checkbox" />
    <Button isPrimary isDestructive />
    <SelectControl multiple inlineTags />
  </>
);
```

And unfortunately, I'm also unable to measure how it raises the degree of making devs being not careful. Is there an obvious difference to be not careful of these patterns?
```js
const view = (
  <>
    <EditRateFormModal />
    <RateFormModal isEdit />
    <RateFormModal mode="edit" />
  </>
);
```

> Having an `isEdit` prop with just true/false to control edit/add is probably not a big deal. However, my bigger concern is about the pattern it brings. Imagine if we have a `mode` prop with values "add" / "edit", and sometime in the future someone wants to add "view" or "duplicate" mode, in the name of reusability, it would be tempting for one to add those modes and add more `if` checks into the component, making the component more and more bloated.

In a development process with code review, I don't think it's something to worry about, like the situation we're discussing. If it's not suitable to expand the `mode`, just let the dev know there're other component designs can be considered. Or, if the new condition and content that need to be added are simple, then this little bit of "bloat" is not a concern and would even be a good example of code reuse. On this premise, adding a ViewRateFormModal instead adds more duplicated code, and isn't this a kind of bloat?

I would like to go back to what has actually happened under the shipping data form scope, what changes will almost happen, and what requirements have a higher chance of happening in the future.

#### Happened:
- It took more effort to do the code review because the duplicated codes were checked repeatedly.
   - Such as the incorrect path of **typedefs.js**. I luckily found that because I tried to use `diff` command to compare AddRateFormModal and EditRateFormModal. But I don't usually need to do this comparison.
   - When a file was moved and the moved file content was changed as well in PR. Sometimes these changes are not identified by the git history, and then it takes extra effort to manually diff to see what has actually changed. Therefore, this also requires more effort for more almost duplicated components.

#### Will almost happen:
- Since TreeSelectControl has more obvious UI/UX problems when used with Modal, it's possible that the Modal would be removed, and TreeSelectControl would be expanded in other ways in the future. So there is a high chance that the three *RateFormModal would have to be adjusted. That's to say, needs more effort in the current implementation.
- It needs to prevent the modal from being close if the dropdown of `TreeSelectControl` is opening too when clicking on the overlay or pressing the ESC key.
   - Fortunately, the needed changes can be limited inside the various shipping modals.
   - But if this processing must be controlled in `EstimatedShippingRatesCard` layer, then in the code scope of shipping rate, the original need to change was two modal, now it becomes three, and even one more layer of props drilling and three times repeated JSDoc to be added.

#### Higher chance of happening in the future:
- If the future requirement wants to enable users to choose currencies, it should inevitably need to change all *RateFormModal files as well.
- For forms, a common convenience processing is to allow users to submit data by pressing the ENTER key when focusing on a `<input>`. However, all shipping modals currently don't support this. If this requirement must be achieved without using the workaround aspect and without having the below warning, then all related *RateFormModal files need to be modified as well.
   image

On these more specific and possible considerations, I really can't agree that the current approach is relatively good than the option of a merged modal.

> With the latest code we have here, with `AddRateFormModal` and `EditRateFormModal` using a base `RateFormModal`, the code is more straightforward. There are no `if` conditional operation and rendering. `RateFormModal` is used by both modals, and the difference between both modals is in the buttons with the `renderButtons` props.
> 
> On the consumption side, it is also easy to see when and where we are using `AddRateFormModal` and `EditRateFormModal`, it is clear with less confusion and less possible mistakes. We would know `EditRateFormModal` with `onDelete` prop would support delete operation, while `AddRateFormModal` does not.
> 

Even though I couldn't imagine how easily made the mistake of missing `onDelete`. With the current implementation of `EditRateFormModal`, missing `onDelete` is nothing different than the merged modal: They both don't reflect the deletion onto UI, there's no error occurred and no test case for checking it. But it should be still easily noticed during development. I believe these can be improved by adding test cases and removing the unnecessary `noop` fallback for `onDelete` to avoid silent errors.

> If someone wants to add "view" or other operations, they can come up with a `ViewRateFormModal` component, and either reuse the base `RateFormModal` or have a totally different implementation without reusing `RateFormModal`. `RateFormModal` would stay lean. On the consumption side, it is easy to understand when we see the usage of `ViewRateFormModal` too.

The merged modal will not block the adding of ViewRateFormModal, and the `buttons` can still be changed to a prop as the **Render Props** pattern if needed.

---

> 2. My second opinion is about the usage of `useState`, setting it with modalProps via `onEditClick`. I think I understand your idea of having `onEditClick`, so that `ShippingRateInputControl` does not have coupling / does not need to know about `EditRateFormModal` (I thought about this too).

The use way of `modalProps` is not the key point. My primary purpose was to demonstrate the concept of how merging them in a simple way and then sharing a single instance of modal. Because it can simplify much more code and remove most duplicated codes. Although there are a few ways to make it clearer than `modalProps`, I still don't aim to focus on this part.

Other reasons are:
- It's much clear how the opening state of a modal is controlled.
- Comparing the visual results on the webpage with the component composition on the code level, it's relatively more intuitive to understand there's a modal that will pop up onto the whole page.  Instead of being more implicitly inside a button of certain input.
- Reduce unnecessary rendering calculations even though no visible difference. But it's a better practice if it can be done with a little effort.

> However, the thing is, within `ShippingRateInputControl` label, we do have an Edit button, and the whole purpose of the edit button is to bring up the `EditRateFormModal` anyway. In other words, I don't see a big difference or advantage of having `onEditClick` over `editButtonModal` props, when `onEditClick` is ultimately used to just display the modal anyway. In fact, by having `editButtonModal` props, it makes it clear that `ShippingRateInputControl` depends on a modal (`EditRateFormModal`), and we have the the explanation for it in the component JSDoc comment.
>
> With `editButtonModal` prop, we do not need to have `useState`, and it make things slightly more straightforward and one less thing to think or worry about.

`ShippingRateInputControl` is indeed better now, allowing me to explain more specifically why they still have bad coupling.

Imagine viewing these components for the first time:
1. First open EditRateFormModal, ok, it requires **onRequestClose** and passes to RateFormModal
1. Then open RateFormModal, ok, it also requires **onRequestClose** and passes to AppModal. Got it, that makes sense.
1. Then check EstimatedShippingRatesCard and notice that EditRateFormModal is passed into ShippingRateInputControl as **editButtonModal** prop, but it's a bit strange: Where is **onRequestClose**? It should be a required prop, right? Why it hasn't done that?
1. Let's keep digging, look at ShippingRateInputControl, oh, the ShippingRateInputControl passes **editButtonModal** into a deeper component AppButtonModalTrigger as **modal** prop.
1. Finally, open AppButtonModalTrigger. Oh okay, that's why. The **onRequestClose** will be injected from here.

That's the reason why I said "they are coupling on the \<AppButtonModalTrigger\>."

---

> Do you think they are acceptable?

I appreciate your explanations and extra effort in adjusting the codes due to my nitpicking. The current revision has reduced some duplication and significantly improved reusability. 🙏

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/edit-rate-form-modal.js
ecgan: Fixed in [`199be5e` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/199be5ed750f729536d252ff207b4184f744db14).

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/rate-form-modal.js
ecgan: Fixed in [`199be5e` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/199be5ed750f729536d252ff207b4184f744db14).

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validate.js
ecgan: Fixed in [`199be5e` (#1424)](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/199be5ed750f729536d252ff207b4184f744db14).

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validate.test.js
ecgan: Let's fix it in another PR. If we were to do it here, this PR is gonna be "refactor + bug fixing". We already have more than enough to test and talk about in this PR. 😂  🙏

ecgan:
@eason9487 ,

> Sorry, I couldn't imagine how a `isEdit` prop or a `mode` prop brings confusion and friction. ...

For example, let's say we see a line like this: `<FooModal mode="add">`.

The questions that I, as a consumer, would have in my mind are:

- What are the possible values of mode? Is there a default mode? What if I didn't specify a value for it?
- What do those modes do within the component? (e.g. Different rendering? Different operations like API calls?)
- What are the different props that I need to provide with different modes? (e.g. `onDelete` props is required for `mode='edit'`)

Note that the impact / effect of the above questions would become bigger when there are more modes, which is likely to grow over time.

When we have a specific dedicated component like `<AddFooModal>`, it removes all these questions about mode, allowing the consumer to focus on the Add operation.

> especially that controling prop can be more described via JSDoc, and devs don't even need to open the file if their code editor could show the JSDoc by a popup tip.

Referring to my above example `<FooModal mode="add">`, we would need to use JSDoc to explain and answer the questions I have above about modes, e.g. "If mode is edit, there will be Update and Cancel buttons. If mode is add, there will be...".

The JSDoc comment would get longer when there are more modes. 

When we have a specific dedicated component like `<AddFooModal>`, JSDoc will just need to focus on the Add operation, without needing to explain modes. JSDoc would be shorter and easier to understand.

> how much confusion and friction it brings when opening the file and seeing the simple `if ( isEdit ) { buttons.unshift( { <button onClick={ onDelete } [...] /> } ) }`?

Option 1: `<FooModal mode='edit'>` with `if ( isEdit ) { buttons.unshift( { <button onClick={ onDelete } [...] /> } ) }` (and possibly more) in it.
Option 2: `<EditFooModal>` with no `if` in it.

I would choose option 2.

Also, while we are on this topic, I would like to also add that the confusion and friction might get amplified in non-ideal situations. For example, imagine there is a highly critical production issue affecting edit modal. It would be easier and have lower risk to work with Option 2 over Option 1.

> do the following components bring confusion and fraction?
> 
> ```js
> const view = (
>   <>
>     <input type="radio" />
>     <input type="checkbox" />
>     <Button isPrimary isDestructive />
>     <SelectControl multiple inlineTags />
>   </>
> );
> ```

The components above are in the generic highly reusable components domain and they would have different design criteria to cater for the masses.

On the other hand, the modal components that we are talking about are in the highly specific business application operation domain, which is different from the above.

(I can share more about this but I think we are getting off topic. 😅  🙏  )

> Is there an obvious difference to be not careful of these patterns?
> 
> ```js
> const view = (
>   <>
>     <EditRateFormModal />
>     <RateFormModal isEdit />
>     <RateFormModal mode="edit" />
>   </>
> );
> ```

If we try to push them a little further like the following:

```js
const View = () => {
	return (
		<>
			<EditFooModal
				prop1={ prop1 }
				prop2={ prop2 }
				prop3={ prop3 }
				prop4={ prop4 }
			/>
			<FooModal
				prop1={ prop1 }
				prop2={ prop2 }
				prop3={ prop3 }
				prop4={ prop4 }
			/>
			<FooModal
				prop1={ prop1 }
				prop2={ prop2 }
				prop3={ prop3 }
				mode="edit"
				prop4={ prop4 }
			/>
		</>
	);
};
```

For the first modal, it is immediately clear that it is for edit operation as the component name implies. For the second and third modal, we have to scan for the mode, and if there isn't one, recall what it means (what is the default mode).

> In a development process with code review, I don't think it's something to worry about, like the situation we're discussing. If it's not suitable to expand the `mode`, just let the dev know there're other component designs can be considered. Or, if the new condition and content that need to be added are simple, then this little bit of "bloat" is not a concern and would even be a good example of code reuse. 

IMO, that places the burden on the code reviewers. Developers and reviewers can make mistakes, especially in non-ideal situation (e.g. highly critical production issue with time pressure). And the developers and reviewers in the future may be other people (not you and I) who are not aware of our discussion here, and would be adding more and more modes into the component, because people tend to follow existing patterns, with little incentive to change existing patterns.

> would even be a good example of code reuse. 

I am actually advocating for code reuse via composition. 😄 

> On this premise, adding a ViewRateFormModal instead **_adds more duplicated code_**, and isn't this a kind of bloat?

I'm curious about what you mean by "adds more duplicated code". 🤔  I see it as code reuse via composition, by reusing `RateFormModal`.

If you are talking about having `AddRateFormModal`, `EditRateFormModal`, `ViewRateFormModal` etc means having more duplicated code - because we need to create these components passing the props to the underlying `RateFormModal` with the difference being just the buttons - I don't think they are duplicated code and a kind of bloat. In fact, I see them as natural and a good design.

The way I come up with these modals is to relate them to the intended operations, i.e.:

- When users click on an Add button, we display an AddFooModal.
- When users click on an Edit button, we display an EditFooModal.
- ... on and on...

With this approach, we are "compartmentalizing" the UI to the operations. When something changes (e.g. business logic or API for the edit operation), we only need to focus on that compartment (EditFooModal). If we have something like `<FooModal mode='edit'>`, my concern is it will get more and more difficult over time.

Also, when the users are performing the clicking actions on the buttons above, it means that they have already made the `if-else` decision in their brain and selected a mode (e.g. "I click on Edit button" because "I want edit mode"), so we can just display the EditFooModal for them. IMO, in a way, having `<FooModal mode='edit'>` with mode checks in the component feels like we are bringing avoidable complexity (even though it is small) to ourselves when we could have capitalize on the user's decision and action (i.e. they have already made the if-else decision on our behalf, we don't need the if-else mode checks in our code). 

Also, in case it helps, the way I think about the modals is not "one modal shared by three operations", but "three operations with their own modals, which happened to share some similar UI". This allows the modals to evolve based on operations.

> * Such as the incorrect path of **typedefs.js**. I luckily found that because I tried to use `diff` command to compare AddRateFormModal and EditRateFormModal. But I don't usually need to do this comparison.

Sorry about that, I was relying on the code editor to fix the path and I didn't notice that. 🤦 

> * When a file was moved and the moved file content was changed as well in PR. Sometimes these changes are not identified by the git history, and then it takes extra effort to manually diff to see what has actually changed. Therefore, this also requires more effort for more almost duplicated components.

Another reason why I prefer `<EditFooModal>` over `<FooModal mode='edit'>` is to have better tracking in git history too. e.g. If I want to find out what causes an edit operation issue in an edit modal, it would be easier to look at git history for `<EditFooModal>`.

> * Since TreeSelectControl has more obvious UI/UX problems when used with Modal, it's possible that the Modal would be removed, and TreeSelectControl would be expanded in other ways in the future. So there is a high chance that _the three *RateFormModal would have to be adjusted_. That's to say, needs more effort in the current implementation.

> If the future requirement wants to enable users to choose currencies, it should inevitably need to change all *RateFormModal files as well.

> * For forms, a common convenience processing is to allow users to submit data by pressing the ENTER key when focusing on a `<input>`. However, all shipping modals currently don't support this. If this requirement must be achieved without using the workaround aspect and without having the below warning, then all related *RateFormModal files need to be modified as well.

🤔  At this moment in this PR, to change the control, we just need to make the change in the base `RateFormModal`, one place only, no?

> * But if this processing must be controlled in `EstimatedShippingRatesCard` layer, then in the code scope of shipping rate, the original need to change was two modal, now it becomes three, and even one more layer of props drilling and three times repeated JSDoc to be added.

My thought is that we just need to make the change in the base `RateFormModal`. I don't think I can't comment too much on it now because I haven't looked that deep into that project with TreeSelectControl yet. 😅 

> > On the consumption side, it is also easy to see when and where we are using `AddRateFormModal` and `EditRateFormModal`, it is clear with less confusion and less possible mistakes. We would know `EditRateFormModal` with `onDelete` prop would support delete operation, while `AddRateFormModal` does not.
> 
> Even though I couldn't imagine how easily made the mistake of missing `onDelete`. ...

What I was suggesting is that if we were to have `<FooModal mode='edit'>`, we would need to refer to the `mode` documentation and find out that it requires `onDelete` props. If there are more modes and more props combinations, it would be more difficult.

> But it should be still easily noticed during development. I believe these can be improved by adding test cases and removing the unnecessary `noop` fallback for `onDelete` to avoid silent errors.

By having `<EditFooModal>`, it can be easily noticed early in the development, with simpler code and JSDoc than `<FooModal mode='edit'>`.

> The merged modal will not block the adding of ViewRateFormModal, and the `buttons` can still be changed to a prop as the **Render Props** pattern if needed.

You are right, but the thing is, most often than not, people will just follow the existing pattern (in the name of reusability and maintainability) and add more `if`s into the shared modal to get their job done, and they have little interest or incentive to make the changes (because it is additional changes that are not their main goal). 

---

> The use way of `modalProps` is not the key point. My primary purpose was to demonstrate the concept of how merging them in a simple way and then sharing a single instance of modal.

Ahh okay, gotcha. 🙂 🙏 

> `ShippingRateInputControl` is indeed better now, allowing me to explain more specifically why they still have bad coupling.
> 
> Imagine viewing these components for the first time: [...]
> 
> That's the reason why I said "they are coupling on the `<AppButtonModalTrigger>`."

This is really helpful, thank you for the detailed explanation with bullet points! 😃 👍 

I think this is actually a specific issue with `AppButtonModalTrigger` component, and the issue with that component kinda leaked over here affecting shipping rates in this PR. I have created issue https://github.com/woocommerce/google-listings-and-ads/issues/1439 with some ideas in it. We can handle it in separate PR.

eason9487:
> For example, let's say we see a line like this: `<FooModal mode="add">`.
> 
> The questions that I, as a consumer, would have in my mind are:
> 
> * What are the possible values of mode? Is there a default mode? What if I didn't specify a value for it?
> * What do those modes do within the component? (e.g. Different rendering? Different operations like API calls?)
> * What are the different props that I need to provide with different modes? (e.g. `onDelete` props is required for `mode='edit'`)

> When we have a specific dedicated component like `<AddFooModal>`, it removes all these questions about mode, allowing the consumer to focus on the Add operation.

I don't see a specifically **dedicated** component like `ShippingRateInputControlLabelText` that removes all these questions. When a dev first time seeing a line like `<ShippingRateInputControlLabelText countries={ countries } />`, how suppose they can have no the same questions without opening the [implementation details](https://github.com/woocommerce/google-listings-and-ads/blob/9e12a8bc60681cd7113ff9a6157fc83edb41528c/js/src/components/shipping-rate-section/estimated-shipping-rates-card/shipping-rate-input-control-label-text.js)?

- What are the possible values of countries? Is there a default value for countries? What if I didn't specify a value for it?
- What do those countries do within the component? (e.g., Different rendering? Different operations like API calls?)
- What are the different props that I need to provide in different countries? (e.g., how to know does it provide a prop for controlling the `firstN` prop of `CountryNamesPlusMore`?)

> Note that the impact / effect of the above questions would become bigger when there are more modes, which is likely to grow over time.

If these questions are the problems that cause confusion and friction, then the reason is still confusing to me. No matter how simple a component is, before getting familiar with any APIs of that component, this should be something that needs to be done frequently when developing: reading its API docs, examples, or practical use cases, and looking at implementation details if want to know more details.

The topic that I had tried to focus on has always been based on the fact that **when a component has only simple logic**, doing so does not bring much of a "confusion" problem. If it's going to become more complex as the props stack up, extracting/separating the components is naturally a higher priority. I don't quite understand why, with an actually simple component, one needs to worry about problems that haven't happened yet, and won't be much of a problem when they do. It doesn't have to be in the same pattern all the time.

> The JSDoc comment would get longer when there are more modes.

As far as I know, I don't know if this will happen. I'd like to know more about how many different variances of shipping modals will be added in the future if you know what will it happen. Otherwise, given the infinite possibilities, it would be very easy to give an example to any component that says it's improperly designed.

> On the other hand, the modal components that we are talking about are in the highly specific business application operation domain, which is different from the above.

It's exactly because these modals are focused on serving shipping data only, so it's not necessary to consider more general uses, which makes it more suitable to internalize simple logic into the component to improve the degree of reuse and remove duplicated codes.

> If we try to push them a little further like the following:
> [...]
> For the first modal, it is immediately clear that it is for edit operation as the component name implies. For the second and third modal, we have to scan for the mode, and if there isn't one, recall what it means (what is the default mode).

I couldn't understand why so much effort was needed to scan the type attribute in the following:
```html
<input attr1="attr1" attr2="attr2" type="password" attr3="attr4" />
```
On the other hand, if a developer is using `EditRateFormModal` for the first time, how do they know they have to set `onDelete` onto it when they only see the component name? It's not immediately clear that the edit operations include the deletion operation, as the component name doesn't imply this. Otherwise, probably the component should be named `EditDeleteRateFormModal`, right? But, what about the countries? The countries are editable as well. Then the next change would be naming it to something like `EditDeleteCountriesAndRateFormModal`?

> IMO, that places the burden on the code reviewers. Developers and reviewers can make mistakes, especially in non-ideal situation (e.g. highly critical production issue with time pressure). And the developers and reviewers in the future may be other people (not you and I) who are not aware of our discussion here, and would be adding more and more modes into the component, because people tend to follow existing patterns, with little incentive to change existing patterns.

I can't predict which situation those future developers will tend to be in and feel less burdened:
1. More dedicated for each shipping modal as possible for two modes:
   - RateFormModal
   - AddRateFormModal
   - EditRateFormModal
   - TimeFormModal
   - AddTimeFormModal
   - EditTimeFormModal
   - MinimumOrderFormModal
   - AddMinimumOrderFormModal
   - EditMinimumOrderFormModal
2. Three dedicated shipping modals for two modes:
   - RateFormModal
   - TimeFormModal
   - MinimumOrderFormModal
3. One dedicated shipping modal for two modes:
   - ShippingFormModal

And if there will be more "modes" to be added, like an example that you have mentioned, the "view" or "duplicate", then it will becomes:
1. More dedicated for each shipping modal as possible for four modes:
   - RateFormModal
   - AddRateFormModal
   - EditRateFormModal
   - ViewRateFormModal
   - DuplicateRateFormModal
   - TimeFormModal
   - AddTimeFormModal
   - EditTimeFormModal
   - ViewTimeFormModal
   - DuplicateTimeFormModal
   - MinimumOrderFormModal
   - AddMinimumOrderFormModal
   - EditMinimumOrderFormModal
   - ViewMinimumOrderFormModal
   - DuplicateMinimumOrderFormModal
2. Three dedicated shipping modals for four modes:
   - RateFormModal
   - TimeFormModal
   - MinimumOrderFormModal
3. One dedicated shipping modal for four modes:
   - ShippingFormModal

It's impossible to tell which one is better based on infinite possibilities. If it only needs additional simple `if`, this makes it even more impossible for me to choose the first pattern. If the new modals are much more different or completed, then it depends on what exactly the requirements are.

---

> If you are talking about having `AddRateFormModal`, `EditRateFormModal`, `ViewRateFormModal` etc means having more duplicated code - because we need to create these components passing the props to the underlying `RateFormModal` with the difference being just the buttons - I don't think they are duplicated code and a kind of bloat. In fact, I see them as natural and a good design.

Let's quantify it. After adjusting some minor naming that could be the same:

### AddRateFormModal
```js
/**
 * External dependencies
 */
import { Button } from '@wordpress/components';
import { noop } from 'lodash';

/**
 * Internal dependencies
 */
import RateFormModal from './rate-form-modal.js';

/**
 * @typedef { import(".~/data/actions").CountryCode } CountryCode
 * @typedef { import("../typedefs.js").ShippingRateGroup } ShippingRateGroup
 */

/**
 * Form to add a new rate for selected country(-ies).
 *
 * @param {Object} props
 * @param {Array<CountryCode>} props.countryOptions Array of country codes, to be used as options in AppCountrySelect.
 * @param {ShippingRateGroup} props.initialValues Initial values for the form.
 * @param {(values: ShippingRateGroup) => void} props.onSubmit Called when the shipping rate group is submitted.
 * @param {() => void} props.onRequestClose Callback to close the modal.
 */
export default function AddRateFormModal( {
  countryOptions,
  initialValues,
  onSubmit,
  onRequestClose = noop,
} ) {
  return (
    <RateFormModal
      countryOptions={ countryOptions }
      initialValues={ initialValues }
      renderButtons={ ( formProps ) => {
        const { isValidForm, handleSubmit } = formProps;

        const handleSubmitClick = () => {
          onRequestClose();
          handleSubmit();
        };

        return [
          <Button
            key="submit"
            isPrimary
            disabled={ ! isValidForm }
            onClick={ handleSubmitClick }
          >
            { __( 'Add shipping rate', 'google-listings-and-ads' ) }
          </Button>,
        ];
      } }
      onSubmit={ onSubmit }
      onRequestClose={ onRequestClose }
    />
  );
}
```
### EditRateFormModal

```js
/**
 * External dependencies
 */
import { Button } from '@wordpress/components';
import { noop } from 'lodash';

/**
 * Internal dependencies
 */
import RateFormModal from './rate-form-modal.js';

/**
 * @typedef { import(".~/data/actions").CountryCode } CountryCode
 * @typedef { import("../typedefs.js").ShippingRateGroup } ShippingRateGroup
 */

/**
 * Form modal to edit or delete shipping rate group.
 *
 * @param {Object} props
 * @param {Array<CountryCode>} props.countryOptions Array of country codes, to be used as options in AppCountrySelect.
 * @param {ShippingRateGroup} props.initialValues Initial values for the form.
 * @param {(values: ShippingRateGroup) => void} props.onSubmit Called when the shipping rate group is submitted.
 * @param {() => void} props.onDelete Called when users clicked on the Delete button.
 * @param {() => void} props.onRequestClose Callback to close the modal.
 */
export default function EditRateFormModal( {
  countryOptions,
  initialValues,
  onSubmit,
  onDelete = noop,
  onRequestClose = noop,
} ) {
  const handleDeleteClick = () => {
    onRequestClose();
    onDelete();
  };

  return (
    <RateFormModal
      countryOptions={ countryOptions }
      initialValues={ initialValues }
      renderButtons={ ( formProps ) => {
        const { isValidForm, handleSubmit } = formProps;

        const handleSubmitClick = () => {
          onRequestClose();
          handleSubmit();
        };

        return [
          <Button
            key="delete"
            isTertiary
            isDestructive
            onClick={ handleDeleteClick }
          >
            { __( 'Delete', 'google-listings-and-ads' ) }
          </Button>,
          <Button
            key="submit"
            isPrimary
            disabled={ ! isValidForm }
            onClick={ handleSubmitClick }
          >
            { __( 'Update shipping rate', 'google-listings-and-ads' ) }
          </Button>,
        ];
      } }
      onSubmit={ onSubmit }
      onRequestClose={ onRequestClose }
    />
  );
}
```

### Their diff
```diff
 /**
- * Form to add a new rate for selected country(-ies).
+ * Form modal to edit or delete shipping rate group.
  * @param {(values: ShippingRateGroup) => void} props.onSubmit Called when the shipping rate group is submitted.
+ * @param {() => void} props.onDelete Called when users clicked on the Delete button.
  * @param {() => void} props.onRequestClose Callback to close the modal.
-export default function AddRateFormModal( {
+export default function EditRateFormModal( {
  onSubmit,
+ onDelete = noop,
  onRequestClose = noop,
 } ) {
+ const handleDeleteClick = () => {
+   onRequestClose();
+   onDelete();
+ };
+
  return (
        return [
+         <Button
+           key="delete"
+           isTertiary
+           isDestructive
+           onClick={ handleDeleteClick }
+         >
+           { __( 'Delete', 'google-listings-and-ads' ) }
+         </Button>,
          >
-           { __( 'Add shipping rate', 'google-listings-and-ads' ) }
+           { __( 'Update shipping rate', 'google-listings-and-ads' ) }
          </Button>,
```

Considering the diff result, an aspect of the duplication percentage in terms of total lines of code to be **84%**, calculated as `1 - 21 / (60 + 75) * 100%`. If we take into account four other files (time and threshold in the same pattern) that will be similar to them, these are the ones I was talking about: there are still a lot of avoidable duplicated codes. If 84% of avoidable duplication is not a kind of bloat for you, it's probably because we have different perceptions of what is duplication or bloat.

> Also, when the users are performing the clicking actions on the buttons above, it means that they have already made the `if-else` decision in their brain and selected a mode (e.g. "I click on Edit button" because "I want edit mode"), so we can just display the EditFooModal for them. IMO, in a way, having `<FooModal mode='edit'>` with mode checks in the component feels like we are bringing unnecessary trouble to ourselves when we could have capitalize on the user's decision and action (i.e. they have already made the if-else decision, why are we making that decision in code again?).

Considering this example:

```js
function Greetings( { role } ) {
  if ( ! role ) {
    return null;
  }

  let msg;
  if ( role === 'interviewee' ) {
    msg = 'Please go to the lounge and wait for the interview to begin.';
  } else {
    msg = 'Please have a seat, the interview will start soon.';
  }

  return (
    <div>
      {/*Here we put a large Hero block and many beautiful effects are attached here.*/}
      <div>Welcome!</div>
      <div>{ msg }</div>
    </div>
  );
}

function Welcome() {
  const [ role, setRole ] = useState();

  return (
    <div>
      <button onClick={ () => setRole( 'interviewee' ) }>
        I am the interviewee
      </button>
      <button onClick={ () => setRole( 'journalist' ) }>
        I am a journalist
      </button>

      <Greetings role={ role } />
    </div>
  );
}
```

If I understand your opinion right, then the above code is an improper pattern for you? And because the `role` is already selected in `Welcome`, passing `role` to any other children components will be considered as "we are bringing unnecessary trouble to ourselves when we could have capitalized on the user's decision".

If yes, when a user clicks on a button to trigger an async operation, they have already made the `if-else` decision for the "loading" state on the code consumer side. But there are already a lot of uses that pass the already known `loading` to `AppButton`, and a dev might even don't know the `loading` prop is tied onto the `disabled` state. Based on that opinion, we already have a lot of "unnecessary troubles" that those `AppButton` should be replaced with `AppLoadingButton`, `AppDisabledButton`, and `AppLoadingDisabledButton` on the consumer sides to make them with dedicated business logic and untie the hidden logic between `loading` and `disabled` states, right?

https://github.com/woocommerce/google-listings-and-ads/blob/425e86d95cb0088fed4226ad4af124db9b408d31/js/src/components/app-button/index.js#L32-L63

---

> At this moment in this PR, to change the control, we just need to make the change in the base `RateFormModal`, one place only, no?

I don't think so. Especially the one that pressing ENTER key to submit the form by the native support of \<form\>. And the same effort will happen several times for all these possible tasks for all modals of shipping rate, time, and threshold. It also includes an iterative development, writing tests, examination, and reviewing of the consistency and differences between them to prevent the copy-paste typos/problems. Such as the already happened [incorrect `typedefs` paths](https://github.com/woocommerce/google-listings-and-ads/pull/1424/commits/199be5ed750f729536d252ff207b4184f744db14), [incorrect properties in the validation](https://github.com/woocommerce/google-listings-and-ads/pull/1410/commits/c4dccb28ddd49c184d22ba107c3073bc13790d07), and the experiences that [consistency adjustment across all shipping modals had been missed like this](https://github.com/woocommerce/google-listings-and-ads/pull/1265/commits/993eeaaa871a1aedace4722fb238b9ac26633f44).

> > * But if this processing must be controlled in `EstimatedShippingRatesCard` layer, then in the code scope of shipping rate, the original need to change was two modal, now it becomes three, and even one more layer of props drilling and three times repeated JSDoc to be added.
> 
> My thought is that we just need to make the change in the base `RateFormModal`.

Unfortunately, in my limited knowledge, apart from using something like React Context to communicate globally, I wouldn't know how to pass a newly added prop/data from `EstimatedShippingRatesCard` to `AppModal` or `AppInputPriceControl` inside the `EditRateFormModal` without modifying `RateFormModal`.

(P.S. It's possible, but I don't think it's acceptable to use the `initialValues` to pass data that is totally unrelated to the form.)

> What I was suggesting is that if we were to have `<FooModal mode='edit'>`, we would need to refer to the `mode` documentation and find out that it requires `onDelete` props. If there are more modes and more props combinations, it would be more difficult.

Same, I doubt that more and more props combinations will happen in the known future. And if it was an anti-pattern, we already had the same problem in the AppButton.

> By having `<EditFooModal>`, it can be easily noticed early in the development, with simpler code and JSDoc than `<FooModal mode='edit'>`.

Comparing this to those already happened experiences in this comment, in my last comment, and more others in my memory, those efforts against these duplicated six or eight times shipping modals made me not lean toward this rather much smaller benefit.

> You are right, but the thing is, most often than not, people will just follow the existing pattern (in the name of reusability and maintainability) and add more `if`s into the shared modal to get their job done, and they have little interest or incentive to make the changes (because it is additional changes that are not their main goal).

I disagree with your point of view. Assuming what other engineers were like and what attitude they were working with, I don't think it's fair for them and even to be one of the bases for us to make a decision. 😕 They will have their own strategy to make their decision then.

> I have created issue #1439 with some ideas in it. We can handle it in separate PR.

Sounds great! Thanks for creating the follow-up!

---

I think I've gotten all your explanations and have read them at least three times. Even though most of them, I'm still not quite sure to agree they are relatively applicable to the facts that have happened now and in the past, or are relatively applicable to those changes that have a higher happening chance.

Anyway, thanks for your time and clarification! While I left more questions in this comment, I guess that would take more of your time, so **it's totally fine to skip over all my questions in this comment**. Unless there are unclear thoughts that need my clarification, and if you don't mind, I'd like to stop more replies from here. Thanks again! 🙏

ecgan:
> Anyway, thanks for your time and clarification! While I left more questions in this comment, I guess that would take more of your time, so **it's totally fine to skip over all my questions in this comment**. Unless there are unclear thoughts that need my clarification, and if you don't mind, I'd like to stop more replies from here. Thanks again! 🙏

Well, okay, I was thinking to reply your questions, but I guess we can talk about them some other time, like over a zoom call, or over a drink during a meetup one day. 😄  Thank you for your time and comments! 🙏

eason9487 (approved):
The main goal of this PR is to refactor and simplify technical debts, and the adjustments in this PR reduced component coupling, increased partial reuse and improved partial consistency.

Although there are still doubts that could be further improved by reference to some software design principles, which made me not quite sure that the overall adjustment meets the main goal - Refactor and Simplify, perhaps those doubts do not need to be resolved in this PR. So, I think we could still move forward and let these considerations settle for some more time.

---

📜 
My summary for future reference when doing related refactoring and simplification.

Although internalizing logical conditions into components seems not so consistent with the Open-Closed principle of software design, the internalization approach could still be considered for some simple and stable logic based on the following fact in the case of shipping modals:
- In software design principles, multi-hierarchy with multi-branch components is an anti-pattern.
- From requirements, those differences between the shipping modals that could be controlled by the internalized conditions have not changed since the first version. They are relatively stable.
- If they will need to be changed to an easily extensible design, then using methods such as Render Props can be made to satisfy the Open-Closed principle with relatively small adjustments.
- There are currently 8 similar shipping modals, and if the pattern proposed in this PR is applied to all of them, there will be 3 base modals and 6 modals for controlling differences such as buttons. Even though the reusability of the code has improved, there is still a high percentage of duplication, making maintenance less than ideal. Such as code implementation, writing tests, examination, and code review for the consistency and differences between them when iterative development.
- When designing changes, if the change is expected, prefer a design that makes it change easy. If the change is not expected for now, we can learn from the next change when it actually happens.

ecgan:
@eason9487 , thank you for the approval! 😄 👍  I think we can talk about the doubts and our opinions later some time in the future, in a venue where it is better for both of us to have those discussions. For now, it is good to get this moving. Thanks so much again for all your input, I appreciate our discussions. 🙂  🙏

js/src/components/shipping-rate-section/estimated-shipping-rates-card/getHandlers.js
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Get handlers for EstimatedShippingRatesCard.
+ *
+ * The handlers are to convert shipping rate group into shipping rates that can be propagated up via `onChange`.
+ *
+ * @param {Object} props
+ * @param {Array<ShippingRate>} props.value Array of individual shipping rates to be used as the initial values of the form.
+ * @param {(newValue: Array<ShippingRate>) => void} props.onChange Callback called with new data once shipping rates are changed.
+ */
+const getHandlers = ( { value, onChange } ) => {
```
tomalec, line 27: :nail_care:
Personally, I find `getHandlers` pretty vague. All I can read from that name is that it "returns something that's used to handle something" which is not much helpful.
WDYT of something more descriptive that states what functions it returns, like `getShipingRatesOperations`

Then, it's somewhat alerting to me that `getHandlers` returns on `handler` and two `get*Handlers`. `getHandler` that returns `get*Handler` isn't it `getGetHandler` ;) - that sounds quite redundant

Also, what is the value of having one function (`getHandlers`) that returns a set of three disjointed functions? They do not need access to each other or share any logic, do they?

Maybe we can just `export { getAddHandler, getChangeHandler, getDeleteHandler }` directly? Then we will save on constructing & destructing, and avoid the naming problem mentioned above. That will also make them more atomic and easier to test and maintain independently.

tomalec: :nail_care: Plus, I'd make it at least 
```suggestion
const getHandlers = ( value, onChange ) => {
```

As we call it as `getHandlers( { value, onChange } );` anyway, so constructing and deconstructing only blurs the picture for me

js/src/components/shipping-rate-section/estimated-shipping-rates-card/getHandlers.test.js
```diff
@@ -0,0 +1,328 @@
...
+
+		mockOnChange = jest.fn();
+		handlers = getHandlers( {
+			value,
+			onChange: mockOnChange,
+		} );
+	} );
+
+	describe( 'handleAddSubmit', () => {
+		it( 'returns value with the newly added group', () => {
```
tomalec, line 50: :construction: Actually, it does not "return value", but "calls the `onChange` callback with the value containing newly added group"

(I don't know why I cannot make a GH suggestion here)

js/src/components/shipping-rate-section/estimated-shipping-rates-card/getHandlers.test.js
```diff
@@ -0,0 +1,328 @@
...
+					currency: 'USD',
+					rate: 30,
+					options: {},
+				},
+			] );
+		} );
+	} );
+
+	describe( 'getChangeHandler', () => {
+		it( 'returns value updated based on changed group rate', () => {
```
tomalec, line 110: :construction: (here and below)
"When called with a value, returns a function, which when called with a new value, calls `onChange` callback with a changed...."

:nail_care: :grey_question: 
Just writing the above gives me a smell that maybe this state keeping logic may be a tad overcomplicated

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/add-rate-form-modal.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {Array<CountryCode>} props.countryOptions Array of country codes, to be used as options in AppCountrySelect.
+ * @param {ShippingRateGroup} props.initialValues Initial values for the form.
+ * @param {(values: ShippingRateGroup) => void} props.onSubmit Called with submitted value.
+ * @param {() => void} props.onRequestClose Callback to close the modal.
+ */
+const AddRateFormModal = ( {
+	countryOptions,
+	initialValues,
+	onSubmit,
+	onRequestClose = noop,
```
tomalec, line 31: :scroll: I got myself head-scratching for some time, thinking: "hey [the only use of `AddRateFormModal` we have](https://github.com/woocommerce/google-listings-and-ads/pull/1424/files#diff-5bc59001786c70aecac2fb120aec5a8fa59cfd84a6784ef80ba27a7c14c04c70R155-R164) does not set `onRequestClose`, so why don't we remove it?"

That's already reported in https://github.com/woocommerce/google-listings-and-ads/issues/1439

:nail_care: But I think the default value of `noop` is not needed here, as we don't use it without given `onRequestClose` prop. So maybe it's YAGNI, plus we can fails fast if something is unexpectedly missing.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/edit-rate-form-modal.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param {(values: ShippingRateGroup) => void} props.onSubmit Called when the shipping rate group is submitted.
+ * @param {() => void} props.onDelete Called when users clicked on the Delete button.
+ * @param {() => void} props.onRequestClose Callback to close the modal.
+ */
+const EditRateFormModal = ( {
+	countryOptions,
+	initialValues,
+	onSubmit,
+	onRequestClose = noop,
+	onDelete = noop,
```
tomalec, line 33: :nail_care: :grey_question: If we don't have a use case when we do not provide those, I'd not set the default value, to fail fast in case of an unexpected lack of those.

tomalec (commented):
I left a few post-mergem comments, as I was getting the context before https://github.com/woocommerce/google-listings-and-ads/pull/1410

Issue #1165, opened 2021-12-21
GMC Account Review Request


puntope:
## User story

As a merchant, I want a Request Account Review button directly in the Issues section of the plugin. I don't want to proactively log in to and navigate through Google Merchant Center in order to request an account review for re-approval. 

## Internal Project Thread Link
pcTzPl-bx-p2

## Figma link
https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=3736%3A72890

## Acceptance criteria
* [x] Split Product feed Issues between `Account issues` and `Product Issues`. #1176 
* [x] Refactor the Account Issues API including ([freelistingsprogram](https://href.li/?https://developers.google.com/shopping-content/reference/rest/v2.1/freelistingsprogram/requestreview) and [shoppingadsprogram](https://href.li/?https://developers.google.com/shopping-content/reference/rest/v2.1/shoppingadsprogram/requestreview))
* [x] Create the No Issues component. In case of issues in one section but not in the other... the one with no issues will show a success message. https://github.com/woocommerce/google-listings-and-ads/pull/1196
* [x] Create The Request Review Notice component. It only appears in case there are issues on  `Account issues`
* [x] Create the Request Review Modal component that renders when "Request Review" is clicked. It shows a confirmation checkbox and a summary of issues to review. 
* [x] When the Review is in progress the Request Review Notice shows that in their state
* [x] When the Review is successful the  Request Review Notice is not shown
* [x] When the Review is not successful it can be Requested again only after 7 days.  Request Review Notice has state for showing both. Unsuccessful in cooldown and Unsuccessful out cooldown.
* [x] When there are no issues (both account & product), the Issues section is hidden with feed setup status updated to say ‘No issues to resolve’. #1176 

### Non functional requirements
* [x] Unit testing
* [x] Tracking

## Event tracking - TBC
- `gla_modal_request_review_open` - When the user clicks on the `Request review` button and modal is opened.
- `gla_modal_request_review_closed` - When the user closes the modal.
- `gla_request_account_review` - When the user finally clicks on `Request account review ` button inside the modal. Potentially send the issues here.

PR #1427, opened 2022-04-14
Map Issue Codes with MC issues

puntope:

Closes part of #1165 

Maps the codes provided by the new API endpoint with the Issues with description obtained in the `useMCIssuesTypeFilter` dispatcher in order to show them in the Request Review Modal.

### Screenshots:

Screenshot 2022-04-14 at 17 24 55

1. Checkout PR
2. First of all, take a look at `src/API/Site/Controllers/MerchantCenter/AccountReviewController.php::get_review_read_callback` you can see there some issues codes like `#1 Issue one` those are dummy hardcoded Issue codes to be shown in the Request review Modal they should match with any MC Issue (in database). 
4. Probably you want to test with some issues creating them in `wp_gla_merchant_issues` just be sure the code column match with some of the issues in `src/API/Site/Controllers/MerchantCenter/AccountReviewController.php::get_review_read_callback`, in the column "issue" you can write some dummy description.
5. Go to product Feed `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed` 
6. See The notice in the Account Issues section and click over Request Review
7. You will see the description of those reviews matching the database. in the issue list of the modal. 
8. If there is no issue matching a database issue code... no issue list neither checkbox will be shown.

### DIFF
--- a/js/src/product-feed/review-request/review-request-issues.js
+++ b/js/src/product-feed/review-request/review-request-issues.js
@@ -32,7 +32,7 @@ const ReviewRequestIssues = ( { issues = [] } ) => {
 			</Text>
 			<ul className="gla-review-request-modal__issue-list">
 				{ issuesToRender.map( ( issue ) => (
-					<li key={ issue }>{ issue }</li>
+					<li key={ issue.code }>{ issue.issue }</li>
 				) ) }
 			</ul>
 			{ issues.length > COLLAPSED_ISSUES_SIZE && (
--- a/js/src/product-feed/review-request/review-request-issues.js
+++ b/js/src/product-feed/review-request/review-request-issues.js
@@ -10,6 +10,7 @@ const COLLAPSED_ISSUES_SIZE = 5;
 
 const ReviewRequestIssues = ( { issues = [] } ) => {
 	const [ expanded, setExpanded ] = useState( false );
+	if ( ! issues.length ) return null;
 
 	const toggleExpanded = () => {
 		recordEvent( 'gla_request_review_issue_list_toggle_click', {
--- a/js/src/product-feed/review-request/review-request-modal.js
+++ b/js/src/product-feed/review-request/review-request-modal.js
@@ -22,7 +22,7 @@ const ReviewRequestModal = ( {
 } ) => {
 	const [ checkBoxChecked, setCheckBoxChecked ] = useState( false );
 
-	if ( ! issues.length || ! isActive ) {
+	if ( ! isActive ) {
 		return null;
 	}
 
--- a/js/src/product-feed/review-request/review-request-modal.js
+++ b/js/src/product-feed/review-request/review-request-modal.js
@@ -50,7 +50,7 @@ const ReviewRequestModal = ( {
 				<AppButton
 					key="primary"
 					isPrimary
-					disabled={ ! checkBoxChecked }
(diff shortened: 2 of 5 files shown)


js/src/product-feed/review-request/review-request-modal.js
```diff
@@ -87,15 +87,17 @@ const ReviewRequestModal = ( {
...
-			<CheckboxControl
-				className="gla-review-request-modal__checkbox"
-				label={ __(
-					'I have resolved all the issue(s) listed above.',
-					'google-listings-and-ads'
-				) }
-				checked={ checkBoxChecked }
-				onChange={ handleCheckboxChange }
-			/>
+			{ issues.length > 0 && (
```
jorgemd24, line 100: ❓

Is it not better to check if there are issues on the top of this file? If issues.length === 0 do not render this modal? 

Or maybe is even better if we do it just [before this line](https://github.com/woocommerce/google-listings-and-ads/blob/6f863dcf3ef87074ac973f11cc51bfd6e9261d36/js/src/product-feed/review-request/index.js#L58) and we do not render the `<ReviewRequestModal/>` component at all.

js/src/product-feed/review-request/review-request-issues.js
```diff
@@ -32,7 +33,7 @@ const ReviewRequestIssues = ( { issues = [] } ) => {
 			</Text>
 			<ul className="gla-review-request-modal__issue-list">
 				{ issuesToRender.map( ( issue ) => (
-					<li key={ issue }>{ issue }</li>
+					<li key={ issue.code }>{ issue.issue }</li>
```
jorgemd24, line 36: ❔

I suppose that each `issue.code` is unique and can not be repeated otherwise we will have the warning: Encountered two children with the same key

js/src/product-feed/review-request/review-request-modal.js
```diff
@@ -22,7 +22,7 @@ const ReviewRequestModal = ( {
 } ) => {
```
jorgemd24, line 32: 💅

I think it would be nice to add in the future some JSDocs so it is easier to understand some variables.

jorgemd24 (approved):
Thanks, @puntope, LGTM 👍 

I have just left some comments about rendering the modal and other minor things.

js/src/product-feed/review-request/review-request-issues.js
puntope: Yes, I filtered by Unique [in the backend Part PR](https://github.com/woocommerce/google-listings-and-ads/pull/1413)

Thanks for confirming

js/src/product-feed/review-request/review-request-modal.js
puntope: Let me explain the rationale behind this: 

Before this PR we actually hide the modal if there were no issues:

`if ( ! issues.length || ! isActive ) {` 

Right now we still render the modal if there is no issues. That's pretty much defensive since in case the account shows the request button it means there are issues but in the worst case scenario that for some reason the status was DISAPPROVED but without providing any issue (maybe a Google API issue or a bug on my side) it'd still show the modal allowing the user to request the review. 

If we avoid rendering the modal if no issues, in case this rare error happens it will show the Request Review button but it will never open the modal since there are no issues.

Issue #1392, opened 2022-04-01
UX Improvements on Get Started Page


ianlin:
### Goal

To better explain the benefits and value of the Google Listings & Ads integration to the merchants, in order to reduce the current drop-off rate of ~31%.

### Overview

1. Highlight more of the benefits of this extension.
    - The current welcome page is underselling the potential value of this extension.
2. Communicate and educate the user about Smart Shopping Campaigns.
    - By failing to mention “Smart Shopping Campaigns” in enough depth at the beginning, users doubt the total value of the extension. This means their expectations are also misled and they miss out on an opportunity to learn about “Smart Shopping Campaigns” (something few knew about)
3. Provide an estimate of how long setup will take.
    - Eg. 15 minutes.
4. Include a place for our marketing video.
    - You can watch the video [here](https://woocommerce.com/products/google-listings-and-ads/)
5. Include a place for case studies / customer quotes.

From the new design, there are several new sections in the get started page:

1. Get started CTA with video.
2. Benefits.
3. Features.
4. Case studies.
5. Get started CTA again.
6. FAQ

<img width="451" alt="Screenshot 2022-04-01 at 12 55 20" src="https://user-images.githubusercontent.com/914406/161197443-58be67f3-35d2-4f77-87d7-7b155425f7e7.png">

### The current state of the system

In our current design, the in-product Get Started page is pretty bare, communicating just a little about the integration. Based on the funnel analysis in Sept 2021, the drop-off here is pretty steep; 31% who view this screen do not proceed with their setup.

There are only three sections in the current get started page:

1. Get started CTA.
2. Features.
3. FAQ.

<img width="888" alt="Screenshot 2022-04-01 at 12 59 48" src="https://user-images.githubusercontent.com/914406/161197840-36433faf-e043-4abc-8f96-cda3a222b64c.png">

## Technical

💡 We will create a shared branch [feature/1392-ux-improvements-on-get-started-page](https://github.com/woocommerce/google-listings-and-ads/tree/feature/1392-ux-improvements-on-get-started-page). All relevant developments should be branched from it and merged back to it. And open a PR to merge the shared branch to develop once the overall feature gets ready to release.
💡 This project is mainly focusing on the UI changes, no backend modification required.

Currently there are two cards, `get-started-card` and `features-card`, and one FAQ section. We need to update it to five cards and one FAQ section according to the new design.

The five cards and the FAQ in the code structure would be:

[...]

PR #1432, opened 2022-04-18
Get started page UX improvements: add get started with video card

ianlin:

Part of #1392, this PR add the first card of the new get started page. The first card includes the same marketing video in the [Google Listings & Ads WooCommerce product page](https://woocommerce.com/products/google-listings-and-ads/).

To be able to play the video, I referred to the implementation from 10517-gh-Automattic/woocommerce.com but using [Wistia's JavaScript Player API](https://wistia.com/support/developers/player-api#with-iframe-embeds) to get the video state in order to expand the video height after the user clicks on the video.

To be able to use Wistia's API, I add a custom hook `useScript` and use it to load Wistia's script dynamically in the component.

Note that the video will not be expanded in the mobile version.

Design for desktop: eQ9O4m2flzrcAiDMXkOW0m-fi-7350%3A120914
Design for mobile: eQ9O4m2flzrcAiDMXkOW0m-fi-7476%3A122828

#### Brave (Chromium)

https://user-images.githubusercontent.com/914406/163765567-62a6b434-8359-4f07-b492-4df607bb9085.mov

### Firefox

https://user-images.githubusercontent.com/914406/163766064-938019f3-b02b-4183-96e8-248e9575ae3c.mov

### Safari

https://user-images.githubusercontent.com/914406/163767314-cdf16c48-8f95-4dd5-a4e6-6c111308d767.mov

2. Check the first card of the get started page, the UI should align with the design.
3. The video will be expanded when playing the video on the desktop.
4. The video won't be expanded when playing the video on the mobile.
5. Click on the `Start listing products` and it will go to the set up page.

### Changelog entry

> Update - Add the first card with a CTA and a video for UX improvements on get started page.


js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { glaData } from '.~/constants';
+import AppButton from '.~/components/app-button';
+import AppDocumentationLink from '.~/components/app-documentation-link';
+import WistiaVideo from '.~/components/wistia-video';
+import footerImageURL from './footer.svg';
+import googleImageURL from './google.svg';
```
eason9487: ❔ 
There is already a Google logo at [here](https://github.com/woocommerce/google-listings-and-ads/blob/0cfd8f16e7cd061e974c62cb3430b38c0918b380/js/src/product-feed/submission-success-guide/google-logo.svg). It could be reused by setting the needed image width or height.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+import { getNewPath } from '@woocommerce/navigation';
+
+/**
+ * Internal dependencies
+ */
+import { glaData } from '.~/constants';
+import AppButton from '.~/components/app-button';
+import AppDocumentationLink from '.~/components/app-documentation-link';
+import WistiaVideo from '.~/components/wistia-video';
+import footerImageURL from './footer.svg';
```
eason9487: ❓ 
This icon and the right-side description could be replaced with \<Tip\> provided by `@wordpress/components`.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+import googleImageURL from './google.svg';
+import './index.scss';
+
+const GetStartedWithVideoCard = () => {
+	const disableNextStep = ! glaData.mcSupportedLanguage;
+
+	return (
+		<Card
+			className="woocommerce-marketing-google-get-started-with-video-card"
```
eason9487: ❔ 
I guess the *woocommerce-marketing-googl* prefix was only used at the beginning of this plugin, but it's too long then and didn't be used other than the Get started page. Currently, most CSS class names are started with *gla-*.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+			isBorderless
+		>
+			<Flex>
+				<FlexBlock className="header">
+					<Text variant="body" className="title">
```
eason9487: ❓ 
Suggest avoiding using CSS names that easily name collisions.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+						{ __(
+							'The official extension for WooCommerce, built in collaboration with',
+							'google-listings-and-ads'
+						) }
+					</Text>
+					<div id="google-img">
```
eason9487: ❓ 
Suggest using `className` instead because `id` usually is only adopted when there is a concrete reason. From the perspective of component design, most components are not designed with the premise of unique existing on the whole page, even though sometimes they are the only instance in the runtime.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+						{ __(
+							'Sync your products directly to Google, manage your product feed, and create Google Ad campaigns–without leaving your WooCommerce dashboard. The official extension, built in collaboration with Google.',
+							'google-listings-and-ads'
+						) }
+					</Text>
+					<AppButton
+						className="button"
+						isPrimary
+						disabled={ disableNextStep }
+						href={ getNewPath( {}, '/google/setup-mc' ) }
```
eason9487: 💅  It could be replaced with https://github.com/woocommerce/google-listings-and-ads/blob/0cfd8f16e7cd061e974c62cb3430b38c0918b380/js/src/utils/urls.js#L50-L52

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+						eventName="gla_setup_mc"
+						eventProps={ {
+							target: 'set_up_free_listings',
+							trigger: 'click',
+						} }
```
eason9487: 💅  Suggest adding event JSDoc for auto-generating the tracking docs.

https://github.com/woocommerce/google-listings-and-ads/blob/0cfd8f16e7cd061e974c62cb3430b38c0918b380/js/src/get-started-page/get-started-card/index.js#L24-L26

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+					>
+						{ __(
+							'Start listing products →',
+							'google-listings-and-ads'
+						) }
+					</AppButton>
+					<Text className="hint">
```
eason9487: ❓  Same concern for the CSS naming collisions.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+										context="get-started"
+										linkId="wp-terms-of-service"
+										href="https://wordpress.com/tos/"
+									/>
+								),
+							}
+						) }
+					</Text>
+				</FlexBlock>
+				<FlexBlock className="footer">
```
eason9487: ❓ Same concern for the CSS naming collisions.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+						) }
+					</Text>
+					<Text variant="body" className="description">
+						{ __(
+							'Sync your products directly to Google, manage your product feed, and create Google Ad campaigns–without leaving your WooCommerce dashboard. The official extension, built in collaboration with Google.',
+							'google-listings-and-ads'
+						) }
+					</Text>
+					<AppButton
+						className="button"
```
eason9487: ❓ Same concern for the CSS naming collisions.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+										href="https://wordpress.com/tos/"
+									/>
+								),
+							}
+						) }
+					</Text>
+				</FlexBlock>
+				<FlexBlock className="footer">
+					<img src={ footerImageURL } alt="" />
+					<Text variant="body" className="hint">
```
eason9487: ❓ Same concern for the CSS naming collisions.

js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -0,0 +1,122 @@
...
+						} }
+					>
+						{ __(
+							'Start listing products →',
+							'google-listings-and-ads'
+						) }
+					</AppButton>
+					<Text className="hint">
+						{ __(
+							'Estimated setup time: 15 min ',
```
eason9487: 💅  The space char at the end could be removed.

js/src/get-started-page/get-started-with-video-card/index.scss
```diff
@@ -0,0 +1,120 @@
...
+				.title {
+					padding: 0;
+					font-size: 20px;
+					line-height: $default-line-height;
+				}
+			}
+
+			.footer {
+				margin: $grid-unit-20;
+			}
```
eason9487: 💅 
According to the visual design, the icon is aligned to the top in the mobile view.
image

js/src/get-started-page/get-started-with-video-card/index.scss
```diff
@@ -0,0 +1,120 @@
+.woocommerce-marketing-google-get-started-with-video-card {
+	.components-flex {
+		flex-direction: column;
+
+		.header {
```
eason9487: 💅  or 📜 
With the CardHeader, CardBody, and CardFooter components provided by `@wordpress/components`, I think some styles in this file could be omitted and be more consistent with the codebase.

js/src/get-started-page/index.js
```diff
@@ -5,12 +5,14 @@ import './index.scss';
...
 import GetStartedCard from './get-started-card';
+import GetStartedWithVideoCard from './get-started-with-video-card';
 import FeaturesCard from './features-card';
 
 const GetStartedPage = () => {
 	return (
 		<div className="woocommerce-marketing-google-get-started-page">
 			<UnsupportedNotices />
+			<GetStartedWithVideoCard />
 			<GetStartedCard />
```
eason9487, line 16: ❔ or 💅 
Would the GetStartedCard components be removed in another PR, or we could remove them along with this PR?

js/src/hooks/useScript.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+import { useEffect } from '@wordpress/element';
+
+/**
+ * Load an external script and call the provided callback function when the script is loaded.
+ *
+ * @param {string}   url      The script's url.
+ * @param {Function} callback A function to be called after the script is loaded.
+ */
+const useScript = ( url, callback = () => {} ) => {
```
eason9487: 💅  The default value of `callback` is not needed if it's a required parameter. Or, if the `callback` is optional, it'd be better to indicate that in the JSDoc and use the same reference for the default value. Otherwise, the callback of `useEffect` will be triggered when every time rendering.

Usually. The *noop* can be imported from `lodash` or declared at the module scope.

```js
import { noop } from 'lodash';
// or
const noop = () => {};
```

js/src/components/wistia-video/index.js
```diff
@@ -0,0 +1,97 @@
...
+	useEffect( () => {
+		if ( wistiaVideoHandle ) {
+			const intervalId = setInterval( () => {
+				if ( wistiaVideoHandle.state() === 'playing' ) {
+					setPlaying( true );
+					clearInterval( intervalId );
+				}
+			}, 100 );
+		}
+	}, [ wistiaVideoHandle ] );
```
eason9487: 💅  This handling could be replaced with the following in the callback of `useScript`:
```js
window._wq.push( {
	id,
	onReady: ( video ) => {
		video.bind( 'play', () => setPlaying( true ) );
	},
} );
```

js/src/components/wistia-video/index.js
```diff
@@ -0,0 +1,97 @@
...
+	useScript( WISTIA_API_SCRIPT_URL, () => {
+		if ( window._wq ) {
+			window._wq.push( {
+				id,
+				onReady: ( video ) => {
+					setWistaVideoHandle( video );
+				},
+			} );
+		}
+	} );
```
eason9487: 🚧 
The callback function is a different reference in each rendering, which leads the script might be appended to and removed from the document body several times, and the initial processing callback itself is called several times. Even though it appears that wistia internally handles the problem of script reloading and reinitialization, this still added some potential concerns. Suggestions to avoid unexpected changes in the callback reference.

Kapture 2022-04-18 at 19 33 35

js/src/components/wistia-video/index.scss
```diff
@@ -0,0 +1,43 @@
+.wistia-responsive-padding {
```
eason9487: ❔ Wondering why it's named with `-padding` postfix?

js/src/components/wistia-video/index.js
```diff
@@ -0,0 +1,97 @@
...
+	};
+
+	return (
+		<div
+			className={ classNames(
+				'wistia-responsive-padding',
+				isPlaying ? 'is-playing' : ''
+			) }
+		>
+			<div className="wistia-responsive-wrapper">
```
eason9487: 💅
In GLA codebase, the naming convention of subcomponent is usually named in the form of gla-main-component-name__part-name. Using __ as a separator makes the main and sub distinguishable.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks, updated in 40c6663.

js/src/get-started-page/index.js
ianlin: Thanks for the reminder, there is a new card get `GetStartedCard` as well like in the design: eQ9O4m2flzrcAiDMXkOW0m-fi-7350%3A121253, so I will update it in another PR.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Refactored the css part in 6eb1e68 and 0aba591.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Refactored the css part in 6eb1e68 and 0aba591.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Refactored the css part in 6eb1e68 and 0aba591.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Refactored the css part in 6eb1e68 and 0aba591.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Refactored the css part in 6eb1e68 and 0aba591.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Refactored the css part in 6eb1e68 and 0aba591.

js/src/get-started-page/get-started-with-video-card/index.scss
ianlin: Thanks! Refactored the css part in 6eb1e68 and 0aba591.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Updated in 5759c51.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Updated in e84172b.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Updated in 1b95995.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: Thanks! Added in b0f903d.

js/src/get-started-page/get-started-with-video-card/index.js
ianlin: 👀  Thanks! Removed in dadeb91.

js/src/get-started-page/get-started-with-video-card/index.scss
ianlin: Thanks! Updated in df9220a.

js/src/hooks/useScript.js
ianlin: Thanks! Updated in c426ee5.

js/src/components/wistia-video/index.js
ianlin: Awesome! Thanks for the suggestion, updated in 71a9913.

js/src/components/wistia-video/index.js
ianlin: Thanks for spotting this! Updated in 910e112 and confirmed it won't be loading the script again when re-render.

js/src/components/wistia-video/index.scss
ianlin: It was copied from the Wistia's sample code 😅. Updated in 2c31424, thanks!

js/src/components/wistia-video/index.js
ianlin: Thanks, also updated in 2c31424.

ianlin:
Hey @eason9487! Thanks for all the review, it's super useful. I've updated the code based on your suggestions and it's ready for another round of the review.

js/src/get-started-page/get-started-with-video-card/index.scss
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	// Repeat selector to make it higher priority.
+	& &__description {
+		margin-bottom: $grid-unit-40;
+		padding: 0 $grid-unit-10;
+	}
+
+	// Repeat selector to make it higher priority.
+	& &__button {
```
eason9487: 💅 

Some styling details could be considered.
- The padding of `.components-card__header` still has a lower priority than `@wordpress/components`.
- Specify the proper content height to avoid overriding the spacing and hardcoded height on the container. This also brings a flexible layout when the content changes or is dynamic.
- Some CSS styles could be omitted.
- Remove some code comments since `& &__` is not a repeated-selector pattern.

I added a code suggestion block because it might be clearer than using a lot of comments. 🙏 
```suggestion
	// Adjust <CardHeader> imported from @wordpress/components.
	// Repeat selector to make it higher priority.
	.components-card__header.components-card__header {
		justify-content: flex-end;
		padding: $grid-unit * 2.5;

		& > p {
			margin: 0 1em;
			line-height: 16px;
			font-size: $helptext-font-size;
			color: $gray-600;
		}

		@media (max-width: $break-small) {
			justify-content: center;
			text-align: center;
			padding: $grid-unit-20;

			& > p::after {
				content: " Google";
			}

			& > img {
				display: none;
			}
		}
	}

	.motivation-video {
		min-width: 100%;
		margin: 0;
	}

	// Adjust <CardBody> imported from @wordpress/components.
	// Repeat selector to make it higher priority.
	.components-card__body.components-card__body {
		display: flex;
		flex-direction: column;
		align-items: center;
		text-align: center;
		margin: calc(var(--main-gap) * 1.67) calc(var(--main-gap) * 4.67);
		padding: 0;

		@media (max-width: $break-small) {
			margin: $grid-unit-30;
		}
	}

	& &__title {
		padding: 0 $grid-unit-30;
		margin-bottom: $grid-unit-15;
	}

	& &__description {
		margin-bottom: $grid-unit-40;
		padding: 0 $grid-unit-10;
	}

	& &__button {
```

eason9487 (approved):
Thanks for the updates! Testing well. LGTM. Left a comment about the styling, and it's not a blocker.

js/src/get-started-page/get-started-with-video-card/index.scss
eason9487: I missed the other two code comments below, which are also removable.

ianlin: Thanks for the suggested changes @eason9487 , it's much more cleaner now! I've got a question regarding:

> & &__ is not a repeated-selector pattern.

Agree it's different from the repeating form like `className.className`. But using `& &__` actually solved the priority problem to me in this case. I checked the [MDN CSS Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#how_is_specificity_calculated) doc and it seems like `& &__` matches either the rule of

1. Make better use of the CSS cascade, or
2. Use more specific rules. By indicating one or more elements before the element you're selecting, the rule becomes more specific and gets higher priority

So it could have higher priority to override the `Text` component's style from `@wordpress/components`. I'd like to add some comments like `// Use more specific rules to make it higher priority.` to those use `& &__`, what do you think about it?

eason9487: > So it could have higher priority to override the `Text` component's style from `@wordpress/components`. I'd like to add some comments like `// Use more specific rules to make it higher priority.` to those use `& &__`, what do you think about it?

It might seem a bit strange at first glance to see something like `& &__title`, but the purpose of this syntax becomes clear after thinking about the compiled result like `.card .card__title`. And its purpose is exactly the either or both reasons you mentioned. Since these two purposes are very common rules for writing/reading CSS, they usually don't need to be commented on.

But on the other hand, when styling for a component that is entirely under our control, there is no need to override some other styles, then the use of `&__title` is usually sufficient. So even if the reason for these selectors can be deduced here is to raise the priority over `Text`, it would be still fine to leave comments for this concrete reason.

Issue #717, opened 2021-06-02
Override the "enhanced free listings" warning


j111q:
## Context

A significant number of beta testers have been seeing this message the second they complete their setup:
![image](https://user-images.githubusercontent.com/65582817/120415886-de764f80-c38e-11eb-9c34-9a813b6c9007.png)

They have assumed that their whole setup isn't working, because they don't understand that this is for **enhanced** listings. 

The Google team will be providing feedback internally to their own team, but in the meantime, they've given us the green light to override it for v1 if we can. cc @JamesAntisdel 

## Changes

1. Change it from a red disapproval icon > to a yellow warning icon. 
2. Change the messaging to `Issue: Show products on additional surfaces across Google through enhanced free listings` and `Suggested action: Read about enhanced free listings`

![image](https://user-images.githubusercontent.com/65582817/120416215-5fcde200-c38f-11eb-9f07-92be2d90f528.png)

## Figma link
https://www.figma.com/file/jZUpa8eTrnrK1Lwt2ry7zk/Native-Google-Integration?node-id=9712%3A844

PR #1437, opened 2022-04-20
GLA Request Review - Hide warning issues when critical issues

puntope:

Closes part of #1165 .

We want to hide warning issues in case some critical issues are present. 

1. Go to database and in `wp_gla_merchant_issues` create some issues, at least one warning and one error.
2. Go to Product Feed
3. Only Critical issues are shown.
4. Go back to database and remove the error one.
5. Go to Product Feed
6. Warning issues are shown.
7. Paginator and issues totals are consistent

### Additional details:

Some issues are overrided in the code: See -> https://github.com/woocommerce/google-listings-and-ads/issues/717

We changed here the moment when we override some issues

This is because the override is happening after we query the issues. Hence we are asking for critical issues but then maybe they are overrided after to warning. 

So now we are doing this when we insert the issue in DB. that way we can filter correctly the issues.

Note that if you have already the issues in your DB maybe you have the previous version of the override issue and it will show a warning issue on critical issues. This will be solved on refresh in that case. 

I didn't delete `woocommerce_gla_merchant_issue_override` filter tho... since we don't know if is being used by a consumer.

### DIFF
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -124,8 +124,10 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 	/**
 	 * Retrieve the Merchant Center issues and total count. Refresh if the cache issues have gone stale.
 	 * Issue details are reduced, and for products, grouped by type.
-	 * Issues can be filtered by type, searched by name or ID (if product type) and paginated.
+	 * Issues can be filtered by type, severity and searched by name or ID (if product type) and paginated.
 	 * Count takes into account the type filter, but not the pagination.
+	 *
+	 * In case there are issues with severity Error we hide the other issues with lower severity.
 	 *
 	 * @param string|null $type To filter by issue type if desired.
 	 * @param int         $per_page The number of issues to return (0 for no limit).
@@ -137,7 +139,12 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 	 */
 	public function get_issues( string $type = null, int $per_page = 0, int $page = 1, bool $force_refresh = false ): array {
 		$this->maybe_refresh_status_data( $force_refresh );
-		return $this->fetch_issues( $type, $per_page, $page );
+
+		// Get only critical issues
+		$critical_issues = $this->fetch_issues( $type, $per_page, $page, self::SEVERITY_ERROR );
+
+		// In case there are critical issue we show only those, otherwise we show all the issues.
+		return $critical_issues['total'] > 0 ? $critical_issues : $this->fetch_issues( $type, $per_page, $page );
 	}
 
 	/**
@@ -245,11 +252,12 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 	 * @param string|null $type To filter by issue type if desired.
 	 * @param int         $per_page The number of issues to return (0 for no limit).
 	 * @param int         $page The page to start on (1-indexed).
+	 * @param string|null $severity Filters the issues by severity level.
 	 *
 	 * @return array The requested issues and the total count of issues.
 	 * @throws InvalidValue If the type filter is invalid.
 	 */
-	protected function fetch_issues( string $type = null, int $per_page = 0, int $page = 1 ): array {
+	protected function fetch_issues( string $type = null, int $per_page = 0, int $page = 1, string $severity = null ): array {
 		/** @var MerchantIssueQuery $issue_query */
 		$issue_query = $this->container->get( MerchantIssueQuery::class );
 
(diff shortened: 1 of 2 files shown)


src/MerchantCenter/MerchantStatuses.php
```diff
@@ -137,7 +139,12 @@ public function get_product_statistics( bool $force_refresh = false ): array {
 	 */
 	public function get_issues( string $type = null, int $per_page = 0, int $page = 1, bool $force_refresh = false ): array {
 		$this->maybe_refresh_status_data( $force_refresh );
-		return $this->fetch_issues( $type, $per_page, $page );
+
+		// Get only critical issues
+		$critical_issues = $this->fetch_issues( $type, $per_page, $page, self::SEVERITY_ERROR );
```
mikkamp: The naming here is slightly confusing since in the Google API `error` and `critical` are different severity levels. 
It also doesn't work for critical errors since in the DB we save the severity as it's retrieved from the API.

The function `get_issue_severity` combine the severity into just two levels, but that's only returned in the responses. If we filter for errors, we need to look for severity "critical OR error" (possibly more?). Or alternatively we need to save them in the DB with just two levels.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -782,4 +796,22 @@ protected function get_issue_severity( array $row ): string {
...
+
+	/**
+	 * In very rare instances, issue values need to be overridden manually.
+	 *
+	 * @param array $issue
+	 *
+	 * @return array The original issue with any possibly overridden values.
+	 */
+	private function maybe_override_issue_values( array $issue ): array {
+		if ( 'merchant_quality_low' === $issue['code'] ) {
```
mikkamp, line 808: Do we have documentation to show that this code is only used for the message/severity we want to override? `merchant_quality_low` sounds like it could be generic but I don't find much information about that code.

mikkamp (commented):
Thanks for the changes, overall they look good. Although in my DB the errors are saved with the severity "critical" so it doesn't find only errors and always returns the full list.

src/MerchantCenter/MerchantStatuses.php
puntope: Good spot @mikkamp 

As far as i read here https://developers.google.com/shopping-content/guides/how-tos/severity-mapping

We only have critical, error, and suggestion. Seems like the documentation is different about the statuses we are handling in code. I think i will go for passing a boolean $only_error_issues inside the function and do the where with `error` and `critical` issue.

src/MerchantCenter/MerchantStatuses.php
puntope: In here they only have one id per issue:

https://developers.google.com/shopping-content/guides/account-issues

And it clearly says it's not eligible for free listing. My worries abut using issue prop as a criteria for the override is that in case Google for some reason modifies the title the conditional wont work anymore. By using the issue code we are kinda sure we are catching it.

src/MerchantCenter/MerchantStatuses.php
puntope: Hey @mikkamp 

I applied your suggestion here  https://github.com/woocommerce/google-listings-and-ads/pull/1437/commits/f04dbbbbd692ac32c3a23e9a0b3e1d574acd2ef6

puntope:
@mikkamp  All the comments were discussed and required adjustments applied. This PR is ready for another round of reviewing. Thanks in advance.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -137,7 +139,12 @@ public function get_product_statistics( bool $force_refresh = false ): array {
 	 */
 	public function get_issues( string $type = null, int $per_page = 0, int $page = 1, bool $force_refresh = false ): array {
 		$this->maybe_refresh_status_data( $force_refresh );
-		return $this->fetch_issues( $type, $per_page, $page );
+
+		// Get only critical issues
```
mikkamp: Minor: We can adjust this comment to just mention error issues

mikkamp (approved):
Thanks for the changes, it's working well now. I'm getting a combination of error and critical.

Just left one small minor note about a comment, but it looks ready to merge.

src/MerchantCenter/MerchantStatuses.php
puntope: Thanks 🙏 Updated here https://github.com/woocommerce/google-listings-and-ads/pull/1437/commits/71823072f6fac6cd65562d21f2d5acfb389aab7f  (still thinking about how I didn't see that one 🤔  tho)

PR #1196, opened 2022-01-14
Improved Default Issue Type and Issues Solved state

puntope:
### Changes proposed in this Pull Request:

Partially Closes #1165 

This PR improves 2 things:

- Improves the Active Issue type logic as follow up for the comment here pcTzPl-da-p2#comment-430  and here https://github.com/woocommerce/google-listings-and-ads/pull/1176#discussion_r777647374

- Adds the Solved Issues state following the design direction in this Figma https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=4337%3A105188

### Screenshots:

Screenshot 2022-01-14 at 19 12 47

### Detailed test instructions:

1. Enter in GLA Product Feed section in an account with both Product and Account issues
2. See that Account Issues is selected by default and it shows the issues as normal
3. Switch to product Issues and see that it shows the Product issues as normal
4. Reload the page with Product issues selected. After reload. It should be Product Issues selected.
5. Delete in the database All your Account issues. You can use this command `DELETE FROM wp_gla_merchant_issues WHERE type = 'account';` 
6. Reload the page (still Product Issues selected) see that they load correctly. 
7. Switch to account issues. You should see a message similar to the Screenshot one. Saying that Account issues are solved and you should check Product issues tab. 
8. Refresh the page (Now Account issue is selected) after reload Account issues are selected. 
9. Delete from the URL the searchQuery `&issueType=account`. Enter to the page.
10. Product issues is selected by default since Account issues has no issues.
11. Delete your product issues.  `DELETE FROM wp_gla_merchant_issues WHERE type = 'product';` 
12. There is no Issues block anymore. And you see No issues to resolve 🎉 in the summary.

### Changelog entry

> Tweak - Informative message when here are no issues to solve
> Tweak - Issue type with issues is active by default

PR #1438, opened 2022-04-21
Get started page UX improvements: add benefits card

ianlin:

Part of #1392, this PR adds the benefits card to the new get started page.

Design for desktop: eQ9O4m2flzrcAiDMXkOW0m-fi-7350%3A121106
Design for mobile: eQ9O4m2flzrcAiDMXkOW0m-fi-7476%3A122821

#### Desktop

Screenshot 2022-04-21 at 15 00 21

#### Mobile

Screenshot 2022-04-21 at 15 00 32

2. Check the second card of the get started page, the UI should align with the design.

### Changelog entry

> Update - Add the benefits card for UX improvements on get started page.


js/src/get-started-page/benefits-card/index.scss
```diff
@@ -0,0 +1,57 @@
...
+
+				&:last-child {
+					margin-bottom: unset;
+				}
+			}
+
+			&__title::after {
+				content: " 50%";
+				font-weight: bold;
+			}
```
eason9487: ❓ 
It should avoid composing i18n string programmatically if it's possible. To apply style onto i18n strings, it can be done by using the `createInterpolateElement` function.

Related discussion: https://github.com/woocommerce/google-listings-and-ads/pull/1196#discussion_r791352611

js/src/get-started-page/benefits-card/index.scss
```diff
@@ -0,0 +1,57 @@
...
+			padding: $grid-unit-50;
+			max-width: 51.5%;
+
+			> p {
+				margin-bottom: $grid-unit-20;
+				padding: 0 $grid-unit-10;
+
+				&:last-child {
+					margin-bottom: unset;
+				}
```
eason9487: 💅 
- The `margin-bottom` between `<p>` could be achieved by setting `gap`, which works with Flexbox, on the container layer.
- The `padding` could be removed since the width of this layout is dynamic, so it's okay not to accurately align to the same visual design.

js/src/get-started-page/benefits-card/index.scss
```diff
@@ -0,0 +1,57 @@
+.gla-get-started-benefits-card {
+	// Adjust <CardBody> imported from @wordpress/components.
+	// Repeat selector to make it higher priority.
+	.components-card__body.components-card__body {
+		display: flex;
+		flex-direction: row;
```
eason9487: 💅  The default value of `flex-direction` is the same.

js/src/get-started-page/benefits-card/index.js
```diff
@@ -0,0 +1,62 @@
...
+						'Google Listings & Ads Benefits',
+						'google-listings-and-ads'
+					) }
+					width="100%"
+					height="100%"
+				/>
+				<FlexItem>
+					<Text
+						variant="title.medium"
+						className="components-flex__item__title"
```
eason9487: ❓ 
It looks a little confusing by using the core component class name to extend the sub-name. Probably extending from `gla-get-started-benefits-card` would be nicer.

js/src/get-started-page/benefits-card/index.scss
```diff
@@ -0,0 +1,57 @@
...
+		@media (max-width: $break-small) {
+			flex-direction: column;
+		}
+
+		.components-flex__item {
+			display: flex;
+			flex-direction: column;
+			justify-content: center;
+			padding: $grid-unit-50;
+			max-width: 51.5%;
```
eason9487: ❔ The layout looks the same without setting `max-width`. What is `max-width` supposed to handle?

js/src/get-started-page/benefits-card/index.scss
ianlin: Thanks and TIL! Updated in b90aaae.

js/src/get-started-page/benefits-card/index.scss
ianlin: Thanks and TIL for Flexbox's `gap`, updated in 22271b1.

> since the width of this layout is dynamic, so it's okay not to accurately align to the same visual design.

Gotcha, I was desperately trying to make it looks like the design 🤣

js/src/get-started-page/benefits-card/index.scss
ianlin: Thanks, updated in 4a4b4b7.

js/src/get-started-page/benefits-card/index.js
ianlin: Gotcha, it makes more sense. Updated in 2c531e5 and thanks!

js/src/get-started-page/benefits-card/index.scss
ianlin: Ah, I think it was previously we used a SVG file for the image and from the design the image should be slightly smaller than the text on the right hand side. Thanks for spotting it, updated in 9181eb1.

ianlin:
Hey @eason9487 thanks for the review! I've added some commits based on the review comments.

js/src/get-started-page/benefits-card/index.scss
eason9487: > I was desperately trying to make it looks like the design

In the case of dynamic content with a dynamic layout, using only CSS could easily run into its limits and not achieve pixel-perfect results. 🙈

eason9487 (approved):
Thanks for the adjustments! Testing well, and LGTM.

PR #1442, opened 2022-04-26
Improve the interactive UI for a dropdown inside a modal

eason9487:

Address [two addtional UI problems](https://github.com/woocommerce/google-listings-and-ads/issues/1356#issuecomment-1109239892) of #1356 

- Add the `overflowY` prop to \<AppModal\> component for specifying the `overflow-y` of the modal.
- Add the `onDropdownVisibilityChange` callback prop to \<TreeSelectControl\> component for getting the visible state of the dropdown options.
- Make the viewport of all shipping modals not cover the dropdown options.
- Avoid closing the shipping modals by clicking on overlay or pressing ESC when the inside dropdown is opened
- 💡  Extra changes: remove unneeded `classNames` and scss files for two modals:
   1. js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/countries-time-input/edit-time-button/edit-time-modal/index.js
   2.  js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/countries-time-input/edit-time-button/edit-time-modal/index.js

#### 📷  The dropdown is not covered by the viewport of modal.

image

#### 📹  Improved interaction UI for a dropdown inside a modal.

https://user-images.githubusercontent.com/17420811/165212064-4483c2eb-d015-415b-aa2e-3b1f83ec2c10.mp4

1. Go to step 2 of editing the free listings.
1. Check every country selectors of all adding/editing shipping modals (rate, time, and minimum order) with the following steps.
   1. Open a modal and open the inside dropdown options of the tree selector.
   1. Expend some options to see if the dropdown is no longer covered by the viewport of modal.
   1. After clicking on the overlay, which is the outside area of the modal, it should close the opened dropdown only.
   1. After clicking on the overlay again, it should close the modal.
   1. Open that modal and the dropdown again.
   1. After pressing the ESC key, it should close the opened dropdown only.
   1. After pressing the ESC key again, it should close the modal.
1. Go to step 3 of the onboarding setup.
1. Check the country selectors of the adding/editing shipping **time** modals by the same steps in step 2.

### DIFF
--- a/js/src/components/app-modal/index.js
+++ b/js/src/components/app-modal/index.js
@@ -10,6 +10,16 @@ import classnames from 'classnames';
 import './index.scss';
 
 const overflowYStyleName = {
+	/**
+	 * In @wordpress/components 14.2.0, the overflow of Modal container will be
+	 * changed from `auto` to `hidden`, and the change also adds `auto` to Modal body.
+	 * Here using `auto` as the default overflow to make it have a stable overflow style
+	 * across these versions.
+	 *
+	 * References:
+	 * - https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4012.0.9/packages/components/src/modal/style.scss#L29
+	 * - https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4014.2.0/packages/components/src/modal/style.scss#L24
+	 */
 	auto: 'app-modal__styled--overflow-y-auto',
 	visible: 'app-modal__styled--overflow-y-visible',
 };
--- a/js/src/components/app-modal/index.js
+++ b/js/src/components/app-modal/index.js
@@ -9,9 +9,16 @@ import classnames from 'classnames';
  */
 import './index.scss';
 
-const AppModal = ( props ) => {
-	const { buttons = [], className, children, ...rest } = props;
-
+/**
+ * Renders a Modal component with additional commonly used features.
+ *
+ * @param {Object} props React props.
+ * @param {string} [props.className] Additional CSS class name to be appended.
+ * @param {Array<JSX.Element>} [props.buttons] Buttons to be rendered at the modal footer.
+ * @param {JSX.Element} props.children Content to be rendered.
+ * @param {Object} props.rest Props to be forwarded to Modal.
+ */
+const AppModal = ( { className, buttons = [], children, ...rest } ) => {
 	return (
 		<Modal className={ classnames( 'app-modal', className ) } { ...rest }>
 			{ children }
--- a/js/src/components/app-modal/index.js
+++ b/js/src/components/app-modal/index.js
@@ -9,19 +9,19 @@ import classnames from 'classnames';
  */
 import './index.scss';
 
-const overflowYStyleName = {
+const overflowStyleName = {
 	/**
 	 * In @wordpress/components 14.2.0, the overflow of Modal container will be
 	 * changed from `auto` to `hidden`, and the change also adds `auto` to Modal body.
-	 * Here using `auto` as the default overflow to make it have a stable overflow style
-	 * across these versions.
+	 * The `auto` is set as the default overflow in this component's CSS to make it
+	 * have a stable overflow style across these versions.
 	 *
 	 * References:
 	 * - https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4012.0.9/packages/components/src/modal/style.scss#L29
 	 * - https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4014.2.0/packages/components/src/modal/style.scss#L24
 	 */
-	auto: 'app-modal__styled--overflow-y-auto',
-	visible: 'app-modal__styled--overflow-y-visible',
+	auto: false,
+	visible: 'app-modal__styled--overflow-visible',
 };
 
 /**
@@ -29,21 +29,21 @@ const overflowYStyleName = {
  *
  * @param {Object} props React props.
  * @param {string} [props.className] Additional CSS class name to be appended.
- * @param {'auto'|'visible'} [props.overflowY='auto'] Y-axis overflow style of modal.
+ * @param {'auto'|'visible'} [props.overflow='auto'] Overflow style of modal.
  * @param {Array<JSX.Element>} [props.buttons] Buttons to be rendered at the modal footer.
  * @param {JSX.Element} props.children Content to be rendered.
  * @param {Object} props.rest Props to be forwarded to Modal.
  */
 const AppModal = ( {
 	className,
-	overflowY = 'auto',
+	overflow = 'auto',
 	buttons = [],
 	children,
 	...rest
 } ) => {
 	const modalClassName = classnames(
 		'app-modal',
-		overflowYStyleName[ overflowY ],
+		overflowStyleName[ overflow ],
 		className
 	);
 
--- a/js/src/components/app-modal/index.js
+++ b/js/src/components/app-modal/index.js
@@ -9,18 +9,36 @@ import classnames from 'classnames';
  */
 import './index.scss';
 
+const overflowYStyleName = {
+	auto: 'app-modal__styled--overflow-y-auto',
+	visible: 'app-modal__styled--overflow-y-visible',
+};
+
 /**
  * Renders a Modal component with additional commonly used features.
  *
  * @param {Object} props React props.
  * @param {string} [props.className] Additional CSS class name to be appended.
+ * @param {'auto'|'visible'} [props.overflowY='auto'] Y-axis overflow style of modal.
  * @param {Array<JSX.Element>} [props.buttons] Buttons to be rendered at the modal footer.
  * @param {JSX.Element} props.children Content to be rendered.
  * @param {Object} props.rest Props to be forwarded to Modal.
  */
-const AppModal = ( { className, buttons = [], children, ...rest } ) => {
+const AppModal = ( {
+	className,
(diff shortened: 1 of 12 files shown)


js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/countries-time-input/edit-time-button/edit-time-modal/index.js
```diff
@@ -80,7 +81,9 @@ const EditTimeModal = ( props ) => {
 
 				return (
 					<AppModal
-						className="gla-edit-time-modal"
+						overflowY="visible"
+						shouldCloseOnEsc={ ! dropdownVisible }
+						shouldCloseOnClickOutside={ ! dropdownVisible }
```
puntope: 💅  What about just a property like `dropdownVisible ={ dropdownVisible }` ??

My concerns with the current implementation are that we need to set up 3 props in order to make the modal support the dropdown.

- Set overflowY="visible" (is only "visible" if there is a dropdown)
- Set shouldCloseOnEsc={ ! dropdownVisible } (related to dropdown)
- Set shouldCloseOnClickOutside={ ! dropdownVisible } (Related to dropdown)

Hence, having that `withDropdown={ dropdownVisible }` or creating a wrapper component `AppWithDropdown` will be better in terms of code reusability. 

```
const AppModal = ( {
	className,
	dropdownVisible,
	buttons = [],
	children,
	...rest
} ) => {
	const modalClassName = classnames(
		'app-modal',
		dropdownVisible && `app-modal__styled--overflow-y-visible`,
		className
	);

    <Modal 
          className={ modalClassName } 
          shouldCloseOnEsc={ ! dropdownVisible }
	      shouldCloseOnClickOutside={ ! dropdownVisible }>
```

js/src/components/app-modal/index.js
```diff
@@ -9,11 +9,36 @@ import classnames from 'classnames';
  */
 import './index.scss';
 
-const AppModal = ( props ) => {
-	const { buttons = [], className, children, ...rest } = props;
+const overflowYStyleName = {
+	auto: 'app-modal__styled--overflow-y-auto',
```
puntope: ❓  Why is the auto class necessary?

puntope (commented):
Thanks for the changes, I left some feedback

🚧  / ❓ 

One thing I notice was that the inner class `.components-modal__content` has `overflow:auto` set, hence it didn't work for me.

See video:


https://user-images.githubusercontent.com/5908855/165267886-8309f462-1cd3-4b07-b466-b944279bb7e3.mov

js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/countries-time-input/edit-time-button/edit-time-modal/index.js
eason9487: > What about just a property like `dropdownVisible ={ dropdownVisible }` ??

Usually, the more basic the component, the less interconnectivity between its props internally, and these interconnections are more appropriate for higher-level (shared) components. 

The AppModal is a basic component for generic uses. Using a `dropdownVisible` prop to control the other three props will reduce flexibility. For example, a modal that doesn't allow to be closed by clicking on outside area or pressing ESC, regardless of whether the dropdown is displayed or not.

> or creating a wrapper component `AppWithDropdown` will be better in terms of code reusability.

Adding another layer of wrapper components might help the reusability, but I lean toward reducing the code conflicts to the in-progress adjustments of shipping modals.

js/src/components/app-modal/index.js
eason9487: The `overflow` of `.components-modal__frame` will be changed from `auto` to `hidden` in `@wordpress/components` **14.2.0**. I intended to make it have a stable overflow style across these versions. I added a code comment to clarify the intention in https://github.com/woocommerce/google-listings-and-ads/pull/1442/commits/1684e0e2469dbbc203f75214f137796cc1cb9f2c.

eason9487:

> One thing I notice was that the inner class `.components-modal__content` has `overflow:auto` set, hence it didn't work for me.

Referring to some context in the https://github.com/woocommerce/google-listings-and-ads/pull/1442#discussion_r858568993, I noticed that `overflow` change would happen on Modal container, but I didn't find it would also happen on Modal body. Thanks for catching this. 🙏 

I made an adjustment in https://github.com/woocommerce/google-listings-and-ads/commit/1684e0e2469dbbc203f75214f137796cc1cb9f2c tried to fix it, but I have only used a simulation to reproduce the problem you encountered. Could you help me with confirming if the problem is fixed on your end?

puntope:
> Could you help me with confirming if the problem is fixed on your end?

Confirmed this is fixed! 🙏 🆗

js/src/components/app-modal/index.js
puntope: ❓ 

Yes, thats good, but I still find the extra class unnecessary, why not set ` overflow-Y: auto` to the modal without any need of that `app-modal__styled--overflow-y-auto`

js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/countries-time-input/edit-time-button/edit-time-modal/index.js
puntope: I disagree that the current implementation reduces complexity. Adding  `shouldCloseOnEsc` &   `shouldCloseOnClickOutside` in the AppModal component based on the dropdown property would reduce the complexity for the implementation since it would needed to send 1 parameter instead of 3. Besides the fact that adding that those parameters seems for me potentially forgettable. 

> The AppModal is a basic component for generic uses. Using a dropdownVisible prop to control the other three props will reduce flexibility. For example, a modal that doesn't allow to be closed by clicking on outside area or pressing ESC, regardless of whether the dropdown is displayed or not.

Adding  `shouldCloseOnEsc` &   `shouldCloseOnClickOutside` in the AppModal component based on the dropdown property is not reducing flexibility since we can use also the ...rest feature and add directly `shouldCloseOnEsc` &   `shouldCloseOnClickOutside` as before.

Saying that and since its not a blocking issue, I rely on your criteria (or further PRs) to consider that or not

puntope (approved):
✅ LGTM

Some non blocking comments regarding the used approach and potentially unnecessary classname

js/src/components/app-modal/index.js
eason9487: Since one of `overflow` axes is set to `visible` alone, it has a hidden rule in CSS spec that if the other axis is not `visible`, then the `visible` one is interpreted as `auto`.

I originally wanted to use CSS's own default value as much as possible and specify the Y-axis overflow only, but this also made the above rule implicit.

Considering there are code comments added in the last round review, and it's good to make it a more explicit way, I changed the implementation in https://github.com/woocommerce/google-listings-and-ads/pull/1442/commits/28935f133511ea024a8844ac13a053ff47720102.

js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/countries-time-input/edit-time-button/edit-time-modal/index.js
eason9487: > I disagree that the current implementation reduces complexity.

I meant that the flexibility would be reduced if using a prop to control the other three props internally and make interconnections on them. I supposed I didn't mention which complexity is lower. If I were to evaluate the complexity, I don't think the difference is significant between both.

> Adding `shouldCloseOnEsc` & `shouldCloseOnClickOutside` in the AppModal component based on the dropdown property would reduce the complexity for the implementation since it would needed to send 1 parameter instead of 3. Besides the fact that adding that those parameters seems for me potentially forgettable.

> Adding `shouldCloseOnEsc` & `shouldCloseOnClickOutside` in the AppModal component based on the dropdown property is not reducing flexibility since we can use also the ...rest feature and add directly `shouldCloseOnEsc` & `shouldCloseOnClickOutside` as before.

Surely this could make it less needed props for the cases in this PR, but I meant it doesn't make the opposite use case easier. If a floating block is showing and inside a modal, but the scenario doesn't need to prevent that modal from being closed by pressing ESC or by clicking on outside, then it still needs to specify the AppModal like:
```js
<AppModal
  dropdownVisible={ dropdownVisible }
  shouldCloseOnEsc={ true }
  shouldCloseOnClickOutside={ true }
>
```

The need of adding props like `shouldCloseOnEsc` and `shouldCloseOnClickOutside` is still the same as "potentially forgettable". Preserving the independence of their logic is a way to retain flexibility for a basic level component.

---

But considering the current instances are the same use case, I would say it's still fine to use a single prop approach you suggested. However, I will have two more concerns with that approach:
1. The prop name `dropdownVisible` might not be ideal. Suppose the AppModal should not need to be too specific and literal, such as **dropdown**. It would be more appropriate to use a name that describes the intent, otherwise coupling in the naming is higher.
2. Since the given single prop is a dynamic state, and its rendering is not updated in the same tick as the updating of Modal `className`, this causes visible flickering on slower computer hardware.
   Kapture 2022-04-27 at 12 19 17

I'd like to know if you have a better idea to avoid these concerns.

PR #1448, opened 2022-04-27
GLA Request Review - Adding time in Cooldown period date

puntope:

After implementing the cooldown period feature I was expecting to have today (27 April) my cooldown period finished.

However, when I accessed at 9am my account was still in Cooldown period and the notice was showing that the end of cooldown period is 27 April. Checking the cooldown timestamp seems like google take also in consideration the time, that means we should add as well the time together with the date to avoid confusion to the user.

In brief, this PR will add `27 April, 2022, 16:40pm` instead of just `27 April, 2022`

### Screenshots:

Screenshot 2022-04-27 at 11 58 39

```
$response =  [
				                 'freeListingsProgram' => [
					                 'status' => 200,
					                 'data'   => [
						                 'regionStatuses' => [
							                 [
								                 'regionCodes'                      => [ 'US' ],
								                 'eligibilityStatus'                => 'DISAPPROVED',
								                 'reviewEligibilityStatus'          => 'INELIGIBLE',
								                 'reviewIneligibilityReasonDetails' => [
									                 'cooldownTime' => "2022-04-27T10:58:51Z" // 27/04/2022
								                 ]
							                 ],
							                 [
								                 'regionCodes'                      => [ 'NL' ],
								                 'eligibilityStatus'                => 'DISAPPROVED',
								                 'reviewEligibilityStatus'          => 'INELIGIBLE',
								                 'reviewIneligibilityReasonDetails' => [
									                 'cooldownTime' => "2022-04-25T10:58:51Z" // 25/04/2022
								                 ]
							                 ],
							                 [
								                 'regionCodes'             => [ 'IT' ],
								                 'eligibilityStatus'       => 'DISAPPROVED',
								                 'reviewEligibilityStatus' => 'ELIGIBLE',
							                 ],
						                 ]
					                 ]
				                 ]
			                 ]
```
1. Checkout this PR
2. Go to `src/API/Site/Controllers/MerchantCenter/RequestReviewController.php::get_review_read_callback()` and hardcode the response provided above to return cooldown period
3. Go to Product feed. Check that the notice shows `Your account is under cool down period. You can request a new review on April 27, 2022 10.58 am.` (notice the time is in GMT so it could vary depending on your time zone) you can use https://www.epochconverter.com/ to get the time for your zone.

### DIFF
--- a/js/src/product-feed/review-request/review-request-notice.js
+++ b/js/src/product-feed/review-request/review-request-notice.js
@@ -35,7 +35,10 @@ const ReviewRequestNotice = ( {
 				'Your account is under cool down period. You can request a new review on %s.',
 				'google-listings-and-ads'
 			),
-			formatDate( glaData.dateFormat, new Date( account.cooldown ) )
+			formatDate(
+				`${ glaData.dateFormat }, ${ glaData.timeFormat }`,
+				new Date( account.cooldown )
+			)
 		);
 
 	return (
--- a/js/src/product-feed/review-request/review-request-notice.test.js
+++ b/js/src/product-feed/review-request/review-request-notice.test.js
@@ -50,7 +50,7 @@ describe( 'Request Review Notice', () => {
 		);
 		expect(
 			queryByText(
-				'Your account is under cool down period. You can request a new review on April 27, 2022.'
+				'Your account is under cool down period. You can request a new review on April 27, 2022, 12:11 pm.'
 			)
 		).toBeTruthy();
 
--- a/js/src/product-feed/review-request/review-request-notice.test.js
+++ b/js/src/product-feed/review-request/review-request-notice.test.js
@@ -2,12 +2,12 @@
  * External dependencies
  */
 import { fireEvent, render } from '@testing-library/react';
+import { format as formatDate } from '@wordpress/date';
 
 /**
  * Internal dependencies
  */
 import ReviewRequestNotice from '.~/product-feed/review-request/review-request-notice';
-import { format as formatDate } from '@wordpress/date';
 
 describe( 'Request Review Notice', () => {
 	it.each( [ 'DISAPPROVED', 'WARNING' ] )(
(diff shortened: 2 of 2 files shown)


js/src/product-feed/review-request/review-request-notice.js
```diff
@@ -35,7 +35,10 @@ const ReviewRequestNotice = ( {
 				'Your account is under cool down period. You can request a new review on %s.',
 				'google-listings-and-ads'
 			),
-			formatDate( glaData.dateFormat, new Date( account.cooldown ) )
+			formatDate(
+				`${ glaData.dateFormat }, ${ glaData.timeFormat }`,
+				new Date( account.cooldown )
+			)
```
eason9487, line 41: 💅 
I think users hardly ever save the format of date or time in WC Settings as blank, but WC Settings still allows this.

image

And it could cause the formatting result like:
image

Probably this code could share the same logic in
https://github.com/woocommerce/google-listings-and-ads/blob/0ebe02d22d1e9621ef4e96b0aa33e8867b0da364/js/src/product-feed/product-statistics/status-box/sync-status.js#L63-L70

eason9487 (approved):
Left a comment about logic reuse. Although there is a minor problem, it should rarely occur and would not cause page crashes. So it's not a blocker for this PR.

js/src/product-feed/review-request/review-request-notice.js
puntope: To be adjusted in a separated PR 🙏  Thanks.

puntope: I'm also thinking about if we should allow empty time in GLA. In my mind having blank time is some kinda of patch to don't see maybe the time in the  post date or some kind of dates in orders and products?

So I'm wondering if we should just apply a default format (like h:s) instead of just removing the comma and allowing the user to remove the time that can create confusion to HE in case some customer says the GLA notice says 27 April and he is still unavailable. If I empathise with the HE seems hard for me to think WC settings could be the reason creating the issue.  

What do you think?

eason9487: Usually, I would say it should respect the users' WC Settings. If they set it to blank, it could be considered they want to. But your point is thorough and both the **date** and **time** are important in this use case. So, I think it would be nicer to have fallbacks for the blank formats of WC Settings here.

PR #1449, opened 2022-04-27
Get started page UX improvements: add customer quotes card

ianlin:

Part of #1392, this PR adds the customer quotes card to the new get started page.

Design for desktop: eQ9O4m2flzrcAiDMXkOW0m-fi-7350%3A121231
Design for mobile: eQ9O4m2flzrcAiDMXkOW0m-fi-7476%3A123025

#### Desktop

Screenshot 2022-04-27 at 18 47 38

#### Mobile

Screenshot 2022-04-27 at 18 48 29

2. Check the fourth card of the get started page, the UI should align with the design.

### Changelog entry

> Update - Add the customer quotes card for UX improvements on get started page.


js/src/get-started-page/index.js
```diff
@@ -14,6 +15,7 @@ const GetStartedPage = () => {
 			<GetStartedWithVideoCard />
 			<BenefitsCard />
 			<FeaturesCard />
+			<CaseStudiesCard />
```
puntope: 💅 

CaseStudies? In this case I guess it will be more accurate to use CustomerQuotes or Testimonials.

js/src/get-started-page/case-studies-card/index.scss
```diff
@@ -0,0 +1,53 @@
...
+	// Repeat selector to make it higher priority.
+	.components-card__header.components-card__header {
+		flex-direction: column;
+		padding: $grid-unit * 5.75 $grid-unit * 14 $grid-unit-50;
+		text-align: center;
+
+		@media (max-width: $break-small) {
+			padding: $grid-unit-30;
+		}
+	}
```
puntope: 💅 / 🚧 

Inserting this selector inside `.components-flex` also works and we can avoid that double selector patch.

```
.components-flex {
		gap: $grid-unit * 3.5;
		padding: 0 $grid-unit * 7.75 $grid-unit * 5.75;

		&.components-card__header {
			flex-direction: column;
			... 
```

js/src/get-started-page/case-studies-card/index.js
```diff
@@ -0,0 +1,75 @@
...
+					variant="title.medium"
+				>
+					{ __(
+						'21,000+ WooCommerce store owners like you already list products on Google for free',
+						'google-listings-and-ads'
+					) }
+				</Text>
+			</CardHeader>
+			<Flex gap={ 0 }>
+				<ContentBlock
```
puntope: 💅  I'm not convinced neither by ContentBlock as name for this. Maybe `TestimonialItem` or `Quote` or `Testimonial`?

js/src/get-started-page/case-studies-card/index.scss
```diff
@@ -0,0 +1,53 @@
...
+
+	.components-flex {
+		gap: $grid-unit * 3.5;
+		padding: 0 $grid-unit * 7.75 $grid-unit * 5.75;
```
puntope: 💅 / ❓ 

Not sure if this is already discussed but maybe we can create some SCSS variables for the fonts and some spacing is being reused all over `get-started-page` directory.

puntope (commented):
Looking good ✅ thanks for working on this. 

I'd love to discuss about that "Repeat selector to make it higher priority." hack.  I guess there here are more standard ways to do so.

Would be nice also to evaluate the use and creation of variables for the common css values.

Rest of my comments are just opinions about naming.

js/src/get-started-page/index.js
ianlin: Both terms of `case studies` and `customer quotes` appear in the design brief document in this post: pcTzPl-i7-p2#design-requirements. But I agree that `customer quotes` is more straightforward, updated in ca06c61, thanks!

js/src/get-started-page/case-studies-card/index.js
ianlin: Thanks for the suggestion! Updated in 6404822.

js/src/get-started-page/case-studies-card/index.scss
ianlin: Thanks for the suggestion! I added a commit 13e1ac1 to extract some shared variables for font sizes and line heights as they've got more common uses. 

There are too many different sizes of the spacing that I've used in get started page based on the design, for example:

```
10px;
20px;
28px;
40px;
44px;
46px;
50px;
55px;
62px;
88px;
112px;
```

It's a bit tricky to make them as variables tho.

js/src/get-started-page/case-studies-card/index.scss
ianlin: Thanks for it! Using this form is clearer, updated in c673732. Also updated for other cards in fd58db1.

Note that it works for `CardHeader` since it has `.components-flex` class, while `CardBody` doesn't have that class so in `js/src/get-started-page/benefits-card/index.scss` I still use the repeating selectors hack in order to override the `CardBody`'s original styles.

ianlin:
Hey @puntope, thanks for the review. I pushed some commits based on your suggestions and it's ready for another round of review.

puntope (commented):
Thanks for the adjustments it looks accurate with design and cleaner now.

Sorry. I want to mention 2 details more before approve. 1 of them seems to be regarding previous PRs since its related to other section. 

- I saw that in my tablet in vertical some resolutions the image on benefits-card is not covering the container. Wrapping the image in flex and playing with object-fit: cover maybe solves it.

Screenshot 2022-04-28 at 18 31 46

- In the quotes, if one paragraph is higher than the other (because resolution or content length) it looks unbalanced

Screenshot 2022-04-28 at 18 33 43

Maybe `align-items: stretch` ??

Screenshot 2022-04-28 at 18 36 46

ianlin:
Thanks @puntope! Miss the tablet's resolution, noted! I pushed two commits a369c68 and 0011c43 to address the problem.

puntope:
> commits https://github.com/woocommerce/google-listings-and-ads/commit/a369c68903f1c6975b19f080dabe02ad7082d4b7

Thanks for the change, it is no space anymore in the container, however now the image gets stretched. Adding `object-fit: cover` to the img element may solve this issue.

Screenshot 2022-04-29 at 11 02 36


> and https://github.com/woocommerce/google-listings-and-ads/commit/0011c433def228fb8768c00c4e3d967282b8ac1a

it works! 👌 Thanks

ianlin:
@puntope 

I did try `object-fit: cover` and it looks like it's being stretched, while it looks normal when `object-fit: cover` is not added.

#### With `object-fit: cover`

Screenshot 2022-04-29 at 18 08 33

#### Without `object-fit: cover`

Screenshot 2022-04-29 at 18 08 18

But I was using the dev tool's device toolbar to check the result, it might not be accurate as using the real tablet so I still added `object-fit: cover` in 8f321ed as it doesn't look harm if we apply it.

Thanks for the detailed review!

puntope (approved):
✅  APPROVED good work! 

Thanks for all the adjustments. The comment about `_shared.scss` here https://github.com/woocommerce/google-listings-and-ads/pull/1460#discussion_r861573029

I guess will be addressed in future PR.

Issue #861, opened 2021-07-06
Product Sync batches can be empty


mikkamp:
### Describe the bug:
When we sync a lot of products we divide it into batches (default 100) per batch. Each batch can be a mix of single products + variations. In our [original query](https://github.com/woocommerce/google-listings-and-ads/blob/1.1.0/src/Product/ProductRepository.php#L204-L217) to get the included ID's we exclude any products that are set to `dont-sync-and-show`, as well as exclude any other products which are not ready for syncing. Once we get a batch of products we run them through an [additional filter](https://github.com/woocommerce/google-listings-and-ads/blob/1.1.0/src/Product/ProductRepository.php#L175-L199) to make sure they are fully sync ready. When this filtering is done it can choose to exclude products based on additional criteria. In theory it will just make the batch size smaller, but in theory it's possible to make a batch size reduce to zero which could cause it to trigger `handle_complete` and stop processing subsequent batch jobs.

One way we can see this happening is if we have a large amount of variable products which we set to `dont-sync-and-show`. This value is only saved for the parent product. So when we query simple products and variations we will also include the variations whose parent is set to `dont-sync-and-show`. If we happen to have 4 variable products in a row with 25 variations each that can reduce a complete batch to empty and triggered it to stop processing subsequent batches.

I'm not sure whether it would warrant saving the visibility status for each variable product, or whether we would be better off detecting the total amount of batches in a different way. Not as common, but in theory it could also be reproduced with a large amount of bundled products which are not eligible.

### Steps to reproduce
It's a little hard to reproduce since we don't determine the order in which the products are synced, but here is the general steps to reproduce it:
1. It's easier to reproduce with a smaller batch size, such as:
```
add_filter(
	'woocommerce_gla_batched_job_size',
	function( $size ) {
		return 10;
	}
);
```
2. Have a large amount of variable products set to `dont-sync-and-show`
3. Since we don't determine the order the variable products would have to come before any later created products
4. When syncing all products observe the batches and notice that later batches can be missing once a larger variable product is skipped from syncing

mikkamp:
Not only does an empty batch trigger a `handle_complete`, if any of the products are filtered / removed from the batch then the next batch won't be scheduled.

This is because of the check [here](https://github.com/woocommerce/google-listings-and-ads/blob/1.1.0/src/Jobs/AbstractBatchedActionSchedulerJob.php#L78):

```
if ( count( $items ) >= $this->get_batch_size() )
```

[...]

PR #1451, opened 2022-04-27
Adding update all product test to catch the bug when scheduled job wa…

### LINKED ISSUE #1450: Products are not synced to Google Merchant Center.
### Describe the bug:

The scheduled job is expecting an array of ids, but got an array of WC_Product objects.
https://github.com/woocommerce/google-listings-and-ads/blob/422de5e11115f6bcb2090309bff239f2bd9508e0/src/Jobs/UpdateAllProducts.php#L110

The bug was introduced with this PR https://github.com/woocommerce/google-listings-and-ads/pull/1361 and tests did not catch it. 

### Expected behavior:

The scheduled job to receive product ids instead of objects.

message-dimke:
With PR https://github.com/woocommerce/google-listings-and-ads/pull/1361 an issue was introduced. Scheduled job began to receive item objects instead of object ids and product stopped syncing.

Test did not catch this behaviour and were missing for update all products.

Adding corresponding tests and fixing the code so the jobs appear working properly again.

Closes #1450.

1. Disconnect and connect again creating a new account in Google.
2. You will see in the Product Feed the Status: “Syncing”
3. After a few minutes, I reload the page and I get the Status “Automatically synced to GoogleLast updated: April 26, 2022, 4:22 pm, containing 0 products”

screenshot-2022-04-26-at-16 16 06

screenshot-2022-04-26-at-16 16 10

4. Apply the fix from PR and see products syncing.

### Changelog entry

> Fix - Update all products job syncing products

### DIFF
--- a/tests/Unit/Jobs/UpdateAllProductsTest.php
+++ b/tests/Unit/Jobs/UpdateAllProductsTest.php
@@ -5,6 +5,8 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Jobs;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\ActionScheduler\ActionScheduler;
 use Automattic\WooCommerce\GoogleListingsAndAds\ActionScheduler\ActionSchedulerInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\BatchProductEntry;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\BatchProductResponse;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\ActionSchedulerJobMonitor;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\UpdateAllProducts;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
@@ -76,7 +78,7 @@ class UpdateAllProductsTest extends UnitTest {
 		$this->assertEquals( self::JOB_NAME, $this->job->get_name() );
 	}
 
-	public function test_schedule_schedules_a_single_batched_job_with_items() {
+	public function test_single_batched_job_with_items() {
 		$filtered_product_list = new FilteredProductList( $this->generate_simple_products_set( 4 ), 4 );
 
 		$this->action_scheduler->expects( $this->exactly( 2 ) )
@@ -95,7 +97,7 @@ class UpdateAllProductsTest extends UnitTest {
 		do_action( self::CREATE_BATCH_HOOK, 1 );
 	}
 
-	public function test_schedule_schedules_a_single_batched_job_with_no_items() {
+	public function test_single_batched_job_with_no_items() {
 		$filtered_product_list = new FilteredProductList( [], 0 );
 		$this->action_scheduler->expects( $this->once() )
 			->method( 'schedule_immediate' )
@@ -109,7 +111,7 @@ class UpdateAllProductsTest extends UnitTest {
 		do_action( self::CREATE_BATCH_HOOK, 1 );
 	}
 
-	public function test_schedule_schedules_two_batches_of_items_and_empty_one() {
+	public function test_multiple_batches_of_items_and_empty_one() {
 
 		/* adding a filter to make batch smaller for testing */
 		add_filter(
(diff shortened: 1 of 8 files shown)


tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -0,0 +1,184 @@
...
+	public function setUp(): void {
+		parent::setUp();
+
+		$this->action_scheduler      = $this->createMock( ActionSchedulerInterface::class );
+		$this->monitor               = $this->createMock( ActionSchedulerJobMonitor::class );
+		$this->product_syncer        = $this->createMock( ProductSyncer::class );
+		$this->product_repository    = $this->createMock( ProductRepository::class );
+		$this->product_helper        = $this->createMock( BatchProductHelper::class );
+		$this->merchant_center       = $this->createMock( MerchantCenterService::class );
+		$this->job                   = new UpdateAllProducts(
```
mikkamp: Just a small note these lines are indented too far.

tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -0,0 +1,184 @@
...
+			->with( $filtered_product_list->get() );
+		$this->product_syncer
+			->method( 'update_by_batch_requests' )
+			->willReturn( $response );
+
+		do_action( self::PROCESS_ITEM_HOOK, $filtered_product_list->get_product_ids() );
+	}
+
+	/**
+	 * Helper function to crate array filled with product mocks for test purposes
```
mikkamp: Small typo 'create array'

tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -0,0 +1,184 @@
...
+	 *
+	 * @param int $number Number of elements to fill an array with.
+	 * @return WC_Product[]
+	 */
+	private function generate_simple_products_set( int $number ) {
```
mikkamp: Can we move this to the ProductTrait and indicate in the name that it creates a set of mocked products. Alternatively we also have `create_simple_product_set` in ProductTrait which can be used instead, although it will create "real" products instead of mocked ones.

tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -0,0 +1,184 @@
...
+		$batch_c = new FilteredProductList( [], 0 );
+
+		$this->action_scheduler->expects( $this->exactly( 5 ) )
+			->method( 'schedule_immediate' )
+			->withConsecutive(
+				array( self::CREATE_BATCH_HOOK, [ 1 ] ),
+				array( self::PROCESS_ITEM_HOOK, [ $batch_a->get_product_ids() ] ),
+				array( self::CREATE_BATCH_HOOK, [ 2 ] ),
+				array( self::PROCESS_ITEM_HOOK, [ $batch_b->get_product_ids() ] ),
+				array( self::CREATE_BATCH_HOOK, [ 3 ] ),
```
mikkamp: can we consistently use the short array syntax. There are several other locations in this file where the long array syntax is used.

tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -0,0 +1,184 @@
...
+		$this->job->schedule();
+
+		do_action( self::CREATE_BATCH_HOOK, 1 );
+		do_action( self::CREATE_BATCH_HOOK, 2 );
+		do_action( self::CREATE_BATCH_HOOK, 3 );
+	}
+
+	public function test_process_item() {
+		$filtered_product_list = new FilteredProductList( $this->generate_simple_products_set( 1 ), 1 );
+		$product_id            = $filtered_product_list->get_product_ids()[0];
```
mikkamp: This works fine here, but just wanted to mention that we'd need to be careful with this kind of pattern outside of tests. `array_map` preserves the array keys if it's passed only one array, so array key 0 might not always exist especially if some products are filtered out of the list.

mikkamp (commented):
Thanks for the code changes, the revised param comments are helpful to explain what we are working with. The initial code changes look good and I see the products are syncing now.

I left a few smaller comments, mainly about the tests, but if we want to cover all of UpdateAllProducts we might need to add a few more tests. Since this has a high priority I'd suggest to do that in a followup PR. Some of the tests we might want to add:

- Filter out some products from the batches and confirm that batches continue even if some batch sizes are smaller or empty, see #861
- Test a final batch smaller than batch size (this will identify that `handle_complete` is not always called, which isn't an issue because we never use it)
- Test `process_items` with an exception

Another thought was that we should call the UpdateAllProducts job on plugin update, like [here](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.5/src/Jobs/Update/PluginUpdate.php#L41-L46). We broke the sync since 1.12.2 so all new users that onboarded after that would have run a "bad" sync. After updating they would need to go through the onboarding process again to retrigger a sync if we don't do it for them.

tests/Unit/Jobs/UpdateAllProductsTest.php
message-dimke: Thanks! Adjusted.

tests/Unit/Jobs/UpdateAllProductsTest.php
message-dimke: Renamed and moved into trait.

tests/Unit/Jobs/UpdateAllProductsTest.php
message-dimke: Thanks! Done. Reformatted into short syntax.

tests/Unit/Jobs/UpdateAllProductsTest.php
message-dimke: Got it

message-dimke:
@mikkamp , I've added exception test on process items to test rescheduling.

mikkamp:
Thanks for adding the additional test, that looks good. Before approving though I'm concerned about the following comment:

> Another thought was that we should call the UpdateAllProducts job on plugin update, like [here](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.5/src/Jobs/Update/PluginUpdate.php#L41-L46). We broke the sync since 1.12.2 so all new users that onboarded after that would have run a "bad" sync. After updating they would need to go through the onboarding process again to retrigger a sync if we don't do it for them.

If we don't address that then we need to provide instructions upon release for the steps that need to be followed to get their products to sync (so I think it would be easier to automate that).

mikkamp (approved):
Thanks for the additional update procedure. I can confirm that when I bump the version to 1.12.6 the UpdateAllProducts job is run.

PR #1452, opened 2022-04-27
Campaign Conversion Status

mikkamp:

This PR adds a helper function to retrieve the campaign conversion status by checking the list of campaigns. It will return the following results:

- unconverted - Still need to convert some older campaigns
- converted - All campaigns are converted to PMax campaigns
- not-applicable - User never had any older campaign types

If we detect that the user converted we add the suffix ` (Old)` for any campaigns that are now deleted and were not PMax campaigns. This is updated in the reports results, as well as when we retrieve a list of campaigns (note: we need to specifically include removed campaigns)

A parameter `exclude_removed` was added to the `GET ads/campaigns` endpoint, by default this will be set to true. Setting this to false can be used when requesting campaigns in the reports page filter (we also need to be able to filter removed campaigns).

#### Unconverted campaigns
1. Start with an already linked Ads account (containing both PMax and SSC campaigns)
2. Send a request `GET https://domain.test/wp-json/wc/gla/ads/campaigns?exclude_removed=0`
3. Confirm that we get campaigns with the status "removed"
4. If we have a mix of campaigns (not converted yet) we should not get the suffix ` (Old)` added to any campaign names
5. Check the `wp_options` table for `gla_campaign_convert_status`, it should show the status is `unconverted`

#### No campaign conversion needed
1. Disconnect ads account and reconnect a new (another) account which doesn't have any old SSC campaigns
2. Send a request `GET https://domain.test/wp-json/wc/gla/ads/campaigns?exclude_removed=0`
3. We should not get the suffix ` (Old)` added to any campaign names
4. Check the `wp_options` table for `gla_campaign_convert_status`, it should show the status is `not-applicable`

#### Converted campaigns
1. Disconnect ads account and reconnect a new (another) account with all the SSC campaigns removed and at least one enabled PMax campaign
2. Send a request `GET https://domain.test/wp-json/wc/gla/ads/campaigns?exclude_removed=0`
3. We should get the suffix ` (Old)` added to any campaign names that were removed and are a SSC
4. Check the `wp_options` table for `gla_campaign_convert_status`, it should show the status is `converted`

#### Converted campaigns in reports
1. Check the report data to confirm that the old campaigns have the suffix ` (Old`
2. Send a request `GET https://domain.test/wp-json/wc/gla/ads/reports/programs?after=2022-04-01&before=2022-04-14&interval=day&fields[0]=clicks`
3. We should get the suffix ` (Old)` added to any campaign names that were removed and are a SSC

Note that this scenario is a little harder to test because we need actual reporting data for those two campaigns. I tested it with the `tests/proxy` utility and modified some of the mocked responses accordingly.

### Changelog entry
* Add - Campaign Conversion Status for detecting converted campaigns.


src/API/Google/AdsCampaign.php
```diff
@@ -329,6 +339,65 @@ public function delete_campaign( int $campaign_id ): int {
...
+						if ( CampaignStatus::REMOVED === $campaign['status'] ) {
+							$old_removed_campaigns++;
+						} else {
+							$old_campaigns++;
+						}
+					}
+				}
+			} catch ( Exception $e ) {
+				// Error when retrieving campaigns, do not handle conversion.
+				$convert_status['status'] = 'not-applicable';
```
jorgemd24: If the `$this->get_campaigns()` fails only once for X reason, we would not give the opportunity to re-try as we are only checking the status if `$convert_status['status'] === 'unconverted'`, what do you think?

tests/Unit/API/Google/AdsCampaignTest.php
```diff
@@ -464,4 +464,119 @@ public function test_delete_campaign_exception() {
 		}
 	}
 
+	public function test_get_campaign_convert_status_unconverted() {
+		$campaigns_data = [
+			[
+				'id'      => self::TEST_CAMPAIGN_ID,
+				'name'    => 'Campaign One',
+				'status'  => 'enabled',
+				'type'    => 'shopping',
```
jorgemd24: is it not better to use the constants `CampaignType::PERFORMANCE_MAX` and `CampaignType::PERFORMANCE_MAX` instead of writing this manually?

tests/Unit/API/Google/AdsCampaignTest.php
```diff
@@ -464,4 +464,119 @@ public function test_delete_campaign_exception() {
...
+
+	public function test_get_campaign_convert_status_fetch_cached() {
+		$this->options->method( 'get' )
+			->with( OptionsInterface::CAMPAIGN_CONVERT_STATUS )
+			->willReturn(
+				[
+					'status'  => 'unconverted',
+					'updated' => time(),
+				]
+			);
```
jorgemd24, line 553: We could test as well the case when the option `OptionsInterface::CAMPAIGN_CONVERT_STATUS` is not set or the `updated`field is empty.

jorgemd24 (commented):
Thanks, @mikkamp,

✅  I have tested all the scenarios and they are working as expected. 

I have left a comment about what would happen if it fails to get the campaigns, for example, if it fails once for a network issue, the status will stay in `not-applicable` permanently.

src/API/Google/AdsCampaign.php
mikkamp: I was originally thinking the worst case scenario would be that those users don't see the update notifications, which isn't going to block anything for them. But I suppose we can introduce a 4th status `unknown`, which will also retry after the cache expires.

tests/Unit/API/Google/AdsCampaignTest.php
mikkamp: In all the tests for a specific status we don't mock the option return value, so it will be equal to not being set. I added another test for when the updated field is empty.

src/API/Google/AdsCampaign.php
jorgemd24: >I was originally thinking the worst case scenario would be that those users don't see the update notifications

Yes, it will only affect the notifications and renaming of the old campaigns. 

> But I suppose we can introduce a 4th status unknown, which will also retry after the cache expires.

Yes, or we just leave the status as `unconverted`. But I am not sure if this could cause that we end in an infinite loop if the get_campaigns() fails constantly. What do you think?

mikkamp: Already added the `unknown` status. An infinite loop would be pretty rare, and then it would still cache it for an hour.
If we want we can always show the notice if it's unknown or unconverted. But since it will retry after an hour and the notices run once a day, I think it will eventually show up with just unconverted. Unless they are really unlucky that they can't retrieve any campaigns, but then they have a greater problem to fix and we don't necessarily need to be prompting them that their campaigns are going to update.

mikkamp:
@jorgemd24 Thanks for the initial review and the comments. It's ready for another round.

jorgemd24 (approved):
@mikkamp, thanks for the changes. LGTM 👍

PR #1454, opened 2022-04-28
Clean up after upgrading wp-scripts to 22.1.0 and webpack config to v5

eason9487:

It's stage 3 of #1333 and also includes some relevant enhancements.

- Replace `file-loader` with the built-in Asset Modules of webpack to handle the exporting of images.
- Remove the `injectPolyfill: true` config of DEWP to drop support for IE.
   - Ref: https://github.com/WordPress/gutenberg/pull/35436
- Generate source map of SCSS in development.
- Use `<rootDir>` for paths in jest config, and ignore the **tests** folder when generating the coverage report.
- Remove temporary peer packages that have been correctly listed in those packages that require them.
   - `@wordpress/core-data`
   - `md5`
   - `postcss`
- Remove two `@wordpress/scripts` related packages that can be specified by `import/core-modules` settings of eslint.
   - `stylelint`
   - `@wordpress/stylelint-config`
- Deduplicate the content of **package-lock.json** file via `npm dedupe`.
   - 💡 After the deduplication, the package-lock file will have some diff when back and forth running `npm install` and `npm dedupe`. So the revision be87fa922f66ec901fea526389ef567eb79f89f4 was committed with the result of `npm dedupe` and then `npm install`.

1. Run `npm install` to check if there is no error.
1. Run `npm run test:js`. Check if the **tests** folder is no longer included in the coverage report.
1. Run `npm run lint:js`. The eslint should be able to run.
1. Run/restart `npm start`. Browse GLA pages to check if the images are still shown.
1. Inspect some DOM elements to see if their CSS classes are referenced to the relatively correct number of lines of SCSS source code rather than the number of lines after compilation.
   - 💡 Please note that the number of lines displayed on DevTool will be the actual number of lines in the source file plus 1. Because each scss file will have an extra line inserted with SCSS `at-rules` at the file head via [this webpack config](https://github.com/woocommerce/google-listings-and-ads/blob/be87fa922f66ec901fea526389ef567eb79f89f4/webpack.config.js#L180-L185).


webpack.config.js
```diff
@@ -179,7 +173,7 @@ if ( hasReactFastRefresh ) {
 
 const sassTest = /\.(sc|sa)ss$/;
 const updatedSassOptions = {
-	sourceMap: process.env.NODE_ENV === 'production',
+	sourceMap: ! isProduction,
```
ecgan, line 176: Just want to be sure and leaving a note: here we are turning on sourceMap when it is not production mode, which is the opposite of the code before this PR change.

ecgan (commented):
Tested based on the test instructions and compared the outcome between `develop` branch and this PR branch, things are looking quite good.

Some questions below:

❓ When I was running `npm start`, I see a warning that says `"No entry file discovered in the "src" directory."`. I do not see this in the `develop` branch. @eason9487 , do you know about this? Is this expected?
 
image

❓ I checkout this PR branch, ran `npm ci` and then `npm dedupe`, and I see there are some changes for package-lock.json. Is that expected? 🤔 

❔ In issue https://github.com/woocommerce/google-listings-and-ads/issues/1333, under "Stage 3" section, there is this task: "Revert https://github.com/woocommerce/google-listings-and-ads/commit/0167ec4fc531e89df0b607aec5f7e5957a7f09fe after getting approval and before merging." Just want to be sure, should the revert be in this PR, or outside of this PR?

webpack.config.js
eason9487: Yes, it's the "Generate source map of SCSS in development." item in the PR description.

Originally, the config here is the opposite of the `css-loader` in the [default webpack config](https://github.com/WordPress/gutenberg/blob/%40wordpress/scripts%4012.5.0/packages/scripts/config/webpack.config.js#L31-L33) in `wp-script`, so the source map generated during development does not point to the source scss files. And the production build will not generate any source map in the end because of the `css-loader` config. So I think it's more effective to change it to a consistent config so that the source map can point to the correct scss source files during development.

eason9487:
@ecgan, thanks for the review!

> ❓ When I was running `npm start`, I see a warning that says `"No entry file discovered in the "src" directory."`. I do not see this in the `develop` branch. @eason9487 , do you know about this? Is this expected?

It's a feature of `wp-script` as the [fallback file for the `block.json` config](https://github.com/WordPress/gutenberg/tree/%40wordpress/scripts%4022.1.0/packages/scripts#build), but GLA doesn't use that. So there will [show a warning](https://github.com/WordPress/gutenberg/blob/%40wordpress/scripts%4022.1.0/packages/scripts/utils/config.js#L270-L278) when building. I think it should be ok to ignore it in the GLA use.

> ❓ I checkout this PR branch, ran `npm ci` and then `npm dedupe`, and I see there are some changes for package-lock.json. Is that expected? 🤔

I didn't look deeper into its reason, except to guess that npm's internal processing might be causing inconsistent results. These differences cannot be eliminated with back and forth running `npm dedupe` and `npm install`.

Since we use `npm install` more often than `npm dedupe`, I ended up running `npm install` again after `npm dedupe` in order not to keep causing diff in package-lock.json, and then committing that results to version control.

> ❔ In issue #1333, under "Stage 3" section, there is this task: "Revert [0167ec4](https://github.com/woocommerce/google-listings-and-ads/commit/0167ec4fc531e89df0b607aec5f7e5957a7f09fe) after getting approval and before merging." Just want to be sure, should the revert be in this PR, or outside of this PR?

It will be appended to this PR.

ecgan:
> It's a feature of `wp-script` as the [fallback file for the `block.json` config](https://github.com/WordPress/gutenberg/tree/%40wordpress/scripts%4022.1.0/packages/scripts?rgh-link-date=2022-05-03T08%3A25%3A31Z#build), but GLA doesn't use that. So there will [show a warning](https://github.com/WordPress/gutenberg/blob/%40wordpress/scripts%4022.1.0/packages/scripts/utils/config.js?rgh-link-date=2022-05-03T08%3A25%3A31Z#L270-L278) when building. I think it should be ok to ignore it in the GLA use.

Gotcha. Thank you. 🙂 👍 

> I didn't look deeper into its reason, except to guess that _npm's internal processing might be causing inconsistent results_. These differences cannot be eliminated with back and forth running `npm dedupe` and _`npm install`_.
> 
> Since we use `npm install` more often than `npm dedupe`, _I ended up running `npm install` again after `npm dedupe`_ in order not to keep causing diff in package-lock.json, and then committing that results to version control.

I think that may be the issue. `npm install` may modify package-lock.json ([`npm install` documentation](https://docs.npmjs.com/cli/v6/commands/npm-install)). If we want a reproducible build without modifying package-lock.json, then we should use `npm ci`. 

So with the above said, I think we should have `npm dedupe`, and then run `npm ci` instead of `npm install` so that package-lock.json is consistent and not modified. WDYT?

eason9487:
@ecgan

> So with the above said, I think we should have `npm dedupe`, and then run `npm ci` instead of `npm install` so that package-lock.json is consistent and not modified. WDYT?

I'm not pretty sure about the benefits of aligning to the result of `npm dedupe`.

After running `npm dedupe`, the diff is deleting the `@storybook/mdx1-csf` dependency, but it's actually a dependency of `@storybook/addon-docs` and no other packages rely on it. Thereby, it could be assumed that the result of `npm dedupe` is not so correct.

In addition, the `npm dedupe` attempts to simplify the structure of package tree and maybe reduce a bit of the installation time and total file size of *node_modules*. The `npm ci` is meant to be used in automated environments. And eventually, the *package-lock.json* will be changed when bumping/changing packages.

I think it doesn't have to align with the result of `npm dedupe`. Using the result of `npm install` should be ok.

ecgan:
> I'm not pretty sure about the benefits of aligning to the result of `npm dedupe`.

Actually my concern isn't so much about "aligning to the result of `npm dedupe`". In the PR description, it says "deduplicate the content of **package-lock.json** file via `npm dedupe`". Because of that, I have an expectation that "If I run `npm dedupe`, I should see no changes", however the actual result shows there are some diff changes in package-lock.json, which gives me a bit of doubt, e.g. was it something wrong with the code in this PR, or was it something wrong in my environment, etc.

So I'm just thinking the outcome of this PR should match the PR description there. Maybe we can just rephrase / remove the "npm dedupe" statement in the PR description, or add another commit with the result of `npm dedupe` so that it matches the expectation that we would have when reading PR description. 

---

The following is a side note discussion which I think is outside the scope of this PR, just chipping in my opinion. 😄 

> The `npm ci` is meant to be used in automated environments.
> [...]
> Using the result of `npm install` should be ok.

I'd like to mention that it is _not just_ for automated envinronments. We use that when we want to have consistent reproducible environment. [`npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci) means "clean install". I often use that in my local machine when I switch to some different branches, or when I want to troubleshoot some issues. For example, if I want to reproduce an issue in an old version, we should checkout the version and run `npm ci`. Another example is that a PR build runs fine in the cloud, but when we checkout the PR and run tests in our local machine, the tests might fail with `npm install` but runs fine with `npm ci`.

ecgan (approved):
All the questions in my previous PR review https://github.com/woocommerce/google-listings-and-ads/pull/1454#pullrequestreview-959304147 are answered, there isn't major blocker in this PR and things are working as expected, I'm approving this. 🙂 👍 

There is only one comment https://github.com/woocommerce/google-listings-and-ads/pull/1454#issuecomment-1116152867 about `npm dedupe`, which I think is a minor issue and not a dealbreaker in this PR.

Friendly reminder: don't forget about the task "Revert [0167ec4](https://github.com/woocommerce/google-listings-and-ads/commit/0167ec4fc531e89df0b607aec5f7e5957a7f09fe) after getting approval and before merging." 😄  (Maybe next time we could have just include the revert into the PR, instead of doing it after the approval?)

Thanks @eason9487 for your work here.

eason9487:
@ecgan

> So I'm just thinking the outcome of this PR should match the PR description there. Maybe we can just rephrase / remove the "npm dedupe" statement in the PR description, or add another commit with the result of `npm dedupe` so that it matches the expectation that we would have when reading PR description.

I updated the PR description.

> I'd like to mention that it is _not just_ for automated envinronments. We use that when we want to have consistent reproducible environment. [`npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci) means "clean install". I often use that in my local machine when I switch to some different branches, or when I want to troubleshoot some issues. [...]

I believe I didn't mean it was *just for* automated environments. I meant to clarify that it would be less than ideal to restrict developers to use `npm ci` as much as possible in the development environment. Regardless, thanks for sharing your ways of using it.

> Friendly reminder: don't forget about the task "Revert [0167ec4](https://github.com/woocommerce/google-listings-and-ads/commit/0167ec4fc531e89df0b607aec5f7e5957a7f09fe) after getting approval and before merging."

Thanks!

> (Maybe next time we could have just include the revert into the PR, instead of doing it after the approval?)

Sounds good!

Issue #1228, opened 2022-02-01
WooCommerce Shipping Zones Integration - Syncing rates for states/provinces and postal codes


nima-karimi:
## Objective
The aim is to provide integration with WooCommerce shipping zones so that the shipping rates are set for various regions such as states, provinces, and postal/zip codes can be automatically synchronized with Google Merchant Center.

## Overview
In this milestone, we will set up the system required to automatically map the supported shipping methods defined for each state/province or zip code in all WooCommerce shipping zones into their respective Merchant Center settings.

## The current state of the system
At the end of milestone one, we currently have an integration with WooCommerce shipping zones. However, it only works at a country level. Meaning that the shipping methods are all grouped together and synchronized for each country. Even if they are defined at a more granular level such as a state or zip code.

Currently, the user has three options when it comes to setting their store’s shipping settings during the onboarding process:

> - Recommended: Automatically sync my store’s shipping settings to Google.
> - My shipping settings are simple. I can manually estimate flat shipping rates.
> - My shipping settings are complex. I will enter my shipping rates manually in Google Merchant Center

![image](https://user-images.githubusercontent.com/73110514/151979310-67c9e8b6-4332-4ca7-bac9-9cda4bcf3ac1.png)
Three options are presented to users for setting the shipping rates during onboarding.

The automatic syncing will happen only if the user chooses the first option. In that case, we map each shipping method defined in WooCommerce shipping zones to a Merchant Center shipping service.

## WooCommerce Shipping Zones
A Shipping Zone is a geographical area to which you ship items. You can get as specific as you need, right down to regions and zip codes, or you can leave it more general. Each customer matches only one zone.

To set up shipping in WooCommerce, you first set up shipping zones, then add methods to those zones, and lastly, rates to your methods. You can create as many zones as you like, and add multiple methods and rates to each zone.

![image](https://user-images.githubusercontent.com/73110514/151979416-a034a285-b045-4639-ad7d-853f0954472b.png)
Defining a shipping zone

Each zone can have multiple regions along with multiple postcodes associated with it. The regions can be defined as continents, countries, or states and provinces of each country. The postcodes can also be defined either as ranges (90301-90305), wildcards (9003*), or one number (90210).

[...]

PR #1455, opened 2022-04-28
Shipping Integration - Sync shipping rates for subdivisions of a country

### LINKED ISSUE #1231: Shipping Integration - Syncing the rates for provinces/states and postal codes
As part of #1228

This issue is dependent on #1230.

Once we have the rates for supported subregions of shipping countries, we can submit them to Merchant Center by modifying the `API/Google/Settings` class.

The rates for sub-regions and postcodes will be submitted only if the user has chosen the “automatic” shipping option during onboarding.

The `update_shipping_rates` method and other related methods will be modified to take into account the new property returned by the `ShippingZone::get_shipping_rates_for_country` method. This property includes the list of shipping rates for each subregion (state, province, postcode, etc.).

The list of returned postcodes should be provided as an array of [PostalCodeGroup](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#PostalCodeGroup) objects under the postalCodeGroups property of [ShippingSettings](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#resource:-shippingsettings). This will allow us to refer to it in our shipping rates table.

Currently, to store the rates, we create one [RateGroup](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#RateGroup) as the main flat rate and then optionally multiple other [RateGroup](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#RateGroup) objects for different shipping classes/labels.

Submitting the rates for postcodes or subregions is only possible by creating an instance of the [Table](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Table) class and filling it with [Headers](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Headers) and [Rows](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Row). The [RateGroup](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#RateGroup) has a `mainTable` property that accepts an instance of the [Table](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Table) class.

For example, if WooCommerce has the following shipping zones and methods:

Zone name | Region | Shipping methods
-- | -- | --
Local | California ZIP 90210 | Flat rate ($10)
State | California | Flat rate ($20

This will translate to the following table set in the [RateGroup](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#RateGroup) object:

locations/postalCodeGroupNames | 21137(California state location ID fetched from the Google CSV)
-- | --
90210 | $10
"all other locations" | $20

The "all other locations" string is a special string that can be used in the `postalCodeGroupNames` property to specify the rates for all other postcodes in the area.

A table can have two dimensions with two headers. It can have as many [Rows](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Row) as the number of `rowHeaders`. Each [Row](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Row) can contain as many cells as the number of `columnHeaders`.

In the example above, the `rowHeaders` object has the following values set for its `postalCodeGroupNames` property: 90210 and “all other locations”. Consequently, the `columnHeaders` property has the following value set for its `locations` property: 21137. The table `rows` property has an array of two [Row](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Row) objects each containing one cell with the shipping cost: $10 and $20.

nima-karimi:

Closes #1231.

This PR implements the system for parsing and syncing the suggested shipping rates for states/provinces and postal codes fetched from WC Shipping Zones (provided by #1341) with Google Merchant Center.

Two new adapter classes are created for mapping the store's shipping configuration to Google's `ShippingSettings`:

- **`DBShippingSettingsAdapter`**: Maps the rates obtained from the database to a ShippingSettings object that can be sent to the Google API. It accepts a `db_rates` argument that contains the rates in the same format as set in the `gla_shipping_rates` table.
- **`WCShippingSettingsAdapter`**: Maps the `LocationRate` objects fetched by parsing the WC Shipping Zone (by the `ShippingZone` class). It accepts a `rates_collections` parameter, which is an array of `CountryRatesCollection` containing location rates for each country supported by the store.

We group the shipping rates by their country and applicable shipping area first. An applicable area can be a whole country, a state, or specific postcodes of a country. Then for each group, we separate the rates that have any minimum order amount constraints (specifically free shipping rates with a min order amount threshold).

After the rates are all grouped together, we map them according to their applicable shipping area:

- If the rates only apply to a country, then we create a simple RateGroup with just one flat rate for the whole country.
- If the rates only apply to a state, then we create a table rate group and set the shipping rates for each state by mapping the rate to their associated location ID.
- If the rates apply to postcodes, we again create a table rate group for that country and set the rates for each group of postcodes.

An important note is that Google assumes all postal codes are unique within a country; therefore, the API doesn't accept different shipping rates associated with the same postcode but in different states/provinces. WooCommerce, on the other hand, allows the user to set different postcode values for any country+state combinations so there's a slight difference in how we handle that here. This shouldn't be a problem for us since we only support a handful of countries at the moment and AFAIK all those countries have unique postcodes. However, if we later add support for countries that don't adhere to this rule (like France) we (and Google) might need to change how this mapping works.

1. Create multiple types of shipping settings, rates, and zones in WooCommerce settings
2. Edit your free listing program or go through the plugin onboarding and select the "Recommended" option for syncing the shipping settings
3. Complete setup and confirm in MC that all your shipping settings are synced (only the supported countries and methods will be synced as mentioned in #1228)
4. Edit your free listing program again and this time choose the "My shipping settings are simple." option
5. Enter shipping rates for the countries and complete setup
6. Check in MC that the shipping rates are now updated to reflect the new ones set in the previous step.

### Changelog entry

> Add - Sync the shipping rates for states/provinces and postal codes to Merchant Center.


nima-karimi:
Still a work-in-progress because I have hit an error when trying to sync the shipping rates that belong to specific postcodes of a state and forwarded the following question to the Google team:

```
I’m hitting the following error from the Google shipping API when I set a [table](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Table) for a RateGroup containing a ‘locationIdSets’ header for its rows and ‘postalCodeGroupNames` for the columns:

[services[4].rateGroups[0].mainTable] Both dimensions of the table are of type delivery location.

This happens when, for example, we set the following table:

+----------------------------------------+
Location ID \ Postcodes    ||  1001 - 1009
123 (CA)                             ||  20
456 (NV)                            ||  10
+----------------------------------------+

I assume this means that we can’t use locationIds AND postal codes in the same table but there are valid use cases for this, for example, when the user sets a shipping rate for just a few postcodes of a certain state/province.

This constraint probably comes from the assumption that all postcodes are unique within a country and we should simply set them on a country level instead of on a state/province level. However, this is not true for all countries and there’s no such constraint defined in the WooCommerce shipping settings.

The question is: should we just assume that all postcodes are unique within a country and simply set them on a country level? Since we currently only support a select number of countries, this might be feasible but it’s something we might need to change later once we support more countries.
```

I'll update the PR description with more details once the above issue is resolved and this is ready for a review.

nima-karimi:
This PR is now unblocked and ready for review. The Google team responded to my question above and their requirements are as I suspected. The summary is: Postcodes are unique within a country; therefore, a rate group cannot define rates for both states and postcodes in the same table.

Since we only support a handful of countries, the above constraint is fine for now. It might bite us back when we add support for more countries that don't adhere to this rule. That said, there's nothing we can do now since Google API is enforcing this constraint on their side.

To make the mapping code compatible with the above, I've ignored the `state` parameter for a shipping location if the location already has `postcodes` defined. This means that if the user defines shipping rates for "US_CA::90210", we set the rate for "US::90210" by creating a table rate.

src/Shipping/CountryRatesCollection.php
```diff
@@ -0,0 +1,108 @@
...
+
+		foreach ( $this->location_rates as $location_rate ) {
+			$country          = $location_rate->get_location()->get_country();
+			$shipping_area    = $location_rate->get_location()->get_applicable_area();
+			$min_order_amount = $location_rate->get_shipping_rate()->get_min_order_amount();
+
+			// Group rates by their applicable country and affecting shipping area
+			$service_key = $country . $shipping_area;
+			// If the rate has a min order amount constraint, then it should be under a new service
+			if ( ! is_null( $min_order_amount ) ) {
```
mikkamp: Wouldn't it be safer to just use `if ( $min_order_amount )` or do we expect this value to never be set to 0? Or since it's a float compare for it to be larger than 0.0? In WC it's valid to set the minimum order amount to `0` or `0.0`.

Same comment for the function `has_min_order_amount`.

tests/Unit/Shipping/GoogleAdapter/DBShippingSettingsAdapterTest.php
```diff
@@ -0,0 +1,182 @@
...
+		$this->assertCount( 1, $services[0]->getRateGroups() );
+		$this->assertCount( 1, $services[1]->getRateGroups() );
+
+		foreach ( $services as $service ) {
+			// Assert that the delivery country of both services is either US or AU
+			$this->assertTrue( in_array( $service->getDeliveryCountry(), [ 'US', 'AU' ], true ) );
+
+			if ( 'US' === $service->getDeliveryCountry() ) {
+				$this->assertEquals( 'USD', $service->getRateGroups()[0]->getSingleValue()->getFlatRate()->getCurrency() );
+				$this->assertEquals( 10, $service->getRateGroups()[0]->getSingleValue()->getFlatRate()->getValue() );
```
mikkamp, line 50: Since we have different delivery times per country, why aren't we asserting here that those are correct as well?

tests/Unit/Shipping/GoogleAdapter/DBShippingSettingsAdapterTest.php
```diff
@@ -0,0 +1,182 @@
...
+				'currency'       => 'USD',
+				'delivery_times' => [ 'US' => 1 ],
+				'db_rates'       => $db_rates,
+			]
+		);
+
+		$this->assertEmpty(  $settings->getServices() );
+	}
+
+	public function test_sets_free_shipping_threshold_on_free_rates() {
```
mikkamp, line 82: Should we also test free shipping without any threshold?

tests/Unit/Shipping/PostcodeRangeTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Shipping\PostcodeRange;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
+
+/**
+ * Class PostcodeRangeTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Shipping
+ */
+class PostcodeRangeTest extends UnitTest {
```
mikkamp, line 14: It looks like each postcode is mapped through this range class but we aren't actually testing ranges. Should we be testing single postcode ranges here as well? Or what about a postcode containing a wildcard?

mikkamp (commented):
That's a lot of changes, at initial glance it looks like the code is functioning as expected without any errors.

It's not showing up as one of the changes in this PR. But the function `LocationRatesProcessorTest::setUp` should have a void return type to be compatible with the PHP 8 tests.

Maybe not part of this PR, but I noticed that ShippingRateSuggestionsController no longer uses the ShippingZone parameter.

I tested some different scenarios with various zones/rates and they appear to be coming across as expected. For the US I have a mix of both rates for States and for postcode ranges. When I call the funciton `generate_shipping_settings` it creates an instance of `Google\Service\ShoppingContent\Table` with the same postcode ranges:

```
["postalCodeGroupNames"]=>array(1) {
    [0]=>string(30) "US - 99501...99950,1234*,23456"
}

...

["postalCodeRanges"]=>
      array(3) {
        [0]=>
        object(Google\Service\ShoppingContent\PostalCodeRange)#17084 (5) {
          ["postalCodeRangeBegin"]=>string(5) "99501"
          ["postalCodeRangeEnd"]=>string(5) "99950"
        }
        [1]=>
        object(Google\Service\ShoppingContent\PostalCodeRange)#17085 (5) {
          ["postalCodeRangeBegin"]=>string(5) "1234*"
          ["postalCodeRangeEnd"]=>NULL
        }
        [2]=>
        object(Google\Service\ShoppingContent\PostalCodeRange)#17086 (5) {
          ["postalCodeRangeBegin"]=>string(5) "23456"
          ["postalCodeRangeEnd"]=>NULL
        }
      }
```

But I don't see these appear in my Merchant Account:

image

Do I need to test postcode ranges separately without rates for zones? Or is there something else I'm missing?

It seems like all those formats should be supported:
image

mikkamp:
Update: It did sync the postcode as a separate table, I just failed to notice that it was only showing the first 5 services.

image

image

src/Shipping/CountryRatesCollection.php
nima-karimi: My main reason for this change was that comparing floats usually [can't be trusted](https://www.php.net/manual/en/language.types.float.php) as they work differently on different systems. So PHP might think 0.001 is 0. That might still be fine given the context I guess but I think simply checking whether the `min_order_amount` has been set or not would work here because even if we get two rates with min_order_amount of 0, it'd be a result of user misconfiguration and not a mapping error.

tests/Unit/Shipping/GoogleAdapter/DBShippingSettingsAdapterTest.php
nima-karimi: Done in https://github.com/woocommerce/google-listings-and-ads/pull/1455/commits/c0d1a954a3d9ddcc8715e32cfd09b30ebbcdf7f0

tests/Unit/Shipping/GoogleAdapter/DBShippingSettingsAdapterTest.php
nima-karimi: The mapping logic would be the same as any other rate but for completeness, it might be a good idea. Added in https://github.com/woocommerce/google-listings-and-ads/pull/1455/commits/c67fcd98b9df0b862081568183c73aeed49b6615

tests/Unit/Shipping/PostcodeRangeTest.php
nima-karimi: Hmm. Do you mean we should test the `set_` and `get_` methods to see whether they can accept and return single postcodes and wildcards?

Given that the PostcodeRange class has limited logic and is more used as a value object, we're only testing its string conversion functionalities. The `set_` and `get_` and construct functionalities are also being tested as part of the string casting tests. A postcode containing a wildcard would still be treated as any other string since this class doesn't handle any formating.

nima-karimi:
Thanks, @mikkamp 🙏 And sorry for the large changeset. I've responded to your comments and made some changes to the code. Ready for another review.

> It's not showing up as one of the changes in this PR. But the function LocationRatesProcessorTest::setUp should have a void return type to be compatible with the PHP 8 tests.

Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1455/commits/a309d6754a5e9ffbf1c015ac7414a2bd00ceb3f5

 > Maybe not part of this PR, but I noticed that ShippingRateSuggestionsController no longer uses the ShippingZone parameter.

This was a change introduced in #1422. I've fixed it in there and merged the commit here.

tests/Unit/Shipping/PostcodeRangeTest.php
mikkamp: The comment was more intended for the `PostcodesRateGroupAdapterTest` where it only tests mapping with a postcode range. Which I suppose is fine since `PostcodeRangeTest`, tests a single range. I was also a bit confused about why it tests a postcode range for `DE` (I was under the impression that subdivisions was only supported for some countries). I guess it's just the API that has this restriction, and the mapping should work regardless.

I was originally under the impression that the class `Google\Service\ShoppingContent\PostalCodeRange` would also handle syntax checking, but it seems it just allows any value for begin and end, and it's only when it's passed to the API that it might complain about invalid syntax. I was looking for the limitations specified [here](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#PostalCodeRange), while both begin and end can be a string it mentions specific formatting.

I mainly say this is missing, because WooCommerce does not prevent you from saving a postcode rate like `1234?`, Internally it will never match that location when calculating shipping. However I tested that with the mapping and it doesn't complain since it accepts all strings, but when we try to sync that to Google it responds with the following error:
```
{
	"code": "gla_setting_sync_error",
	"message": "{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"[postalCodeGroups[0].postalCodeRanges[3]] None of the pattern matched by postal code: 1234?.\",\n    \"errors\": [\n      {\n        \"message\": \"[postalCodeGroups[0].postalCodeRanges[3]] None of the pattern matched by postal code: 1234?.\",\n        \"domain\": \"global\",\n        \"reason\": \"invalid\"\n      }\n    ]\n  }\n}\n",
	"data": {
		"status": 500,
		"message": "",
		"data": {
			"error": {
				"code": 400,
				"message": "[postalCodeGroups[0].postalCodeRanges[3]] None of the pattern matched by postal code: 1234?.",
				"errors": [
					{
						"message": "[postalCodeGroups[0].postalCodeRanges[3]] None of the pattern matched by postal code: 1234?.",
						"domain": "global",
						"reason": "invalid"
					}
				]
			}
		}
	}
}
```

mikkamp:
Thanks for those extra changes, that looks good and it's passing the unit tests now.

I just had one worry left, as mentioned [here](https://github.com/woocommerce/google-listings-and-ads/pull/1455#discussion_r865653189). I just think we shouldn't make an error like that prevent syncing from happening, especially since the sync errors aren't surfaced in the UI, they would have to check the logs to find out what to fix.

nima-karimi:
> The comment was more intended for the PostcodesRateGroupAdapterTest where it only tests mapping with a postcode range. Which I suppose is fine since PostcodeRangeTest, tests a single range.

Yes, since that functionality is not defined in the `PostcodesRateGroupAdapter` class it would make more sense to be tested in the `PostcodeRangeTest`.

> I was also a bit confused about why it tests a postcode range for DE (I was under the impression that subdivisions was only supported for some countries). I guess it's just the API that has this restriction, and the mapping should work regardless.

We don't have any checks in the code since we automatically filter only the supported countries when we build the mapping and the list of suggested shipping countries won't come from external/user source. I changed that country to `AU` for the time being but as you said it wouldn't have any effect on the code since that list is dynamic and can be changed later.

> I was originally under the impression that the class Google\Service\ShoppingContent\PostalCodeRange would also handle syntax checking, but it seems it just allows any value for begin and end, and it's only when it's passed to the API that it might complain about invalid syntax. I was looking for the limitations specified [here](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#PostalCodeRange), while both begin and end can be a string it mentions specific formatting.
> I mainly say this is missing, because WooCommerce does not prevent you from saving a postcode rate like 1234?, Internally it will never match that location when calculating shipping. However I tested that with the mapping and it doesn't complain since it accepts all strings, but when we try to sync that to Google it responds with the following error
 
Yes, we're merely passing the values from WC to MC API and no validation happening here. This is kind of a trade-off, whether we want to add the same rules defined by Google API in our own code or simply let Google handle the validation and relay that back to the user. I decided to go with the API way because it would make our code simpler. Do you think checking them locally would be more beneficial and worth the extra logic? We could argue that it reduces the # of invalid requests but this isn't like product sync and this endpoint is only called once by user action.

mikkamp (approved):
> Yes, we're merely passing the values from WC to MC API and no validation happening here. This is kind of a trade-off, whether we want to add the same rules defined by Google API in our own code or simply let Google handle the validation and relay that back to the user. I decided to go with the API way because it would make our code simpler. Do you think checking them locally would be more beneficial and worth the extra logic? We could argue that it reduces the # of invalid requests but this isn't like product sync and this endpoint is only called once by user action.

If the UI would return the error I think it would be easier for them to correct and resync, but all they would get to see is "Unable to complete your setup. Please try again later.", for the actual error they would need to check the logs. It was just on my mind that we ignore other rates (for example it uses a formula for the rate), so it would seem logical to ignore the postcode line.

Granted though, this won't happen often, so I'll leave it up to you to decide. We could also move it to a low priority issue, so I'll go ahead and approve the PR since I didn't come across any other blockers.

nima-karimi:
Thanks for the thorough review, @mikkamp 🙏 

> If the UI would return the error I think it would be easier for them to correct and resync, but all they would get to see is "Unable to complete your setup. Please try again later.", for the actual error they would need to check the logs. 

This is a valid point and a nice improvement that applies to the rest of the API endpoints as well. It's just, as you mentioned, beyond the scope of this PR. I think we can work on improving the error handling and relaying the API errors in a separate PR.

PR #1456, opened 2022-04-28
GLA Request review - Add Onboarding notice

puntope:

This PR adds the Onboarding Status to the Account Status as well as other minor refactoring. 

- The onboarding status will be there when the status of any of the program regions is Onboarding or if any of the global states is `NO_OFFERS_UPLOADED`

Its present in the frontend like this
Screenshot 2022-04-28 at 17 26 42

- The statuses calculation was refactored in a less verbose way. 
- Fixed an error in which array_unique returned an associative version of the issues array sometimes. This was making Frontend part to fail because associative array is parsed as an object in JSON instead of an array. 
- Tweak: In case we find unknown statues or disabled global states we will hide the notice and the status.
- Fix: Added cooldown property instead of `nextAttempt` in the default state that was missed in the last PR. 

```
$response = [
				                 'programTypeA' => [
					                 'status' => 200,
					                 'data'   => [
						                 "globalState" =>  "ENABLED",
						                 'regionStatuses' => [
							                 [
								                 'reviewEligibilityStatus' => 'INELIGIBLE',
								                 'eligibilityStatus'       => 'APPROVED',
							                 ],
							                 [
								                 'reviewEligibilityStatus' => 'ELIGIBLE',
								                 'eligibilityStatus'       => 'DISAPPROVED',
								                 'reviewIssues'            => [ 'one' ]
							                 ],
						                 ]
					                 ]
				                 ],
				                 'programTypeB' => [
					                 'status' => 200,
					                 'data'   => [
						                 "globalState" =>  "ENABLED",
						                 'regionStatuses' => [
							                 [
								                 'reviewEligibilityStatus' => 'INELIGIBLE',
								                 'eligibilityStatus'       => 'WARNING',
								                 'reviewIssues'            => [ 'two' ]
							                 ],
							                 [
								                 'reviewEligibilityStatus' => 'ELIGIBLE',
								                 'eligibilityStatus'       => 'DISAPPROVED',
								                 'reviewIssues'            => [ 'one' ]
							                 ],
							                 [
								                 'reviewEligibilityStatus' => 'INELIGIBLE',
								                 'eligibilityStatus'       => 'UNDER_REVIEW'
							                 ],
						                 ]
					                 ]
				                 ]
			                 ];
```
1. Checkout this issue and run `npm run dev` to generate the assets
2. In `src/API/Site/Controllers/MerchantCenter/RequestReviewController.php::get_review_read_callback()` you can hardcode the response with something like the provided above 
3. Test that if any globalState is NO_OFFERS_UPLOADED we get ONBOARDING status
4. Check that if any statues has ONBOARDING as eligibilityStatus we show ONBOARDING as status
5. Check that if any statues has DISAPPROVED as eligibilityStatus, and no ONBOARDING in any of them we show DISAPPROVED as status
6. Check that if any statues has WARNING as eligibilityStatus, and no ONBOARDING or DISAPPROVED in any of them we show WARNING as status
7. Check that if any statues has UNDER_REVIEW as eligibilityStatus, and no ONBOARDING or DISAPPROVED or WARNING in any of them we show UNDER_REVIEW as status
8. Check that if any statues has PENDING_REVIEW as eligibilityStatus, and no ONBOARDING or DISAPPROVED or WARNING or UNDER_REVIEW in any of them we show PENDING_REVIEW as status
9.  Check that if any statues has APPROVED as eligibilityStatus, and no ONBOARDING or DISAPPROVED or WARNING or UNDER_REVIEW or PENDING_REVIEW in any of them we show APPROVED as status
10. Rest of cases will show null as status. this could be tested by adding for example "globalState" =>  "UNKNOWN" in any of the programs inside the response.

### DIFF
--- a/src/Google/RequestReviewStatuses.php
+++ b/src/Google/RequestReviewStatuses.php
@@ -125,7 +125,7 @@ class RequestReviewStatuses implements Service {
 	 */
 	private function get_cooldown( int $cooldown ) {
 		if ( $cooldown ) {
-			$cooldown = ( $cooldown + self::get_account_review_lifetime() ) * 1000;
+			$cooldown = ( $cooldown + $this->get_account_review_lifetime() ) * 1000;
 		}
 
 		return $cooldown;
--- a/src/Google/RequestReviewStatuses.php
+++ b/src/Google/RequestReviewStatuses.php
@@ -50,14 +50,14 @@ class RequestReviewStatuses implements Service {
 			// otherwise we compute the new status, issues and cooldown period
 			foreach ( $program_type['data']['regionStatuses'] as $region_status ) {
 				$issues = array_merge( $issues, $region_status['reviewIssues'] ?? [] );
-				self::maybe_update_cooldown_period( $region_status, $cooldown );
-				self::maybe_update_status( $region_status['eligibilityStatus'], $status );
+				$this->maybe_update_cooldown_period( $region_status, $cooldown );
+				$this->maybe_update_status( $region_status['eligibilityStatus'], $status );
 			}
 		}
 
 		return [
 			'issues'   => array_map( 'strtolower', array_values( array_unique( $issues ) ) ),
-			'cooldown' => self::get_cooldown( $cooldown ), // add lifetime cache to cooldown time
+			'cooldown' => $this->get_cooldown( $cooldown ), // add lifetime cache to cooldown time
 			'status'   => $status,
 		];
 	}
--- a/src/Google/RequestReviewStatuses.php
+++ b/src/Google/RequestReviewStatuses.php
@@ -97,7 +97,7 @@ class RequestReviewStatuses implements Service {
 		];
 
 		$current_status_priority = array_search( $status, $status_priority_list, true );
-		$new_status_priority = array_search( $new_status, $status_priority_list, true );
+		$new_status_priority     = array_search( $new_status, $status_priority_list, true );
 
 		if ( $new_status_priority === false ) {
 			return;
--- a/src/Google/RequestReviewStatuses.php
+++ b/src/Google/RequestReviewStatuses.php
@@ -9,78 +9,96 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
  */
 class RequestReviewStatuses implements Service {
 
-	public const DISAPPROVED    = 'DISAPPROVED';
-	public const WARNING        = 'WARNING';
-	public const ELIGIBLE       = 'ELIGIBLE';
-	public const UNDER_REVIEW   = 'UNDER_REVIEW';
+	public const DISAPPROVED = 'DISAPPROVED';
+	public const WARNING = 'WARNING';
+	public const ELIGIBLE = 'ELIGIBLE';
+	public const UNDER_REVIEW = 'UNDER_REVIEW';
 	public const PENDING_REVIEW = 'PENDING_REVIEW';
-	public const ONBOARDING     = 'ONBOARDING';
+	public const ONBOARDING = 'ONBOARDING';
+	public const APPROVED = 'APPROVED';
+	public const NO_OFFERS = 'NO_OFFERS_UPLOADED';
+
 
 	/**
-	 * Merges the different program statuses based on priority
+	 * Merges the different program statuses, issues and cooldown period date.
 	 *
-	 * @param array $response
-	 *
-	 * @return array
+	 * @param array $response Associative array containing the response data from Google API
+	 * @return array The computed status, with the issues and cooldown period.
 	 */
 	public function get_statuses_from_response( array $response ) {
 		$issues   = [];
 		$cooldown = 0;
-		$status   = 'APPROVED';
+		$status   = null;
+
+		$valid_program_states = array( 'ENABLED', self::NO_OFFERS );
 
 		foreach ( $response as $program_type ) {
+
+			// In case any Program is with no offers we consider it Onboarding
+			if ( $program_type['data']['globalState'] == self::NO_OFFERS ) {
+				$status = self::ONBOARDING;
+				break;
+			}
+
+			// In case any Program is not enabled or there are no regionStatuses we return null status
+			if ( ! isset( $program_type['data']['regionStatuses'] ) || ! in_array( $program_type['data']['globalState'], $valid_program_states ) ) {
+				$status = null;
+				break;
+			}
+
+			// otherwise we compute the new status, issues and cooldown period
 			foreach ( $program_type['data']['regionStatuses'] as $region_status ) {
-
-				if (
-					isset( $region_status['reviewIneligibilityReasonDetails'] ) &&
-					isset( $region_status['reviewIneligibilityReasonDetails']['cooldownTime'] )
-				) {
-					$region_cooldown = intval( strtotime( $region_status['reviewIneligibilityReasonDetails']['cooldownTime'] ) );
-
-					if ( ! $cooldown || $region_cooldown > $cooldown ) {
-						$cooldown = $region_cooldown;
-					}
-				}
-
-				if ( $region_status['eligibilityStatus'] === self::DISAPPROVED || $region_status['eligibilityStatus'] === self::WARNING ) {
-					if ( $status !== self::DISAPPROVED ) {
-						$status = $region_status['eligibilityStatus'];
-					}
-
-					$issues = array_merge( $issues, $region_status['reviewIssues'] ?? [] );
-				}
-
-				if ( $status === self::DISAPPROVED ) {
-					continue;
-				}
-
-				if ( $region_status['eligibilityStatus'] === self::UNDER_REVIEW ) {
-					$status = self::UNDER_REVIEW;
-				}
-
-				if ( $status === self::UNDER_REVIEW ) {
-					continue;
-				}
-
-				if ( $region_status['eligibilityStatus'] === self::PENDING_REVIEW ) {
-					$status = self::PENDING_REVIEW;
-				}
-
-				if ( $status === self::PENDING_REVIEW ) {
-					continue;
-				}
-
-				if ( $region_status['eligibilityStatus'] === self::ONBOARDING ) {
-					$status = self::ONBOARDING;
-				}
+				$issues = array_merge( $issues, $region_status['reviewIssues'] ?? [] );
+				self::maybe_update_cooldown_period( $region_status, $cooldown );
+				self::maybe_update_status( $region_status['eligibilityStatus'], $status );
 			}
 		}
 
 		return [
-			'issues'   => array_map( 'strtolower', array_unique( $issues ) ),
+			'issues'   => array_map( 'strtolower', array_values( array_unique( $issues ) ) ),
 			'cooldown' => $cooldown * 1000,
 			'status'   => $status,
 		];
 	}
+	/**
+	 * Updates the cooldown period in case the new cooldown period date is available and later than the current cooldown period.
+	 *
+	 * @param array $region_status Associative array containing (maybe) a cooldown date property.
+	 * @param int $cooldown Referenced current cooldown to compare with
+	 */
+	private function maybe_update_cooldown_period( $region_status, &$cooldown ) {
+		if (
+			isset( $region_status['reviewIneligibilityReasonDetails'] ) &&
+			isset( $region_status['reviewIneligibilityReasonDetails']['cooldownTime'] )
+		) {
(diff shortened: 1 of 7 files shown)


src/Google/RequestReviewStatuses.php
```diff
@@ -11,79 +11,96 @@ class RequestReviewStatuses implements Service {
...
+			}
 
-				if ( $region_status['eligibilityStatus'] === self::ONBOARDING ) {
-					$status = self::ONBOARDING;
-				}
+			// otherwise we compute the new status, issues and cooldown period
+			foreach ( $program_type['data']['regionStatuses'] as $region_status ) {
+				$issues = array_merge( $issues, $region_status['reviewIssues'] ?? [] );
+				self::maybe_update_cooldown_period( $region_status, $cooldown );
+				self::maybe_update_status( $region_status['eligibilityStatus'], $status );
```
jorgemd24: ❓

Why `self::maybe_update_cooldown_period` instead of $this->maybe_update_cooldown_period? Usually self is used with static methods but in this case both methods are not static.

src/Google/RequestReviewStatuses.php
```diff
@@ -11,79 +11,96 @@ class RequestReviewStatuses implements Service {
...
+			}
 
-				if ( $status === self::PENDING_REVIEW ) {
-					continue;
-				}
+			// In case any Program is not enabled or there are no regionStatuses we return null status
+			if ( ! isset( $program_type['data']['regionStatuses'] ) || ! in_array( $program_type['data']['globalState'], $valid_program_states, true ) ) {
+				$status = null;
+				break;
+			}
```
jorgemd24, line 48: ❓

Shouldn't this part of the code go first than checking `$program_type['data']['globalState'] === self::NO_OFFERS`?  what do you think?

```
			// In case any Program is with no offers we consider it Onboarding
			if ( $program_type['data']['globalState'] === self::NO_OFFERS ) {
				$status = self::ONBOARDING;
				break;
			}
```

src/Google/RequestReviewStatuses.php
```diff
@@ -11,79 +11,96 @@ class RequestReviewStatuses implements Service {
...
+			if ( $program_type['data']['globalState'] === self::NO_OFFERS ) {
+				$status = self::ONBOARDING;
+				break;
+			}
 
-				if ( $status === self::PENDING_REVIEW ) {
-					continue;
-				}
+			// In case any Program is not enabled or there are no regionStatuses we return null status
+			if ( ! isset( $program_type['data']['regionStatuses'] ) || ! in_array( $program_type['data']['globalState'], $valid_program_states, true ) ) {
```
jorgemd24, line 46: ❔

Just curious, why if one of the programTypes does not have regionsStatus we just return null and we do not use the regionStatus from the other programTypes?

src/Google/RequestReviewStatuses.php
```diff
@@ -11,79 +11,96 @@ class RequestReviewStatuses implements Service {
...
-
-				if ( $region_status['eligibilityStatus'] === self::DISAPPROVED || $region_status['eligibilityStatus'] === self::WARNING ) {
-					if ( $status !== self::DISAPPROVED ) {
-						$status = $region_status['eligibilityStatus'];
-					}
+		$status   = null;
 
-					$issues = array_merge( $issues, $region_status['reviewIssues'] ?? [] );
-				}
+		$valid_program_states = [ 'ENABLED', self::NO_OFFERS ];
```
jorgemd24: ❓ 

Shouldn't 'ENABLED' be a constant like self::NO_OFFERS?

src/Google/RequestReviewStatuses.php
```diff
@@ -11,79 +11,96 @@ class RequestReviewStatuses implements Service {
...
+		}
+	}
+
+	/**
+	 * Updates the status reference in case the new status has more priority.
+	 *
+	 * @param String $new_status New status to check has more priority than the current one
+	 * @param String $status Referenced current status
+	 */
+	private function maybe_update_status( $new_status, &$status ) {
```
jorgemd24: ❓

Why is it needed to pass the second argument ($status) as a reference?

src/Google/RequestReviewStatuses.php
```diff
@@ -11,79 +11,96 @@ class RequestReviewStatuses implements Service {
...
 			'status'   => $status,
 		];
 	}
+	/**
+	 * Updates the cooldown period in case the new cooldown period date is available and later than the current cooldown period.
+	 *
+	 * @param array $region_status Associative array containing (maybe) a cooldown date property.
+	 * @param int   $cooldown Referenced current cooldown to compare with
+	 */
+	private function maybe_update_cooldown_period( $region_status, &$cooldown ) {
```
jorgemd24: ❓

Similar question, why is it needed to pass $cooldown as a reference?

tests/Unit/API/Site/Controllers/MerchantCenter/RequestReviewControllerTest.php
```diff
@@ -116,6 +119,87 @@ public function test_merged_response() {
...
+								                 'eligibilityStatus'       => 'DISAPPROVED',
+								                 'reviewIssues'            => [ 'one' ]
+							                 ],
+						                 ]
+					                 ]
+				                 ],
+				                 'programTypeB' => [
+					                 'status' => 200,
+					                 'data'   => [
+						                 "globalState" =>  "NO_OFFERS_UPLOADED"
```
jorgemd24: ❓

Isn't it better to use the constants `RequestReviewStatuses::NO_OFFERS` etc instead of doing this manually?

tests/Unit/API/Site/Controllers/MerchantCenter/RequestReviewControllerTest.php
```diff
@@ -116,6 +119,87 @@ public function test_merged_response() {
...
+			                 ]
+		                 );
+
+		$response = $this->do_request( self::ROUTE_GET_REQUEST );
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals( [
+			'status'   => null,
+			'issues'   => [],
+			'cooldown' => 0
+		], $response->get_data() );
```
jorgemd24, line 295: 💅

We could test as the scenario when `regionStatuses` is not set and it should return null as well.

jorgemd24 (commented):
Thanks, @puntope, in overall is working as expected. I left some comments about the code.

Something that I saw when I was testing is that for example the `eligibilityStatus` = 'TEST' the status will be 'TEST', I was expecting to see `null`.

src/Google/RequestReviewStatuses.php
puntope: It was only using once, thats why I found adding the constant a bit redundant. However, I added the constant here https://github.com/woocommerce/google-listings-and-ads/pull/1456/commits/dfef02f8cc33ccfdfe80172b3c42a915daf0d4c4

src/Google/RequestReviewStatuses.php
puntope: Because we are modifying the variable status without doing the equivalent `$status = maybe_update_status($new_status, $status);` 

https://www.php.net/manual/en/language.references.pass.php

src/Google/RequestReviewStatuses.php
puntope: Same as previous question

src/Google/RequestReviewStatuses.php
puntope: Doing that would make null the NO_OFFERS state, instead of OBOARDING, since it doesn't have regionStatus

src/Google/RequestReviewStatuses.php
puntope: Thats a fair question, I wanted to avoid inconsistencies. But your'e right that maybe it's better to allow processing of other statuses.

src/Google/RequestReviewStatuses.php
puntope: As far as I know self could be used as well to refer to class scope based methods. Not necessarily static.

src/Google/RequestReviewStatuses.php
jorgemd24: Aaah ok thanks, I thought that all `$program_type`'s had a `regionStatus`

puntope:
> Something that I saw when I was testing is that for example the eligibilityStatus = 'TEST' the status will be 'TEST', I was expecting to see null.

Good spot @jorgemd24  I was not checking that, I modified the function as well as the tests

puntope:
@jorgemd24  I implemented most of the changes, I guess this is ready for another round.

src/Google/RequestReviewStatuses.php
jorgemd24: Yeah, but it is confusing, if you check the codebase usually the self is used for static methods or for example if you have a parent class with the same method, so it is useful to use the self:: or parent:: to the tell which method you want to call but in this case, it does not seem necessary to use self. 

Some examples: 

https://github.com/woocommerce/google-listings-and-ads/blob/abe75172c58526ba326e56ee8fb94ebf63fed1ce/src/Product/Attributes/AttributeManager.php#L220

https://github.com/woocommerce/google-listings-and-ads/blob/de511f904ba86d20584af209ec5d039c48c676f6/src/Shipping/ShippingZone.php#L192

https://github.com/woocommerce/google-listings-and-ads/blob/de511f904ba86d20584af209ec5d039c48c676f6/src/Shipping/ShippingZone.php#L234

puntope: For me doesn't seem confusing to use self for access class methods. But I changed it to make it less confusing since I don't really have a strong preference over it.

src/Google/RequestReviewStatuses.php
jorgemd24: Thanks, however, I do not see the benefit in this case to pass-by-reference instead of pass-by-value. Using the pass-by-reference makes the code more susceptible to creating bugs as you are updating variables outside of the function scope. Looking into the codebase you see that mostly all the cases that pass-by-reference are used, it is when we are updating the actual values of an array, in that case, I think it makes more sense to use the reference. 

Example: 

https://github.com/woocommerce/google-listings-and-ads/blob/fb70c94c21037c37352ee7d8d30d90362b1d6fc4/src/API/Site/Controllers/MerchantCenter/IssuesController.php#L92

puntope: I don't think passing a var by reference will create bugs. Also notice that both are private functions using internally. 

However, I update it to by-value to get the PR moving and make it clearer.

tests/Unit/API/Site/Controllers/MerchantCenter/RequestReviewControllerTest.php
```diff
@@ -40,10 +40,11 @@ public function test_route() {
 				                 'programType' => [
 					                 'status' => 200,
 					                 'data'   => [
+						                 "globalState" =>  "ENABLED",
```
jorgemd24: 💅

We could use the ENABLE constant here.

tests/Unit/API/Site/Controllers/MerchantCenter/RequestReviewControllerTest.php
```diff
@@ -110,12 +113,188 @@ public function test_merged_response() {
...
+	public function test_unset_region_statuses() {
+
+		$this->middleware->expects( $this->once() )
+		                 ->method( 'get_account_review_status' )
+		                 ->willReturn(
+			                 [
+				                 'programTypeA' => [
+					                 'status' => 200,
+					                 'data'   => [
+						                 "globalState"    => "TEST",
```
jorgemd24: 💅

I think the globalState should be a valid state for example 'RequestReviewStatuses::ENABLE' or 'RequestReviewStatuses::NO_OFFERS' to cover the case of a valid globalState without region but as you mentioned in a previous comment, that will depend on:

>Thats a fair question, I wanted to avoid inconsistencies. But your'e right that maybe it's better to allow processing of other statuses.

jorgemd24 (approved):
Thanks, @puntope for the changes, it is working as expected. I left a couple of comments related to tests.

Regards to this comment: https://github.com/woocommerce/google-listings-and-ads/pull/1456#discussion_r864580918 I think we need more context to check if it is convenient to allow processing of other statuses.

tests/Unit/API/Site/Controllers/MerchantCenter/RequestReviewControllerTest.php
puntope: Thats handled in `test_unexpected_state_and_good_states_response` test case

PR #1460, opened 2022-04-29
Get started page UX improvements: add get started card

ianlin:
❗ **Change the base branch to `feature/1392-ux-improvements-on-get-started-page` when #1449 is merged.**

Part of #1392, this PR adds the get started card to the new get started page.

Design for desktop: eQ9O4m2flzrcAiDMXkOW0m-fi-7350%3A121253
Design for mobile: eQ9O4m2flzrcAiDMXkOW0m-fi-7476%3A123161

#### Desktop

The image should appear at the right-hand side.

Screenshot 2022-04-29 at 14 47 23

#### Mobile

The image should appear at the top.

Screenshot 2022-04-29 at 14 47 44

#### Tablet

The image should appear at the right-hand side.

Screenshot 2022-04-29 at 14 48 02

2. Check the fifth card of the get started page, the UI should align with the design.

#### Testing for Tracking Events

1. Open your console and run `localStorage.setItem('debug', 'wc-admin*')`.
2. Enable verbose mode in dev tool's console.
3. Go to GLA's get started page: `https://domain.test/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fstart`
4. Click on the `Start listing products →` on the fifth card, it should open a new tab to the set up MC page.
    - `https://domain.test/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-mc`
6. From the dev console you should see an event is sent with `target` = `set_up_free_listings`.

### Changelog entry

> Update - Add the get started card for UX improvements on get started page.


js/src/get-started-page/get-started-card/index.js
```diff
@@ -28,66 +27,61 @@ const GetStartedCard = () => {
...
-							'List your products on Google, for free and with ads',
+		<Card className="gla-get-started-card" isBorderless>
+			<FlexItem className="motivation-image">
+				<img
+					src={ motivationImageURL }
+					alt={ __(
+						'Google Shopping search results example',
+						'google-listings-and-ads'
+					) }
+					width="279.05"
```
eason9487: 💅  `.05` could be removed since it doesn't affect the render size of the image.

js/src/get-started-page/get-started-card/index.js
```diff
@@ -28,66 +27,61 @@ const GetStartedCard = () => {
...
+				</Text>
+				<AppButton
+					isPrimary
+					disabled={ disableNextStep }
+					href={ getSetupMCUrl() }
+					eventName="gla_setup_mc"
+					eventProps={ {
+						target: 'set_up_free_listings',
+						trigger: 'click',
+					} }
```
eason9487: ❓ 
Should this event tracking be distinguished from the GetStartedWithVideoCard one? Otherwise, they may be aggregated into the same event from the tracking data.
https://github.com/woocommerce/google-listings-and-ads/blob/587422aaa78b275c173009c14ea87a5dc9498816/js/src/get-started-page/get-started-with-video-card/index.js#L79-L83

js/src/get-started-page/get-started-card/index.js
```diff
@@ -28,66 +27,61 @@ const GetStartedCard = () => {
...
-						{ createInterpolateElement(
-							__(
-								'By clicking ‘Set up free listings in Google’, you agree to our <link>Terms of Service.</link>',
-								'google-listings-and-ads'
+						),
+						{
+							link: (
+								<AppDocumentationLink
+									context="get-started"
+									linkId="wp-terms-of-service"
```
eason9487, line 79: ❓ Same tracking concern as above.

js/src/get-started-page/get-started-card/index.scss
```diff
@@ -1,41 +1,60 @@
-.woocommerce-marketing-google-get-started-card {
-	.components-flex {
-		flex-direction: column;
-		align-items: flex-end;
+@import "../_shared";
```
eason9487: 🚧 
Although **_shared.scss** was not created in this PR, it's worth mentioning that Sass `@import` makes all scss variables in the imported file globally accessible across all scss files. Variables pollution globally further cause the collisional variables to overwrite each other in the order they are imported while compiling, making debugging and maintenance difficult.

Although Sass `@use` can keep scss variables within the scope of individual scss files, the [current webpack config](https://github.com/woocommerce/google-listings-and-ads/blob/587422aaa78b275c173009c14ea87a5dc9498816/webpack.config.js#L127-L131) inserts other `at-rules` at the first line of every scss file, making it impossible to use `@use` for individual scss files. It's recommended to add a `gla-` prefix to variables and move all shared variables to **js/src/css/_variables.scss** for better maintainability.

Ref: https://sass-lang.com/documentation/at-rules/import

js/src/get-started-page/get-started-card/index.scss
```diff
@@ -1,41 +1,60 @@
...
 
-		@include break-large() {
-			flex-direction: row;
-			align-items: center;
-		}
+.gla-get-started-card {
+	display: flex;
+	flex-direction: row-reverse;
+	padding: $grid-unit-50 $grid-unit-50 $grid-unit-20 $grid-unit-50;
```
eason9487: 💅  The last `$grid-unit-50` could be omitted when it's the same as the second one.

js/src/get-started-page/get-started-card/index.scss
```diff
@@ -1,41 +1,60 @@
...
+			align-items: center;
 		}
 	}
 
-	.components-flex &__terms-notice {
-		margin-top: $grid-unit-20;
-		font-style: italic;
+	// Use more specific rules to make it higher priority to override <Text> component.
+	& &__terms-notice {
+		// margin-top: $grid-unit-20;
```
eason9487: 💅  The commented-out style seems to be left?

eason9487 (commented):
Left some concerns and suggestions. The primary concern of sharing variables via Sass `@import` without prefixing could be addressed in another PR since it needs to change a larger scope of files.

js/src/get-started-page/get-started-card/index.scss
puntope: :til: Thanks for spotting this @eason9487 I didn't know that one neither and missed that one in my review.

js/src/get-started-page/get-started-card/index.js
ianlin: Thanks for raising this up @eason9487!

@j111q as Eason's question, we now have two `Start listing products →` CTA in the new get started page, do we see the two CTA clicks are two different events, or were they actually the same? Thanks!

@eason9487

> Otherwise, they may be aggregated into the same event from the tracking data.

I don't quite understand what does this mean. My understanding of events aggregation is if we send the events with the same event name and same props, they would be seen as the same one, otherwise not. So in this case, if the answer from Jill is we don't need to distinguish the two events, we just keep the code as is. But if we need to distinguish them, we should send two different event names from each click. Do I understand it correctly?

js/src/get-started-page/get-started-card/index.js
ianlin: Hey @j111q, same as the above question, do we see the two ToS CTA as two different click events?

js/src/get-started-page/get-started-card/index.js
eason9487: > My understanding of events aggregation is if we send the events with the same event name and same props, they would be seen as the same one, otherwise not.

I think so. Although the tracking function will capture more data internally, not 100% sure if they could be still distinguished by those ones. It would be easier to make them differentiate from the tracking data GLA sent.

> So in this case, if the answer from Jill is we don't need to distinguish the two events, we just keep the code as is.

Yes.

> But if we need to distinguish them, we should send two different event names from each click. Do I understand it correctly?

I think we could add another event property. For instance, [the `gla_chart_tab_click` event](https://github.com/woocommerce/google-listings-and-ads/blob/a819c2ea56bfe9a4b871c0b5f379e4aa560c033b/js/src/reports/summary-section.js#L23-L25). In this way, if an analyst wants to know "What is the difference in conversion rates between two CTA buttons to the onboarding setup?", they can aggregate data into two groups by the same event name and, for example, `context`. And if an analyst wants to know "What the conversion rate from the get started page to the onboarding setup", they can just use the same event name to aggregate data.

ianlin: Thanks for the comprehensive explanation! I'll see what would be the path to go next week.

js/src/get-started-page/get-started-card/index.js
ianlin: Thanks, updated in 45622a3.

js/src/get-started-page/get-started-card/index.js
ianlin: I updated the code in 5faf81d, I think it's more flexible as we can both see them as the same event or two different events simply filtering by the new property `context`. Thanks @eason9487!

js/src/get-started-page/get-started-card/index.js
ianlin: As above, I updated the code in 5faf81d and it's more flexible as we can both see them as the same event or two different events simply filtering by the new property `context`. Thanks @eason9487!

js/src/get-started-page/get-started-card/index.scss
ianlin: Thanks for catching this @eason9487 as well. I ended up adding the scss file under `js/src/css/abstracts` and named it `_gla_variables.scss` and also added `@import "_gla_variables"` in `webpack.config.js` as this was the only way I could make it work without extra `@import` from other scss files. The commit for this is 302d769.

js/src/get-started-page/get-started-card/index.scss
ianlin: Thanks! Updated in a819c2e.

js/src/get-started-page/get-started-card/index.scss
ianlin: Ooops, updated in a819c2e, thanks!

ianlin:
Hey @eason9487, thanks for the review as always. I've added some commits based on your suggestion and it's ready for another review.

js/src/get-started-page/get-started-card/index.js
eason9487: Thanks for the change!

To make the tracking doc up-to-date, it would be nice to have the corresponding updates in their event JSDoc.

https://github.com/woocommerce/google-listings-and-ads/blob/5faf81dc65cd2ce9bdb855098905143a5e85ca97/js/src/utils/recordEvent.js#L160-L166
https://github.com/woocommerce/google-listings-and-ads/blob/5faf81dc65cd2ce9bdb855098905143a5e85ca97/js/src/get-started-page/get-started-card/index.js#L24
https://github.com/woocommerce/google-listings-and-ads/blob/5faf81dc65cd2ce9bdb855098905143a5e85ca97/js/src/get-started-page/get-started-with-video-card/index.js#L27

eason9487 (approved):
Thanks for addressing my concerns! Testing well, and LGTM. Left a non-blocking comment about updating the event tracking doc.

js/src/get-started-page/get-started-card/index.js
ianlin: Ah, miss the event JSDoc again. Added in d5cf7d3 and noted, thanks!

PR #1461, opened 2022-04-29
Fix syntax errors in JSDoc and the incorrect relative paths when generating the tracking README

eason9487:

- Fix syntax errors in the JSDoc of `TreeSelectControl`.
- Fix the incorrect relative paths when generating the tracking README.
- Prevent the false warnings `@ignore tag does not permit a value` from showing when running `doc:tracking` script.
- Update the tracking README file.

1. Run `npm run doc:tracking` to generate tracking README. The script should be able to run without errors.
2. Go to the [tracking README](https://github.com/woocommerce/google-listings-and-ads/blob/fb034f12fbc1411fb8568362a54794d66076f4c2/src/Tracking/README.md) to check if the hyperlinks point to the right source file.


ecgan (approved):
Summary: LGTM 🙂 👍 

I checkout `develop` branch and run `npm run doc:tracking`, there was an error that says `Error: Invalid docstring *`.

I checkout this PR branch, and I can run the command successfully. However, there were two warnings in the command line like below:

```text
WARNING: The @ignore tag does not permit a value; the value will be ignored. File: index.js, line: 1
WARNING: The @ignore tag does not permit a value; the value will be ignored. File: index.js, line: 42
```

I search for `@ignore` in the code base but I don't find anything related. I'm not sure if this is just me? @eason9487 , do you see the same warning?

The generated tracking readme doc looks good. 

There are two findings that is outside the scope of this PR, just want to note them down here:

1. Default path: In woo-tracking-jsdoc [readme](https://github.com/woocommerce/google-listings-and-ads/blob/fb034f12fbc1411fb8568362a54794d66076f4c2/woo-tracking-jsdoc/README.md) and [publish.js](https://github.com/woocommerce/google-listings-and-ads/blob/fb034f12fbc1411fb8568362a54794d66076f4c2/woo-tracking-jsdoc/publish.js#L47), we have default path set to "TRACKING.md". I think it would be nice to make use of that default path, instead of inside our current "src/Tracking" folder, which is usually for PHP things.

2. Exports: The generated [tracking README](https://github.com/woocommerce/google-listings-and-ads/blob/fb034f12fbc1411fb8568362a54794d66076f4c2/src/Tracking/README.md) contains quite a number of "exports" (see screenshot below), and they point to the usage of code like `export default function EditStoreAddress()`. It just makes me wonder that maybe we shouldn't do `export default function ...`, or maybe there is a way to tweak woo-tracking-jsdoc to handle it better.

image

Since the PR looks good and nothing is blocking, I'm approving this. 👍

eason9487:
@ecgan, thanks for the review!

> There are two findings that is outside the scope of this PR, just want to note them down here:
> 1. Default path: In woo-tracking-jsdoc [readme](https://github.com/woocommerce/google-listings-and-ads/blob/fb034f12fbc1411fb8568362a54794d66076f4c2/woo-tracking-jsdoc/README.md) and [publish.js](https://github.com/woocommerce/google-listings-and-ads/blob/fb034f12fbc1411fb8568362a54794d66076f4c2/woo-tracking-jsdoc/publish.js#L47), we have default path set to "TRACKING.md". I think it would be nice to make use of that default path, instead of inside our current "src/Tracking" folder, which is usually for PHP things.

I don't have a concrete opinion about it and it's out of this PR scope. Maybe it could be a 📜  topic to be discussed separately.
 
> 2. Exports: The generated [tracking README](https://github.com/woocommerce/google-listings-and-ads/blob/fb034f12fbc1411fb8568362a54794d66076f4c2/src/Tracking/README.md) contains quite a number of "exports" [...]. It just makes me wonder that maybe we shouldn't do `export default function ...`, or maybe there is a way to tweak woo-tracking-jsdoc to handle it better.

It's `jsdoc` bug, and there's a related discussion and workaround in #1369:

image

I searched the `jsdoc` GH repo and found many related issues reported. Unless this causes problems for the readers of this document, I think we could wait for it to be fixed from `jsdoc`.

> I checkout this PR branch, and I can run the command successfully. However, there were two warnings in the command line like below:
> [...]
> I search for `@ignore` in the code base but I don't find anything related. I'm not sure if this is just me?

Looks like It's caused by the same `jsdoc` bug above. The false warnings can be removed by using another syntax for the `CompareFilter` component. I updated it in https://github.com/woocommerce/google-listings-and-ads/pull/1461/commits/1c031b50330a284a50b4499cf20df6707dc68c1e.

ecgan:
@eason9487 ,

> The false warnings can be removed by using another syntax for the `CompareFilter` component. I updated it in [1c031b5](https://github.com/woocommerce/google-listings-and-ads/commit/1c031b50330a284a50b4499cf20df6707dc68c1e).

Thanks for the update! I checkout the merged `develop` branch and run `npm run doc:tracking`, and I don't see the warnings anymore. 🙂 👍

PR #1464, opened 2022-04-29
GLA Request Review - Adding Cache to Review Request Controller

puntope:

Closes part of #1165 

I decided to implement cache in the responses. The advantages of this will be:

- Avoid hitting unnecessarily the Google API and VIPGO server
- Allowing to save the status after requesting a review. This will be useful since Google takes a few minutes to update between statuses (ie, from ELIGIBLE to UNDER_REVIEW) hence the user would be able to perform several requests. With this cache we can save the new status manually after the request giving google time to update the status.

The way is works is vary basic, just a transient with 20 mins lifetime that stores the response from Google. 

Screenshot 2022-04-29 at 14 53 13

In case no transient available or expired transients we just perform the response normally and save the status in the transient.

Notice we also added the lifetime to the cooldown period. Since the cooldown period is now showing the time, we want to show the current cooldown + the 20 mins cache to be sure the user is not getting a cached cooldown time in the past. 

```

$response = [
				                 'freeListingsProgram' => [
					                 'status' => 200,
					                 'data'   => [
						                 'regionStatuses' => [
							                 [
								                 'regionCodes'                      => [ 'US' ],
								                 'eligibilityStatus'                => 'DISAPPROVED',
								                 'reviewEligibilityStatus'          => 'INELIGIBLE',
								                 'reviewIneligibilityReasonDetails' => [
									                 'cooldownTime' => "2022-04-27T10:58:51Z" // 27/04/2022
								                 ]
							                 ]
						                 ]
					                 ]
				                 ]
			                 ]
```
1. Checkout this PR
2. Go to product feed and wait until the Account status appears.
3. Go to DB and see that `_transient_gla_mc_account_review`  as well as `_transient_timeout_gla_mc_account_review` is present in `wp_options` table 
4. Go to `src/API/Site/Controllers/MerchantCenter/RequestReviewController.php::get_review_read_callback()` 
5. In line 72 Use the response provided above instead of the API call using `$this->middleware->get_account_review_status()`. 
6. Refresh product feed and wait until the Account status appears. You still see the previous response not the new one hardcoded.
7.  Go to DB again and change  `_transient_timeout_gla_mc_account_review` to 0 to force refresh (you can also delete the transients)
8. Refresh product feed again, see that the new response hardcoded appears.
9. Go to DB again and see how the transients were updated.

### DIFF
--- a/src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
+++ b/src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
@@ -10,7 +10,6 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Google\RequestReviewStatuses;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\Transients;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
-use Psr\Container\ContainerInterface;
 use WP_REST_Request as Request;
 use WP_REST_Response as Response;
 use Exception;
@@ -25,30 +24,18 @@ defined( 'ABSPATH' ) || exit;
 class RequestReviewController extends BaseOptionsController {
 
 	/**
-	 * @var ContainerInterface
-	 */
-	protected $container;
-
-	/**
-	 * @var Middleware
-	 */
-	protected $middleware;
-
-	/**
-	 * @var RequestReviewStatuses
-	 */
-	protected $request_review_statuses;
-
-	/**
 	 * RequestReviewController constructor.
 	 *
-	 * @param ContainerInterface $container
+	 * @param RESTServer            $server
+	 * @param Middleware            $middleware
+	 * @param RequestReviewStatuses $request_review_statuses
+	 * @param TransientsInterface   $transient
 	 */
-	public function __construct( ContainerInterface $container ) {
-		$this->container = $container;
-		parent::__construct( $container->get( RESTServer::class ) );
-		$this->middleware              = $container->get( Middleware::class );
-		$this->request_review_statuses = $container->get( RequestReviewStatuses::class );
+	public function __construct( RESTServer $server, Middleware $middleware, RequestReviewStatuses $request_review_statuses, TransientsInterface $transient ) {
+		parent::__construct( $server );
+		$this->middleware              = $middleware;
+		$this->request_review_statuses = $request_review_statuses;
+		$this->transient = $transient;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
```diff
@@ -21,18 +23,19 @@
...
 	 * @param Middleware            $middleware
 	 * @param RequestReviewStatuses $request_review_statuses
+	 * @param TransientsInterface   $transient
 	 */
-	public function __construct( RESTServer $server, Middleware $middleware, RequestReviewStatuses $request_review_statuses ) {
+	public function __construct( RESTServer $server, Middleware $middleware, RequestReviewStatuses $request_review_statuses, TransientsInterface $transient ) {
 		parent::__construct( $server );
 		$this->middleware              = $middleware;
 		$this->request_review_statuses = $request_review_statuses;
+		$this->transient               = $transient;
```
mikkamp: This is the interface for transients, so I think it's best to name it plural `transients`. Also we should probably declare it in the class so we can make it protected (otherwise it defaults to public).

src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
```diff
@@ -116,4 +124,28 @@ protected function get_schema_properties(): array {
...
 	}
+
+	/**
+	 * Save the Account Review Status data inside a transient for caching purposes.
+	 *
+	 * @param array $value The Account Review Status data to save in the transient
+	 */
+	private function set_cached_review_status( $value ): void {
+		$this->transient->set(
+			Transients::MC_ACCOUNT_REVIEW,
```
mikkamp: In other classes we use the TransientsInterface for this (which means we can have one import less at the top of the file). We use the same pattern for options which means we are more flexible in case we ever use a different implementation to store options/transients.

src/Google/RequestReviewStatuses.php
```diff
@@ -78,9 +80,31 @@ public function get_statuses_from_response( array $response ) {
...
+	public function get_account_review_lifetime(): int {
+		return apply_filters( 'woocommerce_gla_mc_account_review_lifetime', self::MC_ACCOUNT_REVIEW_LIFETIME );
+	}
+
+	/**
+	 * @param int $cooldown The cooldown in PHP format (seconds)
+	 *
+	 * @return int The cooldown in miliseconds and adding the lifetime cache
+	 */
+	private function parse_cooldown( $cooldown ) {
```
mikkamp: Probably left over from a refactor, but it seems the `$cooldown` is already passed through `intval( strtotime() )` before calling the function, so it's always passed to this function as an int.
So we can force the function parameter to int and probably rename the function to `calculate` or `get` instead of `parse`.

mikkamp (commented):
Thanks for the changes, they are working as expected, I'm getting the results saved in the transient.

I just left a few smaller comments about some parts which could be renamed to make it clearer what we are doing.

puntope:
Howdy @mikkamp, thanks for the review :) 

I implemented all of your suggestions. This PR is ready for another review.

mikkamp (approved):
Thanks for those changes. It looks good to me.

PR #1465, opened 2022-04-29
Fix TreeSelectControl label UI to be consistent with other input label

### LINKED ISSUE #1453: TreeSelectControl label is different from other input control label
### Describe the bug:

When I look at the shipping rate modal, minimum order modal and shipping time modal, the TreeSelectControl label UI is different from the other input control label. See screenshot below:

image

Another place is in the Setup Ads flow:

image

This issue happens in the `develop` branch (commit a8c6c13279f5e68397a1e2d485bd5bdc9790a53d).

### Steps to reproduce:

1. Go to Setup MC or Edit Free Listings, open shipping rate modal, minimum order modal and shipping time modal.
2. Go to Setup Ads, Step 2, Audience section. 

### Expected behavior:

TreeSelectControl label UI should look consistent with other input control label.

### Actual behavior:

TreeSelectControl label is different from other input control label - it has bigger font-size and bigger padding-bottom.

ecgan:

Closes #1453.

In this PR, we fix the TreeSelectControl / SupportedCountrySelect label UI so that it is consistent with other input control label UI, i.e. have the same font-size and padding-bottom.

Impacted areas are in the screenshots below.

### Screenshots:

Shipping rate - modification is done in SupportedCountrySelect component, see commit 7f8d8e65c48a6416d63f1b6ad85f46cd3f1ac24a: 

image

Note: 

- Minimum order modals also uses SupportedCountrySelect component.
- Shipping time modals in Edit Free Listings also uses SupportedCountrySelect component.

Shipping times add modal in Setup MC - this is impacted because it was not using the SupportedCountrySelect label, I have changed it in 3a5b89d09f4f075dbd3a7e691f5acaa18d5fd0fc:

image

Shipping times edit modal in Setup MC - this is impacted because it was not using the SupportedCountrySelect label, I have changed it in 3a5b89d09f4f075dbd3a7e691f5acaa18d5fd0fc:

image

Create paid ads campaign - this is impacted because I have removed the unneeded CSS for the campaign form content in 8b47081160893010a80863cc44610aa96edad1ba:

image

Edit paid ads campaign - this is impacted because I have removed the unneeded CSS for the campaign form content in 8b47081160893010a80863cc44610aa96edad1ba:

image

1. Go to Setup MC Step 3, look into shipping rate add and edit modals, make sure the labels look consistent as per screenshots above.
2. Look into shipping times add modal, make sure the labels look consistent as per "shipping times add modal" screenshots above.
3. Look into shipping times edit modal, make sure the labels look consistent as per "shipping times edit modal" screenshots above.
4. Complete the Setup MC flow, and go to create a new paid ads campaign. The labels should be as per "create paid ads campaign" screenshot above.
5. Go to edit a paid ads campaign. The labels should be as per "edit paid ads campaign" screenshot above.

### Changelog entry

> Fix - Label UI for selecting countries (TreeSelectControl / SupportedCountrySelect).

### DIFF
--- a/js/src/components/paid-ads/campaign-form-content.scss
+++ b/js/src/components/paid-ads/campaign-form-content.scss
@@ -25,12 +25,4 @@
 		margin: 0;
 		font-style: italic;
 	}
-
-	.woocommerce-tree-select-control__label {
-		font-size: 13px;
-	}
-
-	.woocommerce-tree-select-control__help {
-		font-style: italic;
-	}
 }
--- a/js/src/components/supported-country-select/supported-country-select.js
+++ b/js/src/components/supported-country-select/supported-country-select.js
@@ -3,12 +3,14 @@
  */
 import { __ } from '@wordpress/i18n';
 import { useViewportMatch } from '@wordpress/compose';
+import classnames from 'classnames';
 
 /**
  * Internal dependencies
  */
 import TreeSelectControl from '.~/components/tree-select-control';
 import useMCCountryTreeOptions from './useMCCountryTreeOptions';
+import './supported-country-select.scss';
 
 /**
  * @typedef {import('.~/data/actions').CountryCode} CountryCode
@@ -22,10 +24,11 @@ import useMCCountryTreeOptions from './useMCCountryTreeOptions';
  *
  * @param {Object} props React props.
  * @param {Array<CountryCode>} [props.countryCodes] Country codes for converting to tree-based option structure. It will use all MC supported countries if not specified.
- * @param {Object} props.restProps Props to be forwarded to TreeSelectControl.
+ * @param {string} props.className Class name. Note that when there is a Form's `{...getInputProps('fieldname')}` being passed into this component, the className might be null.
  */
 export default function SupportedCountrySelect( {
 	countryCodes,
+	className,
 	...restProps
 } ) {
 	const maxVisibleTags = useViewportMatch( 'medium', '<' ) ? 5 : 10;
@@ -33,6 +36,10 @@ export default function SupportedCountrySelect( {
 
 	return (
 		<TreeSelectControl
+			className={ classnames(
+				'gla-supported-country-select',
+				className
+			) }
 			placeholder={ __(
 				'Start typing to filter countries…',
 				'google-listings-and-ads'
--- a/js/src/components/supported-country-select/supported-country-select.js
+++ b/js/src/components/supported-country-select/supported-country-select.js
@@ -22,6 +22,8 @@ import './supported-country-select.scss';
  * This component is for selecting countries under Google Merchant Center supported countries.
  * And the selectable countries can be further limited by the `countryCodes` prop.
  *
+ * All other props (except `countryCodes` prop) are forwarded to the underlying TreeSelectControl component.
+ *
  * @param {Object} props React props.
  * @param {Array<CountryCode>} [props.countryCodes] Country codes for converting to tree-based option structure. It will use all MC supported countries if not specified.
  * @param {string} [props.className] Class name. Note that when there is a Form's `{...getInputProps('fieldname')}` being passed into this component, the className might be null.
--- a/js/src/components/supported-country-select/supported-country-select.js
+++ b/js/src/components/supported-country-select/supported-country-select.js
@@ -24,7 +24,7 @@ import './supported-country-select.scss';
  *
  * @param {Object} props React props.
  * @param {Array<CountryCode>} [props.countryCodes] Country codes for converting to tree-based option structure. It will use all MC supported countries if not specified.
- * @param {string} props.className Class name. Note that when there is a Form's `{...getInputProps('fieldname')}` being passed into this component, the className might be null.
+ * @param {string} [props.className] Class name. Note that when there is a Form's `{...getInputProps('fieldname')}` being passed into this component, the className might be null.
  */
 export default function SupportedCountrySelect( {
 	countryCodes,
--- a/js/src/components/supported-country-select/supported-country-select.scss
+++ b/js/src/components/supported-country-select/supported-country-select.scss
@@ -1,6 +1,10 @@
 .gla-supported-country-select {
-	label {
+	.woocommerce-tree-select-control__label {
 		padding-bottom: 4px;
 		font-size: 13px;
+	}
+
+	.woocommerce-tree-select-control__help {
+		font-style: italic;
 	}
 }
--- a/js/src/components/supported-country-select/supported-country-select.scss
+++ b/js/src/components/supported-country-select/supported-country-select.scss
@@ -1,0 +1,6 @@
+.gla-supported-country-select {
+	label {
+		padding-bottom: 4px;
+		font-size: 13px;
+	}
+}
--- a/js/src/components/tree-select-control/index.scss
+++ b/js/src/components/tree-select-control/index.scss
@@ -28,7 +28,6 @@ $muriel-box-shadow-8dp: 0 5px 5px -3px rgba(0, 0, 0, 0.2),
 	&__help {
 		margin-top: $gap-smallest;
 		line-height: 16px;
-		font-style: italic;
 		font-size: 12px;
 		color: $gray-700;
 	}
--- a/js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
+++ b/js/src/setup-mc/setup-stepper/setup-free-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
@@ -82,20 +82,16 @@ const AddTimeModal = ( props ) => {
 						onRequestClose={ onRequestClose }
 					>
 						<VerticalGapLayout>
-							<div>
-								<div className="label">
-									{ __(
-										'If customer is in',
-										'google-listings-and-ads'
-									) }
-								</div>
-								<AudienceCountrySelect
-									onDropdownVisibilityChange={
-										setDropdownVisible
-									}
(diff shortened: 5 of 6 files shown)


js/src/components/paid-ads/campaign-form-content.scss
```diff
@@ -25,12 +25,4 @@
...
 		font-style: italic;
 	}
-
-	.woocommerce-tree-select-control__label {
-		font-size: 13px;
-	}
-
-	.woocommerce-tree-select-control__help {
-		font-style: italic;
-	}
```
ecgan, line 35: It seems like these are not needed, so I have removed them.

js/src/components/supported-country-select/supported-country-select.js
```diff
@@ -22,17 +24,22 @@ import useMCCountryTreeOptions from './useMCCountryTreeOptions';
  *
  * @param {Object} props React props.
  * @param {Array<CountryCode>} [props.countryCodes] Country codes for converting to tree-based option structure. It will use all MC supported countries if not specified.
- * @param {Object} props.restProps Props to be forwarded to TreeSelectControl.
+ * @param {string} props.className Class name. Note that when there is a Form's `{...getInputProps('fieldname')}` being passed into this component, the className might be null.
```
ecgan: As stated in the JSDoc comment here, the `className` props is `null` because it was passed in by `Form`'s `{ ...getInputProps('...') }`. See: https://github.com/woocommerce/woocommerce-admin/blob/78d0e10a99676c92650a5248917207e7304f5c1b/packages/components/src/form/index.js#L130

js/src/components/supported-country-select/supported-country-select.js
```diff
@@ -22,17 +24,22 @@ import useMCCountryTreeOptions from './useMCCountryTreeOptions';
...
 } ) {
 	const maxVisibleTags = useViewportMatch( 'medium', '<' ) ? 5 : 10;
 	const treeOptions = useMCCountryTreeOptions( countryCodes );
 
 	return (
 		<TreeSelectControl
+			className={ classnames(
+				'gla-supported-country-select',
+				className
+			) }
```
ecgan, line 44: We use `classnames` function here, it is able to handle the null `className` prop, and it also allows us to support custom className passed in by future SupportedCountrySelect consumer.

ecgan (commented):
Added some comments below on the findings I have when I was working on this PR.

js/src/components/supported-country-select/supported-country-select.js
puntope: Just pointing out here that in Tree Select Component we also use classnames internally.

js/src/components/supported-country-select/supported-country-select.js
puntope: This comment is a bit confusing for me. Shouldn't be then `{string|null}` instead of just string and then explaining the case with getInputProps? As well as  set up the prop as optional. 

Also I don't understand why we removed  the line regarding

`* @param {Object} props.restProps Props to be forwarded to TreeSelectControl.`

puntope (approved):
✅  LGTM thanks for the fix

💅  Just one comment regarding your comment and the deleted line regarding `restProps`.

Also I discovered a weird behaviour on the dropdown open/close when there is a help prop. 

Reported here: https://github.com/woocommerce/google-listings-and-ads/issues/1468 with the suggestion of move the helper prop to the consumer `<SupportedCountrySelect>` instead of the `<TreeSelectControl>`

js/src/components/supported-country-select/supported-country-select.js
ecgan: > Shouldn't be then `{string|null}` instead of just string and then explaining the case with getInputProps? 

IMO, from the perspective of `SupportedCountrySelect`, className should always be string, and I see WC Form passing className null as a bug.

> As well as set up the prop as optional.

Okay, I can make it into `[props.className]`. 

> Also I don't understand why we removed the line regarding
> 
> `* @param {Object} props.restProps Props to be forwarded to TreeSelectControl.`

Because there isn't a prop called `restProps`? 😅 

```jsx
<SupportedCountrySelect
	restProps={ {} }  // no such prop...?
/>
```

ecgan: > Okay, I can make it into `[props.className]`.

I have updated this in [`ab675ab` (#1465)](https://github.com/woocommerce/google-listings-and-ads/pull/1465/commits/ab675ab8aab972a7a62740d67f1910df5a25b9fa).

js/src/components/paid-ads/campaign-form-content.scss
eason9487: Since the help prop in several control components of `@wordpress/components` and `@wocommerce/components` has not set the italic font style as default, I intended to set the italic here and remove the italic style from TreeSelectControl, but I missed that. 🤦‍♂️ 

Ref:
- The [base help style](https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4012.0.8/packages/components/src/base-control/styles/base-control-styles.js#L29-L34) in `@wordpress/components` has no `italic` style.
- The [help of SelectControl](https://github.com/woocommerce/woocommerce-admin/blob/v2.7.2/packages/components/src/select-control/control.js#L254-L259) in `@wocommerce/components` doesn't have any default style.

ecgan: @eason9487 , thanks for the context, not a problem, it happens to everyone of us. 😄 I remember the italic style was specifically there for Google Listings and Ads, and WooCommerce Design Library, but I just check WC Design Library figma (LcsjAJaTHAQBgIzDfP9BtK-fi-3523%3A9319), apparently there is no more italic style for helper text. 🤔 

> I intended to set the italic here and remove the italic style from TreeSelectControl, but I missed that.

Should we do it in this PR and remove the italic style from TreeSelectControl? Right now we still have the italic style in TreeSelectControl (see screenshot in the PR description). I'm not sure if we should really do this, I remember seeing italic helper text in TreeSelectControl P2 and figma design, but couldn't find it now. The closest thing I find is Edit Paid Campaign figma design (https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=5519%3A115701).

eason9487: Currently, there are three places having the helper text under the selector:
1. Step 2 of onboarding setup
2. Step 1 of editing free listings
3. Edit a paid campaign.

But the first two don't use TreeSelectControl's `help`. Therefore, if they are expected to be changed to use TreeSelectControl's `help` in the future, I think we could move the `italic` style to supported-country-select.scss. Otherwise, it should be ok to keep it here. And both ways remove the `italic` style from the tree-select-control/index.scss.

js/src/components/supported-country-select/supported-country-select.js
puntope: > Because there isn't a prop called restProps? 😅

It's actually.

Check Line 50 of `js/src/components/supported-country-select/supported-country-select.js`

PS. I don't really mind if we document `restProps` or not in JSDOC. I guess the attempt here was just indicating that `SupportedCountrySelect` send `...restProps` into `TreeSelect` component.

js/src/components/paid-ads/campaign-form-content.scss
ecgan: > Therefore, if they are expected to be changed to use TreeSelectControl's help in the future, I think _we could move the italic style to supported-country-select.scss_. Otherwise, it should be ok to keep it here. And _both ways remove the italic style from the tree-select-control/index.scss_.

Okay, gotcha. Since the WordPress and WooCommerce components don't have the italic style, yeah I think it makes sense to remove the italic style from TreeSelectControl, and I think it is better to move it into SupportedCountrySelect. I'll look into this.

js/src/components/supported-country-select/supported-country-select.js
ecgan: > It's actually.

The `restProps` in the code is "an object containing all other props except countryCodes and className". `restProps` itself is not a prop that we can access like `props.restProps`, and we do not assign `restProps` like `<SupportedCountrySelect restProps={ {} } />`.

> I guess the attempt here was just indicating that `SupportedCountrySelect` send `...restProps` into `TreeSelect` component.

Good point. I added JSDoc about this in [`5983adc` (#1465)](https://github.com/woocommerce/google-listings-and-ads/pull/1465/commits/5983adceebb5cbcefb4f985a6b090da72e78e9d7). 🙂

js/src/components/paid-ads/campaign-form-content.scss
ecgan: I have made the changes in [`aab170e` (#1465)](https://github.com/woocommerce/google-listings-and-ads/pull/1465/commits/aab170e033f2ba5046ed0cef1f36c4446f60d483) and [`8e45a3e` (#1465)](https://github.com/woocommerce/google-listings-and-ads/pull/1465/commits/8e45a3e7d28c80524cb7a5769769dd3d0e992f5b). @eason9487 , would you like to have a review on this?

eason9487: > I have made the changes in [`aab170e` (#1465)](https://github.com/woocommerce/google-listings-and-ads/pull/1465/commits/aab170e033f2ba5046ed0cef1f36c4446f60d483) and [`8e45a3e` (#1465)](https://github.com/woocommerce/google-listings-and-ads/pull/1465/commits/8e45a3e7d28c80524cb7a5769769dd3d0e992f5b). @eason9487 , would you like to have a review on this?

The changes look good. Thanks for the help! 🙏

PR #1466, opened 2022-05-01
Update tested to WordPress 6.0 and WooCommerce 6.5

layoutd:

Bumping the WooCommerce and WordPress tested up to tags after testing with **WooCommerce 6.5.0-rc.1** and **WordPress 6.0.0-beta.1**.

### Changelog entry

> Tweak - WC 6.5 compatibility.
> Tweak - WordPress 6.0 compatibility.

### DIFF
--- a/google-listings-and-ads.php
+++ b/google-listings-and-ads.php
@@ -8,11 +8,11 @@
  * Author URI: https://woocommerce.com/
  * Text Domain: google-listings-and-ads
  * Requires at least: 5.7
- * Tested up to: 5.9
+ * Tested up to: 6.0
  * Requires PHP: 7.3
  *
  * WC requires at least: 5.8
- * WC tested up to: 6.3
+ * WC tested up to: 6.5
  * Woo:
  *
  * @package WooCommerce\Admin


mikkamp (approved):
Looks good, thanks for all the compatibility testing.
We are pretty up to date with this extension on compatibility so this looks like a pretty safe thing to bump.

Do we need a follow up issue to remember to bump WP versions for unit tests once 6.0 has been released? For WC I think we should be fine since we always use `latest`.

layoutd:
Thanks for the 👀  @mikkamp 

> Do we need a follow up issue to remember to bump WP versions for unit tests once 6.0 has been released? For WC I think we should be fine since we always use latest.

Is this where we specify WP version? If so, it looks like it has `latest` as well ([example](https://github.com/woocommerce/google-listings-and-ads/actions/runs/2293229439)):
https://github.com/woocommerce/google-listings-and-ads/blob/6b357b73ab869481d53caba73060d47b5453de25/.github/workflows/php-unit-tests.yml#L49

We _would_ need to remove 5.7, however (Or, ideally, find a way to automate L, L-1, and L-2).

mikkamp:
Yes, I meant to update the older versions in the matrix. Also PHP 8.0 is compatible with WP 5.9+ so it could test on both 5.9 and 6.0 once it's released.
An automated way to do L-1 and L-2 would be nice, as long as it isn't too much work to implement, could be a nice cooldown project. If we do rework this, we could decide to revisit what matrix we want to be testing and whether we want to test WC L-2 at least for some versions of PHP/WP.

PR #1202, opened 2022-01-18
Change the shipping rate API response and request args

nima-karimi:
### Changes proposed in this Pull Request:

Implements part of #1117. Based on #1171 and #1192

This PR changes both the response and request arguments for the shipping rate API along with the batch API. Some of the prominent changes include:

- Accept the country code as part of the shipping rate data structure in requests instead of as a separate argument
- Rename the `country_code` parameter in both the response and request to `country` for **ALL shipping rates endpoints**. This is to align it with how it's represented internally and in the DB.
- Return the shipping rate ID in the response. This is useful because we are gradually moving toward having multiple rates per country and relying on the country_code as an identifier isn't gonna work anymore.
- Change the `DELETE` endpoint to only work with IDs instead of country codes (because of the above).
- Add an edit rate endpoint to allow editing a shipping rate by its ID

The request schema for creating a shipping rate is modified to the following (note that the `country_code` parameter is renamed to `country` and it's moved inside the rate data structure):

`POST mc/shipping/rates`
```JSON
{
      "country": "PT",
      "method": "flat_rate",
      "currency": "USD",
      "rate": 80,
      "options": {
        "free_shipping_threshold": "200"
      }
}
```

The batch create API request schema is also changed to the following:

`POST mc/shipping/rates/batch`
```JSON
{
  "rates": [
    {
      "country": "PT",
      "method": "flat_rate",
      "currency": "USD",
      "rate": 80,
      "options": {
        "free_shipping_threshold": "2000"
      }
    },
    {
      "country": "DE",
      "method": "flat_rate",
      "currency": "USD",
      "rate": 50,
      "options": {
        "free_shipping_threshold": "1200"
      }
    },
    {
      "country": "US",
      "method": "flat_rate",
      "currency": "USD",
      "rate": 10
    }
  ]
}
```

---

The response for getting all the available shipping rates is an array of shipping rate objects (without any `rates` parameter):

`GET mc/shipping/rates/`
```JSON
[
    {
        "id": "21",
        "country": "DE",
        "method": "flat_rate",
        "currency": "USD",
        "rate": "50",
        "options": {
            "free_shipping_threshold": 1200
        }
    },
    {
        "id": "33",
        "country": "PT",
        "method": "flat_rate",
        "currency": "USD",
        "rate": "80",
        "options": {
            "free_shipping_threshold": 2000
        }
    },
    {
        "id": "32",
        "country": "US",
        "method": "flat_rate",
        "currency": "USD",
        "rate": "10"
    }
]
```

[...]

PR #1473, opened 2022-05-03
Fix validation for empty shipping rates and shipping times

### LINKED ISSUE #1435: The shipping rates and times form misses the invalid value check
### Describe the bug:

The shipping rates and times form misses the invalid value check. This problem happens on:
- Onboarding setup page
   - Shipping rate edit modal
   - Shipping setup form
- Free listings edit page
   - Shipping rate edit modal
   - Shipping time edit modal
   - Shipping setup form

### Steps to reproduce:

https://user-images.githubusercontent.com/17420811/164143584-85875c2b-3657-4f47-9cc7-bff61142f3af.mp4

1. Go to step 2 of the edit free listings page.
1. Click on the "Edit" link to open the edit modal and delete the shipping times. Repeat this step till all shipping times are deleted.
1. Click on the "Edit" link again to open the edit modal.
1. When the shipping time value is empty, the update button on the modal is still enabled.
1. Click on the update button on the modal.
1. When there is an empty shipping time value, the "Save changes" button at the bottom of page is still enabled.

### Expected behavior:

When any shipping rate/time value is empty, the save/update button on the related modal/form should be disabled.

- Onboarding setup page
   - [ ] Shipping rate edit modal
   - [ ] Shipping setup form
- Free listings edit page
   - [ ] Shipping rate edit modal
   - [ ] Shipping time edit modal
   - [ ] Shipping setup form

ecgan:

Closes #1435.

This PR modifies some validation handling for shipping rates and shipping times in Setup MC and Edit Free Listings to fix the issue #1435.

### Screenshots:

📹  Demo video with my voice:

https://user-images.githubusercontent.com/417342/166510966-7ca9d85a-1c14-4c94-be9f-a616829eafe8.mp4

1. Go to Setup MC Step 3.
2. Delete all the shipping rates and shipping times.
3. Click on the shipping rate "Edit" button to open the edit modal. The update button in the modal should be disabled.
4. Click on the shipping time "Edit" button to open the edit modal. The update button in the modal should be disabled.
5. The "Continue" button at the bottom of page should only be enabled when you have all the required valid values for shipping rates and shipping times.
6. Repeat the above steps for Edit Free Listings.

### Changelog entry

> Fix - Validation for shipping rates and shipping times in Setup MC and Edit Free Listings.

### DIFF
--- a/js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
+++ b/js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
@@ -27,17 +27,14 @@ const AddTimeModal = ( { countries, onRequestClose, onSubmit } ) => {
 	const [ dropdownVisible, setDropdownVisible ] = useState( false );
 
 	const handleSubmitCallback = ( values ) => {
-		onSubmit( {
-			countries: values.countryCodes,
-			time: values.time,
-		} );
+		onSubmit( values );
 		onRequestClose();
 	};
 
 	return (
 		<Form
 			initialValues={ {
-				countryCodes: countries,
+				countries,
 				time: 0,
 			} }
 			validate={ validateShippingTimeGroup }
@@ -80,7 +77,7 @@ const AddTimeModal = ( { countries, onRequestClose, onSubmit } ) => {
 								onDropdownVisibilityChange={
 									setDropdownVisible
 								}
-								{ ...getInputProps( 'countryCodes' ) }
+								{ ...getInputProps( 'countries' ) }
 							/>
 							<AppInputNumberControl
 								label={ __(
--- a/js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
+++ b/js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
@@ -13,6 +13,7 @@ import AppModal from '.~/components/app-modal';
 import AppInputNumberControl from '.~/components/app-input-number-control';
 import VerticalGapLayout from '.~/components/vertical-gap-layout';
 import SupportedCountrySelect from '.~/components/supported-country-select';
+import validateShippingTime from '.~/utils/validateShippingTime';
 
 /**
  * Form to add a new time for selected country(-ies).
@@ -25,38 +26,21 @@ import SupportedCountrySelect from '.~/components/supported-country-select';
 const AddTimeModal = ( { countries, onRequestClose, onSubmit } ) => {
 	const [ dropdownVisible, setDropdownVisible ] = useState( false );
 
-	const handleValidate = ( values ) => {
-		const errors = {};
-
-		if ( values.countries.length === 0 ) {
-			errors.countries = __(
-				'Please specify at least one country.',
-				'google-listings-and-ads'
-			);
-		}
-
-		if ( values.time < 0 ) {
-			errors.time = __(
-				'The estimated shipping time cannot be less than 0.',
-				'google-listings-and-ads'
-			);
-		}
-
-		return errors;
-	};
-
 	const handleSubmitCallback = ( values ) => {
-		onSubmit( values );
+		onSubmit( {
+			countries: values.countryCodes,
+			time: values.time,
+		} );
 		onRequestClose();
 	};
 
 	return (
 		<Form
 			initialValues={ {
-				countries,
+				countryCodes: countries,
 				time: 0,
 			} }
-			validate={ handleValidate }
+			validate={ validateShippingTime }
 			onSubmit={ handleSubmitCallback }
 		>
 			{ ( formProps ) => {
@@ -96,7 +80,7 @@ const AddTimeModal = ( { countries, onRequestClose, onSubmit } ) => {
 								onDropdownVisibilityChange={
 									setDropdownVisible
 								}
-								{ ...getInputProps( 'countries' ) }
+								{ ...getInputProps( 'countryCodes' ) }
 							/>
 							<AppInputNumberControl
 								label={ __(
--- a/js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
+++ b/js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
@@ -13,7 +13,7 @@ import AppModal from '.~/components/app-modal';
 import AppInputNumberControl from '.~/components/app-input-number-control';
 import VerticalGapLayout from '.~/components/vertical-gap-layout';
 import SupportedCountrySelect from '.~/components/supported-country-select';
-import validateShippingTime from '.~/utils/validateShippingTime';
+import validateShippingTimeGroup from '.~/utils/validateShippingTimeGroup';
 
 /**
  * Form to add a new time for selected country(-ies).
@@ -40,7 +40,7 @@ const AddTimeModal = ( { countries, onRequestClose, onSubmit } ) => {
 				countryCodes: countries,
 				time: 0,
 			} }
-			validate={ validateShippingTime }
+			validate={ validateShippingTimeGroup }
 			onSubmit={ handleSubmitCallback }
 		>
 			{ ( formProps ) => {
--- a/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.js
+++ b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.js
@@ -10,7 +10,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * @param {ShippingRateGroup} values
  */
-const validate = ( values ) => {
+const validateShippingRateGroup = ( values ) => {
 	const errors = {};
 
 	if ( values.countries.length === 0 ) {
@@ -37,4 +37,4 @@ const validate = ( values ) => {
 	return errors;
 };
 
-export default validate;
+export default validateShippingRateGroup;
--- a/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.js
+++ b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.js
@@ -20,7 +20,7 @@ const validateShippingRateGroup = ( values ) => {
 		);
 	}
 
-	if ( values.rate === undefined ) {
+	if ( ! Number.isFinite( values.rate ) ) {
 		errors.rate = __(
 			'Please enter the estimated shipping rate.',
 			'google-listings-and-ads'
--- a/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
+++ b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
@@ -30,7 +30,7 @@ describe( 'validateShippingRateGroup', () => {
 		expect( errors ).toHaveProperty( 'countries' );
 	} );
 
-	it( 'has errors.rate when values.rate is empty', () => {
+	it( 'has errors.rate when values.rate is undefined', () => {
 		const values = {
 			...validValues,
 			rate: undefined,
--- a/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
+++ b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
@@ -30,10 +30,32 @@ describe( 'validateShippingRateGroup', () => {
 		expect( errors ).toHaveProperty( 'countries' );
 	} );
 
+	it( 'has errors.rate when values.rate is null', () => {
+		const values = {
+			...validValues,
+			rate: null,
+		};
+
+		const errors = validateShippingRateGroup( values );
+
+		expect( errors ).toHaveProperty( 'rate' );
+	} );
+
 	it( 'has errors.rate when values.rate is undefined', () => {
 		const values = {
 			...validValues,
 			rate: undefined,
+		};
+
+		const errors = validateShippingRateGroup( values );
+
+		expect( errors ).toHaveProperty( 'rate' );
+	} );
+
+	it( `has errors.rate when values.rate is ''`, () => {
+		const values = {
+			...validValues,
+			rate: '',
 		};
 
 		const errors = validateShippingRateGroup( values );
--- a/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
+++ b/js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
@@ -1,20 +1,20 @@
 /**
  * Internal dependencies
  */
-import validate from './validate';
+import validateShippingRateGroup from './validateShippingRateGroup';
 
 const validValues = {
 	countries: [ 'US' ],
 	rate: 0,
 };
 
-describe( 'validate', () => {
+describe( 'validateShippingRateGroup', () => {
 	it( 'has no errors properties with validValues', () => {
 		const values = {
 			...validValues,
 		};
 
-		const errors = validate( values );
+		const errors = validateShippingRateGroup( values );
 
 		expect( errors ).toEqual( {} );
 	} );
@@ -25,7 +25,7 @@ describe( 'validate', () => {
 			countries: [],
 		};
 
-		const errors = validate( values );
+		const errors = validateShippingRateGroup( values );
 
 		expect( errors ).toHaveProperty( 'countries' );
 	} );
@@ -36,7 +36,7 @@ describe( 'validate', () => {
 			rate: undefined,
 		};
 
-		const errors = validate( values );
+		const errors = validateShippingRateGroup( values );
 
 		expect( errors ).toHaveProperty( 'rate' );
 	} );
@@ -47,7 +47,7 @@ describe( 'validate', () => {
 			rate: -1,
 		};
 
-		const errors = validate( values );
+		const errors = validateShippingRateGroup( values );
 
 		expect( errors ).toHaveProperty( 'rate' );
 	} );
--- a/js/src/utils/validateShippingTimeGroup.js
+++ b/js/src/utils/validateShippingTimeGroup.js
@@ -13,7 +13,7 @@ const validateShippingTimeGroup = ( values ) => {
 		);
 	}
 
-	if ( values.time === null ) {
+	if ( ! Number.isInteger( values.time ) ) {
 		errors.time = __(
 			'Please enter the estimated shipping time.',
 			'google-listings-and-ads'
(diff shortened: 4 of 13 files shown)


js/src/utils/validateShippingTimeGroup.test.js
```diff
@@ -0,0 +1,54 @@
...
+		const values = {
+			...validValues,
+		};
+
+		const errors = validateShippingTimeGroup( values );
+
+		expect( errors ).toEqual( {} );
+	} );
+
+	it( 'has errors.countries when values.countryCodes.length is 0', () => {
```
eason9487: 💅  "errors.countries" was missed when renaming.

js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
```diff
@@ -25,38 +26,21 @@ import SupportedCountrySelect from '.~/components/supported-country-select';
...
-			);
-		}
-
-		return errors;
-	};
-
 	const handleSubmitCallback = ( values ) => {
-		onSubmit( values );
+		onSubmit( {
+			countries: values.countryCodes,
```
eason9487: ❓  or ❔ 

Before this PR, the same data were declared with two variable names, `countries` and `countryCodes`, across these shipping forms/modals/input controls/grouping utils/tests. In the most recent PRs, both namings continued to be added to the highly relevant components. Wondering if these two names would gradually converge towards one of them?

This PR adjusted some of them but there are still some inconsistent naming within individual components. The current implementation makes the switching between `countries` and `countryCodes` within and between components more complicated, resulting in lower readability.

If at least these shipping modals interface, internal and its dependencies are adjusted to be consistent in this PR first, it might save time for future code reviews.

js/src/utils/validateShippingTimeGroup.js
```diff
@@ -0,0 +1,33 @@
...
+	const errors = {};
+
+	if ( values.countryCodes.length === 0 ) {
+		errors.countryCodes = __(
+			'Please specify at least one country.',
+			'google-listings-and-ads'
+		);
+	}
+
+	if ( values.time === null ) {
```
eason9487: 💅 
The original "empty" value of shipping rate had been `null` and has been changed to `undefined`, and the "empty" value of minimum order is `undefined`. But the "empty" value of shipping time is `null`. The validation function is also based on these different specific verifications, but in fact other invalid values will not be checked. These inconsistencies don't cause bugs but are somewhat a bit confusing to read.

The "empty" value checking could be changed to check the invalid values and types by `Number.isFinite()`. Thus, it ensures that invalid values of shipping time are detected without the need to validate against the specific empty value only.

js/src/utils/validateShippingTimeGroup.test.js
```diff
@@ -0,0 +1,54 @@
...
+		const errors = validateShippingTimeGroup( values );
+
+		expect( errors ).toHaveProperty( 'countryCodes' );
+	} );
+
+	it( 'has errors.time when values.time is empty', () => {
+		const values = {
+			...validValues,
+			time: null,
+		};
```
eason9487: 💅  For testing purposes, testing only the "empty" value (null) might make the test relatively fragile.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.js
```diff
@@ -20,6 +20,13 @@ const validate = ( values ) => {
 		);
 	}
 
+	if ( values.rate === undefined ) {
```
eason9487: 💅  Similar concern as the second paragraph in https://github.com/woocommerce/google-listings-and-ads/pull/1473#discussion_r864608844.

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
```diff
@@ -25,18 +25,29 @@ describe( 'validate', () => {
...
+		const errors = validateShippingRateGroup( values );
 
 		expect( errors ).toHaveProperty( 'countries' );
 	} );
 
+	it( 'has errors.rate when values.rate is empty', () => {
+		const values = {
+			...validValues,
+			rate: undefined,
+		};
```
eason9487: 💅  Similar concern as https://github.com/woocommerce/google-listings-and-ads/pull/1473#discussion_r864612934

js/src/utils/validateShippingTimeGroup.test.js
ecgan: @eason9487 , could you elaborate more on this, maybe with some suggestions / examples? I'm guessing you are suggesting to test for other possible invalid values like `false`, `""`, `undefined`, and I am not sure if that is really necessary? 🤔

js/src/utils/validateShippingTimeGroup.test.js
ecgan: Fixed in [`b750807` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/b7508073e9bc249fd84605332cc3681bc2dace8c). Thanks. 🙂

js/src/utils/validateShippingTimeGroup.js
ecgan: > The original "empty" value of shipping rate had been `null` and has been changed to `undefined`, and the "empty" value of minimum order is `undefined`. But the "empty" value of shipping time is `null`. The validation function is also based on these different specific verifications, but in fact other invalid values will not be checked. These inconsistencies don't cause bugs but are somewhat a bit confusing to read.

Yeah, I too realized the inconsistencies when I was looking into this issue.

Speaking of this `null` vs `undefined`, do you have a preference / opinion on this? From what I understand, there are two major camps of people: one advocates the use of `null`, and another advocates the use of `undefined`. Wondering what's your take on this. I don't have a strong opinion on this; as long as we are consistent throughout with either one, we are good. We can probably come up with a Wiki page and use an ESLint plugin to enforce this.

> The "empty" value checking could be changed to check the invalid values and types by `Number.isFinite()`. Thus, it ensures that invalid values of shipping time are detected without the need to validate against the specific empty value only.

That is a good idea! One thing: since shipping time is always an integer, I use `Number.isInteger()` instead, done in [`33affe3` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/33affe36a665eec1607a62cbab2e9344da8982d6).

js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
ecgan: > Wondering if these two names would gradually converge towards one of them?

Yeah, I think the ideal situation is that everything converges towards one of them, across shipping rates and shipping times, in both back-end PHP and front-end JS code.

For some background context, the `country_code` parameter in the shipping rate API was renamed to `country` in PR https://github.com/woocommerce/google-listings-and-ads/pull/1202. That's why on the shipping rate JS side, we have `country` / `countries`, to match with the API.

Meanwhile, in shipping time API, we still have `country_code` parameter, that's why we have `countryCode` / `countryCodes` in shipping time JS side.

My preference is to make the naming in JS closely match with the naming in API side.

If we want, we can standardize them to `countries` /  on the JS side, and map `countries` to `country_codes` when we are making API calls. IMO, this is less than ideal, but doable. Let me know your thoughts.

> This PR adjusted some of them but there are still some inconsistent naming within individual components. The current implementation makes the switching between `countries` and `countryCodes` within and between components more complicated, resulting in lower readability.

Yeah, agree. It was quite a deep rabbit hole when I first looked into this issue, I tried to make them nice and consistent but there are too many changes, IMO way too much to fix this issue #1435. In the end, I had to "restart" and make the solution focusing on fixing issue #1435, without looking into other engineering / tech debt concern.  

> If at least these shipping modals interface, internal and its dependencies are adjusted to be consistent in this PR first, it might save time for future code reviews.

I looked into that, and it causes a lot of code changes. 

The idea that I have here is that I would like this PR to focus on fixing issue #1435, so reviewers could (relatively easily) verify that the fix here does fix the issue, without introducing other issues.

Next, we can have a separate PR to address the engineering / tech debt concern, so reviewers could focus on the engineering changes, and no changes to functionalities.

So, in a way, I'm thinking about "save time" from a different perspective. 😄 

I'm thinking I can address these engineering concerns when I work on refactoring the shipping time section. 📜

eason9487: 📜 

> My preference is to make the naming in JS closely match with the naming in API side.

Considering that JS and PHP have different naming in their coding styles, the ideal naming consistency is unlikely to be achieved. Even if this ideal is achievable, I think the strategy of making the naming in JS closely match the API side is a secondary consideration. Having an isolation layer allows for renaming without dripping into the wider code, and allows for flexibility on JS side of the interface design.

> If we want, we can standardize them to `countries` / on the JS side, and map `countries` to `country_codes` when we are making API calls. IMO, this is less than ideal, but doable. Let me know your thoughts.

Yes, this's the way to create an isolation layer.

The isolation layer could be explicitly established between the **wp-data** and the application code, so that if naming consistency or naming mapping is needed, it could be adapted before the fetched data is passed to the **reducer**. Then as you said, when using **actions** to communicate the API, internally map the parameters of the action function to API spec.

---

💅 

> It was quite a deep rabbit hole when I first looked into this issue, I tried to make them nice and consistent but there are too many changes, IMO way too much to fix this issue #1435. In the end, I had to "restart" and make the solution focusing on fixing issue #1435, without looking into other engineering / tech debt concern.

I was thinking it would be good to address some technical debts together within certain scope when fixing the problem or adding new features. Because sometimes if we want to continue modifying the code without dealing with the technical debt, we may therefore be limited by it, unable to use better practices, or even worse, add more technical debt.

After reading through the larger code scope, I understand that it's difficult to make the consistent `country` naming in components within an explicit scope, and that eventually, some components will still have to bear those naming mapping first.

> > If at least these shipping modals interface, internal and its dependencies are adjusted to be consistent in this PR first, it might save time for future code reviews.
> 
> I looked into that, and it causes a lot of code changes.
> 
> The idea that I have here is that I would like this PR to focus on fixing issue #1435, so reviewers could (relatively easily) verify that the fix here does fix the issue, without introducing other issues.
> 
> I'm thinking I can address these engineering concerns when I work on refactoring the shipping time section.

As a code reviewer, I had to check the mapping of these names back and forth in several similar implementations to make sure they are all not missed. And originally I was not sure which of the two variable names would be converged toward, so I was worried that when adjusting the shipping time section, the naming of `countries` or `countryCode` would be changed again where this PR has been changed, so there would be extra review time spent.

If the naming in shipping time would also be changed to `countries`, then maybe the countries -> countryCode mapping direction of this PR could be changed to countryCode -> countries, then it should still save some scope and time for future adjustment.

js/src/utils/validateShippingTimeGroup.js
eason9487: > Speaking of this `null` vs `undefined`, do you have a preference / opinion on this?

Due to the characteristics of the JS language, I don't have an opinion on how they should be genernally interpreted and used, but usually case-by-case.

In this case, I think both are fine. If we want to consider it in more detail, then I would think:
- It's known that we expect its valid value to be a number type.
- The typecasting might occur during data processing, even if these have been well processed in `AppInputNumberControl`.

So one factor to determine would be which is easier to handle and validate, `Number(null)` getting `0` or `Number(undefined)` getting `NaN`.

> That is a good idea! One thing: since shipping time is always an integer, I use `Number.isInteger()` instead, done in [`33affe3` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/33affe36a665eec1607a62cbab2e9344da8982d6).

Using `Number.isInteger()` in this case looks great! Thanks!

js/src/utils/validateShippingTimeGroup.test.js
eason9487: There is no explicit language specification or semantic convention for "empty" for each JS data type, so it's easy to infer that when seeing a valid value is the primitive number type, the given "empty" value may be `null` or `undefined`, sometimes `0` or `-1`, or even `''` or `NaN`.

Testing helps us to identify problems (early) and to have more confidence when making changes. Surely, I don't think it's necessary to enumerate all possible values or types, but if the test case is only for the actual implementation value, `null`, and not for the purpose of meeting the test description, then this test case provides less confidence than it seems.

Specifically, perhaps at least testing with `null` and `undefined` would make it more robust. And maybe even add [the previously used `''`](https://github.com/woocommerce/google-listings-and-ads/blob/1530a66c3db60239d8d83fbb06f627f8915ef3a4/js/src/setup-mc/setup-stepper/setup-free-listings/shipping-rate/shipping-rate-setup/index.js#L52) if we assumed the purpose of data validation should not Implicitly rely on the typecasting in `AppInputNumberControl`.

eason9487 (approved):
Thanks for the adjustments! Testing well. The original❓ concerns have been clarified and will be addressed in future PRs, so there are no blocking concerns now.

js/src/components/free-listings/configure-product-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js
ecgan: > The isolation layer could be explicitly established between the **wp-data** and the application code [...]

Yup. We have the same thinking here. 🙂 👍 

> I was thinking it would be good to address some technical debts together within certain scope when fixing the problem or adding new features. [...]

Yup, agree. In this PR, I already see myself as doing some form of "address some technical debts together within certain scope", by refactoring the shipping time validation into a function shared by the different shipping time UI pieces, and adding tests for the validation function. 😄 

> As a code reviewer, I had to _check the mapping of these names back and forth_ in several similar implementations to make sure they are all not missed. And originally I was not sure _which of the two variable names would be converged toward_, so I was worried that when adjusting the shipping time section, the naming of `countries` or `countryCode` would be changed again where this PR has been changed, so there would be extra review time spent.
> 
> _If the naming in shipping time would also be changed to `countries`_, then _maybe the countries -> countryCode mapping direction of this PR could be changed to countryCode -> countries_, then it should still save some scope and time for future adjustment.

Hey, thanks a lot for this comment here! 😃 👍  Combined with my earlier comment https://github.com/woocommerce/google-listings-and-ads/pull/1473#discussion_r865028069 about "we can standardize them to `countries` on the JS side, and map `countries` to `country_codes` when we are making API calls", I think I realize what you are saying. When I take another look into `validateShippingRateGroup` and `validateShippingTimeGroup`, one is using `countries` and another is using `countryCodes`, which can be a bit confusing. In the future when we work on refactoring shipping time section, I'm leaning towards making them into `countries`, and it make sense for us to change to using `countries` in both `validateShippingRateGroup` and `validateShippingTimeGroup` in this PR. I have made the change in [`b505b33` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/b505b33080e31512fa58e428a6a9c2e0fadf2d0d).

js/src/utils/validateShippingTimeGroup.test.js
ecgan: > There is no explicit language specification or semantic convention for "empty" for each JS data type, so it's easy to infer that when seeing a valid value is the primitive number type, the given "empty" value may be `null` or `undefined`, sometimes `0` or `-1`, or even `''` or `NaN`.

Okay, I think the problem lies in my use of the word "empty" in the test description then. If I change the test description to "... when values.time is null", then we wouldn't have such problem. I have changed it in [`b6cb7c4` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/b6cb7c42f71b88e8d3768d1d7803f57353bea661).

> Specifically, perhaps at least testing with `null` and `undefined` would make it more robust. And _maybe even add [the previously used `''`](https://github.com/woocommerce/google-listings-and-ads/blob/1530a66c3db60239d8d83fbb06f627f8915ef3a4/js/src/setup-mc/setup-stepper/setup-free-listings/shipping-rate/shipping-rate-setup/index.js#L52) if we assumed the purpose of data validation should not Implicitly rely on the typecasting in `AppInputNumberControl`._

Thanks, I think that is fair, especially if the `AppInputNumberControl` is accidentally changed to a normal text input control. I have added tests for this in [`454dcf8` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/454dcf8f9c44f3768fd84aab0a6ceb74b027810f).

js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/validateShippingRateGroup.test.js
ecgan: I have made similar changes in [`1f5e14e` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/1f5e14e990f5d05a915f7be0ed6678d6bd42ec90) and [`a337a1b` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/a337a1b0a5fec3882833a9fc6525b47c6d27207c).

ecgan:
@eason9487 , thanks for the approval. I have made a few more commits based on your comments, could you take a quick round of review please? Specifically, this commit [`b505b33` (#1473)](https://github.com/woocommerce/google-listings-and-ads/pull/1473/commits/b505b33080e31512fa58e428a6a9c2e0fadf2d0d) (mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1473#discussion_r866067306) is the one worth taking another look, I changed `countryCodes` to `countries`.

js/src/utils/validateShippingTimeGroup.js
ecgan: I think the following is getting out of scope of this PR, but I would like to learn more about your thoughts. 😄 

> So one factor to determine would be which is easier to handle and validate, `Number(null)` getting `0` or `Number(undefined)` getting `NaN`.

That is interesting, I didn't think about wrapping the value with `Number()` call, I was thinking to compare value directly. Example:

```js
// Assuming values.time is null.
// Approach 1: compare directly. 
if (values.time === null) { /* set error here. */}

// Approach 2: wrap with Number().
if (Number(values.time) === 0) { /* set error here. */}
```

Is my above interpretation correct? What is the advantage of wrapping with `Number()` here? 🤔 

Or, do you mean that we use `Number()` when assigning? i.e.: 

- `values.time = Number(null)` getting `0`, which means we won't be able to differentiate whether users did not enter anything or users enter value `0`?
-  `values.time = Number(undefined)` getting `NaN`, which means possible values for `values.time` would be integer and `NaN`.

eason9487: > Is my above interpretation correct? What is the advantage of wrapping with `Number()` here?

> Or, do you mean that we use Number() when assigning? [...]

Ahh, I was mainly starting with the initial empty value and thinking about which would be relatively easy to use for subsequent \<input\> processing and data validation, not that add actively typecast somewhere. When values are typecasted in these processes, how many edge cases are to be aware of. Where either the explicit `Number(emptyValue)` that we actively code, or the implicit automatic typecasting of JS, is the focus of attention.

For example:
- When assigning `<input>.value`, `<input>.value = undefined` and `<input>.value = null` will have different results, but of course the better way has been implemented - we will first convert it to an empty string properly and then assign it to the input. So, the focus is on which one will be much easier/more suitable for conversion.
- When validating data, if we want to verify the value range, then the focus is on what possible types will be passed into the validation function when it may be an empty value or a value entered by the user (and may be converted by some input handlers). Assume we were using `if ( ! ( value >= 0 ) )` to check the range here, then the edge case of implicit auto-typecasting happened. The difference of `null >= 0` getting `true` and `undefined >= 0` getting `false` might lead to bug.

eason9487 (approved):
Thanks for the adjustments! The new change looks good to me!

PR #1476, opened 2022-05-04
Move `get_supported_countries_from_continent` method to GoogleHelper

### LINKED ISSUE #1351: Add a helper function to get supported countries of a continent in GoogleHelper
## User story

In this [PR](https://github.com/woocommerce/google-listings-and-ads/pull/1341/files#diff-e5e54e70e3677013fec896bd6a8490847bc64b414e5f573c5945804d85733b95R120) and this [PR](https://github.com/woocommerce/google-listings-and-ads/pull/1344/files#diff-e00da6abb755a70b224c15358214801e8057f047e1ba00ba21f3ba056ce8c480R128-R145), they have similar logic to get only the supported countries for a continent. It would be great to extract the logic as a shared function in GoogleHelper class.

No.

A a shared helper function in GoogleHelper class to get only the supported countries for a continent, and apply in the code in the above PR.

## Acceptance criteria

- The helper function should be working and shared with both callers.
- Unit tests should be complete and cover the case when removing a continent when there are no supported countries.

nima-karimi:

Closes #1351.

This PR moves the `get_supported_countries_from_continent` method, which was previously implemented in `ShippingZone` class to get the list of countries from a continent, to the `GoogleHelper` class. This allows it to be reused by other classes.

1. Nothing much to test except running the related unit tests and confirming that they pass

### Changelog entry

> Tweak - Add helper class to obtain supported countries of a continent

### DIFF
--- a/src/Google/GoogleHelper.php
+++ b/src/Google/GoogleHelper.php
@@ -1339,7 +1339,7 @@ class GoogleHelper implements Service {
 	 */
 	public function get_supported_countries_from_continent( string $continent_code ): array {
 		$countries  = [];
-		$continents = $this->wc->get_wc_countries()->get_continents();
+		$continents = $this->wc->get_continents();
 		if ( isset( $continents[ $continent_code ] ) ) {
 			$countries = $continents[ $continent_code ]['countries'];
 
--- a/src/Google/GoogleHelper.php
+++ b/src/Google/GoogleHelper.php
@@ -1188,7 +1188,7 @@ class GoogleHelper implements Service {
 	 * WooCommerce Countries -> https://github.com/woocommerce/woocommerce/blob/master/i18n/countries.php
 	 * Google Supported Countries -> https://support.google.com/merchants/answer/160637?hl=en
 	 *
-	 * @return string[]
+	 * @return string[] Array of country codes.
 	 */
 	public function get_mc_supported_countries(): array {
 		return array_keys( $this->get_mc_supported_countries_data() );
@@ -1325,6 +1325,32 @@ class GoogleHelper implements Service {
 	 */
 	public function find_subdivision_id_by_code( string $code, string $country_code ): ?int {
 		return self::COUNTRY_SUBDIVISIONS[ $country_code ][ $code ]['id'] ?? null;
+	}
+
+	/**
+	 * Gets the list of supported Merchant Center countries from a continent.
+	 *
+	 * @param string $continent_code
+	 *
+	 * @return string[] Returns an array of country codes with each country code used both as the key and value.
+	 *                  For example: [ 'US' => 'US', 'DE' => 'DE' ].
+	 *
+	 * @since x.x.x
+	 */
+	public function get_supported_countries_from_continent( string $continent_code ): array {
+		$countries  = [];
+		$continents = $this->wc->get_wc_countries()->get_continents();
+		if ( isset( $continents[ $continent_code ] ) ) {
+			$countries = $continents[ $continent_code ]['countries'];
+
+			// Match the list of countries with the list of Merchant Center supported countries.
+			$countries = array_intersect( $countries, $this->get_mc_supported_countries() );
+
+			// Use the country code as array keys.
+			$countries = array_combine( $countries, $countries );
+		}
+
+		return $countries;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


tests/Unit/Google/GoogleHelperTest.php
```diff
@@ -91,6 +92,43 @@ public function test_find_location_id_by_subdivision_code() {
...
+		$wc_countries->expects( $this->any() )
+					 ->method( 'get_continents' )
+					 ->willReturn( [
+						 'EU' => [
+							 'name'      => 'Europe',
+							 'countries' => [
+								 // A random country code, not supported by Merchant Center. This should be ignored.
+								 'OO1',
+								 // Another random country code, not supported by Merchant Center. This should be ignored.
+								 'OO2',
```
mikkamp: These countries codes will get excluded because they aren't countries in the EU. But we aren't testing the array_intersect line (where we filter out unsupported countries), maybe we should add a test for a different continent with the country code `KP`.

My initial thought was that GB is no longer in the EU 😆️, but we'll ignore that since we are looking at the continent.

src/Google/GoogleHelper.php
```diff
@@ -1327,6 +1327,32 @@ public function find_subdivision_id_by_code( string $code, string $country_code
...
+	 * @param string $continent_code
+	 *
+	 * @return string[] Returns an array of country codes with each country code used both as the key and value.
+	 *                  For example: [ 'US' => 'US', 'DE' => 'DE' ].
+	 *
+	 * @since x.x.x
+	 */
+	public function get_supported_countries_from_continent( string $continent_code ): array {
+		$countries  = [];
+		$continents = $this->wc->get_wc_countries()->get_continents();
```
ianlin: 💅 
There is a method `get_continents` in `src/Proxies/WC.php`, I think we can just use that so we don't need to call `get_continents` method from the instance of `WC_Countries`.

https://github.com/woocommerce/google-listings-and-ads/blob/efe2f25e69c5fc740e433212bfca3880bf5286ef/src/Proxies/WC.php#L75

ianlin (approved):
Thanks for adding this! I've tested it and it worked. Just left a non-blocking comment below.

I'll create another that will be using this helper function as well.

mikkamp (commented):
Thanks for the changes, it looks like a pretty straightforward change, I just left one comment about the test.

There is a [todo comment](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.7/src/API/Site/Controllers/MerchantCenter/SupportedCountriesController.php#L128-L130) in the function `SupportedCountriesController::get_supported_continents`. Are we going to refactor that function to use this helper method, or is that going to be done in a followup PR?

ianlin:
@mikkamp 
> There is a [todo comment](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.7/src/API/Site/Controllers/MerchantCenter/SupportedCountriesController.php#L128-L130) in the function SupportedCountriesController::get_supported_continents. Are we going to refactor that function to use this helper method, or is that going to be done in a followup PR?

Thanks for the heads up, I'm working on it and will create a PR tomorrow. 😄

tests/Unit/Google/GoogleHelperTest.php
nima-karimi: > These countries codes will get excluded because they aren't countries in the EU.

Hmm. Not sure I follow.  We're mocking the `get_continents` method and returning these country codes as part of the EU continent. So as far as the `GoogleHelper` class is concerned, these countries are in fact part of the EU 😅 This is the only place that passes the list of countries in a continent to the `GoogleHelper` class. The `array_intersect` part is being tested here because even though the `001` and `002` are allegedly part of EU, they are not supported by MC.

I didn't want to use actual country codes because other EU countries might later be added to the list of supported MC countries and that would make our tests dependent on arbitrary logic unrelated to the method functionality.

nima-karimi:
Thanks, @ianlin, and @mikkamp 🙏 I've made some changes to the code and left some comments. Ready for another review.

tests/Unit/Google/GoogleHelperTest.php
mikkamp: You are right, I was looking at that wrong. Seems I was assuming the list of WC countries wasn't being mocked and it was using the real list of countries. With the mocked list it is working as intended.

mikkamp (approved):
I don't see an additional commit if there were more code changes. I retract my comment on the test. So everything looks good to me.

nima-karimi:
> I don't see an additional commit if there were more code changes. 

Forgot to push the changes 😬

PR #1480, opened 2022-05-05
Refactoring and extending update all products test suit

### LINKED ISSUE #1457: Add more UpdateAllProduct tests.
Some of the tests we might want to add:

- Filter out some products from the batches and confirm that batches continue even if some batch sizes are smaller or empty, see #861
- Test a final batch smaller than batch size (this will identify that `handle_complete` is not always called, which isn't an issue because we never use it)

_Originally posted by @mikkamp in https://github.com/woocommerce/google-listings-and-ads/pull/1451#pullrequestreview-955934017_

message-dimke:

Refactoring and extending update all products tests suit by adding tests to check multiple half filled or empty batches does continue batch scheduling and items processing as expected.

Closes #1457.

### Changelog entry

> Add - Extending Update All Products Test Suit

### DIFF
--- a/tests/Unit/Jobs/UpdateAllProductsTest.php
+++ b/tests/Unit/Jobs/UpdateAllProductsTest.php
@@ -77,7 +77,10 @@ class UpdateAllProductsTest extends UnitTest {
 					return 2;
 				}
 				return $batch_count;
-			}, 10, 2 );
+			},
+			10,
+			2
+		);
 
 		$this->job->init();
 	}
@@ -92,19 +95,18 @@ class UpdateAllProductsTest extends UnitTest {
 	 */
 	public function test_single_empty_batch() {
 		$filtered_product_list = new FilteredProductList( [], 0 );
-		$this->action_scheduler->expects( $this->once() )
-			->method( 'schedule_immediate' )
-			->with( self::CREATE_BATCH_HOOK, [ 1 ] );
 		$this->product_repository->expects( $this->once() )
 			->method( 'find_sync_ready_products' )
 			->willReturn( $filtered_product_list );
-
 		/*
 		 * We expect only single call to `has_scheduled_action` when scheduling
 		 * the batch and no calls further since batch is empty.
 		 */
 		$this->action_scheduler->expects( $this->once() )
 			->method( 'has_scheduled_action' );
+		$this->action_scheduler->expects( $this->once() )
+			->method( 'schedule_immediate' )
+			->with( self::CREATE_BATCH_HOOK, [ 1 ] );
 
 		$this->job->schedule();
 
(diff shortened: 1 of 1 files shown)


message-dimke:
Also I found that when cherrypicking hotfix I did not pick my last updates to the code and styling so those appear into this PR as well.

tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -64,68 +61,216 @@ public function setUp(): void {
...
+			->willReturn( $filtered_product_list );
 
-		$this->filtered_product_list->expects( $this->never() )
-			->method( 'get' );
+		/*
+		 * We expect only single call to `has_scheduled_action` when scheduling
+		 * the batch and no calls further since batch is empty.
+		 */
+		$this->action_scheduler->expects( $this->once() )
+			->method( 'has_scheduled_action' );
```
mikkamp, line 106: I know the order isn't important, but when I was trying to mentally follow what code is executed and what we are expecting, it seemed reversed.

When we first call `job->schedule` it will run:
- `has_scheduled_action`
- `schedule_immediate`

When we call the `do_action` for the batch hook then it will run:
- `find_sync_ready_products`
- `has_scheduled_action`
- `schedule_immediate`

So it would seem logical to put the expectation of `has_scheduled_action` before `schedule_immediate`.
(same comment for the test `test_single_partial_batch`)

tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -64,68 +61,216 @@ public function setUp(): void {
...
 		$this->action_scheduler->expects( $this->exactly( 2 ) )
+			->method( 'has_scheduled_action' );
+
+		$this->job->schedule();
+
+		do_action( self::CREATE_BATCH_HOOK, 1 );
+	}
+
+	/**
+	 * We test two filly loaded batches to make sure batch and process item jobs are correctly scheduled
```
mikkamp: Just a small typo `fully loaded`. Also when we refer to batch and process items, these are separate actions within the full job.
```suggestion
	 * We test two fully loaded batches to make sure batch and process item actions are correctly scheduled
```

tests/Unit/Jobs/UpdateAllProductsTest.php
```diff
@@ -64,68 +61,216 @@ public function setUp(): void {
...
+
+		/* adding a filter to make batch smaller for testing */
+		add_filter(
+			'woocommerce_gla_batched_job_size',
+			function ( $batch_count, $job_name ) {
+				if ( self::JOB_NAME === $job_name ) {
+					return 2;
+				}
+				return $batch_count;
+			}, 10, 2 );
```
mikkamp: If we split the parameters to multiple lines then the PHPCS rules we use require us to put each parameter on a separate line.
```suggestion
			},
			10,
			2
		);
```

mikkamp (approved):
Thanks for adding these additional tests. The tests look good and I'm getting the results as expected.

I just added a few small minor comments, but none of them are blockers, so I'll approve the PR.

tests/Unit/Jobs/UpdateAllProductsTest.php
message-dimke: I changed the order of calls to have expectations aligned to how the corresponding methods are actually called.

tests/Unit/Jobs/UpdateAllProductsTest.php
message-dimke: Fixed the typo and changed jobs to actions. Thanks!

PR #1481, opened 2022-05-05
Fix the comparison of target audience to avoid the unsaved prompt showing when the country sets are the same

### LINKED ISSUE #1431: Edit Free Listings: Unsaved data navigate away prompt might show up when target audience selected countries are still the same after being modified
### Describe the bug:

This issue is found in https://github.com/woocommerce/google-listings-and-ads/pull/1409#discussion_r849359401.

In Edit Free Listings Step 1, if we remove a country and add the removed one back, which means the selected countries are the same as the saved ones after these operations, the unsaved data prompt will be shown.

The problem is that even though in the UI the form fields look the same and there are "no changes" to the users, the underlying data `targetAudience.countries` and `savedTargetAudience.countries` have the same country codes but in different orders, and hence does not satisfy the `isEqual` check on target audience data, causing the navigate away prompt to show up.

https://user-images.githubusercontent.com/417342/163433622-3900e98f-a045-4332-90ab-b9a60e2286e0.mp4

### Steps to reproduce:

Pre-condition: you should have free listings setup with "selected countries only" checked. 

1. Go to Edit Free Listings Step 1.
2. Remove a country from the select dropdown.
3. Add back the removed country in the dropdown.
4. Try to reload the page or click on the top left "go back" button.

### Expected behavior:

The navigate away prompt should not show up.

### Actual behavior:

The navigate away prompt might show up.

eason9487:

Closes #1431 

- When checking if there are unsaved target audiences, the `countries` array values are compared by whether the value sets are the same.

### Screenshots:

https://user-images.githubusercontent.com/17420811/166906096-4bf8f897-d4fa-4e87-b9d0-0637efe0c848.mp4

1. Go to step 1 of the editing free listings page.
   - If there is only one target audience country, it will need to add another one here and save data in step 2, and then go back to step 1.
1. To make the different order of values at the subsequence step, delete the first country, e.g. Japan.
1. Try to reload the page or click on the top left "go back" button. It should pop up a prompt that there are unsaved data.
1. Add back the removed country, e.g. Japan.
1. Try to reload the page or click on the top left "go back" button. It should not pop up the prompt.

### Changelog entry

> Fix - The unsaved prompt might pop up when the countries of the target audience are the same when navigating away from the free listings edit page.

### DIFF
--- a/js/src/edit-free-campaign/index.js
+++ b/js/src/edit-free-campaign/index.js
@@ -121,7 +121,13 @@ export default function EditFreeCampaign() {
 	const { createNotice } = useDispatchCoreNotices();
 
 	// Check what've changed to show prompt, and send requests only to save changed things.
-	const didAudienceChanged = ! isEqual( targetAudience, savedTargetAudience );
+	const didAudienceChanged = ! isEqual(
+		...[ targetAudience, savedTargetAudience ].map( ( el ) => ( {
+			...el,
+			countries: new Set( el?.countries ),
+		} ) )
+	);
+
 	const didSettingsChanged = ! isEqual( settings, savedSettings );
 	const didRatesChanged = hasUnsavedShippingRates(
 		shippingRates,


jorgemd24 (approved):
Thanks @eason9487, LGTM 👍 

✅  The order of the countries is not relevant anymore to detect if the audience has changed or not.

PR #1485, opened 2022-05-06
Use a shared helper method to get supported countries of a continent for /mc/countries API

ianlin:

Related to #1351 and a follow-up of #1476, this PR updates the codes in `/mc/countries` API to use a shared helper method in GoogleHelper in order to get supported countries of a continent.

1. Request to the API `/mc/countries`, the response should remain unchanged before this PR.
2. Run the unit tests, the test results should be passed.

### Changelog entry

> Update - Use a shared helper method to get supported countries of a continent for /mc/countries API


mikkamp (approved):
Thanks for the changes, it works good and is returning the results as expected.

Just noticed that `SupportedCountriesControllerTest` has a few locations where `=>[` is used without a separating space. Not from this PR, but since we are changing things in that test we might as well fix it.

PR #1487, opened 2022-05-06
GMC Request Review - Delete transients when disconnecting account

puntope:

Closes part of #1165 

After implementing the cache in [this PR](https://github.com/woocommerce/google-listings-and-ads/pull/1464) I noticed that when we disconnect Google Account the transient is still there. That means if the user connects with different account it will get the cached result of the previous account.

This PR removes the transients and flushes the cache to avoid that. 

1. Checkout this PR
2. Connect a google account and go to Product Feed (that page will run Request review Status call) you can also do a call to `mc/review` 
3. Go to DB and see that `_transient_gla_mc_account_review` as well as `_transient_timeout_gla_mc_account_review` are present in `wp_options` table
4. Disconnect your Google Account accounts
5. Go to DB and see that `_transient_gla_mc_account_review` as well as `_transient_timeout_gla_mc_account_review` are not there anymore.


jorgemd24 (approved):
Thanks @puntope, LGTM 👍

PR #1488, opened 2022-05-06
GTag tracking for add_to_cart events

mikkamp:

When we create pages using WooCommerce Gutenberg blocks we are unable to track the add_to_cart event since the page content is rendered dynamically. This PR enqueues some JavaScript assets which will hook into the appropriate events so it can track when products are added to the cart.

Since there are many add_to_cart buttons which can show up on archive/single product pages, including in the featured/related products sections, this PR will hook into all buttons on the page to do the tracking. This is less intrusive then injecting a code snippet for each add_to_cart button.

Note: For the Gutenberg blocks we are currently using an action name which is marked as [experimental](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/blocks/feature-flags-and-experimental-interfaces.md#usages-of-experimental-prefix). I've started a discussion to get some feedback on this: p1651828112139989-slack-C02UBB1EPEF

Note: Automated tests haven't been added yet. Since it's being merged into a feature branch I'm going to add tests in a follow up PR.

#### Test Gutenberg blocks
1. Use a site that is fully onboarded including connected to an Ads account.
2. Create a page using the All Products block 
3. View the page on the frontend and add a product to the cart (should be a simple product)
4. Check the network tab in the browser dev tools and confirm we are seeing the events getting sent to `https://googleads.g.doubleclick.net/pagead...`
5. On the products page jump to the next page (must have more products then fit on one page)
6. Click the add cart for one of the products dynamically loaded for another page
7. Confirm that the network tab show the event was tracked

#### Test regular add_to_cart buttons

1. Use a site that is fully onboarded including connected to an Ads account.
2. Go to a product archive page and test the add_to_cart buttons
3. Check the network tab in the browser dev tools and confirm we are seeing the events getting sent to `https://googleads.g.doubleclick.net/pagead...`
4. Confirm the same event works from a single product page
5. Confirm the same event works from related products on a single product page
6. Go to WooCommerce > Settings > Products and set the redirect to cart page option
7. Confirm events still get sent when we add_to_cart (might need to enable "preserve log" in console and network tab)
8. Go to WooCommerce > Settings > Products and disable the redirect to cart page option and disable "AJAX add to cart buttons on archives"
9. Confirm events still get sent when we add_to_cart

### Additional details:

The following points need to be finalised before this PR is ready:
- [x] Confirm usage of the experimental hook
- [x] Align parameters for the `add_to_cart` event to be the same as implemented in #1329 
- [x] Confirm whether we need to add the events `remove_from_cart` and `change_cart_quantity`
- [x] Check the method used for loading script assets on the frontend, and confirm this is the implementation we want to go with
- [x] The script gets loaded depending on `is_page` which means it will be loaded on each page, check if there is any way to know whether WC blocks have been used on the page
- [x] Confirm if we want to also include code for tracking regular cart events (not through blocks) to this script so we don't need to inject as much code on the page
- [ ] Add testing for any of the added code
- [ ] Add an e2e test for confirming the event is fired from a block (this will cover discovering any changes to the experimental hook)

### Changelog entry
* Add - Track add to cart events from all buttons including Gutenberg blocks.


mikkamp:
Discussion summary: Using the "experimental" event can be ok, there will be a deprecation procedure when this is changed so we can catch this during the compatibility testing. Looking into an e2e test which will confirm the event is fired, to facilitate in catching any changes.

mikkamp:
> The script gets loaded depending on is_page which means it will be loaded on each page, check if there is any way to know whether WC blocks have been used on the page

There is the function `has_block` which can be used to check if the "All Products Block" is being used within a specific page. However if the block is not used within the main page content or within a reusable block then this function does not return true. For now we are limited to only tracking `add_to_cart` events from the "All Products Block" but eventually we are going to need to track events from other blocks as well. So I haven't found a better way then continuing to use `is_page` as a conditional to load the tracking JS.

mikkamp:
> Confirm whether we need to add the events `remove_from_cart` and `change_cart_quantity`

Answered [here](https://github.com/woocommerce/google-listings-and-ads/pull/1329#issuecomment-1124390797)

> For current design, we only track "add to cart" action in the product page. We also inject gtag in purchase page which will get accurate item information in cart from there.

js/src/gtag-events/constants.js
```diff
@@ -0,0 +1,3 @@
+export const namespace = 'google-listings-and-ads';
+export const actionPrefix = 'experimental__woocommerce_blocks';
+export const sendToGroup = 'GLA';
```
puntope: 💅 

To consider to use the UPPERCASE_SNAKE_CASE syntax for the constant names as we do in this `constants.js` [here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/constants.js) for example

js/src/gtag-events/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+// Register for jQuery event to update product data.
+if ( typeof jQuery === 'function' ) {
+	jQuery( document ).on(
+		'found_variation',
+		'form.cart',
+		function ( event, variation ) {
+			retrievedVariation( variation );
+		}
+	);
```
puntope, line 93: ❓ 
I take as granted that `found_variation` is triggered by `jQuery.trigger()`. Otherwise I'd suggest to use native `addEventListener` instead

js/src/gtag-events/index.js
```diff
@@ -0,0 +1,83 @@
...
+	const variationId = cartForm.querySelector( '[name=variation_id]' );
+	const quantity = cartForm.querySelector( '[name=quantity]' );
+
+	const product = getProductObject( {
+		id: parseInt( variationId ? variationId.value : addToCart.value, 10 ),
+	} );
+
+	trackAddToCartEvent(
+		product,
+		quantity ? parseInt( quantity.value, 10 ) : 1
```
puntope, line 64: ❓ So quantity can be a string? If so, shouldn't be `trackAddToCartEvent( product, data.quantity || 1 );` in line 30 handling this as well? Same goes for `variationId` case 

💅 

```
 trackAddToCartEvent(
		product,
		parseInt( quantity?.value || 1, 10) 
)
```

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @param {string} eventName
+ * @param {Object} eventParams
+ */
+export const trackEvent = ( eventName, eventParams ) => {
+	if ( typeof gtag !== 'function' ) {
+		throw new Error( 'Function gtag not implemented.' );
+	}
+	// eslint-disable-next-line no-console
+	console.log( `Tracking event ${ eventName }` );
```
puntope: ❓ 
Do we want to show a log with the event? As far as I know there are tools to inspect this without logs

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @param {Object} product
+ * @param {number} quantity
+ * @return {Object} Item object.
+ */
+export const getItemObject = ( product, quantity ) => {
+	const item = {
+		id: 'gla_' + product.id,
+		quantity,
+		google_business_vertical: 'retail',
```
puntope, line 51: ❓  

What is this? `google_business_vertical`

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+	};
+
+	if ( product.name ) {
+		item.name = product.name;
+	}
+
+	if ( 'categories' in product && product.categories.length ) {
+		item.category = product.categories[ 0 ].name;
+	}
```
puntope, line 60: 💅 

```
if ( product?.categories ) {
		item.category = product.categories[ 0 ]?.name;
}
```

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+export const getItemObject = ( product, quantity ) => {
+	const item = {
+		id: 'gla_' + product.id,
+		quantity,
+		google_business_vertical: 'retail',
+	};
+
+	if ( product.name ) {
+		item.name = product.name;
+	}
```
puntope, line 56: 💅 / ❓ 

As far as I see in Gtag. Name is mandatory.

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+
+	if ( 'categories' in product && product.categories.length ) {
```
puntope: 💅  

`if (product?.categories?.length ) {`

also works

❓  Wondering if it's possible to have category without name. Since in that case `item.category` would be undefined.

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+		item.category = product.categories[ 0 ].name;
+	}
+
+	if ( product.prices && product.prices.price ) {
```
puntope: 💅 

`if ( product?.prices?.price ) {`

puntope: ❓ `currency_minor_unit` is not checked so, in case it's undefined, then `item.price` would be NaN

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+};
+
+/**
+ * Formats data into a cart Item object.
+ *
+ * @param {Object} product
+ * @param {number} quantity
+ * @return {Object} Item object.
+ */
+export const getItemObject = ( product, quantity ) => {
```
puntope: 💅 
WDYT of `getCartItemObject` for the name?

js/src/gtag-events/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+addAction(
+	`${ actionPrefix }-cart-add-item`,
+	namespace,
+	( { product, quantity = 1 } ) => {
+		trackAddToCartEvent( product, quantity );
+	}
+);
+
+const addToCartClick = function ( event ) {
```
puntope, line 32: 💅  Docs

js/src/gtag-events/index.js
```diff
@@ -0,0 +1,83 @@
...
+	const data = event.target.dataset;
+	const product = getProductObject( { id: data.product_id } );
+
+	trackAddToCartEvent( product, data.quantity || 1 );
+};
+
+const singleAddToCartClick = function ( event ) {
```
puntope, line 44: 💅  Docs

js/src/gtag-events/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+/* global jQuery */
+
+addAction(
+	`${ actionPrefix }-cart-add-item`,
+	namespace,
+	( { product, quantity = 1 } ) => {
+		trackAddToCartEvent( product, quantity );
+	}
+);
```
puntope: ❓ 
I guess this is what Gutemberg block triggers? 

Would be cool to add just a line explaining so.

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+
+/* global glaGtagData */
+
+/**
+ * Track an event using the global gtag function.
+ *
+ * @param {string} eventName
+ * @param {Object} eventParams
+ */
+export const trackEvent = ( eventName, eventParams ) => {
```
puntope, line 15: 💅  

Missing `@throws` https://jsdoc.app/tags-throws.html

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+		items: [ getItemObject( product, quantity ) ],
+	} );
+};
+
+/**
+ * Formats data into a cart Item object.
+ *
+ * @param {Object} product
+ * @param {number} quantity
+ * @return {Object} Item object.
```
puntope: 💅 Would be nice to add a `typedef` for Item Object instead of just {Object} documenting what kind of props it could have. 

https://jsdoc.app/tags-typedef.html

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return item;
+};
+
+/**
+ * Formats a regular price into a price object.
+ *
+ * @param {number} price
+ * @return {Object} Price object.
```
puntope: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1488/files#r883649128 comment

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+		price: price * 10 ** glaGtagData.currency_minor_unit,
+		currency_minor_unit: glaGtagData.currency_minor_unit,
+	};
+};
+
+/**
+ * Formats a product object to include name and price from global data.
+ *
+ * @param {Object} product
+ * @return {Object} Product object.
```
puntope: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1488/files#r883649128 comment

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+ */
+export const getProductObject = ( product ) => {
```
puntope, line 90: ❓  So` glaGtagData.products` is always available? In case for some reason `glaGtagData.products` is undefined the APP will crash.

js/src/gtag-events/utils.js
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return product;
+};
+
+/**
+ * Updates product data with the retrieved variation.
+ *
+ * @param {Object} variation
+ */
+export const retrievedVariation = ( variation ) => {
```
puntope, line 105: 💅  Since `retrievedVariation` seems to be triggered by a Jquery event somewhere, I would make sure variation is a valid object here as well.

src/Assets/AdminScriptWithBuiltDependenciesAsset.php
```diff
@@ -3,66 +3,25 @@
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Assets;
 
-use Automattic\WooCommerce\GoogleListingsAndAds\Value\BuiltScriptDependencyArray as DependencyArray;
-use Throwable;
-
 defined( 'ABSPATH' ) || exit;
 
 /**
  * Class AdminScriptWithBuiltDependenciesAsset
```
puntope: 💅  This `Class AdminScriptWithBuiltDependenciesAsset` feels a bit redundant for me. We can see already the class name is  `AdminScriptWithBuiltDependenciesAsset` so I'll replace this with a description about what this class is for.

src/Assets/ScriptWithBuiltDependenciesAsset.php
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Assets;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Value\BuiltScriptDependencyArray as DependencyArray;
+use Throwable;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class ScriptWithBuiltDependenciesAsset
```
puntope: 💅  Same as https://github.com/woocommerce/google-listings-and-ads/pull/1488/files#r884728853 comment.

src/Assets/ScriptWithBuiltDependenciesAsset.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Assets
+ */
+class ScriptWithBuiltDependenciesAsset extends ScriptAsset {
+
+	/**
+	 * ScriptWithBuiltDependenciesAsset constructor.
```
puntope, line 19: 💅  Same as previous comments about using the name of the function/class as description. We can maybe replace that one with an explanation about what the constructor exactly does.

src/Google/GlobalSiteTag.php
```diff
@@ -86,14 +115,14 @@ public function register(): void {
 			function () use ( $ads_conversion_id ) {
 				$this->activate_global_site_tag( $ads_conversion_id );
 			},
-			999998
+			999999
```
puntope, line 118: ❓  why this +1?

src/Google/GlobalSiteTag.php
```diff
@@ -109,6 +138,73 @@ function () {
...
+		// Add product data for any add_to_cart link.
+		add_filter(
+			'woocommerce_loop_add_to_cart_link',
+			function ( $link, $product, $args ) {
+				$this->add_product_data( $product );
+				return $link;
+			},
+			10,
+			3
+		);
```
puntope, line 159: 💅 / ❓  

Since we are not really filtering or modifying the data. I wonder if some hook would be more appropriate.

src/Assets/ScriptWithBuiltDependenciesAsset.php
puntope: Also I think would be useful to document more the params ($build_dependency_path, $fallback_dependency_data, $in_footer)

src/Google/GlobalSiteTag.php
```diff
@@ -109,6 +138,73 @@ function () {
...
+
+	/**
+	 * Register and enqueue assets for gtag events in blocks.
+	 */
+	protected function register_assets() {
+		$gtag_events = new ScriptWithBuiltDependenciesAsset(
+			'gla-gtag-events',
+			'js/build/gtag-events',
+			"{$this->get_root_dir()}/js/build/gtag-events.asset.php",
+			new BuiltScriptDependencyArray(
```
puntope, line 181: 💅  Outside scope of this PR but this name is not accurate since we are loading Deps and Version here, not only deps.

src/Google/GlobalSiteTag.php
```diff
@@ -344,6 +440,20 @@ private function display_page_view_event_snippet(): void {
...
 
+	/**
+	 * Add product data to include in JS data.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param WC_Product $product
+	 */
+	protected function add_product_data( $product ) {
+		$this->products[ $product->get_id() ] = [
```
puntope, line 451: 💅  Not sure how strict we are here about using shorthands [] instead of array()

puntope:
> Test Gutenberg blocks

✅  I can confirm I see the console.log -> "Tracking event add_to_cart" when I added a Single product and when I added a single product after navigating to Next page

✅   I can confirm I see network request to `https://googleads.g.doubleclick.net/pagead/...` when I added a Single product and when I added a single product after navigating to Next page

This is how the request look:

Screenshot 2022-05-30 at 13 32 41

Screenshot 2022-05-30 at 13 31 57

puntope:
> Test regular add_to_cart buttons
 
✅ I can confirm I see network request to `https://googleads.g.doubleclick.net/pagead/...` when I added a Single product and when I added a single product after navigating to Next page, I saw also the request when I added to cart in Single product page and also in the Related products 

✅   I can confirm I see the console.log -> "Tracking event add_to_cart" when I added a Single product and when I added a single product after navigating to Next page, I saw also the log when I added to cart in Single product page and also in the Related products 

✅  Same behaviour also when I set redirect to cart page.

Screenshot 2022-05-30 at 13 38 15

❓  I saw tho this extra request to GTM a part of Google Ads which are not present in the blocks test

Screenshot 2022-05-30 at 13 38 39

❓   In the test procedure I see that steps 7 and 9 are same? So I'm not sure what to do in step 9 & 10

puntope (commented):
Howdy @mikkamp 

- I tested the behaviour and I guess is working (I need confirmation about step 9&10) 
- I did some code quality comments tho 

Thanks ❤️

js/src/gtag-events/index.js
mikkamp: Yes this is triggered by jQuery. I tried to depend on jQuery as little as possible, but until WooCommerce rewrites it's JS on product pages this the balance we need to maintain to remain compatible.

js/src/gtag-events/index.js
mikkamp: I'm expecting it to be an int in all cases. The only case where I parse it is when it comes directly from an input box. Although even in that case it's an input with `type=number` and `step=1` so it might not strictly be necessary.

js/src/gtag-events/utils.js
mikkamp: It's code copied from the WooCommerce Blocks extension when it tracks events, but we can remove it.

js/src/gtag-events/utils.js
mikkamp: This is a required parameter for each item, [see docs](https://support.google.com/google-ads/answer/7305793?hl=en#zippy=%2Cretail).
Right now technically only the parameters `id` and `google_business_vertical` are supported, but it's being worked on to include the additional ones, so we are leaving them in for now until they will actually be sent.

js/src/gtag-events/utils.js
mikkamp: As [linked above](https://support.google.com/google-ads/answer/7305793?hl=en#zippy=%2Cretail) only `id` and `google_business_vertical` are being sent. So we need to wait till the other fields are supported / documented before we can confirm which ones are mandatory.

We aren't following the regular Gtag documentation as that's different, but even there it mentions that either `id` or `name` is required.

js/src/gtag-events/utils.js
mikkamp: The categories are populated when the hook `experimental__woocommerce_blocks-cart-add-item` gets triggered. Each category there will have a name. For older add_to_cart buttons we don't track the item categories (since it's optional and it requires traversing and storing additional data per product).

js/src/gtag-events/utils.js
mikkamp: `prices` is a data structure that gets passed internally. Is it common to always check the full structure? My intention was to check if prices is set then convert it.

js/src/gtag-events/utils.js
mikkamp: Do you have some examples of how we define this for objects with optional parameters. And if the same Object is used in multiple files, then how do we share the typedef between them.

js/src/gtag-events/utils.js
mikkamp: It's defined initially as an empty array and then added with wp_localize_script, so any time this code gets loaded it will include the localized data in the global `glaGtagData`.

js/src/gtag-events/utils.js
mikkamp: What kind of validation do you suggest here? Just a simple check if `variation.variation_id` is set or should we check for each parameter we use? At what kind of level can we trust the data that is sent back and forth, or it better to always assume it can be garbled?

src/Assets/ScriptWithBuiltDependenciesAsset.php
mikkamp: This is all existing code that was already present for loading scripts on the backend, but it's now expanded so it can load scripts with dependencies on the frontend as well. But if you think it's important to fully document each parameter then I guess it can be added.

src/Google/GlobalSiteTag.php
mikkamp: If we look back at the original code it hooked twice into wp_head (before feature/gtag-events branch) with two very high priorities to add it (pretty much) at the end of the head block. Now we only hook into wp_head once, but I suppose a comment can be added explaining the high priority number.

src/Google/GlobalSiteTag.php
mikkamp: Do you have a suggestion of which action hook to use instead? Both actions and filters are hooks.

src/Google/GlobalSiteTag.php
mikkamp: In my opinion that would depend on how you look at dependencies. If you see a single dependency as a script + version, then a DependencyArray would be the right name for it.

src/Google/GlobalSiteTag.php
mikkamp: For this extension we have always encouraged it, and most of the code is written using shorthand arrays.

See: https://github.com/woocommerce/google-listings-and-ads/blob/1.13.2/phpcs.xml.dist#L31-L32

mikkamp:
> I tested the behaviour and I guess is working (I need confirmation about step 9&10)

Thanks for catching that. I intended that to be disable that option and the AJAX add to cart as well, I've corrected the instructions now.

mikkamp:
> I saw tho this extra request to GTM a part of Google Ads which are not present in the blocks test

Can you confirm what the full payload of that request is? On my side I see quite a few requests getting sent including a payload including the gtag config parameters, so not each request is tracking an actual event. We also added tracking for page_view/item_view events so you might see a few more which all get sent to the same `AW-<id>`

js/src/gtag-events/utils.js
mikkamp: This is also code copied from WooCommerce blocks. I understand that we have quite high requirements on the backend for JS / browsers. Do we have the same on the frontend, should I be worried at all that optional chaining might not work on some browsers? Have the same question about the exponentiation operator.

mikkamp:
@puntope Thanks for the review, I've gone ahead and addressed some of those comments. Although I've added several other questions to clarify. JS isn't my strong point so I'd need some additional details on how to setup those typedefs and what kind of min-requirements we should have for JS code on the frontend.

js/src/gtag-events/utils.js
puntope: > And if the same Object is used in multiple files, then how do we share the typedef between them.

With this syntax. (Example extracted from `js/src/components/shipping-rate-section/estimated-shipping-rates-card/rate-form-modals/rate-form-modal.js`)

```
/**
 * @typedef { import("@wordpress/components").Button } Button
 * @typedef { import(".~/data/actions").CountryCode } CountryCode
 * @typedef { import("../typedefs.js").ShippingRateGroup } ShippingRateGroup
 */
```

> Do you have some examples of how we define this for objects with optional parameters. 

With brackets. Check `sales` or `conversions` in the example below. (Example extracted from `js/src/data/utils.js`) 
You can also do `@property {fieldType} [fiedlName=value] Description` where `value` is the default value in case `fieldName` is not defined. Otherwise just `@property {fieldType} [fiedlName] Description`  should work

```
/**
 * Report fields fetched from report API.
 *
 * @typedef {Object} ReportFieldsSchema
 * @property {number} clicks Clicks value.
 * @property {number} impressions Impressions value.
 * @property {number} [sales] Sales value. Available for paid type.
 * @property {number} [conversions] Conversions value. Available for paid type.
 * @property {number} [spend] Spend value. Available for paid type.
 */
```

js/src/gtag-events/utils.js
puntope: > Do we have the same on the frontend, should I be worried at all that optional chaining might not work on some browsers? 

We  (wp-scripts) use [babel](https://babeljs.io/) behind the scenes when we build (compile) the code. In brief this converts code using new syntax to be compatible. Besides that, in this case .? chaining operator is [fully supported since 2020 in most browsers.  ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility)

PS. If you're curious you can check `js/build/gtag-events.js` to see how the code compiles.

puntope: >  Have the same question about the exponentiation operator.

Same answer. Actually it has[ full support since 2016](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation#browser_compatibility) (However, I did't use that one before)

js/src/gtag-events/utils.js
puntope: >  Is it common to always check the full structure? 

Depends on how conservative you're. On my side I'm very paranoid and I prefer to excess in my checks. If we are sure `currency_minor_unit` is coming correctly then is not necessary.

js/src/gtag-events/utils.js
puntope: 👌  Other thing I noticed now is that 

`product.id &&` condition in line 91. Is not necessary because in case it's undefined `glaGtagData.products[ product.id ]` will be undefined as well. 

Screenshot 2022-05-31 at 11 52 51

js/src/gtag-events/utils.js
puntope: I guess in this case just checking `variation.variation_id` to be sure variation is there an is an object would be enough to avoid a crash in case variation is not there or is not an object.  In case display_name and display_price I won't check that... since it won't make the app to crash and in the worse case scenario we will send the props as undefined. 

`if (! variation?.variation_id ) return` before line 106 

> At what kind of level can we trust the data that is sent back and forth, or it better to always assume it can be garbled?

In this case since the data is coming from a JQuery event `found_variation` that we don't control I would add the check. But again this is a very 💅  opinionated question based on how defensive your are.

src/Assets/ScriptWithBuiltDependenciesAsset.php
puntope: I definitely understood `in_footer` but I needed to read the callback function and the path to understand what was going on with those params.

src/Google/GlobalSiteTag.php
puntope: Ahhh, so was one with `99998` and other with `999999`  and now since we only use one we maintained the highest. Gotcha

src/Google/GlobalSiteTag.php
puntope: > Both actions and filters are hooks.

Yes, sorry I was not precise, I meant `action` as far as I know `filter` is to modify the value and return that value whereas `action` allows to run some task when something happens (I think this is the case) 

> Do you have a suggestion of which action hook to use instead?

Unfortunately I'm not familiar with the available hooks in this part of the codebase. But I can investigate further 👌

src/Google/GlobalSiteTag.php
puntope: For me one thing is the version of the script and other the dependencies required by that script. But checking the codebase it's how WP defines this... so I think it doesn't worth it the update

js/src/gtag-events/utils.js
puntope: ❓  I'm curious about why `retail`. Maybe we sell courses (Education).

puntope:
> Thanks for catching that. I intended that to be disable that option and the AJAX add to cart as well, I've corrected the instructions now.

I tested again with the new instructions and it  works ! ✅

puntope:
> Can you confirm what the full payload of that request is? 

I couldn't see that one anymore. Maybe I mess with some action in my last test.

puntope (approved):
✅  LGTM

- Behaviour works as expected.
- Most of the code suggestions addressed

💅  Still some optional / opinionated code comments but approved it in advances since they are not blockers IMO. Maybe the only one I'd 100% do is the relative to check if variation.variation_id is present after the JQuery event.

js/src/gtag-events/utils.js
mikkamp: That's what's available in Google. Right now we still only sync simple/variable products to Google so generally any services / courses which use special product types aren't synced. Either way it would be something that needs to be changed in the Gtag scripts which we load from Google, so it's beyond what we can control.

src/Google/GlobalSiteTag.php
mikkamp: Yes, that's correct although they are still very similar and it's quite common in other code to hook into a filter if there is no action hook available, even if we don't change the output.

The most convenient way is to use this filter, the only other way I found was to possibly hook into when it's loading the template. But with themes and plugins overriding templates, I felt it was much simpler to just stick with the filter.

mikkamp:
Thanks for all the additional testing, and the clarifications. I've gone ahead and added those typedefs now (turns out I didn't need them in multiple files). I've changed some of the checks to use optional chaining and added the check for variation_id.

Since it's approved and all the tests pass, I'll go ahead and merge it then.

PR #1491, opened 2022-05-09
GMC Request Review - Request a review frontend

puntope:

Closes part of #1165 

This PR implements the call to Google Merchant Account Request Review in the frontend.

### Screenshots:

https://user-images.githubusercontent.com/5908855/167454479-e0c53026-392b-48d8-b496-93357efa2e37.mov

You can perform a full request review flow but real data already. However, Google takes time to review the request or reject them, so could be interesting to hardcode the responses.

1. Checkout this PR
2. You should also enable WP Tracking in order to check some Recorded events.
3. Go to `src/API/Site/Controllers/MerchantCenter/RequestReviewController.php::get_review_status()` and return this on the function 

```
return [
   'status' => 'DISAPPROVED',
   'issues' => [ 'custom_label' ],
   'cooldown' => 0,
   'reviewEligibleRegions' => [ 'US' ]
];
 
```

Go also to line 131 and comment that line to avoid the request to Google.

4. Go to Product Feed and you should see the notice with the request review button
5. Click on the button, mark the checkbox and request a new review
6. See the event `wcadmin_gla_request_review` in the console  
7. The request button should show a spinner
8. After finishing, the modal is closed, the notice disappears, a system notice appears and the status is changed to Under Review in the Feed Summary
9. 5. See the event `wcadmin_gla_request_review_success` in the console  as well as `wcadmin_gla_modal_closed` with `request-review-success` as action param
10. Restore the changes in `src/API/Site/Controllers/MerchantCenter/RequestReviewController.php::get_review_status()`  and update the page
11. You should see the Under Review status in the Feed Summary

### Additional details

If you have by chance a Google disapproved account you can avoid all the hardcoded changes in the PHP file and just do the real request review.

### DIFF
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -929,3 +929,25 @@ export function* updateMCProductVisibility( ids, visible ) {
 		throw error;
 	}
 }
+
+/**
+ * Request a new review for the connected account
+ */
+export function* mcRequestReview() {
+	try {
+		const response = yield apiFetch( {
+			path: `${ API_NAMESPACE }/mc/request-review`,
+		} );
+
+		return {
+			type: TYPES.RECEIVE_MC_REVIEW_REQUEST,
+			response,
+		};
+	} catch ( error ) {
+		yield handleFetchError(
+			error,
+			__( 'Unable to request a new review.', 'google-listings-and-ads' )
+		);
+		throw error;
+	}
+}
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -941,13 +941,10 @@ export function* mcRequestReview() {
 
 		return {
 			type: TYPES.RECEIVE_MC_REVIEW_REQUEST,
-			response,
+			mcReviewRequest: response,
 		};
 	} catch ( error ) {
-		yield handleFetchError(
-			error,
-			__( 'Unable to request a new review.', 'google-listings-and-ads' )
-		);
+		yield handleFetchError( error, error?.message );
 		throw error;
 	}
 }
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -950,7 +950,7 @@ export function* sendMCReviewRequest() {
 			method: 'POST',
 		} );
 
-		return receiveMCReviewRequest( response );
+		return yield receiveMCReviewRequest( response );
 	} catch ( error ) {
 		yield handleFetchError( error, error?.message );
 		throw error;
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -63,6 +63,16 @@ export function handleFetchError( error, message ) {
  */
 
 /**
+ * Account status data. Indicates the current status for the Google MC account.
+ *
+ * @typedef {Object} AccountStatus
+ * @property {string} status Account status. See the available statuses here https://developers.google.com/shopping-content/reference/rest/v2.1/State
+ * @property {number} cooldown Cooldown period timestamp indicating how long the user should wait until the next request
+ * @property {Array} issues List of issue keys for this accoun
+ * @property {Array} reviewEligibleRegions List of region codes available for review
+ */
+
+/**
  *
  * @return {Array<ShippingRate>} Array of individual shipping rates.
  */
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -68,7 +68,7 @@ export function handleFetchError( error, message ) {
  * @typedef {Object} AccountStatus
  * @property {string} status Account status. See the available statuses here https://developers.google.com/shopping-content/reference/rest/v2.1/State
  * @property {number} cooldown Cooldown period timestamp indicating how long the user should wait until the next request
- * @property {Array} issues List of issue keys for this accoun
+ * @property {Array} issues List of issue keys for this account
  * @property {Array} reviewEligibleRegions List of region codes available for review
  */
 
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -933,16 +933,14 @@ export function* updateMCProductVisibility( ids, visible ) {
 /**
  * Request a new review for the connected account
  */
-export function* mcRequestReview() {
+export function* sendMCReviewRequest() {
 	try {
 		const response = yield apiFetch( {
-			path: `${ API_NAMESPACE }/mc/request-review`,
+			path: `${ API_NAMESPACE }/mc/review`,
+			method: 'POST',
 		} );
 
-		return {
-			type: TYPES.RECEIVE_MC_REVIEW_REQUEST,
-			mcReviewRequest: response,
-		};
+		return receiveMCReviewRequest( response );
 	} catch ( error ) {
 		yield handleFetchError( error, error?.message );
 		throw error;
--- a/js/src/product-feed/review-request/index.js
+++ b/js/src/product-feed/review-request/index.js
@@ -17,6 +17,17 @@ import './index.scss';
 
 const showNotice = ( status ) => !! REVIEW_STATUSES[ status ]?.title;
 
+/**
+ * @typedef { import(".~/data/actions").AccountStatus } AccountStatus
+ */
+
+/**
+ * @fires gla_modal_closed with `action: 'request-review-success' | 'maybe-later' | 'dismiss', context: REQUEST_REVIEW`
+ * @fires gla_modal_open with `context: REQUEST_REVIEW`
+ *
+ * @param {Object} props Component props
+ * @param {AccountStatus} props.account Account object
+ */
 const ReviewRequest = ( { account = {} } ) => {
 	const [ modalActive, setModalActive ] = useState( false );
 	const activeIssueType = useActiveIssueType();
--- a/js/src/product-feed/review-request/index.js
+++ b/js/src/product-feed/review-request/index.js
@@ -35,7 +35,7 @@ const ReviewRequest = ( { account = {} } ) => {
 	if (
 		! mcDataHasFinishedResolution ||
 		! hasFinishedResolution ||
-		! showNotice( accountData.status ) ||
+		! showNotice( accountData?.status ) ||
 		activeIssueType !== ISSUE_TYPE_ACCOUNT
 	) {
 		return null;
@@ -63,7 +63,7 @@ const ReviewRequest = ( { account = {} } ) => {
 				createNotice(
 					'success',
 					__(
-						'Account review was successfully requested.',
+						'Your account review was successfully requested.',
 						'google-listings-and-ads'
 					)
 				);
@@ -71,15 +71,8 @@ const ReviewRequest = ( { account = {} } ) => {
 			} )
 			.catch( ( error ) => {
 				recordEvent( 'gla_request_review_failure', {
-					error: error.toString(),
+					error: error?.message,
 				} );
-				createNotice(
-					'error',
-					__(
-						'And error happened processing the account request review.',
-						'google-listings-and-ads'
-					)
-				);
 			} );
 	};
 
--- a/js/src/product-feed/review-request/index.js
+++ b/js/src/product-feed/review-request/index.js
@@ -26,7 +26,7 @@ const showNotice = ( status ) => !! REVIEW_STATUSES[ status ]?.title;
  * @fires gla_modal_open with `context: REQUEST_REVIEW`
  *
  * @param {Object} props Component props
- * @param {AccountStatus} props.account Account object
+ * @param { { isResolving: boolean, hasFinishedResolution: boolean, data: AccountStatus, invalidateResolution: Function } } props.account Account data payload coming from the data store.
  */
 const ReviewRequest = ( { account = {} } ) => {
 	const [ modalActive, setModalActive ] = useState( false );
@@ -41,7 +41,7 @@ const ReviewRequest = ( { account = {} } ) => {
 	if (
 		! mcDataHasFinishedResolution ||
 		! hasFinishedResolution ||
-		! showNotice( accountData?.status ) ||
+		! showNotice( accountData.status ) ||
 		activeIssueType !== ISSUE_TYPE_ACCOUNT
 	) {
 		return null;
--- a/js/src/product-feed/review-request/index.js
+++ b/js/src/product-feed/review-request/index.js
@@ -3,11 +3,14 @@
  */
 import { useState } from '@wordpress/element';
 import { recordEvent } from '@woocommerce/tracks';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
+import { useAppDispatch } from '.~/data';
 import useActiveIssueType from '.~/hooks/useActiveIssueType';
+import useDispatchCoreNotices from '.~/hooks/useDispatchCoreNotices';
 import ReviewRequestModal from './review-request-modal';
 import ReviewRequestNotice from './review-request-notice';
 import { ISSUE_TYPE_ACCOUNT, REQUEST_REVIEW } from '.~/constants';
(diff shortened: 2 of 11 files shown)


js/src/data/actions.js
```diff
@@ -929,3 +929,22 @@ export function* updateMCProductVisibility( ids, visible ) {
...
+export function* mcRequestReview() {
+	try {
+		const response = yield apiFetch( {
+			path: `${ API_NAMESPACE }/mc/request-review`,
+		} );
+
+		return {
+			type: TYPES.RECEIVE_MC_REVIEW_REQUEST,
+			mcReviewRequest: response,
+		};
```
eason9487: 💅  Looks like [`receiveMCReviewRequest` function](https://github.com/woocommerce/google-listings-and-ads/blob/276cc18615e004d292d304f133e536bfd62a963a/js/src/data/actions.js#L877-L882) could be used here as well.

js/src/data/actions.js
```diff
@@ -929,3 +929,22 @@ export function* updateMCProductVisibility( ids, visible ) {
 		throw error;
 	}
 }
+
+/**
+ * Request a new review for the connected account
+ */
+export function* mcRequestReview() {
```
eason9487: 💅  How about naming this action function starting with a verb?

js/src/data/actions.js
```diff
@@ -929,3 +929,22 @@ export function* updateMCProductVisibility( ids, visible ) {
...
+	try {
+		const response = yield apiFetch( {
+			path: `${ API_NAMESPACE }/mc/request-review`,
+		} );
```
eason9487: ❓ 
Looks like the `mc/request-review` API accepts `GET` method but `POST` method is mentioned/named in its code comment/callback.

https://github.com/woocommerce/google-listings-and-ads/blob/276cc18615e004d292d304f133e536bfd62a963a/src/API/Site/Controllers/MerchantCenter/RequestReviewController.php#L65-L73

By experience, this kind of request would normally be sent with `POST` method. Is it intended to use `GET` for both frontend and backend here?

js/src/product-feed/review-request/index.js
```diff
@@ -29,7 +30,7 @@ const ReviewRequest = ( { account = {} } ) => {
 	if (
 		! mcDataHasFinishedResolution ||
 		! hasFinishedResolution ||
-		! showNotice( accountData.status ) ||
+		! showNotice( accountData?.status ) ||
```
eason9487: 💅 
Probably the optional chaining operator is unnecessary here because this component expects the passed-in `account` to be the **wp-data** store data which is always an object payload. The fallback in `account = {}` could be removed to avoid the need to additionally consider the case where the required prop isn't passed in.

js/src/product-feed/review-request/review-request-modal.js
```diff
@@ -43,6 +46,32 @@ const ReviewRequestModal = ( {
...
+					__(
+						'Your account review was successfully requested.',
+						'google-listings-and-ads'
+					)
+				);
+				setIsRequestingReview( false );
+				onClose( 'request-review-success' );
+				recordEvent( 'gla_request_review_success' );
+			} )
+			.catch( ( error ) => {
```
eason9487: 🚧 

The requesting state is only reset when successful, and it will be unable to retry when the error occurs. The `setIsRequestingReview( false )` could be moved to the `catch` block since it's no need to reset requesting state for the successful request.

https://user-images.githubusercontent.com/17420811/167586972-42f5506c-92d2-4db7-ab8a-32df39814931.mp4

js/src/product-feed/review-request/review-request-modal.js
```diff
@@ -43,6 +46,32 @@ const ReviewRequestModal = ( {
...
+				recordEvent( 'gla_request_review_failure', {
+					error: error?.message,
+				} );
+			} );
+	};
+
 	return (
 		<AppModal
```
eason9487, line 109: ❓ 
Not sure if the modal is expected to be closable while waiting for the API response. Otherwise, I'd suggest avoiding it to be closed by:
- Clicking on the cancel button.
- Clicking on the **X** icon on the top-right side of the modal.
- Clicking on outside of the modal.
- Pressing ESC key.

https://user-images.githubusercontent.com/17420811/167589824-367c33c1-8591-435b-bfe7-3dcd88a66b49.mp4

js/src/product-feed/review-request/review-request-modal.js
```diff
@@ -13,6 +13,8 @@ import AppModal from '.~/components/app-modal';
 import AppButton from '.~/components/app-button';
 import AppDocumentationLink from '.~/components/app-documentation-link';
 import ReviewRequestIssues from './review-request-issues';
+import { useAppDispatch } from '.~/data';
+import useDispatchCoreNotices from '.~/hooks/useDispatchCoreNotices';
```
eason9487, line 18: 💅 
It would be nice to add JSDoc for the new events. Therefore, the tracking docs would have up-to-date docs about them when auto-generating.
- gla_request_review_issues_solved_checkbox_click
   - `action: 'check' | 'uncheck'`
- gla_request_review
- gla_request_review_success
- gla_request_review_failure
   - `error` (Not sure if this should be recorded. Please refer to the relevant comment below.)

js/src/product-feed/review-request/index.js
```diff
@@ -20,6 +20,7 @@ const showNotice = ( status ) => !! REVIEW_STATUSES[ status ]?.title;
 const ReviewRequest = ( { account = {} } ) => {
```
eason9487, line 31: 💅 
It would be nice to add JSDoc for the events and their context/action values. Therefore, the tracking docs would have up-to-date docs about them when auto-generating.
- gla_modal_open
   - `context: 'request-review'`
- gla_modal_closed
   - `context: 'request-review'`
   - `action: 'request-review-success' | 'maybe-later' | 'dismiss'`

js/src/product-feed/review-request/review-request-modal.js
```diff
@@ -43,6 +46,32 @@ const ReviewRequestModal = ( {
...
+						'google-listings-and-ads'
+					)
+				);
+				setIsRequestingReview( false );
+				onClose( 'request-review-success' );
+				recordEvent( 'gla_request_review_success' );
+			} )
+			.catch( ( error ) => {
+				recordEvent( 'gla_request_review_failure', {
+					error: error?.message,
```
eason9487: ❓ For the purpose of event tracking, do we need to record the error message? I'm a bit worried about whether there would be sensitive information in the error message.

js/src/product-feed/review-request/review-request-notice.js
```diff
@@ -66,17 +66,19 @@ const ReviewRequestNotice = ( {
...
-						disabled={ !! account.cooldown }
-						text={ __(
-							'Request review',
-							'google-listings-and-ads'
-						) }
-					/>
-				) }
+				{ accountReviewStatus.requestButton &&
+					( account.cooldown ||
+						account.reviewEligibleRegions?.length > 0 ) && (
```
eason9487: 💅 
Not sure if there is any case that the `account.reviewEligibleRegions` will not be an array? Otherwise, the optional chaining operator should be unnecessary.

js/src/product-feed/review-request/index.js
puntope: Not sure about this tbh. As far as I know data could be `undefined` or `null`. Yes, we are checking hasFinishedResolution to cover that, but I will sleep better if there is not a fatal error crashing the entire feed if there is some kind of malformation in the account. 🙏

js/src/product-feed/review-request/review-request-modal.js
puntope: Good spot

js/src/product-feed/review-request/review-request-modal.js
puntope: Great spot

puntope:
Howdy @eason9487 

Thanks for the useful reviews and all the good spots as always.

I applied most of your suggestions. The only one missed is [this non blocking (💅 ) comment ](https://github.com/woocommerce/google-listings-and-ads/pull/1491#discussion_r868954414)

So I guess this is ready for another round. 

Thanks in advance.

js/src/data/actions.js
```diff
@@ -62,6 +62,16 @@ export function handleFetchError( error, message ) {
  *   Please refer to the descriptions of `country`, `targeted_locations` and `allowMultiple` for more context about this property.
  */
 
+/**
+ * Account status data. Indicates the current status for the Google MC account.
+ *
+ * @typedef {Object} AccountStatus
+ * @property {string} status Account status. See the available statuses here https://developers.google.com/shopping-content/reference/rest/v2.1/State
+ * @property {number} cooldown Cooldown period timestamp indicating how long the user should wait until the next request
+ * @property {Array} issues List of issue keys for this accoun
```
eason9487: 💅 Typo: accoun

js/src/product-feed/review-request/review-request-modal.js
```diff
@@ -91,12 +96,15 @@ const ReviewRequestModal = ( {
...
-				setIsRequestingReview( false );
+
 				onClose( 'request-review-success' );
 				recordEvent( 'gla_request_review_success' );
 			} )
 			.catch( () => {
 				recordEvent( 'gla_request_review_failure' );
+			} )
+			.finally( () => {
+				setIsRequestingReview( false );
```
eason9487: 💅 
Updating state after the modal is closed (unmounted) will cause a false warning.

image

In this case, it does not really cause a memory leak, but the false positives may reduce the maintainability a bit. Therefore, I think it would a bit nicer to have one of these adjustments:
- Move it to the `catch` block.
- Or add a code comment for the false warning.
- Or add [`useIsMounted`](https://github.com/woocommerce/google-listings-and-ads/blob/c503325b6812460c674eeddeb3c216b72665d7ec/js/src/hooks/useIsMounted.js#L8) to prevent the false warning.

js/src/product-feed/review-request/index.js
```diff
@@ -17,9 +17,21 @@ import './index.scss';
...
+/**
+ * @typedef { import(".~/data/actions").AccountStatus } AccountStatus
+ */
+
+/**
+ * @fires gla_modal_closed with `action: 'request-review-success' | 'maybe-later' | 'dismiss', context: REQUEST_REVIEW`
+ * @fires gla_modal_open with `context: REQUEST_REVIEW`
+ *
+ * @param {Object} props Component props
+ * @param {AccountStatus} props.account Account object
```
eason9487: ❓  The type of `props.account` is not the `AccountStatus` type but the payload shape of `useAppSelectDispatch`.

js/src/product-feed/review-request/index.js
eason9487: > As far as I know data could be `undefined` or `null`.

❔ 
Maybe I missed something, so I'm not pretty sure when the data would be `undefined` or `null`. Unless the GET/POST `mc/review` API will respond successfully with other data structures rather than [`AccountStatus`](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/js/src/data/actions.js#L65-L73), otherwise, the `account.data` here is always the `AccountStatus` structure.
- The data structure of `account` prop is the shape of [`useAppSelectDispatch`](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/js/src/hooks/useAppSelectDispatch.js#L28-L36), which is always an object.
- The data structure of `account.data` here is always the [`mc_review_request` data structure](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/js/src/data/reducer.js#L40-L45) provided from the `getMCReviewRequest` selector of wp-data.
- The `get_review_status` function has [handling for the `null` case](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/src/API/Site/Controllers/MerchantCenter/RequestReviewController.php#L235-L239). (It seems not quite accurate that the doc of `get_review_status` says [it may return `null`](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/src/API/Site/Controllers/MerchantCenter/RequestReviewController.php#L229).) And even somehow that function returns `null` to callers, the response data will still be like this:
   ```json
   {
     "status": null,
     "cooldown": null,
     "issues": null,
     "reviewEligibleRegions": null
   }
   ```

eason9487: > As far as I know data could be `undefined` or `null`.

❔ 
Maybe I missed something, so I'm not pretty sure when the data would be `undefined` or `null`. Unless the GET/POST `mc/review` API will respond successfully with other data structures rather than [`AccountStatus`](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/js/src/data/actions.js#L65-L73), otherwise, the `account.data` here is always the `AccountStatus` structure.
- The data structure of `account` prop is the shape of [`useAppSelectDispatch`](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/js/src/hooks/useAppSelectDispatch.js#L28-L36), which is always an object.
- The data structure of `account.data` here is always the [`mc_review_request` data structure](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/js/src/data/reducer.js#L40-L45) provided from the `getMCReviewRequest` selector of wp-data.
- The `get_review_status` function has [handling for the `null` case](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/src/API/Site/Controllers/MerchantCenter/RequestReviewController.php#L235-L239). (It seems not quite accurate that the doc of `get_review_status` says [it may return `null`](https://github.com/woocommerce/google-listings-and-ads/blob/255a575a0d5feefba428fdf8dbaa08dc9449aca6/src/API/Site/Controllers/MerchantCenter/RequestReviewController.php#L229).) And even somehow that function returns `null` to callers, the response data will still be like this:
   ```json
   {
     "status": null,
     "cooldown": null,
     "issues": null,
     "reviewEligibleRegions": null
   }
   ```

puntope:
Hi again @eason9487 

I added new adjustments. Let me now if now the PR is good to go

js/src/product-feed/review-request/index.js
puntope: Right. I tried to log the account payload and is true that is always with the AccountStatus shape. So no undefined. I was kinda sure at some point it was like ` { hasFinishedResolution: false, data: undefined ...}` but seems like it's not the case 🤔

js/src/product-feed/review-request/review-request-modal.js
eason9487: The d0538de6e69d1ad95b70938026054d7472b37cbb change is basically the same logic as the use of `.finally()` here. It still has that false warning when successful.

puntope: I dont see that warning anymore. Checking

js/src/data/actions.js
```diff
@@ -929,3 +939,20 @@ export function* updateMCProductVisibility( ids, visible ) {
...
+ * Request a new review for the connected account
+ */
+export function* sendMCReviewRequest() {
+	try {
+		const response = yield apiFetch( {
+			path: `${ API_NAMESPACE }/mc/review`,
+			method: 'POST',
+		} );
+
+		return receiveMCReviewRequest( response );
```
eason9487: 🚧 

The received `response` will not be passed to wp-data reducer, and it further causes the UI won't update after sending the review request.


https://user-images.githubusercontent.com/17420811/167812912-bd99ee96-36f9-4469-b62d-74750c9050a8.mp4



Since the `receiveMCReviewRequest` is also a generator function, it needs to be called with `yield` here, such as `return yield receiveMCReviewRequest( response );` . Or, change the `receiveMCReviewRequest` to a non-generator function.

puntope: Thanks for catching this 🥇 I totally missed the yield

js/src/product-feed/review-request/review-request-modal.js
puntope: Fixed @eason9487  thanks

eason9487 (approved):
Thanks for the implementation and adjustments. Tested with both simulated data and real API calls to Google side, and work well. LGTM.

PR #1472, opened 2022-05-03
Add banner for PMax migration campaigns 

jorgemd24:
### Changes proposed in this Pull Request:

This PR adds the banner in the GLA Dashboard with information related to the PMax migration status. See more context in the following links:

- [Design Proposal i2: Performance Max Migration (Milestone 2)](https://woodesigngrowcontractor.wordpress.com/2022/04/20/design-proposal-i2-performance-max-migration-milestone-2/)
- [Figma File](https://href.li/?https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=7840%3A127644)

The banner for the reports section, campaign names and inbox notifications will be addressed in different PRs. 

### Before Migration
- If the user has not been converted show the following notice:

image

- Banner goes away when X’d or until all campaigns are migrated

### After Migration
- If the user has been converted show the following notice:

image

- Banner goes away when X’d, or after 1 month after it is first displayed.

### Detailed test instructions:

The conversion status is determined for the wp option: `gla_campaign_convert_status`, it is an object with two properties: `status` & `updated`. The property `updated` is a UNIX time in seconds that indicates when was the last time that the status was checked until it becomes converted. 

#### Unconverted
`a:2:{s:6:"status";s:11:"unconverted";s:7:"updated";i:1651517855;}` //May 02 2022

#### Converted
`a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1651517855;}` //May 02 2022

If you have Smart Shopping campaigns, the starting status will be 'unconverted', if all campaigns are PMax then your status will be 'converted'
 
### Before Migration
1. Go to the GLA dashboard
2. Check that the notice is showing

image

4. If you click X, the banner disappears and if you refresh it will not show again. Remove from the localStorage the item `gla-is-before-migration-notice-dismissed` if you like to see it again.

### After Migration
1. Update manually the wp option to: `a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1651517855;}`
1. Go to the GLA dashboard
2. Check that the notice is showing

image

4. If you click X, the banner disappears. Remove from the localStorage the item `gla-is-after-migration-notice-dismissed` if you like to see it again.
5. Change manually the time in the wp option to be > 1 month ago and the banner should be gone. Example: `a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1646308364;}` //Mar 03 2022

### Additional details:
[...]

PR #1492, opened 2022-05-09
PMax Banner Reports Section

jorgemd24:

This PR adds the banner in the GLA Reports Section with information related to the PMax migration status. See more context in the following links:

- [Design Proposal i2: Performance Max Migration (Milestone 2)](https://woodesigngrowcontractor.wordpress.com/2022/04/20/design-proposal-i2-performance-max-migration-milestone-2/)
- https://github.com/woocommerce/google-listings-and-ads/pull/1472

### Before Migration

- The banner should not be displayed.

### After Migration

- If the user has been converted, show the following notice:

image

- Banner goes away when X’d, or after 1 month after it is first displayed.

The conversion status is determined for the wp option: `gla_campaign_convert_status`, it is an object with two properties: status & updated. The property updated is a UNIX time in seconds that indicates when was the last time that the status was checked until it becomes converted.

#### Unconverted
a:2:{s:6:"status";s:11:"unconverted";s:7:"updated";i:1651517855;} //May 02 2022

#### Converted
a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1651517855;} //May 02 2022

If you have Smart Shopping campaigns, the starting status will be 'unconverted', if all campaigns are PMax then your status will be 'converted'.

### Before Migration

1. Go to the GLA Reports
2. The banner should not appear.

### After Migration

1. Update manually the wp option to: `a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1651517855;}` //May 02 2022
2. Go to the GLA Reports
3. Check that the notice is showing

image

4. If you click X, the banner disappears. Remove from the localStorage the item `gla-is-reports-migration-notice-dismissed` if you like to see it again.
5. Change manually the time in the wp option to be > 1 month ago and the banner should be gone. Example: `a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1646308364;}` //Mar 03 2022

### Additional details:

The link for "Learn more about this upgrade" is pending to be confirmed.

### Changelog entry

> Add - Pmax migration banner reports


js/src/components/campaign-conversion-notice/reports-notice.js
```diff
@@ -0,0 +1,69 @@
...
+						readMoreLink: (
+							<AppDocumentationLink
+								context={ context }
+								//TODO Update link when it is defined
+								href="#"
+								eventName="gla_upgrade_campaign_learn_more_link_click"
+								linkId="campaign-conversion-status-after-migration-reports-read-more"
+							/>
+						),
+					}
```
ianlin: This component was actually very similar to the component `js/src/components/campaign-conversion-notice/index.js` that implemented in the previous PR, I'm thinking maybe we could add a shared component and accept props like `localStorageKey`, `className`, and `children` etc so we could reuse some codes.

js/src/components/campaign-conversion-notice/reports-notice.js
```diff
@@ -0,0 +1,69 @@
...
+						'Your existing campaigns have been upgraded to Performance Max. <readMoreLink>Learn more about this upgrade</readMoreLink>',
+						'google-listings-and-ads'
+					),
+					{
+						readMoreLink: (
+							<AppDocumentationLink
+								context={ context }
+								//TODO Update link when it is defined
+								href="#"
+								eventName="gla_upgrade_campaign_learn_more_link_click"
```
ianlin: I think the event name of `AppDocumentationLink` should always be `gla_documentation_link_click` that is defined in `AppDocumentationLink` component. When we want to distinguish different link clicks we can use the event properties, in this case `linkId` or something else. So an analyst can get the stats of the total document link clicks, as well as clicks per link by filtering out the event properties. It might need to be changed in `js/src/components/campaign-conversion-notice/index.js` from the previous PR.

However from the entire code base there were some places that defined the event name for `AppDocumentationLink`, I will create a follow-up issue for it later.

js/src/components/campaign-conversion-notice/index.scss
```diff
@@ -35,3 +35,16 @@
 		}
 	}
 }
+
+.gla-campaign-conversion-status-reports-notice {
+	margin: $grid-unit-40 0 $grid-unit-10 0;
```
ianlin: 💅  It could be simplified to `margin: $grid-unit-40 0 $grid-unit-10;`

js/src/components/campaign-conversion-notice/index.scss
```diff
@@ -35,3 +35,16 @@
...
+	padding: $grid-unit-15 $grid-unit-40;
```
ianlin: 💅  From the design the left and right padding is just 20px, so maybe we can change the left and right padding to `$grid-unit-30`.

ianlin (commented):
Thanks for adding this, tested locally and the behaviour was working as expected. I left some comments that might be worth discussing.

js/src/components/campaign-conversion-notice/index.scss
jorgemd24: Added [d6b5e5e](https://github.com/woocommerce/google-listings-and-ads/pull/1492/commits/d6b5e5e45a41fea9cbcc92691145da9adb16f4cb)

js/src/components/campaign-conversion-notice/index.scss
jorgemd24: Added [c832e74](https://github.com/woocommerce/google-listings-and-ads/pull/1492/commits/c832e746f984f81da8c9945fd00ba5ac599b4a6d)

js/src/components/campaign-conversion-notice/reports-notice.js
jorgemd24: Thanks, Ian for the explanation! I remove the eventName from AppDocumentationLink. See [d214109](https://github.com/woocommerce/google-listings-and-ads/pull/1492/commits/d214109c486ee8369f27f951e9b24476e1f29784)

jorgemd24:
Thanks, @ianlin for reviewing this and your comments! I have updated the code with your suggestions and I refactor the previous banner and the reports banner so we can reuse more code. 

This PR is ready for another round!

js/src/components/campaign-conversion-notice/migration-notice.js
```diff
@@ -0,0 +1,48 @@
...
+import { useState } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import localStorage from '.~/utils/localStorage';
+
+/**
+ * Shows Notice {@link Notice}
+ * It is the wrapper for the campaign convertion notices
```
ianlin: 💅 Just a typo, should be "conversion".

js/src/components/campaign-conversion-notice/index.js
```diff
@@ -38,31 +33,21 @@ const ExternalIcon = () => (
  * @param {string} props.context Context or page on which the notice is shown, to be forwarded to the link's track event.
  * @return {JSX.Element} {@link Notice} element with the info message and the link to the documentation.
  */
-const CampaignConversionStatusNotice = ( { context } ) => {
+const CampaignConversionDashboardNotice = ( { context } ) => {
```
ianlin, line 35: I couldn't comment on the lines that are not part of the changes, but in the events JSDoc above this, we are no longer sending the event name `gla_upgrade_campaign_learn_more_link_click` so the comment can be removed.

js/src/components/campaign-conversion-notice/reports-notice.js
```diff
@@ -0,0 +1,45 @@
...
+import { glaData } from '.~/constants';
+import CONVERSION_STATUSES from './conversion-statuses';
+import getConversionCampaignStatusNotice from '.~/utils/getConversionCampaignStatusNotice';
+import MigrationCampaignNotice from './migration-notice';
+
+/**
+ * Shows Notice {@link Notice}
+ * providing information about the conversion status of PMax campaigns
+ *
+ * @fires gla_upgrade_campaign_learn_more_link_click with `{ context: 'reports-programs, linkId: 'campaign-conversion-status-after-migration-reports-read-more', href: '#' }`.
```
ianlin: We are no longer sending the event name `gla_upgrade_campaign_learn_more_link_click` so the comment can be removed.

js/src/components/campaign-conversion-notice/conversion-statuses.js
```diff
@@ -66,6 +69,30 @@ const AFTER_CONVERSION = {
...
+	localStorageKey: LOCAL_STORAGE_KEYS.IS_REPORTS_MIGRATION_NOTICE_DISMISSED,
+	content: createInterpolateElement(
+		__(
+			'Your existing campaigns have been upgraded to Performance Max. <readMoreLink>Learn more about this upgrade</readMoreLink>',
+			'google-listings-and-ads'
+		),
+		{
+			readMoreLink: (
+				<AppDocumentationLink
+					context={ 'reports-programs' }
```
ianlin: 💅  The string doesn't need to be wrapped in `{}`.

js/src/components/campaign-conversion-notice/migration-notice.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {string} props.className Class to use in the Notice component.
+ * @param {Object} props.localStorageKey Local Storage Key where is keep the dismiss state.
+ * @return {JSX.Element} {@link Notice} element with the info message and the link to the documentation.
+ */
+const MigrationCampaignNotice = ( {
+	children,
+	className,
+	localStorageKey,
+} ) => {
+	const defaultDismissedValue = !! localStorage.get( localStorageKey );
```
ianlin: I feel like this component could be promoted to a standard shared component in this project, if we name it `DismissibleNotice` and put it under `js/src/components`, and only set the local storage if the provided prop `localStorageKey` is not empty.

But it's not a blocking part for this PR, we can do it later when doing other TODOs in the future. E.g. when the "learn more" link is ready

js/src/components/campaign-conversion-notice/index.js
```diff
@@ -1,22 +1,17 @@
...
 /**
  * Internal dependencies
  */
 import { glaData } from '.~/constants';
-import localStorage from '.~/utils/localStorage';
 import AppDocumentationLink from '.~/components/app-documentation-link';
 import CONVERSION_STATUSES from './conversion-statuses';
 import getConversionCampaignStatusNotice from '.~/utils/getConversionCampaignStatusNotice';
+import MigrationCampaignNotice from './migration-notice';
 import './index.scss';
```
ianlin, line 15: ❔ Question for importing scss: I'm not familiar with css/scss, but we import `index.scss` in this file but did not import it in `campaign-conversion-notice/reports-notice.js`. The styles of the report notice looks great on the browser so it means there is no problem as the styles are loaded by `campaign-conversion-notice/index.js`. I feel like it's not a good way as it's a bit implicit. So maybe we could move the styles of `reports-notice.js` to, for example, `reports-notice.scss` and import it from `reports-notice.js`?

js/src/components/campaign-conversion-notice/report-notice.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Internal dependencies
+ */
+import CampaignConversionReportsNotice from '.~/components/campaign-conversion-notice/reports-notice';
+import getConversionCampaignStatusNotice from '.~/utils/getConversionCampaignStatusNotice';
+
+jest.mock( '.~/utils/getConversionCampaignStatusNotice', () => jest.fn() );
+
+describe( 'Notice Reports Campaign Migration', () => {
```
ianlin: 💅  This test file's name is `report-notice.test.js` while the filename of the component it's testing on is `reports-notice.js`.

ianlin (approved):
Hey @jorgemd24, thanks for all the changes. I approve it in advance but still left some suggestions in the comments. The only comment that I think should be changed before merging is to remove the event JSDoc for `gla_upgrade_campaign_learn_more_link_click`.

js/src/components/campaign-conversion-notice/index.js
jorgemd24: Yeah thats a good idea, I have changed in [ebedc71](https://github.com/woocommerce/google-listings-and-ads/pull/1492/commits/ebedc71cd053261e42eb4f2ce3b74b3fe2bdba86)

js/src/components/campaign-conversion-notice/migration-notice.js
jorgemd24: I had the same feeling! I made some changes to the component, basically:
- I updated the name.
- It checks if the `localStorageKey ` is set or not.

However, I think if we want to leave it as an independent component, it would be convenient to add some callback function on the onRemove function so it is more customized depending on the user's needs. I think for now I will leave it as it is and in a different PR, we can focus to make this component a standard shared component.

jorgemd24:
Thanks, @ianlin for the second review! I have added your suggestions and I left a comment about making the `DismissableNotice` a shared component.

js/src/components/campaign-conversion-notice/migration-notice.js
ianlin: Makes sense, thanks for the explanation!

ianlin (approved):
Thanks for all the changes, it's good to go!

PR #1498, opened 2022-05-11
Fix the unit tests for WooCommerce 6.5

mikkamp:

This PR fixes the unit tests to allow them to run with WooCommerce 6.5 (first version with WC Admin fully included). The following two fixes are applied:

- Generate a feature config file if the install script is present
- Do not call the WC Admin install class (database tables are created as part of the regular install)

1. Now that WC 6.5 is released it will automatically install it for unit tests since we use `latest` 
2. Check that the unit tests successfully run for this PR.

### Changelog entry
* Fix - Unit tests for WooCommerce 6.5


eason9487 (approved):
Tested on local dev env and the unit tests work correctly as well. LGTM.

PR #1499, opened 2022-05-11
Improve tests and event handlers in Minimum Order refactor

tomalec:
_This PR is a proposal/suggestion for an improvement/alternative to the https://github.com/woocommerce/google-listings-and-ads/pull/1410_

- :nail_care:  Extract some hooks to manual mocks, (https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/f0afaab4900cd96f3b2cc6099e7b96c15a5f908f)
   for easier, DRYier, shareable mocking.
   Mock:
   - `useAppSelectDispatch`
   - `useCountryKeyNameMap`
   - `useStoreCurrency`

- Add `MinimumOrderCard.onChange` tests. (https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/4ebd22391627539c9c7182008417ff18bd4a1056)
   Test actual component behavior, instead of its implementation detail.
   Addresses https://github.com/woocommerce/google-listings-and-ads/pull/1410#discussion_r869201399
- Replace `getMinimumOrder*Handler`s with simpler translation functions. (https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/f682f9e2fdb7b20c88804535fe4d80f5891a9f47)
   Make them functions that perform only data type conversion, and operation application, and leave `onChange` call and event handling functionality to `MinimumOrderCard`.

   To make those functions easier to test and reason about.

   Addresses https://github.com/woocommerce/google-listings-and-ads/pull/1410#discussion_r870771644
- Remove add & remove minimum order group handlers/translators. (https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/9d474a5e74f72c08f14e221c50caf37cb99534b6)

   Cover them by `changeGroup` to reduce the amount of duplicated code.

   Address https://github.com/woocommerce/google-listings-and-ads/pull/1410#discussion_r871578823

1. `npm i && npm run test:js` 
2. Smoke test adding, changing, deleting MinimumOrder thresholds
3. Review the final code, to see if it's more or less readable than before.


js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
```diff
@@ -0,0 +1,52 @@
...
+ * Will set `shipingRate.options.free_shipping_threshold` to `newGroup.threshold`
+ * for the shipping rate countries that are listed in `newGroup.countries` (if provided),
+ * and set `threshold` to `undefined` for the countries which were in `oldGroup.countries` (if provided),
+ * but not in the new.
+ *
+ * @param {Array<ShippingRate>} value Shipping rates.
+ * @param {MinimumOrderGroup} [oldGroup] Old minimum order group.
+ * @param {MinimumOrderGroup} [newGroup] New minimum order group.
+ */
+export const changeGroup = ( value, oldGroup, newGroup ) => {
```
puntope: 💅  Since we are doing several refactoring in favor of

```
const changeGroup = () ...

export default changeGroup;

```

I'd suggest to do that way here as well.  Even tho we are not tracking anything in this file we will get more consistency over our codebase

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
```diff
@@ -0,0 +1,52 @@
...
+ * Returns new shipping rates value, with updated group.
+ *
+ * Will set `shipingRate.options.free_shipping_threshold` to `newGroup.threshold`
+ * for the shipping rate countries that are listed in `newGroup.countries` (if provided),
+ * and set `threshold` to `undefined` for the countries which were in `oldGroup.countries` (if provided),
+ * but not in the new.
+ *
+ * @param {Array<ShippingRate>} value Shipping rates.
+ * @param {MinimumOrderGroup} [oldGroup] Old minimum order group.
+ * @param {MinimumOrderGroup} [newGroup] New minimum order group.
```
puntope: 💅  Missed Return prop

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
puntope: 💅  What about `changeMinimumOrderGroup` or `minimumOrderChangeGroup`?

Even tho is located in a folder named like that it could be non descriptive and pretty much generic to use `changeGroup`

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * @typedef { import(".~/data/actions").ShippingRate } ShippingRate
+ * @typedef { import("./typedefs").MinimumOrderGroup } MinimumOrderGroup
+ */
```
puntope: 💅  Maybe moving this to` typedefs.js` file?

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
```diff
@@ -0,0 +1,52 @@
...
+		const newShippingRate = {
+			...shippingRate,
+			options: {
+				...shippingRate.options,
+			},
+		};
+
+		if (
+			newGroup &&
+			newGroup.countries.includes( newShippingRate.country )
```
puntope: 💅 

`newGroup?.countries.includes( newShippingRate.country )`

and

`oldGroup?.countries.includes( newShippingRate.country )`

also works

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
```diff
@@ -0,0 +1,240 @@
...
+			rendered = render(
+				<MinimumOrderCard value={ value } onChange={ onChange } />
+			);
+		} );
+
+		afterEach( () => {
+			jest.clearAllMocks();
+		} );
+
+		test( 'When the new minimum order value is provided for a remianing country, calls the `onChange` callback with the new value containing `shippingRate.options.free_shipping_threshold` set to the given value', async () => {
```
puntope: 💅  Typo "remianing"

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.test.js
```diff
@@ -1,9 +1,9 @@
 /**
  * Internal dependencies
  */
-import getMinimumOrderChangeHandler from './getMinimumOrderChangeHandler';
+import { changeGroup } from './changeGroup';
 
-describe( 'getMinimumOrderChangeHandler( value, onChange, oldGroup )( newGroup )', () => {
+describe( 'changeGroup( value, oldGroup, newGroup )', () => {
```
puntope: 💅  Not sure about this practice of adding the params in the test describe. That means on each change in the function signature we should remember to change them here as well since the test wont fail if the description is not accurate.

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.test.js
```diff
@@ -34,13 +34,88 @@ describe( 'getMinimumOrderChangeHandler( value, onChange, oldGroup )( newGroup )
 			},
 		},
 	] );
-	const mockOnChange = jest.fn();
+	// Pure add.
+	it( "returns new value with `newGroup.country`'s `shippingRate.options.free_shipping_threshold` updated to `newGroup.threshold`, when falsy `oldGroup` is given", () => {
```
puntope: ❓/ 💅   Doesn't matter if `oldGorup` is falsy because value contains the newGroup country

puntope (approved):
✅  LGTM

- I tested the minimum order shipping on free campaigns and work as expected, I delete, create, and update groups.
- I read the code and the changeGroup function improves the legibility of the code.

I left some non blocking comments tho 💅

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
tomalec: > :nail_care: Since we are doing several refactoring in favor of
> ```
> const changeGroup = () ...
> export default changeGroup;
> ```

Do you mean having default export, or just separating `export` from statement, to workaround tracking docs generator issue?

Speaking of `export default` I'm not sure if we discussed it, and came to any conclusion/settled contention?

Personally, I consider `export default` a backward-compatibility layer done in ESM to support easier migration from legacy CJS modules. And I usually prefer named exports. There is a nice listed of prose and cons discussed at pdCMQf-sZ-p2#export-conventions

tomalec: > :nail_care: What about `changeMinimumOrderGroup` or `minimumOrderChangeGroup`?
> Even tho is located in a folder named like that it could be non descriptive and pretty much generic to use `changeGroup`

Thanks for that definitely a good point.
Just to note, the function does not technically change the `MinimumOrderGroup`. It takes one to change the ShippingRates array, to conceptually change the group within. So it's more like `changeMinimumOrderOfAGroup`, but `changeMinimumOrderGroup` for short and conceptual description is fine to me.

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.test.js
tomalec: I just left it as it was to minimize the number of changes, but I agree with your point. Then given, that the end goal is to merge it to 1410, which adds it from scratch, I'd remove them.

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
tomalec: We would still need to import it, like `import("./typedefs").MinimumOrderGroup`

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.test.js
tomalec: > Doesn't matter if `oldGorup` is falsy because value contains the newGroup country

That's why I try to test and express in tests' descriptions that it "deosn't metter" by testing/having a combination when we provide `oldGroup` and don't. Here and in https://github.com/woocommerce/google-listings-and-ads/pull/1499/files/9d474a5e74f72c08f14e221c50caf37cb99534b6#diff-73c3df7a03eb8d9c49534fb11e4c120430035eb3cee5a8fec78b71c03bab225eR207

Maybe we could improve the description of the latter, but I don;t have a good idea.

tomalec:
Thanks for the review @puntope I hope I addressed all your comments, and it's ready for another round.
I'd still wait for @ecgan's approval before merging it.

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
puntope: > Do you mean having default export, or just separating export from statement, to workaround tracking docs generator issue?

I meant this.

js/src/components/shipping-rate-section/minimum-order-card/changeMinimumOrderGroup.js
```diff
@@ -0,0 +1,47 @@
...
+ * for the shipping rate countries that are listed in `newGroup.countries` (if provided),
+ * and set `threshold` to `undefined` for the countries which were in `oldGroup.countries` (if provided),
+ * but not in the new.
+ *
+ * @param {Array<ShippingRate>} value Shipping rates.
+ * @param {MinimumOrderGroup} [oldGroup] Old minimum order group.
+ * @param {MinimumOrderGroup} [newGroup] New minimum order group.
+ * @return {Array<ShippingRate>} New, updated value.
+ */
+export const changeMinimumOrderGroup = ( value, oldGroup, newGroup ) => {
```
ecgan, line 19: 💅 I think I can understand why this is called `changeMinimumOrderGroup` (related to https://github.com/woocommerce/google-listings-and-ads/pull/1499#discussion_r872443466), especially when we view it from UI / user perspective, the users are changing the threshold value of a minimum order group. However, when I look into this function and what it does, I'm leaning towards calling it `getShippingRateNewValue`, as in "get the new `ShippingRate` value based on `value`, `oldGroup` and `newGroup`" (and this is how I would describe or summarize it in JSDoc too).

`changeMinimumOrderGroup` feels like it is going to change `oldGroup` or `newGroup`, but they are not changed in the function.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
```diff
@@ -39,25 +35,27 @@ const MinimumOrderCard = ( props ) => {
 			( shippingRate ) => shippingRate.country
 		);
 
-		/**
-		 * If group length is 1, we render the group,
-		 * regardless of threshold is defined or not.
-		 */
+		// If group length is 1, we render the group,
+		// regardless of threshold is defined or not.
```
ecgan, line 50: 💅 Just curious, is there a reason to change this to `//`? Perhaps it is to reduce the number of lines? Personally I would prefer to have block comments (`/* ... */`) for comments that span multiple lines. 

I understand the previous code `/** ...*/` is a JSDoc comment, and I guess we shouldn't use that for general block comment, although I am not sure if there is any particular downside if we use it that way.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
```diff
@@ -39,25 +35,27 @@ const MinimumOrderCard = ( props ) => {
...
+								groups[ 0 ],
+								newGroup
+							)
+						)
+					}
+					onDelete={ () =>
+						onChange(
+							changeMinimumOrderGroup( value, groups[ 0 ] )
+						)
+					}
```
ecgan: 💅 This comment applies to the add, edit, and delete operations (`onSubmit`, `onChange`, and `onDelete`) in this `MinimumOrderCard` component.

Instead of using inline arrow functions (and having a bit of duplicated code across the two `onChange` and `onDelete`), I would still prefer to have a dedicated "handlers" code block so that it is easier to see how those add/edit/delete operations is mapped to the `changeMinimumOrderGroup` call. In case those operations need to be modified, they just need to be modified in their corresponding handler code instead of the duplicated inline arrow function.

Refer to the screenshot below:

image

Here's the code if you want to copy and paste: 

```js
		const getAddHandler = ( newGroup ) => {
			const newValue = changeMinimumOrderGroup( value, null, newGroup );
			onChange( newValue );
		};

		const getChangeHandler = ( oldGroup ) => ( newGroup ) => {
			const newValue = changeMinimumOrderGroup(
				value,
				oldGroup,
				newGroup
			);
			onChange( newValue );
		};

		const getDeleteHandler = ( oldGroup ) => () => {
			const newValue = changeMinimumOrderGroup( value, oldGroup );
			onChange( newValue );
		};
```

So with the above code, we can easily see how add/change/delete operations is mapped to `changeMinimumOrderGroup`, and how they depend on `oldGroup` and `newGroup`. There are only three `changeMinimumOrderGroup` calls. With the inline arrow functions, IMO it is not as clear, and there are five `changeMinimumOrderGroup` calls.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
```diff
@@ -0,0 +1,240 @@
...
+			jest.clearAllMocks();
+		} );
+
+		test( 'When the new minimum order value is provided for a remaining country, calls the `onChange` callback with the new value containing `shippingRate.options.free_shipping_threshold` set to the given value', async () => {
+			// Open "Add another…" modal.
+			userEvent.click( rendered.getByRole( 'button', { name: /Add/ } ) );
+			// Input some value.
+			const input = screen.getByRole( 'textbox' );
+			// For some reason it's typed backwards.
+			await userEvent.type( input, '03' );
```
ecgan, line 69: 🤯

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
```diff
@@ -0,0 +1,240 @@
...
+				expect( onChange ).toHaveBeenCalledWith( [
+					{
+						...value[ 0 ],
+						options: {
+							free_shipping_threshold: 30,
+						},
+					},
+					value[ 1 ],
+					value[ 2 ],
+				] );
```
ecgan: 💅 About the expected value, we are coding the `value[1]` and `value[2]` there. My concern is that if someone were to change the `value` in the `beforeEach` block, say adding another element into the `value` array (so there would be `value[3]`), this test here would then fail, and one needs to add the `value[3]` here to make it pass. 

I would prefer to have something like this, so that it works with the above scenario: 

```js
// this is the expected new value.
const newValue = [ ...value ];
newValue[ 0 ] = {
	...newValue[ 0 ],
	options: {
		free_shipping_threshold: 30,
	},
};

// Wait for Form component's onSubmit microtask.
await waitFor( () => {
	expect( onChange ).toHaveBeenCalledTimes( 1 );
	expect( onChange ).toHaveBeenCalledWith( newValue );
} );
```

Note that this comment also applies to other tests in this file.

ecgan (approved):
@tomalec , thanks so much for your PR here, I like it! 😍 I especially like how you dedupe the code across the `get*Handlers` with the `changeMinimumOrderGroup` function, and how `changeMinimumOrderGroup` just focus on data transformation without needing to care about `onChange`. When I was working on PR #1410, I already felt there was a bit of duplicated code across the add/edit/delete operations (e.g. `options.free_shipping_threshold = undefined`), but at that time I couldn't think of a way to refactor / simplify it, probably because I was in too deep with the `get*Handlers`. 😂 So thanks for the changes here! 👍 

I tested the functionalities, it is working as expected. I left a few 💅 comments below. Notable ones are:

- `changeMinimumOrderGroup` naming.
- Having simple `get*Handler`s so that we can easily see how add/edit/delete operations are mapped to `changeMinimumOrderGroup`.
- Expected values in tests.

💅 Also, I ran `npm run test:js` and noticed that MinimumOrderCard test has one uncovered branch and line. Wondering if you like to cover it and make it 100%? 💯 😄 

image

Lastly, regarding the merge, which one would you prefer?

Option 1: Merge this PR #1499 into PR #1410 first, and then merge #1410 into `develop`.
Option 2: Merge PR #1410 first into `develop`, and then merge this PR #1499 into `develop`.

Personally, I'm leaning towards option 2, so that each PR is smaller, and because the PRs don't need to go into `develop` together; with option 1, PR #1410 would be bigger with more commits from both of us.

If we go with option 2, you would need to approve #1410 first (with a comment mentioning more refactor enhancement in this PR #1499, of course), then we can merge #1410 into `develop`. 

Since things are working in this PR, I'm approving here. If you make some substantial changes and want another round of review, feel free to request my review again. 🙂 

Thanks so much again for this! 🙏 😄

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
tomalec: Personally, I use `//` for in-line comments that clarify what, why, and what for happens in the code. While block comments `/** */` to describe an entity.

Also, VSCode was catching `/** */` here as a description of a function in https://github.com/woocommerce/google-listings-and-ads/pull/1499/files/3109ee4f7f8d5b0ba6d67f6ff8f1df54df1ab7ca#diff-19741a9dbe0895f99f4b04a6a88e21d3a8445917bea2f62ccf3f2610acff92d8R45 which is totally incorrect.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
tomalec: Yup, I was also :exploding_head: 
 I guess it's either a bug in the testing library or in our code that handles the input. But I didn't want to spend more time here, as it is unrelated to this bit.

js/src/components/shipping-rate-section/minimum-order-card/changeMinimumOrderGroup.js
tomalec: The hardest part of programming - naming ;)

I'm also not happy with `changeMinimumOrderGroup` for the exact reason you mentioned. However, I also don't like `getShippingRateNewValue`:
- It sounds like a getter, so why would I need to pass any args?
- Does it always get up-to-date value?
- `getShippingRateNewValue( value, oldGroup, newGroup ) `:
   - To get a value I need to give a value?
   - "Group" what does a group have to do for shipping rate value? Is that a shipping rates group?
   
 The fact it returns a new value instead of changing the existing one, does not look extremely important to emphasize. Given we write in a framework that tries to embrace functional programming and immutability. (If we already pay a price for this, let's take some benefits ;) )
   
   Personally, I like when the function name is optimized for a consumer's perspective, not implementer's perspective. So instead of stating what the function technically does (like the source code do), it gives hints for the consumer, like:
- What does it do conceptually?
- When would I need/ use it? 

So I would describe this function as: "changeShippingRatesArrayAcordingToChangedMinimumOrderGroup", "changeMinimumOrderOfAGroupWithinShippingRates", "applyMiniumOrderGroupChangeToShippingRates" - but those names are a bit lengthy :joy: 

----


Maybe we can put it that way:

`changeMinimumOrderGroup` is a small utility function, to be used here locally, not plugin wide. This is util used only in `MinimumOrderCard` defined and used only within `ShippingRateSection`. Therefore, maybe we can assume that from the perspective, of the only consumer of this function - `MinimunOrderCard`:
- the "value" is always "shipping rates array" (as specified in props) 
- "group"/"countries group/threshold group" is always "Mimimum Order Group"? (as defined in [code](https://github.com/woocommerce/google-listings-and-ads/pull/1499/files/3109ee4f7f8d5b0ba6d67f6ff8f1df54df1ab7ca#diff-19741a9dbe0895f99f4b04a6a88e21d3a8445917bea2f62ccf3f2610acff92d8R31))
Then we can call it:
`changeGroupWithinValue( value, oldGroup, newGroup )`?

@ecgan , @puntope WDYT?

tomalec:
Speaking of [100% coverage](https://github.com/woocommerce/google-listings-and-ads/pull/1499#pullrequestreview-974077665) and 5 `changeMinimumOrderGroup` [instead of three](https://github.com/woocommerce/google-listings-and-ads/pull/1499#discussion_r873912616) - this is related to the section at 
https://github.com/woocommerce/google-listings-and-ads/blob/feature/refactor-minimum-order-modal-tests/js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js#L38-L61

Which I'd like to reduce in favor of groups rendering below https://github.com/woocommerce/google-listings-and-ads/blob/feature/refactor-minimum-order-modal-tests/js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js#L83-L102

But I'd make it a separate PR as it was not change neither here nor in https://github.com/woocommerce/google-listings-and-ads/pull/1410

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
tomalec: > Instead of using inline arrow functions (and having a bit of duplicated code across the two `onChange` and `onDelete`), I would still prefer to have a dedicated "handlers" code block so that it is easier to see how those add/edit/delete operations is mapped to the `changeMinimumOrderGroup` call. In case those operations need to be modified, they just need to be modified in their corresponding handler code instead of the duplicated inline arrow function.

I'd address duplication as mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1499#issuecomment-1129049643 by removing the duplication in the rendering case when those handlers are called. Then we will end up with 3 different use cases of `changeMinimumOrderGroup` and 3 calls - no duplication whatsoever.

Then, changing:
```jsx
onDelete={ () =>
	  onChange(
		  changeMinimumOrderGroup( value, groups[ 0 ] )
	  )
}
```
to 
```jsx
const getDeleteHandler = ( oldGroup ) => () => {
	const newValue = changeMinimumOrderGroup( value, oldGroup );
	onChange( newValue );
};
// ~ 40 lines of unrelated code
onDelete={ getDeleteHandler( group ) }
```
Does not bring many benefits, as 
- I need to "jump around" to check what `onDelete` actually does, the code is more dislocated and less explicit
- the amount of lines is exactly the same


The one reason that makes me a bit more convinced about that is because we had it that way before #1410, so making it as it was would minimize the changes in this PR(s). Then we can address that together with removing the duplication of renderings.

js/src/components/shipping-rate-section/minimum-order-card/changeMinimumOrderGroup.js
ecgan: > However, I also don't like `getShippingRateNewValue`: [...]

Interesting. For me, I usually just use `get*` as a short and simple way to name a function that returns a value, and it is fine to have args. Alternative naming can be `calculate*`, `calc*`, `generate*`. I also tend to think of these `get*` functions as pure functions that don't mutate the args and can be easily tested with unit tests.

> Personally, I like when the function name is optimized for a _consumer's perspective_, not implementer's perspective. [...]

Yes, agree. I just want to be more specific and say it is for the function code consumer perspective (not the UI end-user consumer perspective; for some reasons I felt the naming was more towards UI end-user perspective, probably because of the word "change" that relates to the users changing the minimum order group in the UI).

---

> Maybe we can put it that way: [...]

Yes, totally agree with those points. 😄 👍 

> * Then we can call it:
>   `changeGroupWithinValue( value, oldGroup, newGroup )`?

I still don't quite like the "changeGroup" naming. From the code consumer perspective, I feel it is like it is going to mutate `oldGroup` or `newGroup` and not going to return a value. 😅 

Anyway, I think if we can't reach an agreement here, we can move on with what we have, either `changeMinimumOrderGroup` or `changeGroupWithinValue` is fine and it doesn't make a big difference.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
ecgan: > I'd address duplication as mentioned in [#1499 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/1499#issuecomment-1129049643) by removing the duplication in the rendering case when those handlers are called. Then we will end up with 3 different use cases of `changeMinimumOrderGroup` and 3 calls - no duplication whatsoever.

I'm not sure how you would like to address the duplication. I'm guessing, maybe you are considering to do some mutations on the `groups` array, so that there is only one `return` in the `renderGroups` function? 🤔 That was the approach we previously had for shipping rates groups (not minimum order groups).

If that is the idea, I actually intentionally want to avoid that, and it is because the `if ( groups.length === 1 )` is a _special case_ where we want to show the `MinimumOrderInputControl` even when there is no MinimumOrderGroup with defined threshold, and we want to return early for such special case. With this, `groups` array is not mutated, and its meaning is always the same, as in "`groups` is the result from calling `groupShippingRatesByMethodFreeShippingThreshold` function". If we were to mutate the `groups` array and use one `return` statement, then the `groups` array meaning would change and wouldn't be as straightforward, it makes it a little harder to follow and understand, and the one `return` statement wouldn't clearly show we have a special rendering for the special case.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
tomalec: In https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/e47c9faa68726d55fb06824588428bb66cffbfba I went with `structuredClone`, to be even less fragile and to emphasize what actually needs to be changed/checked:

```js
// Expect US threshold to be set to 30.
const expectedValue = structuredClone( value );
expectedValue[ 0 ].options.free_shipping_threshold = 30;
expect( onChange ).toHaveBeenCalledWith( expectedValue );
```

Unfortunately, we run an old Node version, so I added a very naive shim, to make it work until we update.

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.js
tomalec: > I'm not sure how you would like to address the duplication. I'm guessing, maybe you are considering to do some mutations on the groups array, so that there is only one return in the renderGroups function? thinking That was the approach we previously had for shipping rates groups (not minimum order groups).

That's something to explore and discuss in a separate PR :)

tomalec: I defined dedicated handlers in https://github.com/woocommerce/google-listings-and-ads/pull/1499/commits/8c31c29840314c02b282212c0711b6f7e0e02193

tomalec:
Also, speaking of code coverage. We do not test whets rendered from `MinimumOrderCard` at all. So adding a test to cover just one conditional rendering path, could leave a false impression, that all rendering paths are covered.

We could add tests for rendered results in a separate PR, not to enlarge those two :)

js/src/components/shipping-rate-section/minimum-order-card/changeMinimumOrderGroup.js
tomalec: >  usually just use `get*` as a short and simple way to name a function that returns a value, and it is fine to have args. Alternative naming can be `calculate*`, `calc*`, `generate*`

Personally, I like when most functions return some value ;) and to me `get*` is more connected to "obtaining" something - like in `localStorage.getItem`, than just returning something. Returning something is a natural characteristic of a function. And here we are not obtaining/accessing/fetching anything.
We are doing state management translation. Having the original state in "local" API and two states in "remote"/"child" API, we rather translate that change and apply it to a given value. So, we either `applyGroupChange`, or  `calculate*` :smiley: the new state out of other states.

> it is for the function code consumer perspective (not the UI end-user consumer perspective;

Good point, to focus more on code customer perspective rather than end-user using UI. (However, I think the latter also has some value. For example new engineers, not familiar with the codebase, who knows what the end product looks like and what is for, but don't know the code yet. They may not know how the state is calculated and what are data types, but they know that the component is meant to change the minimum order for a given group of countries).

Also, as mentioned above, from the code perspective there is also a change. The initial state of the group changed to a new one. Then, this function translates the state change of `MinimumOrderGroup` to the state change in `Array<ShippingRates>`.
But taking the perspective and _"I feel it is like it is going to mutate […]"_ points, I now lean more towards `apply*` or `calculate*`. As for our use case, we do not need to change or mutate anything, we need to return an object with the new state.

WDYT of `applyGroupChange`? 
`calculateValue` does not give much information from what data/in what case/when, so I'd use `calcuateValueFromGroupChange` - which is lengthy

----

But...
As we control both parties: `MinimumOrderCard` and `MinimumOrderInputControl` instead of trying to find a perfect name for a function that translates one API/data type to another, we could make both components use the same API/data structure. Then the problem of naming that function, implementing it, and testing will be simply gone. :smile:

js/src/components/shipping-rate-section/minimum-order-card/changeGroup.js
tomalec: Personally, I find `export const foo = …` a nice syntax, as it gives you the instant info that the thing being defined will be exported.

I wouldn't employ `const foo = … export foo;` convention just because it's a workaround for jsdoc generator bug. I'd use a workaround where needed (where we track events), and try to fix the bug at its source.

tomalec:
> Lastly, regarding the merge, which one would you prefer?
> 
> Option 1: Merge this PR https://github.com/woocommerce/google-listings-and-ads/pull/1499 into PR https://github.com/woocommerce/google-listings-and-ads/pull/1410 first, and then merge https://github.com/woocommerce/google-listings-and-ads/pull/1410 into `develop`.
> Option 2: Merge PR https://github.com/woocommerce/google-listings-and-ads/pull/1410 first into `develop`, and then merge this PR https://github.com/woocommerce/google-listings-and-ads/pull/1499 into `develop`.

Initially, I was thinking of 1. As this PR was meant to simplify #1410. Merging this there would make the `feature/refactor-minimum-order-modal` larger in commits, but speaking of files and changes it will be less
- https://github.com/woocommerce/google-listings-and-ads/compare/develop...feature/refactor-minimum-order-modal
- https://github.com/woocommerce/google-listings-and-ads/compare/develop...feature/refactor-minimum-order-modal-tests

`(1410 & 1499) < 1410 + 1499`

Probably, if I wouldn't add shared mocks for hooks and `structuredClone` shim, it would make it significantly smaller. `(1410 & 1499) < 1410`

Also merging PR to PR, would let us avoid adding 7 soon-to-be-removed files on `develop`.

We didn't add any changelog entries, so it doesn't make a difference for `woorelease`

-----

However, if we merge one soon after another, then the eventual git history would be almost the same anyways. So, I don't mind either way.

tomalec:
@ecgan I believe I addressed all your comments. I think I made some substantial changes, and I'll appreciate another round of review.

I also approved #1410, so feel free to choose whichever merge path you like :)

js/src/components/shipping-rate-section/minimum-order-card/changeMinimumOrderGroup.js
ecgan: > to me `get*` is more connected to "obtaining" something - like in `localStorage.getItem`, than just returning something.

Interesting. I guess you mean `get*` is to get something that already exists somewhere (e.g. localstorage, API, database etc, the things are there, instead of generated / calculated / created in the function itself). Thanks, good to know, and I think / hope I can get use to that 🙂 👍 

> `calculateValue` does not give much information from what data/in what case/when, so I'd use `calculateValueFromGroupChange` - which is lengthy

I'd say both of them are better than `changeMinimumOrderGroup` 😄 👍 Personally I would go for `calculateValue`, because by looking at the function parameters with `oldGroup` and `newGroup`, I would already know it is associated with "group change", so having "*FromGroupChange" doesn't help much for me. But honestly I don't mind `calculateValueFromGroupChange` either, it is not the longest function name in the code base. 😆 We can go with `calculateValueFromGroupChange`.

---

> As we control both parties: `MinimumOrderCard` and `MinimumOrderInputControl` instead of trying to find a perfect name for a function that translates one API/data type to another, we could make both components use the same API/data structure. Then the problem of naming that function, implementing it, and testing will be simply gone. 😄

Hmm, I don't quite get this. I mean, no matter where in the code base, we still need to do that data transformation somewhere, to transform / translate shipping rate data to/from minimum order group data. 🤔 

Maybe a separate PR one day, love to see your idea. 😄

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
ecgan: > I added a very naive shim

I looked into the implementation (https://github.com/woocommerce/google-listings-and-ads/pull/1499/files#diff-c64e291bfa29797411cbe013a64cc82a4a294cb297c5d546ebb5ca69ff53ffa9R1-R10) and at first I was like 🤯 , and then I realized your mention of "naive" and I'm like okay 😂 Thanks for the code comment there. 

And TIL about [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone). Thank you. 🙂 👍

ecgan (approved):
Tested in the UI, working as expected. Unit tests are running good. All comments are addressed. LGTM 👍 🚀 

I have just merged PR #1410, so I'll merge this PR right after.

js/src/components/shipping-rate-section/minimum-order-card/changeMinimumOrderGroup.js
ecgan: I forgot about the discussion we have here and merged this PR after I merged PR #1410. I have created PR https://github.com/woocommerce/google-listings-and-ads/pull/1513 to continue the function naming change and discussion. 🙏

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
ecgan: @tomalec , just FYI, apparently the behavior is different in `develop` branch and it causes tests to fail, the test code here has been changed and fixed in PR https://github.com/woocommerce/google-listings-and-ads/pull/1512.

tomalec: :exploding_head: Thanks!

js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.test.js
tomalec: I decided to keep it super simple. So instead of putting effort to implement, review, tests, maintaining fully-featured polyfill, we can update Node sooner ;)

PR #1500, opened 2022-05-12
Fix the infinite reloading problem for React Fast Refresh

eason9487:

Use file time as the version string of runtime.js script to prevent the React Fast Refresh from infinite reloading.

When developing without checking the "Disable cache" option in the Network tab of DevTool, the infinite reloading problem happens randomly after changing JS code or checking out branches.

The root cause is the runtime.js file is changed but may still be loaded from the local cache, and the React Fast Refresh detects the old version, then triggers the page reloading. However, the runtime.js still be loaded from the local cache, and then the infinite loop occurs.

I didn't find a reliable way to reproduce it but fixed it by this way when I happened to encounter it.

### DIFF
--- a/src/Admin/Admin.php
+++ b/src/Admin/Admin.php
@@ -254,7 +254,9 @@ class Admin implements Service, Registerable, Conditional {
 	 * @param WP_Scripts $scripts WP_Scripts instance.
 	 */
 	private function inject_fast_refresh_for_dev( $scripts ) {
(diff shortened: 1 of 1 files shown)


ianlin (approved):
Thanks for fixing this. As you said I couldn't replicate the issue but the code was pretty straightforward, just adding a version string to the WP_Scrips's add method so LGTM.

PR #1503, opened 2022-05-13
Require URL to match when syncing data to Merchant Center

### LINKED ISSUE #1399: Staging site can override items in the catalog
### Describe the bug:
If the site is copied to a staging site, all the connection settings are also copied over. Unless this is done during the initial sync, it generally won't start automatically syncing all the products as soon as the staging site is created. However since the staging site is also considered as "connected" there are plenty of events that could trigger products to be synced:

- Products have been submitted more than 25 days ago (could trigger a scheduled sync on both the live site and staging site)
- Products are updated on the staging site
- A product update event is triggered by another plugin while changing some of the details

There are plenty of other events that could trigger a product to be removed from the catalog, such as an item goes out of stock, or a staging product becomes a draft or is trashed. This wouldn't trigger staging URL's to be submitted but this could still cause unexpected behaviour of products being removed from the catalog.

Since we use a specific `site_url` for verification (not stored locally), we should be able to add an additional check to only sync when the `site_url` is still the same as the one we used for verification.

### Steps to reproduce:
1. Start with a live site that's fully onboarded and synced all products to the Merchant Center
2. Create a staging site (can use a plugin like [WP Staging](https://wordpress.org/plugins/wp-staging/))
3. Update a product on the staging site
4. Check the action scheduler logs and confirm that the product successfully synced
5. Check the same product in the Merchant Center (might need to wait a while before it's updated) and confirm that the URL's are using that of the staging site

### Expected behaviour:
All syncing to the Merchant Center should only happen if the site URL is the same as the one we originally verified

### Additional details:
Initially reported in #1276

We currently don't save the URL we used for site verification in the onboarding steps. It can be retrieved from the Content API, but it might be better to store it in the local database. We would need to prevent all data that is synced to the content API from happening if the site URL does not match.

mikkamp:

This PR changes the product sync behaviour and requires the URL to match to the one we used to claim the website with. This ensures that the products are only able to sync on the original site, any staging site or copy of the site will not sync product data.

If the user wishes to change the site URL then the onboarding process must be completed again to perform a new site verification/claim.

Closes #1399

1. Start with a live site that's fully onboarded and synced all products to the Merchant Center
2. Create a staging site (can use a plugin like [WP Staging](https://wordpress.org/plugins/wp-staging/))
3. Update a product on the staging site
4. Check the action scheduler logs and confirm that the product was not synced
5. Check the same product in the Merchant Center (might need to wait a while before it's updated) and confirm that the URL's are still from the original site
6. Use the filter `add_filter( 'woocommerce_gla_ready_for_syncing', '__return_true' );`
7. Update a product on the staging site
8. Check the action scheduler logs and confirm that this time the product did sync
9. Check the original site and confirm that product syncing continues to function as it was when updating products

### Additional details:

In addition to the tests for the added / changed code. There was no unit test present for the MerchantCenterService class, so I added it in this PR before adding the additional tests for this class.

### Changelog entry
* Fix - Prevent product sync if the site URL does not match the originally claimed URL.

### DIFF
--- a/src/API/Google/Merchant.php
+++ b/src/API/Google/Merchant.php
@@ -194,6 +194,34 @@ class Merchant implements OptionsAwareInterface {
 	}
 
 	/**
+	 * Get hash of the site URL we used during onboarding.
+	 * If not available in a local option, it's fetched from the Merchant Center account.
+	 *
+	 * @since x.x.x
+	 * @return string|null
+	 */
+	public function get_claimed_url_hash(): ?string {
+		$claimed_url_hash = $this->options->get( OptionsInterface::CLAIMED_URL_HASH );
+
+		if ( empty( $hashclaimed_url_ ) && $this->options->get_merchant_id() ) {
+			try {
+				$account_url = $this->get_account()->getWebsiteUrl();
+
+				if ( empty( $account_url ) || ! $this->get_accountstatus()->getWebsiteClaimed() ) {
+					return null;
+				}
+
+				$claimed_url_hash = md5( $account_url );
+				$this->options->update( OptionsInterface::CLAIMED_URL_HASH, $claimed_url_hash );
+			} catch ( Exception $e ) {
+				return null;
+			}
+		}
+
+		return $claimed_url_hash;
+	}
+
+	/**
 	 * Retrieve the user's Merchant Center account information.
 	 *
 	 * @param int $id Optional - the Merchant Center account to retrieve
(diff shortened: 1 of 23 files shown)


ianlin (approved):
Thanks for fixing this! I could reproduce this issue using `develop` branch, and confirmed the issue is solved using the branch of this PR. The products will only be synced to merchant centre when using the production site.

Just a note while I was testing for it:

Normally I used Ngrok to get a public domain (e.g. `ian-local-test-wp.ngrok.io`) while onboarding setup and using this public domain to claim the url. After that I used a local domain (e.g. `wp59wc61.test`) points to 127.0.0.1 for other operations since it's faster. If I use the local domain `wp59wc61.test` to update the product it will not sync to merchant centre either.

mikkamp:
Thanks for the review.

> Just a note while I was testing for it...
I have the same setup, which is why I added the filter `woocommerce_gla_ready_for_syncing` to override it for local testing. I've also added the filter to our [wiki](https://github.com/woocommerce/google-listings-and-ads/wiki/Testing)

PR #1504, opened 2022-05-16
Simplify the implementation of focus navigation for TreeSelectControl

eason9487:

- Replace the implementation of focus navigation with the `focus` of `@wordpress/dom` for `TreeSelectControl`.
- Fix a "read properties of null" bug that press the Up or Down key when the dropdown is not showing.

   https://user-images.githubusercontent.com/17420811/168525767-b5863d11-304b-4969-bef6-00fbb56eaf28.mp4

### Screenshots:

https://user-images.githubusercontent.com/17420811/168525734-5c20c0d1-c3aa-4229-9ee1-296bf6d914cd.mp4

1. Go to the Edit free listings or the local storybook.
2. Click on the input to open the dropdown.
3. Press Esc key to close the dropdown.
4. Press Up or Down key when the input still has focus. It should not cause the "read properties of null" exception.
5. Open the dropdown and use Up or Down to navigate the options to check if the behavior is the same as before.
6. Click on expanders of parent options, and then the corresponding checkbox should get the focus.

### Changelog entry

> Tweak - Adjust the implementation of focus navigation for the TreeSelectControl component.
> Fix - Address a crash problem of TreeSelectControl component when the dropdown is not showing and press the Up or Down key.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,6 +5,7 @@ const defaultConfig = require( '@wordpress/scripts/config/jest-unit.config' );
 module.exports = {
 	...defaultConfig,
 	testEnvironment: 'jsdom',
+	setupFiles: [ 'core-js' ],
 	moduleNameMapper: {
 		'\\.svg$': '<rootDir>/tests/mocks/assets/svgrMock.js',
 		'\\.scss$': '<rootDir>/tests/mocks/assets/styleMock.js',
--- a/js/src/components/tree-select-control/checkbox.js
+++ b/js/src/components/tree-select-control/checkbox.js
@@ -4,20 +4,19 @@
 import { Icon, check } from '@wordpress/icons';
 
 /**
- * @typedef {import('./index').RepositoryOption} RepositoryOption
+ * @typedef {import('./index').Option} Option
  */
 
 /**
  * Renders a custom Checkbox
  *
  * @param {Object} props Component properties
- * @param {RepositoryOption} props.option Option for the checkbox
- * @param {number} props.index The position in the tree
+ * @param {Option} props.option Option for the checkbox
  * @param {string} props.className The className for the component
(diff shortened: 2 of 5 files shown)


puntope (commented):
Howdy @eason9487 , I checked this PR and looks great, it really simplifies the focus logic! Well done.

One thing I noticed is that the specification was that with TAB we change the focus to the next focusable element outside the tree. But now with TAB we navigate over the inner tree.

eason9487:

> One thing I noticed is that the specification was that with TAB we change the focus to the next focusable element outside the tree. But now with TAB we navigate over the inner tree.

Yes, this is different from before and I forgot to mention it. 🤦‍♂️ 
While the dropdown is showing and the input has the focus, I think it should not be a problem to move focus to a checkbox in the dropdown. Would it be more in line with what is expected when pressing Tab key?

puntope:
> Yes, this is different from before and I forgot to mention it. 🤦‍♂️
While the dropdown is showing and the input has the focus, I think it should not be a problem to move focus to a checkbox in the dropdown. Would it be more in line with what is expected when pressing Tab key?

The problem I see is that when the Tree Select receives focus it **auto-opens** the tree, and you need to navigate over all the tree elements before going to the next input.

I think we should stuck into the old requirement of jumping to the next element instead of navigate in the inner. Or maybe disable that `auto-open` feature on focus.

eason9487:
@puntope 

> I think we should stuck into the old requirement of jumping to the next element instead of navigate in the inner.

Thanks for reminding this! I completely forgot this's a part of the requirements. 🙏  I adjusted it in https://github.com/woocommerce/google-listings-and-ads/pull/1504/commits/2c266ba1d1a195ac3bbf431008fcab81370af674 to fit with the original behavior.

puntope (approved):
✅  LGTM

Thanks for the adjustments

PR #1505, opened 2022-05-16
Includes removed campaigns in Program Reports

jorgemd24:

As part of the migration process of SSC campaigns to PMax campaigns, users need to be able to compare removed SSC campaigns with new campaigns (PMax). This PR includes removed campaigns in the programs reports filter. 

See more context in the following links:

- [Design Proposal i2: Performance Max Migration (Milestone 2)](https://woodesigngrowconsultants.wordpress.com/2022/04/20/design-proposal-i2-performance-max-migration-milestone-2/#in-reports)

Also, I added two small corrections to CSS linting.

### Screenshots:
image

1.  Go to GLA->Reports->Programs
2.  Show -> Single Program

image

3. Search for a removed campaign or go to your dashboard to delete one campaign and check if the filter shows it.

image

### Additional details:

The suffix "(old)" is added if:

- the campaign is an SSC campaign.
- it has been removed.
- The user has been migrated (all SSC campaigns are removed and only PMax campaigns are available). This can be set manually if you add a new option: `gla_campaign_convert_status ` with the following value: `a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1652687806;}`

### Changelog entry

>Add - includes removed campaign in the program report section


js/src/hooks/useAdsCampaigns.js
```diff
@@ -24,34 +25,45 @@ const selectorName = 'getAdsCampaigns';
  * A hook that calls `getAdsCampaigns` selector to load current campaigns
  * from merchant's Google Ads account if connected.
  *
+ * @param {...*} args Arguments to be forwarded to the selector. The only allowed argument is {exclude_removed: true|false}
```
puntope: 💅  This could be clearer using the explicit available params as well as changing from 'args' to 'query' since `query` it's the name of the param that `getAdsCampaigns` will receive as an argument. In case the available query params grows in the future we could consider also creating a typeDef kinda `getAdsCampaignsQuery` with the available params.

```
@param {{exclude_removed: true|false}} query Query to be forwarded to the selector.
const useAdsCampaigns = ( ...query ) => {
```

js/src/data/resolvers.js
```diff
@@ -148,10 +148,37 @@ export function* getTargetAudience() {
...
 }
 
+getAdsCampaigns.shouldInvalidate = ( action, query ) => {
+	return (
+		( action.type === TYPES.UPDATE_ADS_CAMPAIGN ||
+			action.type === TYPES.DELETE_ADS_CAMPAIGN ||
+			action.type === TYPES.CREATE_ADS_CAMPAIGN ) &&
+		query?.exclude_removed === false
+	);
+};
```
puntope, line 180: ❓  What is this for?

js/src/data/test/reducer.test.js
```diff
@@ -38,6 +38,7 @@ describe( 'reducer', () => {
 				contact: null,
 			},
 			ads_campaigns: null,
+			all_ads_campaigns: null,
```
puntope, line 41: 💅 / ❓  

Not sure if this is temporary until finishing the migration but if this is going to be there for a long time maybe we could consider to group the `all` and `available` campaigns  inside `ad_campaigns` instead of creating a separated state.

puntope (commented):
Howdy @jorgemd24 

I reviewed the PR and looks good.

I left a comment 💅  regarding JSDOCS and also I have one question that needs to be clarified before approval:

In the PR description you said:

> Additional details:
The suffix "(old)" is added if:
the campaign is an SSC campaign.
it has been removed.
The user has been migrated. This can be set manually if you add a new option: gla_campaign_convert_status with the following value: a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1652687806;}

So I was testing this and I don't see the (old) suffix. 


https://user-images.githubusercontent.com/5908855/168769940-38f65309-59f2-4529-8b8c-cc76bfb27086.mov

js/src/data/test/reducer.test.js
jorgemd24: Yeah, we could consider that option in the future, we will need to see which parts of the code would need to be updated and what other side effects could have. I can have a look into that in a different PR.

js/src/data/resolvers.js
jorgemd24: In case the ads campaign is updated, deleted or created, we will re-fetch the data for `all_ads_campaigns`. For example, if you ad a new Ads campaign so it will show the new ads campaign in the filter section.

js/src/hooks/useAdsCampaigns.js
jorgemd24: Thanks for the suggestion, I added here: [f7620a9](https://github.com/woocommerce/google-listings-and-ads/pull/1505/commits/f7620a97f8216ce2a0db67dd972bbe75a3a0f79a)

jorgemd24:
Thanks, @puntope for reviewing this! 

>So I was testing this and I don't see the (old) suffix. 

I had to mention that the old suffix is added if the user has the status 'converted', this happens when all Old SSC campaigns are removed. If you see your options in wp_options and look for `gla_campaign_convert_status` there you can see what is your current status. You can update this option manually to converted using this value: `a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1652687806;}`


Let me know if you have any other questions! 😃

puntope (approved):
✅  Approved 

Looks good for me, thanks @jorgemd24 for the changes.

💅  I see now that `argsRefValue` makes reference to the old 'args' naming. So likely it should be `queryRefValue` but approving it in advance.

PR #1507, opened 2022-05-17
MC Account Issues - Populate action with Issue details

puntope:

Design team and Google requested to show the Issue details (if any) in a modal instead of directly link to their documentation. The idea is to replace `action` column content in `wp_gla_merchant_issues` with the issue details provided by Google. Since `action` column was always hardcoded as "Read more about this account issue"  we can replace that with the details with no problem. This means `action` column will be the string with the details field or null.

In the frontend, in case the `issue.action` is null, we will show "Read more about this account issue" linking to the docs. (Same behaviour as now). Otherwise, the link will open a modal with the issue description as well as a "read more" link to the documentation. 

This PR implements the backend part with the following changes:

- Populate `action` column in `wp_gla_merchant_issues` with the Issue details, or null. (This depends on Google response)
- Added some Test for Merchant Statuses relative to the account issues.

1.  Create an account with Account Issues. In case you have already account issues you can just delete the transient `gla_mc_statuses` in `wp_options` to force refresh.
2.  Go to and check that action column is null (no value available from google) or its filled with the issue details (I only was able to localise one issue with details `editorial_and_professional_standards_destination_url_down_policy`).

### DIFF
--- a/tests/Unit/MerchantCenter/MerchantStatusesTest.php
+++ b/tests/Unit/MerchantCenter/MerchantStatusesTest.php
@@ -29,7 +29,7 @@ class MerchantStatusesTest extends UnitTest
 	private $merchant_center_service;
 	private $account_status;
 	private $product_meta_query_helper;
-	private MerchantStatuses $merchant_statuses;
+	private $merchant_statuses;
 
 	/**
 	 * Runs before each test is executed.
--- a/tests/Unit/MerchantCenter/MerchantStatusesTest.php
+++ b/tests/Unit/MerchantCenter/MerchantStatusesTest.php
@@ -1,0 +1,142 @@
+<?php
+declare(strict_types=1);
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Merchant;
+use Automattic\WooCommerce\GoogleListingsAndAds\DB\ProductMetaQueryHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\MerchantIssueQuery;
+use Automattic\WooCommerce\GoogleListingsAndAds\DB\Table\MerchantIssueTable;
+use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
+use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantStatuses;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductRepository;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\League\Container\Container;
+use Google\Service\ShoppingContent;
+
+defined('ABSPATH') || exit;
+
+/**
+ * Test Suit for MerchantStatuses
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\MerchantCenter
+ * @group MerchantCenterStatuses
+ */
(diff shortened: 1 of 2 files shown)


tests/Unit/MerchantCenter/MerchantStatusesTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+defined('ABSPATH') || exit;
+
+/**
+ * Test Suit for MerchantStatuses
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\MerchantCenter
+ * @group MerchantCenterStatuses
+ */
+class MerchantStatusesTest extends UnitTest
+{
```
ianlin: 💅  There is a newline before the `{`. It would be great to remove the newline so it's more consistent with the current codebase.

tests/Unit/MerchantCenter/MerchantStatusesTest.php
```diff
@@ -0,0 +1,142 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\League\Container\Container;
+use Google\Service\ShoppingContent;
+
+defined('ABSPATH') || exit;
+
+/**
+ * Test Suit for MerchantStatuses
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\MerchantCenter
+ * @group MerchantCenterStatuses
+ */
```
ianlin: 💅  Missing properties in doc. For example: https://github.com/woocommerce/google-listings-and-ads/blob/e17daaf1526493eeac1febce6b8d60f57b9ffe0f/tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php#L18-L20

tests/Unit/MerchantCenter/MerchantStatusesTest.php
```diff
@@ -0,0 +1,142 @@
...
+	private $merchant_center_service;
+	private $account_status;
+	private $product_meta_query_helper;
+	private $merchant_statuses;
+
+	/**
+	 * Runs before each test is executed.
+	 */
+	public function setUp(): void
+	{
```
ianlin: 💅  Same as above, It would be great to remove the newline before `{`.

tests/Unit/MerchantCenter/MerchantStatusesTest.php
```diff
@@ -0,0 +1,142 @@
...
+		$container->share(ProductRepository::class, $product_repository);
+		$container->share(ProductMetaQueryHelper::class, $this->product_meta_query_helper);
+		$container->share(MerchantIssueTable::class, $merchant_issue_table);
+
+		$this->merchant_statuses = new MerchantStatuses();
+		$this->merchant_statuses->set_container($container);
+	}
+
+	public function test_refresh_account_issues()
+	{
```
ianlin: 💅  Same as above, It would be great to remove the newline before `{`.

ianlin (approved):
Thanks for adding this. Tested locally and I can see some of the value in `Suggested action` are null. LGTM 👍 

Screenshot 2022-05-18 at 16 05 33

Just left some non-blocking comments. And a question for you that where can I find this issue or anywhere I can find the request from the design and the Google team? 🤔

puntope:
> Just left some non-blocking comments. And a question for you that where can I find this issue or anywhere I can find the request from the design and the Google team? 🤔

Thanks for the review @ianlin.  Since this is the Backend part only, we will create the modal as well as the fallback link in frontend. This was just to populate the issue details in the action field in database.

ianlin:
> Since this is the Backend part only, we will create the modal as well as the fallback link in frontend. This was just to populate the issue details in the action field in database.

Hey @puntope thanks for the clarification and sorry for not being clearly on my question. What I'd like to know is is there any project thread or discussion regarding this request? As in your PR description you mentioned:

> Design team and Google requested to show the Issue details (if any) in a modal instead of directly link to their documentation.

But I couldn't find any related request or discussion regarding this while I was reviewing. But now I found it in your another PR #1511 and now I know it's in this project thread: pcTzPl-ps-p2. Thanks!

PR #1508, opened 2022-05-17
Inbox notifications for PMax migrations

jorgemd24:

As part of the migration process of SSC campaigns to PMax campaigns, users need to be notified about the migration through the inbox notifications. This PR adds the notifications for before and after migration. 

See more context in the following links:

- [Design Proposal i2: Performance Max Migration (Milestone 2)](https://woodesigngrowconsultants.wordpress.com/2022/04/20/design-proposal-i2-performance-max-migration-milestone-2/#in-reports)

The original design for the notification **before** the migration, included a top banner however after some investigation, I found that WC Admin is no longer supporting this option so I will ask the design team for a change in the design.

image

#### Before Migration

1. Add the following option in `wp_options`: `gla_campaign_convert_status`  with the following value: `a:2:{s:6:"status";s:11:"unconverted";s:7:"updated";i:1652800816;}`
2. Go to tools->Scheduled Actions search for a pending job named: `wc_gla_cron_daily_notes `
3. Run the job:

image

4. Go to the Inbox notification and the following notification it should display:

image

#### After Migration
1. Update the option `gla_campaign_convert_status` to `a:2:{s:6:"status";s:9:"converted";s:7:"updated";i:1652800816;}`
2. Go to tools->Scheduled Actions search for a pending job named: `wc_gla_cron_daily_notes `
3. Run the job
4. Go to the Inbox notification and the following notification it should display:

image

### Additional details:
The links for learn/read more about this upgrade are pending to be confirmed. 

### Changelog entry

>Add - Inbox notification for PMax migration


src/Notes/AfterCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @return string
+	 */
+	public function get_name(): string {
+		return 'gla-after-campaign-migration';
+	}
+
+	/**
+	 * Get the note entry.
+	 */
+	public function get_entry(): NoteEntry {
```
puntope: 💅  Missed return type

src/Notes/AfterCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+
+		return $note;
+	}
+
+
+	/**
+	 * Checks if a note can and should be added.
+	 *
+	 * - checks if the campaign migration is completed
+	 *
```
puntope: 💅  To consider removing the extra spaces as well as the one element list.

src/Notes/AfterCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @return bool
+	 */
+	public function should_be_added(): bool {
+		if ( $this->has_been_added() ) {
+			return false;
+		}
+
+		return $this->ads_service->is_migration_completed();
```
puntope: 💅  To consider one liner option

`return ! $this->has_been_added() && $this->ads_service->is_migration_completed();`

src/Notes/BeforeCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	public function get_name(): string {
+		return 'gla-before-campaign-migration';
+	}
+
+	/**
+	 * Get the note entry.
+	 */
+	public function get_entry(): NoteEntry {
+		$note = new NoteEntry();
```
puntope, line 43: 💅  Missed return type

src/Notes/BeforeCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @return bool
+	 */
+	public function should_be_added(): bool {
+		if ( $this->has_been_added() ) {
+			return false;
+		}
+
+		return ! $this->ads_service->is_migration_completed();
+
+	}
```
puntope: 💅  Same as the other note

Unique element list and maybe one liner return.

tests/Unit/Ads/AdsServiceTest.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Class AdsServiceTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Ads
+ *
+ * @property MockObject|AdsAccountState     $state
+ * @property AdsService                     $ads_service
+ * @property OptionsInterface               $options
+ */
```
puntope, line 21: Thanks for adding tests ❤️ 

💅  One thing I normally do and is useful for me is to add

`@group AdsService`

That way we could run the tests on this file (without writing the entire path) and other files in same group by using

`../phpunit --group AdsService`

puntope: 💅 For A++ would be cool to have test as well for the 2 created notes.

src/Ads/AdsService.php
```diff
@@ -63,4 +63,15 @@ public function is_connected(): bool {
...
 
+	/**
+	 * Get whether Ads Campaign are migrated.
+	 *
+	 * @since x.x.x
+	 * @return bool
+	 */
+	public function is_migration_completed(): bool {
+		$convert_status = $this->options->get( OptionsInterface::CAMPAIGN_CONVERT_STATUS );
+		return is_array( $convert_status ) && isset( $convert_status['status'] ) && $convert_status['status'] === 'converted';
```
puntope: 💅 

`is_array( $convert_status )` is unnecessary since its covered by `isset( $convert_status['status'] )`

src/Notes/AfterCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class AfterCampaignMigration
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notes
+ */
+class AfterCampaignMigration extends AbstractNote implements AdsAwareInterface {
```
puntope: 💅  IMHO (and I'm aware they are plenty of these in the codebase) comment a class with `// class [className]` is redundant since we already see it in the PHP syntax. I think it would be more helpful to explain when this Notice is appearing in the Inbox and why giving context for future devs.

src/Notes/BeforeCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Class BeforeCampaignMigration
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notes
+ */
+class BeforeCampaignMigration extends AbstractNote implements AdsAwareInterface {
+
+	use AdsAwareTrait;
+	use PluginHelper;
+	use Utilities;
```
puntope: 💅 / ❓  

Are we really using `Utilities`? Or it is only for loading `OptionsAwareTrait`? If so, I guess we can just add `OptionsAwareTrait` instead of `Utilities`.

src/Notes/BeforeCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+		$note = new NoteEntry();
+
+		$note->set_title( __( 'Your Google Listings & Ads campaigns will soon be automatically upgraded', 'google-listings-and-ads' ) );
+		$note->set_content(
+			__(
+				'From July through September, Google will be upgrading your existing campaigns from Smart Shopping to Performance Max, giving you the same benefits, plus expanded reach. There will be no impact to your spend or campaign settings due to this upgrade.',
+				'google-listings-and-ads'
+			)
+		);
+		$note->set_content_data( (object) [] );
```
puntope: ❓ Just curious. What is this for?

src/Notes/BeforeCampaignMigration.php
```diff
@@ -0,0 +1,79 @@
...
+		$note->set_type( NoteEntry::E_WC_ADMIN_NOTE_INFORMATIONAL );
+		$note->set_layout( 'plain' );
+		$note->set_image( '' );
```
puntope: 💅 / ❓ 

Checking the default constructor for `Note.php`. I guess this 3 are not necessary??

puntope (commented):
howdy @jorgemd24. Thanks for the PR, looking good. It works as expected. I tested different convos of converted, unconverted, dismissing the notices etc. Thanks also for adding tests.

I left a few comments and questions on it.

src/Notes/BeforeCampaignMigration.php
jorgemd24: `set_content_data` is used to add extra information to the Inbox Notification, in this case, we do not need to add extra information so I was passing an empty object as per other Notes in our codebase, however, looking at the default values of Notes::class this field is not needed because the default value is an empty object.

See more here: 

> In addition to content, notes also support structured content.  You can use this property to re-localize notes on the fly, but that is just one use.  You can store other data here too.  This is backed by a longtext column in the database.

Reference: https://developer.woocommerce.com/2020/10/16/using-the-admin-notes-inbox-in-woocommerce/

src/Notes/BeforeCampaignMigration.php
jorgemd24: Thanks, @puntope, yeah the default constructor is already using those 3 values and `set_content_data` is not necessary either. I have updated in: [23d338d](https://github.com/woocommerce/google-listings-and-ads/pull/1508/commits/23d338d4d379c6d57e48dfb75a63c907203e6576)

src/Notes/BeforeCampaignMigration.php
jorgemd24: Thanks for catching this! I have updated in [09c5a4f](https://github.com/woocommerce/google-listings-and-ads/pull/1508/commits/09c5a4fd03729ca2960c1214cf4a25a325779d47)

tests/Unit/Ads/AdsServiceTest.php
jorgemd24: Thanks for sharing this! I did not know about @group! I have added: [b05eae9](https://github.com/woocommerce/google-listings-and-ads/pull/1508/commits/b05eae987d482f4dd20afb3b1f231dc3e055f6e5)

>would be cool to have test as well for the 2 created notes.

I have added the two tests here: [e645f2a](https://github.com/woocommerce/google-listings-and-ads/pull/1508/commits/e645f2a1dcc30ad5a248d26dcdb5cff1e8bfd6fa) 😄

jorgemd24:
Thanks, @puntope for your useful suggestions! I have updated the code and I left a comment about the tests and`set_content_data`. Let me know what you think

puntope (approved):
Thanks for the changes @jorgemd24 

This is ✅  approved! 

I tested again all the procedure and it works. It looks super clean and clear. Huge thanks for the tests. Good work

PR #1511, opened 2022-05-18
Account Issues -  Show the resolution steps in the UI 

puntope:

This PR implements the design proposal of having a modal with the issue information instead of a link to the documentation.

Design can be found here https://googleforwooextension.wordpress.com/2022/03/17/gmc-account-review-request-designs-i4/

(In the section 2)

It's related with #1507  (that PR implemented the backend)

### Screenshots:

https://user-images.githubusercontent.com/5908855/169072768-17a8652b-5d90-4eaa-af88-be574cd0cede.mov

1.  You need to have an account with 2 issues (doesn't matter for now if they are product or account), also you will need to check track events, so please allow tracking and run `localStorage.setItem( 'debug', 'wc-admin:*' );` in verbose mode in Dev Tools Console.  
2. One issue should have an action as null in database, other with action filled with some content. (Normally product issues has content in action and account issues not) But you can create or update at your own in `wp_gla_merchant_issues` 
3. Go to Product Feed, you will see the `Read ore about this issue` link on both.
4. If you click in the one with null value, it will go to the link provided in action_url column. opening the link in a new tab and triggering `gla_documentation_link_click` event
5. If you click in the one with value, it will open a modal with the issue info. Issue title in the modal header as well as the issue description in the body. Also, the event `gla_click_read_more_about_issue` will be triggered. 
6.  Clicking on Learn more will open the link in a new tab and triggering `gla_documentation_link_click` event

### Additional details:

⚠️ I need to confirm wit Design if they want this functionality with Products and Account or only account issues. But that could be adjusted if so in another PR. 

- Some refactors in the issue table header, since we needed in the testing files I decided to separate it in a new file. 
- Created some tests in Data table and Modal
- Created a mock function inside test. If you have a better suggestion please let me know since I'm not sure where to store it.

### Changelog entry

> Tweak - Show the account issue resolution steps in the UI

### DIFF
--- a/js/src/product-feed/issues-table-card/issues-table-data-modal.js
+++ b/js/src/product-feed/issues-table-card/issues-table-data-modal.js
@@ -1,0 +1,55 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import GridiconExternal from 'gridicons/dist/external';
+
+/**
+ * Internal dependencies
+ */
+import AppModal from '.~/components/app-modal';
+import AppButton from '.~/components/app-button';
+
+/**
+ * Renders a modal showing information about a Google Merchant Center issue
+ *
+ * @fires gla_documentation_link_click with { context: 'issues-data-table-modal' }
+ * @param {Object} params Component params
+ * @param {Object} params.issue The issue to be rendered in the modal
+ * @param {Function} params.onRequestClose Onnclose callback function
+ */
+const IssuesTableDataModal = ( { issue, onRequestClose = () => {} } ) => {
+	if ( ! issue ) return null;
+	return (
+		<AppModal
+			className="gla-issues-table-data-modal"
+			title={ issue.issue }
+			onRequestClose={ onRequestClose }
+			buttons={ [
+				<AppButton
+					key="learn-more"
+					isPrimary
+					target="_blank"
+					href={ issue.action_url }
+					text={ __( 'Learn more', 'google-listings-and-ads' ) }
+					eventName={ 'gla_documentation_link_click' }
+					eventProps={ {
+						context: 'issues-data-table-modal',
+						linkId: issue.code,
+						href: issue.action_url,
+					} }
+					icon={ <GridiconExternal /> }
+				/>,
+			] }
+		>
+			<p>
+				<strong>
+					{ __( 'What to do?', 'google-listings-and-ads' ) }
+				</strong>
+			</p>
+			<p>{ issue.action }</p>
+		</AppModal>
+	);
+};
+
+export default IssuesTableDataModal;
--- a/js/src/product-feed/issues-table-card/issues-table-data-modal.js
+++ b/js/src/product-feed/issues-table-card/issues-table-data-modal.js
@@ -14,12 +14,11 @@ import AppButton from '.~/components/app-button';
  * Renders a modal showing information about a Google Merchant Center issue
  *
  * @fires gla_documentation_link_click with { context: 'issues-data-table-modal' }
- * @param {Object} params Component params
- * @param {Object} params.issue The issue to be rendered in the modal
- * @param {Function} params.onRequestClose Onnclose callback function
+ * @param {Object} props React props
+ * @param {Object} props.issue The issue to be rendered in the modal
+ * @param {Function} [props.onRequestClose] Callback function when closing the modal
  */
 const IssuesTableDataModal = ( { issue, onRequestClose = () => {} } ) => {
-	if ( ! issue ) return null;
 	return (
 		<AppModal
 			className="gla-issues-table-data-modal"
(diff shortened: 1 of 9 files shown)


js/src/product-feed/issues-table-card/issues-table-data-modal.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Renders a modal showing information about a Google Merchant Center issue
+ *
+ * @fires gla_documentation_link_click with { context: 'issues-data-table-modal' }
+ * @param {Object} params Component params
+ * @param {Object} params.issue The issue to be rendered in the modal
+ * @param {Function} params.onRequestClose Onnclose callback function
+ */
+const IssuesTableDataModal = ( { issue, onRequestClose = () => {} } ) => {
+	if ( ! issue ) return null;
```
ianlin: 💅  While it would be impossible to happen, but when `issue` is an empty object `{}` it will be treated as `true` so it will shows the modal like this:

Screenshot 2022-05-19 at 13 22 12

We could probably use `Object.keys( issue ).length === 0` to check if it's an empty object, or use `empty` function from Lodash.

js/src/product-feed/issues-table-card/issues-table-data-modal.js
```diff
@@ -0,0 +1,55 @@
...
+import AppModal from '.~/components/app-modal';
+import AppButton from '.~/components/app-button';
+
+/**
+ * Renders a modal showing information about a Google Merchant Center issue
+ *
+ * @fires gla_documentation_link_click with { context: 'issues-data-table-modal' }
+ * @param {Object} params Component params
+ * @param {Object} params.issue The issue to be rendered in the modal
+ * @param {Function} params.onRequestClose Onnclose callback function
```
ianlin: 💅  This prop is optional.

js/src/product-feed/issues-table-card/issues-table-data-modal.js
```diff
@@ -0,0 +1,55 @@
...
+ * Internal dependencies
+ */
+import AppModal from '.~/components/app-modal';
+import AppButton from '.~/components/app-button';
+
+/**
+ * Renders a modal showing information about a Google Merchant Center issue
+ *
+ * @fires gla_documentation_link_click with { context: 'issues-data-table-modal' }
+ * @param {Object} params Component params
```
ianlin: 💅 In other files we use `@param {Object} props React props`. Would be great to maintain the consistency.

js/src/product-feed/issues-table-card/issues-table-data-modal.js
```diff
@@ -0,0 +1,55 @@
...
+			title={ issue.issue }
+			onRequestClose={ onRequestClose }
+			buttons={ [
+				<AppButton
+					key="learn-more"
+					isPrimary
+					target="_blank"
+					href={ issue.action_url }
+					text={ __( 'Learn more', 'google-listings-and-ads' ) }
+					eventName={ 'gla_documentation_link_click' }
```
ianlin: 💅 Strings don't need to be wrapped in `{}`.

js/src/product-feed/issues-table-card/issues-table-data.js
```diff
@@ -54,13 +64,31 @@ const IssuesTableData = ( { data, headers } ) => {
...
 				{
-					display: (
+					display: el.action ? (
+						<AppButtonModalTrigger
+							button={
+								<AppButton
+									isLink
+									eventName={
+										'gla_click_read_more_about_issue'
+									}
```
ianlin: 💅  Strings don't need to be wrapped in `{}`.

ianlin (approved):
Thanks for adding this! I've tested locally and it worked as expected. Here's the video of my test, I happened to have issues with both empty and non-empty actions:

https://user-images.githubusercontent.com/914406/169230528-944eea6e-72ed-4f4b-aa47-9f3336829356.mov

Also confirmed the events have been sent correctly on each click.

I've also left some non-blocking comments that were nice-to-have.

js/src/product-feed/issues-table-card/issues-table-data-modal.js
puntope: You are right, I think is actually better tor remove that line that adds confusion to the code, since its actually impossible to had an empty issue or non issue as a param, as you mentioned. 

Since we are iterating over the issues we know for sure the issue exist, also since having the `issue.action` prop is a condition to show the component we now as well that prop is there. 

We can definitely check if the issue has `issue.code `and `issue.href.` But that would impact as well in previous stages, so I guess that would be a bit of over-engineering.

puntope:
Howdy @ianlin I applied your suggestions in regards JSDOCS, thanks for catching that and review the PR ❤️

ianlin (approved):
Thanks for the update, LGTM 👍

PR #1495, opened 2022-05-10
Cleanup unused Google Ads services

mikkamp:
### Changes proposed in this Pull Request:

The main purpose of this PR is to reduce the size of the Google Ads library. The large size of the library, and the many generated files cause problems on some hosts that timeout while installing/updating the plugin.
Since we only use a small part of the library this PR creates a cleanup script which removes services and enums from the library which we don't use.

Running the script reduces the size in the following way:
- Final zip file: 10.7MB > 7MB = 3.7MB smaller
- Files in zip file: 33.6MB > 22.2MB = 11.4MB smaller
- Googleads directory: 18.3MB > 8.2MB = 10.1MB smaller

image

Resolves part of #314 

### Screenshots:
image

### Detailed test instructions:
1. Run composer install (which will run the cleanup script) 
2. Run the unit tests to confirm that all the needed files from the Ads library are still available
3. Run `npm run build`
4. Check the final zip file and confirm it's reduced in size

### Additional details:
The cleanup script is directly tied to the library version. If we update the library version the cleanup script will need to be modified and any newly added services will need to be added to the list. The list of services was taken from vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/Lib/V9/ServiceClientFactoryTrait.php

### Changelog entry
* Tweak - Cleanup unused Google Ads services.

PR #1520, opened 2022-05-23
Include ShoppingPerformanceView

### LINKED ISSUE #1519: PHP Error: ShoppingPerformanceView not found
### Describe the bug:
From Marketing -> Google Listings & Ads, it doesn't show reports but gives an error:
"We're having trouble loading this data. Try again later, or track your performance in Google Ads."
image
image

PHP Error:
2022-05-23T12:06:14+00:00 CRITICAL Uncaught Error: Class "Google\Ads\GoogleAds\V9\Resources\ShoppingPerformanceView" not found in /home/username/public_html/wp-content/plugins/google-listings-and-ads/vendor/google/protobuf/src/Google/Protobuf/Internal/Message.php:824

### Steps to reproduce:
On a site with Paid Ads running and connected to Google Listings & Ads version 1.13.0: 
1. Navigate to Marketing -> Google Listings & Ads
2. Select most recent month.

### Expected behavior:
Report shown for dates selected.

### Actual behavior:
From Marketing -> Google Listings & Ads, it doesn't show reports but gives an error:
"We're having trouble loading this data. Try again later, or track your performance in Google Ads."

### Additional details:
Error occurs when using version 1.13.0.
When rolling back to version 1.12.8 it works.

mikkamp:

In #1495 the ads library was cleaned up to reduce it's size. However for the reports this fails because of the missing resource ShoppingPerformanceView. We don't directly use any data from this class, but the resource name is included in the reports so it needs to remain in place.

This PR declares it's usage in the AdsReportQuery class so it is not automatically removed. The unit tests were also modified to mock the responses to include the ShoppingPerformanceView so it fails if the class is not present.

Closes #1519

1. Remove the vendor directory and run `composer install` to get a cleaned up copy of the ads library
2. Test with an ads account that is fully onboarded and has report data
3. View the dashboard in Marketing > Google Listings & Ads > Dashboard
4. Confirm that there are no errors in the dashboard or the reports page
5. If testing with the test proxy (no account with report data) then the modified file `tests/proxy/mocks/ads/reports/programs.json` is needed to reproduce the error without the fix in this PR.

### Changelog entry
* Fix - Missing ShoppingPerformanceView error when viewing report data.

### DIFF
--- a/src/API/Google/Query/AdsReportQuery.php
+++ b/src/API/Google/Query/AdsReportQuery.php
@@ -3,6 +3,8 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query;
 
+use Google\Ads\GoogleAds\V9\Resources\ShoppingPerformanceView;
+
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -16,6 +18,7 @@ abstract class AdsReportQuery extends AdsQuery {
 
 	/**
 	 * Query constructor.
+	 * Uses the resource ShoppingPerformanceView.
 	 *
 	 * @param array $args Query arguments.
 	 */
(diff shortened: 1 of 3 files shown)


findingsimple:
Thanks @mikkamp for the speedy fix

ianlin (approved):
Thanks for fixing this, tested locally using the mocked programs.json file and the error no longer appears. LGTM 👍

PR #1521, opened 2022-05-23
Use `woo-tracking-jsdoc` from external package

tomalec:
:warning: This PR requires https://github.com/woocommerce/grow/pull/13

Replace  `woo-tracking-jsdoc` with its external npm package

1. Remove a few lines from https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/README.md?plain=1#L76-L90 to see that the result is generated
2. ```sh
   npm i
   npm run doc:tracking
   ```
3. Check that `Tracking/README.md` is generated correctly

### Additional details:

- Personally, I'm not very convinced that the bundling of jsdoc & co. done in https://github.com/woocommerce/google-listings-and-ads/compare/1e85b2a90d73d9fa22cdaea93f3ff046629452d2...55ebac9ec7c62429931f52474146b469e17b9ffb is the best solution, but it seems to me like an acceptable solution to limit the number of dependencies shared across multiple extensions.

### More to be done:

- [x] update branch name once https://github.com/woocommerce/grow/pull/6 gets merged

<!--
Optional.
Each line should start with `(Fix|Add|Tweak|Update) - `, for example:
> Fix - I took care of something that wasn't working.
> Add - I added something new that's pretty cool.
> Tweak - I made a small change.
> Update - I made big changes to something that wasn't broken.

Leave the "Changelog entry" header in place completely empty, without any summary if no changelog entry is needed.
If you remove the "Changelog entry" header, the title of Pull Request will be used as the changelog entry.
-->
### Changelog entry


package.json
```diff
@@ -64,13 +64,10 @@
...
 		"h2o2": "^8.2.0",
 		"jest-environment-jsdom": "^27.5.1",
-		"jsdoc": "^3.6.10",
-		"jsdoc-advanced-types-plugin": "git+https://github.com/tomalec/jsdoc-advanced-types-plugin.git#add/return-support",
-		"jsdoc-plugin-intersection": "^1.0.4",
-		"jsdoc-plugin-typescript": "^2.0.6",
 		"path-browserify": "^1.0.1",
 		"prettier": "npm:wp-prettier@2.2.1-beta-1",
 		"stylelint-config-standard-scss": "^3.0.0",
+		"woocommerce-grow-jsdoc": "https://gitpkg.now.sh/woocommerce/grow/packages/js/jsdoc?69d8a735b6c07bc601ead0a68cd2b08428c79f8b",
```
eason9487: ❔ I guess this sha should be changed to `c99cb5af2f6d73c3d0c0173e5f8b0b681770a52e`?

eason9487 (approved):
Tested with both revisions `69d8a735b6c07bc601ead0a68cd2b08428c79f8b` and `c99cb5af2f6d73c3d0c0173e5f8b0b681770a52e` of `woocommerce-grow-jsdoc`, all are work well. Therefore, I assumed it should not need a new round of review even if the correct sha should be the latter one. Approved in advance.

💅 
I happened to notice a missed ending backtick `` ` `` at
https://github.com/woocommerce/google-listings-and-ads/blob/e314b2ec5ba0f0d0f7df0b82767933cd7a62acd5/js/src/product-feed/review-request/review-request-modal.js#L48


It's not introduced by this PR but has a bit of association since this PR also updates the tracking README file. Probably it could be fixed along with this PR.

PR #1522, opened 2022-05-24
Simplify and centralize the processing of internal states for TreeSelectControl

eason9487:

To simplify and centralize the processing of updating expanded, checked, and partially checked states, this PR:
- Replace `cloneDeep` with array `reduce` when generating filtered options.
- Calculate the expanded, checked, and partially checked states of options in the main component.
- Move the processing of updating the expanded state to the main component.
- Simplify the handling of updating the state of selected values.

1. Go to the Edit free listings or the local storybook.
2. Test if the states and functionality of checking/unchecking options work correctly.
3. Test if the states and functionality of expanding/collapsing options work correctly.

### Changelog entry

> Tweak - Simplify and centralize the processing of internal states for the TreeSelectControl component.

### DIFF
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -297,6 +297,19 @@ const TreeSelectControl = ( {
 	};
 
 	/**
+	 * Expands/Collapses the Option
+	 *
+	 * @param {InnerOption} option The option to be expanded or collapsed.
+	 */
+	const handleToggleExpanded = ( option ) => {
+		setNodesExpanded(
+			option.expanded
+				? nodesExpanded.filter( ( el ) => option.value !== el )
+				: [ ...nodesExpanded, option.value ]
+		);
+	};
+
+	/**
 	 * Handles a change on the Tree options. Could be a click on a parent option
 	 * or a child option
 	 *
@@ -432,9 +445,8 @@ const TreeSelectControl = ( {
 						options={ filteredOptions }
 						value={ value }
 						onChange={ handleOptionsChange }
-						nodesExpanded={ nodesExpanded }
-						onNodesExpandedChange={ setNodesExpanded }
 						onExpanderClick={ handleExpanderClick }
+						onToggleExpanded={ handleToggleExpanded }
 					/>
 				</div>
 			) }
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -195,6 +195,11 @@ const TreeSelectControl = ( {
 		};
 
 		const descriptors = {
+			hasChildren: {
+				get() {
+					return this.children?.length > 0;
+				},
+			},
 			allLeaves: {
 				/**
 				 * Returns all (filtered) leaf options under this option.
@@ -263,27 +268,22 @@ const TreeSelectControl = ( {
 		};
 
 		const reduceOptions = ( acc, { children = [], ...option } ) => {
-			option.hasChildren = children.length > 0;
-
-			if ( option.hasChildren ) {
+			if ( children.length ) {
 				option.children = children.reduce( reduceOptions, [] );
 
-				if ( option.children.length ) {
-					Object.defineProperties( option, descriptors );
-					acc.push( option );
+				if ( ! option.children.length ) {
+					return acc;
 				}
-			} else {
-				if ( isFiltered ) {
-					const match = option.label.toLowerCase().indexOf( filter );
-					if ( match === -1 ) {
-						return acc;
-					}
-					option.label = highlightOptionLabel( option.label, match );
+			} else if ( isFiltered ) {
+				const match = option.label.toLowerCase().indexOf( filter );
+				if ( match === -1 ) {
+					return acc;
 				}
-
-				Object.defineProperties( option, descriptors );
-				acc.push( option );
+				option.label = highlightOptionLabel( option.label, match );
 			}
+
+			Object.defineProperties( option, descriptors );
+			acc.push( option );
 
 			return acc;
 		};
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -169,8 +169,8 @@ const TreeSelectControl = ( {
 	 * 5. Finally we set the cache with the obtained results and apply the filters
 	 */
 	const filteredOptions = useMemo( () => {
-		const { current } = cacheRef;
-		const cachedFilteredOptions = current.filteredOptionsMap.get( filter );
+		const { current: cache } = cacheRef;
+		const cachedFilteredOptions = cache.filteredOptionsMap.get( filter );
 
 		if ( cachedFilteredOptions ) {
 			return cachedFilteredOptions;
@@ -232,7 +232,7 @@ const TreeSelectControl = ( {
 					if ( this.hasChildren ) {
 						return this.leaves.every( ( opt ) => opt.checked );
 					}
-					return current.selectedValues.includes( this.value );
+					return cache.selectedValues.includes( this.value );
 				},
 			},
 			partialChecked: {
@@ -266,7 +266,7 @@ const TreeSelectControl = ( {
 					return (
 						isSearching ||
 						this.value === ROOT_VALUE ||
-						current.expandedValues.includes( this.value )
+						cache.expandedValues.includes( this.value )
 					);
 				},
 			},
@@ -294,7 +294,7 @@ const TreeSelectControl = ( {
 		};
 
 		const filteredTreeOptions = treeOptions.reduce( reduceOptions, [] );
-		current.filteredOptionsMap.set( filter, filteredTreeOptions );
+		cache.filteredOptionsMap.set( filter, filteredTreeOptions );
 
 		return filteredTreeOptions;
 	}, [ treeOptions, filter ] );
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -63,6 +63,8 @@ import { ARROW_DOWN, ARROW_UP, ENTER, ESCAPE, ROOT_VALUE } from './constants';
  * @typedef {Object} BaseInnerOption
  * @property {string|JSX.Element} label The label string or label with highlighted react element for the option.
  * @property {InnerOption[]|undefined} children The (filtered) children options.
+ * @property {boolean} hasChildren Whether this option has children.
+ * @property {boolean} expanded Whether this option is expanded.
  *
  * @typedef {CommonOption & BaseInnerOption} InnerOption
  */
@@ -112,6 +114,8 @@ const TreeSelectControl = ( {
 
 	// We will save in a REF previous search filter queries to avoid re-query the tree and save performance
 	const filteredOptionsCache = useRef( {} );
+	const cacheStatesRef = useRef( {} );
+	cacheStatesRef.current.expandedValues = nodesExpanded;
 
 	const showTree = ! disabled && treeVisible;
 
@@ -152,15 +156,14 @@ const TreeSelectControl = ( {
 		return repository;
 	}, [ treeOptions ] );
 
-	/**
+	/*
 	 * Perform the search query filter in the Tree options
 	 *
 	 * 1. Check if the search query is already cached and return it if so.
-	 * 2. Deep Copy the tree. Since we are going to modify its children and labels recursively.
+	 * 2. Deep copy the tree with adding properties for rendering.
 	 * 3. In case of filter, we apply the filter option function to the tree.
 	 * 4. In the filter function we also highlight the label with the matching letters
 	 * 5. Finally we set the cache with the obtained results and apply the filters
-	 *
 	 */
 	const filteredOptions = useMemo( () => {
 		const cachedFilteredOptions = filteredOptionsCache.current[ filter ];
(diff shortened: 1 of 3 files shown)


js/src/components/tree-select-control/index.js
```diff
@@ -92,14 +110,16 @@ const TreeSelectControl = ( {
...
 
 	const dropdownRef = useRef();
 	const onDropdownVisibilityChangeRef = useRef();
 	onDropdownVisibilityChangeRef.current = onDropdownVisibilityChange;
 
 	// We will save in a REF previous search filter queries to avoid re-query the tree and save performance
 	const filteredOptionsCache = useRef( {} );
+	const cacheStatesRef = useRef( {} );
+	cacheStatesRef.current.expandedValues = nodesExpanded;
+	cacheStatesRef.current.selectedValues = value;
```
puntope: 💅  WDTY of saving all the caches in

`cacheRef.current.filteredOptions`
`cacheRef.current.expandedValues `
`cacheRef.current.selectedValues `

js/src/components/tree-select-control/index.js
```diff
@@ -140,23 +159,29 @@ const TreeSelectControl = ( {
...
 	 */
-	useEffect( () => {
+	const filteredOptions = useMemo( () => {
 		const cachedFilteredOptions = filteredOptionsCache.current[ filter ];
 
+		if ( cachedFilteredOptions ) {
+			return cachedFilteredOptions;
+		}
+
+		const isFiltered = Boolean( filter );
```
puntope: ❓  Why is this necessary? 

Just using `filter` should work I guess

js/src/components/tree-select-control/index.js
```diff
@@ -169,30 +194,104 @@ const TreeSelectControl = ( {
...
+					const match = option.label.toLowerCase().indexOf( filter );
+					if ( match === -1 ) {
+						return acc;
+					}
+					option.label = highlightOptionLabel( option.label, match );
+				}
+
+				Object.defineProperties( option, descriptors );
+				acc.push( option );
+			}
```
puntope: 💅 

Just an idea, I guess the reducer could be simplified with: 

```
const descriptors = {
            hasChildren: {
				get() {
					return this.children?.length > 0;
				},
			},
			allLeaves: {
```
```
const reduceOptions = ( acc, { children = [], ...option } ) => {
			if ( children.length ) {
				option.children = children.reduce( reduceOptions, [] );
			} else if ( isFiltered ) {
				const match = option.label.toLowerCase().indexOf( filter );
				if ( match === -1 ) {
					return acc;
				}
				option.label = highlightOptionLabel( option.label, match );
			}

			if ( option.children?.length || ! children.length ) {
				Object.defineProperties( option, descriptors );
				acc.push( option );
			}
			return acc;
		};
```

js/src/components/tree-select-control/index.js
```diff
@@ -46,13 +46,31 @@ import { ARROW_DOWN, ARROW_UP, ENTER, ESCAPE, ROOT_VALUE } from './constants';
...
+
+/**
+ * @typedef {Object} BaseOption
  * @property {string} label The label for the option
  * @property {Option[]} [children] The children Option objects
- * @property {string} [key] Optional unique key for the Option. It will fallback to the value property if not defined
- **/
+ *
+ * @typedef {CommonOption & BaseOption} Option
+ */
```
puntope, line 60: 💅  I will use just `Option` instead of CommonOption and BaseOption. (Or maybe I missed the intent of this separation). 

@typedef {Object} Option
@property {string} label The label for the option
@property {string} value The value for the option
@property {string} [key] Optional unique key for the Option. It will fallback to the value property if not defined
@property {Option[]} [children] The children Option objects

💅  WDYT of renaming `InnerOption` as well as `BaseInnerOption` to `Leaf` and `BaseLeaf`

In my mind I see it this way:
Option is the "pure" item we send to the tree, the data itself, with label, value, and optional key and children.
Leaf is and enhanced Option with Tree props (BaseLeaf).

js/src/components/tree-select-control/index.js
```diff
@@ -169,30 +194,104 @@ const TreeSelectControl = ( {
 			);
 		};
 
-		const filterOption = ( option ) => {
-			if ( hasChildren( option ) ) {
-				option.children = option.children.filter( filterOption );
-				return !! option.children.length;
-			}
+		const descriptors = {
+			allLeaves: {
```
puntope: 💅  I will keep it as `leaves` instead of `allLeaves`. When I saw `all` highlighted in the name of the prop, I was expecting to have just a generic `leaves` somewhere in the code. 

Also the description

`@property {InnerOption[]} allLeaves All (filtered) leaf options under this option.`

I'd suggest to replace it with something like: 

`@property {Leaf[]} leaves All the filtered option children flattened.`

PD. I guess InnerOption is not accurate since we have for example the option.label as well??? 🤔

js/src/components/tree-select-control/index.js
puntope: 💅 WDYT of adding a small comment in the reducer?

Something like "This reducer gets the options matching the filter and also adds Leaf properties to them"

puntope (commented):
Howdy @eason9487 

I did a first check for this PR and it works amazing, also it simplifies a lot the relationship between the Options component and the tree ❤️ 

In left some comments tho which I think deserve discussion.

Thanks!

js/src/components/tree-select-control/index.js
eason9487: Changed in https://github.com/woocommerce/google-listings-and-ads/pull/1522/commits/e2a6a827289312f796e202e68b628ed34f0485d4.

js/src/components/tree-select-control/index.js
eason9487: In this context, `isFiltered` is more semantic than `filter`, and the explicit boolean type allows the below return value to always be a boolean type.

https://github.com/woocommerce/google-listings-and-ads/blob/afa8de0035bd66217caa96b5212fb51ff1ea2be9/js/src/components/tree-select-control/index.js#L258-L266

js/src/components/tree-select-control/index.js
eason9487: Thanks for the nice suggestion! Based on your example, I adjusted it to a similar implementation in https://github.com/woocommerce/google-listings-and-ads/pull/1522/commits/da745d8b36ba4e9e020ae00604f85d8fcf238630.

---

> WDYT of adding a small comment in the reducer?

The processing of this reducer is also included in the code comments above, so I guess it's already sufficient?

https://github.com/woocommerce/google-listings-and-ads/blob/afa8de0035bd66217caa96b5212fb51ff1ea2be9/js/src/components/tree-select-control/index.js#L163-L169

js/src/components/tree-select-control/index.js
eason9487: > I will use just `Option` instead of CommonOption and BaseOption. (Or maybe I missed the intent of this separation).

Option and InnerOption both have `label` and `children` properties, but the type and description of the properties are different. JSDoc does not override the same attributes when using `@typedef {Option & BaseInnerOption} InnerOption`, but combines them.

For example, using this JSDoc: 

```js
/**
 * @typedef {Object} Option
 * @property {string} label The label for the option.
 * @property {string} value The value for the option.
 * @property {string} [key] Optional unique key for the Option. It will fallback to the value property if not defined.
 * @property {Option[]} [children] The children Option objects.
 */

/**
 * @typedef {Object} BaseInnerOption
 * @property {string|JSX.Element} label The label string or label with highlighted react element for the option.
 * @property {InnerOption[]|undefined} children The children options. The options are filtered if in searching.
 * (omitted)
 * @typedef {Option & BaseInnerOption} InnerOption
 */
```

There will be a combined type and description:



Furthermore, the nested reference produces inconsistent results.




---

> WDYT of renaming `InnerOption` as well as `BaseInnerOption` to `Leaf` and `BaseLeaf`

In a tree data structure, the terminology of *leaf* node refers to the node with no children. But an `InnerOption` could be a parent node, both parent and child node, or a leaf node. Therefore, I don't think `Leaf` is a proper name.

js/src/components/tree-select-control/index.js
eason9487: Renamed the `allLeaves` and adjusted some JSDocs in https://github.com/woocommerce/google-listings-and-ads/pull/1522/commits/afa8de0035bd66217caa96b5212fb51ff1ea2be9.

> PD. I guess InnerOption is not accurate since we have for example the option.label as well???

Sorry, I don't quite understand this. The `InnerOption` does have the `label` property. And what does *PD.* refer to? Could you elaborate more?

js/src/components/tree-select-control/index.js
puntope: When I saw it I thought for a moment that was kind of a flag indicating that a node was filtered or not. 

💅  Maybe consider to change it to `hasFilters`

js/src/components/tree-select-control/index.js
```diff
@@ -169,30 +194,104 @@ const TreeSelectControl = ( {
...
-			if ( hasChildren( option ) ) {
-				option.children = option.children.filter( filterOption );
-				return !! option.children.length;
-			}
+		const descriptors = {
+			hasChildren: {
+				get() {
+					return this.children?.length > 0;
+				},
+			},
```
puntope: 💅  Missing docs

js/src/components/tree-select-control/index.js
eason9487: > When I saw it I thought for a moment that was kind of a flag indicating that a node was filtered or not.

The naming was taken from the former prop used in the Options component.

https://github.com/woocommerce/google-listings-and-ads/blob/65d232a5aab91dad7910ec352a6389ff4c6f140c/js/src/components/tree-select-control/options.js#L26

> Maybe consider to change it to `hasFilters`

How about `isSearching`?

js/src/components/tree-select-control/index.js
puntope: > In a tree data structure, the terminology of leaf node refers to the node with no children. But an InnerOption could be a parent node, both parent and child node, or a leaf node. Therefore, I don't think Leaf is a proper name.

👌  Right. Maybe `node` then. 

> There will be a combined type and description:

Interesting. In my IDE it does the intersection properly.

Screenshot 2022-05-25 at 12 31 34

My worries are that Inner, Base, Common prefixes could be kinda confusing but this is a non blocking issue.

js/src/components/tree-select-control/index.js
puntope: > Sorry, I don't quite understand this. The InnerOption does have the label property. And what does PD. refer to? Could you elaborate more?

Apologies. I missed the label property inside `InnerOption`. All good 👌 (with PD I meant PS, had a typo)

js/src/components/tree-select-control/index.js
eason9487: > Maybe `node` then.

`Node` is a more general and abstract naming. When viewing the code like `@param {Node} option`, I don't think it's clearer, and it also seems more irrelevant to the parameter name.

> My worries are that Inner, Base, Common prefixes could be kinda confusing

Sorry, I'm not quite sure how they cause confusion in the scope of these type definitions since the Base and Common prefixes are only used for combining the Option and InnerOption. But I'd like to know if you have other naming ideas to replace them to reduce your confusion.

puntope (approved):
✅  All the discussion were addressed! This is good to go. Thanks @eason9487 

I left some opinionated comments 💅  that could improve even more the code IMO. Those are optional or could be done in future PR. So this is approved.

js/src/components/tree-select-control/index.js
puntope: isSearching looks pretty accurate for me yes! 🤝

js/src/components/tree-select-control/index.js
puntope: All the names I have in mind rename Option to Node for defining the items of the tree and differentiate them from the Options provided by the user.

Option = { OptionValue & OptionData } 
Node = { OptionValue & BaseNode } 

`<Options options="filteredOptions"> to be renamed to <Nodes nodes="filteredNodes">`

But that requires more refactor of the Options component that was named wrongly from scratch by me. (notice in the Options.js component we refer to Nodes in ArrowLeft/ArrowRight docs) 

But this is pretty much a subjective and minor issue (unsure even if it worth it all the refactor in renaming).

js/src/components/tree-select-control/index.js
eason9487: Changed in https://github.com/woocommerce/google-listings-and-ads/pull/1522/commits/6b13282a3297613e3d68cd3589315b9c6721e339.

js/src/components/tree-select-control/index.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/1522/commits/6dfadfc6f50e9938f0f5434a1b4fdad1a9bcf93c.

js/src/components/tree-select-control/index.js
eason9487: I think the overall naming adjustments could be a separate PR. Together with, for example, the `value` prop could be internally renamed to `values` or `selectedValues`, and the `nodesExpanded` to `expandedValues`, etc.

PR #1525, opened 2022-05-24
Don't take in consideration order when comparing Shipping Rate objects

### LINKED ISSUE #1433: Edit Free Listings: Unsaved data navigate away prompt might show up when shipping times are still the same after being modified
### Describe the bug:

- This issue is a follow up from PR https://github.com/woocommerce/google-listings-and-ads/pull/1409#pullrequestreview-944049302.
- This issue is similar to issue https://github.com/woocommerce/google-listings-and-ads/issues/1431.

In Edit Free Listings Step 2 shipping time section, if we modify the shipping time setting and the values remain the same after modification, the "unsaved data" prompt might show up.

This is because even though the values look the same in the UI, the underlying array data `shippingTimes` and `savedShippingTimes` may have the shipping times in different order, and hence does not satisfy the `isEqual` check, causing the navigate away prompt to show up.

https://github.com/woocommerce/google-listings-and-ads/blob/8e0ae939a647850e74babe1cb4d1497dee1b9be3/js/src/edit-free-campaign/index.js#L131

📹  Demo video with my voice:

https://user-images.githubusercontent.com/417342/163783991-d7fa293d-d33a-4929-907d-b85f4e14302f.mov

### Steps to reproduce:

1. Go to Edit Free Listings, Step 2. 
2. Make some modifications for shipping times, e.g. delete shipping times and add the same shipping times back.
3. Try to reload the page or click on the top left "go back" button.

### Expected behavior:

The navigate away prompt should not show up.

### Actual behavior:

The navigate away prompt might show up.

puntope:

Closes #1433.

Avoiding to show the Unsaved Data prompt in case the values of Shipping rates are the same.

This bug happens for example when the user deletes a Shipping rate row and add it again. Since the order is different the isEqual function considers them different objects.

### Screenshots:

https://user-images.githubusercontent.com/5908855/170033769-cd2efac2-4c8c-4c20-91f0-db9fe2f8e333.mov

1. Checkout PR and go to Edit Free Listings
2. Setup a target audience of at least 2 countries
3. In the step 2, in shipping rates you should have at least 2 different shipping times rules.
4. Save the changes
5. Check that you can go back without any prompt after saving changes
6. Go again to Edit free listings step 2
7. Modify the first row value in Shipping times
8. Try to go back, and see that the prompt appears
9. Click cancel
10. Modify the first row again in Shipping Times to the original value
11. Try to go back, you don't see any prompt
12. Go again to Edit free listings step 2
13. Remove the first row in Shipping Times
14. Try to go back, and see that the prompt appears
15. Add a new row with the same value as the original 
16. Try to go back, you don't see any prompt

### Additional details:

We use `new Set()` in order to ignore the order of elements.

### Changelog entry

> Fix - Avoid to show Unsaved Values confirmation in Edit Free Listing when no values has been changed.

### DIFF
--- a/js/src/edit-free-campaign/index.js
+++ b/js/src/edit-free-campaign/index.js
@@ -134,7 +134,7 @@ const EditFreeCampaign = () => {
 		savedShippingRates
 	);
 
-	// Dont take in consideration the order of the rules when comparing the Shipping times.
+	// Check what've changed to show prompt. Dont take in consideration the order when comparing the Shipping times.
 	const didTimesChanged = ! isEqual(
 		new Set( shippingTimes ),
 		new Set( savedShippingTimes )
--- a/js/src/edit-free-campaign/index.js
+++ b/js/src/edit-free-campaign/index.js
@@ -134,7 +134,12 @@ const EditFreeCampaign = () => {
 		savedShippingRates
 	);
 
-	const didTimesChanged = ! isEqual( shippingTimes, savedShippingTimes );
+	// Dont take in consideration the order when comparing objects
+	const didTimesChanged = ! isEqual(
+		new Set( shippingTimes ),
+		new Set( savedShippingTimes )
+	);
+
 	const didAnythingChanged =
 		didAudienceChanged ||
 		didSettingsChanged ||
--- a/js/src/edit-free-campaign/index.js
+++ b/js/src/edit-free-campaign/index.js
@@ -134,7 +134,7 @@ const EditFreeCampaign = () => {
 		savedShippingRates
 	);
 
-	// Dont take in consideration the order when comparing objects
+	// Dont take in consideration the order of the rules when comparing the Shipping times.
 	const didTimesChanged = ! isEqual(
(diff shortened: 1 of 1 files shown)


js/src/edit-free-campaign/index.js
```diff
@@ -134,7 +134,12 @@ const EditFreeCampaign = () => {
 		savedShippingRates
 	);
 
-	const didTimesChanged = ! isEqual( shippingTimes, savedShippingTimes );
+	// Dont take in consideration the order of the rules when comparing the Shipping times.
```
jorgemd24: 💅

I would say "Dont take in consideration the order of the shipping times" as I think they are not rules. 

Or something similar to:

https://github.com/woocommerce/google-listings-and-ads/blob/99bfa1819e680a35aee18e1da0a9e2fa998db41f/js/src/edit-free-campaign/index.js#L123

jorgemd24 (approved):
Thanks, @puntope for fixing this. I tested it locally and it is working as expected. LGTM 👍

PR #1526, opened 2022-05-24
Switch to `@woocommerce/dependency-extraction-webpack-plugin`

### LINKED ISSUE #916: Consider switching to **`@woocommerce/`**`dependency-extraction-webpack-plugin`
Currently we use `@wordpress/dependency-extraction-webpack-plugin` in quite outdated version, to [wrangle](https://github.com/woocommerce/google-listings-and-ads/blob/trunk/webpack.config.js#L19-L20) WooCommerce components.

So far there is no problem directly connected to it. But I recall struggling with some `/components` dependencies in the past (like https://github.com/woocommerce/google-listings-and-ads/pull/234).

After reading a comment https://github.com/saucal/pinterest-for-woocommerce/pull/105/files#r678799560 I think it may be worth changing it here, to save us some problems in the future, plus maybe workaround WP package update related problems https://github.com/woocommerce/google-listings-and-ads/pull/821

tomalec:

Switch to `@woocommerce/dependency-extraction-webpack-plugin`,
from `@wordpress/dependency-extraction-webpack-plugin`.

To simplify `webpack.config` and consistently only opt-out bundled packages.
Instead of opting out WP ones, and optin in WC ones.

Fixes: #916 

1. ```
   npm i
   npm run dev
   ```
2. Compare `.externalized.json` & `js/build/index.js` before and after the above operation


### Additional details:

- :warning: This PR currently does not change the effective bundle, to make it safer to merge But now we explicitly opt-out following packages:
   - `@woocommerce/currency`
   - `@woocommerce/date`
   - `@woocommerce/number`
   - `@woocommerce/tracks`
  We could consider removing those from the bundle as well in the next cooldown, to limit the risk of unexpected bugs.

<!--
Optional.
Each line should start with `(Fix|Add|Tweak|Update) - `, for example:
> Fix - I took care of something that wasn't working.
> Add - I added something new that's pretty cool.
> Tweak - I made a small change.
> Update - I made big changes to something that wasn't broken.

Leave the "Changelog entry" header in place completely empty, without any summary if no changelog entry is needed.
If you remove the "Changelog entry" header, the title of Pull Request will be used as the changelog entry.
-->
### Changelog entry


eason9487 (approved):
Tested locally by randomly viewing GLA pages and the pages work well. Confirmed the content of `.externalized.json` and the bundle size of index.js are the same. LGTM.

tomalec:
Created a follow-up issue to revisit bundled packages: https://github.com/woocommerce/google-listings-and-ads/issues/1527

PR #1529, opened 2022-05-25
Prevent repeated account url retrieval

mikkamp:

In PR #1503 we implemented the retrieval of account_url to ensure the site name matches before syncing. If the account_url is not previously stored for sites it will need to fetch it from the account data.

However the problem with that change is that it only stores the URL after successfully fetching it from the account. For sites that assume they are connected (mc_setup_completed_at and google_connected are valid), but in reality their token or refresh token is revoked or not available, the site will get an error when trying to fetch the URL from the account.
This scenario then results in it repeatedly retrying, to prevent this the PR adds a transient for storing when the url matches. We store `yes/no` values in the transients because a boolean can be detected as transient not available.

#### Reproduce the issue with develop branch
1. Start off with a copy of develop (to reproduce the failure) 
2. If testing with a local WCS then we can temporarily remove the refresh token and set the token as expired in the table `wcc_token`
3. Clear out any saved option with the name `gla_claimed_url_hash`
4. Confirm that with every page request the function `get_claimed_url_hash` is called
5. In your local WCS you should see it return errors
6. In WooCommerce > Settings > Logs you should see an entry get added every time a page is loaded:
```
2022-05-25T12:18:14+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\Internal\DependencyManagement\GoogleServiceProvider::handle_unauthorized_error Client error: `GET http://localhost:5500/google/google-mc/content/v2.1/12345678/accounts/12345678` resulted in a `401 Unauthorized` response:
{"statusCode":401,"error":"Unauthorized","message":"Site is not connected to Google account for google-mc"}
```

#### Test the PR
1. Leave the tokens in the local WCS in the invalid state
2. Clear out any saved option with the name `gla_claimed_url_hash`
3. Confirm that the function `get_claimed_url_hash` is called only once even after repeated page refreshes
4. Confirm the error in WooCommerce > Settings > Logs is only added once
5. Clear the transient with the name `gla_url_matches`
6. Refresh a page and confirm the error is again only added once even after multiple refreshes

### Changelog entry
* Fix - Prevent repeated account URL retrievals.

### DIFF
--- a/src/MerchantCenter/MerchantCenterService.php
+++ b/src/MerchantCenter/MerchantCenterService.php
@@ -104,6 +104,7 @@ class MerchantCenterService implements ContainerAwareInterface, OptionsAwareInte
 	 * @return boolean
 	 */
 	public function is_ready_for_syncing(): bool {
+		/** @var TransientsInterface $transients */
 		$transients  = $this->container->get( TransientsInterface::class );
 		$url_matches = $transients->get( TransientsInterface::URL_MATCHES );
 
--- a/src/MerchantCenter/MerchantCenterService.php
+++ b/src/MerchantCenter/MerchantCenterService.php
@@ -16,6 +16,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Options\MerchantAccountState;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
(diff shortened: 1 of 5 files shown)


src/MerchantCenter/MerchantCenterService.php
```diff
@@ -95,15 +97,24 @@ public function is_google_connected(): bool {
...
+	 * URL matches is stored in a transient to prevent it from being refetched in cases
+	 * where the site is unable to access account data.
 	 *
 	 * @since 1.13.0
 	 * @return boolean
 	 */
 	public function is_ready_for_syncing(): bool {
-		$claimed_url_hash = $this->container->get( Merchant::class )->get_claimed_url_hash();
-		$site_url_hash    = md5( $this->get_site_url() );
+		$transients  = $this->container->get( TransientsInterface::class );
```
JPry, line 108: Minor: Let's add a comment to help out IDEs.

```suggestion
		/** @var TransientsInterface $transients */
		$transients  = $this->container->get( TransientsInterface::class );
```

JPry (approved):
Looks good, just one tiny tweak recommend.

PR #1533, opened 2022-05-27
JS unit tests for Gtag events

mikkamp:

This PR is a followup for #1488 and adds JS unit tests for the utility functions of the Gtag events code.

The rounding function was added when converting the price to combat floating point precision. Without it 10.12 is converted to 1011.99999

1. Run the unit tests with `npm run test:js` 
2. Confirm that all the tests pass without any issues

### Changelog entry
* Add - JS Unit testing for Gtag events.


js/src/gtag-events/gtag-event.test.js
```diff
@@ -0,0 +1,147 @@
...
+			send_to: 'GLA',
+		} );
+	} );
+
+	it( 'track add to cart event', () => {
+		const product = {
+			id: 1234,
+			name: 'Test name',
+			prices: {
+				price: 1012,
```
puntope, line 44: ❓ Why 1012 and not 10.12?

puntope: Never mind, I see this is what `currency_minor_unit` does

puntope (approved):
✅  LGTM

Test suit covers utils and constants definitions. I guess the addEvent listeners will be tested by e2e

mikkamp:
Thanks for the review, I'm going to hold off with merging this until the parent PR #1488 has been merged. So far the only change needed here is that we no longer need to mock `console.log`

> I guess the addEvent listeners will be tested by e2e

That was the intention, although it looks like we might need to do some more work on our e2e tests for that to be a reality. In order for this code to be enqueued the merchant and ads account must have been fully onboarded otherwise there is no tracking_id present.

For now I'm going to expand our Wiki with steps on how to manually test tracking and how to find the recorded data in the Ads account.

PR #1534, opened 2022-05-30
Fix the slow GitHub actions

eason9487:

To fix the unsupported protocol used in `npm ci` that causes timeout and retry, this repo replaces the unsupported protocol of unencrypted `git://` with `https://` for:
- `jsdoc-advanced-types-plugin` package
- `@automattic/puppeteer-utils` dependency of `@woocommerce/e2e-environment` package

1. Run `npm ci --ignore-scripts`. It should be faster than the 82d8e5fe6b2dfcdf7ea3a8274a84519523aeded8 revision.
2. Go to the Check tab of this PR to see if the **Bundle Size** and **E2E Tests** are faster than before.

### Additional details:

Since Friday, May 13, 2022, the time required to finish GitHub actions that run `npm ci`  is significantly slower. The reason for this should be that [GitHub had formally turned off the unencrypted `git://` protocol](https://github.blog/2021-09-01-improving-git-protocol-security-github/).

When running `npm ci`, if there are packages installed directly from GitHub, it will fetch the heads and tags of those packages by `git ls-remote`. And the terminated protocol causes `git ls-remote` to attempt three times and spend 75 seconds each until timeout. The timeout protocol and alternative protocols can be verified with these commands:

- `git ls-remote -h -t git://github.com/Automattic/puppeteer-utils.git` (timeout)
- `git ls-remote -h -t ssh://github.com/Automattic/puppeteer-utils.git` (works)
- `git ls-remote -h -t https://github.com/Automattic/puppeteer-utils.git` (works and faster)

### Changelog entry

>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,8 +15,8 @@
 			}
 		},
 		"@automattic/puppeteer-utils": {
-			"version": "github:Automattic/puppeteer-utils#0f3ec50fc22d7bd2a4bd69fc172e8a66d958ef2d",
-			"from": "github:Automattic/puppeteer-utils#0f3ec50",
+			"version": "git+https://github.com/Automattic/puppeteer-utils.git#0f3ec50fc22d7bd2a4bd69fc172e8a66d958ef2d",
+			"from": "git+https://github.com/Automattic/puppeteer-utils.git#0f3ec50",
 			"dev": true,
 			"requires": {
 				"@babel/cli": "^7.8.3",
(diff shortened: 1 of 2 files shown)


mikkamp (approved):
Thanks for catching this I see it make a big difference when I compare PR's (the prepare node step).

Develop branch:
image

This branch:
image

I also tried it locally and see a big difference:
```
develop branch:
$ time npm ci --ignore-scripts
npm WARN prepare removing existing node_modules/ before installation
added 4282 packages in 402.417s

real	6m42.670s
user	0m31.902s
sys	0m12.760s

this PR:
$ time npm ci --ignore-scripts
npm WARN prepare removing existing node_modules/ before installation
added 4282 packages in 21.166s

real	0m21.481s
user	0m26.461s
sys	0m13.048s
```

PR #1535, opened 2022-05-30
Allow unicode for Manufacturer Part Number (MPN ) value

### LINKED ISSUE #1510: Allow Unicode characters in the MPN value when syncing a product

Currently, it allows alphanumeric in the MPN value when syncing a product to Google Merchant Center. Products have other characters in the MPN value won't be synced.

Ref:
- https://github.com/woocommerce/google-listings-and-ads/blob/1.12.8/src/Product/WCProductAdapter.php#L767
- https://github.com/woocommerce/google-listings-and-ads/blob/1.12.8/src/Product/BatchProductHelper.php#L211-L224

According to [the Google document](https://support.google.com/merchants/answer/6324482), the format of MPN value allows Unicode characters. Therefore, it's expected that the validation of product MPN should allow Unicode characters.

### Additional details

The format description of MPN value in [this Google document](https://support.google.com/merchants/answer/160161) is inaccurate.

image

### Customer Reports

p1652172122856809-slack-C7U3Y3VMY

- 4915339-zen
- 5219956-zen

ianlin:

Closes #1510. This PR adds unicode support for the [MPN](https://support.google.com/merchants/answer/6324482) value. Originally it only allows alphanumeric but from the document we can see it should supports unicode characters as well.

#### Reproduction Steps

1. Use the latest version as of now `1.13.2`.
2. In the admin page, go to WooCommerce Products `/wp-admin/edit.php?post_type=product`.
3. Update an existing product's price under `Product data` -> `General`.
4. Add a MPN number with unicode characters (e.g. `MPN-TEST-123`) under `Product data` -> `Google Listings and Ads`.
5. Click `Update`
6. Wait for a few moments and go to Google Merchant Centre's [All products](https://merchants.google.com/mc/items) section, the product you updated for the price shouldn't be synced to it, i.e. the price remain unchanged.

#### Test this PR

1. Follow the steps above but using the branch of this PR, you should see the price is synced to the Google Merchant Centre.

### Additional details:

I spent quite a few time trying to add automated tests to this one-line code change PR, but couldn't find a way to implement it. I think since the Type string are already tested in [Symfony's Constraints](https://github.com/symfony/symfony/blob/a10071bd657c350bb8f995361643072a97ff5819/src/Symfony/Component/Validator/Tests/Constraints/TypeValidatorTest.php) and we have already tested when the product has validation error (see below code), we don't need to test if the provided value is actually a string or not.

https://github.com/woocommerce/google-listings-and-ads/blob/82d8e5fe6b2dfcdf7ea3a8274a84519523aeded8/tests/Unit/Product/BatchProductHelperTest.php#L215-L227

### Changelog entry

> Fix - Allow unicode for Manufacturer Part Number (MPN ) value

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -764,7 +764,7 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 		$metadata->addConstraint( new Assert\Callback( 'validate_availability' ) );
 
 		$metadata->addPropertyConstraint( 'gtin', new Assert\Regex( '/^\d{8}(?:\d{4,6})?$/' ) );
-		$metadata->addPropertyConstraint( 'mpn', new Assert\Type( 'alnum' ) ); // alphanumeric
+		$metadata->addPropertyConstraint( 'mpn', new Assert\Type( 'string' ) );
 		$metadata->addPropertyConstraint( 'mpn', new Assert\Length( null, 0, 70 ) ); // maximum 70 characters
 
 		$metadata->addPropertyConstraint(


mikkamp (approved):
Looks good, thanks for the changes.

I previously got an error when trying to sync:
```
2022-05-31T07:49:57+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\BatchProductHelper::validate_and_generate_update_request_entries Skipping product (ID: 1433) because it does not pass validation: {"woocommerce_id":1433,"errors":["[mpn] This value should be of type alnum."]}
```

Now it's syncing successfully and it accepts any characters:
image

Regarding the tests, I think we've mentioned it before that it's not worth it to retest functionality from a library that's already tested. So it should be fine to leave as is.

PR #1152, opened 2021-12-13
Add CES prompts for initial setup and campaign creation

ianlin:
### Changes proposed in this Pull Request:

Closes #882.

This PR implements two CES prompts that will displayed when the user

**Prompt 1**: successfully completes initial setup flow and enters the product feed page for the first time
**Prompt 2**: successfully completes Ads creation flow and clicks "Got it" in the success modal.

#### Prompt 1

This can be split into two requirements:

1. Display it when the user enters product feed for the first time after setting up
2. Don't display it unless the success modal is not shown. I.e. dismiss the success modal will display the CES prompt

When the user sees the success modal, there are different paths that they can go:

1. Dismiss the success modal by clicking `X` button, clicking anywhere outside of the modal, or clicking `Maybe later` button.
2. Go to the product feed page in a new tab by clicking `Manage and edit your product feed in WooCommerce`.
3. Go to the page to create paid campaign, and they will be redirected to the dashboard after finish creating campaign.
  a. Eventually the user will go to the product feed page by clicking it, so the behaviour will be just like the point 2 above

In oder to fulfil the requirements above, we need to store a flag `gla-can-onboarding-setup-ces-prompt-open` in the local storage to help us to decide when can we display the CES prompt. The flows will be like this:

**Flow 1** ([Video 1](#video-1))
1. User sees the success modal
2. At the mean time the CES prompt will not be displayed since the success modal is shown
3. The success modal set `gla-can-onboarding-setup-ces-prompt-open` to `true` (actually any value is fine)
4. User dismisses the success modal, then the CES prompt is shown
5. The CES prompt will delete the flag `gla-can-onboarding-setup-ces-prompt-open` from the local storage

**Flow 2** ([Video 2](#video-2))
1. User sees the success modal
2. The success modal set `gla-can-onboarding-setup-ces-prompt-open` to `true`
3. User go to the product feed page in a new tab
4. In the new tab there are no success modal and the flag is in the local storage, the CES prompt is shown
5. The CES prompt deletes  `gla-can-onboarding-setup-ces-prompt-open`

#### Prompt 2

Prompt 2 is simpler, the prompt will only be shown when the user clicks on "Got it" of the campaign creation success modal, so we don't need the flag in local storage.

([Video 3](#video-3))
([Video 4](#video-4))

#### <a href="#video-1" id="video-1">Video 1</a> - Dismiss success modal and show the CES prompt

[...]

PR #1073, opened 2021-11-02
Avoid including a nodejs library to reduce the index.js bundle size by 42%

eason9487:
### Changes proposed in this Pull Request:

Found this issue when looking into https://github.com/woocommerce/google-listings-and-ads/pull/1068#discussion_r740796135.

- Specify the `@woocommerce/data` version to **1.2.0** to avoid bundling the nodejs `crypto` into index.js
- Rebalance the max size config of bundlewatch to 1.2 times 220.5 kB for index.js

#### Root cause

When looking into the result of bundle analysis, I notice that the obvious `bn.js` appears several times and takes up a significant file size. After some investigations, it turns out the root cause is that [`@woocommerce/data` 1.1.1 imports the nodejs built-in library `crypto`](https://github.com/woocommerce/woocommerce-admin/blob/v1.6.0/packages/data/src/export/utils.js#L4), and it doesn't set the `sideEffects` option to enable the more advanced tree shaking. Therefore, all its dependencies are included into index.js, the GLA doesn't need it though.

The overall unnecessary codes are highlighted by red background:

#### Solution

Fortunately, the issue has been addressed by https://github.com/woocommerce/woocommerce-admin/pull/5768. And we have `@woocommerce/data` relevant changes these days:

- https://github.com/woocommerce/google-listings-and-ads/pull/1042
- https://github.com/woocommerce/google-listings-and-ads/pull/1055

So currently, we have three places using the `@woocommerce/data`:
1. A constant `NAMESPACE` for concatenating the API endpoint. [The constant was there two years ago.](https://github.com/woocommerce/woocommerce-admin/blob/v2.2.1/packages/data/src/constants.js#L2)
    https://github.com/woocommerce/google-listings-and-ads/blob/e5cf5bd8b01cc5041b3933a0a90bb10e67f908d7/js/src/reports/products/async-requests.js#L15
2. The `getItems` selector for checking where a product is a variant.
   https://github.com/woocommerce/google-listings-and-ads/blob/e5cf5bd8b01cc5041b3933a0a90bb10e67f908d7/js/src/reports/products/products-report-filters.js#L102-L104
3. The `getSetting` selector for getting store's contry.
    https://github.com/woocommerce/google-listings-and-ads/blob/e5cf5bd8b01cc5041b3933a0a90bb10e67f908d7/js/src/hooks/useStoreCountry.js#L19-L22

After specifying `@woocommerce/data` to version 1.2.0 and checking:

- [x] All relevant usages in GLA are working well
[...]

Issue #882, opened 2021-07-14
Add two CES prompts to get more product feedback 


j111q:
## What is a CES prompt?
It's a little toast that appears on the bottom left of the user's screen, asking the user for feedback on a task they've just completed. [Link to CES GitHub issue](https://github.com/woocommerce/woocommerce-admin/issues/5259)

## Prompt 1
Prompt 1 should appear immediately after User successfully completes initial setup flow.

![Setup1](https://user-images.githubusercontent.com/65582817/125259488-efbd5f00-e331-11eb-9db4-d56b2d2ebf34.png)

## Prompt 2
Prompt 2 should appear immediately after User successfully completes Ads creation flow and clicks "Got it" in the success modal.

![image](https://user-images.githubusercontent.com/65582817/125585386-2f3042c6-0cb2-44b8-a541-030cc9a17c0a.png)

## Figma link
[URL](https://mc.a8c.com/pb/29be3/)

eason9487:
Hey @j111q, I have two questions:

1. For **prompt 1**, after completing the setup, it will show [the success modal](https://github.com/woocommerce/google-listings-and-ads/pull/632) first. And user might 1) dismiss/close the modal, 2) open the product feed page in a new tab, 3) browse to the ads setup page. If a user goes the 3) path, then they would run into **prompt 2** after finishing the ads setup continuously without going through the product feed page. So, the question is about the display timing of toast, which one is expected for **prompt 1**?
    - Show toast when the first time user entering the product page regardless of the paths they were going through
    - Show toast after the user 1) dismissing/closing the modal or 2) opening a new tab, but it would be bypassed if they 3) proceed to set up ads
    - Show toast with the modal at the same time
    - Others?
2. The survey data of CES implementation in WC-admin was recorded via event tracking. Are we going to record in the same way? If yes, should the two prompts show up only when the event tracking option is checked off?

eason9487:
Dev note

- Library: https://github.com/woocommerce/woocommerce-admin/tree/main/packages/customer-effort-score
- Usage in WC-admin: https://github.com/woocommerce/woocommerce-admin/blob/38fbb4aedd0f21cf21ddaa6af93e48564fd35324/client/customer-effort-score-tracks/customer-effort-score-tracks.js#L107-L135

j111q:
> Show toast when the first time user entering the product page regardless of the paths they were going through

This is the one we should go with, thank you for checking!! 🙏 

[...]

Issue #555, opened 2021-05-06
Changes to Success Modal after first setup


j111q:
## User story
After the user clicks `Complete setup`, a success modal appears:

![Submission success - first submit a](https://user-images.githubusercontent.com/65582817/117238086-d8a84f80-ae5e-11eb-929e-1cbf88ba23a9.png)
![Submission success - first submit b](https://user-images.githubusercontent.com/65582817/117238091-dba34000-ae5e-11eb-9b57-0193f2f0ab15.png)
![image](https://user-images.githubusercontent.com/65582817/117239120-c4fde880-ae60-11eb-9cfb-69b3d4a3f033.png)
![image](https://user-images.githubusercontent.com/65582817/117239137-cb8c6000-ae60-11eb-8f53-10674f659de0.png)

## Figma link
https://www.figma.com/file/jZUpa8eTrnrK1Lwt2ry7zk/Native-Google-Integration?node-id=1853%3A2413

PR #1543, opened 2022-05-31
Add useEffect to avoid multiple CES prompts on the Dashboard Page

### LINKED ISSUE #1428: CES Prompts Showing several times
### Describe the bug:

While I was testing other PR I noticed the CES Prompt "How easy was it to create a Google Ad campaign?" (maybe some other Prompts as well) are being shown multiple times.

In my opinion, if the Prompt is already there, we should avoid showing it again. 

https://user-images.githubusercontent.com/5908855/163542330-a2ae0723-7911-4bae-9d3e-8b84de001330.mov

### Steps to reproduce (actual behaviour):

1. WP Admin as well as Analytics Feature should be enabled 
2. Edit a campaign and save
3. You will see the prompt "How easy was it to create a Google Ad campaign?" 
4. Dont do anything and just go to Edit a campaign and save again
5. You will see another prompt "How easy was it to create a Google Ad campaign?" so 2 prompts in total.

### Expected behavior:

1. WP Admin as well as Analytics Feature should be enabled 
2. Edit a campaign and save
3. You will see the prompt "How easy was it to create a Google Ad campaign?" 
4. Dont do anything and just go to Edit a campaign and save again
5. You WONT see another prompt "How easy was it to create a Google Ad campaign?" so 1 prompt in total.

jorgemd24:

Closes #1428 .

This PR fixes the issue with the CES prompts in the dashboard page after the campaign creation success modal is shown. The issue was caused because the status `isCESPromptOpen` stays always `true` when navigating through the dashboard sub-components such as `EditFreeCampaign`, `EditPaidAdsCampaign` and `CreatePaidAdsCampaign`, therefore dispatching a new CustomerEffortScore notice when returns to the dashboard page.

https://github.com/woocommerce/google-listings-and-ads/blob/74949bc72a41ff79d47138c9e319739ff4ba36aa/js/src/dashboard/index.js#L35

### Before this PR

https://user-images.githubusercontent.com/2488994/171261339-72fdff1a-c1da-4e02-ab9a-63ef19bd2022.mp4

### After this PR

https://user-images.githubusercontent.com/2488994/171262126-d2165848-1dca-47d7-9518-6a5bb476de5e.mov

1. Go to `wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&guide=campaign-creation-success`
2. Click in "Got it"
3. One CES prompt should be displayed.
4. Visit Edit free listings, Edit Ads Campaigns and Create Ads campaigns and return to the dashboard page and it should show only one CES prompt.

### Changelog entry

> Fix - Multiple CES prompts in the Dashboard Page

### DIFF
--- a/js/src/dashboard/index.js
+++ b/js/src/dashboard/index.js
@@ -26,14 +26,12 @@ import EditFreeCampaign from '.~/edit-free-campaign';
 import EditPaidAdsCampaign from '.~/pages/edit-paid-ads-campaign';
 import CreatePaidAdsCampaign from '.~/pages/create-paid-ads-campaign';
 import { CTA_CREATE_ANOTHER_CAMPAIGN, CTA_CONFIRM } from './constants';
-import useUrlQuery from '.~/hooks/useUrlQuery';
 
 /**
  * @fires gla_modal_closed when CES modal is closed.
  */
 const Dashboard = () => {
 	const [ isCESPromptOpen, setCESPromptOpen ] = useState( false );
-	const query = useUrlQuery();
 
 	const handleCampaignCreationSuccessGuideClose = useCallback(
 		( e, specifiedAction ) => {
@@ -58,6 +56,7 @@ const Dashboard = () => {
 		[ setCESPromptOpen ]
 	);
 
+	const query = getQuery();
 	switch ( query.subpath ) {
 		case subpaths.editFreeListings:
 			return <EditFreeCampaign />;
--- a/js/src/dashboard/index.js
+++ b/js/src/dashboard/index.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { Button } from '@wordpress/components';
-import { useState, useCallback, useEffect } from '@wordpress/element';
+import { useState, useCallback } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { Link } from '@woocommerce/components';
 import { getNewPath, getQuery, getHistory } from '@woocommerce/navigation';
@@ -34,10 +34,6 @@ import useUrlQuery from '.~/hooks/useUrlQuery';
 const Dashboard = () => {
 	const [ isCESPromptOpen, setCESPromptOpen ] = useState( false );
 	const query = useUrlQuery();
-
-	useEffect( () => {
-		setCESPromptOpen( false );
-	}, [ query.subpath ] );
 
 	const handleCampaignCreationSuccessGuideClose = useCallback(
 		( e, specifiedAction ) => {
--- a/js/src/dashboard/index.js
+++ b/js/src/dashboard/index.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { Button } from '@wordpress/components';
-import { useState, useCallback } from '@wordpress/element';
+import { useState, useCallback, useEffect } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { Link } from '@woocommerce/components';
 import { getNewPath, getQuery, getHistory } from '@woocommerce/navigation';
@@ -26,12 +26,18 @@ import EditFreeCampaign from '.~/edit-free-campaign';
 import EditPaidAdsCampaign from '.~/pages/edit-paid-ads-campaign';
 import CreatePaidAdsCampaign from '.~/pages/create-paid-ads-campaign';
 import { CTA_CREATE_ANOTHER_CAMPAIGN, CTA_CONFIRM } from './constants';
+import useUrlQuery from '.~/hooks/useUrlQuery';
 
 /**
  * @fires gla_modal_closed when CES modal is closed.
  */
 const Dashboard = () => {
 	const [ isCESPromptOpen, setCESPromptOpen ] = useState( false );
+	const query = useUrlQuery();
+
+	useEffect( () => {
+		setCESPromptOpen( false );
+	}, [ query.subpath ] );
 
 	const handleCampaignCreationSuccessGuideClose = useCallback(
 		( e, specifiedAction ) => {
@@ -56,7 +62,6 @@ const Dashboard = () => {
 		[ setCESPromptOpen ]
 	);
 
-	const query = getQuery();
 	switch ( query.subpath ) {
 		case subpaths.editFreeListings:
 			return <EditFreeCampaign />;
--- a/js/src/hooks/useCESNotice.js
+++ b/js/src/hooks/useCESNotice.js
@@ -1,0 +1,41 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { noop } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import useUnmountableNotice from '.~/hooks/useUnmountableNotice';
+
+/**
+ * Hook to create a CES notice
+ *
+ * @param {string} label CES prompt label.
+ * @param {JSX.Element} icon Icon (React component) to be shown on the notice.
+ * @param {Function} [onClickCallBack] Function to call when Give feedback is clicked.
+ * @param {Function} [onNoticeDismissedCallback] Function to call when the notice is dismissed.
+ *
+ * @return {Function} a function that will create the CES notice.
+ */
+const useCESNotice = (
+	label,
+	icon,
+	onClickCallBack = noop,
+	onNoticeDismissedCallback = noop
+) => {
+	return useUnmountableNotice( 'success', label, {
+		actions: [
+			{
+				label: __( 'Give feedback', 'woocommerce-admin' ),
+				onClick: onClickCallBack,
+			},
+		],
+		icon,
+		explicitDismiss: true,
+		onDismiss: onNoticeDismissedCallback,
+	} );
+};
+
+export default useCESNotice;
--- a/js/src/hooks/useCESNotice.js
+++ b/js/src/hooks/useCESNotice.js
@@ -28,7 +28,7 @@ const useCESNotice = (
 	return useUnmountableNotice( 'success', label, {
 		actions: [
 			{
-				label: __( 'Give feedback', 'woocommerce-admin' ),
+				label: __( 'Give feedback', 'google-listings-and-ads' ),
 				onClick: onClickCallBack,
 			},
 		],
--- a/js/src/components/customer-effort-score-prompt/customer-effort-unmountable-notice.js
+++ b/js/src/components/customer-effort-score-prompt/customer-effort-unmountable-notice.js
@@ -15,9 +15,9 @@ import useCESNotice from '.~/hooks/useCESNotice';
  *
  * ## Motivation
  *
- * The CustomerEffortScore component include in the package @woocommerce/customer-effort-score does not remove the notice
+ * The CustomerEffortScore component included in the package @woocommerce/customer-effort-score does not remove the notice
  * when the component has been unmounted, throwing an error when the "Give Feedback" button has been clicked. This new
- * component is removing the notice if the CES component does not exist.
+ * component removes the notice if the CES component does not exist.
  *
  * @param {Object} props                               Component props.
  * @param {string} props.label                         The label displayed in the modal.
--- a/js/src/components/customer-effort-score-prompt/customer-effort-unmountable-notice.js
+++ b/js/src/components/customer-effort-score-prompt/customer-effort-unmountable-notice.js
@@ -1,0 +1,74 @@
+/**
+ * External dependencies
+ */
+import { useEffect, useState } from '@wordpress/element';
+import CustomerFeedbackModal from '@woocommerce/customer-effort-score/build/customer-feedback-modal';
+import { noop } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import useCESNotice from '.~/hooks/useCESNotice';
+
+/**
+ * Renders an unmountable CES notice to gather a customer effort score.
+ *
+ * ## Motivation
+ *
+ * The CustomerEffortScore component include in the package @woocommerce/customer-effort-score does not remove the notice
+ * when the component has been unmounted, throwing an error when the "Give Feedback" button has been clicked. This new
+ * component is removing the notice if the CES component does not exist.
+ *
+ * @param {Object} props                               Component props.
+ * @param {string} props.label                         The label displayed in the modal.
+ * @param {JSX.Element} props.icon                     Icon (React component) to be shown on the notice.
+ * @param {Function} [props.recordScoreCallback]       Function to call when the score should be recorded.
+ * @param {Function} [props.onNoticeShownCallback]     Function to call when the notice is shown.
+ * @param {Function} [props.onNoticeDismissedCallback] Function to call when the notice is dismissed.
+ * @param {Function} [props.onModalShownCallback]      Function to call when the modal is shown.
+ */
+const CustomerEffortScoreUnmountableNotice = ( {
+	label,
+	icon,
+	recordScoreCallback = noop,
+	onNoticeShownCallback = noop,
+	onNoticeDismissedCallback = noop,
+	onModalShownCallback = noop,
+} ) => {
+	const [ shouldCreateNotice, setShouldCreateNotice ] = useState( true );
+	const [ visible, setVisible ] = useState( false );
+	const onClickFeedBack = () => {
+		setVisible( true );
+		onModalShownCallback();
+	};
+
+	const createCESNotice = useCESNotice(
+		label,
+		icon,
+		onClickFeedBack,
+		onNoticeDismissedCallback
+	);
+
+	useEffect( () => {
+		if ( ! shouldCreateNotice ) {
+			return;
+		}
+
+		createCESNotice();
+		setShouldCreateNotice( false );
+		onNoticeShownCallback();
+	}, [ shouldCreateNotice, createCESNotice, onNoticeShownCallback ] );
+
+	if ( shouldCreateNotice || ! visible ) {
+		return null;
+	}
+
+	return (
+		<CustomerFeedbackModal
+			label={ label }
+			recordScoreCallback={ recordScoreCallback }
+		/>
+	);
+};
+
+export default CustomerEffortScoreUnmountableNotice;
--- a/js/src/components/customer-effort-score-prompt/customer-effort-unmountable-notice.test.js
+++ b/js/src/components/customer-effort-score-prompt/customer-effort-unmountable-notice.test.js
@@ -1,0 +1,99 @@
+/**
+ * External dependencies
+ */
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+/**
+ * Internal dependencies
+ */
+import useCESNotice from '.~/hooks/useCESNotice';
+import CustomerEffortScoreUnmountableNotice from '.~/components/customer-effort-score-prompt/customer-effort-unmountable-notice';
+
+const mockCreateCESNoticeShowModal = jest
+	.fn()
+	.mockImplementation( ( onClickCallBack ) => onClickCallBack() );
+
+const mockCreateCESNotice = jest.fn();
+
+jest.mock( '.~/hooks/useCESNotice', () => {
+	return jest.fn();
+} );
+
+describe( 'Customer Effort Notice component', () => {
+	let defaultProps;
+
+	beforeEach( () => {
+		defaultProps = {
+			label: 'label test',
+			icon: <span>my icon</span>,
+			recordScoreCallback: jest.fn(),
+			onNoticeDismissedCallback: jest.fn(),
+			onNoticeShownCallback: jest.fn(),
+			onModalShownCallback: jest.fn(),
+		};
+	} );
+
+	test( 'Rendering the CES Notice without the CES modal', async () => {
+		useCESNotice.mockImplementation( () => mockCreateCESNotice );
+
+		const { queryByText } = render(
+			<CustomerEffortScoreUnmountableNotice { ...defaultProps } />
+		);
+
+		//Creates the notice
+		expect( mockCreateCESNotice ).toHaveBeenCalledTimes( 1 );
+		expect( defaultProps.onNoticeShownCallback ).toHaveBeenCalledTimes( 1 );
+		expect( defaultProps.onModalShownCallback ).toHaveBeenCalledTimes( 0 );
+
+		expect( useCESNotice ).toHaveBeenLastCalledWith(
+			defaultProps.label,
+			defaultProps.icon,
+			expect.any( Function ),
+			defaultProps.onNoticeDismissedCallback
+		);
+
+		//The modal should not be displayed
+		expect(
+			queryByText( 'Please share your feedback' )
+		).not.toBeInTheDocument();
+
+		expect( defaultProps.recordScoreCallback ).toHaveBeenCalledTimes( 0 );
+	} );
+
+	test( 'Rendering the CES Notice with the CES modal', async () => {
+		useCESNotice.mockImplementation(
+			( label, icon, onClickCallBack ) => () =>
+				mockCreateCESNoticeShowModal( onClickCallBack )
+		);
+
+		const { getByText } = render(
+			<CustomerEffortScoreUnmountableNotice { ...defaultProps } />
+		);
+
+		//Creates the notice
+		expect( mockCreateCESNotice ).toHaveBeenCalledTimes( 1 );
+		expect( defaultProps.onModalShownCallback ).toHaveBeenCalledTimes( 1 );
+		expect( useCESNotice ).toHaveBeenLastCalledWith(
+			defaultProps.label,
+			defaultProps.icon,
+			expect.any( Function ),
+			defaultProps.onNoticeDismissedCallback
+		);
+
+		//The modal should be displayed
+		expect( getByText( 'Please share your feedback' ) ).toBeInTheDocument();
+
+		//record score
+		const score = getByText( 'Very easy' );
+		const send = getByText( 'Send' );
+
+		expect( score ).toBeInTheDocument();
+		expect( send ).toBeInTheDocument();
+
+		userEvent.click( score );
+		userEvent.click( send );
+		expect( defaultProps.recordScoreCallback ).toHaveBeenCalledTimes( 1 );
+	} );
+} );
--- a/js/src/components/customer-effort-score-prompt/index.js
+++ b/js/src/components/customer-effort-score-prompt/index.js
@@ -2,7 +2,6 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import CustomerEffortScore from '@woocommerce/customer-effort-score';
 import { recordEvent } from '@woocommerce/tracks';
 
 /**
@@ -10,6 +9,7 @@ import { recordEvent } from '@woocommerce/tracks';
  */
 import { LOCAL_STORAGE_KEYS } from '.~/constants';
 import localStorage from '.~/utils/localStorage';
+import CustomerEffortScoreUnmountableNotice from './customer-effort-unmountable-notice';
 
 /**
  * CES prompt snackbar open
@@ -81,7 +81,7 @@ const CustomerEffortScorePrompt = ( { eventContext, label } ) => {
 	};
 
 	return (
-		<CustomerEffortScore
+		<CustomerEffortScoreUnmountableNotice
 			label={ label }
 			recordScoreCallback={ recordScore }
 			onNoticeShownCallback={ onNoticeShown }
--- a/js/src/hooks/useCESNotice.test.js
+++ b/js/src/hooks/useCESNotice.test.js
@@ -1,0 +1,52 @@
+/**
+ * External dependencies
+ */
+import { renderHook } from '@testing-library/react-hooks';
+
+/**
+ * Internal dependencies
+ */
+import useCESNotice from './useCESNotice';
+import useUnmountableNotice from './useUnmountableNotice';
+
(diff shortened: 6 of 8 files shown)


js/src/dashboard/index.js
```diff
@@ -26,12 +26,18 @@ import EditFreeCampaign from '.~/edit-free-campaign';
...
 /**
  * @fires gla_modal_closed when CES modal is closed.
  */
 const Dashboard = () => {
 	const [ isCESPromptOpen, setCESPromptOpen ] = useState( false );
+	const query = useUrlQuery();
+
+	useEffect( () => {
+		setCESPromptOpen( false );
+	}, [ query.subpath ] );
```
ecgan: 💅 Can we add in a code comment to explain why this is needed? Something like: "When users navigate into one of the subpaths (edit free listings, edit paid ads, and create paid ads), we call setCESPromptOpen( false ) via effect, so that when users come back into dashboard, the CustomerEffortScorePrompt component will not be rendered and will not trigger another notice. There will be only one notice throughout."

ecgan (approved):
Tested the code before and after the PR code change, I can reproduce the issue, and the code change does fix the issue. 

I left one 💅 comment below, a suggestion to add code comment to explain why the `useEffect` is needed.

A note for the future: I think the fix here is more like a quick fix workaround. For a long term better solution, I think we should have an AppShell that contains the whole plugin, and things like the `CustomerEffortScorePrompt`, `DifferentCurrencyNotice` and `NavigationClassic` components should be in the AppShell, so that there will only be one instance of them throughout the whole app. With that approach, we wouldn't need to have this `useEffect` call. This approach would also solve the issue where the notice appears to keep re-appearing when we switch tabs, see demo video of the issue below:

https://user-images.githubusercontent.com/417342/171477682-b656fcda-6216-4d2d-b61d-cc585de4127c.mov

I also looked into the `CustomerEffortScorePrompt`, and the more I look into it, the deeper rabbit hole it gets, and the more questions I have: e.g. 

- Why do we need the use of `localStorage`? It is introduced in PR https://github.com/woocommerce/google-listings-and-ads/pull/1152.
- I found the reason in commit [`0ff1b0f` (#1152)](https://github.com/woocommerce/google-listings-and-ads/pull/1152/commits/0ff1b0fa66812894b25c08662c30e753f39c6c7e), and it is because we want to have control of state between two window tabs.
- Based on the commit message, I tested around and it gives me another question: we are already in product feed page, why do we need to open a new tab that shows product feed page again? I'm thinking it is because we want to retain the modal in the first tab so that users can click on the next button, but from the user experience perspective, it feels weird (and maybe even appear buggy) to open a new tab showing the same page. See demo video below: 

https://user-images.githubusercontent.com/417342/171481350-5ec54228-978a-4b9a-8b8a-c59d02c9a54e.mov

In the end, since the code here fixes the issue, I'm approving this PR. I think it would be good to get a second review from @ianlin too, since he worked on the PR #1152 above, he might have a bit more insight and ideas that I miss. 😄 🙏

ianlin:
Thanks @ecgan for the detailed testing!

@jorgemd24 I just tested it and found the CES prompt will become nonfunctional after going to the subpath. However it's not because of the changes in your PR, it's an existing but undiscovered problem in my PR #1152. Check the video:

https://user-images.githubusercontent.com/914406/171572446-a5158e98-301f-4fef-944f-69cfb0eca4f4.mov

I don't have a solution in my head at the moment, if you think it's fun to solve it please go ahead otherwise I'd try to solve it after my sprint tasks are done as it's a bug I created 🐛 😅.

@ecgan 
> I'm thinking it is because we want to retain the modal in the first tab so that users can click on the next button, but from the user experience perspective, it feels weird (and maybe even appear buggy) to open a new tab showing the same page. 

I searched a bit from the previous commits and found it's a requirement in #555, quotes from @j111q:

> This link opens the underlying Product Feed page for Google Listings & Ads in a new tab. 

> I know that’s a little weird, but we’re trying to achieve two things here - people’s desire to manage their feed AND our objective to get them to create paid campaigns in the next screen. 

> So if they do want to edit their product feed, they have it in another tab.

jorgemd24:
@ecgan,

Thanks for your suggestions 🙂 , related to:

>A note for the future: I think the fix here is more like a quick fix workaround. For a long term better solution, I think we should have an AppShell that contains the whole plugin, and things like the CustomerEffortScorePrompt, DifferentCurrencyNotice and NavigationClassic components should be in the AppShell, so that there will only be one instance of them throughout the whole app. With that approach, we wouldn't need to have this useEffect call. 

I agree with you, the problem is because every time that CustomerEffortScorePrompt is rendered is dispatching a new Notice, using the AppShell, the dispatch action would be triggered only once. 

Package: `@woocommerce/customer-effort-score`
```
export default compose(
	withDispatch( ( dispatch ) => {
		const { createNotice } = dispatch( 'core/notices2' );

		return {
			createNotice,
		};
	} )
)( CustomerEffortScore );
```
I would suggest creating a separate issue where we can discuss how to implement the AppShell and which components could get the advantage of that improvement. Also, it would be convenient to see how other plugins are implementing this. 

>This approach would also solve the issue where the notice appears to keep re-appearing when we switch tabs, see a demo video of the issue below:

That issue is affecting others' notices, I am not sure if this is a "bug" in our app or if it is something more related to `core/notices`. See video (I intentionally blocked the ads/campaigns endpoint so we can get the prompt):


https://user-images.githubusercontent.com/2488994/171616024-3fb3352b-7dd3-4b10-b75e-cbdd2a3b15c1.mp4




@ianlin 

>I don't have a solution in my head at the moment, if you think it's fun to solve it please go ahead otherwise I'd try to solve it after my sprint tasks are done as it's a bug I created 🐛 😅.

Thanks for your comments, I think I know what is the problem:

1. The `CustomerEffortScorePrompt` component is mounted.
2. The CES prompt is dispatched.
3. 'Give feedback' has two actions linked to the `CustomerEffortScore` component:   setVisible(true) and                   onModalShownCallback().

Package: `@woocommerce/customer-effort-score`
```
	useEffect( () => {
		if ( ! shouldCreateNotice ) {
			return;
		}

		createNotice( 'success', label, {
			actions: [
				{
					label: __( 'Give feedback', 'woocommerce-admin' ),
					onClick: () => {
						setVisible( true );
						onModalShownCallback();
					},
				},
			],
			icon,
			explicitDismiss: true,
			onDismiss: onNoticeDismissedCallback,
		} );

		setShouldCreateNotice( false );

		onNoticeShownCallback();
	}, [ shouldCreateNotice ] );
```

3. The IU changes to a different page, for example, editing Campaign ads.
4. `CustomerEffortScorePrompt` and `CustomerEffortScore` are unmounted . 
4. The prompt will still show but is missing its `CustomerEffortScore` component because was `unmounted` 
5. Go back to the dashboard, a new `CustomerEffortScorePrompt` is mounted, 'Give feedback' works but the previous CES prompt does not work because does not have any `CustomerEffortScore` component associated with it (it was unmounted before). 

This issue will affect the current solution too, I will see how we can fix this as well. 

Let me know your thoughts!

ecgan:
@jorgemd24 ,

> I would suggest creating a separate issue where we can discuss how to implement the AppShell and which components could get the advantage of that improvement.

Yup, I have the same thought too. I have created issue https://github.com/woocommerce/google-listings-and-ads/issues/1548 for this. It is an "early proposal" issue, feel free to refine the issue description and add more things into it. 😄 

> I think I know what is the problem: [...]
> * The prompt will still show but is missing its `CustomerEffortScore` component because was `unmounted`

I haven't looked detailed into the code there yet, but I have a feeling that `@woocommerce/customer-effort-score` has a leaky abstraction. It is nice to have a `<CustomerEffortScorePrompt>` React component that will display a notice when we render the component, but if we use the "React way of doing things", when `<CustomerEffortScorePrompt>` is unmounted / not rendered, the notice should be dismissed / not displayed, too. Otherwise, it gives us this sort of problems because we are mixing React and non-React way of doing things.

An idea to address the above is we can probably come up with a wrapper around `CustomerEffortScore`, so that it works like React way. When it works good enough for us, we can contribute it upstream into `@woocommerce/customer-effort-score` package.

jorgemd24:
@ecgan,

> I haven't looked detailed into the code there yet, but I have a feeling that @woocommerce/customer-effort-score has a leaky abstraction. It is nice to have a <CustomerEffortScorePrompt> React component that will display a notice when we render the component, but if we use the "React way of doing things", when <CustomerEffortScorePrompt> is unmounted / not rendered, the notice should be dismissed / not displayed, too

Yes, currently (using develop/trunk) the prompt is throwing an error when you are changing the page and clicking on "Give feedback".

```
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

See the following videos:

https://user-images.githubusercontent.com/2488994/172468682-d2ab6435-e495-4e34-89f1-78f4d8cf3538.mp4


https://user-images.githubusercontent.com/2488994/172468307-cc176a25-2823-430a-b080-00b5db8b8b78.mp4



@ianlin 
Does the prompt need to be shown on every page? or only in the dashboard/product feed page?


I have been playing with `CustomerEffortScore`component from `@woocommerce/customer-effort-score`, and I think the easiest way to have the `CustomerEffortScore` working as "React", so when `CustomerEffortScore` is unmounted, the notice dissapered too, it is the following:

```
import { dispatch } from '@wordpress/data';
import { useEffect, useRef } from '@wordpress/element';
import { CustomerEffortScore } from '@woocommerce/customer-effort-score';
import { uniqueId } from 'lodash';

const CustomerEffortScoreWithCustomNotice = ( {
	customOptions = {},
	...rest
} ) => {

        // core/notices2 was used in @woocommerce/customer-effort-score so I left it as it was.
	const { createNotice, removeNotice } = dispatch( 'core/notices2' );

	const idRef = useRef( customOptions.id || uniqueId() );

	//remove notice when the component is unmounted
	useEffect( () => () => removeNotice( idRef.current ), [] );

	return (
		<CustomerEffortScore
			{ ...rest }
			createNotice={ ( type, label, options ) =>
				createNotice( type, label, {
					...options,
					...customOptions,
					id: idRef.current,
				} )
			}
		/>
	);
};
```

1. When `CustomerEffortScorePrompt` is unmounted, we could use [removeNotice](https://github.com/WordPress/gutenberg/blob/9efab2cbe695ff831c0c8b099b3193629c4989a7/packages/notices/src/store/actions.js#L159-L165), to dismiss the prompt/snackbar.
2. For this we need the notice ID, unfortunately `<CustomerEffortScore>` component creates the ID randomly and it does not give the option to set the ID externally. 
3. As the default `<CustomerEffortScore>` component does not give the option to set a custom ID, we create a custom notice so we are able to dismiss the notice if the component is unmounted.

The behaviour will be similar than it was before, except that the prompt will be only visible if it is mounted in the current view, so if the user changes the page, the prompt will be gone as the component it is not mounted anymore.

What do you think?

ianlin:
@jorgemd24 

> Does the prompt need to be shown on every page? or only in the dashboard/product feed page?

Hmm, that's a good question. I've just read through the original feature request in #882 and still cannot have a conclusion about it. Basically the requirements are:

### CES prompt for setting up GLA
- When the user goes to the product page for the first time, it will show the success modal. When the success modal is closed, the CES prompt will show up.

### CES prompt for creating a Google Ad campaign
- When the user create an Ad campaign, it will show the campaign creation success modal. When the user clicks `Got it` on the success modal it will be closed and the CES prompt will show up.

In my view if the CES prompt shows up and the user did not click on it and they go to other pages instead, the CES prompt should be closed automatically. Simply because I think the user doesn't care about the prompt and would pretty likely to just click the close (X) button of it.

@j111q What do you think? Do both prompts need to be shown on every page, or only in the dashboard / product feed pages?

ecgan:
@jorgemd24 ,

> I think the easiest way to have the `CustomerEffortScore` working as "React", so when `CustomerEffortScore` is unmounted, the notice dissapered too, it is the following: [...]

In your code example, we provide `createNotice` prop to `CustomerEffortScore` component, but there isn't a `createNotice` prop? 🤔  Source: https://github.com/woocommerce/woocommerce/blob/be15a3503837dabe74d78407810ed426f819b3ec/packages/js/customer-effort-score/src/customer-effort-score.tsx#L25-L46

I think your idea there is pretty much in the right direction, as in we use our own `createNotice` and `removeNotice`. I'm thinking, instead of using `CustomerEffortScore` component, we use the [`CustomerFeedbackModal` component](https://github.com/woocommerce/woocommerce/blob/5db5c8b110412556806b54b34ecd8ba513db1387/packages/js/customer-effort-score/src/customer-feedback-modal/index.tsx#L32), like [how it is used in `CustomerEffortScore` component](https://github.com/woocommerce/woocommerce/blob/be15a3503837dabe74d78407810ed426f819b3ec/packages/js/customer-effort-score/src/customer-effort-score.tsx#L84-L89), but without the problem of `createNotice` and `useEffect` in `CustomerEffortScore`.

---

@ianlin ,

> > Does the prompt need to be shown on every page? or only in the dashboard/product feed page?
>
> In my view if the CES prompt shows up and the user did not click on it and they go to other pages instead, the CES prompt should be closed automatically. _Simply because I think the user doesn't care about the prompt and would pretty likely to just click the close (X) button of it._

I have mixed thoughts about this. On one hand, I think your reason is valid. On the other hand, I also think it may feel buggy, like "I switched to other pages / other tabs, why did it dismiss the notice at the bottom of the page? I wanted to give feedback after looking at the other tabs, and now I can't give feedback anymore because the notice is dismissed when I switched tabs." 🤔 

I'll leave this design decision to @j111q . 😄

jorgemd24:
Thanks @ecgan 

>In your code example, we provide createNotice prop to CustomerEffortScore component, but there isn't a createNotice prop? 🤔 Source: https://github.com/woocommerce/woocommerce/blob/be15a3503837dabe74d78407810ed426f819b3ec/packages/js/customer-effort-score/src/customer-effort-score.tsx#L25-L46

Ah! It seems that createNotice prop has been removed when it was migrated from JS to TS. See [PR 33110](https://github.com/woocommerce/woocommerce/pull/33110/files) 

In [the previous versions](https://github.com/woocommerce/woocommerce-admin/blob/661dd8d181cacca9a5c29ed26b0fd6224c83b85c/packages/customer-effort-score/src/index.js#L26) and, in my local package, the CustomerEffortScore has a createNotice prop.

>I think your idea there is pretty much in the right direction, as in we use our own createNotice and removeNotice. I'm thinking, instead of using CustomerEffortScore component, we use the [CustomerFeedbackModal component](https://github.com/woocommerce/woocommerce/blob/5db5c8b110412556806b54b34ecd8ba513db1387/packages/js/customer-effort-score/src/customer-feedback-modal/index.tsx#L32), like [how it is used in CustomerEffortScore component](https://github.com/woocommerce/woocommerce/blob/be15a3503837dabe74d78407810ed426f819b3ec/packages/js/customer-effort-score/src/customer-effort-score.tsx#L84-L89), but without the problem of createNotice and useEffect in CustomerEffortScore.

Thanks for this, I am working on it and as well for a reusable component to create notices that can be removed when the component is unmount.

>I have mixed thoughts about this. On one hand, I think your reason is valid. On the other hand, I also think it may feel buggy, like "I switched to other pages / other tabs, why did it dismiss the notice at the bottom of the page? I wanted to give feedback after looking at the other tabs, and now I can't give feedback anymore because the notice is dismissed when I switched tabs."

If we would like to show on every page, we will need to think about how to render the component globally in the app. I am not sure if we already have something similar.

ecgan:
@jorgemd24 ,

> Ah! It seems that createNotice prop has been removed when it was migrated from JS to TS. See [PR 33110](https://github.com/woocommerce/woocommerce/pull/33110/files)

In PR #33110, the `createNotice` prop was already not there in the `CustomerEffortScore` component, the devs removed supplying the prop in test in https://github.com/woocommerce/woocommerce/pull/33110/files#diff-d50cc71cacb845f1663f3426683cb4f488ca1fbeff99185706c218ae77599e56L64-L68.

I looked deeper and the prop was actually removed from the component in https://github.com/woocommerce/woocommerce/pull/32538/files#diff-68a3f6fa9ff5ed6b1718a40d9ca3188b7fd0dba8a816462c94f7a456d5451384R30-R37.

> If we would like to show on every page, we will need to think about how to render the component globally in the app. I am not sure if we already have something similar.

I believe we don't have that. If we want to show on every page nicely, I think we need to have the AppShell implementation (issue https://github.com/woocommerce/google-listings-and-ads/issues/1548).

jorgemd24:
@ecgan, @ianlin 

I have updated the code following the previous discussion: https://github.com/woocommerce/google-listings-and-ads/pull/1543#issuecomment-1149791189.

With these changes we are:
- Removing the prompt when the component is unmounted.
- We are fixing the issue of showing multiple notices
- We are fixing the issue that @ianlin mentioned before: https://github.com/woocommerce/google-listings-and-ads/pull/1543#issuecomment-1144518273

There are some points that need to be discussed for the Dashboard prompt:
- If the prompt is showing and the user goes to a subpath (for example edit campaign), and then it goes back to the dashboard, should we re-render the prompt in the dashboard?
- Should the prompt be shown on every page? Outside of the dashboard/product feed?

The tests are failing, because of the bundle size, I am not sure why this is happening but I think it is related to importing [CustomerFeedbackModal component](https://github.com/woocommerce/woocommerce/blob/5db5c8b110412556806b54b34ecd8ba513db1387/packages/js/customer-effort-score/src/customer-feedback-modal/index.tsx#L32), any idea?

js/src/hooks/useCESNotice.js
```diff
@@ -0,0 +1,41 @@
...
+const useCESNotice = (
+	label,
+	icon,
+	onClickCallBack = noop,
+	onNoticeDismissedCallback = noop
+) => {
+	return useUnmountableNotice( 'success', label, {
+		actions: [
+			{
+				label: __( 'Give feedback', 'woocommerce-admin' ),
```
ianlin: Should it be `google-listings-and-ads` rather than `woocommerce-admin`?

ianlin (commented):
Thanks, it tested as expected on my end: the CES notice will disappear when user goes to a sub-page, and it will show again when the user goes back to the main page, and clicks on the `Give feedback` will show the feedback modal.

For the two question I think we still need @j111q's input:
> - If the prompt is showing and the user goes to a subpath (for example edit campaign), and then it goes back to the dashboard, should we re-render the prompt in the dashboard?
> - Should the prompt be shown on every page? Outside of the dashboard/product feed?

For the bundle size, I also think it's related to importing `'@woocommerce/customer-effort-score/build/customer-feedback-modal'`, but I don't have any clues at the moment 😓.

js/src/hooks/useCESNotice.js
jorgemd24: Thanks @ianlin, it is updated: [f674ca7](https://github.com/woocommerce/google-listings-and-ads/pull/1543/commits/f674ca76ee1dbd9950ad178e319d0f3b3a591b75)

>For the bundle size, I also think it's related to importing '@woocommerce/customer-effort-score/build/customer-feedback-modal', but I don't have any clues at the moment

It is related to @wordpress/components, I think the issue is that @woocommerce/customer-effort-score/build/customer-feedback-modal uses a different version of @wordpress/components, and this increases the bundle size. 

I am looking into possible solutions, maybe @ecgan have seen this issue before.

ecgan:
@jorgemd24 @ianlin ,

> The tests are failing, because of the bundle size, I am not sure why this is happening but I think it is related to importing [CustomerFeedbackModal component](https://github.com/woocommerce/woocommerce/blob/5db5c8b110412556806b54b34ecd8ba513db1387/packages/js/customer-effort-score/src/customer-feedback-modal/index.tsx#L32), any idea?

This is interesting. I looked into the issue and I think I have the explanation for it.

Prior to @jorgemd24's updated code (as mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1543#issuecomment-1152381803), bundle size check was running okay, but now it fails with the updated code.

Before the updated code, we were using import `@woocommerce/customer-effort-score`. In the updated code, we are using `@woocommerce/customer-effort-score/build/customer-feedback-modal`.

In GLA, we are using this https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/dependency-extraction-webpack-plugin webpack plugin (aka. DEWP): 

https://github.com/woocommerce/google-listings-and-ads/blob/29a724b2c4e2dd770c29f12d1035a60ae8263ead/webpack.config.js#L4

This means that when we have a package import that matches the defined list of packages (including `@woocommerce/customer-effort-score`; see https://github.com/woocommerce/woocommerce/blob/1e03b68598ef6e9758d09c61424bf4b730d26fc5/packages/js/dependency-extraction-webpack-plugin/assets/packages.js for the complete list), the package will not be bundled, instead GLA will use the package available in the global scope injected by WC. See example screenshot of the global `wc.customerEffortScore` below:

image

The "before" code was passing bundlesize check because it does not have the `@woocommerce/customer-effort-score` bundled. In the updated code, because `@woocommerce/customer-effort-score/build/customer-feedback-modal` does not match the defined list in DEWP, it is bundled together, causing bundlesize check to fail. We can run `npm run env -- NODE_ENV=production wp-scripts build --webpack-bundle-analyzer` and see that it is bundled in a visualization like below (I learned this from Eason's PR https://github.com/woocommerce/google-listings-and-ads/pull/1073). 

image

After discovering the above, the next question is: why do we have to import from `@woocommerce/customer-effort-score/build/customer-feedback-modal`? Why can't we import from `@woocommerce/customer-effort-score`, which should then solve the bundle size issue?

When I first suggested using `CustomerFeedbackModal` in https://github.com/woocommerce/google-listings-and-ads/pull/1543#issuecomment-1149791189, I looked into the package index.ts file https://github.com/woocommerce/woocommerce/blob/5db5c8b110412556806b54b34ecd8ba513db1387/packages/js/customer-effort-score/src/index.ts, and it does export `CustomerFeedbackModal`.

Now, I looked deeper into it, and I realized that the component export was just done recently (see PR https://github.com/woocommerce/woocommerce/pull/32538/files#diff-22de08ccb210a9adf3a9b8b57c9be9fd58bafb47212e1b64ea0e7c7f7677373b), and it has not been published on npm yet. In npm, the latest version is 2.0.1 published 3 months ago. In GLA, we are using version 1.1.0 published 10 months ago. That's why we don't have the `CustomerFeedbackModal` exported from the package.

So, to solve this issue, I think we would want to get the relevant team to publish a new version for the `@woocommerce/customer-effort-score` package, and then we use that new version in GLA, then everything should work nicely. You can also test by changing `@woocommerce/customer-effort-score` version in GLA package.json and make it point to your local git checkout copy of https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/customer-effort-score directory in your machine (so GLA doesn't use the published npm package) and see if the idea works.

js/src/hooks/useCESNotice.js
ecgan: > I am looking into possible solutions, maybe @ecgan have seen this issue before.

See my comment https://github.com/woocommerce/google-listings-and-ads/pull/1543#issuecomment-1154241021 below.

eason9487:
Hey, I'd like to chip in some suggestions. 🙏

### About the cause of bloat bundle size

The main problem is that `@woocommerce/customer-effort-score` didn't build its ES Module correctly when publishing version 1.1.0.

#### Incorrect build for ES Modules

From the node module directory, we can find it only has the build of CommonJS Module in the `build` directory but no `build-module`. That's why it needs to use `import CustomerFeedbackModal from '@woocommerce/customer-effort-score/build/customer-feedback-modal';` in this PR.



#### DEWP and GLA's webpack config

As @ecgan mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1543#issuecomment-1154241021, `@woocommerce/customer-effort-score/build/customer-feedback-modal` will be bundled since it doesn't match the packages provided by DEWP.

The [dependent packages in \<CustomerFeedbackModal\>](https://github.com/woocommerce/woocommerce-admin/blob/v2.9.4/packages/customer-effort-score/src/customer-feedback-modal/index.tsx#L4-L13) will be provided via DEWP as well if it matches. When webpack resolves the `@wordpress/components`, we state [it should be bundled in GLA wepback config](https://github.com/woocommerce/google-listings-and-ads/blob/f674ca76ee1dbd9950ad178e319d0f3b3a591b75/webpack.config.js#L19). Ideally, it should not lead to the bundle size problem since GLA has already bundled `@wordpress/components` from the beginning.

#### The root cause

Originally, it should use the same bundled `@wordpress/components` as other codes are dependent on, but the problem is the wrong build of the module type. When compiling, **webpack couldn't optimize the bundle size via tree shacking if the package only provides the build of CommonJS Module**. So in the end, all \<CustomerFeedbackModal\> dependencies are bundled into the compiled result. That's the root cause of why its size bloats so much.

Ref: the warning block in https://webpack.js.org/guides/author-libraries/#final-steps

### Possible workarounds

Ideally, the solution will be using the version that officially exports the \<CustomerFeedbackModal\> or asking upstream to make changes to the showing/closing behavior of the notice. But considering the L-2 support policy, we probably need to wait for several months till the L-2 hits the WC 6.7.

So currently, we have two possible workarounds that need a smaller change scope.

#### Workaround A

Create another PR that branches off `develop`, and remove the notice in the \<CustomerEffortScorePrompt\> component. Adding code snippets as below to the component allows us to experiment with the feasibility. And the bundle size of index.js is ≈ 197.6 KB.

```js
useEffect( () => {
  return () => {
    const notice = wp.data
      .select( 'core/notices2' )
      .getNotices()
      .find( ( el ) => el.content === label );

    if ( notice ) {
      wp.data.dispatch( 'core/notices2' ).removeNotice( notice.id );
    }
  };
}, [ label ] );
```

(💡 It's for experiment only. The selector and action of `'core/notices2'` should be used in a similar way to `useUnmountableNotice` in this PR.)

And this PR could be set on hold till a release version of WC exports \<CustomerFeedbackModal\> and hits the L-2 policy. I guess this PR might be reused directly as the better solution then.

#### Workaround B

It's about to release the WC 6.6 in a few days, so the L-2 of WC could be bumped to 6.4, which uses [WC-admin 3.3.2](https://github.com/woocommerce/woocommerce/blob/6.4.0/plugins/woocommerce/composer.json#L24). And `@woocommerce/customer-effort-score` in that WC-admin is [version 2.0.0](https://github.com/woocommerce/woocommerce-admin/blob/v3.3.2/packages/customer-effort-score/package.json#L3), which has fixed the incorrect build of module types.

So we could bump the `@woocommerce/customer-effort-score` to 2.0.0 and use the correct ES Module build in this PR as below. The bundle size of index.js is ≈ 222.0 KB.

```js
import CustomerFeedbackModal from '@woocommerce/customer-effort-score/build-module/customer-feedback-modal';
```

### Suggestion

Comparing the two workarounds, both rely on internal implementation and might be incompatible someday. But the workaround A has a smaller bundle size. IMO, I would suggest the workaround A for now. And we could still have a better solution by holding this PR and waiting for the WC 6.7 to hit the L-2.

j111q:
Sorry for late reply, just reading through this thread ☝️ 

> If the prompt is showing and the user goes to a subpath (for example edit campaign), and then it goes back to the dashboard, should we re-render the prompt in the dashboard?

What does "re-render the prompt" mean? Does it mean, for example, you hide the first existing prompt and show a new second prompt? I don't really have a preference on whether the user is seeing the first prompt or a second prompt -- I think ultimately they should only be seeing one prompt at a time. 

> Does the prompt need to be shown on every page? or only in the dashboard/product feed page?
> >
>> In my view if the CES prompt shows up and the user did not click on it and they go to other pages instead, the CES prompt should be closed automatically. Simply because I think the user doesn't care about the prompt and would pretty likely to just click the close (X) button of it.
> >>
>>> I have mixed thoughts about this. On one hand, I think your reason is valid. On the other hand, I also think it may feel buggy, like "I switched to other pages / other tabs, why did it dismiss the notice at the bottom of the page? I wanted to give feedback after looking at the other tabs, and now I can't give feedback anymore because the notice is dismissed when I switched tabs." 🤔

^ I think it makes sense that the notice doesn't automatically disappear, even if you go to another page. Let's allow the user to close it if they want to close it, otherwise it can stay there. 

> I tested around and it gives me another question: we are already in product feed page, why do we need to open a new tab that shows product feed page again? I'm thinking it is because we want to retain the modal in the first tab so that users can click on the next button, but from the user experience perspective, it feels weird (and maybe even appear buggy) to open a new tab showing the same page.

Happy to share that @dominiquevias is working on a better design here, and in an upcoming iteration, we will NOT be opening the product feed in a new tab from the modal. 😊

jorgemd24:
Thanks, @eason9487 and @ecgan for your comments,

>Workaround A
Create another PR that branches off develop, and remove the notice in the <CustomerEffortScorePrompt> component. Adding code snippets as below to the component allows us to experiment with the feasibility. And the bundle size of index.js is ≈ 197.6 KB.

I agree with you, that for now, this is the best workaround. I am preparing the PR and implementing it. When the customer effort package is updated we can look back to this PR and see which options are better.

@j111q, thanks for your input!

>What does "re-render the prompt" mean? Does it mean, for example, you hide the first existing prompt and show a new second prompt? I don't really have a preference on whether the user is seeing the first prompt or a second prompt -- I think ultimately they should only be seeing one prompt at a time.

The "re-render" is more from the React point of view but means exactly what you explained. The thing is that for now, we do not have a way to show the prompt on every page with its CES modal,  so the prompt will disappear when it goes to a subpath and it will show again if it goes to the Dashboard, this will only happen in the dashboard CES prompt. 

>I think ultimately they should only be seeing one prompt at a time.

Yes, this will be fixed. Before this PR, it was showing duplicated notices.

We are thinking to implement an AppShell (see https://github.com/woocommerce/google-listings-and-ads/issues/1548) to be able to show the same component, on every page in the GLA app without the need to remove and display the component every time that the page changes.

Issue #1545, opened 2022-06-02
Error getting account review status


mikkamp:
### Describe the bug:
We are getting several reports of users getting the error that they can't retrieve their account review status:
- 5256564-zen
- [Forum post](https://wordpress.org/support/topic/there-was-an-error-loading-your-merchant-center-product-review-request-status/)
- [Forum post](https://wordpress.org/support/topic/there-was-an-error-loading-your-merchant-center-product-review-request-status-2/)

In the UI just a generic message is shown, but the log files show a more specific error:
```
2022-05-31T21:27:51+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Middleware::get_account_review_status
Client error: GET https://api-vipgo.woocommerce.com/google/manager/account-review-status/<accound-id> resulted in a 400 Bad Request response:
{“statusCode”:400,”error”:”Bad Request”,”message”:”User cannot access account <account-id>"}
```

The request for the account review status is a proxied request because the API requires privileged access. However this means that we can only request these for child accounts (accounts created through the extension). Based on the error messages, it looks like these are all standalone accounts which have been linked. Which doesn't provide enough access permissions to perform a proxied request.

### Steps to reproduce:
1. Go through the onboarding using a standalone account
2. Once redirected to the product feed page, observe the error that the account review status can't be retrieved
3. Go to WooCommerce > Status > Logs and view the log file to confirm the detailed error

### Expected behavior:
I'd expect the review status / request to be unavailable for users with a standalone account. I'd expect this to happen silently in the background and it shouldn't affect any followup requests or prevent syncing from happening.

### Additional details:
It seems the error is followed by:
```
2022-05-31T21:43:25+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer::validate_merchant_center_setup Cannot sync any products before setting up Google Merchant Center.
```

I'm not able to fully reproduce this scenario but since we are getting a 400 error returned, it looks like it's triggering the reconnect code which means their site will not be ready for syncing products.

ianlin:
@mikkamp Thanks for reporting this issue.

> However this means that we can only request these for child accounts (accounts created through the extension). Based on the error messages, it looks like these are all standalone accounts which have been linked. Which doesn't provide enough access permissions to perform a proxied request.

I'm not quite sure what's a "standalone account", does it mean it's an account created outside of the extension, like creating from the Google merchant centre directly?

[...]

PR #1546, opened 2022-06-02
Do not store URL matches transient until fully connected

mikkamp:

The transient for a matching URL was being retrieved (and updated) before the connected check. Which means that for sites that were not connected it was storing the URL as "non matching" because it wasn't able to retrieve the URL from the account yet. Since it's a transient it is rechecked after 12 hours, but this means that after onboarding the sync would only be able to start 12 hours later.

In this PR we first check if we are connected, and if not it won't check/update the transient to prevent it from being set until onboarding has been fully completed. I also added a test scenario to confirm that the transient is not retrieved when the site is not connected.

Resolves part of #1545

#### Reproduce error with develop branch:
1. Go to the Connection Test page and disconnect the Merchant Account (deletes transients and connection details) 
2. Load the page Marketing > Google Listings & Ads
3. Check the `wp_options` table for the option `_transient_gla_url_matches` and confirm it's set to `no`

#### Confirm it's fixed with this PR:
1. Repeat the steps from above, this time the transient should not be set at all
2. Continue with the onboarding process
3. At the end of the onboarding process the transient should appear and it should be set to `yes`

### Changelog entry
* Fix - Do not store URL matches transient until fully connected.

### DIFF
--- a/src/MerchantCenter/MerchantCenterService.php
+++ b/src/MerchantCenter/MerchantCenterService.php
@@ -104,6 +104,10 @@ class MerchantCenterService implements ContainerAwareInterface, OptionsAwareInte
 	 * @return boolean
 	 */
 	public function is_ready_for_syncing(): bool {
+		if ( ! $this->is_connected() ) {
+			return false;
+		}
+
 		/** @var TransientsInterface $transients */
 		$transients  = $this->container->get( TransientsInterface::class );
 		$url_matches = $transients->get( TransientsInterface::URL_MATCHES );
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Thanks for fixing this!

Tested using your instructions and all passed. Also remove and install the extension then go through the onboarding step, confirmed the transient `_transient_gla_url_matches` is set to `yes` and the product syncing action is scheduled. 👍 

Screenshot 2022-06-03 at 12 15 13

gla-scheduled-actions-all-products

PR #1550, opened 2022-06-03
End to End tests for Gtag Events

mikkamp:

This PR adds End to End tests to confirm that the following Gtag events are sent on the frontend pages:

- Page view
- View item on a single product page
- Add to cart on a single product page
- Add related product to cart from a single product page
- Add to cart from a shop page
- Cart page view
- Conversion on order completed
- Purchase on order completed

1. Confirm the e2e run in this PR without errors 
2. Run the e2e tests locally and confirm they run without errors

### Additional details:
Still to be added is to confirm the experimental hook is fired when products are added to cart from the all products block (will add in a followup PR).

To be able to visit shop/cart pages using the `@woocommerce/e2e-utils` functions we needed to switch to pretty permalinks. This had the site affect that the tests in `setup-mc` where failing because the API URLs changed from using `%2F` to regular `/`. So the matching patterns had to be altered. Also we can't redirect to links like `/google/auth/` because this returns a 404, so instead a query parameter link was used.

### Changelog entry
- Add - End to end tests for Gtag events.

### DIFF
--- a/js/src/gtag-events/index.js
+++ b/js/src/gtag-events/index.js
@@ -69,7 +69,7 @@ const singleAddToCartClick = function ( event ) {
 window.onload = function () {
 	document
 		.querySelectorAll(
-			'.add_to_cart_button:not( .product_type_variable, .product_type_grouped )'
+			'.add_to_cart_button:not( .product_type_variable ):not( .product_type_grouped )'
 		)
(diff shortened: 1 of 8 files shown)


ianlin (approved):
LGTM, the e2e and e2e-dev tests passed on my local machine.

PR #1551, opened 2022-06-03
Always compare site URL hash without trailing slash

### LINKED ISSUE #1549: Site URL mismatch with trailing slash
### Describe the bug:
We came across a site where the site URL is verified and claimed in Merchant Center as `https://domain.com/` (with trailing slash). However in the site settings it's configured without a slash. Which results in it seeing the URL as a mismatch which prevents product syncing.

In the Settings > General if we save the Site URL with a trailing slash it strips it off and displays it without a trailing slash. Although if we look in the DB or the hidden `wp-admin/options.php ` page we can see it with a trailing slash.
**Note: In the DB it's saved with the option name `home`, not `siteurl`.**

### Steps to reproduce:
1. Go to Settings > General and add a trailing slash to the URL (confirm it's saved that way in the DB since we can't see it on this screen) .
2. Go through the onboarding and verify / claim the URL (if using ngrok you will need to make sure the adjusted URL has a trailing slash `define( 'WP_HOME', 'https://' . $_SERVER['HTTP_X_ORIGINAL_HOST'] . '/' );`).
3. Go back to the DB and change the `home` option in the `wp_options` table, removing the trailing slash.
4. Wait for the initial sync after onboarding to complete so it doesn't conflict with us testing syncing after changing the URL.
5. Go to the Connection Test page and trigger a product sync (all or individual) making sure async is checked.
6. Check the scheduled actions and confirm the sync did not occur because the URL does not match.

### Expected behavior:
For the URL to be compared without a trailing slash so in cases like the one described above it will still consider it as a match.

### Additional details:
Discovered the mismatch in 5256564-zen

mikkamp:

This PR always removes the trailing slash from the URL before calculating the MD5. This is to ensure that we continue syncing even if the account and site URL differ with just a trailing slash.

Closes #1549 

1. Go to Settings > General and add a trailing slash to the URL (confirm it's saved that way in the DB since we can't see it on this screen) .
2. Go through the onboarding and verify / claim the URL (if using ngrok you will need to make sure the adjusted URL has a trailing slash define( 'WP_HOME', 'https://' . $_SERVER['HTTP_X_ORIGINAL_HOST'] . '/' );).
3. Wait for the initial sync after onboarding to complete so it doesn't conflict with us testing syncing after changing the URL.
4. Go back to the DB and change the home option in the wp_options table, removing the trailing slash.
5. Go to the Connection Test page and trigger a product sync (all or individual) making sure async is checked.
6. Check the scheduled actions and confirm that they still sync even though the URL's don't fully match.

### Changelog entry
* Tweak - Always compare site URL hash without trailing slash.

### DIFF
--- a/src/API/Google/Merchant.php
+++ b/src/API/Google/Merchant.php
@@ -211,7 +211,7 @@ class Merchant implements OptionsAwareInterface {
 					return null;
 				}
 
-				$claimed_url_hash = md5( $account_url );
+				$claimed_url_hash = md5( untrailingslashit( $account_url ) );
 				$this->options->update( OptionsInterface::CLAIMED_URL_HASH, $claimed_url_hash );
 			} catch ( Exception $e ) {
 				return null;
--- a/src/MerchantCenter/MerchantCenterService.php
+++ b/src/MerchantCenter/MerchantCenterService.php
@@ -114,7 +114,7 @@ class MerchantCenterService implements ContainerAwareInterface, OptionsAwareInte
 
 		if ( null === $url_matches ) {
 			$claimed_url_hash = $this->container->get( Merchant::class )->get_claimed_url_hash();
-			$site_url_hash    = md5( $this->get_site_url() );
+			$site_url_hash    = md5( untrailingslashit( $this->get_site_url() ) );
 			$url_matches      = apply_filters( 'woocommerce_gla_ready_for_syncing', $claimed_url_hash === $site_url_hash ) ? 'yes' : 'no';
 			$transients->set( TransientsInterface::URL_MATCHES, $url_matches, HOUR_IN_SECONDS * 12 );
 		}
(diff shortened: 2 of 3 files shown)


layoutd (approved):
LGTM 👍🏻 

Had a time testing it: I wanted to confirm that it didn't work with `develop` and _did_ with the PR, but `gla_claimed_url_hash` was generated the first time and I didn't disconnect/reconnect, so the original value was still used for matching (with the trailing `/` from the first test), so even with PR changes, it wouldn't sync 😞  . Finally solved it by erasing the options.

It would be a problem after this PR is merged is if a merchant already has GL&A set up with a trailing-slash URL in GMC, and then updates to this new version – the changes wouldn't affect the hash stored in `wp_options`, so it syncing would continue to fail. But a GMC disconnect-reconnect would erase and create the URL hash again, so that one very-edge case could be resolved easily.

mikkamp:
Thanks for the review.

> But a GMC disconnect-reconnect would erase and create the URL hash again, so that one very-edge case could be resolved easily.

Even if we would automatically detect the change, the automatic syncing doesn't restart, since it's only triggered once after onboarding. So to ensure it goes through all the steps it's best to do a reconnect, which is what I assumed they would do to resolve an issue like this. Same happens if they genuinely change their site URL, a full reconnect is needed.

PR #1554, opened 2022-06-06
Test add to cart event from a block shop page

mikkamp:

This is a followup to #1550 
Another test is added to test the experimental hook is fired from a blocks page and the resulting Gtag event is sent.

1. Confirm the e2e run in this PR without errors
2. Run the e2e tests locally and confirm the extra test for add_to_cart event from block shop page runs without errors

### Additional details:
We don't have fixtures in the WC e2e test utils, but part of the implementation in the WooCommerce Blocks extension was copied to create a page with block content.

### Changelog entry
* Add - End to end tests for Gtag events from a block shop page.


ianlin (approved):
LGTM. The e2e and e2e-dev tests for add_to_cart event from block shop page passed in my local machine.

PR #1555, opened 2022-06-06
Disable Review Request in Standalone Accounts

### LINKED ISSUE #1545: Error getting account review status

puntope:

Closes #1545 

This PR checks if the account is standalone, if so, returns null in the status hiding the request review feature. That prevent to show a notice dialog indicating that there was an error requesting Account review status. 

1. Checkout PR
2. Disconnect all your google account in settings to clear the cache 
3. Connect GLA with a Standalone account (this is an account created outside the plugin)
4. When you finish the setup you wont see the Account status in product feed summary anymore, neither any error notice.
5. Disconnect again the accounts
6. Connect a normal account (created in GLA plugin)
7. After completing setup you will see the account status in product feed summary.

### Additional details:

- Added Test suit for Account Review Status Middleware (get)
- Restore changes made in resolvers hiding all the errors. 
- Created  helper in Merchant.php to check if the account is standalone following indications [in here](https://developers.google.com/shopping-content/reference/rest/v2.1/accounts/authinfo). It seems that Standalone account doesn't have the `aggregatorId` field set

### Changelog entry

> Tweak - Disable Review Request in Standalone Accounts

### DIFF
--- a/src/API/Google/Merchant.php
+++ b/src/API/Google/Merchant.php
@@ -348,20 +348,4 @@ class Merchant implements OptionsAwareInterface {
 	public function update_merchant_id( int $id ): bool {
 		return $this->options->update( OptionsInterface::MERCHANT_ID, $id );
 	}
-
-	/**
-	 * A standalone account is an account created directly in MC without using our plugin.
-	 * Standalone accounts have no access to WCS proxy and other features.
-	 *
-	 * This function detects if the account is standalone
-	 * We use this API for that https://developers.google.com/shopping-content/reference/rest/v2.1/accounts/authinfo
-	 *
-	 * We consider an accou t standalone if no aggregatorId is present or no accountIdentifiers are available.
-	 *
-	 * @return bool True if it's a standalone account.
-	 */
-	public function is_standalone(): bool {
-		$type = $this->service->accounts->authinfo();
-		return ! isset( $type['accountIdentifiers'] ) || empty( $type['accountIdentifiers'] ) || is_null( $type['accountIdentifiers'][0]['aggregatorId'] );
-	}
 }
--- a/src/API/Google/Merchant.php
+++ b/src/API/Google/Merchant.php
@@ -348,4 +348,20 @@ class Merchant implements OptionsAwareInterface {
 	public function update_merchant_id( int $id ): bool {
 		return $this->options->update( OptionsInterface::MERCHANT_ID, $id );
 	}
+
+	/**
+	 * A standalone account is an account created directly in MC without using our plugin.
+	 * Standalone accounts have no access to WCS proxy and other features.
+	 *
+	 * This function detects if the account is standalone
+	 * We use this API for that https://developers.google.com/shopping-content/reference/rest/v2.1/accounts/authinfo
+	 *
+	 * We consider an accou t standalone if no aggregatorId is present or no accountIdentifiers are available.
+	 *
+	 * @return bool True if it's a standalone account.
+	 */
+	public function is_standalone(): bool {
+		$type = $this->service->accounts->authinfo();
+		return ! isset( $type['accountIdentifiers'] ) || empty( $type['accountIdentifiers'] ) || is_null( $type['accountIdentifiers'][0]['aggregatorId'] );
+	}
 }
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -540,16 +540,15 @@ class Middleware implements OptionsAwareInterface {
 				$this->get_manager_url( 'account-review-status/' . $this->options->get_merchant_id() ),
 			);
 
-			 $response = json_decode( $result->getBody()->getContents(), true );
+			$response = json_decode( $result->getBody()->getContents(), true );
 
 			if ( 200 === $result->getStatusCode() && isset( $response['freeListingsProgram'] ) && isset( $response['shoppingAdsProgram'] ) ) {
 				do_action( 'woocommerce_gla_request_review_response', $response );
 				return $response;
 			}
-
-			 do_action( 'woocommerce_gla_guzzle_invalid_response', $response, __METHOD__ );
-			 $error = $response['message'] ?? __( 'Invalid response getting account review status', 'google-listings-and-ads' );
-			 throw new Exception( $error, $result->getStatusCode() );
+			do_action( 'woocommerce_gla_guzzle_invalid_response', $response, __METHOD__ );
+			$error = $response['message'] ?? __( 'Invalid response getting account review status', 'google-listings-and-ads' );
+			throw new Exception( $error, $result->getStatusCode() );
 		} catch ( ClientExceptionInterface $e ) {
 			do_action( 'woocommerce_gla_guzzle_client_exception', $e, __METHOD__ );
 
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -36,7 +36,6 @@ defined( 'ABSPATH' ) || exit;
  */
 class Middleware implements OptionsAwareInterface {
 
-
 	use ApiExceptionTrait;
 	use OptionsAwareTrait;
 	use PluginHelper;
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -529,13 +529,9 @@ class Middleware implements OptionsAwareInterface {
 	 */
 	public function get_account_review_status() {
 		try {
-			/** @var Merchant $merchant */
-			$merchant = $this->container->get( Merchant::class );
 
-			if ( $merchant->is_standalone() ) {
-				$response = [];
-				do_action( 'woocommerce_gla_request_review_standalone', $response );
-				return $response;
+			if ( ! $this->is_subaccount() ) {
+				return [];
 			}
 
 			/** @var Client $client */
@@ -621,5 +617,23 @@ class Middleware implements OptionsAwareInterface {
 				$e->getCode()
 			);
 		}
+	}
+
+	/**
+	 * This function detects if the current account is a sub-account
+	 *
+	 * @return bool True if it's a standalone account.
+	 */
+	public function is_subaccount(): bool {
+		$merchant_id = $this->options->get_merchant_id();
+		$accounts    = $this->get_merchant_accounts();
+
+		foreach ( $accounts as $account ) {
+			if ( $account['id'] === $merchant_id && $account['subaccount'] ) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 }
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -498,7 +498,7 @@ class Middleware implements OptionsAwareInterface {
 	 */
 	protected function default_account_name(): string {
 		return sprintf(
-		/* translators: 1: current date in the format Y-m-d */
+			/* translators: 1: current date in the format Y-m-d */
 			__( 'Account %1$s', 'google-listings-and-ads' ),
 			( new DateTime() )->format( 'Y-m-d' )
 		);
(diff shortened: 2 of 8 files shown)


ecgan:
Just want to note that the commit https://github.com/woocommerce/google-listings-and-ads/pull/1555/commits/da5452d12b2d94050d74cc22c46203b6013ddb2e essentially reverts the PR https://github.com/woocommerce/google-listings-and-ads/pull/1547. 

I'm wondering if we should revert it - we can have the PHP code changes without having the JS reverted code change. One thing good about the revert is that it makes it consistent with other code in the wp-data resolvers, but on the other hand, I also think we should not revert as we should move away from that `handleFetchError` / `createNotice` in wp-data (though it may be done later, not now). What do you think? 🤔

src/API/Google/Merchant.php
```diff
@@ -348,4 +348,20 @@ public function has_access( string $email ): bool {
...
+	 * We use this API for that https://developers.google.com/shopping-content/reference/rest/v2.1/accounts/authinfo
+	 *
+	 * We consider an accou t standalone if no aggregatorId is present or no accountIdentifiers are available.
+	 *
+	 * @return bool True if it's a standalone account.
+	 */
+	public function is_standalone(): bool {
+		$type = $this->service->accounts->authinfo();
+		return ! isset( $type['accountIdentifiers'] ) || empty( $type['accountIdentifiers'] ) || is_null( $type['accountIdentifiers'][0]['aggregatorId'] );
+	}
```
mikkamp: I don't think this function is going to catch all the scenarios we need. There are other "manager" accounts which can create sub accounts, we need to specifically detect if our WooCommerce account is the aggregator account.

Since we don't store the ID in the plugin itself we moved the request to get Merchant Accounts to the WCS so we can determine whether it's a sub account or not. See [here](https://github.com/Automattic/woocommerce-connect-server/blob/woogle/lib/google/index.js#L499)

We should consider whether we want to add another end point to check if it's a sub account. Or if we want to cache the isSubAccount value during onboarding.

Also I think the naming might be confusing since in our case `isSubAccount` means it is both a sub account and a child of our WooCommerce account. All other cases are the ones we want to exclude but the word "standalone" doesn't seem to cover sub accounts that have a different parent account.

src/API/Google/Middleware.php
```diff
@@ -529,6 +529,15 @@ protected function get_site_timezone_string(): string {
 	 */
 	public function get_account_review_status() {
 		try {
+			/** @var Merchant $merchant */
+			$merchant = $this->container->get( Merchant::class );
+
+			if ( $merchant->is_standalone() ) {
+				$response = [];
+				do_action( 'woocommerce_gla_request_review_standalone', $response );
+				return $response;
```
mikkamp: May I ask what the intention of this action hook is? It looks like if it will ever be useful it might be better to have this as a filter instead. Always passing an empty array to the action hook doesn't provide any benefit to the function that hook into this.

In my opinion I wouldn't have added any action hooks in the function `get_account_review_status` unless we foresee us specifically needing them for anything specific.

mikkamp (commented):
Thanks for the PR. I haven't fully tested it, but I left some comments as I think we need to change some of the logic as it wouldn't catch sub-accounts with a different "manager" account.

puntope:
> I'm wondering if we should revert it 

IMO. I think is useful to have the error handler to catch potential errors in the API call.

src/API/Google/Merchant.php
puntope: > Since we don't store the ID in the plugin itself we moved the request to get Merchant Accounts to the WCS so we can determine whether it's a sub account or not. See [here](https://github.com/Automattic/woocommerce-connect-server/blob/woogle/lib/google/index.js#L499)
We should consider whether we want to add another end point to check if it's a sub account. 

Thanks for pointing this out.

I modified a bit the approach. Now I use get_merchant_accounts to get if its a subaccount. 

> Or if we want to cache the isSubAccount value during onboarding.

Yes would be cool to create a future PR maybe to cache that endpoint if needed.

src/API/Google/Middleware.php
puntope: Right, wanted a filter but did the action instead. I just removed it at the end.

src/API/Google/Middleware.php
```diff
@@ -613,4 +618,22 @@ public function account_request_review( $regions ) {
...
+		$accounts    = $this->get_merchant_accounts();
+
+		foreach ( $accounts as $account ) {
+			if ( $account['id'] === $merchant_id && $account['subaccount'] ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
```
mikkamp: The only thing I don't like about this approach is that the WCS turns this into several requests (one for account list + one per account for details). I might be missing something, but as far as I can see this is requested every time we view the Product Feed page, since it isn't cached.

Can we cache this in a transient to prevent it from being requested repeatedly. As far as I'm aware an account can't change from sub account to standalone, at least not easily. So repeatedly asking the account type isn't necessary.
Or do you think it's worth it to have a separate endpoint in WCS to get the sub-account status. Or even rewrite the get merchants endpoint to flag whether we want account detail requests to be sent.

puntope: > I might be missing something, but as far as I can see this is requested every time we view the Product Feed page, since it isn't cached.

This is cached  for 20 mins on `get_review_status` function inside `RquestReviewController`

I can do as well another cache only for the standalone function without expiration time.

That is done here https://github.com/woocommerce/google-listings-and-ads/pull/1555/commits/ff14ff68afe26612b665cc0fe7875098b9049ecd

src/API/Google/Middleware.php
```diff
@@ -529,22 +529,27 @@ protected function get_site_timezone_string(): string {
...
 				do_action( 'woocommerce_gla_request_review_response', $response );
 				return $response;
 			}
 
-			do_action( 'woocommerce_gla_guzzle_invalid_response', $response, __METHOD__ );
-			$error = $response['message'] ?? __( 'Invalid response getting account review status', 'google-listings-and-ads' );
-			throw new Exception( $error, $result->getStatusCode() );
+			 do_action( 'woocommerce_gla_guzzle_invalid_response', $response, __METHOD__ );
+			 $error = $response['message'] ?? __( 'Invalid response getting account review status', 'google-listings-and-ads' );
+			 throw new Exception( $error, $result->getStatusCode() );
```
mikkamp: Looks like these lines accidentally got an extra space.

tests/Unit/API/Google/MiddlewareTest.php
```diff
@@ -353,4 +358,79 @@ public function test_check_tos_accepted_exception() {
 		$this->assertEquals( 'error', $tos->message() );
 	}
 
+	public function test_get_account_review_status() {
+
+		$this->options->expects( $this->exactly( 2 ) )->method('get_merchant_id')->willReturn(self::TEST_MERCHANT_ID);
```
mikkamp: While we don't run these test files through PHPCS it's still nice to at least be consistent with adding spaces before and after brackets.

src/API/Google/Middleware.php
```diff
@@ -613,4 +618,34 @@ public function account_request_review( $regions ) {
...
+			foreach ( $accounts as $account ) {
+				if ( $account['id'] === $merchant_id && $account['subaccount'] ) {
+					$is_subaccount = 1;
+				}
+			}
+
+			$transients->set( $transients::MC_IS_SUBACCOUNT, $is_subaccount );
+		}
+
+		// since transients doesn't support booleans, we save them as 0/1 and do the conversion here
```
mikkamp:
```suggestion
		// since transients don't support booleans, we save them as 0/1 and do the conversion here
```

I've seen other solutions and in WC where the values "yes" / "no" are saved to get around this problem, but integer values work as well.

src/API/Google/Middleware.php
```diff
@@ -498,7 +498,7 @@ protected function new_account_name(): string {
 	 */
 	protected function default_account_name(): string {
 		return sprintf(
-			/* translators: 1: current date in the format Y-m-d */
+		/* translators: 1: current date in the format Y-m-d */
```
mikkamp: Any specific reason this indent was changed? The comment belongs to the line below, so I think it should match it's indent.

src/API/Google/Middleware.php
```diff
@@ -122,16 +123,15 @@ public function create_merchant_account(): int {
...
-	 * @param string $name     Site name
+	 * @param string $name Site name
 	 * @param string $site_url Website URL
 	 *
 	 * @return int Created merchant account ID
 	 *
 	 * @throws Exception   When an Exception is caught or we receive an invalid response.
 	 * @throws InvalidTerm When the account name contains invalid terms.
 	 * @throws InvalidDomainName When the site URL ends with an invalid top-level domain.
+	 * @since 1.5.0
```
mikkamp, line 133: Are these auto changed in your IDE? I don't think we've decided on a strict standard where to put the `@since` tag, but it would seem like they can be left alone if we aren't actually changing anything for those functions.

src/API/Google/Middleware.php
```diff
@@ -35,6 +36,7 @@
  */
 class Middleware implements OptionsAwareInterface {
 
+
```
mikkamp: One blank line should be enough here.

mikkamp (approved):
Thanks for all those changes.

I missed the 20 minute transient in the controller, but I think the never expiring transient is better to reduce requests, since that shouldn't change until they disconnect.

When testing with a standalone account I get the following response from the controller:
```json
{
    "status": null,
    "cooldown": 0,
    "issues": [],
    "reviewEligibleRegions": []
}
```

And I no longer get an error. On the account that I was still getting an internal error I still see the error message:
> There was an error loading your merchant center product review request status.

Which is what I expected.

I just added a few small comments, but none of them are blocking so I'll go ahead and approve the PR.

src/API/Google/Middleware.php
puntope: I would say PHPBF fixed that? Not sure to be honest, I will change it back

src/API/Google/Middleware.php
puntope: Not sure if I get you on this one, this since was in older PRs

mikkamp: That's what I meant, since the `@since` tags are from older PR's then why are thy being moved to a different location in this PR. Not a big deal I was just curious if that was intended or not.

PR #1557, opened 2022-06-09
Tree Select Control - Clear search input after an option changes

### LINKED ISSUE #1489: Enhancement - TreeSelectComponent should clear the previous search when an option is selected.
### Describe the enhancement:
When I am looking for an option in the TreeSelectComponent, I am expecting that once one option is selected, the search should be clear so I can search for another item without deleting the previous search manually.  

### Steps to reproduce:

1. Go to Edit Free listings -> Update the Audience country
2. Search for any country, select the country, for example Italy
3. Search for another country, here it is needed to delete Italy manually and then search for the new country.

https://user-images.githubusercontent.com/2488994/167125778-55246232-63b2-4251-bd58-87bb672d04da.mp4

### Expected behaviour:
Once an option is selected, the previous search should be clear.

### Actual behaviour:
Once an option is selected, the previous search is still showing.

puntope:

Closes #1489 .

After discussion, this PR removes the filter input after selecting an option. The component still maintains the **user** expanded nodes. 

The Tree Select acts a bit like [Ant Design system does it here](https://ant.design/components/tree-select/)

### Screenshots:

https://user-images.githubusercontent.com/5908855/172848902-86b2c578-2bba-49c8-93e7-5eb615de1e9a.mov

1. Checkout this PR and execute `npm run storybook`
2. Go to http://localhost:6006/?path=/story/tree-select-control--base
3. Don't expand any node and just search for a child item (like Kuala Lumpur, typing "Lump")
4. See how Kuala Lumpur appears expanded and other irrelevant nodes are hiden.
5. Select Kuala Lumpur
6. See how the search input filter is cleared, the tag Kuala Lumpur is added and the tree goes back to the initial state
7. Expand some child nodes, Like America
8. Search something including those child nodes (Like Canada)
9. Select Canada
10. See how the tag Canada is added, the search input is cleared and the expanded nodes are still showing America 
11. Search for a selected node (for example again Kuala Lumpur)
12. Deselect the node
13. See how the tag disappears, the tree goes to the previous state and the input is cleared as well.
14. Check the code, check the tests

### Changelog entry

> Tweak - Clear input search filter after selecting an option

### DIFF
--- a/js/src/components/tree-select-control/control.js
+++ b/js/src/components/tree-select-control/control.js
@@ -22,6 +22,7 @@ import { BACKSPACE } from './constants';
  * @param {boolean} props.isExpanded True if the tree is expanded
  * @param {boolean} props.disabled True if the component is disabled
  * @param {number} props.maxVisibleTags The maximum number of tags to show. Undefined, 0 or less than 0 evaluates to "Show All".
+ * @param {string} props.value The current input value
  * @param {Function} props.onFocus On Focus Callback
  * @param {Function} props.onTagsChange Callback when the Tags change
  * @param {Function} props.onInputChange Callback when the Input value changes
@@ -35,6 +36,7 @@ const Control = ( {
 	isExpanded,
 	disabled,
 	maxVisibleTags,
+	value = '',
 	onFocus = () => {},
 	onTagsChange = () => {},
 	onInputChange = () => {},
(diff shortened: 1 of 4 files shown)


jorgemd24 (approved):
Thanks, @puntope, LGMT 👍  

💅 

I was expecting to see a test in `options.test.js`, testing when the option is selected, the input value should be cleared, what do you think?

puntope:
> I was expecting to see a test in `options.test.js`, testing when the option is selected, the input value should be cleared, what do you think?

Right @jorgemd24 . I just added a test checking that :) thanks!

PR #1559, opened 2022-06-10
Fix cannot disconnect Jetpack when other activated plugins are using Jetpack connection

### LINKED ISSUE #1434: Jetpack connection can't be disconnected via the hidden page
### Describe the bug:

The extension hidden page is used to connect as well as disconnect existing connections. However, while testing the recent update, the Jetpack connection can't be disconnected from that page: https://d.pr/i/amHuCz

This suggests the Jetpack connection can't be disconnected at all in the extension (Except all connections are disconnected using the "disconnect all button"). Being able to disconnect and reconnect Jetpack is important because disconnecting existing connections from the hidden page, etc, is known to help when there's a connection/setup issue (e.g the setup can't be completed).

1. Setup the plugin 
2. Go to the hidden page. - example.com/wp-admin/admin.php?page=connection-test-admin-page
3. Click the "Disconnect Jetpack" button
4. The connection remains connected so the prompt to reconnect Jetpack isn't shown

### Expected behavior:

* Clicking "Disconnect Jetpack" from the hidden page should disconnect the existing Jetpack connection, allowing for the new prompt to reconnect Jetpack to be shown on the extension's page.

### Actual behavior:

The connection remains despite the button to "disconnect Jetpack" being clicked.

ianlin:
Project thread: pcTzPl-AG-p2

## Context

Previously the behaviour `cannot disconnect Jetpack when other activated plugins are using Jetpack connection` was considered as an expected "soft disconnect" feature, meaning that a plugin would not actually disconnect Jetpack connection when there were other activated plugins that are using the connection. However, according to p9dueE-55t-p2 the Jetpack crew has decided to change this behaviour, here's one quote from the post:

> There’s no longer such a thing as “soft” connection/disconnection. Meaning that either the site is connected and all plugins are connected, or the site is disconnected and all plugins are disconnected.

## <a href="#updates-15-june-2022" id="updates-15-june-2022">#</a> Updates on 15th June 2022

According to the suggestions from @leogermani in https://github.com/woocommerce/google-listings-and-ads/pull/1559#issuecomment-1153916454 as well as in the project post pcTzPl-AG-p2#comment-1247, we have decided to keep the original "soft disconnect" behaviour when the user tries to disconnect Jetpack connection in both GLA's settings page and the connection test page. If the user wants to fully disconnect Jetpack connection while there are other plugins using the connection, [they should use "My Jetpack" from the Jetpack plugin](https://jetpack.com/support/getting-started-with-jetpack/my-jetpack/).

#### Updates on 21st June 2022

In the connection test page I've added some changes to show the reason why Jetpack connection cannot be disconnected, it will look like this:

Screenshot 2022-06-21 at 14 38 21

Closes #1434, this PR fixes a problem that from GLA's connection test page (`/wp-admin/admin.php?page=connection-test-admin-page`) and GLA's settings page (`/wp-admin/admin.php?page=wc-admin&path=/google/settings`), the Jetpack connection cannot be disconnected when there were activated plugins that are using Jetpack connection as well.

The Jetpack connection's new behaviour was first introduced in [1.39.0](https://github.com/Automattic/jetpack-connection/blob/382337b0de23f88ed0a325b3ee0f02b32d5923b5/src/class-manager.php#L1645-L1650) and is remained unchanged in the latest [1.40.5](https://github.com/Automattic/jetpack-connection/blob/461d33c0aa9b8f706d1b62a317a6c770f990ce4d/src/class-manager.php#L1645-L1650) as of now, you can see the `remove_connection()` method's second parameter `ignore_connected_plugins` is set to `false` by default, and the method just calls `disconnect_site()` method without doing any other things. The `remove_connection()` method is for backward compatible as it keeps the "soft disconnect" behaviour if we just call it without any parameters, just like what we did in `src/ConnectionTest.php` before changes in this PR.

Since `disconnect_site()` method's second parameter `ignore_connected_plugins` is set to `true` by default, in `src/ConnectionTest.php` we can just replace `remove_connection()` with `disconnect_site()` and it will solve the problem.

In this PR we have changed three things:

1. Upgrade `jetpack-connection` package to `^1.40.0`, because currently we're using `^1.20` and the new behaviour was introduced since `1.39.0`.
2. Replace `remove_connection()` with `disconnect_site()`.
3. Remove calling a deprecated method `enable_plugin()`.

#### Disconnect Jetpack from the connection test page.

1. Use `develop` branch.
2. Install [Jetpack Boost](https://wordpress.org/plugins/jetpack-boost/) plugin and activate it.
3. Go to connection test page: `/wp-admin/admin.php?page=connection-test-admin-page`
4. Click `Disconnect Jetpack` button
    Screenshot 2022-06-10 at 15 39 03
5. The Jetpack connection is not disconnected.
    - Go to connection test page, the `Disconnect Jetpack` button is still there.
    - Go to Jetpack Boost page `/wp-admin/admin.php?page=jetpack-boost`, you will see it is still running:
        Screenshot 2022-06-10 at 15 44 46

#### Disconnect Jetpack from GLA's settings page

1. Use `develop` branch.
2. Install [Jetpack Boost](https://wordpress.org/plugins/jetpack-boost/) plugin and activate it.
3. Go to GLA's settings page: `/wp-admin/admin.php?page=wc-admin&path=/google/settings`.
4. Click `Disconnect from all accounts` button.
    Screenshot 2022-06-10 at 16 38 31
5. The Jetpack connection is not disconnected.
    - Click `Start listing products` from the GLA get started page, the WordPress account is already connected.
        Screenshot 2022-06-10 at 16 35 19
    - Go to Jetpack Boost page `/wp-admin/admin.php?page=jetpack-boost`, you will see it is still running:
        Screenshot 2022-06-10 at 15 44 46

#### Disconnect Jetpack from the connection test page.

1. Use the branch of this PR `fix/cannot-disconnect-jetpack-when-other-jetpack-plugins-activated`.
2. Repeat the step 2, 3, and 4 from the `steps to reproduce of Disconnect Jetpack from the connection test page` above.
3. The Jetpack connection should be disconnected.
    - Go to connection test page, the `Connect to Jetpack` button should be there:
        Screenshot 2022-06-10 at 15 48 27
    - Go to Jetpack Boost page `/wp-admin/admin.php?page=jetpack-boost`, you will see it is asking you to get started:
        Screenshot 2022-06-10 at 15 42 28

#### Disconnect Jetpack from GLA's settings page

1. Use the branch of this PR `fix/cannot-disconnect-jetpack-when-other-jetpack-plugins-activated`.
2. Repeat the step 2, 3, and 4 from the `steps to reproduce of Disconnect Jetpack from GLA's settings page` above.
3. The Jetpack connection should be disconnected.
    - Click `Start listing products` from the GLA get started page, the WordPress account is not connected:
        Screenshot 2022-06-10 at 16 40 16
    - Go to Jetpack Boost page `/wp-admin/admin.php?page=jetpack-boost`, you will see it is asking you to get started:
        Screenshot 2022-06-10 at 15 42 28

### Changelog entry

> Fix - Cannot disconnect Jetpack when other activated plugins are using Jetpack connection.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
 		"php": ">=7.3",
 		"automattic/jetpack-autoloader": "^2.4",
 		"automattic/jetpack-config": "^1.4",
-		"automattic/jetpack-connection": "^1.20",
+		"automattic/jetpack-connection": "^1.40",
 		"google/apiclient": "^2.12",
 		"google/apiclient-services": "~0.230",
 		"googleads/google-ads-php": "^12.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,28 +4,36 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "16f2430cdfa4afda4cf867622f1e7c27",
+    "content-hash": "5c1f22c9184e8e10a38504425a3949b1",
     "packages": [
         {
             "name": "automattic/jetpack-a8c-mc-stats",
-            "version": "1.4.1",
+            "version": "v1.4.13",
             "source": {
                 "type": "git",
                 "url": "https://github.com/Automattic/jetpack-a8c-mc-stats.git",
-                "reference": "b2f6b20aa3046848391593e6a18f87ae3797e888"
+                "reference": "64ee9a83861c6b2c8744e2ebc25157806f3548e0"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/Automattic/jetpack-a8c-mc-stats/zipball/b2f6b20aa3046848391593e6a18f87ae3797e888",
-                "reference": "b2f6b20aa3046848391593e6a18f87ae3797e888",
+                "url": "https://api.github.com/repos/Automattic/jetpack-a8c-mc-stats/zipball/64ee9a83861c6b2c8744e2ebc25157806f3548e0",
+                "reference": "64ee9a83861c6b2c8744e2ebc25157806f3548e0",
                 "shasum": ""
             },
             "require-dev": {
-                "yoast/phpunit-polyfills": "0.2.0"
+                "automattic/jetpack-changelogger": "^3.1",
+                "yoast/phpunit-polyfills": "1.0.3"
             },
-            "type": "library",
+            "type": "jetpack-library",
             "extra": {
-                "mirror-repo": "Automattic/jetpack-a8c-mc-stats"
+                "autotagger": true,
+                "mirror-repo": "Automattic/jetpack-a8c-mc-stats",
+                "changelogger": {
+                    "link-template": "https://github.com/Automattic/jetpack-a8c-mc-stats/compare/v${old}...v${new}"
+                },
+                "branch-alias": {
+                    "dev-master": "1.4.x-dev"
+                }
             },
             "autoload": {
                 "classmap": [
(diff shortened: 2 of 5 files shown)


mikkamp (commented):
I haven't fully tested the code, based on the new recommendations the implementation looks right for the Connection Test page.

However most users won't be using the Connection Test page and will use the UI (which only has the option to disconnect all accounts including Jetpack).
I don't see the same changes in `src/API/Site/Controllers/Jetpack/AccountController.php` for the connect and disconnect functions, is that supposed to be part of this PR as well?

ianlin:
@mikkamp thanks for the quick review! I was actually modifying `src/API/Site/Controllers/Jetpack/AccountController.php` at the moment, I'll let you know once it's done.

ianlin:
Hey @mikkamp I've pushed a new commit for updating `src/API/Site/Controllers/Jetpack/AccountController.php` as well as updated the description for reproduction and test instructions.

mikkamp (approved):
Thanks for the changes, I tested them according to the instructions and was able to confirm that the connection was fully disconnected.

I was just wondering if we got any input from design on the "Disconnect all accounts" feature. Now that it's a hard disconnect I might want to disconnect only my MC and Google account if I'm using the Jetpack connection for WCPay. We also don't warn them explicitly that it could for example cause issues with checkout if WCPay is no longer connected.

jorgemd24:
I was just reading about the JetPack updates, and I saw that `is_active` has been deprecated:

> is_active() has also been deprecated a while ago. Instead you should use the following methods to verify if the site is connected:
> 
> is_connected() – meaning it is connected at a site level (aka has a blog token) and
> has_connected_owner() – meaning at least one admin has authorized their WPCOM account.

See the full comment here: p9dueE-55t-p2#comment-7911

It seems that `is_active`is only used in one place in the GLA codebase:

https://github.com/woocommerce/google-listings-and-ads/blob/33a1be14045ea88633a67e958d3b240f44b0279c/src/API/Site/Controllers/Jetpack/AccountController.php#L194

I think we should replace this deprecated method too.

mikkamp:
> I was just reading about the JetPack updates, and I saw that is_active has been deprecated

Nice catch, I had tested the connection status and didn't receive any deprecated notices so I assumed it was alright. Seems like it's [intentional](https://github.com/Automattic/jetpack-connection/blob/v1.40.5/src/class-manager.php#L519) not to show a notice though:

> This method is deprecated since version 1.25.0 of this package. Please use has_connected_owner instead.
> Since this method has a wide spread use, we decided not to throw any deprecation warnings for now.

Still nicer to use a future proof method.

ianlin:
@jorgemd24 Good catch, thanks. I've updated it in 3937bfd.

ianlin:
> I was just wondering if we got any input from design on the "Disconnect all accounts" feature. 

@mikkamp Thanks for the reminder, I'll ask in the project post pcTzPl-AG-p2.

leogermani:
Hi all, thanks for working on this!

I want to add two clarifications:

* Even though we don't do "soft disconnect" anymore, we still use `remove_connection` as the default callback in Plugin Deactivation hooks. Meaning that if you have a routine that will remove all connections when the plugin is deactivated, it should not kill the Jetpack connection if other plugins are using it. So `remove_connection` is not only for BC, it is a shortcut to be used in those cases.

* Since `disconnect_site` will kill the connection for all plugins, we are removing this user action from all Jetpack family plugins, and leaving it in one shared place called "My Jetpack" where users can manage their connections. If you sill have a button that will kill the connection from your plugin's UI, **we ask you that you warn users that other plugins are using the connection and will stop working properly**. You can see an example of this when you try to Disconnect from "My Jetpack" and there are multiple plugins using it.

ianlin:
Thanks @leogermani for clarifying! We will discuss with the designer which approaches we're gonna go. 👍

ianlin:
Thanks @leogermani for the suggestions, I have updated the code to still use `remove_connection()` when disconnecting Jetpack in order to keep the soft disconnect behaviour. Also [updated the PR description](#updates-15-june-2022) about the new changes.

jorgemd24:
@ianlin, thanks for the adjustments, I was reading about all the updates in Jetpack and I have some questions:

1. If for example, I have only the following plugins installed:

image

And I click disconnect Jetpack from the "Test Page", shouldn't the Jetpack get disconnected? As there are no other plugins using JetPack? I see that my JetPack connection is still activated.

2. Related to the issue: https://github.com/woocommerce/google-listings-and-ads/issues/1434, I understand that from the user's point of view everything would still work as before, so maybe we should add a comment in the issue, so it is clear that it will disconnect the Jetpack connection only if the GLA plugin was the only plugin using it. what do you think?

ianlin:
Hey @jorgemd24 thanks for the review and suggestions. From your screenshot you've installed and activated `Jetpack Boost` plugin which uses Jetpack connection as well, that's the reason why the connection cannot be disconnected from the connection test page. However I think it will be useful to display the reason why the connection cannot be disconnected in the test page so I added some changes in 8c1550a. It will look like this:

Screenshot 2022-06-21 at 14 38 21

> so maybe we should add a comment in the issue, so it is clear that it will disconnect the Jetpack connection only if the GLA plugin was the only plugin using it. what do you think?

Thanks for the suggestion and it's definitely helpful to leave an update in the issue. I was actually waiting for this PR is merged before leaving a comment there.

jorgemd24 (approved):
Thanks, @ianlin for the changes! The message that you added in the connection test will be really useful! LGTM 👍

PR #1494, opened 2022-05-10
Shipping Integration - Syncing rates for states/provinces and postal codes

nima-karimi:
### Changes proposed in this Pull Request:

Closes #1228.
Closes #1230.
Closes #1231.

This is the main feature branch for the shipping integration milestone two. All of the work in this PR has already been reviewed, tested, and merged from other PRs:

- https://github.com/woocommerce/google-listings-and-ads/pull/1341
- https://github.com/woocommerce/google-listings-and-ads/pull/1422
- https://github.com/woocommerce/google-listings-and-ads/pull/1455
- https://github.com/woocommerce/google-listings-and-ads/pull/1445
- https://github.com/woocommerce/google-listings-and-ads/pull/1560
- https://github.com/woocommerce/google-listings-and-ads/pull/1574

### Changelog entry

> Add - Option to automatically sync the shipping rates based on the store shipping zone configurations

PR #1560, opened 2022-06-10
Automatically sync WooCommerce shipping settings with Merchant Center

nima-karimi:

This PR adds automatic syncing functionality for synchronizing the WooCommerce shipping zone settings to Google Merchant Center whenever there's a change applied by the user in their WooCommerce store settings.

There's a new async ActionSchedule job called `update_shipping_settings`. This job will run whenever we detect any changes to WooCommerce shipping settings and it will upload the updated settings to Google Merchant Center. Before scheduling and before processing (two times) it checks whether the MC account is connected AND the user has chosen the "automatic" syncing option for their shipping settings during onboarding.

To watch for WooCommerce settings changes, a new SyncerHooks class is introduced under the `Shipping` namespace. This class attaches an action to various WooCommerce and WordPress hooks and whenever it detects a change it schedules the `update_shipping_settings` job. WooCommerce uses various methods for updating the shipping settings (ajax, rest API, etc.); therefore, it was necessary to watch for multiple hooks. However, there is a flag in place to prevent the job from being scheduled more than once on each request.

This PR is based on the [main feature branch](https://github.com/woocommerce/google-listings-and-ads/pull/1494) for shipping integration milestone two.

1. Make sure you have your MC account connected and finished the onboarding steps
2. Confirm that you have chosen the "Recommended" option for syncing your shipping settings automatically from WooCommerce
3. Go to WooCommerce settings and add a new shipping zone.
4. Add new methods to it. Edit then and add prices and save.
5. Add new locations and postcodes. Save the shipping zone.
6. Edit previously added methods or delete them. Save the shipping zone.
7. Edit the zone's locations or delete them and save the shipping zone. 
8. Add new shipping classes, assign prices to them in your shipping methods and then rename the classes or delete them.
9. Confirm that in all the test cases mentioned in steps 3-8, the `update_shipping_settings` job is scheduled in ActionScheduler.
10. Confirm in your MC account that the new shipping settings are reflected in your account's shipping section
11. Edit your free listings settings and change your shipping option to anything other than the "Recommended" option.
12. Repeat steps 3-8 and confirm that no update job is scheduled in ActionScheduler and your shipping settings remained unchained in MC.
13. Disconnect your MC account and check steps 3-8. No job should be scheduled to update the shipping settings.

### Changelog entry

> Add - Automatically sync WooCommerce shipping settings with Merchant Center


ianlin (commented):
I was trying to go through the onboarding again to ensure I was using the "Recommended" option, but I got an error in the last step of onboarding when calling `POST mc/settings/sync` API:

> [postalCodeGroups[0].postalCodeRanges[0]] Location type postal code is not supported in TW.

It worked as expected if I used `develop` branch, but using the branch of this PR got me the error above. I also locally merged `develop` into this branch but still got the same error. I'm not sure if I've missed something, any idea about it?

nima-karimi:
Postal codes are currently only supported in a few countries. Namely US, AU, and JP. But you shouldn't be getting any errors for TW because the API should ignore that country. Can you please post your shipping zone setup in WC and your GLA target country? Thanks.

ianlin:
@nima-karimi Thanks for pointing this out. It was because my shipping zone setup in WC is `TW` with a postcode, I did some tests with different combinations and here's the result for the last step of onboarding setup (i.e. calling `POST mc/settings/sync`):

GLA target country: All countries.

WC Shipping zone:
- TW _without_ postcodes: success ✅ 
- US _with_ postcodes: success ✅ 
- TW _with_ postcodes: failed ❌ 
  - Error message: `[postalCodeGroups[0].postalCodeRanges[0]] Location type postal code is not supported in TW.`

While you're looking into this I will review the PR for automatically sync shipping settings with GMC. Thanks!

src/Shipping/SyncerHooks.php
```diff
@@ -0,0 +1,131 @@
...
+		if ( ! $this->merchant_center->is_connected() || ! $this->google_settings->should_get_shipping_rates_from_woocommerce() ) {
+			return;
+		}
+
+		$update_settings = function () {
+			$this->handle_update_shipping_settings();
+		};
+
+		// After a shipping zone object is saved to database.
+		add_action( 'woocommerce_after_shipping_zone_object_save', $update_settings, 90 );
```
ianlin, line 75: ❔ Just a question for a WP/WC newbie like me, how could we find this hook name? I couldn't find it from [WooCommerce Action and Filter Hook Reference](https://woocommerce.github.io/code-reference/hooks/hooks.html) 🤔.

ianlin (commented):
Thanks for adding this. I've tested locally and it worked as expected. I can confirm the shipping settings have been changed in Google Merchant Centre after modifying from the WooCommerce shipping settings. Also confirm that it won't be syncing if the merchant centre account is not connected, or the free listing's shipping option is not "Recommended".

Before approving the PR I have two questions:

1. Asked in https://github.com/woocommerce/google-listings-and-ads/pull/1560#issuecomment-1157200627, the onboarding will fail if I set up shipping postcodes for a postcodes unsupported country, e.g. TW.
2. After deleting a `Shipping Zone`, the job wasn't scheduled immediately. Not sure it's a requirement or not. If we need the job to be scheduled after deleting a "Shipping Zone", maybe we could add an action `woocommerce_delete_shipping_zone` in `src/Shipping/SyncerHooks.php`?

src/Shipping/SyncerHooks.php
nima-karimi: Yeah, it's a bit hard to find because its name is generated dynamically. The `do_action` part is:

```php
do_action( 'woocommerce_after_' . $this->object_type . '_object_save', $this, $this->data_store );
```

It's called in `\WC_Data::save` among other places.

nima-karimi:
Thanks for the review @ianlin. 

> After deleting a Shipping Zone, the job wasn't scheduled immediately.

Good point. I've added the action for this in https://github.com/woocommerce/google-listings-and-ads/pull/1560/commits/b0d8fb805a7b8f0ec0fa945f8e36591c5353734c

> the onboarding will fail if I set up shipping postcodes for a postcodes unsupported country, e.g. TW.

I've modified this part quite a bit in the other related PR #1574. Can you please test this in that PR? It should fix this issue once merged.

ianlin (approved):
Thanks for the updates, tested again and the `update_shipping_settings` job was scheduled right away after a shipping zone was deleted, thanks.

Also I've locally merged the branch in #1574 into this PR and the onboarding passed if I have set up postcodes for the countries that don't support postcodes. I'll start reviewing #1574 then.

PR #1562, opened 2022-06-10
Tree Select Control - Fix control focus when the selected element is not expanded

### LINKED ISSUE #1561: Tree Select Control - Losing focus after clearing search input
### Describe the bug:

Introduced in #1557 

### Steps to reproduce:
1. Run storybook `npm run storybook`
2. On Tree Select Control without expanding any node search for "Kuala"
3. Kuala Lumpur will appear in the tree
4. Select Kuala Lumpur
5. Focus is lost

### Expected behavior:
Focus goes to input search if no node is available. In case node is available focus stays in that node.

https://user-images.githubusercontent.com/5908855/173115709-4bdd830e-acfa-419f-8c14-1e0a7c0a482c.mov

### Actual behavior:
If the element is not visible, the component lost the focus

https://user-images.githubusercontent.com/5908855/173115795-5d11ff9c-e417-4fd2-ad53-845c5d1076b0.mov

puntope:

Closes #1561 

This PR fixes an error in Tree Select Control introduced in https://github.com/woocommerce/google-listings-and-ads/pull/1557

When the Search filter shows a node that is not expanded after selecting it, the node is hidden and the Tree Select Control lost the focus.

To solve this a Ref was created and in case the parent Node is not expanded we set the focus in the input, otherwise we maintain the focus in the node.

### Screenshots:

https://user-images.githubusercontent.com/5908855/173116046-52b0eb26-8d58-4b3e-b07e-9d879792a3aa.mov

1. Run storybook `npm run storybook`
2. On Tree Select Control without expanding any node search for "Kuala"
3. Kuala Lumpur will appear in the tree
4. Select Kuala Lumpur
5. Input is cleared
6. Focus is  again in input search
7. Expand Europe
8. Search for Spain
9. Click on Spain
10. Input is cleared and focus stays in Spain

### Changelog entry

> Fix - Avoid losing focus when selecting an option in Tree Select Control

### DIFF
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -392,6 +392,7 @@ const TreeSelectControl = ( {
 		if ( ! nodesExpanded.includes( option.parent ) ) {
 			controlRef.current.focus();
 		}
+
 	};
 
 	/**
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -111,6 +111,7 @@ const TreeSelectControl = ( {
 	const [ nodesExpanded, setNodesExpanded ] = useState( [] );
 	const [ inputControlValue, setInputControlValue ] = useState( '' );
 
+	const controlRef = useRef();
 	const dropdownRef = useRef();
 	const onDropdownVisibilityChangeRef = useRef();
 	onDropdownVisibilityChangeRef.current = onDropdownVisibilityChange;
@@ -381,7 +382,22 @@ const TreeSelectControl = ( {
 			handleSingleChange( checked, option );
 		}
 
+		clearSearchInput( option );
+	};
+
+	const clearSearchInput = ( option ) => {
 		setInputControlValue( '' );
+		const optionID = option.key || option.value;
+
+		const currentFocusedElement = focus.focusable
+			.find( dropdownRef.current )
+			.filter( ( el ) => el.value === optionID )[ 0 ];
+
+		const treeGroup = currentFocusedElement?.closest( '[role=treegroup]' );
+		const parentId = treeGroup?.querySelector( 'input' );
+		if ( ! nodesExpanded.includes( parentId?.value ) ) {
+			controlRef.current.focus();
+		}
 	};
 
 	/**
@@ -461,6 +477,7 @@ const TreeSelectControl = ( {
 			) }
 
 			<Control
+				ref={ controlRef }
 				disabled={ disabled }
 				tags={ getTags() }
 				isExpanded={ showTree }
--- a/js/src/components/tree-select-control/index.js
+++ b/js/src/components/tree-select-control/index.js
@@ -68,6 +68,7 @@ import { ARROW_DOWN, ARROW_UP, ENTER, ESCAPE, ROOT_VALUE } from './constants';
  * @property {boolean} checked Whether this option is checked.
  * @property {boolean} partialChecked Whether this option is partially checked.
  * @property {boolean} expanded Whether this option is expanded.
+ * @property {boolean} parent The parent of the current option
  *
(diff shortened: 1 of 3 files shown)


js/src/components/tree-select-control/index.js
```diff
@@ -382,6 +389,9 @@ const TreeSelectControl = ( {
 		}
 
 		setInputControlValue( '' );
+		if ( ! nodesExpanded.includes( option.parent ) ) {
```
jorgemd24, line 392: ❓

Why is this needed? For example, If I expand Asia, then Malaysia, and then I search for Kuala Lumpur and I click on Kuala Lumpur, why shouldn't it focus on the search input?

jorgemd24 (commented):
Thanks, @puntope for fixing this. I tested however it is not clear to me why we do not refocus if the option parent is expanded. I left a comment in the code. See video:

https://user-images.githubusercontent.com/2488994/173636128-00255c90-0225-4d53-8ae4-173c91049d0c.mp4

I would expect to be able to search again in the search component.

💅 

Could we test that the function `onFocus` is called when the search term is cleared?

Thanks!

puntope:
> Thanks, @puntope for fixing this. I tested however it is not clear to me why we do not refocus if the option parent is expanded. I left a comment in the code. See video:
> 
>  Screen.Capture.on.2022-06-14.at.19-06-37.mp4 
> I would expect to be able to search again in the search component.
> 
> 💅
> 
> Could we test that the function `onFocus` is called when the search term is cleared?
> 
> Thanks!

If you do that, each time you select/deselect options you will get focus on the input. 

That would be a 🚧  in accessibility. So right now I only give back the focus to the search input in case the checkbox is not visible anymore. As Ant Design Component does.

js/src/components/tree-select-control/index.js
puntope: Same answer as https://github.com/woocommerce/google-listings-and-ads/pull/1562#issuecomment-1155630818 

We only give the focus to the search when the current focus is lost

puntope:
Hey @jorgemd24 I answered your questions :) Let me know if now its clear. 

In brief I we want is to emulate more  https://ant.design/components/tree-select/  we can just do the focus if there is some text in the search input to clear.  To do what Ant design does will require more work. Since it seems to have the focus in the input when searching and in the options when navigating

Since this solves a critical bug on previous PR https://github.com/woocommerce/google-listings-and-ads/pull/1557 I guess this solves that issue and then we can merge and release tomorrow. Then we can open a new discussion/pr  if you think we still need other tweaks enhancements in the component.

Let me know what do you think  

Otherwise I can also rollback to the previous implementation without clearing the input because if not we cannot release other code that is ready.

jorgemd24 (approved):
> In brief I we want is to emulate more https://ant.design/components/tree-select/ we can just do the focus if there is some text in the search input to clear. To do what Ant design does will require more work. Since it seems to have the focus in the input when searching and in the options when navigating

Thanks, @puntope for the clarification, I was confused because when I was comparing with the ant design I saw the difference between both components, see below:

https://user-images.githubusercontent.com/2488994/173684822-c6b1990d-c61e-4766-a6eb-cf32ffd86273.mp4

Anyway, I am approving this PR as I see that fixing the issue with the focus if the element was not visible.

Thanks, @puntope!

puntope:
> Thanks, @puntope for the clarification, I was confused because when I was comparing with the ant design I saw the difference between both components, see below:

Yeah, ANT design has 2 focus controllers. 1 for the search (when you type characters) and other for navigating over the three. We don't have that in Tree Select yet. I created [a new issue](https://github.com/woocommerce/google-listings-and-ads/issues/1566) in case we want to go ahead over that one.  However that will require to replace SPACE as a key for select the checkbox, since SPACE is a valid key in the search input.

PR #1532, opened 2022-05-26
Onboarding policy check client side change

songchen2022:
### Changes proposed in this Pull Request:

This PR will check for the list of info in the policy check.

- GetAllowedCountries
- GetPaymentGateways
- IsSsl
- GetRefundReturnPolicyPageContent

### Screenshots:
Screen Shot 2022-05-25 at 11 51 38 PM

### Detailed test instructions:

As the API endpoints are not available now , we only could test it by running the unit test
`npm run test:js` : Run the JS test suite

We could test it when end to end is ready.

PR #1552, opened 2022-06-06
Use commit instead of branch for storybook dependency

puntope:
### Changes proposed in this Pull Request:

This PR uses a commit version instead of a generic branch for loading `woocommerce-grow-storybook` dependency. Loading just the branch reates integrity inconsistencies.

### Detailed test instructions:

1. Checkout the branch
2. Run `npm i`
3. It finish successfully
4. Run `npm run dev` 
5. It finish successfully
6. You can navigate over all the pages in GLA without compilation errors

PR #1563, opened 2022-06-14
Add PolicyComplianceCheckController 

songchen2022:

This PR is based on [**Onboarding policy check client side change**](https://github.com/woocommerce/google-listings-and-ads/pull/1532)
This PR adds PolicyComplianceCheckController in server side which will communicate with actions(fetchAllowedCountries, fetchSsl, fetchPaymentGateways, fetchReturnRefundPolicyPageContent, etc.) in client side to provide the corresponding data. 

### Screenshots:

image

We only test it by running the unit test
$ vendor/bin/phpunit

In the Dev console:
We could test it works or not by running

`wp.data.resolveSelect( 'wc/gla' ).getPolicyCheck().then((data) => {
    console.log("policy_check", data)
}).catch((error) => console.log("error",error))
`

Screen Shot 2022-07-11 at 10 26 55 PM


songchen2022:
@jorgemd24 Could you please help me take a look this PR when you have any chance? Thanks so much!

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	/**
+	 * Get the allowed countries for the controller.
+	 *
+	 * @return array
+	 */
+	protected function get_allowed_countries(): array {
+		return $this->wc->get_allowed_countries();
+	}
```
jorgemd24: The callback functions should return a function, the same applies for `has_payment_gateways`, `get_is_store_ssl`, `has_refund_return_policy_page`, See example: https://github.com/woocommerce/google-listings-and-ads/blob/fb70c94c21037c37352ee7d8d30d90362b1d6fc4/src/API/Site/Controllers/MerchantCenter/ContactInformationController.php#L84-L100

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'mc/policy_check/allowed_countries',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_allowed_countries(),
```
jorgemd24: Usually, the callback functions are named with the suffix callback, for example, `get_allowed_countries_callback`

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Class PolicyComplianceCheckController
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter
+ */
+class PolicyComplianceCheckController extends BaseController {
+
+	use CountryCodeTrait;
+	use EmptySchemaPropertiesTrait;
```
jorgemd24: We should add a schema for this controller, see the example here: https://github.com/woocommerce/google-listings-and-ads/blob/fb70c94c21037c37352ee7d8d30d90362b1d6fc4/src/API/Site/Controllers/MerchantCenter/ContactInformationController.php#L125 and you can read more about it here: https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+	 * BaseController constructor.
+	 *
+	 * @param RESTServer   $server
+	 * @param WC           $wc
+	 * @param GoogleHelper $google_helper
+	 */
+	public function __construct( RESTServer $server, WC $wc, GoogleHelper $google_helper ) {
+		parent::__construct( $server );
+		$this->wc            = $wc;
+		$this->google_helper = $google_helper;
```
jorgemd24: Are we going to use `$this->google_helper`? it seems that is unused.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+	 * Check if the store has refund return policy page for the controller.
+	 *
+	 * @return bool
+	 */
+	protected function has_refund_return_policy_page(): bool {
+		if ( $this->the_slug_exists( 'refund_returns' ) ) {
+			return true;
+		}
+		return false;
+	}
```
jorgemd24: We should move the logic to a new class in `src/MerchantCenter/PolicyComplianceCheck.php` so we leave the controller as clean as possible.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+
+	/**
+	 * Check if the slug exists or not.
+	 *
+	 * @param PostName $post_name
+	 */
+	protected function the_slug_exists( $post_name ) {
+		global $wpdb;
+
+		if ( $wpdb->get_row( $wpdb->prepare( 'SELECT post_name FROM wp_posts WHERE post_name = %s', $post_name ) ) ) {
```
jorgemd24: Should we not check the status of that page? What happens if it is not published and it is a draft? 

Maybe we could do something like this:

```
$args = [
  'name'        => $the_slug,
  'post_type'   => 'page',
  'post_status' => 'publish',
  'numberposts' => 1
];
$my_posts = get_posts($args);
```

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+	protected function get_schema_title(): string {
+		return 'policy_check';
+	}
+
+	/**
+	 * Check if the slug exists or not.
+	 *
+	 * @param PostName $post_name
+	 */
+	protected function the_slug_exists( $post_name ) {
```
jorgemd24: Missing the return type

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\CountryCodeTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\EmptySchemaPropertiesTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\GoogleHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
+use WC_Install;
```
jorgemd24: This seems that is not used.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+	protected function get_allowed_countries(): array {
+		return $this->wc->get_allowed_countries();
+	}
+
+	/**
+	 * Check if the payment gateways is empty or not for the controller.
+	 *
+	 * @return bool
+	 */
+	protected function has_payment_gateways(): bool {
```
jorgemd24: I believe this will be implemented in future PR, but just in case, I believe we will need to check which available payment gateways are valid, I think Google does not accept the store if it is using bank transfer for example.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+		return true;
+	}
+
+	/**
+	 * Check if the store is using SSL for the controller.
+	 *
+	 * @return bool
+	 */
+	protected function get_is_store_ssl(): bool {
+		return is_ssl();
```
jorgemd24: For the cases where the websites are behind a load balancer and possibly without SSL, should we return false?

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,168 @@
...
+			]
+		);
+	}
+
+	/**
+	 * Get the allowed countries for the controller.
+	 *
+	 * @return array
+	 */
+	protected function get_allowed_countries(): array {
```
jorgemd24: I was reading the PRD, but I can't find anything related to allowed countries, how is this going to be used? Shouldn't this be allowed to countries by GMC?

jorgemd24 (commented):
@songchen2022 thanks for this PR, I left some comments about the code and I see that currently, the client can not make calls to the controller for two reasons:

1. The controller callbacks should be a function.
2. The client is expecting to receive a response of the following structure: `data.policy_check.allowed_countries`.

if you run the following code in your browser console, you will the error in the reducer.js:

```
wp.data.resolveSelect( 'wc/gla' ).getIsStoreSsl().then((data) => {
    console.log("data", data)
}).catch((error) => console.log("error",error))
```

I was reading the PR and I see that there is a section related to `validate common destination URL`, shouldn't we add this checking as well? 

Also, it is missing the REST controllers tests, in the following link there are some examples:

https://github.com/woocommerce/google-listings-and-ads/blob/10cc3fdf568abe66d450c3eb7afb0c5f1b55d9e1/tests/Unit/API/Site/Controllers/MerchantCenter/PhoneVerificationControllerTest.php#L22

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Thanks for the suggestions. Updated.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Updated.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Deleted.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Thanks for the suggestions. Done.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Updated.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: We are planning to check if the store is live and accessible to all users. There is one check for this policy which is to check for access geo restrictions. Then we would like to check if the store set up the allowed countries. Please correct me if I'm wrong.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: You're right. This is the corner case. We are planning to check if the store has a secure checkout process through checking the ssl. Do you have any good suggestions? Thanks!

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: That's great suggestions! Thanks! We are planning to add this payment gateways check and auto fix feature with this valid check.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Done.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Thanks for your suggestions! Checking the status of the page is very helpful. Woo doesn't require a refunds and returns policy page to be setup. And the page slug can be translated. Therefore, we don't check the detailed info of the returns and refunds page here. Please correct me if I'm wrong.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Updated.

jorgemd24:
@songchen2022, let me know when this is ready for another round 🙂  I see some JS tests are failing because of the `tarball data for woocommerce-grow-storybook`, this was fixed in PR https://github.com/woocommerce/google-listings-and-ads/pull/1552. Maybe you would like to merge `develop` into this branch or rebase it.

songchen2022:
> 

Thanks so much for your suggestions and reviewing! Really appreciate that. I've add the unit test part. For  "validate common destination URL", I'm not sure if you mean that checking "My store is live and accessible to all users". This is the [design doc](https://docs.google.com/document/d/1qDunog-GagI9xXhecyPH_SllRc8003sXWn9-kIBPVE4/edit#) and this is [checking](https://docs.google.com/document/d/1qDunog-GagI9xXhecyPH_SllRc8003sXWn9-kIBPVE4/edit#bookmark=id.3g9qjj59zg9q) what we will add. Welcome any comments and suggestions.

songchen2022:
@jorgemd24 Could you please help me take a look again when you have any chance? Thanks!

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,198 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\CountryCodeTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\EmptySchemaPropertiesTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\GoogleHelper;
```
jorgemd24: This import `GoogleHelper` is unused.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,198 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\PolicyComplianceCheck;
+use Exception;
+use WP_REST_Request as Request;
+use WP_REST_Response as Response;
```
jorgemd24, line 12: `WP_REST_Response` is not used.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,105 @@
...
+namespace Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Merchant;
+use Automattic\WooCommerce\GoogleListingsAndAds\Exception\InvalidValue;
+use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
+use Automattic\WooCommerce\GoogleListingsAndAds\Utility\ISOUtility;
+use Automattic\WooCommerce\GoogleListingsAndAds\Value\PhoneNumber;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
+use Google\Service\Exception as GoogleServiceException;
```
jorgemd24: Multiple imports are unused.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,198 @@
...
+		);
+	}
+
+	/**
+	 * Get the allowed countries for the controller.
+	 *
+	 * @return callable
+	 */
+	protected function get_allowed_countries_callback(): callable {
+		return function ( Request $request ) {
```
jorgemd24: If we are not going to use `Request`, it can be removed from this callback and the other callbacks.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,198 @@
...
+		$this->register_route(
+			'mc/policy_check/refund_return_policy',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->has_refund_return_policy_page_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+			]
+		);
```
jorgemd24: Do you think that it is necessary to have an endpoint for each check? Or can we use one endpoint like `mc/policy_check` that returns all the checks?
 ```
 {
    allowed_countries: ....,
    payment_gateways: ....,
    store_ssl: ....,
    refund_return_policy: ....
}
```

And from the client we do only one request instead of 4 requests?

File: src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php [hunk 17]
```diff
@@ -0,0 +1,198 @@
...
+				'description' => __( 'The store ssl associated with onboarding policy checking.', 'google-listings-and-ads' ),
+				'context'     => [ 'view' ],
+			],
+			'refund_returns'    => [
+				'type'        => 'boolean',
+				'description' => __( 'The refund returns policy associated with onboarding policy checking.', 'google-listings-and-ads' ),
+				'context'     => [ 'view' ],
+			],
+		];
+	}
```
jorgemd24: If we use one endpoint as I mentioned before, we could use this schema for the full endpoint.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
jorgemd24: If we use the schema this is not needed. `use EmptySchemaPropertiesTrait;`

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,105 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class PolicyComplianceCheck
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter
+ *
+ * @since 1.5.0
```
jorgemd24: This since is incorrect. We should use the placeholder `@since x.x.x` so it is replaced automatically

tests/Unit/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckControllerTest.php
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @property RESTServer                   $rest_server
+ * @property PolicyComplianceCheck|MockObject $policy_check
+ */
+class PolicyComplianceCheckControllerTest extends RESTControllerUnitTest {
+
+	protected const ALLOWED_COUNTRIES     = '/wc/gla/mc/policy_check/allowed_countries';
+	protected const STORE_SSL     = '/wc/gla/mc/policy_check/store_ssl';
+	protected const PAYMENT_GATEWAYS     = '/wc/gla/mc/policy_check/payment_gateways';
+	protected const REFUND_RETURN     = '/wc/gla/mc/policy_check/refund_return_policy';
```
jorgemd24: I know that tests are excluded from PHPCS but we should at least follow the same identations so all the code looks well structure. 

https://github.com/woocommerce/google-listings-and-ads/blob/2e56301df9695200ff6b3207d224272b76370a18/phpcs.xml.dist#L117

jorgemd24 (commented):
@songchen2022 thanks for the adjustments! I left some comments about the code.

Besides that, I am wondering why did you choose to make an endpoint for each check? I think the client could fetch this information using only one request instead of 4 requests, what do you think?

Also at the moment, the client is not ready to communicate with the endpoints as it is expecting to receive a different type of response. As I mentioned before, you could test it without the UI using the following code in your dev console:

```
wp.data.resolveSelect( 'wc/gla' ).getIsStoreSsl().then((data) => {
    console.log("data", data)
}).catch((error) => console.log("error",error))
```

When I ran the previous code, I got the following error:
`error TypeError: Cannot read properties of undefined (reading 'policy_check')`

The package-lock.json should not be committed in this PR. 

I have not reviewed in deep the logic of the `PolicyComplianceCheck` as I understand that the intention of this PR is to create the endpoints and the connection between the client and the API, I believe that in future PRs the logic will be updated to match with the PRD. Please, let me know if my understanding is correct. 

Thanks again!

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Deleted.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Deleted.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Deleted.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Good suggestions! Updated.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Updated.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Updated.

tests/Unit/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckControllerTest.php
songchen2022: Updated.

songchen2022:
> am wondering why did you choose to

Thanks for your suggestions. I've update the endpoints to one endpoint for the whole policy check. And upload the console test screenshot in the description. Could you please help me to rebase [feature/policy-compliance-checks] branch with the develop branch and help me take a look again when you have any chance?

jorgemd24:
@songchen2022, 

>Could you please help me to rebase [feature/policy-compliance-checks] branch with the develop branch

It is not needed to merge "develop" branch into this PR that often. We did it before, because the unit tests were failing and the fix was in the develop branch, anyway I have merged the develop branch into the feature branch and it is now showing the correct files in this PR.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,126 @@
...
+	 */
+	protected function get_policy_check_callback(): callable {
+		return function () {
+			try {
+				return [
+					'policy_check' => array_merge(
+						$this->policy_compliance_check->get_allowed_countries(),
+						$this->policy_compliance_check->has_payment_gateways(),
+						$this->policy_compliance_check->get_is_store_ssl(),
+						$this->policy_compliance_check->has_refund_return_policy_page(),
```
jorgemd24: What do you think if the policy compliance methods returns the data instead of returning an array with the response format, for example `get_is_store_ssl` should return a boolean - true or false instead of an array-object: ['ssl' => `true` ].

`get_policy_check_callback` function could look likes this:

return [
    allowed_countries => $this->policy_compliance_check->get_allowed_countries(),
    payment_gateways  => $this->policy_compliance_check->get_allowed_countries(),
    .... etc
]

See example: https://github.com/songchen2022/google-listings-and-ads/blob/b0d1ec71792f663c9de5fe8c8bd2ad4f12ce75ad/src/API/Site/Controllers/MerchantCenter/ProductFeedController.php#L64-L68

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,99 @@
...
+		return [ 'payment_gateways' => true ];
+	}
+
+	/**
+	 * Check if the store is using SSL for the controller.
+	 *
+	 * @return array
+	 */
+	public function get_is_store_ssl(): array {
+		return [ 'ssl' => is_ssl() ];
```
jorgemd24: In the schema we set that this field should be named `store_ssl` however here it is `ssl`.

https://github.com/songchen2022/google-listings-and-ads/blob/b0d1ec71792f663c9de5fe8c8bd2ad4f12ce75ad/src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php#L102

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,99 @@
...
+	/**
+	 * Check if the store has refund return policy page for the controller.
+	 *
+	 * @return array
+	 */
+	public function has_refund_return_policy_page(): array {
+		if ( $this->the_slug_exists( 'refund_returns' ) ) {
+			return [ 'refund_return_policy_page' => true ];
+		}
+		return [ 'refund_return_policy_page' => false ];
```
jorgemd24: Similar thing than before, in the schema we set that this field should be named refund_returns however here it is refund_return_policy_page

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,99 @@
...
+		$this->wc = $wc;
+	}
+
+	/**
+	 * Get the allowed countries for the controller.
+	 *
+	 * @return array
+	 */
+	public function get_allowed_countries(): array {
+		return [ 'allowed_countries' => $this->wc->get_allowed_countries() ];
```
jorgemd24: Is it necessary to send the full list of allowed countries? Or can we check if the store's country is in the allowed countries list?

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,126 @@
...
+	/**
+	 * Get the allowed countries, payment gateways info, store ssl and refund return policy page for the controller.
+	 *
+	 * @return callable
+	 */
+	protected function get_policy_check_callback(): callable {
+		return function () {
+			try {
+				return [
+					'policy_check' => array_merge(
```
jorgemd24: I do not think it is needed to wrap the response inside of `policy_check` because the endoint is alrady mentioning that the response is the `mc/policy_check`. The response could look like this:

{
    allowed_countries: ...,
    payment_gateways: ....,
    .....
}

If we do the change, the schema should be updated too.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,99 @@
...
+			return [ 'refund_return_policy_page' => true ];
+		}
+		return [ 'refund_return_policy_page' => false ];
+	}
+
+
+	/**
+	 * Check if the slug exists or not.
+	 *
+	 * @param PostName $post_name
```
jorgemd24: `PostName` is not a valid value here and it is not defined, I think this should be replace with `string`

js/src/hooks/usePolicyCheck.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * ```json
+ * {
+ * 		 "AR": {
+ * 			"name": "Argentina",
+ * 			"currency": "ARS"
+ * 		}
+ * }
+ * ```
+ */
```
jorgemd24: This comments are incorrect, the response is now different.

js/src/hooks/usePolicyCheck.js
```diff
@@ -0,0 +1,25 @@
...
+const usePolicyCheck = () => {
+	return useAppSelectDispatch( 'usePolicyCheck' );
```
jorgemd24: useAppSelectDispatch is expecting as first parameter the selector's name, in this case should be `getPolicyCheck`

jorgemd24 (commented):
@songchen2022, Thanks for the changes, it is looking good. I think that having a single endpoint simplifies the code. The package-lock.json should not be committed as we have not installed any new dependency.

When we progress with the logic for the compliance check, I guess in a different PR, we will need to add unit tests for the class `PolicyComplianceCheck`.

Also, I see that there are some conflicts, can you resolve them when you have time? Thanks!

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: The schema has updated based on the response.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: To check the store's country sounds more efficient. By that way, we might need add the store's country in the request. Could we update this later to improve after we done the UI?

songchen2022:
> @songchen2022, Thanks for the changes, it is looking good. I think that having a single endpoint simplifies the code. The package-lock.json should not be committed as we have not installed any new dependency.
> 
> When we progress with the logic for the compliance check, I guess in a different PR, we will need to add unit tests for the class `PolicyComplianceCheck`.
> 
> Also, I see that there are some conflicts, can you resolve them when you have time? Thanks!

Thanks so much for taking time to review!  I will add the unit test for PolicyComplianceCheck in a different PR. Could you please help me take a look again?

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,126 @@
...
+	protected function get_policy_check_callback(): callable {
+		return function () {
+			try {
+				return [
+					'policy_check' => [
+						'allowed_countries' => $this->policy_compliance_check->get_allowed_countries(),
+						'payment_gateways'  => $this->policy_compliance_check->has_payment_gateways(),
+						'store_ssl'         => $this->policy_compliance_check->get_is_store_ssl(),
+						'refund_returns'    => $this->policy_compliance_check->has_refund_return_policy_page(),
+					],
```
jorgemd24: It is not needed to send the response inside of the policy_check, you can just return the array like this:

				return [
						'allowed_countries' => $this->policy_compliance_check->get_allowed_countries(),
						'payment_gateways'  => $this->policy_compliance_check->has_payment_gateways(),
						'store_ssl'         => $this->policy_compliance_check->get_is_store_ssl(),
						'refund_returns'    => $this->policy_compliance_check->has_refund_return_policy_page(),
					],

src/MerchantCenter/PolicyComplianceCheck.php
jorgemd24: I think it would make more sense to add it in this PR so the UI knows what type of data will receive from the API endpoint `mc/policy_check`, I think we should return a boolean here depending if the store's country is in the allowed country list. 


>By that way, we might need add the store's country in the request

You could get the store's country using the following function: [wc_get_base_location](https://woocommerce.github.io/code-reference/namespaces/default.html#function_wc_get_base_location)

jorgemd24 (commented):
@songchen2022, thanks for all the changes, I have just left a couple of comments that may be worth it to discuss.

Thanks again for your work!

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,122 @@
...
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\CountryCodeTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\PolicyComplianceCheck;
+use Exception;
+use WP_REST_Request as Request;
```
jorgemd24: This import `WP_REST_Request` is unused.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,122 @@
...
+	public function register_routes(): void {
+		$this->register_route(
+			'mc/policy_check',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_policy_check_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+			]
```
jorgemd24, line 56: Missing the schema callback: 

`'schema' => $this->get_api_response_schema_callback()`

See example: https://github.com/songchen2022/google-listings-and-ads/blob/7ab93b44532479442ee504c121d77afd36136368/src/API/Site/Controllers/Ads/BudgetRecommendationController.php#L63

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
```diff
@@ -0,0 +1,122 @@
...
+	 */
+	protected function get_policy_check_callback(): callable {
+		return function () {
+			try {
+				return [
+					'allowed_countries' => $this->policy_compliance_check->get_allowed_countries(),
+					'payment_gateways'  => $this->policy_compliance_check->has_payment_gateways(),
+					'store_ssl'         => $this->policy_compliance_check->get_is_store_ssl(),
+					'refund_returns'    => $this->policy_compliance_check->has_refund_return_policy_page(),
+				];
```
jorgemd24: The response needs to be wrapped inside of the Response class.

```
Example: new Response ([
  allowed_countries...
  payment_gateways....
  etc....
]);
```

See example: https://github.com/songchen2022/google-listings-and-ads/blob/7ab93b44532479442ee504c121d77afd36136368/src/API/Site/Controllers/Ads/AccountController.php#L99
See more info about this here: https://developer.wordpress.org/plugins/responses-2/

jorgemd24 (commented):
Thanks, @songchen2022 for the adjustments. I left three comments about the code. 

As was discussed before, the requirements for the `allowed_countries` check needs to be clarified with the Google team as per the [doc](https://docs.google.com/document/d/1qDunog-GagI9xXhecyPH_SllRc8003sXWn9-kIBPVE4/edit#heading=h.wttttrv5ie69) and I understand that this logic will be implemented in a different PR. 

Also, can you please not merge develop branch into this PR if it is not needed? It makes the review harder because it is adding files from different PRs and I needed to merge  `develop` branch into the feature branch so I get the correct files.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Thanks for your suggestions. Updated.

src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php
songchen2022: Thanks for pointing this out. Updated.

songchen2022:
@jorgemd24 Thanks for reviewing. I updated and added the country checking. Could you please help to take a look again? Thanks so much!

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,115 @@
...
+		$all_countries = $this->wc->get_countries();
+		$mc_countries  = $this->google_helper->get_mc_supported_countries();
+
+		foreach ( $mc_countries as $country ) {
+			if ( ! array_key_exists( $country, $all_countries ) ) {
+				return false;
+			}
+		}
+		return true;
+	}
```
mikkamp: Would you be able to clarify how this check is supposed to work. `get_countries` just returns a list of all the countries in WC, and `get_mc_supported_countries` just returns a hardcoded list of countries we support. It is possible to filter the list of countries (although this isn't common), but for most sites comparing those two lists of countries will always return true.

In the document it seems to be referring to the function `get_allowed_countries` which can differ per site because they can either include or exclude a list of countries they don't wish to sell to. However blocking a product from being sold in a country only prevents them from checking out with a billing/shipping address matching the blocked country. The landing pages for the product are still visible.

So doesn't it make more sense to compare allowed countries with target audience countries? That will at least confirm that all the target countries they wish to advertise to aren't being blocked from being sold to.

It still doesn't confirm if the landing pages are visible for those countries, but that seems to need some more extensive checking (possibly through an external service)

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * Check if the store is using SSL for the controller.
+	 *
+	 * @return bool
+	 */
+	public function get_is_store_ssl(): bool {
+		return is_ssl();
```
mikkamp: While this is the correct way to check if it's using SSL, what this is actually doing is checking if the current admin page is using SSL. Now that SSL is a lot more common almost every site has both frontend and backend as the same, but it is still possible to [setup WordPress](https://developer.wordpress.org/reference/functions/force_ssl_admin/) with only the admin area using SSL.

It seems what we intend to check here is that the product landing pages are using SSL. In that case wouldn't it be best to use `PluginHelper::get_site_url` and confirm that the URL is https.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Check if the store has refund return policy page for the controller.
+	 *
+	 * @return bool
+	 */
+	public function has_refund_return_policy_page(): bool {
+		if ( $this->the_slug_exists( 'refund_returns' ) ) {
```
mikkamp: This works for English sites that published the page. Should we run this through the same translation function as WooCommerce to allow it to support languages that translate this slug?

```
_x( 'refund_returns', 'Page slug', 'woocommerce' )
```

src/Proxies/WC.php
```diff
@@ -47,11 +47,12 @@ class WC {
...
-		$countries          = $countries ?? new WC_Countries();
-		$this->wc_countries = $countries;
-		$this->base_country = $countries->get_base_country() ?? 'US';
-		$this->countries    = $countries->get_countries() ?? [];
-		$this->continents   = $countries->get_continents() ?? [];
+		$countries               = $countries ?? new WC_Countries();
+		$this->wc_countries      = $countries;
+		$this->base_country      = $countries->get_base_country() ?? 'US';
+		$this->countries         = $countries->get_countries() ?? [];
+		$this->allowed_countries = $countries->get_allowed_countries() ?? [];
```
mikkamp, line 54: We added this helper variable and function, but in the end it's currently not being used. I see there was some back and forth about whether this was needed, so I guess this should only be left in if we are going to actually use it.

js/src/data/test/reducer.test.js
```diff
@@ -418,87 +413,42 @@ describe( 'reducer', () => {
...
 				data,
 			};
 			const state = reducer( prepareState(), action );
 
 			state.assertConsistentRef();
 			expect( state ).toHaveProperty(
-				'mc.policy_check.store_ssl',
-				data.policy_check.store_ssl
+				'mc.policy_check.refund_returns',
+				data.refund_returns
```
mikkamp, line 432: This test is going to pass because we return data in the request and then compare if it matches the same data. But since we changed the code in the PHP it should be looking for a value that's `true` here. Wouldn't it be better to explicitly test for `true` or `false` here? Same for checking all 4 of the fields.

js/src/data/test/reducer.test.js
```diff
@@ -418,87 +413,42 @@ describe( 'reducer', () => {
...
-			const data = {
-				policy_check: {
-					store_ssl: true,
+				store_ssl: true,
+				payment_gateways: {
+					id: 'wc_custom_pg',
+					title: 'Custom Payment Gateway',
+					method_description: 'Description of the payment gateway',
 				},
+				refund_returns: 'Refund and Returns Policy',
```
mikkamp: These values still pass in the test, but they don't match up with what it was changed to in the PHP code. Each of these properties should return a boolean value. Same for the `allowed_countries`.

tests/Unit/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckControllerTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+		$response = $this->do_request( self::POLICY_CHECK, 'GET', [] );
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals(
+		[
+			'allowed_countries'    	=> true,
+			'store_ssl'         	=> true,
+			'payment_gateways'  	=> true,
+			'refund_returns' 	=> true,
```
mikkamp: Not a big deal, since we don't run PHPCS on test code. But it does look a bit nicer if the spacing is used correctly.

mikkamp (commented):
Thanks @songchen2022 for all the additional updates and changes. I went through the latest changes, and added some comments. It would be helpful to get some more clarity on what we actually want to check with the allowed_countries check.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Thanks for pointing this out. We definitely need to compare the allowed countries with target audience countries. If the audience country is in the allowed country list, in my understanding the landing page has to be shown. Updated this to wc_get_allowed_countries.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Thanks for your suggestions. Updated.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Updated.

js/src/data/test/reducer.test.js
songchen2022: Updated.

src/Proxies/WC.php
songchen2022: I added allowed countries here and now use it in the policyComplianceCheck file for checking if the mc county list is in this allowed country list.

js/src/data/test/reducer.test.js
songchen2022: Updated.

songchen2022:
@mikkamp Thanks so much for reviewing! Could you please help to take a look again?

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	/**
+	 * Check if the store website is accessed by all users for the controller.
+	 *
+	 * @return bool
+	 */
+	public function is_accessible(): bool {
+		$all_allowed_countries = $this->wc->get_allowed_countries();
+		$mc_countries          = $this->google_helper->get_mc_supported_countries();
```
mikkamp: You mentioned:
> We definitely need to compare the allowed countries with target audience countries.

However `get_mc_supported_countries` gets all the countries that are supported by the extension, not the list of target countries that the user selected during onboarding. Looks like an instance of `TargetAudience` and calling the function `get_target_countries` is needed here.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,124 @@
...
+	 * @return bool
+	 */
+	public function get_is_store_ssl(): bool {
+		global $wp_filesystem;
+		WP_Filesystem();
+		$ssl_check = $wp_filesystem->get_contents( 'ssl://' . $this->get_site_url(), 443, $errno, $errstr, 30 );
+		if ( empty( $ssl_check ) ) {
+			return false;
+		}
+		return ! ! $ssl_check;
```
mikkamp: I personally don't think that this is the best way to perform the check as this looks like it would have to explicitly have ssl:// enabled which might not be true for all hosts. If we actually want to load a copy of the page, I would again recommend to use functions like [wp_remote_get](https://developer.wordpress.org/apis/handbook/making-http-requests/). However in this case I think it's sufficient to just check the URL, similar to how it's done in the function [wp_is_site_url_using_https](https://developer.wordpress.org/reference/functions/wp_is_site_url_using_https/), but then instead pass the URL from `$this->get_site_url()`, something like this:

```suggestion
		return 'https' === wp_parse_url( $this->get_site_url(), PHP_URL_SCHEME );
```

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+
+	/**
+	 * Check if the store has refund return policy page for the controller.
+	 *
+	 * @return bool
+	 */
+	public function has_refund_return_policy_page(): bool {
+		if ( $this->the_slug_exists( _x( 'refund_returns', 'Page slug', 'google-listings-and-ads' ) ) ) {
```
mikkamp: Even though PHPCS recommends to change the text domain name, in this case we need to keep the text domain name as `woocommerce` otherwise we won't be making use of what the slug is translated to in WooCommerce (the GLA translation files could have a different slug or not translated it). So we'll need to have PHPCS ignore the different text domain name, I'd suggest to also add a comment to clarify what we are doing.

```suggestion
		// Check the slug as it's translated by the "woocommerce" text domain name.
		// phpcs:ignore WordPress.WP.I18n.TextDomainMismatch
		if ( $this->the_slug_exists( _x( 'refund_returns', 'Page slug', 'woocommerce' ) ) ) {
```

mikkamp (commented):
Thanks for the changes, it's looking better. I just left a few comments on some additional things to change.

I also noticed we only have a unit test for `PolicyComplianceCheckController` and not `PolicyComplianceCheck`. Since in the controller we have all the functions in the service just return true, and then compare if true was returned there is little that we test there besides confirming that the route was registered and the request can be sent. Would it be possible to get some code up to confirm the `PolicyComplianceCheck` class is returning the expected results?

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: get_mc_supported_countries gets all the [countries](https://support.google.com/merchants/answer/160637?hl=en) that are supported by the extension. get_allowed_countries will get the list of countries that the merchant selected during onboarding. We want to make sure all the supported countries is in the list of get_allowed_countries. However, an instance of TargetAudience makes more sense.

songchen2022:
> Thanks for the changes, it's looking better. I just left a few comments on some additional things to change.
> 
> I also noticed we only have a unit test for `PolicyComplianceCheckController` and not `PolicyComplianceCheck`. Since in the controller we have all the functions in the service just return true, and then compare if true was returned there is little that we test there besides confirming that the route was registered and the request can be sent. Would it be possible to get some code up to confirm the `PolicyComplianceCheck` class is returning the expected results?

Thanks @mikkamp for reviewing! As discussed before in this PR, we would like to add unit tests for PolicyComplianceCheck. I'll put the PR link here when it is ready.

Could you please take a look this PR again? Thanks!

mikkamp (approved):
Thanks for all the additional changes, they all seem to be working as intended. So I'll go ahead and approve and merge it to the feature branch.

PR #1569, opened 2022-06-20
Workaround for CES component to be compatible with WC 6.6.0 and older versions

jorgemd24:

While I was testing other PR with WC 6.6.0 (which was released last week), I realized that the new version of WC was breaking the GLA app, the reason for this is that the package `@woocommerce/customer-effort-score` does not have a default export anymore. Currently, we are importing the component as a default: 

https://github.com/woocommerce/google-listings-and-ads/blob/43ab87ae36c7b1a94147d20af567145766844767/js/src/components/customer-effort-score-prompt/index.js#L5

The usage instructions of `@woocommerce/customer-effort-score` are incorrect too:
https://github.com/woocommerce/woocommerce/blob/6.6.0/packages/js/customer-effort-score/README.md#customereffortscore-component

See here the difference between versions:

- WC 6.6.0 - https://github.com/woocommerce/woocommerce/blob/6.6.0/packages/js/customer-effort-score/src/index.ts
- WC 6.5.1 - https://github.com/woocommerce/woocommerce/blob/6.5.1/packages/js/customer-effort-score/src/index.js

### Screenshots:

https://user-images.githubusercontent.com/2488994/174679022-bb2eaac1-2857-461c-ad77-65aba8042569.mp4

1. Check out this branch.
2. Install the latest version of WC [6.6.0](https://downloads.wordpress.org/plugin/woocommerce.6.6.0.zip) 
3. Go to wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&guide=campaign-creation-success
4. Click on "Got it"
5. No errors should be displayed.

Repeat the same process with [WC 6.5.1](https://downloads.wordpress.org/plugin/woocommerce.6.5.1.zip)

You would get the error using develop branch with the latest WC 6.6.0

### Changelog entry

> Fix - Compatibility CES prompts with WC 6.6.0

### DIFF
--- a/js/src/components/customer-effort-score-prompt/index.js
+++ b/js/src/components/customer-effort-score-prompt/index.js
@@ -14,9 +14,10 @@ import { LOCAL_STORAGE_KEYS } from '.~/constants';
 import localStorage from '.~/utils/localStorage';
 import useEffectRemoveNotice from '.~/hooks/useEffectRemoveNotice';
 
-// WC 6.6.0 uses @woocommerce/customer-effort-score v2.0.1 which does not include a default export, therefore
+// WC 6.6.0 updated the package @woocommerce/customer-effort-score, which does not include a default export anymore, therefore
 // breaking the page for older versions of WC. This is a temporal workaround to be compatible with older WC versions
 // and with our L-2 policy.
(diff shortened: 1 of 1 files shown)


eason9487 (approved):
LGTM.

PR #1574, opened 2022-06-23
Generate random ID for postcode regions when syncing shipping settings

nima-karimi:

To sync the shipping rates for postal codes of a country, we need to define them as "[regions](https://developers.google.com/shopping-content/reference/rest/v2.1/regions#Region)" through the Google API. Even though we do not need to do this explicitly, Google automatically creates "[regions](https://developers.google.com/shopping-content/reference/rest/v2.1/regions#Region)" when we present the postal codes as part of the [ShippingSettings API under `postalCodeGroups`](https://developers.google.com/shopping-content/guides/shippingsettings#set_postal_code_groups). 

The ["name" property of PostalCodeGroup](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#PostalCodeGroup) does not have any explicit requirements defined for it. However, since Google maps PostalCodeGroups to Regions, the [PostalCodeGroup's](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#PostalCodeGroup) `name` parameter maps directly to a [Region's](https://developers.google.com/shopping-content/reference/rest/v2.1/regions#Region) `regionId` parameter.

The `regionId` parameter has some [requirements defined by Google](https://support.google.com/merchants/answer/9698880?hl=en#requirements). One of the main requirements is the following:

> For privacy reasons, the region ID value must be a randomized set of numbers (minimum 6 digits), rather than a readable name, such as "NY_City". Readable names or names that look like postal codes aren't accepted.

Previously, we added the country and state information in all region IDs. However, in this PR, the `regionId` parameter is set to a randomly generated number containing >= 6 digits.

Note that this random ID is generated on the fly whenever the user syncs their shipping settings and we do not store this ID in the local DB. This means that whenever the shipping settings are updated, a new random ID will be generated for each region. 

The above won't affect our current functionality because we do not use this ID to identify a region in our code. However, another [requirement](https://support.google.com/merchants/answer/9698880?hl=en#requirements) is:

> You’re able to accept the region ID parameter passed by Google into your product landing pages and render the correct regional product pages with corresponding regional pricing and availability.

This functionality is not implemented right now and I don't believe we have any plans for it. Moreover, all shipping data is stored on the Google side and we can retrieve them from the [`regions` API](https://developers.google.com/shopping-content/reference/rest/v2.1/regions) if we ever need to identify the regions.

1. Make sure you have a shipping zone for the "US" country that includes zip codes and a flat rate method.
2. Edit your GLA free listings settings and choose the "Recommended" settings for the shipping option.
3. Save the free listing so that the shipping settings are synced to MC.
4. Open your MC and go to the shipping & return settings.
5. Check the details of the shipping service that was just synced. It will have a name similar to "[63a5] Google Listings and Ads generated service - US"
6. Go to the "Cost" section and click on "All products"
7. Under the "Destination" column of "Shipping rate table" there should be a random number with more than 6 digits.
8. Click on the pencil mark (Edit) right next to that random number to open the "Edit destination" popup.
9. Confirm that the postcodes you entered in your shipping zone all appear under "Postal codes" in that popup.  

### Changelog entry

> Tweak - Generate random ID for postcode regions when syncing shipping settings

### DIFF
--- a/src/Shipping/GoogleAdapter/PostcodesRateGroupAdapter.php
+++ b/src/Shipping/GoogleAdapter/PostcodesRateGroupAdapter.php
@@ -30,7 +30,12 @@ class PostcodesRateGroupAdapter extends AbstractRateGroupAdapter {
 		$postal_codes = [];
 		$rows         = [];
 		foreach ( $location_rates as $location_rate ) {
-			$postcode_name                  = $location_rate->get_location()->get_postcode_group_name();
+			$region = $location_rate->get_location()->get_shipping_region();
+			if ( empty( $region ) ) {
+				continue;
+			}
+
+			$postcode_name                  = $region->get_id();
 			$postal_codes[ $postcode_name ] = $postcode_name;
 
 			$rows[ $postcode_name ] = new Row( [ 'cells' => [ $this->create_value_object( $location_rate->get_shipping_rate()->get_rate(), $currency ) ] ] );
--- a/src/Shipping/GoogleAdapter/WCShippingSettingsAdapter.php
+++ b/src/Shipping/GoogleAdapter/WCShippingSettingsAdapter.php
@@ -188,12 +188,13 @@ class WCShippingSettingsAdapter extends AbstractShippingSettingsAdapter {
 
 		foreach ( $location_rates as $location_rate ) {
 			$location = $location_rate->get_location();
-			if ( empty( $location->get_postcodes() ) ) {
+			if ( empty( $location->get_shipping_region() ) ) {
 				continue;
 			}
+			$region = $location->get_shipping_region();
 
 			$postcode_ranges = [];
-			foreach ( $location->get_postcodes() as $postcode_range ) {
+			foreach ( $region->get_postcode_ranges() as $postcode_range ) {
 				$postcode_ranges[] = new PostalCodeRange(
 					[
 						'postalCodeRangeBegin' => $postcode_range->get_start_code(),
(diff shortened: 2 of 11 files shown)


ianlin (commented):
Thanks for adding this and the good findings on

> the [PostalCodeGroup's](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#PostalCodeGroup) name parameter maps directly to a [Region's](https://developers.google.com/shopping-content/reference/rest/v2.1/regions#Region) regionId parameter.

and 

> For privacy reasons, the region ID value must be a randomized set of numbers (minimum 6 digits), rather than a readable name, such as "NY_City". Readable names or names that look like postal codes aren't accepted.
> 
> Previously, we added the country and state information in all region IDs. However, in this PR, the regionId parameter is set to a randomly generated number containing >= 6 digits.

I've tested locally and it worked as expected, I can see on the MC the region ID is >= 6 digits and the postcodes are displayed correctly.

However there was an error when syncing to Google if the combination is:

Regions: `US` and `California, US`
Postcodes: `94110`

gla-US-CA-94110

gla-US-CA-94110-error

I understand that one can only choose `Destination by postal code` or `Destination by state` from the MC but not both, and this combination may not be common in the real world. Then I tested some different combinations and the results are shown below:

## Two Regions without Postcodes

gla-US-CA

It will create two records in MC:

gla-US-CA-MC

gla-US-CA-MC-ALL

gla-US-CA-MC-CA

## One Region (`US`) with Postcodes

gla-US-94110

Expected results according to this PR.

gla-US-94110-MC-94110

## One Region (`California, US`) with Postcodes

gla-CA-94110

Expected results according to this PR.

gla-CA-94110-MC-94110

## Using `develop` branch

All of the above combinations (including the one returns error) have the same result on MC:

Screenshot 2022-06-27 at 18 14 25

nima-karimi:
https://github.com/woocommerce/google-listings-and-ads/pull/1574/commits/49e4ee96499f16f9433c061b4559b815594a46fc should fix the issue with duplicate postcodes.

Ready for another review, @ianlin  🙏

ianlin (approved):
Thanks for the changes, I've tested again and the error has gone, I can see the postcodes are synced in the MC. 👍

PR #1584, opened 2022-07-04
Normalize image urls

### LINKED ISSUE #1582: Valid product image URLs not being detected in Google Merchant Center
### Describe the bug:

Several users are reporting their products not syncing due to invalid image URLs error. GMC is unable to detect product images on some sites and return the following errors:

`[imageLink] This value is not a valid URL.`

`[additionalImageLinks[0]] This value is not a valid URL.`

There's been a number of support requests related to this issue all of which we've not been able to identify the cause.

### Steps to reproduce:

There's no specific steps to reproduce this issue as it appears to happen randomly.

### Additional details:

This issue was discussed in p1654635110739929-slack-C01E9LB4X61

### Support requests

5338712-zen, 5263382-zen, 5203580-zen, 5208482-zen Forum 1, 4915339-zen, 4821349-zen

mikkamp:

In 3 of the linked support issues I was able to find image names with special characters, like:
- `El-viaje-Fantástico-del-Osito-Gominola-Mockup.jpg`
- `Copenhagen-Rags-Cashmere-toerklaede-blåt-closeup.jpeg`
- `Minselfie-50-årsdag-bursdag-photobooth.webp`

By default WordPress will remove accents when uploading an image and replace the name with it's non-accented equivalents. However this is not the case if UTF8-NFD characters (combined accented characters) are used, which seems to be most common on MACOS. WordPress has a long standing bug issue for this (currently scheduled for version 6.1): https://core.trac.wordpress.org/ticket/24661
However until that's resolved we need to make sure that these image URL's are not blocked when validating. According to this [bug report](https://github.com/symfony/symfony/issues/29953) it recommends to normalize the URL's before validating. So this PR ensures they are normalized so the above mentioned images pass validation.

We aren't currently testing the validation rules since this is tested in an external library. However I thought it was important to test these types of URL's since it depends on how we use the library whether it validates or not. So this PR also includes a test to validate the image URL.

Closes #1582

1. Create an image with a name like `ïmágê-ñåmè.jpg` (make sure to copy exactly as general UTF8 characters will be converted, [test page](https://minaret.info/test/normalize.msp) for getting the right string).
2. Confirm after uploading that the special characters are still in the image name.
image
3. Create a test product with this image (make sure to add description and price to confirm it syncs).
4. On the Connection Test page, sync the product using it's ID (non async).
5. Confirm it syncs without any validation errors (if there are validation errors they need to be retrieved from the log).
6. After a while check the [Google Dashboard](https://merchants.google.com) to confirm the image URL is valid.

### Additional details:
This PR would only resolve the validation error for sites that are actually using accented characters in the image URL's. Looking at the linked issues there were several sites which appeared to be using images with normal characters. However I wasn't able to retrieve additional details from those tickets to track down specific reasons why it might be failing. If after this PR is released we can still find cases where the images are not synced correctly, the issue can be reopened with additional details containing specific image URLs which are failing to validate.

### Changelog entry
* Fix - Normalize image URLs before validation.

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -753,8 +753,17 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 		$metadata->addPropertyConstraint( 'link', new Assert\Url() );
 
 		$metadata->addPropertyConstraint( 'imageLink', new Assert\NotBlank() );
-		$metadata->addPropertyConstraint( 'imageLink', new Assert\Url() );
-		$metadata->addPropertyConstraint( 'additionalImageLinks', new Assert\All( [ 'constraints' => [ new Assert\Url() ] ] ) );
+		$metadata->addPropertyConstraint( 'imageLink', new Assert\Url( [ 'normalizer' => 'Normalizer::normalize' ] ) );
+		$metadata->addPropertyConstraint(
+			'additionalImageLinks',
+			new Assert\All(
+				[
+					'constraints' => [
+						new Assert\Url( [ 'normalizer' => 'Normalizer::normalize' ] ),
+					],
+				]
+			)
+		);
 
 		$metadata->addGetterConstraint( 'price', new Assert\NotNull() );
 		$metadata->addGetterConstraint( 'price', new GooglePriceConstraint() );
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Test locally and it worked as expected.

- I could reproduce the problem using `develop` branch, the product using the image with UTF8-NFD characters couldn't be synced.
  - The error from `wp-admin/admin.php?page=wc-status&tab=logs` showed `[imageLink] This value is not a valid URL."]`.
  - From the GMC dashboard the product did not exist.
- The problem has been solved using the branch of this PR, the product was synced to GMC and from GMC dashboard I could see it showed the image is waited to be crawled, no invalid image link error appeared in the dashboard.
    - Screenshot 2022-07-05 at 14 14 28

Issue #1331, opened 2022-03-16
Edit Free Listings: Success notification is shown when there is an API error and saving is not successful


ecgan:
### Describe the bug:

- This is mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1313#pullrequestreview-911315044.

> Found another issue that already existed before this project. When there is a failed API call, the notification at the bottom still shows "Your changes to your Free Listings have been saved and will be synced to your Google Merchant Center account."

https://github.com/woocommerce/google-listings-and-ads/blob/154565b1739ba5a9982caaab60447a05503d00ba/js/src/edit-free-campaign/index.js#L188-L208

https://user-images.githubusercontent.com/17420811/158565390-9cf4b921-0adf-4eda-b37e-a6ec65f972a2.mp4

### Steps to reproduce:

1. Modify some API path (e.g. upsert shipping rate API path) so that it will throw error when users are trying to save changes in Edit Free Listings.
2. Go to Edit Free Listings and make some changes.
3. Click on "Save changes" button.

### Expected behavior:

Error message notification should be displayed, e.g. `'Something went wrong while saving your changes. Please try again later.'`.

### Actual behavior:

Success message notification is displayed: `'Your changes to your Free Listings have been saved and will be synced to your Google Merchant Center account.'`.

### Additional details:

I think this is due to the wrong use of [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled). It should be [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) instead. We should investigate a lil bit deeper on why we used `allSettled` in the first place, whether it is really intentional.

jorgemd24:
I was having a look at this issue and I have two questions:

1. I understand that the use of [Promise.allSettled()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) is because we allow to sync the data even when it is partially successful but is it OK to sync the data even if the rates are not saved successful? Should not it be [Promise.all()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) ? I am tagging @eason9487 😄   as I see that you reviewed this part of the code before https://github.com/woocommerce/google-listings-and-ads/pull/468#discussion_r613085411

[...]

PR #1586, opened 2022-07-05
Propagate errors for saveTargetAudience 

jorgemd24:

This is a follow-up of https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1109566842 & https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1110084066

This PR removes the `try` and `catch` from `saveTargetAudience`, allowing the error to bubble up and be handled by the consumer.  In the following example:

https://github.com/woocommerce/google-listings-and-ads/blob/189a55a29a368b0a2ce9150deb3ca70c6077de29/js/src/edit-free-campaign/index.js#L203-L208

if `saveTargetAudience` failed, the consumer was not able to tell if `saveTargetAudience` failed because the error was already caught. Using this PR, `saveTargetAudience` will be shown as `rejected`.

For more context see issue #1331.

I found only two places where `saveTargetAudience` is called:

1.  https://github.com/woocommerce/google-listings-and-ads/blob/189a55a29a368b0a2ce9150deb3ca70c6077de29/js/src/edit-free-campaign/index.js#L204 (This error will be handle by a custom hook, see https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1109566842

2. https://github.com/woocommerce/google-listings-and-ads/blob/a2e05d7361a86efb33dd6dc2d86053ba62230039/js/src/setup-mc/setup-stepper/choose-audience/useAutoSaveTargetAudienceEffect.js#L26-L27  This PR updates the useAutoSaveTargetAudienceEffect to catch the error.

1. Go to the onboarding page step 2.

image

2. Update the countries, no errors should be shown.
3. If you provoke an error, a notice will be displayed. For example, setting an incorrect URL in the following line: https://github.com/woocommerce/google-listings-and-ads/blob/a892de55a797a0feaf20e2d7b88ad25755614417/js/src/data/actions.js#L725 

image

### Additional details:
- The next step will be to remove the try and catch from `saveSettings` so we can handle the errors in situations like the following:

```
const results = await Promise.allSettled( [
	saveTargetAudience( targetAudience ), // remove its try/catch in actions.js. (This PR)
	saveSettings( settings ), // TODO: remove its try/catch in actions.js.
	saveShippingRates( shippingRates ), // already throws error when it fails.
	saveShippingTimes( shippingTimes ), // already throws error when it fails.
] );
```
 
### Changelog entry

>Tweak - Remove try and catch in saveTargetAudience action.

### DIFF
--- a/js/src/setup-mc/setup-stepper/choose-audience/form-content.js
+++ b/js/src/setup-mc/setup-stepper/choose-audience/form-content.js
@@ -3,7 +3,7 @@
  */
 import { Button, RadioControl } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
-import { createInterpolateElement } from '@wordpress/element';
+import { createInterpolateElement, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -32,8 +32,12 @@ const FormContent = ( props ) => {
 	const { formProps } = props;
 	const { values, isValidForm, getInputProps, handleSubmit } = formProps;
 	const { locale, language } = values;
+	const [ isAutoSaved, setAutoSaved ] = useState( true );
 
-	useAutoSaveTargetAudienceEffect( values );
+	const autoSaveCallback = ( resultAutoSave ) =>
+		setAutoSaved( resultAutoSave );
+
+	useAutoSaveTargetAudienceEffect( values, autoSaveCallback );
 	useAutoClearShippingEffect( values.location, values.countries );
 
 	return (
@@ -138,7 +142,7 @@ const FormContent = ( props ) => {
 			<StepContentFooter>
 				<Button
 					isPrimary
-					disabled={ ! isValidForm }
+					disabled={ ! isValidForm || ! isAutoSaved }
 					onClick={ handleSubmit }
 				>
 					{ __( 'Continue', 'google-listings-and-ads' ) }
--- a/js/src/setup-mc/setup-stepper/choose-audience/form-content.js
+++ b/js/src/setup-mc/setup-stepper/choose-audience/form-content.js
@@ -34,10 +34,7 @@ const FormContent = ( props ) => {
 	const { locale, language } = values;
 	const [ isAutoSaved, setAutoSaved ] = useState( true );
 
-	const autoSaveCallback = ( resultAutoSave ) =>
-		setAutoSaved( resultAutoSave );
-
-	useAutoSaveTargetAudienceEffect( values, autoSaveCallback );
+	useAutoSaveTargetAudienceEffect( values, setAutoSaved );
 	useAutoClearShippingEffect( values.location, values.countries );
 
 	return (
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -73,6 +73,14 @@ export function handleFetchError( error, message ) {
  */
 
 /**
+ * @typedef {Object} TargetAudienceData
+ * @property {string} locale The locale for the site. Example: 'en_US'.
+ * @property {string} language The language to use for product listings. Example: 'English'.
+ * @property {string} location Type of location, There are two possible values: 'selected' countries or 'all' countries.
+ * @property {Array<CountryCode>} countries Array of audience countries.
+ */
+
+/**
  *
  * @return {Array<ShippingRate>} Array of individual shipping rates.
  */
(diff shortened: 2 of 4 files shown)


puntope:
I was testing this and I couldn't render the error notice in Edit Free Listings.

I provoke a 404 error like this editing `saveTargetAudience` in `actions.js:`

```
yield apiFetch( {
		path: `${ API_NAMESPACE }/mc/target_audiencesssss`,
		method: 'POST',
		data: targetAudience,
	} );
```

Then I still see the success message

Screenshot 2022-07-07 at 18 17 56

Notice in Onboarding I see it working perfectly.

ecgan:
@puntope , I believe this PR covers only the Setup MC onboarding flow. The Edit Free Listings flow will be covered / handled by a custom hook later.

jorgemd24:
Thanks, @puntope, for reviewing this.

> I was testing this and I couldn't render the error notice in Edit Free Listings.

That's expected, as it is mentioned on this https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1110084066, the idea was to split the changes in different PRs to minimize the risks: first to propagate the errors for the `saveTargetAudience`, the same with `saveSettings` and finally create the custom hook to handle cases such as `Edit Free Listings` where multiple promises are called with `allSettled`. Currently, If `saveSettings` fails it would still show the success message independently of `saveTargetAudience`. 

I am happy to discuss if it is convenient to add the custom hook in this PR, however, we need to keep in mind some requirements that could make this PR a bit more complex and could divert from its original intention which is to propagate the errors of `saveTargetAudience`:

- The error notice needs to be able to show multiple errors and be coherent with i18n. See the second part of this https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1105050752 for more context.
- In the case that we would like to show multiple errors as a bullet list, we will need to use a different type of notice. As far as I know, the `createNotice` only allows string content and single-line.

```
There are errors in the following actions: 
 - There was an error trying to save the target audience
 - There was an error trying to save settings
```

- When evaluating the results from `allSettled` we will need to match it with its user-friendly error.

The hook could look something like the code below, however we will need to find a way to make it more coherent with i18n and more generic:

```
const useCustomAllSettled = () => {
	const { createNotice } = useDispatchCoreNotices();

	return async ( promises, errorsText ) => {

		const results = await Promise.allSettled( promises );

		const rejected = results.reduce( ( acc, item, index ) => {
			if ( item.status === 'rejected' && errorsText[ index ] ) {
				acc.push( errorsText[ index ] );
			}
			return acc;
		}, [] );

		const hasErrors = !! rejected.length;

		if ( hasErrors ) {
			createNotice(
				'error',
				`There were errors trying to save: ${ rejected }. Other changes have been saved. Please try again later.`
			);
		}

		return hasErrors;
	};
};
```


 Let me know what you think!

puntope:
> Let me know what you think!

We can handle it in separate PR no problem. I wan't sure if we were covering those as well.

puntope:
> * In the case that we would like to show multiple errors as a bullet list, we will need to use a different type of notice. As far as I know, the `createNotice` only allows string content and single-line.

I like the idea of concat errors like ... actionA, actionB, actionC as you defined them in the hook.

```There were errors trying to save: target audience, taxes and budget. Other changes have been saved. Please try again later.``` 

I personally hate when some APP show me "try again later..." and also I guess we should add an "and" in the last item somehow. But I guess the copy is more a question for design team. The idea looks good for me. 

Ok also to have it in other PR

**Comment by puntope** [comment]
One thing I saw is with a fresh onboarding, in case we force the error I see the message but I can still continue to next step, however, I just saw the spinner running forever. 

Then I removed the error, completed step 2 and the problem is not happening anymore even if I force again the error in target audience. 

https://user-images.githubusercontent.com/5908855/177945394-3dd0c94e-ad91-4eac-a388-5f9aaf26255f.mov

Wondering if we should block the continue button in case of error. But this could also be done/discussed in further PR

**Comment by puntope** [comment]
💅  Wondering if we could extend our tests adding some checks...  for example checking that no exception is thrown by the action.

**Comment by jorgemd24** [comment]
@puntope,

>Wondering if we should block the continue button in case of error. But this could also be done/discussed in further PR

I have added it in this PR, the continue button is disabled if the autosaving fails. See video:

https://user-images.githubusercontent.com/2488994/178262677-2d8f0483-871f-4439-bca2-9baef7544713.mp4

>Wondering if we could extend our tests adding some checks... for example checking that no exception is thrown by the action.

The action throws an exception and it is caught by the consumer, (the autosaving hook in this case ). I added some tests for the autosaving hook. 

@puntope, I guess this is ready for another round! Thanks for your comments!

File: js/src/setup-mc/setup-stepper/choose-audience/form-content.js [hunk 1]
```diff
@@ -32,8 +32,12 @@ const FormContent = ( props ) => {
 	const { formProps } = props;
 	const { values, isValidForm, getInputProps, handleSubmit } = formProps;
 	const { locale, language } = values;
+	const [ isAutoSaved, setAutoSaved ] = useState( true );
 
-	useAutoSaveTargetAudienceEffect( values );
+	const autoSaveCallback = ( resultAutoSave ) =>
+		setAutoSaved( resultAutoSave );
+
+	useAutoSaveTargetAudienceEffect( values, autoSaveCallback );
```
puntope: 💅 

`useAutoSaveTargetAudienceEffect( values, setAutoSaved );` 

Also works ☺️

puntope (approved):
✅  LGTM

Thanks for the changes, tweaks and tests 👨‍💻

jorgemd24:
Thanks, @puntope for reviewing this and your suggestions 💯

PR #1587, opened 2022-07-07
Fix encoding problems in product issues table

### LINKED ISSUE #1581: Product name has wrongly HTML-encoded en dash `"&#8211;"` in Issues to resolve table
### Describe the bug:

Reproducible in Google Listings and Ads version: [1.13.6 (release)](https://github.com/woocommerce/google-listings-and-ads/releases/tag/1.13.6)

When I look into the "Issues to resolve" table, some of the product names have `&#8211;` in it, e.g. "Hoodie `&#8211;` blue, No".

image

The product name comes from the backend API side `GET /mc/issues/product`:

image

I noticed this when I was doing WC 6.7 compatibility testing in https://github.com/woocommerce/google-listings-and-ads/issues/1577. I think this issue may have been happening for some time now and we didn't notice this.

### Steps to reproduce:

1. Go to Product Feed page > Issue to resolve table > Product issues tab.
2. Look for Hoodie and V-next t-shirt product (i.e. variable product).

### Expected behavior:

The product name should not have HTML-encoded `"&#8211;"` text in it.

### Actual behavior:

The product name has HTML-encoded `"&#8211;"` text in it.

puntope:

Closes #1581 .

This PR as escape function to render correctly the HMLT entities received from Google API.

### Screenshots:

Before
image

After
Screenshot 2022-07-07 at 13 00 14

1. Checkout this PR
2. Load sample products, or have some variant in your products to force the dash (-) symbol.
3. Go to product Feed, maybe you need to force refresh, You can do that in `src/MerchantCenter/MerchantStatuses.php` changing the line 141 to `$this->maybe_refresh_status_data( true );`
4. Check that’ll  no html entity appears, instead you see the character.

### Additional details:

In brief the problem is that we receive encoded HTML entities from Google API, so they are saved like that in DB. I added a native PHP function to prevent that.

### Changelog entry

> Fix - Encoding product names in Issues Table

### DIFF
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -463,7 +463,7 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 			}
 
 			$product_issue_template = [
-				'product'              => $this->product_data_lookup[ $wc_product_id ]['name'],
+				'product'              => html_entity_decode( $this->product_data_lookup[ $wc_product_id ]['name'] ),
 				'product_id'           => $wc_product_id,
 				'created_at'           => $created_at,
 				'applicable_countries' => [],


jorgemd24 (approved):
Thanks, @puntope, The HTML entities are decoded correctly. LGTM 👍

PR #579, opened 2021-05-11
Replacement polyfills for mbstring

mikkamp:
### Changes proposed in this Pull Request:

To validate products before they sync to the Merchant Center API, we use the `symfony/validator` library. This in turn installs `polyfill-mbstring` as a fallback for sites that do not have the `mbstring` extension enabled.

The polyfill works fine but it produces an error when submitting to WordPress.org. To get around this, we provide our own polyfill so the symfony one is no longer needed.

Closes #331

### Detailed test instructions:

1. Configure PHP to skip loading of the mbstring extension `extension=mbstring.so`
2. Remove the vendor folder and run `composer install --no-dev` to make sure the `polyfill-mbstring` package is no longer loaded (dev dependencies require the mbstring PHP extension)
3. Tools > Health Check will confirm if the mbstring extension is missing (make sure the GLA plugin is not active as it provides an alternative)
image
4. Edit one of the products and set one of the MPN numbers (this will trigger a string length validation when syncing)
5. Sync all the products on the connection test page
6. Confirm there are no fatal errors or warnings

### Changelog Note:
- Fix - Replacement polyfills for mbstring.

Issue #331, opened 2021-03-11
Release Process - release to WP.org


mikkamp:
As a followup to #327 

Setup the plugin for release to WP.org in a way that's compatible with WooRelease 
There are quite a few examples of how we do this in our other plugins (Google Analytics Integration), which we can copy.

PR #1588, opened 2022-07-09
Add Normalizer Polyfill

jorgemd24:

When I was trying to sync products I was getting the following error:

```
The "normalizer" option must be a valid callable ("string" given). in /var/www/html/wp-content/plugins/google-listings-and-ads/vendor/symfony/validator/Constraints/Url.php:54 Stack trace: #0 /var/www/html/wp-content/plugins/google-listings-and-ads/src/Product/WCProductAdapter.php(756):
```

The error was caused because the [Int](https://php.net/intl) extension was not installed hence the `Normalizer` class was not found.  

This PR adds the Normalizer polyfill, based off https://github.com/symfony/polyfill-intl-normalizer, unfortunately we can't add the package as a dependency using `composer require symfony/polyfill-intl-normalizer` as it causes an error when using with woorelease. See more context: https://github.com/woocommerce/google-listings-and-ads/issues/331#issuecomment-838372148 and https://github.com/woocommerce/google-listings-and-ads/pull/579.

Because of this issue, the polyfill needs to be added manually to our repository. I added only the methods needed to run the plugin.

### Screenshots:

https://user-images.githubusercontent.com/2488994/178124584-29ceadfa-1f42-4d8d-ac86-c069f190fd77.mp4

1. Disable the extension `intl` extension by adding a semicolon at the begging, for example: `;extension=intl.so`. In my case I disabled it going to the following file: `/usr/local/etc/php/conf.d/docker-php-ext-intl.ini`
2. Go to `/wp-admin/admin.php?page=connection-test-admin-page`
3. Sync a product, without clicking `async`
4. The following error should be displayed: `The "normalizer" option must be a valid callable.....`
5. Check out this PR
6. run `composer dump-autoload` to create the autoload files.
7. Repeat steps from 2 to 4
8. No errors should be displayed

### Additional details:

- I applied phpcs in the new files but I observed that adding `Resources/unidata` files was increasing considerably the time to run phpcs. In my local setup went from 15s to more than 1min therefore I set `// phpcs:ignoreFile` in those files.
- I applied phpcs `Normalizer.php` in the places where it was safe to do it, I did not want it to replace all variables to be snake_case as I think it could increase the risk to create an unnecessary bug. 
- I added a new polyfill `mb_internal_encoding` to the class MBString as they are used in the `Normalizer` class.

### Changelog entry

>Add - Normalizer Polyfill


composer.json
```diff
@@ -27,7 +27,8 @@
 		"yoast/phpunit-polyfills": "^1.0"
 	},
 	"replace": {
-		"symfony/polyfill-mbstring": "*"
+		"symfony/polyfill-mbstring": "*",
+		"symfony/polyfill-intl-normalizer": "*"
```
mikkamp, line 31: I don't believe this is necessary. The line was added for `polyfill-mbstring` since it's a [direct requirement](https://github.com/symfony/validator/blob/5.2/composer.json#L22) in the `symfony/validator` package. Either way it shouldn't cause any harm and could be an extra protection so it isn't loaded twice.

composer.json
```diff
@@ -38,12 +39,13 @@
...
 			"vendor/guzzlehttp/promises/src/functions_include.php"
-		]
+		],
+		"classmap": [ "src/Polyfills" ]
 	},
 	"autoload-dev": {
 		"psr-4": {
 			"Automattic\\WooCommerce\\GoogleListingsAndAds\\Tests\\": "tests/"
-		}
+		}
```
mikkamp: Minor: shouldn't need to add this extra spacing

src/Polyfills/bootstrap.php
```diff
@@ -58,3 +58,18 @@ function mb_substr( $string, $start, $length = null, $encoding = null ) {
...
+	 * Get internal encoding.
+	 *
+	 * @param string $encoding
+	 *
+	 * @return string|bool
+	 *
+	 * @since x.x.x
+	 */
+	function mb_internal_encoding( $encoding = null ) {
+		return Mbstring::mb_internal_encoding( $encoding );
```
mikkamp: PHP isn't case sensitive but it would be nice to use the same case for the class name `MBString`.

mikkamp (approved):
Thanks for the prompt PR.

I disabled the `intl` extension and was able to reproduce the fatal error, when I switch to this PR, the error is fixed and I'm able to sync the product using a normalized name.

I added a few smaller comments, but the polyfill is working as intended so I'll go ahead and mark it as approved.

jorgemd24:
Thanks, @mikkamp for the review,

I addressed the minor comments. Thanks!

richbscott:
Hi this is resulting in an undefined function `Uncaught Error: Call to undefined method Normalizer::isNormalized()` when used with the Wordpress Cloudflare Plugin https://github.com/cloudflare/Cloudflare-WordPress which uses the actual symfony library as it seems to now be picking up this class rather than the one in their vendor directory.

Can this be switched to use the dependency so that all the functions are available?  I note that there was an issue with woorelease as per https://github.com/woocommerce/google-listings-and-ads/issues/331#issuecomment-838372148

Can this be solved in the same way as the Cloudflare wordpress plugin with a script in the composer file here https://github.com/cloudflare/Cloudflare-WordPress/blob/master/composer.json to remove the boostrap8.0.php files?

jorgemd24:
Thanks @richbscott for reporting this.

>Can this be solved in the same way as the Cloudflare wordpress plugin with a script in the composer file here https://github.com/cloudflare/Cloudflare-WordPress/blob/master/composer.json to remove the boostrap8.0.php files?

Yeah, that could be one possible solution for the issue with woorelease.

The other possible solutions are:

- The GLA plugin does not load the class globally and we use our own Normalize class.
- The other option is to add the method `isNormalized` and the other functions to the GLA Normalize class so the method is available. 

Let me look into this a bit more so we can find a good solution for this!

mikkamp:
I do like the suggestion of using the actual symfony packages so we don't have to maintain duplicate code. Although we do need to run our extension on sites with PHP 8. Just removing the PHP 8 bootstrap file will cause an error here: https://github.com/symfony/polyfill-intl-normalizer/blob/v1.26.0/bootstrap.php#L14-L16

We do have other scripts which modify libraries as needed, so I don't think it's a bad approach to write a script which removes both the bootstrap80.php file and the `loading condition`.

PR #1589, opened 2022-07-10
Add GH workflow to set PR labels

tomalec:

Add GH workflow to set PR labels according to branch name convention, to denote changelog type.

Currently, those labels do not make much. They show up in [pulls list](https://github.com/woocommerce/google-listings-and-ads/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc), and let GitHub release notes categorize changes. But then woorelease is ([will be](https://github.com/woocommerce/woorelease/pull/374)) using them regardless of the category ignoring the labels.

But, I plan to change WR to use those labels, to set changelog type prefix when using PR title as the entry.

So eventually, the developer, could choose to go the conventional path: use branch name prefix, write a good PR title, to get a changelog entry, without the need to manually state it.

### Screenshots:

image

1. See the workflow log for this PR https://github.com/woocommerce/google-listings-and-ads/runs/7272844970?check_suite_focus=true
2. Check the added labels

### Additional details:

- I also added label's descriptions to this repo: 
   image

### Changelog entry

> Add - GH workflow to set PR labels


eason9487 (approved):
Viewed the GHA job and the added labels. LGTM.

PR #1595, opened 2022-07-14
Disable identifier_exists field

### LINKED ISSUE #1592: Remove the identiier_exist logic entirely
### Background

Every year Google comes out with [new changes to the product feed spec](https://support.google.com/merchants/answer/11905642). 

The one that is most applicable to Woo merchants using the plugin is updated guidance for unique product identifiers. 

Generally speaking, Woo merchants should fill out the identifiers if they can to get the most out of shopping ads + free listings. Previously, if there was a missing GTIN and our systems identified this, it would lead to a disapproval of the offer. 

Now it leads to a demotion only, which means the product can serve but may not get as many clicks/impressions.

### Describe the bug:

Despite this new product update, the issue is... there will still be scenarios where there this [disapproval](https://support.google.com/merchants/answer/9464748?hl=en) appears because of the logic of how identifier_exist = false is implemented. 

Essentially, identifier_exist = false occurs when a merchant doesn't update any identifiers within the product editor. 

This is then sent to MC.

The Google team have advised that we should **remove the identiier_exist logic entirely in GLA**. 

This means the `identifier_exist` property won’t be submitted to Google. 

They have confirmed that there shouldn't be any issues with removing identifier_exists. 

Products that lack unique identifiers should still serve just in a demoted state as opposed to being disapproved for identifier exists misuse item error.

So to clarify - `identifier_exist` should not be set to either true or false. The property should be removed completely.

This is aimed at addresses issues like "Incorrect value: identifier exists [identifier_exists]"  [reference](https://support.google.com/merchants/answer/9464748?hl=en)

### Expected behavior:
The `identifier_exist` property is never set or sent to Google

### Actual behavior:
Currently, we set `identifier_exist` to false by default. If an identifier is set e.g. "GTIN" the we set `identifier_exist` to true.

puntope:

Closes #1592

This PR disables `identifier_exists` as requested from Google.

TL;DR

This field was true if GTIN or MPN fields were filled up. False otherwise. Now we never send this field anymore.
 

### Screenshots:

Screenshot 2022-07-14 at 11 46 11

1. Connect your Google Account and go to one of your Single Products
2. In the Tab google Listing And Ads, fill [GTIN number](https://en.wikipedia.org/wiki/Global_Trade_Item_Number) should be in the right format. Otherwise the product wont sync.   
3. Go to Connection Test page and Sync that product
4. You will see successfully synced with google. 
5. Go to Google MC Dashboard -> All Products 
6. Click on the Product you just synced 
7. At the end of the page you will see a tab "Raw feed attributes: Content API". Click on it
8. Check that  identifier_exists field is not being sent anymore.

### Additional details:

- I found only one lace when we send this field. SO I just removed the statement.
- I updated the tests to assert null 

### Changelog entry

> Fix - Disable identifier_exists field

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -123,7 +123,6 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			 ->map_wc_product_shipping()
 			 ->map_wc_prices();
 
-		$this->setIdentifierExists( ! empty( $this->getGtin() ) || ! empty( $this->getMpn() ) );
 	}
 
 	/**
--- a/tests/Unit/Product/WCProductAdapterTest.php
+++ b/tests/Unit/Product/WCProductAdapterTest.php
@@ -861,7 +861,7 @@ DESCRIPTION;
 		$this->assertNull( $adapted_product->getShippingWidth() );
 	}
 
-	public function test_identifier_exists_is_false_if_gtin_and_mpn_not_provided() {
+	public function test_identifier_exists_is_null_if_gtin_and_mpn_not_provided() {
 		$product         = WC_Helper_Product::create_simple_product( false );
 		$adapted_product = new WCProductAdapter(
 			[
@@ -869,10 +869,10 @@ DESCRIPTION;
 				'targetCountry' => 'US',
 			]
 		);
-		$this->assertFalse( $adapted_product->getIdentifierExists() );
+		$this->assertNull( $adapted_product->getIdentifierExists() );
 	}
 
-	public function test_identifier_exists_is_true_if_gtin_provided() {
+	public function test_identifier_exists_is_null_if_gtin_provided() {
 		$product         = WC_Helper_Product::create_simple_product( false );
 		$adapted_product = new WCProductAdapter(
 			[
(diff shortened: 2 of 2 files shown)


mikkamp (approved):
Thanks, the changes look good. I tested the code and can confirm that it's showing up in the raw attributes view without the identifier exists field.

The tests also pass, but I was just wondering how useful it is to test for attributes that haven't been set. The WCProductAdapter extends GoogleProduct and there are a lot of other attributes we don't set without testing. I would have personally just removed the tests since they are no longer relevant.

puntope:
> Thanks, the changes look good. I tested the code and can confirm that it's showing up in the raw attributes view without the identifier exists field.
> 
> The tests also pass, but I was just wondering how useful it is to test for attributes that haven't been set. The WCProductAdapter extends GoogleProduct and there are a lot of other attributes we don't set without testing. I would have personally just removed the tests since they are no longer relevant.

Thanks @mikkamp I was thinking the same. My thought was. If this property being set is going to create problems, better to be sure is not being set.

findingsimple:
Thanks @puntope for knocking this one over

PR #1602, opened 2022-07-20
Hook for handling multiple errors for rejected promises

jorgemd24:

The intention of this PR is to provide a react hook to create an error notice displaying multiple errors such as rejected promises invoked by [allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled).

As was discussed in the following comment: https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1105050752, there are two possible ways to show the error messages:

1. Display the error messages in one line.
2. Display the error messages in a list format.

My intention was to create a list of errors but `createNotice` only allows string content and single line. There is an option to insert HTML: `__unstableHTML` however it has the following comment:

```
Notice message as raw HTML. Intended to
serve primarily for compatibility of
server-rendered notices, and SHOULD NOT
be used for notices. It is subject to
removal without notice
```

See code here: https://github.com/WordPress/WordPress/blob/master/wp-includes/js/dist/notices.js#L357-L361

So I decided not to use that option and I tried to implement the first option, showing the error messages in one line. 

This PR is partially closing https://github.com/woocommerce/google-listings-and-ads/issues/1331 as I realized that `saveShippingTimes` does not propagate the error because `upsertShippingTimes` catches the error. I will work on this in a different PR. 

https://github.com/woocommerce/google-listings-and-ads/blob/c978d3596c35f7d2f5bec0985ead8912b5d96d8b/js/src/data/actions.js#L232-L259

### Screenshots:

Before this PR:

Target audience, MC settings and Shipping Rates failed but still show that the changes were sync successful.

image
[https://d.pr/i/PLasQ6](https://d.pr/i/PLasQ6)

After this PR:

Target audience, MC settings and Shipping Rates failed but it will inform the user which actions failed. 

image
[https://d.pr/i/jVwePF](https://d.pr/i/jVwePF)

1. Go to edit Edit free listings.
2. Try to cause an error in one or all the following endpoints: 

```
POST - wp-json/wc/gla/mc/target_audience
POST - wp-json/wc/gla/mc/settings
POST - wp-json/wc/gla/mc/shipping/rates (This will only be called if the shipping rates have been updated)
```
I tested blocking the request:

image
 
3. Click on save changes.

image

4. If you are in the `develop` branch it will show you that the changes have been sync successful.
5. If you checked out this PR it will show you that there were some errors.

### Additional details:
There is some complexity in adding the singular and plural for error messages, for example:

`There **are** errors in the following action**s**`
`There **is an error** for the following **action**`

I think this could be improve in a different PR and I am happy to hear about tools or functions that could help us to improve this part.

### Changelog entry

>Fix - Handle multiple errors in the Edit free listings page.


js/src/edit-free-campaign/index.js
```diff
@@ -205,23 +209,34 @@ const EditFreeCampaign = () => {
...
 				saveSettings( settings ),
 				saveShippingRates( shippingRates ),
 				saveShippingTimes( shippingTimes ),
+			];
+
+			const rejected = await createNoticeForRejectedPromises( promises, [
+				'Target audience',
+				'MC Settings',
+				'Shipping rates',
+				'Shipping times',
```
eason9487: ❓ I think these texts should be still wrapped by `__` function.

js/src/edit-free-campaign/index.js
```diff
@@ -205,23 +209,34 @@ const EditFreeCampaign = () => {
...
+			const promises = [
 				saveTargetAudience( targetAudience ),
 				saveSettings( settings ),
 				saveShippingRates( shippingRates ),
 				saveShippingTimes( shippingTimes ),
+			];
+
+			const rejected = await createNoticeForRejectedPromises( promises, [
+				'Target audience',
+				'MC Settings',
```
eason9487: ❔ Not sure if merchants will recognize MC as an abbreviation for Merchant Center. I would suggest using a clearer name to reduce the possibility of confusion.

js/src/hooks/useCreateNoticeForMultipleErrors.js
```diff
@@ -0,0 +1,59 @@
...
+			const listErrors = concatenateListOfWords( errorMessages );
+
+			const content = sprintf(
+				// translators: 1: list of errors 2: optional text if some promises have been invoked successfully.
+				__(
+					'There are errors in the following actions: %1$s. %2$s Please try again later.',
+					'google-listings-and-ads'
+				),
+				listErrors,
+				isPartiallySuccessful ? 'Other changes have been saved.' : ''
```
eason9487: ❔ I think `'Other changes have been saved.'` should be wrapped by `__` function as well.

js/src/utils/concatenateListOfWords.js
```diff
@@ -0,0 +1,27 @@
...
+ * @return {string} The concatenated string.
+ */
+
+export default function concatenateListOfWords(
+	listWords,
+	separator = ', ',
+	separatorLastElement = 'and'
+) {
+	if ( listWords.length === 1 ) {
+		return listWords.slice( -1 ).join( '' );
```
eason9487: 💅 Looks like this line could be simplified to `return listWords[ 0 ]`;

js/src/utils/concatenateListOfWords.js
```diff
@@ -0,0 +1,27 @@
...
+	}
+	return `${ listWords
+		.slice( 0, -1 )
+		.join( separator ) } ${ separatorLastElement } ${ listWords.slice(
+		-1
+	) }`;
```
eason9487: 💅 or ❔ 
> There is some complexity in adding the singular and plural for error messages, for example:
> 
> `There **are** errors in the following action**s**` `There **is an error** for the following **action**`
> 
> I think this could be improve in a different PR and I am happy to hear about tools or functions that could help us to improve this part.

There is the `_n` in `@wordpress/i18n` package but it doesn't perfectly fit with this implementation. If it's ok to internalize this handling to `useCreateNoticeForMultipleErrors`, it could be replaced with

```js
const listErrors = sprintf(
	_n(
		'There is an error in the following action: %1$s%2$s',
		'There are errors in the following actions: %1$s and %2$s',
		listWords.length,
		'google-listings-and-ads'
	),
	listWords.slice( 0, -1 ).join( separator ),
	listWords.at( -1 )
);
```

js/src/hooks/useCreateNoticeForRejectedPromises.js
```diff
@@ -0,0 +1,50 @@
...
+	) => {
+		const results = await Promise.allSettled( promises );
+		const rejectedMessages = results.reduce( ( acc, item, index ) => {
+			if ( item.status === 'rejected' && promiseNames[ index ] ) {
+				acc.push( promiseNames[ index ] );
+			}
+			return acc;
+		}, [] );
+
+		const isPartiallySuccessful = rejectedMessages.length < promises.length;
```
eason9487: 🚧 
When all promises are rejected, it will make the `useCreateNoticeForMultipleErrors` show an error notice with "Other changes have been saved." in its text.

js/src/hooks/useCreateNoticeForMultipleErrors.js
```diff
@@ -0,0 +1,59 @@
...
+			);
+
+			createNotice( 'error', content );
```
eason9487: ❔ 
I'm curious Is there any reason to call `createNotice` in this function? As I understand, `useCreateNoticeForMultipleErrors` and `useCreateNoticeForRejectedPromises` could be non-hook functions because they don't have any states and are not triggered by any React life cycle. Look like they are hook functions just because this hook calls `createNotice` here.

Moving the call to *js/src/edit-free-campaign/index.js* could make these two functions simpler and be used easier as well, and even merge them to single util if there is no other use case that needs to use them separately. Another point is that the implementation of error handling would have the same abstraction level for the `createNotice` calls if put them in the same place.

js/src/hooks/useCreateNoticeForRejectedPromises.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @return {CreateNoticeForRejectedPromises} - A function that will create the error notice for the rejected promises.
+ */
+export default function useCreateNoticeForRejectedPromises() {
+	const {
+		createNoticeForMultipleErrors,
+	} = useCreateNoticeForMultipleErrors();
+
+	/**
+	 * @type {CreateNoticeForRejectedPromises} CreateNoticeForRejectedPromises	 *
```
eason9487: ❔ Looks like the description was missing or the `*` at the tail are leftover chars.

js/src/utils/concatenateListOfWords.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * A function that joins a list of words using the sepators provided as the second and third argument.
+ *
+ * Example:
+ *
+ * concatenateListOfWords(['A', 'B', 'C']) will return 'A, B and C'
+ *
+ * @param  {Array<string>} listWords A list of words to concatenate.
+ * @param  {string} [separator] The separator between words.
+ * @param  {string} [separatorLastElement] The separator for the last word.
```
eason9487: 💅 It would be good to also indicate the default values for the optional arguments.

js/src/hooks/useCreateNoticeForRejectedPromises.js
jorgemd24: Is it not because of this?

>This PR is partially closing https://github.com/woocommerce/google-listings-and-ads/issues/1331 as I realized that saveShippingTimes does not propagate the error because upsertShippingTimes catches the error. I will work on this in a different PR.

I can't reproduce it 🤔 , as far as I can see `isPartiallySuccessful` will be `false` if `rejectedMessages.length` has the same length that the promises array.

js/src/hooks/useCreateNoticeForMultipleErrors.js
jorgemd24: My main reason was to facilitate the consumer to create the notice including the error messages without the need to add the extra step.  Just a hook with two arguments: the promises & the error messages and the notice would display. 

I gave it a try to your suggestion, as I think you have valid points too. Here is the commit: [f83704c](https://github.com/woocommerce/google-listings-and-ads/pull/1602/commits/f83704c62dfcf384f91764a946cc07999438f8f2)

jorgemd24:
Thanks, @eason9487 for reviewing this! I added your suggestions. I guess this is ready for another round!

js/src/edit-free-campaign/index.js
```diff
@@ -205,23 +206,38 @@ const EditFreeCampaign = () => {
...
-			] );
+			];
+
+			const errorMessage = await createErrorMessageForRejectedPromises(
+				promises,
+				[
+					__( 'Target audience' ),
+					__( 'Merchant Center Settings' ),
+					__( 'Shipping rates' ),
+					__( 'Shipping times' ),
```
eason9487: 💅 
These `__` calls need the second parameter `'google-listings-and-ads'` to indicate the i18n text domain.

js/src/hooks/useCreateNoticeForRejectedPromises.js
eason9487: Oh, sorry. 🙏 🤦‍♂️ 

I thought "Would it be a problem if pass `isPartiallySuccessful` as `true` to call `createNoticeForMultipleErrors` when all promises are resolved?" when viewing this line of code. But I got the logic upside down when writing the comment after switching back and forth several files.

💅 
And I learned that `useCreateNoticeForMultipleErrors` would check the length of `errorMessages`, so it won't be a problem. But I'm still unsure about the original comment I was supposed to ask - Should `isPartiallySuccessful` be `true` to match its semantic when all promises are resolved? Or maybe it would be clearer with:
```js
// `useCreateNoticeForRejectedPromises` - Move the length check to here and rename isPartiallySuccessful to isAllFailed or something more semantically.
if ( rejectedMessages.length ) {
	const isAllFailed = rejectedMessages.length === promises.length;
	createNoticeForMultipleErrors( rejectedMessages, isAllFailed );
}

// `useCreateNoticeForMultipleErrors` - Corresponding changes to above.
const createNoticeForMultipleErrors = (
	errorMessages,
	isAllFailed = true
) => {
	const listErrors = concatenateListOfWords( errorMessages );

	const content = sprintf(
		__(
			'%1$s. %2$s Please try again later.',
			'google-listings-and-ads'
		),
		listErrors,
		isAllFailed ? '' : 'Other changes have been saved.'
	);

	createNotice( 'error', content );
};
```

eason9487 (approved):
Testing well and LGTM. Left a comment about the missing parameter of i18n text domain but it's not a blocker.

js/src/utils/createMessageForMultipleErrors.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {Array<string>} errorMessages Error messages to be joined.
+ * @param {boolean} [isPartiallySuccessful='true'] Whether some actions were successful.
+ * @return {string|null} - Return the error message for a list of errors otherwise null if there are no error messages
+ */
+export default function createMessageForMultipleErrors(
+	errorMessages,
+	isPartiallySuccessful = true
+) {
+	if ( errorMessages.length ) {
+		const listErrors = concatenateListOfWords( errorMessages );
```
tomalec: :construction: We are not providing i18nized separators here, so when using Polish translation I will see:
> Podane akcje niepowiodły się: Koszt Dostawy **and** Cena Dostawy.  Spróbuj ponownie.

instead of 
> Podane akcje niepowiodły się: Koszt Dostawy **i** Cena Dostawy. Spróbuj ponownie.

eason9487: Thanks for pointing out this.

The suggestion in https://github.com/woocommerce/google-listings-and-ads/pull/1602#discussion_r928705066 that includes `'and'` text in the i18n function could be considered.

js/src/utils/concatenateListOfWords.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @param  {Array<string>} listWords A list of words to concatenate.
+ * @param  {string} [separator=', '] The separator between words.
+ * @param  {string} [separatorLastElement='and'] The separator for the last word.
+ * @return {string} The concatenated string.
+ */
+
+export default function concatenateListOfWords(
+	listWords,
+	separator = ', ',
```
ecgan: 💅 Note that separator like this should also be handled with i18n. Not all languages use a comma and space `, ` as separator. For example, in Chinese, a comma is `，`, and there is even an enumeration comma `、`.

More info:

- [Chinese punctuation](https://en.wikipedia.org/wiki/Chinese_punctuation)
- [Comma in languages other than Western European](https://en.wikipedia.org/wiki/Comma#Languages_other_than_Western_European) - Contains mention of enumeration or ideographic comma in Chinese, Japanese and Korean.

(Yup, i18n can be quite complicated to get it right. 😂 )

js/src/utils/createMessageForMultipleErrors.js
eason9487: Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1602/commits/340a524876435634fea495ca977d082f74f67ea3.

js/src/utils/concatenateListOfWords.js
eason9487: Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1602/commits/340a524876435634fea495ca977d082f74f67ea3.

js/src/edit-free-campaign/index.js
eason9487: Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1602/commits/0aa19ae398b6e591257ff9b77ccf7b76ffce6ad2.

eason9487:
Hi @tomalec and @ecgan, I continued this PR to address the rest issues of i18n texts. Could you help with the new round of code reviews?

js/src/utils/createMessageForMultipleErrors.js
```diff
@@ -0,0 +1,55 @@
...
+	if ( errorMessages.length ) {
+		const separator = _x(
+			', ',
+			'the separator for concatenating the messages of failed actions',
+			'google-listings-and-ads'
+		);
+		const listErrors = sprintf(
+			// translators: 1: optional string when there are multiple failed actions, and it's a concatenated text of failed actions except for the last one. 2: the last one or the only failed action.
+			_n(
+				'There is an error in the following action: %1$s%2$s.',
```
tomalec, line 27: :nail_care: 
`%1S` is always `''` isn't it? Personally, I'd avoid one interpolation, 
```suggestion
				'There is an error in the following action: %2$s.',
```

js/src/utils/createErrorMessageForRejectedPromises.test.js
```diff
@@ -0,0 +1,88 @@
...
+describe( 'createErrorMessageForRejectedPromises', () => {
+	const successPromise = () => new Promise( ( resolve ) => resolve( 'OK' ) );
+	const rejectedPromise = () =>
+		new Promise( ( r, reject ) => reject( 'Failed' ) );
+
+	afterEach( () => {
+		jest.clearAllMocks();
+	} );
+
+	it( 'No rejected Promises', async () => {
```
tomalec: :nail_care: I'd be a bit more descriptive
```suggestion
	it( 'Called with no rejected promises, calls `createMessageForMultipleErrors` with an empty array and `true`', async () => {
```

js/src/utils/createErrorMessageForRejectedPromises.test.js
```diff
@@ -0,0 +1,88 @@
...
+		const promisesName = [ 'Promise A', 'Promise B', 'Promise C' ];
+
+		await createErrorMessageForRejectedPromises( promises, promisesName );
+
+		expect( createMessageForMultipleErrors ).toBeCalledTimes( 1 );
+		expect( createMessageForMultipleErrors ).toBeCalledWith(
+			[ 'Promise A', 'Promise B', 'Promise C' ],
+			false
+		);
+	} );
```
tomalec: :nail_care: :scroll: 
I think it would be nice to test also the return values of `createErrorMessageForRejectedPromises`, as that is what we use in real scenarios, but do not test here.

js/src/utils/createErrorMessageForRejectedPromises.test.js
```diff
@@ -0,0 +1,88 @@
...
+		];
+		const promisesName = [ 'Promise A', 'Promise B', 'Promise C' ];
+
+		await createErrorMessageForRejectedPromises( promises, promisesName );
+
+		expect( createMessageForMultipleErrors ).toBeCalledTimes( 1 );
+		expect( createMessageForMultipleErrors ).toBeCalledWith( [], true );
+	} );
+
+	it( 'One rejected Promise', async () => {
```
tomalec:
```suggestion
	it( 'Called with one rejected promise, calls `createMessageForMultipleErrors` with its name, and `true`', async () => {
```

js/src/utils/createErrorMessageForRejectedPromises.test.js
```diff
@@ -0,0 +1,88 @@
...
+		await createErrorMessageForRejectedPromises( promises, promisesName );
+
+		expect( createMessageForMultipleErrors ).toBeCalledTimes( 1 );
+		expect( createMessageForMultipleErrors ).toBeCalledWith(
+			[ 'Promise B' ],
+			true
+		);
+	} );
+
+	it( 'Multiple rejected promises', async () => {
```
tomalec:
```suggestion
	it( 'Called with multiple rejected promises, calls `createMessageForMultipleErrors` with their names, and `true`', async () => {
```

js/src/utils/createErrorMessageForRejectedPromises.test.js
```diff
@@ -0,0 +1,88 @@
...
+		await createErrorMessageForRejectedPromises( promises, promisesName );
+
+		expect( createMessageForMultipleErrors ).toBeCalledTimes( 1 );
+		expect( createMessageForMultipleErrors ).toBeCalledWith(
+			[ 'Promise B', 'Promise C', 'Promise D' ],
+			true
+		);
+	} );
+
+	it( 'All rejected promises', async () => {
```
tomalec:
```suggestion
	it( 'Called with all rejected promises, calls `createMessageForMultipleErrors` with their names, and `false`', async () => {
```

tomalec (approved):
Reviewed the code, tested locally, by blocking 0-all responses, LGTM :) thanks.

Left few cosmetic ideas. I think the most important one is to test the return value of `createErrorMessageForRejectedPromises` as that is what we use in the real scenario, not the fact it calls some other function.

js/src/utils/createMessageForMultipleErrors.js
```diff
@@ -0,0 +1,55 @@
...
+				'google-listings-and-ads'
+			),
+			listErrors,
+			isPartiallySuccessful
+				? __(
+						'Other changes have been saved.',
+						'google-listings-and-ads'
+				  )
+				: ''
+		);
```
ecgan: ❔ If `isPartiallySuccessful` is `false`, then there will be two spaces before the word "Please".

Maybe separating the string into two different translation strings like below would help? 🤔 

```js
__(
	'%1$s Please try again later.',
	'google-listings-and-ads'
)

__(
	'%1$s Other changes have been saved. Please try again later.',
	'google-listings-and-ads'
)
```

ecgan (commented):
Just adding one minor comment below, in response to @eason's https://github.com/woocommerce/google-listings-and-ads/pull/1602#discussion_r941135744. 😄

js/src/utils/concatenateListOfWords.js
ecgan: Thanks @eason9487 . I added a comment about your change in https://github.com/woocommerce/google-listings-and-ads/pull/1602#discussion_r943764746.

js/src/utils/createMessageForMultipleErrors.js
eason9487: Yes, it's always `''` but the eslint is unhappy about it. 🤷‍♂️
image

js/src/utils/createErrorMessageForRejectedPromises.test.js
eason9487: I tried adding locally but it looks redundant since it tests the almost same test cases in `js/src/utils/createMessageForMultipleErrors.test.js`. Although I agree this test is too specific to the implementation details rather than its inputs and outputs, I'm not quite sure about the reason for adding the same test cases.

js/src/utils/createMessageForMultipleErrors.js
eason9487: Thanks for the suggestion. Changed and also updated the mismatched function names in its tests along with https://github.com/woocommerce/google-listings-and-ads/pull/1602/commits/b226cc7b77b04febdf039a59a9fd0449df316cc9.

js/src/utils/createErrorMessageForRejectedPromises.test.js
eason9487: I added assertions for the return values in https://github.com/woocommerce/google-listings-and-ads/pull/1602/commits/66bb465f0cdaf0420dff2d51a95a064063cf7e9b.

eason9487:
Thank @tomalec and @ecgan for the suggestions! I added some commits as per the new comments. I guess it might need a new round of code reviews.

js/src/utils/createMessageForMultipleErrors.js
tomalec: I don't quite understand why we have such a rule, but let's keep it then. :)

js/src/utils/createErrorMessageForRejectedPromises.test.js
tomalec: Thanks. Alternatively, I was thinking to test it like 
```js
it( 'Called with one rejected promise, returns the value of `createMessageForMultipleErrors` called with its name, and `true`' 
...
expect( message ).toEqual( createMessageForMultipleErrors( [ 'Promise B' ],	true ) );
```

It still describes more of an implementation detail, but is less redundant and fragile to `createMessageForMultipleErrors` and message changes.

tomalec:
Re-reviewed the code tested locally, LGTM :)

js/src/utils/createErrorMessageForRejectedPromises.test.js
eason9487: Thanks for the suggestion. This approach looks better. I reverted the https://github.com/woocommerce/google-listings-and-ads/commit/66bb465f0cdaf0420dff2d51a95a064063cf7e9b and adopted this in https://github.com/woocommerce/google-listings-and-ads/pull/1602/commits/2426b2af44f3c1028991d5abd28525103043d5b6.

PR #1608, opened 2022-07-21
Add utilities to remove PHP 8.0 specific code of Symfony polyfills

jorgemd24:

Until now we have been importing the Symfony polyfills manually due of the issues with woorelease, see: https://github.com/woocommerce/google-listings-and-ads/issues/331#issuecomment-838372148. This solution was not ideal because it make it harder to maintain and import new polyfills.

Recently an issue with the Cloudflare plugin was reported related to the Normalizer polyfill, see comment: https://github.com/woocommerce/google-listings-and-ads/pull/1588#issuecomment-1190472845. It was loading the GLA Normalizer Polyfill instead of the actual Symfony polyfill that Cloudflare is using in its plugin.

An easier way to make compatible the Symfony polyfills with `woorelease` was deleting the files `bootstrap80.php` and the reference to PHP 8.0 specific code. These files are:

- bootstrap80.php (this is the file that creates the issue with woorelease). This PR deletes this file.
- bootstrap.php - Contains the conditional statement, `if PHP_VERSION >= 80000 require bootstrap80.php`. This PR is deleting the conditional statement.

### Case 1
1. Remove your vendor folder.
3. Run `composer install`.
4. Check these folder `/vendor/symfony/polyfill-intl-normalizer/` and `/vendor/symfony/polyfill-mbstring/`
5. Those folders should not contain the file `bootstrap80.php`
6. The bootstrap.php should not contain the following statement: 

```
if (\PHP_VERSION_ID >= 80000) {
    return require __DIR__.'/bootstrap80.php';
}
```

### Case 2
1. Disable your extensions: `;extension=mbstring.so` and `;extension=intl.so`.
2. Go to `/wp-admin/admin.php?page=connection-test-admin-page`.
3. Sync a product, without clicking async.
4. Not fatal errors should be displayed.

### Additional details:

I wanted to make the script to remove automatically the PHP 8.0 specific code for all folders starting with `sympfony/polyfill-*`  but I saw that some of them, such as `polyfill-ctype` (used by `symfony/validator`) does not have the issue mentioned on https://github.com/woocommerce/google-listings-and-ads/issues/331#issuecomment-838372148 so I think it is better to have a list of polyfills so we can update it manually and have more control over the process.

### Changelog entry

>Tweak - Remove PHP 8.0 specific code of Symfony polyfills


bin/SymfonyPolyfillCleanup.php
```diff
@@ -0,0 +1,234 @@
...
+			return;
+		}
+
+		$contents = file_get_contents( $file ); // phpcs:ignore WordPress.WP.AlternativeFunctions
+
+		$pattern = '/' . preg_quote( $inner_pattern, '/' ) . '/';
+		if ( ! preg_match( $pattern, $contents, $matches, PREG_OFFSET_CAPTURE ) ) {
+			$this->warning( sprintf( 'Inner pattern %s not found in %s', $inner_pattern, $file ) );
+			return;
+		}
```
mikkamp: I would have expected just searching for the if block would have been enough, but the double check to make sure it contains the right import statement works as well.

Also since inner pattern is a regular expression pattern I'd expect it to be something a little more flexible like:
`require .*/bootstrap80.php`
Either one is fine though, since they are consistent across packages.

bin/SymfonyPolyfillCleanup.php
```diff
@@ -0,0 +1,234 @@
...
+
+		unlink( $file );
+	}
+
+	/**
+	 * Find a specific filename pattern within the polyfill folder.
+	 *
+	 * @param string $pattern Regexp pattern to match.
```
mikkamp: I realise that this is copied from the other file, but technically this would be a glob pattern and not a regexp.

bin/SymfonyPolyfillCleanup.php
```diff
@@ -0,0 +1,234 @@
...
+				$this->remove_statement_with_pattern( $bootstrap, "return require __DIR__.'/bootstrap80.php", 'if (\PHP_VERSION_ID >= 80000)' );
+			}
+
+			// Search for bootstrap80.php
+			$bootstraps80 = array_unique(
+				$this->find_library_file_pattern( 'bootstrap80', $polyfill )
+			);
+
+			foreach ( $bootstraps80 as $bootstrap80 ) {
+				$this->remove_file( $bootstrap80 );
```
mikkamp, line 96: This will remove the file, even if the remove statement call fails. It's not an issue since it finds both, but should we prevent the file from being removed if the statement is not found?

mikkamp (approved):
Thanks for the changes. I ran all the tests with and without the PHP extensions enabled and everything is working as ecpected.

I just left a few smaller comments about things I noticed in the code, but none of them are blockers.

jorgemd24:
Thanks, @mikkamp for reviewing this, I addressed all your suggestions!

bin/SymfonyPolyfillCleanup.php
```diff
@@ -0,0 +1,243 @@
...
+	 */
+	protected function remove_statement_with_pattern( string $file, string $inner_pattern, string $statement_pattern ) {
+		if ( ! file_exists( $file ) ) {
+			$this->warning( sprintf( 'File does not exist: %s', $file ) );
+			return false;
+		}
+
+		$contents = file_get_contents( $file ); // phpcs:ignore WordPress.WP.AlternativeFunctions
+
+		$pattern = '/' . $inner_pattern . '/';
```
mikkamp, line 118: I guess it's debatable whether the pattern should be escaped within or outside of this function. Although I thought it was clearer to do it within, because you need to know which delimiter is added within the function to properly escape.
For example if this line was:

```php
$pattern = '#' . $inner_pattern . '#';
```

Then the inner pattern would look different if it needed to include a `/`

bin/SymfonyPolyfillCleanup.php
```diff
@@ -0,0 +1,243 @@
...
+			$bootstraps = array_unique(
+				$this->find_library_file_pattern( 'bootstrap', $polyfill )
+			);
+
+			foreach ( $bootstraps as $bootstrap ) {
+				$statement_results[] = $this->remove_statement_with_pattern( $bootstrap, 'require .*\/bootstrap80.php', 'if (\PHP_VERSION_ID >= 80000)' );
+			}
+
+			// If the statement is not found, skip removing bootstrap80.php
+			if ( ! in_array( true, $statement_results, true ) ) {
```
mikkamp: This check works fine, but it looks a little odd because if there were multiple bootstrap files then it would still continue if at least one of them passed. Since we aren't passing a pattern but a constant string to `find_library_file_pattern` we can always expect only one result to be returned. Maybe it's easier to just take the first result to simplify the code a little:

```php
			// Search for bootstrap.php
			$bootstraps = $this->find_library_file_pattern( 'bootstrap', $polyfill );
			$bootstrap  = reset( $bootstraps );

			if ( ! $this->remove_statement_with_pattern( $bootstrap, 'require .*\/bootstrap80.php', 'if (\PHP_VERSION_ID >= 80000)' ) ) {
				continue;
			}
```

Similar can be done for `bootstrap80` since that will also always return only one file.

mikkamp (commented):
Thanks for the additional changes, they still work as intended. I just left a small comment about escaping the pattern and one about just handling the first entry.

bin/SymfonyPolyfillCleanup.php
jorgemd24: The issue I found with the previous implementation was that `preg_quote` was escaping `.` and `*` from this string `require .*\/bootstrap80.php`, that's why I decided to remove it, however, if you know a better way to pass regex expressions to `preg_quote` I will be happy to add it.

mikkamp: Makes sense, leaving it as is, would probably be best then.

PR #1611, opened 2022-08-02
Check if the store is accessible to all users

songchen2022:

This PR is based on [Add PolicyComplianceCheckController #1563](https://github.com/woocommerce/google-listings-and-ads/pull/1563)

This PR adds three checks in the server side to check if the website is accessible to all users:
1. Check for restrictions in robots.txt
2. Check if sample of product landing pages leads to a 404 error 
3. Check if sample product landing pages have redirects through 3P domains
4. Check if the store has a complete checkout process
5. Check if the store has a secure checkout process
6. Check if the store set the default refund return policy page
7. Check if the store has the Geo restrictions

### Screenshots:
image

We only test it by running the unit test
$ vendor/bin/phpunit

In the Dev console:
We could test it works or not by running

wp.data.resolveSelect( 'wc/gla' ).getPolicyCheck().then((data) => { console.log("policy_check", data) }).catch((error) => console.log("error",error))

Screen Shot 2022-08-01 at 11 45 07 PM

Screen Shot 2022-08-21 at 11 40 51 PM


js/src/hooks/usePolicyCheck.js
```diff
@@ -0,0 +1,30 @@
...
+ *	'policy_check' =>	[
+ *				'allowed_countries'    	=> true,
+ *                              'robots_restriction'    => false,
+ *                              'page_not_found_error'  => false,
+ *                              'page_restricts'    	=> false,
+ *				'store_ssl'         	=> true,
+ *				'payment_gateways'  	=> true,
+ *				'refund_returns' 	=> true,
+ *				]
+ * }
```
mikkamp: Can we use the right spacing here so it lines up nicely.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+		return false;
+	}
+
+	/**
+	 * Check if the merchant set the restrictions in robots.txt or not in the store.
+	 *
+	 * @return bool
+	 */
+	public function has_restriction(): bool {
+		return ! $this->robots_allowed( get_site_url() );
```
mikkamp: We have the `PluginHelper::get_site_url` function, can we use that so we always use the same URL throughout the plugin.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+
+		$agents = [ preg_quote( '*', null ) ];
+		if ( $useragent ) {
+			$agents[] = preg_quote( $useragent, null );
+		}
+		$agents = implode( '|', $agents );
+
+		// location of robots.txt file
+
+		$robotstxt = file( "http://{$parsed['host']}/robots.txt" );
```
mikkamp: Why are we switching to http here? We are already requiring them to use SSL. Also the site URL we pass to this function will already be `https://domain.test` so why can't we use something like `trailingslashit( $url ) . 'robots.txt'`. That means we no longer need to use `wp_parse_url` either.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+	 */
+	public function has_restriction(): bool {
+		return ! $this->robots_allowed( get_site_url() );
+	}
+
+	/**
+	 * Check if the robots.txt has restrictions or not in the store.
+	 *
+	 * @param string $url
+	 * @param bool   $useragent
```
mikkamp: This parameter is only by default a boolean, but it's usage within the function implies it's a string containing the user agent. Is it useful to have this parameter if we are only going to check for user agent `*`?

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+	 */
+	private function robots_allowed( $url, $useragent = false ) {
+		// parse url to retrieve host and path
+		$parsed = wp_parse_url( $url );
+
+		$agents = [ preg_quote( '*', null ) ];
+		if ( $useragent ) {
+			$agents[] = preg_quote( $useragent, null );
+		}
+		$agents = implode( '|', $agents );
```
mikkamp: We use `$agents` in a regular expression where we use `/` as the delimiter. So we should call `preg_quote` with the delimiter.
```suggestion
		$agents = [ preg_quote( '*', '/' ) ];
		if ( $useragent ) {
			$agents[] = preg_quote( $useragent, '/' );
		}
		$agents = implode( '|', $agents );
```

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+				$rules[] = preg_quote( trim( $regs[1] ), '/' );
+			}
+		}
+
+		foreach ( $rules as $rule ) {
+			// check if page is disallowed to us
+			if ( preg_match( "/^$rule/", $parsed['path'] ) ) {
+				return false;
+			}
+		}
```
mikkamp: Site URL is returned without a trailing slash, which means that the `wp_parse_url` won't have a path. So this code generates a warning:

```
PHP Warning:  Undefined array key "path"
```

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+		$rule_applies = false;
+		foreach ( $robotstxt as $line ) {
+			// skip blank lines
+			$line = trim( $line );
+			if ( ! $line ) {
+				continue;
+			}
+
+			// following rules only apply if User-agent matches $useragent or '*'
+			if ( preg_match( '/^\s*User-agent: (.*)/i', $line, $match ) ) {
```
mikkamp: Here we are explicitly requiring one space after the colon. whereas in the "Dissallow" regex it would allow 0 or multiple spaces which will be trimmed.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+	 */
+	public function has_page_not_found_error(): bool {
+		$ids = wc_get_products(
+			[
+				'return' => 'ids',
+				'limit'  => 1,
+			]
+		);
+		if ( ! empty( $ids ) ) {
+			$url     = get_permalink( $ids[0] );
```
mikkamp: Since we are using `wc_get_products` it would be nice to return the object (remove return parameter) and then use the crud function to [get the permalink](https://woocommerce.github.io/code-reference/classes/WC-Product.html#method_get_permalink). Ex: `$ids[0]->get_permalink()`

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+			$headers = get_headers( $url );
```
mikkamp: I would suggest to use the standard WordPress functions here for performing a request, wp_remote_get in combination with wp_remote_retrieve_headers or even wp_remote_retrieve_response_code, since we seem to be comparing a response code (might also be more accurate if the protocol version is different). These functions also run a lot of filters which means plugins can hook into them to make sure these requests are sent correctly.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+			if ( ! empty( $headesr ) && $headers[0] !== 'HTTP/1.1 200 OK' ) {
```
mikkamp: There is a typo here `$headesr`

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,230 @@
...
+		}
+		return false;
+	}
+
+	/**
+	 * Check if the store sample product landing pages has redirects through 3P domains.
+	 *
+	 * @return bool
+	 */
+	public function has_redirects(): bool {
```
mikkamp, line 92: This function is a repeat of the previous one with the only difference that it's comparing a different header, would a common helper function be useful where we can pass the required header or a comparison function if one needs to equal and the other not equal.

mikkamp (commented):
I set the base branch to the feature branch since the branch `songchen2022:policy_controller` isn't available in this repo. I'm going to cover only the changes introduced in this branch, so I won't repeat comments from PR #1563 

Thanks for adding the additional checks, I added some comments as I went through the code.

js/src/hooks/usePolicyCheck.js
songchen2022: Done.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Thanks for your suggestions! trailingslashit( $url ) . 'robots.txt' is much better.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,240 @@
...
+	 */
+	public function get_landing_page_headers(): array {
+		$ids = wc_get_products(
+			[
+				'return' => 'ids',
+				'limit'  => 1,
+			]
+		);
+		if ( ! empty( $ids ) ) {
+			$response = wp_remote_get( $ids[0]->get_permalink() );
```
eason9487: When I tried to check the implementation of the frontend wp-data store after getting the `mc/policy_check` API response, it responded with 500 error after running `wp.data.select('wc/gla').getPolicyCheck()` via the Console on the browser DevTool. I guess there might be a problem here.

PHP error log:
```
CRITICAL Uncaught Error: Call to a member function get_permalink() on int in /var/www/html/wp-content/plugins/google-listings-and-ads/src/MerchantCenter/PolicyComplianceCheck.php:115
Stack trace:
#0 /var/www/html/wp-content/plugins/google-listings-and-ads/src/MerchantCenter/PolicyComplianceCheck.php(79): Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\PolicyComplianceCheck->get_landing_page_headers()
#1 /var/www/html/wp-content/plugins/google-listings-and-ads/src/API/Site/Controllers/MerchantCenter/PolicyComplianceCheckController.php(72): Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\PolicyComplianceCheck->has_page_not_found_error()
#2 /var/www/html/wp-includes/rest-api/class-wp-rest-server.php(1140): Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter\PolicyComplianceCheckController->Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter\{closure}()
#3 /var/www/html/wp-includes/rest-api/class-wp-rest-server.php(987): WP_REST_Server->respond_ in /var/www/html/wp-content/plugins/google-listings-and-ads/src/MerchantCenter/PolicyComplianceCheck.php on line 115
```

js/src/data/test/reducer.test.js
```diff
@@ -418,87 +413,50 @@ describe( 'reducer', () => {
 	} );
 
 	describe( 'Policy Check', () => {
-		it( 'should return with allowed countries', () => {
+		it( 'should return with policy check info', () => {
```
eason9487: Since the reducer only takes care of placing the received data of the `TYPES.POLICY_CHECK` action to a specific position in the data store tree, this whole test case could be simplified by adding a new tuple `[ TYPES.POLICY_CHECK, 'data', 'mc.policy_check' ],` to the below test case:

https://github.com/woocommerce/google-listings-and-ads/blob/0197013965630d1563fb65119f93f9227b00c262/js/src/data/test/reducer.test.js#L775-L793

js/src/hooks/usePolicyCheck.js
```diff
@@ -0,0 +1,30 @@
...
+ *				'allowed_countries'	=> true,
+ *                              'robots_restriction'	=> false,
+ *                              'page_not_found_error'	=> false,
+ *                              'page_restricts'	=> false,
+ *				'store_ssl'		=> true,
+ *				'payment_gateways'	=> true,
+ *				'refund_returns'	=> true,
+ *				]
+ * }
+ * ```
```
eason9487: The indentations are mixed of spaces and tabs, and the syntax of this example is not a valid JSON format nor fits with the actual data structure. (`data` should be a flat object without the `policy_check` layer.)

eason9487 (commented):
Thanks for the changes. I added some comments that mainly related to the frontend side files.

src/MerchantCenter/PolicyComplianceCheck.php
mikkamp: As mentioned [here](https://github.com/woocommerce/google-listings-and-ads/pull/1611#discussion_r935744640) the return parameter needs to be removed, which defaults to returning an array of WC_Product instead of an array of ID's.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -0,0 +1,240 @@
...
+		return false;
+	}
+
+	/**
+	 * Check if the store sample product landing pages has redirects through 3P domains.
+	 *
+	 * @return bool
+	 */
+	public function has_redirects(): bool {
+				$headers = $this->get_landing_page_headers();
```
eason9487: Even though the PHPCS config of this repo doesn't check indentations with the `exact` ruleset, it would be great if the indentations could be aligned here.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Thanks for pointing this out. Updated.

js/src/data/test/reducer.test.js
songchen2022: Thanks for your suggestions! Updated.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Done.

js/src/hooks/usePolicyCheck.js
songchen2022: Updated.

findingsimple:
Hi @songchen2022 👋 

Thanks for your work on this. Just some general feedback to help us review your work more easily.

1. Best to try to avoid merging the `develop` branch into an in-progress pr. Because you have done a few merges you've ended up with 192 files at the moment that are showing as being changed - which as you can appreciate makes it tricky for any fresh eyes coming in to review or see what the actual changes are.

2. Remember to double-check code linting to pick up [incorrect indents, spaces, tabs etc.](https://github.com/woocommerce/google-listings-and-ads/pull/1611#discussion_r943244813)  - it's an easy thing to do before a request for a review.

3. Remember to test the code to make sure it works as expected (at least make sure the detailed test instructions work as intended) - it can be quite frustrating for reviewers if they have to figure out how to get things to work as expected or block testing completely. [I think one of the examples is this.](https://github.com/woocommerce/google-listings-and-ads/pull/1611#discussion_r943218796)

4. Be extra careful when copy and pasting code to update it completely to reflect the new usage. [An example is copy and pasting event tracking codes with a number of properties](https://github.com/woocommerce/google-listings-and-ads/pull/1612#discussion_r942187588) - it's important we don't mix these up.

5. Similarly double check the JSDoc are updated correctly - either after a copy of paste or when adding new ones.

Focussing on the basics before requesting the review will help keep the review process running smoothly and save time in the long run.

songchen2022:
@jconroy Thanks for your suggestions! I‘ll try my best to work on it. 
I usually don't merge to `develop`. However, I kept getting the "version 6.8.1 cannot be found" error this time. So I merged to develop to test it. I won't do the merge like this next time. 
image
Could you please help merge this branch `[woocommerce:feature/policy-compliance-checks]`with develop branch?
Sorry for any inconvenience.

mikkamp:
Hi @songchen2022
I've gone ahead and merged develop into `feature/policy-compliance-checks` so it's only showing the relevant commits/changes in this PR. However you'll still need to resolve the conflicts.

Regarding the error with not finding the 6.8.1 branch. That was an oversight in the WooCommerce repo, the tag was added yesterday, so that should be resolved now.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -68,6 +70,116 @@ public function is_accessible(): bool {
...
+	public function get_landing_page_headers(): array {
+		$products = wc_get_products(
+			[
+				'limit' => 1,
+			]
+		);
+		if ( ! empty( $products ) ) {
+			$response = wp_remote_get( $products[0]->get_permalink() );
+			if ( is_array( $response ) && ! is_wp_error( $response ) ) {
+				return $response['headers'];
```
mikkamp: If we run this code and it actually fetches the URL then this fails with a fatal error because the `headers` returns an instance of [Requests_Utility_CaseInsensitiveDictionary](https://developer.wordpress.org/reference/classes/requests_utility_caseinsensitivedictionary).

Since what we really want to know though is the response code, I don't think we should have this helper function, instead we should create a helper function for fetching the landing page URL (see next comment).

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -68,6 +70,116 @@ public function is_accessible(): bool {
...
+	 * Check if the store sample product landing pages lead to a 404 error.
+	 *
+	 * @return bool
+	 */
+	public function has_page_not_found_error(): bool {
+		$headers = $this->get_landing_page_headers();
+		if ( ! empty( $headers ) && $headers[0] !== 'HTTP/1.1 200 OK' ) {
+			return true;
+		}
+		return false;
```
mikkamp: If we send the request with `wp_remote_get` then `$headers[0]` won't be set since it's not an array. As previously mentioned we should be using the function `wp_remote_retrieve_response_code` so we can compare just an integer.

So I suggest the function gets rewritten like this to compare the response code isn't a 404.
```suggestion
		$url      = $this->get_landing_page_url();
		$response = wp_remote_get( $url );

		if ( 200 !== wp_remote_retrieve_response_code( $response ) ) {
			return true;
		}

		return false;
```

The helper function `get_landing_page_url` can be created as follows:
```php
	/**
	 * Returns a product page URL, uses homepage as a fallback.
	 *
	 * @return string Landing page URL.
	 */
	private function get_landing_page_url(): string {
		$products = wc_get_products( [ 'limit' => 1 ] );
		if ( ! empty( $products ) ) {
			return $products[0]->get_permalink();
		}

		return $this->get_site_url();
	}
```

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -68,6 +70,116 @@ public function is_accessible(): bool {
...
+	 */
+	public function has_redirects(): bool {
+		$headers = $this->get_landing_page_headers();
+		if ( ! empty( $headesr ) && isset( $headers[0] ) ) {
+			if ( $headers[0] === 'HTTP/1.1 302 Found' ) {
+				// this is the URL where it's redirecting
+				return true;
+			}
+		}
+			return false;
```
mikkamp: As mentioned in the previous comment it's better to use `wp_remote_retrieve_response_code` to retrieve the response code. Also the string comparison isn't very accurate because there are [many redirect codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_redirection). Lastly by default a `wp_remote_get` will follow redirects so we need to set the `redirection` argument to 0 to detect a redirect response code.

So I suggest to create this function like this:
```suggestion
		$url      = $this->get_landing_page_url();
		$response = wp_remote_get( $url, [ 'redirection' => 0 ] );
		$code     = wp_remote_retrieve_response_code( $response );

		if ( $code >= 300 && $code <= 399 ) {
			return true;
		}

		return false;
```

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -68,6 +70,116 @@ public function is_accessible(): bool {
...
+		foreach ( $robotstxt as $line ) {
+			// skip blank lines
+			$line = trim( $line );
+			if ( ! $line ) {
+				continue;
+			}
+
+			// following rules only apply if User-agent matches '*'
+			if ( preg_match( '/^\s*User-agent:\s*(.*)/i', $line, $match ) ) {
+				$rule_applies = preg_match( "/($agents)/i", $match[1] );
```
mikkamp: If we are not going to pass agents to the function and are going to only use compare to agent *, then the two lines above to escape and implode become unnecessary and we can just do a simple comparison here:

```suggestion
				$rule_applies = '*' === $match[1];
```

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -68,6 +70,116 @@ public function is_accessible(): bool {
...
+	 *
+	 * @param string $url
+	 * @return bool
+	 */
+	private function robots_allowed( $url ) {
+		$agents = [ preg_quote( '*', '/' ) ];
+		$agents = implode( '|', $agents );
+
+		// location of robots.txt file
+		$robotstxt = file( trailingslashit( $url ) . 'robots.txt' );
```
mikkamp: On my test server this request is blocked (unless I override the configuration), and there could be other server configurations which do not allow file access to an external URL. It's probably better to switch to a `wp_remote_get` since that's generally confirmed to work. Note that we'll need to break up the content into lines, if we want to loop through it line by line (which is done by default by the `file` function).

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -68,6 +70,116 @@ public function is_accessible(): bool {
...
+			if ( preg_match( '/^\s*User-agent:\s*(.*)/i', $line, $match ) ) {
+				$rule_applies = preg_match( "/($agents)/i", $match[1] );
+			}
+			if ( $rule_applies && preg_match( '/^\s*Disallow:\s*(.*)/i', $line, $regs ) ) {
+				// an empty rule implies full access - no further tests required
+				if ( ! $regs[1] ) {
+					return true;
+				}
+				// add rules that apply to array for testing
+				$rules[] = preg_quote( trim( $regs[1] ), '/' );
```
mikkamp: Here again if all we are going to compare is the disallow of `/` then we don't need to add it to an array of rules and we can just compare if `'/' === trim( $regs[1] )`.

Although combined with matching a user agent of `*`, I'm wondering how useful this check will be. Because it will need to have a robots file with the following content to match:

```
User-agent: *
Disallow: /
```

Which would imply that they are restricting the full site.
Even if that pattern is not in the robots.txt file it's still possible to for example, restrict all product pages.

mikkamp (commented):
Hi @songchen2022 

I went ahead and tested the policy checks and noticed it was returning a fatal error because the return type of `get_landing_page_headers` was not an array. So I went through each function individually and suggested some alternatives to use.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Thanks for your suggestions! Based on this [PRD](https://docs.google.com/document/d/1NxMLzh5rb6KBx7xP7KZF2q4WeoAQJmnPTxE3B8Ss3jY/edit?resourcekey=0-QRep8wnQJgcCQRudP770GA#bookmark=id.km9dsgduk5uq), we could only do the general checking. Do you have any good suggestions?

mikkamp: I don't have any suggestions for improving the checks. I was mainly pointing out that it will catch very few sites that don't match those patterns. Since these are only preliminary checks to get them through the onboarding and it will be checked again through an external service, it seems like this is as far as we can go for now.

js/src/data/test/reducer.test.js
```diff
@@ -405,6 +405,9 @@ describe( 'reducer', () => {
 		it( 'should return with policy check info', () => {
 			const data = {
 				allowed_countries: true,
+				robots_restriction: false,
```
eason9487: Based on [this comment](https://github.com/woocommerce/google-listings-and-ads/pull/1611#discussion_r943238539), [this test case](https://github.com/songchen2022/google-listings-and-ads/blob/e9e5ec823168b40908e61cea8fb2ef2e80ddba31/js/src/data/test/reducer.test.js#L404-L451) could be removed since it's already covered [here](https://github.com/songchen2022/google-listings-and-ads/blob/e9e5ec823168b40908e61cea8fb2ef2e80ddba31/js/src/data/test/reducer.test.js#L842).

js/src/data/test/reducer.test.js
```diff
@@ -780,18 +795,51 @@ describe( 'reducer', () => {
...
+				'target_audience',
+				'mc.target_audience',
+			],
 			[ TYPES.RECEIVE_MC_SETUP, 'mcSetup', 'mc_setup' ],
-			[ TYPES.RECEIVE_MC_PRODUCT_STATISTICS, 'mcProductStatistics', 'mc_product_statistics' ],
+			[
+				TYPES.RECEIVE_MC_PRODUCT_STATISTICS,
+				'mcProductStatistics',
+				'mc_product_statistics',
+			],
```
eason9487: It'd be great if we could revert these changes to the previous format for a bit better readability and also to conform to the [code comment above](https://github.com/songchen2022/google-listings-and-ads/blob/e9e5ec823168b40908e61cea8fb2ef2e80ddba31/js/src/data/test/reducer.test.js#L793-L795).

js/src/hooks/usePolicyCheck.js
```diff
@@ -11,12 +11,13 @@ import useAppSelectDispatch from './useAppSelectDispatch';
...
- *				'payment_gateways'  	=> true,
- *				'refund_returns' 	=> true,
- *				]
+ *		"allowed_countries": true,
+ *		"robots_restriction": false,
+ *		"page_not_found_error": false,
+ *		"page_restricts": false,
+ *   		"store_ssl": true,
+ * 		"payment_gateways": true,
+ *		"refund_returns": true
```
eason9487: Some indentations are still mixed with spaces and tabs. And I believe the key `page_restricts` should be `page_redirects`.

```suggestion
 *   "allowed_countries": true,
 *   "robots_restriction": false,
 *   "page_not_found_error": false,
 *   "page_redirects": false,
 *   "store_ssl": true,
 *   "payment_gateways": true,
 *   "refund_returns": true
```

eason9487 (commented):
Thanks for the updates. I viewed the client-related changes and left two comments about the test case, and another one for fixing mismatched content in JSDoc.

js/src/data/test/reducer.test.js
songchen2022: Fixed. Thanks for reminder!

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Thanks for your reply! I totally agree with you. This is only preliminary checks now.

eason9487 (approved):
@songchen2022 thanks for the adjustments. The changes to .js files look good to me.

src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -68,6 +70,109 @@ public function is_accessible(): bool {
...
+		}
+
+		$body      = wp_remote_retrieve_body( $response );
+		$robotstxt = preg_split( "/\r\n|\n|\r/", $body );
+
+		if ( empty( $robotstxt ) ) {
+			return true;
+		}
+
+		$rules        = [];
```
mikkamp: This variable is no longer used so we can remove this line.

mikkamp (approved):
Thanks for all the changes, it's working well for me now. I just left one small comment on a variable that's no longer used.

Besides that it looks ready to merge to the feature branch.

src/MerchantCenter/PolicyComplianceCheck.php
songchen2022: Thanks for pointing this out. Updated.

PR #1612, opened 2022-08-02
Policy Check UI 

songchen2022:

Add policy check UI 

This PR is based on this [Check if the store is accessible to all users #1611](https://github.com/woocommerce/google-listings-and-ads/pull/1611)

### Screenshots:
Screen Shot 2022-08-02 at 1 06 39 PM

https://drive.google.com/file/d/1uSKtVZXgU5z-Kalr1jJ4WD5mRjitbTvk/view?usp=sharing&resourcekey=0-uLWjTmR3upIHojUlFweTaQ

 Figma:
https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=7450%3A120810


eason9487:
Hi @songchen2022, I changed the base to `feature/policy-compliance-checks` to reduce some commits that belong to other PRs, but it turned out there are two conflicts. Could you check if those conflicts should be resolved? Or the changed base branch is not the expected operation?

songchen2022:
> 

Thanks so much for changing the base to `feature/policy-compliance-checks`. Conflicts have been solved.

js/src/setup-mc/setup-stepper/store-requirements/index.js
```diff
@@ -96,7 +100,7 @@ export default function StoreRequirements() {
 		<StepContent>
 			<StepContentHeader
 				title={ __(
-					'Confirm store requirements',
+					'[internal] Confirm store requirements',
```
eason9487: As per the visual design in Figma, this title seems not to be changed. Is this an expected change in this project?

js/src/setup-mc/setup-stepper/store-requirements/index.js
```diff
@@ -22,7 +22,11 @@ import ContactInformation from '.~/components/contact-information';
 import AppButton from '.~/components/app-button';
 import AppSpinner from '.~/components/app-spinner';
 import PreLaunchChecklist from './pre-launch-checklist';
-import checkErrors from './pre-launch-checklist/checkErrors';
+
+function _checkErrors() {
+	const errors = {};
+	return errors;
+}
```
eason9487: I'm not quite sure if this is a temporary implementation, but the form validation should not be changed to an empty check here. It leads to the onboarding setup can be completed without accepting all required items in the pre-launch checklist.

And about the convention in this codebase, a function name usually doesn't start with `_` if there is no particular reason.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +103,443 @@ const PreLaunchChecklist = ( props ) => {
...
+								<div id="checkbox">
+									<CheckboxControl
+										{ ...getInputProps( 'website_live' ) }
+									></CheckboxControl>
+									<div id="panel">
+										<Panel>
+											<PanelBody
+												key={
+													'why-do-i-need-a-wp-account'
+												}
```
eason9487: The `key` prop in these `<PanelBody>` is not required when they are not rendered via a loop.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +103,443 @@ const PreLaunchChecklist = ( props ) => {
...
+												title={ __(
+													'My store is live and accessible to all users',
+													'google-listings-and-ads'
+												) }
+												initialOpen={ true }
+												onToggle={ getPanelToggleHandler(
+													'gla_faq',
+													'why-do-i-need-a-wp-account',
+													'setup-mc-accounts'
```
eason9487: The name of event tracking seems to be not so accurate since this is not like a pure FAQ section. I'd suggest using another name to distinguish it from the existing event trackings of FAQ.

In addition, using `'why-do-i-need-a-wp-account'` and `'setup-mc-accounts'` as the `id` and `context` of this event tracking properties is incorrect. The `id` should be associated with this check, and the `context` could be omitted if these event trackings use another name, otherwise, the `context` should be associated with this onboarding step as well.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +103,443 @@ const PreLaunchChecklist = ( props ) => {
...
+															link: (
+																<AppDocumentationLink
+																	context="setup-mc-checklist"
+																	linkId="check-checkout-process"
+																	type="external"
+																	href="https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#store-is-live"
+																/>
+															),
+														}
+													) }
```
eason9487: This link is not required to be wrapped by `createInterpolateElement` when the content is not a hybrid of i18n text and HTML that needs to be replaced dynamically. And, the `linkId` prop is not correct. Suggest replacing it with:

```js
<AppDocumentationLink
	context="setup-mc-checklist"
	linkId="check-website-is-live"
	type="external"
	href="https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#store-is-live"
>
	{ __(
		'Learn more about common landing page issues and how to fix them',
		'google-listings-and-ads'
	) }
</AppDocumentationLink>
```

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +103,443 @@ const PreLaunchChecklist = ( props ) => {
 				<Section.Card>
 					<Section.Card.Body>
 						<VerticalGapLayout size="large">
-							<div className="gla-pre-launch-checklist__checkbox_popover">
+							{ ! values.website_live && (
+								<div id="checkbox">
```
eason9487: The `id` attribute of HTML tags should be unique in a page. Suggest using a CSS class name or replacing it with a `@wordpress/components` component `<Flex align="normal">` for composing the flexbox layout.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +103,443 @@ const PreLaunchChecklist = ( props ) => {
...
+									<CheckboxControl
+										{ ...getInputProps( 'website_live' ) }
+									></CheckboxControl>
+									<div id="panel">
```
eason9487: The concern of `id` attribute is similar to the above. And it could be replaced with `<FlexBlock>` as well.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +103,443 @@ const PreLaunchChecklist = ( props ) => {
...
+																/>
+															),
+														}
+													) }
+												</p>
+												<Button
+													isPrimary
+													disabled={
+														values.website_live
+													}
```
eason9487: This could be omitted since it won't be rendered when `values.website_live` is true.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +103,443 @@ const PreLaunchChecklist = ( props ) => {
...
+													onClick={ () => {
+														setValue(
+															'website_live',
+															true
+														);
+														saveSettings(
+															values.website_live,
+															true
+														);
```
eason9487: After calling `setValue`, it will trigger `<Form>`'s `onChange` callback in the parent component, so the `saveSettings` call here can be removed.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -3,24 +3,72 @@
...
+		if ( data.refund_returns !== values.refund_tos_visible ) {
+			setValue( 'refund_tos_visible', data.refund_returns );
+		}
+
+		if ( data.payment_gateways !== values.payment_methods_visible ) {
+			setValue( 'payment_methods_visible', data.payment_gateways );
+		}
+
+		setValue( 'checked', true );
+	}
```
eason9487: I'm not pretty sure about the intention of this part. Please clarify if I get it wrong.

It would be better to finish the auto-checked handling before setting the `initialValues` prop of Form. Changing values here will cause the `<Form>`'s `onChange` callback to be triggered several times and further trigger the API requests several times. In addition, setting a non-existing value `checked` to Form here leads to the above problems as well, and the `checked` value is also passed as one of data properties in the API request payload.

If it's expected to be run one time in the initial phase, I'd suggest moving this to `StoreRequirements` component, and changing the implementation to use React hooks or a custom hook.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.scss
```diff
@@ -8,4 +8,13 @@
 			vertical-align: text-bottom;
 		}
 	}
+
+	#checkbox {
+		display: flex;
+		justify-content: stretch;
+
+		#panel {
+			min-width: 600px;
```
eason9487: I think it's not necessary to set `min-width` since it breaks the layout.

image

eason9487 (commented):
Thanks for the changes.

There are some concerns that do not belong to specific code fragments or affect the wider scope:
1. All parameters of each `getPanelToggleHandler` function call need to be checked.
2. The `linkId` prop of each `<AppDocumentationLink>` needs to be checked if its value has the correct semantic meaning or identity.
3. It would be great if the JSDoc of new tracking events could be added [here](https://github.com/songchen2022/google-listings-and-ads/blob/0a6ea96987adf66456ad0a6a6ba6b8b60564bcc7/js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js#L26-L28) as well.
4. It's not a problem, but it would be a bit nicer to create a new component for the composition of CheckboxControl and Panel components. In this way, it could reduce some repetitive code in the `PreLaunchChecklist` component.

And I found these requirements in the Figma file. I wonder will they be added in a subsequent PR or are they missed in this PR?

1. > When an item is confirmed, it moves up and pushes down any unconfirmed items above it. In this way, checked and unchecked items are grouped together.
2. > Clicking the ’Confirm’ button brings it to a **Confirmed state** with no ability to uncheck or go back.
3. > **Auto-checked state**: When an item is detected as completed, the wording is in the 1st person. i.e. “MY store is...”, “I have a complete...”
4. > **Unchecked/unconfirmed state**: When an item is unchecked and needs to be confirmed, the wording is in the 2nd person. i.e. “Confirm YOUR store is...”, “Confirm YOU have a...”

js/src/setup-mc/setup-stepper/store-requirements/index.js
songchen2022: This is not expected change. Updated.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Thanks for your suggestions. Updated.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Thanks for your suggestion! Updated.

js/src/setup-mc/setup-stepper/store-requirements/index.js
songchen2022: Thanks for pointing this out. Updated.

songchen2022:
> Thanks for the changes.
> 
> There are some concerns that do not belong to specific code fragments or affect the wider scope:
> 
> 1. All parameters of each `getPanelToggleHandler` function call need to be checked.
> 2. The `linkId` prop of each `<AppDocumentationLink>` needs to be checked if its value has the correct semantic meaning or identity.
> 3. It would be great if the JSDoc of new tracking events could be added [here](https://github.com/songchen2022/google-listings-and-ads/blob/0a6ea96987adf66456ad0a6a6ba6b8b60564bcc7/js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js#L26-L28) as well.
> 4. It's not a problem, but it would be a bit nicer to create a new component for the composition of CheckboxControl and Panel components. In this way, it could reduce some repetitive code in the `PreLaunchChecklist` component.
> 
> And I found these requirements in the Figma file. I wonder will they be added in a subsequent PR or are they missed in this PR?
> 
> 1. > When an item is confirmed, it moves up and pushes down any unconfirmed items above it. In this way, checked and unchecked items are grouped together.
> 2. > Clicking the ’Confirm’ button brings it to a **Confirmed state** with no ability to uncheck or go back.
> 3. > **Auto-checked state**: When an item is detected as completed, the wording is in the 1st person. i.e. “MY store is...”, “I have a complete...”
> 4. > **Unchecked/unconfirmed state**: When an item is unchecked and needs to be confirmed, the wording is in the 2nd person. i.e. “Confirm YOUR store is...”, “Confirm YOU have a...”

Thanks so much for reviewing and suggestions! I will discuss with internal team about the JSDoc part. For the comment 4, It's a good suggestions. Creating a new component for the composition of CheckboxControl and Panel will reduce some repetitive code but I found we need to pass around 10 parameters to the component which will look not very easily readable. 

For the other 4 comments, we don't have the requirement as 1. I updated the code to follow your comments 2,3,4. 
Thanks again. Please correct me if I'm wrong.

songchen2022:
@eason9487  Hi Eason, could you please help me take a look this PR again? Thanks!

js/src/setup-mc/setup-stepper/store-requirements/index.js
```diff
@@ -91,6 +94,36 @@ export default function StoreRequirements() {
...
+	}
+
+	if ( ! policyChecked ) {
+		const websiteLive =
+			policyCheckData.allowed_countries &&
+			! policyCheckData.robots_restriction &&
+			! policyCheckData.page_not_found_error &&
+			! policyCheckData.page_redirects;
+		if ( websiteLive !== settings.website_live ) {
+			settings.website_live = policyCheckData.website_live;
```
eason9487: I think the `policyCheckData.website_live` at line 108 should be `websiteLive`.

js/src/setup-mc/setup-stepper/store-requirements/index.js
```diff
@@ -91,6 +94,36 @@ export default function StoreRequirements() {
...
+		}
+
+		if (
+			policyCheckData.payment_gateways !==
+			settings.payment_methods_visible
+		) {
+			settings.payment_methods_visible = policyCheckData.payment_gateways;
+		}
+		setPolicyChecked( true );
+	}
```
eason9487: The `settings` data got from the `.~/data` store is not expected to be mutated directly, and this way doesn't guarantee a mutated data/state is updated on the rendering UI since React won't know an object's property value was changed. In addition, it won't trigger form's `onChange` to save the `settings` data to DB.

I'd suggest using something like the following for preprocessing the auto-checked state and data saving.

```js
	// Preprocess the auto-checked state and data saving.
	useEffect( () => {
		if ( preprocessed || ! settings || ! policyCheckData ) {
			return;
		}

		const newSettings = { ...settings };

		const websiteLive =
			policyCheckData.allowed_countries &&
			! policyCheckData.robots_restriction &&
			! policyCheckData.page_not_found_error &&
			! policyCheckData.page_redirects;

		if ( websiteLive !== settings.website_live ) {
			newSettings.website_live = websiteLive;
		}

		if ( policyCheckData.store_ssl !== settings.checkout_process_secure ) {
			newSettings.checkout_process_secure = policyCheckData.store_ssl;
		}

		if ( policyCheckData.refund_returns !== settings.refund_tos_visible ) {
			newSettings.refund_tos_visible = policyCheckData.refund_returns;
		}

		if (
			policyCheckData.payment_gateways !==
			newSettings.payment_methods_visible
		) {
			newSettings.payment_methods_visible =
				policyCheckData.payment_gateways;
		}

		const promise = isEqual( newSettings, settings )
			? Promise.resolve()
			: saveSettings( newSettings );

		promise.finally( () => setPreprocessed( true ) );
	}, [ preprocessed, policyCheckData, settings, saveSettings ] );
```

In this example, the `preprocessed` and `setPreprocessed` were renamed from `policyChecked` and `setPolicyChecked`.

Also, with this approach, these two loading status controls
```js
if ( ! settings ) {
	return <AppSpinner />;
}
if ( ! policyCheckData ) {
	return <AppSpinner />;
}
```

could be replaced with a single one:
```js
	if ( ! preprocessed ) {
		return <AppSpinner />;
	}
```

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+										<Panel>
+											<PanelBody
+												title={ __(
+													'My store is live and accessible to all users',
+													'google-listings-and-ads'
+												) }
+												initialOpen={ true }
+												onToggle={ getPanelToggleHandler(
+													'website_live',
+													'pre-launch-checklist',
```
eason9487: I think all calls to `getPanelToggleHandler` function have the first two parameters reversed. The tracking event name could be the same for grouping these similar event tracking and be distinguished by a unique `id`.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+										{ ...getInputProps( 'website_live' ) }
+									></CheckboxControl>
+									<FlexBlock>
+										<Panel>
+											<PanelBody
+												title={ __(
+													'My store is live and accessible to all users',
+													'google-listings-and-ads'
+												) }
+												initialOpen={ true }
```
eason9487: Based on the design in Figma, looks like this initial state is closed?

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+											</PanelBody>
+										</Panel>
+									</FlexBlock>
+								</div>
+							) }
+
+							{ values.payment_methods_visible && (
+								<CheckboxControl
+									label={
+										<span className="gla-pre-launch-checklist__checkbox_popover">
```
eason9487: It would be great if we could remove all `<span className="gla-pre-launch-checklist__checkbox_popover">` layers and the related CSS styles because it's no longer a popover UI and it's an extra layer that can be omitted.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+											<span className="checkbox-label">
```
eason9487: Similar to above. All `<span className="checkbox-label">` layers could be removed since it's no longer within a popover.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+								</div>
+							) }
+							{ values.website_live && (
 								<CheckboxControl
 									label={
 										<span className="gla-pre-launch-checklist__checkbox_popover">
 											<span className="checkbox-label">
 												{ __(
-													'My store website is live.',
+													'My store is live and accessible to all users.',
```
eason9487: According to the description in Figma, this wording should remain in the second person after an item is confirmed, but here doesn't implement that logic. All other check items also have the same concerns.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+							{ ! values.website_live && (
+								<div className="gla-pre-launch-checklist__checkbox">
+									<CheckboxControl
+										{ ...getInputProps( 'website_live' ) }
+									></CheckboxControl>
+									<FlexBlock>
+										<Panel>
+											<PanelBody
+												title={ __(
+													'My store is live and accessible to all users',
```
eason9487: If I understood it right, this wording should be in the first person: "Confirm your store is live and accessible to all users". Some other check items also have the same concerns.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+												initialOpen={ true }
+												onToggle={ getPanelToggleHandler(
+													'website_live',
+													'pre-launch-checklist',
+													'onboarding-policy-check'
+												) }
+											>
+												<PanelRow>
+													{ __(
+														'We use a WordPress.com account to connect your site to the WooCommerce and Google servers. It ensures that requests (e.g. product feed, clicks, sales, etc) from your site are securely and correctly attributed to your store. It enables a connection to your self-hosted site, and provides a common authentication interface across disparate server configurations and architectures.',
```
eason9487: This copywriting seems to be incorrect.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+												<PanelRow>
+													{ __(
+														"Update your website to ensure that every webpage that collects a customer's personal information is processed through a secure SSL server. Any page on your website that collects any personal information from the user needs to be SSL protected.",
+														'google-listings-and-ads'
+													) }
+												</PanelRow>
+
+												<PanelRow>
+													{ __(
+														"Use a secure server: Make sure to use a secure processing server when processing customer's personal information (SSL-protected, with a valid SSL certificate). With SSL, your webpage URL will appear with https:// instead of http://,",
```
eason9487: There is an extra comma at the end of this description.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+														'google-listings-and-ads'
+													) }
+												</PanelRow>
+												<p>
+													<AppDocumentationLink
+														context="setup-mc-checklist"
+														linkId="check-checkout-process-secure"
+														type="external"
+														href="https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#payment-methods"
```
eason9487: I'd like to double-confirm if this URL should be https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#complete-checkout because the SSL relevant content is placed in the "I have a complete checkout process" section.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+									</FlexBlock>
+								</div>
+							) }
+							{ values.contact_info_visible && (
+								<CheckboxControl
+									label={
+										<span className="gla-pre-launch-checklist__checkbox_popover">
+											<span className="checkbox-label">
+												{ __(
+													"My store's phone number, email and/or address are visible on your website.",
```
eason9487: The copywriting is inconsistent with the one in Figma: "My store’s phone number, email and/or address are visible on my website"

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +71,384 @@ const PreLaunchChecklist = ( props ) => {
...
+												initialOpen={ true }
+												onToggle={ getPanelToggleHandler(
+													'contact_info_visible',
+													'pre-launch-checklist',
+													'onboarding-policy-check'
+												) }
+											>
+												<PanelRow>
+													{ __(
+														'Allow your customers to contact you for product inquiries by including contact information on your website (i,e, contact us form, business profile link, social media, email or phone number.',
```
eason9487: The closing bracket was missed at the end.

eason9487 (changes requested):
Hi @songchen2022, thanks for fixing some concerns.

I left some comments about issues, inconsistent strings, and missing implementations. And there are some other questions:

1. The sixth checklist item seems to be missed - "Confirm your store does not have placeholder images or text."
2. Does the auto-checked mechanism include the direction of changing the state from checked to unchecked? Because the current implementation will also auto-uncheck items when refreshing the page.
3. The copywriting of all checklist items doesn't have a period at the end, perhaps they can be changed to be consistent.

Please have a look and let me know if any of my comments are ambiguous.

---
> > 4. It's not a problem, but it would be a bit nicer to create a new component for the composition of CheckboxControl and Panel components. In this way, it could reduce some repetitive code in the PreLaunchChecklist component.

> For the comment 4, It's a good suggestions. Creating a new component for the composition of CheckboxControl and Panel will reduce some repetitive code but I found we need to pass around 10 parameters to the component which will look not very easily readable.

Summarizing some of the issues mentioned in the comments of this code review round, as well as a suggestion mentioned in the last time, here is a practical example.

Component for checklist items:
```js
const getPanelToggleHandler = ( id ) => ( isOpened ) => {
	recordEvent( 'pre-launch-checklist', {
		id,
		action: isOpened ? 'expand' : 'collapse',
	} );
};

function PreLaunchCheckItem( {
	formProps,
	fieldName,
	firstPersonTitle,
	secondPersonTitle,
	children,
} ) {
	const { getInputProps, setValue, values } = formProps;
	const checked = values[ fieldName ];
	const initialCheckedRef = useRef( checked );

	if ( checked ) {
		return (
			<CheckboxControl
				label={
					initialCheckedRef.current
						? firstPersonTitle
						: secondPersonTitle
				}
				{ ...getInputProps( fieldName ) }
				disabled
			/>
		);
	}

	return (
		<div className="gla-pre-launch-checklist__checkbox">
			<CheckboxControl { ...getInputProps( fieldName ) } />
			<Panel>
				<PanelBody
					title={ secondPersonTitle }
					initialOpen={ false }
					onToggle={ getPanelToggleHandler( fieldName ) }
				>
					<PanelRow>
						{ children }
						<Button
							isPrimary
							onClick={ () => setValue( fieldName, true ) }
						>
							{ __( 'Confirm', 'google-listings-and-ads' ) }
						</Button>
					</PanelRow>
				</PanelBody>
			</Panel>
		</div>
	);
}
```

Related component style:
```scss
.gla-pre-launch-checklist {
	.components-panel {
		flex: 1 1;

		&__row {
			display: flex;
			flex-direction: column;
			align-items: flex-start;
			gap: 1em;
		}
	}
}
```

Using with `<PreLaunchChecklist>`:
```js
return (
	// ... Omitted
	<Section.Card>
		<Section.Card.Body>
			<VerticalGapLayout size="large">
				<PreLaunchCheckItem
					formProps={ formProps }
					fieldName="website_live"
					firstPersonTitle={ __(
						'My store is live and accessible to all users',
						'google-listings-and-ads'
					) }
					secondPersonTitle={ __(
						'Confirm your store is live and accessible to all users',
						'google-listings-and-ads'
					) }
				>
					{ __(
						'Your Merchant Center account can be suspended if your store is not functional. Ensure that your domain and product pages are not leading to an under construction webpage, or an error page that displays a status code beginning with a 4 or 5 (such as a 405 error).',
						'google-listings-and-ads'
					) }
					<AppDocumentationLink
						context="setup-mc-checklist"
						linkId="check-website-is-live"
						type="external"
						href="https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#store-is-live"
					>
						{ __(
							'Learn more about common landing page issues and how to fix them',
							'google-listings-and-ads'
						) }
					</AppDocumentationLink>
				</PreLaunchCheckItem>
				<PreLaunchCheckItem
					formProps={ formProps }
					fieldName="payment_methods_visible"
					// ... Omitted
```

This example includes some simplifications on styling and mockup structure, removes some unnecessary parameters, and adds a missing implementation of the wording in the first and second person. Although it requires passing 4 parameters/props, I think it still bring a good readability and a higher degree of code reuse. Hope it could help with solving some issues.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Didn't add the figma link here. Sorry about that. Just added it in the description. This state should be true here. We will open the card if the check doesn't pass, otherwise it will be closed.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Thanks for pointing this out. However, we should keep as first person if the check is passed based on the PRD and [Mock link](https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=7450%3A120810). Please correct me if I'm wrong.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Updated. Thanks.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
eason9487: > Just added it in the description.

image

Thanks for adding the document. The added Figma link points to the same document I was referring to, and the first visual design shows all the unchecked items are collapsed. And that's why I inferred their initial state might be closed.

> This state should be true here. We will open the card if the check doesn't pass, otherwise it will be closed.

Sorry, I couldn't find the relevant document/spec in Figma, could you point me where it's?

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
eason9487: > However, we should keep as first person if the check is passed based on the PRD and [Mock link](https://www.figma.com/file/eQ9O4m2flzrcAiDMXkOW0m/Google-Listings-%26-Ads-v1.x?node-id=7450%3A120810). Please correct me if I'm wrong.

Yes, it will be the first person wording if the auto-check is passed. However, I meant the spec in Figma mentioned the wording remains in the second person when an item is confirmed (from unchecked status with the 2nd person wording). Either here or the relative processing code fragment of `! values.website_live`, there is no related implementation.

image

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Good catch. I matched the link based on the title which is not correct. Thanks!

js/src/setup-mc/setup-stepper/store-requirements/index.js
songchen2022: Thanks so much for your detailed suggestions ! Updated.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Sorry. I don't find the doc recording this. I confirmed this with PM before. I will confirm them again and bring the doc with you soon about this.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Thanks for your suggestions. Updated.

songchen2022:
> 4\. create a new component for the composition of CheckboxControl and Panel components

@eason9487 Thanks so much for taking time to review and giving very detailed suggestions! Really appreciate that!
1. 

> create a new component for the composition of CheckboxControl and Panel components

 Confirmed with internal team, we won't support this in the automatic check. Please refer this : https://drive.google.com/file/d/1NB4zWV63_3H0cStU0jxY_ASl6oVnptLt/view?usp=sharing&resourcekey=0-NhBPoF4Sdx97qlLghJlHxw
2. 

> Does the auto-checked mechanism include the direction of changing the state from checked to unchecked? Because the current implementation will also auto-uncheck items when refreshing the page.

I think this is as expected. Auto-check will overwrite the previous state when refreshing the page.

> 3. Create a new component for the composition of CheckboxControl and Panel components
Thanks for your great suggestions! Didn't know we could create children in the parameters before.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: [Confirmed](https://docs.google.com/document/d/1dcyZN6t0AgG36TCqV29luRLzrCI9ArS6VUHDsVoU5Yo/edit#bookmark=id.ey9n5s26n5n3) this with UX team, we will keep it close initially.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +51,151 @@ const PreLaunchChecklist = ( props ) => {
...
+									"My store's phone number, email and/or address are visible on my website",
+									'google-listings-and-ads'
+								) }
+								secondPersonTitle={ __(
+									"Confirm your store's phone number, email and/or address are visible on your website",
+									'google-listings-and-ads'
+								) }
+							>
+								{ __(
+									'Allow your customers to contact you for product inquiries by including contact information on your website (i,e, contact us form, business profile link, social media, email or phone number.',
```
eason9487: The closing parenthesis was still missed at the end.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +51,151 @@ const PreLaunchChecklist = ( props ) => {
...
+								<div></div>
+								{ __(
+									"Use a secure server: Make sure to use a secure processing server when processing customer's personal information (SSL-protected, with a valid SSL certificate). With SSL, your webpage URL will appear with https:// instead of http://",
+									'google-listings-and-ads'
+								) }
+								<AppDocumentationLink
+									context="setup-mc-checklist"
+									linkId="check-checkout-process-secure"
+									type="external"
+									href="https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#payment-methods"
```
eason9487: Looks like this link is still incorrect. Ref: https://github.com/woocommerce/google-listings-and-ads/pull/1612#discussion_r964680799

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
```diff
@@ -55,170 +51,151 @@ const PreLaunchChecklist = ( props ) => {
...
+							>
+								{ __(
+									'Ensure that all customers are able to complete the full checkout process on your site with an eligible payment method. Include a confirmation of the purchase after completion of the checkout process. ',
+									'google-listings-and-ads'
+								) }
+								<AppDocumentationLink
+									context="setup-mc-checklist"
+									linkId="check-payment-methods-visible"
+									type="external"
+									href="https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#payment-methods"
```
eason9487: This link was changed from https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#complete-checkout to https://woocommerce.com/document/google-listings-and-ads/compliance-policy/#payment-methods in 18e9ff25e8c91f23f096d163ebe17e05e8275447. However, it seems not to be an expected change.

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.scss
```diff
@@ -8,4 +8,9 @@
 			vertical-align: text-bottom;
 		}
 	}
+
+	.gla-pre-launch-checklist__checkbox {
+		display: flex;
+		justify-content: stretch;
+	}
```
eason9487: It would be great if this style is placed along with the corresponding component's file - js/src/components/pre-launch-check-item/index.scss

js/src/components/pre-launch-check-item/index.js
```diff
@@ -0,0 +1,74 @@
...
+		);
+	}
+
+	return (
+		<div className="gla-pre-launch-checklist__checkbox">
+			<CheckboxControl { ...getInputProps( fieldName ) } />
+			<Panel>
+				<PanelBody
+					title={ secondPersonTitle }
+					initialOpen={ true }
```
eason9487: Thanks for confirming the initial state in https://github.com/woocommerce/google-listings-and-ads/pull/1612#discussion_r970181132. Adding this PR comment for following up that the `initialOpen` value should be changed to `false`.

eason9487 (commented):
Hi @songchen2022, thanks for the adjustments. Left a few comments on things that seem to be missed from the last code review, and some are new suggestions. In addition, it would be nicer if [these styles](https://github.com/songchen2022/google-listings-and-ads/blob/6a05b0e86b03c622f81d6ad8b42b7af86ddb5891/js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.scss#L2-L10) could be removed as well because they are no longer in use.

js/src/components/pre-launch-check-item/index.js
songchen2022: Just confirmed this with this afternoon and haven't pushed at that time. Updated now. Thanks!

js/src/setup-mc/setup-stepper/store-requirements/pre-launch-checklist/index.js
songchen2022: Thanks for the offline suggestions. Updated.

eason9487 (approved):
@songchen2022, thanks for resolving all my concerns. LGTM.

📝 There is an extra change in this [commit](https://github.com/woocommerce/google-listings-and-ads/pull/1612/commits/b57f38b8b1330844f16beb8fb87ce1377f73d582#diff-25e8174bfde194d8659ec4930eb6771f4ac6333268b807db41cc35505c39aedeL187-L197). Since that item doesn't have a "Learn about" document link on Figma, I guess it's an expected removal.

Issue #1610, opened 2022-07-28
Streamlined Free Listings plus Paid Ads for the onboarding flow


eason9487:
## Objective

- An onboarding flow for the integration offers merchants a choice of selecting Free Listings plus Paid Ads.
- The ability for merchants to:
   - preview the ads or campaigns they are about to launch,
   - know how many products will be synced to Google Merchant Center,
   - and set up billing for their Google Ads account via a popup.

💡 We will create a shared branch `feature/1610-streamlined-onboarding`. All relevant developments should be branched from it and merged back into it. And open a PR to merge the shared branch to develop once the overall features get ready to release.

## Terminology

- FL: Free Listings

## Technical - Frontend

The UI changes are mainly applied to the onboarding flow, and some shared components used in the Edit free listings page and Set up paid campaign page might be changed respectively.

Most requirements are listed in pcTzPl-Kc-p2#the-flow-breakdown and the related Figma document. This list only outlines other changes besides the p2 post.

### 📌 Adjust the setting sections in the onboarding steps

- [x] Merge the original step 3 into step 2 - #1616
- [x] Remove the auto-fill mechanism of shipping rates - #1616
- [x] Add another submit button to step 4 for skipping paid ads creation - #1619

### 📌 Update other UI changes

Refer to **Landing Page**, **Step 1: Set up accounts**, **Step 2: Configure product listings**, **Step 3: Confirm store requirements**, and **Submission Success** in pcTzPl-Kc-p2#the-flow-breakdown

- [x] Logo icons in some account cards - #1618
- [x] Layouts and copywriting on the landing page - #1618
- [x] Layouts and copywriting of steps 1~3 in the onboarding flow - #1618
- [x] Add [CSS disclaimer](https://github.com/woocommerce/google-listings-and-ads/issues/1610#issuecomment-1246442668) to step 1 - #1679 
- [x] FAQ card in step 4 - #1679 
- [x] Submission success modal in the product feed page - #1679 

### 📌 Product feed status section

Refer to **Step 4.1: Setting up Paid Ads** in pcTzPl-Kc-p2#the-flow-breakdown

- [x] Implement this section UI - #1638 

### 📌 Boost product listings section

Refer to **Card 2 in Step 4: Complete your campaign with paid ads** in pcTzPl-Kc-p2#the-flow-breakdown

- [x] Section UI and the interactions of "skip" and "continue" buttons - #1649 
- [x] Dynamically composite the ad previews - #1650

[...]

Issue #232, opened 2021-02-22
Setup MC: changing target audience countries in Step 2 after setting up shipping rates and shipping times in Step 3


ecgan:
This is a follow up to https://github.com/woocommerce/google-listings-and-ads/pull/212#issuecomment-781858035.

When users proceeded through Step 2 and Step 3 in the Setup MC flow, and then go back to Step 2 and change the target audience countries, what should happen to the shipping rates and shipping times that are already entered in Step 3?

I supposed for the target audience countries that are being removed, we can simply remove the associated shipping rates and shipping times.

Do we want to show some kind of prompt to inform users?

Tagging @j111q  @jconroy for your inputs.

PR #927, opened 2021-08-02
Move the pre-launch checklist from step 3 to 4 for the MC setup

eason9487:
### Changes proposed in this Pull Request:

Partial of #863 and based on #926

- Move `js/src/setup-mc/setup-stepper/setup-free-listings/pre-launch-checklist` in the folder of `js/src/setup-mc/setup-stepper/store-requirements/`
- Separate the pre-launch checks into another validation function
- Move the pre-launch checklist from step 3 to 4 for the MC setup and relevant checking and data updating

### Screenshots:

https://user-images.githubusercontent.com/17420811/127865542-96a7a46b-c524-4cd7-be06-a9e25f949e00.mp4

### Detailed test instructions:

1. Go to the MC setup and proceed to step 3
2. The pre-launch checklist should no longer be in step 3
3. Change some product listings settings to see if the disable state of "Continue" button works correctly
4. Proceed to step 4 and the pre-launch checklist should be at the bottom
5. Fill in the contact information
6. Tick or untick each pre-launch checklist to see if the disable state of "Complete setup" button works correctly

PR #893, opened 2021-07-21
Shows a Jetpack connected success text when viewing MC setup and settings pages as a non jetpack owner

eason9487:
### Changes proposed in this Pull Request:

Closes #624 

- Add a util function to get the connected jetpack info
- Show jetpack connected info according to whether the current user is the jetpack owner on the MC setup and Setting pages

#### MC setup

image

#### Settings

image

### Detailed test instructions:

1. Log in as the jetpack owner user
2. Go to the MC setup and settings pages to see if it shows the owner email
3. Create a new admin user and log in as that user
4. Go to the MC setup and settings pages to see if it shows "Successfully connected through Jetpack"

### Changelog entry

> Fix - Shows a Jetpack connected success text instead of blank when viewing the onboarding setup and the settings pages as a non jetpack owner account.

PR #1616, opened 2022-08-04
Free Listings + Paid Ads: Combine the audience and shipping steps

### LINKED ISSUE #274: Refactor useGetRemainingCountryCodes
This is a follow up from https://github.com/woocommerce/google-listings-and-ads/pull/264#discussion_r585243370.

Currently we have two similar copies of `useGetRemainingCountryCodes` hook, one for shipping rate and one for shipping time. There is just a bit of difference between both.

This issue is aimed to refactor the code to reduce the duplication.

### LINKED ISSUE #288: Better implementation of Shipping Rates and Shipping Times component for MC Account Edit Settings and Setup MC Step 3
This is related to https://github.com/woocommerce/google-listings-and-ads/pull/226#issuecomment-790215574 . Last night @tomalec and I had a discussion on the current shipping rates and shipping times auto save implementation in Setup MC onboarding flow and its impact on @tomalec's work on MC account edit settings page. I'm opening this issue to document the ideas in my head, to serve as a guide for @tomalec's work and for future code refactoring changes in Setup MC onboarding flow.

## Current implementation

In Setup MC Step 3 page, we have a few places that calls API that saves data: 

- `js/src/setup-mc/listings-stepper/setup-free-listings/form-content.js` which uses `useAutoSaveSettingsEffect` to save automatically save settings data. This does not include shipping rate and shipping time. (This is fine and not a problem.)
- `js/src/setup-mc/listings-stepper/setup-free-listings/shipping-time/shipping-time-setup/countries-time-input-form/index.js` which will call upsertShippingTime in a debounced manner when users modify the shipping time in the inline edit input.
- `js/src/setup-mc/listings-stepper/setup-free-listings/shipping-time/shipping-time-setup/countries-time-input/edit-time-button/edit-time-modal/index.js` which will call upsert and delete shipping time when users edit or delete shipping time in the pop-up modal.
- `js/src/setup-mc/listings-stepper/setup-free-listings/shipping-time/shipping-time-setup/add-time-button/add-time-modal/index.js` which will call upsert shipping time when users add new shipping time.
- `js/src/setup-mc/listings-stepper/setup-free-listings/shipping-rate/shipping-rate-setup/countries-price-input-form/index.js`
- `js/src/setup-mc/listings-stepper/setup-free-listings/shipping-rate/shipping-rate-setup/countries-price-input/edit-rate-button/edit-rate-modal/index.js`
- `js/src/setup-mc/listings-stepper/setup-free-listings/shipping-rate/shipping-rate-setup/add-rate-button/add-rate-modal/index.js`

For MC account edit settings page, we do not want to have auto save, instead we only want to save upon clicking save button. The last six files above present a problem and cannot be easily reused in MC account edit settings page.

### For MC account edit settings

We duplicate the last six files above and remove the auto save logic (essentially making the component dumb). We should put the initial values for the shipping rates and shipping times into Form component (`js/src/setup-mc/listings-stepper/setup-free-listings/index.js`) and pass it down to the child components. When shipping rates and shipping times change, the changes should be propagated up and stored in the Form component. (Note that the above was how I implemented it in the early beginning, but along the way I changed it when I did the integration with shipping rates and shipping times API.)

Because we duplicate the files, this ensures the Setup MC onboarding flow do not break.

When users click the Save button, we can save the settings, shipping rates and shipping times in the `Form`'s `onSubmitCallback` event handler.

### For Setup MC Step 3 (future work)

With the above duplicated dumb components, we should then use them to replace the smart components in the Setup MC Step 3 page. The smart components should be deleted.

To implement the shipping rate and shipping time auto save feature, we can use similar technique like `useAutoSaveSettingsEffect`. Since the shipping rates and shipping times are stored in Form value and passed down to FormContent, we can get the values there and then come up with `useAutoSaveShippingRatesEffect` and `useAutoSaveShippingTimesEffect`.

### LINKED ISSUE #1078: Shipping values flash during the onboarding setup
### Describe the bug:

@layoutd found this issue.

When entering shipping rates/times, sometimes the input values will flash between empty, old value, and new value.

### Expected behavior:

Apart from the currency/number formatting that might change the input value, it should not flash the old/empty value on the `<input>` field after moving out the input cursor.

#### Appear empty value

Screen Capture on 2021-11-02 at 11-10-04

#### Appear old value

Kapture 2021-11-05 at 16 29 48

### Steps to reproduce:

1. Go to step 2 of onboarding setup page.
1. Select "All countries" for audience location.
1. Move to step 3
1. Open browser DevTool and throttle network with "Slow 3G"
    image
1. In step 3, enter values to shipping rates and times quickly, and maybe keep entering values round trip. Then sometimes the flash will happen.

### Additional details:

Eventually the saved values are correct, but it might confuse users.

### LINKED ISSUE #1335: Replace Form-related "auto" effects with Form `onChange` handler
## User story

- This is related to https://github.com/woocommerce/google-listings-and-ads/pull/1313#discussion_r828739436.

Currently in our code base, we have four "auto" effects that are related to Form:

- `useAutoSaveTargetAudienceEffect`
- `useAutoClearShippingEffect`
- `useAutoSaveSettingsEffect`
- `useAutoSaveShippingRatesEffect`

The above four effects are used in the Setup MC flow. They were created in the first place because at that time it was not known and not documented that the WooCommerce [Form](https://github.com/woocommerce/woocommerce-admin/tree/main/packages/components/src/form) component has an `onChange` prop (previously named `onChangeCallback`).

Now that we know about Form's `onChange` prop, we should make use of it to replace the above auto effects. 

Note: there is also an "auto" effect called `useAutoCheckBillingStatusEffect` which is not relevant to Form and not related to this issue.

## Technical

The four effects uses debounced callbacks. For most of them, it probably makes sense, because users might keep on changing values in the form (e.g. text input, radio options, checkboxes), and we don't want to call the APIs every time the values change.

For `useAutoSaveShippingRatesEffect` however, because the shipping rates value in the form would only change when the input control fires the `onBlur` event, we can just make the API call in Form's `onChange` without debouncing. (See https://github.com/woocommerce/google-listings-and-ads/pull/1313#discussion_r828739436 and https://github.com/woocommerce/google-listings-and-ads/pull/1267#discussion_r820441914.)

eason9487:

It's a part of **📌 Adjust the setting sections in the onboarding steps** in #1610. Mainly focus on combining the audience and shipping steps for the onboarding flow and the editing free listings page. Most visual and wording changes are not included.

- Move the `SetupFreeListings` component of editing free listings to the shared directory.
- Change the shared `ChooseAudience` to `ChooseAudienceSection` and merge it into `SetupFreeListings` with corresponding validations for the editing free listings.
- Replace `SetupFreeListings` with the shared one for the onboarding flow.
- Remove unused files and an npm package `use-debounce`.

It also handles some issues:

- There are no `useGetRemainingCountryCodes` hooks after this PR.
   - #274
- Components of shipping times are changed to use the shared ones. Each group of shipping rates and shipping times components is no longer duplicated.
   - #288
- The last shipping values flash issue only happens on the inputs of shipping time during onboarding setup. With the replacement of using shared shipping time components, the form state management was isolated from wp-store.
   - #1078
- After replacing `ChooseAudience` and `SetupFreeListings` with the shared components, the auto-saving mechanism is implemented with `on*Change` callbacks instead. All auto-saving hooks and debouncing are no longer in need.
   - #1335

#### 📷 Onboarding flow

wp test_wp-admin_admin php_page=wc-admin path=%2Fgoogle%2Fsetup-mc google-mc=connected

#### 📷 Edit free listings

wp test_wp-admin_admin php_page=wc-admin path=%2Fgoogle%2Fproduct-feed google-mc=connected guide=submission-success (2)

#### Onboarding flow

1. Disconnect all accounts or run these SQLs to delete related GLA settings
   ```sql
   DELETE FROM wp_options WHERE option_name IN ('gla_mc_setup_completed_at', 'gla_target_audience', 'gla_merchant_center');
   DELETE FROM wp_gla_shipping_rates;
   DELETE FROM wp_gla_shipping_times;
   ```
2. Start the onboarding setup and check if the combined step for audience and shipping works correctly.
   - Auto-saving mechanism for audience, MC settings, shipping rates, and shipping times
   - All possible ways to add/edit shipping times
   - Switch the location options of audience to see if **all** countries and **selected** countries options affect the corresponding shipping sections correctly.
3. Change the audience countries to see if the removed countries are removed from shipping sections as well.
   - 💡 The implementation is not the same as before. Please refer to the originally expected behavior in https://github.com/woocommerce/google-listings-and-ads/issues/232#issuecomment-785687459
4. Edit settings to check if the disabled status of the "Continue" button is correct.
5. Continue or switch between steps to view whether it works as expected.
6. Refer to the replicate steps in #1078 to check if the value flash problem in shipping time fields is fixed.
   - 💡 The problem in shipping rate fields has been fixed in another PR before.

#### Edit free listings

1. Go to the Edit Free Listings page.
2. It should be changed to a one-pager without steps.
3. Simiar to the tests in the onboarding flow except for the auto-saving mechanism, operate the combined audience and shipping to see if they work well.
4. Edit some values to see if the unsaved data prompt pops up when navigating away.
5. Edit the above values back to see if there is no prompt when navigating away.

### Additional details:

- The auto-fill feature of flat shipping rates in the onboarding flow is removed from the frontend side. Ref: p1659494048432789-slack-C01E9LB4X61
- Not sure why but `mc/target_audience` will respond with an 400 error if saving with empty countries. From my memory, it should be ok before because the UI allows merchants to unselect all. 🤔 

### Changelog entry

> Add - Combine the audience and shipping steps for the onboarding flow and the editing free listings page.
> Fix - Shipping time values flash during the onboarding setup.


js/src/components/free-listings/configure-product-listings/__snapshots__/checkErrors.test.js.snap
```diff
@@ -1,5 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`checkErrors Audience When the audience countries array is empty and the value of audience location option is 'selected', should not pass 1`] = `"Please select at least one country."`;
```
puntope, line 3: 💅  Out of this PR and opinionated

Maybe to consider migration to unit testing?

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
```diff
@@ -25,6 +33,49 @@ import stepNameKeyMap from './stepNameKeyMap';
...
+	}, [ targetAudience, suggestedAudience, saveTargetAudience ] );
+
+	// Auto-save the default values for shipping options to fall back with the original implementation.
+	// Ref: https://github.com/woocommerce/google-listings-and-ads/blob/2.0.2/js/src/setup-mc/setup-stepper/setup-free-listings/form-content.js#L33
+	useEffect( () => {
+		if ( settings?.shipping_rate === null ) {
+			saveSettings( {
+				...settings,
+				shipping_rate: 'automatic',
+				shipping_time: 'flat',
```
puntope, line 74: ❓ Not a suggestion, just curious. Why we are not adding the default values inside data store DEFAULT_STATE ?

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
```diff
@@ -58,6 +100,15 @@ const SavedSetupStepper = ( { savedStep, onRefetchSavedStep = () => {} } ) => {
 		}
 	};
 
+	function handleFormChange( errorMessage, newValue ) {
+		this( newValue ).catch( () => createNotice( 'error', errorMessage ) );
```
puntope, line 115: 💅  it was a bit confusing to understand what `this` was doing here. I saw after  in the code that we bind  the action  to be called in `SetupFreeListings`. 

However, would be nice to add a JSDOC in this function explaining whats going on.

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
puntope: I saw we use null values as checkers for other stuff... probably because of that

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
```diff
@@ -58,6 +100,15 @@ const SavedSetupStepper = ( { savedStep, onRefetchSavedStep = () => {} } ) => {
...
+	}
+
+	const initShippingRates = hasResolvedShippingRates ? shippingRates : null;
+	const initShippingTimes = hasResolvedShippingTimes ? shippingTimes : null;
+	const initTargetAudience = targetAudience?.location ? targetAudience : null;
+	const initSettings = settings?.shipping_rate ? settings : null;
```
puntope, line 121: ❓ 💅  I did some degugging. Settings is `null` by default until populate the object values... so maybe we can just pass `settings` directly ????

puntope: Same for `initTargetAudience ` I guess

puntope (commented):
- I checked the behaviour. Tested locally and I didn't find any error. All the combos worked fine for me. 
- I checked also the code. Since there is a lot of files deleted and moved around mixed with changes in those files. It was a bit hard to differentiate what was added/changed and what was just a copy paste for the deleted files. However everything looks good for me... left a few comments and minor questions

puntope:
> Not sure why but mc/target_audience will respond with an 400 error if saving with empty countries. From my memory, it should be ok before because the UI allows merchants to unselect all. 🤔

I was unable to save with empty value in the audience selector. 🤔 I guess do you mean it shows an error when the autosave runs when the selector is empty?

js/src/components/free-listings/configure-product-listings/__snapshots__/checkErrors.test.js.snap
eason9487: Do you mean replacing `expect( errors.countries ).toBe( 'Please select at least one country.' );` with `expect( errors.countries ).toMatchSnapshot();` ? If so, some related discussions can be found in https://github.com/woocommerce/google-listings-and-ads/pull/893#discussion_r674040117 and https://github.com/woocommerce/google-listings-and-ads/pull/927#discussion_r681107540.

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
eason9487: The intent here is to trigger API to save the form's default values. It would be easier and more suitable to handle the intent here.

In addition, moving the default values of form layer to wp-data store layer will also bring different abstraction of data/states isolation. Since `isResolving` and `hasFinishedResolution` are not enough to cover all statuses of a state in wp-store (see issue #438.), and considering the "statuses" of states in wp-data haven't yet had a consistent definition/convention, for now, the default value in a wp-data state is more like the status that tells the state is not ready to be used. And after receiving the data from API, the state will be changed to that data, which will also change the status of that state to "the status of the state is ready, and the data is the state itself".

IMO, before a consistent definition/convention of the statuses, states, and data in wp-data is created, it's probably hard to say how to set a default value would be a good practice.

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
eason9487: It's waiting for completing the above processing to auto-save the default values for `settings`. Passing `settings` directly will lead to the `<Form>` receiving `initialValues` without those default values of `settings` when the `GET mc/settings` request is the last one to respond.

The problem can be simulated by adding delay to [fetchSettings](https://github.com/woocommerce/google-listings-and-ads/blob/43a2f76deda2218b5943be595bb6a1d0b1c5ee47/js/src/data/actions.js#L290):

```js
import { __unstableAwaitPromise } from '@wordpress/data-controls';

export function* fetchSettings() {
	yield __unstableAwaitPromise(
		new Promise( ( resolve ) => setTimeout( resolve, 3000 ) )
	);

	// ... omitted
}
```

using `settings` directly:

```js
<SetupFreeListings
	targetAudience={ initTargetAudience }
	settings={ settings }
	// ... omitted
>
```

deleting saved settings:

```sql
DELETE FROM wp_options WHERE option_name = 'gla_merchant_center';
```

and then refresh page to view the initial shipping options in step 2:

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/1616/commits/d8ecc0824ce15ca8cbfcae5e49c0896ceb517f1b.

eason9487:
Thanks for the review, @puntope

> * I checked also the code. Since there is a lot of files deleted and moved around mixed with changes in those files. It was a bit hard to differentiate what was added/changed and what was just a copy paste for the deleted files.

Sorry for the inconvenience, I was trying to separate this as smaller PRs but the second one would only be including file deletions. So I ended up choosing to use deal with it altogether.

There are no deleted file changes that were caused by copy-and-paste. All deleted files were actually deleted (20e758fda9ddbfb22e5d2037177e9d69deaad056). And two deleted-and-added files are completely different:
- `js/src/components/free-listings/choose-audience-section/index.js`
- `js/src/components/free-listings/choose-audience-section/choose-audience-section.scss`

> I guess do you mean it shows an error when the autosave runs when the selector is empty?

Yes. UI allows the operation to clear the value in the audience selector but then it causes an error. It may confuse merchants. I will check it later while working on this project and open a follow-up issue if needed.

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
puntope: Gotcha thanks.

puntope:
> Yes. UI allows the operation to clear the value in the audience selector but then it causes an error. It may confuse merchants. I will check it later while working on this project and open a follow-up issue if needed.

I agree

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
puntope: Now is perfect! thanks

puntope (approved):

PR #1622, opened 2022-08-09
Add unit tests for PolicyComplianceCheck

songchen2022:

This PR is based on https://github.com/woocommerce/google-listings-and-ads/pull/1563
This PR adds the unit tests for [PolicyComplianceCheck.php](https://github.com/woocommerce/google-listings-and-ads/pull/1611/files/ce444bd47a3832c8fce0301ed3eb621711bd9b40#diff-adf729a3445cfe44b0dbddce26a7ff32b7a350fcf46367c71222174945e9a48e)

### Screenshots:

image

### DIFF
--- a/src/MerchantCenter/PolicyComplianceCheck.php
+++ b/src/MerchantCenter/PolicyComplianceCheck.php
@@ -62,8 +62,8 @@ class PolicyComplianceCheck implements Service {
 		$all_allowed_countries = $this->wc->get_allowed_countries();
 		$target_countries      = $this->target_audience->get_target_countries();
 
-		foreach ( $target_countries as $country ) {
-			if ( ! array_key_exists( $country, $all_allowed_countries ) ) {
+		foreach ( $target_countries as $country_key => $country ) {
+			if ( ! array_key_exists( $country_key, $all_allowed_countries ) ) {
 				return false;
 			}
 		}
--- a/src/MerchantCenter/PolicyComplianceCheck.php
+++ b/src/MerchantCenter/PolicyComplianceCheck.php
@@ -62,8 +62,8 @@ class PolicyComplianceCheck implements Service {
 		$all_allowed_countries = $this->wc->get_allowed_countries();
 		$target_countries      = $this->target_audience->get_target_countries();
 
-		foreach ( $target_countries as $country_key => $country ) {
-			if ( ! array_key_exists( $country_key, $all_allowed_countries ) ) {
+		foreach ( $target_countries as $country ) {
+			if ( ! array_key_exists( $country, $all_allowed_countries ) ) {
 				return false;
 			}
 		}
--- a/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
+++ b/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
@@ -97,6 +97,7 @@ class PolicyComplianceCheckTest extends WPRequestUnitTest {
 	}
 
 	public function test_not_has_page_not_found_error() {
+		$this->mock_wp_request( 'http://example.org', '' );
 		$this->assertFalse( $this->policy_compliance_check->has_page_not_found_error() );
 	}
 
@@ -106,6 +107,7 @@ class PolicyComplianceCheckTest extends WPRequestUnitTest {
 	}
 
 	public function test_not_has_redirects() {
+		$this->mock_wp_request( 'http://example.org', '' );
 		$this->assertFalse( $this->policy_compliance_check->has_redirects() );
 	}
 
@@ -128,11 +130,29 @@ class PolicyComplianceCheckTest extends WPRequestUnitTest {
 	}
 
 	public function test_with_store_ssl() {
-		define( 'WP_HOME', 'https://example.org' );
+		$override_url = function () {
+			return 'https://secure.url';
+		};
+		add_filter( 'woocommerce_gla_site_url', $override_url );
 		$this->assertTrue( $this->policy_compliance_check->get_is_store_ssl() );
+		remove_filter( 'woocommerce_gla_site_url', $override_url );
 	}
 
 	public function test_not_has_refund_page() {
 		$this->assertFalse( $this->policy_compliance_check->has_refund_return_policy_page() );
+	}
+
+	public function test_has_refund_page() {
+		$page_id = wp_insert_post(
+			[
+				'post_title'  => 'Refund and Returns',
+				'post_name'   => 'refund_returns',
+				'post_type'   => 'page',
+				'post_status' => 'publish',
+			]
+		);
+
+		$this->assertTrue( $this->policy_compliance_check->has_refund_return_policy_page() );
+		wp_delete_post( $page_id, true );
 	}
 }
--- a/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
+++ b/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
@@ -84,6 +84,11 @@ class PolicyComplianceCheckTest extends WPRequestUnitTest {
 		$this->assertFalse($this->policy_compliance_check->has_page_not_found_error());
 	}
 
+	public function test_has_page_not_found_error2() {
+		$this->mock_wp_request( 'http://example.org', '', 404 );
+		$this->assertTrue($this->policy_compliance_check->has_page_not_found_error());
+	}
+
 	public function test_has_redirects() {
 		$this->assertFalse($this->policy_compliance_check->has_redirects());
 	}
--- a/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
+++ b/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
@@ -88,6 +88,11 @@ class PolicyComplianceCheckTest extends WPRequestUnitTest {
 		$this->assertFalse($this->policy_compliance_check->has_redirects());
 	}
 
+	public function test_has_redirects2() {
+		$this->mock_wp_request( 'http://example.org', '', 301 );
+		$this->assertTrue($this->policy_compliance_check->has_redirects());
+	}
+
 	public function test_has_restrictions() {
 		$this->assertFalse($this->policy_compliance_check->has_restriction());
 	}
--- a/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
+++ b/tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
@@ -81,23 +81,19 @@ class PolicyComplianceCheckTest extends WPRequestUnitTest {
 	}
 
 	public function test_not_has_page_not_found_error() {
-		define( 'WP_HOME', 'http://example.org' );
 		$this->assertFalse($this->policy_compliance_check->has_page_not_found_error());
 	}
 
 	public function test_has_page_not_found_error() {
-		define( 'WP_HOME', 'http://example.org' );
 		$this->mock_wp_request( 'http://example.org', '', 404 );
 		$this->assertTrue($this->policy_compliance_check->has_page_not_found_error());
 	}
 
 	public function test_not_has_redirects() {
-		define( 'WP_HOME', 'http://example.org' );
 		$this->assertFalse($this->policy_compliance_check->has_redirects());
 	}
 
 	public function test_has_redirects() {
-		define( 'WP_HOME', 'http://example.org' );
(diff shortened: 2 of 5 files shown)


src/MerchantCenter/PolicyComplianceCheck.php
```diff
@@ -60,8 +60,8 @@ public function is_accessible(): bool {
 		$all_allowed_countries = $this->wc->get_allowed_countries();
 		$target_countries      = $this->target_audience->get_target_countries();
 
-		foreach ( $target_countries as $country ) {
-			if ( ! array_key_exists( $country, $all_allowed_countries ) ) {
+		foreach ( $target_countries as $country_key => $country ) {
+			if ( ! array_key_exists( $country_key, $all_allowed_countries ) ) {
```
mikkamp: The target countries function returns a list of country codes, it is not the same as the array of allowed countries which is a list of countries indexed by country code. So this change should be reverted, as it was working correctly.

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+
+
+	public function test_website_accessible() {
+		$this->wc->expects( $this->any() )
+					   ->method( 'get_allowed_countries' )
+					   ->willReturn( ["AU" => "Australia", "AT" => "Austria", "CA" => "Canada", "US" => "United States"] );
+		$this->target_audience->expects( $this->any() )
+					   ->method( 'get_target_countries' )
+					   ->willReturn( ["AU" => "Australia", "US" => "United States"] );
```
mikkamp: As mentioned in the previous comment, the target audience should be an array of country codes like:

```php
[ 'AU', 'US' ]
```

Also generally we only use double quotes when we need to add variables in the string. If it's just a plain text string it's recommended to use single quotes.

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+		$this->assertEquals($this->policy_compliance_check->is_accessible(), true);
```
mikkamp: It's a slight bit more readable if we use `assertTrue`, same for the other comparisons where we can use `assertFalse`.

```suggestion
		$this->assertTrue( $this->policy_compliance_check->is_accessible() );
```

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,82 @@
...
+		$this->assertEquals($this->policy_compliance_check->is_accessible(), false);
+	}
+
+
+	public function test_payment_gateways() {
+		$this->wc->expects( $this->any() )
+					   ->method( 'get_available_payment_gateways' )
+					   ->willReturn( ["PayPal", "Stripe"] );
+
+		$this->assertEquals($this->policy_compliance_check->has_payment_gateways(), true);
```
mikkamp: The function `get_available_payment_gateways` returns a different array for the payment gateways. In this case since we are only testing if the array is empty it doesn't matter much, but I'd suggest to populate the response with the correct structure. This could be done with something like this:

```suggestion
		$payment_gateways = [
			'stripe' => $this->createMock( WC_Payment_Gateway::class ),
			'paypal' => $this->createMock( WC_Payment_Gateway::class ),
		];

		$this->wc->expects( $this->any() )
			->method( 'get_available_payment_gateways' )
			->willReturn( $payment_gateways );

		$this->assertTrue( $this->policy_compliance_check->has_payment_gateways() );
```

> **Note**
An import will be needed at the top of the file `use WC_Payment_Gateway;`

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,82 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\GoogleHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
```
mikkamp: This import is not used

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,82 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
```
mikkamp: This import is not used

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class PolicyComplianceCheckTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\MerchantCenter
+ *
+ * @property  MockObject|Merchant $merchant
+ * @property  MockObject|Settings $google_settings
```
mikkamp: We create three different mock objects as properties in this class: `$wc, $google_helper, $target_audience`

mikkamp (commented):
Thanks for adding the unit test for the PolicyComplianceCheck class. I've left some comments on various things I've noticed when running the tests.
I also noticed that there is quite a mix of alignment using spaces and tabs. Can those be adjusted so it sticks to the [WordPress coding standard guidelines](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation).

mikkamp (commented):
Thanks for all the changes. The code for the tests look good, although it tests only two functions. The PolicyComplianceCheck has 7 checks now, are the additional checks going to be added in this PR or do you have a series of followup PR's ready for this?

songchen2022:
@mikkamp Thanks for taking time to review. How do you think what the best way to test other checks which are using functions like wp_remote_get or wc_get_products is? Thanks!

mikkamp:
For `wc_get_products` you can create a single test product with `WC_Helper_Product::create_simple_product` like is done [here](https://github.com/woocommerce/google-listings-and-ads/blob/2.1.1/tests/Unit/Product/ProductFactoryTest.php#L31).

For `wp_get_requests` we don't have any existing examples, but I'd suggest to create a `WPRequestUnitTest` class similar to `RESTControllerUnitTest`. There are some suggestions [here](https://wordpress.stackexchange.com/questions/133414/how-do-i-mock-http-requests-for-phpunit) that show how to mock those requests. The setUp function can be used to add the filter and the tearDown can be used to remove the filter, which checks if it has a mocked response for the matching URL. Then a `mock_request` function can take a URL and response, and add it to the list of mocked responses.

I can write up some example code if you like.

mikkamp:
Here is an example [WPRequestUnitTest](https://gist.github.com/mikkamp/0b4f7bf6f81fa79c9f4a536af9ad89b6) class.

The `PolicyComplianceCheckTest` can extend `WPRequestUnitTest` instead of just `UnitTest` and then to mockup a request you can just call the helper function:

```php
$this->mock_wp_request( 'https://foo.bar', 'foobar' );
```

If you want to mock up an error/redirect response it can be used like:

```php
$this->mock_wp_request( 'https://foo.bar', '', 301 );
```

> **Note**
I didn't fully test the class but it should give an idea what's needed to mock the requests. For example it can also be used to mock a robots.txt file and passing the URL in the first argument and the contents of an example robots.txt file in the second parameter.

songchen2022:
@mikkamp could you help me sync this PR with the latest develop branch?  Thanks!

mikkamp:
@songchen2022 I've merged the latest changes from develop into the base branch: https://github.com/woocommerce/google-listings-and-ads/tree/feature/policy-compliance-checks

You can rebase your branch onto this base branch. For the unit tests I'd recommend to revert the changes to composer.json and composer.lock as version 9.5 is not compatible with older versions of WordPress. If the E2E tests fail you can ignore those as we are still working on making those a bit more stable, as they fail sometimes without any changes to the related code.

Also as mentioned the unit tests now have required PHPCS rules so you might have to fix some of the spacing in the tests to comply to those rules. Let me know if there is anything else I can help with and when this PR is ready for another review.

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,138 @@
...
+	public function test_empty_payment_gateways() {
+		$this->wc->expects( $this->any() )
+					   ->method( 'get_available_payment_gateways' )
+					   ->willReturn( [] );
+
+		$this->assertFalse( $this->policy_compliance_check->has_payment_gateways() );
+	}
+
+	public function test_not_has_page_not_found_error() {
+		$this->assertFalse( $this->policy_compliance_check->has_page_not_found_error() );
```
mikkamp, line 101: I'm not sure what's setup when these get run in GitHub actions. However when I run the tests locally this fails because it tries to get `http://example.org` which does not return a 200 response.

Can we add mocking of a successful response:
```php
$this->mock_wp_request( 'http://example.org', '' );
```

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	public function test_has_page_not_found_error() {
+		$this->mock_wp_request( 'http://example.org', '', 404 );
+		$this->assertTrue( $this->policy_compliance_check->has_page_not_found_error() );
+	}
+
+	public function test_not_has_redirects() {
+		$this->assertFalse( $this->policy_compliance_check->has_redirects() );
```
mikkamp, line 111: Also for this test, it doesn't fail. But the `wp_remote_get` returns a WP_Error which means the $code will be 0. So can we also mock a valid 200 response here.

```php
$this->mock_wp_request( 'http://example.org', '' );
```

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,138 @@
...
+		$this->assertTrue( $this->policy_compliance_check->has_restriction() );
+	}
+
+	public function test_without_store_ssl() {
+		$this->assertFalse( $this->policy_compliance_check->get_is_store_ssl() );
+	}
+
+	public function test_with_store_ssl() {
+		define( 'WP_HOME', 'https://example.org' );
+		$this->assertTrue( $this->policy_compliance_check->get_is_store_ssl() );
```
mikkamp: I think it would be better to use an alternative method to override the URL. When we define a constant in PHP we can't undefine it or re-define it so this change would affect other tests.

Since we have the filter `woocommerce_gla_site_url` I'd suggest to use that instead so we can unset the override.

```suggestion
		$override_url = function () {
			return 'https://secure.url';
		};

		add_filter( 'woocommerce_gla_site_url', $override_url );
		$this->assertTrue( $this->policy_compliance_check->get_is_store_ssl() );
		remove_filter( 'woocommerce_gla_site_url', $override_url );
```

tests/Unit/MerchantCenter/PolicyComplianceCheckTest.php
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	public function test_not_has_refund_page() {
```
mikkamp, line 141: It would be nice to test for an existing refund page as well.

Something like this should work:

```php
	public function test_has_refund_page() {
		$page_id = wp_insert_post(
			[
				'post_title'  => 'Refund and Returns',
				'post_name'   => 'refund_returns',
				'post_type'   => 'page',
				'post_status' => 'publish',
			]
		);

		$this->assertTrue( $this->policy_compliance_check->has_refund_return_policy_page() );
		wp_delete_post( $page_id, true );
	}
```

mikkamp (commented):
Thanks for the adding those tests. I suggested a few tweaks and an additional test.

mikkamp (approved):
Thanks for adding the final changes. It looks good to go so I'll go ahead and merge this to the feature branch.

PR #1206, opened 2022-01-19
Replace FullScreen and FullContainer components with useLayout hook

eason9487:
### Changes proposed in this Pull Request:

Closes #311 

- Move layout CSS styles to primary css folder.
- Implement the `useLayout` hook.
- Replace all `<FullScreen>` and `<FullContainer>` with `useLayout` hook.
- Move the CSS for hiding the `<Spinner>` of `<Suspense>` to global scope to globally prevent the flickering when initial loading GLA admin pages.
- Use CSS approach instead to override **#wpbody**'s inline `margin-top`. To prevent the flickering when overriding the `margin-top`.

#### Full page layout

1. Go to the Get started page `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fstart`
2. Click "Set up free listings in Google" to enter the onboarding page in full-page layout.
    - The whole page should be the same visual result as before.
    - There should be no flickering above the header with "Get started with Google Listings & Ads" content. The flickering problem looks like this:
    Kapture 2022-01-18 at 18 42 00
3. Refresh the current page.
    - The whole page should be the same visual result as before.
    - There should be no flickering above the header with "Get started with Google Listings & Ads" content.
4. Go back to the Get started page. The full-page layout should not be applied.

#### Full content layout

1. Go to the Dashboard page `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard`
2. Click "Add paid campaign" to enter the campaign creation page in full-content layout.
    - The whole page should be the same visual result as before.
    - There should be no flickering above the header with "Create your paid campaign" content.
3. Refresh the current page.
    - The whole page should be the same visual result as before.
    - There should be no flickering above the header with "Create your paid campaign" content.
4. Go back to the Dashboard page. The full-content layout should not be applied.

#### The patch of hiding the `<Spinner>` of `<Suspense>`

1. Go to any of GLA admin pages, for example, the Reports page.
2. Refresh the current page.
3. There should be no flickering above the navigation tab. The flickering problem looks like this:
    Kapture 2022-01-18 at 18 46 06

#### With new WooCommerce navigation

1. Install and activate [the WooCommerce Admin extension](https://wordpress.org/plugins/woocommerce-admin/).
2. Go to WC settings page to enable the new navigation feature `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`
3. Repeat the above tests with new WooCommerce navigation.

### Additional details:

[...]

PR #1626, opened 2022-08-11
Use visibility:hidden to hide the element but preserving offset

### LINKED ISSUE #1625: Incorrect width in GLA .woocommerce-layout__header
### Describe the bug:

I noticed there is a message in the WooCommerce header overlapping the back button in GLA header, making it impossible to go back. Since the position is fixed it seems like the style is broken after setting the width to 0 to make it invisible, this approach since to be not enough to hide the element.

Screenshot 2022-08-11 at 11 41 23

Screenshot 2022-08-11 at 11 50 33

Screenshot 2022-08-11 at 11 50 38

### Steps to reproduce:

1. Setup WooCommerce store without completing all the steps
2. It should appear a message like " You're doing great! 3 steps left  to get your store up and running. "
3. Go to GLA edit free listings
4. See how the message is over the back button and it doesn't allow to click on it.

### Expected behavior:
- Woo message should be completely hidden.

### Actual behavior:
- Woo message is not completely hidden and doesn't allow to click on back button.

### Additional details:

Adding visibility hidden could make the job instead of width 0. `visibility: hidden` hides the element, but it still takes up space in the layout

puntope:

Closes #1625 

This PR implements a fix to avoid the content of the message bar in the header to overflow in the fixed element. Instead of width 0 (initial solution) we use now visibility:hidden. The goal is to preserve the offset of the DOM element without render its content. 

### Screenshots:

before:

Screenshot 2022-08-11 at 11 41 23

after:

Screenshot 2022-08-11 at 16 31 48

1.  Create a store in WooCommerce without finishing all the config
2. Go to Edit free listings
3. See that the message appears in code but is invisible. However, the offset preserves.
4. Back button is clickable

### Changelog entry

> Fix - Hide WooCommerce System messages in the plugin screen.

### DIFF
--- a/js/src/css/shared/_woocommerce-admin.scss
+++ b/js/src/css/shared/_woocommerce-admin.scss
@@ -10,12 +10,11 @@
 	.woocommerce-layout {
 		padding-top: 0;
 
-		// set the layout header width to 0 to hide it.
 		// do not set the header to display: none,
 		// because we want the offsetHeight value
 		// and set it to wpbody-content margin-top to counter back.
 		.woocommerce-layout__header {
-			width: 0;
+			visibility: hidden;
 
 			// only display the top left WC navigation,
 			// do not display the h1 page title element


js/src/css/shared/_woocommerce-admin.scss
```diff
@@ -10,12 +10,11 @@
 	.woocommerce-layout {
 		padding-top: 0;
 
-		// set the layout header width to 0 to hide it.
 		// do not set the header to display: none,
 		// because we want the offsetHeight value
 		// and set it to wpbody-content margin-top to counter back.
 		.woocommerce-layout__header {
-			width: 0;
+			visibility: hidden;
```
eason9487, line 17: 💅 
I think it's fine to use `display: none;` since the original use in `useWPBodyMarginOffsetEffect` (a75ed9f8ed825af7f880d1db005a5dfb45beac8d and ddcaecd96657f835c36750b30951cdb193e34761) had been replaced with another approach in #1206.

eason9487 (approved):
Tested locally and works well. LGTM.

js/src/css/shared/_woocommerce-admin.scss
puntope: Thanks for the review and the suggestion @eason9487 . Since `visibility:hidden` causes no damage I prefer to maintain the code as closest as previous implementation. But totally agree to have a further PR with `display:none` approach if needed.

eason9487: I happened to notice the code comment of this CSS style was outdated. Keeping `visibility: hidden;` is ok as well but the outdated code comment might be confusing.

PR #1627, opened 2022-08-12
Test - Set default timeout

puntope:

Attempts to fix random timeouts in our e2e tests causing by 2 issues:

1. WaitingFor timeout of 500ms with some selectors -> Solution to increase default timeout to 3000ms
2. Test timeout of 30000ms potentially caused by divergences in google domain in regards `trackGtagEvent` -> Solution to modify the function to check the path instead of the domain.

### Additional details

The reason of using timeout of 3000ms instead of something lower like 1000ms is because the default timeout also have impact in navigationTimeout used by page navigation. 

1. Observe the tests in future PR to see if this PR solves the problem
2. Unfortunately there is not a consistent way to test the issue is solved. 

### Changelog entry

### DIFF
--- a/tests/e2e/specs/gtag-events/gtag-events.test.js
+++ b/tests/e2e/specs/gtag-events/gtag-events.test.js
@@ -33,7 +33,8 @@ let simpleProductID;
 
 describe( 'GTag events', () => {
 	beforeAll( async () => {
-		page.setDefaultTimeout( 200 );
+		page.setDefaultTimeout( 3000 );
+		jest.setTimeout( 60000 );
 		await saveConversionID();
 		await createBlockShopPage();
 		simpleProductID = await createSimpleProduct();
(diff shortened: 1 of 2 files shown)


mikkamp (approved):
Thanks for the changes, it looks good to me. I can see them pass in the PR, and when I run it locally they all pass as well:

```
Test Suites: 3 passed, 3 total
Tests:       15 passed, 15 total
Snapshots:   0 total
Time:        38.903s
Ran all test suites.
Test results written to: test-results.json
Jest exit code: 0
```

tomalec:
It seems the test is still flaky, as it failed for https://github.com/woocommerce/google-listings-and-ads/runs/7828290896?check_suite_focus=true before re-run.

mikkamp:
That looks like a different failure as it's not a timeout, but instead the product ID's don't match. Not quite sure why that's happening, we use the shopper util class from WooCommerce to add a specific product to the cart from the shop page, but it appears to be adding a different one.

mikkamp:
I did some additional testing and I seem to have been able to identify that it was sending the event while the previous page was still loaded (hence the wrong product ID). Hopefully [this PR](https://github.com/woocommerce/google-listings-and-ads/pull/1629) will resolve it.

PR #1628, opened 2022-08-14
Remove wc-admin installation from E2E env setup

tomalec:

Remove wc-admin installation from E2E env setup, it should be now merged to WooCommerce.

So the step is not needed and potential performance and debugging obstacles.

1. Check E2e run by github actions.
2. Run E2E locally `npm i && npm run dev && npm run docker:up && npm test:e2e`

### Changelog entry

> Dev - Remove wc-admin installation from E2E env setup


mikkamp (approved):
Looks good and it's working as expected when I run it locally.

I don't think it actually caused any harm because WooCommerce was deactivating it, but at least I now no longer see the message:
> The WooCommerce Admin plugin has been deactivated as the latest improvements are now included with the WooCommerce plugin.

And the plugin is no longer installed on the test site.

PR #1629, opened 2022-08-15
E2E Fix for redirecting to single product page.

mikkamp:

Apparently the E2E tests are still flaky and sometimes fail. I was able to reproduce it some of the times locally, and it seems that the tracking event was being sent from the wrong page. This happened after redirecting to a single product page. We originally used shopper.goToProduct which links to a product page like `?p=<product_id>`, and then waits till `networkIdle`. The problem is that we enabled pretty permalinks, which means this will right away return a redirect to the pretty permalink. In most cases we had resolved this by waiting for the element `form.cart`, however if the previous page also had this element then it would return instantly causing the event to sometimes be sent from the previous page.

This PR resolves this by explicitly waiting for a `form.cart` element on a single product page.

1. Test the E2E tests locally and confirm they pass (was able to reproduce the best locally when running tests 5 seconds after docker:up. 
2. Check the E2E tests in the PR and make sure they haven't failed on any runs.

### Changelog entry
* Dev - E2E Fix for redirecting to single product page.

### DIFF
--- a/tests/e2e/specs/gtag-events/gtag-events.test.js
+++ b/tests/e2e/specs/gtag-events/gtag-events.test.js
@@ -23,6 +23,7 @@ import {
 	relatedProductAddToCart,
 } from '../../utils/cart';
 import { createBlockShopPage } from '../../utils/block-page';
+import { goToSingleProduct } from '../../utils/visit-product-page';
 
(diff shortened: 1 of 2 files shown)


puntope (approved):
Interesting. Great catch.

Approved to test if this PR fix the e2e from now on.

PR #1631, opened 2022-08-16
Onload conflict when tracking events

### LINKED ISSUE #1630: Version 2.0.3 conflicting with site theme
### Describe the bug:

There are reports of Google Listings & Ads 2.0.3 breaking sites. This issue has already been reported on Slack and is being Investigated.

I'm creating this report to help keep track and follow up on affected users.

### Forum threads

https://wordpress.org/support/topic/checkout-not-loading-after-update/
https://wordpress.org/support/topic/version-2-0-3-broke-headerst-but-2-0-2-works-fine/
https://wordpress.org/support/topic/last-update-caused-problems-in-sites-functions/
https://wordpress.org/support/topic/latest-update-has-broken-the-site/

### Additional details:

p1660563720862579-slack-C01E9LB4X61

mikkamp:

This PR changes the onload function to use addEventListener, there were various sites which reported broken behaviour because `window.onload` can only be defined once. I was able to confirm that three out of the four linked issues contained code which assigned a function to `window.onload`. On the fourth site I wasn't able to find the explicit usage of the code, but there are still chances it's what caused the conflict.

The PR also changes the inline code snippets to be wrapped in a loading function so the event is run when the page has loaded.

Closes #1630

1. Test on a site with a theme that overrides `window.onload` or use a code snippet like this:
```
add_action(
	'wp_head',
	function () {
		wp_print_inline_script_tag( 'window.onload = function() { console.log("running"); }' );
	},
	1
);
```
2. Ensure Merchant and Ads onboarding has completed (to allow for event tracking).
3. Confirm that with latest develop we don't see the console output "running". 
4. Try again with this PR (rebuild the JS after switching) and confirm that the console output is shown.
5. Try viewing various pages and click add to cart buttons.
6. Confirm that for each of these we see a tracking event sent to `https://google.com/pagead` (URL might vary slightly depending on region).

### Changelog entry
* Fix - Onload conflict when tracking events.

### DIFF
--- a/js/src/gtag-events/index.js
+++ b/js/src/gtag-events/index.js
@@ -66,7 +66,7 @@ const singleAddToCartClick = function ( event ) {
 };
 
 // Register for add_to_cart click events.
-window.onload = function () {
+window.addEventListener( 'load', function () {
 	document
 		.querySelectorAll(
 			'.add_to_cart_button:not( .product_type_variable ):not( .product_type_grouped )'
@@ -80,7 +80,7 @@ window.onload = function () {
 		.forEach( ( button ) => {
 			button.addEventListener( 'click', singleAddToCartClick );
 		} );
-};
+} );
 
 // Register for jQuery event to update product data.
 if ( typeof jQuery === 'function' ) {
--- a/js/src/gtag-events/index.js
+++ b/js/src/gtag-events/index.js
@@ -66,7 +66,7 @@ const singleAddToCartClick = function ( event ) {
 };
 
 // Register for add_to_cart click events.
-window.addEventListener( 'load', function () {
+document.defaultView.addEventListener( 'DOMContentLoaded', function () {
 	document
 		.querySelectorAll(
 			'.add_to_cart_button:not( .product_type_variable ):not( .product_type_grouped )'
--- a/src/Google/GlobalSiteTag.php
+++ b/src/Google/GlobalSiteTag.php
@@ -293,7 +293,7 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 			esc_js( $order->get_currency() ),
 			esc_js( $order->get_id() ),
 		);
-		$this->print_inline_gtag_script( $conversion_gtag_info );
+		wp_print_inline_script_tag( $conversion_gtag_info );
 
 		// Get the item info in the order
 		$item_info = [];
@@ -354,7 +354,7 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 			esc_js( $language ),
 			join( ',', $item_info ),
 		);
-		$this->print_inline_gtag_script( $purchase_page_gtag );
+		wp_print_inline_script_tag( $purchase_page_gtag );
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


mikkamp:
I haven't been able to track down why the E2E tests are failing for this PR, they pass locally and I can confirm that the event is sent when adding to cart from a block page. Since this change has a bit of urgency I think we can ignore the tests failing.

puntope (dismissed):
For me this PR is working:

- Checked locally (i see running log ad well as the pageads events)

puntope (commented):
Sorry, I approved it so fast. I think it could be issues in related product (e2e test) I want to take a look again on those before approval

js/src/gtag-events/index.js
```diff
@@ -66,7 +66,7 @@ const singleAddToCartClick = function ( event ) {
 };
 
 // Register for add_to_cart click events.
-window.onload = function () {
+window.addEventListener( 'load', function () {
```
tomalec: To make linter happy, and to execute once content is there (as suggested by @eason9487 
```suggestion
document.defaultView.addEventListener( 'DOMContentLoaded', function () {
```

mikkamp: Would you suggest to make the same change here: https://github.com/woocommerce/google-listings-and-ads/blob/fix/1630-onload-conflict/src/Google/GlobalSiteTag.php#L465

tomalec: I think so.

However, I'm still not sure about the timing of all those `maybe_display_conversion_and_purchase_event_snippets`. Are we guaranteed they are stamped before the document loads? (So we won't attach the event listener after it was fired)

mikkamp: Those events update some order meta when the page is generated in PHP so it isn't tracked a second time. We will be sending the event a little later then before, but previously it was also a problem that even though we mark it as tracked we can't guarantee that the event actually got sent when the page was loaded.
It's the same pattern that's used in Google Analytics Integration, so it is widely used.

mikkamp: > So we won't attach the event listener after it was fired

The function `print_inline_gtag_script` wraps the event code in the onload function, so it can't fire the event before the listener is attached.

tomalec: I meant a different thing.

My question was about the timing of `maybe_display_conversion_and_purchase_event_snippets` itself. When it is called, are we sure it's not called after the content is loaded. In cases like the streamed HTML, or some other tricks to insert an executable `script` tag.
Then the timing would be:
1. Script A does `window.addEventListener('DOMContentLoaded', doSmth );`
2. Content is loaded
3. `doSmth` is executed
4. Script B does `window.addEventListener('DOMContentLoaded', doSmthElse );`

Then `doSmthElse` is never fired

tomalec: But I guess we don't do much magic about those scripts and just concatenate them to the returned document, right?

mikkamp: Yes it's printed directly in the document. Although there are plugins that can compress / lazy load inline scripts as well do you think in that case it's safer to not wrap the inline scripts?

tomalec (approved):
Reviewed the code, tested locally LGTM.

PR #1633, opened 2022-08-16
Add a filter to disable GTag tracking

### LINKED ISSUE #1609: Add filter to disable event tracking

There are users of the Google Listings and Ads plugin that want to use different plugins for tracking than the integrated one in the Google Listings and Ads plugin. Such a third party tracking plugin is the [Pixel Manager for WooCommerce](https://wordpress.org/plugins/woocommerce-google-adwords-conversion-tracking-tag/) (of which I am the developer of). There are numerous advantages in using a third party plugin like the Pixel Manager for WooCommerce over the integrated tracking. I'll be happy to explain that in more detail if you require me to. 

In order to be able to use the Google Listings and Ads plugin together with the Pixel Manager for WooCommerce it would be great if you could add a filter with which the customer (or the Pixel Manager) could deactivate the integrated tracking in the Google Listings and Ads plugin. 

Could you add such a filter?

mikkamp:

This PR adds a filter for disabling the internal GTag tracking, the intention is for the filter to only be used when replacement tracking has been enabled. See #1609 for valid scenarios.

Closes #1609

1. Ensure Merchant and Ads onboarding has completed (to allow for event tracking).
2. Try viewing various pages and click add to cart buttons.
3. Confirm that for each of these we see a tracking event sent to https://google.com/pagead (URL might vary slightly depending on region).
4. Enable the filter to disable tracking
```php
add_filter( 'woocommerce_gla_disable_gtag_tracking', '__return_true' );
```
5. Confirm that the events are no longer sent when we perform the same actions.
6. Confirm that the frontend script `js/build/gtag-events.js` is no longer loaded.

### Changelog entry
* Tweak - Add a filter to disable GTag tracking.

### DIFF
--- a/src/Google/GlobalSiteTag.php
+++ b/src/Google/GlobalSiteTag.php
@@ -460,6 +460,10 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 	 * @return bool True if the Global Site Tag framework should be included.
 	 */
 	public static function is_needed(): bool {
+		if ( apply_filters( 'woocommerce_gla_disable_gtag_tracking', false ) ) {
+			return false;
+		}
+
 		return true;
 	}
 


ianlin (approved):
Thanks for adding this, the gtag event js isn't loaded when I add the disable filter. However I wasn't sure how to access https://google.com/pagead, and I did not see the gtag events being sent (by checking the network tab in dev tool) when the disable filter isn't added. I think it could be my environment problem so I approved the PR first. (I have set up merchant and ads account and enable event tracking in WC.)

mikkamp:
These are the steps to test the events manually: https://github.com/woocommerce/google-listings-and-ads/wiki/Smoke-tests#gtag-events-covered-by-e2e-tests-but-can-be-tested-manually
You should at least see some of the JS being added in the source code when it's enabled. In the network tab try just searching for `/pagead` depending on region it sometimes uses a slightly different URL. For example I've had it send to `https://region1.google.com/pagead`

ianlin:
Thanks for clarifying. I have tried several times and I can confirm that when the disable filter is not added, I can see some of the JS being added in the page source, also there is a comment `Global site tag` in the page source. I can also see the `gtag-events.js` is loaded via network tab. When I add the disable filter, all of the above mentioned things were gone. The only thing I can't confirm is when disable filter isn't added, the gtag event wasn't being sent when adding a simple product to the cart. I couldn't see anything by searching `pagead` in the network tab. However I think it's a bit out of the topic for this PR. the filter this PR added can prevent the gtag JS from being added and I think it's good to go.

mikkamp:
Yes you are correct that it doesn't affect the changes in this PR. I was just curious if it highlighted some other problem with the tracking code which could be reported in a followup issue.
If you do have some extra time to check why the code in `gtag-events.js` is not triggering or maybe it's still loading an older cached version. Then it would be helpful to report it in a separate issue, in case anyone else is experiencing the same.
I'll go ahead and merge this PR now.

ianlin:
Just found out the reason I couldn't see the events being triggered: I used Brave browser and it blocks all the tracking and ads by default 🤦 Turned it off and I could see the events being sent without any problems.

PR #1636, opened 2022-08-17
E2E Testing: Reduce the false positive rate and adjust the running events on GitHub Actions

eason9487:

- Change the E2E test to only run for `release/*` branches automatically and to accept `workflow_dispatch` event for running it manually.
- By comparing the latest 3 pages of [the e2e test results](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/e2e-tests.yml) and [the results of this PR](https://github.com/woocommerce/google-listings-and-ads/actions/runs/3021878781), it shows this PR reduces the false positive rate from ≈40% (25 failed in 62 runs) to 10% (10/60).

The following are the detail of adjustments.

#### Tweak the selector in `blockProductAddToCart` function to avoid including the "Add to Cart" buttons of Gutenberg blocks.

- Originally, it had also added the click event to the buttons from Gutenberg blocks, and it would trigger the `'add_to_cart'` event twice. The `product_id` property in the second one would be `'gla_undefined'`.
- The above problem has been fixed in #1631 by changing `load` to `DOMContentLoaded` since the product items are not yet rendered by React when the `DOMContentLoaded` event is fired.
- This adjustment is to make the selector not include those buttons even though it happens to have no problems for now.

#### Adjust some failure-prone points to reduce the possibility of e2e tests failure.

- Wait for the loaded status (no `disabled` attribute) when clicking on the "Add to Cart" button for a blocks product in an e2e test.
   - I suspect the test might click on the button while the products are still loading. (see [e2e test run 7857596997](https://github.com/woocommerce/google-listings-and-ads/runs/7857596997?check_suite_focus=true))
   - 
- Set timeout for `waitForRequest` call in `trackGtagEvent` function to cover a wider range of time it may take to perform e2e tests.
    - When the moment of a test calls `const event = trackGtagEvent( 'add_to_cart' );`, it already started timing. It might result in a timeout error due to a long run time of e2e test it reaches and fulfills the `await event.then()`. (see [e2e test run 7855437517](https://github.com/woocommerce/google-listings-and-ads/runs/7855437517?check_suite_focus=true))
- Increase the default timeout for all gtag e2e tests to reduce the possibility of timeout errors.
- Adjust the slash in the all-products-block page URL to prevent from 301 redirects in an e2e test.

1. Run e2e tests locally.
2. View [the e2e tests results for validation](https://github.com/woocommerce/google-listings-and-ads/actions/runs/3021878781) of this PR on GitHub Actions.
3. Check if the e2e tests are not run with the last commit of this PR.

### Additional details:

According to the discussion on slack, we decided to run the e2e tests only on `release/*` branches when pulling requests or pushing, and also make it can be run manually. Ref: p1660734637321149/1660734005.228109-slack-CK365S85V

There is still a failure case that I could not reproduce or simulate locally, so maybe it needs to be investigated further.
- Mismatched product ID: https://github.com/woocommerce/google-listings-and-ads/runs/7875505899?check_suite_focus=true#step:8:61

And sometimes the test crashed with this error: `Puppeteer - Protocol error (Page.navigate): Target closed`.
- This type of error log may even look like a timeout error.
- https://stackoverflow.com/questions/51629151/puppeteer-protocol-error-page-navigate-target-closed

### Changelog entry


tests/e2e/utils/track-event.js
```diff
@@ -15,11 +15,12 @@
  */
 export function trackGtagEvent( eventName ) {
 	const eventPath = '/pagead';
+	const option = { timeout: 15000 };
```
mikkamp: According to the documentation the default timeout for waitForRequest is 30 seconds, but it can be overwritten with `page.setDefaultTimeout`. When originally implementing I tested it with the 30 second timeout which seemed long enough.
Ever since PR #1627 the default timeout was set at 3 seconds (bumped up to 10 seconds in this PR). Shouldn't we just set it back to the default of 30 seconds here? Or set the page default timeout to 30 seconds so we only need to set it once, I don't think it's a big issue that it takes longer to timeout if it will be more stable.

eason9487: Thanks for the suggestion, adjusted it in two commits: https://github.com/woocommerce/google-listings-and-ads/pull/1636/commits/25944aed3404b8692fcd64d2bc662bbf14fb6483 and https://github.com/woocommerce/google-listings-and-ads/pull/1636/commits/9c3d297f48eca3dd9e7305743f06edefcc035d99

README.md
```diff
@@ -4,7 +4,6 @@
 [![JavaScript Unit Tests](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/js-unit-tests.yml/badge.svg)](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/js-unit-tests.yml)
 [![PHP Coding Standards](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/php-coding-standards.yml/badge.svg)](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/php-coding-standards.yml)
 [![JavaScript and CSS Linting](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/js-css-linting.yml/badge.svg)](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/js-css-linting.yml)
-[![E2E Tests](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/e2e-tests.yml/badge.svg)](https://github.com/woocommerce/google-listings-and-ads/actions/workflows/e2e-tests.yml)
```
eason9487, line 7: Removed this badge as it won't stay with the latest e2e testing result from time to time.

tomalec (approved):
Reviewed the changes, run locally, and checked the PR checks: LGTM.

tomalec:
For future reference. 
I faced some problems running E2E on my Linux setup with docker, which resulted in `wc-e2e test:e2e-dev` throwing `UnhandledPromiseRejectionWarning: Error: Page crashed!`. It seems there is some Docker privileges problem. 
I managed to work it around by adding `--no-sandbox` in [`tests/e2e/config/jest-puppeteer.config.js`](https://github.com/woocommerce/google-listings-and-ads/blob/develop/tests/e2e/config/jest-puppeteer.config.js#L7)

PR #1637, opened 2022-08-17
Free Listings + Paid Ads: Add mc setup complete API

ianlin:

This PR implements the [backend task 0](https://github.com/woocommerce/google-listings-and-ads/issues/1610#be-task0) **📌 Add a new API for onboarding completed** in the epic issue #1610.

- Add an API `mc/setup/complete` which will be called at the step 4 of new onboarding flow, no matter the merchant skip or set up a paid campaign.
- This new API does only one thing: set the timestamp to the option `gla_mc_setup_completed_at`.
- This PR also modifies an existing API `mc/settings/sync` by not setting the timestamp to the option `gla_mc_setup_completed_at`.

1. Use Postman, make a `POST` request to `mc/setup/complete`.
2. In the database, the value of the option `gla_mc_setup_completed_at` should be added or updated.
3. Use Postman, make a `POST` request to `mc/settings/sync`.
4. In the database, the value of the option `gla_mc_setup_completed_at` should not be added or updated.

### DIFF
--- a/src/API/Site/Controllers/MerchantCenter/SetupCompleteController.php
+++ b/src/API/Site/Controllers/MerchantCenter/SetupCompleteController.php
@@ -1,0 +1,68 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\EmptySchemaPropertiesTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use WP_REST_Request as Request;
+use WP_REST_Response as Response;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class SetupCompleteController
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter
+ */
+class SetupCompleteController extends BaseController {
+
+	use EmptySchemaPropertiesTrait;
+
+	/**
+	 * Registers the routes for the objects of the controller.
+	 */
+	public function register_routes() {
+		$this->register_route(
+			'mc/setup/complete',
+			[
+				[
+					'methods'             => TransportMethods::CREATABLE,
+					'callback'            => $this->get_setup_complete_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+			]
+		);
+	}
+
+	/**
+	 * Get the callback function for marking setup complete.
+	 *
+	 * @return callable
+	 */
+	protected function get_setup_complete_callback(): callable {
+		return function( Request $request ) {
+			do_action( 'woocommerce_gla_mc_setup_completed' );
+
+			return new Response(
+				[
+					'status'  => 'success',
+					'message' => __( 'Successfully marked merchant center setup as completed.', 'google-listings-and-ads' ),
+				],
+				201
+			);
+		};
+	}
+
+	/**
+	 * Get the item schema name for the controller.
+	 *
+	 * Used for building the API response schema.
+	 *
+	 * @return string
+	 */
+	protected function get_schema_title(): string {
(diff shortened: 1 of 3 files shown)


src/API/Site/Controllers/MerchantCenter/SetupCompleteController.php
```diff
@@ -0,0 +1,68 @@
...
+	protected function get_setup_complete_callback(): callable {
+		return function( Request $request ) {
+			do_action( 'woocommerce_gla_mc_setup_completed' );
+
+			return new Response(
+				[
+					'status'  => 'success',
+					'message' => __( 'Successfully marked merchant center setup as completed.', 'google-listings-and-ads' ),
+				],
+				201
```
mikkamp, line 53: I know this is a copy of the previous controller. But 201 means something was successfully created, I think 200 would fit better in this case. Also in the sync settings controller we are syncing and not creating, so 200 would suit better there.

src/API/Site/Controllers/MerchantCenter/SetupCompleteController.php
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	/**
+	 * Get the callback function for marking setup complete.
+	 *
+	 * @return callable
+	 */
+	protected function get_setup_complete_callback(): callable {
+		return function( Request $request ) {
+			do_action( 'woocommerce_gla_mc_setup_completed' );
```
mikkamp, line 46: This works fine, but it just seems to me it kind of hides the purpose of what this controller is doing. I think the action hook to trigger additional tasks is handy. But I don't see why we can't call `set_completed_timestamp` directly here if it's changed to a public function. I have the same opinion about `set_contact_information_setup`, being set directly and `MerchantSetupCompleted` not needing to be a Registerable. Just my opinion though, we don't need to necessarily rewrite all this code.

mikkamp (commented):
Should the HooksDocsGenerator script be run as part of this PR to add the new hook to `src/Hooks/README.md`?

The StartProductSync event is hooked into `woocommerce_gla_mc_settings_sync`, if it's the intention to call the settings sync endpoint multiple times, then shouldn't we switch the product sync to the hook `woocommerce_gla_mc_setup_completed`?

Thanks for the changes, the two controllers work as expected. I mostly left opinionated comments, I think the only important one is to change when the product sync is triggered.

mikkamp:
Just wanted to add that I'm not really familiar with the new flow, and it seems like it's still intentional to start the product sync at step 3 (although we would want to prevent this from happening multiple times), if step 1-3 are repeated, or we would need to restart the product sync and cancel the previous if target countries change. But it gets kinda tricky to start a product sync and settings sync (auto generate shipping), and then we change the settings half way.

But the main thing is, that as it is now, product sync won't work without `gla_mc_setup_completed_at` being set, so we are going to need some refactoring there.

Personally I think even though the two flows are being added in one we should still keep setting `gla_mc_setup_completed_at` and `gla_ads_setup_completed_at` happening at the same times as they where previously, because there are too many dependent things currently attached. Also if they go back to step 1-3 and change anything we should consider that as a MC disconnect so it can clear up previous data/restart product sync.

ianlin:
> But the main thing is, that as it is now, product sync won't work without `gla_mc_setup_completed_at` being set, so we are going to need some refactoring there.

Absolutely spot on, thanks a lot @mikkamp. Sorry I didn't catch this problem 😓, in this case this approach can't work. I will close this PR and update the epic issue #1610 with the new approach.

Short explanation about what we wanted to achieve and the potential new approach:

- In step 4 (new flow, see fqR0EHi63lWahRcVTKCcba-fi-449%3A80582), we want to display the number of products that being uploaded.
- Frontend plans to call the API `mc/product-statistics` to get the number.
- The API `mc/product-statistics` depends on `mc/settings/sync` is done.
- The API `mc/settings/sync` will set `gla_mc_setup_completed_at`.
- When `gla_mc_setup_completed_at` is set, frontend will go to the product feed page, so the step 4 is never showing.
- That's the reason this PR was created, but I missed one important thing that product sync won't work without `gla_mc_setup_completed_at` being set.

Since the original goal is for frontend to know the number of products are being uploaded (or will be uploaded), we decide to create a new API for it, rather than calling `mc/product-statistics`. E.g. `GET wc/gla/onboarding/products` (naming TBD).

PR #1638, opened 2022-08-17
Free Listings + Paid Ads: Add the product feed status section UI

eason9487:

This PR implements the UI part of **📌 Product feed status section** in #1610.

- Extract a code fragment from the sync-status.js as `getNumberOfSyncProducts` function.
- Implement `ProductFeedStatusSection` component and add it to the paid ads setup step of onboarding flow.

#### 📷 Normal view

image

#### 📷 Hover on the "# products" text

image

1. Check if the visual result of the "Product feed status" section is close to the design in Figma.
2. Hover on the "# products" text to view the tooltip.

### DIFF
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -96,6 +96,24 @@ export function handleFetchError( error, message ) {
  */
 
 /**
+ * @typedef {Object} ProductStatisticsDetails
+ * @property {number} active Number of active products.
+ * @property {number} expiring Number of expiring products.
+ * @property {number} pending Number of pending products.
+ * @property {number} disapproved Number of disapproved products.
+ * @property {number} not_synced Number of not synced products.
+ */
+
+/**
+ * Product status statistics on Google Merchant Center
+ *
+ * @typedef {Object} ProductStatistics
+ * @property {number} scheduled_sync Number of scheduled jobs which will sync products to Google.
+ * @property {number} timestamp Timestamp reflecting when the product status statistics were last generated.
+ * @property {ProductStatisticsDetails} statistics Statistics information of product status on Google Merchant Center.
+ */
+
+/**
  *
  * @return {Array<ShippingRate>} Array of individual shipping rates.
  */
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -889,6 +889,13 @@ export function* receiveMCSetup( mcSetup ) {
 	};
 }
 
+/**
+ * Creates a wp-data action with data payload to be dispatched the received
+ * MC product statistics to wp-data store.
+ *
+ * @param {ProductStatistics} mcProductStatistics The received MC product statistics data.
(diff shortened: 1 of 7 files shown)


js/src/data/actions.js
```diff
@@ -95,6 +95,24 @@ export function handleFetchError( error, message ) {
...
+ * @property {number} expiring Number of expiring products.
+ * @property {number} pending Number of pending products.
+ * @property {number} disapproved Number of disapproved products.
+ * @property {number} not_synced Number of not synced products.
+ */
+
+/**
+ * Product status statistics on Google Merchant Center
+ *
+ * @typedef {Object} ProductStatistics
```
ianlin, line 110: 💅  Since the type of `ProductStatistics` is defined, it would be nice to add some doc comments in the below function that is using this type.

https://github.com/woocommerce/google-listings-and-ads/blob/9333fa5a0c2a69dc0aa177cd119be07bafcd3698/js/src/data/actions.js#L892

js/src/setup-mc/setup-stepper/setup-paid-ads/product-feed-status-section.js
```diff
@@ -0,0 +1,113 @@
...
+ * Renders a section layout to elaborate on how the product listings will be processed
+ * and show the number of products will be synced to Google Merchant Center.
+ */
+export default function ProductFeedStatusSection() {
+	/*
+	const { data, hasFinishedResolution } = useAppSelectDispatch(
+		'getMCProductStatistics'
+	);
+	*/
+	// TODO: Replace the dummy data with the above code later to use the adjusted API.
```
ianlin, line 60: ❔ Just wanted to make sure my understanding is right. The API `mc/product-statistics` would only be returning data after we call `mc/settings/sync`, so the comment `Replace the dummy data with the above code later to use the adjusted API.` means we would make a POST request to  `mc/settings/sync` later, right?

ianlin (approved):
LGTM.

✅  The UI looks close to the design.
✅  The tooltip was displayed when hovering on the `# products`.

js/src/setup-mc/setup-stepper/setup-paid-ads/product-feed-status-section.js
eason9487: Yes, as per my test from the frontend side, it has to call `mc/settings/sync` before getting statistics from `mc/product-statistics`, so the frontend side will call `mc/settings/sync` when entering step 4. And the `mc/product-statistics` seems to check whether the `MC_SETUP_COMPLETED_AT` flag is set before returning, thus, we will need the related adjustment in this project as well.

js/src/data/actions.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/1638/commits/456668361dccc992d835445f05ae3f24cfebd8fb.

PR #1650, opened 2022-08-24
Free Listings + Paid Ads: Add the paid ads previews to the boost product listings section

eason9487:

This PR implements the ad previews part of **📌 Boost product listings section** in #1610.

- Dynamically compositing the ad previews with product and shop data.

- [x] 💡Please note that there is a very small white border in the fallback image of product cover. It's not a CSS style problem and I will check with the design team later.
   - Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1650/commits/6fe6ba71e1b43c78b8519ff2fc5cdd5f3ca40cc4.

#### 📹 All ad previews and blur animation

https://user-images.githubusercontent.com/17420811/186405954-5cf6471b-7889-4fcc-9c69-88c8d2d14037.mp4

1. Go to step 4 of the onboarding flow.
1. Check if the visual result of the ad previews in the "Boost product listings with paid ads" section is close to the design in Figma.
1. Change [these values](https://github.com/woocommerce/google-listings-and-ads/blob/05779b932a21b71b8d7aa41321fe097b6f122bdb/js/src/components/paid-ads/campaign-preview/campaign-preview.js#L29-L48) locally to see if the layouts of ad previews are still rendered properly.


.eslintrc.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
 			'stylelint',
 			'@wordpress/stylelint-config',
 			'@pmmmwh/react-refresh-webpack-plugin',
+			'react-transition-group',
```
puntope, line 31: ❓ Just curiosity. Not an issue. Why is needed to add this dependency here?

js/src/components/paid-ads/campaign-preview/components/placeholder.js
```diff
@@ -0,0 +1,52 @@
...
+	'gray-500': 'gla-ads-mockup__placeholder--gray-500',
+};
+
+/**
+ * Renders a placeholder to draw a line.
+ *
+ * @param {Object} props React props.
+ * @param {boolean} [props.thinnest] Whether to draw a thinnest placeholder.
+ * @param {boolean} [props.thinner] Whether to draw a thinner placeholder.
+ * @param {boolean} [props.thicker] Whether to draw a thicker placeholder.
```
puntope: 💅  I see this just set the height of the placeholder?? 

Why not just type="thinnest|thinner|thicker" ? Or just do height={unit}  and set up the height as we do with the width.

js/src/components/paid-ads/campaign-preview/components/scaled-text.js
```diff
@@ -0,0 +1,40 @@
...
+	'gray-800': 'gla-ads-mockup__scaled-text--gray-800',
+};
+
+/**
+ * Renders a placeholder to draw a line.
+ *
+ * @param {Object} props React props.
+ * @param {boolean} [props.adBadge=false] Whether to show the 'AD' badge.
+ * @param {boolean} [props.smaller] Whether to draw a smaller text size.
+ * @param {boolean} [props.larger] Whether to draw a larger text size.
```
puntope: 💅  Same as the case of thinner/ticker... Why not just set a type prop and add the class name modifier? Or do the inline style mdoification?

puntope:
❓  Regarding google shopping logo (sorry somehow I don't know how to comment in the file) 

Shall we check if that is still the Shopping logo we want?[ I just modified them recently with the new logo ](https://github.com/woocommerce/google-listings-and-ads/pull/1653)

js/src/components/paid-ads/campaign-preview/index.js
```diff
@@ -0,0 +1,13 @@
+export { default } from './campaign-preview';
+
+/**
+ * Data for compositing ad preview mockups.
+ *
+ * @typedef {Object} AdPreviewData
+ * @property {string} title Price of the product.
```
puntope: ❓ Should this be "Title of the product"?

js/src/components/paid-ads/campaign-preview/campaign-preview.js
```diff
@@ -0,0 +1,94 @@
...
+	}, [] );
+
+	useEffect( () => {
+		if ( second === 0 ) {
+			if ( callCount > 0 ) {
+				moveBy( 1 );
+			}
+			startCountdown( 5 );
+		}
+	}, [ second, callCount, startCountdown, moveBy ] );
```
puntope, line 79: 💅  This maybe can be wrapped in some kind of `useTimingIndex`

puntope:
💅  Maybe some adjustments needed for resolutions between 500-700px

Screenshot 2022-08-29 at 17 36 50

.eslintrc.js
eason9487: This PR uses the same `react-transition-group` dependency as the one `@woocommerce/components` depends on, so `react-transition-group` is not listed in this repo's package.json file. Therefore, this change is to let eslint know it's one of the core modules and to prevent from the eslint error `import/no-extraneous-dependencies`.

js/src/components/paid-ads/campaign-preview/campaign-preview.js
eason9487: Wrapping this to another custom hook sometimes also increases the complexity. In this case, the `CampaignPreview` component doesn't have much code, therefore, I think it ok to stay with the current implementation.

js/src/components/paid-ads/campaign-preview/index.js
eason9487: Thanks for catching this. Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1650/commits/8e72981c45e5ba8bc62f09bb2c1bb8bf1ebb7fbb.

js/src/components/paid-ads/campaign-preview/components/placeholder.js
eason9487: I forgot the original reason. I probably started with three different values of height and was inspired by the props of `Button` component (`isPrimary`, `isSecondary`, `isLink`, etc.). But when adding the color, it was not so ideal to use the same pattern, so I changed to using a single prop with different values.

And it's definitely better to have consistent props for this component. I changed them in https://github.com/woocommerce/google-listings-and-ads/pull/1650/commits/c6908de6d3f4a2abe4ee388a410ebb3884b41e39.

P.S. I have double-checked the visual results again with Figma file when doing this change. And some prop values were different from their previous value in this commit since I found them are not the closest ones to the visual design. Not sure if Figma file was changed or if I set the wrong values in the beginning. 😅

js/src/components/paid-ads/campaign-preview/components/scaled-text.js
eason9487: Changed in https://github.com/woocommerce/google-listings-and-ads/pull/1650/commits/631e8fe5f5f680ea82ee2c427a6d51617817669a.

eason9487:

> ❓ Regarding google shopping logo (sorry somehow I don't know how to comment in the file)
> Shall we check if that is still the Shopping logo we want?[ I just modified them recently with the new logo](https://github.com/woocommerce/google-listings-and-ads/pull/1653)

I checked this with the design team, and it's confirmed that the logo is correct for this project. Because it's supposed to be a Google Shopping logo on the ad preview of Google Shopping.

> 💅 Maybe some adjustments needed for resolutions between 500-700px

Thanks for catching this. I added some CSS styles for this in https://github.com/woocommerce/google-listings-and-ads/pull/1650/commits/656aeb992ca1e175747dde95c4810d966c345a50.

---

I believe all comments were replied to or fixed. Could you help with another round of code reviews?

puntope (approved):
✅ LGTM

💅  One thing I'm missing here and in this feature are the tests

Issue #1648, opened 2022-08-24
Custom Attribute Mapping


puntope:
## User story

This feature implements GLA Attribute Mapping

Project Thread: pcTzPl-HY-p2

## Figma link

https://www.figma.com/file/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=400%3A80229

## Acceptance criteria

* [ ] Attribute Mapping Page in GLA
* [ ] Endpoint in order to get the "sources" options
* [ ] Endpoint in order to get the "destinations" options
* [ ] Attribute Mapping Table
* [ ] New Attribute mapping modal
* [ ] Save Attribute mapping config in WP Options
* [ ] Override GLA Product info with attribute mapping
* [ ] Test coverage
* [ ] Tracking

## Event tracking

- `gla_modal_open` - Triggered when opening the new attribute mapping modal is open 
- `gla_modal_closed` - Triggered when opening the new attribute mapping modal is closed 
- `add_attribute_mapping` - Triggered when the user creates a new attribute mapping
- `delete_attribute_mapping` - Triggered when the user deletes an attribute mapping
- `update_attribute_mapping` - Triggered when the user updates an attribute mapping

PR #1656, opened 2022-08-26
Attribute mapping - Endpoints for Sources and Destinations

puntope:

Closes part of #1648.

This PR implements 2 new endpoints related to Attribute Mapping Feature

**Destinations Endpoint:**

`GET /wp-json/wc/gla/mc/attribute-mapping/destinations`

⚠️  After Review new endpoint is: `GET /wp-json/wc/gla/mc/mapping/attributes`
Gets all the available attributes. This is, all the Google attributes we allow to do a mapping.

(Notice for now we only send 2, as a dummy data, since it is still under discussion which attributes we allow finally) 

**Source Endpoint:**

`GET /wp-json/wc/gla/mc/attribute-mapping/sources?destination={destination}`
⚠️  After Review new endpoint is: `GET /wp-json/wc/gla/mc/mapping/sources?attribute={attribute}`

Param Attribute: String identifying the attribute ID we want to get the sources

Gets all the available sources. This is, all the Custom fields, Taxonomies or options we allow to do a mapping for that destination.

(Notice for now we only send 2, as a dummy data, since it is still under discussion which attributes we allow finally) 

1. Checkout PR
2. Go to browser and load `/wp-json/wc/gla/mc/mapping/attributes`
3. You should receive the dummy data like

```
{
"data": {
"adult": "Adult",
"ageGroup": "Age group"
}
}
```

4. Go to browser and load `/wp-json/wc/gla/mc/mapping/sources?attribute=adult`
5. You should receive the dummy data like

```
{
"data": {
"yes": "Yes",
"no": "No"
}
}
```

6. Go to browser and load `/wp-json/wc/gla/mc/mapping/sources?attribute=ageGroup`
7. You should receive the dummy data like

```
{
"data": {
"newborn": "Newborn",
"infant": "Infant",
"toddler": "Toddler",
"kids": "Kids",
"adult": "Adult"
}
}
```

8. Go to browser and load `/wp-json/wc/gla/mc/mapping/sources?attribute=nothing`
7. You should receive the dummy data like

```
{
"data": []
}
```

9. Go to browser and load `/wp-json/wc/gla/mc/mapping/sources`
7. You should receive the dummy data like

```
{
"data": []
}
```

### DIFF
--- a/src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
+++ b/src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
@@ -1,0 +1,157 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\AttributeMappingHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use WP_REST_Request as Request;
+use WP_REST_Response as Response;
+use Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class for handling API requests for getting source and destination data for Attribute Mapping
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter
+ */
+class AttributeMappingController extends BaseOptionsController {
+
+	/**
+	 * @var AttributeMappingHelper
+	 */
+	private AttributeMappingHelper $attribute_mapping_helper;
+
+
+	/**
+	 * AttributeMappingController constructor.
+	 *
+	 * @param RESTServer             $server
+	 * @param AttributeMappingHelper $attribute_mapping_helper
+	 */
+	public function __construct( RESTServer $server, AttributeMappingHelper $attribute_mapping_helper ) {
+		parent::__construct( $server );
+		$this->attribute_mapping_helper = $attribute_mapping_helper;
+	}
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		/**
+		 * GET the destination fields for Google Shopping
+		 */
+		$this->register_route(
+			'mc/attribute-mapping/destinations',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_attribute_mapping_destinations_read_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+				'schema' => $this->get_api_response_schema_callback(),
+			],
+		);
+
+		/**
+		 * GET for getting the source data for a specific destination
+		 */
+		$this->register_route(
+			'mc/attribute-mapping/sources',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_attribute_mapping_sources_read_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+				'schema' => $this->get_api_response_schema_callback(),
+			],
+		);
+	}
+
+	/**
+	 * Get the callback function for returning the destinations data
+	 *
+	 * @return callable
+	 */
+	protected function get_attribute_mapping_destinations_read_callback(): callable {
+		return function ( Request $request ) {
+			try {
+				return $this->prepare_item_for_response( $this->get_destinations(), $request );
+			} catch ( Exception $e ) {
+				return new Response( [ 'message' => $e->getMessage() ], $e->getCode() ?: 400 );
+			}
+		};
+	}
+
+	/**
+	 * Get the callback function getting destination data.
+	 *
+	 * @return callable
+	 */
+	protected function get_attribute_mapping_sources_read_callback(): callable {
+		return function( Request $request ) {
+			try {
+				$destination = $request['destination'];
+				return $this->get_sources_for_destination( $destination );
+			} catch ( Exception $e ) {
+				return $this->response_from_exception( $e );
+			}
+		};
+	}
+
+	/**
+	 * Get the item schema properties for the controller.
+	 *
+	 * @return array
+	 */
+	protected function get_schema_properties(): array {
+		return [
+			'data' => [
+				'type'        => 'array',
+				'description' => __( 'The data with the different destinations or sources.', 'google-listings-and-ads' ),
+				'context'     => [ 'view' ],
+				'readonly'    => true,
+			],
+		];
+	}
+
+
+	/**
+	 * Get the item schema name for the controller.
+	 *
+	 * Used for building the API response schema.
+	 *
+	 * @return string
+	 */
+	protected function get_schema_title(): string {
+		return 'attribute_mapping';
+	}
+
+	/**
+	 * Destinations getter
+	 *
+	 * @return array The destinations
+	 */
+	private function get_destinations(): array {
+		return [
+			'data' => $this->attribute_mapping_helper->get_destinations(),
+		];
+	}
+
+	/**
+	 * Sources getter
+	 *
+	 * @param string $destination The destination to get the sources for
+	 * @return array[] Array with sources
+	 */
+	private function get_sources_for_destination( string $destination ): array {
+		$sources = $this->attribute_mapping_helper->get_sources();
+		return [
+			'data' => $sources[ $destination ] ?? [],
+		];
+	}
+}
--- a/src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
+++ b/src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
@@ -45,11 +45,11 @@ class AttributeMappingController extends BaseOptionsController {
 		 * GET the destination fields for Google Shopping
 		 */
 		$this->register_route(
-			'mc/attribute-mapping/destinations',
+			'mc/mapping/attributes',
 			[
 				[
 					'methods'             => TransportMethods::READABLE,
-					'callback'            => $this->get_attribute_mapping_destinations_read_callback(),
+					'callback'            => $this->get_mappping_attributes_read_callback(),
 					'permission_callback' => $this->get_permission_callback(),
 				],
 				'schema' => $this->get_api_response_schema_callback(),
@@ -60,12 +60,19 @@ class AttributeMappingController extends BaseOptionsController {
 		 * GET for getting the source data for a specific destination
 		 */
 		$this->register_route(
-			'mc/attribute-mapping/sources',
+			'mc/mapping/sources',
 			[
 				[
 					'methods'             => TransportMethods::READABLE,
-					'callback'            => $this->get_attribute_mapping_sources_read_callback(),
+					'callback'            => $this->get_mapping_sources_read_callback(),
 					'permission_callback' => $this->get_permission_callback(),
+					'args'                => [
+						'attribute' => [
+							'description'       => __( 'The attribute key to get the sources.', 'google-listings-and-ads' ),
+							'type'              => 'string',
+							'validate_callback' => 'rest_validate_request_arg',
+						],
+					],
 				],
 				'schema' => $this->get_api_response_schema_callback(),
 			],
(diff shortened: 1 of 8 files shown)


src/Product/Attributes/AgeGroup.php
```diff
@@ -66,4 +66,22 @@ public static function get_input_type(): string {
...
+		return __( 'Age group', 'google-listings-and-ads' );
+	}
+
+	/**
+	 * Returns the attribute sources
+	 *
+	 * @return array
+	 */
+	public static function get_sources(): array {
+		return self::get_value_options();
```
mikkamp, line 85: I realize these were already added, but since these are just labels in a dropdown, would it be helpful to include the age ranges:

```json
{
	"newborn": "Newborn (0–3 months)",
	"infant": "Infant (3–12 months)",
	"toddler": "Toddler (1–5 years)",
	"kids": "Kids (5–13 years)",
	"adult": "Adult (13+ years)"
}
```

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingControllerTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter\AttributeMappingController;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\AttributeMappingHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\RESTControllerUnitTest;
+
+/**
+ * Test suit for AttributeMappingController
```
mikkamp: Just a small typo here `suite`

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingControllerTest.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers\MerchantCenter
+ * @group AttributeMapping
```
mikkamp, line 13: I'm not against adding groups as it can help to run a specific set of tests, although I personally don't use it. Instead I run something like:

`vendor/bin/phpunit --filter "AttributeMappingControllerTest|AccountControllerTest"` or just the name of one class. Maybe a bit more typing but I can run any set of tests regardless of how they have been grouped.

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingControllerTest.php
```diff
@@ -0,0 +1,68 @@
...
+						'yes' => 'Yes',
+						'no'  => 'No',
+					],
+				],
+			);
+
+		$response = $this->do_request( self::ROUTE_REQUEST_SOURCES, 'GET', [ 'destination' => 'adult' ] );
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertNotNull( $response->get_data() );
```
mikkamp: Do we want to be a bit more specific here? At the very least we are going to want to have a response containing `data: [somearray]`. Just checking not null means it can return anything in that response.

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingControllerTest.php
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	public function test_destinations_route() {
+		$this->attribute_mapping_helper->expects( $this->once() )
+			->method( 'get_destinations' );
+
+		$response = $this->do_request( self::ROUTE_REQUEST_DESTINATIONS );
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertNotNull( $response->get_data() );
```
mikkamp: Same as comment for next test. We want to at the very least check that we are getting back an object with the data property set to an array.

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
```diff
@@ -0,0 +1,164 @@
...
+	/**
+	 * Get the item schema properties for the controller.
+	 *
+	 * @return array
+	 */
+	protected function get_schema_properties(): array {
+		return [
+			'data' => [
+				'type'        => 'array',
+				'description' => __( 'The data with the different destinations or sources.', 'google-listings-and-ads' ),
```
mikkamp: Would it be clearer to state:

> List of attributes or attribute sources.

The word "data" is very ambiguous and doesn't describe anything.

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
```diff
@@ -0,0 +1,164 @@
...
+
+	/**
+	 * Get the item schema properties for the controller.
+	 *
+	 * @return array
+	 */
+	protected function get_schema_properties(): array {
+		return [
+			'data' => [
+				'type'        => 'array',
```
mikkamp: Is this the right type to use? While in PHP the two are interchangeable, when this gets sent as a JSON response it's actually an object. Types: https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#primitive-types

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
```diff
@@ -0,0 +1,164 @@
...
+		 * GET for getting the source data for a specific destination
+		 */
+		$this->register_route(
+			'mc/attribute-mapping/sources',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_attribute_mapping_sources_read_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
```
mikkamp: If we are going to allow query arguments then we should define the schema for them with a helper function like we have done here: https://github.com/woocommerce/google-listings-and-ads/blob/2.1.1/src/API/Site/Controllers/Ads/CampaignController.php#L58

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
```diff
@@ -0,0 +1,164 @@
...
+
+	/**
+	 * Get the callback function getting destination data.
+	 *
+	 * @return callable
+	 */
+	protected function get_attribute_mapping_sources_read_callback(): callable {
+		return function( Request $request ) {
+			try {
+				$destination = $request['destination'];
```
mikkamp: While this works it's better to use the helper functions.

```suggestion
				$destination = $request->get_param( 'destination' );
```

src/Product/Attributes/Adult.php
```diff
@@ -59,4 +59,25 @@ public static function get_value_type(): string {
...
+
+	/**
+	 * Returns the attribute sources
+	 *
+	 * @return array
+	 */
+	public static function get_sources(): array {
+		return [
+			'yes' => __( 'Yes', 'google-listings-and-ads' ),
+			'no'  => __( 'No', 'google-listings-and-ads' ),
```
mikkamp, line 80: The API expects a boolean type and doesn't support `yes/no` strings. Are we going to keep all mapping values as strings and only convert to the right type when we are setting the attributes?

mikkamp (commented):
Thanks for the PR, I tested those endpoints and they are returning the results as expected. 

Although I had some second thought about the naming, to me the `destinations` name sounds confusing, I think an alternative of `attributes` would be clearer. If we don't want to repeat the word attributes in the URL we can always call it just `mapping`, ex:

```
/wp-json/wc/gla/mc/mapping/attributes 
/wp-json/wc/gla/mc/mapping/sources?attribute=ageGroup 
```

I added mainly minor comments / questions.

src/Product/Attributes/AgeGroup.php
puntope: Do you mean to update `get_value_options()` or override them in `get_sources()` ?

Notice this `get_value_options()` is also used in Product Edition - GLA Attributes

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingControllerTest.php
puntope: I Agree that we can a filter with the naming but for sets of 3 or more files it starts to be big, plus the potential typo 

With group it's just `vendor/bin/phpunit --group AttributeMapping` which is very convenient at least for me.

Since it's not harmful,  I guess we can keep it?

mikkamp: Yes, by all means keep it (especially if it's convenient for you). I was merely pointing out how I run tests since the group isn't always added.

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
puntope: TIL

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
puntope: TIL

src/Product/Attributes/Adult.php
puntope: With API I take as granted you mean Google API.

> Are we going to keep all mapping values as strings and only convert to the right type when we are setting the attributes?

Yes, that was my initial idea since Frontend will store it as strings. We can do also "true" and "false"  it that makes more sense and it's easier (maybe even automatic) to convert to boolean.

puntope: However, it would be more convenient for me to wait til the future PR to load the attributes in the Frontend selector to verify this and bet for the most elegant solution.

puntope:
Howdy @mikkamp , thanks for the review

I updated the PR with almost all of your suggestions. I guess this is ready for another round 👍

src/Product/Attributes/AgeGroup.php
mikkamp: I meant to update them in both, it will be helpful in both circumstances for selecting the right one.

src/Product/Attributes/Adult.php
mikkamp: Yes I meant the Google API. `yes/no` is good, it's what WC also uses for boolean settings that are stored in the options table. Was mainly curious, no need to change anything here.

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingControllerTest.php
```diff
@@ -0,0 +1,68 @@
...
+ * Test suite for AttributeMappingController
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers\MerchantCenter
+ * @group AttributeMapping
+ */
+class AttributeMappingControllerTest extends RESTControllerUnitTest {
+
+
+	protected const ROUTE_REQUEST_SOURCES      = '/wc/gla/mc/mapping/sources';
+	protected const ROUTE_REQUEST_DESTINATIONS = '/wc/gla/mc/mapping/attributes';
```
mikkamp: Can we rename all the destinations variables in this test to attributes to remain consistent?

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingControllerTest.php
```diff
@@ -0,0 +1,68 @@
...
+						'yes' => 'Yes',
+						'no'  => 'No',
+					],
+				],
+			);
+
+		$response = $this->do_request( self::ROUTE_REQUEST_SOURCES, 'GET', [ 'attribute' => 'adult' ] );
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertIsArray( $response->get_data() );
```
mikkamp: This is a little more specific, but since we are getting the data from a JSON response, an array can be multiple things. Since in the schema you specifically set the response to 'data' => array containing response. We can at the very least test for the `data` array key to confirm the response conforms to the expected schema.

```suggestion
		$this->assertArrayHasKey( 'data', $response->get_data() );
```

Or we could go all the way and test for an exact response.

```suggestion
		$this->assertEquals(
			[
				'data' => [
					'yes' => 'Yes',
					'no'  => 'No',
				],
			],
			$response->get_data()
		);
```

Same applies for the `test_destinations_route` function.

mikkamp (approved):
Thanks for addressing those changes. This pretty much looks ready to go, just added a few small comments and answered your questions.

PR #1658, opened 2022-08-29
Fix fatal error if a null rate specified for flat rate methods with shipping classes

### LINKED ISSUE #1657: An error occurred while using WC Shipping settings in the target audience suggestions API
### Describe the bug:

A specific WP Shipping setting will cause the `mc/target_audience/suggestions` API to respond with 500 error.

### Steps to reproduce:

1. Go to WooCommver -> Settings -> Shipping -> Shipping classes `/wp-admin/admin.php?page=wc-settings&tab=shipping&section=classes`
2. Create a shipping class. For example:
   

3. Switch to shipping zones and add a setting like the below screenshots:
   
   
4. (Optional) Run SQL to delete GLA's MC settings if want to reproduce the issue on the UI of onboarding flow: `DELETE FROM wp_options WHERE option_name IN ('gla_mc_setup_completed_at', 'gla_target_audience', 'gla_merchant_center');`.
5. Request the `mc/target_audience/suggestions` API via a tool, or proceed with the onboarding flow till step 2.

https://user-images.githubusercontent.com/17420811/187180407-19d90b5b-bd16-409f-90b6-d1eee10bfc7d.mp4

### Expected behavior:

- [ ] The API should return without a 500 error.
- [ ] The onboarding flow should be able to use.

### Actual behavior:

The onboarding flow shows a blank page when entering step 2.

nima-karimi:

Closes #1657.

This will prevent a fatal error on the shipping suggestions API (`GET mc/shipping/rates/suggestions?country_codes[]=...`), which was thrown if the user has entered a null value for their Flat rate shipping in WooCommerce shipping zone settings while specifying a rate for their shipping classes.

1. Check the steps described in #1657 and confirm that the error no longer exists.

### Changelog entry

> Fix - Fatal error if a null rate specified for flat rate methods with shipping classes

### DIFF
--- a/src/Shipping/ZoneMethodsParser.php
+++ b/src/Shipping/ZoneMethodsParser.php
@@ -71,7 +71,7 @@ class ZoneMethodsParser implements Service {
 					return [];
 				}
 
-				$shipping_rates[] = new ShippingRate( $flat_rate );
+				$shipping_rates[] = new ShippingRate( (float) $flat_rate );
 
 				if ( ! empty( $shipping_class_rates ) ) {
 					foreach ( $shipping_class_rates as ['class' => $class, 'rate' => $rate] ) {


eason9487 (approved):
Testing well. Thanks for the fix. LGTM.

PR #1662, opened 2022-08-31
Adding Attribute Mapping Page under Settings

puntope:

Closes part of #1648.

This PR creates Attribute mapping page under Settings tab.

Figma: fqR0EHi63lWahRcVTKCcba-fi-883%3A114583

### Screenshots:

Screenshot 2022-09-02 at 15 44 19

1. Complete an onboarded setup and see the new tab Attribute Mapping i
3. Click on Attributes Mapping Tab and see the new Attribute Mapping page
5. In  Attribute Mapping check that UI looks similar to Figma design
6. Enter in `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fattribute-mapping`  and see that you can also see the Attributes mapping tab


js/src/utils/urls.js
```diff
@@ -19,6 +19,7 @@ export const pagePaths = {
 };
 
 export const subpaths = {
+	attributeMapping: '/attribute-mapping',
```
eason9487: 💅 
Looks like this was missed to be removed when changing to the new design in https://github.com/woocommerce/google-listings-and-ads/pull/1662/commits/0c33769985eb17f1ce8daa95939a9f7e8793e390?

js/src/attribute-mapping/index.scss
```diff
@@ -0,0 +1,17 @@
...
+
+	&__notice {
+		background: $gla-color-light-blue;
+		margin: 0;
+		margin-bottom: $grid-unit-20;
+
+		&.is-dismissible {
+			padding-right: 10px;
+		}
+	}
```
eason9487: 💅
Looks like this CSS style and `$gla-color-light-blue` were missed to be removed when changing to the new design in https://github.com/woocommerce/google-listings-and-ads/pull/1662/commits/5d506d02bf6b45cfb303b0de47d0d91a7e6af36b?

js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -0,0 +1,78 @@
...
+						numberOfRows={ 1 }
+						rows={ [] } // TODO: Implement data getter
+					/>
+				</CardBody>
+				<CardFooter
+					align="start"
+					className="gla-settings-attribute-mapping__table-footer"
+				>
+					<AppButton
+						className="gla-settings-attribute-mapping__button"
```
eason9487: 💅 Perhaps the `className` could be renamed because this page is no longer under the Settings tab, but it still contains `-settings-`.

js/src/attribute-mapping/index.scss
```diff
@@ -0,0 +1,17 @@
...
+
+	&__table-footer {
+		justify-content: flex-start;
```
eason9487: 💅 Looks like this style could be omitted since it's already specified at:
https://github.com/woocommerce/google-listings-and-ads/blob/5cbc2f804bd11cf5afea7a06b3de2fc840d4be7b/js/src/attribute-mapping/attribute-mapping-table.js#L60

js/src/attribute-mapping/index.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+const AttributeMapping = () => {
+	return (
+		<div className="gla-settings-attribute-mapping">
+			<NavigationClassic />
+			<Section
+				title={ __( 'Attribute Mapping', 'google-listings-and-ads' ) }
+				description={ <AttributeMappingDescription /> }
+			>
+				<VerticalGapLayout size="overlap">
```
eason9487: 💅 
If the blue background notice has been removed from the requirements, this `VerticalGapLayout` could be removed as well, or at least `size="overlap"` could be removed.

eason9487 (approved):
❔ 
Not sure which mockup on Figma this PR is referencing, but some copywriting seems not to be the same as the ones on Figma. For example:
- Attribute Mapping -> Feed Management or Product Feed Management
- Add new attribute mapping -> Add new configuration
- The content on the left side of the Product Feed Management section

---

Left some minor comments for codes that were omittable or missed to be removed. Approved in advance since there are no blocker items.

puntope:
> some copywriting seems not to be the same as the ones on Figma. For example:

You're right. This is because we are still in progress on design. So copy will likely change in the future. We can address that in other PR

PR #1664, opened 2022-09-02
Attribute Mapping -  List Taxonomies and Global Attributes in sources 

puntope:

ℹ️  It includes #1656  . Please review that PR first.

Closes part of #1648.

This PR includes a helper function to get Taxonomies and Global Attributes as a source for a destination.

ℹ️  In the example we used the attribute `Color` in the future should be defined in more attributes when we now exactly the direction to follow.

### Screenshots:

Screenshot 2022-09-02 at 16 24 45

1. Do `GET /wp-json/wc/gla/mc/attribute-mapping/sources?destination=color`
2. You should get a list with attributes and taxomomies matching your store.
3. Try creating a new attribute in Products --> Attributes and see that it appears in global Attributes
4. Install a plugin like WooCommerce Brands or create a new Taxomomy to see it appears in the Taxonomies group.

### DIFF
--- a/src/Product/AttributeMappingHelper.php
+++ b/src/Product/AttributeMappingHelper.php
@@ -54,6 +54,11 @@ class AttributeMappingHelper implements Service {
 		return $sources;
 	}
 
+	/**
+	 * Gets the taxonomies and global attributes to render them as options in the frontend.
+	 *
+	 * @return array An array with the taxonomies and global attributes
+	 */
 	public static function get_source_taxonomies(): array {
 		$taxonomies = get_object_taxonomies( 'product' );
 		$taxes      = [];
--- a/src/Product/AttributeMappingHelper.php
+++ b/src/Product/AttributeMappingHelper.php
@@ -67,30 +67,42 @@ class AttributeMappingHelper implements Service {
 	 * @return array An array with the taxonomies and global attributes
 	 */
 	public static function get_source_taxonomies(): array {
-		$taxonomies = get_object_taxonomies( 'product' );
-		$taxes      = [];
-		$attributes = [];
+		$taxonomies        = get_object_taxonomies( 'product', 'objects' );
+		$parsed_taxonomies = [];
+		$attributes        = [];
+		$sources           = [];
+
 		foreach ( $taxonomies as $taxonomy ) {
-			$tax = get_taxonomy( $taxonomy );
-			if ( taxonomy_is_product_attribute( $taxonomy ) ) {
-				$attributes[ 'tax:' . $taxonomy ] = $tax->labels->name;
+			if ( taxonomy_is_product_attribute( $taxonomy->name ) ) {
+				$attributes[ 'taxonomy:' . $taxonomy->name ] = $taxonomy->label;
 				continue;
 			}
 
-			$taxes[ 'tax:' . $taxonomy ] = $tax->labels->name;
+			$parsed_taxonomies[ 'taxonomy:' . $taxonomy->name ] = $taxonomy->label;
 		}
-		asort( $taxes );
+
+		asort( $parsed_taxonomies );
 		asort( $attributes );
 
-		return array_merge(
-			[
-				'disabled:attributes' => __( '- Global attributes -', 'google-listings-and-ads' ),
-			],
-			$attributes,
-			[
-				'disabled:taxes' => __( '- Taxonomies -', 'google-listings-and-ads' ),
-			],
-			$taxes
-		);
+		if ( ! empty( $attributes ) ) {
+			$sources = array_merge(
+				[
+					'disabled:attributes' => __( '- Global attributes -', 'google-listings-and-ads' ),
+				],
+				$attributes
+			);
+		}
+
+		if ( ! empty( $parsed_taxonomies ) ) {
+			$sources = array_merge(
+				$sources,
+				[
+					'disabled:taxonomies' => __( '- Taxonomies -', 'google-listings-and-ads' ),
+				],
+				$parsed_taxonomies
+			);
+		}
+
+		return $sources;
 	}
 }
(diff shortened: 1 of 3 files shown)


src/Product/AttributeMappingHelper.php
```diff
@@ -0,0 +1,89 @@
...
+		asort( $taxes );
+		asort( $attributes );
+
+		return array_merge(
+			[
+				'disabled:attributes' => __( '- Global attributes -', 'google-listings-and-ads' ),
+			],
+			$attributes,
+			[
+				'disabled:taxes' => __( '- Taxonomies -', 'google-listings-and-ads' ),
```
mikkamp: Making the abbreviation `tax` plural doesn't really work here as it has a different meaning. I'd suggest to stick with `taxonomies`. Would suggest the same for the variable `$taxes` as it's a bit confusing when reading the code.

src/Product/AttributeMappingHelper.php
```diff
@@ -0,0 +1,89 @@
...
+	 * Gets the taxonomies and global attributes to render them as options in the frontend.
+	 *
+	 * @return array An array with the taxonomies and global attributes
+	 */
+	public static function get_source_taxonomies(): array {
+		$taxonomies = get_object_taxonomies( 'product' );
+		$taxes      = [];
+		$attributes = [];
+		foreach ( $taxonomies as $taxonomy ) {
+			$tax = get_taxonomy( $taxonomy );
```
mikkamp: This function call can return false, which if it happens will cause errors since `$tax` won't be an object. I'd suggest to use the second parameter in `get_object_taxonomies` to retrieve all the objects outside of the loop.

src/Product/AttributeMappingHelper.php
```diff
@@ -0,0 +1,89 @@
...
+		return $sources;
+	}
+
+	/**
+	 * Gets the taxonomies and global attributes to render them as options in the frontend.
+	 *
+	 * @return array An array with the taxonomies and global attributes
+	 */
+	public static function get_source_taxonomies(): array {
+		$taxonomies = get_object_taxonomies( 'product' );
```
mikkamp: I'd personally recommend to use `get_taxonomies` here, although it seems to have some issues which is why `get_object_taxonomies` is recommended. The advantage of the `get_taxonomies` call is that a filter can be used. For example product type and product visibility are both non public taxonomies. So I think it would be better to exclude non public ones. It can still be done by comparing the public parameter though.

src/Product/AttributeMappingHelper.php
```diff
@@ -0,0 +1,89 @@
...
+				$attributes[ 'tax:' . $taxonomy ] = $tax->labels->name;
+				continue;
+			}
+
+			$taxes[ 'tax:' . $taxonomy ] = $tax->labels->name;
+		}
+		asort( $taxes );
+		asort( $attributes );
+
+		return array_merge(
```
mikkamp: This list can potentially get really long. What kind of UI elements are we going to have for this? Does it include matching upon typing the word. Or would it be better to pick the most recommended attribute from the list.

For example I have a product attribute called `color` but I don't want to scroll through a large list of attributes to find it. If an attribute with matching name is found, should we pre-select it or bump it to the top of the list?

mikkamp (commented):
I'm getting the results as expected when testing:

```json
{
	"data": {
		"disabled:attributes": "- Global attributes -",
		"tax:pa_aut": "Product Aut",
		"tax:pa_color": "Product Color",
		"tax:pa_numeric-size": "Product Numeric Size",
		"tax:pa_quidem": "Product Quidem",
		"tax:pa_size": "Product Size",
		"disabled:taxes": "- Taxonomies -",
		"tax:product_cat": "Product categories",
		"tax:product_shipping_class": "Product shipping classes",
		"tax:product_tag": "Product tags",
		"tax:product_type": "Product type",
		"tax:product_visibility": "Product visibility"
	}
}
```

I left some comments on a few points I noticed.

src/Product/AttributeMappingHelper.php
puntope: Unfortunately as far as I know we don't receive yet from Google /.UX which attributes and sources we should have available.

This is actually how Google Feed manages this... so thats why I used this strategy.

>  What kind of UI elements are we going to have for this? Does it include matching upon typing the word.

Initially... Just a standard selector without search as I see in Figma... but we are still on i2 so. It could be a suggestion to add search in the selector. 

> Or would it be better to pick the most recommended attribute from the list.

This is a cool solution in my opinion to have a "Recommended" section in the selector. 

I suggest to elevate this to P2 discussion and modify this function in a future PR if needed.

Thoughts?

src/Product/AttributeMappingHelper.php
puntope: Interesting... when I use get_taxonomies with `public` I miss the product attributes (some of them created by myself)

Screenshot 2022-09-16 at 11 28 12

However without `public` it gives me the attributes:

Screenshot 2022-09-16 at 11 26 33

Is there an issue by offering the user the possibility of  using the non public ones as source?

puntope: Another strange output is that in `get_object_taxonomies` I get some extra taxes like shipping_class or product_visibility

Screenshot 2022-09-16 at 11 34 29

Even though, using `get_taxonomies` (without public attribute set as true) doesn't give me that ones...

mikkamp: We have to use `get_object_taxonomies` because of this issue: https://core.trac.wordpress.org/ticket/27918 (note: temporarily unavailable).

But I meant since we are going through the objects in a loop we can continue if `! $taxonomy->public`
It's not a big issue to include public, but I think it returns a cleaner list since I can't think of any specific cases where you would need to use product_type or product_visibility.

src/Product/AttributeMappingHelper.php
mikkamp: Elevate to P2 and future PR sounds like a good plan.

puntope:
Howdy @mikkamp  I added your suggestions and answered your questions.

Also I added some check for preventing showing the disabled entries in case there is no attributes/taxomonies returned.

src/Product/AttributeMappingHelper.php
puntope: > But I meant since we are going through the objects in a loop we can continue if ! $taxonomy->public

I tried to do so... but the attributes disappeared.

mikkamp: In my testing I added the check for public after the check `taxonomy_is_product_attribute`, so it would still add the product attributes.

But either way my main intent was to filter out product_type and product_visibility. So I'll go ahead and approve this PR and we can tweak later if needed.

mikkamp (approved):

PR #1665, opened 2022-09-05
Fix an infinite re-rendering loop for the campaign budget input component

### LINKED ISSUE #1652: WC 6.9 Compatibility: Cannot type daily budget cost when creating paid campaigns
### Describe the bug:

Using [WooCommerce 6.9.0-beta.2](https://developer.woocommerce.com/2022/08/24/woocommerce-6-9-beta-2/), the input filed of the daily average budget cost cannot be typed in any numbers when creating a paid campaign, while using WooCommerce `6.8.2` works as expected.

### Steps to reproduce:

1. Use GLA 2.1.0 and WooCommerce 6.9.0-beta.2, download from [here](https://developer.woocommerce.com/2022/08/24/woocommerce-6-9-beta-2/).
2. Finish the onboarding.
3. In GLA dashboard, click `Add paid campaign`.
4. Connect Ads account if needed.
5. Type anything in `Daily average cost` input field.
6. You can see it's not editable.

### Expected behavior:

The budget input field should be editable.

Video using WC `6.8.2`:

https://user-images.githubusercontent.com/914406/186626962-76bdb32d-6728-48d8-9213-12c563b30983.mov

### Actual behavior:

The budget input field is not editable.

Video using WC `6.9.0-beta.2`:

https://user-images.githubusercontent.com/914406/186627652-8c4248c2-f1f2-461a-8c46-9b1d03fd670b.mov

eason9487:

Closes #1652 

- Fix an infinite re-rendering loop for the campaign budget input component

### Screenshots:

https://user-images.githubusercontent.com/17420811/188428012-d974f233-b7c9-4941-b9f1-e80e45896770.mp4

#### Root cause

With the `@woocommerce/components`'s \<Form\> component provided by WC 6.9-beta.2, the reference changing of `setValue` function got from \<Form\> will not be the same as before. When \<Form\> values are changed, the reference of `setValue` is also changed. Therefore, it may lead to an infinite re-rendering loop when using it in such case:
1. Have `setValue` in a `useEffect` and it's also one of `useEffect` deps.
1. Call `setValue` to change form values, and then the form passes down a new reference of `setValue` function.
1. The change of `setValue` function triggers that `useEffect` again, and then loops back to the above to call the `setValue` to change form values again.

See: https://github.com/woocommerce/woocommerce/blob/78c28ae9f3a70e996e40c4f173c032dfa0d0551a/packages/js/components/src/form/form.tsx#L176

💡 I searched the codebase of this repo, and it should be the only case that has this issue.

1. Use a WC version under 6.9.0
3. Follow the replication steps in #1652 to test the input of campaign budget.
4. Switch to WC 6.9.0-beta.2
5. Repeat the replication steps again
6. Check if this fix is compatible with both WC versions.

### Changelog entry

> Fix - A compatibility issue with WooCommerce 6.9 to prevent the input of the paid campaign budget from being not interactable.

### DIFF
--- a/js/src/components/paid-ads/budget-section/index.js
+++ b/js/src/components/paid-ads/budget-section/index.js
@@ -38,6 +38,12 @@ const BudgetSection = ( props ) => {
 	// Display the currency code that will be used by Google Ads, but still use the store's currency formatting settings.
 	const currency = googleAdsAccount?.currency;
 
+	// Wrapping `useRef` is because since WC 6.9, the reference of `setValue` may be changed
+	// after calling itself and further leads to an infinite re-rendering loop if used in a
+	// `useEffect`.
+	const setValueRef = useRef();
+	setValueRef.current = setValue;
+
 	/**
 	 * In addition to the initial value setting during initialization, when `disabled` changes
 	 * - from false to true, then clear filled amount to `undefined` for showing a blank <input>.
@@ -46,8 +52,8 @@ const BudgetSection = ( props ) => {
 	const initialAmountRef = useRef( amount );
 	useEffect( () => {
 		const nextAmount = disabled ? undefined : initialAmountRef.current;
-		setValue( 'amount', nextAmount );
-	}, [ disabled, setValue ] );
+		setValueRef.current( 'amount', nextAmount );
+	}, [ disabled ] );
 
 	return (
 		<div className="gla-budget-section">


ianlin (approved):
Thanks for fixing this, it works as expected for both WC 6.8 and WC 6.9.

❔ Looks like t's caused by the `<Form>` component was changed from a class component to a functional component, and `setValue`'s dependencies has `values` state so the reference of `setValue` would be changed every time it's being called. Just curious is this inevitable when using functional components with state? Or maybe there would be a better way to implement `<Form>` component?

eason9487:
@ianlin, thanks for the review.

> Looks like t's caused by the `<Form>` component was changed from a class component to a functional component, and `setValue`'s dependencies has `values` state so the reference of `setValue` would be changed every time it's being called.

Yes.

> Just curious is this inevitable when using functional components with state? Or maybe there would be a better way to implement `<Form>` component?

Usually either a class component or a functional component, the function interface of a fundamental component will be a stable reference, changing only when necessary, but this isn't a mandatory rule for React component design. It can be done on the Form side by wrapping needed states with `useRef` internally.

PR #1672, opened 2022-09-14
Free Listings + Paid Ads: Fix a compatibility issue with the Form component in WC 6.9

eason9487:

It relates to #1610 

The `<Form ref={ formRef }>` use was added in 93249e65336667bc7372bb66973587440323ca92 (#1616) and was the only one use case. However, WC 6.9 changes the exposed interfaces of `Form` component, so it needs another way to do the same logic.

This PR fixes the exception in the target audience setup of onboarding flow.

image

### Screenshots:

https://user-images.githubusercontent.com/17420811/190137850-a7f02161-6920-402c-bb91-12ec81f437a1.mp4

1. Use WC 6.9
2. Go to step 2 of the onboarding flow.
3. Remove a target audience country that has set the relevant shipping rate and time.
4. Check if the removed country is also removed from the shipping rate and time settings.

### DIFF
--- a/js/src/components/free-listings/setup-free-listings/index.js
+++ b/js/src/components/free-listings/setup-free-listings/index.js
@@ -88,7 +88,7 @@ const SetupFreeListings = ( {
 	submitLabel,
 } ) => {
 	const [ saving, setSaving ] = useState( false );
-	const formRef = useRef();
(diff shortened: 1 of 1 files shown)


puntope (approved):
✅ LGTM 

Tested locally

PR #1681, opened 2022-09-20
Free Listings + Paid Ads: Update product feed API to return the bestselling product

ianlin:

This PR implements the [backend task 4](https://github.com/woocommerce/google-listings-and-ads/issues/1610#be-task4) **_📌 Return merchants key product data (e.g. bestselling)_** from the epic issue #1610. Here are what have changed:

- Extend an existing API `mc/product-feed` by adding a new condition `total_sales` to the parameter `orderby`.
  - `total_sales` is one of the meta keys of a product.
  - Combining with other parameters we can get the product which has the most sales. E.g.
    - `GET mc/product-feed?per_page=1&orderby=total_sales&order=desc`
- Extend an existing API `mc/product-feed` to return price and image url of a product.
- Add `siteLogoUrl` to `glaData` for the frontend.

1. Make a GET request to `mc/product-feed` API with parameters:
    - per_page=1
    - orderby=total_sales
    - order=desc
    - Full example: `GET mc/product-feed?per_page=1&orderby=total_sales&order=desc`
2. The response should contain only one product, with new response data `price` and `image_url`.
3. To verify if the returned product has the most sales, use this query in your local database:
    - `SELECT * FROM wp_postmeta WHERE meta_key = 'total_sales' ORDER BY CAST( meta_value AS unsigned ) DESC;`
4. Go to GLA dashboard, open the dev tool console, and type `window.glaData`
    - You should see there is a new data `siteLogoUrl`.
    - It would be `null` if you did not upload a custom site logo.
5. If the `siteLogoUrl` is `null`, go to `wp-admin/admin.php?page=wc-admin&task=appearance` and upload a logo.
6. Open the dev tool console and type `window.glaData` again, you should see the value of `siteLogoUrl`.

### DIFF
--- a/src/DB/ProductFeedQueryHelper.php
+++ b/src/DB/ProductFeedQueryHelper.php
@@ -88,11 +88,13 @@ class ProductFeedQueryHelper implements ContainerAwareInterface, Service {
 			$errors = $product_helper->get_validation_errors( $product );
 
 			$products[ $id ] = [
-				'id'      => $id,
-				'title'   => $product->get_name(),
-				'visible' => $product_helper->get_channel_visibility( $product ) !== ChannelVisibility::DONT_SYNC_AND_SHOW,
-				'status'  => $product_helper->get_mc_status( $product ) ?: $product_helper->get_sync_status( $product ),
-				'errors'  => array_values( $errors ),
+				'id'        => $id,
+				'title'     => $product->get_name(),
+				'visible'   => $product_helper->get_channel_visibility( $product ) !== ChannelVisibility::DONT_SYNC_AND_SHOW,
+				'status'    => $product_helper->get_mc_status( $product ) ?: $product_helper->get_sync_status( $product ),
+				'image_url' => wp_get_attachment_image_url( $product->get_image_id(), 'full' ),
+				'price'     => $product->get_price(),
+				'errors'    => array_values( $errors ),
 			];
 		}
 
--- a/src/DB/ProductFeedQueryHelper.php
+++ b/src/DB/ProductFeedQueryHelper.php
@@ -165,6 +165,10 @@ class ProductFeedQueryHelper implements ContainerAwareInterface, Service {
 				$args['meta_key'] = $this->prefix_meta_key( ProductMetaHandler::KEY_MC_STATUS );
 				$args['orderby']  = [ 'meta_value' => $this->get_order() ] + $args['orderby'];
 				break;
+			case 'total_sales':
+				$args['meta_key'] = ProductMetaHandler::KEY_TOTAL_SALES;
+				$args['orderby']  = [ 'meta_value_num' => $this->get_order() ] + $args['orderby'];
+				break;
 			default:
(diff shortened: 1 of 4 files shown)


src/DB/ProductFeedQueryHelper.php
```diff
@@ -165,6 +167,10 @@ protected function prepare_query_args(): array {
 				$args['meta_key'] = $this->prefix_meta_key( ProductMetaHandler::KEY_MC_STATUS );
 				$args['orderby']  = [ 'meta_value' => $this->get_order() ] + $args['orderby'];
 				break;
+			case 'total_sales':
+				$args['meta_key'] = ProductMetaHandler::KEY_TOTAL_SALES;
+				$args['orderby']  = [ 'meta_value_num' => $this->get_order() ] + $args['orderby'];
+				break;
 			default:
 				throw InvalidValue::not_in_allowed_list( 'orderby', [ 'title', 'id', 'visible', 'status' ] );
```
mikkamp: We should add `total_sales` as that's an allowed value.

src/DB/ProductFeedQueryHelper.php
```diff
@@ -165,6 +167,10 @@ protected function prepare_query_args(): array {
 				$args['meta_key'] = $this->prefix_meta_key( ProductMetaHandler::KEY_MC_STATUS );
 				$args['orderby']  = [ 'meta_value' => $this->get_order() ] + $args['orderby'];
 				break;
+			case 'total_sales':
+				$args['meta_key'] = ProductMetaHandler::KEY_TOTAL_SALES;
```
mikkamp: I personally think we should just use a constant string of `total_sales` here. I don't think it really fits within the `ProductMetaHandler` class as that's to manage product meta fields we add for the extension, also it means we can call update/get/delete for this field, but it won't target the right entry because it will be adding the prefix. So it can be removed from the `ProductMetaHandler` class.

mikkamp (approved):
Thanks for the changes, I'm getting the response as expected.

```json
{
	"products": [
		{
			"id": 1695,
			"title": "Incredible Steel Coat",
			"visible": true,
			"status": "not_synced",
			"errors": []
		}
	],
	"total": 316,
	"page": 1
}
```

Just double checking if the product status makes a difference? In my case it picks a product that's not synced (excluded).

I just left a few smaller comments but they aren't blocking so I'll go ahead and approve the PR.

ianlin:
@mikkamp 
> Just double checking if the product status makes a difference? In my case it picks a product that's not synced (excluded).

I think for most new users when they reach that step of the onboarding flow their products haven't been synced to GMC, so filtering only for synced products might have some problems.

ianlin:
The original base branch this PR was trying to merge into was `update/1610-add-api-to-get-sync-ready-products`. However due to PR #1667 [will be closed](https://github.com/woocommerce/google-listings-and-ads/pull/1667#issuecomment-1253523564) so I changed the base branch to `feature/1610-streamlined-onboarding`.

PR #1682, opened 2022-09-20
Attribute Mapping - Add API Enums

puntope:

Closes part of #1648 .

This PR adds enum prop in the API Response

Also registers all the ENUMS in the Available Attributes 

1. Checkout PR and request `GET /wp-json/wc/gla/mc/mapping/attributes`
2. See you receive all the ENUM values with enum prop set as true.
3. Check each of them with `GET wp-json/wc/gla/mc/mapping/sources?attribute={attribute_name}` and see they return the enum values

### DIFF
--- a/src/Product/Attributes/SizeType.php
+++ b/src/Product/Attributes/SizeType.php
@@ -67,4 +67,31 @@ class SizeType extends AbstractAttribute implements WithValueOptionsInterface {
 		return SizeTypeInput::class;
 	}
 
+	/**
+	 * Returns the attribute name
+	 *
+	 * @return string
+	 */
+	public static function get_name(): string {
+		return __( 'Size Type', 'google-listings-and-ads' );
+	}
+
+	/**
+	 * Returns true if the attribute is an enum
+	 *
+	 * @return bool
+	 */
+	public static function is_enum(): bool {
+		return true;
+	}
+
+	/**
+	 * Returns the attribute sources
+	 *
+	 * @return array
+	 */
+	public static function get_sources(): array {
+		return self::get_value_options();
+	}
+
 }
--- a/src/Product/Attributes/SizeType.php
+++ b/src/Product/Attributes/SizeType.php
@@ -14,6 +14,8 @@ defined( 'ABSPATH' ) || exit;
  */
 class SizeType extends AbstractAttribute implements WithValueOptionsInterface, WithMappingInterface {
 
+	use IsEnumTrait;
+
 	/**
 	 * Returns the attribute ID.
 	 *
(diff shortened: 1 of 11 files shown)


src/Product/Attributes/WithMappingInterface.php
```diff
@@ -19,6 +19,13 @@ interface WithMappingInterface {
 	 */
 	public static function get_name(): string;
 
+	/**
+	 * Returns true if the attribute is enum type
+	 *
+	 * @return boolean
+	 */
+	public static function is_enum(): bool;
```
mikkamp, line 27: Wouldn't it be cleaner to define the function `is_enum` in a trait. So each class can both implement WithMappingInterface and use IsEnumTrait. That way we won't need to repeat the function in each class. We can also create an IsNotEnumTrait but that can be added when needed (as it doesn't seem required for this PR).

Ex:
```php
class SizeType extends AbstractAttribute implements WithValueOptionsInterface, WithMappingInterface {

	use IsEnumTrait;
	...
```

src/Product/Attributes/SizeType.php
```diff
@@ -67,4 +67,31 @@ public static function get_input_type(): string {
...
+		return true;
+	}
+
+	/**
+	 * Returns the attribute sources
+	 *
+	 * @return array
+	 */
+	public static function get_sources(): array {
+		return self::get_value_options();
```
mikkamp, line 87: I realize these values were already present, but I just noticed that they are different then the [help article](https://support.google.com/merchants/answer/6324497?hl=en). Do we need to double check with Google to make sure all the enums have the right set of values?

mikkamp (commented):
Looks good and I'm getting the results I'd expect. I just left a comment about using a trait so we don't have to repeat the same function in each enum class. And also noticed a small difference in the list of values.

src/Product/Attributes/SizeType.php
puntope: Makes sense to double check. Shall I do that in the questions doc?

src/Product/Attributes/WithMappingInterface.php
puntope: Goos suggestion @mikkamp 

I added the refactor here: https://github.com/woocommerce/google-listings-and-ads/pull/1682/commits/6a546b8c9b73bcca18d67a3d3e452540103fded1

src/Product/Attributes/SizeType.php
mikkamp: Yes that would seem like the best place.

mikkamp (approved):
Thanks for the additional refactor, that's working well so I'll go ahead and approve this PR.

As mentioned we can still add the difference in `size_type` to the Google Q&A doc, but any tweaks needed there can be handled in a follow up PR.

src/Product/Attributes/SizeType.php
puntope: Done

PR #1683, opened 2022-09-20
Free Listings + Paid Ads: Store and restore the states of paid ads setup

eason9487:

This PR implements two items of the **📌 Client states management and API integration** in #1610.

- Store/restore the selected countries and the entered budget when leaving/entering step 4.
- Store/restore the switched state of skipping or continuing the paid ads setup when leaving/entering step 4.

### Screenshots:

https://user-images.githubusercontent.com/17420811/191235325-3eeb6a0b-92cf-412e-b8cd-834842dae927.mp4

1. Go to step 4 of onboarding flow.
2. Continue the paid ads setup by clicking on the "Create a paid ad campaign" button.
3. Refresh page to see if the paid ads setup keeps showing after page is loaded.
4. Edit the countries of ads audience and campaign budget.
5. Refresh page to see if the countries of ads audience and campaign budget are the same as in step 4 after page is loaded.
6. Switch to step 2 and remove some countries, that were selected as **ads** audience, from **target** audience.
7. Continue to step 4 and check if the removed target audience countries are also removed from the selected countries of ads audience.

### DIFF
--- a/js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
+++ b/js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
@@ -35,24 +35,16 @@ const clientSession = {
 	 * @param {Array<CountryCode>} data.countryCodes Country codes of the campaign.
 	 */
 	setCampaign( { amount, countryCodes } ) {
-		try {
-			const json = JSON.stringify( { amount, countryCodes } );
-			sessionStorage.setItem( KEY_PAID_ADS, json );
-		} catch ( e ) {
-			// noop
-		}
+		const json = JSON.stringify( { amount, countryCodes } );
+		sessionStorage.setItem( KEY_PAID_ADS, json );
 	},
 
 	/**
 	 * @return {CampaignData|null} The stored campaign data. It will return `null` if stored data is not available.
 	 */
 	getCampaign() {
-		try {
-			const json = sessionStorage.getItem( KEY_PAID_ADS );
-			return JSON.parse( json );
-		} catch ( e ) {
-			return null;
-		}
+		const json = sessionStorage.getItem( KEY_PAID_ADS );
+		return json === null ? null : JSON.parse( json );
 	},
 };
 
--- a/js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
+++ b/js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
@@ -1,0 +1,59 @@
+/**
+ * @typedef { import(".~/data/actions").CountryCode } CountryCode
+ *
+ * @typedef {Object} CampaignData
+ * @property {number|undefined} amount Daily average cost of the paid ads campaign.
+ * @property {Array<CountryCode>} countryCodes Audience country codes of the paid ads campaign. Example: 'US'.
+ */
+
+const KEY_SHOW_PAID_ADS_SETUP = 'gla-onboarding-show-paid-ads-setup';
+const KEY_PAID_ADS = 'gla-onboarding-paid-ads';
+
+const { sessionStorage } = window;
+
+const clientSession = {
+	/**
+	 * @param {boolean} isShowing Whether the paid ads setup is showing.
+	 */
+	setShowPaidAdsSetup( isShowing ) {
+		const showing = Boolean( isShowing ).toString();
+		sessionStorage.setItem( KEY_SHOW_PAID_ADS_SETUP, showing );
+	},
+
+	/**
+	 * @param {boolean} defaultValue The default value to be returned if stored value is not available.
+	 * @return {boolean} Returns the stored value. It will return `defaultValue` if stored value is not available.
+	 */
+	getShowPaidAdsSetup( defaultValue ) {
+		const showing = sessionStorage.getItem( KEY_SHOW_PAID_ADS_SETUP );
+		return showing === null ? defaultValue : showing === 'true';
+	},
+
+	/**
+	 * @param {CampaignData} data Campaign data to be stored.
+	 * @param {number|undefined} data.amount Daily average cost of the campaign.
+	 * @param {Array<CountryCode>} data.countryCodes Country codes of the campaign.
+	 */
+	setCampaign( { amount, countryCodes } ) {
+		try {
+			const json = JSON.stringify( { amount, countryCodes } );
+			sessionStorage.setItem( KEY_PAID_ADS, json );
+		} catch ( e ) {
+			// noop
+		}
+	},
+
+	/**
+	 * @return {CampaignData|null} The stored campaign data. It will return `null` if stored data is not available.
+	 */
+	getCampaign() {
+		try {
+			const json = sessionStorage.getItem( KEY_PAID_ADS );
+			return JSON.parse( json );
+		} catch ( e ) {
+			return null;
+		}
+	},
+};
+
+export default clientSession;
--- a/js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
+++ b/js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
@@ -40,6 +40,7 @@ const defaultPaidAds = {
 
 /**
  * Resolve the initial paid ads data from the given paid ads data with the loaded target audience.
+ * Parts of the resolved data are used in the `initialValues` prop of `Form` component.
  *
  * @param {PaidAdsData} paidAds The paid ads data as the base to be resolved with other states.
  * @param {Array<CountryCode>} targetAudience Country codes of selected target audience.
@@ -50,6 +51,8 @@ function resolveInitialPaidAds( paidAds, targetAudience ) {
 
 	if ( targetAudience ) {
 		if ( nextPaidAds.countryCodes === defaultPaidAds.countryCodes ) {
+			// Replace the country codes with the loaded target audience only if the reference is
+			// the same as the default because the given country codes might be the restored ones.
 			nextPaidAds.countryCodes = targetAudience;
 		} else {
 			// The selected target audience may be changed back and forth during the onboarding flow.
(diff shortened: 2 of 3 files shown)


js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
```diff
@@ -0,0 +1,59 @@
...
+	 * @param {number|undefined} data.amount Daily average cost of the campaign.
+	 * @param {Array<CountryCode>} data.countryCodes Country codes of the campaign.
+	 */
+	setCampaign( { amount, countryCodes } ) {
+		try {
+			const json = JSON.stringify( { amount, countryCodes } );
+			sessionStorage.setItem( KEY_PAID_ADS, json );
+		} catch ( e ) {
+			// noop
+		}
```
tomalec: :question: 
Why do we need this `try...catch` here? What kind of errors do we want to silence?

js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
```diff
@@ -0,0 +1,59 @@
...
+	/**
+	 * @return {CampaignData|null} The stored campaign data. It will return `null` if stored data is not available.
+	 */
+	getCampaign() {
+		try {
+			const json = sessionStorage.getItem( KEY_PAID_ADS );
+			return JSON.parse( json );
+		} catch ( e ) {
+			return null;
+		}
```
tomalec: :question: Do we expect some errors here? or is this `try..catch` just to cover lack of data, like `return json &&  JSON.parse( json );`?

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -30,6 +31,39 @@ import {
...
+	isValid: false,
+	isReady: false,
+};
+
+/**
+ * Resolve the initial paid ads data from the given paid ads data with the loaded target audience.
+ *
+ * @param {PaidAdsData} paidAds The paid ads data as the base to be resolved with other states.
+ * @param {Array<CountryCode>} targetAudience Country codes of selected target audience.
+ * @return {PaidAdsData} The resolved paid ads data.
```
tomalec: :question: I'm not 100% sure I understand what and why happens here. I think I may be misled by the _"initial"_ here.

Do I get it right that?:
- This function is meant to resolve the new state of paid ads countries, after the target audience changes
- `paidAds` argument is the current/previous state
- The state is "initial" from the perspective of a customer, that (re-)visits the "Complete your campaign" step. But is not initial, but rather "new"/"next" from the perspective of the data management underneath? Or more precisely it's "initial" from the page load/render perspective, as for the user's session it was initially different, and this is also, yet another state.
Or maybe it was called "initial" just to match the `Form`'s param name?

tomalec (commented):
Tested locally and works as expected.
However, I left some questions regarding the code

js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
eason9487: > Why do we need this `try...catch` here?

To tolerate unexpected/unknown errors, I sometimes add a `try catch` when converting JSON.

> What kind of errors do we want to silence?

The unexpected/unknown error can not be recovered but the failure doesn't make the paid ads setup unworkable, so I left it to silence.

---

Considering this isn't common util for the whole codebase but only for the paid ads setup, it should be fine without the `try catch`. Removed the `try catch` in https://github.com/woocommerce/google-listings-and-ads/pull/1683/commits/239b45f78e36fa23addf1294b0bdc98c608a12f0

js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
eason9487: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/1683#discussion_r976021485 and removed the `try catch` in https://github.com/woocommerce/google-listings-and-ads/commit/239b45f78e36fa23addf1294b0bdc98c608a12f0

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
eason9487: > * This function is meant to resolve the new state of paid ads countries, after the target audience changes

To be more precise a little bit, it only intends to resolve the loaded `targetAudience`. When the `PaidAdsSetupSections` is rendered with paid ads setup step, the loaded `targetAudience` won't be changed in the runtime.

> * The state is "initial" from the perspective of a customer, [...] Or more precisely it's "initial" from the page load/render perspective, as for the user's session it was initially different, and this is also, yet another state. 
>   Or maybe it was called "initial" just to match the `Form`'s param name?

It resolves the starting state of the `const [ paidAds, setPaidAds ] = useState( ...` with the stored campaign data, and also resolves the `paidAds` state via the below `useEffect` if the `targetAudience` is loaded after the `PaidAdsSetupSections` component has been already rendered. And the main purpose of this function is to prepare the value for the `initialValues` prop of `Form`.

---

I added and rephrased the code comments in https://github.com/woocommerce/google-listings-and-ads/pull/1683/commits/9f583b56dd2c3274732e8aa6f217e67e9cfe0e16. Let me know if it's still confusing or unclear.

eason9487:
@tomalec, thanks for the code review. I added some adjustments based on the comments. Could you help with a new round of code reviews?

js/src/setup-mc/setup-stepper/setup-paid-ads/clientSession.js
tomalec: :) Thanks for the explanations. I think as you mentioned this is local util, that's us to provide `amount, countryCodes` and we have no actual means to fall back from such error. I'd rather not silence the error and fails fast and explicitly, so it would be easier to identify and report for HEs

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -30,6 +31,42 @@ import {
...
 
+const defaultPaidAds = {
+	amount: 0,
+	countryCodes: [],
+	isValid: false,
+	isReady: false,
+};
+
+/**
+ * Resolve the initial paid ads data from the given paid ads data with the loaded target audience.
```
tomalec, line 42: :nail_care: WDYT of
```suggestion
 * Resolve the updated state of paid ads data from the previous state and the loaded target audience.
```
is it accurate? sound better or worse?

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
tomalec: > When the `PaidAdsSetupSections` is rendered with paid ads setup step, the loaded `targetAudience` won't be changed in the runtime.

This is true as we now have it in separate steps. If we hypothetically happen to place those two components side-by-side, this may be changed in the runtime. But that shouldn't matter. What matters here is that from `PaidAdsSetupSections` perspective and per current implementation, the `targetAudience` is changed.

> It resolves the starting state of the const `[ paidAds, setPaidAds ] = useState( ...` with the stored campaign data, and also resolves the `paidAds` state via the below `useEffect` if the `targetAudience` is loaded after the `PaidAdsSetupSections` component has been already rendered. And the main purpose of this function is to prepare the value for the `initialValues` prop of `Form`.

So If I get it right, it's not "initial" for us - `PaidAdsSetupSections`. For `PaidAdsSetupSections` it's the actual - current state. It is "initial" only for `Form`.

:nail_care: Then maybe we could rename it to just `resolvePaidAds( previousState, targetAudience )` and use "initial" word only  in `<Form initialValues={
		countryCodes: paidAds.countryCodes,
		amount: paidAds.amount,
	};`.
	
But I'm fine with it as is :)

tomalec (approved):
Thank you for the explanations & clarifications :) LGTM

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
eason9487: > This is true as we now have it in separate steps. If we hypothetically happen to place those two components side-by-side, this may be changed in the runtime. But that shouldn't matter. What matters here is that from `PaidAdsSetupSections` perspective and per current implementation, the `targetAudience` is changed.

Ideally, it should be nicer to make the design of a component from its perspective rather than serving some hidden business logic across several components.

But based on the state management of Form component and its consumer side are separated, if there is a changing external-state must update a form's internal-state accordingly, then this needs additional processing according to the business logic.

Therefore, if the separate steps happen to be placed together and the external state would be changed in the runtime, then it would need something similar to the following processing to handle the runtime change of `targetAudience`. Otherwise, the calling of `setPaidAds` in [this useEffect](https://github.com/woocommerce/google-listings-and-ads/blob/9f583b56dd2c3274732e8aa6f217e67e9cfe0e16/js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js#L120-L124) may make unsynced states between this component and its Form.

https://github.com/woocommerce/google-listings-and-ads/blob/5b6522ca10ad75556e6b2de7c120cc712aab70b1/js/src/components/free-listings/setup-free-listings/index.js#L120-L134

https://github.com/woocommerce/google-listings-and-ads/blob/5b6522ca10ad75556e6b2de7c120cc712aab70b1/js/src/components/free-listings/setup-free-listings/index.js#L172-L186

Hence, I think keeping the "initial" word would still be a relatively accurate description for now, since here it's not expected that the loaded `targetAudience` will change in the runtime. I added a note in 5572a4f7d646059e6618da72b6a2a6994fac314b in case the runtime change happens someday.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
eason9487: Please refer to https://github.com/woocommerce/google-listings-and-ads/pull/1683#discussion_r979548972.

PR #1687, opened 2022-09-21
Attribute Mapping - Add logic for Saving Rules in Database with API

puntope:

Closes part of #1648 .

This PR adds the logic for:

- Database setup for `gla_attribute_mapping_rules` table
- API endpoint for getting the rules from database
- API endpoint for insert a rule in database
- API endpoint for update a rule in database
- API endpoint for delete a rule in database
- Tests

⚠️  Pagination / limit / offsets in `get_rules` will be added in future PRs

⚠️  Is important to run the table installer.  The value of `WC_GLA_VERSION` is higher than your value in `gla_db_version` (you can increase `WC_GLA_VERSION` in the plugin file or decrease `gla_db_version` in `wp_options` table

1. Once this is done. Enter in dashboard in GLA and check `wp_gla_attribute_mapping_rules` is created  with the attributes like this columns (id, attribute, source, category_condition_type, categories) 
2. Request to `GET /wp-json/wc/gla/mc/mapping/rules` and see that you get an empty array
3. Request to `POST /wp-json/wc/gla/mc/mapping/rule` adding this in the body:

```
{
    "rule": { "attribute": "adult", "source": "yes", "category_condition_type": "ALL" }
}
```

You should get the rule back with an autogenerated ID.

4. Request to `GET /wp-json/wc/gla/mc/mapping/rules` and see that you get the rule you created
5.  Request to `POST /wp-json/wc/gla/mc/mapping/rule` adding this in the body:

```
{
    "rule": { "id": 1,  "attribute": "adult", "source": "no", "category_condition_type": "ALL" }
}
```

You should get the rule back modified. (source is now "no")

6. Request to `GET /wp-json/wc/gla/mc/mapping/rules` and see that you get the rule modified
7. Request to `DELETE /wp-json/wc/gla/mc/mapping/rule?rule_id=1`. You should get `True`
8. Request to `GET /wp-json/wc/gla/mc/mapping/rules` and see that you get empty array
9. Request to `DELETE /wp-json/wc/gla/mc/mapping/rule?rule_id=1`. You should get `False`
10. Request to `DELETE /wp-json/wc/gla/mc/mapping/rule`. You should get `400 Request error`
11.  Request to `POST /wp-json/wc/gla/mc/mapping/rule` with empty body. You should get `400 Request error`
12.  Request to `POST /wp-json/wc/gla/mc/mapping/rule` with this body. 

```
{
    "rule": { "attribute": "not_correct", "source": "no", "category_condition_type": "ALL" }
}
```
You should get `400 Request error`

13.  Request to `POST /wp-json/wc/gla/mc/mapping/rule` with this body. 

```
{
    "rule": { "attribute": "adult", "source": "no", "category_condition_type": "BAD" }
}
```
You should get `400 Request error`


src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
```diff
@@ -77,6 +78,97 @@ public function register_routes(): void {
...
+					'callback'            => $this->get_mapping_rules_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+			],
+		);
+		/**
+		 * POST - Upsert an Attribute mapping rule
+		 */
+		$this->register_route(
+			'mc/mapping/rule',
```
mikkamp: These routes don't really follow the generic pattern which is used in WP API's. Usually for CRUD we would see something like:

```
CREATE : POST mc/mapping/rules
READ (collection)  : GET mc/mapping/rules
READ (single not always added) : GET mc/mapping/rules/<id>
UPDATE : POST mc/mapping/rules/<id>
DELETE : DELETE mc/mapping/rules/<id>
```

You can see an example of how that's done with [campaigns](https://github.com/woocommerce/google-listings-and-ads/blob/2.1.2/src/API/Site/Controllers/Ads/CampaignController.php#L51-L91)

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
```diff
@@ -77,6 +78,97 @@ public function register_routes(): void {
...
+									'type'              => 'string',
+									'validate_callback' => 'rest_validate_request_arg',
+									'required'          => true,
+									'enum'              => $this->attribute_mapping_helper->get_category_condition_types(),
+								],
+								'categories'              => [
+									'description'       => __( 'Comma separated categories for this rule.', 'google-listings-and-ads' ),
+									'type'              => 'string',
+									'validate_callback' => 'rest_validate_request_arg',
+								],
```
mikkamp: The schema here is great, but we want to use that schema both for when we create and when we read a rule (or set of rules). This means it would be best to create a separate controller just for the rules, so the schema can be moved to the function `get_schema_properties`.

Also the validation of the categories could be a bit stricter, or passed through a custom function. Based on the test data it looks like we intend to store a comma separated string of category id's, but the schema here doesn't clarify that they can't be category slugs.

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
```diff
@@ -117,6 +209,68 @@ protected function get_mapping_sources_read_callback(): callable {
...
+
+	/**
+	 * Callback function for getting the Attribute Mapping rules from DB
+	 *
+	 * @return callable
+	 */
+	protected function get_mapping_rules_callback(): callable {
+		return function() {
+			try {
+				return $this->attribute_mapping_helper->get_rules();
```
mikkamp: Another advantage of moving it to it's own controller is that you can pass it through the prepare for response functions as seen [here](https://github.com/woocommerce/google-listings-and-ads/blob/2.1.2/src/API/Site/Controllers/Ads/CampaignController.php#L104-L110).
That will ensure the response conforms to the schema.

src/Product/AttributeMappingHelper.php
```diff
@@ -105,4 +123,67 @@ public static function get_source_taxonomies(): array {
...
+	}
+
+	/**
+	 * Gets a specific rule from Database
+	 *
+	 * @param int $rule_id The rule ID to get from Database
+	 * @return array The rules from database
+	 */
+	public function get_rule( int $rule_id ): array {
+		return $this->rules_query->where( 'id', $rule_id )->get_row();
```
mikkamp: I'm not really sure these "helper" functions should be placed in this class. Wouldn't it be better to pass an instance of `AttributeMappingRulesQuery` to the controller and move these functions within that class.

Also in this case these functions have a very specific name, but a lot of this is functionality that can be reused for any table with an ID column. Should they be part of the `Query` class?
Or if we don't want them as part of the `Query` class they could be added as a trait so they can be reused for any table with an ID column (which is currently all our tables).

Something like:
```php
function get_row_by_id( int $id ): array;
function insert_and_get_row( array $data ): ?array;
function update_and_get_row( array $data, int $id ): ?array;
function delete_by_id( int $id ): bool;
```

mikkamp (commented):
I had a look through the code, and left some comments on ways to possibly restructure it. I didn't look through the tests, but I think they will substantially change if it gets restructured.

src/Product/AttributeMappingHelper.php
puntope: Ok! 

Implemented here https://github.com/woocommerce/google-listings-and-ads/pull/1687/commits/b44b973913631ab341ee22abe1c9335c8fbf5276

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
puntope: Ok!

Implemented here https://github.com/woocommerce/google-listings-and-ads/commit/b44b973913631ab341ee22abe1c9335c8fbf5276

src/API/Site/Controllers/MerchantCenter/AttributeMappingController.php
puntope: I tried to implement the custom validation here https://github.com/woocommerce/google-listings-and-ads/commit/b44b973913631ab341ee22abe1c9335c8fbf5276

But it is not calling the function somehow, do I miss something here? 🤔

puntope:
Howdy @mikkamp Thanks for the review. I adapted the code to the suggested changes.

This is ready for a new round

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+			'mc/mapping/rules',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_mapping_rules_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+				[
+					'methods'             => TransportMethods::CREATABLE,
+					'callback'            => $this->post_mapping_create_rule_callback(),
```
mikkamp: Now that we've refactored this I think the names can be "simpler" to make them a bit more descriptive.

post_mapping_create_rule_callback > create_mapping_rule_callback
post_mapping_update_rule_callback > update_mapping_rule_callback

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+		$this->attribute_mapping_helper      = $attribute_mapping_helper;
+		$this->attribute_mapping_rules_query = $attribute_mapping_rules_query;
+	}
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		/**
+		 * GET - Receive All Attribute mapping rules from database
```
mikkamp: This comment should cover both request types if we want to help clarify what we are registering, although I personally don't think the comment is needed. Same for the POST comment.

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @return callable
+	 */
+	protected function post_mapping_create_rule_callback(): callable {
+		return function( Request $request ) {
+			try {
+				if ( ! $this->attribute_mapping_rules_query->insert( $this->prepare_item_for_database( $request ) ) ) {
+					return $this->response_from_exception( new Exception( 'Unable to create the new rule.' ) );
+				}
+
+				return $this->attribute_mapping_rules_query->get_rule( $this->attribute_mapping_rules_query->last_insert_id() );
```
mikkamp: We can add `$this->prepare_item_for_response` to make it consistent with the response from an update call.

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	/**
+	 * @param string $categories  Categories to validate
+	 * @return bool|WP_Error  True if it's validated
+	 *
+	 * @throw Exception when invalid categories are provided
+	 */
+	public function validate_categories_param( string $categories ) {
+			$categories_array = explode( ',', $categories );
```
mikkamp: We can adjust the indentation (as well as the return line) to have one tab less.

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+			],
+			'category_condition_type' => [
+				'description'       => __( 'The category condition type to apply for this rule.', 'google-listings-and-ads' ),
+				'type'              => 'string',
+				'validate_callback' => 'rest_validate_request_arg',
+				'required'          => true,
+				'enum'              => $this->attribute_mapping_helper->get_category_condition_types(),
+			],
+			'categories'              => [
+				'description'       => __( 'Comma separated categories for this rule.', 'google-listings-and-ads' ),
```
mikkamp: Would this description be more specific:

> List of category ID's, separated by commas.

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @throw Exception when invalid categories are provided
+	 */
+	public function validate_categories_param( string $categories ) {
+			$categories_array = explode( ',', $categories );
+
+		foreach ( $categories_array as $category ) {
+			if ( ! is_numeric( $category ) ) {
+				return new WP_Error(
+					'gla_attribute_mapping_invalid_categories_schema',
+					'it should be an string with categories separated by comma, and each of the categories should be in number like format',
```
mikkamp:
```json
{
	"attribute": "attribute is not one of adult, ageGroup, condition, gender, isBundle, sizeSystem, sizeType and color.",
	"category_condition_type": "category_condition_type is not one of ALL, EXCEPT and ONLY.",
	"categories": "it should be an string with categories separated by comma, and each of the categories should be in number like format"
}
```

Should we make the error string match better with the wording used for other errors?

> categories should be a string of category ID's separated by commas.

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\AttributeMappingRulesQuery;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\AttributeMappingHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use Throwable;
```
mikkamp: This import is not used and can be left out.

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @param string $categories  Categories to validate
+	 * @return bool|WP_Error  True if it's validated
+	 *
+	 * @throw Exception when invalid categories are provided
+	 */
+	public function validate_categories_param( string $categories ) {
+			$categories_array = explode( ',', $categories );
+
+		foreach ( $categories_array as $category ) {
+			if ( ! is_numeric( $category ) ) {
```
mikkamp, line 235: PHP 7.4 still has inconsistencies when comparing strings with trailing spaces (fixed in PHP 8). So if we pass categories as a string of `1 , 2 , 3` it will fail validation. Is it worth it to run it through `trim` or do you think we should be fine because the frontend won't be inserting additional spaces?

Tested the following code at https://onlinephp.io to compare the differences
```php
$cats = '1 , 2 , 3';
$nums = explode( ',', $cats );

foreach( $nums as $num ) {
	var_dump( is_numeric( $num ), $num );
}
```

src/DB/Query/AttributeMappingRulesQuery.php
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		return $value;
+	}
+
+	/**
+	 * Gets a specific rule from Database
+	 *
+	 * @param int $rule_id The rule ID to get from Database
+	 * @return array The rules from database
```
mikkamp: This would be returning a singular rule.

src/DB/Table/AttributeMappingRulesTable.php
```diff
@@ -0,0 +1,61 @@
...
+	 *
+	 * @return string
+	 */
+	protected function get_install_query(): string {
+		return <<< SQL
+CREATE TABLE `{$this->get_sql_safe_name()}` (
+    `id` bigint(20) NOT NULL AUTO_INCREMENT,
+    `attribute` varchar(255) NOT NULL,
+    `source` varchar(255) NOT NULL,
+    `category_condition_type` varchar(255) NOT NULL,
```
mikkamp: We only support three types, is there a better length to use here?

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingRulesControllerTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+		$this->attribute_mapping_rules_query->expects( $this->once() )
+			->method( 'update' )->willReturn( 0 );
+
+		$response = $this->do_request( self::ROUTE_RULE, 'post', self::TEST_RULE );
+		$this->assertEquals( 400, $response->get_status() );
+	}
+
+	private function validate_post_route( $route, $method ) {
+		$rule_with_id = array_merge( [ 'id' => self::TEST_RULE_ID ], self::TEST_RULE, [ 'categories' => null ] );
```
mikkamp, line 125: This has been refactored, since `id` is marked as readonly it doesn't produce an error. But we shouldn't be sending the ID in the data to mimic what a real request would look like.

Maybe we also don't need a shared validate function anymore. Since the create and update endpoints share the same schema we only need to check the validation rules once.

And on the topic of validation, I'm not seeing the custom category validation function being tested. Nor an incorrect enum value for either `attribute` or `category_condition_type`

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingRulesControllerTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+		// delete works
+		$response = $this->do_request( self::ROUTE_RULE, 'delete' );
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals( $data, $response->get_data() );
+	}
+
+	public function test_delete_rule_route_error() {
+		$this->attribute_mapping_rules_query->expects( $this->once() )
+			->method( 'delete' )->willReturn( 0 );
```
mikkamp, line 223: According to the definition of the delete function it `throws InvalidQuery When there is an error deleting data`
Have we formatted the delete route incorrectly? We should only have to catch the errors and never need to check for a `0` value being returned.

mikkamp (commented):
Thanks for all the changes, I can see the validation happening now and the endpoints are working as expected.

I've added a few comments on some things we could improve, they would all be minor things to fix.

src/API/Site/Controllers/MerchantCenter/AttributeMappingRulesController.php
puntope: In theory the categories tree this will come from another API call rendering a TreeSelectControl with the category IDs saved as an array. Then I will use `join(,)` function (the JS implode equivalent). So I guess it's safe.

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingRulesControllerTest.php
puntope: > This has been refactored, since id is marked as readonly it doesn't produce an error. But we shouldn't be sending the ID in the data to mimic what a real request would look like.

`rule_with_id` is used only for return values, not for the input being send. For the returned value yes we have the id.

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingRulesControllerTest.php
puntope: It returns 0 when no rows are deleted. Which is not an error. OFC We are catching error as well by the exception.

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingRulesControllerTest.php
puntope: > Maybe we also don't need a shared validate function anymore. Since the create and update endpoints share the same schema we only need to check the validation rules once.

Wouldn't this make unsafe the controller tests? For example if we would delete the schema lines accidentally in the controller for one of the update/create routes.  The test will pass silently. Checking both we are sure that both of them are working.

puntope:
Thanks @mikkamp 

I applied the suggested minor changes (or clarified them). Let me know if this is ready to go.

tests/Unit/API/Site/Controllers/MerchantCenter/AttributeMappingRulesControllerTest.php
mikkamp: I guess I should clarify. In the `wpdb::delete` function and the `Query::delete` function not deleting a row (maybe because it was previously deleted) is not considered an error. It will only throw an error if it encounters a problem like connecting to the database or the SQL table is missing or something related.

However in the `AttributeMappingRulesController` we explicitly make the delete return an error if it didn't actually delete a row. That's different then the behaviour in for example deleting a shipping rate or time. If that's intentional then that's fine, I was mainly pointing out the difference.

mikkamp (approved):
Thanks for the last adjustments. I added some clarifications on what I meant in regards to the delete. However everything is working so I'll go ahead and approve the PR.

PR #1690, opened 2022-09-22
Free Listings + Paid Ads: Return null statistics when onboarding is not finished

ianlin:

This is a follow-up for #1681 and found by @eason9487. In this PR it modifies the `MerchantStatuses`'s `maybe_refresh_status_data()` method so it will not return `400` error when the onboarding flow is not finished (i.e. the option `gla_mc_setup_completed_at` is not being set). The PR also modifies `get_product_statistics()` method to return `null` statistics if the onboarding flow is not finished.

The reason we add this is the API `GET mc/product-feed` will be called during the step 4 of the new onboarding flow, in `mc/product-feed` it will call `maybe_refresh_status_data()` from `MerchantStatuses`, and the previous behaviour would return `400` if the option `gla_mc_setup_completed_at` is not being set.

The code this PR added was actually added in a closed PR #1667's 8bf41533833b3763ac9055ebde220688a70dbe66 commit. In that PR `mc/product-statistics` needs to be called during the new onboarding flow as well.

### Steps to reproducce:

1. Use the branch `feature/1610-streamlined-onboarding`
2. Remove the option `gla_mc_setup_completed_at`, `_transient_timeout_gla_mc_statuses`, and `_transient_gla_mc_statuses` from the table `wp_options`.
3. Call the API `GET mc/product-feed?per_page=1&orderby=total_sales&order=desc`
4. It should return `400` error with the message `Merchant Center account is not set up.`.

1. Use the branch of this PR: `update/1610-return-null-statistics-when-onboarding-not-finish`.
2. Do the steps 2 and 3 from the above.
3. It should return the correct product data.

### DIFF
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -108,15 +108,18 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 	public function get_product_statistics( bool $force_refresh = false ): array {
 		$this->maybe_refresh_status_data( $force_refresh );
 
-		$counting_stats = $this->mc_statuses['statistics'];
-		$counting_stats = array_merge(
-			[ 'active' => $counting_stats[ MCStatus::PARTIALLY_APPROVED ] + $counting_stats[ MCStatus::APPROVED ] ],
-			$counting_stats
-		);
-		unset( $counting_stats[ MCStatus::PARTIALLY_APPROVED ], $counting_stats[ MCStatus::APPROVED ] );
+		$counting_stats = isset( $this->mc_statuses['statistics'] ) ? $this->mc_statuses['statistics'] : null;
+
+		if ( is_array( $counting_stats ) ) {
+			$counting_stats = array_merge(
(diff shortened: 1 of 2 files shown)


mikkamp (commented):
While the code changes here work (and they have already been reviewed), I personally think this is changing the wrong code. If we no longer need to call `gla/mc/product-statistics` before onboarding completed then we shouldn't need to have that endpoint return null data.

As far as I can see we only call `maybe_refresh_status_data` to ensure that we have the latest status for each product we return. But if we aren't onboarded then there is nothing to retrieve from MC. So why can't we just add a condition around `maybe_refresh_status_data` and only call it if we are onboarded? That would be a smaller change and would allow the endpoint to be called at any time.

ianlin:
>  So why can't we just add a condition around `maybe_refresh_status_data` and only call it if we are onboarded? That would be a smaller change and would allow the endpoint to be called at any time.

Thanks for the suggestion, completely agree with it. I was rushing to added the functionality back so did not think too much about it. I will close this PR and create a new one for it.

PR #1680, opened 2022-09-19
GLA Attribute Mapping - Add Create Rule Modal

puntope:
### Changes proposed in this Pull Request:

Closes part of #1648.

This PR ads the modal and the form for adding the new Attachment tribute Mapping rule when the user clicks on "Add new Attribute Mapping rule"

⚠️ The data for the attributes, categories and sources is still dummy until the backend side is finished (soon)
⚠️ The button "Save Rule" doesn't have functionality yet. (in a future PR to save the settings and reload the data in the table)
⚠️  Test will be added once implemented the API endpoints

### Screenshots:

https://user-images.githubusercontent.com/5908855/191014123-1a7fb93e-d73b-4c02-9c31-ad319f8459ff.mov

### Detailed test instructions:

1.   Enter into Attribute Mapping Tab
2.  Click on "Create attribute rule"
3.  See the modal appearing matching the design flow and appearance in Figma (some differences in copy/links are ok since UX it's WIP)
4. Select Adult (an enum field) an see that only default value input and categories selector appear
5. Select Brand and see how Field Type selector appears
6. Select Set Fixed value and see how the view change to text input
7. Check that categories selector works fine
8. Check that the modal close/open works fine
9. Check code quality 

### Additional details:

Project:  https://googleforwooextension.wordpress.com/2022/07/06/project-thread-custom-attribute-mapping/
Figma: https://www.figma.com/file/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=1573%3A141294

PR #1691, opened 2022-09-22
Attribute Mapping - Table Data

puntope:

Closes  part of #1648 .

This PR renders the rows in the table data according to Figma Design.

⚠️  For now the data is Dummy until PR's regarding Data are fully approved. See #1687 
⚠️  Button for Adding new Rule is implemented here #1680 

### Screenshots:

Screenshot 2022-09-22 at 18 32 42

1.  Enter into Attribute Mapping tab
2.  See the table looks like the desing in Figma
3.  Hover the mause in the categories column and see the tooltip appearing with the category names.
4. If there are more than 5 categories, only 5 categories are shown and `...` is added. 

### Additional details:

Project: https://googleforwooextension.wordpress.com/2022/07/06/project-thread-custom-attribute-mapping/
Figma: https://www.figma.com/file/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=1573%3A141294

### DIFF
--- a/js/src/attribute-mapping/attribute-mapping-table-categories.js
+++ b/js/src/attribute-mapping/attribute-mapping-table-categories.js
@@ -1,0 +1,47 @@
+/**
+ * External dependencies
+ */
+import { _n, sprintf } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import AppTooltip from '.~/components/app-tooltip';
+
+const DUMMY_CATEGORIES = [
+	{ id: 1, name: 'Category 1' },
+	{ id: 2, name: 'Category 2' },
+	{ id: 3, name: 'Category 3' },
+];
+
+const AttributeMappingTableCategories = ( { categories } ) => {
+	const categoryNames = categories
+		.slice( 0, 5 )
+		.map( ( category ) => {
+			return DUMMY_CATEGORIES.find(
+				( e ) => e.id === parseInt( category, 10 )
+			).name;
+		} )
+		.join( ', ' );
+
+	const maybeDots = categories.length > 5 ? '...' : '';
+
+	return (
+		<AppTooltip text={ `${ categoryNames }${ maybeDots }` }>
+			<div className="gla-attribute-mapping__table-categories-help">
+				{ sprintf(
+					// translators: %d: number of categories.
+					_n(
+						'%d category',
+						'%d categories',
+						categories.length,
+						'google-listings-and-ads'
+					),
+					categories.length
+				) }
+			</div>
+		</AppTooltip>
+	);
+};
+
+export default AttributeMappingTableCategories;
--- a/js/src/attribute-mapping/attribute-mapping-table-categories.js
+++ b/js/src/attribute-mapping/attribute-mapping-table-categories.js
@@ -46,7 +46,7 @@ const AttributeMappingTableCategories = ( { categories, condition } ) => {
 		} )
 		.join( SEPARATOR );
 
-	const maybeDots =
+	const more =
 		categoryArray.length > CATEGORIES_TO_SHOW
 			? sprintf(
 					// translators: %d: The number of categories.
@@ -66,7 +66,7 @@ const AttributeMappingTableCategories = ( { categories, condition } ) => {
 				text={
 					<div className="gla-attribute-mapping__table-categories-tooltip">
 						{ categoryNames }
-						{ maybeDots && <span>{ maybeDots }</span> }
+						{ more && <span>{ more }</span> }
 					</div>
 				}
 			>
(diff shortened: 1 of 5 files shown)


js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -2,29 +2,29 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
+import { createInterpolateElement } from '@wordpress/element';
 import { Table } from '@woocommerce/components';
-import { CardBody, CardFooter } from '@wordpress/components';
-
+import { CardBody, CardFooter, Flex } from '@wordpress/components';
+import GridIconTrash from 'gridicons/dist/trash';
```
eason9487: It would be a bit nicer to use a lowercase i for the GridiconTrash naming so that all relevant `gridicons` imports in this codebase would have consistent naming.

js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -52,8 +94,47 @@ const AttributeMappingTable = () => {
...
+											{ __(
+												'Manage',
+												'google-listings-and-ads'
+											) }
+										</AppButton>
+										<AppIconButton
+											icon={
+												<GridIconTrash size={ 18 } />
+											}
+										/>
```
eason9487: WDYT of using `AppButton` directly because `Button` already supports rendering an icon with `icon` and `iconSize` props. And the icon size seems to be more close to the default size of 24.

js/src/attribute-mapping/attribute-mapping-table-categories.js
```diff
@@ -0,0 +1,47 @@
...
+	const categoryNames = categories
+		.slice( 0, 5 )
+		.map( ( category ) => {
+			return DUMMY_CATEGORIES.find(
+				( e ) => e.id === parseInt( category, 10 )
+			).name;
+		} )
+		.join( ', ' );
+
+	const maybeDots = categories.length > 5 ? '...' : '';
```
eason9487: Is this expected to use "+# more" rather than "..."?

image

js/src/attribute-mapping/attribute-mapping-table-categories.js
```diff
@@ -0,0 +1,47 @@
...
+
+const AttributeMappingTableCategories = ( { categories } ) => {
+	const categoryNames = categories
+		.slice( 0, 5 )
+		.map( ( category ) => {
+			return DUMMY_CATEGORIES.find(
+				( e ) => e.id === parseInt( category, 10 )
+			).name;
+		} )
+		.join( ', ' );
```
eason9487: It would be nice if we could wrap `', '` with an i18n function to provide more context to translators. For example:

https://github.com/woocommerce/google-listings-and-ads/blob/340a524876435634fea495ca977d082f74f67ea3/js/src/utils/createMessageForMultipleErrors.js#L19-L23

js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -34,14 +34,56 @@ const ATTRIBUTE_MAPPING_TABLE_HEADERS = [
...
+	},
+];
+
+const DUMMY_DESTINATIONS = [
+	{ id: 'adult', name: 'Adult' },
+	{ id: 'brands', name: 'Brands' },
+	{ id: 'color', name: 'Color' },
 ];
 
+const parseCategories = ( condition, categories ) => {
```
eason9487: WDYT of moving this processing into `AttributeMappingTableCategories` and using `AttributeMappingTableCategories` in the table directly? In this way, the whole picture of how the content of a table cell is composited would be in the same place and might be a bit clearer.

eason9487 (commented):
Thanks for the implementation. Viewed and tested the code, and it works correctly. Left a few comments that could be tweaked a little bit, and the main concern is the different implementations in the category cell.

puntope:
Thank for the review @eason9487 

I addressed all your suggestions, so this is ready for another review!

Thanks!

js/src/attribute-mapping/attribute-mapping-table-categories.js
```diff
@@ -0,0 +1,90 @@
...
+	const categoryNames = categoryArray
+		.slice( 0, CATEGORIES_TO_SHOW )
+		.map( ( category ) => {
+			return DUMMY_CATEGORIES.find(
+				( e ) => e.id === parseInt( category, 10 )
+			).name;
+		} )
+		.join( SEPARATOR );
+
+	const maybeDots =
```
eason9487: It would be great if the variable could be renamed as well for the respective change in https://github.com/woocommerce/google-listings-and-ads/pull/1691/commits/f3bf7a8a4cc8974d58e9de5869d6b41e76cde3d7

eason9487 (approved):
Thanks for all the further adjustments. LGTM. Left a comment for a very minor thing but it's not a blocker.

PR #1695, opened 2022-09-23
Override Website Claim Lost issue info

### LINKED ISSUE #1688: Customize the AccountStatus -> accountLevelIssues for website not claimed title and support link
### Describe the bug:

Google changed how they check website verification on an ongoing basis. Once we identify a website is not verified, google send an email to warn the merchant they need to verify and claim their website. If they don't verify, they get an account issue which appears informing the merchant they need to claim their website. This current account issue directs the merchant to MC to fix, but we want to change it so it is clear to the Woo merchant they can before the verification + claim in the woo platform.

**Before the change:** It was possible for a merchant to successfully verify and claim their site through our onboarding, and then lose that verification (for example, if they've changed their store URL, or[ see other reasons here](https://support.google.com/merchants/answer/7527436)), yet the store would still be claimed.

**After the change:** Google launched a new feature in early August that will periodically check if a store's verification is still valid. Merchants who are found to have invalid verifications will then receive emails telling them to re-verify their sites, and they have up to 60 days before their claims are lost and their products are no longer listed on Google's surfaces.

### Steps to reproduce:

1. Merchant doesn't verify their website, they will get an email from Google merchant center asking to claim their website
2. If merchant doesn't claim their website through the plugin after 60 days, they will get this account suspension (here)
Screenshot 2022-09-12 1 54 39 PM

### Expected behavior:

1. We want to change the accountLevelIssues to the following:

- Issue: Website claim is lost, need to re verify and claim your website. Please reference the support link 
- Support link: https://woocommerce.com/document/google-listings-and-ads-faqs/#reverify-website

puntope:

Closes #1688.

This PR overrides the Issue data for `home_page_issue` with

Title: Website claim is lost, need to re verify and claim your website. Please reference the support link
Support Link: https://woocommerce.com/document/google-listings-and-ads-faqs/#reverify-website

### Screenshots:

Screenshot 2022-09-23 at 12 50 47

1. Make the site website claim to be lost
2. Refresh the issues and go to Product feed
3. See the Website not claimed issue with the new info.

### Changelog entry

> Tweak - Update Website not Claimed issue information.

### DIFF
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -814,12 +814,20 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 		 *
 		 * TODO: Remove the condition of matching the $issue['issue']
 		 *       if its issue code is the same as 'merchant_quality_low'
-		 *       after Free and Enhanced Listings merge.
+		 *       after Google replaces the issue title on their side.
 		 */
 		if ( 'merchant_quality_low' === $issue['code'] || "Account isn't eligible for free listings" === $issue['issue'] ) {
 			$issue['issue']      = 'Show products on additional surfaces across Google through free listings';
 			$issue['severity']   = self::SEVERITY_WARNING;
 			$issue['action_url'] = 'https://support.google.com/merchants/answer/9199328?hl=en';
+		}
+
+		/**
+		 * Reference: https://github.com/woocommerce/google-listings-and-ads/issues/1688
+		 */
+		if ( 'home_page_issue' === $issue['code'] ) {
+			$issue['issue']      = 'Website claim is lost, need to re verify and claim your website. Please reference the support link';
+			$issue['action_url'] = 'https://woocommerce.com/document/google-listings-and-ads-faqs/#reverify-website';
 		}
 
 		return $issue;


mikkamp (approved):
The code looks good and the issue is being overwritten.

image

Although the original text of the issue is "Website claim pending expiration". From my understanding that means it hasn't expired yet. Whereas the message we overwrite with indicates that it has been lost. Do we need to clarify what kind of text we can get back for `home_page_issue` and whether it always needs to be overwritten?

puntope:
> The code looks good and the issue is being overwritten.
> 
> image
> 
> Although the original text of the issue is "Website claim pending expiration". From my understanding that means it hasn't expired yet. Whereas the message we overwrite with indicates that it has been lost. Do we need to clarify what kind of text we can get back for `home_page_issue` and whether it always needs to be overwritten?

Do you mean that `home_page_issue` ID could have several issue titles?

mikkamp:
> Do you mean that `home_page_issue ID` could have several issue titles?

Yes it seems like it can have several titles. I'm assuming we want to overwrite all of them. However I think we should clarify with Google whether the text "Website claim pending expiration" means it's actually lost and should be re-verified, or whether it's something that will happen soon.

puntope:
> Yes it seems like it can have several titles. I'm assuming we want to overwrite all of them. However I think we should clarify with Google whether the text "Website claim pending expiration" means it's actually lost and should be re-verified, or whether it's something that will happen soon.

Ok. I will ask in the google doc

PR #912, opened 2021-07-28
Contact info notification and issue

mikkamp:
### Changes proposed in this Pull Request:

This PR checks if contact information has been set for users that already went through the onboarding process.
If it hasn't been set it prompts them to do so with an Inbox Notifcation and an entry added to the Issues table.

Closes #910

### Detailed test instructions:
1. Edit `gla_db_version` in the options table and set it to an older version (like 1.2.1), to trigger the DB table to update
2. Confirm we already completed Merchant Center setup
3. Send a POST request to `/wc/gla/mc/contact-information` containing a valid phone number (will also update store address)
4. On the connection test page clear the transient for the Merchant Statuses
5. Reload the Product Feed page and confirm there is no issue showing up for updating contact info
6. Go to WooCommerce > Home and confirm there is no notice prompting to update the contact info
7. Send a POST request to `/wc/gla/mc/contact-information` with the `phone_number` set to `null`
8. On the connection test page clear the transient for the Merchant Statuses
9. Delete the option `gla_contact_info_setup` in the `wp_options` table
10. Reload the Product Feed page and confirm there is an issue showing up for updating contact info
image
11. Check the issue link and confirm it's pointing to the full link `...?page=wc-admin&path=/google/settings&subpath=/edit-contact-information` (it was previously limited to 100 characters in the DB table)
12. Go to WooCommerce > Home and confirm there is a notice prompting to update the contact info
image
13. Remove the previous note from the DB table `wp_wc_admin_notes` with name `gla-contact-information`
14. Reload WooCommerce > Home and confirm the notice is gone
15. Update the address in the Merchant Center dashboard and make it different to the one stored in WC settings
16. Manually clear the transient `mc_contact_info` in the `wp_options` table
17. Reload WooCommerce > Home and confirm the notice appears again

### Changelog entry

> Add - Prompt users to setup their contact information.

PR #1696, opened 2022-09-23
Migrate column data type in MC account Issues table

### LINKED ISSUE #1694: Action column in MC issues is truncated when more than 100 characters.
### Describe the bug:

The action column in the MC issues is truncated when the size of the action is higher than 100 characters.

This is causes because the action column was not used previously and the size of it is still `varchar(100)`

### Steps to reproduce:

1. Try to add some data in `action` column inside `wp_gla_merchant_issues `. See the error `Data too long for column 'action' at row 1`

### Expected behavior:

- Data size is enough to render properly the issue title. 

### Actual behavior:

- Data size is not enough to render properly the issue title. And the system truncates the issue title showing it as in the screenshot below:

Screenshot 2022-09-23 at 11 41 13

puntope:

Closes #1694.

This PR migrates the data type for `wp_gla_merchant_issues`.`action` 
We migrate from VARCHAR(100) to TEXT

Context: The action column was not used and being defined as 100 Chars limit. Now the issues action contains the description... so sometimes the description seems truncated.

### Screenshots:

Before:
Screenshot 2022-09-23 at 11 41 13

After:
Screenshot 2022-09-23 at 13 35 12

1. Change the version in `src/DB/Migration/Migration20220923T1663910364.php:get_applicable_version()` to match the version defined in the plugin file.
2. Enter to the plugin dashboard
3. Check the Table data type in `wp_gla_merchant_issues`.`action`  was changed to TEXT
4. Check that you can add more than 100 characters in the column
5. Check that Product Feed Issues is working as before (if action is set it shows a modal after clicking on Read more about this issue) 

### Changelog entry

> Fix - Avoid truncate for issues with more than 100 characters length


mikkamp (changes requested):
While this migration works I think it's overkill. We use [`db_delta`](https://developer.wordpress.org/reference/functions/dbdelta/) for updating / creating database tables. So unless we change a key or actually convert the data then we don't need a migrate script.

All we need to do is change the SQL as is done in [this commit](https://github.com/woocommerce/google-listings-and-ads/pull/912/commits/42e9ce69b93a3ecc725c3c8b991800a0eab03d2a). You can change the column to `text` there and as soon as the plugin version is bumped, DB delta will take care of updating the table to the new structure.

puntope:
> While this migration works I think it's overkill. We use [`db_delta`](https://developer.wordpress.org/reference/functions/dbdelta/) for updating / creating database tables. So unless we change a key or actually convert the data then we don't need a migrate script.
> 
> All we need to do is change the SQL as is done in [this commit](https://github.com/woocommerce/google-listings-and-ads/pull/912/commits/42e9ce69b93a3ecc725c3c8b991800a0eab03d2a). You can change the column to `text` there and as soon as the plugin version is bumped, DB delta will take care of updating the table to the new structure.

Interesting, thanks for clarifying.  Didn't know about that one.

I updated the code here: https://github.com/woocommerce/google-listings-and-ads/pull/1696/commits/d3696b5e261ba3a9fac85a7ee420cb065b5a15ea

mikkamp (approved):
Thanks for the changes, it's working as expected. When I bump/adjust the version the column length is changed, and I can now see a value there that is longer than a 100 characters.

PR #1697, opened 2022-09-23
Fix Size types ENUM options

puntope:

This PR updates the options available for Size type attribute.

Based on this  docs https://support.google.com/merchants/answer/6324497 seems that these values were updated. Looks like the Google product data feed-spec was updated in 2021 to support the new fields for the `size_type` product attribute. The announcement article on the Merchant Center Help Center for the 2021 Feed-spec update is [here](https://support.google.com/merchants/answer/10451760).

⚠️  Notice the current values are "working" still and Google seems like converting `Oversize` to `Plus`

Google confirmed that 

### Screenshots:

Screenshot 2022-09-23 at 14 02 25

Screenshot 2022-09-23 at 14 05 40

Screenshot 2022-09-23 at 14 06 42

1. Go to a Product  and check that the new values are added to the GLA Attributes tab Size Type selector.
2. Select one of the new values, for example "Tall"
3. Sync the product and see how the data correctly appears in Merchant center

### Changelog entry

> Fix - Update Size Type Attribute available values

### DIFF
--- a/src/Product/Attributes/SizeType.php
+++ b/src/Product/Attributes/SizeType.php
@@ -48,8 +48,8 @@ class SizeType extends AbstractAttribute implements WithValueOptionsInterface {
 			'regular'      => __( 'Regular', 'google-listings-and-ads' ),
 			'petite'       => __( 'Petite', 'google-listings-and-ads' ),
 			'plus'         => __( 'Plus', 'google-listings-and-ads' ),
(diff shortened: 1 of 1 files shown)


mikkamp (approved):
Thanks for the changes, I can confirm that the final attributes appear in the Merchant Center.
image

For values that were previously saved as oversize, if they update the product it will revert back to default (no size type) because oversize no longer exists. I think that behaviour is fine, but that means some merchants will need to update their products after this change.

PR #1699, opened 2022-09-26
Free Listings + Paid Ads: Add two APIs to retrieve the syncable products count and schedule the job to calculate the count

ianlin:

This is the [backend task 1](https://github.com/woocommerce/google-listings-and-ads/issues/1610#be-task1) 📌 **_Syncable Products Count_** from the epic issue #1610.

This PR creates two APIs:
- `GET mc/syncable-products-count`, to retrieve the number of syncable products from the database.
- `POST mc/syncable-products-count`, to schedule a job to calculate the number of syncable products and save it in the database.

The job scheduled by the POST API is a batched async scheduler job, it has mostly the same logic as an existing job `UpdateAllProducts`, so this PR also extracts the similar code between both jobs to prevent the code from being repetitive.

1. Disconnect all the accounts from GLA setting page.
2. Request to `GET mc/syncable-products-count`, it should returns
    ```json
    {
        "count": null
    }
    ````
3. Request to `POST mc/syncable-products-count` to schedule the job to calculate the # of syncable products.
4. Go to **Scheduled Actions** in **WooCommerce Tools**, search for `update_syncable_products_count` and order by scheduled date DESC
    - The path: `wp-admin/admin.php?page=wc-status&tab=action-scheduler&s=syncable_products&action=-1&action2=-1&orderby=schedule&order=desc&paged=1`
5. The actions `gla/jobs/update_syncable_products_count/create_batch` and `gla/jobs/update_syncable_products_count/process_item` should be scheduled.
6. After the job is completed, request to `GET mc/syncable-products-count` again, it should returns the number, e.g.
    ```json
    {
        "count": 100
    }
    ````

Also verify if the existing job `UpdateAllProducts` is working as expected by either calling `POST mc/settings/sync`, or go through the onboarding again, then check the job `upate_all_products` from **Scheduled Actions** in **WooCommerce Tools** to see if there is any error.


src/Jobs/SyncableProductsBatchedActionSchedulerJobTrait.php
```diff
@@ -0,0 +1,87 @@
...
+			// if no more items the job is complete
+			$this->handle_complete( $batch_number );
+		} else {
+			// if items, schedule the process action
+			if ( count( $items ) ) {
+				$this->schedule_process_action( $items->get_product_ids() );
+			}
+
+			// Add another "create_batch" action to handle unfiltered items.
+			$this->schedule_create_batch_action( $batch_number + 1 );
```
ianlin: Self comment for the reviewers: I remove the `if` condition as below code because I noticed that that last batch would never be processed if the size of it is less than the batch size.

https://github.com/woocommerce/google-listings-and-ads/blob/6fc608dbfde001ce36c2c36e3a76cad31e07ae17/src/Jobs/UpdateAllProducts.php#L94-L97

mikkamp: This problem has been identified in the past, but up until now we've never actually used `handle_complete` so we've avoided it, see: https://github.com/woocommerce/google-listings-and-ads/pull/1451#pullrequestreview-955934017

The last batch is getting processed though, what removing that `if` statement does is, it will register an additional batch after the last one. When we get to running the batch it can't find any products and then calls `handle_complete`.

I think it should be alright to add the extra empty batch, the alternative would mean that we need to pass an extra parameter within the batch that it's the last one, so it can know it should call `handle_complete`. If processing the batch and `handle_complete` can both have "heavy" tasks attached to them, then it would be safer to call the `handle_complete` with the empty batch. So I'm a bit on the fence which solution would be right. Either way though can we add some extra comments to clarify exactly what we are doing. And it's probably best to make the same change in `AbstractBatchedActionSchedulerJob` since that follows the same pattern.

src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php
```diff
@@ -0,0 +1,141 @@
...
+		$this->register_route(
+			'mc/syncable-products-count',
+			[
+				[
+					'methods'             => TransportMethods::CREATABLE,
+					'callback'            => $this->update_syncable_products_count_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+			]
+		);
```
mikkamp: Since the endpoint is the same we can combine these like we do [here](https://github.com/woocommerce/google-listings-and-ads/blob/2.1.2/src/API/Site/Controllers/Ads/CampaignController.php#L51-L68)

src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php
```diff
@@ -0,0 +1,141 @@
...
+
+	/**
+	 * Get the callback for syncing shipping.
+	 *
+	 * @return callable
+	 */
+	protected function update_syncable_products_count_callback(): callable {
+		return function( Request $request ) {
+			$job = $this->job_repository->get( UpdateSyncableProductsCount::class );
+			$job->schedule();
```
mikkamp, line 93: Should we be clearing the transient, when we trigger this job? Otherwise it might get a stale count thinking it's already completed if we call a GET request right after.

src/Jobs/UpdateSyncableProductsCount.php
```diff
@@ -0,0 +1,115 @@
...
+		return $this->count;
+	}
+
+	/**
+	 * Set the count instance variable.
+	 *
+	 * @param int $count The number that will be set to $this->count.
+	 */
+	public function set_count( $count ) {
+		$this->count = $count;
```
mikkamp: This seems to only work because several batches are handled quickly in one request. But if the batches are spread out then it should have a new instance of the class so it would no longer persist the counted data.

We need to somehow persist the in between data, which I'm assuming needs to be a list of product ID's to properly aggregate product variations (or is that planned for a followup PR?)

src/Jobs/UpdateSyncableProductsCount.php
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Called when the job is completed.
+	 *
+	 * @param int $final_batch_number The final batch number when the job was completed.
+	 *                                If equal to 1 then no items were processed by the job.
+	 */
+	protected function handle_complete( int $final_batch_number ) {
+		$this->transients->set( TransientsInterface::SYNCABLE_PRODUCTS_COUNT, $this->get_count(), HOUR_IN_SECONDS );
```
mikkamp: Is an hour long enough? If we aren't going to recount until we restart onboarding, then why does it need an expiry time at all? Should it be moved to an option?

mikkamp (commented):
I left some comments as I went through the code.

When testing it seemed to only count the products that were run within the same process as the `handle_complete` function. But if the batches are spread out it doesn't persist the data.

I got a total result of 1561 but I should be getting 2285

src/Jobs/UpdateSyncableProductsCount.php
ianlin: Thanks for the suggestion. I think using an option here is more appropriate than using a transient, as you said we would only recalculate when we restart onboarding.

src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php
ianlin: According to the below comment https://github.com/woocommerce/google-listings-and-ads/pull/1699#discussion_r980712187 I'm going to change to use an option rather than a transient. I will still add a line here to remove the option before we run the job so the frontend would not get a stale count.

src/Jobs/UpdateSyncableProductsCount.php
ianlin: > This seems to only work because several batches are handled quickly in one request. But if the batches are spread out then it should have a new instance of the class so it would no longer persist the counted data.

Thanks, I was also thinking about it but I did not run into the problem so I assumed async scheduler would always use the same process for the same batched job. Obviously I was wrong so I will update it to keep the intermediate data in the transient.

> which I'm assuming needs to be a list of product ID's to properly aggregate product variations (or is that planned for a followup PR?)

Yeah, aggregating the product variations will be in a followup PR. In this PR I'm going to store a list of product IDs in a transient and append the list in each batch until the job is done.

ianlin: > In this PR I'm going to store a list of product IDs in a transient

Update: will use options instead of transients, since the spirit of transients is, [it might disappear one second after you set them, or 24 hours, but they will never be around after the expiration time](https://developer.wordpress.org/apis/handbook/transients/).

src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php
ianlin: Updated in 0094870.

src/Jobs/UpdateSyncableProductsCount.php
ianlin: Updated in 25f3ee3.

src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php
ianlin: Updated in 25f3ee3 and 8a48750.

src/Jobs/UpdateSyncableProductsCount.php
ianlin: Updated in aaf27ae.

src/Jobs/SyncableProductsBatchedActionSchedulerJobTrait.php
ianlin: Thanks for the explanation, you're right it's `handle_complete` not being called rather than the last batch would never be processed. Update `AbstractBatchedActionSchedulerJob` to have the same pattern and add more comments in 42f3a27.

ianlin:
Thanks for the review @mikkamp. I've updated the code based on your suggestion and it's ready for another review.

src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php
```diff
@@ -0,0 +1,130 @@
...
+		return function( Request $request ) {
+			$response = [
+				'count' => null,
+			];
+
+			$count = (int) $this->options->get( OptionsInterface::SYNCABLE_PRODUCTS_COUNT );
+
+			if ( $count ) {
+				$response['count'] = $count;
+			}
```
mikkamp: This means that on a new site with 0 products (or a site with no syncable products) it will continue returning `null`. Since the status `null` means it's still counting is that going to be a problem? Will that mean it continues polling the count even though it might be completed?

tests/Unit/Jobs/UpdateSyncableProductsCountTest.php
```diff
@@ -0,0 +1,118 @@
...
+				$batch_a->get_product_ids(),
+				$syncable_products
+			);
+
+		$this->options->expects( $this->exactly( 3 ) )
+			->method( 'update' )
+			->withConsecutive(
+				[ OptionsInterface::SYNCABLE_PRODUCTS_COUNT_INTERMEDIATE_DATA, $batch_a->get_product_ids() ],
+				[ OptionsInterface::SYNCABLE_PRODUCTS_COUNT_INTERMEDIATE_DATA, $syncable_products ],
+				[ OptionsInterface::SYNCABLE_PRODUCTS_COUNT, $product_count ]
```
mikkamp: I was personally looking for a hardcoded value of us expecting 4 products to be counted. I'm not against using a calculated total, but if we accidentally forger to add some products to `$syncable_products` then product count would still pass.

Also are we going to throw some "non syncable" products in the mix (non published or out of stock).

mikkamp (commented):
Thanks for the additional changes, it's working as expected.

Although I'd like to clarify a bit more about how it would work for sites with 0 syncable products. And it would be nice to add some unsyncable products in the tests.

tests/Unit/Jobs/UpdateSyncableProductsCountTest.php
ianlin: > if we accidentally forger to add some products to $syncable_products then product count would still pass.

Good point, thanks for the heads up. Updated in d01a43c.

src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php
ianlin: Thanks, this is quite a bug. Updated in 963b327 and 13e1a57. The former updated the job `UpdateSyncableProductsCount` so it will store `0` in the database when there is no syncable products, also added tests for unsyncable products. The latter updated the controller to return `0` when there is no syncable products, and return `null` when the job hasn't finished.

ianlin:
@mikkamp

Thanks for reviewing for the second round, I've fixed the bugs and added the tests for unsyncable products.

mikkamp (approved):
Thanks for all the changes. I've ran the tests again and I'm getting all the expected results.

PR #1703, opened 2022-09-28
 Attribute Mapping - Integrate Modal API (Frontend)

puntope:

Closes part of #1648 .

Implements the API for getting the sources and attributes in the Create Rule modal as well as mapping the names of the source in the table.

### Screenshots:

https://user-images.githubusercontent.com/5908855/192836685-c390ea16-b27b-496e-a894-e68920a50a57.mov

1. Enter to Attend tribute Mapping and see how a placeholder table appears when the data is being loaded
2. After data is being loaded you see the right names in the Target columns. (Notice one empty because that attribute doesn't exists)
3.  Click on create modal. You see the modal open
4.  Click on the selector, you will see real Attributes  from the API
5. Select Adult
6. A spinner will appear until request is finished
7. See the enum selector appearing
8. Select Color attribute. Another spinner will appear
9. See how the field / fixed value selector appears
10. In the field selector you can see some of your Fields/Taxonomies listed
11. Select again Adult. No spinner/request runs. But the right  enum selector appears

### Additional details

⚠️  The test for Category selector will be added when API for gets ting categories is approved
⚠️  We are returning an object in sources. This Will be refactored soon.


js/src/data/actions.js
```diff
@@ -949,3 +949,70 @@ export function* sendMCReviewRequest() {
...
 	}
 }
+
+/**
+ * Individual shipping time.
+ *
+ * @typedef {Object} ShippingTime
+ * @property {CountryCode} countryCode Destination country code.
+ * @property {number} time Shipping time.
+ */
```
eason9487: This `typedef` doesn't seem to be expected in this PR.

js/src/data/actions.js
```diff
@@ -949,3 +949,70 @@ export function* sendMCReviewRequest() {
...
+				'google-listings-and-ads'
+			)
+		);
+	}
+}
+
+/**
+ * Action for fetching available sources for mapping a specific attribute.
+ *
+ * @param {string} attributeKey Attribute id to receive the sources
```
eason9487: I'm not pretty sure about the meaning of "receive" in this description. From the implementation, is it more like "Attribute ID to query the sources"?

js/src/data/actions.js
```diff
@@ -949,3 +949,70 @@ export function* sendMCReviewRequest() {
...
+ */
+export function* fetchMappingSources( attributeKey ) {
+	try {
+		if ( ! attributeKey ) {
+			return {
+				type: TYPES.RECEIVE_MAPPING_SOURCES,
+				sources: {},
+				attributeKey,
+			};
```
eason9487: I guess this intends to handle the empty string value from the default option. If so, it could `return;` directly, so that the wp-data store tree won't have the following empty key and plain object.

```js
{
  attributes: [ /* ... */ ],
  sources: {
    "": {}
  }
}
```

Or, if this did want to use wp-data store for the state of the default option, then the relevant processing in selector could be omitted. Otherwise, this state is held by the store tree but won't ever be used.

https://github.com/woocommerce/google-listings-and-ads/blob/193f643d318e6f735bd22c491ca0c72406681640/js/src/data/selectors.js#L282-L284

js/src/data/resolvers.js
```diff
@@ -308,3 +310,18 @@ export function* getReportByApiQuery( category, type, reportQuery ) {
...
+export function* getMappingSources( attributeKey ) {
+	yield fetchMappingSources( attributeKey );
+}
+
+/*getMappingSources.shouldInvalidate = ( action, query ) => {
+	return (
+		action.type === TYPES.RECEIVE_MAPPING_SOURCES &&
+		action.attributeKey !== query.attributeKey
+	);
+};*/
```
eason9487: It seems to be a temporary test code. Was it accidentally added?

js/src/data/resolvers.js
```diff
@@ -308,3 +310,18 @@ export function* getReportByApiQuery( category, type, reportQuery ) {
 		);
 	}
 }
+
+export function* getMappingAttributes() {
+	yield fetchMappingAttributes();
```
eason9487: The implementation of fetching processing could be placed here instead of placed in the actions.js, although it's not a standard nor a convention when using wp-data, from [some points of view](https://github.com/woocommerce/google-listings-and-ads/pull/1358#discussion_r836052841), it would be a bit nicer.

js/src/data/reducer.js
```diff
@@ -369,6 +373,18 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+		case TYPES.RECEIVE_MAPPING_ATTRIBUTES: {
+			return setIn( state, 'mc.mapping.attributes', action.attributes );
+		}
+
+		case TYPES.RECEIVE_MAPPING_SOURCES: {
+			const { attributeKey, sources } = action;
+			return setIn( state, 'mc.mapping.sources', {
+				...state.mc.mapping.sources,
+				[ attributeKey ]: sources,
+			} );
```
eason9487: Not sure what the structure in the path `mc.mapping.sources` of store tree will be when the passed-in `sources` is changed to an array, but it's worth mentioning that if it's still an object structure, `setIn` already handles shallow copies, so this could be replaced with one of the following:
1. `setIn( state, [ 'mc', 'mapping', 'sources', attributeKey ], sources );`
2. ``setIn( state, `mc.mapping.sources.${ attributeKey }`, sources );``
3. `chainState( state, 'mc.mapping.sources' ).setIn( attributeKey, sources ).end();`

js/src/attribute-mapping/index.test.js
```diff
@@ -1,15 +1,47 @@
...
+				],
+			};
+		} ),
+} ) );
+
+jest.mock( '.~/hooks/useMappingAttributesSources', () => ( {
+	__esModule: true,
+	default: jest
+		.fn()
+		.mockName( 'useMappingAttributes' )
```
eason9487: Typo
```suggestion
		.mockName( 'useMappingAttributesSources' )
```

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -26,13 +27,45 @@ import AttributeMappingCategoryControl from '.~/attribute-mapping/attribute-mapp
...
+		hasFinishedResolution: sourcesHasFinishedResolution,
+	} = useMappingAttributesSources( selectedAttribute );
+
+	const isEnum =
+		attributes.find( ( { id } ) => id === selectedAttribute )?.enum ||
+		false;
+
+	const sourcesOptions = [
+		// Todo: Check this in the future. (Due to an error on my side returning object in the backend)
+		...Object.keys( sources ).map( ( sourceKey ) => {
```
eason9487, line 45: Wonder if there will be other options to be added together with the array literal of `sourcesOptions`? If not, it could be replaced with `const sourcesOptions = Object.keys( sources ).map // ...` without spreading the mapped results again.

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -66,23 +99,18 @@ const AttributeMappingRuleModal = ( { rule, onRequestClose = noop } ) => {
...
-						};
-					} ) }
+					options={ attributesOptions }
 				/>
 			</Subsection>
 
-			{ sources.length > 0 && (
+			{ ! sourcesHasFinishedResolution && <AppSpinner /> }
+
+			{ sourcesOptions.length > 1 && sourcesHasFinishedResolution && (
```
eason9487: Will the `sourcesOptions.length` always be at least 2? Or is it possible the fetched attribute source array has only one item?

js/src/attribute-mapping/attribute-mapping-rule-modal.js
puntope: Yeah. I will update that chunk after the fix in Backend. So this code will be removed.

js/src/attribute-mapping/attribute-mapping-rule-modal.js
puntope: I'm counting the empty one as well "Select Option, coming from backend" 

Note: This is maybe related with the fix on the sources so I need to equalise a bit the response format in backend (including in both empty option and also same value/label format) 

Thanks!

puntope:
Howdy @eason9487 

Thanks for your review. I addressed all your comments and this is ready for another round. 

Thanks in advance.

js/src/data/actions.js
```diff
@@ -949,3 +949,29 @@ export function* sendMCReviewRequest() {
...
+		type: TYPES.RECEIVE_MAPPING_ATTRIBUTES,
+		attributes,
+	};
+}
+
+/**
+ * Receive Mapping Sources action
+ *
+ * @param {Array} sources The sources to update in the state.
+ * @param {string} attributeKey The key for teh attribute we are querying the sources.
```
eason9487: Typo

```suggestion
 * @param {string} attributeKey The key for the attribute we are querying the sources.
```

js/src/attribute-mapping/attribute-mapping-rule-modal.js
puntope: Sorry, I said this comes (or should come from backend) but it's in `<AttributeMappingFieldSourcesControl>`

So you were right. its ` length > 0`

puntope:
@eason9487 

I also chunked the `Add new Attribute mapping button` long test in two. This if for avoiding potential jest timeouts.

Sorry for the "after review request" updates.

js/src/attribute-mapping/attribute-mapping-rule-modal.js
eason9487: Thanks for clarifying!

eason9487 (approved):
Thanks for the implementation and the further adjustments. Testing well. LGTM. Only a typo needs a tweak.

PR #1706, opened 2022-09-29
Free Listings + Paid Ads: Fetch the number of syncable products for the product feed status section

eason9487:

This PR implements an item of the **📌 Client states management and API integration** in #1610.

- Trigger the calculation of the number of syncable products in the Get Started page for later use during the onboarding flow.
- Fetch and show the number of syncable products in the `ProductFeedStatusSection` component.

### Screenshots:

Kapture 2022-09-30 at 11 37 11

1. Go to the Get Started page to trigger the calculation.
2. Proceed with the onboarding flow until step 4.
3. Check if the number of syncable products is shown on the product feed status section.
4. Testing the UI status of waiting for the calculation:
   - It could be done by commenting out the following line to let it return `null` locally.
     https://github.com/woocommerce/google-listings-and-ads/blob/2afa78f34330183bc2b33345a1d90437994aeed4/src/API/Site/Controllers/MerchantCenter/SyncableProductsCountController.php#L76
   - Reload the current page. Check if the UI shows the loading status and wording.
   - In the Network tab of DevTool, it should be able to inspect there is a request to `wc/gla/mc/syncable-products-count` API being issued every 5 seconds.

### DIFF
--- a/js/src/hooks/useSyncableProductsCalculation.js
+++ b/js/src/hooks/useSyncableProductsCalculation.js
@@ -1,0 +1,62 @@
+/**
+ * External dependencies
+ */
+import { useEffect, useCallback } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { API_NAMESPACE } from '.~/data/constants';
+import useApiFetchCallback from './useApiFetchCallback';
+import useCountdown from './useCountdown';
+
+/**
+ * @typedef {Object} SyncableProductsCalculation
+ * @property {number|null} count The number of syncable products. `null` if it's still in the calculation.
+ * @property {()=>Promise} request The function requesting the start of a calculation.
+ * @property {()=>Promise} retrieve The function retrieving the result of a requested calculation by polling with 5-second timer.
+ */
+
+const getOptions = {
+	path: `${ API_NAMESPACE }/mc/syncable-products-count`,
+};
+const postOptions = {
+	...getOptions,
+	method: 'POST',
+};
+
+/**
+ * A hook for communicating with the calculation of the syncable products and for returning the result.
+ *
+ * If a shop has a large number of products, requesting the result with a single API may encounter
+ * a timeout or out-of-memory problem. Therefore, we use an API to schedule a batch for calculating,
+ * and another one to poll the result.
+ *
+ * @return {SyncableProductsCalculation} Payload containing the result of calculation and the functions for requesting and retrieving a calculation.
+ */
+export default function useSyncableProductsCalculation() {
+	const { second, callCount, startCountdown } = useCountdown();
+	const [ fetch, { data } ] = useApiFetchCallback( getOptions );
+	const [ request ] = useApiFetchCallback( postOptions );
+
+	// The number 0 is a valid value.
+	const count = data?.count ?? null;
+
+	const retrieve = useCallback( () => {
+		const promise = fetch();
+		promise.finally( () => startCountdown( 5 ) );
+		return promise;
+	}, [ fetch, startCountdown ] );
+
+	useEffect( () => {
+		if ( second === 0 && callCount > 0 && count === null ) {
+			retrieve();
+		}
+	}, [ second, callCount, count, retrieve ] );
+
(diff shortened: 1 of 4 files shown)


js/src/hooks/useSyncableProductsCalculation.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+import { API_NAMESPACE } from '.~/data/constants';
+import useApiFetchCallback from './useApiFetchCallback';
+import useCountdown from './useCountdown';
+
+/**
+ * @typedef {Object} SyncableProductsCalculation
+ * @property {number|null} count The number of syncable products. `null` if it's still in the calculation.
+ * @property {()=>Promise} request The function requesting the start of a calculation.
+ * @property {()=>Promise} retrieve The function retrieving the result of a requested calculation by polling with 5-second timer.
```
puntope, line 17: Not sure but I guess instead of `by polling` you mean `by pulling`?

js/src/hooks/useSyncableProductsCalculation.js
```diff
@@ -0,0 +1,62 @@
...
+	...getOptions,
+	method: 'POST',
+};
+
+/**
+ * A hook for communicating with the calculation of the syncable products and for returning the result.
+ *
+ * If a shop has a large number of products, requesting the result with a single API may encounter
+ * a timeout or out-of-memory problem. Therefore, we use an API to schedule a batch for calculating,
+ * and another one to poll the result.
```
puntope, line 33: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1706/files#r985802794

I think the verb is [pull ](https://www.webopedia.com/definitions/pull)

js/src/hooks/useSyncableProductsCalculation.js
```diff
@@ -0,0 +1,62 @@
...
+	const count = data?.count ?? null;
+
+	const retrieve = useCallback( () => {
+		const promise = fetch();
+		promise.finally( () => startCountdown( 5 ) );
+		return promise;
+	}, [ fetch, startCountdown ] );
+
+	useEffect( () => {
+		if ( second === 0 && callCount > 0 && count === null ) {
```
puntope, line 52: ❓  Why we need `second === 0 && callCount > 0` and the dependency of second?

puntope: I'm not sure if `callCount` is each time we call `startCountdown` or it is each time the counter reset (after reaching 5)

puntope (approved):
LGTM ✅ 

I place some comment in regards one potential typo and some curiosity question. 

I tested locally with and without forcing null and it works fine :)

js/src/hooks/useSyncableProductsCalculation.js
mikkamp: Just my 2 cents, to me polling is usually used when a request is periodically repeated (which seems to be the case here). Either one could be correct though.

eason9487: Since the client side has no way of knowing when the needed data will be ready, it queries the data by asking periodically. The implementation here matches the scenario, so the term polling is used.

js/src/hooks/useSyncableProductsCalculation.js
eason9487: > Why we need `second === 0 && callCount > 0` and the dependency of second?

It's up to the hook's consumer to decide when to call `retrieve`, and then `retrieve` will request data and schedule a timer for the next call of `retrieve` triggered by this `useEffect`. But `useEffect` will always be called a time once the associated component is mounted, therefore, this condition is to skip the call when mounted.

eason9487: > I'm not sure if `callCount` is each time we call `startCountdown` or it is each time the counter reset (after reaching 5)

https://github.com/woocommerce/google-listings-and-ads/blob/8511198f241152b9b5fb92e128e2fff3c0923cbf/js/src/hooks/useCountdown.js#L9

PR #1686, opened 2022-09-21
Feature - Policy Compliance Checks

mikkamp:
Changes for the Policy Compliance Checks:

- #1532 
- #1563 
- #1611 
- #1612 
- #1622

PR #1707, opened 2022-09-30
Policy Compliance Checks: Remove unused JS hooks

eason9487:

Found four unused JS hooks when testing #1686. This PR removes:

- js/src/hooks/useAllowedCountries.js
- js/src/hooks/usePaymentGateways.js
- js/src/hooks/useRefundReturnPolicyPage.js
- js/src/hooks/useStoreSsl.js

1. Search the codebase to confirm if the four hooks are not used anywhere.
2. `npm run dev` to check if it can build the client files without errors.


mikkamp (approved):
Thanks for removing these hooks. They were originally added before the PolicyCompliance checks switched to using just one endpoint for all the checks. I don't see any errors when building or going through the onboarding with these removed.

PR #1710, opened 2022-10-03
Attribute Mapping - CRUD Rules Frontend

puntope:

Closes part of #1648.

This PR implements the Create/Read/Update/Delete features in the Attribute Mapping Rules table.

Pagination is also included.

### Screenshots:

https://user-images.githubusercontent.com/5908855/193570814-2f71fdea-6a08-400e-94f2-18a31ab05153.mov

https://user-images.githubusercontent.com/5908855/194066435-f577e5e8-439b-4958-8a5e-4ef84bd1e75d.mov

1. GO to Attribute Mapping Tab
2. See that when data is loading a placeholder table appears
3. Click Create a new rule and se how it appears in the table (notice save is disabled if the form is not correct)
4. Save the rule (see Saving... state)
5. Refresh and see how the data appears after reloading
6. Edit the rule and see that Save button is disabled if form is not correct or didn't change
7. After saving, see how it is modified in the table
8. Refresh and see how the data appears correctly after reloading
9. Delete some rule and see the modal 
10. Confirm deletion and see how the rule is deleted
11. Test the pagination by creating more than 10 rules and deleting some of them.

### Additional details:

⚠️  Notice `source_name` (with the nice name for the sources) is not yet implemented. 

Figma: https://www.figma.com/file/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=1610%3A156293


puntope:
ℹ️  Since the pagination backend and the tweaks for the backend are already approved I merged directly here everything.

No need to test the backend part. Just the Frontend.

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -84,13 +129,19 @@ const AttributeMappingRuleModal = ( { rule, onRequestClose = noop } ) => {
 					{ __( 'Cancel', 'google-listings-and-ads' ) }
```
eason9487, line 156: While saving rule, perhaps we could also disable the cancel button and also ignore the `onRequestClose` called from `AppModal`.

https://user-images.githubusercontent.com/17420811/194277120-e73ede29-3e67-4891-b309-da1532830269.mp4

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -141,6 +197,26 @@ const AttributeMappingRuleModal = ( { rule, onRequestClose = noop } ) => {
...
+									...newRule,
+									category_condition_type: type,
+								} );
+							} }
+							onCategoriesChange={ ( categories ) => {
+								setNewRule( {
+									...newRule,
+									categories: categories.join( ',' ),
+								} );
+							} }
```
eason9487: https://user-images.githubusercontent.com/17420811/194274331-774a0337-c2f7-42b7-81bc-6e69b8cf051f.mp4

With these operation steps in the video, the saved rule will have a non-empty value in `categories`. For example:
```json
{
  "id": "2",
  "attribute": "adult",
  "source": "yes",
  "category_condition_type": "ALL",
  "categories": "1,2",
}
```

Wonder if it's the expected result that a rule item has ALL `category_condition_type` and non-empty `categories` at the same time?

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -27,46 +55,63 @@ import AppSpinner from '.~/components/app-spinner';
...
+			}
+			onRequestClose();
+		} catch ( error ) {
+			setSaving( false );
+		}
+	};
+
+	const onSourceUpdate = ( source ) => {
+		setNewRule( { ...newRule, source } );
+	};
```
eason9487: If select an existing product field first and then switch to set a fixed value, it will show the value of selected existing product field in raw format. I guess it might confuse the merchants.

1. select an existing product field
   image
2. Switch to set a fixed value
    image

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -27,46 +55,63 @@ import AppSpinner from '.~/components/app-spinner';
...
+		...mapOptions( attributes ),
 	];
 
+	const isValidRule =
+		newRule.source &&
+		newRule.attribute &&
+		( newRule.category_condition_type ===
+			CATEGORY_CONDITION_SELECT_TYPES.ALL ||
+			newRule.categories?.length > 0 ) &&
+		! isEqual( newRule, rule );
```
eason9487, line 107: It's very minor but maybe the same set in a different order can be considered equal.

https://user-images.githubusercontent.com/17420811/194281505-38ce1628-6f6c-4530-ac11-53ccb8664b41.mp4

js/src/attribute-mapping/attribute-mapping-delete-rule-modal.js
```diff
@@ -0,0 +1,82 @@
...
+	};
+
+	return (
+		<AppModal
+			onRequestClose={ onRequestClose }
+			title={ __( 'Delete attribute rule?', ' google-listings-and-ads' ) }
+			buttons={ [
+				<AppButton key="cancel" isLink onClick={ onRequestClose }>
+					{ __( 'Cancel', 'google-listings-and-ads' ) }
+				</AppButton>,
```
eason9487: Similar to the saving rule state in `AttributeMappingRuleModal`, perhaps we could also disable the cancel button and also ignore the `onRequestClose` called from `AppModal`.

js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -58,7 +68,24 @@ const AttributeMappingTable = ( { rules } ) => {
...
+	/**
+	 * Prevent to stay in a page without rules.
+	 * This is because maybe the user is in the page 2 which has only one rule.
+	 * If the user deletes that rule we don't want to stay in page 2 anymore, since it doesn't exists.
+	 */
+	useEffect( () => {
+		if ( rulesHasFinishedResolution && rules?.length === 0 && page > 1 ) {
+			setPage( page - 1 );
+		}
+	}, [ page, rules, rulesHasFinishedResolution ] );
```
eason9487: It would be nicer to avoid the eslint warning:

```
react-hooks/exhaustive-deps - React Hook useEffect has a missing dependency: 'setPage'. Either include it or remove the dependency array.
```

js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -107,15 +135,52 @@ const AttributeMappingTable = ( { rules } ) => {
...
+											<FlexItem>
+												<AppButtonModalTrigger
+													button={
+														<AppButton
+															isLink
+															text={ __(
+																'Manage',
+																'google-listings-and-ads'
+															) }
+															eventName="gla_attribute_mapping_new_rule_click"
```
eason9487: The `new` in this event name seems to be a typo.

js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -45,11 +50,16 @@ const ATTRIBUTE_MAPPING_TABLE_HEADERS = [
...
  * @return {JSX.Element} The component
  */
-const AttributeMappingTable = ( { rules } ) => {
+const AttributeMappingTable = () => {
+	const { page, setPage } = usePagination( 'attribute-mapping' );
+
+	const {
+		data: { rules, total },
+		hasFinishedResolution: rulesHasFinishedResolution,
+	} = useMappingRules( { page } );
```
eason9487: WDYT of explicitly specifying the `perPage` value here with the same value as it on the `Pagination`?

https://github.com/woocommerce/google-listings-and-ads/blob/fa831690d66733490bd7b40d87291dfe34dd1189/js/src/attribute-mapping/attribute-mapping-table.js#L211-L213

js/src/attribute-mapping/index.scss
```diff
@@ -80,4 +80,8 @@
 		margin-left: $grid-unit-05;
 		word-wrap: break-word;
 	}
+
+	&__manage {
```
eason9487: Looks like an unused CSS style?

js/src/data/reducer.js
```diff
@@ -387,6 +392,52 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
 			);
 		}
 
+		case TYPES.RECEIVE_MAPPING_RULES: {
+			const { rules, pagination } = action;
+			const newRulesState = [ ...state.mc.mapping.rules.items ];
+
+			const start = ( pagination.page - 1 ) * pagination.perPage;
+			const end = pagination.perPage;
```
eason9487: It would be more precise a bit to name this variable with `count`, `length` or `size`.

js/src/data/reducer.js
```diff
@@ -387,6 +392,52 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+
+			if ( ruleIndex >= 0 ) {
+				newRulesState[ ruleIndex ] = rule;
+			} else {
+				newRulesState.push( rule );
+			}
+
+			return chainState( state, 'mc.mapping.rules' )
+				.setIn( 'items', newRulesState )
+				.setIn( 'total', newRulesState.length )
```
eason9487: When there are more than 10 mapping rules, e.g. 30 rules, the value of `total` state will be set to an incorrect value of 11 for a moment if only the first page is loaded.

https://user-images.githubusercontent.com/17420811/194299248-da3e0db5-509e-4225-b507-63439c4b7194.mp4

js/src/data/reducer.js
```diff
@@ -387,6 +392,52 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+		}
+
+		case TYPES.DELETE_MAPPING_RULE: {
+			const rules = state.mc.mapping.rules.items.filter(
+				( el ) => el.id !== action.rule.id
+			);
+
+			return chainState( state, 'mc.mapping.rules' )
+				.setIn( 'items', rules )
+				.setIn( 'total', rules.length )
```
eason9487: Similar to the above when deleting an item if only the first page is loaded.

js/src/data/resolvers.js
```diff
@@ -350,3 +354,52 @@ export function* getMappingSources( attributeKey ) {
...
+ * @see AttributeMappingRulesController.php
+ */
+export function* getMappingRules( pagination ) {
+	try {
+		const response = yield fetchWithHeaders( {
+			path: addQueryArgs( `${ API_NAMESPACE }/mc/mapping/rules`, {
+				page: pagination.page,
+				per_page: pagination.perPage,
+			} ),
+			parse: false,
```
eason9487: `fetchWithHeaders` will directly specify/override the `parse` option, which probably doesn't need to be specified here.

js/src/data/resolvers.js
```diff
@@ -350,3 +354,52 @@ export function* getMappingSources( attributeKey ) {
 		);
 	}
 }
+
+/**
+ * Fetches the Attribute Mapping Rules and calls receive action
+ *
+ * @param {Object} pagination Object containing client pagination parameters like page and per_page
```
eason9487: Seems to be a typo.
```suggestion
 * @param {Object} pagination Object containing client pagination parameters like page and perPage
```

js/src/attribute-mapping/attribute-mapping-rule-modal.js
puntope: Since this is not having any impact in the behaviour. I guess it's ok if we allow  the user to do the save request n case the order changes.

js/src/attribute-mapping/attribute-mapping-table.js
puntope: 10 is the default value for WP per_page. But I will add it if that is clearer.

js/src/data/reducer.js
puntope: Renamed to `deleteCount` following the splice argument specification.

puntope:
Howdy @eason9487 

Thanks very much for review this PR and for your nice suggestions. 

Thanks

js/src/attribute-mapping/attribute-mapping-table.js
eason9487: Thanks for the adjustment! Yes, I think explicitly specifying them would have a bit better readability and maintainability even if the default value is the same.

eason9487 (commented):
Thanks for the further adjustments. The rule creation, updates and deletions work correctly. However, there is still one thing that might need to be checked.

It is possible to issue an API request for updating a rule under specific operation steps, but the rule data doesn't actually have changes and gets an error response.

https://user-images.githubusercontent.com/17420811/194463220-13c7b5f6-a4d5-48e6-917c-65963075e10a.mp4

puntope:
Thanks for that nice catch @eason9487 

I fixed it and now YES I hope is working fine. 🙏

eason9487 (approved):
Tested locally and all the CRUD operations to the attribute rules work well. Thanks for the implementation. LGTM.

PR #1711, opened 2022-10-03
Attribute Mapping - Tweak Sources & Data Types in backend

puntope:

Closes part of  #1648 

This PR implements minor tweaks in backend as follows:

- In sources API route we return now an array instead of object for consistency with attributes API route
- We set  '' as default value for categories column. Also we allow '' as  valid data for categories in the API route. 
- Limited sources size to 100 in DB 

1. Test source route, ie. `GET /wp-json/wc/gla/mc/mapping/sources?attribute=color` 
2. See that it returns an array with id & value params.
3. Do a POST request to `/wp-json/wc/gla/mc/mapping/rules` using this payload

`{ "attribute": "adult", "source": "No", "category_condition_type": "ALL", "categories" : "" }``

4. See how it works.
5. Do a POST request again to `/wp-json/wc/gla/mc/mapping/rules` using this payload

`{ "attribute": "adult", "source": "No", "category_condition_type": "ALL"  }``

6. See how it works

### DIFF
--- a/src/API/Site/Controllers/MerchantCenter/AttributeMappingDataController.php
+++ b/src/API/Site/Controllers/MerchantCenter/AttributeMappingDataController.php
@@ -129,23 +129,6 @@ class AttributeMappingDataController extends BaseOptionsController {
 				'description' => __( 'The list of attributes or attribute sources.', 'google-listings-and-ads' ),
 				'context'     => [ 'view' ],
 				'readonly'    => true,
-				'items'       => [
-					'id' => [
-						'type'        => 'integer',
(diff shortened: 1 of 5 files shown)


ianlin (approved):
Tested good and the results were expected.

✅  I can see the response data of `GET /mc/mapping/sources` is an array rather than an object.
✅  For `POST /mc/mapping/rules`, the `categories` will be saved as an empty string `''` if I either don't provide it in the payload, or provide `categories=""` in the payload.
✅  The validation of categories still work if I provide `categories="anything"` in the payload.

PR #1714, opened 2022-10-04
Attribute Mapping - Rule pagination

puntope:

Closes part of #1648 

This PR implements pagination in the Attribute Mapping Rules API 

1. Checkout PR and run this Request --> `GET /wp-json/wc/gla/mc/mapping/rules`
2. See how it returns the WP default pagination as page = 1, per_page = 10
3. See how the headers contain `X-WP-Total` (with the total rules you have) and `X-WP-TotalPages` (with the right number of pages)
4. Run this Request --> `GET /wp-json/wc/gla/mc/mapping/rules?page=1&per_page=1`
5. See how only one Rule is returned, also the header `X-WP-Total`  and  `X-WP-TotalPages`  are correct.
6. Run some other queries playing with `page` and `per_page` params 
7. Run one request with `GET /wp-json/wc/gla/mc/mapping/rules?page=-1`
8. See the error in the response
9. Run one request with `GET /wp-json/wc/gla/mc/mapping/rules?per_page=101`
10. See the error in the response


ianlin (approved):
Tested well and the results were expected with multiple `page` `per_page` combination. 👍

PR #1718, opened 2022-10-06
Attribute Mapping - Sources

puntope:

Closes part of #1648 .

This PR implements more sources in the Attributes. Including Metas and product attributes.
Also it implements a cache API system for the Meta Query.
Finally it implements Yoast SEO integration for GTIN and MPN 

1. Check all the attributes with `GET wp-json/wc/gla/mc/mapping/sources?attribute={attribute}` and see if if returns the right values. Try to create new metas and attributes to see they appear in the results.
2. Check that after querying an attrubute of type Field like `GET wp-json/wc/gla/mc/mapping/sources?attribute=brand` a new transient is created `_transient_gla_attribute_mapping_meta_fields` with the met values. 
3. Call `DELETE wp-json/wc/gla/mc/mapping/cache` and check transient is deleted. 
4. Query `GET wp-json/wc/gla/mc/mapping/sources?attribute=gtin` and `GET wp-json/wc/gla/mc/mapping/sources?attribute=mpn` See how they don't include Yoast fields.
5. "Install" Yoast SEO for WooCommerce. Since it's a premium plugin you can just add this in your functions.php `define( 'WPSEO_WOO_VERSION', '19.7.2' );`
6. Query again `GET wp-json/wc/gla/mc/mapping/sources?attribute=gtin` and `GET wp-json/wc/gla/mc/mapping/sources?attribute=mpn` See how they now include Yoast fields.


src/API/Site/Controllers/MerchantCenter/AttributeMappingCacheController.php
```diff
@@ -0,0 +1,103 @@
...
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		/**
+		 * DELETE Flush the cache
+		 */
+		$this->register_route(
+			'mc/mapping/cache',
```
mikkamp: Since we are using a transient for this it will be regenerated after an hour. When would we need to forcefully clear this value (except for testing)? To me it seems we don't need to add this class/controller at all, as it can be handled internally. In the past we've added some code to the ConnectionTest class if we want to have a simple tool to clear the cache for testing.

src/Product/AttributeMapping/AttributeMappingHelper.php
```diff
@@ -0,0 +1,316 @@
...
+	 *
+	 * @return Transients A single instance of Transients
+	 */
+	public static function get_transients_object(): Transients {
+		if ( is_null( self::$transients ) ) {
+			self::$transients = new Transients();
+		}
+
+		return self::$transients;
+	}
```
mikkamp: Why are we not getting transients from the container instead of getting a new instance? The Transients class is shared meaning we want to have only one instance within the plugin, this will create a new instance just for attribute mapping. It also makes it harder to pass a mocked Transients class for testing.

src/Product/AttributeMapping/AttributeMappingHelper.php
```diff
@@ -0,0 +1,316 @@
...
+		$attributes        = [];
+		$sources           = [];
+
+		foreach ( $taxonomies as $taxonomy ) {
+			if ( taxonomy_is_product_attribute( $taxonomy->name ) ) {
+				$attributes[ 'taxonomy:' . $taxonomy->name ] = $taxonomy->label;
+				continue;
+			}
+
+			$parsed_taxonomies[ 'taxonomy:' . $taxonomy->name ] = $taxonomy->label;
```
mikkamp: Just a minor, but this variable called `parsed` is slightly confusing. To me it looks like we are doing the opposite of parsing, and instead structuring the taxonomies to be in the right format. To me `formatted_taxonomies` would sound better, or I'd rename the response from `get_object_taxonomies` so we can just use `taxonomies`.

src/Product/AttributeMapping/AttributeMappingHelper.php
```diff
@@ -0,0 +1,316 @@
...
+		}
+
+		asort( $parsed_taxonomies );
+		asort( $attributes );
```
mikkamp: Is sorting before filtering intentional?

src/Product/AttributeMapping/AttributeMappingHelper.php
```diff
@@ -0,0 +1,316 @@
...
+
+		return apply_filters( 'woocommerce_gla_attribute_mapping_fields_custom_fields', $sources );
+	}
+
+	/**
+	 * Allowing to register custom attributes by using a filter.
+	 *
+	 * @return array The custom attributes
+	 */
+	public static function get_source_custom_attributes() {
```
mikkamp: I assume the sole purpose of this filter/function is to allow third party plugins to add extra attributes. How does it get the matching data from this attribute? And if I'm adding extra mapping fields, then why would I need to use this instead of `woocommerce_gla_attribute_mapping_fields` like is being done for Yoast.

src/Product/AttributeMapping/Traits/IsFieldTrait.php
```diff
@@ -0,0 +1,34 @@
...
+		return false;
+	}
+
+	/**
+	 * Returns the attribute sources
+	 *
+	 * @return array The available sources
+	 */
+	public static function get_sources(): array {
+		return AttributeMappingHelper::get_fields( self::get_id() );
```
mikkamp: Since this is contained within a trait it will only be defined once, so we can copy the contents of `AttributeMappingHelper::get_fields` here without needing to add an extra helper function (less confusion between fields and sources).

In my opinion we can also rename the filters since we called them `mapping/sources` in the API we want that to match internally so we don't get confused between fields and sources. So the filters could be:

```
woocommerce_gla_attribute_mapping_sources
woocommerce_gla_attribute_mapping_sources_product_fields
etc...
```

mikkamp (commented):
It's looking good and the API is returning the results as expected in the test instructions. I added some comments as I went through the code.

I also noticed that for the meta I get the field: `meta:wpseo_global_identifier_values` but this is an array of identifier values which we can't map to directly. Should we exclude this meta when Yoast SEO is active?

The global product attributes are also showing up in meta data, for example I have both:
- `meta:attribute_pa_color`
- `taxonomy:pa_color`

This is because for variable products the taxonomy is assigned to the variable product (making it easier for filtering), but the post meta identifies what each variation is set to. Should we also filter out meta that starts with the prefix `attribute_pa`?

The same goes for custom attributes, we aren't loading the custom attributes which are created per product, but they do still show up in meta. For example `meta:attribute_size`. If we aren't going to include them as attributes, should they still be selectable in meta?

src/API/Site/Controllers/MerchantCenter/AttributeMappingCacheController.php
puntope: This is (as discussed) for us to clear the cache after entering into the page.  So I was thinking about calling this on the frontend.

If you know any way to hook when the page is visited in the backend could be even better

src/Product/AttributeMapping/AttributeMappingHelper.php
puntope: The meta query is called per each attribute under Product/Attributes folder which as far as I know has not access to container.

src/Product/AttributeMapping/AttributeMappingHelper.php
puntope: Yes. my initial idea was to show the taxonomies/attributes added by the hook after everything else. Relying on the client to decide how it should be sorted. But also fine for me to force our sorting after.

src/Product/AttributeMapping/AttributeMappingHelper.php
puntope: Those Custom attributes are under the umbrella of custom attributes and have the prefix `attribute` that we will use to get the attribute by calling:

`$product->get_attribute( $attribute_key );`

In case of Yoast has its own prefixes because we need to get the Yoast attributes differently.

src/API/Site/Controllers/MerchantCenter/AttributeMappingCacheController.php
mikkamp: From my understanding we discussed that the clear on page visit would only be needed if the cache expiry was long. But if we are sticking with a short expiry (1 hour) then we don't need to trigger it.
They would only end up with stale data if they open up the mapping rules, add new meta keys for their products and then move back to the mapping rules. Which doesn't sound like a common scenario, and if it would happen it would eventually correct itself or a tool to clear transients can be used.

src/Product/AttributeMapping/AttributeMappingHelper.php
mikkamp: The AttributeMappingHelper is a shared class: https://github.com/woocommerce/google-listings-and-ads/blob/add/attribute-mapping/sources/src/Internal/DependencyManagement/CoreServiceProvider.php#L396
We can pass the TransientsInterface as a dependency there.

puntope:
> I also noticed that for the meta I get the field: meta:wpseo_global_identifier_values but this is an array of identifier values which we can't map to directly. Should we exclude this meta when Yoast SEO is active?

Interesting. Do you mean Yoast SEO or WooCommerce Yoast SEO?

puntope:
> This is because for variable products the taxonomy is assigned to the variable product (making it easier for filtering), but the post meta identifies what each variation is set to. Should we also filter out meta that starts with the prefix attribute_pa?

Agree, it seems redundant

puntope:
> The same goes for custom attributes, we aren't loading the custom attributes which are created per product, but they do still show up in meta. For example meta:attribute_size. If we aren't going to include them as attributes, should they still be selectable in meta?

What do you think of removing entirely the meta sources section? Sounds radical but thinking about it... what more than custom attributes per product are going to being added? And since is per product seems for me that is going to be hard to be sure all of the products are having that meta... as well building the logic for discarding (or not) specific metas.  

Besides that... we can get rid of the cache thing and the transients as well.

mikkamp:
> Interesting. Do you mean Yoast SEO or WooCommerce Yoast SEO?

WooCommerce Yoast SEO, it's the same meta field we get the data from here: https://github.com/woocommerce/google-listings-and-ads/blob/2.1.4/src/Integration/YoastWooCommerceSeo.php#L130

mikkamp:
> What do you think of removing entirely the meta sources section? Sounds radical but thinking about it... what more than custom attributes per product are going to being added? And since is per product seems for me that is going to be hard to be sure all of the products are having that meta... as well building the logic for discarding (or not) specific metas.
> Besides that... we can get rid of the cache thing and the transients as well.

I think the need comes to light mainly when looking at other plugins that extend WooCommerce, it's very common to just set some extra data in product meta. And if you are importing your products through CSV (possibly from another platform) and aren't mapping data to product attributes, then the next place it will end up is in product meta.
So I think removing it completely would take away quite a bit of flexibility. I don't have any data to back up how much one is used over the other. Maybe it would be worthwhile to scan through some common plugins and see how the data is stored to help decide if we should push this to a later version.

src/Product/AttributeMapping/AttributeMappingHelper.php
mikkamp: Thanks that makes sense. Although I can't seem to think of any extension or use case where it would add data as a custom attribute and not as a global attribute. Maybe we should research a bit more on how useful this is and which extensions we are going to need this for.

src/Product/AttributeMapping/AttributeMappingHelper.php
mikkamp: It's fine to leave as is, was just wondering if there was a specific need to have them sorted. Most extensions that will hook into this will probably just add without sorting.

puntope:
> I don't have any data to back up how much one is used over the other. Maybe it would be worthwhile to scan through some common plugins and see how the data is stored to help decide if we should push this to a later version.

So just to confirm, we can remove it in this version and check if we need it in next milestones?

src/Product/AttributeMapping/AttributeMappingHelper.php
puntope: What I mean is that even if we pass the Transient object in the constructor and the Provider. It would be able to be called inside a static function? 🤔

src/API/Site/Controllers/MerchantCenter/AttributeMappingCacheController.php
puntope: Ah ok, I didn't catch that one. However, now we deleted all the metas, so that endpoint is not needed anymore.

puntope:
Howdy @mikkamp 

I applied your suggestions and removed the Meta sources feature (+ API and Transients) at the end.

Hence, some of the review comments are (for now) not applicable. I took note of them in case we integrate metas as sources at some point in the future. 

This PR is ready for another round

src/Product/AttributeMapping/AttributeMappingHelper.php
mikkamp: Ah ok I see what you mean. I missed that the attributes were all static, another reason to move the `get_fields` function to the trait to reduce confusion. Anyways, if we are removing the Transient for now we don't need to worry about this yet.

src/Internal/DependencyManagement/RESTServiceProvider.php
```diff
@@ -51,8 +51,9 @@
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantStatuses;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\ContactInformation;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\PhoneVerification;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\Transients;
```
mikkamp: This line can get removed since we no longer use transients.

src/Product/AttributeMapping/Traits/IsEnumTrait.php
```diff
@@ -1,14 +1,14 @@
 <?php
 declare( strict_types=1 );
 
-namespace Automattic\WooCommerce\GoogleListingsAndAds\Product\Attributes;
+	namespace Automattic\WooCommerce\GoogleListingsAndAds\Product\AttributeMapping\Traits;
```
mikkamp: Extra tab here.

src/Product/AttributeMapping/Traits/IsFieldTrait.php
```diff
@@ -0,0 +1,42 @@
...
+	 *
+	 * @return array The available sources
+	 */
+	public static function get_sources(): array {
+		return apply_filters(
+			'woocommerce_gla_attribute_mapping_sources',
+			array_merge(
+				AttributeMappingHelper::get_source_product_fields(),
+				AttributeMappingHelper::get_source_taxonomies(),
+				AttributeMappingHelper::get_source_custom_attributes()
```
mikkamp: These three functions belong specifically to "field" type attributes. As far as I can see they don't need anything from AttributeMappingHelper so it would make more sense to place them within the Trait. Although this makes me wonder if we should be creating an AbstractFieldAttribute class instead of placing these functions in a trait.
Based on that same argument we could create a matching AbstractEnumAttribute.

That gives less of a reason to have all the functions static, but that seems to be a decision that was made previously.

> **Note**
This might be opinionated so feel free to suggest otherwise.

src/Product/Attributes/Color.php
```diff
@@ -15,6 +15,8 @@
  */
 class Color extends AbstractAttribute {
```
mikkamp: This class should also implement WithMappingInterface, although if you follow the advice from the previous comment to create AbstractFieldAttribute then we can add the implementation in the abstract class to keep these simple. Same comment for the AgeGroup class.

src/Product/Attributes/GTIN.php
```diff
@@ -4,6 +4,8 @@
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Product\Attributes;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Admin\Product\Attributes\Input\GTINInput;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\AttributeMapping\AttributeMappingHelper;
```
mikkamp: This class is no longer used.

src/Product/Attributes/MPN.php
```diff
@@ -4,6 +4,8 @@
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Product\Attributes;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Admin\Product\Attributes\Input\MPNInput;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\AttributeMapping\AttributeMappingHelper;
```
mikkamp: This class is no longer used.

mikkamp (commented):
Thanks for all the changes, however I noticed a few more things we might need to address.

In `AttributeMappingHelper::get_sources` we loop through all "mapable" attributes, to retrieve the sources. However in the API I see we only request one specific attribute. Are we ever going to need to load all of them at once? Wouldn't it be better to validate if it's a mapable attribute and then return only the sources for one?

It is going to be cached, however it means we are going to be repeat calls to `get_object_taxonomies` for each "field" type attribute. When we really only need it to be called once. It will also repeat all the filters for each attribute. Which brings me to another point, the filters like  `woocommerce_gla_attribute_mapping_sources_taxonomies` might need a little more context added to them, like the attribute id.

For example the plugin WooCommerce Brands. Since it registers a taxonomy it's already included in the list, but what if I only want it to show up when the sources for the Brand attribute are being requested (and exclude it for the rest). I'd need to know the attribute id when the filter gets called.

puntope:
Howdy @mikkamp 

I applied all the changes.

- Remove unused imports
- Move Sources logic inside traits
- Refactor get_sources in he Data Controller to get only the specific attribute source without querying the rest of attributes.

puntope:
> Which brings me to another point, the filters like `woocommerce_gla_attribute_mapping_sources_taxonomies` might need a little more context added to them, like the attribute id.

The taxonomies are the same for all the attributes. For changing sources available per attribute there is 

`woocommerce_gla_attribute_mapping_sources`

src/Product/Attributes/Color.php
puntope: Since we are using interface methods in the `get_sources` function I think is better a Trait. 

I tried to setup `AbstractFieldAttribute` and `AbstractEnumAttribute` extending `AbstractAttribute` and implementing    `WithMappingInterface` but I got an error in execution because we are calling `self:get_id()` and `self::get_value_options()` from different interfaces. 

We can switch to Abstract maybe with some extra work but for now I've kept them with traits.

src/Product/AttributeMapping/Traits/IsFieldTrait.php
puntope: Answered bellow

src/Product/AttributeMapping/AttributeMappingHelper.php
```diff
@@ -0,0 +1,150 @@
...
+				function ( $class ) use ( $attribute_id ) {
+					/**
+					 * @var AttributeInterface $class
+					 */
+					return $class::get_id() === $attribute_id;
+				}
+			)
+		);
+
+		return ! empty( $attribute ) ? $attribute[0] : null;
```
mikkamp: Since we are only going to return the first match we could simplify this code to make it more readable.

```suggestion
		foreach ( self::ATTRIBUTES_AVAILABLE_FOR_MAPPING as $class ) {
			if ( $class::get_id() === $attribute_id ) {
				return $class;
			}
		}

		return null;
```

mikkamp (approved):
Thanks for all the changes, it's working as expected now and only retrieving the sources once. I just left a small comment about a refactor for making the function more legible.

Also `AgeGroup` only implements one interface it should also implement `WithMappingInterface`.

PR #1719, opened 2022-10-06
Attribute Mapping - Category Tree API

puntope:

Closes part of #1648 .

This PR implements an API for returning the Store categories in hierarchical format. Ready for being used by Tree Select Control in the Create Rule Modal.

### Screenshots:

Screenshot 2022-10-06 at 18 29 09

1. Create some nested categories in Product -> Categories
2. Do `GET /wp-json/wc/gla/mc/mapping/categories`
3. You should get a list of categories with `value` (ID), `label` (name) and `children` (the children)


src/API/Site/Controllers/MerchantCenter/AttributeMappingCategoriesController.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	/**
+	 * Get the item schema properties for the controller.
+	 *
+	 * @return array The Schema properties
+	 */
+	protected function get_schema_properties(): array {
+		return [
+			'value'    => [
```
ianlin: ❔ What do you think if we just call it `id`?

src/API/Site/Controllers/MerchantCenter/AttributeMappingCategoriesController.php
```diff
@@ -0,0 +1,141 @@
...
+				'description'       => __( 'The Category ID.', 'google-listings-and-ads' ),
+				'type'              => 'integer',
+				'validate_callback' => 'rest_validate_request_arg',
+				'readonly'          => true,
+			],
+			'label'    => [
```
ianlin: ❔ Same as the above, maybe we could just call it `name`?

src/API/Site/Controllers/MerchantCenter/AttributeMappingCategoriesController.php
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Class for handling API requests for getting category tree in Attribute Mapping
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter
+ */
+class AttributeMappingCategoriesController extends BaseOptionsController {
+
+
+	/**
+	 * AttributeMappingController constructor.
```
ianlin: 💅  It should be `AttributeMappingCategoriesController`.

src/API/Site/Controllers/MerchantCenter/AttributeMappingCategoriesController.php
```diff
@@ -0,0 +1,141 @@
...
+
+	/**
+	 * Get the item schema name for the controller.
+	 *
+	 * Used for building the API response schema.
+	 *
+	 * @return string
+	 */
+	protected function get_schema_title(): string {
+		return 'attribute_mapping_rules';
```
ianlin: 💅  It should be `attribute_mapping_categories`.

ianlin (approved):
Thanks for adding it. I've tested locally and I can see the expected result with nested categories. I also added some minor comments.

puntope:
Howdy @ianlin  Thanks for checking this PR. 

After working in frontend I decided to build the category tree in the frontend instead of here. So besides your suggestions I modified the controller function a bit. Could you test again?

src/API/Site/Controllers/MerchantCenter/AttributeMappingCategoriesController.php
```diff
@@ -0,0 +1,137 @@
...
+use Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class for handling API requests for getting category tree in Attribute Mapping
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter
+ */
+class AttributeMappingCategoriesController extends BaseOptionsController {
```
ianlin: 💅  Just a minor one, this controller class doesn't use `options` at all, so we could just extends the `BaseController`.

ianlin (approved):
LGTM. Tested again and I got the expected results of categories tree without nested children.

PR #1723, opened 2022-10-10
Attribute Mapping - Categories Frontend

puntope:

Closes part of #1648 .

This PR implements the Category Tree Selector in the frontend.

### Screenshots:

https://user-images.githubusercontent.com/5908855/194890357-9d6b0343-891e-446c-9d90-291ca748330f.mov

1. Create some categories in your Store
2. Go to Attribute Mapping and create some rules, see that the category selector shows your categories
3. Save the category and see that in the table, you can hover in the categories number and see the category names
4. Check that you can modify categories 
5. Create a new test category and assign it in a new rule
6. Delete the category afterwards
7. Check that the category appears as "Category ID %id% (deleted)


js/src/hooks/useCategories.js
```diff
@@ -0,0 +1,113 @@
...
+const SEPARATOR = _x(
+	', ',
+	'the separator for concatenating the categories where the Attribute mapping rule is applied.',
+	'google-listings-and-ads'
+);
+
+const getDeletedCategoryName = ( categoryId ) => {
+	return sprintf(
+		// translators: %d: number of categories.
+		'Category ID %s (deleted)',
```
eason9487: An i18n function is missed to be applied.

js/src/hooks/useCategories.js
```diff
@@ -0,0 +1,113 @@
...
+		[ categoryId ],
+		'google-listings-and-ads'
```
eason9487: I guess this domain string is for the missed `__()` function since `sprintf()` function doesn't ask for it.

js/src/hooks/useCategories.js
```diff
@@ -0,0 +1,113 @@
...
+	);
+};
+
+/**
+ * Returns the category tree used for Tree Select Control
+ * It also maps the deleted (and previously selected) categories
+ *
+ * @param {Array} [selected] The selected category ID's
+ * @return {{tree: *[], names: string, hasFinishedResolution: *}} The tree ready to insert in Tree Select Control, the categories separated by commas and the resolution state
```
eason9487: Looks like the possible types of `hasFinishedResolution` only has the boolean type?

js/src/hooks/useCategories.js
```diff
@@ -0,0 +1,113 @@
...
+const getSelectedNames = ( selected, allCategories ) => {
+	return selected
+		.slice( 0, CATEGORIES_TO_SHOW_IN_TOOLTIP )
+		.map( ( category ) => {
+			return (
+				allCategories.find( ( e ) => e.id.toString() === category )
+					?.name || getDeletedCategoryName( category )
+			);
+		} )
+		.filter( ( x ) => x !== undefined )
```
eason9487: If I understood it right, it shouldn't have any undefined values. Perhaps it doesn't need to filter.

js/src/hooks/useCategories.js
```diff
@@ -0,0 +1,113 @@
...
+		[ categoryId ],
+		'google-listings-and-ads'
+	);
+};
+
+/**
+ * Returns the category tree used for Tree Select Control
+ * It also maps the deleted (and previously selected) categories
+ *
+ * @param {Array} [selected] The selected category ID's
```
eason9487: It would be a bit nicer to indicate the type of `selected` since it's compared to other string type values internally.

```suggestion
 * @param {Array<string>} [selected] The selected category ID's
```

js/src/data/resolvers.js
```diff
@@ -402,3 +403,25 @@ getMappingRules.shouldInvalidate = ( action ) => {
...
+		yield receiveStoreCategories( response );
+	} catch ( error ) {
+		yield handleFetchError(
+			error,
+			__(
+				'There was an error getting the store categories.',
+				'google-listings-and-ads'
+			)
+		);
+		throw error;
```
eason9487: Since the selectors of wp-data won't catch errors internally nor re-throw out them to somewhere for the library consumer side to catch them, this `throw` statement could be omitted.

js/src/attribute-mapping/attribute-mapping-table-categories.js
```diff
@@ -64,27 +70,20 @@ const AttributeMappingTableCategories = ( { categories, condition } ) => {
...
-							'%d category',
-							'%d categories',
-							categoryArray.length,
-							'google-listings-and-ads'
-						),
-						categoryArray.length
-					) }
-				</div>
-			</AppTooltip>
+			{ names.length ? (
```
eason9487: Wonder when the `names` will be 0 lengths?

js/src/attribute-mapping/attribute-mapping-table-categories.js
```diff
@@ -33,27 +47,19 @@ const CATEGORIES_TO_SHOW = 5;
  * @return {JSX.Element|string} The component
  */
 const AttributeMappingTableCategories = ( { categories, condition } ) => {
+	const categoryArray = categories?.split( ',' ) || [];
+	const { names } = useCategories( categoryArray );
```
eason9487, line 51: Although the final rendering result is correct, the `categoryArray` may be `['']` and `names` may be `'Category ID  (deleted)'` if a rule is like `{ "category_condition_type": "ALL", "categories": "" }`. We could consider avoiding unintended values.

eason9487 (commented):
Tested some scenarios and most of them were work well except for an unsure concern described below. I also left some comments and all are minor things.

#### Unsure concern:

A saved rule with a leaf category may cause an unexpected UI state in the tree selector if that leaf category becomes a parent category. Does this feature expect the parent categories are also selectable options?

https://user-images.githubusercontent.com/17420811/196384850-d70eb0b6-2b0d-4e99-b13a-7743052015af.mp4

puntope:
> A saved rule with a leaf category may cause an unexpected UI state in the tree selector if that leaf category becomes a parent category. Does this feature expect the parent categories are also selectable options?

Yes. I'll need to tweak Tree Select control in order to allow that. But it could be done in a separated PR

js/src/attribute-mapping/attribute-mapping-table-categories.js
puntope: If `category_condition_type` is `ALL`, we return ` __( 'All', 'google-listings-and-ads' );`

In any other case, it will have at least one category and that issue doesn't apply as far as I understand.

puntope:
> Yes. I'll need to tweak Tree Select control in order to allow that. But it could be done in a separated PR

At the end I did it in this one since it's quite a minor change. Please let me know it it looks ok for you

puntope:
Howdy @eason9487 . Thanks for the review

I implemented mostly all your changes. This is ready for another round.

js/src/attribute-mapping/attribute-mapping-table-categories.js
eason9487: Yes, I meant we could still avoid the unintended values, they won't be used in the rendering for `ALL` rules though. Just to reduce potential problems in order to improve future maintainability.

js/src/components/tree-select-control/index.js
```diff
@@ -202,6 +205,16 @@ const TreeSelectControl = ( {
 		};
 
 		const descriptors = {
+			isRoot: {
+				/**
+				 * Returns whether this option is the root.
+				 *
+				 * @return {boolean} Trie if the option is the root.
```
eason9487: Typo

```suggestion
				 * @return {boolean} True if the option is the root.
```

eason9487 (commented):
Thanks for the adjustments. After roughly testing the tree selector with the new option `allowsSelectParents`, it seems to break a few of the original UI states. For example:

1. The select all children of a parent is no longer work
2. The checkbox state for showing "all children are selected" is different from before. Now it's changed to the partially selected state.
   image
3. The original state of selected all options is overridden by the state of selected the parent. When the children are collapsed, it couldn't distinguish whether there are any selected children.
   Kapture 2022-10-18 at 19 20 55
4. The parent items are unsearchable.
   Kapture 2022-10-18 at 19 07 59

Probably it needs more consideration if it needs to add this new function to the tree selector.

puntope:
Howdy @eason9487 . Thanks for the extra review

In regards the issue 1 and 2,3. I decided for now to NOT doing them. 

In regards issue 1 . Seems for me that selecting all the children categories when selecting the root will be annoying. In my mind I imagine a user who only wants Clothing root category, but not the children. So User would need to deselect one by one all the children. 

In regards issue 2, since now we have a 3rd state (option selected) I guess it's ok to sacrifice the partial check and promote it to "some children are selected, but not the parent". Otherwise we need to create a 3rd state for "option selected" and other different for "all children selected". This also impacts on issue 3. Thats why I decided to expand the level if a parent option is selected, to say like "hey, there are more options here".   
 
Issue 4 I agree is critical and was solved in last commit. 

I guess since this feature is exclusive for Category Tree Select and it shouldn't have impact in the rest of Tree Select Controls, this is not a blocker. However, I agree that we should discuss this and spend some time with UX to decide the best behaviour.

Let me know your thoughts.

puntope:
If the discussed [here](https://github.com/woocommerce/google-listings-and-ads/pull/1723#issuecomment-1282826765) is ok for you, Could you please test again?

I did this fixes:

- Fix search functionality
- Fix selector for ALL option and for parents as well.
- Remove All option in the Category tree Select, since is not necessary here 

please test with and without allowsSelectParents and with and without ALL Option (`selectAllLabel={ false }` for disable)

Thanks in advance 🙏

js/src/components/tree-select-control/index.js
```diff
@@ -87,6 +88,7 @@ import { ARROW_DOWN, ARROW_UP, ENTER, ESCAPE, ROOT_VALUE } from './constants';
  * @param {Option[]} [props.options] Options to show in the component
  * @param {string[]} [props.value] Selected values
  * @param {number} [props.maxVisibleTags] The maximum number of tags to show. Undefined, 0 or less than 0 evaluates to "Show All".
+ * @param {boolean} [props.allowsSelectParents=false] If true, when selecting a parent it also adds its key in the onChange callback.
```
eason9487: Looks like it's supposed to be *its value*.

```suggestion
 * @param {boolean} [props.allowsSelectParents=false] If true, when selecting a parent it also adds its value in the onChange callback.
```

js/src/components/tree-select-control/index.js
```diff
@@ -272,30 +289,50 @@ const TreeSelectControl = ( {
...
 		};
 
 		const reduceOptions = ( acc, { children = [], ...option } ) => {
+			let match = 0;
+
+			// its a searchable option if this two conditions apply:
+			// 1. its not Root and
+			// 2. is a leaf or selecting parents is allowed.
+			const isSearchable =
+				( ! option.children || allowsSelectParents ) &&
```
eason9487: `option.children` is always undefined at this moment, thus except for the root node, the `isSearchable` of all other options is always true. Similar to https://github.com/woocommerce/google-listings-and-ads/pull/1723#discussion_r998048106, it can show the wanted results eventually, but it would be nicer to avoid the unintended value.

js/src/components/tree-select-control/index.js
```diff
@@ -272,30 +289,50 @@ const TreeSelectControl = ( {
...
 				}
-				option.label = highlightOptionLabel( option.label, match );
 			}
 
-			Object.defineProperties( option, descriptors );
-			acc.push( option );
+			if ( ( match >= 0 && isSearchable ) || option.children?.length ) {
+				// only add the option to the tree if:
+				// a. there are matching children for this option.
+				// b. there is a match for the option itself, and its a searchable option.
```
eason9487: It took me a while to understand these codes. When `isSearching` is false, this code fragment doesn't go through `indexOf` to do the match comparison plus the conditions `(match >= 0 && isSearchable)` to check if there is a match, which makes it confusing to read the code. Using `0` as the default initial value of `match` with the code comment of **b. there is a match**, it may seem like when the code is running to here, the `match` value seems to be the result of matching, but in fact it's still the default initial value when not in searching.

Code comment **a.** also has similar confusion since the children are not always processed through the matching processing but the "match" word makes it looks like they have been processed the matching.

I'd suggest if it's possible, keeping the conditional branches as simple as possible would be easier to read and understand, especially since this function includes recursive calls as well. For example, the original implementation has only three primary conditional branches to separate:
1. Processing for a parent.
2. Processing for searching children.
3. Accumulate a node if it's not returned early by the above two processing.

It could still use similar conditional branches by adjusting it to:

```js
const reduceOptions = ( acc, { children = [], ...option } ) => {
	if ( children.length ) {
		option.children = children.reduce( reduceOptions, [] );

		if ( ! option.children.length && ! allowsSelectParents ) {
			return acc;
		}
	}

	if ( isSearching && option.value !== ROOT_VALUE ) {
		const match = option.label.toLowerCase().indexOf( filter );
		if ( match === -1 ) {
			return acc;
		}
		option.label = highlightOptionLabel( option.label, match );
	}

	Object.defineProperties( option, descriptors );
	acc.push( option );

	return acc;
};
```

js/src/components/tree-select-control/index.js
```diff
@@ -415,8 +452,31 @@ const TreeSelectControl = ( {
...
+				children.map( ( child ) =>
+					flattenChildren( child.children || [] )
+				)
+			);
+
 		let newValue;
-		const changedValues = option.leaves
+		const selectableValues = allowsSelectParents
+			? flattenChildren( option.children ).map(
+					( child ) => delete child.children && child
```
eason9487: Wonder why it needs to perform `delete child.children`? I think usually it's not recommended to mutate the properties of an object which is shared across other scopes, even though it doesn't affect other places for now.

js/src/components/tree-select-control/index.js
```diff
@@ -415,8 +452,31 @@ const TreeSelectControl = ( {
...
+			  )
```
eason9487: WDYT of moving this getter to `descriptors` above, so that all relevant getters would be at the same place, and here could focus on the logic of clicking on a parent node.

js/src/components/tree-select-control/index.js
```diff
@@ -415,8 +452,31 @@ const TreeSelectControl = ( {
...
+			return;
+		}
+
+		const flattenChildren = ( children ) =>
+			Array.prototype.concat.apply(
+				children,
+				children.map( ( child ) =>
+					flattenChildren( child.children || [] )
+				)
+			);
```
eason9487: It's ok to keep it as it's. Just an idea came to my mind that it could be done by `flatMap` to do the array concatenation:
```js
const flattenChildren = ( children ) =>
	children.flatMap( ( child ) => [
		child,
		...flattenChildren( child.children || [] ),
	] );
```

eason9487 (commented):
> I guess since this feature is exclusive for Category Tree Select and it shouldn't have impact in the rest of Tree Select Controls, this is not a blocker.

I also assumed that the business logic of Category selector requires a parent node to be a selectable node. However, the current way brings inconsistent results and confusing UI.

Unfortunately, I'm not pretty sure if the current changes don't have any impact on the rest of the tree select component. Also, from the perspective of component design, these changes are blockers to me. For example, after adding `allowsSelectParents`:
1. The statuses of a checkbox become inconsistent and may confuse users. Especially they may also use the same UI when selecting target audience or shipping settings but the UIs show different behavior and statuses.
2. The results of clicking on the root node and a parent node are also inconsistent. One selects all children and another doesn't.
3. The definition of `option.partialChecked` becomes unstable and its doc is inaccurate. When a parent has a checked leaf, its `partialChecked` could be `true` or `false`, and it depends on `allowsSelectParents` and whether itself is checked.
   https://github.com/woocommerce/google-listings-and-ads/blob/7d6deaca2d6c146d085df1c1807a9ef3ca85c3fd/js/src/components/tree-select-control/index.js#L262-L266
   And it leads to another confusing checkbox state that the upper parents show different partial check states when selecting a parent or a leaf.

   https://user-images.githubusercontent.com/17420811/196722512-5fc5f3d7-c214-4d0b-a1bb-2106c62b6724.mp4
4. The definition of `option.leaves` becomes confusing since it doesn't include parents under itself when `allowsSelectParents` is true.
5. Category selector doesn't need the result of selecting all leaves when clicking on a parent, but other tree selector use cases might need the opposite result for their business logic. It will be harder to add this conflict need and maintain it even if it could be done by adding more props.
6. I think it may be harder to keep pushing these changes upstream if the TreeSelectControl component will be sourced from `@woocommerce/components` eventually.

As a user, I expect the UI offers consistent or at least confusion-less functionalities. As a developer who uses/maintains the `TreeSelectControl` component, I expect the component is easy to know how to use/maintain.

---

If it's possible that we don't have to stick with the completely same UI or component as the current TreeSelectControl, alternative methods could be:

1. Add another standalone leaf option upon each parent node by the `getTree` function in `useCategories`. It looks dumb but at least the logic and states are still consistent and robust.
   
2. Use a similar way as WC's Category management page, show a flattened selector instead of a tree selector.
   
3. I don't know if the business logic allows aggregating the selected values to their nearest parent value when all leaves are checked. If yes, then we could add a new prop for this merging function rather than the `allowsSelectParents` prop.
   Kapture 2022-10-19 at 14 40 45

puntope:
Howdy @eason9487  thanks for your comments.

After reading it we have 4 options I guess. 

1. Modify the Tree Select adding a 3rd state in the UX (option selected, some children selected, all children selected) and tweak documentation and bugs. The complexity will be increased. 

2. Modify the Options data for adding the parent option separately from the parent with children option. This is the easiest and faster to implement. However could be also confusing for the user to see duplications.

3. Use a similar way as WC's Category management page, show a flattened selector instead of a tree selector. Implies re-building the component from scratch again, as well as changing the (already validated) UX. Potentially also implies more work implementing search functionality and further UX requests due context change.

4. I don't know if the business logic allows aggregating the selected values to their nearest parent value when all leaves are checked. If yes, then we could add a new prop for this merging function rather than the allowsSelectParents prop. Could work but we have still the problem of changes in the structure after the categories are selected. For example, when a category was root and after was converted to child.  

----

Saying this I would go to option 2. I can ask for approval from UX. In future I would go for option 1 but it needs more work on that.

eason9487:
> 3\. Use a similar way as WC's Category management page, show a flattened selector instead of a tree selector. Implies re-building the component from scratch again, as well as changing the (already validated) UX. Potentially also implies more work implementing search functionality and further UX requests due context change.

The `SelectControl` component of `@woocommerce/components` should fulfill needed functions like multiple-select, search, custom options, etc. It might not need to rebuild a component from scratch.

puntope:
Howdy @eason9487 

I changed the code to use SelectControl instead of TreeSelectControl as temporary solution. Nikki approved this changes. She will think about tree select control in order to have a 3rd state and then implement the parent selector capabilities. 

This is ready for another round

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -241,7 +237,9 @@ const AttributeMappingRuleModal = ( { rule, onRequestClose = noop } ) => {
 							onCategoriesChange={ ( categories ) => {
 								updateRule( {
 									...newRule,
-									categories: categories.join( ',' ),
+									categories: categories
+										.map( ( category ) => category.value )
```
eason9487: WDYT of moving the mapping inside `AttributeMappingCategoryControl`? In this way, the component asks for a string array on `selectedCategories` prop and the callback value of `onCategoriesChange` could be a string array as well, which has a consistent data shape on the component APIs.

eason9487 (approved):
Thanks for the further adjustments. Left a minor thing about the consistent data shape of component APIs and it's not a blocker.

PR #428, opened 2021-04-06
Add extra product attributes

nima-karimi:
### Changes proposed in this Pull Request:

This PR adds some of the extra product attributes like GTIN, MPN, etc. when syncing them to Google Merchant Center. It adds a new tab to the product edit page and allows the custom attributes to be entered by the user via a form. It also checks for integrations by providing filters to modify the attribute's "acceptable values" and the actual value.

Part of #16.

### Testing instructions
1. Create a new product or edit an existing one
2. There should be a new tab called "Google listings and ads" on the product data section of the edit product page
3. There should be a field to enter MPN and GTIN for the product
4. You can select one of the following for these attributes
        - `Default`: This means that the value will be set to the default value specified on a global/category level (not yet implemented)
        - __Values from integrations__: For now we have only one integration for the Yoast WooCommerce Seo extension, so if you have that installed there will be a value called "From Yoast WooCommerce SEO" displayed there
        - `Enter a custom value`: If this is selected a new text box will be displayed under the current attribute to allow the user to enter their own custom value for this product
5. Choose any of the values and update/publish the product
6. Wait until the product gets synced to Google (you can check the action scheduler job or just sync it manually via the Connection Test page)
7. Check the synced product details on Google Merchant Center and make sure that the attributes you set for them are visible in there
8. Repeat the testing steps for variable products and this time set the attributes per each variation

### Additional information
- We do a simple validation on the attributes on our end. For example, GTIN length is validated but we don't validate its correctness (e.g. check digit validation).
- To test the Yoast WooCommerce SEO integration without installing it (since it's not free), you could set a global constant in `wp-config` called `WPSEO_WOO_VERSION` (with any value) and then create a new metadata for your product in the `wp_postmeta` table with the key `wpseo_global_identifier_values` and the following value as an example:
```
a:6:{s:5:"gtin8";s:8:"12345678";s:6:"gtin12";s:0:"";s:6:"gtin13";s:0:"";s:6:"gtin14";s:0:"";s:4:"isbn";s:0:"";s:3:"mpn";s:9:"123456789";}
```

PR #1727, opened 2022-10-12
Attribute Mapping - Mapping function

puntope:

Closes part of #1648 .

This PR implements the mapping function for Product Syncing.

a. For individual product sync 
b. For batch syncing

It implements the next mappings:

1.  Fixed values
2.  Enum values
3.  Taxonomies
4.  Product fields
5.  Custom created attributes 
6. Yoast Seo GTIN and MPN 

Notice Attributes set in GLA Attributes at product level have priority over rules. So on esting be sure the value is not overrided by some attributes already set in your product

- Empty al your rules in `wp_gla_attribute_mapping_rules `
- check out this PR and Go to Attribute Mapping
- For testing the sync you can use for now connection test page
- You can use debugger to see the final data or just check in your Merchant Center. (notice it takes a few minutes to update)

**Testing Category Conditions**

1. Add one rule with Category condition ALL. Check that the rule applies to all the products
2. Add a category in some products. Add one rule with Category condition ONLY and that category . Check that the rule only applies to the products with that category.
3. Add one rule with Category condition EXCEPT and that category . Check that the rule only applies to the products that are not in that category.

**Testing product field rules:**

1. Create type ALL rules using this fields: Allow backorders, Product title, SKU, Stock Qty, Stock Status, Tax class, Variation title, Weight (raw value, no units), Weight (with units). See that the values are properly loaded in the attribute of your choice.

**Testing taxonomy rules:**

1. Create type ALL rules using Color taxomomy, Size taxonomy and WooCommerce Brands (you should install woocommerce brands)
2. Create a new attribute and see that it works as well.

**Testing YOAST rules:**

 1. Follow https://github.com/woocommerce/google-listings-and-ads/pull/428 to simulate you have WC YOAST installed
 2. Test GTIN Yoast and MPN Yoast for MPN and GTIN

**Testing fixed values**

1. Assign a fixed value to some rule and see it works. 

**Testing Custom Hook attributes** 

1. Create an attribute at product level
2. Load that attribute via filter hook `woocommerce_gla_attribute_mapping_sources_custom_attributes`. Example

```
add_filter(
		'woocommerce_gla_attribute_mapping_sources_custom_attributes',
		function( $values ) {
			return ['total_sales'];
		}
	);
```

3. Test that the attribute appears in the selector and works

**Testing GLA Attributes priority** 

1. Add some attribute in GLA Attributes Tab for a product.
2. Create some rule that applies that attribute. 
3. See how at the end GLA Attribute takes place.


src/Integration/YoastWooCommerceSeo.php
```diff
@@ -96,7 +96,7 @@ protected function add_value_option( array $value_options ): array {
 	 * @return mixed
 	 */
 	protected function get_mpn( $value, WC_Product $product ) {
-		if ( self::VALUE_KEY === $value ) {
+		if ( str_contains( $value, self::VALUE_KEY ) ) {
```
mikkamp: I'm assuming we need to change this match, because we set it to `yoast_seo:mpn` in a mapping rule. Shouldn't we be checking if it starts with the string then?

Also `str_compare` (as well as `str_starts_with`) is a PHP 8.0 function. WordPress added polyfills for this, but it was added in version 5.9. Since we are L-2 compatible we have to still support WordPress 5.8 a little longer.

src/Integration/YoastWooCommerceSeo.php
```diff
@@ -179,7 +179,7 @@ protected function get_yoast_seo_attribute_mapping_group_source(): array {
 	 * @return array The GTIN sources
 	 */
 	protected function get_yoast_seo_attribute_mapping_gtin_sources(): array {
-		return array_merge( self::get_yoast_seo_attribute_mapping_group_source(), [ 'yoast:gtin' => __( 'GTIN Field', 'google-listings-and-ads' ) ] );
+		return array_merge( self::get_yoast_seo_attribute_mapping_group_source(), [ self::VALUE_KEY . ':gtin' => __( 'GTIN Field', 'google-listings-and-ads' ) ] );
```
mikkamp, line 182: For consistency should we change `disabled:yoast` to also match the same key?

src/Product/AttributeMapping/Traits/IsFieldTrait.php
```diff
@@ -94,12 +94,12 @@ public static function get_source_taxonomies(): array {
...
 			'product:backorders'       => __( 'Allow backorders setting', 'google-listings-and-ads' ),
-			'product:product_title'    => __( 'Product title', 'google-listings-and-ads' ),
+			'product:title'            => __( 'Product title', 'google-listings-and-ads' ),
 			'product:sku'              => __( 'SKU', 'google-listings-and-ads' ),
-			'product:stock_qty'        => __( 'Stock Qty', 'google-listings-and-ads' ),
+			'product:stock_quantity'   => __( 'Stock Qty', 'google-listings-and-ads' ),
 			'product:stock_status'     => __( 'Stock Status', 'google-listings-and-ads' ),
 			'product:tax_class'        => __( 'Tax class', 'google-listings-and-ads' ),
-			'product:variation_title'  => __( 'Variation title', 'google-listings-and-ads' ),
+			'product:name'             => __( 'Variation title', 'google-listings-and-ads' ),
```
mikkamp, line 102: Why are we moving variation out of the slug? To me this seems confusing, outside of this trait how do I know whether it's `title` or `name` that points to the parent product name or it's the name of the variation?

src/Product/WCProductAdapter.php
```diff
@@ -903,4 +907,149 @@ public function setTargetCountry( $targetCountry ) {
...
+	 * @param string $source The source to get the value
+	 * @return string|null The source value for this product
+	 */
+	protected function get_source( string $source ) {
+		$source_type = null;
+
+		if ( strpos( $source, ':' ) ) {
+			$source_parts = explode( ':', $source );
+			$source_type  = $source_parts[0];
+			$source_value = $source_parts[1];
```
mikkamp: What if we have some product meta called `gla:brand? This code won't work because it will have set that source as `attribute:gla:brand`

src/Product/WCProductAdapter.php
```diff
@@ -903,4 +907,149 @@ public function setTargetCountry( $targetCountry ) {
...
+	 * @param array $rule The attribute mapping rule
+	 * @return bool True if the rule is applicable
+	 */
+	protected function rule_match_conditions( array $rule ): bool {
+		$attribute                 = $rule['attribute'];
+		$category_condition_type   = $rule['category_condition_type'];
+		$categories                = explode( ',', $rule['categories'] );
+		$wc_product                = $this->get_wc_product();
+		$terms                     = get_the_terms( $wc_product->get_id(), 'product_cat' );
+		$wc_product_categories     = $terms ? wp_list_pluck( $terms, 'term_id' ) : [];
```
mikkamp: The helper function `get_wc_product` is there for when we need to get it externally, but internally it just seems like overkill since we can just access `$this->wc_product`. That part doesn't matter so much, but what I think should be addressed is that it's probably best to cache these category ID's within the class. Probably a wrapper function to get category ID's and then store them in `$this->category_ids`, right now we are extracting the categories as we loop through each rule.

I would personally refactor this code so we don't need to compare categories unless the condition type requires it. So something along the lines of:

```php
		switch ( $category_condition_type ) {
			case AttributeMappingHelper::CATEGORY_CONDITION_TYPE_ONLY:
				return $this->match_categories( $rule['categories'] );
			case AttributeMappingHelper::CATEGORY_CONDITION_TYPE_EXCEPT:
				return ! $this->match_categories( $rule['categories'] );
			default:
				return true;
		}
```

The helper function could then handle all the exploding and intersecting of categories, since we don't need to run all that for certain conditions.

src/Product/WCProductAdapter.php
```diff
@@ -903,4 +907,149 @@ public function setTargetCountry( $targetCountry ) {
...
+		$attributes = [];
+
+		if ( empty( $mapping_rules ) ) {
+			return $this;
+		}
+
+		foreach ( $mapping_rules as $mapping_rule ) {
+			if ( $this->rule_match_conditions( $mapping_rule ) ) {
+				$attribute_id                = $mapping_rule['attribute'];
+				$attributes[ $attribute_id ] = self::format_attribute( apply_filters( "woocommerce_gla_product_attribute_value_{$attribute_id}", $this->get_source( $mapping_rule['source'] ), $this->get_wc_product() ), $attribute_id );
```
mikkamp: We can call `$this->format_attribute` it's not a static function.

Formatting it across multiple lines makes it easier to see what it's doing:
```suggestion
				$attributes[ $attribute_id ] = self::format_attribute(
					apply_filters(
						"woocommerce_gla_product_attribute_value_{$attribute_id}",
						$this->get_source( $mapping_rule['source'] ),
						$this->get_wc_product()
					),
					$attribute_id
				);
```

Also I had a question about the filter name, this allows us to override the mapped value, but as far as I understand if it's also set in the product data, then the filter will be called twice. Should we filter the `mapping_rules` with `gla_attributes` (which is a list of all attributes which are set in the product)? Technically we can skip mapping rules if we know they are going to be overwritten right after.

src/Product/WCProductAdapter.php
```diff
@@ -903,4 +907,149 @@ public function setTargetCountry( $targetCountry ) {
...
+			$source_parts = explode( ':', $source );
+			$source_type  = $source_parts[0];
+			$source_value = $source_parts[1];
+		}
+
+		switch ( $source_type ) {
+			case 'product':
+				return self::get_product_field( $source_value );
+			case 'taxonomy':
+				return self::get_product_taxonomy( $source_value );
```
mikkamp: These aren't static functions, so `$this->` would be nicer.

tests/Tools/HelperTrait/ProductTrait.php
```diff
@@ -313,6 +313,120 @@ public function get_sample_attributes(): array {
...
+			],
+			[
+				'attribute'               => 'ageGroup',
+				'source'                  => 'newborn',
+				'category_condition_type' => 'ALL',
+				'categories'              => '',
+			],
+			[
+				'attribute'               => 'multipack',
+				'source'                  => '0',
```
mikkamp: Isn't it better to test with an actual multipack value here? `'0'` is a falsey value, and assertEquals doesn't strictly type check so I'm not sure how reliable that check would be (possibly change to assertSame)?

Maybe some specific tests to confirm `format_attribute` is mapping all the value.

mikkamp (commented):
Thanks for adding the mapping. I've ran the tests and the data does get sent to the Merchant Center. However I've added some comments as I went through the code.

src/Product/AttributeMapping/Traits/IsFieldTrait.php
puntope: As far as I tested it, when I use `name` it gets the variation title. Whereas `title`, gets always the parent title. So we are not removing variation, it's just that with `name` key it seems to be enough to select the variation...

src/Product/WCProductAdapter.php
puntope: > The helper function get_wc_product is there for when we need to get it externally, but internally it just seems like overkill since we can just access $this->wc_product.

The potential issue of calling directly `$this->wc_product` is that if the getter had some kind of decoration (Let's say a filter.) We could have inconsistencies. Yes, it's a small chance of that happen. But as small it is the "overkill" for using the getter.

puntope:
Howdy @mikkamp Thanks for reviewing the PR

This is ready for another round!

src/Integration/YoastWooCommerceSeo.php
mikkamp: Maybe I'm missing a commit, but I don't see this point addressed in the recent changes. It will still call a non existing function on PHP 7 with WordPress 5.8.

src/Product/AttributeMapping/Traits/IsFieldTrait.php
mikkamp: Thanks for clarifying, I see we are actually using those to retrieve the product props and `variation_title` isn't a function. Although if you choose that mapping it can still be confusing to users what's actually going to be displayed, since it depends on the product type and the length/count of attributes.

```
Simple product:
title = "Simple Product Name", name = "Simple Product Name"

Variation product (1 attribute):
title = "Parent Product Name", name = "Parent Product Name - Attribute Name"

Variation product (more than 2 attributes or long attribute names):
title = "Parent Product Name", name = "Parent Product Name"
```

But I'm fine with leaving it in, although it will probably rarely be used.

src/Product/WCProductAdapter.php
```diff
@@ -903,4 +912,166 @@ public function setTargetCountry( $targetCountry ) {
...
+	 * @param array $mapping_rules The set of rules to apply
+	 */
+	protected function map_attribute_mapping_rules( array $mapping_rules ) {
+		$attributes = [];
+
+		if ( empty( $mapping_rules ) ) {
+			return $this;
+		}
+
+		$this->set_product_category_ids();
```
mikkamp: This at least won't call it for every mapping rule, but it's still not guaranteed that we need those categories. We've already called the function `map_product_categories` couldn't we just cache the ID's from there?

Also I noticed it calls `wc_get_product_cat_ids` which includes the hierarchy (fetching category ancestors), which is different to the way it's done in `set_product_category_ids`. I'm not sure how the category selection works but would it provide better matching?

For example if we have categories like:
```
Clothes
   > Shirts
   > Jackets
```

If the rule is set to apply to clothes, but the product is assigned to shirts, then we'll only have a match if we fetch categories including ancestors.

src/Product/WCProductAdapter.php
mikkamp: Makes sense, that part was just a minor comment. I just noticed that in most other places in that same class it just called the variable directly, so if we would add that filter then we would need to make it consistent everywhere.

src/Product/WCProductAdapter.php
```diff
@@ -903,4 +912,166 @@ public function setTargetCountry( $targetCountry ) {
...
+
+		switch ( $source_type ) {
+			case 'product':
+				return $this->get_product_field( $source_value );
+			case 'taxonomy':
+				return $this->get_product_taxonomy( $source_value );
+			case 'attribute':
+				return $this->get_wc_product()->get_meta( $source_value );
+			default:
+				return $source;
```
mikkamp, line 979: I found this a little confusing (maybe needs a comment to clarify), as I would expect if we can't fetch a value for this mapping then I'd expect to return a value of null.

But apparently if we add mappings with `woocommerce_gla_attribute_mapping_sources` we need to also add a filter for the value and if the value equals the custom source we added then we should return a custom value. Seems confusing but I suppose it was already done that way for YoastSEO, so we can't change that behaviour.

src/Product/WCProductAdapter.php
```diff
@@ -903,4 +912,166 @@ public function setTargetCountry( $targetCountry ) {
...
+	/**
+	 * Get taxonomy source type for attribute mapping
+	 *
+	 * @param string $taxonomy The taxonomy to get
+	 * @return string The taxonomy value
+	 */
+	protected function get_product_taxonomy( $taxonomy ) {
+		$product = $this->get_wc_product();
+		$values  = get_the_terms( $product->get_id(), $taxonomy );
+		return implode( '|', wp_list_pluck( $values, 'name' ) );
```
mikkamp: Is it correct to always combine the values with a `|`?

For example if we use the WooCommerce Brands plugin and then add a rule for mapping the `product_brand` taxonomy. If I assign multiple brands to a product, in the schema it will only include [the first brand](https://github.com/woocommerce/woocommerce-brands/blob/1.6.38/includes/class-wc-brands.php#L366-L370). [This article](https://support.google.com/merchants/answer/6324351?hl=en) also doesn't seem to indicate we should allow multiple brands. When Google does the product check it scans the schema, so I'm assuming it should match or it would flag it as a warning/error.

For [color](https://support.google.com/merchants/answer/6324487?hl=en) it seems to be a little more confusing:
> If you provide multiple color attributes for the same variant, only one will be applied. Additional color attributes will be discarded, and you will receive a warning.

> If your product is made up of several colors, you can specify 1 primary color followed by up to 2 secondary colors that are each separated by a slash ( / ). For example: Red/Green/Black. Don’t submit multiple colors with a comma ( , ). For example, for a red shoe with green and black accents list:"Red/Green/Black". Don’t combine colors together "RedGreenBlack". Don’t submit multiple colors with a comma ("Red", "Green", "Black". If you use commas when entering colors, only one color will be applied.

Size is also another one that mentions using a `/`. Should we maybe replace the `|` with a `/`?

mikkamp (commented):
Thanks for the refactoring that looks and performs a lot better. Was still missing the commit for changing the usage of `str_contains` though. And I left a few additional comments.

src/Product/WCProductAdapter.php
puntope: At this point, in case the source type is not in product|taxonomy|attribute we take the source as static value. 

The custom attributes like yoast: should be treated independently as you mentioned

src/Product/WCProductAdapter.php
puntope: That was a missed piece of code. Actually we DON'T support yet multiple values.

That would require more work in Attribute Definition. So set if it allows multiple values, and the format they should be sent.

src/Product/WCProductAdapter.php
puntope: Btw. I added the comment

src/Integration/YoastWooCommerceSeo.php
puntope: You're right. I changing in other chunk on the code and missed this one.

puntope:
Howdy @mikkamp 

This is ready for another round of reviews! Thanks in advance.

mikkamp (approved):
Thanks for all the changes, I ran some additional tests and I can see the attributes being mapped based on the rules I setup.

As for the unit tests I see that it covers generic mapping, but I don't see any tests to only apply mapping if the categories match. I was especially interested in seeing a test where we confirm the ancestor categories match when checking if the mapping should apply. However it's ok to add this testing in a follow up PR.

puntope:
> However it's ok to add this testing in a follow up PR.

This is noted. Thanks @mikkamp

PR #1238, opened 2022-02-10
Fix retries when deleting products

mikkamp:
### Changes proposed in this Pull Request:

The task in the initial issue was to limit the amount of retries so if we would receive internal errors it wouldn't endlessly retry. However after additional testing it appears the `can_schedule` will check if there is a scheduled task running with the same hook name and the same list of arguments. When we call the action `woocommerce_gla_batch_retry_delete_products` we are still running the scheduled task with the same hook name. So if all the products in the delete batch fail we will have the same list of arguments and the retry will not be scheduled. This can occur easier if we are only deleting one product at a time (for example when we set a single product to not sync or show, or it goes out of stock).

So this PR changes the following things:
- Always reschedule when we called the action `woocommerce_gla_batch_retry_delete_products`
- Reschedule a retry after 1 minute (no use to retry an internal error right away)
- Limit retries to the product syncer threshold (5 tries)
- Filter the delete ready products when processing the DeleteProducts job, this removes any products that aren't ready (like the ones that reached the failure threshold)
- Add unit tests for the added products

Closes #1193

### Detailed test instructions:

Since we can't simulate internal errors in the Google API's we will need to test this PR with our test proxy in the `delete_error` mode so it can return mocked error responses. If you are using docker see instructions [here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/tests/proxy/proxy.md#connect-test-site-to-proxy) on how to get it running.

1. Run the proxy with a command like: `PROXY_MODE=delete_error cd tests/proxy && npm run test-proxy`
2. Use a code snippet to get our site to connect to the proxy:
```php
define( 'WOOCOMMERCE_GLA_CONNECT_SERVER_URL', 'http://localhost:5500' );
```
3. Connect your Merchant account and sync all or some products (ensure we waited for the syncing to complete)
4. Edit one of the synced products and mark it as "Don't Sync and show" and save it
5. This will trigger a product delete, since we mocked the errors it will trigger a retry until it reaches the threshold of 5
6. Wait for at least 6 minutes till it completes all the retries
7. View the GLA log file in WooCommerce > Status > Logs
8. Confirm that you see 4 failures in the log with a message like (the 5th time it won't be retried so it will only show debug messages that it failed):
```
2022-02-10T08:12:24+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer::handle_delete_errors Internal API errors while deleting the following products: Array
(
    [online:en:US:gla_1234] => 1234
)
``` 
[...]

PR #880, opened 2021-07-13
Don't sync invisible products to MC

nima-karimi:
### Changes proposed in this Pull Request:

This PR modifies the `ProductHelper::is_sync_ready` method so that it also checks the product's visibility. In addition to the previous conditions, it now returns false for a product when:
 - `woocommerce_hide_out_of_stock_items` option is enabled and the product is out of stock
 - If it's an invisible variation (i.e. have no price or it's disabled)

The WooCommerce core `woocommerce_hide_invisible_variations` filter is also used here to check whether the variation should be displayed even if it's invisible.

These conditions were already checked for variations in `SyncerHooks` class and a few other places where the method accepts a variable product and tries to fetch its "available" variations via the `\WC_Product_Variable::get_available_variations` method. By adding these conditions to `ProductHelper::is_sync_ready` we'll also check for the "availability" (or visibility) of all other product types just like we do for variations and make things more consistent.

One major effect of this change is that if the user has enabled the `woocommerce_hide_out_of_stock_items` we no longer sync the `out-of-stock` products to Merchant Center and also remove the previously synced products once the product goes out of stock. 

However, if the user already has some out-of-stock products and then enables the `woocommerce_hide_out_of_stock_items` option, we do not remove anything. This should probably be done by watching the WooCommerce events and trigger a refresh job if any settings change. I didn't include this part in the PR because we also don't do this for any other settings changes (e.g. country, language, etc.) and I first wanted to discuss whether we need to.

### Detailed test instructions:

1. Create an out-of-stock product
2. Check the `Out of stock visibility` option  under "WooCommerce > Settings > Products > Inventory"
3. Sync the out of stock product via Connect Test page and check that it's NOT synced to MC
4. Uncheck the `Out of stock visibility` option  under "WooCommerce > Settings > Products > Inventory"
5. Sync the out of stock product via Connect Test page and check that it's synced to MC
6. Repeat the above with a variation product
7. Disable a variation by unchecking the "Enabled" option
8. Sync it via Connect Test page and check that it's NOT synced to MC
9. Remove the variation price
10. Sync it via Connect Test page and check that it's NOT synced to MC
11. Add the following filter and then disable a variation
     ```
     add_filter( 
	     'woocommerce_hide_invisible_variations', 
	     function () {
		     return false;
	     }
     );
     ```
12. Sync it via Connect Test page and check that it's synced to MC
13. Run phpunit and make sure all the tests pass

### Changelog entry

> Update - Stop syncing invisible products and variations

PR #1728, opened 2022-10-14
Fix incorrect product statistics count

ianlin:
## Changes proposed in this Pull Request:

Project post: pcTzPl-1a9-p2

This is a follow up for #1713, which is part of the project #1610.

The number returned by product statistics API are not aligned with the syncable products count because it includes the variation products. There are two parts when getting product statistics: counting `synced` products and `not_synced` products, we need to group the variations as one for the same parent for both parts.

### 1. Group variation products when counting `synced` products.

The solution is inspired by @mikkamp's [comments in the PR 1713](https://href.li/?https://github.com/woocommerce/google-listings-and-ads/pull/1713#discussion_r987921062) by using two existing data `$this->product_data_lookup` and `$this->product_statuses` from `MerchantStatuses.php`, the format of both data are:

**product_statuses**

```json
{
  "products": {
    "602": {
      "disapproved": 1
    },
    "603": {
      "approved": 1
    }
    "604": {
      "disapproved": 1
    }
  },
  "parents": {
    "589": {
      "approved": 1,
      "disapproved": 1
    }
  }
}
```

**product_data_lookup**

```json
{
  "602": {
    "name": "Name A",
    "visibility": [],
    "parent_id": 589,
  },
  "603": {
    "name": "Name B",
    "visibility": [],
    "parent_id": 589,
  },
  "604": {
    "name": "Name C",
    "visibility": [],
    "parent_id": 0,
  },
}
```

The modified code loop through `product_statuses` and check `product_data_lookup` to see if a product has a parent. From the above data we can get two `synced` products: `589` and `604`, where `589` is a variable product that has two variations `602` and `603`.

@mikkamp also suggested in pcTzPl-I1-p2#comment-1839 that we should consider a case where one variation is `approved` and the other is `disapproved`. So we add a new data `product_statistics_priority` to assign each status a priority (in this case `disapproved` has higher priority) then we can get the correct status for each product. The new data's format looks like this:

```php
$product_statistics_priority = [
  MCStatus::APPROVED           => 6,
  MCStatus::PARTIALLY_APPROVED => 5,
  MCStatus::EXPIRING           => 4,
  MCStatus::PENDING            => 3,
  MCStatus::DISAPPROVED        => 2,
  MCStatus::NOT_SYNCED         => 1,
];
```

### 2. Get **syncable but not synced** products

- **GMC status**: product statuses fetched from GMC and stored in `wp_postmeta` with meta key `_wc_gla_mc_status` and has statues:
  - approved
  - partially_approved
  - expiring
  - pending
  - disapproved
  - not_synced
- **Sync status**: product sync statuses stored in `wp_postmeta` with meta key `_wc_gla_sync_status` and has statuses:
  - synced
  - not-synced
  - has-errors
  - pending

The original implementation to get the `not_synced` products is: finding all the `simple` and `variable` products where their GMC status is `not_synced`. While it already excludes the `variation` products but the result would contains **unsyncable** products. Our goal is to get the **syncable but not synced** products, to achieve this we could simply get all the `simple` and `variable` products where their `SyncStatus` is not `synced`. The reason why we can use `SyncStatus` is, only the **syncable** products would have a status recorded in `SyncStatus`.

### <a href="#trash-products" id="trash-products">3.</a> Fix a problem that "trashing" a product would never really trigger delete batch requests

In [DeleteProducts.php](https://github.com/woocommerce/google-listings-and-ads/blob/d1cb95b835a248f3035026f417aecd416facf28f/src/Jobs/DeleteProducts.php#L40), the format of `$product_id_map` is:

```php
[
  "online:en:TW:gla_1234" => 1234,
  "online:en:TW:gla_5678" => 5678,
]
```

However in `ProductRepository.php` the method `find_delete_product_ids()` accepts an array of WC product IDs. But even though we use `array_values` on `$product_id_map` to pass the array of IDs, the product is already in `trash` status so [WC_Product_Query](https://github.com/woocommerce/woocommerce/wiki/wc_get_products-and-WC_Product_Query#general) still couldn't get the product unless we set the args `'status' => 'trash'` in the query.

### Use an existing Merchant Centre account

1. Disconnect all accounts from GLA settings.
2. Clear `SyncStatus` and `MCStatus` from the DB.
    ```sql
    DELETE FROM wp_postmeta WHERE meta_key = '_wc_gla_sync_status';
    DELETE FROM wp_postmeta WHERE meta_key = '_wc_gla_mc_status';
    ```
3. Go through the onboarding process. In the step 1 `Set up your accounts`, choose an existing Merchant Centre account.
4. In the step 4 `Complete your campaign` you will see `Your product listings are ready to be uploaded
N products`.
5. Complete the onboarding by clicking `Skip this step for now`, as we don't need the Ads account for the test.
6. Check if the products sync jobs is completed from `/wp-admin/admin.php?page=wc-status&tab=action-scheduler&s=update_all_products&action=-1&action2=-1&orderby=schedule&order=desc&paged=1`
7. When the product sync job is done, call `GET /mc/product-statistics/refresh` or go to GLA product feed page, the total number should be identical to the `N` products in step 4.

### Create a new Merchant Centre account

Follow the above steps but in step 3 create a new Merchant Centre account.

### Add, update, or delete a product

- Adding, updating, or deleting a product, the number returned by `GET /mc/product-statistics/refresh` should be changed accordingly.

## Changelog entry

> Update - Fix incorrect product statistics count

### DIFF
--- a/src/Jobs/DeleteProducts.php
+++ b/src/Jobs/DeleteProducts.php
@@ -37,11 +37,6 @@ class DeleteProducts extends AbstractProductSyncerJob implements StartOnHookInte
 	 * @throws ProductSyncerException If an error occurs. The exception will be logged by ActionScheduler.
 	 */
 	public function process_items( array $product_id_map ) {
-		$ready_ids = $this->product_repository->find_delete_product_ids( $product_id_map );
-
-		// Exclude any ID's which are not ready to delete.
-		$product_id_map = array_intersect( $product_id_map, $ready_ids );
-
 		$product_entries = BatchProductIDRequestEntry::create_from_id_map( new ProductIDMap( $product_id_map ) );
 		$this->product_syncer->delete_by_batch_requests( $product_entries );
 	}
--- a/src/Jobs/DeleteProducts.php
+++ b/src/Jobs/DeleteProducts.php
@@ -37,6 +37,12 @@ class DeleteProducts extends AbstractProductSyncerJob implements StartOnHookInte
 	 * @throws ProductSyncerException If an error occurs. The exception will be logged by ActionScheduler.
 	 */
 	public function process_items( array $product_id_map ) {
+		$product_ids = array_values( $product_id_map );
+		$ready_ids   = $this->product_repository->find_delete_product_ids( $product_ids );
+
+		// Exclude any ID's which are not ready to delete.
+		$product_id_map = array_intersect( $product_id_map, $ready_ids );
+
 		$product_entries = BatchProductIDRequestEntry::create_from_id_map( new ProductIDMap( $product_id_map ) );
 		$this->product_syncer->delete_by_batch_requests( $product_entries );
 	}
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -295,9 +295,9 @@ class ProductHelper implements Service {
 	 *
 	 * @param int $product_id
 	 *
-	 * @return WP_Post
+	 * @return WP_Post|null
 	 */
-	public function get_wc_product_by_wp_post( int $product_id ): WP_Post {
+	public function get_wc_product_by_wp_post( int $product_id ): ?WP_Post {
 		return get_post( $product_id );
 	}
 
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -14,6 +14,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Value\SyncStatus;
 use Google\Service\ShoppingContent\Product as GoogleProduct;
 use WC_Product;
 use WC_Product_Variation;
+use WP_Post;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -287,6 +288,17 @@ class ProductHelper implements Service {
 	 */
 	public function get_wc_product( int $product_id ): WC_Product {
 		return $this->wc->get_product( $product_id );
+	}
+
+	/**
+	 * Get WooCommerce product by WP get_post
+	 *
+	 * @param int $product_id
+	 *
+	 * @return WP_Post
+	 */
+	public function get_wc_product_by_wp_post( int $product_id ): WP_Post {
+		return get_post( $product_id );
 	}
 
 	/**
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -103,7 +103,11 @@ class ProductHelper implements Service {
 	 */
 	public function mark_as_unsynced( WC_Product $product ) {
 		$this->meta_handler->delete_synced_at( $product );
-		$this->meta_handler->update_sync_status( $product, SyncStatus::NOT_SYNCED );
+		if ( ! $this->is_sync_ready( $product ) ) {
+			$this->meta_handler->delete_sync_status( $product );
+		} else {
+			$this->meta_handler->update_sync_status( $product, SyncStatus::NOT_SYNCED );
+		}
 		$this->meta_handler->delete_google_ids( $product );
 		$this->meta_handler->delete_errors( $product );
 		$this->meta_handler->delete_failed_sync_attempts( $product );
--- a/src/Product/ProductRepository.php
+++ b/src/Product/ProductRepository.php
@@ -6,7 +6,7 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\Product;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
 use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\ChannelVisibility;
-use Automattic\WooCommerce\GoogleListingsAndAds\Value\MCStatus;
+use Automattic\WooCommerce\GoogleListingsAndAds\Value\SyncStatus;
 use WC_Product;
 
 defined( 'ABSPATH' ) || exit;
@@ -264,9 +264,9 @@ class ProductRepository implements Service {
 			'type'       => $types,
 			'meta_query' => [
 				[
-					'key'     => ProductMetaHandler::KEY_MC_STATUS,
-					'compare' => '=',
-					'value'   => MCStatus::NOT_SYNCED,
+					'key'     => ProductMetaHandler::KEY_SYNC_STATUS,
+					'compare' => '!=',
+					'value'   => SyncStatus::SYNCED,
 				],
 			],
 		];
--- a/src/Product/ProductRepository.php
+++ b/src/Product/ProductRepository.php
@@ -81,12 +81,13 @@ class ProductRepository implements Service {
 	 * Find and return an array of WooCommerce product objects based on the provided product IDs.
 	 *
 	 * @param int[] $ids    Array of WooCommerce product IDs
+	 * @param array $args   Array of WooCommerce args (except 'return'), and product metadata.
 	 * @param int   $limit  Maximum number of results to retrieve or -1 for unlimited.
 	 * @param int   $offset Amount to offset product results.
 	 *
 	 * @return WC_Product[] Array of WooCommerce product objects
 	 */
-	public function find_by_ids( array $ids, int $limit = -1, int $offset = 0 ): array {
+	public function find_by_ids( array $ids, array $args = [], int $limit = -1, int $offset = 0 ): array {
 		$args['include'] = $ids;
 
 		return $this->find( $args, $limit, $offset );
@@ -163,7 +164,8 @@ class ProductRepository implements Service {
 	 * @return array
 	 */
 	public function find_delete_product_ids( array $ids, int $limit = - 1, int $offset = 0 ): array {
-		$results = $this->find_by_ids( $ids, $limit, $offset );
+		$args    = [ 'status' => 'trash' ];
+		$results = $this->find_by_ids( $ids, $args, $limit, $offset );
 		return $this->product_filter->filter_products_for_delete( $results )->get_product_ids();
 	}
 
--- a/src/Product/ProductRepository.php
+++ b/src/Product/ProductRepository.php
@@ -270,6 +270,11 @@ class ProductRepository implements Service {
 					'compare' => '!=',
 					'value'   => SyncStatus::SYNCED,
 				],
+				[
+					'key'     => ProductMetaHandler::KEY_VISIBILITY,
+					'compare' => '=',
+					'value'   => ChannelVisibility::SYNC_AND_SHOW,
+				],
 			],
 		];
 
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -350,7 +350,7 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 				continue;
 			}
 
-			$wc_product = get_post( $wc_product_id );
+			$wc_product = $product_helper->get_wc_product_by_wp_post( $wc_product_id );
 			if ( ! $wc_product || 'product' !== substr( $wc_product->post_type, 0, 7 ) ) {
 				// Should never reach here since the products IDS are retrieved from postmeta.
 				do_action(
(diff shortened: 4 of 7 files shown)


src/Product/ProductHelper.php
```diff
@@ -289,6 +290,17 @@ public function get_wc_product( int $product_id ): WC_Product {
...
 	}
 
+	/**
+	 * Get WooCommerce product by WP get_post
+	 *
+	 * @param int $product_id
+	 *
+	 * @return WP_Post
+	 */
+	public function get_wc_product_by_wp_post( int $product_id ): WP_Post {
```
mikkamp: After we call this function we double check to make sure the product has a value (is not null), but the return type of this function does not allow null. Do we need a nullable return type?

src/Jobs/DeleteProducts.php
```diff
@@ -37,11 +37,6 @@ public function get_name(): string {
 	 * @throws ProductSyncerException If an error occurs. The exception will be logged by ActionScheduler.
 	 */
 	public function process_items( array $product_id_map ) {
-		$ready_ids = $this->product_repository->find_delete_product_ids( $product_id_map );
-
-		// Exclude any ID's which are not ready to delete.
-		$product_id_map = array_intersect( $product_id_map, $ready_ids );
-
```
mikkamp: By removing these lines we are reverting the PR #1238
I can see the flaws in the filter, but I don't think we should revert it completely otherwise we'll reintroduce the endless retry upon failure. I can see the logic of adding the status equals trash, but if we look at the sync hooks. It's hooked into both when a product is trashed and when it's deleted completely. In the latter case it won't be able to query the product data at all. Which means:
- we should still try the initial `delete_by_batch_request`
- we should not retry a product that is not available in the database

In that case maybe we should move the filtering from process_items to the code block where the retries are scheduled. If we filter the ID map at that point then we can remove any no longer existing products and ones that retried too many times.

Although I would move this to a separate issue/PR.

mikkamp (commented):
I left some comments as I went through the code. When testing I was unable to get the two amounts to match up (I've got quite the mix of products).

For the syncable product count I get back:
```json
{
	"count": 301
}
```

And for the product statistics I get:
```json
{
	"timestamp": 1668615754,
	"statistics": {
		"active": 0,
		"expiring": 0,
		"pending": 0,
		"disapproved": 205,
		"not_synced": 111
	},
	"scheduled_sync": 0
}
```

I'll take another look tomorrow to see if I can pull up some exact product ID's and see why they aren't matching up.

src/Jobs/DeleteProducts.php
ianlin: Ah, my brain wasn't working properly yesterday, I actually forgot to add back the deleted code so even though I added "status equals trash" logic in `find_delete_product_ids` method from `ProductRepository.php`, there is no one calling this method. So if I added it back like below It would try to find the product with status `trash` then exclude the ones which are not ready to delete. Note that `array_intersect( $product_id_map, $ready_ids );` can work as `array_intersect` is working through the array values.

```diff
--- a/src/Jobs/DeleteProducts.php
+++ b/src/Jobs/DeleteProducts.php
@@ -37,6 +37,12 @@ class DeleteProducts extends AbstractProductSyncerJob implements StartOnHookInte
   * @throws ProductSyncerException If an error occurs. The exception will be logged by ActionScheduler.
   */
  public function process_items( array $product_id_map ) {
+   $product_ids = array_values( $product_id_map );
+   $ready_ids   = $this->product_repository->find_delete_product_ids( $product_ids );
+
+   // Exclude any ID's which are not ready to delete.
+   $product_id_map = array_intersect( $product_id_map, $ready_ids );
+
    $product_entries = BatchProductIDRequestEntry::create_from_id_map( new ProductIDMap( $product_id_map ) );
    $this->product_syncer->delete_by_batch_requests( $product_entries );
  }
```

> but if we look at the sync hooks. It's hooked into both when a product is trashed and when it's deleted completely. In the latter case it won't be able to query the product data at all.

Correct me if I'm wrong, from the WooCommerce admin products page I can only delete a `trash` product completely. Which means based on the above code, we would call `delete_by_batch_request` when a product status is becoming `trash`, and we would not call `delete_by_batch_request` again when the product is deleted completely. Is it correct?

mikkamp: A product can still be force deleted (without going to trash) either through WP-CLI, API or making use of a plugin.
In that case we have the ID mapping scheduled for removal, but we can't load any product details when the scheduled task actually starts running. So I imagine in that case the filter wouldn't work and the product would get excluded from the batch.

mikkamp:
> I'll take another look tomorrow to see if I can pull up some exact product ID's and see why they aren't matching up.

So I collected the arrays of product ID's and did some comparisons. For counting syncable products I should have had 300 instead of 301. The extra one was because I have a variable subscription product, but the Subscription plugin was deactivated, in which case it detects the variations as a regular variable product. (This might be acceptable to leave as an edge case)

For the counts I still had 16 products too many in the unsynced count, it seems this was from products that were synced once upon a time but the channel visibility was later toggled to "do not sync and show". Modifying the meta query for unsynced products seems to resolve that:

```php
			'meta_query' => [
				[
					'key'     => ProductMetaHandler::KEY_SYNC_STATUS,
					'compare' => '!=',
					'value'   => SyncStatus::SYNCED,
				],
				[
					'key'     => ProductMetaHandler::KEY_VISIBILITY,
					'compare' => '=',
					'value'   => ChannelVisibility::SYNC_AND_SHOW,
				],
			],
```

That gets me closer but I'm still off by one product. Which is a product that previously synced but I set the catalog visibility to hidden, which means it becomes unsyncable. Not sure what the best way to resolve that is. Should either the sync status or the channel visibility be reset in such a case?

ianlin:
Thanks @mikkamp, those are very valuable results. I will update the meta query for unsynced products.

> Which is a product that previously synced but I set the catalog visibility to hidden, which means it becomes unsyncable. Not sure what the best way to resolve that is. Should either the sync status or the channel visibility be reset in such a case?

When setting a product's `catalog visibility` to hidden, the `channel visibility` on product UI display `Don't sync and show`, while the product's meta value `_wc_gla_visibility` in the database is still `sync-and-show`. I think it would be more consistent to update the channel visibility meta to `dont-sync-and-show` as well in such case, what do you think?

src/Jobs/DeleteProducts.php
ianlin: Right, I think in this case it would be better to move this into different issue/PR so we don't block this PR. Do you think I should revert the commits 3f7a2cc and 7dfd462 then create an issue, or adding a new commit including the change in my [previous comment](https://github.com/woocommerce/google-listings-and-ads/pull/1728#discussion_r1024828110) and create an issue?

mikkamp:
> When setting a product's catalog visibility to hidden, the channel visibility on product UI display Don't sync and show, while the product's meta value _wc_gla_visibility in the database is still sync-and-show. I think it would be more consistent to update the channel visibility meta to dont-sync-and-show as well in such case, what do you think?

If I recall correctly there was a reason why we didn't update the visibility. In PR #880 is where it was implemented not to sync products hidden from the catalog. This PR also mentions products out of stock, I don't think I had any test products for this, but the same thing would probably apply. So we might want to test how that scenario works, what we don't want is the following scenario:

1. Product is purchased and goes out of stock
2. Stock update triggers removal of synced product from Merchant Center
3. Since it is no longer syncable we set the channel visibility to do-not-sync-and-show
4. Stock is added (often done through external system) but channel visibility is not restored (so it requires a manual action to restore)

Step number 3 is not being done right now, so we'll have to be careful we don't introduce unwanted side effects. Maybe it's those cases that we want to have an extra sync status available like TEMPORARILY_UNSYNCED, or just delete the sync status (which throws it in the group of unsyncable products).

ianlin:
Thanks for the detailed explanation on the history of why we made it. I think removing the sync status for such case (catalog visibility becomes hidden) so it falls into the group of unsyncable products is a good idea, as the product becomes unsyncable for sure.

I'm looking for where is the best place to remove the sync status, at first I found

https://github.com/woocommerce/google-listings-and-ads/blob/a9825c16f0535b0da5c5512e1c6bffe7fc336950/src/Product/SyncerHooks.php#L191-L193

Setting catalog visibility to hidden the synced product becomes unsyncable (not sync ready) so the code above will trigger delete products job to delete a product from GMC. However adding the code to remove sync status here would be overridden by the code in the `ProductSyncer.php`:

https://github.com/woocommerce/google-listings-and-ads/blob/a9825c16f0535b0da5c5512e1c6bffe7fc336950/src/Product/ProductSyncer.php#L210

There are some options we could do:

1. In `ProductSyncer.php`'s `delete_by_batch_requests()` method, remove sync status if the product's `catalog_visibility` is `hidden` (or maybe checking `product->is_visible()` is `false`).
    - However this could leads to some performance problems as we only have product ID in method `delete_by_batch_requests()`, we need to get all the product objects from the array.
2. Not sure if it makes sense, another thought I have is to just remove the sync status in `delete_by_batch_requests()` method as when the product is trashed or deleted it becomes unsyncable.

mikkamp:
I don't think just checking `$product->is_visible()` is enough. What I meant to imply with the going out of stock scenario is that there are many reasons why a product might become `unsyncable` after it was previously `syncable`. So I think the best source of truth is to call `$product_helper->is_sync_ready( $product )`.

`batch_helper->mark_as_unsynced` already loads the product object so we won't be using extra resources there. Maybe we should pass a parameter to `mark_as_unsynced` which if true will check `is_sync_ready` and if it's not will remove sync status. Something like:

```php
if ( $remove_sync_status && $this->is_sync_ready( $product ) {
	$this->meta_handler->delete_sync_status( $product );
} else {
	$this->meta_handler->update_sync_status( $product, SyncStatus::NOT_SYNCED );
}
```

We might not even need the `$remove_sync_status` variable (would need to check what effect that would have), but I think it could also be safer to only call that after delete_by_batch to limit the scope of this change.

src/Jobs/DeleteProducts.php
mikkamp: I think fixing it temporarily by changing the status to trash (as long as we aren't reverting previous behaviour) is a good option. And then create a separate followup issue for filtering later, since that's covering behaviour that was already broken.

src/Product/ProductHelper.php
ianlin: Updated in 6c84398.

ianlin:
Hello @mikkamp, I've added three commits 40f189a, 41c52e0, and 39e3190 to address the problems you've mentioned above so it's ready for another round of review.

src/Jobs/DeleteProducts.php
ianlin: The follow up issue is created in #1785.

src/Product/ProductRepository.php
```diff
@@ -163,7 +164,8 @@ public function find_sync_ready_products( array $args = [], int $limit = - 1, in
 	 * @return array
 	 */
 	public function find_delete_product_ids( array $ids, int $limit = - 1, int $offset = 0 ): array {
-		$results = $this->find_by_ids( $ids, $limit, $offset );
+		$args    = [ 'status' => 'trash' ];
```
mikkamp, line 167: This is now working for items that need to be deleted because they are in the trash. But for items that get deleted from Merchant Center for other reasons are not found/updated. For example I set a product to catalog visibility = hidden, which will trigger it to be deleted from the Merchant Center, but it's status remains "published".

Without any status it was previously querying the [default WP statuses](https://github.com/woocommerce/woocommerce/wiki/wc_get_products-and-WC_Product_Query#general) so it seems we should include `trash` with the default array.

```suggestion
		$args    = [ 'status' => [ 'draft', 'pending', 'private', 'publish', 'trash' ] ];
```

mikkamp (approved):
Thanks for the additional changes. I rechecked the calculations and I'm now getting matching numbers.

I just found one additional problem with it querying non-trash products. I added a suggestion to fix that and we can continue adding some tests for it in the followup PR.

This PR looks good to go then, so I'll approve it.

src/Product/ProductRepository.php
ianlin: Thanks for catching this, I've updated the issue #1785 so it will be handled in the next PR. Merging this PR first.

PR #1731, opened 2022-10-14
Attribute Mapping - Syncer

puntope:

Closes part of #1648 .

This PR implements a Syncer logic to Update the products once the Attribute Mapping Rules change.

- Create hooks on delete/update/create rule
- Create logic for run delayed create batch for updating the products
- Create API to check if the current process is syncing in order to show that information in the frontend

### Steps for testing

1. Checkout this PR and delete/update/create a new rule in Attribute Mapping.
2. Go to WooCommerce- Status - Scheduled Actions and see how `gla/jobs/update_all_products/create_batch` was scheduled delayed in 30 minutes.
3. Request  `GET /wp-json/wc/gla/mc/mapping/sync`. Yo will receive `is_syncing: true` as response
4. Run manually the action (or wait 30 minutes)
5. Request  `GET /wp-json/wc/gla/mc/mapping/sync`. Yo will receive `is_syncing: false` as response


src/Jobs/UpdateAllProducts.php
```diff
@@ -36,7 +36,26 @@ public function get_name(): string {
...
+		}
+	}
+
+	/**
+	 * Check if the current job is currently running
+	 *
+	 * @return bool True if the job is running.
+	 */
+	public function is_syncing(): bool {
+		return $this->is_running( [ 1 ] );
```
mikkamp: For a batched job this just checks if the first batch job with the create hook is scheduled. The name might be confusing but that means we are actually just getting the status "is scheduled". But realistically it schedules batches of the following two jobs:

```
gla/jobs/update_syncable_products_count/create_batch
gla/jobs/update_syncable_products_count/process_items
```

Syncing is only completed when all those jobs have been run, not just the first one. Do you think it would be better to rely on the [ProductSyncStats](https://github.com/woocommerce/google-listings-and-ads/blob/2.2.0/src/Jobs/ProductSyncStats.php) class, it's what we use for the product statistics endpoint? It does check more jobs but that might not be a bad thing as it will just reflect that it's syncing products in the background.

> **Note**
Having looked at that class I think we could improve the search by adding the parameter `'group' => 'gla'` otherwise we could potentially be looping through 1000's of pending subscriptions.

src/Event/StartProductSync.php
```diff
@@ -41,6 +41,13 @@ function() {
 				$this->on_settings_sync();
 			}
 		);
+
+		add_action(
+			'woocommerce_gla_rules_change',
```
mikkamp: Since this is a global action do we want to clarify it's a mapping rules change?

```suggestion
			'woocommerce_gla_mapping_rules_change',
```

src/Jobs/UpdateAllProducts.php
```diff
@@ -36,7 +36,26 @@ public function get_name(): string {
...
 		$this->product_syncer->update( $products );
 	}
+
+	/**
+	 * Schedules a delayed batched job
+	 *
+	 * @param int $delay The delay time in seconds
+	 */
+	public function schedule_delayed( int $delay ) {
+		if ( $this->can_schedule( [ 1 ] ) ) {
```
mikkamp, line 51: Just noting that this means we can schedule a new batch during the time it's syncing. However the more I think of it, it's probably an edge case that they update rules after 30 minutes and their products take a long time to sync, so we can leave this as is.

mikkamp (commented):
Thanks for the changes, it's working as expected according to the test instructions.

Although the endpoint `gla/mc/mapping/sync` only returns true during the time the first task is scheduled and not during syncing, which doesn't seem to match with the frontend PR where it shows the message "Your products are synced with your rules". It would show this message too early.

Also I'm not sure what the final design decision was but it could still take another 30 minutes for the newly synced changes to show up in the Merchant Center, does that need to be communicated in that message?

src/Jobs/UpdateAllProducts.php
puntope: > Having looked at that class I think we could improve the search by adding the parameter 'group' => 'gla' otherwise we could potentially be looping through 1000's of pending subscriptions.

What do you mean with this? And where shall I add that group? As far as I see it's automatically added in the search method.

puntope: As far as I see it's automatically added in the search method.

puntope:
Howdy @mikkamp 

I implemented the suggested solution based on ProductSyncStats.

Notice now the API returns:

is_scheduled --> True if there is a number greater than 0 in the SyncStats count
last_sync --> Timestamp for the last completed job. Handled by creating an Option on `handle_complete` function.

puntope:
Btw, notice now a normal UpdateAllProducts job also creates the Option for the last sync. 

I guess this is ok, if not, let me know and I can create a new Job extending UpdateAllProducts.

src/Jobs/UpdateAllProducts.php
mikkamp: Yup, you are right, I missed [the line where the group is added](https://github.com/woocommerce/google-listings-and-ads/blob/2.2.0/src/ActionScheduler/ActionScheduler.php#L147) in the ActionScheduler class.

src/API/Site/Controllers/AttributeMapping/AttributeMappingSyncerController.php
```diff
@@ -0,0 +1,116 @@
...
+			'is_scheduled' => [
+				'description'       => __( 'Indicates if the products are currently syncing', 'google-listings-and-ads' ),
+				'type'              => 'boolean',
+				'validate_callback' => 'rest_validate_request_arg',
+				'readonly'          => true,
+				'context'           => [ 'view' ],
+			],
+			'last_sync'    => [
+				'description'       => __( 'Timestamp with the last sync.', 'google-listings-and-ads' ),
+				'type'              => 'string',
```
mikkamp, line 97: Not that it really matters but a timestamp doesn't have to be string. Making it numeric would be consistent with how it's done here: https://github.com/woocommerce/google-listings-and-ads/blob/2.2.0/src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php#L128-L133

Also the description might be clearer with something like:
> Timestamp when the last product sync completed.

mikkamp (approved):
Thanks for the changes, it's working as expected.

I just left a small comment about the schema for the sync controller, but that's not blocking.

Regarding "a normal UpdateAllProducts job also creates the Option for the last sync", I see that as a benefit because it's still the last time the products were synced.

src/API/Site/Controllers/AttributeMapping/AttributeMappingSyncerController.php
puntope: Noted. To be implemented in future tweaks.

PR #1733, opened 2022-10-17
Attribute Mapping - Sync Frontend

puntope:

Closes part of #1648 .

This PR implements the Frontend par of https://github.com/woocommerce/google-listings-and-ads/pull/1731

That is to Poll the sync API endpoint and update the message in accordance. 

### Screenshots:

https://user-images.githubusercontent.com/5908855/196202508-c0345216-b5a1-4f93-9ef6-92dcd017cc6f.mov

1. Go to Attribute Mapping. See the message "Your products are synced with your rules"
2. Create a new rule and see how the message changes to "Your products are being synced with your rules" after a few seconds.
3. Without closing the window.  Go to WooCommerce - Status - Scheduler in another window an look for `gla/jobs/update_all_products/create_batch`. Scheduled for 30 minutes. Run it manually. 
4. Go back to the other window in Attribute mapping and see how after a few seconds it updates back to "Your products are synced with your rules"


js/src/hooks/usePolling.js
```diff
@@ -0,0 +1,44 @@
...
+		const promise = fetch();
+		promise.finally( () => startCountdown( delay ) );
+		return promise;
+	}, [ fetch, startCountdown, delay ] );
+
+	useEffect( () => {
+		if ( second === 0 && callCount > 0 && ( ! data || ! stopOnResolved ) ) {
+			start();
+		}
+	}, [ second, callCount, start, stopOnResolved, data ] );
```
eason9487: As `data` is also in the deps array of this useEffect, if an API call responds within one second and changes the reference of `data`, it will invoke `start()` again which results in a duplicate API call.

https://user-images.githubusercontent.com/17420811/196396125-9c5a02f2-272a-4937-a78f-ebb8957eec07.mp4

js/src/hooks/usePolling.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {number} [delay=20] Time in seconds to wait for each polling request
+ * @param {boolean} [stopOnResolved=false] If true. It stops the polling when the data is received
+ * @return {Object} Payload containing the result of the API call as well as a trigger function for starting the polling.
+ */
+export default function usePolling(
+	options,
+	delay = 5,
+	stopOnResolved = false
+) {
+	const { second, callCount, startCountdown } = useCountdown();
```
eason9487, line 25: After leaving the Attribute Mapping page, it will cause an error `Cannot read properties of undefined (reading 'id')` in `useCountdown` and lead to a blank page. Considering `useCountdown` was changed to have the default `handle` value in the `develop` branch, we could cherry-pick f3cc0437ae9dd124aaac702c3b14099490e121ec into this PR or merge the `develop` branch to the shared branch of this project and rebase related PRs. So that the use case here could keep no parameters.

eason9487 (commented):
Found two issues that may need adjustments. Could you point me to the Figma document or mockup that this implementation is referred to? I couldn't find the match one from both design i1 and i2 documents on Figma.

puntope:
Howdy @eason9487 

Thanks for cactus hing both bugs! 

I fixed them and this is ready for another round

js/src/hooks/usePolling.js
```diff
@@ -0,0 +1,48 @@
...
+			return promise;
+		},
+		[ fetch, startCountdown, delay ]
+	);
+
+	useEffect( () => {
+		if ( second === 0 && callCount > 0 ) {
+			start( ! data || ! stopOnResolved );
+		}
+	}, [ second, callCount, start, stopOnResolved ] );
```
eason9487: There is an eslint warning `react-hooks/exhaustive-deps`. Unfortunately, the [issue of duplicate API calls](https://github.com/woocommerce/google-listings-and-ads/pull/1733#discussion_r997974904) still appears if adding `data` to useEffect deps array.

image

With the revision [9003b36](https://github.com/woocommerce/google-listings-and-ads/blob/9003b36ca713b699db55f8eca7948942aaa095ab/js/src/hooks/usePolling.js#L34-L38), replacing `stopOnResolved` and `data` in deps with a combined variable such as `const shouldRun = ! data || ! stopOnResolved;`, which is declared out of the useEffect scope, could be a solution.

eason9487 (commented):
Thanks for updating this PR with `develop` branch. After some tests, I think the issue of duplicate API calls might not yet be solved. Furthermore, quoted from https://github.com/woocommerce/google-listings-and-ads/pull/1733#pullrequestreview-1145451070:
> Could you point me to the Figma document or mockup that this implementation is referred to? I couldn't find the match one from both design i1 and i2 documents on Figma.

I'm still not 100% sure which design document is the target one to be referred to. In fqR0EHi63lWahRcVTKCcba-fi-2305%3A162769, the most likely design documents I can find are the following. However, they are quite different from the actual implementation in this PR.

- Statuses
   image
- Tooltip
   image

puntope:
Howdy @eason9487 

I updated the PR again with your awesome suggestions. 

Thanks for the tip in regards the useEffect problem

I updated the logic adapting the table to the new UX. Now it looks accurate (notice pagination in the center)

ℹ️  Since backend has changed after feedback, for getting the last sync date you will need to add this option in the Options Table `gla_update_all_products_last_sync`. You can add a timestamp like `1667338631` as value. Or just update an attribute rule and run manually the Job in Action Scheduler (or wait 30 mins) otherwise the date will be `never`. Notice also when there is an action scheduled or running, we don't show  the date but instead "Scheduled for sync"

eason9487 (approved):
Thanks for the updates. LGTM. There are a few styling differences that could be further adjusted but they are not blockers.

---

The ways of line break and alignment in the table footer are a little different from the design on Figma.
image
image

Figma:
image

puntope:
> The ways of line break and alignment in the table footer are a little different from the design on Figma

Thanks @eason9487 . This is noted for future tweaks

Issue #1548, opened 2022-06-02
UI AppShell containing the whole plugin


ecgan:
## User story

This is a proposal and a follow-up to https://github.com/woocommerce/google-listings-and-ads/pull/1543#pullrequestreview-992499432:

> For a long term better solution, I think we should have an AppShell that contains the whole plugin, and things like the `CustomerEffortScorePrompt`, `DifferentCurrencyNotice` and `NavigationClassic` components should be in the AppShell, so that there will only be one instance of them throughout the whole app. With that approach, we wouldn't need to have this `useEffect` call. This approach would also solve the issue where the notice appears to keep re-appearing when we switch tabs, see demo video of the issue below:

Here are a few problems that we have in our existing code without AppShell:

- Problem with the `CustomerEffortScorePrompt`, as shown in https://github.com/woocommerce/google-listings-and-ads/pull/1543#pullrequestreview-992499432.
- We have multiple instances of the tab navigation component in multiple components. Because it is not a single tab component, we lose the fancy tab switch animation.
- We have multiple instances of the `DifferentCurrencyNotice` in multiple components.

### Describe the solution you'd like 

Have an AppShell that contains the whole plugin, so that the above mentioned components can be placed in the AppShell, thus addressing the problems.

tomalec:
:bulb: (probably side-tracking) Having the AppShell reminds me of the concept of [app shell from PWAs](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/App_structure#app_shell). I think it would be good to have some (preferably service worker) infrastructure from WP-admin/WC to the app shell. A one that's heavily cached in user's browser, and available offline.

eason9487:
> Here are a few problems that we have in our existing code without AppShell:
> 
> * [...]
> * We have multiple instances of the tab navigation component in multiple components. Because it is not a single tab component, we lose the fancy tab switch animation.
> * [...]

I tried tweaking `NavigationClassic`, which is the most used on this extension's page. In the ways I've experimented, moving it into `withAdminPageShell` requires an additional layer of abstraction or a certain level of indirect conditions to control whether or not to display it.

In comparison, I think the intuitive use of a `<NavigationClassic />` on the needed pages is preferable to the added complexity of these tweaks.

---

I'm going to close this issue since the conceptual equivalent of AppShell has been added in #1739.
[...]

PR #1739, opened 2022-10-26
WP 6.1 Compatibility: Fix the giant icons in the `DatePicker` component

### LINKED ISSUE #1736: WP 5.9 Compatibility: The style of `VisuallyHidden` component is not working
### Describe the bug:

When using `ExternalLink` component, the originally invisible text is shown since WP 5.9.

### Steps to reproduce:

1. Install a WP version >= 5.9
2. Set up paid ads or go to GLA Settings page to view the text "(opens in a new tab)" on the card UI of the connected Google Ads account.

### Expected behavior:

The text "(opens in a new tab)" should be invisible.

### Actual behavior:

The text "(opens in a new tab)" is visible.

### Additional details:

- [WP 5.8.6](https://github.com/WordPress/wordpress-develop/blob/5.8.6/package.json#L90) use `@wordpress/components` [14.1.11](https://github.com/WordPress/gutenberg/tree/%40wordpress/components%4014.1.11/packages/components/src/visually-hidden).
- [WP 5.9.0](https://github.com/WordPress/wordpress-develop/blob/5.9.0/package.json#L90) use `@wordpress/components` [19.2.2](https://github.com/WordPress/gutenberg/tree/%40wordpress/components%4019.2.2/packages/components/src/visually-hidden).
- The style of `VisuallyHidden` component was changed to apply `style` prop directly.

eason9487:

Closes #1736 

- ~Fix that the style of `VisuallyHidden` component is not working since WP 5.9.~
- Fix the giant icons in the `DatePicker` component
   image

This PR also implements the base of the app shell (#1548):

- ~Add a new value of `layoutName` to `useLayout` hook for isolating GLA's style fixes.~
- Wrap all top-level admin pages with an app shell and CSS class `gla-admin-page`.
- Add CSS class `gla-admin-page` to the root of `Modal` as well.

### Screenshots:

image

~1. Install a WP version >= 5.9~
~2. Set up paid ads or go to GLA Settings page to view the external link on the card UI of the connected Google Ads account.~
~3. The text "(opens in a new tab)" within the external link should be invisible.~

1. Install WP 6.1.
2. Go to the Dashboard page.
3. Click the date range option to open datepicker. Switch to custom tab to verify if the icon size of navigation buttons is appropriate.

### Changelog entry

> Fix - WordPress 6.1 Compatibility: The size of navigation icons in Datepicker component should not be a giant size.

### DIFF
--- a/js/src/components/app-modal/index.js
+++ b/js/src/components/app-modal/index.js
@@ -42,6 +42,9 @@ const AppModal = ( {
 	...rest
 } ) => {
 	const modalClassName = classnames(
+		// gla-admin-page is for scoping particular styles to components that are used by
+		// a GLA admin page and are rendered via ReactDOM.createPortal.
+		'gla-admin-page',
 		'app-modal',
 		overflowStyleName[ overflow ],
 		className
--- a/js/src/components/withAdminPageShell.js
+++ b/js/src/components/withAdminPageShell.js
@@ -1,0 +1,25 @@
+/**
+ * External dependencies
+ */
+import { createHigherOrderComponent } from '@wordpress/compose';
+
+/**
+ * Internal dependencies
+ */
+import useLayout from '.~/hooks/useLayout';
+
+/**
+ * A higher-order component for wrapping the app shell on top of the GLA admin page.
+ * Cross-page shared things could be handled here.
+ *
+ * @param {JSX.Element} Page Top-level admin page component to be wrapped by app shell.
+ */
+const withAdminPageShell = createHigherOrderComponent(
+	( Page ) => ( props ) => {
+		useLayout( 'admin-page' );
+		return <Page { ...props } />;
+	},
+	'withAdminPageShell'
+);
+
+export default withAdminPageShell;
--- a/js/src/components/withAdminPageShell.js
+++ b/js/src/components/withAdminPageShell.js
@@ -4,11 +4,6 @@
 import { createHigherOrderComponent } from '@wordpress/compose';
 
 /**
- * Internal dependencies
- */
-import useLayout from '.~/hooks/useLayout';
-
-/**
  * A higher-order component for wrapping the app shell on top of the GLA admin page.
  * Cross-page shared things could be handled here.
  *
@@ -16,8 +11,12 @@ import useLayout from '.~/hooks/useLayout';
  */
 const withAdminPageShell = createHigherOrderComponent(
 	( Page ) => ( props ) => {
-		useLayout( 'admin-page' );
-		return <Page { ...props } />;
+		return (
+			// gla-admin-page is for scoping particular styles to a GLA admin page.
+			<div className="gla-admin-page">
+				<Page { ...props } />;
+			</div>
+		);
 	},
 	'withAdminPageShell'
 );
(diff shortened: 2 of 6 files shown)


tomalec (approved):
Tested locally, inspected the code, LGTM

tomalec:
I really like the introduction of the app shell concept.

:grey_question: 
However, I think it would be nice to specify what we consider the "app shell", as I'd expect it to be an actual element enclosing the app's HTML subtree, or something narrowing the scope of elements to the ones created by our extension.

Whereas if you visit `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings` and do `document.querySelectorAll('.gla-admin-page button')` you will get buttons from WP side menu, `#activity-panel-tab-activity` from WooCommerce chrome, etc. which are beyond our control.

Is this really what we wanted? Why don't we just wrap every GLA's page with `<gla-page>`/`<div class="gla-page">` or attach this class to the enclosing container/direct `.parentElement`?

tomalec:
Made an alternative fix in https://github.com/woocommerce/google-listings-and-ads/pull/1762

eason9487:
@tomalec, thanks for the review.

> However, I think it would be nice to specify what we consider the "app shell", as I'd expect it to be an actual element enclosing the app's HTML subtree, or something narrowing the scope of elements to the ones created by our extension.

> Is this really what we wanted? Why don't we just wrap every GLA's page with `<gla-page>`/`<div class="gla-page">` or attach this class to the enclosing container/direct `.parentElement`?

Since some components are created through ReactDOM.createPortal, they are usually mounted directly under the `<body>`, but do not provide identifiers for precise scoping, such as `Modal`, so I chose to add `.gla-full-page` to `<body>` to ensure those components would also apply CSS fixes. But as you mentioned, it's indeed not ideal.

I searched the GLA codebase and the only components that use `ReactDOM.createPortal` are `Modal` and `Snackbar` (used through `createNotice`). The former can be scoping CSS class in `AppModal`, the latter probably doesn't need to be adjusted at the moment, so I made some adjustments to scope GLA admin pages more accurately.

- https://github.com/woocommerce/google-listings-and-ads/pull/1739/commits/9decd2908953aa6a11a5f7315237987e49690f4e : Move CSS class `gla-admin-page` to nodes that are closer to the root of GLA pages in the DOM tree.
- https://github.com/woocommerce/google-listings-and-ads/pull/1739/commits/52ef95f5a0a237ee41d20a061e019476779324c2 : Revert the change in the `useLayout` hook.

eason9487:
@tomalec, thanks for proposing [a nicer solution](https://github.com/woocommerce/google-listings-and-ads/pull/1739#issuecomment-1310850630).

I adjusted this PR as per https://github.com/woocommerce/google-listings-and-ads/pull/1762#issuecomment-1311437832 by https://github.com/woocommerce/google-listings-and-ads/pull/1739/commits/3293fb01d198cd69bb6ed03454a0707891dba963 and https://github.com/woocommerce/google-listings-and-ads/pull/1739/commits/aecd97f4124bc626e8f03eea26aa539a6df82613. PR description was also updated. Could you help with a new round of code reviews?

tomalec (approved):
Thanks for the research and explanations. Regarding `createNotice` I think it's better we try to avoid tempering its styles to sustain consistent experience for merchants, so leaving it without `gla-admin-page` scope seems fine to me.

Tested locally, reviewed the code LGTM

PR #1750, opened 2022-11-07
WC 6.9 compatibility: Fix the incorrect form states of shipping settings

### LINKED ISSUE #1747: WC 6.9 Compatibility: Shipping time settings should not appear after selecting the "complex" shipping option
### Describe the bug:

If the "My shipping settings are complex. I will enter my shipping rates and times manually in Google Merchant Center" option is selected, the Shipping time should be hidden but appears.

### Steps to reproduce:

1. Use a WooCommerce version >= 6.9.0
2. Go to the Edit free listings page
3. Select the shipping option "My shipping settings are complex. I will enter my shipping rates and times manually in Google Merchant Center."

### Expected behavior:

image

### Actual behavior:

image

### Additional details:

Maybe #1738 is also related to this.

### LINKED ISSUE #1753: WC 6.9 - Offer Free Shipping Setting is not updated
### Describe the bug:

If all shipping rates are 0, the state `values.offer_free_shipping` should be updated to `undefined` so the free shipping option is not required allowing the user to continue with the onboarding or saving the data. 

https://github.com/woocommerce/google-listings-and-ads/blob/e464c76de235de9fd1bcb22d1917d66e48d396e8/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js#L25-L29

### Steps to reproduce:

1.  Go to Edit Free listings 
2. Add shipping rates > 0 
3. Offer Free Shipping...: 'Yes'
4. Minimum Order Quantity...: any value bigger than 0
5. Edit the shipping rates to 0
6. "Save Changes" button is disabled (should be enabled)

### Expected behavior:

If the shipping rates are 0 it should not require a free shipping threshold. 

https://github.com/woocommerce/google-listings-and-ads/blob/e464c76de235de9fd1bcb22d1917d66e48d396e8/js/src/components/free-listings/configure-product-listings/checkErrors.js#L73

https://user-images.githubusercontent.com/2488994/200633714-7519b445-cb27-4613-abbd-35ae45c528ec.mp4

### Actual behavior:

https://user-images.githubusercontent.com/2488994/200632912-4687367c-828a-4084-acfc-b859c4b74576.mp4

### Additional details:

The issue happens with WC >= 6.9

eason9487:

Closes #1747

- Fix the WC 6.9 compatibility problem that the `shipping_time` of the form state will be incorrect when the selection of shipping rate is 'complex'.

1. Use a WooCommerce version >= 6.9.0
2. Go to the Edit free listings page
3. Change the shipping options:
   - When the recommended or simple shipping rate is selected, shipping time settings should be displayed.
   - When the complex shipping rate is selected, shipping time settings should be hidden.
4. Save changes and then inspect the `gla/mc/settings` API request to see if the saved data are correct.
   https://github.com/woocommerce/google-listings-and-ads/blob/e464c76de235de9fd1bcb22d1917d66e48d396e8/js/src/components/free-listings/setup-free-listings/index.js#L126-L127

### Additional details:

- This fix is **NOT** compatible with WC < 6.9 but I think it should be fine since we are going to bump the minimum support WC version to 6.9.
- There is a warning as the following as it did request state updates while rendering. Given the continued use of WC `Form` component and the need to satisfy business logic, this warning might not be avoided, even though the final form state will still be synchronized.
   image

### Changelog entry

> Fix - WC 6.9 compatibility: Shipping time settings should not appear after selecting the "complex" shipping option.
> Fix - WC 6.9 compatibility: The free shipping threshold should be cleared after selecting the "No" free shipping option.
> Fix - WC 6.9 compatibility: The selected free shipping option should be reset after setting all shipping rates to 0.

### DIFF
--- a/js/src/components/free-listings/setup-free-listings/index.js
+++ b/js/src/components/free-listings/setup-free-listings/index.js
@@ -123,6 +123,24 @@ const SetupFreeListings = ( {
 					formProps.setValue( 'offer_free_shipping', undefined );
 				} );
 			}
+		} else if ( change.name === 'offer_free_shipping' ) {
+			// After selecting the 'No' option of the free shipping threshold,
+			// Reset all shipping_country_rates.options.free_shipping_threshold.
+			if ( ! change.value ) {
+				const nextValue = values.shipping_country_rates.map(
+					( rate ) => ( {
+						...rate,
+						options: {
+							...rate.options,
+							free_shipping_threshold: undefined,
+						},
+					} )
+				);
+
+				formPropsDelegateeRef.current.push( ( formProps ) => {
+					formProps.setValue( 'shipping_country_rates', nextValue );
+				} );
+			}
 		} else if ( change.name === 'shipping_country_times' ) {
 			onShippingTimesChange( values.shipping_country_times );
 		} else if ( settingsFieldNames.includes( change.name ) ) {
--- a/js/src/components/free-listings/setup-free-listings/index.js
+++ b/js/src/components/free-listings/setup-free-listings/index.js
@@ -115,7 +115,29 @@ const SetupFreeListings = ( {
 		} else if ( change.name === 'shipping_country_times' ) {
 			onShippingTimesChange( values.shipping_country_times );
 		} else if ( settingsFieldNames.includes( change.name ) ) {
-			onSettingsChange( getSettings( values ) );
+			// The value of `shipping_time` option is determined by the value of `shipping_rate` option.
+			// So if the current form change is considered it needs to change `shipping_time` as well,
+			// it schedules the processing with `formPropsDelegateeRef` and also skips the call of
+			// `onSettingsChange` this time, and lets the call of `onSettingsChange` be triggered
+			// when the form change of `shipping_time` happens.
+			let shouldTriggerOnChange = true;
+
+			if ( change.name === 'shipping_rate' ) {
+				// When shipping rate is 'manual', shipping time should be 'manual' as well;
+				// When shipping rate is 'automatic' or 'flat', shipping time should be 'flat'.
+				const nextValue = change.value === 'manual' ? 'manual' : 'flat';
+
+				if ( nextValue !== values.shipping_time ) {
+					shouldTriggerOnChange = false;
+					formPropsDelegateeRef.current.push( ( formProps ) =>
+						formProps.setValue( 'shipping_time', nextValue )
+					);
+				}
+			}
+
+			if ( shouldTriggerOnChange ) {
+				onSettingsChange( getSettings( values ) );
+			}
 		} else if ( targetAudienceFields.includes( change.name ) ) {
 			onTargetAudienceChange( pick( values, targetAudienceFields ) );
 
--- a/js/src/components/free-listings/setup-free-listings/index.js
+++ b/js/src/components/free-listings/setup-free-listings/index.js
@@ -12,6 +12,7 @@ import AppSpinner from '.~/components/app-spinner';
 import Hero from '.~/components/free-listings/configure-product-listings/hero';
 import checkErrors from '.~/components/free-listings/configure-product-listings/checkErrors';
 import getOfferFreeShippingInitialValue from '.~/utils/getOfferFreeShippingInitialValue';
+import isNonFreeShippingRate from '.~/utils/isNonFreeShippingRate';
 import FormContent from './form-content';
 
 /**
@@ -112,6 +113,16 @@ const SetupFreeListings = ( {
 	const handleChange = ( change, values ) => {
 		if ( change.name === 'shipping_country_rates' ) {
 			onShippingRatesChange( values.shipping_country_rates );
+
+			// If all the shipping rates are free shipping,
+			// we set the offer_free_shipping to undefined,
+			// so that when users add a non-free shipping rate,
+			// they would need to choose "yes" / "no" for offer_free_shipping.
+			if ( ! change.value.some( isNonFreeShippingRate ) ) {
+				formPropsDelegateeRef.current.push( ( formProps ) => {
+					formProps.setValue( 'offer_free_shipping', undefined );
+				} );
+			}
 		} else if ( change.name === 'shipping_country_times' ) {
 			onShippingTimesChange( values.shipping_country_times );
 		} else if ( settingsFieldNames.includes( change.name ) ) {
--- a/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js
+++ b/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js
@@ -8,34 +8,10 @@ import MinimumOrderCard from './minimum-order-card';
 
 const FlatShippingRatesInputCards = ( props ) => {
 	const { audienceCountries, formProps } = props;
-	const { getInputProps, values, setValue } = formProps;
+	const { getInputProps, values } = formProps;
 	const displayFreeShippingCards = values.shipping_country_rates.some(
 		isNonFreeShippingRate
 	);
-
-	const getOfferFreeShippingChangeHandler = ( onChange ) => (
-		newOfferFreeShippingValue
-	) => {
-		/**
-		 * When users select 'No', we remove all the
-		 * shippingRate.options.free_shipping_threshold.
-		 */
-		if ( ! newOfferFreeShippingValue ) {
-			const newValue = values.shipping_country_rates.map( ( el ) => {
-				return {
-					...el,
-					options: {
-						...el.options,
-						free_shipping_threshold: undefined,
-					},
-				};
-			} );
-
-			setValue( 'shipping_country_rates', newValue );
-		}
-
-		onChange( newOfferFreeShippingValue );
-	};
 
 	return (
 		<>
@@ -47,9 +23,6 @@ const FlatShippingRatesInputCards = ( props ) => {
 				<>
 					<OfferFreeShippingCard
 						{ ...getInputProps( 'offer_free_shipping' ) }
-						onChange={ getOfferFreeShippingChangeHandler(
-							getInputProps( 'offer_free_shipping' ).onChange
-						) }
 					/>
 					{ values.offer_free_shipping && (
 						<MinimumOrderCard
--- a/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js
+++ b/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js
@@ -13,22 +13,6 @@ const FlatShippingRatesInputCards = ( props ) => {
 		isNonFreeShippingRate
 	);
 
-	const getShippingRatesChangeHandler = ( onChange ) => (
-		newShippingRates
-	) => {
-		/**
-		 * If all the shipping rates are free shipping,
-		 * we set the offer_free_shipping to undefined,
-		 * so that when users add a non-free shipping rate,
-		 * they would need to choose "yes" / "no" for offer_free_shipping.
-		 */
-		if ( ! newShippingRates.some( isNonFreeShippingRate ) ) {
-			setValue( 'offer_free_shipping', undefined );
-		}
-
-		onChange( newShippingRates );
-	};
-
 	const getOfferFreeShippingChangeHandler = ( onChange ) => (
 		newOfferFreeShippingValue
 	) => {
@@ -58,9 +42,6 @@ const FlatShippingRatesInputCards = ( props ) => {
 			<EstimatedShippingRatesCard
 				audienceCountries={ audienceCountries }
 				{ ...getInputProps( 'shipping_country_rates' ) }
-				onChange={ getShippingRatesChangeHandler(
-					getInputProps( 'shipping_country_rates' ).onChange
-				) }
 			/>
 			{ displayFreeShippingCards && (
 				<>
--- a/js/src/components/shipping-rate-section/shipping-rate-section.js
+++ b/js/src/components/shipping-rate-section/shipping-rate-section.js
@@ -20,23 +20,8 @@ import FlatShippingRatesInputCards from './flat-shipping-rates-input-cards';
  */
 
 const ShippingRateSection = ( { formProps, audienceCountries } ) => {
-	const { getInputProps, values, setValue } = formProps;
+	const { getInputProps, values } = formProps;
 	const inputProps = getInputProps( 'shipping_rate' );
-
-	const getShippingRateOptionChangeHandler = ( onChange ) => ( value ) => {
-		switch ( value ) {
-			case 'automatic':
-			case 'flat':
-				setValue( 'shipping_time', 'flat' );
-				break;
-
-			case 'manual':
-				setValue( 'shipping_time', 'manual' );
-				break;
-		}
-
-		onChange( value );
-	};
 
 	return (
 		<Section
@@ -78,9 +63,6 @@ const ShippingRateSection = ( { formProps, audienceCountries } ) => {
 								) }
 								value="automatic"
 								collapsible
-								onChange={ getShippingRateOptionChangeHandler(
-									inputProps.onChange
-								) }
 							>
 								<RadioHelperText>
 									{ __(
@@ -97,9 +79,6 @@ const ShippingRateSection = ( { formProps, audienceCountries } ) => {
 								) }
 								value="flat"
 								collapsible
-								onChange={ getShippingRateOptionChangeHandler(
-									inputProps.onChange
-								) }
 							/>
 							<AppRadioContentControl
 								{ ...inputProps }
@@ -109,9 +88,6 @@ const ShippingRateSection = ( { formProps, audienceCountries } ) => {
 								) }
 								value="manual"
 								collapsible
-								onChange={ getShippingRateOptionChangeHandler(
-									inputProps.onChange
-								) }
 							>
 								<RadioHelperText>
 									{ createInterpolateElement(


tomalec:
Tested locally, the UI behaves as expected, however, `onSettingsChange` callback is called 15 times after a single change. That looks like a big overhead. Is this expected, or could we avoid it?

js/src/components/free-listings/setup-free-listings/index.js
```diff
@@ -115,7 +115,29 @@ const SetupFreeListings = ( {
...
+			// when the form change of `shipping_time` happens.
+			let shouldTriggerOnChange = true;
+
+			if ( change.name === 'shipping_rate' ) {
+				// When shipping rate is 'manual', shipping time should be 'manual' as well;
+				// When shipping rate is 'automatic' or 'flat', shipping time should be 'flat'.
+				const nextValue = change.value === 'manual' ? 'manual' : 'flat';
+
+				if ( nextValue !== values.shipping_time ) {
+					shouldTriggerOnChange = false;
```
tomalec, line 160: :question: 
I found it counterintuitive, that if the next value is different we do not trigger `onChange`, but where it is the same (nothing changed) we do trigger it.

tomalec: Sorry, it's explained above

>             // it schedules the processing with `formPropsDelegateeRef` and also skips the call of
> 			// `onSettingsChange` this time, and lets the call of `onSettingsChange` be triggered
> 			// when the form change of `shipping_time` happens.

jorgemd24 (approved):
Thanks, @eason9487, I have tested the code and it is fixing the issue with the `shipping_time` state 👍  

However I have some concerns about the root of this problem, when I was debugging this PR I saw that we have a similar issue with the `values.offer_free_shipping` state, for example: 

- when the rate is 0 and  `values.offer_free_shipping` was previously set to "Y"

The state is not updated correctly, `values.offer_free_shipping` should be updated to 'undefined'.  See video:

https://user-images.githubusercontent.com/2488994/200619413-8c234011-f119-4c95-a979-f16918886854.mp4

From what I can see it seems that there is an issue when `setValue` and `getInputProps( 'name' ).onChange` are combined and the state is somehow overridden. 

https://github.com/woocommerce/google-listings-and-ads/blob/e464c76de235de9fd1bcb22d1917d66e48d396e8/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js#L25-L29

A similar case in the `ShippingRateSection` prior this PR:

https://github.com/woocommerce/google-listings-and-ads/pull/1750/files#diff-4594bb5f450b25cad952c53f192d5aaa3e33cb73798f578cfa85695c6f4165e2L26-L40

What do you think?

tomalec:
Give, we have conceptually the components as follows:
```
SetupFreeListings:
   <Form>
      <ShippingRatesType>
      <FlatShippingRate>
      <FlatShippingTime>
  </Form>
```

The way I understood the desired data flow here looks like this:
##### conceptually
- user **changes** shipping **rate type** to a `newValue`
   - SetupFreeListings updates its (direct or indirect) state to have `shipping_rate: newValue`
   - SetupFreeListings updates its state to have `shipping_time: 'manual'` if `newValue == 'manual'`, otherwise `flat`
   - SetupFreeListings (directly, or indirectly via FormContent) shows `FlatShippingRate` only if `newValue == 'flat'`
   - SetupFreeListings (directly, or indirectly via FormContent) shows `FlatShippingTime` only if `newValue == 'flat' | 'automatic'`
   - SetupFreeListings triggers its onSettingsChanged callback with an updated state
   
However, the way I read the implementation it's more like this:

##### implementation

- user **changes** shipping **rate type** to a `newValue`
   1. SetupFreeListings > FormContent updates its state to have `shipping_rate: newValue`
   2. SetupFreeListings > FormContent shows `FlatShippingRate` only if `newValue == 'flat'`
   3. if the change does not result in changing the `shippint_time`, `SetupFreeListings` trigger `onSettingChanged`
   4. if the change results in changing `shipping_time`, do not trigger callback, but `SetupFreeListings` updates the state of `FormContent`
   5. SetupFreeListings > FormContent shows `FlatShippingTime` only if `newValue == 'flat' | 'automatic'`
   6. `Form`  notifies `SetupFreeListings` on the change it just made.
   7. `SetupFreeListings` triggers the callback.
   

Steps 3-6 seem to me like unnecessary ping pong. Can we make it simpler by putting state management into a single component?
Is that due to some limitations in the external components we use? If so maybe it's worth rising an issue upstream and linking it in the code, to hopefully, eventually improve that.
I'm afraid, especially considering [the comment above](https://github.com/woocommerce/google-listings-and-ads/pull/1750#pullrequestreview-1172507359) that if we keep on patching it may eventually hit us with more bugs.

tomalec (approved):
Tested locally, reviewed the code, LGTM

eason9487:
Thanks for the reviews, @tomalec @jorgemd24 

---

@tomalec 

> Tested locally, the UI behaves as expected, however, `onSettingsChange` callback is called 15 times after a single change. That looks like a big overhead. Is this expected, or could we avoid it?

I was unable to reproduce this. Could you provide the replication steps?

And I will find another time to circle back this https://github.com/woocommerce/google-listings-and-ads/pull/1750#issuecomment-1307967729 🙏 

---

@jorgemd24 

> However I have some concerns about the root of this problem, when I was debugging this PR I saw that we have a similar issue with the `values.offer_free_shipping` state

Nice finding! I will push a new commit to this PR for fixing that one together.

eason9487:
### Additional fix:

Issue #1753 should be fixed by https://github.com/woocommerce/google-listings-and-ads/pull/1750/commits/415bb91835903cd0bb99f966e5938535a886d4b5.

### Screenshots:
https://user-images.githubusercontent.com/17420811/200836112-65ea68d1-2d9b-41a7-8f85-23a4ebe5ebc2.mp4

Please refer to the replication steps in #1753

eason9487:
### One more additional fix:

[This processing](https://github.com/woocommerce/google-listings-and-ads/blob/e464c76de235de9fd1bcb22d1917d66e48d396e8/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js#L39-L51) has the same issue as well. It should be fixed by https://github.com/woocommerce/google-listings-and-ads/pull/1750/commits/2daefb2a9f160a1030729bfe026232f55ae53606.

### Actual behavior:

The value of Minimum Order Quantity...:  is not cleared after selecting the "No" option for free shipping.

https://user-images.githubusercontent.com/17420811/200836158-e235028e-54f7-4aa8-aa5f-96e9b44c2a18.mp4

### Screenshots (Expected behavior):

https://user-images.githubusercontent.com/17420811/200836142-93923659-a209-4c6b-adfb-d8303f5cee65.mp4

1. Go to Edit Free listings
1. Add shipping rates > 0
1. Offer Free Shipping...: 'Yes'
1. Minimum Order Quantity...: any value bigger than 0
1. Offer Free Shipping...: 'No'
1. Offer Free Shipping...: 'Yes' again
1. The value of Minimum Order Quantity...:  entered in step 4 should be cleared

eason9487:
Hi @tomalec, as we discussed during the dev call. To make all relevant issues of WC `Form` be released sooner, I pushed related commits to this PR. Could you help with the additional round of code reviews?

tomalec:
:construction: After [the last commit](https://github.com/woocommerce/google-listings-and-ads/pull/1750/commits/2daefb2a9f160a1030729bfe026232f55ae53606)
once I change anything on the settings page, it gets blank, and the console throws:
```
Too many re-renders. React limits the number of renders to prevent an infinite loop
```
[Screencast from 09.11.2022 21:21:38.webm](https://user-images.githubusercontent.com/17435/200934187-846ce584-fa41-437d-a845-057f5235e7a4.webm)

tomalec:
I don't see why https://github.com/woocommerce/google-listings-and-ads/pull/1750/commits/2daefb2a9f160a1030729bfe026232f55ae53606 is different than previous ones. Seems we reached the limit of workarounds for colliding versions of React components

tomalec:
>> Tested locally, the UI behaves as expected, however, onSettingsChange callback is called 15 times after a single change. That looks like a big overhead. Is this expected, or could we avoid it?
>
> I was unable to reproduce this. Could you provide the replication steps?

Here is how I logged it 
https://github.com/woocommerce/google-listings-and-ads/compare/fix/1747-shipping-time-form-state...test/1750-multiple-onSettingsChange-calls?expand=1

Repro steps:
1. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Ffree-listings%2Fedit`
2. Change shipping rate from automatic to manual

eason9487:
@tomalec, thanks for the review!

> once I change anything on the settings page, it gets blank [...]

I guess you were testing with WC 7.1? I have to say I was very lucky because I only tested it with WC 6.9 and 7.0. However, I was just aware of the `Form` in WC 7.1 may have new bugs or breaking changes that caused new incompatibilities.

I couldn't run the code to confirm it right now but the codes below look like lead to trigger `onChange` for each item in the `values` regardless whether a value change is did happened, even if we only call `setValue` for updating a single value.

- https://github.com/woocommerce/woocommerce/blob/7.1.0/packages/js/components/src/form/form.tsx#L209-L211
- https://github.com/woocommerce/woocommerce/blob/7.1.0/packages/js/components/src/form/form.tsx#L182-L197

Probably it needs more hacks since WC 7.1 has been released. 😂

tomalec:
> I guess you were testing with WC 7.1?

Yes, exactly.


>  I couldn't run the code to confirm it right now but the codes below look like lead to trigger onChange for each item in the values regardless whether a value change is did happened, even if we only call setValue for updating a single value.

Confirmed, that's the case. :facepalm: 

> Probably it needs more hacks since WC 7.1 has been released. joy

Our code already consists of mostly workarounds for WC & WP quirks, maybe it's time to finally clean the field.

eason9487:
@tomalec, thanks for confirming the root cause

> 🚧 After [the last commit](https://github.com/woocommerce/google-listings-and-ads/pull/1750/commits/2daefb2a9f160a1030729bfe026232f55ae53606) once I change anything on the settings page, it gets blank, and the console throws:
> 
> ```
> Too many re-renders. React limits the number of renders to prevent an infinite loop
> ```

Since this issue belongs to the WC 7.1 compatibility issue, I was addressing it in a separate PR #1759. If #1759 is feasible and gets approved, I think this PR could be merged together in order by then.

tomalec (approved):
Reviewed and tested together with https://github.com/woocommerce/google-listings-and-ads/pull/1759/files# LGTM

eason9487:
Hi @tomalec, thanks for your patience in waiting for me to get back to the question raised almost two months ago.

Quoted from myself:

> And I will find another time to circle back this [#1750 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/1750#issuecomment-1307967729) 🙏

---

## Topic

Quoted from https://github.com/woocommerce/google-listings-and-ads/pull/1750#issuecomment-1307967729:

> However, the way I read the implementation it's more like this:
> 
> ##### implementation
> * user **changes** shipping **rate type** to a `newValue`
>   
>   1. SetupFreeListings > FormContent updates its state to have `shipping_rate: newValue`
>   2. SetupFreeListings > FormContent shows `FlatShippingRate` only if `newValue == 'flat'`
>   3. if the change does not result in changing the `shippint_time`, `SetupFreeListings` trigger `onSettingChanged`
>   4. if the change results in changing `shipping_time`, do not trigger callback, but `SetupFreeListings` updates the state of `FormContent`
>   5. SetupFreeListings > FormContent shows `FlatShippingTime` only if `newValue == 'flat' | 'automatic'`
>   6. `Form`  notifies `SetupFreeListings` on the change it just made.
>   7. `SetupFreeListings` triggers the callback.
> 
> Steps 3-6 seem to me like unnecessary ping pong. Can we make it simpler by putting state management into a single component? Is that due to some limitations in the external components we use? If so maybe it's worth rising an issue upstream and linking it in the code, to hopefully, eventually improve that. I'm afraid, especially considering [the comment above](https://github.com/woocommerce/google-listings-and-ads/pull/1750#pullrequestreview-1172507359) that if we keep on patching it may eventually hit us with more bugs.

Thanks for bringing up this question. For me, this question has always been considered a code smell, and it has been buried in my mind for many months.

## Possible causes

I think the primary cause for this result is the strategy for managing data and states, and the secondary is the limitations imposed by the WC `Form` component.

- Management strategy of data and states: In GLA frontend codebase, data and state are usually not managed in a hierarchical way, which once made the logic of a field highly coupled, ranging from the backend API, to the frontend data retrieval, wp-data store, and then all the way through the form state to the form's input components.
- `Form` component: The functionality provided for operating `values` is very limited. Before WC 7.1, there is only a `setValue` function and only a flat field can be updated at a time.

## Context

Before going further, I would like to mention some context of this case based on the current implementation.

- Currently, the `'automatic'` application of shipping time settings is not supported in GLA, so there is a special logic that needs to be handled additionally.
- In the API communication layer, the `shipping_time` is data.
- In the form layer, `shipping_time` could be only a state rather than data.
- In the form section layer, `shipping_time` could be more like a prop for determining the special logic rather than data.
- From the technical design point of view, should the changes of multiple data fields in the custom `on*Change` event be triggered once combined or multiple times individually?

## Possible alternatives

Next, there are several alternatives.

1. Since WC 7.1, the `Form` has added `setValues` to support updating multiple fields at once.
2. Let it trigger `onSettingsChange` two times directly, and debounce the callbacks of `onSettingsChange` in its parent components or the calls of `saveSettings`.
3. Adopt a strategy of hierarchical management of data and states across the frontend codebase.

## Thoughts on the first two possible alternatives above

### 1️⃣ Use `setValues`

In a location similar to [this revision](https://github.com/woocommerce/google-listings-and-ads/blob/25a7d8ae714ed7e0495a77c0507eeb113ff4c167/js/src/components/shipping-rate-section/shipping-rate-section.js#L26-L39), change the handler to an implementation like this:

```js
const { getInputProps, values, setValues } = formProps;

const handleShippingRateOptionChange = ( value ) => {
	switch ( value ) {
		case 'automatic':
		case 'flat':
			setValues( {
				'shipping_rate', value,
				'shipping_time', 'flat',
			} );
			break;

		case 'manual':
			setValues( {
				'shipping_rate', value,
				'shipping_time', value,
			} );
			break;
	}
};
```

This solution looks intuitive and simple, but relies on the functionality of `Form`, and the code smells of high coupling still exists, even more than at present. I personally wouldn't recommend this approach given that the entire Onboarding and Free Listings settings were already very complicated.


### 2️⃣ Debounce the callbacks of `onSettingsChange` or the calls of `saveSettings`

It could be something like these removed hooks:

- [useAutoSaveSettingsEffect](https://github.com/woocommerce/google-listings-and-ads/blob/d23fdf59e3d9d3b0e3eed80f647b282a5a709ddb/js/src/setup-mc/setup-stepper/setup-free-listings/useAutoSaveSettingsEffect.js)
- [useDebouncedCallbackEffect](https://github.com/woocommerce/google-listings-and-ads/blob/d23fdf59e3d9d3b0e3eed80f647b282a5a709ddb/js/src/hooks/useDebouncedCallbackEffect.js)

In earlier versions, several places originally had debouncing handling. I intentionally removed them when developing [Free Listings + Paid Ads](https://github.com/woocommerce/google-listings-and-ads/pull/1616), because those uses of debouncing were more like workarounds, in my opinion.

- It is not a mandatory solution and could be better handled by controlling the triggering of `on*Change` See issue [#1335](https://github.com/woocommerce/google-listings-and-ads/issues/1335) and PR [#1616](https://github.com/woocommerce/google-listings-and-ads/pull/1616).
- Some issues like the [flashing shipping values](https://github.com/woocommerce/google-listings-and-ads/issues/1078) were indirectly caused by it.
- It also complicates error handling and the reference management of React hook, and reduces flexibility. See [#1313 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/1313#discussion_r825075174) and [#1313 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/1313#discussion_r828740923)
- From a certain point of view, it's an unnecessary ping pong as well.

A more suitable use case would be, for example, debouncing the continuous typing on an `<input>` and then trigger the API request for the proposed results.

I'd say this approach is not ideal as well.

## My proposal for the strategy of hierarchical management of data

I have tried to advocate this concept, and some possible problems I mentioned at the time have actually happened. Related discussions can be found here: https://github.com/woocommerce/google-listings-and-ads/pull/1266#discussion_r820377015 and https://github.com/woocommerce/google-listings-and-ads/pull/1317#discussion_r827607006. Therefore, I would like to try to advocate the same idea again.

### The reasons

- With the handling of data, states, errors, validations and a large variety of events, forms are inherently complex in general, as applications continue to iteratively add more features, improve UI/UX, and become easier for users to use, things might get trickier and trickier for devs.
- Reduce coupling
   - If data obtained from API must be maintained the consistency of its data structure (shape) across the API communication like creation or updating, in `wp-data` layer, in form layer, and in form input components layer, then a relevant change to a field may involve all of them.
   - It also reduces the reusability of components since they are highly interdependent in the data processing.
- Separation of concerns
   - There are no clear boundaries for what a component is concerned with, e.g. a form input component updates field A while updating another field B that is not directly related to it but used in another component.
   - Clear boundaries can also bring clarity to the test objectives and scope.

### Hierarchy

- In the `wp-data` store, it's in charge of the data bridges between backend APIs and application.
   - Resolvers and Actions focus on how to communicate with APIs and what bi-directional data adaptations are needed.
   - Reducer focuses on how to maintain the store tree.
   - Selectors/Actions is definitely the interfaces for retrieving/mutating the store.
- In the specific implementation of a component like `SetupFreeListings` based on the `Form` component
   - Separately manage data and states used in multiple sub-components.
   - The structure of `initialValues` depends more on the form UI instead of sticking with the data shape got from `wp-data`.
   - In complex use cases, propagate the data out via custom event callbacks for the component consumer side to handle the `wp-data` communications rather than communicating with `wp-data` directly.
   - Managing data correlation internally, such as changing target countries, may also change shipping settings, but the data correlation required by the backend API should be handled in the `wp-data` layer.
- In the specific implementation of a component for interacting with user inputs such as `ShippingRateSection`
   - Focuses on how to handle user inputs and callbacks the processed data to its form.
   - Should be agnostic to other irrelevant data as possible. For example, in `TargetAudience` component, changing shipping rates data should be considered an anti-pattern.
   - Of course, it should try to avoid changing external data via wp-data as well.

### Demo

As the actual practice of my thought is easier to demonstrate with code, please refer to add851c88d709329cc6b0bc75d891a61624fda9d for the changes.

In this way,

- Data processing for `shipping_rate` and `shipping_time` is restricted to actions.js.
- The confusing ping pong process in `SetupFreeListings` has been completely eliminated.
- `onSettingsChange` is still not triggered continuously.
- The form and related components no longer have the `shipping_time` field they don't need to know or even to mutate.
- Lower coupling, of course.

Obviously, there is one problem that remains unresolved. The special logic of `'automatic'` for shipping time is still scattered in `FormContent` and `checkErrors`.

However, using the same concept, the logic could be further moved to a util or `useSettings`, and provided as a prop to `FormContent` and as an external parameter to `checkErrors`. That is, we could also give that: Special logic is expected to be handled in [some] layer/place.

## Other solutions?

This comment depends mainly on my subjective thoughts and personal experience. I would be happy to learn more if there are other solutions for this case.

PR #1751, opened 2022-11-08
Tours API Endpoint

puntope:

This PR implements an API endpoint for handling the Tour Visualisation.

Tours are going to be a new experience provided by [Woo Spotlight component.](https://ghidorahp2.wordpress.com/2022/04/13/shared-in-app-spotlight-component/) 

When the user is entering the first time, is going to see this tour spotlights. When the user clicks on "Got it" button. We won't show this anymore. So we will save this as an option in the Database. 

For now it will be only 1 tour for Attribute Mapping. But in next iteration we will show a new one in case the user has invalid product data. So, if in the future there are more tours. We can add them here as well.

1.  Checkout this PR
2.  Create a new tour by requesting `POST wp-json/wc/gla/tours/` with this body:
```
{
    id: 'test',
    checked: true,
}
```
3. Get the tour by requesting `GET wp-json/wc/gla/tours/test`
4. You should get the tour you just created
5. Try to get a not existing tour, and try to send invalid data like missing props or props with wrong types.


src/API/Site/Controllers/TourController.php
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		/**
+		 * GET The tour visualizations
+		 */
+		$this->register_route(
+			'/tours/(?P<id>[a-z]+)',
```
ianlin: ❔ Not a blocking one, just thinking it might be a bit strict for the ID to be only `a-z`, which means the ID such as `test1` or `test-1` would not have a matching route.

Side note: TIL though we use `[a-z]+` here, it still work if we pass the upper case letters.

src/API/Site/Controllers/TourController.php
```diff
@@ -0,0 +1,171 @@
...
+	/**
+	 * Get the item schema properties for the controller.
+	 *
+	 * @return array The Schema properties
+	 */
+	protected function get_schema_properties(): array {
+		return [
+			'id'      => [
+				'description'       => __( 'The Id for the tour.', 'google-listings-and-ads' ),
+				'type'              => 'string',
```
ianlin, line 147: 💅  We should add a pattern validation for `POST /tours` to align with the API `GET /tours/(?P<id>[a-z]+)`. If we want to restrict the ID to be only letters it would be great adding a validation here so the ID can be verified when creating a tour.

```suggestion
				'type'              => 'string',
				'pattern'           => '^[a-z]+$',
```

tests/Unit/API/Site/Controllers/ToursControllerTest.php
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	public function test_post_tour_invalid() {
+		$response = $this->do_request( self::ROUTE, 'POST', [] );
+		$this->assertEquals( 400, $response->get_status() );
+	}
+
+	public function test_get_tour_not_found() {
+		$response = $this->do_request( self::ROUTE . '/not_exists' );
+		$this->assertEquals( 404, $response->get_status() );
```
ianlin: 💅  Maybe we could add a new assertion to check the error message. In this case the message should be `No route was found matching the URL and request method.`.

Also we might need to add one more test for testing the error message 'Tour not found'.

ianlin (commented):
Thanks for adding the new Tour API, I can create tours and get the data I just created. I left some comments regarding the validations.

src/API/Site/Controllers/TourController.php
puntope: Interesting, thanks. I ended up adding A-Z in the regex. Even the a-z supports them as you mentioned  I think it will be clear to explicitly add it in the regex.

src/API/Site/Controllers/TourController.php
puntope: Thanks. I added [a-zA-Z0-9-_]

puntope:
Howdy @ianlin . Thanks for the review. I added your suggestions. This is ready for another round.

src/API/Site/Controllers/TourController.php
```diff
@@ -21,6 +21,11 @@
 class TourController extends BaseOptionsController {
 
 
+	/**
+	 * Defines the regex to be used for the Tour ID
+	 */
+	private const TOUR_REGEX = '[a-zA-z0-9-_]+';
```
ianlin: This is an unused constant.

src/API/Site/Controllers/TourController.php
```diff
@@ -148,6 +153,7 @@ protected function get_schema_properties(): array {
 				'type'              => 'string',
 				'validate_callback' => 'rest_validate_request_arg',
 				'required'          => true,
+				'pattern'           => $this->get_tour_id_regex(),
```
ianlin: I can still create a tour with some symbols like `@` in the tour name. It only works if the regex pattern has a begin and ending sign, e.g. `^[a-zA-z0-9-_]+$`.

ianlin (commented):
Thanks for the changes. Just one more comment regarding the regex pattern when creating a tour.

src/API/Site/Controllers/TourController.php
puntope: Amazing catch! Thanks

I adjusted the code again.

ianlin (approved):
Thanks for all the changes, tested good locally.

PR #1752, opened 2022-11-08
Attribute Mapping - Spotlight Component

puntope:

This PR integrates TourKit in Attribute Mapping page and in the navigation 

### Screenshots:

Screenshot 2022-12-14 at 18 07 37

Screenshot 2022-12-14 at 18 07 51

1. Enter into the GLA Dashboard. You will see the tour over Attribute mapping tab.
2. Navigate over different tabs. (Except Attribute Mapping) and see how the tour persists.
3. Navigate to Attribute Mapping. It doesn't show anymore the tour for the navigation. Instead It shows another tour for the feature.
4. Navigate back to other tab. See how tour for the navigation appears again.
5. Click on close or "Got it" and see how the tour disappears.
6. Refresh the page and see how it won't show the tour.
7. Go to Attribute Mapping tab. See the tour. Click on Got it and see how it disappears.
8. Refresh the page and see how it won't show the tour.

### Additional details:

⚠️  Test four the tours are not implemented in this PR since they are still creating some issues and needs further investigation. It will be wrangled in cooldown.

### Changelog entry

> Add - Interactive discovery message for Attribute mapping feature.


tomalec:
The failure seems like a deeper issue.

> TypeError: list.addEventListener is not a function


The new `/viewport` used by `/components` checks matchmedia https://github.com/WordPress/gutenberg/blob/trunk/packages/viewport/src/listener.js#L41-L44 then attaches a listener to it, which is deprecated https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener

I think it's worth rising an issue upstream.

tomalec:
Check if you updated jest presets => `wp/scripts` to match so they mock those in testing environment.

Like https://github.com/WordPress/gutenberg/commit/2c45ac18051aa4735558a0da8e02da1c0a86986d

tomalec:
This is really strange. I put a debugger in 
`node_modules/@wordpress/viewport/src/listener.js` at `list.addEventListener('change', setIsMatching);`

and calling `list.addEventListener()` from console works, but the error is thrown anyway.

I'm afraid it's some webpack/compiler/jest-transform issue

tomalec:
Got it :facepalm: it's `addListener` not `addEventListener`
> In older browsers MediaQueryList did not yet inherit from [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget), so this method was provided as an alias of [EventTarget.addEventListener()](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). Use addEventListener() instead of addListener() if it is available in the browsers you need to support.

tomalec:
Maybe we can try to resolve it by juggling `@woocommerce/components` and `@wordpress/components` versions.

tomalec:
I tried mocking it locally, but for some reason, if I only use a function there, Jest fails with ESM transforms :confused: :shrug: 
https://github.com/woocommerce/google-listings-and-ads/compare/add/attribute-mapping/spotlight...add/attribute-mapping/spotlight-mock?expand=1

I think the only way is to juggle versions, or maybe you have an idea hot to solve that Jest thingy.
(I wish, we could finally use ESM everywhere without transpiling back and forth)

puntope:
Thanks Tomek for looking at this! 

> I think it's worth rising an issue upstream.

Do you know which team is responsible for this? Ghiorah?

puntope:
I think it would be faster to fix it ourselves directly in Woo Repo. In case it's just that error. However, I'm afraid there are more errors after that one. I did a mockup of viewport in jest.config, but new errors happened.

> I think the only way is to juggle versions, or maybe you have an idea hot to solve that Jest thingy.
(I wish, we could finally use ESM everywhere without transpiling back and forth)

Could you elaborate this?

tomalec:
> Do you know which team is responsible for this? Ghiorah?

I don't know, I'd just post it on the GitHub repo, as any other 3PD extension developer would do :)

tomalec:
> I think it would be faster to fix it ourselves directly in Woo Repo. In case it's just that error. However, I'm afraid there are more errors after that one. I did a mockup of viewport in jest.config, but new errors happened.

Most probably, it will be faster for us here. but by flagging it upstream, we may save some time fo other teams, and rise awareness for the maintainers for similar problems in the future. (However, given the enormous number of issues in Gutenbergs repo, this may be missed drop in the ocean)

> > I think the only way is to juggle versions, or maybe you have an idea hot to solve that Jest thingy.
(I wish, we could finally use ESM everywhere without transpiling back and forth)
>
> Could you elaborate this?

- "I think the only way is to juggle versions" - I guess in Gutenberg at the point when the change to `/viewport` was introduced, the `/jest-preset-default` was also updated to match. From Gutenberg **monorepo** perspective that was easy, same point in git history. But for us - external consumers of `@woocommerce/components@a.b.c` & `@wordpress/scripts@x.y.z` it's a complex puzzle to solve.


> or maybe you have an idea how to solve that Jest thingy

Those Jest error messages about transforming ESM <-> CJS are always confusing to me, and I feel lost there. But maybe you get that better than me.  So we could use local mock as in https://github.com/woocommerce/google-listings-and-ads/compare/add/attribute-mapping/spotlight...add/attribute-mapping/spotlight-mock?expand=1, but using `addEventListener: () => {}` and somewhat easily setup jest to run.

> (I wish, we could finally use ESM everywhere without transpiling back and forth)

I wish we could use JavaScript modules, `import/export` syntax, and let those files be used as such by the JS engine, without a need for a stack of tools that transpires it.
I see no value in being attached to CJS legacy. 

The Node version we use locally supports ESM. The browser we use for testing supports ESM. The browsers we target for WP admin support ESM. Why then at any point of development or build process we need to use babel or any other transpiler?

It only complicates the stack and process. It prolongs the build, and increases the number of dependencies, increasing the potential points of failure.

puntope:
> I wish we could use JavaScript modules, import/export syntax, and let those files be used as such by the JS engine, without a need for a stack of tools that transpires it.
I see no value in being attached to CJS legacy

In which config do you change this? I don't fully understand how to disable the transpiler if that solves the poroblem.

puntope:
So the error in Viewport was solved by mocking that component. 
Then I got another error in regards `es6 export` with `is-plain-obj` dependency. I think is solved with adding the path to. the `transformIgnore`  

Now the new issue happened in which seems like the test doesn't like the Hooks, so I guess is due to some mismatching react/react-dom version. But I'm unsure at all.

tomalec:
> In which config do you change this? I don't fully understand how to disable the transpiler if that solves the poroblem.

Sorry for not making it clear enough. I didn't mean any change that could be done here.

That was a high-level idea/goal to remove transpiling from our build process. That would mean changing webpack, jest, and probably our dependencies which are still using CJS.

Also, I'm not sure if that would solve the problem we have here. I just think it would simplify the debugging and reduce the bug-prone layers.

tomalec:
Also, I'd try doing one smaller step before, and update `@woocommerce/componenets` to the [version used by 6.9 - 10.3.0](https://github.com/woocommerce/woocommerce/blob/6.9.4/packages/js/components/package.json#L3).
So the version we tested manually E2E, and see if that works with less conflicts.

tomalec:
FWIW, you may find https://github.com/woocommerce/google-listings-and-ads/pull/1774 useful for checking minimum versions of WC dependencies.

eason9487:
After several rounds of investigation, I think one workable solution for solving all issues in running jest is to align all `@wordpress/*` and `@woocommerce/*` packages with a specific WC version. Therefore, the react version across these packages would be almost consistent.

A separate PR based on #1775 should make this upgrading easier to move forward, so I handled it separately with #1779, and more details were recorded there.

js/src/data/resolvers.js
```diff
@@ -462,3 +463,26 @@ export function* getStoreCategories() {
...
+ * Resolver for getting the Tours.
+ *
+ * @param {string} tourId The tour to get
+ */
+export function* getTour( tourId ) {
+	const response = yield apiFetch( {
+		path: `${ API_NAMESPACE }/tours/${ tourId }`,
+	} );
+
+	yield receiveTour( response );
```
eason9487: Not sure if I missed anything but an uncaught error occurred here when the first time I viewed the Dashboard page. The GET `tours/AttributeMappingNavigationTour` request responded with 404 status and an error message "Tour not found".

js/src/data/resolvers.js
```diff
@@ -462,3 +463,26 @@ export function* getStoreCategories() {
...
+
+/**
+ * Refresh getTour if UPSERT happens in the APP.
+ *
+ * @param {Object} action The performed action
+ * @return {boolean} True if the action should be invalidated
+ */
+getTour.shouldInvalidate = ( action ) => {
+	return action.type === TYPES.UPSERT_TOUR;
+};
```
eason9487: This invalidator could be omitted as the `upsertTour` action will send `TYPES.UPSERT_TOUR` and the latest tour to reducer to make the `tours` state in wp store up-to-date. Or, maybe also match the tour ID here. Otherwise, once a tour is updated, this invalidator will trigger all selected tours to be reloaded, which means triggering many API requests to fetch the already exactly same data for all using tour entities.

The concern can be simulated after addressing the uncaught error above.
1. Navigate to the dashboard page.
2. Open the Console tool.
3. Run `wp.data.select('wc/gla').getTour('rrrr')`
4. Run `wp.data.dispatch('wc/gla').upsertTour({ id: 'rrrr', checked: false })`
5. It shows that there is a new GET request `tours/AttributeMappingNavigationTour` although the updating tour is another one.
6. The "Introducing Attributes" popover also disappears for a moment after the irrelevant tour is updated.

https://user-images.githubusercontent.com/17420811/217758581-4591ba43-ae6c-485a-a9b7-c6bf8707fbbc.mp4

js/src/attribute-mapping/attribute-mapping-feature-tour.js
```diff
@@ -0,0 +1,65 @@
...
+ * Internal dependencies
+ */
+import useTour from '.~/hooks/useTour';
+const TOUR_ID = 'AttributeMappingFeatureTour';
+
+const AttributeMappingFeatureTour = () => {
+	const { setTour, showTour } = useTour( TOUR_ID );
+
+	const closeHandler = async () => {
+		await setTour( { id: TOUR_ID, checked: true } );
```
eason9487, line 17: It might occur uncaught error here as the `setTour` (`upsertTour`) could throw errors.

js/src/attribute-mapping/attribute-mapping-navigation-tour.js
```diff
@@ -0,0 +1,58 @@
...
+ * Internal dependencies
+ */
+import useTour from '.~/hooks/useTour';
+const TOUR_ID = 'AttributeMappingNavigationTour';
+
+const AttributeMappingNavigationTour = () => {
+	const { setTour, showTour } = useTour( TOUR_ID );
+
+	const closeHandler = async ( _data, _step, origin ) => {
+		await setTour( { id: TOUR_ID, checked: true } );
```
eason9487, line 18: It might occur uncaught error here as the `setTour` (`upsertTour`) could throw errors.

js/src/data/actions.js
```diff
@@ -1120,3 +1120,44 @@ export function* receiveStoreCategories( storeCategories ) {
 		storeCategories,
 	};
 }
+
+/**
+ * Action to receive the Tours.
+ *
+ * @param {Array} tour The tour to receive
```
eason9487: The type seems to be an object.

js/src/tests/dependencies/wordpress/viewport.js
```diff
@@ -0,0 +1,5 @@
+export function withViewportMatch() {
```
eason9487, line 1: This file is no longer in use.

js/src/hooks/useTour.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+/**
+ * @typedef {import('.~/data/selectors').Tour} Tour
+ */
+
+/**
+ * @typedef {Object} TourHook
+ * @property { { data: Tour, hasFinishedResolution: boolean, isResolving: boolean, invalidateResolution: Function } } tour The tour object
+ * @property {React.Dispatch<React.SetStateAction<Tour>>} setTour Setter for the hook
```
eason9487: The type of this property probably needs to be clarified. I don't think the dispatcher of wp-data is the same dispatcher as the updater of the `useState` in React. The `useDispatch` of wp-data is more likely a hook wrapper to connect the actions of wp-data via the React Context of wp-store. See:
- https://github.com/WordPress/gutenberg/blob/%40wordpress/data%406.15.0/packages/data/src/components/use-dispatch/use-dispatch.js#L53-L54
- https://github.com/WordPress/gutenberg/blob/%40wordpress/data%406.15.0/packages/data/src/components/registry-provider/use-registry.js#L50-L51
- https://github.com/WordPress/gutenberg/blob/%40wordpress/data%406.15.0/packages/data/src/registry.js#L168-L174

js/src/hooks/useTour.js
```diff
@@ -0,0 +1,41 @@
...
+ * @typedef {import('react').React} React
+ */
+
+/**
+ * @typedef {import('.~/data/selectors').Tour} Tour
+ */
+
+/**
+ * @typedef {Object} TourHook
+ * @property { { data: Tour, hasFinishedResolution: boolean, isResolving: boolean, invalidateResolution: Function } } tour The tour object
```
eason9487: The type of the `data` could be undefined before resolving or the querying tour doesn't exist after resolving.

js/src/hooks/useTour.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+const useTour = ( tourId ) => {
+	const tour = useAppSelectDispatch( 'getTour', tourId );
+	const { upsertTour } = useAppDispatch();
+
+	const showTour = tour.hasFinishedResolution && ! tour.data?.checked;
+
+	return {
+		tour,
+		setTour: upsertTour,
```
eason9487: As the interfaces of this hook, the function signature of `setTour` could only accept the `checked` because the `tourId` is already known when calling the hook. So that,
- The callers would not need to shape the actual request body like `setTour( { id: TOUR_ID, checked: true } )` but only need to call it like `setTourChecked( true )`.
- Avoid callers accidentally calling `setTour` with inconsistent tour IDs. For example:
   ```js
   const { setTour, tour } = useTour( 'ID-A' );

   const closeHandler = async () => {
     const id = tour.data?.id;
     // id could be `undefined`
     await setTour( { id, checked: true } );

     // Or even calls with another ID
     await setTour( { id: 'ID-B', checked: true } );
   };
   ```

eason9487 (commented):
Thanks for the work.

There may be some problems since both tours didn't show up while testing without any local changes.

image

I also viewed the code and left some comments. The main concerns are about three places that may occur uncaught errors, and the whole tour states in wp-data store will be re-obtained after updating single one tour. The rest comments are minor.

eason9487:
Just noticed the **status: on hold** label after finishing the code review.

I originally tried to base on this branch and then implement the tour for PMax Assets project, but then encountered some issues so I turned to start the code review.

js/src/data/resolvers.js
eason9487: Fixed by https://github.com/woocommerce/google-listings-and-ads/commit/e2878dd3bb1a3fd9a137f534764a9ee4e9358606 in #1879.

js/src/data/resolvers.js
eason9487: Removed by https://github.com/woocommerce/google-listings-and-ads/commit/96863790cec82ef253fafe56fe95214a2f48e647 in #1879.

js/src/data/actions.js
eason9487: Had been fixed by https://github.com/woocommerce/google-listings-and-ads/commit/7e2a3187b14b6454a74719bca02811ba88c82a1c and then updated by https://github.com/woocommerce/google-listings-and-ads/commit/fce415882f4a25d7dcded71467405bd7767c30c7 in #1879.

js/src/hooks/useTour.js
eason9487: Had been fixed by https://github.com/woocommerce/google-listings-and-ads/commit/7e2a3187b14b6454a74719bca02811ba88c82a1c and then removed by https://github.com/woocommerce/google-listings-and-ads/commit/2aa7988850f990b2480251e42d8c111772bb71fe in #1879.

js/src/hooks/useTour.js
eason9487: Had been fixed by https://github.com/woocommerce/google-listings-and-ads/commit/b42e36865bd885ee23500512edad211be9a16367 and then removed by https://github.com/woocommerce/google-listings-and-ads/commit/2aa7988850f990b2480251e42d8c111772bb71fe in #1879.

js/src/hooks/useTour.js
eason9487: Added `setTourChecked` setter by https://github.com/woocommerce/google-listings-and-ads/commit/c06bcda0fa9d8761b387607553fe94735e789eca and then removed `setTour` by https://github.com/woocommerce/google-listings-and-ads/commit/2aa7988850f990b2480251e42d8c111772bb71fe in #1879.

eason9487:
As per the last follow-up step in https://github.com/woocommerce/google-listings-and-ads/pull/1879#issuecomment-1434094218, all changes made in #1879 have been synced to this PR.

js/src/attribute-mapping/attribute-mapping-feature-tour.js
eason9487: This is no longer applicable after adopting https://github.com/woocommerce/google-listings-and-ads/commit/2aa7988850f990b2480251e42d8c111772bb71fe in #1879.

js/src/attribute-mapping/attribute-mapping-navigation-tour.js
eason9487: This is no longer applicable after adopting https://github.com/woocommerce/google-listings-and-ads/commit/2aa7988850f990b2480251e42d8c111772bb71fe in #1879.

puntope:
No longer needed

PR #1759, opened 2022-11-10
WC 7.1 Compatibility: Fix Form's onChange is called unexpected times which leads to infinite requesting state updates

eason9487:

- Create a new component `AdaptiveForm` for wrapping workarounds for the `Form` of `@woocommerce/components` in order to be compatible with WC 6.9 to 7.1.
- Fix the issue of unanticipated number of calls Form's `onChange`.
   - Ref: https://github.com/woocommerce/google-listings-and-ads/pull/1750#issuecomment-1309300428

1. Install WooCommerce 7.1.0
2. Go to the Edit free listings page
3. Change the shipping options:
   - When the recommended or simple shipping rate is selected, shipping time settings should be displayed.
   - When the complex shipping rate is selected, shipping time settings should be hidden.
4. Free Shipping selection
   1. Add shipping rates > 0
   1. Select "Yes" option for the Free Shipping
   1. Add the Minimum order to qualify for free shipping with any value bigger than 0
   1. Edit the shipping rates to 0
   1. "Save Changes" button should be enabled
5. Minimum order settings
   1. Add shipping rates > 0
   1. Select "Yes" option for the Free Shipping
   1. Add the Minimum order to qualify for free shipping with any value bigger than 0
   1. Select "No" option for the Free Shipping and then back to "Yes"
   1. The value of Minimum order... should be cleared
6. Target audience and shipping
   1. Remove a target audience country that has set the relevant shipping rate and time.
   1. Check if the removed country is also removed from the shipping rate and time settings.
7. During the test, there should not be a blank page, and there should not be unexpected callbacks if inspect the  `onChange` calls of `<From>`.
8. Install WC [7.0, 6.9]. Repeat steps 2 - 7.

💡 With this solution, it should be also compatible with some WC versions < 6.9, such as 6.8.

### Additional details:

The codes below lead to trigger `onChange` for each item in the values regardless of whether a value change did happen, even if we only call `setValue` for updating a single value.
- https://github.com/woocommerce/woocommerce/blob/7.1.0/packages/js/components/src/form/form.tsx#L209-L211
- https://github.com/woocommerce/woocommerce/blob/7.1.0/packages/js/components/src/form/form.tsx#L182-L197

### Changelog entry

> Fix - WC 7.1 compatibility: Fixing the forms in the free listings setup may cause infinite requesting state updates which lead to a blank page or issue a lot of API requests.

### DIFF
--- a/js/src/components/adaptive-form/adaptive-form.js
+++ b/js/src/components/adaptive-form/adaptive-form.js
@@ -23,7 +23,7 @@ function isEvent( value ) {
  */
 
 /**
- * Renders an adapted form component that wraps the `Form` of `@wordpress/element` with
+ * Renders an adapted form component that wraps the `Form` of `@woocommerce/components` with
  * several workarounds in order to be compatible with WC 6.9 to 7.1.
  *
  * @param {Object} props React props.
--- a/js/src/components/adaptive-form/adaptive-form.js
+++ b/js/src/components/adaptive-form/adaptive-form.js
@@ -125,8 +125,8 @@ function AdaptiveForm( { children, ...props }, ref ) {
 					// Use `setImmediate` to avoid the warning of request state updates while rendering.
 					// Mutating a React hook state is an anti-pattern in most cases. Here is done intentionally
 					// because it's necessary to ensure this component will be triggered re-rendering through
-					// `setBatchQueue`, but also to avoid calling `settingBatchQueue` here and triggering
-					// additional rendering again.
+					// `setBatchQueue`, but also to avoid calling `setBatchQueue` here and triggering additional
+					// rendering again.
 					setImmediate( () => setDelegation( batchQueue.shift() ) );
 				}
 
--- a/js/src/components/adaptive-form/adaptive-form.js
+++ b/js/src/components/adaptive-form/adaptive-form.js
@@ -121,9 +121,12 @@ function AdaptiveForm( { children, ...props }, ref ) {
 
 				// Related to WC 6.9. Only one delegate can be consumed at a time in this render prop to
 				// ensure the updating states will always be the latest when calling.
-				//const { batchQueue } = adapterRef.current;
 				if ( batchQueue.length ) {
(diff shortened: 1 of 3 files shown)


tomalec:
Looks promissing, but running in with WC7.1 and WP6.1, still breaks onboarding:
image
```
Uncaught TypeError: Cannot read properties of null (reading 'length')
    at checkErrors (checkErrors.js:95:46)
    at Object.handleValidate [as validate] (index.js:104:21)
    at index.js?ver=1668027741:2:599361
    at index.js?ver=1668027741:2:599428
    at invokePassiveEffectCreate (react-dom.js?ver=17.0.1:23517:22)
    at HTMLUnknownElement.callCallback (react-dom.js?ver=17.0.1:3942:16)
    at Object.invokeGuardedCallbackDev (react-dom.js?ver=17.0.1:3991:18)
    at invokeGuardedCallback (react-dom.js?ver=17.0.1:4053:33)
    at flushPassiveEffectsImpl (react-dom.js?ver=17.0.1:23604:11)
    at unstable_runWithPriority (react.js?ver=17.0.1:2764:14)
react-dom.js?ver=17.0.1:20115 The above error occurred in the <ForwardRef> component:

    at https://wp600wc700.test/wp-content/plugins/woocommerce/assets/client/admin/components/index.js?ver=1668027741:2:598957
    at AdaptiveForm (https://wp600wc700.test/wp-content/plugins/google-listings-and-ads/js/build/index.js?ver=73999960da3a0bcb6b4ec95ddf622a9e:24424:5)
    at div
    at SetupFreeListings (https://wp600wc700.test/wp-content/plugins/google-listings-and-ads/js/build/index.js?ver=73999960da3a0bcb6b4ec95ddf622a9e:29520:5)
```

tomalec (approved):
Reviewed the code, tested locally, LGTM.

Thank you for the elaborate code comments, they helped me to understand what's happening, and I hope they will help us in the future to navigate reduce/adapt those workarounds as WC moves forward.

PR #1755, opened 2022-11-09
WP 6.1 Compatibility: Fix styles of `Popover`, `Tooltip` and `DatePicker` components

eason9487:
### Changes proposed in this Pull Request:

Closes #1748 

Fix WP 6.1 compatibility:

- Add a prebuild processing for importing the needed old styles.
   - Import popover style in the following way won't work since [some parent selectors were placed in the middle](https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4012.0.8/packages/components/src/popover/style.scss#L174-L215). Therefore, the workaround is to build those scss files first and then import the built css file directly under the `.gla-admin-page` CSS class.
      ```scss
      .gla-admin-page {
        @import "node_modules/@wordpress/components/src/popover/style";
      }
      ```
   - Please refer to the "Heads up!" block in the Nesting section: https://sass-lang.com/documentation/at-rules/import#nesting
- The styles of `Popover` and `Tooltip` components.
- Extra: The `DatePicker` components.
   image

### Screenshots:

https://user-images.githubusercontent.com/17420811/200822442-ad0f776b-ba4f-44e8-8974-7f9271e426b1.mp4

### Detailed test instructions:

1. Install WP 6.1.
2. Check if the build scripts can run successfully and if the development scripts can rebuild changes.
   - `npm start`. Make some changes to JS files and the prebuild.scss. Refresh page to view the changed parts.
   - `npm run start:hot`. Make some changes to JS files and the prebuild.scss. Check if the page is updated accordingly without refreshing the page on the browser level.
   - `npm run dev`
   - `npm run build`
3. Go to the Dashboard page.
   - Click the ellipsis menu in the Programs table to check if the position and styles of the popover content are correct.
   - Click the date range option to open datepicker. Switch to custom tab to verify if the icon size of navigation buttons is appropriate.
4. Go to the Product Feed page.
   - Click the question mark icon in the Overview header. Check if the popover is displayed as floating.
   - Hover on the edit icon in the Product Feed table. Check if the tooltip is displayed.
5. Install a WP version >=5.9 and < 6.1
6. Repeat steps 2-4 to verify if this fix also works for other support WP versions.

### Additional details:

[...]

PR #1762, opened 2022-11-10
Import from an externalized package if explicitly asked for

### LINKED ISSUE #1736: WP 5.9 Compatibility: The style of `VisuallyHidden` component is not working

### LINKED ISSUE #1748: WP 6.1 Compatibility: The popover content is not displayed as floating
### Describe the bug:

The popover content of `HelpPopover` is not displayed as floating but is rendered after its anchor and breaks the layout.

### Steps to reproduce:

1. Install WordPress 6.1
2. Go to the Product Feed page
3. Click the question mark icon in the Overview header
4. Select at least one product in the Product Feed table, and then hover on the "Apply to # selected" button in the table header

### Expected behavior:

image

image

### Actual behavior:

image

image

tomalec:
This is an alternative/complimentary PR to https://github.com/woocommerce/google-listings-and-ads/pull/1739

- [Import from an externalized package if explicitly asked for,](https://github.com/woocommerce/google-listings-and-ads/commit/a57b0be588dd5c0c4dc4ca33b7b1bd10594401e1) 
   even if the package is bundled as well.
   So we could selectively externalize specific imports and gradually externalize all cases.

- [Import ExternalLink from DEWPed @wordpress/components,](https://github.com/woocommerce/google-listings-and-ads/commit/ad10229ada8c08d6542ac66260bee5b24a26242e) 
   even though we bundle that package.
   Fixes non-working `VisuallyHidden` styles for Google Ads settings https://github.com/woocommerce/google-listings-and-ads/issues/1736,
  and billing setup.

### Screenshots:

image

1. Checkout
2. `npm i && npm run build`
3. Go to paid ads setup `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads` with non-approved billing
4. Check The external links  on step "1 Set up your accounts" & "3 Set up billing"

### Additional details:
1. I'm using `@wordpress/dewp` implementation details here. This could be unstable. If we'd like to move on with this change, I'd propose a change upstream to improve DEWP to be more extendable. as IMHO, currently, it's not, that's why `@woocommerce/DEWP` needs to extend the class, and more-or-less duplicate `@wordpress`es code. 

   I believe `@woocommerce/DEWP` and our case here could be simplified if the exposed API, would allow doing something like:
```js
requestToExternal = ( request, defaultAction ) => {
	if ( request.startsWith( explicitlyExtractPrefix ) ) {
		 defaultAction( request.substr( explicitlyExtractPrefix.length ) )
   }
   // ...
}
```
or
```js
requestToExternal = ( request ) => {
	if ( request.startsWith( explicitlyExtractPrefix ) ) {
		 DependencyExtractionWebpackPlugin.requestToExternal( request.substr( explicitlyExtractPrefix.length ) )
   }
   // ...
}
```

2. Currently, this setup supports only packages extracted by `@wordpress/DEWP`, as `@woo…`'s one does not export its `requestToExternal`. So we would have to duplicate its code, which IMHO is even worse than the plugin into implementation details like `@…/dependency-extraction-webpack-plugin/lib/util`. I believe that could be solved in a cleaner fashion if the proposal from the point above gets accepted. Plus, being able to do so just for `@wordpress/` package is already a value.
3. Alternatively, we could explicitly bundle, instead of extract. Then the implementation would not face the problems of the points above, as we'd just do the current opt-out selectively.
   However, that would require us to prefix every import we have with `bundled/` which may bring lots of noise.
4. ~I added `eslint-import-resolver-alias` as I didn't manage to configure an alias for `node_modules` with `eslint-import-resolver-webpack`. I think ideally, we should stick to one. As we already have quite many places responsible for setting aliases: `webpack.config.js`, `js/src/jsconfig.json`, `jest.config.js`, now `.eslintrc.js`. I'd appreciate if somebody could clean this up.~ solved in https://github.com/woocommerce/google-listings-and-ads/pull/1762#issuecomment-1311438333
5. [Externalized report](https://github.com/woocommerce/google-listings-and-ads/pull/1762/files#diff-2ccc451e09095e78419f3f696655eda1da857eb2213c97996783d2accc6f042c) shows `extracted/@wordpress/components` as and externalized package. With more extendable DEWP, that exposes its primitives instead of simplifying everything to returning `undefined` on path, we could avoid that by doing something like:
```js
requestToExternal = ( request ) => {
	if ( request.startsWith( explicitlyExtractPrefix ) ) {
		 return externalizeWpDeps( request.substr( explicitlyExtractPrefix.length ) )
    }
   // filter bundled opt-outs.
```

### Changelog entry

> Fix - WordPress 5.9 Compatibility: Visually hide descriptions of external link icons.
> Fix - WordPress 6.1 Compatibility: Popover and Tooltip components should be displayed as floating.
> Dev - Selectively externalize bundled packages.


eason9487:
Thanks for the solution. I think it's better than importing the old styles as #1739 and #1755. More importantly, it's accompanied by the externalization processing of dependencies gradually. 💯

I tried to fix the incompatible `Popover` and `Tooltip` (#1755) with this approach, and also tried it on `Button`, which is the most widely used component in GLA. All worked well. 👍 

Although fixing `DatePicker` would still need to import the old style since its incompatibility is between `@woocommerce/components` and WP 6.1, we could drop the workaround of prebuild SCSS and do this:

```scss
.gla-admin-page {
  @import "node_modules/@wordpress/components/src/date-time/datepicker";
}
```

I think we could adopt this solution. If so, the next actions could be:
- #1739
   - Replace the fix of `VisuallyHidden` with the one for `DatePicker`
- #1755
   - Close it since we won't need it unless someday we still need to prebuild SCSS files, it could be referred to. (Hope it won't happen)
- This PR
   - After searching GLA codebase, looks like fixing `Popover` and `Tooltip` only needs to do the same `extracted/` prepending for `HelpPopover` and `AppTooltip`. Therefore, we could fix them in this PR as well.

WDYT?

eason9487:
> 4\. I added `eslint-import-resolver-alias` as I didn't manage to configure an alias for `node_modules` with `eslint-import-resolver-webpack`. I think ideally, we should stick to one.

Adding the following to [webpack config here](https://github.com/woocommerce/google-listings-and-ads/blob/d3807cad5b1f8cbbfb5f32910cf80bbe65df26ee/webpack.config.js#L79-L81) should make it.

```js
extracted: path.resolve( __dirname, 'node_modules' ),
```

tomalec:
> Adding the following to [webpack config here](https://github.com/woocommerce/google-listings-and-ads/blob/d3807cad5b1f8cbbfb5f32910cf80bbe65df26ee/webpack.config.js#L79-L81) should make it.
> ```
> extracted: path.resolve( __dirname, 'node_modules' ),
> ```

Thanks! I was trying `'extracted/` ;) :facepalm:

tomalec:
@eason9487 Thanks for the review and new ideas!

Unfortunately, `Popover` are still missing something:
image


I think I addressed all your suggestions. Would you mind another review?

eason9487:
> Unfortunately, `Popover` are still missing something:

The inner spacing style was added in GLA's `HelpPopover`. Probably the DOM structure of externalized `Popover` is different from the bundled one. May need some tweaks.

eason9487 (approved):
Thanks for the great solution. LGTM. The [style of Popover](https://github.com/woocommerce/google-listings-and-ads/pull/1762#issuecomment-1311615910) might need a bit of adjustment but is not a blocker.

I will handle the rest of action items in https://github.com/woocommerce/google-listings-and-ads/pull/1762#issuecomment-1311437832 later. 🤝

PR #539, opened 2021-05-02
Proposal/RFC scoped styles

tomalec:
This is an experiment/proposal/RFC for a style scoping solution to address https://github.com/woocommerce/google-listings-and-ads/issues/65
- leaking & conflicting styles,
- class names conflicts
- CSS-in-JS ergonomics

(Could be considered a native alternative to libraries like styled-components / emotion https://github.com/woocommerce/google-listings-and-ads/pull/71)

This change plays well with https://github.com/woocommerce/google-listings-and-ads/pull/538, but could be totally separated.

### Live demo

https://codesandbox.io/s/scoped-shadow-styles-for-react-82wrh?file=/src/components/QuoteCard.js

### Changes proposed in this Pull Request:

Encapsulate `ContentButtonLayout`'s styles.
Move them as inline styles to element's shadow root.

Add `useShadowStyles` hook,  to separate style encapsulation logic.

I'm using inline styles here to simplify the PR, we can use `CSSStyleSheets` but making them work in all browsers https://caniuse.com/mdn-api_cssstylesheet_replacesync would require a little bit more code that could blur the picture. I guess with our heavy cannon of webpack, generating stylesheet objects, adding css imports, adding them as chunked styles, should be a big deal.

See styleSheet version at [fix/65-encapsulated-stylesheets](https://github.com/woocommerce/google-listings-and-ads/compare/fix/65-encapsulated-stylesheets)

Please note, that components `children` are still rendered in the light DOM, so are accessible for external use, styling, and manipulation as there were before.

### Screenshots:

Screenshot of devtools showing `:host` styles for `ContentButtonLayout`

### Detailed test instructions:

1. Go to any page that uses `ContentButtonLayout` component, like https://gla1.test/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads
2. Open dev tools, highlight the element.

### Pros:

1. Styles do not leak. No other element would be styled with them, no matter what.
2. Element could be still easily styled/themed/specialized, with regular light DOM styles.
3. There is no class name to collide with something else.
4. Better DevX:
	1. Clear distinction, which styles pertains to the class which are instance-specific
		Shadow, `:host ...` selector styles are the class ones - applies to all instances.
		`.gla-some-class` indicates a special subset of instances to be selected.
	2. You don't have to think about the hardest thing in programming - a new class name.
	3. You can use CSS-in-JS to build, import, etc. your CSS string, or `CSSStyleSheet` object
[...]

PR #538, opened 2021-05-02
Proposal/RFC for custom tag names

tomalec:
This is an example for https://github.com/woocommerce/google-listings-and-ads/issues/84

### Changes proposed in this Pull Request:

Add `ContentButtonLayout` a custom tag name.

To reduce the div-soup, and make use of modern browser features.

I believe such change could be valuable for the components that render a separate wrapping container. Like `<div class="foo">`, as opposed to components that render a document fragment, or to-be-collapsed element. As we render a separate HTML element, why don't we put some sematic into its name, instead of creating 100s of `divs`.

Please note, that this PR does **not** define **a custom element** - it should work in IE8+.
I use a `custom-name` casing, to suppress React warning for `customname`.

### Screenshots:

Chrom Dev tools highlighting `woo-gla-content-button-layout` element

### Detailed test instructions:

1. Go to any page that uses `ContentButtonLayout` component, like https://gla1.test/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads
2. Open dev tools, highlight the element.

### Pros:

1. Better DevX:
	1. Less blurred picture with 1000s hypnotizing `div`s.
	2. Shorter rendered code `<woo-gla-content-button-layout>` vs `<div class=".woo-gla-content-button-layout">`
	2. Shorter working code [`update/84-custom-name?expand=1`#diff-794cef5df7](https://github.com/woocommerce/google-listings-and-ads/compare/update/84-custom-name?expand=1#diff-794cef5df71654781ea327d6ce357f0b0a19247add18a4df6a420f6c072a4cea)
	3. More semantic/informative HTML tree
	4. Makes it easier to find all instances of the rendered component with: `document.querySelectorAll('my-component-name')`
2. Works even without React. (if it's just a layout component as this one). No dependencies.
3. With the custom tag name, it is easier to relate it with React component, as opposed to reading the class names. See example below.

	  ```js
	  // What we have currently:
	  // it is not immediately clear this is rendered by which React component.
	  <div class="gla-content-button-layout gla-title-button-layout many-more-class-names">
	  
	  // What this PR proposes:
	  // from the custom tag name, you can guess it is coming from ContentButtonLayout component.
	  <woo-gla-content-button-layout class="class-names-here">
	  ```

### Cons:
It's hard for me to find significant ones, please shoot if you have any.

1. It does not follow the existing "make everything a `div`" convention. 
	- This is an argument we could put against any change, and I believe the current conventions do not bring much value themselves.
[...]

Issue #1391, opened 2022-04-01
Prevent the GLA button styles from overriding other styles or being overridden by other styles


eason9487:
### Describe the bug:

1. The GLA button styles might override other styles out of GLA pages.
2. The GLA button styles are overridden by other styles.

#### Overrides others
1. Go to Marketing -> Overview
1. Run script via DevTool console:
   ```js
   {
   	const button = document.createElement('button');
   	button.innerText = 'Howdy!';
   	button.classList.add( 'components-button', 'is-destructive', 'is-link' );
   	document.querySelector('#woocommerce-layout__notice-list').appendChild( button );
   }
   ```
1. The styles of just created "Howdy!" button is overridden by GLA styles.
   ![2022-04-01 12 45 38](https://user-images.githubusercontent.com/17420811/161196602-4d9c2a57-36b1-430d-ba52-849622d3d657.png)

#### Being overridden by others

https://github.com/woocommerce/google-listings-and-ads/pull/1374#issuecomment-1084623791

> Crazy, it comes from Automatewoo
> 
> <img alt="Screenshot 2022-03-31 at 15 52 46" width="338" src="https://user-images.githubusercontent.com/5908855/161072353-2a89a16c-d02c-4c52-bd30-76e4d45780d3.png">
> 
> <img alt="Screenshot 2022-03-31 at 15 55 04" width="371" src="https://user-images.githubusercontent.com/5908855/161072386-26bb5433-60e9-457e-9abe-06fdc3def2c3.png">

### Expected behavior:

The the GLA button styles should avoid conflict with other styles

### Actual behavior:

Conflicts between the GLA button styles and other styles

eason9487:
Closing this issue:

> 1. The GLA button styles might override other styles out of GLA pages.

This extension has been made to not override the button styles on other pages by #1834.

> 2. The GLA button styles are overridden by other styles.

With the release of AutomateWoo 6.0.6, AutomateWoo no longer overrides the button styles on other pages. In other extensions maintained by the team Automata, there are no known button style conflicts for now.

PR #528, opened 2021-04-28
Fix the giant icon problem in the warning messages

eason9487:
### Changes proposed in this Pull Request:

This problem was discovered in #527.

image

### Root cause
Libraries depend on a different version of `@wordpress/icons`, causing the react component instance checking `icon.type === SVG` to get false result.

https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4012.0.8/packages/components/src/icon/index.js#L41-L50

### Solution

Upgrade `@wordpress/icons` to the same version used in `@wordpress/components`.

### Screenshots:

image

### Detailed test instructions:

1. `npm install` and restart dev build `npm start`
2. Set the store location to “South Sudan” in WooCommerce → Settings
3. Head to the Get Started page and the external link icon should be displayed with a proper size.

PR #1764, opened 2022-11-11
Proposal/RFC scoped styles with custom elements

tomalec:

This is another proposal to use Shadow DOM to encapsulate at least some of our styles.

This is an alternative to https://github.com/woocommerce/google-listings-and-ads/pull/539 that joins https://github.com/woocommerce/google-listings-and-ads/pull/538
And has all the pros and cons of both :) 

#### Declutters HTML
- We no longer need to autogenerate class names and Emotion dependency to do scoping
- Element is denoted by its name, so we do not mix the purpose of class attribute of selecting a group of elements of a kind.
- 
image
#### Declutters CSS
- We now know which rules comes from element definition, which are applied by the consumer of an element.
image

image
#### Coding flexibility
- We can use single file components
- We are CSS-in-JS ready
- But we can still use separate CSS files if we want to
- We can now explicitly import and export just styles for elements, to extend/re-use them
image

1. checkout, `npm run build`
2. go to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings`
3. Inspect the "Connected" element. 

### Additional details:
1. To make it work in Safari, we'd need either use a CSS polyfill, like https://www.npmjs.com/package/construct-style-sheets-polyfill - still lighter, and more bulletproof than emotion. Or we can stamp `<style>` tag to every root to play safe.
2.  Also, in long run we could enrich the `css` tagged template, or replace it with something from an external library, like lit.
3. To get <code>css`</code> highlighted in VSCode, you need a plugin like https://marketplace.visualstudio.com/items?itemName=bashmish.es6-string-css or `lit-html`
4. If we would like to avoid custom element name collisions. In case somebody else happen to register an element with `<gla-connected-icon-label>` or we want to use two different versions of the same component. **What we currently cannot with `@wordpress/` and `@woocommerce/components`** we can simply add generated suffix to `defineShadowStylesHost` and make it return the name. Styles will keep their isolation thanks to Shadow DOM.

### Changelog entry

> Dev - scope styles with shadow hosting custom element.

### DIFF
--- a/js/src/components/connected-icon-label/index.js
+++ b/js/src/components/connected-icon-label/index.js
@@ -1,33 +1,39 @@
 /**
  * External dependencies
  */
-import classnames from 'classnames';
 import { __ } from '@wordpress/i18n';
-import { Flex, FlexItem } from '@wordpress/components';
 import GridiconCheckmarkCircle from 'gridicons/dist/checkmark-circle';
 
 /**
  * Internal dependencies
  */
-import './index.scss';
+import { css, defineShadowStylesHost } from '.~/utils/defineShadowStylesHost';
 
-const ConnectedIconLabel = ( props ) => {
-	const { className } = props;
+export const ConnectedIconLabelStyle = css`
+	:host {
+		display: inline-flex;
+		fill: currentcolor;
+		color: #23a713;
+		gap: var( --wp-grid-gap, 4px );
+		align-items: center;
 
-	return (
-		<Flex
-			className={ classnames( 'gla-connected-icon-label', className ) }
-			align="center"
-			gap={ 1 }
-		>
-			<FlexItem>
-				<GridiconCheckmarkCircle />
-			</FlexItem>
-			<FlexItem>
-				{ __( 'Connected', 'google-listings-and-ads' ) }
-			</FlexItem>
-		</Flex>
-	);
-};
+		direction: row;
+		justify-content: space-between;
+	}
+	::slotted( svg ) {
+		display: block;
+	}
+`;
+
+defineShadowStylesHost( 'gla-connected-icon-label', [
+	ConnectedIconLabelStyle,
+] );
+
+const ConnectedIconLabel = ( { className, ...props } ) => (
+	<gla-connected-icon-label class={ className } { ...props }>
+		<GridiconCheckmarkCircle />
+		{ __( 'Connected', 'google-listings-and-ads' ) }
+	</gla-connected-icon-label>
+);
 
 export default ConnectedIconLabel;
--- a/js/src/components/connected-icon-label/index.scss
+++ b/js/src/components/connected-icon-label/index.scss
@@ -1,8 +1,0 @@
-.gla-connected-icon-label {
-	fill: currentcolor;
-	color: #23a713;
-
-	svg {
-		display: block;
-	}
-}
--- a/js/src/utils/defineShadowStylesHost.js
+++ b/js/src/utils/defineShadowStylesHost.js
@@ -1,0 +1,50 @@
+/**
+ * Utility to define scoped styles using a custom element with a shadow root.
+ */
+/* global CSSStyleSheet, HTMLElement */
+/**
+ * A template literal tag to create stylesheet objects.
+ *
+ * It's overly simplify, in future we may introduce more features, like:
+ * - Safari support
+ * - lazy constructing - not to create object untill the first instance is used
+ * - safe rules nesting
+ * - etc.
+ *
+ * @param {TemplateStringsArray} strings
+ * @param  {Array} values
+ * @return {CSSStyleSheet} Constructed stylesheet.
+ */
+export const css = ( strings, ...values ) => {
+	const cssText =
+		strings.length === 1
+			? strings[ 0 ]
+			: values.reduce(
+					( accumulator, value, index ) =>
+						accumulator + value + strings[ index + 1 ],
+					strings[ 0 ]
+			  );
+	const sheet = new CSSStyleSheet();
+	sheet.replaceSync( cssText );
+	return sheet;
+};
+
+/**
+ * Creates a custom element that will act as the host for given styles.
+ *
+ * @param {string} name Valid custom element name.
+ * @param {Array<CSSStyleSheet>} styles Style sheets to be adoptes by the host.
+ */
+export const defineShadowStylesHost = ( name, styles ) => {
+	window.customElements.define(
+		name,
+		class extends HTMLElement {
+			constructor() {
+				super();
+				const shadowRoot = this.attachShadow( { mode: 'open' } );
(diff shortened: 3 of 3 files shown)


eason9487 (commented):
If I understood correctly, this solution is mainly to reduce/avoid the style naming conflicts with upstream or other extensions, and this brings a further benefit that the children's style naming within a scoped component will be easier, right?

If I remember rightly, the style conflicts that had/have happened in practice were probably these situations:
1. Conflicts between WP and WC components themselves due to the current use of DEWP
   - #528
   - #1739
2. GLA's style affects others
   - #1391
   - #1765
3. Others' style affects GLA
   - #1391
   - https://github.com/woocommerce/google-listings-and-ads/pull/1880#issuecomment-1432763685
   - #1909

In the first case, after WP and WC packages are externalized, it should not occur because of GLA's DEWP setting. In the second case, the only direct importing of `node_modules/@wordpress/components/src/panel/style` will also be removed soon, and with the `.gla-` prefix and `.gla-admin-page` scoping support already in place, this is unlikely to happen in the future.

However, the third one is still not effectively addressed, since most GLA components are heavily based on WP and WC components, there seems no way to completely avoid style conflicts from outside on its own CSS name (e.g. `.components-button`). This solution does not seem to be applicable to solving this problem either?

In addition, the Fast Refresh development mode (`npm run start:hot`) seems to conflict with this approach. Once a code change occurs within a component that uses `defineShadowStylesHost`, subsequent code changes will not make Fast Refresh update the page. Must refresh the page to resume.

tomalec:
> If I understood correctly, this solution is mainly to reduce/avoid the style naming conflicts with upstream or other extensions, and this brings a further benefit that the children's style naming within a scoped component will be easier, right?

Not quite.

This PRs aims to improve the DevX to make our developer life merrier, help to avoid bugs (incl. style conflicts), and speed up debugging. The improvements are listed in bullet points in OP. The ones I'd highlight:

- We have a way of denoting styles for a component (set in component's definition), distinct from a way to add styles for s sub`class` of instances by the consumers of those elements. (Currently, we use the `class` attribute for both)
- It's easier to find the component in the elements tab in dev tools
- Styles are modular and reusable in a more JSy way
- Styles are actually and technically isolated. So there is a technical solution protecting us from the style leak. As opposed `gla-` prefix, which is just a convention that reduces the chance of such leak but gives no technical protection.

Also, I'd like to emphasize one thing: this solution is mostly to **specify and encapsulate the styles for** an HTML **Element** (`<div>` or other) **we do create and govern**, not the ones created by external components (`<WooCommerceButton>`) (Unless we start using those in Shadow DOM, but it has more severe implications, that I'd like to avoid on this baby step). 
We are defining styles for a host we created.

So I do not aim to fix the issues you link. 

---------
 

> 1. Conflicts between WP and WC components themselves due to the current use of DEWP

As long as we do not have the ambition to change WP & WC conflict, this proposal does not aim to fix that problem. As you said DEWPing all the things should be the solution.

> 2. GLA's style affects others

As you mentioned, `.gla-` & `.gla-admin-page` fix those specific bugs. However, the solution proposed here aims to protect agins our GLA's styles from leaking and affecting others and has some advantages over `.gla-` prefix convention.

- It does not clutter the code with more class names and high CSS specificity, nor it requires and parent or wrapper element.
- Allows using/importing our component outside of the GLA scope and still be encapsulated.

> 3. Others' style affects GLA

The solution proposed here is to style only the specific, GLA-defined host element (`gla-connected-icon-label`).
And as you said
> most GLA components are heavily based on WP and WC components

So "This solution does not seem to be applicable to solving" others' styles styling elements like `.components-button`.

But the aim here is to encapsulate the styles of `<gla-connected-icon-label>`. And after this PR, for others to affect those, they would need to either use a selector like `*`, or `gla-connected-icon-label` to affect our element, which is unlikely, or they could attach a class attribute to our element. And then it means they do that on purpose and probably mean to do so.

tomalec:
> In addition, the Fast Refresh development mode (`npm run start:hot`) seems to conflict with this approach. Once a code change occurs within a component that uses `defineShadowStylesHost`, subsequent code changes will not make Fast Refresh update the page. Must refresh the page to resume.

I didn't use Fast Refresh. I'll take a look and check what happens there

eason9487:
> So I do not aim to fix the issues you link.

Thanks for clarifying.

I originally thought it was related to those style conflict cases because this PR came up in this slack thread: p1668008895895869/1668000686.809249-slack-C0410KV3YLW

> Also, I'd like to emphasize one thing: this solution is mostly to **specify and encapsulate the styles for** an HTML **Element** (`<div>` or other) **we do create and govern**, not the ones created by external components (`<WooCommerceButton>`) (Unless we start using those in Shadow DOM, but it has more severe implications, that I'd like to avoid on this baby step).
> We are defining styles for a host we created.

After a quick browsing of the .~/components directory, there are probably only a few components completely hosted by GLA:

- `AppTableCardDiv`
- `ContentButtonLayout`
- `VerticalGapLayout`
- `WistiaVideo`
- .~/components/paid-ads/campaign-preview/mockup-*.js
- .~/components/paid-ads/campaign-preview/components
- .~/components/stepper

Considering that
- there are not many cases that can be applied in practice,
- GLA style leak problem has been generally solved, although `.gla-` and `.gla-admin-page` are not complete technical guarantees, they are low cost and actually effective,
- this solution might not be effective for the major style conflict problem,
   > 3\. Others' style affects GLA 

Currently, I think this is probably still an experimental-stage method and a bit far from the actual adoption. Or,  maybe try `WistiaVideo` as a starting point to have a taste of this. (Before solving the problem of Fast Refresh mode, campaign preview is not so suitable, because the development of it requires a lot of style adjustment, so manually refreshing the page would reduce productivity too much.)

eason9487:
Thank you, @tomalec! As always, I appreciate your various technical ideas and attempts at introduction. Since this RFC has stopped discussion and no conclusion has been reached, and there are currently some alternative solutions that can achieve part of the goals, I'm closing it.

If someone happens to find this PR while searching for a solution to reduce the possibility of CSS conflicts, please refer to:
- CSS modules supported natively in `@wordPress/scripts` by prefixing `.module` to the SCSS extension. For example:
   - https://github.com/woocommerce/google-listings-and-ads/blob/3.5.0/js/src/components/help-icon-button/index.module.scss
- Utilize the `gla-admin-page` CSS class attached to each page via the [`withAdminPageShell` component](https://github.com/woocommerce/google-listings-and-ads/blob/3.5.0/js/src/components/withAdminPageShell.js), along with classical and native CSS ways to achieve scoping. For example:
   - https://github.com/woocommerce/google-listings-and-ads/blob/3.5.0/js/src/css/shared/_gutenberg-components.scss#L1-L2

📌 Please note that the above solutions do not guarantee complete CSS encapsulation.

PR #1767, opened 2022-11-14
Remove unused GRPC packages

### LINKED ISSUE #1643: Remove grpc package
The grpc package is a dependency of the `googleads/google-ads-php` package. However we don't use grpc and instead opted for [REST requests](https://developers.google.com/google-ads/api/docs/concepts/call-structure).

The `grpc` package is currently flagged as a false positive in a PHP 8 compatibility check:
```
FILE: ./google-listings-and-ads/vendor/grpc/grpc/src/lib/RpcServer.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
On line 75 | ERROR | Using 'continue' outside of a loop or switch structure is invalid and will throw a fatal error since PHP 7.0
```

It is being used inside a while loop. But it's defined as `while (true) try {` which seems like it's not getting detected correctly. So we shouldn't have to worry about the false report.

However in the light of us trying to keep the plugin/zip file as small as possible, it can be beneficial to remove unused code. So I'm creating this issue to investigate the possibility of removing this dependency.

mikkamp:

This PR removes the unused GRPC composer packages. They are added as dependency of `googleads/google-ads-php`, however since we opted to use REST request we do not use this package. Removing the packages reduces the final plugin zip file.

Closes #1643 

1. Remove and previous copy of the vendors folder and run `composer install`.
2. Confirm that the packages `grpc/grpc` and `google/grpc-gcp` are no longer found in the vendor folder.
3. Send some test requests to the Google Ads API such as load the GLA dashboard screen (make sure we have a MC and Ads account onboarded).
4. Run the unit tests locally and confirm we don't have any warnings/errors.
5. Check all the tests of this PR passed and the bundle size decreased.
image

### Changelog entry
* Tweak - Remove unused GRPC packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,10 @@
 		"wp-coding-standards/wpcs": "^2.3",
 		"yoast/phpunit-polyfills": "^1.0"
 	},
+	"replace" : {
+		"google/grpc-gcp": "*",
+		"grpc/grpc": "*"
+	},
 	"license": "GPL-3.0",
 	"autoload": {
 		"psr-4": {
--- a/bin/cleanup-vendor-files.sh
+++ b/bin/cleanup-vendor-files.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
 
 echo Removing unused vendor files to reduce space
-rm vendor/google/grpc-gcp/cloudprober/bins/opt/grpc_php_plugin || true
 rm vendor/symfony/validator/Resources/translations/*.xlf || true
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c870e5441a3f1fbae10f20fc5abc35c6",
+    "content-hash": "d3a8b2414bdede67a099175f82c3b62a",
     "packages": [
         {
             "name": "automattic/jetpack-a8c-mc-stats",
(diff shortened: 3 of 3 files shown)


composer.json
```diff
@@ -29,6 +29,10 @@
 		"wp-coding-standards/wpcs": "^2.3",
 		"yoast/phpunit-polyfills": "^1.0"
 	},
+	"replace" : {
+		"google/grpc-gcp": "*",
+		"grpc/grpc": "*"
+	},
```
puntope, line 35: ❓  Wondering why are we using this syntax instead of just adding the `rm -rf` command into [cleanup-vendor-files.sh](https://github.com/woocommerce/google-listings-and-ads/blob/develop/bin/cleanup-vendor-files.sh)

IMO I think it's better to maintain this cleanups in the sh file to have a single source of truth. Plus the fact that replace `*`doesn't seems very declarative for me. I needed to research in order to understand that '*' will make the package version to match "everything" and hence composer won't install the package.

In any case... if we are using this strategy... then we can get rid of the Line 4 or an output will happen on composer install

```
rm: vendor/google/grpc-gcp/cloudprober/bins/opt/grpc_php_plugin: No such file or directory
```

puntope (changes requested):
One of this 2 actions should be done before approval:

a. Remove L4 from [cleanup-vendor-files.sh](https://github.com/woocommerce/google-listings-and-ads/blob/develop/bin/cleanup-vendor-files.sh)

b. Remove the vendor files inside [leanup-vendor-files.sh](https://github.com/woocommerce/google-listings-and-ads/blob/develop/bin/cleanup-vendor-files.sh) instead of using `replace: *` strategy.

puntope:
Thanks @mikkamp for working on this.

I left a comment in regards one issue I think needs to be addressed.

composer.json
mikkamp: Great catch on the extra line to remove.

I personally think preventing a package from being installed is better than installing and removing. This method of replacing the package is something [we were doing before](https://github.com/woocommerce/google-listings-and-ads/blob/0d68e01fdf8acfa8d3faeccc627b46e82d867f1e/composer.json#L30), so it's not a new concept. We can add a note in the composer.json file if it's helpful to understand what it does.

mikkamp:
Thanks for the review @puntope 
I added [a comment](https://github.com/woocommerce/google-listings-and-ads/pull/1767#discussion_r1025000260) about why I think it's best to prevent the packages from being installed. I tried adding a comment in `composer.json` but I got the following error:

> Comments are not permitted in JSON.

So I decided to leave it out, it should be pretty easy to find the attached commit/PR where the explanation can be found, in case someone wants to understand what this code is doing.

I've also added another commit to remove the line from `cleanup-vendor-files.sh`.

It's ready for another round.

puntope (approved):
✅  LGTM 

I tested locally

PR #1768, opened 2022-11-14
Fix Yoast global identifiers for variable products

### LINKED ISSUE #1763: Missing gtin in woo variable products
### Describe the bug:
Hi, in admin under Marketing -> Google Listings & Ads -> Product Feed I’m getting lots of Product Issues with the message “Limited performance due to missing identifiers [gtin, mpn, brand]”. Same list in Google Merchant center.

The problem is it can’t find the gtin13. I’m only getting this with variable products. In each of the affected variable products, I’ve the Google Listings & Ads GTIN setting set to “From Yoast WooCommerce SEO”, which I have defined in the variable product settings under “Yoast SEO options” as GTIN13/EAN.

This works fine with simple products, but apparently it doesn’t pick it up for variable products. It used to work.

I use the latest WP/WOO and plugin versions with PHP 8.1.

### Steps to reproduce:
1. Create product with variations
2. Add GTIN13 in variation in the Yoast field
3. Set GTIN in the variation Google field to “From Yoast WooCommerce SEO” (see screenshot below)

### Expected behavior:
GTIN of variable products to be fed to Google

### Actual behavior:
Google Merchant center gives errors that GTIN of variable products isn't fed to them

### Additional details:
Bug should be somewhere here: src/Integration/YoastWooCommerceSeo.php
I reported this bug earlier here: https://wordpress.org/support/topic/missing-gtin-in-woo-variable-products/
But I met no intelligence.
Yoast also investigated the problem, they said the problem is at your end as I suspected.

Product issues:
gla1

Entered the GTIN in Yoast, selected "From Yoast WooCommerce SEO" in GL&A:
gla2

mikkamp:

This PR fixes the global identifiers for variable products. While it's possible they are stored in the parent product, it's also possible to set the identifiers per variation. This PR takes the latter part into consideration and will fallback to the identifiers stored in the variable product if they are not set.

Closes #1763

1. Install the Yoast SEO plugin: https://wordpress.org/plugins/wordpress-seo/
2. Install the Yoast SEO: WooCommerce plugin: https://github.com/Yoast/wpseo-woocommerce/releases/tag/15.3
3. Edit a variable product and under the variation data set both the Yoast identifier values and the GLA attributes to be loaded from Yoast.
image
4. Save the product (will trigger a scheduled sync).
5. Wait till the sync completes (and some time for the product to update in Google).
6. Confirm the identifier values have been synced.
image

### Changelog entry
* Fix - Yoast global identifiers for variable products.

### DIFF
--- a/src/Integration/YoastWooCommerceSeo.php
+++ b/src/Integration/YoastWooCommerceSeo.php
@@ -151,7 +151,7 @@ class YoastWooCommerceSeo implements IntegrationInterface {
 		if ( $product instanceof WC_Product_Variation ) {
 			$identifiers = $product->get_meta( 'wpseo_variation_global_identifiers_values', true );
 
-			if ( empty( $identifiers ) ) {
+			if ( ! is_array( $identifiers ) || empty( array_filter( $identifiers ) ) ) {
 				$parent_product = wc_get_product( $product->get_parent_id() );
 				$identifiers    = $this->get_identifier_meta( $parent_product );
 			}
--- a/src/Integration/YoastWooCommerceSeo.php
+++ b/src/Integration/YoastWooCommerceSeo.php
@@ -116,6 +116,8 @@ class YoastWooCommerceSeo implements IntegrationInterface {
 	}
 
 	/**
+	 * Get the identifier value from cache or product meta.
+	 *
 	 * @param string     $key
 	 * @param WC_Product $product
 	 *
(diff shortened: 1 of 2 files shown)


src/Integration/YoastWooCommerceSeo.php
```diff
@@ -125,11 +127,38 @@ protected function get_identifier_value( string $key, WC_Product $product ) {
...
+			return null;
+		}
+
+		if ( $product instanceof WC_Product_Variation ) {
+			$identifiers = $product->get_meta( 'wpseo_variation_global_identifiers_values', true );
+
+			if ( empty( $identifiers ) ) {
+				$parent_product = wc_get_product( $product->get_parent_id() );
+				$identifiers    = $this->get_identifier_meta( $parent_product );
+			}
```
jorgemd24, line 157: ❔

Can the user set identifiers for the parent product? I see only the option in the variations tabs.

src/Integration/YoastWooCommerceSeo.php
```diff
@@ -125,11 +127,38 @@ protected function get_identifier_value( string $key, WC_Product $product ) {
...
+	 */
+	protected function get_identifier_meta( WC_Product $product ) {
+		if ( ! $product ) {
+			return null;
+		}
+
+		if ( $product instanceof WC_Product_Variation ) {
+			$identifiers = $product->get_meta( 'wpseo_variation_global_identifiers_values', true );
+
+			if ( empty( $identifiers ) ) {
```
jorgemd24: `$identifiers` will never be empty as every time that the variation/product is created Yoast is adding the meta `wpseo_variation_global_identifiers_values` with empty values so `$product->get_meta( 'wpseo_variation_global_identifiers_values', true );` will return an array with the format of `['gtin18' => '', 'gtin13' => '' ....]`.

We should check if all values are empty so we can use the parent's identifier.  

Suggested change:

```
if ( empty( array_filter( $identifiers ) ) ) {
 .....
}
```

jorgemd24 (changes requested):

src/Integration/YoastWooCommerceSeo.php
mikkamp: Thanks for the feedback, that makes sense. Although in cases where the variations haven't been re-saved after activating the extension $identifiers could be null. So we'll probably need a combination of both checks or some type casting otherwise array_filter will fail if we don't pass it an array.

mikkamp:
Thanks for the review @jorgemd24. I added your suggestion to filter the array for empty values. Ready for another round.

> **Note**
Because of the way the caching is done (full meta array is cached), if only the MPN number is set for the variation, it won't be able to fallback to the parent value for GTIN. However I think that's acceptable behaviour as the merchant can get around that by setting both the GTIN and the MPN in the variation data (and is most likely an edge case).

jorgemd24 (approved):
Thanks, @mikkamp  for the adjustments. LGTM 👍

PR #1769, opened 2022-11-15
Attribute Mapping - UX and Text tweaks

puntope:

Closes part of #1648 .

This PR implements minor changes in the texts and also applies the last changes in develop fixing UX issues in the Popover. Also, we removed the links for now since they are not available yet. (this was requested by UX team)

Check figma https://www.figma.com/file/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=2581%3A178755&t=zCwUFuRdEgokgzj7-0 and verify texts are the same.

### DIFF
--- a/js/src/components/help-popover/index.js
+++ b/js/src/components/help-popover/index.js
@@ -24,7 +24,7 @@ import './index.scss';
  * @param {Array<JSX.Element>} props.children The Popover’s content
  * @fires gla_tooltip_viewed with the given `id`.
  */
-const HelpPopover = ( { id, children } ) => {
+const HelpPopover = ( { id, children, ...props } ) => {
 	const [ showPopover, setShowPopover ] = useState( false );
 
 	const handleButtonClick = () => {
@@ -47,6 +47,7 @@ const HelpPopover = ( { id, children } ) => {
 				<Popover
 					focusOnMount="container"
 					onClose={ handlePopoverClose }
+					{ ...props }
 				>
 					{ children }
 				</Popover>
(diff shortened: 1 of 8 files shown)


js/src/components/help-popover/index.js
```diff
@@ -47,6 +47,7 @@ const HelpPopover = ( { id, children } ) => {
 				<Popover
 					focusOnMount="container"
 					onClose={ handlePopoverClose }
+					{ ...props }
```
eason9487, line 50: It's nice to forward props to `Popover` so that we can fully control it. And we could also handle `onClick` in [the same way in `AppButton`](https://github.com/woocommerce/google-listings-and-ads/blob/ef2ed8bd4c4ea721401026a53468eacfaba23bb0/js/src/components/app-button/index.js#L40-L50), thus, the following use would work smoothly.
```js
<HelpPopover
	onClose={ () => { /* do something */ } }
/>
```

eason9487 (approved):
Thanks for the adjustment. Left a comment about the flexibility of `HelpPopover` but it's minor. LGTM.

js/src/components/help-popover/index.js
puntope: Thanks for the review @eason9487 

I'm not sure if I fully understand this. I'm passing actually the `...props` to the Popover, hence you can define the props you want.

I guess you suggested to add `onClick` and `onClose` callbacks in the `handleButtonClick` and `handlePopoverClose`,  but not 100% sure.

eason9487: > And we could also handle `onClick` in [...]

Sorry for my typo. 🙏  It should be `onClose` and I meant this processing:

```js
const HelpPopover = ( { id, onClose = noop, children, ...props } ) => {
	const [ showPopover, setShowPopover ] = useState( false );

	// ... omitted

	const handlePopoverClose = () => {
		setShowPopover( false );
		onClose();
	};

	// ... omitted
};
```

PR #1773, opened 2022-11-15
Remove custom implementation of `<input type="number">`

### LINKED ISSUE #663: Use number input field for positive integer
### Describe the issue:
Follow up from https://github.com/woocommerce/google-listings-and-ads/pull/631#discussion_r638580539

The multipack attribute uses an integer as a form input value. This attribute indicates when you are selling multiple items in one, for example a set of 4 pens. This means the input value should always be a positive integer.

However for the input field we use `<input type="text">`. In combination with that we use some jQuery code to trigger the function `replaceNonInteger` on `change` and `keyup` events. This means any character we enter in the field that is not a `0-9` is deleted. Personally I think this gives an odd effect as we can see the character entered and then quickly deleted afterwards.

In my opinion it would make sense to switch this to `<input type="number" min="0">` to let the browser handle the input validation. I'm not sure if we need to enforce all `integer` input fields to be strictly positive, at the moment we only have one attribute that needs to be both an integer and positive.

### Steps to reproduce:
1. Edit a simple product and open the GLA tab in Product Data
2. Type some random text into the Multipack field, and notice how it only allows numeric characters to remain

tomalec:

Reduce usage of jQuery, use plain HTML instead.
Improve UX and accessibility, by allowing arrow up/down changes, and enabling assistive tools integration.

Fixes: https://github.com/woocommerce/google-listings-and-ads/issues/663

### Screenshots:
image

1.  Edit or add a product `/wp-admin/post.php?post=37&action=edit`
2. Open `Google Listings and Ads` Product Data tab
3. Change the value of `multipack` attribute.
4. Try putting some non-number characters or negative values.

### Changelog entry

> Update - Change multipack attribute input to be native number input, to improve accessibility.


mikkamp:
Love the simplification. I'm not sure if it was inspired by this issue but we can link it to #663 to make sure it's closed once we merge this.

tomalec:
> I'm not sure if it was inspired by this issue but we can link it to https://github.com/woocommerce/google-listings-and-ads/issues/663

Oh, thanks for mentioning that. I totally forgot about it :)

It was actually inspired by your comment about dropping jQuery, and my curiosity about where I could sneak an improvement based on Web Components without conflicting with existing env ;)  Plus, my brainless day spent looking for some easy thing do do and where to bring more tranquility to our code.

views/inputs/integer.php
```diff
@@ -12,7 +12,10 @@
  */
 $input = $this->input;
 
-$input['type']  = 'text';
-$input['class'] = sprintf( '%s %s', $input['class'] ?? '', 'gla-input-integer' );
+$input['type'] = 'number';
+// Not so "custom" but standard `<input type="numer">` attribute.
```
mikkamp: Just a tiny typo here.

```suggestion
// Not so "custom" but standard `<input type="number">` attribute.
```

mikkamp (approved):
Thanks for the simplification, everything seems to be working as expected.

I'm a little uncertain about the re-usability of setting the input element to a positive number in the view, but the attribute itself is only restricted to an integer. Maybe something to leave for the future though, as we don't have any attributes that allow negative numbers, so not necessary to account for all these edge cases at the moment.

tomalec:
I was thinking of removing jQuery from `SelectWithTextInput` as well, but couldn't find where it's used and how to test it.

tomalec:
> I'm a little uncertain about the re-usability of setting the input element to a positive number in the view, but the attribute itself is only restricted to an integer. Maybe something to leave for the future though, as we don't have any attributes that allow negative numbers, so not necessary to account for all these edge cases at the moment.

Also, that's the way it behaved before. But that's definitely something to think of in the future.

PR #1774, opened 2022-11-16
Add a script to list woo/DEWPed versions

tomalec:

[Add a script to list woo/DEWPed versions](https://github.com/woocommerce/google-listings-and-ads/commit/c90a0b4871f81997cadba2ade84efa9a59a7df98)

So we will not have to browse `woocommerce/woocommerce` repo manually every time to check what version given WC uses.

Slightly mitigates tiny part of https://github.com/WordPress/gutenberg/issues/35630

 
[Add some docs about DEWP tricks](https://github.com/woocommerce/google-listings-and-ads/commit/4073e4b904be5ec2607a5082f7faa73b2ca4a153)

### Screenshots:

image

1. Checkout this branch
2. `nvm use 19` (or at least 17)
3. `npm run dewps:woo 6.9.4`

### Additional details:
1. Node 17+ is required to use the clean `fetch` syntax. I hope additional `nvm use 17` is not a big price to pay for cleaner code to maintain. Hopefully, someday we could get there with `.nvmrc` as well.

### Changelog entry

> Dev - Add script to list DEWPed dependencies' versions for a given WC version.


package.json
```diff
@@ -18,6 +18,7 @@
 		"check-engines": "wp-scripts check-engines",
 		"check-licenses": "wp-scripts check-licenses",
 		"dev": "NODE_ENV=development wp-scripts build",
+		"dewps:woo": "node --experimental-fetch bin/list-woo-dewped.mjs",
```
tomalec: `--experimental-fetch` is required for Node 17, for 18+ we would not need that. But as we are far behind with our 14 and WooCommerce's 16, I'd stick to the lowest possible.

Working with DEWP.md
```diff
@@ -0,0 +1,55 @@
...
+
+In a regular project to track outdated packages, you would use `npm outdated`. However, with DEWP updating a bundled package is not the same as updating externalized one. For example, if we aim to support WC L-2, we theoretically need to test our extensions with all the package versions used by that version, and we cannot update it to the latest as we go. So we have two scripts to separate those cases:
+
+- `npm run outdated:dewp` Check for outdated packages that are listed in `.externalized.json`
+- `npm run outdated:nondewp` Check for those not listed there.
+
+### What is the version of the DEWPed package?
+
+As we externalize packages we lose control of the version of the package that will run in the wild (on WordPress instance).
+However, with the current platform, we not only loos control, but we also lose the information of what is actually the range of package we should expect.
```
mikkamp:
```suggestion
However, with the current platform, we not only lose control, but we also lose the actual information on the range of package we should expect.
```

Working with DEWP.md
```diff
@@ -0,0 +1,55 @@
...
+So, even if we know we support WooCommerce 6.9+, what are the `@woocommerce/components` there?
+
+There is no published table of those, and finding them usually requires quite a lot of manual digging. To mitigate that we created another script.
+
+- `npm run dewps:woo 6.9.4` - where `6.9.4` is the version of WooCommerce you would like to check.
+
+Please note this simple script still have several limitations.
```
mikkamp:
```suggestion
Please note this simple script still has several limitations.
```

bin/list-woo-dewped.mjs
```diff
@@ -0,0 +1,54 @@
...
+'use strict';
+import * as fs from 'node:fs';
+
+const externalizedListPath = '.externalized.json';
+const wooPackage = ( tag, packageName ) =>
+	`https://raw.githubusercontent.com/woocommerce/woocommerce/${ tag }/packages/js/${ packageName }/package.json`;
+
+const myArgs = process.argv.slice( 2 );
+const wcTag = myArgs[ 0 ];
+if ( ! wcTag || ! wcTag.match( /(\d+).(\d+).(\d+).*/ ) ) {
```
mikkamp: I guess since this just prints a warning it's not really a blocker. But to better match the WC versions we use the regex should start with a `$`. Otherwise `foobar1.2.3` is also a valid tag.

package.json
mikkamp: If we need to `nvm use <version>` any ways before running this command then I think it would be nicer to leave the `--experimental-fetch` parameter out and require 18+.

Reasoning for this is that if you still have the node version from `.nvmrc` installed it won't run because the parameter `--experimental-fetch` is not supported. But if we leave the parameter out then you can at least run the script. Which means we can do a quick compare at the beginning of the script and check if `process.versions.node` is smaller than 18 and if so just output a warning to use a newer node version and exit. My opinion is that it makes it slightly more friendly.

mikkamp (approved):
Thanks for adding the helpful utility and all the documentation around it. After reading those docs I feel I have a better understanding of how it's working.

The script runs well and I get the results as expected. I left a few minor comments and some corrections for the wording. But it looks pretty much good to go, so I'll approve it.

package.json
tomalec: Fair point. 

Initially, I thought that given `woocommerce/woocommerce` is already at node 16, then we could get to 17 relatively sooner than to 18. Then, if our `.nvmrc` would point to 17, it won't be a problem any longer.

Interestingly, despite the Node changelog https://nodejs.org/uk/blog/release/v17.5.0/#add-fetch-api, I got the script running successfully on Node 16. Which is the one used by WC and AW, so maybe we can bump it here as well.

tomalec: :disappointed: But we cannot update Node to 16 = version Woo uses, as `@woocommerce/e2e-environment` does not work with it. https://github.com/woocommerce/woocommerce/issues/35650

PR #1777, opened 2022-11-17
Attribute Mapping - Add tracking

puntope:

Closes part of #1648 .

This PR adds the tracking for the feature:

`gla_modal_closed`
`gla_modal_open`
`gla_attribute_mapping_save_rule_click`
`gla_attribute_mapping_delete_rule_click`
`gla_attribute_mapping_create_rule`
`gla_attribute_mapping_update_rule`
`gla_attribute_mapping_delete_rule`

1. Enable event tracking logging by running `localStorage.setItem( 'debug', 'wc-admin:*' )` in the Console tab of DevTool, and refresh page to make it effective.
2. Go to Attribute Mapping page and click on Create rule button. It should trigger `gla_modal_open` with the context `attribute-mapping-create-rule-modal`.
3. Close the modal using cancel button, X button and clicking outside. It should trigger the event `gla_modal_closed` with the context `attribute-mapping-create-rule-modal` and the action `dismiss`
4.  Click again on Create rule button and click on save rule (after filling the form). It should trigger  `gla_attribute_mapping_save_rule_click` with context  `attribute-mapping-create-rule-modal` and also `gla_attribute_mapping_create_rule` with the same context and the rule props.
5. Click on Manage on the rule you just created. It should trigger  `gla_modal_open` with context  `attribute-mapping-manage-rule-modal`
6. Close the modal using cancel button, X button and clicking outside. It should trigger the event `gla_modal_closed` with the context `attribute-mapping-manage-rule-modal` and the action `dismiss`
7. Click again on Manage on the rule you just created and click on Save rule (after modifying the rule). It should trigger  `gla_attribute_mapping_save_rule_click` with context  `attribute-mapping-manage-rule-modal` and also `gla_attribute_mapping_update_rule` with the same context and the rule props.
8. Click on Delete on the rule you just created. It should trigger  `gla_modal_open` with context  `attribute-mapping-delete-rule-modal`
9. Close the modal using cancel button, X button and clicking outside. It should trigger the event `gla_modal_closed` with the context `attribute-mapping-delete-rule-modal` and the action `dismiss`
10. 8. Click again on Delete on the rule you just created. Confirm deletion. It should trigger  `gla_attribute_mapping_delete_rule_click` with context  `attribute-mapping-delete-rule-modal` and also `gla_attribute_mapping_delete_rule` with the same context and the rule props.

### Additional details:

Some  errors happened when I run `npm run doc:tracking`. Most of them are fixable by changing JSDOC definition but the one at the end:

```
google-listings-and-ads/node_modules/jsdoc-advanced-types-plugin/index.js:105
				throw "infinite loop!"
```

Requires more investigation and might be solved in future PRs

### DIFF
--- a/js/src/attribute-mapping/attribute-mapping-delete-rule-modal.js
+++ b/js/src/attribute-mapping/attribute-mapping-delete-rule-modal.js
@@ -4,6 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { noop } from 'lodash';
 import { useState } from '@wordpress/element';
+import { recordEvent } from '@woocommerce/tracks';
 
 /**
  * Internal dependencies
@@ -12,6 +13,34 @@ import AppModal from '.~/components/app-modal';
 import AppButton from '.~/components/app-button';
 import { useAppDispatch } from '.~/data';
 
+/**
+ * Deletes the rule successfully
+ *
+ * @event gla_attribute_mapping_delete_rule
+ * @property {string} context Indicates where this event happened
+ * @property {string} id Rule id
+ * @property {string} attribute Rule attribute
+ * @property {string} source Rule source
+ * @property {string} category_condition_type Rule category condition type
+ * @property {string} categories Rule categories
+ */
+
+/**
+ * Clicks on delete rule button
+ *
+ * @event gla_attribute_mapping_delete_rule_click
+ * @property {string} context Indicates where this event happened
+ */
+
+/**
+ * Renders a Modal to confirm Attribute Mapping Rule deletion
+ *
+ * @param {Object} props Component props
+ * @param {Function} props.onRequestClose Callback function when the modal is closed
+ * @param {Object} props.rule The rule to be deleted
+ * @fires gla_attribute_mapping_delete_rule When the rule is successfully deleted
+ * @fires gla_attribute_mapping_delete_rule_click When user clicks on delete rule button
+ */
 const AttributeMappingDeleteRuleModal = ( { onRequestClose = noop, rule } ) => {
 	const [ deleting, setDeleting ] = useState( false );
 	const { deleteMappingRule } = useAppDispatch();
@@ -21,7 +50,11 @@ const AttributeMappingDeleteRuleModal = ( { onRequestClose = noop, rule } ) => {
 
 		try {
 			await deleteMappingRule( rule );
-			onRequestClose();
+			recordEvent( 'gla_attribute_mapping_delete_rule', {
+				context: 'attribute-mapping-delete-rule-modal',
+				...rule,
+			} );
+			onRequestClose( 'confirm' );
 		} catch ( error ) {
 			setDeleting( false );
 		}
@@ -29,7 +62,7 @@ const AttributeMappingDeleteRuleModal = ( { onRequestClose = noop, rule } ) => {
 
 	const handleClose = () => {
 		if ( deleting ) return;
-		onRequestClose();
+		onRequestClose( 'dismiss' );
 	};
 
 	return (
@@ -57,9 +90,9 @@ const AttributeMappingDeleteRuleModal = ( { onRequestClose = noop, rule } ) => {
 									'google-listings-and-ads'
 							  )
 					}
-					eventName="gla_attribute_mapping_delete_rule"
+					eventName="gla_attribute_mapping_delete_rule_click"
 					eventProps={ {
-						context: 'attribute-mapping-rule-modal',
+						context: 'attribute-mapping-delete-rule-modal',
 					} }
 					onClick={ onDelete }
 				/>,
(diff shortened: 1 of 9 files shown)


puntope:
In regard the error for infinite loop in JSDOC tracking I created an issue

https://github.com/woocommerce/grow/issues/43

Also, I updated manually the `npm_modules` dependency in order to generate the updated README.md for this PR.

js/src/attribute-mapping/attribute-mapping-table.js
```diff
@@ -52,6 +53,8 @@ const ATTRIBUTE_MAPPING_TABLE_HEADERS = [
 /**
  * Renders the Attribute Mapping table component
  *
+ * @fires gla_modal_closed When any of the modals is closed
+ * @fires gla_modal_open When any of the modals is open with `context: attribute-mapping-manage-rule-modal`
```
eason9487: Two possible context values `attribute-mapping-create-rule-modal` and `attribute-mapping-delete-rule-modal` seem to be missed.

js/src/components/adaptive-form/adaptive-form.js
```diff
@@ -27,7 +27,7 @@ function isEvent( value ) {
  * several workarounds in order to be compatible with WC 6.9 to 7.1.
  *
  * @param {Object} props React props.
- * @param {JSX.Element | (formProps: Object) => JSX.Element} props.children Children to be rendered. Could be a render prop function.
+ * @param {JSX.Element | Function} props.children Children to be rendered. Could be a render prop function.
```
eason9487: Swapping the order could make it parsable as well.

```suggestion
 * @param {(formProps: Object) => JSX.Element | JSX.Element} props.children Children to be rendered. Could be a render prop function.
```

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -111,10 +145,18 @@ const AttributeMappingRuleModal = ( { rule, onRequestClose = noop } ) => {
...
 				await updateMappingRule( newRule );
+				recordEvent( 'gla_attribute_mapping_update_rule', {
+					context: 'attribute-mapping-manage-rule-modal',
+					...newRule,
+				} );
 			} else {
 				await createMappingRule( newRule );
+				recordEvent( 'gla_attribute_mapping_create_rule', {
+					context: 'attribute-mapping-create-rule-modal',
+					...newRule,
```
eason9487: Do we intend to record the rule data? Not sure if it would violate some Terms. In addition, the `id` and the content of `categories` are only numbers representing their IDs in DB tables, so meaningful information may not be available on the aggregation or statistics gathered from event tracking.

image

js/src/attribute-mapping/attribute-mapping-rule-modal.js
```diff
@@ -34,8 +65,8 @@ const attributeSelectorLabel = __(
 /**
  * Map the format received from the backend into the format needed in the SelectControl
  *
- * @param {{id: string, label: string}[]} data The array with the values from the backend
- * @return {{label: string, value: string}[]} The data formatted
+ * @param {Array} data The array with the values from the backend
+ * @return {Array} The data formatted
```
eason9487:
```js
/**
 * @param {Array<{id: string, label: string}>}
 * @param {Array<{label: string, value: string}>}
 */
```
This format could keep the object structure and be a bit nicer to have clearer JSDoc.

eason9487 (commented):
Testing well. However, I have some doubts about recording the content of attribute rule. Other comments are minor, but it would be a bit better if they could be adjusted.

> In regard the error for infinite loop in JSDOC tracking I created an issue [...]

I believe it was caused by the following `@return` tag as their types are missed.

https://github.com/woocommerce/google-listings-and-ads/blob/27763a1604b83fd5394efa91a576ac1dead46a9f/js/src/hooks/useCategories.js#L114

js/src/attribute-mapping/attribute-mapping-delete-rule-modal.js
```diff
@@ -21,15 +50,19 @@ const AttributeMappingDeleteRuleModal = ( { onRequestClose = noop, rule } ) => {
 
 		try {
 			await deleteMappingRule( rule );
-			onRequestClose();
+			recordEvent( 'gla_attribute_mapping_delete_rule', {
+				context: 'attribute-mapping-delete-rule-modal',
+				...rule,
+			} );
```
eason9487: Same doubt as https://github.com/woocommerce/google-listings-and-ads/pull/1777#discussion_r1026220206.

puntope:
Howdy @eason9487 

Thanks for the review, I applied all your suggestions. This is ready for another round.

BTW; Notice I modified one extra PHP file, since the types returned where wrong. I noticed it when your suggestion in https://github.com/woocommerce/google-listings-and-ads/pull/1777#pullrequestreview-1185747125

eason9487 (approved):
Thanks for the adjustment! LGTM.

The only minor thing is we could run `npm run doc:tracking` again to make the *src/Tracking/README.md* file align with the latest changes in this PR.

puntope:
Updated! Thanks @eason9487

PR #1778, opened 2022-11-17
Attribute Mapping - Add Inbox Notification

puntope:

Closes part of #1648 .

This PR add the inbox notification informing existing users about the new feature.

### Screenshots:

Screenshot 2022-11-17 at 12 36 08

1. Be sure `gla_install_timestamp` in your `wp_options` table is `> 1669766400`
2. Go to WooCommerce->Status->AcctionSchedule->Pending and run the task `wc_gla_cron_daily_notes`
3. Go to WooCommerce->Home. The notification should NOT be added
4. Change `gla_install_timestamp` in your `wp_options` table to `<= 1669766400`
5. Go to WooCommerce->Status->AcctionSchedule->Pending and run the task `wc_gla_cron_daily_notes`
6. Go to WooCommerce->Home. The notification should YES be added
7. Verify the content match [this content](https://docs.google.com/document/d/1hVV-mFeSD4qsnIv-G6vFJ6zdYDMRZLiagquGqY_gpqg/edit). 
8. Click on Learn more. It should go to Attribute Mapping feature
9.  Go to WooCommerce->Status->AcctionSchedule->Pending and run the task `wc_gla_cron_daily_notes`
10. Go to WooCommerce->Home. The notification should NOT be added


src/Notes/AttributeMappingNewFeature.php
```diff
@@ -0,0 +1,72 @@
...
+				'google-listings-and-ads'
+			)
+		);
+
+		$note->set_name( $this->get_name() );
+		$note->set_source( $this->get_slug() );
+		$note->add_action(
+			'learn-more-attribute-mapping',
+			__( 'Learn more', 'google-listings-and-ads' ),
+			'admin.php?page=wc-admin&path=/google/attribute-mapping'
```
jorgemd24: WDYT if we add the url in the PluginHelper class, similar to: 

https://github.com/woocommerce/google-listings-and-ads/blob/e121e5593a66be087e83d8a95de139c9b34a5fe5/src/PluginHelper.php#L127-L129

src/Notes/AttributeMappingNewFeature.php
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+
+	/**
+	 * Checks if a note should be added and the user is an existing user.
+	 *
+	 * @return bool
+	 */
+	public function should_be_added(): bool {
+		return ! $this->has_been_added() && $this->options->get( OptionsInterface::FILE_VERSION ) <= '2.2.1';
```
jorgemd24: `$this->options->get( OptionsInterface::FILE_VERSION )` could be replaced with `$this->get_version()`. See: https://github.com/woocommerce/google-listings-and-ads/blob/e121e5593a66be087e83d8a95de139c9b34a5fe5/src/PluginHelper.php#L68-L70

jorgemd24: WDTY if we compare versions using `version_compare` function? should it not be GLA version > '2.2.1' and then we show that notification? Or is this feature available in 2.2.1 and previous versions?

jorgemd24 (commented):
Thanks, @puntope for adding this. I left some comments about the code. 

Also, it is not clear to me why we want to show the notification for versions <= 2.2.1 if the feature is not enabled yet.

src/Notes/AttributeMappingNewFeature.php
puntope: You're right. 

At the end, I'm using INSTALLED_TIMESTAMP to check if the user is new or existing (comparing the timestamp with 30 November)

src/Notes/AttributeMappingNewFeature.php
puntope: Added in https://github.com/woocommerce/google-listings-and-ads/pull/1778/commits/34aaf7621b77aa5452eaa80fb272341cf0445aa7

puntope:
Howdy @jorgemd24 

I added your suggestions.

In regards version, you were right. Is not the best way to check if it's an existing user. hence I added the logic based on INSTALL_TIMESTAMP 

Thanks

jorgemd24 (commented):
Thanks, @puntope for the changes, the code looks good. I have just some doubts about what we can consider "existing" users. 

For instance, if a new user installs the plugin **for the first time** on  23/11 with the version that includes the attribute mapping feature, the user will be treated as an "**existing**" user, right? Even though technically it is a newcomer. If that's fine I think we can go ahead with this PR otherwise maybe you could add a flag indicating if it is an existing user. For example here:

https://github.com/woocommerce/google-listings-and-ads/blob/c04d28d8ac973f984de5ca228920b0bc7c38e028/src/Installer.php#L85-L89

puntope:
> For instance, if a new user installs the plugin for the first time on 23/11 with the version that includes the attribute mapping feature, the user will be treated as an "existing" user, right? Even though technically it is a newcomer. If that's fine I think we can go ahead with this PR 

Yes, I was considering existing user those after 30th November in regards Attribute mapping Feature and as that was the initial deadline. I think thats ok since 30th is there he date in when we will announce it in WWIW. It's not a big issue to receive the notification inbox for those users before 30th, but yes is more accurate to use 23th, so I modified it.

 > otherwise maybe you could add a flag indicating if it is an existing user. For example here:

The problem with that approach I guess is that after the install the user will be treated as "existing"  so if the inbox notification runs after that, the user will get the notification even if its new user. With timestamp approach we can cover as well those cases.

puntope:
@jorgemd24 I adjusted the timestamp to 23/11/2022 so for tomorrow shipping. Let me know i now this is ready to go

jorgemd24 (approved):
Thanks @puntope LGTM 👍

Issue #1760, opened 2022-11-10
Please vendor prefix Google-related composer dependencies


ecaron:
### Describe the bug:
If other plugins use Google SDKs, there's a namespace conflict that happens where other plugins using (for example) [apiclient](https://packagist.org/packages/google/apiclient) and are expecting `GuzzleHttpClientInterface` but receive `AutomatticWooCommerceGoogleListingsAndAdsVendorGuzzleHttpClient`.

A detailed example of the error condition:
`Fatal error: Uncaught TypeError: GoogleClient::setHttpClient(): Argument #1 ($http) must be of type GuzzleHttpClientInterface, AutomatticWooCommerceGoogleListingsAndAdsVendorGuzzleHttpClient given in /code/wp-content/plugins/customer-plugin/vendor/google/apiclient/src/Client.php:1152 Stack trace: #0 [internal function]: GoogleClient->setHttpClient(Object(AutomatticWooCommerceGoogleListingsAndAdsVendorGuzzleHttpClient)) #1 /code/wp-content/plugins/google-listings-and-ads/vendor/league/container/src/Definition/Definition.php(273): call_user_func_array(Array, Array) #2 /code/wp-content/plugins/google-listings-and-ads/vendor/league/container/src/Definition/Definition.php(216): AutomatticWooCommerceGoogleListingsAndAdsVendorLeagueContainerDefinitionDefinition->invokeMethods(Object(GoogleClient))`

Tickets #314 and #1643 might be impacted by this.

### Steps to reproduce:
1. Install another plugin using stock [apiclient](https://packagist.org/packages/google/apiclient)
2. Install Google Listings and Ads

### Expected behavior:
Both plugins work.

### Actual behavior:
Site crashes because of type errors due to inconsistent SDK behavior between the installed plugins.

### Additional details:
Other plugins that have similar "add vendor prefixing" tickets:
* [WooCommerce](https://github.com/woocommerce/woocommerce/pull/27395)
* [Yoast](https://github.com/Yoast/wordpress-seo/issues/15522)

mikkamp:
Thanks for the report @ecaron 

Is it just the `apiclient` library that is being used by the plugin? Or are there other Google packages that are conflicting as well? We do already add a vendor prefix to a select list of composer packages, however `apiclient` is not on that list. So it would be helpful to know a bit more about the plugin to decide which additional packages to include.

[...]

PR #1783, opened 2022-11-18
Composer `install-scripts` are now OS-indifferent, path safe, and mention modified files

ecaron:

For occasions where `composer install` is run on a non-*NIX operating system, or in a path that included a " ", the install had been failing. This modifies the calls to happen entirely within PHP, so it works on any OS and within any path.

The script also tweaks `prefix-vendor-namespace.php` to mention files that have their class dependencies altered, but retain their original namespace (with intent of ensuring future package changes to introduce new namespace conflicts similar to #1760)

### Changelog entry

> Dev - Adjusted parts of the post-install process to work on machines without `grep` and `find`. 
> Dev - Adjusted post-install process to mention when files have their class-expectations modified but retain their original namespace.

### DIFF
--- a/bin/GoogleAdsCleanupServices.php
+++ b/bin/GoogleAdsCleanupServices.php
@@ -193,15 +193,17 @@ class GoogleAdsCleanupServices {
 		try {
 			$rdi = new RecursiveDirectoryIterator( $path );
 		} catch ( UnexpectedValueException $e ) {
-			$this->output_text( sprintf(
-				'Expected directory "%s" was not found',
-				$path
-			) );
+			$this->output_text(
+				sprintf(
+					'Expected directory "%s" was not found',
+					$path
+				)
+			);
 			exit( 1 );
 		}
 
-		$rii = new RecursiveIteratorIterator( $rdi );
-		$rri = new RegexIterator( $rii, $match );
+		$rii   = new RecursiveIteratorIterator( $rdi );
+		$rri   = new RegexIterator( $rii, $match );
 		$files = [];
 		foreach ( $rri as $file ) {
 			$files[] = $file->getPathname();
@@ -219,12 +221,12 @@ class GoogleAdsCleanupServices {
 	 * @return array List of names that match.
 	 */
 	protected function find_used_pattern( string $pattern ): array {
-		$files  = $this->get_dir_contents( $this->code_path,  '/\.php$/i');
+		$files  = $this->get_dir_contents( $this->code_path, '/\.php$/i' );
 		$output = [];
-		foreach ( $files as $file) {
-			preg_match_all( '/' . $pattern . '/', file_get_contents( $file ), $matches); // phpcs:ignore WordPress.WP.AlternativeFunctions
+		foreach ( $files as $file ) {
+			preg_match_all( "/{$pattern}/", file_get_contents( $file ), $matches ); // phpcs:ignore WordPress.WP.AlternativeFunctions
 			if ( isset( $matches[1] ) ) {
-				foreach ($matches[1] AS $match) {
+				foreach ( $matches[1] as $match ) {
 					$output[] = $match;
 				}
 			}
@@ -234,7 +236,7 @@ class GoogleAdsCleanupServices {
 			return [];
 		}
 
-		return array_unique($output);
+		return array_unique( $output );
 	}
 
 	/**
--- a/bin/GoogleAdsCleanupServices.php
+++ b/bin/GoogleAdsCleanupServices.php
@@ -179,6 +179,35 @@ class GoogleAdsCleanupServices {
 	}
 
 	/**
+	 * Find a list of files in a path matching a pattern.
+	 *
+	 * @param string $path Package path
+	 * @param string $match Regex pattern to match
+	 * @return array Matching files
+	 */
+	protected function get_dir_contents($path, $match) {
+		try {
+			$rdi = new \RecursiveDirectoryIterator($path, \RecursiveDirectoryIterator::SKIP_DOTS);
+		} catch ( Exception  $e ) {
+			printf(
+				'Expected directory "%s" was not found' . PHP_EOL,
+				$path
+			);
+			exit( 1 );
+		}
+
+		$rii = new \RecursiveIteratorIterator($rdi);
+		$rri = new \RegexIterator($rii, $match);
+		$files = [];
+		foreach ($rri as $file) {
+			$files[] = $file->getPathname();
+		}
+
+		return $files;
+	}
+
+
+	/**
 	 * Find a specific pattern used within the extension.
 	 *
 	 * @param string $pattern Regexp pattern to match.
@@ -186,24 +215,22 @@ class GoogleAdsCleanupServices {
 	 * @return array List of names that match.
 	 */
 	protected function find_used_pattern( string $pattern ): array {
-		$command = "grep -rE --include=*.php '{$pattern}' {$this->code_path}";
-
-		// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.system_calls_exec
-		exec( $command, $output );
+		$files  = $this->get_dir_contents( $this->code_path,  '/\.php$/i');
+		$output = [];
+		foreach ( $files AS $file) {
+			preg_match_all( '/' . $pattern . '/', file_get_contents( $file ), $matches);
+			if ( isset( $matches[1] ) ) {
+				foreach ($matches[1] AS $match) {
+					$output[] = $match;
+				}
+			}
+		}
 
 		if ( empty( $output ) ) {
 			return [];
 		}
 
-		return array_unique(
-			array_map(
-				function( $line ) use ( $pattern ) {
-					preg_match( "/{$pattern}/", $line, $matches );
-					return $matches[1];
-				},
-				$output
-			)
-		);
+		return array_unique($output);
 	}
 
 	/**
--- a/bin/GoogleAdsCleanupServices.php
+++ b/bin/GoogleAdsCleanupServices.php
@@ -4,6 +4,10 @@ declare( strict_types=1 );
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Util;
 
 use Composer\Script\Event;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use RegexIterator;
+use UnexpectedValueException;
 
 /**
  * Utilities to remove Google Ads API services in the library.
@@ -179,27 +183,27 @@ class GoogleAdsCleanupServices {
 	}
 
 	/**
-	 * Find a list of files in a path matching a pattern.
+	 * Find a list of files in a path, including subdirectories, matching a pattern.
 	 *
 	 * @param string $path Package path
 	 * @param string $match Regex pattern to match
 	 * @return array Matching files
 	 */
-	protected function get_dir_contents($path, $match) {
+	protected function get_dir_contents( $path, $match ) {
 		try {
-			$rdi = new \RecursiveDirectoryIterator($path, \RecursiveDirectoryIterator::SKIP_DOTS);
-		} catch ( Exception  $e ) {
-			printf(
-				'Expected directory "%s" was not found' . PHP_EOL,
+			$rdi = new RecursiveDirectoryIterator( $path );
+		} catch ( UnexpectedValueException $e ) {
+			$this->output_text( sprintf(
+				'Expected directory "%s" was not found',
 				$path
-			);
+			) );
 			exit( 1 );
 		}
 
-		$rii = new \RecursiveIteratorIterator($rdi);
-		$rri = new \RegexIterator($rii, $match);
+		$rii = new RecursiveIteratorIterator( $rdi );
+		$rri = new RegexIterator( $rii, $match );
 		$files = [];
-		foreach ($rri as $file) {
+		foreach ( $rri as $file ) {
 			$files[] = $file->getPathname();
 		}
 
@@ -217,8 +221,8 @@ class GoogleAdsCleanupServices {
 	protected function find_used_pattern( string $pattern ): array {
 		$files  = $this->get_dir_contents( $this->code_path,  '/\.php$/i');
 		$output = [];
-		foreach ( $files AS $file) {
-			preg_match_all( '/' . $pattern . '/', file_get_contents( $file ), $matches);
+		foreach ( $files as $file) {
+			preg_match_all( '/' . $pattern . '/', file_get_contents( $file ), $matches); // phpcs:ignore WordPress.WP.AlternativeFunctions
 			if ( isset( $matches[1] ) ) {
 				foreach ($matches[1] AS $match) {
 					$output[] = $match;
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -90,21 +90,20 @@ foreach ( $replacements as $namespace => $path ) {
 					-1,
 					$direct_change
 				);
-				if ( 0 < $direct_change ) {
+				if ( $direct_change ) {
 					$uses_change += $direct_change;
 				}
 			}
 		}
-		if ( 0 === $namespace_change && 0 < $uses_change ) {
+		if ( ! $namespace_change && $uses_change ) {
 			$file_notices[] = $file;
 		}
 
 		file_put_contents( $file, $contents );
 	}
 
-	// Update the namespace in the composer.json files.
-
-	$composer_files = get_dir_contents($vendor_dir . DIRECTORY_SEPARATOR . $path, '/composer.json/');
+	// Update the namespace in the composer.json files, recursively finding all files named explicitly "composer.json".
+	$composer_files = get_dir_contents($vendor_dir . '/' . $path, '/' . preg_quote(DIRECTORY_SEPARATOR . 'composer.json') . '$/');
 
 	array_map(
 		function( $file ) use ( $namespace, $new_namespace ) {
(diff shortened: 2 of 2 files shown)


ecaron:
@mikkamp I didn't know if there was prior art from other plugins that `prefix-vendor-namespace.php` or `GoogleAdsCleanupServices.php` were modeled after, or if the intent is that plugin build only happens on Linux machines. These adjustments have been helping me make progress in isolating what we're discussing in #1760, so I thought it could help others.

mikkamp:
Thanks for the contribution.

The script `prefix-vendor-namespace` was originally a [shell script based on an older implementation in WooCommerce](https://github.com/woocommerce/google-listings-and-ads/pull/66) where the Container library was namespaced.
We successfully build on MacOs and Linux environments, which are the main two we strive to work with for development.

ecaron:
@mikkamp Thanks for that backstory. Does that mean you will or won't consider merging this PR? Windows-support aside, the "spaces in the directory path" issue can come up for developers using tools like [Local](https://localwp.com/) to isolate their dev environments (e.g.l, localwp creates a folder called "Local Sites").

bin/GoogleAdsCleanupServices.php
```diff
@@ -178,6 +178,35 @@ protected function remove_enum( string $enum ) {
...
+	/**
+	 * Find a list of files in a path matching a pattern.
+	 *
+	 * @param string $path Package path
+	 * @param string $match Regex pattern to match
+	 * @return array Matching files
+	 */
+	protected function get_dir_contents($path, $match) {
+		try {
+			$rdi = new \RecursiveDirectoryIterator($path, \RecursiveDirectoryIterator::SKIP_DOTS);
```
mikkamp: We prefer to add a `use` statement at the top of the file so we don't have to explicitly specify the global namespace (same for other classes from the global namespace).

Also from my testing we can leave the second parameter out, by default skip dots is one of the enabled flags. Is there any specific reason it's needed on a different platform?

bin/GoogleAdsCleanupServices.php
```diff
@@ -178,6 +178,35 @@ protected function remove_enum( string $enum ) {
...
+		} catch ( Exception  $e ) {
+			printf(
```
mikkamp: It would be nice to use `$this->output_text` here to remain consistent and it also outputs differently depending on how the script was called. It also means `PHP_EOL` can be omitted.

bin/GoogleAdsCleanupServices.php
```diff
@@ -178,6 +178,35 @@ protected function remove_enum( string $enum ) {
...
+	 * Find a list of files in a path matching a pattern.
+	 *
+	 * @param string $path Package path
+	 * @param string $match Regex pattern to match
+	 * @return array Matching files
+	 */
+	protected function get_dir_contents($path, $match) {
+		try {
+			$rdi = new \RecursiveDirectoryIterator($path, \RecursiveDirectoryIterator::SKIP_DOTS);
+		} catch ( Exception  $e ) {
```
mikkamp: `Exception` is not imported so you are asking it to catch `\Automattic\WooCommerce\GoogleListingsAndAds\Util\Exception` which won't happen when there is an error.

Also a minor note that if the path is ever an empty string it will throw an uncaught ValueError (ok to ignore since it's a dev tool and we most likely won't encounter that scenario).

bin/GoogleAdsCleanupServices.php
```diff
@@ -186,24 +215,22 @@ protected function remove_enum( string $enum ) {
 	 * @return array List of names that match.
 	 */
 	protected function find_used_pattern( string $pattern ): array {
-		$command = "grep -rE --include=*.php '{$pattern}' {$this->code_path}";
-
-		// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.system_calls_exec
-		exec( $command, $output );
+		$files  = $this->get_dir_contents( $this->code_path,  '/\.php$/i');
+		$output = [];
+		foreach ( $files AS $file) {
```
mikkamp: lowercase `as` is preferred for WP standards.

bin/prefix-vendor-namespace.php
```diff
@@ -57,39 +60,51 @@
...
+				$replace       = preg_quote( $replace, '#' );
+				$contents      = preg_replace(
 					"#({$replace})#m",
 					"{$new_namespace}\\\\\$1",
-					$contents
+					$contents,
+					-1,
+					$direct_change
 				);
+				if ( 0 < $direct_change ) {
```
mikkamp: [Yoda conditions](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#yoda-conditions) are generally not recommended for larger or smaller than comparisons. I think since we are explicitly working with a count variable we can simplify the comparisons to make it easier to read.

```suggestion
				if ( $direct_change ) {
```

bin/prefix-vendor-namespace.php
```diff
@@ -57,39 +60,51 @@
...
+					$uses_change += $direct_change;
+				}
 			}
 		}
+		if ( 0 === $namespace_change && 0 < $uses_change ) {
```
mikkamp: Same here the counts will never be negative numbers.

```suggestion
		if ( ! $namespace_change && $uses_change ) {
```

bin/GoogleAdsCleanupServices.php
```diff
@@ -178,6 +178,35 @@ protected function remove_enum( string $enum ) {
 		$this->remove_file( "/{$this->version}/Enums/{$enum}.php" );
 	}
 
+	/**
+	 * Find a list of files in a path matching a pattern.
```
mikkamp: Would be nice to mention that this function searches in the path recursively.

bin/prefix-vendor-namespace.php
```diff
@@ -57,39 +60,51 @@
...
 
 	// Update the namespace in the composer.json files.
-	$composer_files = array_filter(
-		explode(
-			"\n",
-			`find "{$vendor_dir}/{$path}" -iname 'composer.json'`
-		)
-	);
+
+	$composer_files = get_dir_contents($vendor_dir . DIRECTORY_SEPARATOR . $path, '/composer.json/');
```
mikkamp: Is the `DIRECTORY_SEPARATOR` explicitly needed on Windows? There are other locations in this file where we are explicitly using `/` as the directory separator.

Also since we want to explicitly match the full file name shouldn't the pattern be: `/^composer.json$/`

mikkamp:
I think you have a valid point about the `spaces in the directory path`. Although at initial glance I had some comments about the way debug messages are output.

Since you mentioned you have an alternative solution in #1760 I didn't want to invest too much time. When priorities shift, I'd like to do some more research on whether we will be switching over to using Mozart, which is what WC did: https://github.com/woocommerce/woocommerce/pull/33703

I'll go ahead and add an initial review, once that's approved we can consider merging.

mikkamp (commented):
Thanks for the contribution, I tested the code and it seems to run well without needing `grep and find`. I've added some comments as I went through the code. The PHPCS errors have also been added after I ran the checks, it would be nice to address those as well. Since it is a dev tool it is ok to set some of them to be ignored.

bin/GoogleAdsCleanupServices.php
ecaron: `use` statements have been added to the file. I missed that FilesystemIterator included `SKIP_DOTS` by default, so that has been removed now.

bin/GoogleAdsCleanupServices.php
ecaron: Fair point. I've moved this catch to focus specifically on `UnexpectedValueException`

bin/prefix-vendor-namespace.php
ecaron: Regarding `DIRECTORY_SEPARATOR`, for reading file paths it isn't necessary (only when matching file paths - more on that later). So I reverted these constants to just `'/'` ([nice WordPress conversation on it here](https://core.trac.wordpress.org/ticket/44654)).

Regarding the explicit match of `composer.json`, you're right that the regex needs to be tweaked to ensure `sample-composer.json` isn't matched. However the `^` caused the match to fail, because there was a full path before it. Prefixing it with `/` only works on OS' that use that as the path separator, so using a `preg_quote` around `DIRECTORY_SEPARATOR` creates a way to ensure the file name is just `composer.json` while matching it on any OS.

bin/prefix-vendor-namespace.php
```diff
@@ -57,39 +60,50 @@
...
 
-	// Update the namespace in the composer.json files.
-	$composer_files = array_filter(
-		explode(
-			"\n",
-			`find "{$vendor_dir}/{$path}" -iname 'composer.json'`
-		)
-	);
+	// Update the namespace in the composer.json files, recursively finding all files named explicitly "composer.json".
+	$composer_files = get_dir_contents($vendor_dir . '/' . $path, '/' . preg_quote(DIRECTORY_SEPARATOR . 'composer.json') . '$/');
```
mikkamp: This line doesn't work on platforms where DIRECTORY_SEPARATOR is `/`. That's because the default deliminator for preg_quote is null. So unless you pass the deliminator which is being used in the regex `/` then it doesn't escape it.

It's building a regex like this:
```
//composer.json$/
```

As you can see this causes all the automated tests to fail because it thinks the `c` is a regex modifier.

```suggestion
	$composer_files = get_dir_contents(
		"{$vendor_dir}/{$path}",
		'/' . preg_quote( DIRECTORY_SEPARATOR, '/' ) . 'composer.json$/'
	);
```

bin/prefix-vendor-namespace.php
```diff
@@ -193,7 +234,7 @@ function remove_file_autoloads( string $file, array $composer_autoload, string $
 
 	$modified = false;
 	foreach ( $json['packages'] as $key => $package ) {
-		if ( 0 !== strpos( $package['name'], $package_name ) ) {
+		if ( strpos( $package['name'], $package_name ) ) {
```
mikkamp: I would leave this as a strict check because `strpos` can return false which means something completely different to found at 0. It's also causing the following error when I run the script:

```
Autoloaded file "vendor/google/apiclient/src/aliases.php" should be included in composer.json
```

bin/prefix-vendor-namespace.php
```diff
@@ -110,6 +124,43 @@ function( $file ) use ( $namespace, $new_namespace ) {
 	);
 }
 
+if ( count( $file_notices ) ) {
+	printf(
+		'Several files were modified without changes to namespace: %s' . PHP_EOL,
+		implode('; ', $file_notices)
```
mikkamp, line 133: This file does have a phpcs ignore but it would be nice to stick to the default WP spacing rules for easy legibility.
```suggestion
		implode( '; ', $file_notices )
```

bin/prefix-vendor-namespace.php
```diff
@@ -110,6 +124,43 @@ function( $file ) use ( $namespace, $new_namespace ) {
...
+ * @since 2.2.2
+ *
+ * @param string $path Package path
+ * @param string $match Regex pattern to match
+ * @return array Matching files
+ */
+function get_dir_contents( $path, $match ) {
+	try {
+		$rdi = new RecursiveDirectoryIterator( $path );
+	} catch ( UnexpectedValueException  $e ) {
```
mikkamp, line 149:
```suggestion
	} catch ( UnexpectedValueException $e ) {
```

bin/prefix-vendor-namespace.php
```diff
@@ -110,6 +124,43 @@ function( $file ) use ( $namespace, $new_namespace ) {
...
+			'Expected directory "%s" was not found' . PHP_EOL,
+			$path
+		);
+		exit( 1 );
+	}
+
+	$rii = new RecursiveIteratorIterator($rdi);
+    $rri = new RegexIterator($rii, $match);
+    $files = [];
+    foreach ($rri as $file) {
```
mikkamp:
```suggestion
	$rii = new RecursiveIteratorIterator( $rdi );
    $rri = new RegexIterator( $rii, $match );
    $files = [];
    foreach ( $rri as $file ) {
```

bin/prefix-vendor-namespace.php
```diff
@@ -122,21 +173,11 @@ function( $file ) use ( $namespace, $new_namespace ) {
 function find_files( string $path ): array {
 	global $vendor_dir, $dependencies;
 
-	$files = array_filter(
-		explode(
-			"\n",
-			`find "{$vendor_dir}/{$path}" -iname '*.php'`
-		)
-	);
+	$files = get_dir_contents($vendor_dir . '/' . $path, '/\.php$/i');
```
mikkamp:
```suggestion
	$files = get_dir_contents( "{$vendor_dir}/{$path}", '/\.php$/i' );
```

bin/prefix-vendor-namespace.php
```diff
@@ -122,21 +173,11 @@ function( $file ) use ( $namespace, $new_namespace ) {
...
 
 	if ( ! empty( $dependencies[ $path ] ) ) {
 		foreach ( $dependencies[ $path ] as $dependency ) {
-			$dependent_files = array_filter(
-				explode(
-					"\n",
-					`find "{$vendor_dir}/{$dependency}" -iname '*.php'`
-				)
-			);
+			$dependent_files = get_dir_contents($vendor_dir . '/' . $dependency, '/\.php$/i');
```
mikkamp:
```suggestion
			$dependent_files = get_dir_contents( "{$vendor_dir}/{$dependency}", '/\.php$/i' );
```

bin/GoogleAdsCleanupServices.php
```diff
@@ -178,6 +182,35 @@ protected function remove_enum( string $enum ) {
...
+	 * @return array Matching files
+	 */
+	protected function get_dir_contents( $path, $match ) {
+		try {
+			$rdi = new RecursiveDirectoryIterator( $path );
+		} catch ( UnexpectedValueException $e ) {
+			$this->output_text( sprintf(
+				'Expected directory "%s" was not found',
+				$path
+			) );
```
mikkamp:
```suggestion
			$this->output_text(
				sprintf(
					'Expected directory "%s" was not found',
					$path
				)
			);
```

bin/GoogleAdsCleanupServices.php
```diff
@@ -178,6 +182,35 @@ protected function remove_enum( string $enum ) {
...
+			exit( 1 );
+		}
+
+		$rii = new RecursiveIteratorIterator( $rdi );
+		$rri = new RegexIterator( $rii, $match );
```
mikkamp:
```suggestion
		$rii   = new RecursiveIteratorIterator( $rdi );
		$rri   = new RegexIterator( $rii, $match );
```

bin/GoogleAdsCleanupServices.php
```diff
@@ -186,24 +219,22 @@ protected function remove_enum( string $enum ) {
 	 * @return array List of names that match.
 	 */
 	protected function find_used_pattern( string $pattern ): array {
-		$command = "grep -rE --include=*.php '{$pattern}' {$this->code_path}";
-
-		// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.system_calls_exec
-		exec( $command, $output );
+		$files  = $this->get_dir_contents( $this->code_path,  '/\.php$/i');
```
mikkamp:
```suggestion
		$files  = $this->get_dir_contents( $this->code_path, '/\.php$/i' );
```

bin/GoogleAdsCleanupServices.php
```diff
@@ -186,24 +219,22 @@ protected function remove_enum( string $enum ) {
...
+		$output = [];
+		foreach ( $files as $file) {
+			preg_match_all( '/' . $pattern . '/', file_get_contents( $file ), $matches); // phpcs:ignore WordPress.WP.AlternativeFunctions
+			if ( isset( $matches[1] ) ) {
+				foreach ($matches[1] AS $match) {
```
mikkamp:
```suggestion
		foreach ( $files as $file ) {
			preg_match_all( "/{$pattern}/", file_get_contents( $file ), $matches ); // phpcs:ignore WordPress.WP.AlternativeFunctions
			if ( isset( $matches[1] ) ) {
				foreach ( $matches[1] as $match ) {
```

bin/GoogleAdsCleanupServices.php
```diff
@@ -186,24 +219,22 @@ protected function remove_enum( string $enum ) {
...
-		return array_unique(
-			array_map(
-				function( $line ) use ( $pattern ) {
-					preg_match( "/{$pattern}/", $line, $matches );
-					return $matches[1];
-				},
-				$output
-			)
-		);
+		return array_unique($output);
```
mikkamp:
```suggestion
		return array_unique( $output );
```

mikkamp (commented):
Thanks for the changes, looks like we are getting close, I just came across 2 errors which need to be addressed (preg_quote and strpos check).

The rest are just minor phpcs changes to stick to WP spacing standards.

ecaron:
@mikkamp Do you run the `phpcs.xml.dist` as-is for local linting with `phpcs` and `phpcbf`, or do you run different versions? I've been running `phpcs` without a `phpcs.xml` present, and I think that's causing the lint issues that I should've been catching.

mikkamp (approved):
Thanks for the additional changes. It's working as expected now, and all the tests are passing. So I'll go ahead and approve this PR.

mikkamp:
> Do you run the phpcs.xml.dist as-is for local linting with phpcs and phpcbf, or do you run different versions? I've been running phpcs without a phpcs.xml present, and I think that's causing the lint issues that I should've been catching.

I use VS code with the phpcs extension and the option to auto search config enabled. So it automatically picks up the right set of rules. Seems the extension is deprecated now but it's still working fine for me.

Issue #1780, opened 2022-11-17
PMax Assets


jorgemd24:
This feature implements PMax Assets.

Project Thread: [pcTzPl-HY-p2](pcTzPl-12o-p2)
Technical Exploration: pcTzPl-1c9-p2

## Terminology
[Assets](https://href.li/?https://developers.google.com/google-ads/api/fields/v11/asset): headlines, descriptions, logos, images, Youtube links etc.
[Asset Group](https://href.li/?https://developers.google.com/google-ads/api/fields/v11/asset_group): It’s a group that contains the assets and is linked to a campaign and a Final URL.
[Assets Group Assets](https://href.li/?https://developers.google.com/google-ads/api/fields/v11/asset_group_asset): It’s the link between the Assets and Asset Groups.

## Figma link

[fqR0EHi63lWahRcVTKCcba-fi.x?t=whbSXeSxxvs1voWR-0](fqR0EHi63lWahRcVTKCcba-fi-3472%3A185864)

### Acceptance criteria - Backend

- [X] Endpoint to fetch  "Final URLs" from post/product/terms pages. https://github.com/woocommerce/google-listings-and-ads/pull/1784
- [x] Endpoint to fetch assets from specific URL. 
  * [x] Fetch assets for Posts. https://github.com/woocommerce/google-listings-and-ads/pull/1795
    * [x] Suggest inserted HTML Images https://github.com/woocommerce/google-listings-and-ads/pull/1860
  * [x] Fetch assets for Terms. https://github.com/woocommerce/google-listings-and-ads/pull/1797
- [x] Resize images https://github.com/woocommerce/google-listings-and-ads/pull/1804
- [x] Endpoint to fetch AssetGroup and Assets using the Ads API. https://github.com/woocommerce/google-listings-and-ads/pull/1820
- [x] Endpoint to update AssetGroups.  https://github.com/woocommerce/google-listings-and-ads/pull/1832
- [x] Fetch assets from others asset groups. https://github.com/woocommerce/google-listings-and-ads/pull/1837
- [x] Add default assets for a non-static homepage.  https://github.com/woocommerce/google-listings-and-ads/pull/1847
- [x] Endpoint to create AssetGroups https://github.com/woocommerce/google-listings-and-ads/pull/1851
- [x] Assets batching - https://github.com/woocommerce/google-listings-and-ads/pull/1864
[...]

PR #1784, opened 2022-11-18
PMax-Asset: Final URL Suggestions Endpoint

jorgemd24:

Closes part of #1780 

This PR adds a new endpoint `assets/final-url/suggestions` that returns a list of `posts` and `terms`. The list will be used to display a dropdown in the frontend so the user can populate assets from a specific URL. 

image

The response will contain an array of items formatted in the following way:

```
[
    {
        "id": post|term ID,
        "type": post|term,
        "title": post|term title,
        "url": post|term URL.
    },
    .....
 ]
```

1. Make a request to `/wc/gla/assets/final-url/suggestions?search=YOUR_SEARCH&per_page=30&order_by=title`
2. The list should display all your posts/pages/products/categories URLs that match your search.
3. The response should contain the name and the URL for each item.
4. If the search param is empty it will return the Homepage if it is setup as a `static page` and the Shop page if it is setup in WC.

See:  [How to create a Static Front Page](https://wordpress.org/support/article/creating-a-static-front-page/#how-to-create-a-static-front-page) and [Tell WooCommerce what pages to use](https://woocommerce.com/document/woocommerce-pages/#tell-woocommerce-what-pages-to-use)


src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,128 @@
...
+	/**
+	 * Return a list of pages that can be used to suggest assets.
+	 *
+	 * @return array Array of pages
+	 */
+	public function get_page_suggestions(): array {
+		$posts = $this->get_post_suggestions();
+		$terms = $this->get_terms_suggestion();
+		return array_merge( $posts, $terms );
+	}
```
puntope: It was confusing for me the term "page" being used for this function since in WP terminology Page is a type of Post. So I thought we were going to return post_type pages in here. But is actually getting all the suggestions posts + terms

💅  What about renaming this as `get_suggestions` or `get_all_suggestions`?

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,130 @@
...
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Assets
+ */
+class AssetSuggestionsController extends BaseController {
+
+	/**
+	 * Service used to populate ads suggestions data.
+	 *
+	 * @var AssetSuggestionsService
+	 */
+	protected $asset_group;
```
puntope: Why `asset_group` and not `asset_suggestion_service`?

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,130 @@
...
+				'context'     => [ 'view' ],
+				'enum'        => [ 'post', 'term' ],
+				'readonly'    => true,
+			],
+			'post_type' => [
+				'type'        => 'string',
+				'description' => __( 'The post type', 'google-listings-and-ads' ),
+				'context'     => [ 'view' ],
+				'readonly'    => true,
+			],
```
puntope: ❓ Reading a bit the project thread and explorations, as well as checking Figma file. I don't see the need of this 2 fields besides the need of future calls to other endpoints for getting the assets (I guess that would be different for post and taxonomies???). Is that correct?

Btw, maybe `type` could be computed and hence omitted from the API. For example, if `post_type` is not NULL, then is type `post`, otherwise is a `term`

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,130 @@
...
+		parent::__construct( $server );
+		$this->asset_group = $asset_suggestion;
+	}
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'assets/page/suggestions',
```
puntope: In line with my previous comment here https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1027503068. This seems not accurate for what we got. 

I would name it `assets/suggestions`

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,130 @@
...
+			]
+		);
+	}
+
+	/**
+	 * Get the callback function for the list of final-url suggestions request.
+	 *
+	 * @return callable
+	 */
+	protected function get_pages_suggestions_callback(): callable {
```
puntope: Same as in https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1027509284.

The term pages, seems not accurate for me.

src/Proxies/WP.php
```diff
@@ -177,4 +178,76 @@ public function number_format_i18n( float $number, int $decimals = 0 ): string {
...
+	 *                                 function parameter will be parsed as a taxonomy or array of taxonomies.
+	 *                                 Default empty.
+	 * @return WP_Term[]|int[]|string[]|string|WP_Error Array of terms, a count thereof as a numeric string,
+	 *                                                  or WP_Error if any of the taxonomies do not exist.
+	 *                                                  See the function description for more information.
+	 */
+	public function get_terms( $args = [], $deprecated = '' ) {
+		return get_terms( $args, $deprecated );
+	}
 }
```
puntope: Do we really need to Proxy these functions?

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,128 @@
...
+				'public'              => true,
+			]
+		);
+
+		// Exclude attachment post_type
+		$filtered_post_types = array_diff( $post_types, $excluded_post_types );
+
+		$args = [
+			'post_type'      => $filtered_post_types,
+			'posts_per_page' => -1,
```
puntope: ❓ Don't take this as a 100% assured assumption... but are we sure we want to return all the posts for a site? (+ all terms) That could be a huge number of posts/terms. I'm unsure about the potential performance leaks on this approach. 

As far as I read in the project, this is going to be a selector in where the user can type the URL or Post/Term name. 
So I wonder if we can just return a limited number of Posts/Terms and maybe have an extra endpoint for querying them. 🤔 

What are your thoughts about this?

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,128 @@
...
+				'public'       => true,
+				'show_in_menu' => true,
+			],
+		);
+
+		// Skip empty terms
+		$terms = $this->wp->get_terms(
+			[
+				'taxonomy'   => $taxonomies,
+				'hide_empty' => true,
```
puntope: I guess this is true by default. 

Btw. Why we don't allow to show empty terms?

src/Proxies/WP.php
puntope: Maybe is for testing them easier???

puntope (commented):
Thanks for the PR @jorgemd24 

The behaviour is working as described. I left some concerns about performance and some comments regarding code quality.

Thanks

src/Proxies/WP.php
jorgemd24: Yes, it makes it easier to mock those functionalities.

src/Ads/AssetSuggestionsService.php
jorgemd24: Yes, I had the same concern, I talked to @eason9487  about this last week, and for now, the idea is to return the complete list of URLs and depending on how the WP/WC component works we could add some type of matching URLs function depending on the input. Also, the final design still needs to be determined. See comment here: pcTzPl-1eS-p2#comment-2048

src/Ads/AssetSuggestionsService.php
jorgemd24: > I guess this is true by default.

Removed https://github.com/woocommerce/google-listings-and-ads/pull/1784/commits/77c3744533071f3f008bc91408e9d245457c36e4

> Btw. Why we don't allow to show empty terms?

Because if it does not contain any product/page/post etc, the assets that we can populate like images are limited and I am not sure if the user would like to promote a category/term with no items.

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
jorgemd24: >❓ Reading a bit the project thread and explorations, as well as checking Figma file. I don't see the need of this 2 fields besides the need of future calls to other endpoints for getting the assets (I guess that would be different for post and taxonomies???). Is that correct?

Correct, to fetch the assets from a specific post or term I will need to know the type because they work differently.

I removed the post_type as it is not required anymore from the frontend perspective.  https://github.com/woocommerce/google-listings-and-ads/pull/1784/commits/f79f190ba63471b729cce93407396ea242d4b42a

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
jorgemd24: I have updated it to `assets/final-url/suggestions` as `assets/suggestions` is too generic and can be interpreted as suggesting assets instead of final URLs.

src/Ads/AssetSuggestionsService.php
jorgemd24: I moved this conversation to pcTzPl-1eS-p2#comment-2057 , depending on the outcome I will adjust this PR.

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
puntope: Thanks for this tweak.

It's going to be any other suggestion endpoints or other `final-url` related endpoints? I like more `final-url` than `pages` for sure...  but having a "directory" `/final-url` made me think that it was going to be more endpoints here (as  children of `/final-url/`) 

If it's not the case, maybe worth it to consider just `assets/final-url-suggestions` or just `assets/url-suggestions` (I'm unsure what "final" means here)

src/Ads/AssetSuggestionsService.php
puntope: Ok, I understand the logic. I was thinking in the case of a user want to fill the campaigns for the terms and after that adding products, not sure if this is a real use case tho.

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
jorgemd24: > It's going to be any other suggestion endpoints or other final-url related endpoints? I like more final-url than pages for sure... but having a "directory" /final-url made me think that it was going to be more endpoints here (as children of /final-url/)

For now, the other endpoint will be `assets/suggestions` and the response will include the title, descriptions, and images which can be used for creating ads.

I have seen that we have similar patterns in other controllers see here:

https://github.com/woocommerce/google-listings-and-ads/blob/19c800c34619f4c767bc11024ccac4cf3bf71393/src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php#L32

https://github.com/woocommerce/google-listings-and-ads/blob/19c800c34619f4c767bc11024ccac4cf3bf71393/src/API/Site/Controllers/MerchantCenter/TargetAudienceController.php#L93

> I'm unsure what "final" means here 

It is because the field in the Ads API and in the Ads interface is called Final URL. See here more information https://support.google.com/google-ads/answer/6080568?hl=en

src/Ads/AssetSuggestionsService.php
jorgemd24: I have changed it to 'hide_empty' => false and it will be up to the user if they want to promote the page or not.

jorgemd24:
@puntope, I have updated the PR. The endpoint is now limiting the results and it uses a `search` param to query for title posts or terms names.

See the PR description for more detailed information!

Thanks for your time!

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,201 @@
...
+		$per_page_posts = (int) ceil( $per_page / 2 );
+		$per_page_terms = (int) $per_page - $per_page_posts;
+
+		$posts = $this->get_post_suggestions( $search, $per_page_posts );
+
+		// If not posts results, try to get all results from terms.
+		if ( count( $posts ) === 0 ) {
+			$result = $this->get_terms_suggestion( $search, $per_page );
+			return $this->sort_results( $result, $order_by );
+		}
```
puntope: As far as I understand this code, if the client enters `$per_page = 10` this code will try to get 5 posts, and 5 terms.

Let's suppose we have 10 terms. In case there are exactly 0 posts, the code will try to get 10 terms to show 10 items. Ok

❓  What if there  is 1-4 posts?  Then the results will be less than 10, even tho, this is not true, since we have 10 terms.

I guess instead of  `if ( count( $posts ) === 0 ) ` it should be `if ( count( $posts ) <  $per_page_posts)` 

However, this piece of code seems unnecessary for me, since we can compute that in the terms call directly.

`$per_page_terms = $per_page - count($posts);`
`$terms  = $this->get_terms_suggestion( $search, $per_page_terms);`

However, we can have still the case in which there is 10 or more posts, and there is less than 5 terms. In that case shouldn't we fill it with the posts?

src/Proxies/WP.php
```diff
@@ -177,4 +178,114 @@ public function number_format_i18n( float $number, int $decimals = 0 ): string {
...
+	public function get_posts( array $args ): array {
+		return get_posts( $args );
+	}
+
+	/**
+	 * Gets a list of all registered post type objects.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array|string $args     Optional. An array of key => value arguments to match against
```
puntope, line 204: In the PHPDOCS it says `array|string` for `$args` param. However, in the method signature it claims only `array` for `$args` attribute.

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,201 @@
...
+
+	}
+
+	/**
+	 * Get defaults final urls suggestions.
+	 *
+	 * @return array Return Array with the default final urls
+	 */
+	protected function get_defaults_final_urls_suggestions(): array {
+		// We can only offer assets if the page is static.
```
puntope: 💅 

// We can only offer assets if the homepage is static

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,201 @@
...
+		$result = array_merge( $posts, $terms );
+
+		return $this->sort_results( $result, $order_by );
+
+	}
+
+	/**
+	 * Get defaults final urls suggestions.
+	 *
+	 * @return array Return Array with the default final urls
```
puntope: 💅 
Redundant Return Array In the comment

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		return $defaults;
+	}
+
+	/**
+	 *  Order suggestions alphabetically
+	 *
+	 *  @param array  $array associative array
+	 *  @param string $field Sort by a specific field
```
puntope, line 180: Missed return directive in PHPDOCS

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,201 @@
...
+
+	/**
+	 * Return an assotiave array with the page suggestion response format.
+	 *
+	 * @param int    $id post|term ID
+	 * @param string $type post|term
+	 * @param string $title page|term title
+	 * @param string $url page|term url
+	 *
+	 * @return array
```
puntope: Missed return directive description

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,158 @@
...
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Assets
+ */
+class AssetSuggestionsController extends BaseController {
+
+	/**
+	 * Service used to populate ads suggestions data.
+	 *
+	 * @var AssetSuggestionsService
+	 */
+	protected $asset_suggestion_service;
```
puntope: 💅 

Minor, but the class is in plural `asset_suggestions` but the variable is `asset_suggestion`

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,158 @@
...
+	 *
+	 * @return array
+	 */
+	public function get_collection_params(): array {
+		return [
+			'search'   => [
+				'description'       => __( 'Search for post title or term name', 'google-listings-and-ads' ),
+				'type'              => 'string',
+				'default'           => '',
+				'validate_callback' => 'rest_validate_request_arg',
```
puntope: Maybe we can do some sanitisation as the original search param does.

https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api/endpoints/class-wp-rest-controller.php#L364

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,158 @@
...
+			],
+			'per_page' => [
+				'description'       => __( 'The number of items to be return', 'google-listings-and-ads' ),
+				'type'              => 'number',
```
puntope, line 75: We could add sanitisation 

https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api/endpoints/class-wp-rest-controller.php#L358

Also declaring the `'minimum' => 1` directive could allow to omit the [$per_page < 0  check in get_final_urls_suggestions](https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#diff-23f729894d673b96e8ca50cf015ed612fce8328ff39e6f5132e4dab3da3bdfcbR114-R116)

puntope (commented):
Hey there @jorgemd24 

Thanks for the adjustments!

I left some additional comments for the new changes.

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
puntope: I actually just noticed I texted wrong my previous comment in https://github.com/woocommerce/google-listings-and-ads/pull/1784#discussion_r1027503401

And thats why maybe is that way. :sorry:

src/Ads/AssetSuggestionsService.php
puntope: Was thinking about this... Not sure if this strategy of 50%/50%  is something defined as a requirement but maybe it is easier to just return all the posts we can until we complete the "per page" param and fill with terms if necessary (when there are less posts than the per_page param).

jorgemd24: >However, we can have still the case in which there is 10 or more posts, and there is less than 5 terms. In that case shouldn't we fill it with the posts?

That could be the case, however, every time that the search is more specific, the wanted result will come up if it does not show up in the first request. I tweaked a bit the code to try to get more results if we have some margin.

See: https://github.com/woocommerce/google-listings-and-ads/pull/1784/commits/fcf2045afa0bf8076101753c7271fd11c7f8aff9

> it is easier to just return all the posts we can until we complete the "per page" param and fill with terms if necessary (when there are less posts than the per_page param).

The problem with this approach is that if we have X posts matching the query and per_page = X always will return `posts` and the user will not be able to search for terms.

src/Proxies/WP.php
jorgemd24: I removed the `array` type as union types are only allowed in PHP >=8.0

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
jorgemd24: Probably was that 😆

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
jorgemd24: Thanks for this! TIL 😃

jorgemd24:
Thanks, @puntope for reviewing this! This is ready for another round!

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,211 @@
...
+		// Try to get more results using the terms
+		$per_page_terms = $per_page - count( $posts );
+
+		$terms = [];
+
+		// get_terms  evaluates $per_page_terms = 0 as a falsy, therefore it will not add the LIMIT clausure returning all the results.
+		// See: https://github.com/WordPress/WordPress/blob/abe134c2090e84080adc46187884201a4badd649/wp-includes/class-wp-term-query.php#L868
+		if ( $per_page_terms > 0 ) {
+			$terms = $this->get_terms_suggestion( $search, $per_page_terms );
+		}
```
puntope: 💅 

Adding 

```
if ( $per_page <= 0 ) {
  return [];
}
```
Inside `get_terms_suggestion` instead of here, could reduce the verbosity of this code to just:

`$terms = $this->get_terms_suggestion( $search, $per_page_terms );`

Omitting the conditional check and the `$terms = []` initialisation

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,211 @@
...
+
+		$pending_results = $per_page - count( $posts ) - count( $terms );
+		$more_results    = [];
+
+		// Try to get more results using posts
+		if ( $pending_results > 0 && count( $posts ) === $per_page_posts ) {
+			$more_results = $this->get_post_suggestions( $search, $pending_results, $per_page_posts );
+		}
```
puntope, line 146: 💅  

Adding

```
if ( $per_page <= 0 ) {
  return [];
}
```
Inside `get_post_suggestions` instead of here, could reduce the verbosity of this code omitting `$pending_results > 0`

puntope: 💅 

Applying the previous suggestion we could do just

`$more_results = $this->get_post_suggestions( $search, $pending_results, count( $posts ) );`

Omitting the condition and the `$more_results    = [];` initialization.

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,211 @@
...
+
+	/**
+	 * Get terms that can be used to suggest assets
+	 *
+	 * @param string $search The search query
+	 * @param int    $per_page Number of items per page
+	 *
+	 * @return array
+	 */
+	protected function get_terms_suggestion( $search, $per_page ): array {
```
puntope: 💅 

Minor, but just noticed that `get_terms_suggestion` is in singular. Shouldn't be `suggestions`?

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,162 @@
...
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'assets/final-url/suggestions',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_final_urls_suggestions_callback(),
```
puntope: 💅  minor

`get_final_urls_suggestions_callback `has `final_urls` (plural) in the name whereas the endpoint is in singular `final-url`

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,162 @@
...
+		return function( Request $request ) {
+			$search   = $request->get_param( 'search' );
+			$per_page = $request->get_param( 'per_page' );
+			$order_by = $request->get_param( 'order_by' );
+			return array_map(
+				function( $item ) use ( $request ) {
+					$data = $this->prepare_item_for_response( $item, $request );
+					return $this->prepare_response_for_collection( $data );
+				},
+				$this->asset_suggestions_service->get_final_urls_suggestions( $search, $per_page, $order_by )
```
puntope: 💅  minor

Same as in https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1031025484  there are some plural and singular inconsistencies

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * Get the item schema name for the controller.
+	 *
+	 * Used for building the API response schema.
+	 *
+	 * @return string
+	 */
+	protected function get_schema_title(): string {
+		return 'asset_final_urls_suggestions';
```
puntope: 💅  Minor

Same as in https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1031025484

Plural/singular inconsistencies

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,211 @@
...
+	/**
+	 * Return a list of final urls that can be used to suggest assets.
+	 *
+	 * @param string $search The search query
+	 * @param int    $per_page Number of items per page
+	 * @param string $order_by Order by: id, type, title, url
+	 *
+	 * @return array final urls with their title, id & type.
+	 */
+	public function get_final_urls_suggestions( $search = '', $per_page = 30, $order_by = 'title' ): array {
```
puntope: 💅  minor

Same as 
https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1031025484

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,211 @@
...
+		return $this->sort_results( $result, $order_by );
+
+	}
+
+	/**
+	 * Get defaults final urls suggestions.
+	 *
+	 * @return array default final urls.
+	 */
+	protected function get_defaults_final_urls_suggestions(): array {
```
puntope: 💅  minor

Same as

https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1031025484

tests/Unit/API/Site/Controllers/Assets/AssetSuggestionsControllerTest.php
```diff
@@ -0,0 +1,68 @@
...
+			'url'   => 'http://localhost:8082/product-category/clothing/',
+		],
+
+	];
+	protected const TEST_NO_FINAL_URL_SUGGESTIONS = [];
+
+	public function setUp(): void {
+		parent::setUp();
+
+		$this->asset_suggestion = $this->createMock( AssetSuggestionsService::class );
```
puntope: 💅  Minor

This could be changed as well 

`$this->asset_suggestions`

src/API/Site/Controllers/Assets/AssetSuggestionsController.php
```diff
@@ -0,0 +1,162 @@
...
+	/**
+	 * AssetSuggestionsController constructor.
+	 *
+	 * @param RESTServer              $server
+	 * @param AssetSuggestionsService $asset_suggestion
+	 */
+	public function __construct( RESTServer $server, AssetSuggestionsService $asset_suggestion ) {
+		parent::__construct( $server );
+		$this->asset_suggestions_service = $asset_suggestion;
+	}
```
puntope: 💅 

Still some naming as `asset_suggestion` instead of `asset_suggestions`

src/Ads/AssetSuggestionsService.php
```diff
@@ -0,0 +1,211 @@
...
+			$array,
+			function ( $a, $b ) use ( $field ) {
+				return strcmp( (string) $a[ $field ], (string) $b[ $field ] );
+			}
+		);
+
+		return $array;
+
+	}
+
```
puntope: This is not working for `order_by=id`.  Unsure how essential is to order by id. Maybe it is just ok to remove that `order_by=id` instead of adding complexity in the sort function. 

Another thing to consider is that this will place lowercase after uppercase. See last entries of my response ordering by Title. Maybe we could lowercase the fields before comparing.

```
{
        "id": 182,
        "type": "post",
        "title": "Term Post",
        "url": "https://wp23.test/term-post/"
    },
    {
        "id": 35,
        "type": "post",
        "title": "WordPress Pennant",
        "url": "https://wp23.test/product/wordpress-pennant/"
    },
    {
        "id": 166,
        "type": "post",
        "title": "p4", // starts with p, but appears after W.
        "url": "https://wp23.test/p4/"
    }
```

puntope (commented):
Howdy @jorgemd24  Thanks again for the changes. We are getting there.

Only one thing that kept my attention was this comment in regards sorting logic.
https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1031032331

Also, I left some additional 💅  comments. Especially would be nice to have the refactor over the offset logic in `get_final_urls_suggestions ` method to reduce the verbosity.

Rest is just naming issues with some inconsistencies between `suggestion` and `suggestions` or `final_url` and `final_urls` in code and tests.

For now what I tested so far:

- Tested with Postman with different sets of Post/Terms and works as expected ✅ 
- Tested default params ✅ 
- Tested no search params (static homepage) ✅ 
- Tested orderby ⚠️ 
- Tested with wrong  API params ✅

src/Ads/AssetSuggestionsService.php
jorgemd24: I excluded the order by id as it is not essential. I think ordering by title will help the user to find the desired result. See [a3c8b5a](https://github.com/woocommerce/google-listings-and-ads/pull/1784/commits/a3c8b5a9fe9eb5d9a2c298590cf81dae836816af)

src/Ads/AssetSuggestionsService.php
jorgemd24: Not sure about this one, if we omit `$pending_results > 0`, it seems that we always want to fetch more results if `count ( $posts ) === $per_page_posts`, which is not the case because it will depend on the terms results.

jorgemd24:
Thanks @puntope,

I added your suggestions about the plural names and the ordering.

Thanks!

src/Ads/AssetSuggestionsService.php
puntope: If you add

```
if ( $per_page <= 0 ) {
  return [];
}
```

Inside `get_post_suggestions` it would return `[]` in case `$pending_results <= 0` so it won't fetch more results.

puntope (approved):
✅  LGTM

Thanks for all the adjustments and working on this. Now everything works perfect! 

I left a follow-up comment in https://github.com/woocommerce/google-listings-and-ads/pull/1784/files#r1031020057. But is not a blocker.

PR #1779, opened 2022-11-17
Bump min supported WP version to 5.9 and align all WP and WC packages to WC 6.9

eason9487:
### Changes proposed in this Pull Request:

In order to use the `TourKit` component provided in `@woocommerce/components` 10.3.0, which is available from a minimum of WC 6.9, it's time to upgrade it.

Moreover, to avoid conflict between WC and WP packages and different `react` versions, we also have to bump all `@woocommerce/*` packages to align with WC 6.9, and all `@wordpress/*` packages need to be aligned with the corresponding versions used in `@woocommerce/components` 10.3.0. Otherwise, the JS unit test would encounter a lot of "Invalid hook call" errors due to multiple `react` versions.

Furthermore, to align the upgraded `@wordpress/*` packages, this PR also set the minimum supported WP version to 5.9.

Finally, since the upgrading of packages was accompanied by many errors in the JS unit test, several fixes were added.

### Detailed test instructions:

1. `npm ci` or `npm i`
2. `npm run test:js` to check if all JS tests could pass.
3. `npm start`, `npm run start:hot`, or `npm run dev`
4. Look around all GLA pages to see if there are UI problems or errors in the DevTool Console

### Additional details:

The ["Invalid hook call" errors](https://github.com/woocommerce/google-listings-and-ads/actions/runs/3469909930/jobs/5797467733) in #1752 were caused by multiple `react` versions while running jest. The following is what this PR did to upgrade all `@woocommerce/*` and all `@wordpress/*` packages:

```bash
npm i --save @woocommerce/components@^10.3.0 @woocommerce/currency@^4.2.0 @woocommerce/data@^4.1.0 @woocommerce/date@^4.2.0 @woocommerce/number@^2.4.0 @woocommerce/tracks@^1.3.0
npm i --save @wordpress/api-fetch@^6.0.1 @wordpress/base-styles@^4.1.1 @wordpress/components@^19.5.0 @wordpress/compose@^5.1.2 @wordpress/data-controls@^2.3.2 @wordpress/data@^6.2.1 @wordpress/date@^4.3.1 @wordpress/dom@^3.3.2 @wordpress/element@^4.1.1 @wordpress/hooks@^3.5.0 @wordpress/html-entities@^3.3.1 @wordpress/i18n@^4.3.1 @wordpress/icons@^8.1.0 @wordpress/primitives@^3.1.1 @wordpress/url@^3.4.1
npm i --save-dev react@^17.0.2 react-dom@^17.0.2
rm package-lock.json
[...]

PR #1790, opened 2022-11-24
Bump WC/WP Components: Component Text fix

puntope:

This PR implements our own version of `<Text>` component instead of using `__experimentalText` from `wordpress/components`

The idea is:

A- Fix the UX issues in regards <Text> components after upgrading the dependencies. 
B- Avoid warnings in `js:lint` tasks
C- Avoid future problems regarding further changes in `__experimentalText` component 
D- Simplify our version of <Text> components since we are not using any of the enhancements provided by `__experimentalText`
E- Avoid massive replacement of previous <Text> implementations over all the codebase.

1. Check the components affected by this change are rendering the text component properly.
2. As a reference you can use the Getting Started page in `develop` branch that include most of the <Text> variants.

ℹ️  Notice this PR doesn't fix other UX issues, like spacing in Flex items.

### DIFF
--- a/js/src/components/app-text/index.js
+++ b/js/src/components/app-text/index.js
@@ -1,0 +1,41 @@
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import './index.scss';
+
+/**
+ * Renders a paragraph with some predefined styles based on the variant
+ *
+ *
+ * ```jsx
+ * <AppText variant="title-small">
+ * 		My text
+ * </AppText>
+ * ```
+ *
+ * @param {Object} props Component props
+ * @param {string} props.variant The variant to use for the text
+ * @param {string} props.className Custom classname for this component
+ * @param {JSX.Element} props.children The content for this component
+ * @return {JSX.Element} The component
+ */
+const AppText = ( { variant, className, children } ) => {
+	return (
+		<p
+			className={ classnames(
+				'gla-app-text',
+				`gla-app-text--${ variant }`,
+				className
+			) }
+		>
+			{ children }
+		</p>
+	);
+};
+
+export default AppText;
--- a/js/src/components/app-text/index.js
+++ b/js/src/components/app-text/index.js
@@ -21,20 +21,28 @@ import './index.scss';
  * @param {Object} props Component props
  * @param {string} [props.variant] The variant to use for the text
  * @param {string} [props.className] Custom classname for this component
+ * @param {keyof JSX.IntrinsicElements} [props.as] as Custom HTML tag for the component (by default <p>)
  * @param {JSX.Element} props.children The content for this component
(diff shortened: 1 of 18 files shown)


eason9487 (commented):
Thanks for the adjustment.

There are three use cases that have `as` prop, for example:
https://github.com/woocommerce/google-listings-and-ads/blob/3d40f09eeadede59595361ab4802c838c4d0050e/js/src/product-feed/issues-table-card/index.js#L83

Wonder if this was intentionally not implemented or missed. If we are not going to port the `as` prop, then perhaps removing it from the use cases, or adding some description that it is not supported would be nicer.

puntope:
Thanks @eason9487  It was a miss on my side.

I added support for `as` tag
This is ready for a new round

js/src/components/app-text/index.js
```diff
@@ -0,0 +1,49 @@
...
+ * ```jsx
+ * <AppText variant="title-small">
+ * 		My text
+ * </AppText>
+ * ```
+ *
+ * @param {Object} props Component props
+ * @param {string} [props.variant] The variant to use for the text
+ * @param {string} [props.className] Custom classname for this component
+ * @param {keyof JSX.IntrinsicElements} [props.as] as Custom HTML tag for the component (by default <p>)
```
eason9487: Looks like the current JSDoc config supports `keyof` but the `npm run doc:tracking` script doesn't. Probably it needs more investigation.

puntope: Thanks for catching this @eason9487. 

Is this a blocker? Or it's possible to set this prop as "string"  and continue investigation in future PR on the `doc:tracking` problem?

puntope: Alternative options are:

- Fix the JSDOC tracking to support keyof. Ideal but it will need more time and me getting familiar with that plugin. So it will delay everything a bit. 
- Add just the common tags like 'p'|'h1'|'h2'...|'h6'|'div'|'span'...
- Like the previous option, but adding all the tags.

eason9487: > Is this a blocker?

Since `{keyof JSX.IntrinsicElements}` is very precise, so this made me hesitate.

> Or it's possible to set this prop as "string" and continue investigation in future PR on the `doc:tracking` problem?

I vote for this one since it appears to be the most ideal option at present. Using `string` type and maybe adding one more sentence like ", e.g. h2, span." to its description would be very enough for devs to understand this prop quickly.

eason9487: > continue investigation in future PR on the `doc:tracking` problem?

Maybe open a follow-up issue first and circle back someday since the issue is not very urgent.

eason9487 (approved):
Thanks for the further adjustment. There is only [one concern](https://github.com/woocommerce/google-listings-and-ads/pull/1790#discussion_r1034479870) left but I believe there are already several workable solutions for now, so I'll be approving this PR in advance.

PR #1791, opened 2022-11-24
Bump WC/WP Components: Fix Component Card CSS

puntope:

It fixes the CSS in Card component:

A- Changing `components-flex__item` selector to the new  `components-flex-item`
B- Changing `components-flex__block` selector to the new  `components-flex-block`
C- Adding `>div:first-child ` in the `gla-get-started-card` 

1. Check that the Cards in getting Started Page are correct. 
2. Use develop branch as reference

ℹ️  It's possible that other issues with Card component are still there. This PR doesn't cover that.

### DIFF
--- a/js/src/dashboard/index.scss
+++ b/js/src/dashboard/index.scss
@@ -45,6 +45,12 @@
 			display: flex;
 			flex-direction: column;
 			flex: 1;
+
+			> div:first-of-type {
+				display: flex;
+				flex-direction: column;
+				flex: 1;
+			}
 		}
 	}
 }
--- a/js/src/dashboard/index.scss
+++ b/js/src/dashboard/index.scss
@@ -42,14 +42,11 @@
 		}
 
 		.components-card {
-			display: flex;
-			flex-direction: column;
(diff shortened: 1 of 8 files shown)


js/src/dashboard/index.scss
```diff
@@ -45,6 +45,12 @@
 			display: flex;
 			flex-direction: column;
 			flex: 1;
+
+			> div:first-of-type {
+				display: flex;
+				flex-direction: column;
+				flex: 1;
+			}
```
eason9487: A part of the style looks redundant. Maybe this would be enough?
```suggestion
			flex: 1;

			> div:first-of-type {
				display: flex;
				flex-direction: column;
			}
```

eason9487 (approved):
Thanks for the fix. LGTM. There is a part of style that could be simplified a bit but not blocking this PR.

puntope:
> Thanks for the fix. LGTM. There is a part of style that could be simplified a bit but not blocking this PR.

Thanks @eason9487 . I applied the suggestion and merged this PR

PR #1792, opened 2022-11-24
Bump WC/WP Components:  Fix AppRadioContentControl

puntope:

It fixes the AppRadioContentControl component by fixing the check prop. In my investigations, it's returning always checked `true`, overriding the checked prop in RadioControl component

1. Go to Onboarding Setup
2. In step 2 "Configure product listings" check that the form radio controls works as expected
3. Use develop branch as reference


eason9487 (approved):
Thanks for fixing this. LGTM.

PR #1793, opened 2022-11-25
Delete products in GMC when force delete a product or change catalog visibility to hidden

### LINKED ISSUE #1785: Force delete a product without trash it first will not delete the product from GMC
### Describe the bug:

This is a follow up issue from [#1728](https://github.com/woocommerce/google-listings-and-ads/pull/1728#trash-products), and the comments from the same issue as well: https://github.com/woocommerce/google-listings-and-ads/pull/1728#discussion_r1024115911.

The problem is, in method `find_delete_product_ids()` from `ProductRepository.php` it gets the products which their status are trash. But in the case where products get deleted by WP-CLI, API or making use of other plugins, it would not find out the products that is already deleted so the initial `delete_by_batch_request` is not called.

The second problem is, when setting a product's catalog visibility to `hidden`, its status still remains `publish`, so the method `find_delete_product_ids()` would not get any products since it only looking for products with status `trash`.

### Steps to reproduce:

1. Force delete a product using WP-CLI
    - `wp wc product delete <product_id> --force=1 --user=admin`
2. The `delete_products` action should be scheduled
    - `/wp-admin/admin.php?page=wc-status&tab=action-scheduler&s=delete_products&action=-1&action2=-1&orderby=schedule&order=desc&paged=1`
3. After the `delete_products` action is completed, go the Google Merchant Center dashboard
4. The product is still in the GMC.

#### Setting a product's catalog visibility to hidden

1. Set a product's catalog visibility to `hidden`
2. Follow the steps 2 and 4 above.
3. The catalog hidden product is still in the GMC.

### Expected behavior:

1. The product should also be removed from GMC.

### Actual behavior:

1. The product is still in GMC.

ianlin:
## Changes proposed in this Pull Request:

Closes #1785.

This PR fixes the problem which in some cases the deleted products are not actually deleted in GMC.

### When we force delete a product by WP-CLI

The problem is when we force delete a product it no longer exist in the database, so the filter in `process_items()` of `DeleteProducts` job would exclude this product before calling `delete_by_batch_request()`.

This is solved by moving the filtering logic from the method `process_items()` in `DeleteProducts` job to the code block where the retries are scheduled. So we will do the initial `delete_by_batch_request()` anyway, then when any internal errors occur returned by GMC we will be excluding any no longer existing products and ones that retried too many times.

However, there is still a case when we got internal error returned by GMC for the force deleted product. The force deleted product will be excluded before retrying so the data is still in GMC. I'm not sure if we need to take care of this case as it seems a bit of an edge case. For a complete data cleaning maybe we could have a daily or weekly job to remove the products that are not available in the local DB but still exists in GMC.

### When we change the product's catalog visibility to `hidden`.

In PR #1728 we added a condition that [only get products with status `trash` for deleting](https://github.com/woocommerce/google-listings-and-ads/pull/1728/files#diff-6934ab807d1707ca88f670c6059b54ae86ae9d22bf3e8151db14fe542c268ba9R167). However for the product that is changed to catalog visibility `hidden` its status is still `publish` so we are excluding this product to be deleted in GMC. This is solved by changing the condition to get all the default status plus status `trash`.

## Detailed test instructions:

1. Follow the steps to reproduce in #1785 for both **force delete a product** and **set a product's catalog visibility to hidded**
2. The product should also be deleted in GMC.
3. The infinite retry when deleting products problem that was solved in #1238 should still behave correctly.

## Changelog entry

> Fix - Products are not deleted in GMC when force delete a product or change catalog visibility to hidden.

### DIFF
--- a/tests/Unit/Product/ProductSyncerTest.php
+++ b/tests/Unit/Product/ProductSyncerTest.php
@@ -368,6 +368,135 @@ class ProductSyncerTest extends ContainerAwareUnitTest {
 		$this->assertEqualsCanonicalizing( $result_product_ids, $delete_ready_product_ids );
 	}
 
+	public function test_delete_by_batch_requests_should_delete_trashed_product() {
+		// $deleted_products:  products that were successfully synced and then deleted from Merchant Center
+		// $rejected_products: products that were synced but deleting them resulted in errors and were rejected by Google API
+		[ $deleted_products, $rejected_products ] = $this->create_multiple_simple_product_sets( 2, 1 );
+
+		$this->mock_google_service( $deleted_products, $rejected_products );
+
+		$products = array_merge( $deleted_products, $rejected_products );
+
+		// generate delete request entries
+		$product_entries = array_map(
+			function ( WC_Product $product ) {
+				return new BatchProductIDRequestEntry( $product->get_id(), $this->generate_google_id( $product ) );
+			},
+			$products
+		);
+
+		// first we mark all products as synced and
+		// set the product status to trash
+		array_walk(
+			$products,
+			function ( WC_Product $product ) {
+				$this->product_helper->mark_as_synced( $product, $this->generate_google_product_mock() );
+				$product->set_status( 'trash' );
+				$product->save();
+			}
+		);
+
+		$results = $this->product_syncer->delete_by_batch_requests( $product_entries );
+
+		$result_product_ids = array_map(
+			function ( $product_entry ) {
+				return $product_entry->get_wc_product_id();
+			},
+			$results->get_products()
+		);
+
+		$delete_ready_product_ids = array_keys( $deleted_products );
+
+		$this->assertEqualsCanonicalizing( $result_product_ids, $delete_ready_product_ids );
+		$this->assert_delete_results_are_valid( $results, $deleted_products, $rejected_products );
+	}
+
+	public function test_delete_by_batch_requests_should_delete_catalog_visibility_hidden_product() {
+		// $deleted_products:  products that were successfully synced and then deleted from Merchant Center
+		// $rejected_products: products that were synced but deleting them resulted in errors and were rejected by Google API
+		[ $deleted_products, $rejected_products ] = $this->create_multiple_simple_product_sets( 2, 1 );
+
+		$this->mock_google_service( $deleted_products, $rejected_products );
+
+		$products = array_merge( $deleted_products, $rejected_products );
+
+		// generate delete request entries
+		$product_entries = array_map(
+			function ( WC_Product $product ) {
+				return new BatchProductIDRequestEntry( $product->get_id(), $this->generate_google_id( $product ) );
+			},
+			$products
+		);
+
+		// first we mark all products as synced and
+		// set the product's catalog visibility to hidden
+		array_walk(
+			$products,
+			function ( WC_Product $product ) {
+				$this->product_helper->mark_as_synced( $product, $this->generate_google_product_mock() );
+				$product->set_catalog_visibility( 'hidden' );
+				$product->save();
+			}
+		);
+
+		$results = $this->product_syncer->delete_by_batch_requests( $product_entries );
+
+		$result_product_ids = array_map(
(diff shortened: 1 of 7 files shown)


tests/Unit/Product/ProductSyncerTest.php
```diff
@@ -325,6 +327,36 @@ public function test_delete_by_batch_requests_throws_exception_if_google_api_cal
...
+		// generate delete request entries
+		$product_entries = array_map(
+			function ( WC_Product $product ) {
+				return new BatchProductIDRequestEntry( $product->get_id(), $this->generate_google_id( $product ) );
+			},
+			$products
+		);
+
+		$this->product_syncer->delete_by_batch_requests( $product_entries );
+		$this->assertEquals( 0, did_action( 'woocommerce_gla_batch_retry_delete_products' ) );
```
mikkamp: Can we assert a little more, to confirm that something happened here? It seems we are just checking that it didn't trigger a retry, but it would be nice to also confirm it successfully processed those delete requests.

Maybe hooking into `woocommerce_gla_batch_deleted_products` and confirming the right ID's got deleted, or taking a look at the return value of `delete_by_batch_requests` and seeing if the BatchProductResponse contains the expected ID's.

mikkamp (approved):
Thanks the code changes look good and they are working as expected. I tested with both force deleting a product and making a product hidden from the catalog. The log entries show that it happened successfully, and I confirmed they were removed from the Google dashboard.

```
2022-11-25T08:16:59+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer::update_by_batch_requests Submitted 1 products:
[{"woocommerce_id":2587,"google_id":"online:en:GB:gla_2587"}]
2022-11-25T08:30:21+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer::delete_by_batch_requests Deleted 1 products:
[{"woocommerce_id":2587}]
2022-11-25T08:34:14+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\WCProductAdapter::map_product_categories Product category (ID: 1342): ["nisi"].
2022-11-25T08:34:15+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer::update_by_batch_requests Submitted 1 products:
[{"woocommerce_id":1342,"google_id":"online:en:GB:gla_1342"}]
2022-11-25T08:35:55+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\SyncerHooks::handle_update_products Deleting product (ID: 1342) from Google Merchant Center because it is not ready to be synced.
2022-11-25T08:35:55+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\SyncerHooks::handle_update_products Deleting product (ID: 1342) from Google Merchant Center because it is not ready to be synced.
2022-11-25T08:35:55+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\SyncerHooks::handle_update_products Deleting product (ID: 1342) from Google Merchant Center because it is not ready to be synced.
2022-11-25T08:36:58+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer::delete_by_batch_requests Deleted 1 products:
[{"woocommerce_id":1342}]
```

What I found odd was that the syncerhooks were triggered 3 times, based on the [code](https://github.com/woocommerce/google-listings-and-ads/blob/2.3.1/src/Product/SyncerHooks.php#L137) it looks like we expect it at least twice, not sure where the third one comes from, but it only scheduled one job to delete the product so looks like we handle that gracefully.

I did think the unit tests were missing some scenarios though so it would be nice to check that the products ID's actually match. I also didn't find any tests for the following scenarios:
1. Sync a product > Trash it (regular) > Confirm the delete batch included that product ID
2. Sync a product > Change it to catalog hidden > Confirm the delete batch included that product ID
2. Sync a product > Change it to draft > Confirm the delete batch included that product ID

If we would have had those tests we would have caught the errors sooner.

Issue #1787, opened 2022-11-22
PMax Assets - Stage 1


eason9487:
## Technical Goals

- Enable merchants to launch PMax campaigns based on their assets.
- Enable merchants to maintain their assets of PMax campaigns.

## Terminology

- **PMax**: Performance Max
   - https://www.google.com/retail/solutions/performance-max/
   - https://support.google.com/google-ads/answer/10724817
- **Assets**: PMax campaigns have some unique characteristics regarding assets.
   - Assets are grouped together in a collection called an AssetGroup, which is unique to PMax campaigns.
- **AssetGroup**: An AssetGroup is made from the images, logos, headlines, descriptions, etc when creating a PMax campaign.
   - Google Ads automatically mixes and matches these assets based on which Google Ad channel (YouTube, Gmail, Search, etc.) the ad is being served.
- **Final URL**: The page URL on merchant's website that people reach when they click Google's ad.
   - A page URL is required when creating an asset group.
   - On WooCommerce side, it refers to a post, page or taxonomy

## <a href="#tech-frontend" id="tech-frontend">Technical - Frontend</a>

The main scope is the campaign management page and the Dashboard page.

### 📌 <a href="#fe-campaign-step2" id="fe-campaign-step2">Add step 2 to the campaign management page</a>

- [x] Creation page [#1794](https://github.com/woocommerce/google-listings-and-ads/pull/1794)
- [x] Editing page [#1799](https://github.com/woocommerce/google-listings-and-ads/pull/1799)
- [x] The routing of editing page can direct to step 2 [#1799](https://github.com/woocommerce/google-listings-and-ads/pull/1799)
- [x] Add FAQs [#1881](https://github.com/woocommerce/google-listings-and-ads/pull/1881)
- [x] Add a basic section for placing the assets form [#1794](https://github.com/woocommerce/google-listings-and-ads/pull/1794)
- [x] Adjust the buttons in the footer for each step [#1794](https://github.com/woocommerce/google-listings-and-ads/pull/1794)

### 📌 <a href="#fe-misc" id="fe-misc">Assets form and related components</a>

- [x] Card for the Final URL [#1805](https://github.com/woocommerce/google-listings-and-ads/pull/1805)
   - [x] Input field for searching pages ([WC SelectControl](https://github.com/woocommerce/woocommerce/tree/6.9.0/packages/js/components/src/select-control)) [#1805](https://github.com/woocommerce/google-listings-and-ads/pull/1805)
[...]

PR #1007, opened 2021-09-10
Fix issues and react warnings for verification code components

eason9487:
### Changes proposed in this Pull Request:

Closes #1002, #1003, #1004

- https://github.com/woocommerce/google-listings-and-ads/commit/305c21c4e0fa92ea25bd3f301a7be3b1bfe3777d - Fix that the first verification code input isn't cleared after reset.
- https://github.com/woocommerce/google-listings-and-ads/commit/2188f6075aac882330a7f327a5ed10d5325fac9d - Fix that backspace doson't remove selected digit in the `<VerificationCodeControl>`.
- https://github.com/woocommerce/google-listings-and-ads/commit/49371b518e96d4babc1d135d26e14ff06a9d55e0 - Fix the warning of updating state on an unmounted component.
- https://github.com/woocommerce/google-listings-and-ads/commit/8f8734125bfa81d4c4e3356315583cf6f14bbdc3 - Fix the warning of missing unique key in the `<AppButton>`.

### Additional notes

- [x] 📝  Revert https://github.com/woocommerce/google-listings-and-ads/commit/f605132381c0ff3494f8b4305c575381e2fa5c8c to removed `<DemoVerifyPhoneNumberCard>` before merging this PR.

#### #1002 - Clear input after reset

Kapture 2021-09-10 at 15 19 51

#### #1003 - Delete selected value when pressing backspace

Kapture 2021-09-10 at 15 16 42

#### #1004 - Fix react warnings

Kapture 2021-09-10 at 15 54 21

### Detailed test instructions:

1. Go to the Setting page and there's a demo component for testing
2. Follow the replication steps in each issue to check if they were fixed
    - #1002
    - #1003
    - #1004

PR #1794, opened 2022-11-25
PMax Assets: Add step 2 to the campaign creation page

eason9487:

This PR implements parts of 📌 [Add step 2 to the campaign management page](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-campaign-step2) in #1787.

- Add a basic section for placing the assets form
- Add step 2 to the campaign creation page
- Adjust the buttons in the footer for each step
- Add `useAdaptiveFormContext` to make `AdaptiveForm`'s context also apply compatible fixes
- Add form submission related states to `AdaptiveFormContext`

### Screenshots:

https://user-images.githubusercontent.com/17420811/204020795-94cb4d0e-28f7-4434-a362-59502f3fcea0.mp4

2. Check if the stepper works correctly.
3. Create campaigns individually using "Launch paid campaign" and "Skip adding assets" buttons.

#### 💡 About the changes in `AdaptiveForm`

- I renamed `formProps` to `formContext` since it should be more precise.
- Since WC 6.9, it added the ability to obtain `Form` context via `useFormContext` hook. IMO, it's helpful to avoid prop drilling, especially since GLA codebase already has many layers of prop drilling happening on `formProps`. But this also needs 9fdd5f31983c92a211d3d640cb82c4bce2cacb6d to apply compatible fixes.
- To control the `disabled` and `loading` states of submission button, we did not have a clear way to separate form submission processing from button states. So usually, they have to be placed together (within the same component), which made data and states at different abstraction levels mixed together, and also made managing data less flexible. I added
f0563372de425278263e400e860a5fa651abce9a to explore a more proper way to use the form in WC plugin development.
   - The [processing of form submission](https://github.com/woocommerce/google-listings-and-ads/blob/28899e31daf53e79105808fc16279ed01d9b3974/js/src/setup-mc/setup-stepper/setup-paid-ads/setup-paid-ads.js#L82-L115) was placed in the last step of onboarding flow rather than in the `onSubmit` callback.
   - [The `saving` state](https://github.com/woocommerce/google-listings-and-ads/blob/28899e31daf53e79105808fc16279ed01d9b3974/js/src/components/free-listings/setup-free-listings/index.js#L225) needs to be passed down to children.
   - [Mutate `formProps` to let `isSubmitting` hitchhike](https://github.com/woocommerce/google-listings-and-ads/blob/28899e31daf53e79105808fc16279ed01d9b3974/js/src/setup-ads/setup-ads-form.js#L96) to a child in several layers deep.


puntope:
>  IMO, it's helpful to avoid prop drilling, especially since GLA codebase already has many layers of prop drilling

Agree

js/src/components/adaptive-form/adaptive-form-context.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * AdaptiveForm's context hook.
+ *
+ * @return {AdaptiveFormContext} AdaptiveForm's context.
+ */
+export function useAdaptiveFormContext() {
+	const adaptiveFormContext = useContext( AdaptiveFormContext );
+
+	return adaptiveFormContext;
+}
```
puntope, line 54: I agree with the use of context for this case, but I found this hook `useAdaptiveFormContext` a bit redundant tho.

Why not to have a context folder with the contexts (this one for now) and just call useContext() in the consumer side?

js/src/components/paid-ads/asset-group/asset-group.js
```diff
@@ -0,0 +1,75 @@
...
+						! isValidForm ||
+						isSubmitted ||
+						currentAction === ACTION_SKIP_ASSETS
+					}
+					loading={
+						isSubmitting && currentAction === ACTION_COMPLETE
+					}
+					onClick={ handleSubmit }
+				>
+					{ __( 'Launch paid campaign', 'google-listings-and-ads' ) }
```
puntope, line 70: 💅 
Maybe this is because of `// TODO: Change to only validate campaign data.` But if not... with a little tweak in the disabled logic we can have a potential refactor here:

i.e.
```
<StepFooterButton action={ACTION_SKIP_ASSETS} isTerciary>{ __( 'Skip adding assets', 'google-listings-and-ads' ) }</StepFooterButton>
<StepFooterButton action={ACTION_COMPLETE} isPrimary>{ __( 'Launch paid campaign', 'google-listings-and-ads' ) }</StepFooterButton>
```

```
[...]
data-action={ action }
disabled={
						! isValidForm ||
						isSubmitted ||
						currentAction !== action
					}
[...]
```

js/src/components/paid-ads/asset-group/asset-group-section.js
```diff
@@ -0,0 +1,57 @@
...
+			className=".gla-asset-group-section"
+			title={ createInterpolateElement(
+				__(
+					'Add assets <gray>(Optional)</gray>',
+					'google-listings-and-ads'
+				),
+				{
+					gray: (
+						<span className="gla-asset-group-section__optional-label" />
+					),
```
puntope: 💅 

Maybe we could change this from `gray` to `optional`

Seems more consistent for me and also if the color of the optional label changes in the future it will maintain consistency, whereas gray won't.

js/src/components/paid-ads/asset-group/asset-group-section.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Section from '.~/wcdl/section';
+import AppDocumentationLink from '.~/components/app-documentation-link';
+import './asset-group-section.scss';
+
+/**
+ * Renders the form content for managing an asset group of a campaign with Section UI.
```
puntope, line 15: 💅 

Feels for me that the `form content for managing asset group of a campaign` will be handed in  another (form?) component (where now is a final url placeholder).

I see this more like a `Renders the section UI for managing asset group of a campaign`

js/src/components/adaptive-form/adaptive-form-context.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * External dependencies
+ */
+import { createContext, useContext } from '@wordpress/element';
+
+/**
+ * @typedef {Object} InputProps
+ * @property {*} value Form value.
+ * @property {boolean} checked Form value converted to boolean.
+ * @property {*} selected Form value.
```
puntope, line 14: This I guess it should be `Selected Form value`. Instead of just `Form value.`

js/src/components/adaptive-form/adaptive-form-context.js
```diff
@@ -0,0 +1,50 @@
...
+ * @property {(value: Event<HTMLInputElement> | *) => void} onChange Function to handle onChange event.
```
puntope, line 15: `(value: Event<HTMLInputElement> | *)` is the same as just `(*)`

js/src/components/adaptive-form/adaptive-form-context.js
```diff
@@ -0,0 +1,50 @@
...
+ * @property {() => void} onBlur Function to handle onBlur event.
+ * @property {'has-error' | undefined} className 'has-error' if the form value is invalid and marked as touched. `undefined` otherwise.
+ * @property {string | undefined | null} help The corresponding value in form `errors` if the form value is marked as touched. `null` otherwise.
```
puntope, line 18: ❓ 

So In which cases it is `undefined`?

js/src/components/adaptive-form/adaptive-form-context.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @typedef {Object} AdaptiveFormContext
+ * @property {Object} values Form values.
+ * @property {Object} errors Object with key-value pairs representing errors for form values. Empty object if no errors.
+ * @property {Object} touched Object with key-value pairs representing form values is touched.
+ * @property {boolean} isValidForm `true` if form values pass the validation.
+ * @property {boolean} isDirty `true` after any of the form values is modified.
+ * @property {(name: string, value: *) => void} setValue Function to set a form value.
+ * @property {(name: string) => InputProps} getInputProps Function to get the corresponding input props by the name of a form value.
```
puntope: The part of `by the name of a form value` seems unclear for me. As far as I understand we have a set of form inputs with props (let's say name, and age). So if we do  `getInputProps('name')` we are getting the props for the `name` input which has a value inside among other props. So i would rephrase this as. 

`Function to get the corresponding InputProps object using the input name as a key.`

js/src/components/adaptive-form/adaptive-form-context.js
```diff
@@ -0,0 +1,50 @@
...
+ * @property {() => Promise<Object>} handleSubmit Function to trigger form submission.
+ * @property {(initialValues: Object) => void} resetForm Function to reset form with given initial values.
+ * @property {(currentTouched: Object) => void} setTouched Function to update the `touched` state.
```
puntope: Is not clear for me how it work this. Do we send the props as true  in `currentTouched`? Or what exactly should be inside `currentTouched`?

js/src/components/adaptive-form/adaptive-form.js
```diff
@@ -130,7 +163,30 @@ function AdaptiveForm( { children, ...props }, ref ) {
 					setImmediate( () => setDelegation( batchQueue.shift() ) );
 				}
 
-				return children( formProps );
+				// Enhancements related.
+				formContext.handleSubmit = function ( event ) {
+					adapterRef.current.submitter = event.currentTarget;
+					return handleSubmit.call( this, event );
+				};
```
puntope: Would be nice to explain this chunk a little bit more. Instead of `// Enhancements related.` that is a bit generic, I would add something more descriptive.

js/src/components/adaptive-form/adaptive-form.js
```diff
@@ -56,9 +63,35 @@ function AdaptiveForm( { children, ...props }, ref ) {
...
+			setSubmission( [ true, false ] );
+
+			await onSubmit.call( this, values, adapterRef.current.submitter );
+
+			if ( isMounted() ) {
+				setSubmission( [ false, true ] );
+				adapterRef.current.submitter = null;
+			}
+		};
+	}
```
puntope: It's unclear for me what this chunk of code really does. 

- Why submission is a tuple of two values? What those values are doing?
- Why we are attaching the `submission` state to the `isSubmitting` state?

js/src/components/adaptive-form/adaptive-form.js
```diff
@@ -56,9 +63,35 @@ function AdaptiveForm( { children, ...props }, ref ) {
...
 	} ) );
 
+	// Enhancements related.
+	// Additional enhancements to make AdaptiveForm have more useful or reusable features.
+	// It could also be the playground of the practical instances before pushing them upstream.
+	const isMounted = useIsMounted();
+	const [ submission, setSubmission ] = useState( [ false, false ] );
+	const [ isSubmitting, isSubmitted ] = submission;
+
+	if ( onSubmit ) {
```
puntope, line 83: ❓ 
In which circumstances we don't need a submit function in the AdaptiveForm?

puntope: (Asking this because is marked as an optional and also we have a check here for it's presence)

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -0,0 +1,68 @@
...
+							<AppDocumentationLink
+								context="create-ads"
+								linkId="see-what-ads-look-like"
+								href="https://support.google.com/google-ads/answer/6275294"
+							/>
+						),
+					}
+				) }
+			/>
+			<CreateCampaignFormContent formProps={ formContext } />
```
puntope, line 56: Since we changed the names from `CreatePaidAdsCampaignForm` to `AdsCampaign`. WDYT of changing this to `AdsCampaignForm` or `AdsCampaignFormContent`?

puntope: Same for the SCSS

puntope: P.S. Reverting kinda the previous name from `AdsCampaign` to `CreateAdsCampaign` also ok for me.

puntope: Actually, thinking about it, I think it would be better `CreatePaidAdsCampaign`. So then we would have `CreatePaidAdsCampaign`, `EditPaidAdsCampaignFormContent` and `CreatePaidAdsCampaignFormContent`. In any case, I miss some consistency here.

puntope:
Howdy @eason9487 

Thanks for working on this.

I left several comments 💅 / ❓  to be discussed in regards code. No blocker but yes would be nice to discuss to understand better the code. 

Besides that, when I clicked on the Ad Campaign button in the dashboard. It still show me the old version (without the Boost Your Campaign Step)  in case my ads account is not connected.

js/src/components/paid-ads/asset-group/asset-group-section.js
eason9487: Thanks for the suggestion. Renamed in https://github.com/woocommerce/google-listings-and-ads/pull/1794/commits/f6e5e9ff668fa1e84c473a93c2e0bf1df2520409

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: Thanks for catching this. Just found the `currentTouched` naming was not correct and another type was missed. Fixed in
https://github.com/woocommerce/google-listings-and-ads/pull/1794/commits/e0b8dc68d52acd78ca26356215a79285f38bf9d4 with some other adjustments to make it clearer.

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: Using "input name" might be confusing since the `name` should be one of the names of the given form `values`. I rephrased it in https://github.com/woocommerce/google-listings-and-ads/pull/1794/commits/c9f90ec2192e4c72e93e55059fb15c6a2296548a. Is this clear now?

js/src/components/adaptive-form/adaptive-form.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/1794/commits/7ccbc8f8e5568a6f74bbd15d337fbae286ca05f6.

js/src/components/adaptive-form/adaptive-form.js
eason9487: Adjusted and added in https://github.com/woocommerce/google-listings-and-ads/pull/1794/commits/7ccbc8f8e5568a6f74bbd15d337fbae286ca05f6.

> What those values are doing?

I didn't add the description of these two state values since `AdaptiveFormContext` already has [their JSDocs](https://github.com/woocommerce/google-listings-and-ads/blob/7ccbc8f8e5568a6f74bbd15d337fbae286ca05f6/js/src/components/adaptive-form/adaptive-form-context.js#L18-L21). Let me know if you think it should still have copies of them here.

> * Why we are attaching the `submission` state to the `isSubmitting` state?

I'm not sure what this question means, could you help clarify?

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: > I agree with the use of context for this case, but I found this hook useAdaptiveFormContext a bit redundant tho.

Since there will be many form-related components that will use this context, there will be many such code fragments repeated across them.

```js
import { useContext } from '@wordpress/element';
import { AdaptiveFormContext } from '.~/components/adaptive-form';

function InputComponent() {
	const formContext = useContext( AdaptiveFormContext );
}
```

So there is a common pattern to avoid the redundancy above.

```js
import { useAdaptiveFormContext } from '.~/components/adaptive-form';

function InputComponent() {
}
```

Comparing these two uses, should the following one be less redundant?

> Why not to have a context folder with the contexts (this one for now) and just call useContext() in the consumer side?

There are usually very few situations where React Context will be used, so I supposed there is still no need for a dedicated directory to centralize the only context at present.

js/src/components/paid-ads/asset-group/asset-group.js
eason9487: Do you mean a code like this?

```js
export default function AssetGroup() {
	const { isValidForm, handleSubmit, adapter } = useAdaptiveFormContext();
	const { isSubmitting, isSubmitted, submitter } = adapter;

	function StepFooterButton( { action, ...props } ) {
		return (
			<AppButton
				{ ...props }
				loading={ isSubmitting && currentAction === action }
				data-action={ action }
				onClick={ handleSubmit }
			/>
		);
	}

	return (
		<StepContent>
			[ ... omitted ]
			<StepContentFooter>
				<StepFooterButton
					isTertiary
					action={ ACTION_SKIP_ASSETS }
					disabled={
						// TODO: Change to only validate campaign data.
						! isValidForm ||
						isSubmitted ||
						currentAction === ACTION_COMPLETE
					}
				>
					{ __( 'Skip adding assets', 'google-listings-and-ads' ) }
				</StepFooterButton>
				<StepFooterButton
					isPrimary
					action={ ACTION_COMPLETE }
					disabled={
						! isValidForm ||
						isSubmitted ||
						currentAction === ACTION_SKIP_ASSETS
					}
				>
					{ __( 'Launch paid campaign', 'google-listings-and-ads' ) }
				</StepFooterButton>
			</StepContentFooter>
		</StepContent>
	);
}
```

If so, however, I'm not pretty sure it's worth it. Because button implementations are instead split into shattered code fragments, which reduces readability and doesn't improve maintainability much.

js/src/components/paid-ads/asset-group/asset-group-section.js
eason9487: This component is at the same hierarchy level as several components in the codebase named `*FormContent`. However, the form "content" is usually split into several components and combined with some other non-form components at different hierarchy levels to composite the final form content, so I'm not keen on having a specific component named `FooFormContent`. The "form content" is used here in place of the `FormContent` part that was originally included in its naming.

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: Thanks for the suggestion. The type declaration here is mainly to make it explicitly clear that this handler is also compatible with the native `onChange` event/prop.

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: When an input field is marked as touched and has no errors, the value of `help` will be `undefined`, which come from the corresponding value in form `errors`.
Ref: https://github.com/woocommerce/woocommerce/blob/6.9.0/packages/js/components/src/form/form.tsx#L258

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: It's the same value as `value`. (Its type declaration above is even wrong.)
https://github.com/woocommerce/woocommerce/blob/6.9.0/packages/js/components/src/form/form.tsx#L249-L251

js/src/components/adaptive-form/adaptive-form.js
eason9487: It is optional in order to keep it consistent with the existing interface of WC's `Form` component. Ref:
- https://github.com/woocommerce/woocommerce/blob/6.9.0/packages/js/components/src/form/form.tsx#L43
- https://github.com/woocommerce/woocommerce/blob/6.9.0/packages/js/components/src/form/form.tsx#L82
- https://github.com/woocommerce/woocommerce/blob/6.9.0/packages/js/components/src/form/form.tsx#L229-L231

js/src/components/paid-ads/ads-campaign.js
eason9487: This component will be also used to replace the `EditPaidAdsCampaignForm` in [edit-paid-ads-campaign/index.js](https://github.com/woocommerce/google-listings-and-ads/blob/7ccbc8f8e5568a6f74bbd15d337fbae286ca05f6/js/src/pages/edit-paid-ads-campaign/index.js#L69) in a subsequent PR. I initially named it `ManageAdsCampaign`, later, considering that most of the component names in [paid-ads](https://github.com/woocommerce/google-listings-and-ads/tree/7ccbc8f8e5568a6f74bbd15d337fbae286ca05f6/js/src/components/paid-ads) don't start with a verb, consistency considerations here led to the name `AdsCampaign`.

In addition, there is no clear consistency in the naming of this aspect in GLA codebase, such as `AttributeMappingRuleModal` doesn't start with a verb either.

And the reason why there is no component name with "FormContent" in this PR is described [here](https://github.com/woocommerce/google-listings-and-ads/pull/1794#discussion_r1035576050). Borrowing from the same example again, I doubt a name like `AttributeMappingFormRuleModal` or `AttributeMappingRuleFormContentModal`, would be more appropriate.

eason9487:

> Besides that, when I clicked on the Ad Campaign button in the dashboard. It still show me the old version (without the Boost Your Campaign Step) in case my ads account is not connected.

I think you meant the Set up paid campaign page (`/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads`). The assets step was not intended to be added to this page, so showing the original version is as expected.

---

I believe I have addressed or replied to all your review comments. Could you help with a new round of code reviews?

js/src/components/paid-ads/asset-group/asset-group.js
puntope: Something like this. Thats why I asked about that TODO thing

```jsx
export default function AssetGroup() {
	const { isValidForm, handleSubmit, adapter } = useAdaptiveFormContext();
	const { isSubmitting, isSubmitted, submitter } = adapter;

	function StepFooterButton( { action, ...props } ) {
		return (
			<AppButton
				{ ...props }
				loading={ isSubmitting && currentAction === action }
				data-action={ action }
				onClick={ handleSubmit }
                                disabled={
						! isValidForm ||
						isSubmitted ||
						currentAction !== action
					}
			/>
		);
	}

	return (
		<StepContent>
			[ ... omitted ]
			<StepContentFooter>
				<StepFooterButton
					isTertiary
					action={ ACTION_SKIP_ASSETS }
				>
					{ __( 'Skip adding assets', 'google-listings-and-ads' ) }
				</StepFooterButton>
				<StepFooterButton
					isPrimary
					action={ ACTION_COMPLETE }
				>
					{ __( 'Launch paid campaign', 'google-listings-and-ads' ) }
				</StepFooterButton>
			</StepContentFooter>
		</StepContent>
	);
}
```

js/src/components/adaptive-form/adaptive-form-context.js
puntope: Maybe worth it to at that case in the property docs

js/src/components/adaptive-form/adaptive-form.js
```diff
@@ -63,10 +67,13 @@ function AdaptiveForm( { onSubmit, children, ...props }, ref ) {
...
-	// Enhancements related.
-	// Additional enhancements to make AdaptiveForm have more useful or reusable features.
-	// It could also be the playground of the practical instances before pushing them upstream.
+	/* === Start of enhancement-related codes === */
+
 	const isMounted = useIsMounted();
+
+	// Add `isSubmitting` and `isSubmitted` states for facilitating across multiple layers of
+	// component controlling, such as disabling inputs or buttons. They are composed in a tuple
+	// to reduce the number of triggered re-rendering.
```
puntope: Ok, reading it now again I understand now this code.

The first element in the tuple is the `isSubmitting` state and the second is the `isSubmitted ` state. The first time I didn't read it properly. I assumed somehow the second line was another different state (maybe because I read it fast). 

I think it will be clearer naming the submission statuses instead of using tuples. 

```jsx
const [ submission, setSubmission ] = useState( null );
const isSubmitting = submission === SUBMITTING ;
const isSubmitted = submission === SUBMITTED ;

if ( onSubmit ) {
		props.onSubmit = async function ( values ) {
			setSubmission( SUBMITTING );
			await onSubmit.call( this, values, adapterRef.current.submitter );
			if ( isMounted() ) {
				setSubmission( SUBMITTED );
				adapterRef.current.submitter = null;
			}
		};
	}
```

puntope: I still don't get why `isMounted` is related to the submission of the form. Maybe I'm missing some of the internals tho...

puntope: My bet is that if user clicks submit, then the form is on isSubmitting, If the user leves this page between that moment and the completion of the onSubmit call , the component is unMounted. So `setSubmission( SUBMITTED );`should not be called. This isMounted() prevents that case. Is that right?

js/src/components/paid-ads/ads-campaign.js
puntope: Let's see then when it's all finish. Hard to discuss without all the components in the folder and knowing what each component finally contains.

Not sure neither whats the relationship with `AttributeMappingRuleModal` here.

js/src/components/adaptive-form/adaptive-form-context.js
puntope: In my IDE at least it marks the new type a s invalid:

I guess we can use

` * @property {React.Dispatch<React.SetStateAction<Object>>}`

puntope (approved):
✅  LGTM

Thanks @eason9487  for further explanations and adjusting the documentation.

I left some comments In regards some gaps understanding 100% the code and also with some discussions.

Besides that, I expected some unit tests, but I guess they are planned to be added in next PRs?

js/src/components/paid-ads/asset-group/asset-group.js
eason9487: Thanks for the explanation.

> Maybe this is because of `// TODO: Change to only validate campaign data.` [...]

> Thats why I asked about that TODO thing

Sorry, I didn't fully understand what you meant originally.

Yes, since that submit button is used to create a new campaign without assets, it will be changed to validate `countryCodes` and `amount` only in a later PR, along with changing the validation in *utils/paid-ads/validateForm.js* to also check asset data.

> But if not... with a little tweak in the disabled logic we can have a potential refactor here: [...]

`currentAction === ACTION_COMPLETE` and `currentAction === ACTION_SKIP_ASSETS` are not exactly logically equivalent to `currentAction !== action` since the `currentAction` could be `undefined`, but using `Boolean( currentAction && currentAction !== action )` could solve this.

---

Taken together, the code could be like this:

```js
export default function AssetGroup() {
	const { isValidForm, handleSubmit, adapter } = useAdaptiveFormContext();
	const { isSubmitting, isSubmitted, submitter } = adapter;

	function StepFooterButton( { action, disabledByValidation, ...props } ) {
		return (
			<AppButton
				{ ...props }
				loading={ isSubmitting && currentAction === action }
				disabled={
					disabledByValidation ||
					isSubmitted ||
					Boolean( currentAction && currentAction !== action )
				}
				data-action={ action }
				onClick={ handleSubmit }
			/>
		);
	}

	return (
		<StepContent>
			[ ... omitted ]
			<StepContentFooter>
				<StepFooterButton
					isTertiary
					action={ ACTION_SKIP_ASSETS }
					disabledByValidation={
						// TODO: Change to only validate campaign data.
						! isValidForm
					}
				>
					{ __( 'Skip adding assets', 'google-listings-and-ads' ) }
				</StepFooterButton>
				<StepFooterButton
					isPrimary
					action={ ACTION_COMPLETE }
					disabledByValidation={ ! isValidForm }
				>
					{ __( 'Launch paid campaign', 'google-listings-and-ads' ) }
				</StepFooterButton>
			</StepContentFooter>
		</StepContent>
	);
}
```

Visually, the shattered code fragments in implementation are a little better, but the number of props being split is still not better, and there is even another layer of logic to make the final `disabled`.

However, the current implementation is concentrated but looks trivial, so I don't think any of it is particularly good. If you prefer the splitting one, I'd be happy to adjust it.

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: > Maybe worth it to at that case in the property docs

I tried this one but looks a bit redundant:
```
The corresponding value in form `errors` if the form value is marked as touched, so whether it's a string or `undefined` depends on the corresponding value in the current form `errors`. `null` otherwise.
```
Do you have any suggestions?

js/src/components/adaptive-form/adaptive-form.js
eason9487: > I think it will be clearer naming the submission statuses instead of using tuples.

Thanks for the nice suggestion. 👍  I changed to use it in https://github.com/woocommerce/google-listings-and-ads/pull/1794/commits/1c80e7d8761c7138c66bd1664abc0b9c29be574a.

> My bet is that if user clicks submit, then the form is on isSubmitting, If the user leves this page between that moment and the completion of the onSubmit call , the component is unMounted. So `setSubmission( SUBMITTED );`should not be called. This isMounted() prevents that case. Is that right?

Yes, It's indeed to deal with this situation. Otherwise, the following warning would appear even though it won't cause an actual memory problem here.

image

js/src/components/adaptive-form/adaptive-form-context.js
eason9487: Thanks for the better suggestion. Updated in https://github.com/woocommerce/google-listings-and-ads/pull/1794/commits/b055ba1b05f6634783e4fd8b57d54a6e0d806b01.

js/src/components/adaptive-form/adaptive-form.js
eason9487: A related discussion about `useIsMounted`: https://github.com/woocommerce/google-listings-and-ads/pull/1007#discussion_r708838891

eason9487:
Thanks for the second round of reviews, @puntope.

I will be merging this PR later. If there is further discussion of these comments, I will address them in a separate PR.

- https://github.com/woocommerce/google-listings-and-ads/pull/1794#discussion_r1035766262
- https://github.com/woocommerce/google-listings-and-ads/pull/1794#discussion_r1035777865

> Besides that, I expected some unit tests, but I guess they are planned to be added in next PRs?

I will gradually record tests that could be added. After the main goal is completed, I could come back to make them up if there is spare capacity. See: 📌 [Add tests](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-tests)

PR #1799, opened 2022-12-01
PMax Assets: Add step 2 to the campaign editing page

eason9487:
### Changes proposed in this Pull Request:

This PR implements parts of 📌 [Add step 2 to the campaign management page](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-campaign-step2) in #1787.

- Add step 2 to the campaign editing page and the support of starting from step 2
- Make `AdsCampaign` component support editing a campaign
- Add some editing logic to `AssetGroup` component

It also adds a check to facilitate debugging.

- Add a check for the existence of a context provider to `useAdaptiveFormContext`

### Screenshots:

https://user-images.githubusercontent.com/17420811/205005011-a6c2db43-9da9-4829-bada-a9c739e22755.mp4

### Detailed test instructions:

1. Edit a campaign via the edit button on its most right side.
   - When entering the editing page, it should start with the first step.
   - Check if the stepper works correctly.
   - Update campaigns individually using "Save changes" and "Skip adding assets" buttons.
1. Go back to the Dashboard page.
1. Edit a campaign via the edit button on its "Assets" column.
   - When entering the editing page, it should start with the second step.
   - Update campaigns individually using "Save changes" and "Skip adding assets" buttons.

PR #1803, opened 2022-12-05
Internalize campaign creation and editing to `AdsCampaign` to eliminate the associated *FormContent components

eason9487:

It's related to https://github.com/woocommerce/google-listings-and-ads/pull/1799#discussion_r1038754145

- Internalize campaign creation and editing to `AdsCampaign`, so the associated **FormContent* components could be removed
   - `CreateCampaignFormContent`
   - `EditPaidAdsCampaignFormContent`
- Remove the almost duplicate component `setup-ads/ads-stepper/create-campaign/index.js` and replace with `AdsCampaign`
- Merge `setup-ads/setup-ads-form.js` and `setup-ads/setup-ads-form-content.js` because the latter only combined two components and was not reused.

This way, it could reduce some naming ambiguities and duplicate code for campaign-related components.

1. Go to the related pages
   - Step 4 of onboarding - `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-mc`
   - Ads setup - `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads`
   - Campaign creation - `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Fcampaigns%2Fcreate`
   - Campaign editing - `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Fcampaigns%2Fedit&programId={CAMPAIGN_ID}`
2. Check if the UI and functionality are the same
   - 💡 The `margin-right` of the currency label is different because it was missed to apply the consistent style while developing the Streamlined FL+Paid Ads project.
   2022-12-05 下午2 08 57
   - 💡 The `CampaignPreviewCard` is expected to show on all campaign creation and editing pages except for the onboarding.
   image


js/src/components/paid-ads/ads-campaign.js
```diff
@@ -60,24 +77,27 @@ export default function AdsCampaign( { campaign, onContinue } ) {
...
-				<CreateCampaignFormContent formProps={ formContext } />
-			) : (
-				<EditPaidAdsCampaignFormContent
-					formProps={ formContext }
-					allowMultiple={ campaign.allowMultiple }
-				/>
-			) }
+			<AudienceSection
+				disabled={ ! isCreation }
+				multiple={ isCreation || campaign.allowMultiple }
```
puntope, line 90: 💅  Since `! isCreation` disables the component. I guess multiple will be always true when the component is enabled.

eason9487: When `allowMultiple` is false, it also renders another selector which matches the wanted visual result.

eason9487:
Hi @puntope, sorry I missed mentioning that the `CampaignPreviewCard` is expected to show on all campaign creation and editing pages except for the onboarding. So it's the second difference compared to the previous revision.

image

(I also added this to the main content of this PR.)

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -26,18 +28,33 @@ import EditPaidAdsCampaignFormContent from '.~/components/paid-ads/edit-paid-ads
...
  * a context provider component (`AdaptiveForm`) to existing in its parents.
  *
- * @fires gla_documentation_link_click with `{ context: 'create-ads', link_id: 'see-what-ads-look-like', href: 'https://support.google.com/google-ads/answer/6275294' }`
- * @fires gla_documentation_link_click with `{ context: 'edit-ads', link_id: 'see-what-ads-look-like', href: 'https://support.google.com/google-ads/answer/6275294' }`
+ * @fires gla_documentation_link_click with `{ context: 'create-ads' | 'edit-ads' | 'setup-ads', link_id: 'see-what-ads-look-like', href: 'https://support.google.com/google-ads/answer/6275294' }`
  *
  * @param {Object} props React props.
  * @param {Campaign} [props.campaign] Campaign data to be edited. If not provided, this component will show campaign creation UI.
  * @param {() => void} props.onContinue Callback called once continue button is clicked.
+ * @param {'create-ads'|'edit-ads'|'setup-ads'} props.trackingContext A context indicating which page this component is used on. This will be the value of `context` in the track event properties.
```
puntope, line 36: 💅 

An alternative solution for knowing the context could be to use the paths?

puntope (approved):
✅  LGTM

Thanks for this nice refactor.

I tested and compared both versions. They look the same for me.

js/src/components/paid-ads/ads-campaign.js
eason9487: Ideally, the `AdsCampaign` component is as agnostic as possible about which pages it would be used. So using the `trackingContext` prop given by the consumer sides of this component would be closer to the purpose.

PR #1804, opened 2022-12-05
Resizing Images Assets Pmax assets

jorgemd24:

This PR resizes the images to match the Ads API requirements. There are three types of images:

- MARKETING_IMAGE: Landscape (1.91:1) 1200 x 628 recommended; 600 x 314 min.
- SQUARE_MARKETING_IMAGE: (1:1) 1200 x 1200 recommended; 300 x 300 min.
- LOGO: (1:1) 1200 x 1200 recommended; 128 x 128 min; 

The PR will try to resize the image to the closest size to the recommended size. For instance, for MARKETING_IMAGE, the resize target will be 1200x628px. Furthermore, an image can be resized if the width **and** height are bigger than the minimum sizes, otherwise, it will be excluded.

### Testings with posts (products/pages/posts/etc).

1. Make a request to GET `wc/gla/assets/suggestions?id=YOUR_POST_ID&type=post`
2. If the images are bigger than the minimum requirements mentioned above, the response should contain the images/galleries that are attached to the post with the closest size to the recommended size.
4. Open your uploads folder and verify that the images created have the correct sizes, if you make the same request the images should not be created again.

### Testings with terms.

1. Go to any taxonomy that is created on your website, for example, Product->Categories.
2. Go to edit one of the terms. For example Hoodies. Get the ID term, it can be found on the URL as the param `tag_id`
3. Make a request to GET `wc/gla/assets/suggestions?id=YOUR_TERM_ID&type=term`
4. Follow steps 3 & 4 from the section of "Testings with posts"

### DIFF
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -164,10 +164,10 @@ class AssetSuggestionsService implements Service {
 
 		if ( $post->post_type === 'product' || $post->post_type === 'product_variation' ) {
 			$product         = $this->wc->maybe_get_product( $id );
-			$attachments_ids = [ ...$attachments_ids, ...$product->get_gallery_image_ids(), $product->get_image_id() ];
+			$attachments_ids = [ ...$attachments_ids, ...$product->get_gallery_image_ids() ];
 		}
 
-		$attachments_ids  = array_slice( [ ...$attachments_ids, ...$this->get_gallery_images_ids( $id ) ], 0, self::DEFAULT_MAXIMUM_MARKETING_IMAGES );
+		$attachments_ids  = array_slice( [ ...$attachments_ids, ...$this->get_gallery_images_ids( $id ), get_post_thumbnail_id( $id ) ], 0, self::DEFAULT_MAXIMUM_MARKETING_IMAGES );
 		$marketing_images = $this->get_url_attachments_by_ids( $attachments_ids );
 		$long_headline    = get_bloginfo( 'name' ) . ': ' . $post->post_title;
 
@@ -208,12 +208,7 @@ class AssetSuggestionsService implements Service {
 		$attachments_ids            = [];
 
 		foreach ( $posts_assigned_to_term as $post ) {
-
-			if ( $post->post_type === 'product' ) {
-				$product           = $this->wc->maybe_get_product( $post->ID );
-				$attachments_ids[] = $product->get_image_id();
-			}
-
+			$attachments_ids[]            = get_post_thumbnail_id( $post->ID );
 			$posts_ids_assigned_to_term[] = $post->ID;
 		}
 
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -439,6 +439,24 @@ class AssetSuggestionsService implements Service {
 	}
 
 	/**
+	 * Filter for the posts_where hook, adds WHERE clause to search
+	 * for the 'search_title' parameter in the post titles (when present).
+	 *
+	 * @param string   $where The WHERE clause of the query.
+	 * @param WP_Query $wp_query The WP_Query instance (passed by reference).
+	 *
+	 * @return string The updated WHERE clause.
+	 */
+	public function title_filter( string $where, WP_Query $wp_query ): string {
+		$search_title = $wp_query->get( 'search_title' );
+		if ( $search_title ) {
+			$title_search = '%' . $this->wpdb->esc_like( $search_title ) . '%';
+			$where       .= $this->wpdb->prepare( " AND `{$this->wpdb->posts}`.`post_title` LIKE %s", $title_search ); // phpcs:ignore WordPress.DB.PreparedSQL
+		}
+		return $where;
+	}
+
+	/**
 	 * Get terms that can be used to suggest assets
 	 *
 	 * @param string $search The search query
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -133,10 +133,7 @@ class AssetSuggestionsService implements Service {
 			$attachments_ids = [ ...$attachments_ids, ...$product->get_gallery_image_ids(), $product->get_image_id() ];
 		}
 
-		$gallery_images_urls = get_post_gallery_images( $id );
-		$marketing_images    = [ ...$this->get_url_attachments_by_ids( $attachments_ids ), ...$gallery_images_urls ];
-		$marketing_images    = array_slice( $marketing_images, 0, self::DEFAULT_MAXIMUM_MARKETING_IMAGES );
-		$long_headline       = get_bloginfo( 'name' ) . ': ' . $post->post_title;
+		$attachments_ids  = [ ...$attachments_ids, ...$this->get_gallery_images_ids( $id ) ];
 
 		return [
 			'headline'                => [ $post->post_title ],
@@ -260,6 +257,23 @@ class AssetSuggestionsService implements Service {
 		$args = wp_parse_args( $args, $defaults );
 
 		return $this->wp->get_posts( $args );
+	}
+
+	/**
+	 * Get gallery images ids.
+	 *
+	 * @param int $post_id Post ID that contains the gallery.
+	 *
+	 * @return array List of gallery images ids.
+	 */
+	protected function get_gallery_images_ids( int $post_id ): array {
+		$gallery = get_post_gallery( $post_id, false );
+
+		if ( ! $gallery || ! isset( $gallery['ids'] ) ) {
+			return [];
+		}
+
+		return explode( ',', $gallery['ids'] );
 	}
 
 	/**
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -328,7 +328,7 @@ class AssetSuggestionsService implements Service {
 
 				// If the original size matches the suggested size with a precision of +-1px.
 				if ( $suggested_size && wp_fuzzy_number_match( $suggested_size->x, $image_size->x ) && wp_fuzzy_number_match( $suggested_size->y, $image_size->y ) ) {
-					$marketing_images[ $size_key ][] = wp_get_attachment_url( $id, $size_key );
+					$marketing_images[ $size_key ][] = wp_get_attachment_url( $id );
 				} elseif ( isset( $metadata['sizes'][ $size_key ] ) ) {
 					// use the sub size.
 					$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -332,7 +332,7 @@ class AssetSuggestionsService implements Service {
 				} elseif ( isset( $metadata['sizes'][ $size_key ] ) ) {
 					// use the sub size.
 					$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
-				} elseif ( $suggested_size && $this->image_utility->try_add_subsize_image( $id, $size_key, $suggested_size ) ) {
+				} elseif ( $suggested_size && $this->image_utility->maybe_add_subsize_image( $id, $size_key, $suggested_size ) ) {
 					// use the resized image.
 					$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
 				} else {
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -433,14 +433,19 @@ class AssetSuggestionsService implements Service {
 		$filtered_post_types = array_diff( $post_types, $excluded_post_types );
 
 		$args = [
-			'post_type'      => $filtered_post_types,
-			'posts_per_page' => $per_page,
-			'post_status'    => 'publish',
-			's'              => $search,
-			'offset'         => $offset,
+			'post_type'        => $filtered_post_types,
+			'posts_per_page'   => $per_page,
+			'post_status'      => 'publish',
+			'search_title'     => $search,
+			'offset'           => $offset,
+			'suppress_filters' => false,
 		];
 
+		add_filter( 'posts_where', [ $this, 'title_filter' ], 10, 2 );
+
 		$posts = $this->wp->get_posts( $args );
+
+		remove_filter( 'posts_where', [ $this, 'title_filter' ] );
 
 		foreach ( $posts as $post ) {
 			$post_suggestions[] = $this->format_final_url_response( $post->ID, 'post', $post->post_title, get_permalink( $post->ID ) );
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -45,10 +45,8 @@ class AssetSuggestionsService implements Service {
 
 	/**
 	 * Image requirements.
-	 *
-	 * @var array
 	 */
-	protected array $image_requirements = [
+	protected const IMAGE_REQUIREMENTS = [
 		self::MARKETING_IMAGE_KEY        => [
 			'minimum'     => [ 600, 314 ],
 			'recommended' => [ 1200, 628 ],
@@ -337,8 +335,8 @@ class AssetSuggestionsService implements Service {
 
 			foreach ( $size_keys as $size_key ) {
 
-				$minimum_size     = new DimensionUtility( ...$this->image_requirements[ $size_key ]['minimum'] );
-				$recommended_size = new DimensionUtility( ...$this->image_requirements[ $size_key ]['recommended'] );
+				$minimum_size     = new DimensionUtility( ...self::IMAGE_REQUIREMENTS[ $size_key ]['minimum'] );
+				$recommended_size = new DimensionUtility( ...self::IMAGE_REQUIREMENTS[ $size_key ]['recommended'] );
 				$image_size       = new DimensionUtility( $metadata['width'], $metadata['height'] );
 				$suggested_size   = $this->image_utility->recommend_size( $image_size, $recommended_size, $minimum_size );
 
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -5,6 +5,8 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\Ads;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
 use Automattic\WooCommerce\GoogleListingsAndAds\Utility\ArrayUtil;
+use Automattic\WooCommerce\GoogleListingsAndAds\Utility\ImageUtility;
+use Automattic\WooCommerce\GoogleListingsAndAds\Utility\DimensionUtility;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
 use Exception;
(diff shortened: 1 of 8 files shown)


src/Ads/AssetSuggestionsService.php
```diff
@@ -97,21 +146,20 @@ protected function get_post_assets( int $id ): array {
 			$attachments_ids = [ ...$attachments_ids, ...$product->get_gallery_image_ids(), $product->get_image_id() ];
 		}
 
-		$gallery_images_urls = get_post_gallery_images( $id );
-		$marketing_images    = [ ...$this->get_url_attachments_by_ids( $attachments_ids ), ...$gallery_images_urls ];
-		$marketing_images    = array_slice( $marketing_images, 0, self::DEFAULT_MAXIMUM_MARKETING_IMAGES );
-		$long_headline       = get_bloginfo( 'name' ) . ': ' . $post->post_title;
+		$attachments_ids  = array_slice( [ ...$attachments_ids, ...$this->get_gallery_images_ids( $id ) ], 0, self::DEFAULT_MAXIMUM_MARKETING_IMAGES );
+		$marketing_images = $this->get_url_attachments_by_ids( $attachments_ids );
```
puntope: 💅 For better readability 

Maybe we can add the `array_slice` logic in the `get_url_attachments_by_ids` scope. 

I see we do some preparation inside `get_url_attachments_by_ids` so maybe we can do something like :

```php
function prepare_image_ids( $ids ) {
  $ids = array_unique( ArrayUtil::remove_empty_values( $ids ) );
  $ids = array_map( 'intval', $ids );
  return [$ids, 0, self::DEFAULT_MAXIMUM_MARKETING_IMAGES];
}

[...]

protected function get_url_attachments_by_ids( $ids, ... ) {
   $ids = $this->prepare_image_ids($ids);
   [...] 
}
```

src/Ads/AssetSuggestionsService.php
```diff
@@ -24,16 +26,63 @@ class AssetSuggestionsService implements Service {
...
-		$this->wp = $wp;
-		$this->wc = $wc;
+	public function __construct( WP $wp, WC $wc, ImageUtility $image_utility ) {
+		$this->wp            = $wp;
+		$this->wc            = $wc;
+		$this->image_utility = $image_utility;
+
+		$this->image_requirements[ self::MARKETING_IMAGE_KEY ]        = self::MARKETING_IMAGE_SIZES;
+		$this->image_requirements[ self::SQUARE_MARKETING_IMAGE_KEY ] = self::MARKETING_SQUARE_IMAGE_SIZES;
+		$this->image_requirements[ self::LOGO_IMAGE_KEY ]             = self::LOGO_IMAGE_SIZES;
```
puntope: ❓  Why is this chunk of code in the constructor and not at the top? As far as I see is not dynamically depending of any constructor argument.

src/Ads/AssetSuggestionsService.php
```diff
@@ -152,23 +200,33 @@ protected function get_term_assets( int $id ): array {
 			$attachments_ids = [ ...$this->get_post_image_attachments( [ 'post_parent__in' => $posts_ids_assigned_to_term ] ), ...$attachments_ids ];
 		}
 
+		$attachments_ids  = array_slice( $attachments_ids, 0, self::DEFAULT_MAXIMUM_MARKETING_IMAGES );
 		$marketing_images = $this->get_url_attachments_by_ids( $attachments_ids );
```
puntope: 💅  

Related to https://github.com/woocommerce/google-listings-and-ads/pull/1804/files#r1045420662

We can avoid verbosity implementing limiting the ids size in  `get_url_attachments_by_ids` scope

src/Utility/ImageUtility.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Checks if the first image is bigger than the other one.
+	 *
+	 * @param DimensionUtility $image First image.
+	 * @param DimensionUtility $target The image to be compared.
+	 *
+	 * @return bool true if the first image is bigger than the other one otherwise false.
+	 */
+	public function is_bigger( DimensionUtility $image, DimensionUtility $target ): bool {
+		return $image->x >= $target->x && $image->y >= $target->y;
+	}
```
puntope: 💅  I personally would add this inside `DimensionUtility`

```php
public function is_bigger(DimensionUtility $target) {
  return $this->x >= $target->x && $this->y >= $target->y;
}
[...]
$size->isBigger($minimum)
```

src/Utility/ImageUtility.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Try to recommend a size using the real size, the recommended and the minimum.
+	 *
+	 * @param DimensionUtility $size Image size.
+	 * @param DimensionUtility $recommended Recommended image size.
+	 * @param DimensionUtility $minimum Minimum image size.
+	 *
+	 * @return DimensionUtility|bool False if does not fulfil the minimum size otherwise returns the suggested size.
+	 */
+	public function recommend_size( DimensionUtility $size, DimensionUtility $recommended, DimensionUtility $minimum ) {
+		if ( ! $this->is_bigger( $size, $minimum ) ) {
```
puntope, line 54: 💅  Maybe worth it to add `is_valid` or check manually if `y` is greater than zero to avoid potential division by zero.

src/Utility/ImageUtility.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @param DimensionUtility $size The new size for the subsize key.
+	 * @param bool             $crop Whether to crop the image.
+	 *
+	 * @return bool True if the subsize has been added to the attachment metadata otherwise false.
+	 */
+	public function try_add_subsize_image( int $attachment_id, string $subsize_key, DimensionUtility $size, bool $crop = true ): bool {
+		// It is required as wp_update_image_subsizes is not loaded automatically.
+		if ( ! function_exists( 'wp_update_image_subsizes' ) ) {
+			include ABSPATH . 'wp-admin/includes/image.php';
+		}
```
puntope: ❓  Would the use of the WP Proxy be beneficial here?

src/Utility/ImageUtility.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Try to add a new subsize image.
+	 *
+	 * @param int              $attachment_id Attachment ID.
+	 * @param string           $subsize_key The subsize key that we are trying to generate.
+	 * @param DimensionUtility $size The new size for the subsize key.
+	 * @param bool             $crop Whether to crop the image.
+	 *
+	 * @return bool True if the subsize has been added to the attachment metadata otherwise false.
+	 */
+	public function try_add_subsize_image( int $attachment_id, string $subsize_key, DimensionUtility $size, bool $crop = true ): bool {
```
puntope: 💅  I think the convention of this would be `maybe_` instead of `try_`

src/Ads/AssetSuggestionsService.php
```diff
@@ -24,16 +26,63 @@ class AssetSuggestionsService implements Service {
...
+	protected $image_requirements = [];
 
 	/**
 	 * AssetSuggestionsService constructor.
 	 *
-	 * @param WP $wp WP Proxy.
-	 * @param WC $wc WC Proxy.
+	 * @param WP           $wp WP Proxy.
+	 * @param WC           $wc WC Proxy.
+	 * @param ImageUtility $image_utility Image utility.
```
puntope, line 96: 💅  I miss also the prop declaration

```
	private ImageUtility $image_utility;
	private WC $wc;
	private WP $wp;
```

src/Ads/AssetSuggestionsService.php
```diff
@@ -24,16 +26,63 @@ class AssetSuggestionsService implements Service {
...
+		'minimum'     => [ 128, 128 ],
+		'recommended' => [ 1200, 1200 ],
+	];
+
+	/**
+	 * Image requirements.
+	 *
+	 * @var array
+	 */
+	protected $image_requirements = [];
```
puntope: 💅  Missed type

src/Ads/AssetSuggestionsService.php
```diff
@@ -227,19 +285,62 @@ protected function get_shop_products( array $args = [] ): array {
...
+
 		foreach ( $ids as $id ) {
-			$marketing_images[] = wp_get_attachment_image_url( $id );
+
+			$metadata = wp_get_attachment_metadata( $id );
+
+			foreach ( $size_keys as $size_key ) {
+
+				$minimum_size    = new DimensionUtility( ...$this->image_requirements[ $size_key ]['minimum'] );
+				$recomended_size = new DimensionUtility( ...$this->image_requirements[ $size_key ]['recommended'] );
```
puntope: 💅  Typo `recommended` instead of `recomended`

src/Ads/AssetSuggestionsService.php
```diff
@@ -227,19 +285,62 @@ protected function get_shop_products( array $args = [] ): array {
...
+				$image_size      = new DimensionUtility( $metadata['width'], $metadata['height'] );
+				$suggested_size  = $this->image_utility->recommend_size( $image_size, $recomended_size, $minimum_size );
+
+				// If the original size matches the suggested size with a precision of +-1px.
+				if ( $suggested_size && wp_fuzzy_number_match( $suggested_size->x, $image_size->x ) && wp_fuzzy_number_match( $suggested_size->y, $image_size->y ) ) {
```
puntope: 💅  I would wrap this into `DimensionUtility`. Maybe an `equals` or `match` method.

```php
function equals(DimensionUtility target) {
  return wp_fuzzy_number_match( $this->x, $target->x ) && wp_fuzzy_number_match( $this->y, $target->y )
}
```

Then we can just do

```
if ( $suggested_size && $suggested_size->equals($image_size) )
```

src/Ads/AssetSuggestionsService.php
```diff
@@ -227,19 +285,62 @@ protected function get_shop_products( array $args = [] ): array {
...
+					$marketing_images[ $size_key ][] = wp_get_attachment_url( $id, $size_key );
```
puntope: ❓ Method call is provided 2 parameters, but `wp_get_attachment_url `  method signature uses 1 parameters. Im unsure if we want to use  `wp_get_attachment_image_url ` instead, or omit the extra parameter.

src/Ads/AssetSuggestionsService.php
```diff
@@ -227,19 +285,62 @@ protected function get_shop_products( array $args = [] ): array {
...
+				} elseif ( isset( $metadata['sizes'][ $size_key ] ) ) {
+					// use the sub size.
```
puntope, line 359: ❓  I don't understand the concept of `subsize`. It's like an image resized to the defined non default size? Like `LOGO_IMAGE_SIZES` for example?

puntope:
❓ Testing with POSTS

- When I added the images in the post body it worked, however, when I add the image in Post Featured image. It doesn't

puntope:
❓  Testing with Terms

- I added the images in the category body Posts->Categories. It was not returned by the API

Screenshot 2022-12-12 at 14 16 19
Screenshot 2022-12-12 at 14 16 11

puntope:
Howdy @jorgemd24 thanks for working on this.

I reviewed the code and the behaviour. I left some comments :)

src/Ads/AssetSuggestionsService.php
jorgemd24: > It's like an image resized to the defined non default size

Correct, the default image sizes are thumbnail, large, medium etc. In our case, we want to create a specific `subsize` for the image assets, so if we need to resize the image, we can add the subsize and `wp_update_image_subsizes` will do the resize for us. 

If the image is suggested again, it will not be needed to create the image again because the subsize already exists.

src/Ads/AssetSuggestionsService.php
jorgemd24: I added your suggestion with one extra argument `$maximum_images` so we can make this function more flexible.
See: https://github.com/woocommerce/google-listings-and-ads/pull/1804/commits/9677071533e50f43fb89beafc6b6d12731089fbd

jorgemd24:
Thanks, @puntope for having a look at this.

>When I added the images in the post body it worked, however, when I add the image in Post Featured image. It doesn't

I have this option only for products but yes other post types can have feature images, so I tweak the code, see here:  [e31c859](https://github.com/woocommerce/google-listings-and-ads/pull/1804/commits/e31c859c5f2950d514fd0f25739c4c78ab5a7ad4)

>I added the images in the category body Posts->Categories. It was not returned by the API

Yes, that is expected, the code will return the images linked to the post/product/pages assigned to the term. Also, I am not sure if the attachment id from the description term can be retrieved easily as it seems that is inserted in the description as an HTML tag but the term does not seem to have its own attachments as `posts` does.

I have added the rest of your suggestions 👍 and I guess this is ready for another round!

src/Ads/AssetSuggestionsService.php
```diff
@@ -227,19 +300,76 @@ protected function get_shop_products( array $args = [] ): array {
...
+				if ( $suggested_size && $suggested_size->equals( $image_size ) ) {
+					$marketing_images[ $size_key ][] = wp_get_attachment_url( $id );
+				} elseif ( isset( $metadata['sizes'][ $size_key ] ) ) {
+					// use the sub size.
+					$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
+				} elseif ( $suggested_size && $this->image_utility->maybe_add_subsize_image( $id, $size_key, $suggested_size ) ) {
+					// use the resized image.
+					$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
+				} else {
+					continue;
```
puntope: 💅  This else could be omitted since it's the last line in the loop

src/Ads/AssetSuggestionsService.php
```diff
@@ -227,19 +300,76 @@ protected function get_shop_products( array $args = [] ): array {
...
+
+				// If the original size matches the suggested size with a precision of +-1px.
+				if ( $suggested_size && $suggested_size->equals( $image_size ) ) {
+					$marketing_images[ $size_key ][] = wp_get_attachment_url( $id );
+				} elseif ( isset( $metadata['sizes'][ $size_key ] ) ) {
+					// use the sub size.
+					$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
+				} elseif ( $suggested_size && $this->image_utility->maybe_add_subsize_image( $id, $size_key, $suggested_size ) ) {
+					// use the resized image.
+					$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
```
puntope, line 363: 💅  Optionally this 2 conditionals could be merged into one

```php
// If the original size matches the suggested size with a precision of +-1px.
if ( $suggested_size && $suggested_size->equals( $image_size ) ) {
	$marketing_images[ $size_key ][] = wp_get_attachment_url( $id );
} elseif ( isset( $metadata['sizes'][ $size_key ] ) || ( $suggested_size && $this->image_utility->maybe_add_subsize_image( $id, $size_key, $suggested_size ) ) ) {
	// use the previously created subsize or created  subsize.
	$marketing_images[ $size_key ][] = wp_get_attachment_image_url( $id, $size_key );
}
```

puntope: This is just an alternative. Separate in different conditions groups also ok.

puntope:
> I have this option only for products but yes other post types can have feature images, so I tweak the code, see here: [e31c859](https://github.com/woocommerce/google-listings-and-ads/pull/1804/commits/e31c859c5f2950d514fd0f25739c4c78ab5a7ad4)

Works!

src/Utility/DimensionUtility.php
```diff
@@ -0,0 +1,61 @@
...
+	/**
+	 * Checks if the dimension is bigger or equal than the other one.
+	 *
+	 * @param DimensionUtility $target The dimension to be compared.
+	 *
+	 * @return bool true if the dimension is bigger or equal than the other one otherwise false.
+	 */
+	public function is_bigger( DimensionUtility $target ): bool {
+		return $this->x >= $target->x && $this->y >= $target->y;
+	}
```
puntope: 💅  Thinking about this again. I wonder if we should omit `or equals` on the `is_bigger` name. WDTY?

puntope (approved):
Thanks for the adjustments @jorgemd24 

✅  LGTM

- Tested locally with terms and posts using galleries, featured images and attached images with different sizes.
- Tested code quality

Left some 💅  final comments but this is approved in advance

src/Utility/DimensionUtility.php
jorgemd24: I added `or equals` because the condition is `>=`  or we could rename this function to `is_bigger_or_equal` if it is a bit confusing. WDYT?

src/Ads/AssetSuggestionsService.php
jorgemd24: I will keep it as it is, as I think it is a bit easier to read it.

src/Ads/AssetSuggestionsService.php
```diff
@@ -20,20 +22,88 @@
...
+	 * Image requirements.
+	 *
+	 * @var array
+	 */
+	protected array $image_requirements = [
+		self::MARKETING_IMAGE_KEY        => self::MARKETING_IMAGE_SIZES,
+		self::SQUARE_MARKETING_IMAGE_KEY => self::MARKETING_SQUARE_IMAGE_SIZES,
+		self::LOGO_IMAGE_KEY             => self::LOGO_IMAGE_SIZES,
+
+	];
```
mikkamp: By moving this here wouldn't we also want to make it a `const` or are we going to filter this at some point to allow additional sizes?

Also I understand the need for needing a unique prefixed key name where we want to have a const, but do we also need a const for each of the image size arrays? Probably just a nitpick but I have the feeling I need to scroll back and forth to figure out what all the images sizes are. So defining it all in one seems more readable:
```suggestion
	protected const $image_requirements = [
		self::MARKETING_IMAGE_KEY        => [
			'minimum'     => [ 600, 314 ],
			'recommended' => [ 1200, 628 ],
		],
		self::SQUARE_MARKETING_IMAGE_KEY => [
			'minimum'     => [ 300, 300 ],
			'recommended' => [ 1200, 1200 ],
		],
		self::LOGO_IMAGE_KEY             => [
			'minimum'     => [ 128, 128 ],
			'recommended' => [ 1200, 1200 ],
		],
	];
```

src/Utility/DimensionUtility.php
mikkamp: We are checking if it meets the required minimum size, so what about `is_minimum_size`

src/Ads/AssetSuggestionsService.php
jorgemd24: > By moving this here wouldn't we also want to make it a const or are we going to filter this at some point to allow additional sizes?

I thought to leave it more flexible, thinking that maybe in the future, we could apply a filter so the user can update the sizes by themselves. 

> Also I understand the need for needing a unique prefixed key name where we want to have a const, but do we also need a const for each of the image size arrays?

Yes, I think it makes it easier to read all the sizes. I have updated the code: [f3a1a00](https://github.com/woocommerce/google-listings-and-ads/pull/1804/commits/f3a1a001127604f38fe52788e186c2ea3d65ab2f)

src/Utility/DimensionUtility.php
jorgemd24: @mikkamp I missed this comment yesterday, yes that it is a more convenient name. See [da76e8e](https://github.com/woocommerce/google-listings-and-ads/pull/1804/commits/da76e8e6d4018b64a3ef17264ede41604f224999)

src/Ads/AssetSuggestionsService.php
jorgemd24: I updated the variable `$image_requirements` to be a constant, if in the future we would like to apply a filter o to make it more dynamic we can always change it back.

PR #1805, opened 2022-12-06
PMax Assets: Add the final URL card to the assets section

eason9487:

This PR implements parts of 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) in #1787.

- Card for the Final URL
   - Input field for searching pages
   - Button for loading assets
   - Button for selecting another page

💡 The item "Once an asset group has been saved, the Final URL cannot be edited" is not included in this PR.

And also implements parts of 📌 [Integrating with backend APIs](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-integrating-apis) in #1787.

- Query website pages by page title
- Load suggested assets by a page

### Screenshots:

https://user-images.githubusercontent.com/17420811/205898260-26083537-69a3-466c-b09e-dd50a84c105b.mp4

2. Enter some keywords to query pages.
3. Select a page and click "Scan for assets".
4. Click "Or, select another page" to select another page.

### Additional details:

📌 Please note that the number of options on UI may be different from the number of the returned by `assets/final-url/suggestions` API.

This is because the `post_content` field in DB is also included in the search scope for now, and the `SelectControl` component has internal processing to filter out the returned results by matching them to the query string.

The API would be changed to not include `post_content`, so I didn't add code comments for this temporary difference.


js/src/data/types.js
```diff
@@ -0,0 +1,16 @@
+/**
```
puntope, line 1: 💅 

I have doubts about including this file here at `./data` level. Maybe we can move it inside `paid-ads/assets-group`.

js/src/components/paid-ads/asset-group/final-url-card.js
```diff
@@ -0,0 +1,84 @@
...
+ * @param {(assetGroup: AssetGroup | null) => void} props.onAssetsChange Callback function when the suggested assets are changed or reset to `null`.
+ * @param {string} [props.initialFinalUrl] The initial final URL.
+ * @param {boolean} [props.hideFooter=false] Whether to hide the card footer.
+ */
+export default function FinalUrlCard( {
+	onAssetsChange,
+	initialFinalUrl,
+	hideFooter = false,
+} ) {
+	const [ finalUrl, setFinalUrl ] = useState( initialFinalUrl || null );
```
puntope, line 35: 💅  This `| null` could be omitted I guess. `undefined` will have the same result.

js/src/components/paid-ads/asset-group/assets-loader.js
```diff
@@ -0,0 +1,158 @@
...
+	// The selector allows only one option to be selected which is expected, and the array is used
+	// here to get the entire data of the selected option rather than getting its key only.
+	// Ref: https://github.com/woocommerce/woocommerce/blob/6.9.0/packages/js/components/src/select-control/index.js#L137-L141
+	const [ selectedOptions, setSelectedOptions ] = useState( [] );
+	const [ fetching, setFetching ] = useState( false );
+	const { createNotice } = useDispatchCoreNotices();
+
+	const handleSearch = ( _, rawSearch ) => {
+		const cache = cacheRef.current;
+		const search = rawSearch.trim();
```
puntope: ❓  Is the search query case sensitive? 

If not we can maybe lowercase the `rawSearch` to avoid refetching. ie. (LOGO, logo)

js/src/components/paid-ads/asset-group/assets-loader.js
```diff
@@ -0,0 +1,158 @@
...
+ * @typedef {import('.~/data/types.js').AssetGroup} AssetGroup
+ */
+
+function allowSuggestedPages( search ) {
+	// Querying with empty `search` means getting the suggestion pages by default.
+	// Make it result in `new RegExp('.', 'i')` to avoid the default results being filtered out.
+	if ( search === '' ) {
+		return '.';
+	}
+	return search;
```
puntope: 💅  Maybe we can shorten this as `return search || '.' `

puntope:
❓ This is probably to discuss with UX. There is no fallback for "no results found" in the URL search input. 


https://user-images.githubusercontent.com/5908855/206623031-3eb03f06-38d8-46df-879e-1fb468db44cf.mov

puntope:
❓Also UX.  Feels strange the behaviour when focusing out the control.

Current: 

- Input is removed, then when focusing again, it show the previous results for a while and then fallback to default search.

Expected (imo):

- Input is not removed from the control

https://user-images.githubusercontent.com/5908855/206623361-2f1fc22d-4e7d-4f7c-b7bc-108fcc4dad55.mov

puntope:
❓  In slow networks it could be confusing there is not any spinner/indicator. In addition with  https://github.com/woocommerce/google-listings-and-ads/pull/1805#issuecomment-1343826955 is hard to know if there are no results or is still fetching. 

https://user-images.githubusercontent.com/5908855/206623779-030b7bca-66f5-4e18-ba21-d1f215549240.mov

puntope (commented):
Thanks for working on this @eason9487 

I tested it locally and review the code. I left some comments regarding both behaviour and code.

Thanks in advance.

js/src/data/types.js
eason9487: Since there is no consensus on where the type definitions should be placed, and there are currently scattered in many places, especially actions.js and selectors.js have accumulated many.

Given that types often need to be shared across different directories and files, I would like to centralize the type definitions of the data fetched from APIs in this file instead of adding them to actions.js or selectors.js.

js/src/components/paid-ads/asset-group/final-url-card.js
eason9487: The intention is to align with the state reset below:

https://github.com/woocommerce/google-listings-and-ads/blob/f084e1a13d30b8243e8f19de624fca7b54993540/js/src/components/paid-ads/asset-group/final-url-card.js#L52

So it won't have three possible types: `string`, `null`, and `undefined`.

Of course, it's possible to change the type possibilities to `string` and `undefined`, but the semantics of `setFinalUrl();` and `setFinalUrl( undefined );` are relatively unclear.

js/src/components/paid-ads/asset-group/assets-loader.js
eason9487: Thanks for the suggestion. Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/1805/commits/ac2ea585e40efa9e3bdbd3cfb97e3b5bb26dee70.

js/src/components/paid-ads/asset-group/assets-loader.js
eason9487: The original implementation was changed in https://github.com/woocommerce/google-listings-and-ads/pull/1805/commits/f084e1a13d30b8243e8f19de624fca7b54993540 while addressing another review comment https://github.com/woocommerce/google-listings-and-ads/pull/1805#issuecomment-1343826955

eason9487:
@puntope

> ❓ This is probably to discuss with UX. There is no fallback for "no results found" in the URL search input.

I added a possible solution by https://github.com/woocommerce/google-listings-and-ads/pull/1805/commits/f084e1a13d30b8243e8f19de624fca7b54993540, and I will loop the designer to see if it wokrs.

image

eason9487:
@puntope 

> ❓Also UX. Feels strange the behaviour when focusing out the control.
> 
> Current:
> 
> * Input is removed, then when focusing again, it show the previous results for a while and then fallback to default search.
> 
> Expected (imo):
> 
> * Input is not removed from the control
> 

> ❓ In slow networks it could be confusing there is not any spinner/indicator. In addition with [#1805 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/1805#issuecomment-1343826955) is hard to know if there are no results or is still fetching.
> 

Based on the goals set for this project, the development of a dedicated selector is not currently preferred, so only the components provided by the WC/WP were adapted to be as usable as possible.

However, due to the limitations of WC `SelectControl`, it doesn't provide the ability to show the fetching indicator and keep the search value of the input.

I tried many different ways and finally only got a relatively possible solution. The code change can be viewed in c3b65fc7302514bb41003d8b2799d6da72ec9d30 (`add/1787-final-url-card-workaround` branch).

https://user-images.githubusercontent.com/17420811/206687160-91cdf6df-12a2-47a9-be07-11bfbf5b65f9.mp4


### 📌 But the concerns are:

- It very depends on the internal implementation of `SelectControl`, which also means a higher chance of becoming incompatible in the future.
- The fetching indicator can only be shown outside of the dropdown options.
- Keeping the entered search value may overwrite the selected option.

https://user-images.githubusercontent.com/17420811/206688051-b8db981c-7184-48aa-a92c-5095b66e9e45.mp4

As these two are also not in the project requirements, I will ping the designer to get some input as well.

puntope:
Thanks for the changes! Looking good 

re - Your solution with spinner + outfocus. Is also working nice imo. But it's true that increases the complexity. So up to you and the UX to decide. My main concern was the "No results" issue but now it's solved.

nikivias:
> 

Hi, I was hoping we use an existing solution and this 'No matching results' solution seems to work. The enabling and disabling of the button is good feedback too.

nikivias:
Eason the spinner next to the label looks good to me :)

eason9487:
Hi @puntope, I merged c3b65fc7302514bb41003d8b2799d6da72ec9d30 (add/1787-final-url-card-workaround branch) for adding the searching spinner and keeping the entered search value. Could you help with a new round of code reviews?

puntope (approved):
✅ LGTM

Amazing Eason, looking good. Thanks for the tweaks

PR #1814, opened 2022-12-14
Improve image validation error messages

### LINKED ISSUE #1582: Valid product image URLs not being detected in Google Merchant Center

mikkamp:

As [outlined within the issue](https://github.com/woocommerce/google-listings-and-ads/issues/1582#issuecomment-1350674620), Google doesn't support images names with non standard characters. So the purpose of this PR is to improve the error messages that were returned to make it easier for users to identify which images need to be changed. This involves a custom validator to change the generic message `This value is not a valid URL` to specifically naming the image within the error message.

image

This PR added the following changes:
- A custom validator (based on the regular url validator) which by default enables normalization and customizes the error message.
- Improve error messages in the product issues table, previously errors were showing up with a leading bracket `] This value is not a valid URL [additionalImageLinks[0]`
- Add unit tests for custom image validation.

Closes #1582 

1. Create a png image with the name `Untitled-1000-×-1000-px-7-2.png` and upload it to a publicly accessible folder on your site (I placed it in wp-content/uploads).
2. Create a csv file to update the image name using the importer (replace the product ID and the image URL in the following example):
```csv
ID,Type,Images
1234,simple,"https://domain.test/wp-content/uploads/Untitled-1000-×-1000-px-7-2.png,https://domain.test/wp-content/uploads/æą€×.jpg"
```
3. Go to Products > Import and use the created CSV file to update the product (make sure to check update product).
image
4. Go to the connection test page `wp-admin/admin.php?page=connection-test-admin-page` and sync the product ID (or wait till the scheduled sync completes.
5. Confirm that there is an error syncing the product because the images are not valid.
6. Go to Marketing > Google Listings & Ads > Product Feed and select the Product Issues table (clear the transients if it's not loading a fresh list of issues).
7. Step through the pages till you find the product we used during the test (helps to name it starting with A if you have a lot of product issues).
8. Confirm that the errors are not correctly formatted, showing the image name, to help identify which one should be changed.
image

### Additional details:
As a follow up to these improvements it would be nice to add a [FAQ in the docs](https://woocommerce.com/document/google-listings-and-ads-faqs/) to explain the steps users need to take to re-upload an image with a valid name. cc: @sukafia 

### Changelog entry
* Tweak - Improve image validation error messages.

### DIFF
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -793,6 +793,13 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 	/**
 	 * Parse the code and formatted issue text out of the presync validation error text.
 	 *
+	 * Converts the error strings:
+	 * "[attribute] Error message." > "Error message [attribute]"
+	 *
+	 * Note:
+	 * If attribute is an array the name can be "[attribute[0]]".
+	 * So we need to match the additional set of square brackets.
+	 *
 	 * @param string $text
 	 *
 	 * @return string[] With indexes `code` and `issue`
@@ -807,12 +814,12 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 			];
 		}
 
-		// Convert imageLink to image
+		// Convert attribute name "imageLink" to "image".
 		if ( 'imageLink' === $matches[1] ) {
 			$matches[1] = 'image';
 		}
 
-		// Convert additionalImageLinks[] to galleryImage.
+		// Convert attribute name "additionalImageLinks[]" to "galleryImage".
 		if ( str_starts_with( $matches[1], 'additionalImageLinks' ) ) {
 			$matches[1] = 'galleryImage';
 		}
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -799,7 +799,7 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 	 */
 	protected function parse_presync_issue_text( string $text ): array {
 		$matches = [];
-		preg_match( '/^\[([^\]]+)\]\s*(.+)$/', $text, $matches );
+		preg_match( '/^\[([^\]]+\]?)\]\s*(.+)$/', $text, $matches );
 		if ( count( $matches ) !== 3 ) {
 			return [
 				'code'  => 'presync_error_attrib_' . md5( $text ),
@@ -811,6 +811,12 @@ class MerchantStatuses implements Service, ContainerAwareInterface {
 		if ( 'imageLink' === $matches[1] ) {
 			$matches[1] = 'image';
 		}
+
+		// Convert additionalImageLinks[] to galleryImage.
+		if ( str_starts_with( $matches[1], 'additionalImageLinks' ) ) {
+			$matches[1] = 'galleryImage';
+		}
+
 		$matches[2] = trim( $matches[2], ' .' );
 		return [
 			'code'  => 'presync_error_' . $matches[1],
--- a/src/Validator/ImageUrlConstraintValidator.php
+++ b/src/Validator/ImageUrlConstraintValidator.php
@@ -21,8 +21,8 @@ class ImageUrlConstraintValidator extends UrlConstraintValidator {
 	/**
 	 * Checks if the passed value is valid.
 	 *
-	 * @param GooglePrice $value The value that should be validated
-	 * @param Constraint  $constraint
+	 * @param string     $value The value that should be validated
+	 * @param Constraint $constraint
 	 *
 	 * @throws UnexpectedTypeException If invalid constraint provided.
 	 * @throws UnexpectedValueException If invalid value provided.
--- a/src/Validator/ImageUrlConstraintValidator.php
+++ b/src/Validator/ImageUrlConstraintValidator.php
@@ -1,0 +1,60 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Validator;
+
+use Normalizer;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Constraints\UrlValidator as UrlConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+use Symfony\Component\Validator\Exception\UnexpectedValueException;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class ImageUrlConstraintValidator
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Validator
+ */
+class ImageUrlConstraintValidator extends UrlConstraintValidator {
+
+	/**
+	 * Checks if the passed value is valid.
+	 *
+	 * @param GooglePrice $value The value that should be validated
+	 * @param Constraint  $constraint
+	 *
+	 * @throws UnexpectedTypeException If invalid constraint provided.
+	 * @throws UnexpectedValueException If invalid value provided.
+	 */
+	public function validate( $value, Constraint $constraint ) {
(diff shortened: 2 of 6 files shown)


src/MerchantCenter/MerchantStatuses.php
```diff
@@ -799,7 +799,7 @@ protected function get_valid_issue_types(): array {
 	 */
 	protected function parse_presync_issue_text( string $text ): array {
 		$matches = [];
-		preg_match( '/^\[([^\]]+)\]\s*(.+)$/', $text, $matches );
+		preg_match( '/^\[([^\]]+\]?)\]\s*(.+)$/', $text, $matches );
```
puntope, line 809: I think this is just checking if `$text` contains `[{something}]`? I know it's out scope but it's nice for me to add a little comment in the regex about what is checking

puntope: Also would be nice to add what text is inside the PHPDOC `$text` param

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -811,6 +811,12 @@ protected function parse_presync_issue_text( string $text ): array {
 		if ( 'imageLink' === $matches[1] ) {
 			$matches[1] = 'image';
 		}
+
+		// Convert additionalImageLinks[] to galleryImage.
```
puntope: 💅  Was confusing for me at first glance this comment. AdditionalImageLinks[] is syntactically an array and I thought we were converting some kind of variable data into other format. After reading the context I understood that `Convert additionalImageLinks[] to galleryImage` means `replace "additionalImageLinks[]" in the issue description to "galleryImage"`

Same goes with the old comment in line 810

src/Validator/ImageUrlConstraintValidator.php
```diff
@@ -0,0 +1,60 @@
...
+		$value   = Normalizer::normalize( $value );
+		$pattern = sprintf( static::PATTERN, implode( '|', $constraint->protocols ) );
+
+		if ( ! preg_match( $pattern, $value ) ) {
+			$this->context->buildViolation( $constraint->message )
+				->setParameter( '{{ name }}', wp_basename( $value ) )
+				->setCode( ImageUrlConstraint::INVALID_URL_ERROR )
+				->addViolation();
+
+			return;
```
puntope, line 57: This `return` could be omitted since it's the last statement in the method

src/Validator/ImageUrlConstraintValidator.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @throws UnexpectedValueException If invalid value provided.
+	 */
+	public function validate( $value, Constraint $constraint ) {
+		if ( ! $constraint instanceof ImageUrlConstraint ) {
+			throw new UnexpectedTypeException( $constraint, ImageUrlConstraint::class );
+		}
+
+		if ( null === $value || '' === $value ) {
+			return;
+		}
```
puntope, line 37: ❓ Pretty minor. But not sure 100% about this chunk of code. Seems for me that it's covered in L39 and L43?

tests/Unit/Product/WCProductAdapterTest.php
```diff
@@ -1577,6 +1577,35 @@ public function test_valid_image_name_with_utf8_nfd() {
...
+		$adapted_product = new WCProductAdapter(
+			[
+				'wc_product'    => $product,
+				'targetCountry' => 'US',
+			]
+		);
+
+		$violation = $this->validate_product_property( $adapted_product, 'imageLink' );
+		$this->assertEquals(
+			'Product image "600-×-300.png" is not a valid name.',
```
puntope, line 1600: ❓  Just curious why here it appears `Product image "600-×-300.png" is not a valid name.` and in the screenshot and code we are adding `Product image "600-×-300.png" is not a valid name. [image]`

puntope (approved):
✅ LGTM

Tested locally and the descriptions are correctly shown the error in `/wp-content/uploads/Untitled-1000-×-1000-px-7-2.jpg`. 

For  `/wp-content/uploads/æą€×.png` I received an error in the import file: 

`Error getting remote image http://wp23.test/wp-content/uploads/æą€×.png. Error: A valid URL was not provided.`

I tried with just `wp-content/uploads/æą€.png` and i worked, however it was a valid upload since all of them were ASCII characters


I left some comments, but they are minor 💅 . So approved in advance

src/MerchantCenter/MerchantStatuses.php
mikkamp: This comes from the fact that we map the errors to a string `[attributeName] Error message` as seen here: https://github.com/woocommerce/google-listings-and-ads/blob/2.3.3/src/Google/BatchInvalidProductEntry.php#L84
They then get saved as a collection of errors in product meta data.

However when we display the product issues table we decided later on to include errors from multiple sources. To make them match other errors the format needed to be the other way around. So this function is basically converting:
`[attributeName] Error message` > `Error message [attributeName]`

What my fix does is catch the extra bracket since the attributeName can be `[name[0]]` in the case of an image array. I'll go ahead and add some more info in the doc block to make that clearer though.

src/Validator/ImageUrlConstraintValidator.php
mikkamp: Totally correct, although I'd prefer to leave it in place as we are overwriting/customizing this function: https://github.com/symfony/validator/blob/v6.2.0/Constraints/UrlValidator.php#L74-L81
I chose to keep the structure relatively identical in case we need to update something later.

src/Validator/ImageUrlConstraintValidator.php
mikkamp: These lines are ignoring the validation rule and returns early so it can move on to the next validation rule (we can have multiple validation rules per attribute). In this case that's important because for the main image we also check the Constraint NotBlank, whereas we don't want that constraint to apply for gallery images.

The validator docs explain that a bit clearer: https://symfony.com/doc/current/validation/custom_constraint.html#creating-the-validator-itself

tests/Unit/Product/WCProductAdapterTest.php
mikkamp: In the test we are directly catching the validation error.

If we actually sync the product (not done in this particular test) the validation error is mapped to `[attributeName] Error message` and then stored in product meta.

When we collect issues to include in the issue DB table we convert the errors from product meta to `Error message [attributeName]`, attributeName is directly converted to the string `image`.

mikkamp:
> For /wp-content/uploads/æą€×.png I received an error in the import file:
> Error getting remote image `http://wp23.test/wp-content/uploads/æą€×.png`. Error: A valid URL was not provided.
> I tried with just `wp-content/uploads/æą€.png` and i worked, however it was a valid upload since all of them were ASCII characters

Did you get the import error because I forgot to mention in the steps to place both images in the uploads folder? Or was there some other issue with that image?

I also noticed that the importer still converts some characters in the image name, for example `æą€` > `aeaE` which means we can't reproduce the error unless we add other characters that aren't mapped. I didn't really dig deeper exactly which characters the importer maps and which ones it doesn't. But I assumed there are other importers which don't do this conversion so there can still be sites that end up with images where those characters aren't mapped.

puntope:
> Did you get the import error because I forgot to mention in the steps to place both images in the uploads folder? Or was there some other issue with that image?

Was a problem in my import file. I had some typo in the extension.

PR #1819, opened 2022-12-16
PMax Assets: Integrate WP Media library to implement the asset images selector

eason9487:

This PR implements a part of 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) in #1787.

- Multiple images selector

#### 📹 Crop, skip cropping, check image size, remove selected image, and replace selected image

https://user-images.githubusercontent.com/17420811/208123491-7a403bb0-17e8-41a5-8edf-62f01b39e9bd.mp4

#### 📹 Pre-select when replacing image and check the size of the newly uploaded image

https://user-images.githubusercontent.com/17420811/208124129-29c80288-314d-421e-a271-a42ea5d2ddb9.mp4

#### 📹 Remove the deleted image from the selected images

https://user-images.githubusercontent.com/17420811/208124578-711f1ef8-2772-4412-aa3e-66f4f167d29b.mp4

#### 📹 Swap images if the selected image already exists while replacing

https://user-images.githubusercontent.com/17420811/208124807-a5ca07cf-ccf8-447c-b4ed-a5efbd5a03c7.mp4

#### 📹 The image to be replaced has been deleted

https://user-images.githubusercontent.com/17420811/208124844-b7c72ac1-bedb-4e78-810f-163f7ad83438.mp4

#### 📷 Disable "Add image" button and multiline images

image

2. Use the asset image selector to see if most of the operations are working properly.

### DIFF
--- a/js/src/components/paid-ads/asset-group/images-selector.js
+++ b/js/src/components/paid-ads/asset-group/images-selector.js
@@ -58,7 +58,7 @@ export default function ImagesSelector( {
 					// If the selected image already exists while replacing, it's considered a swap position.
 					nextImages.splice( index, 1, { ...replacingImage } );
 				}
-				// If index gets -1 here, it means the image to be replaced has been deleted.
+				// If index gets -1 here, it means the image to be replaced has been removed via the `onDelete` callback.
 				index = nextImages.indexOf( replacingImage );
 			}
 
--- a/js/src/components/paid-ads/asset-group/images-selector.js
+++ b/js/src/components/paid-ads/asset-group/images-selector.js
@@ -36,7 +36,8 @@ export default function ImagesSelector( {
 } ) {
 	const [ replacingImage, setReplacingImage ] = useState( null );
 	const [ images, setImages ] = useState( () =>
-		initialImageUrls.map( ( url ) => ( { url, alt: '' } ) )
+		// The asset images fetched from Google Ads are only URLs.
+		initialImageUrls.map( ( url ) => ( { url, id: url, alt: '' } ) )
 	);
 
 	const removeImage = ( deletedImage ) => {
--- a/js/src/components/paid-ads/asset-group/images-selector.js
+++ b/js/src/components/paid-ads/asset-group/images-selector.js
@@ -41,6 +41,9 @@ export default function ImagesSelector( {
 	);
 
 	const removeImage = ( deletedImage ) => {
+		if ( deletedImage.id === awaitingActionImage?.id ) {
+			setAwaitingActionImage( null );
+		}
 		setImages( images.filter( ( { id } ) => id !== deletedImage.id ) );
 	};
 
--- a/js/src/components/paid-ads/asset-group/images-selector.js
+++ b/js/src/components/paid-ads/asset-group/images-selector.js
@@ -61,7 +61,7 @@ export default function ImagesSelector( {
 					// If the selected image already exists while replacing, it's considered a swap position.
 					nextImages.splice( index, 1, { ...awaitingActionImage } );
 				}
-				// If index gets -1 here, it means the image to be replaced has been removed via the `onDelete` callback.
+				// Find the index to be replaced with the selected image.
 				index = nextImages.indexOf( awaitingActionImage );
 			}
 
--- a/js/src/components/paid-ads/asset-group/images-selector.js
+++ b/js/src/components/paid-ads/asset-group/images-selector.js
@@ -1,0 +1,125 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useState } from '@wordpress/element';
+import GridiconPlusSmall from 'gridicons/dist/plus-small';
+import GridiconCrossCircle from 'gridicons/dist/cross-circle';
+
+/**
+ * Internal dependencies
+ */
+import AppButton from '.~/components/app-button';
+import useCroppedImageSelector from '.~/hooks/useCroppedImageSelector';
+import './images-selector.scss';
+
+/**
+ * @typedef {Object} AssetImageConfig
+ * @property {number} minWidth The minimum width.
+ * @property {number} minHeight The minimum height.
+ * @property {number} suggestedWidth The suggested width.
+ * @property {number} suggestedHeight The suggested height.
+ */
+
+/**
+ * Renders a selector for asset images.
+ *
+ * @param {Object} props React props.
+ * @param {AssetImageConfig} props.imageConfig The config of the asset image.
+ * @param {string[]} props.initialImageUrls The initial image URLs.
+ * @param {number} [props.maxNumberOfImages=0] The maximum number of images. 0 by default and it means unlimited number.
+ */
+export default function ImagesSelector( {
+	imageConfig,
+	initialImageUrls = [],
+	maxNumberOfImages = 0,
+} ) {
+	const [ replacingImage, setReplacingImage ] = useState( null );
+	const [ images, setImages ] = useState( () =>
+		initialImageUrls.map( ( url ) => ( { url, alt: '' } ) )
+	);
+
+	const removeImage = ( deletedImage ) => {
+		setImages( images.filter( ( { id } ) => id !== deletedImage.id ) );
+	};
+
+	const handle = useCroppedImageSelector( {
+		...imageConfig,
+		onDelete: removeImage,
+		onSelect( image ) {
+			const nextImages = [ ...images ];
+
+			// Find if there is a duplicate image first.
+			let index = nextImages.findIndex( ( { id } ) => id === image.id );
+
+			if ( replacingImage ) {
+				if ( index !== -1 && image.id !== replacingImage.id ) {
+					// If the selected image already exists while replacing, it's considered a swap position.
+					nextImages.splice( index, 1, { ...replacingImage } );
+				}
+				// If index gets -1 here, it means the image to be replaced has been deleted.
+				index = nextImages.indexOf( replacingImage );
+			}
+
+			if ( index === -1 ) {
+				nextImages.push( image );
+			} else {
+				nextImages.splice( index, 1, image );
+			}
+
+			setReplacingImage( null );
+			setImages( nextImages );
+		},
+	} );
+
+	const handleUpsertImageClick = ( event, image = null ) => {
+		setReplacingImage( image );
+		handle.openSelector( image?.id );
+	};
+
+	return (
+		<div className="gla-images-selector">
+			<div className="gla-images-selector__image-list">
+				{ images.map( ( image ) => {
+					return (
+						<div
+							key={ image.url }
+							className="gla-images-selector__image-item"
+						>
+							<AppButton
+								className="gla-images-selector__replace-image-button"
+								onClick={ () =>
+									handleUpsertImageClick( null, image )
+								}
+							>
+								<img
+									className="gla-images-selector__image"
+									alt={ image.alt }
+									src={ image.url }
+								/>
+							</AppButton>
+							<AppButton
+								className="gla-images-selector__remove-image-button"
+								icon={ <GridiconCrossCircle /> }
+								iconSize={ 20 }
+								onClick={ () => removeImage( image ) }
+							/>
+						</div>
+					);
+				} ) }
+			</div>
+			<AppButton
+				className="gla-images-selector__add-image-button"
+				isLink
+				disabled={
+					maxNumberOfImages !== 0 &&
+					images.length >= maxNumberOfImages
+				}
+				icon={ <GridiconPlusSmall /> }
+				iconSize={ 16 }
+				text={ __( 'Add image', 'google-listings-and-ads' ) }
+				onClick={ handleUpsertImageClick }
+			/>
+		</div>
+	);
+}
--- a/js/src/components/paid-ads/asset-group/images-selector.js
+++ b/js/src/components/paid-ads/asset-group/images-selector.js
@@ -34,7 +34,7 @@ export default function ImagesSelector( {
 	initialImageUrls = [],
 	maxNumberOfImages = 0,
 } ) {
-	const [ replacingImage, setReplacingImage ] = useState( null );
+	const [ awaitingActionImage, setAwaitingActionImage ] = useState( null );
 	const [ images, setImages ] = useState( () =>
 		// The asset images fetched from Google Ads are only URLs.
 		initialImageUrls.map( ( url ) => ( { url, id: url, alt: '' } ) )
@@ -53,13 +53,13 @@ export default function ImagesSelector( {
 			// Find if there is a duplicate image first.
 			let index = nextImages.findIndex( ( { id } ) => id === image.id );
 
-			if ( replacingImage ) {
-				if ( index !== -1 && image.id !== replacingImage.id ) {
+			if ( awaitingActionImage ) {
+				if ( index !== -1 && image.id !== awaitingActionImage.id ) {
 					// If the selected image already exists while replacing, it's considered a swap position.
-					nextImages.splice( index, 1, { ...replacingImage } );
+					nextImages.splice( index, 1, { ...awaitingActionImage } );
 				}
 				// If index gets -1 here, it means the image to be replaced has been removed via the `onDelete` callback.
-				index = nextImages.indexOf( replacingImage );
+				index = nextImages.indexOf( awaitingActionImage );
 			}
 
 			if ( index === -1 ) {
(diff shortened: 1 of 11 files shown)


js/src/hooks/types.js
```diff
@@ -0,0 +1,12 @@
+/**
```
puntope, line 1: 💅  Since is there not a consensus yet about this `types.js` standard and `ImageMedia` is only being used in `useCroppedImageSelector.js` I would add it there directly deleting this file.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
```diff
@@ -0,0 +1,289 @@
...
+	const ratio = widthScale / heightScale;
+
+	if ( width / height > ratio ) {
+		width = Math.round( height * ratio );
+	} else {
+		height = Math.round( width / ratio );
+	}
+
+	return [ width, height ];
+}
```
puntope: ❓  I guess this just calculate the maximum cropping size but is not actually cropping the image.

💅  If so, I would rename the function as  `calculateMaxCropByFixedRatio` or something similar. IMO without reading the function internals `cropByFixedRatio` could be understood as we are actually cropping the image.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
```diff
@@ -0,0 +1,289 @@
...
+			// - https://github.com/WordPress/wordpress-develop/blob/5.9.0/src/js/media/controllers/customize-image-cropper.js#L14
+			// - https://github.com/WordPress/wordpress-develop/blob/5.9.0/src/js/media/controllers/cropper.js#L65
+			// - https://github.com/WordPress/wordpress-develop/blob/5.9.0/src/js/media/views/cropper.js#L48-L54
+			const imgSelectOptions = ( attachment, controller ) => {
+				const width = attachment.get( 'width' );
+				const height = attachment.get( 'height' );
+				const args = [ width, height, minWidth, minHeight ];
+				const ratioPercentError = calcRatioPercentError( ...args );
+
+				if ( ratioPercentError < allowedRatioPercentError ) {
```
puntope: 💅  I guess this should be `<= `since we "allowed" that percent error, so equals is valid. Otherwise we could rename it to `ratioPercentErrorTolerance` or just `ratioPercentError`

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
```diff
@@ -0,0 +1,289 @@
...
+
+	if ( width / height > ratio ) {
+		width = Math.round( height * ratio );
+	} else {
+		height = Math.round( width / ratio );
+	}
+
+	return [ width, height ];
+}
+
```
puntope, line 46: 💅  Even tho you added the refs of the extended library. I missed some explanations on each file function, just explaining what we are doing there is enough, no need to be super detailed. I guess tat would save time of checking parent files for the developers landing on this code.

js/src/components/paid-ads/asset-group/images-selector.js
```diff
@@ -0,0 +1,126 @@
...
+			// Find if there is a duplicate image first.
+			let index = nextImages.findIndex( ( { id } ) => id === image.id );
+
+			if ( replacingImage ) {
+				if ( index !== -1 && image.id !== replacingImage.id ) {
+					// If the selected image already exists while replacing, it's considered a swap position.
+					nextImages.splice( index, 1, { ...replacingImage } );
+				}
+				// If index gets -1 here, it means the image to be replaced has been deleted.
+				index = nextImages.indexOf( replacingImage );
```
puntope: ❓  I guess this means the we clicked on "Delete permanently" in the media selector?

js/src/components/paid-ads/asset-group/images-selector.js
```diff
@@ -0,0 +1,126 @@
...
+	const handle = useCroppedImageSelector( {
+		...imageConfig,
+		onDelete: removeImage,
+		onSelect( image ) {
+			const nextImages = [ ...images ];
+
+			// Find if there is a duplicate image first.
+			let index = nextImages.findIndex( ( { id } ) => id === image.id );
+
+			if ( replacingImage ) {
```
puntope: 💅  I found `replacingImage` a bit confusing in order to identify if `replacingImage` was

A:  the new selected image to replace the previous one. 
B:  the previous image to be replaced by the new selected image. 

After inspecting all the code I guess its B. So I'd suggest replacing `image` with `newSelectedImage` and `previousSelectedImage`

puntope:
Howdy @eason9487 . I tested this PR, thanks for all the code and work here:

I did this checks besides my suggestion in code:

✅  A red message appears near the button when the image size are below the minimum. And it disables the selection.
✅  Images are cropped correctly preserving the desired aspect ratio and the min height and width. 
✅  When image is correctly sized, it allows to skip crop.
✅  The maximum number of image limitation works.
✅  Swap images works (see conflict with remove image at the bottom)
✅  Remove image works
✅  Replace image works
✅  Replace image and then delete the image works.  (see conflict with swap image at the bottom)
✅  Squared images behaviour (works good as previous checked steps)
✅  Pre-select when replacing image and check the size of the newly uploaded image works
✅  Multiline images

❓ **Topic 1**: Set initial crop to suggested size if possible. 

I somehow expected this behaviour, not sure if thats planned or it could be omitted:

- Image squared: Suggested size 1200x1200 
- Image A squared: Original size 1500x1500
- Select image A, open cropper with 1200x1200 preselected.

Current behaviour: 

- Image squared: Suggested size 1200x1200 
- Image A squared: Original size 1500x1500
- Select image A, open cropper with 1500x1500 preselected.

❓ **Topic 2**: Disable crop button if no crop is done. 

 I somehow expected this behaviour, not sure if thats planned or it could be omitted:

- Image squared: Suggested size 1200x1200 
- Image A squared: Original size 1200x1200
- Select image A, open cropper with 1200x1200 preselected.
- No crop was done, so Crop Button is disabled.

Current behaviour: 

- Image squared: Suggested size 1200x1200 
- Image A squared: Original size 1200x1200
- Select image A, open cropper with 1200x1200 preselected.
- No crop was done, but "Crop" button is enabled
- Click on "Crop"
- New cropped image with same dimensions is created. 

**⚠️  The image to be replaced has been deleted in combination with swap images doesn't work as expected:**

I expected this behaviour:

- Add Image A 
- Add Image B
- Click on Image B 
- Delete Image B permanently
- Select image A
- Save with Skip Crop
- Image B is deleted. Image A is in the image list

Current behaviour: 

- Add Image A 
- Add Image B
- Click on Image B 
- Delete Image B permanently
- Select image A
- Save with Skip Crop
- Image B and Image A are in the image list, when clicking on Image B again, there is no selection.


https://user-images.githubusercontent.com/5908855/208607953-e42f2bbb-cc2c-4feb-91b8-139f779c5ff2.mov

js/src/hooks/types.js
eason9487: The reason is similar to https://github.com/woocommerce/google-listings-and-ads/pull/1805#discussion_r1044303992. I would like to advocate that the type definitions of hooks be centralized in this file.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
eason9487: Renamed in https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/d11fc0ec9a053e95a77aadb8dcfd6fca1f76bea3.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
eason9487: Thanks for the suggestion. Renamed in https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/4f321138ba65ec29d6246d4f68d9352edde72702.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
eason9487: Since the WP Media library seems to have no documents and it's mainly built on top of Backbone.js and mixed some WP custom mechanisms,  the need for additional explanations very depends on the readers' understanding of this background knowledge. Could you help point out which places need more explanation?

js/src/components/paid-ads/asset-group/images-selector.js
eason9487: Yes, but ImagesSelector should be agnostic to the WP Media library, which relies only on the interface of useCroppedImageSelector. I adjusted the code comment a little bit in https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/0932206b4476c5bce5080f0798f897af380243a9. Does this make it clearer?

js/src/components/paid-ads/asset-group/images-selector.js
eason9487: Consider that the `replacingImage` may not only be replaced but may also be swapped places or removed without further processing. I renamed it to `awaitingActionImage` in https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/84f63e919ad5d4e5a12d80475f3c73c560785beb.

eason9487:
@puntope, thanks for the detailed code review and testing.

With this comment, I believe all review comments were addressed or replied. Could you help with a new round of code reviews?

---

> ❓ **Topic 1**: Set initial crop to suggested size if possible. [...]

Since the size and content of the original images have usually been adjusted by users, using the recommended size to pre-select the cropping might annoying to users.

For example, if the user already has many 1500x1500 images and the appropriate gaps between the content and borders have been retained within images. The pre-selected 1200x1200 area in the middle of images may instead always crop onto the content of images, causing the user to keep reselecting to the largest area.

Of course, the opposite case is also true, and there are advantages and disadvantages to either approach, so I think we could go with the current implementation for now.

---

> ❓ **Topic 2**: Disable crop button if no crop is done. [...]

This is a great suggestion! I added this processing by https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/54da6a7d95648b4b0f8f7f8576960699fc1420df.

https://user-images.githubusercontent.com/17420811/209540786-e73f3a84-0754-4907-904a-6e023bacddd3.mp4

---

>  **⚠️  The image to be replaced has been deleted in combination with swap images doesn't work as expected:** [...]

Thanks for catching this. 👍 

Fixed by https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/adb4734264ea9ac39e5c87c11b0229c6952757e0.

https://user-images.githubusercontent.com/17420811/209540876-aa0b52bf-6914-472d-90f9-671711507697.mp4

js/src/hooks/types.js
puntope: Yes. I understand the idea. But I would suggest to have a team consensus. Maybe a P2 or an agenda question to be all in the same page about the new proposal.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
puntope: >  Could you help point out which places need more explanation?

Especially I had doubts about these 2. (the second one in L171, for sure)


- In the chunk of code for handling the invalid size and showing error message: (Line 216)

maybe something like; 
// Check if the selected image has a valid width and height. If it has invalid date show an error message in the toolbar. 


- Line 171, about the `states`. Maybe explain why we are loading here this controllers. 

// Maybe like override the WP Media adding suggested width and height for the image and a custom image cropper. 

(Unsure if its accurate)  

Optionally to document the handlers and other internal functions.

js/src/components/paid-ads/asset-group/images-selector.js
puntope: Thanks for the adjustment. I'm still not sure about the new name `awaitingActionImage`. The word `await` and `action` in the name could have ambiguous meanings and I found it confusing. 

I don't want to sound pushy about my suggestion. But what is against `newSelectedImage` and `previousSelectedImage`?  I think it clearly defines what the user is doing in the UI.

js/src/components/paid-ads/asset-group/images-selector.js
puntope: > Yes, but ImagesSelector should be agnostic to the WP Media library,

Could you elaborate this? Does it mean we can use another media Library rather the WP Media like a 3rd party one?

puntope: ❓  If so, should we test the PR against one 3rd party media library to see if it works?

puntope:
Howdy @eason9487 

I tested again the behaviour and now is looking good. Thanks for the adjustments.

Looks good to me. My only concern besides my 💅  comments is this one https://github.com/woocommerce/google-listings-and-ads/pull/1819#discussion_r1057182329

Just to be sure if it would be convenient to test with a 3rd party media lib or not before approval.

js/src/hooks/types.js
eason9487: Surely it's nicer to have a team consensus for this, and the discussion of this convention perhaps could be covered in #195.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js
eason9487: Not sure line numbers 171 and 216 are referred to which code revision. Based on the comment content, I guess it means:
- https://github.com/woocommerce/google-listings-and-ads/blob/54da6a7d95648b4b0f8f7f8576960699fc1420df/js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js#L233
- https://github.com/woocommerce/google-listings-and-ads/blob/54da6a7d95648b4b0f8f7f8576960699fc1420df/js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.js#L198

I added some code comments and adjusted some naming in https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/bd25f054a082d990ea94d76759ceac193bd52ef9. Most internal functions still don't have additional documents because the function names should be sufficient to describe themselves. Hope it has covered where the explanations were needed.

js/src/components/paid-ads/asset-group/images-selector.js
eason9487: > > Yes, but ImagesSelector should be agnostic to the WP Media library,
> 
> Could you elaborate this? Does it mean we can use another media Library rather the WP Media like a 3rd party one?

It's simply due to the management/practice of interdependence.

Since the JS part of WP Media library is based on jQuery, Backbone.js and some `wp.*` libraries, and it also has its own lifecycle and state management, one of the purposes is to isolate the dependency on WP Media in `useCroppedImageSelector`.

`ImagesSelector`, as a hook user of `useCroppedImageSelector`, relies only on the interfaces provided by the hook for both code implementation and comments, and does not need to be coupled with the WP Media library. For example, `ImagesSelector` doesn't need to know how an image would be deleted, or by whom, but it only needs to know when and which image was deleted.

Therefore, that's the thought of "ImagesSelector should be agnostic to the WP Media library".

The possibility of replacing the implementation of `useCroppedImageSelector` with some other library is the result of this design approach, but not the purpose. In this case, I didn't expect WP Media would be replaced by a 3rd-party library, because in practice this probably won't happen.

eason9487: BTW, I adjusted the code comment in https://github.com/woocommerce/google-listings-and-ads/pull/1819/commits/4b111e86bdd3c1869b50316094daae730d7e8273 since the -1 `index` case originally described no longer occurs after the fix in adb4734264ea9ac39e5c87c11b0229c6952757e0.

js/src/components/paid-ads/asset-group/images-selector.js
eason9487: > I'm still not sure about the new name `awaitingActionImage`. The word `await` and `action` in the name could have ambiguous meanings and I found it confusing.

It's precisely to avoid ambiguity, the word "selected" is used only in the `onSelect` callback and refers to "the selected image", and the word was avoided in the rest of `ImagesSelector`.

Otherwise, for example, would the `previousSelectedImage` literally mean the previously selected image in the `onSelect` callback, or the previously selected image in the callback of the replacement buttons?

The `awaitingActionImage` keeps the last image data sent from the associated click handler of a replacement button. After "the selected image" is called back from `onSelect`, the next action for `awaitingActionImage` could be:
- pointing out which image to be replaced with the selected image.
- pointing out which image to be swapped the position with the selected image.
- nothing if the reset action for `awaitingActionImage` has occurred in the `onDelete` callback.

Thus, the `awaitingActionImage` is an image **awaiting** something to happen, and when it does, it takes **action** according to the situation at the moment.

---

Besides, I didn't rename `image` to `newSelectedImage` because the `image` as the parameter to `onSelect` callback should be enough to indicate that it is the "selected image", and the `onSelect` implementation is only a few lines long, so I thought it's probably not confusing. But it's just my opinion. I'm happy to rename `image` to `selectedImage` if it's still confusing you.

eason9487:
Hi @puntope, thanks for the comments. I pushed two commits for adjusting code comments and replied the comments in the second round of code reviews. Could you take another look at this PR?

puntope (approved):
✅  Approved thanks @eason9487

PR #1820, opened 2022-12-16
PMax Assets - Fetch assets groups

jorgemd24:

This PR adds the functionality to fetch the asset groups for a specific campaign ID. The asset group will contain assets of the following type (when present):

Asset Field Type|
--|
HEADLINE|
LONG_HEADLINE|
DESCRIPTION|
BUSINESS_NAME|
MARKETING_IMAGE|
SQUARE_MARKETING_IMAGE|
LOGO|
CALL_TO_ACTION_SELECTION|

Example response:

```
 [
 {
        "id": 6443481181,
        "final_url": "https://mystagingwebsite.com",
        "display_url_path": [
            "",
            ""
        ],
        "assets": {
            "square_marketing_image": [
                {
                    "id": 23933203451,
                    "content": "https://tpc.googlesyndication.com/simgad/184308248126859863561"
                },
                {
                    "id": 23946157339,
                    "content": "https://tpc.googlesyndication.com/simgad/70392509749665497401"
                }
            ],
            "marketing_image": [
                {
                    "id": 23933203454,
                    "content": "https://tpc.googlesyndication.com/simgad/113463390769957791581"
                }
            ],
         .....more assets
 }
]
```

1. As we are using some new imports from the GoogleAds Library, you will need to remove your vendor folder `rm -rf vendor/*` and install again the dependencies `composer install`
2. Go to your Ads Account, choose one campaign and get the campaignID. The campaign ID can be found in the URL with the param `campaignId`
3. Go to the Asset Group Sections and choose an asset group.

image

5. Make a request to `GET ads/campaigns/CAMPAIGN_ID/assetgroups`
6. Compare the API results with the Asset Group in your Ads Account.
7. If you do not have assets, add some of them and compare the results. By default the asset group is empty when is created using the GL&A Plugin.

 
### Additional details:
 - For the MVP, the number of asset groups returned is limited to 1.
 - If there are multiple asset groups the Ads API returns by default the oldest Asset Group.


src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -0,0 +1,204 @@
...
+		parent::__construct( $rest_server );
+		$this->ads_asset_group = $ads_asset_group;
+	}
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'ads/campaigns/(?P<id>[\d]+)/assetgroups',
```
puntope: Some naming/location about this controller. Probably I'm missing concepts. But just to be sure :) 
 
❓  Why this file is not under Controller/Assets as `AssetSuggestionsController`
❓  I would name it under `assets/(?P<id>[\d]+)/groups` 

💅  In any case, instead of `assetgroups` I'd suggest `asset-groups`

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -0,0 +1,204 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\AdsAssetGroup;
+use WP_REST_Request as Request;
+use Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class AssetGroupController
```
puntope: 💅  Instead of adding the trivial `Class AssetGroupController` I'd suggest to add some explanation about the responsibility of this controller.

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -0,0 +1,204 @@
...
+						'long_headline'            => [
+							'type'     => 'array',
+							'items'    => $this->get_schema_asset(),
+							'required' => false,
+						],
+						'call_to_action_selection' => [
+							'type'     => 'array',
+							'items'    => $this->get_schema_asset(),
+							'required' => false,
+						],
```
puntope: 💅  What about wrap all of these into:

```php
... 
'long_headline'  => $this->get_schema_asset_group(),
'call_to_action_selection' => $this->get_schema_asset_group(),
... 

protected function get_schema_asset_group() {
 return [
   'type'     => 'array',
   'items'    => $this->get_schema_asset(), 
   'required' => false,
 ]
}



```

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+
+	/**
+	 * @return callable
+	 */
+	protected function get_asset_groups_by_campaign_id_callback(): callable {
+		return function( Request $request ) {
+			try {
+				$campaign_id = $request['id'];
```
puntope: 💅  Ideally we can use `$request->get_param('id);`

src/API/Google/AdsAssetGroup.php
```diff
@@ -190,11 +205,80 @@ protected function listing_group_delete_operations(): array {
...
+		$asset_group_results = ( new AdsAssetGroupQuery( [ 'pageSize' => self::PAGE_SIZE ] ) )
+			->set_client( $this->client, $this->options->get_ads_id() )
+			->add_columns( [ 'asset_group.path1', 'asset_group.path2', 'asset_group.id', 'asset_group.final_urls' ] )
+			->where( 'campaign.id', $campaign_id )
+			->where( 'asset_group.status', 'REMOVED', '!=' )
+			->get_results();
+
+		/** @var GoogleAdsRow $row */
+		foreach ( $asset_group_results->getPage()->getIterator() as $row ) {
+			$asset_groups_converted[ $row->getAssetGroup()->getId() ] = $this->convert_asset_group( $row );
```
puntope: ❓  Why we are saving the assets in `$asset_groups_converted` as an associative array if we return only the values ?

src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -0,0 +1,103 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\Ads\GoogleAdsClient;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
+use Google\Ads\GoogleAds\V11\Services\GoogleAdsRow;
+use Google\Ads\GoogleAds\V11\Resources\AssetGroupAsset;
+
+
+
+/**
+ * Class AdsAssetGroupAsset
```
puntope, line 19: 💅  Same as other files. It would be cool to add some brief explanation about the file.

src/API/Google/AssetFieldType.php
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Mapping between Google and internal AssetFieldTypes
+ * https://developers.google.com/google-ads/api/reference/rpc/v11/AssetFieldTypeEnum.AssetFieldType
+ *
+ * @since x.x.x
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Google
+ */
+class AssetFieldType extends StatusMapping {
```
puntope, line 19: 💅  WDYT about renaming this to `AssetFieldTypeMapping`?

src/API/Google/AdsAssetGroup.php
```diff
@@ -190,11 +205,80 @@ protected function listing_group_delete_operations(): array {
...
+	 * Get Asset Groups for a specific campaign.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param int  $campaign_id The campaign ID.
+	 * @param bool $include_assets Whether to include the assets in the response.
+	 *
+	 * @return array The asset groups for the campaign.
+	 */
+	public function get_asset_groups_by_campaign_id( int $campaign_id, bool $include_assets = true ): array {
```
puntope, line 230: ❓  In regards `$include_assets` param. Is there any case we don't want to get the assets?

src/API/Google/Query/AdsAssetGroupAssetQuery.php
```diff
@@ -0,0 +1,22 @@
...
+ * Class AdsAssetGroupAssetQuery
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query
+ */
+class AdsAssetGroupAssetQuery extends AdsQuery {
+
+	/**
+	 * Query constructor.
+	 */
+	public function __construct() {
```
puntope, line 18: 💅  Would be cool to ad some ref about the parameters available here in case another dev needs to update this query.

puntope:
Howdy @jorgemd24  I tested this PR. Thanks for your work here.

✅  I checked to get the assets of an existen campaign with no asset inside the asset group and it gave me this result;

```json
[
    {
        "id": 6445482928,
        "final_url": "",
        "display_url_path": [
            "",
            ""
        ],
        "assets": []
    }
]
```

✅  I checked to get the assets of an existen campaign with a single asset group and it gave me the right result.
✅  I checked to get the assets of an existen campaign with a two asset groups and it gave me the oldest asset group result.
✅  I checked to get a result of a non existent campaign and it gave me empty array []

💅  / ❓  I left some comments in regards the code

src/API/Google/AssetFieldType.php
jorgemd24: The other classes that are implementing `StatusMapping` do not use the suffix Mapping. See: 

- https://github.com/woocommerce/google-listings-and-ads/blob/0ffb057cbe73a3b1989bb35babed13899da1c4cf/src/API/Google/CampaignType.php#L17
- https://github.com/woocommerce/google-listings-and-ads/blob/0ffb057cbe73a3b1989bb35babed13899da1c4cf/src/API/Google/BillingSetupStatus.php#L15

So I think it is ok if we leave it as it is.

src/API/Google/AdsAssetGroup.php
jorgemd24: I think it is handy to leave the `id` as a `key` because it allows retrieving the `ids` easily using `array_keys( $asset_groups );`.  For example in this case: https://github.com/woocommerce/google-listings-and-ads/blob/2ac0403841d865a5530546d8c87cdccbc364d8bf/src/API/Google/AdsAssetGroup.php#L258

src/API/Google/AdsAssetGroup.php
jorgemd24: Currently, we are showing by default only one asset group including the assets however in the future if we decide to work with multiple asset groups maybe could be necessary to list all the asset groups linked to a campaign and later load the assets for a specific asset group. 

Also, I added the param because it was not adding complexity to the function and was making the function more flexible.

src/API/Google/Query/AdsAssetGroupAssetQuery.php
jorgemd24: Do you mean the methods and properties available from the `AdsQuery` and `Query` class?

src/API/Site/Controllers/Ads/AssetGroupController.php
jorgemd24: > ❓ Why this file is not under Controller/Assets as AssetSuggestionsController

I have some doubts about the location too, my thinking was Asset Groups are linked to a campaign, campaign are in the `Ads` directory, and asset groups belongs as well to the "Ads" section, and I was thinking as well about moving the `AssetSuggestionsController ` to the `Ads` directory. WDYT? Or do you think it is better to leave a specific folder for the `Assets`?

> ❓ I would name it under assets/(?P<id>[\d]+)/groups

Not sure about this one, because it seems that groups are under `assets` but the hierarchy is more like this:

Campaign -> Asset Group -> Asset Group Asset (the link between Asset Group and Asset) -> Assets. Also, the terminology in the Ads API is `Asset Group`.

> 💅 In any case, instead of assetgroups I'd suggest asset-groups

👍

jorgemd24:
Thanks, @puntope for reviewing this!

I addressed your comments and replied to your questions. I also realized that we could add a try/catch in the query requests, so I went ahead and I added it.

src/API/Google/Query/AdsAssetGroupAssetQuery.php
puntope: I was thinking about a link where you found the documentation for the available columns and params in the query.

puntope (approved):
✅  Approved

Thanks for the adjustments and answering my questions 👍

PR #1821, opened 2022-12-19
Tweak Copy for Attribute Mapping

### LINKED ISSUE #1813: Copy and styling issues in Attribute Mapping page
### Describe the bug:
Some copy needs to be revised in various places within the Attribute mapping page for better clarity.

The tooltip width needs to be shortened as well for better readability.
 

### Steps to reproduce:
1. Go to Google Listings & Ads > Attributes
2. Create an attribute rule
3. Select an attribute
4. Select the source field / set a default value
5. Set up category filters
6. Hover over tooltips

### Expected behavior:
I will share screenshots from the live site and type the correct text or correct format.

#### 1. Change text from `No data to display` to `You have no attribute rules`
Screenshot 2022-12-13 at 2 12 28 PM

Screenshot 2022-12-13 at 2 12 18 PM

#### 2. Change where it says `one` to `an` in dropdowns
Screenshot 2022-12-13 at 2 17 26 PM
Screenshot 2022-12-13 at 2 17 18 PM
Screenshot 2022-12-13 at 2 16 27 PM

#### 3. Correct typo and formatting of text in the category filter

The correct text is:
Apply to all categories
Apply to all categories EXCEPT
Apply ONLY to these categories

Group 2459

#### 4. Remove period for `Assign value` options. There is also an extra period to remove after _`Can't find an appropriate field?`_

Group 2

#### 5. Shorten width of all tooltips from 480px to 350px for better readability

image

image

#### 6. Correct some text in the tooltips

Remove the extra 'with'.

I'm looking at this text and now I think it's too wordy. Please change text to: `Auto-populate the target attribute with the value of the field you link it to.`

Group 3

Change the format of White to _'White'_ (with single quotes and italicized) if possible

image 7

puntope:

Closes #1813 .

This PR implements suggested changes from 2 to 6.

### Screenshots

Screenshot 2022-12-19 at 16 42 00

Screenshot 2022-12-19 at 16 41 53

Screenshot 2022-12-19 at 16 41 48

Screenshot 2022-12-19 at 16 41 43

1. Go to Attribute Mapping and click on Create new rule
2. See the selector appears as "Select an attribute"
3. Select Brands
4. See the new field selector appears as  "Select an option"
5. Check the tooltips are max 350px
6. Check there is no period after the copy sentences
7. Check that tooltip content in "Use value from existing product field" matches the specifications in #1813  point 6
8. Check that tooltip content in "Set a fixed value" matches the specifications in  #1813  point 6 especially the <em> markup in the word "White" 

### Additional details:

Notice point 1 was not implemented: Change text from No data to display to You have no attribute rules in the table when no rules

This is because `<Table>` component doesn't support changing that text. So we can use `<EmptyTable>` component but it will remove the table headers. 

### Changelog entry

> Tweak - Adjust copy in Attribute Mapping section

### DIFF
--- a/js/src/attribute-mapping/attribute-mapping-category-control.js
+++ b/js/src/attribute-mapping/attribute-mapping-category-control.js
@@ -35,21 +35,21 @@ const AttributeMappingCategoryControl = ( {
 					{
 						value: CATEGORY_CONDITION_SELECT_TYPES.ALL,
 						label: __(
-							'Apply to All categories',
+							'Apply to all categories',
 							'google-listings-and-ads'
 						),
 					},
 					{
 						value: CATEGORY_CONDITION_SELECT_TYPES.EXCEPT,
 						label: __(
-							'Apply to All categories EXCEPT',
+							'Apply to all categories EXCEPT',
 							'google-listings-and-ads'
 						),
 					},
 					{
 						value: CATEGORY_CONDITION_SELECT_TYPES.ONLY,
 						label: __(
-							'Apply ONLY to this categories',
+							'Apply ONLY to these categories',
 							'google-listings-and-ads'
 						),
 					},
--- a/js/src/attribute-mapping/attribute-mapping-field-sources-control.js
+++ b/js/src/attribute-mapping/attribute-mapping-field-sources-control.js
@@ -27,7 +27,7 @@ const AttributeMappingFieldSourcesControl = ( {
 					{
 						value: '',
 						label: __(
-							'Select one option',
+							'Select an option',
 							'google-listings-and-ads'
 						),
 					},
--- a/js/src/attribute-mapping/attribute-mapping-rule-modal.js
+++ b/js/src/attribute-mapping/attribute-mapping-rule-modal.js
@@ -118,7 +118,7 @@ const AttributeMappingRuleModal = ( { rule, onRequestClose = noop } ) => {
 	const attributesOptions = [
 		{
 			value: '',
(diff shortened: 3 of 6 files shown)


mikkamp (approved):
Thanks for the changes, I can confirm that each of those text changes is showing correctly now. The max width of the tooltips have also been reduced now.

As for the first point I initially thought we could try copying what's done [here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/product-feed/product-feed-table-card/index.js#L213). But apparently `EmptyTable` ignores the headers property.
What kind of effort would it be to submit a PR upstream which would change the component [here](https://github.com/woocommerce/woocommerce/blob/7.2.1/packages/js/components/src/table/table.js#L347-L350) to support either a value from a property or the hard coded message? While  it might not be available to us right away, eventually it could benefit others or our own extension.

puntope:
> What kind of effort would it be to submit a PR upstream which would change the component [here](https://github.com/woocommerce/woocommerce/blob/7.2.1/packages/js/components/src/table/table.js#L347-L350) to support either a value from a property or the hard coded message? While it might not be available to us right away, eventually it could benefit others or our own extension.

DEVELOPMENT: ~1h
WAITING FOR REVIEW: ~1 week

I will do it today so I can send it before AFK. I will open separate issue for it in Woo

puntope:
I created PR for the requested change @mikkamp 

https://github.com/woocommerce/woocommerce/pull/36124

Issue #1822, opened 2022-12-20
Improve landing page URL for policy compliance checks


mikkamp:
### Describe the bug:
When we retrieve the landing page URL, we do so by retrieving the latest product and checking it's URL returns a 200 status. If there are no products available then it falls back to using the home page.
However the default wc_get_products queries products with all statuses (including draft or private). If the last product happens to be private then it won't return a 200 status when we try to query the product.

The recommendation to fix this would be to query only products with `status = published`.

### Steps to reproduce:
1. Create only one product and set it to a private product. 
2. Go through the onboarding steps and confirm that the check "Confirm your store is live and accessible to all users" does not automatically complete.

### Expected behavior:
Always to check the store using a public product.

mikkamp:
This issue was resolved in PR #1825

PR #1825, opened 2022-12-21
Improve landing page URL for policy compliance checks 

songchen2022:

We should only consider the url for the products with publish status

Reference from this [issue](https://github.com/woocommerce/google-listings-and-ads/issues/1822)

We add this fix to help query only products with publish status.

1. Create only one product and set it to a publish product.
2. Go through the onboarding steps and confirm that the check "Confirm your store is live and accessible to all users" is passed.

### Changelog entry
* Tweak - Retrieve a published product as a landing page URL.
>

### DIFF
--- a/src/MerchantCenter/PolicyComplianceCheck.php
+++ b/src/MerchantCenter/PolicyComplianceCheck.php
@@ -105,7 +105,12 @@ class PolicyComplianceCheck implements Service {
 	 * @return string Landing page URL.
 	 */
 	private function get_landing_page_url(): string {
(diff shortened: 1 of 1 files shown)


mikkamp (approved):
Thanks for the change, I can confirm that I get a successful response when creating a private product.

The only suggestion I have is to add a changelog entry in the PR, something like this.
```
* Tweak - Retrieve a published product as a landing page URL.
```

songchen2022:
> 

Thanks for reviewing. I added the changelog as you suggested.

PR #251, opened 2021-02-25
Report endpoints for free listings and campaigns

mikkamp:
### Changes proposed in this Pull Request:

This PR adds two reports endpoints for the merchant account (free listings) and ad account (campaigns). In addition to the reports it also refactors the way search queries are run for both Ads and Merchant. The interface is similar to how the DB queries are constructed, but instead it builds GAQL queries and sends a request to the API to retrieve the results.

Closes #236 
Closes #237

### Notes
1. The date selector (copied from WooCommerce Admin) compares the current period with a previous period. That works fine here as well, but it means we need to request the report data for two date ranges (which can be significantly slower since we are requesting data from an API).
2. Because of potential delays in getting report data, should we implement a caching system? Or do we need a pre-fetch method?

### Pending points (still waiting on answers in Q&A doc)
1. The reports from ads allow segmenting by day/week/month/quarter/year whereas the merchant reports only allow segmenting by day
2. I'm not able to retrieve the product name for merchant reports
3. When segmenting by date and by product the merchant reports don't return all products

### Request parameters
Both report endpoints `/gla/mc/reports` or `/gla/ads/reports`support common parameters for retrieving the report data.

#### Report type
The type of report can be specified in the endpoint:
- `reports/programs` - Will segment the data by campaign (for Merchant reports they are all combined under `free_listings`)
- `reports/products` - Will segment the data by product

#### Pagination
This is not a traditional pagination, instead it limits the amount of rows to return and provides a next page token for retrieving the next batch of results. This can be used to breakdown the report data into batches. WCAdmin breaks this down into batches of a 100, I've set the default to 200 (Google will take longer to return the first page, but subsequent pages will be retrieved much faster).
`next_page` will be returned in the response data containing a token to retrieve another page. If this value is null, there are no more pages available.
- `per_page` - Set the limit of rows per page
- `next_page` - Set the next page token for retrieving another page of results

#### Filtering
Filtering can be done by providing a list of ids. Note that we use strings for product ids (currently in the format `gla_<wc_product_id>`). Note that for campaigns we can use either a string or an number, but campaign ids are always a positive integer.
- `ids` - Comma separated list of product or campaign ids

#### Ordering
- `order` - `asc` or `desc`
- `orderby` - Column we would like to order the results by, this must be one of the fields we are requesting

#### Interval
This allows us to specify how we want to segment the data, which is intended for populating a graph. Note that the Merchant data only supports segmenting by day, but the ads report allows the interval to be set to any of the following values:
[...]

PR #1826, opened 2022-12-21
Simplify report controller parameters

mikkamp:

Based on the discussion [here](https://github.com/woocommerce/automatewoo/pull/1226#discussion_r1054443474), this PR cleans up the report controllers to make the code reusable and take advantage of the existing function [WP_REST_Request::get_default_params](https://developer.wordpress.org/reference/classes/wp_rest_request/get_default_params/)

Unit testing is still not implemented for the Report Controllers (I started an attempt, but it's going to take a substantial effort to unit test these classes so I left it out for now).

The reports can be tested manually as outlined in PR #251, although this PR doesn't change the report data so we would only want to confirm that the parameter defaults still get mapped in the same way.

The easiest way to test that is to set a break point in `prepare_query_arguments` and confirm that the arguments are as expected:
image

> **Note**
The DateTime's will vary but it should cover the time range from `-7 days` to `now`.

A GET request without query parameters can be sent to `https://domain.test/wp-json/wc/gla/ads/reports/programs`.

### Changelog entry
* Tweak - Simplify report controller parameters.

### DIFF
--- a/src/API/Site/Controllers/BaseReportsController.php
+++ b/src/API/Site/Controllers/BaseReportsController.php
@@ -120,30 +120,16 @@ abstract class BaseReportsController extends BaseController {
 	 * @return array
 	 */
 	protected function prepare_query_arguments( Request $request ): array {
-		$params   = $this->get_collection_params();
-		$defaults = $this->get_defaults( $params );
-		$args     = wp_parse_args( array_intersect_key( $request->get_query_params(), $params ), $defaults );
+		$args = wp_parse_args(
+			array_intersect_key(
+				$request->get_query_params(),
+				$this->get_collection_params()
+			),
+			$request->get_default_params()
+		);
 
 		$this->normalize_timezones( $args );
 		return $args;
-	}
-
-	/**
-	 * Get parameter defaults.
-	 *
-	 * @param array $params List of parameters.
-	 *
-	 * @return array
-	 */
-	protected function get_defaults( array $params ): array {
-		$defaults = [];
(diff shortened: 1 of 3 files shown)


src/API/Site/Controllers/BaseReportsController.php
```diff
@@ -39,7 +39,7 @@ public function __construct( ContainerInterface $container ) {
 	 * @return array
 	 */
 	public function get_collection_params(): array {
-		$params = [
```
tomalec, line 42: :question: 
Shouldn't we call the parent here as well to get `context, page, per_page, and search` from there?
image

tomalec (approved):
Reviewed the code, smoke tested locally, and checked by fiddling with the debugger.
Left on a note, for a potential improvement https://github.com/woocommerce/google-listings-and-ads/pull/1826#discussion_r1057264122

src/API/Site/Controllers/BaseReportsController.php
mikkamp: I don't think we really can because while we try to reuse as much as possible from the parent controller, we get our data from the Google API instead of from a local DB. Which means some of those query parameters are different.

- context: we redefine this one, although we don't really use context much internally
- page: we can't reuse this because the Google API expects a next page token so we can't query a specific page
- per_page: we redefine this since the Google API sets some different limitations on how many records you can retrieve per request
- search: for the parent controller this is for searching for posts that match a particular search string (again this isn't really appropriate for our controllers)

PR #1832, opened 2022-12-29
Update Assets Groups

jorgemd24:

This PR allows updating the Asset Groups and their assets. The fields that are allowed to be updated in the asset group are:

- The Final URL - The Asset Group final URL.
- Path 1 - The first part of the text that may appear appended to the URL displayed in the ad.
- Path 2 - Second part of the text that may appear appended to the URL displayed in the ad. This field can only be set when path1 is set.
- And their assets.

Assets are [immutable](https://github.com/googleads/google-ads-php/blob/main/src/Google/Ads/GoogleAds/V11/Resources/Asset.php#L22-L25) and are linked to the Asset Group through the `AssetGroupAsset`, this means that in the case of updating an asset it is required to create a new `Asset` and its link in the `AssetGroupAsset`. 

### Get an Asset Group ID

1. As the branch `feature/pmax-assets` is using some new imports from the GoogleAds Library, you may need to remove your vendor folder `rm -Rf vendor/*` and install again the dependencies using `composer install`.
2. You can get an asset group id for a specific campaign by making a request to  `/gla/ads/campaigns/YOUR_CAMPAIGN_ID/asset-groups` and get the `id` from the response.

### Update Asset Group

1.  Make a request: `PUT gla/ads/campaigns/asset-groups/Your_AssetGroupID` with a JSON body.  The JSON body can be formed with the following fields:
```
{
"final_url": string, // The final URL and shopping merchant URL needs to have the same domain.
"path1": string,
"path2": string,
"assets": [
   {
      "id": string|null,
      "field_type":  string. // See enums values: https://github.com/woocommerce/google-listings-and-ads/pull/1832/files#diff-387aaa2a15a475b5e67f27d02bb3c638ca8c574f914a880bb4e95b54ce0a751cR236-R245
      "content": string|null. // If the field type is logo|square_marketing_image|marketing_images should be the url.
   }, 
   .... More assets
]
```
- If the `id` is null, it is considered as a new asset.
- If the `content` is null, it is considered as a deleted asset.
- If `id` and `content` are present, the asset will be updated.

4. Check that the Asset Group has been updated. You can check it in the Google Ads interface or call the endpoint: `/gla/ads/campaigns/YOUR_CAMPAIGN_ID/asset-groups`.

### Additional details:
- If there is an image with similar content loaded in the asset group previously, the Ads API will skip it.
- I realize that WCS has a maximum of 1 MB for POST requests, so when we are uploading multiple images, it will fail. The solution is increasing the maximum payload by adding the param `maxBytes` in the following line: https://github.com/Automattic/woocommerce-connect-server/blob/3f51d1ba3b90145137de5390366619621cfcd989/lib/google-proxy/index.js#L56-L58 . I will create a PR to increase this payload size.
- If you are updating images such as `logo`, `square_marketing_image` or `marketing_image` you need to use the right size image. You can use the endpoint `gla/assets/suggestions?id=POST_ID|TERM_ID&type=post|term` and it will suggest images with the right size if the post or the category has images.
- The image URL needs to be accessible. Localhost links may not work.
- See Assets requirements here: https://developers.google.com/google-ads/api/docs/performance-max/assets

### DIFF
--- a/src/API/Google/AdsAsset.php
+++ b/src/API/Google/AdsAsset.php
@@ -83,7 +83,7 @@ class AdsAsset implements OptionsAwareInterface {
 	 * @return MutateOperation The text asset mutation.
 	 * @throws Exception If the asset type is not supported.
 	 */
-	public function create_operation_asset( array $data, int $temporary_id = self::TEMPORARY_ID ): MutateOperation {
+	public function create_operation( array $data, int $temporary_id = self::TEMPORARY_ID ): MutateOperation {
 		$asset = new Asset(
 			[
 				'resource_name' => $this->temporary_resource_name( $temporary_id ),
--- a/src/API/Google/AdsAsset.php
+++ b/src/API/Google/AdsAsset.php
@@ -119,33 +119,39 @@ class AdsAsset implements OptionsAwareInterface {
 	}
 
 	/**
-	 * Convert Asset data to an array.
+	 * Returns the asset content for the given row.
 	 *
 	 * @param GoogleAdsRow $row Data row returned from a query request.
 	 *
-	 * @return array
+	 * @return string The asset content.
 	 */
-	public function convert_asset( GoogleAdsRow $row ): array {
+	public function get_asset_content( GoogleAdsRow $row ): string {
 		/** @var Asset $asset */
 		$asset = $row->getAsset();
 
 		switch ( $asset->getType() ) {
 			case AssetType::IMAGE:
-				$data = $asset->getImageAsset()->getFullSize()->getUrl();
-				break;
+				return $asset->getImageAsset()->getFullSize()->getUrl();
 			case AssetType::TEXT:
-				$data = $asset->getTextAsset()->getText();
-				break;
+				return $asset->getTextAsset()->getText();
 			case AssetType::CALL_TO_ACTION:
-				$data = CallToActionType::label( $asset->getCallToActionAsset()->getCallToAction() );
-				break;
+				return CallToActionType::label( $asset->getCallToActionAsset()->getCallToAction() );
 			default:
-				$data = '';
+				return '';
 		}
+	}
 
+	/**
+	 * Convert Asset data to an array.
+	 *
+	 * @param GoogleAdsRow $row Data row returned from a query request.
+	 *
+	 * @return array The asset data converted.
+	 */
+	public function convert_asset( GoogleAdsRow $row ): array {
 		return [
-			'id'      => $asset->getId(),
-			'content' => $data,
+			'id'      => $row->getAsset()->getId(),
+			'content' => $this->get_asset_content( $row ),
 		];
 	}
 }
--- a/src/API/Google/AdsAsset.php
+++ b/src/API/Google/AdsAsset.php
@@ -15,6 +15,7 @@ use Google\Ads\GoogleAds\Util\V11\ResourceNames;
 use Google\Ads\GoogleAds\V11\Common\TextAsset;
 use Google\Ads\GoogleAds\V11\Common\ImageAsset;
 use Google\Ads\GoogleAds\V11\Common\CallToActionAsset;
+use Exception;
 
 /**
  * Class AdsAsset
@@ -49,27 +50,64 @@ class AdsAsset implements OptionsAwareInterface {
 	}
 
 	/**
-	 * Returns a set of operations to create multiple text assets.
+	 * Returns the asset type for the given field type.
 	 *
-	 * @param array $asset The assets to use the text asset.
+	 * @param string $field_type The field type.
+	 *
+	 * @return int The asset type.
+	 * @throws Exception If the field type is not supported.
+	 */
+	protected function get_asset_type_by_field_type( $field_type ): int {
+		switch ( $field_type ) {
+			case AssetFieldType::LOGO:
+			case AssetFieldType::MARKETING_IMAGE:
+			case AssetFieldType::SQUARE_MARKETING_IMAGE:
+				return AssetType::IMAGE;
+			case AssetFieldType::CALL_TO_ACTION_SELECTION:
+				return AssetType::CALL_TO_ACTION;
+			case AssetFieldType::HEADLINE:
+			case AssetFieldType::LONG_HEADLINE:
+			case AssetFieldType::DESCRIPTION:
+			case AssetFieldType::BUSINESS_NAME:
+				return AssetType::TEXT;
+			default:
+				throw new Exception( 'Asset Field type not supported' );
+		}
+
+	}
+
+	/**
+	 * Returns an operations to create a text asset.
+	 *
+	 * @param array $data The assets to use the text asset.
 	 * @param int   $temporary_id The temporary ID to use for the asset.
 	 * @return MutateOperation The text asset mutation.
+	 * @throws Exception If the asset type is not supported.
 	 */
-	public function create_operation_text_asset( array $asset, int $temporary_id = self::TEMPORARY_ID ): MutateOperation {
-		return new MutateOperation(
+	public function create_operation_asset( array $data, int $temporary_id = self::TEMPORARY_ID ): MutateOperation {
+		$asset = new Asset(
 			[
-				'asset_operation' => new AssetOperation(
-					[
-						'create' => new Asset(
-							[
-								'resource_name' => $this->temporary_resource_name( $temporary_id ),
-								'text_asset'    => new TextAsset( [ 'text' => $asset['content'] ] ),
-							]
-						),
-					]
-				),
+				'resource_name' => $this->temporary_resource_name( $temporary_id ),
 			]
 		);
+
+		switch ( $this->get_asset_type_by_field_type( $data['field_type'] ) ) {
+			case AssetType::CALL_TO_ACTION:
+				$asset->setCallToActionAsset( new CallToActionAsset( [ 'call_to_action' => $data['content'] ] ) );
+				break;
+			case AssetType::IMAGE:
+				$asset->setImageAsset( new ImageAsset( [ 'data' => wp_remote_get( $data['content'] )['body'] ] ) );
+				$asset->setName( basename( $data['content'] ) );
+				break;
+			case AssetType::TEXT:
+				$asset->setTextAsset( new TextAsset( [ 'text' => $data['content'] ] ) );
+				break;
+			default:
+				throw new Exception( 'Asset type not supported' );
+		}
+
+		$operation = ( new AssetOperation() )->setCreate( $asset );
+		return ( new MutateOperation() )->setAssetOperation( $operation );
 
 	}
 
--- a/src/API/Google/AdsAsset.php
+++ b/src/API/Google/AdsAsset.php
@@ -14,6 +14,7 @@ use Google\Ads\GoogleAds\Util\V11\ResourceNames;
 use Google\Ads\GoogleAds\V11\Common\TextAsset;
 use Google\Ads\GoogleAds\V11\Common\ImageAsset;
 use Google\Ads\GoogleAds\V11\Common\CallToActionAsset;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
 use Exception;
 
 /**
@@ -29,6 +30,22 @@ use Exception;
 class AdsAsset implements OptionsAwareInterface {
 
 	use OptionsAwareTrait;
+
+	/**
+	 * WP Proxy
+	 *
+	 * @var WP
+	 */
+	protected WP $wp;
+
+	/**
+	 * AdsAsset constructor.
+	 *
+	 * @param WP $wp The WordPress proxy.
+	 */
+	public function __construct( WP $wp ) {
+		$this->wp = $wp;
+	}
 
 	/**
 	 * Temporary ID to use within a batch job.
@@ -97,7 +114,7 @@ class AdsAsset implements OptionsAwareInterface {
 				$asset->setCallToActionAsset( new CallToActionAsset( [ 'call_to_action' => CallToActionType::number( $data['content'] ) ] ) );
 				break;
 			case AssetType::IMAGE:
-				$image_data = wp_remote_get( $data['content'] );
+				$image_data = $this->wp->wp_remote_get( $data['content'] );
 
 				if ( is_wp_error( $image_data ) || empty( $image_data['body'] ) ) {
 					throw new Exception( 'Incorrect image asset url.' );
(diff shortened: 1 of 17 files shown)


src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -121,4 +133,84 @@ public function get_assets_by_asset_group_ids( array $asset_groups_ids ): array
...
+
+	/**
+	 * Edit assets group assets.
+	 *
+	 * @param int   $asset_group_id The asset group id.
+	 * @param array $assets The assets to create.
+	 *
+	 * @return array The asset group asset operations.
+	 */
+	public function edit_operations_assets_group_assets( int $asset_group_id, array $assets ): array {
```
puntope: 💅 

What do you think about renaming this as `edit_operations`? Otherwise we can rename  `create_operation` as `create_operation_assets_group_assets` and `delete_operation` as `delete_operation_assets_group_assets`.

In brief I was expecting some consistency in naming.

src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -121,4 +133,84 @@ public function get_assets_by_asset_group_ids( array $asset_groups_ids ): array
...
+				'asset_group' => ResourceNames::forAssetGroup( $this->options->get_ads_id(), $asset_group_id ),
+				'field_type'  => AssetFieldType::number( $asset_field_type ),
+			]
+		);
+
+		return ( new MutateOperation() )->setAssetGroupAssetOperation( $operation->setCreate( $new_asset_group_asset ) );
+	}
+
+	/**
+	 * Returns a campaign delete operation.
```
puntope: ❓  Shouldn't this be "Returns a delete operation for asset group asset"?

puntope:
❓ 

When I tried to create a new Call to Action I got 500 error

```json
{
"final_url": "https://805e-136-144-33-28.ngrok.io",
"path1": "test1",
"path2": "test2",
"assets": [
   {
      "id": null,
      "field_type": "call_to_action_selection",
      "content": "test"
   }
]
}
```

jorgemd24:
Thanks, @puntope for having a look at this.

> When I tried to create a new Call to Action I got 500 error.

The issue was caused because `array_search` returns a boolean if the item is not found, in this case `test` is not a valid value for the call_to_action field. I fix the issue here: [9a24355](https://github.com/woocommerce/google-listings-and-ads/pull/1832/commits/9a243553fa7c8862cc5a7ca5c7f8c1459184843f)

You can find the possible values for `call_to_action` here: https://github.com/woocommerce/google-listings-and-ads/blob/374fdf8888da4a47c232af4e0a6a3f15273265e9/src/API/Google/CallToActionType.php#L34-L95.

src/API/Google/AdsAssetGroupAsset.php
jorgemd24: That sounds right, I have changed it to `edit_operations`. See: [374fdf8](https://github.com/woocommerce/google-listings-and-ads/pull/1832/commits/374fdf8888da4a47c232af4e0a6a3f15273265e9)

puntope:
Hello @jorgemd24 I was testing again the call to action type but still I'm not able to do it. See screenshot. What I'm doing wrong? 

Screenshot 2023-01-04 at 17 27 11

jorgemd24:
@puntope,

The message is referring that the maximum number of assets for that field type (`call_to_action_selection`) is exceeded.  In this case, the maximum number for `call_to_action_selection` assets  is `1`, so probably the asset group that you are trying to update already has one `call_to_action_selection`, therefore is failing. Once is created you can update/delete it using the asset id. If you delete it then you will be able to create it again.

src/API/Google/AdsAsset.php
```diff
@@ -18,7 +26,97 @@
...
+	protected const TEMPORARY_ID = -5;
+
+	/**
+	 * Return a temporary resource name for the asset.
+	 *
+	 * @param int $temporary_id The temporary ID to use for the asset.
+	 *
+	 * @return string The Asset resource name.
+	 */
+	protected function temporary_resource_name( $temporary_id = self::TEMPORARY_ID ): string {
```
puntope: 💅  We can add the int type here

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -42,6 +43,17 @@ public function __construct( RESTServer $rest_server, AdsAssetGroup $ads_asset_g
 	 * Register rest routes with WordPress.
 	 */
 	public function register_routes(): void {
+		$this->register_route(
+			'ads/campaigns/asset-groups/(?P<id>[\d]+)',
```
puntope, line 47: ❓   Why here in the PUT the ID param is at the end and in the GET is before the asset-groups URL segment?

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -42,6 +43,17 @@ public function __construct( RESTServer $rest_server, AdsAssetGroup $ads_asset_g
...
+			[
+				[
+					'methods'             => TransportMethods::EDITABLE,
+					'callback'            => $this->edit_asset_group_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+					'args'                => $this->get_edit_params(),
```
puntope: 💅  For consistency maybe we could rename this to `edit_asset_group_params`

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -56,6 +68,50 @@ public function register_routes(): void {
...
+				'type'        => 'string',
+				'description' => __( 'Asset Group path 2.', 'google-listings-and-ads' ),
+			],
+		];
+	}
+
+	/**
+	 * Get the edit params to update an asset group.
+	 */
+	public function get_edit_params() {
```
puntope: 💅  Missed return type

src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -121,4 +133,84 @@ public function get_assets_by_asset_group_ids( array $asset_groups_ids ): array
...
 
 	}
+
+	/**
+	 * Edit assets group assets.
+	 *
+	 * @param int   $asset_group_id The asset group id.
+	 * @param array $assets The assets to create.
+	 *
+	 * @return array The asset group asset operations.
```
puntope, line 207: 💅  Seems like `create_operation` could potentially throw an extension which is not handled neither added to the phpdocs.

src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -44,6 +48,14 @@ class AdsAssetGroupAsset implements OptionsAwareInterface {
 	 */
```
puntope, line 47: In this file

`use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\AdsAsset;`
Is not being used.

src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -121,4 +133,84 @@ public function get_assets_by_asset_group_ids( array $asset_groups_ids ): array
...
+			// As Assets are inmmutable, we delete the old link between the asset and the asset group if exists.
+			if ( ! empty( $asset['id'] ) ) {
+				$delete_asset_group_assets_operations[] = $this->delete_operation( $asset_group_id, $asset['field_type'], $asset['id'] );
+			}
+		}
+
+		// Delete asset group assets operations must be executed last so we are never under the minimum quantity.
+		$operations = array_merge( $assets_operations, $asset_group_assets_operations, $delete_asset_group_assets_operations );
+
+		return $operations;
```
puntope: 💅  We can just return `array_merge( $assets_operations, $asset_group_assets_operations, $delete_asset_group_assets_operations )`

src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -121,4 +133,84 @@ public function get_assets_by_asset_group_ids( array $asset_groups_ids ): array
...
+	/**
+	 * Creates an operation for an asset group asset.
+	 *
+	 * @param int    $asset_group_id The ID of the asset group.
+	 * @param string $asset_field_type The field type of the asset.
+	 * @param int    $asset_id The ID of the asset.
+	 *
+	 * @return AssetGroupAssetOperation The create operation for the asset group asset.
+	 */
+	protected function create_operation( int $asset_group_id, string $asset_field_type, int $asset_id ): MutateOperation {
```
puntope, line 248: ❓  In the type it says MutateOperation, in the PHPDOC AssetGroupAssetOperation

src/API/Google/AdsAsset.php
```diff
@@ -18,7 +26,97 @@
...
+		switch ( $this->get_asset_type_by_field_type( $data['field_type'] ) ) {
+			case AssetType::CALL_TO_ACTION:
+				$asset->setCallToActionAsset( new CallToActionAsset( [ 'call_to_action' => CallToActionType::number( $data['content'] ) ] ) );
+				break;
+			case AssetType::IMAGE:
+				$image_data = wp_remote_get( $data['content'] );
+
+				if ( is_wp_error( $image_data ) || empty( $image_data['body'] ) ) {
+					throw new Exception( 'Incorrect image asset url.' );
+				}
```
puntope: ❓  Wondering if we can wrap this in the WP Proxy

src/API/Google/AdsAssetGroup.php
```diff
@@ -283,6 +284,69 @@ protected function get_assets( array $asset_groups ): array {
 		return $asset_groups;
 	}
```
puntope, line 284: 💅  `use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\AdsAssetGroupAsset;`

Not being used

src/API/Google/AdsAssetGroup.php
```diff
@@ -283,6 +284,69 @@ protected function get_assets( array $asset_groups ): array {
...
+
+			if ( ! empty( $data ) ) {
+				$operations[] = $this->edit_operation( $asset_group_id, $data );
+			}
+
+			if ( ! empty( $operations ) ) {
+				$this->client->getGoogleAdsServiceClient()->mutate( $this->options->get_ads_id(), $operations );
+			}
+
+			return $asset_group_id;
```
puntope, line 314: ❓ I take as granted that if the client request is not successful it will trow exception?

tests/Unit/API/Google/AdsAssetTest.php
```diff
@@ -0,0 +1,141 @@
...
+			'id'         => self::TEST_ASSET_ID,
+		];
+		$row  = $this->generate_asset_row_mock( $data );
+
+		$expected = [
+			'id'      => self::TEST_ASSET_ID,
+			'content' => $data['content'],
+		];
+
+		$this->assertEquals( $expected, $this->asset->convert_asset( $row ) );
```
puntope: 💅  Minor since it's a test. But potentially we could refactor these as 

`assertAssetTypeConversion( $type, $data)`

tests/Unit/API/Google/AdsAssetTest.php
```diff
@@ -0,0 +1,141 @@
...
+		$operation       = $this->asset->create_operation( $data, self::TEMPORARY_ID );
+		$asset_operation = $operation->getAssetOperation();
+
+		$this->assertTrue( $asset_operation->hasCreate() );
+		$this->assertEquals( ResourceNames::forAsset( $this->options->get_ads_id(), self::TEMPORARY_ID ), $asset_operation->getCreate()->getResourceName() );
+		$this->assertEquals( $data['content'], $asset_operation->getCreate()->getTextAsset()->getText() );
+
+	}
+
+	public function test_create_operation_call_to_action_asset() {
```
puntope, line 110: ❓ I guess we missed here Image type asset create operation test?

puntope (commented):
Howdy @jorgemd24 

Behaviour looking good. I did finally some deeper checks in regards code. Mostly minor. 

Thanks in advance.

src/API/Site/Controllers/Ads/AssetGroupController.php
jorgemd24: In the `PUT` the `id` is the `asset group id` and in the `GET` the `id` is the `campaign id`, however, I am considering renaming the `GET` endpoint ( in a different PR), so instead of using  `ads/campaigns/CAMPAIGN_ID/asset-groups` I think it is more convenient something like: `ads/campaigns/asset-groups?campaign_id=123445`, WDYT?

src/API/Google/AdsAssetGroupAsset.php
jorgemd24: I am not sure about this, what I can see the `MutateOperation` or any of their methods `setAssetGroupAssetOperation `or `setCreate`, would not throw an exception. The exception will throw when the mutating operations are executed in this line:  https://github.com/woocommerce/google-listings-and-ads/blob/1d9f8deda7f031738c8fa457948a65fae99a7ece/src/API/Google/AdsAssetGroup.php#L311

src/API/Google/AdsAsset.php
jorgemd24: I have added `wp_remote_get` to the WP proxy as it makes easier the testing for the Image assets, however, I am not sure if we should include throwing the `exception`  in the WP Proxy as it will depend on the consumer how it wants to treat the case where exists a `wp_error` or if we want to consider an empty body as an `Exception`. See [b7664f0](https://github.com/woocommerce/google-listings-and-ads/pull/1832/commits/b7664f060f631c6f110959f10667e94b7ad314f1)

src/API/Google/AdsAssetGroup.php
jorgemd24: Correct 👍

jorgemd24:
Thanks, @puntope for having a look at it. 

I have addressed your comments and replied to your questions. I guess this is ready for another round! 

Thanks!

src/API/Site/Controllers/Ads/AssetGroupController.php
puntope: Ahhh gotcha. I missed that. For me both options are valid tho.

src/API/Google/AdsAssetGroupAsset.php
puntope: What I meant is that inside `$this->asset->create_operation(`) we throw an exception in the default method. 

```
default:
			throw new Exception( 'Asset type not supported' );
```

This exception is not being handled inside `edit_operation` method. So either we handled that exception inside `edit_operation` or we can just add  `* @throws Exception` in PHPDOC

puntope (approved):
✅  Thanks for the changes @jorgemd24 

This is looking good for me

PR #1835, opened 2022-12-30
i18n for View Reports button

ecgan:

The "View Reports" button in the dashboard page was not wrapped in i18n function, so the text cannot be translated.

In this PR, we wrapped the text with i18n function `__()`, so that it can be translated.

### Screenshots:

image

1. Run `npm run i18n`.
2. Look into the generated `languages/google-listings-and-ads.pot` file. There should be a line with "View Reports".

### Additional details:

I found this when I was working on PR https://github.com/woocommerce/google-listings-and-ads/pull/1834.

### Changelog entry

> Fix - i18n for "View Reports" button.


eason9487 (approved):
LGTM.

PR #1797, opened 2022-11-29
Fetch WP Assets from specific term

jorgemd24:
### Changes proposed in this Pull Request:

Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/1780

This PR adds the functionality to return a list of Ads assets for a specific term/category.  The assets will be mapped using the below table:

Google text asset | Existing post/taxonomy meta
-- | --
Final URL | Permalink
(1) Headline Var1 Max. 15 chars |Term title
(2) Headline Var2 Max. 30 chars | Term title
(3) Long headline Max. 90 chars | Site title: Term title
(4) Short description Max. 60 chars | Term description
(5) Description Max. 90 chars | Term description, tagline
(5) Business name Max. 25 chars | Site title
(7) Call-to-action | –
(8) Display URL path | Slug

See original table: pcTzPl-12o-p2#comment-1849

- The Headlines (1 & 2) are grouped as one headline, as the Ads API only use one type which is [HEADLINE](https://github.com/googleads/google-ads-php/blob/151bab9bd37d46543bbeb2c3b0caedf91cb91b6b/src/Google/Ads/GoogleAds/V11/Enums/AssetFieldTypeEnum/AssetFieldType.php#L177).
- The Descriptions (4 & 5) are grouped as one description, as the Ads API only use one type which is [DESCRIPTION](https://github.com/googleads/google-ads-php/blob/151bab9bd37d46543bbeb2c3b0caedf91cb91b6b/src/Google/Ads/GoogleAds/V11/Enums/AssetFieldTypeEnum/AssetFieldType.php#L178).
- When fetching suggestions from WP/WC the field `call_to_action` will be `null` as it is not possible to offer a suggestion, however, when we suggest assets used in other campaigns using the Ads API, the field will be filled.

In addition to the mapping table, the endpoint returns the post's images linked to that term.

For instance, if the category Hoodies contains 5 products, it will show the images of those 5 products.

At the moment we have three types of images:

- Square Marketing images.
- Marketing Images (landscape).
- Logo

The endpoint will return the same images for Square and Landscape images until we decide how to resize/format the images. See comment: pcTzPl-12o-p2#comment-2090

### API Response:

```
{
    "headline": [
        "Hoodies"
    ],
    "long_headline": [
        "Wordpress: Hoodies"
    ],
    "description": [
        "Super Hoodies!",
        "I sell clothing!"
    ],
    "logo": [
[...]

PR #1795, opened 2022-11-27
Fetch WP Assets from specific post.

jorgemd24:
### Changes proposed in this Pull Request:

Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/1780

This PR adds a new endpoint `assets/suggestions` that returns a list of assets for a specific post (page/product/post/etc),  with the below mapping:

Google text asset | Existing post/taxonomy meta
-- | --
Final URL | Permalink
(1) Headline Var1 Max. 15 chars | Post/taxonomy title
(2) Headline Var2 Max. 30 chars | Post/taxonomy title
(3) Long headline Max. 90 chars | Site title: Post/taxonomy title
(4) Short description Max. 60 chars | Post excerpt/taxonomy description
(5) Description Max. 90 chars | Post excerpt/taxonomy description, tagline
(5) Business name Max. 25 chars | Site title
(7) Call-to-action | –
(8) Display URL path | Slug

See original table: pcTzPl-12o-p2#comment-1849

- The Headlines (1 & 2) are grouped as one headline, as the Ads API only use one type which is [HEADLINE](https://github.com/googleads/google-ads-php/blob/151bab9bd37d46543bbeb2c3b0caedf91cb91b6b/src/Google/Ads/GoogleAds/V11/Enums/AssetFieldTypeEnum/AssetFieldType.php#L177).
- The Descriptions (4 & 5) are grouped as one description, as the Ads API only use one type which is [DESCRIPTION](https://github.com/googleads/google-ads-php/blob/151bab9bd37d46543bbeb2c3b0caedf91cb91b6b/src/Google/Ads/GoogleAds/V11/Enums/AssetFieldTypeEnum/AssetFieldType.php#L178).
- When fetching suggestions from WP/WC the field `call_to_action` will be `null` as it is not possible to offer a suggestion, however when we suggest assets used in other campaigns using the Ads API, the field will be filled. 

In addition to the mapping table, the endpoint returns the image assets linked to the post and the website logo.

At the moment we have three types of images:
- Square Marketing images.
- Marketing Images (landscape).
- Logo

The endpoint will return the same images for Square and Landascape images until we decide how to to resize/format the images. See comment: pcTzPl-12o-p2#comment-2090

The images that will return for a post are a combination of:
- Images attached to the post.
- Gallery images
- In case that is a product, images from the product gallery.
- If the page is the Shop page, it will return images of published products. 

### API Response:

```
{
    "headline": [
        "Sample Page"
    ],
    "long_headline": [
[...]

PR #1837, opened 2023-01-03
Fetch assets from other Asset Groups using the final url

jorgemd24:

This PR allows populating assets (images, descriptions, headlines..) from other asset groups using the same final URL than the URL from a specific product, page, post or category. In case there are no assets available in other asset groups it will fetch the WP Assets.

See the following PRs for more context: 

- https://github.com/woocommerce/google-listings-and-ads/pull/1795
- https://github.com/woocommerce/google-listings-and-ads/pull/1797

1. Look for a post, page, product or category and get the ID. 
3. Make a request to `gla/assets/suggestions?type=post|term&id=POST_ID|TERM_ID`. If it is category or taxonomy the `type` is `term` otherwise it will be `post`
4. If there are not asset groups with the post/term URL, the response should include the assets included in the WP post or term.
5. If there was an asset group with the same `final_url` (see `final_url` from the previous response) it should show the assets from that asset group.
6. If you do not have any asset group using that `final_url`, you can go to [Google Ads ](https://ads.google.com/intl/en_EN/home/), choose a Campaign -> Asset Groups -> Edit Asset Group -> Update the final URL with the post or term URL that you got earlier.

image

7. Repeat the steps and you should get the Assets from the Asset group.

### Additional details:
- If there is more than one asset group using the same final URL, we would suggest the assets found in the first asset group returned from the Ads API, the reason for this is that we do not want to exceed the maximum allowed assets and to avoid repeated assets. See here all the assets requirements: https://developers.google.com/google-ads/api/docs/performance-max/assets.
- If you are using a local URL and you would like to change it temporarily so it can match with the URL in the asset group you can use the following filter:

```
add_filter('home_url', function ($url) {
   return  str_replace('http://localhost', 'https://test.ngrok.io', $url);
}, 10, 1);
```

### DIFF
--- a/src/API/Google/AdsAssetGroupAsset.php
+++ b/src/API/Google/AdsAssetGroupAsset.php
@@ -135,6 +135,51 @@ class AdsAssetGroupAsset implements OptionsAwareInterface {
 	}
 
 	/**
+	 * Get Assets for specific final URL.
+	 *
+	 * @param string $url The final url.
+	 *
+	 * @return array The assets for the asset groups with a specific final url.
+	 * @throws ExceptionWithResponseData When an ApiException is caught.
+	 */
+	public function get_assets_by_url( string $url ): array {
+		try {
+
+			$asset_group_assets = [];
+			$asset_results      = ( new AdsAssetGroupAssetQuery() )
+				->set_client( $this->client, $this->options->get_ads_id() )
+				->where( 'asset_group.final_urls', [ $url ], 'CONTAINS ANY' )
+				->where( 'asset_group_asset.field_type', $this->get_asset_field_types_query(), 'IN' )
+				->where( 'asset_group_asset.status', 'REMOVED', '!=' )
+				->get_results();
+
+			/** @var GoogleAdsRow $row */
+			foreach ( $asset_results->iterateAllElements() as $row ) {
+
+				/** @var AssetGroupAsset $asset_group_asset */
+				$asset_group_asset = $row->getAssetGroupAsset();
+
+				$field_type                          = AssetFieldType::label( $asset_group_asset->getFieldType() );
+				$asset_group_assets[ $field_type ][] = $this->asset->get_asset_content( $row );
+			}
+
+			return $asset_group_assets;
+		} catch ( ApiException $e ) {
+			do_action( 'woocommerce_gla_ads_client_exception', $e, __METHOD__ );
+
+			$errors = $this->get_api_exception_errors( $e );
+			throw new ExceptionWithResponseData(
+				/* translators: %s Error message */
+				sprintf( __( 'Error retrieving asset groups assets: %s', 'google-listings-and-ads' ), reset( $errors ) ),
+				$this->map_grpc_code_to_http_status_code( $e ),
+				null,
+				[ 'errors' => $errors ]
+			);
+		}
+
+	}
+
+	/**
 	 * Edit assets group assets.
 	 *
 	 * @param int   $asset_group_id The asset group id.
--- a/src/API/Google/AdsAssetGroupAsset.php
+++ b/src/API/Google/AdsAssetGroupAsset.php
@@ -146,9 +146,11 @@ class AdsAssetGroupAsset implements OptionsAwareInterface {
 		try {
 
 			$asset_group_assets = [];
-			$asset_results      = ( new AdsAssetGroupAssetQuery() )
+
+			// Search urls with and without trailing slash.
+			$asset_results = ( new AdsAssetGroupAssetQuery() )
 				->set_client( $this->client, $this->options->get_ads_id() )
-				->where( 'asset_group.final_urls', [ $url ], 'CONTAINS ANY' )
+				->where( 'asset_group.final_urls', [ rtrim( $url, '/' ), rtrim( $url, '/' ) . '/' ], 'CONTAINS ANY' )
 				->where( 'asset_group_asset.field_type', $this->get_asset_field_types_query(), 'IN' )
 				->where( 'asset_group_asset.status', 'REMOVED', '!=' )
 				->get_results();
--- a/src/API/Google/AdsAssetGroupAsset.php
+++ b/src/API/Google/AdsAssetGroupAsset.php
@@ -142,7 +142,7 @@ class AdsAssetGroupAsset implements OptionsAwareInterface {
 	 * @return array The assets for the asset groups with a specific final url.
 	 * @throws ExceptionWithResponseData When an ApiException is caught.
 	 */
-	public function get_assets_by_url( string $url ): array {
+	public function get_assets_by_final_url( string $url ): array {
 		try {
 
 			$asset_group_assets = [];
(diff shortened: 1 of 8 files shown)


src/API/Google/AdsAssetGroupAsset.php
```diff
@@ -134,6 +134,66 @@ public function get_assets_by_asset_group_ids( array $asset_groups_ids ): array
...
+	public function get_assets_by_final_url( string $url ): array {
+		try {
+
+			$asset_group_assets = [];
+
+			// Search urls with and without trailing slash.
+			$asset_results = ( new AdsAssetGroupAssetQuery() )
+				->set_client( $this->client, $this->options->get_ads_id() )
+				->add_columns( [ 'asset_group.id', 'asset_group.path1', 'asset_group.path2' ] )
+				->where( 'asset_group.final_urls', [ rtrim( $url, '/' ), rtrim( $url, '/' ) . '/' ], 'CONTAINS ANY' )
```
mikkamp: Can we use the [trailingslashit](https://developer.wordpress.org/reference/functions/trailingslashit/) and untrailingslashit functions here? They do pretty much the same thing.

src/API/Google/Query/Query.php
```diff
@@ -238,6 +238,7 @@ protected function validate_compare( string $compare ) {
 			case 'NOT IN':
 			case 'BETWEEN':
 			case 'IS NOT NULL':
+			case 'CONTAINS ANY':
```
mikkamp, line 241: Just curious, what's the difference between `CONTAINS ANY` and `IN`? I don't seem to find that mentioned in the [docs](https://developers.google.com/google-ads/api/docs/query/grammar).

src/Ads/AssetSuggestionsService.php
```diff
@@ -145,7 +216,7 @@ protected function get_post_assets( int $id ): array {
 
 		if ( ! $post || $post->post_status === 'trash' ) {
 			throw new Exception(
-				/* translators: 1: is an integer representing an unknown Post ID */
+			/* translators: 1: is an integer representing an unknown Post ID */
```
mikkamp: Is there a reason these 2 indents got removed? I prefer the indented version as the comment belongs to the `sprintf` line. There is also one more in this file which you left indented, so whatever format gets decided it's nice to stay consistent.

src/Ads/AssetSuggestionsService.php
```diff
@@ -111,10 +122,70 @@ public function __construct( WP $wp, WC $wc, ImageUtility $image_utility, wpdb $
...
+	 * @param int    $id Post or Term ID.
+	 * @param string $type Only possible values are post or term.
+	 */
+	protected function get_asset_group_asset_suggestions( int $id, string $type ): array {
+		$final_url = $this->get_url( $id, $type );
+
+		$assets = array_values( $this->asset_group_asset->get_assets_by_final_url( $final_url ) );
+
+		// Suggest the assets from the first asset group if exists.
+		$asset_group_assets = array_shift( $assets );
```
mikkamp: We don't really need to remove the indexes with array_values if all we are going to do is shift the first value off the array. My personal preference is also to use array_reset since it won't change the original array or trigger a reindex. Probably negligible in this case though.

Looking at this though if we are only interested in the first set of assets then couldn't we make that the responsibility of `AdsAssetGroupAsset` with a function like `get_first_asset_group_assets_by_final_url`? Or even just leaving it as `get_assets_by_final_url` is fine but then clarify in the docblock that it's getting the assets from the first asset group it encounters.

If we feel like we are going to reuse this more often we could create a specific helper (like [get_row](https://github.com/woocommerce/google-listings-and-ads/blob/2.3.5/src/DB/QueryInterface.php#L77-L82)) function in Query to only fetch the first result. This can be done by setting the [pageSize to 1](https://github.com/woocommerce/google-listings-and-ads/blob/2.3.5/src/API/Google/Query/ReportQueryTrait.php#L50). Or alternatively we can just process the first `$row` and return early, so we don't have to go through the process of converting all the assets.

mikkamp (commented):
Thanks for the PR, it's working good and I'm getting the results as expected.

I mainly left some minor comments. Although I think it would be nicer to move the responsibility of getting the first asset group. If we ever need to fetch all assets groups in the future, we can always refactor that with a specific function.

src/API/Google/Query/Query.php
jorgemd24: It depends on the field that you are querying,  for instance `asset_group.final_urls` does not allow the `IN` operator. If we try with the `IN` operator we get the following error: `IN is not a valid operator to use with 'asset_group.final_urls' in WHERE clause`. 

I guess that is related to how Google stores the field internally.

src/Ads/AssetSuggestionsService.php
jorgemd24: > We don't really need to remove the indexes with array_values if all we are going to do is shift the first value off the array. My personal preference is also to use array_reset since it won't change the original array or trigger a reindex. Probably negligible in this case though.

Thanks, yes array `reset` is a better option here.

> Looking at this though if we are only interested in the first set of assets then couldn't we make that the responsibility of AdsAssetGroupAsset with a function like get_first_asset_group_assets_by_final_url? Or even just leaving it as get_assets_by_final_url is fine but then clarify in the docblock that it's getting the assets from the first asset group it encounters.

> If we feel like we are going to reuse this more often we could create a specific helper (like [get_row](https://github.com/woocommerce/google-listings-and-ads/blob/2.3.5/src/DB/QueryInterface.php#L77-L82)) function in Query to only fetch the first result. This can be done by setting the [pageSize to 1](https://github.com/woocommerce/google-listings-and-ads/blob/2.3.5/src/API/Google/Query/ReportQueryTrait.php#L50). Or alternatively we can just process the first $row and return early, so we don't have to go through the process of converting all the assets.

I could move the responsibility however `get_assets_by_final_url ` will still need to process all assets because the query returns an asset for each `row` (each `row` contains and `asset` and `asset_group`) , so in this case using `pageSize = 1` or processing the first `row` would not work. That was the reason why I leave the responsibility to get the first asset group to the function `get_asset_group_asset_suggestions`. 

Another alternative, if we would like to move the responsibility, could be to add a param to the function `get_assets_by_final_url ` something like `get_assets_by_final_url( $url, $only_first_asset_group)`{ , so it is clear that we are specifically getting the assets from the first asset group.  WDYT?

src/Ads/AssetSuggestionsService.php
jorgemd24: This was unintentional - I changed it here: [d81a9bb](https://github.com/woocommerce/google-listings-and-ads/pull/1837/commits/d81a9bb1860f5d9c4d1afc7f854bf7a21cd8c3d2)

src/API/Google/AdsAssetGroupAsset.php
jorgemd24: Thanks, I have update it here: [8be6c67](https://github.com/woocommerce/google-listings-and-ads/pull/1837/commits/8be6c67f696ca4134bc1099565d4a923e07e6c12)

src/Ads/AssetSuggestionsService.php
mikkamp: Thanks for the clarification I see what you mean now, they are all being grouped by the asset group ID's and could be in a random order. I think the `$only_first_asset_group` option makes sense in this case.
That can either be implemented by then returning only the first array result or converting only the assets that match the first group we encounter (second option would be less processing, but probably negligible since we already fetched all assets anyway).

jorgemd24:
Thanks, @mikkamp for reviewing this PR. I added the param `$only_first_asset_group` to the function `get_assets_by_final_url ` and the other suggestions have been addressed.

mikkamp (approved):
Thanks for the additional changes, it's working as expected.

PR #1840, opened 2023-01-04
Tweak response asset group

jorgemd24:

Part of https://github.com/woocommerce/google-listings-and-ads/issues/1780

This PR changes the response format for the asset types: `BUSINESS_NAME` and `CALL_TO_ACTION_SELECTION` when making a request to `GET ads/campaigns/CAMPAIGN_ID/assetgroups`  , currently, it was returning an array of objects but I realized that is not needed to send an array as the maximum number of these assets is `1`.

I will add the test for these cases when the PR https://github.com/woocommerce/google-listings-and-ads/pull/1832 is merged, I created there a function that allows me to create the mocks easily. See https://github.com/woocommerce/google-listings-and-ads/pull/1832/files#diff-ce3988e96ebfe4e08831d1a3498e17570451c07f0b87b8b22460b2db0e638919R745

1. Follow the steps mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/1820
3. Check that the fields `BUSINESS_NAME` and `CALL_TO_ACTION_SELECTION` are objects and not an array of objects.
4. The Ads API does not return the `CALL_TO_ACTION_SELECTION` field if it is set as `Automated`, otherwise it will be included in the response.

### DIFF
--- a/src/API/Google/AdsAssetGroupAsset.php
+++ b/src/API/Google/AdsAssetGroupAsset.php
@@ -101,9 +101,16 @@ class AdsAssetGroupAsset implements OptionsAwareInterface {
 
 				/** @var AssetGroupAsset $asset_group_asset */
 				$asset_group_asset = $row->getAssetGroupAsset();
+				$field_type        = AssetFieldType::label( $asset_group_asset->getFieldType() );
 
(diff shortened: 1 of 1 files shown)


puntope (approved):
✅  LGTM

It works as expected. Thanks @jorgemd24

PR #1844, opened 2023-01-09
Prevent the "import/no-duplicates" warning for the externalization process

eason9487:

Quotes from @ecgan's https://github.com/woocommerce/google-listings-and-ads/pull/1834#issuecomment-1373854933:

> Three of the external-components files have eslint warnings that say "node_modules/@wordpress/components/build-module/index.js' imported multiple times. eslint [import/no-duplicates](https://github.com/import-js/eslint-plugin-import/blob/v2.26.0/docs/rules/no-duplicates.md)", because we have imports from both `@wordpress/components` and `extracted/@wordpress/components`. See screenshot below:
> 
> image

This PR changes to use the "no-duplicate-imports" rule to prevent the warnings.

### Screenshots:

It can still find the basic duplicated imports:

image

1. `npm run lint:js`
2. Check if there are no "import/no-duplicates" warnings.

### Additional details:

"no-duplicate-imports" doesn't do the actual file path resolving, so something like `import Foo from '. /foo'` and `import Foo from '. /foo.js'` are not recognized as duplicates.

However, such rules are for code clearer purposes only, and these undetected duplicates should not cause errors.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,5 +40,14 @@ module.exports = {
 				additionalHooks: 'useSelect',
 			},
 		],
+		// When a file imports from both `@wordpress/components` and `extracted/@wordpress/components`
+		// at the same time, it would get the "no-duplicates" warning. It should be considered a
+		// false positive when working on the externalization process. So here we temporarily change
+		// to use "no-duplicate-imports" instead.
+		//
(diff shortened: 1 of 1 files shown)


puntope (approved):
Thanks @eason9487 

When I run the lint I cannot see anymore the warnings.

PR #1847, opened 2023-01-10
PMax - Assets for non static homepage

jorgemd24:

Until now it was only possible to fetch assets from the homepage if it was linked to a static page (post, page, etc). This PR adds the functionality of getting assets from a non-static homepage by adding a new asset suggestion type: `homepage`.

In case that is a non-static homepage, the assets will be populated using one of the below options:

- Get assets from other asset groups  (see https://github.com/woocommerce/google-listings-and-ads/pull/1837), 
- A list of default assets.

1. Go to WP Settings -> Reading -> Select "Your latest posts" which means that you are not selecting a static homepage.
2. Make a GET request: `wc/gla/assets/final-url/suggestions` or search for `homepage` (`wc/gla/assets/final-url/suggestions?search=homepage`),  it should return the homepage with the new `homepage type`: 

```
    {
        "id": 0,
        "type": "homepage", 
        "title": "Homepage",
        "url": "https://example.io"
    },
```

3. Make a GET request  `gla/assets/suggestions?id=0&type=homepage`
4. If you have an asset group with the same homepage URL, it should return the assets from that asset group.
5. If you do not have an asset group it should return a list of default assets.
6. If you change it to a static homepage and there are no asset groups with the homepage URL, it should return the assets from the static page.

### Additional details:
- In the case there are more than one asset group with the same URL, it will show the assets from only one asset group. See https://github.com/woocommerce/google-listings-and-ads/pull/1837
- If you are using a local URL and you would like to change it temporarily so it can match with the URL in the asset group you can use the following filter:
```
add_filter('home_url', function ($url) {
   return  str_replace('http://localhost', 'https://test.ngrok.io', $url);
}, 10, 1);
```

### DIFF
--- a/src/API/Site/Controllers/Ads/AssetSuggestionsController.php
+++ b/src/API/Site/Controllers/Ads/AssetSuggestionsController.php
@@ -110,8 +110,10 @@ class AssetSuggestionsController extends BaseController {
 		return [
 			'id'   => [
 				'description'       => __( 'Post ID or Term ID.', 'google-listings-and-ads' ),
-				'type'              => [ 'integer', 'null' ],
+				'type'              => 'number',
+				'sanitize_callback' => 'absint',
 				'validate_callback' => 'rest_validate_request_arg',
+				'required'          => true,
 			],
 			'type' => [
 				'description'       => __( 'Type linked to the id.', 'google-listings-and-ads' ),
@@ -119,6 +121,7 @@ class AssetSuggestionsController extends BaseController {
 				'sanitize_callback' => 'sanitize_text_field',
 				'enum'              => [ 'post', 'term', 'homepage' ],
 				'validate_callback' => 'rest_validate_request_arg',
+				'required'          => true,
 			],
 		];
 	}
--- a/src/API/Site/Controllers/Ads/AssetSuggestionsController.php
+++ b/src/API/Site/Controllers/Ads/AssetSuggestionsController.php
@@ -179,7 +179,7 @@ class AssetSuggestionsController extends BaseController {
 			],
 			'type'  => [
 				'type'        => 'string',
-				'description' => __( 'Post or term', 'google-listings-and-ads' ),
+				'description' => __( 'Post, term or homepage', 'google-listings-and-ads' ),
 				'context'     => [ 'view' ],
 				'enum'        => [ 'post', 'term', 'homepage' ],
 				'readonly'    => true,
--- a/src/API/Site/Controllers/Ads/AssetSuggestionsController.php
+++ b/src/API/Site/Controllers/Ads/AssetSuggestionsController.php
@@ -110,15 +110,14 @@ class AssetSuggestionsController extends BaseController {
 		return [
 			'id'   => [
 				'description'       => __( 'Post ID or Term ID.', 'google-listings-and-ads' ),
-				'type'              => 'number',
-				'sanitize_callback' => 'absint',
+				'type'              => [ 'integer', 'null' ],
 				'validate_callback' => 'rest_validate_request_arg',
 			],
 			'type' => [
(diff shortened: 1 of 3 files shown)


src/API/Site/Controllers/Ads/AssetSuggestionsController.php
```diff
@@ -179,7 +181,7 @@ protected function get_schema_properties(): array {
 				'type'        => 'string',
 				'description' => __( 'Post or term', 'google-listings-and-ads' ),
```
puntope: 💅  This I guess could be updated to "Post, term or homepage"

src/Ads/AssetSuggestionsService.php
```diff
@@ -134,20 +139,22 @@ public function get_assets_suggestions( int $id, string $type ): array {
...
 			$url = get_permalink( $id );
-		} else {
+		} elseif ( $type === 'term' ) {
 			$url = get_term_link( $id );
+		} else {
+			$url = get_bloginfo( 'url' );
 		}
 
-		if ( $url === false || is_wp_error( $url ) ) {
+		if ( $url === false || is_wp_error( $url ) || empty( $url ) ) {
```
puntope: 💅  since [empty()](https://www.php.net/manual/en/function.empty.php) checks also if $url is falsy we can consider to  omit `$url === false`

src/Ads/AssetSuggestionsService.php
```diff
@@ -134,20 +139,22 @@ public function get_assets_suggestions( int $id, string $type ): array {
...
 			throw new Exception(
 				/* translators: 1: is an integer representing an unknown Term ID */
 				sprintf( __( 'Invalid Term ID or Post ID %1$d', 'google-listings-and-ads' ), $id )
```
puntope: 💅  This could be also an invalid/empty site URL

src/Ads/AssetSuggestionsService.php
```diff
@@ -187,17 +194,48 @@ protected function get_asset_group_asset_suggestions( int $id, string $type ): a
...
+	 * @throws Exception If the homepage id is invalid.
+	 */
+	protected function get_homepage_assets(): array {
+		$home_page = $this->wp->get_static_homepage();
+
+		// Static homepage.
+		if ( $home_page ) {
+			return $this->get_post_assets( $home_page->ID );
+		}
+		// Non static homepage.
```
puntope, line 226: 💅  I guess those are default values. Since now static homepage will be potentially overrided with the Asset Group with the same URL

src/Ads/AssetSuggestionsService.php
```diff
@@ -187,17 +194,48 @@ protected function get_asset_group_asset_suggestions( int $id, string $type ): a
...
 	 *
-	 * @param int    $id Post or Term ID.
+	 * @param int    $id Post or Term ID, or self::HOMEPAGE_KEY_ID.
 	 * @param string $type Only possible values are post or term.
 	 *
-	 * @return array All assets available for specific term or post.
+	 * @return array All assets available for specific term, post or homepage.
 	 */
 	protected function get_wp_assets( int $id, string $type ): array {
 		if ( $type === 'post' ) {
```
puntope, line 204: Missing either adding `* @throws Exception` or handling the Exception in the function

puntope (approved):
✅  LGTM

✅  If you have an asset group with the same homepage URL, it should return the assets from that asset group.
✅  If you do not have an asset group it should return a list of default assets.
✅  If you change it to a static homepage and there are no asset groups with the homepage URL, it should return the assets from the static page.

I left some minor comments mainly regarding docs.

jorgemd24:
Thanks, @puntope for taking the time to review this PR. 

I addressed your suggestions 👍

PR #1850, opened 2023-01-12
Preselect first existing merchant account during onboarding

### LINKED ISSUE #1639: Preselect first existing merchant account during onboarding
## User story
We are going to tryout with some flows where the merchant account is created before going through onboarding. One point of friction is when the user is selecting a Merchant Center account, as they may have forgotten/not realized that an account was already created for them previously.

Since we don't have the data available to select the last created account the other option is to select the first existing account in the list. This means the step will change from:

image

to: 

image

> **Note**
If it's a new account it might not have the name or website linked yet.

## Additional
Detailed information pcTzPl-PH-p2

puntope:

Closes #1639 

- Preselect the first account (after sorting) in the onboarding select MC account Step
- Added tests

1. Go to Onboarding 1st step
2. Connect Wordpress.com and Google account
3. In the MC account see that the first account is selected
4. Check connect button is enabled you can connect to this account without changing the selector
5. Check you can connect with other accounts in the selector

### Additional details:

Even this is a very simple change, the main challenge was to trigger the selected value in the parent component. I thought about moving the `useExistingGoogleMCAccounts`  hook at the top, but at the end I decided to have a useEffect that calls the onChange when the value is not set. In the practice, this is only in the init. 

💅  Please if you have a more elegant solution just comment in the PR 

### Changelog entry

> Tweak - Pre-select a default MC account

### DIFF
--- a/js/src/components/merchant-center-select-control/index.js
+++ b/js/src/components/merchant-center-select-control/index.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { __, sprintf } from '@wordpress/i18n';
+import { useEffect } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -9,9 +10,18 @@ import { __, sprintf } from '@wordpress/i18n';
 import useExistingGoogleMCAccounts from '.~/hooks/useExistingGoogleMCAccounts';
 import AppSelectControl from '.~/components/app-select-control';
 
-const MerchantCenterSelectControl = ( props ) => {
+/**
+ * @param {Object} props The component props
+ * @param {string} [props.value] The selected value. IF no value is defined, then the first option is selected and onChange function is triggered.
+ * @param {Function} [props.onChange] Callback when the select value changes.
+ * @return {JSX.Element} An enhanced AppSelectControl component.
+ */
+const MerchantCenterSelectControl = ( {
+	value,
+	onChange = () => {},
+	...props
+} ) => {
 	const { data: existingAccounts = [] } = useExistingGoogleMCAccounts();
-
 	const options = existingAccounts.map( ( acc ) => {
 		return {
 			value: acc.id,
@@ -27,6 +37,13 @@ const MerchantCenterSelectControl = ( props ) => {
 	options.sort( ( a, b ) => {
 		return a.label.localeCompare( b.label );
 	} );
+
+	useEffect( () => {
+		// Triggers the onChange event in order to pre-select the initial value
+		if ( value === undefined ) {
+			onChange( options[ 0 ].value );
+		}
+	}, [ options, onChange, value ] );
 
 	return <AppSelectControl options={ options } { ...props } />;
 };
--- a/js/src/components/merchant-center-select-control/index.js
+++ b/js/src/components/merchant-center-select-control/index.js
@@ -45,7 +45,14 @@ const MerchantCenterSelectControl = ( {
 		}
 	}, [ options, onChange, value ] );
 
-	return <AppSelectControl options={ options } { ...props } />;
+	return (
+		<AppSelectControl
+			options={ options }
+			onChange={ onChange }
+			value={ value }
+			{ ...props }
+		/>
+	);
 };
 
 export default MerchantCenterSelectControl;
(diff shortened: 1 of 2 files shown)


js/src/components/merchant-center-select-control/index.js
```diff
@@ -2,16 +2,26 @@
...
+ * @param {Object} props The component props
+ * @param {string} [props.value] The selected value. IF no value is defined, then the first option is selected and onChange function is triggered.
+ * @param {Function} [props.onChange] Callback when the select value changes.
+ * @return {JSX.Element} An enhanced AppSelectControl component.
+ */
+const MerchantCenterSelectControl = ( {
+	value,
+	onChange = () => {},
+	...props
+} ) => {
```
jorgemd24, line 23: Because we are extracting the `value` and the `onChange` params from the `props`, the `AppSelectControl` does not receive the `value` nor `onChange` therefore when you update the dropdown to choose a different account, the MC account is not updated.

js/src/components/merchant-center-select-control/index.test.js
```diff
@@ -0,0 +1,52 @@
...
+	} );
+
+	test( 'When a value is defined, it doesnt call onChange init method', () => {
+		const onChange = jest.fn().mockName( 'onChange' );
+		render(
+			<MerchantCenterSelectControl value="2" onChange={ onChange } />
+		);
+		expect( onChange ).not.toHaveBeenCalled();
+	} );
+} );
```
jorgemd24, line 61: I think we are missing a test for updating the select component with a different option. For example, if the option selected is `1`, and if we change it to option `2`  the `onChange` should be triggered.

js/src/components/merchant-center-select-control/index.js
```diff
@@ -27,10 +37,13 @@ const MerchantCenterSelectControl = ( props ) => {
...
-	} );
+
+	useEffect( () => {
+		// Triggers the onChange event in order to pre-select the initial value
+		if ( value === undefined ) {
+			onChange( options[ 0 ].value );
+		}
+	}, [ options, onChange, value ] );
 
 	return <AppSelectControl options={ options } { ...props } />;
```
jorgemd24: ~~What do you think if we do the following to add the default value:~~

```suggestion
	return		<AppSelectControl
			options={ options }
			{ ...props }
			value={
				props.value === undefined ? options[ 0 ].value : props.value
			}
		/>;
```

I see that it would not work because of the parent component would not get updated. 

Ideally, we could just set the default value here:
https://github.com/woocommerce/google-listings-and-ads/blob/51210551db7d2950008acd408b3615b113d3cab9/js/src/components/google-mc-account-card/connect-mc/index.js#L42

However, that would require moving the logic and the `useExistingGoogleMCAccounts` hook to the parent component, otherwise, I think your solution would work if we pass the `onChange` and the `value` prop to the `AppSelectControl`

jorgemd24 (commented):
Howdy @puntope,

Thanks for working on this. I was testing this PR but I saw that when the user chooses a different option, the `value` does not get updated therefore the user would choose the wrong MC account.

See video:


https://user-images.githubusercontent.com/2488994/212089874-42eb6261-f2a1-41d7-b94f-f05a39219e53.mp4

puntope:
Thanks for Call catching the error in the props @jorgemd24 I forgot to add them after refactoring the variables.

Now it should work as expected. This is ready for another round.

Ps. I also added the suggested test

js/src/components/merchant-center-select-control/index.js
```diff
@@ -27,12 +37,22 @@ const MerchantCenterSelectControl = ( props ) => {
...
-	options.unshift( {
-		value: '',
-		label: __( 'Select one', 'google-listings-and-ads' ),
-	} );
 
-	return <AppSelectControl options={ options } { ...props } />;
+	useEffect( () => {
+		// Triggers the onChange event in order to pre-select the initial value
+		if ( value === undefined ) {
+			onChange( options[ 0 ].value );
```
jorgemd24: I know that when the options are empty this component is not rendered, but we could change it to
```suggestion
			onChange( options[ 0 ]?.value );
```

So if this component is reused and the options are empty, the component will not break.

jorgemd24 (approved):
@puntope,

Thanks for the further adjustments, it is now working as expected. I have just left a minor comment.

PR #1851, opened 2023-01-13
Add Endpoint- Create Asset Group

jorgemd24:

This PR adds a new endpoint to create asset groups for a specific ad campaign. By default, all campaigns are created with an asset group but the user can remove the asset group later on, so in case the user removes all asset groups we will be able to create a new one if needed. 

Also, this PR has updated the endpoint route for getting assets from asset groups, the route has been changed from `ads/campaigns/CAMPAIGN_ID/asset-groups` to `ads/campaigns/asset-groups?campaign_id=CAMPAIGN_ID`, I made the change because I think the endpoint name will be more aligned with the other controllers that we have. This endpoint is not yet used for the front-end.

1. As the branch `feature/pmax-assets` is using some new imports from the GoogleAds Library, you may need to remove your vendor folder `rm -Rf vendor/*` and install again the dependencies using `composer install`.
2. Make a request to `POST wc/gla/ads/campaigns/asset-groups` with the following body:
```
{
	"campaign_id" : YOUR_CAMPAIGN_ID
}
```
3. The response should look likes this:
```
{
    "status": "success",
    "message": "Successfully created asset group.",
    "id": 6460020972
}
```
4. Check in your Ads account if the asset group has been created or make a request to `GET ads/campaigns/asset-groups?campaign_id=YOUR_CAMPAIGN_ID` (the endpoint only returns one asset group)

### DIFF
--- a/src/API/Google/AdsAssetGroup.php
+++ b/src/API/Google/AdsAssetGroup.php
@@ -96,7 +96,7 @@ class AdsAssetGroup implements OptionsAwareInterface {
 	 * @return id The asset group id.
 	 * @throws ExceptionWithResponseData ExceptionWithResponseData When an ApiException is caught.
 	 */
-	public function create_asset_group( $campaign_id ): int {
+	public function create_asset_group( int $campaign_id ): int {
 		try {
 			$campaign_resource_name = ResourceNames::forCampaign( $this->options->get_ads_id(), $campaign_id );
 			$current_date_time      = ( new DateTime( 'now', wp_timezone() ) )->format( 'Y-m-d H:i:s' );
@@ -439,7 +439,7 @@ class AdsAssetGroup implements OptionsAwareInterface {
 	 *
 	 * @param string $name Resource name containing ID number.
 	 *
-	 * @return int
+	 * @return int The asset group ID.
 	 * @throws Exception When unable to parse resource ID.
 	 */
 	protected function parse_asset_group_id( string $name ): int {
--- a/src/API/Google/AdsAssetGroup.php
+++ b/src/API/Google/AdsAssetGroup.php
@@ -93,7 +93,7 @@ class AdsAssetGroup implements OptionsAwareInterface {
 	 *
 	 * @param int $campaign_id
 	 *
-	 * @return int The asset group ID.
+	 * @return id The asset group id.
 	 * @throws ExceptionWithResponseData ExceptionWithResponseData When an ApiException is caught.
 	 */
 	public function create_asset_group( $campaign_id ): int {
@@ -101,25 +101,20 @@ class AdsAssetGroup implements OptionsAwareInterface {
 			$campaign_resource_name = ResourceNames::forCampaign( $this->options->get_ads_id(), $campaign_id );
 			$current_date_time      = ( new DateTime( 'now', wp_timezone() ) )->format( 'Y-m-d H:i:s' );
 			$name                   = sprintf(
-			/* translators: %s: current date time. */
-				__( 'Asset Group %s', 'google-listings-and-ads' ),
+				/* translators: %s: current date time. */
+				__( 'PMax %s', 'google-listings-and-ads' ),
 				$current_date_time
 			);
 
-			$operations     = $this->create_operations( $campaign_resource_name, $name );
-			$asset_group_id = $this->mutate( $operations );
-
-			return [
-				'id'   => $asset_group_id,
-				'name' => $name,
-			];
+			$operations = $this->create_operations( $campaign_resource_name, $name );
+			return $this->mutate( $operations );
 
 		} catch ( ApiException $e ) {
 			do_action( 'woocommerce_gla_ads_client_exception', $e, __METHOD__ );
 
 			$errors = $this->get_api_exception_errors( $e );
 			throw new ExceptionWithResponseData(
-			/* translators: %s Error message */
+				/* translators: %s Error message */
 				sprintf( __( 'Error creating asset group: %s', 'google-listings-and-ads' ), reset( $errors ) ),
 				$this->map_grpc_code_to_http_status_code( $e ),
 				null,
@@ -357,7 +352,7 @@ class AdsAssetGroup implements OptionsAwareInterface {
 			}
 
 			if ( ! empty( $operations ) ) {
-				$this->client->getGoogleAdsServiceClient()->mutate( $this->options->get_ads_id(), $operations );
+				$this->mutate( $operations );
 			}
 
 			return $asset_group_id;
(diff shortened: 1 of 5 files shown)


src/API/Google/AdsAssetGroup.php
```diff
@@ -82,17 +86,56 @@ public function __construct( GoogleAdsClient $client, AdsAssetGroupAsset $asset_
...
+	 * @param int $campaign_id
+	 *
+	 * @return id The asset group id.
+	 * @throws ExceptionWithResponseData ExceptionWithResponseData When an ApiException is caught.
+	 */
+	public function create_asset_group( int $campaign_id ): int {
+		try {
+			$campaign_resource_name = ResourceNames::forCampaign( $this->options->get_ads_id(), $campaign_id );
+			$current_date_time      = ( new DateTime( 'now', wp_timezone() ) )->format( 'Y-m-d H:i:s' );
+			$name                   = sprintf(
```
puntope: 💅  I'd personally would change `$name` to `$asset_group_name` since $name is quite generic. (Especially for the function signature)

src/API/Google/AdsAssetGroup.php
```diff
@@ -82,17 +86,56 @@ public function __construct( GoogleAdsClient $client, AdsAssetGroupAsset $asset_
...
 	/**
 	 * Returns a set of operations to create an asset group.
 	 *
 	 * @param string $campaign_resource_name
-	 * @param string $campaign_name
+	 * @param string $name The asset group name.
 	 * @return array
 	 */
-	public function create_operations( string $campaign_resource_name, string $campaign_name ): array {
+	public function create_operations( string $campaign_resource_name, string $name ): array {
```
puntope: Same as in https://github.com/woocommerce/google-listings-and-ads/pull/1851/files#r1071013296

💅  I'd personally would change `$name` to `$asset_group_name` since $name is quite generic.

src/API/Google/AdsAssetGroup.php
```diff
@@ -82,17 +86,56 @@ public function __construct( GoogleAdsClient $client, AdsAssetGroupAsset $asset_
...
 	}
 
+	/**
+	 * Create an asset group.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param int $campaign_id
+	 *
+	 * @return id The asset group id.
```
puntope: 💅 Missing type in the return directive

src/API/Google/AdsAssetGroup.php
```diff
@@ -82,17 +86,56 @@ public function __construct( GoogleAdsClient $client, AdsAssetGroupAsset $asset_
...
+			$name                   = sprintf(
+				/* translators: %s: current date time. */
+				__( 'PMax %s', 'google-listings-and-ads' ),
+				$current_date_time
+			);
+
+			$operations = $this->create_operations( $campaign_resource_name, $name );
+			return $this->mutate( $operations );
+
+		} catch ( ApiException $e ) {
```
puntope: 💅  Since DateTime constructor could potentially throw an Exception maybe we could change this `ApiException` to just `Exception` and check the instance of exception in order to send it to the `get_api_exception_errors`

src/API/Google/AdsAssetGroup.php
```diff
@@ -362,4 +405,50 @@ protected function convert_asset_group( GoogleAdsRow $row ): array {
...
+	protected function mutate( array $operations ): int {
+		$responses = $this->client->getGoogleAdsServiceClient()->mutate(
+			$this->options->get_ads_id(),
+			$operations
+		);
+
+		foreach ( $responses->getMutateOperationResponses() as $response ) {
+			if ( 'asset_group_result' === $response->getResponse() ) {
+				$asset_group_result = $response->getAssetGroupResult();
+				return $this->parse_asset_group_id( $asset_group_result->getResourceName() );
```
puntope, line 438: 💅  This could throw an Exception but only ApiException is being reflected in the PHPDOCS

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -121,7 +127,7 @@ public function edit_asset_group_params(): array {
 	 */
 	public function get_asset_group_params(): array {
 		return [
-			'id' => [
+			'campaign_id' => [
```
puntope, line 130: 💅   I guess we should make this param required? 🤔

src/API/Site/Controllers/Ads/AssetGroupController.php
```diff
@@ -55,14 +55,20 @@ public function register_routes(): void {
...
 			[
 				[
 					'methods'             => TransportMethods::READABLE,
 					'callback'            => $this->get_asset_groups_by_campaign_id_callback(),
 					'permission_callback' => $this->get_permission_callback(),
 					'args'                => $this->get_asset_group_params(),
 				],
+				[
+					'methods'             => TransportMethods::CREATABLE,
+					'callback'            => $this->create_asset_group_callback(),
```
puntope, line 68: 💅 I personally will rename the previous `get_asset_groups_by_campaign_id_callback()` as just `get_asset_groups_callback` to maintain consistency and keep it shorter.

src/API/Site/Controllers/Ads/AssetGroupController.php
puntope: Otherwise I got a fatal when running the endpoint without `campaing_id` param.

```
 2023/01/16 09:57:51 [error] 30#30: *4801 FastCGI sent in stderr: 
"PHP message: PHP Fatal error:  Uncaught TypeError: Argument 1 passed to 

Automattic\WooCommerce\GoogleListingsAndAds\API\Google\AdsAssetGroup::get_asset_groups_by_campaign_id() must be of the type int, null given, 
called in /var/www/html/wp-content/plugins/google-listings-and-ads/src/API/Site/Controllers/Ads/AssetGroupController.php 
on line 153 and defined in /var/www/html/wp-content/plugins/google-listings-and-ads/src/API/Google/AdsAssetGroup.php:273
```

puntope (commented):
Howdy @jorgemd24 

I tested this PR, thanks for working on it.

- Behaviour looking good. I was able to fetch the asset group with the new edited endpoint.
- I was able to create an asset group

Screenshot 2023-01-16 at 17 14 35

- 💅  Left some comments. One blocker regarding the "required" for the `campaign_id` param.

jorgemd24:
Thanks, @puntope for having a look at it.

I have added your suggestions 👍   I guess this is ready for another round!

Thanks in advance!

puntope (approved):
✅  LGTM thanks for the adjustments @jorgemd24 

- GET asset groups by campaign ID ✅ 
- Show error when wrong campaign id format, no campaign_id, or not found campaign ✅ 
- POST create new asset group, by campaign id ✅ 
- Show error when wrong campaign id format, no campaign_id, or not found campaign ✅ 
- Checked the code ✅

PR #1854, opened 2023-01-16
PMax Assets: TextsEditor for managing the asset texts

eason9487:

This PR implements a part of 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) in #1787.

- Multiple texts editor
   - Set the min/max number of texts
   - Set the max of character counting for each input
   - Add/remove text input
   - Get updated texts via callback
- Characters length calculation
   - The results should be the same with the text input fields of assets management on Google Ads admin page.
   - 💡 Please note that currently there are no detailed rules for this section, and the ported code retains all unintended inconsistent results. (See its tests)

### Screenshots:

https://user-images.githubusercontent.com/17420811/212669335-378b8691-ba92-41f4-a357-0b674fbbbd0a.mp4

2. Use the TextsEditor demo to test.
3. Go to an assets management on Google Ads admin page
   1. Open Google Ads admin page
   1. Go to **Campaigns** list via the main menu on left side
   1. Click an active campaign to view the details
   1. View the **Asset groups** via the menu on left side
   1. Click **Edit assets** button to enter the assets management. The page should look like the following:
      image
4. Enter the same strings on both sides to see if the results of character counting are the same.

### Additional details:

The logic of character counting was ported from the actual implementation found on Google Ads admin page. As a record, the code can be partially found by the following steps.

1. Go to an assets management on Google Ads admin page.
2. Find a JS file ends with `/aw_cm/editing/main.dart.js`.
3. Format the minified JS.
4. The first part is the main logic and it can be found by searching the keyword `65023`.
5. The second part is listing the zero-width char codes and it can be found by searching the keyword `2304, null`.

#### The first part:
```js
{
    aUt(a) {
      var s, r
      if (a == null)
        return 0
      for (s = new A.aLc(a),
      s = new A.eK(s,s.gav(s),t.iY.p("eK<br.E>")),
      r = 0; s.am(); )
        r += A.DIV(s.d)
      return r
    },
    DIV(a) {
      var s
      if (a >= 2304 && a <= 3455)
        return J.ii(B.mgJ.a, a) ? 0 : 1
      if (!(a <= 1273))
        if (a !== 1470)
          if (!(a >= 1488 && a <= 1514))
            if (a !== 1523)
              if (a !== 1524)
                if (!(a >= 1536 && a <= 1791))
                  if (!(a >= 1872 && a <= 1919))
                    if (!(a >= 64336 && a <= 65023))
                      if (!(a >= 65136 && a <= 65279))
                        if (!(a >= 7680 && a <= 8383))
                          if (!(a >= 8448 && a <= 8506))
                            if (!(a >= 3584 && a <= 3711))
                              s = a >= 65377 && a <= 65500
                            else
                              s = !0
                          else
                            s = !0
                        else
                          s = !0
                      else
                        s = !0
                    else
                      s = !0
                  else
                    s = !0
                else
                  s = !0
              else
                s = !0
            else
              s = !0
          else
            s = !0
        else
          s = !0
      else
        s = !0
      if (s)
        return 1
      return 2
    },
}
```

#### The second part:

```js
B.kl3 = new A.bk([2304, null, 2305, null, 2306, null, 2362, null, 2364, null, 2369, null, 2370, null, 2371, null, 2372, null, 2373, null, 2374, null, 2375, null, 2376, null, 2381, null, 2385, null, 2386, null, 2387, null, 2388, null, 2389, null, 2390, null, 2391, null, 2402, null, 2403, null],A.c("bk<x*,an>"))
```

### Changelog entry

>

### DIFF
--- a/js/src/components/paid-ads/asset-group/asset-group-section.js
+++ b/js/src/components/paid-ads/asset-group/asset-group-section.js
@@ -12,6 +12,7 @@ import AppDocumentationLink from '.~/components/app-documentation-link';
 import VerticalGapLayout from '.~/components/vertical-gap-layout';
 import FinalUrlCard from './final-url-card';
 import ImagesSelector from './images-selector';
+import TextsEditor from './texts-editor';
 import AssetField from './asset-field';
 import './asset-group-section.scss';
 
@@ -58,7 +59,7 @@ export default function AssetGroupSection() {
 			<VerticalGapLayout size="medium">
 				<FinalUrlCard onAssetsChange={ setAssetGroup } />
 				<div>Assets card will be added here</div>
-				<h3>Temporary demo for AssetField:</h3>
+				<h3>Temporary demo for AssetField and TextsEditor:</h3>
 				<div>
 					<AssetField
 						heading={ __(
@@ -135,7 +136,70 @@ export default function AssetGroupSection() {
 						numOfIssues={ 0 }
 						initialExpanded
 					>
-						Initially expanded asset field
+						<TextsEditor
+							minNumberOfTexts={ 3 }
+							maxNumberOfTexts={ 5 }
+							maxCharacterCounts={ [ 15, 30, 30, 30, 30 ] }
+							placeholder={ __(
+								'Headline',
+								'google-listings-and-ads'
+							) }
+							addButtonText={ __(
+								'Add headline',
+								'google-listings-and-ads'
+							) }
+							initialTexts={ [ 'Hello', 'World' ] }
+							onChange={ ( v ) =>
+								// eslint-disable-next-line no-console
+								console.log( 'Headline:onChange', v )
+							}
+						/>
+					</AssetField>
+					<AssetField
+						heading={ __(
+							'Long headlines',
+							'google-listings-and-ads'
+						) }
+						subheading={ __(
+							'At least 1 required. Add up to 5.',
+							'google-listings-and-ads'
+						) }
+						help={
+							<>
+								<p>
+									{ __(
+										'The long headline is the first line of your ad, and appears instead of your short headline in larger ads. Long headlines can be up to 90 characters, and may appear with or without your description.',
+										'google-listings-and-ads'
+									) }
+								</p>
+								<p>
+									{ __(
+										'The length of the rendered headline will depend on the site it appears on. If shortened, it will end with an ellipsis(…).',
+										'google-listings-and-ads'
+									) }
+								</p>
+							</>
+						}
+						numOfIssues={ 0 }
+					>
+						<TextsEditor
+							minNumberOfTexts={ 1 }
+							maxNumberOfTexts={ 5 }
+							maxCharacterCounts={ 90 }
+							placeholder={ __(
+								'Long headlines',
+								'google-listings-and-ads'
+							) }
+							addButtonText={ __(
+								'Add long headline',
+								'google-listings-and-ads'
+							) }
+							initialTexts={ [] }
+							onChange={ ( v ) =>
+								// eslint-disable-next-line no-console
+								console.log( 'Long headlines:onChange', v )
+							}
+						/>
 					</AssetField>
 				</div>
 			</VerticalGapLayout>
--- a/js/src/components/paid-ads/asset-group/texts-editor.js
+++ b/js/src/components/paid-ads/asset-group/texts-editor.js
@@ -2,7 +2,8 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState } from '@wordpress/element';
+import { noop } from 'lodash';
+import { useState, useEffect, useRef } from '@wordpress/element';
 import GridiconCrossSmall from 'gridicons/dist/cross-small';
 
 /**
@@ -12,6 +13,16 @@ import AppButton from '.~/components/app-button';
 import AppInputControl from '.~/components/app-input-control';
 import AddAssetItemButton from './add-asset-item-button';
 import './texts-editor.scss';
+
+function normalizeNumberOfTexts( texts, minNumberOfTexts, maxNumberOfTexts ) {
+	const shortage = Math.max( minNumberOfTexts - texts.length, 0 );
+	const supplement = Array.from( { length: shortage }, () => '' );
+	const sliceArgs = [ 0 ];
+	if ( maxNumberOfTexts > 0 ) {
+		sliceArgs.push( maxNumberOfTexts );
+	}
+	return texts.concat( supplement ).slice( ...sliceArgs );
+}
 
 /**
  * Renders a list of text inputs for managing the single type of asset texts.
@@ -23,7 +34,7 @@ import './texts-editor.scss';
  * @param {number|number[]} props.maxCharacterCounts Maximum number of characters for each text. If the limits are the same, a single number can be used instead of an array.
  * @param {string} props.addButtonText Text for the button to add a new text input.
  * @param {string} [props.placeholder] Placeholder text.
- * @param {(texts: Array<string>) => void} props.onChange Callback function to be called when the texts are changed.
+ * @param {(texts: Array<string>) => void} [props.onChange] Callback function to be called when the texts are changed.
  */
 export default function TextsEditor( {
 	initialTexts = [],
@@ -32,22 +43,41 @@ export default function TextsEditor( {
 	maxCharacterCounts,
 	addButtonText,
 	placeholder,
-	onChange,
+	onChange = noop,
 } ) {
+	const updateTextsRef = useRef();
 	const [ texts, setTexts ] = useState( () => {
-		const shortage = minNumberOfTexts - initialTexts.length;
-		if ( shortage <= 0 ) {
-			return initialTexts;
+		const nextTexts = normalizeNumberOfTexts(
+			initialTexts,
+			minNumberOfTexts,
+			maxNumberOfTexts
+		);
+		if ( nextTexts.length !== initialTexts.length ) {
+			onChange( nextTexts );
 		}
-
-		const supplement = Array.from( { length: shortage }, () => '' );
-		return initialTexts.concat( supplement );
+		return nextTexts;
 	} );
 
 	const updateTexts = ( nextTexts ) => {
 		setTexts( nextTexts );
 		onChange( nextTexts );
 	};
+	updateTextsRef.current = updateTexts;
+
+	useEffect( () => {
+		if (
+			( maxNumberOfTexts > 0 && texts.length > maxNumberOfTexts ) ||
+			( minNumberOfTexts > 0 && texts.length < minNumberOfTexts )
+		) {
+			updateTextsRef.current(
+				normalizeNumberOfTexts(
+					texts,
+					minNumberOfTexts,
+					maxNumberOfTexts
+				)
+			);
+		}
+	}, [ texts, maxNumberOfTexts, minNumberOfTexts ] );
 
 	const handleChange = ( text, { event } ) => {
 		const { index } = event.target.dataset;
--- a/js/src/components/paid-ads/asset-group/texts-editor.js
+++ b/js/src/components/paid-ads/asset-group/texts-editor.js
@@ -46,17 +46,7 @@ export default function TextsEditor( {
 	onChange = noop,
 } ) {
 	const updateTextsRef = useRef();
-	const [ texts, setTexts ] = useState( () => {
-		const nextTexts = normalizeNumberOfTexts(
-			initialTexts,
-			minNumberOfTexts,
-			maxNumberOfTexts
-		);
-		if ( nextTexts.length !== initialTexts.length ) {
-			onChange( nextTexts );
-		}
-		return nextTexts;
-	} );
+	const [ texts, setTexts ] = useState( initialTexts );
 
 	const updateTexts = ( nextTexts ) => {
 		setTexts( nextTexts );
--- a/js/src/components/paid-ads/asset-group/texts-editor.js
+++ b/js/src/components/paid-ads/asset-group/texts-editor.js
@@ -23,7 +23,7 @@ import './texts-editor.scss';
  * @param {number|number[]} props.maxCharacterCounts Maximum number of characters for each text. If the limits are the same, a single number can be used instead of an array.
  * @param {string} props.addButtonText Text for the button to add a new text input.
  * @param {string} [props.placeholder] Placeholder text.
- * @param {Function} props.onChange Callback function to be called when the texts are changed.
+ * @param {(texts: Array<string>) => void} props.onChange Callback function to be called when the texts are changed.
  */
 export default function TextsEditor( {
 	initialTexts = [],
--- a/js/src/components/paid-ads/asset-group/texts-editor.js
+++ b/js/src/components/paid-ads/asset-group/texts-editor.js
@@ -1,0 +1,123 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useState } from '@wordpress/element';
+import GridiconCrossSmall from 'gridicons/dist/cross-small';
+
+/**
+ * Internal dependencies
+ */
+import AppButton from '.~/components/app-button';
+import AppInputControl from '.~/components/app-input-control';
+import AddAssetItemButton from './add-asset-item-button';
+import './texts-editor.scss';
+
+/**
+ * Renders a list of text inputs for managing the single type of asset texts.
+ *
+ * @param {Object} props React props.
+ * @param {string[]} [props.initialTexts=[]] Initial texts.
+ * @param {number} [props.minNumberOfTexts=0] Minimum number of texts.
+ * @param {number} [props.maxNumberOfTexts=0] Maximum number of texts.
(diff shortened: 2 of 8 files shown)


puntope:
❓ Not sure if this should be handled in this PR or not. Having an empty input is no counting as item in Google admin, but in this PR its counting as valid item. In other words, in this PR, if you need 3 headlines, you can have 2 headlines with 10 characters and one with 0 characters and seems it take it as valid. 

**In Google Admin** 

- Go to headlines
- Fill 2 of them
- Have one empty
- It shows an error "At least 3 headlines are required"

Screenshot 2023-01-19 at 09 58 15

**In this PR**

- Go to headlines
- Fill 2 of them
- Have one empty
- It doesn't shows an error and Launch campaign button is enabled. 

Screenshot 2023-01-19 at 10 01 44

js/src/components/paid-ads/asset-group/texts-editor.js
```diff
@@ -0,0 +1,123 @@
...
+
+					return (
+						<div
+							key={ index }
+							className="gla-texts-editor__text-item"
+						>
+							<AppInputControl
+								className="gla-texts-editor__text-input"
+								value={ text }
+								kindCharacterCount="google-ads"
```
puntope, line 108: ❓ I see this is always going to be `google-ads`. Is this going to be extended in the future to other kinds of characters sets?

If not I would just use a boolean. Or even better just relying on `maxCharacterCount` prop for deciding it AppInputControl should care about counting characters.

js/src/components/paid-ads/asset-group/texts-editor.js
```diff
@@ -0,0 +1,123 @@
...
+								) }
+							</div>
+						</div>
+					);
+				} ) }
+			</div>
+			<AddAssetItemButton
+				aria-label={ __( 'Add text', 'google-listings-and-ads' ) }
+				disabled={
+					maxNumberOfTexts !== 0 && texts.length >= maxNumberOfTexts
```
puntope: ❓ Not a big deal tho, since it is likely not happening. But setting negative number here is allowed. Whereas in `js/src/components/app-input-control/index.js`  we check `maxCharacterCount > 0`

js/src/components/paid-ads/asset-group/asset-group-section.js
```diff
@@ -135,7 +136,70 @@ export default function AssetGroupSection() {
...
+									) }
+								</p>
+							</>
+						}
+						numOfIssues={ 0 }
+					>
+						<TextsEditor
+							minNumberOfTexts={ 1 }
+							maxNumberOfTexts={ 5 }
+							maxCharacterCounts={ 90 }
```
puntope, line 188: ❓  What is the rational behind supporting single number as a type for `maxCharacterCounts`?  I don't have by hand the API response format. But I think it would be more consistent to stick into array format as we do in `initialTexts`

js/src/components/paid-ads/asset-group/texts-editor.js
```diff
@@ -0,0 +1,123 @@
...
+	maxCharacterCounts,
+	addButtonText,
+	placeholder,
+	onChange,
+} ) {
+	const [ texts, setTexts ] = useState( () => {
+		const shortage = minNumberOfTexts - initialTexts.length;
+		if ( shortage <= 0 ) {
+			return initialTexts;
+		}
```
puntope: 💅  This could be omitted since `Array.from( { length: shortage }, () => '' );` with  length <= 0 will return an empty array.

js/src/components/paid-ads/asset-group/texts-editor.js
```diff
@@ -0,0 +1,123 @@
...
+	initialTexts = [],
+	minNumberOfTexts = 0,
+	maxNumberOfTexts = 0,
+	maxCharacterCounts,
+	addButtonText,
+	placeholder,
+	onChange,
+} ) {
+	const [ texts, setTexts ] = useState( () => {
+		const shortage = minNumberOfTexts - initialTexts.length;
```
puntope: ⚠️   We are not handling here the `maxNumberOfTexts` constraint. 

Hence if we add a component like this;:

```jsx
<TextsEditor
							minNumberOfTexts={ 3 }
							maxNumberOfTexts={ 3 }
							maxCharacterCounts={ [ 15, 30, 30, 30, 30 ] }
							placeholder={ __(
								'Headline',
								'google-listings-and-ads'
							) }
							addButtonText={ __(
								'Add headline',
								'google-listings-and-ads'
							) }
							initialTexts={ [ 'Hello', 'World', 'kkk', 'ppp' ] }
							onChange={ ( v ) =>
								// eslint-disable-next-line no-console
								console.log( 'Headline:onChange', v )
							}
						/>
```

It will render the 5 values even tho the maximum is 3

Screenshot 2023-01-19 at 10 51 14

js/src/components/paid-ads/asset-group/texts-editor.js
```diff
@@ -0,0 +1,123 @@
...
+		if ( shortage <= 0 ) {
+			return initialTexts;
+		}
+
+		const supplement = Array.from( { length: shortage }, () => '' );
```
puntope: 💅  Not sure about this `supplement` name

js/src/components/paid-ads/asset-group/texts-editor.js
```diff
@@ -0,0 +1,123 @@
...
+		const nextTexts = [ ...texts ];
+		nextTexts.splice( index, 1 );
+		updateTexts( nextTexts );
+	};
+
+	const handleAddClick = () => {
+		updateTexts( texts.concat( '' ) );
+	};
+
+	const normalizedMaxCharacterCounts = [ maxCharacterCounts ].flat();
```
puntope, line 90: ❓  Why is this for? 🤔

puntope (commented):
Howdy @eason9487 

Thanks for the work here. Overall looking great.  

Left some 💅 comments as well as some questions and potential issues

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: The need to specify `kindCharacterCount` was an intentional dev design. As the character count implementation of Google Ads doesn't use a common rule and the `AppInputControl` is a basic component, setting Google Ads rule as the default counter would not be ideal.

Therefore, even if no other character count rules will be added at present, it would be better to avoid misapplying Google Ads rule than to make them the default.

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/1854/commits/1cb97201b4a9e94e3650634c95ecd06de671ae67.

js/src/components/paid-ads/asset-group/asset-group-section.js
eason9487: Based on [Google's asset specs](https://developers.google.com/google-ads/api/docs/performance-max/assets#ensure_minimum_asset_requirements_are_met), the headlines and descriptions each have two different character limits, which are also reflected in the UI design. The single number is used for asset texts with the same character limit.

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: Removed along with the change in https://github.com/woocommerce/google-listings-and-ads/pull/1854/commits/0fe5fc9bc716542590ac57d3f06c7c9dce6b9470.

The reason was that calling `Array.from` with a negative `length` looks semantically buggy. Therefore, I added a `Math.max()`.

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: Fixed in https://github.com/woocommerce/google-listings-and-ads/commit/0fe5fc9bc716542590ac57d3f06c7c9dce6b9470.

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: It's for supplementing the shortage parts of `initialTexts` when the length of `initialTexts` is less than `minNumberOfTexts`. Do you have another semantically more precise name?

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: Normalize `maxCharacterCounts` to an array as it could be a number or an array. Its purpose is equivalent to

```js
const normalizedMaxCharacterCounts = Array.isArray( maxCharacterCounts )
	? maxCharacterCounts
	: [ maxCharacterCounts ];

```

eason9487:
Hi @puntope, thanks for the review. I believe all comments in the first code review have been handled or replied. Could you take another look?

---

> ❓ Not sure if this should be handled in this PR or not. Having an empty input is no counting as item in Google admin, but in this PR its counting as valid item. In other words, in this PR, if you need 3 headlines, you can have 2 headlines with 10 characters and one with 0 characters and seems it take it as valid. [...]

These belong to the form level's validation and error messages and will be added in a subsequence PR.

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: Pushed another related adjustment https://github.com/woocommerce/google-listings-and-ads/pull/1854/commits/ba8b8910905d30494020fda6028732c2868313a7 to avoid a warning:
image

js/src/components/paid-ads/asset-group/texts-editor.js
puntope: I checked again but I still not get it. 

In AppInputControl we check 2 conditions to perform the new counter validation:

1- `kindCharacterCount` should be defined
2- `maxCharacterCount` > 0

So or we remove condition 1 or we could just omit `kindCharacterCount` prop and use always `google` one when condition 2 is satisfied.

js/src/components/paid-ads/asset-group/asset-group-section.js
puntope: So does that mean using `maxCharacterCounts={ 9 } `is a way to say "ok, apply 9 to all the values" whereas  `maxCharacterCounts={ [9,8,7] }` applies the values on input differently by index.  

I see `maxCharacterCounts={ [9] } `do exactly the same as `maxCharacterCounts={ 9 }`. So I wonder if we can reduce complexity (only one type + removing the Array.flat() call) by using only array format.

js/src/components/paid-ads/asset-group/texts-editor.js
puntope: > Do you have another semantically more precise name?

I would say inline shortage in the length prop and `itemsToAdd`  instead of `supplementing`.

But is okey if you keep like i's. Just an opinion.

puntope:
Howdy again @eason9487 

Behaviour looking good now. Thanks. I have still some questions. Mainly about the need of the complexity related with the dual type for `maxCharacterCounts` as well as the need of the kind of character count discussion. If this is rushy to be merged let me know, we can do so and discuss further this in other PRs

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: Suppose now we are going to add a feature to `AppInputControl` so that it will show a hashed input value somewhere after a specific event like`onBlur` or `onChange`. The new props could be:

1. `<AppInputControl showHashAfter="change" />`
   - Set the default prop value for `hash` prop like `{ hash = 'md5' }`
   - Once set the `showHashAfter` prop, it enables the feature 
2. `<AppInputControl showHashAfter="change" hash="md5" />`
   - It always needs to specify the `showHashAfter` prop and a hash algorithm if wants to use the feature

The first way seems to satisfy and will work well when only md5 is actually used. But considering that there are dozens of hash algorithms, probably none of them could be considered by most to be used as the default one. Therefore, even though the first way works correctly for now, it's not an ideal way in terms of technical design.

The situation here is similar in that there is no one character count method that could be taken as the default one, especially since we currently only have a ported implementation from Google Ads, and it's not a common one either. Furthermore, `AppInputControl` isn't a dedicated component for asset form but a general component for many uses, otherwise, a component like `TextsEditor` is dedicated so that the `kindCharacterCount` was specified directly.

js/src/components/paid-ads/asset-group/asset-group-section.js
eason9487: > So does that mean using `maxCharacterCounts={ 9 } `is a way to say "ok, apply 9 to all the values" whereas `maxCharacterCounts={ [9,8,7] }` applies the values on input differently by index.

Yes.

https://github.com/woocommerce/google-listings-and-ads/blob/ba8b8910905d30494020fda6028732c2868313a7/js/src/components/paid-ads/asset-group/texts-editor.js#L34

> I see `maxCharacterCounts={ [9] } `do exactly the same as `maxCharacterCounts={ 9 }`. So I wonder if we can reduce complexity (only one type + removing the Array.flat() call) by using only array format.

It just happened to be possible to specify all the values by `maxCharacterCounts={ [ 9 ] }` as well but it will rely on the internal implementation details rather than the component's interfaces. The interface doesn't guarantee this kind of usage.

Would it work for you if change the internal `maxCharacterCount` from

```js
const maxCharacterCount =
	normalizedMaxCharacterCounts[ index ] ??
	normalizedMaxCharacterCounts[ 0 ];
```
to

```js
const maxCharacterCount =
	maxCharacterCounts?.[ index ] ?? maxCharacterCounts;

// and remove the line of const normalizedMaxCharacterCounts = [ maxCharacterCounts ].flat();
```

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: Thanks for the suggestion. I will keep the current one as I don't see the confusion about the name `supplement`.

eason9487:
Hi @puntope, thanks for the second round. I replied to all the new comments in the second round. Could you help with a new round?

---

> I have still some questions. Mainly about the need of the complexity related with the dual type for `maxCharacterCounts` as well as the need of the kind of character count discussion.

An array that accepts a single value or a value of the same type should be a very common design, and its support requires only 1 to 3 lines of code for implementation. Therefore, I don't think it adds complexity but rather the usability of the components.

Please refer to https://github.com/woocommerce/google-listings-and-ads/pull/1854#discussion_r1082287076 for the need for specifying the kind of character count.

> If this is rushy to be merged let me know, we can do so and discuss further this in other PRs

Thanks for your consideration. I would say all PRs related to this project were urgent in terms of time since the beginning, but it's fine if anything is a blocker from your point of view.

js/src/components/paid-ads/asset-group/texts-editor.js
puntope: I agree that if we are going to have multiple implementations it would be necessary to add that "kind of characters" property. But since it's not even planned, I don't think it's necessary to add it.

js/src/components/paid-ads/asset-group/asset-group-section.js
puntope: For me doesn't matter the 3 solutions do same. And they are necessary only because we allowed two types of data in that prop. I still don't see the advantage of allowing a number as acceptable prop type.

puntope (approved):
✅  Approved

Still some disagreements about two points in this PR but due deadlines and since they don't have direct impact neither in performance nor behaviour this Is good to go.

js/src/components/paid-ads/asset-group/texts-editor.js
eason9487: As mentioned earlier, this is an intentional precaution to avoid misuse. Just because the code can work well in the runtime or is not used, this does not necessarily mean that precaution is always unnecessary. It is a common practice to add precautions to error-prone places unless it's considered overengineering handling.

For example, a `getPageQuery` function could look like this:

```js
function getPageQuery( page ) {
  return {
    page,
    per_page: 10,
    order_by: 'total_sales',
    order: 'desc',
  };
}
```

It certainly works well as all previous page buttons are always disabled or hidden when viewing the first page.

However, `getPageQuery` is a common util function, and even though it's currently called without going below the minimum page number, it's still good practice to add some precautions like this:

```js
function getPageQuery( page ) {
  return {
    // Avoid `page` out of range
    page: Math.max( page, 1),
    per_page: 10,
    order_by: 'total_sales',
    order: 'desc',
  };
}
```

Another benefit is that it also shows that the minimum number of pages starts from 1.

To summarize, `'google-ads'` was intentionally not used as the default value for `kindCharacterCount` here for the following reasons.
1. It's not a common method.
2. Even among the common methods, none of them is widely considered to be the default method.
3. It's a precaution to avoid misuse.
4. Whether to set the default value for `kindCharacterCount` or not, the difference is likely to be only a few lines of code and not complicated, so it probably is not overengineering.

js/src/components/paid-ads/asset-group/asset-group-section.js
eason9487: @puntope, thanks for your thought. To me, there are no obvious pros or cons as it's a trade-off between ease of use and complexity of the prop. Therefore, if this is a code smell or blocker for you, I'm happy to open another PR to change it to a single type of prop. Let me know if it's needed.

PR #1859, opened 2023-01-18
Fix namespace for additional vendor packages

mikkamp:

There have been additional reports of our vendor packages conflicting with other extensions. In this case `woocommerce-bookings` which both share usage of the `google/apiclient` package. 

Following is the list of challenges I came across, which resulted in having to rewrite a substantial amount of the `prefix-vendor-namespace.php` script.

#### Unique namespace match
The `apiclient` package uses the namespace `Google` as well as additional subname spaces. In itself the Google namespace is not unique enough to search for replacements. Which means that our current solution of replacing a namespace prefix was not going to work. To work around this I introduced the concept of matching "strict" or not. When strict is enabled it will match the full namespace, when it's disabled it will match any namespace within the package that starts with the package namespace.

For example:
Not strict and package name `guzzleclient` will match all namespaces like `GuzzleHttp\Psr7`, `GuzzleHttp\Guzzle`, etc...
Whereas strict and package name `apiclient` will only match namespaces that are defined as `Google`.

#### Additional namespaces
As mentioned in the previous point the `apiclient` package also uses additional subname spaces, so strict matching doesn't fully work. To get around this I introduced a list of `extra_namespaces` which can be configured per package which will also be replaced. This list might be a bit of work to maintain as we update packages but it should be relatively easy to add any additional ones.

#### Exclude file autoloads
In the previous version of the script we required any `autoload > files` configuration to be added in the main `composer.json` file. The reason being because the hash of the composer file wouldn't be unique enough if both of the plugins had the extensions in the vendor folder. However in the case of the `apiclient` package it loads an aliases.php file. Which is used for backward compatibility before namespaces were used. Since we don't need this backward compatibility (and we don't fully prefix everything in this fine), the preferable solution is to prevent this file from being autoloaded. I added the `exclude_autoload` parameter which can be configured per package and prevents these files from being required in the main `composer.json` file.

#### String replacements in JSON files
To replace namespaces in JSON files we were previously using a regular `str_replace` however this doesn't work very well to check if a namespace starts with a specific name. For example `GPBMetadata\\Google\\Protobuf\\` was getting matched because we prefix the namespace `Google`. To get around this I merged the prefixing code with the one for removing file autoloads. There we were parsing the JSON and going through it per package, which makes it easier to do some stricter matches (depending on strict mode).

#### Replacements in package composer.json
Previously we were also replacing the namespaces within the packages composer.json files (which wasn't accurate, see previous point). Since the script is called directly from composer the package map has already been created, so we only need to update the package map. So at the moment I left out the code which replaces the local composer.json files. If there is some other case where it's needed we can restore it but it needs to be improved by properly parsing the JSON files.

#### Counting uses and namespace replacements
Recently some additional debugging code was added which would count namespace and uses replacements. And would output a list of files where uses were replaced without a namespace replacement. Since the way we process replacements was changed, this list was no longer accurate since we could run replacements in packages multiple times. So I opted to remove this code.

Addresses part of #1760

Luckily we have unit tests in place which cover pretty much all internal usages of these Google Services. So I relied largely on the unit tests to surface namespaces which had to be renamed. In addition to the unit tests I ran several of the smoke tests included in the WIKI to make sure API calls still functioned with the renamed packages.

To confirm that the plugin no longer conflicts with `woocommerce-bookings` the following test can be run:
1. Checkout this PR and run `rm -rf vendor && composer install`
1. Deactivate the `google-listings-and-ads` plugin
2. Install and activate [`woocommerce-bookings`](https://github.com/woocommerce/woocommerce-bookings/releases)
3. Activate the plugin through CLI `wp --allow-root plugin activate google-listings-and-ads`

> **Note**
Activating through the WP dashboard does a redirect so it doesn't encounter the situation where `woocommerce-bookings` code is active and then `google-listings-and-ads` is loaded later.
The problem is only reproducible when plugins aren't loaded alphabetically which is why we have to reproduce it with the WP CLI command.

### Changelog entry
* Fix - Prefix Google Service packages to prevent plugin conflicts.


bin/prefix-vendor-namespace.php
```diff
@@ -45,92 +96,165 @@
...
-		if ( false !== strpos( $contents, "{$new_namespace}\\{$namespace}" ) ) {
-			continue 2;
-		}
 
-		$namespace_change = 0;
-		$uses_change      = 0;
+/**
+ * Process prefixing in a specific file.
+ *
+ * @since x.x.x
```
puntope, line 122: 💅  Not sure about adding here `@since` tag. Since it's just a modification and the function is not new.

bin/prefix-vendor-namespace.php
```diff
@@ -45,92 +96,165 @@
...
+ *
+ * @param string $file             File name.
+ * @param string $package          Package prefix configuration.
+ * @param bool   $prefix_namespace Whether to prefix namespaces or not.
+ * @param bool   $prefix_uses      Whether to prefix uses of a namespace.
+ */
+function process_file( $file, $package, $prefix_namespace = true, $prefix_uses = false ) {
```
puntope, line 129: seems for me that $prefix_namespace should be an array. But its type is defined as Srtring.

bin/prefix-vendor-namespace.php
```diff
@@ -45,92 +96,165 @@
...
+	global $direct_replacements, $namespace_prefix;
```
puntope: `$namespace_prefix` is not being used.

bin/prefix-vendor-namespace.php
```diff
@@ -45,92 +96,165 @@
...
-		"{$vendor_dir}/composer/installed.json",
-		$composer_json['autoload']['files'] ?? [],
-		$path
+/**
+ * Prefix any import statements and direct uses of namespace.
+ *
+ * @since x.x.x
+ *
+ * @param string $contents File contents.
+ * @param string $package  Package prefix configuration.
```
puntope: Same as in https://github.com/woocommerce/google-listings-and-ads/pull/1859/files#r1085050870

`$package` seems for me an array

bin/prefix-vendor-namespace.php
```diff
@@ -45,92 +96,165 @@
...
+ */
+function prefix_uses( &$contents, $package ) {
```
puntope, line 203: Same as in https://github.com/woocommerce/google-listings-and-ads/pull/1859/files#r1085080705

puntope:
One thing I saw is now the PHPDOC directives like `@see` and `@throw` lost the reference. 

Screenshot 2023-01-24 at 18 34 05

Screenshot 2023-01-24 at 18 32 19


I think those better to change them manually.

puntope:
Also, one file lost the reference in the implements. I guess because is not being used.

Screenshot 2023-01-24 at 18 35 16

puntope (commented):
Howdy @mikkamp I've finished finally the review of this PR. Just some minor comments aded.

bin/prefix-vendor-namespace.php
mikkamp: `$prefix_namespace` should still be a bool, but you are right that the second parameter `$package` got refactored from `string > array`

mikkamp:
Thanks for the review @puntope I've gone ahead and fixed the `$package` changing to an array. Nice catch on finding the missing replace, as you mentioned we no longer use `aliases.php` so it's not a problem, but it does point out that the regexp we use is strictly for a class, I've modified it to work for both a class and an interface.

Regarding the `@throws`, they are all using imported names within our extension, so it's only the docblocks within the vendor folder. In my opinion those are fine to leave as there is plenty of documentation available to explain what's expected (we also haven't replaced these in the past for other packages).
As for the `@see` within our extension, for me it doesn't add any extra benefit to prefix them, I personally would be reviewing the parameters in the public documentation here: https://developers.google.com/shopping-content/reference/rest/v2.1/products#Product
Although if you think it's going to be helpful I don't mind going ahead and changing the references within our extension (not in the vendor folder).

It's ready for another round of review.

bin/prefix-vendor-namespace.php
puntope: Yes. sorry. I messes up the names 😂

puntope:
Thanks @mikkamp  Looking good.

Just one more thing. We added extends/interface support. But not implements. I guess we don't hace any case for this tho. But wonder if in the future we will. 

Approved this in advance

puntope (approved):

mikkamp:
Thanks for the thorough review.

To add `implements` becomes a little more complicated because you can implement multiple classes. For example you can have the following code:
```php
class Two extends \Namespace\One implements \Namespace\Usable, \Namespace\Updatable
```

The regex currently isn't able to replace multiple namespaces in a line. However as you mentioned this hasn't been needed yet, because if each name has an import line at the top of the file then we don't need to handle the replacement in the class declaration. So far we only need it for the `apiclient-services` code like is [done here](https://github.com/googleapis/google-api-php-client-services/blob/v0.284.0/src/ShoppingContent.php#L35). They have `use Google\Client` at the top of the file, if they would have just added `use Google\Service` then we wouldn't have needed this extra replacement. So I think the better approach would be to add additional replacements if needed, but not try to handle patterns that aren't there yet.

PR #1860, opened 2023-01-18
PMax - Suggest HTML inserted images 

jorgemd24:

This PR adds the functionality to suggest image assets that are inserted as HTML tag in the post content. Until now the images that were suggested were attached to the post however it is also possible to add images to a specific post without attaching the image.

The image is inserted as HTML when we select an image that has been previously uploaded to the media library.

image

Also, this PR is limiting the type of attachments to `image/jpeg, image/png, image/jpg, as other types of images are not allowed. See: https://support.google.com/google-ads/answer/10724492?hl=en#zippy=%2Cimage-specifications

1. Choose a page, post or product. Get the ID.
2. Add an image that is already in the media library.
3. Make a request to `GET gla/assets/suggestions?id=POST_ID&type=post`
4. Check that the image is showing in the response. 

### Additional details:

If the image size is less than the minimum requirement, it will not be suggested. For a square marketing image, the minimum is `300x300px` and for marketing_image is `600x314px.`

### DIFF
--- a/src/Ads/AssetSuggestionsService.php
+++ b/src/Ads/AssetSuggestionsService.php
@@ -312,7 +312,7 @@ class AssetSuggestionsService implements Service {
 		$dom->loadHTML( $html );
 		$images     = $dom->getElementsByTagName( 'img' );
 		$images_ids = [];
-		$pattern    = '/-\d+x\d+\.(jpg|jpeg|gif|png|svg|tiff|webp|heif|heic)$/i';
+		$pattern    = '/-\d+x\d+\.(jpg|jpeg|png)$/i';
 		foreach ( $images as $image ) {
 			$url_unscaled = preg_replace(
 				$pattern,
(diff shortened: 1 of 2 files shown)


puntope:
❓ I take as granted this PR is not covering HTML images inserted in the terms? (i.e. Categories)

puntope (approved):
✅ LGTM 

I tested locally:

- Adding JPG image from media library and valid dimensions. 
- Adding PNG image from media library and valid dimensions. 
- Adding PNG image from media library with wrong dimensions.
- Upload an image PNG. It's appearing in suggestion too. 
- Upload an image JPEG. It's appearing in suggestion too. 
- Only one that didn't work (obviously) was entering the image by URL. 

My only question is about the need of supporting this in HTML inside categories terms. But I think we don't as discussed in previous PRs. So approved in advance.

jorgemd24:
Thanks @puntope for your time looking at this.

> ❓ I take as granted this PR is not covering HTML images inserted in the terms. (i.e. Categories)

Yes, that is correct, only the posts.

PR #1862, opened 2023-01-20
PMax Assets: Preprocessing for the assets form

eason9487:

This PR implements the preprocessing for the "Assets form" part of 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) in #1787.

In order to avoid the subsequent PR of the "Assets form" being too large, some are split into preprocessing:

- Move and rename *js/src/utils/paid-ads/validateForm* to *js/src/components/paid-ads/validateCampaign*
   - and a later PR will add the validateAssetGroup in the same directory
- Extract the campaign creation and editing forms out as `CampaignAssetsForm`
   - for adding more shared functions in a later PR
- Add `children` prop to `TextsEditor`
   - for showing error messages in a later PR
- Add `children` and `onChange` props to `ImagesSelector`
   - for showing error messages and getting updated image URLs in a later PR
- Add tests for `ImagesSelector`
- Add tests for the functions in `useCroppedImageSelector`

1. Go to the campaign step of the campaign creation page
2. Check if the "Continue" or "Complete setup" button is disabled when the budget is <= 0
3. Check if the "Continue" or "Complete setup" button is disabled when the ads audience is empty
4. Repeat test step 2 for the campaign editing page
5. Repeat test steps 2 and 3 for the "Create your paid campaign" step of the ads setup flow (`/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads`)
6. Repeat 1~3 steps for the "Complete your campaign" step of the onboarding flow (`/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-mc`)


js/src/components/paid-ads/campaign-assets-form.js
```diff
@@ -0,0 +1,29 @@
...
+import validateCampaign from '.~/components/paid-ads/validateCampaign';
+
+/**
+ * @typedef {import('.~/components/types.js').CampaignFormValues} CampaignFormValues
+ */
+
+/**
+ * Renders a form based on AdaptiveForm for managing campaign and assets.
+ *
+ * @augments AdaptiveForm
```
puntope, line 14: 💅  Not blocker. Even being a sinonym, I would use a more widely used keyword for this `@extends`

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.test.js
```diff
@@ -0,0 +1,73 @@
+/**
```
puntope, line 1: 💅  Thanks for adding tests. It would be nice to cover also the `useCroppedImageSelector` itself and not only the aux functions. But not a blocker and could be done in the future.

puntope (approved):
✅ LGTM

I tested locally as well as inspected the code quality. All good for me... justa minor 💅 
 comments and suggestion about more tests.

js/src/components/paid-ads/campaign-assets-form.js
eason9487: The reason for using `@augments` instead of `@extends` was to convey that the props of `CampaignAssetsForm` are inherited from `AdaptiveForm` and that they are not language-level inheritance. Considering `@augments` isn't quite the right tag, it might be better to remove it and add a code comment instead.

js/src/hooks/useCroppedImageSelector/useCroppedImageSelector.test.js
eason9487: @puntope, thanks for the suggestion.

However, I don't know if there is a proper way to test `useCroppedImageSelector` with `jest` since it heavily relies on `window.wp.media` library directly and the library depends on `jQuery` and `backbone.js`. In runtime, they are prebuilt files loaded via WP core PHP functions. And mocking the entire `wp.media` seems impractical, like trying to test a React component by mocking the entire `react` library.

Or do you know of a proper way to do so?

PR #1863, opened 2023-01-20
PMax Assets: Assets form for managing asset group

eason9487:

This PR implements the "Assets form" part of 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) in #1787.

The assets form includes:

- Add `AssetGroupCard` for managing the an asset group
- Validate asset group and show error messages after clicking on the "Launch paid campaign" button
- After clicking on "Launch paid campaign", scroll to the error asset field which is nearest the top of page
- Adjust `CampaignAssetsForm` to validate the campaign values only

Other required changes:

- Add `extendAdapter` prop to `AdaptiveForm` for customizing the adapter of the form context
- Add `className` prop to `AssetField` and also make `numOfIssues` and `subheading` props optional

### Screenshots:

https://user-images.githubusercontent.com/17420811/213707451-e25eea85-b154-4b17-93b3-8dcd682f2458.mp4

💡 Please note that this PR doesn't include the API communication of creating/updating the asset group.

2. Load an asset group via the final URL.
3. Test the form UI of managing the asset group.
4. Test the form validation and error messages.

### Changelog entry

> Add - Asset group management form for the campaign creation and editing pages.


puntope:
💅  Behaviour looking good. However, I'm thinking if it would be better to show the issues in the very beginning instead of when the user performs a change or clicks on Launch Campaign. I guess this is more a UX question (I didn't check if there is any discussion about in P2)

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -25,8 +25,8 @@ import FaqsSection from './faqs-section';
 /**
  * Renders the container of the form content for campaign management.
  *
- * Please note that this component relies on an AdaptiveForm's context, so it expects
- * a context provider component (`AdaptiveForm`) to existing in its parents.
+ * Please note that this component relies on an CampaignAssetsForm's context and custom adapter,
+ * so it expects a `CampaignAssetsForm` to existing in its parents.
```
puntope, line 29: ❓  I can see `<AdsCampaign>` being rendered in `<AdsStepper>` which is rendered by `<SetupAdsForm>` using just `<AdaptiveForm>`. So seems like it's not true that  `<CampaignAssetsForm>`  is a required wrapping component for  `<AdsCampaign>`.

js/src/components/paid-ads/asset-group/asset-group-card.js
```diff
@@ -0,0 +1,236 @@
...
+	{ label: 'Download', value: 'DOWNLOAD' },
+	{ label: 'Book now', value: 'BOOK_NOW' },
+	{ label: 'Shop now', value: 'SHOP_NOW' },
+];
+
+/**
+ * Renders the Card UI for managing the an asset group.
+ *
+ * Please note that this component relies on an CampaignAssetsForm's context and custom adapter,
+ * so it expects a `CampaignAssetsForm` to existing in its parents.
```
puntope, line 43: Since this is only rendered in `<AssetGroupSection>` I would remove this duplicated docs. I take as granted `<AssetGroupCard>` is not going to be used outside  `<AssetGroupSection>`.

js/src/components/paid-ads/asset-group/asset-group-section.js
```diff
@@ -2,25 +2,27 @@
...
-import ImagesSelector from './images-selector';
-import TextsEditor from './texts-editor';
-import AssetField from './asset-field';
+import AssetGroupCard from './asset-group-card';
 import './asset-group-section.scss';
 
 /**
  * Renders the form content for managing an asset group of a campaign with Section UI.
+ *
+ * Please note that this component relies on an CampaignAssetsForm's context and custom adapter,
```
puntope, line 21: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1863/files#r1087449867

Seems like this is only used in `<AssetGroup>` so I guess we could remove the comment duplication.

js/src/components/paid-ads/asset-group/asset-group-card.js
puntope: Seems like this is only used in `<AssetGroup>`. So IMO better only in `<AssetGroup>` to avoid duplication.

js/src/components/paid-ads/constants.js
```diff
@@ -0,0 +1,282 @@
...
+						`The length of the rendered description will depend on the site it appears on. If it's shortened, it will end with an ellipsis(…). The description doesn't show in all sizes and formats.`,
+						'google-listings-and-ads'
+					) }
+				</p>
+			</>
+		),
+	},
+];
+
+{
```
puntope: ❓  Why we need to wrap this with `{}`

js/src/components/paid-ads/constants.js
```diff
@@ -0,0 +1,282 @@
...
+	},
+	{
+		maxCharacterCount: 15,
+		capitalizedName: _x(
+			'The second display URL path',
+			'Capitalized asset field name as the start of an error message',
+			'google-listings-and-ads'
+		),
+	},
+];
```
puntope: 💅 WDYT about moving this one at the top to avoid the dynamic logic and the private functions `getImageHelpContent` and `getSubheading` to be mixed between the constants.

js/src/components/paid-ads/asset-group/asset-group.js
```diff
@@ -34,9 +34,17 @@ export default function AssetGroup( { campaign } ) {
...
+	const { isValidCampaign, isSubmitting, isSubmitted, submitter } = adapter;
 	const currentAction = submitter?.dataset.action;
 
+	const handleLaunchClick = ( ...args ) => {
+		if ( isValidForm ) {
+			handleSubmit( ...args );
+		} else {
+			adapter.showValidation();
+		}
+	};
```
puntope, line 46: ❓  Maybe to discuss with UX if it would be convenient to show the validation after selecting URL and populate assets the first time?

js/src/components/paid-ads/asset-group/asset-group-card.js
```diff
@@ -0,0 +1,236 @@
...
+				initialExpanded={ isSelectedFinalUrl }
+			>
+				<SelectControl
+					options={ ctaOptions }
+					value={ ctaProps.value || ctaOptions[ 0 ].value }
+					onChange={ ctaProps.onChange }
+				/>
+			</AssetField>
+			<AssetField
+				ref={ refFirstErrorField.bind( 'display_url_path' ) }
```
puntope: 💅  WDYT of using a constant instead of hardcoding the string?

js/src/components/paid-ads/asset-group/asset-group-card.scss
```diff
@@ -0,0 +1,74 @@
...
+				list-style: none;
+				padding-left: 0;
+			}
+		}
+	}
+}
+
+.gla-asset-field-call-to-action {
+	.components-base-control {
+		width: 240px;
```
puntope, line 32: 💅  Maybe we can create a new dimension inside `js/src/css/abstracts/_variables.scss` or even use `$gla-width-small` if it's not very problematic to get that 10px extra.

js/src/components/paid-ads/campaign-assets-form.js
```diff
@@ -1,28 +1,83 @@
...
 import AdaptiveForm from '.~/components/adaptive-form';
 import validateCampaign from '.~/components/paid-ads/validateCampaign';
+import validateAssetGroup from '.~/components/paid-ads/validateAssetGroup';
 
 /**
  * @typedef {import('.~/components/types.js').CampaignFormValues} CampaignFormValues
+ * @typedef {import('.~/components/types.js').AssetGroupFormValues} AssetGroupFormValues
  */
 
+const emptyAssetGroup = {
```
puntope, line 18: 💅  Maybe we can rename this in uppercase `EMPTY_ASSET_GROUP`

js/src/components/paid-ads/constants.js
```diff
@@ -0,0 +1,282 @@
+/**
```
puntope: 💅  I would name this `specs.js` or `specs-config.js` to be less generic

puntope (approved):
✅  LGTM

I tested the behaviour as well as the code quality.

Left some non blocking comments. Approved in advance

js/src/components/paid-ads/ads-campaign.js
eason9487: Thanks for catching this. The form component in `SetupAdsForm` indeed needs to be changed to `CampaignAssetsForm` as well, otherwise the "Continue" button in the ads setup flow will always be disabled. Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1863/commits/716526c20f7b332baf1aafadd249c87d2ee8500f.

js/src/components/paid-ads/asset-group/asset-group-card.js
eason9487: Added this code comment is not based on the composition of components, but on whether a component uses `useAdaptiveFormContext` and the form adapter extended by `CampaignAssetsForm`.

It's independent and unrelated between components, so it's not a duplication but an indication of the component's dependency on the context provider.

js/src/components/paid-ads/asset-group/asset-group-section.js
eason9487: Same as https://github.com/woocommerce/google-listings-and-ads/pull/1863#discussion_r1090437704.

js/src/components/paid-ads/constants.js
eason9487: This scoped block is to signal that dynamic initialization is handled here, and that the functions inside are for initialization purposes only and are not expected to be exposed outside this module. I added this as the code comment in https://github.com/woocommerce/google-listings-and-ads/pull/1863/commits/de0abb4063db89f7db3caa396cdee86cb057c711.

js/src/components/paid-ads/constants.js
eason9487: Moved by https://github.com/woocommerce/google-listings-and-ads/pull/1863/commits/03dd7667ef2c109b7490ba83ca15b482f1afc49a.

js/src/components/paid-ads/asset-group/asset-group-card.js
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/1863/commits/82af988567984bb248bba1d096000f96c30b6971.

js/src/components/paid-ads/asset-group/asset-group-card.scss
eason9487: The `$gla-width-small` is a generic name that does not semantically fit the purpose here. Currently, there is only this component that needs this width value for aligning the `<input>` and `<select>`. I tend to move it to **_variables.scss** when there is a more wide reuse.

js/src/components/paid-ads/campaign-assets-form.js
eason9487: Unless variables will be exposed outside a module, it does not seem to be a clear convention for variables used only inside a module. Searching this codebase globally with `^const [a-z]\w* = [\[\{]` shows plenty of non-all-uppercase naming.

js/src/components/paid-ads/constants.js
eason9487: Renamed to `asset-specs.js` by https://github.com/woocommerce/google-listings-and-ads/pull/1863/commits/856763cd1103f707523f1399a526b7da7baf5ce6.

js/src/components/paid-ads/asset-group/asset-group.js
eason9487: @puntope, thanks for the suggestion.

> Maybe to discuss with UX if it would be convenient to show the validation after selecting URL and populate assets the first time?

> However, I'm thinking if it would be better to show the issues in the very beginning instead of when the user performs a change or clicks on Launch Campaign. I guess this is more a UX question (I didn't check if there is any discussion about in P2) - Quoted from https://github.com/woocommerce/google-listings-and-ads/pull/1863#issuecomment-1403491807

The flow of showing error messages can be found in this design walkthrough video: pcTzPl-1eS-p2#design-walkthrough

https://user-images.githubusercontent.com/17420811/215654469-4357d7bc-b6ef-4f2a-ae63-3b87773cd28e.mp4

The common ways of handling form validation errors are:

1. Errors are displayed after the user leaves an input field.
2. Errors are displayed after the user submits the form.
3. Start displaying errors while the user is entering.

While it's not best practice to display errors after clicking the submit button, displaying errors before the user has entered data or has touched any input fields is considered a bad design (aka premature validation) in many UX recommendations. The reasons for avoiding overly aggressive premature validation are:

- Users may feel frustrated and unfairly reprimanded when a site claims they’ve made mistakes before they’ve had a chance to enter a valid input.
- Users may be distracted from their typing.
- The first glance of the form for users is a bunch of errors that may be very scary.

Therefore, I think it's probably not necessary to discuss this approach with the designer. But if you think this is still the better way to go, I will still discuss it with the designer.

PR #1866, opened 2023-01-27
Validate PHP Architecture 64 bits (Show notice)

### LINKED ISSUE #1849: 64 bit PHP as a prerequisite
## User story
Google Merchant Center is moving it's account ID's to 64 bit (signed) integers, which means we'll have to support any account ID's larger than 32 bit. In the extension we handle all account ID's as the type `int`. As long as a 64 bit PHP version is installed that means an int is a signed 64 bit integer.

Since 64 bit is the norm, we want to add it as a pre-requisite to run the extension. Similar to how we require at least PHP 7.4.
We can use `PHP_INT_SIZE` to determine whether it's a 64 bit version or not.

### Tasks
- [x] Add a requirement check for 64 bit PHP
- [x] Add the requirement to readme.txt, README.md and any other documentation where prerequisites/minimum requirements are mentioned
- [x] Add the requirement to the public documentation: https://woocommerce.com/document/google-listings-and-ads/

### Project details
pcTzPl-1jE-p2

puntope:

Closes #1849 

This PR will show an admin notice in case the PHP version is working on 32 bits.

### Screenshots:

Screenshot 2023-01-27 at 17 30 46

1. Force the conditional to be true in `src/Internal/Requirements/PHPValidator.php::L58` or use PHP with 32 bits 
2. Reload the site with GLA Installed or Install/Activate it from scratch
3. See the admin notice 
4. Force the conditional to be true in `src/Internal/Requirements/PHPValidator.php::L58` or use PHP with 64 bits 
5. Reload the site with GLA Installed or Install/Activate it from scratch
6. You don't See the admin notice 

### Additional details:

I'm unsure how to switch to 32 bits PHP. If you know, please comment bellow 

### Changelog entry

> Tweak - Show admin notice when PHP 32 bits is being used


src/Internal/Requirements/VersionValidator.php
```diff
@@ -8,7 +8,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Class VersionValidator
+ * Class PHPValidator. Validates PHP Requirements like the version or the architecture.
```
jorgemd24: ❓ Why PHPValidator if the class is named VersionValidator? 

> Validates PHP Requirements like the version or the architecture

I would replace `or` with `and`, as we are doing both validations.

src/Exception/InvalidVersion.php
```diff
@@ -56,4 +56,16 @@ public static function requirement_missing( string $requirement, string $minimum
 			)
 		);
 	}
+
+	/**
+	 * Create a new instance of the exception when an invalid architecture is detected.
+	 *
+	 * @since x.x.x
+	 * @return static
```
jorgemd24: ❓ 

In the function signature we say that we return `InvalidVersion` but in the PHPDocs `static`, wouldn't it be more convenient to use `InvalidVersion` in both cases?

src/Internal/Requirements/VersionValidator.php
```diff
@@ -39,4 +40,19 @@ protected function validate_php_version() {
 			throw InvalidVersion::from_requirement( 'PHP', PHP_VERSION, WC_GLA_MIN_PHP_VER );
 		}
 	}
+
+	/**
+	 * Validate the PHP Architecture being 64 Bits.
+	 * This is done by checking PHP_INT_SIZE. In 32 bits this will be 4. In 64 Bits this will be 8
```
jorgemd24: I would say: This is done by checking PHP_INT_SIZE. In 32 bits this will be 4 **( Bytes )**. In 64 Bits this will be 8 **( Bytes )**.

jorgemd24 (approved):
Thanks @puntope for working on this, the code is working well, I tested forcing the conditional. I left some comments about the PHPDocs.

Related to [public information](https://woocommerce.com/document/google-listings-and-ads/), how can it be updated?

src/Internal/Requirements/VersionValidator.php
puntope: > ❓ Why PHPValidator if the class is named VersionValidator?

I forgot to change the doc. Correct name is VersionValidator

PR #1872, opened 2023-02-06
PMax Assets: Preprocessing for API integrations

eason9487:

This PR implements the preprocessing for 📌 [Integrating with backend APIs](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-integrating-apis) in #1787.

In order to avoid the subsequent PR of the API integrations being too large, some are split into preprocessing:

- Fix that the submission of "Skip adding assets" button is blocked when creating and updating campaigns.
- Rename the **asset-specs.js** file to **assetSpecs.js** since most non-component file names were using camel case naming.
- Remove a misplaced test from the **validateAssetGroup.test.js** (https://github.com/woocommerce/google-listings-and-ads/commit/2185ff13acfcf100868d6122ed8e0c3f55c3efb9)
- Create constants for the keys of asset group and related data structures as they will be wildly used later.
- Adjust the type definitions of asset group and related data structures to map to the response of asset groups and suggested assets. (Examples of the different API responses are shown in the **Additional details** section below or can be found in the implementation of #1874.)
- Update the values of the Call To Action options in the `AssetGroupCard` because the actual values should be the small cases.
- Cast the type of the empty `assets` in the response of the asset groups fetching API from array to object. So that the `assets` will always be the JSON object type.

2. Check if the "Continue" is enabled after entering valid audiences and the budget.
3. Go to the second step.
4. Check if the "Skip adding assets" can proceed with the campaign creation.

### Additional details:

An example data structure of `GET assets/suggestions?id=ID` response:

```json
{
  "headline": [
    "Beanie with Logo"
  ],
  "long_headline": [
    "Cat Pota: Beanie with Logo"
  ],
  "description": [
    "This is a simple product.",
    "Just another WordPress site"
  ],
  "logo": [],
  "business_name": "Cat Pota",
  "square_marketing_image": [
    "https://wp.test/wp-content/uploads/2021/01/beanie-with-logo-1.jpg"
  ],
  "marketing_image": [
    "https://wp.test/wp-content/uploads/2021/01/beanie-with-logo-1-800x418.jpg"
  ],
  "call_to_action_selection": null,
  "display_url_path": [
    "beanie-with-logo"
  ],
  "final_url": "https://wp.test/product/beanie-with-logo/"
}
```

An example data structure of `GET ads/campaigns/asset-groups?campaign_id=ID` response:

```json
[
  {
    "id": 6459681012,
    "final_url": "https://easonsu.a8c-localtunnel.cyou",
    "display_url_path": [
      "path-1",
      "path-2"
    ],
    "assets": {
      "square_marketing_image": [
        {
          "id": 63949253641,
          "content": "https://tpc.googlesyndication.com/simgad/13484709713110638229"
        }
      ],
      "marketing_image": [
        {
          "id": 63960463580,
          "content": "https://tpc.googlesyndication.com/simgad/6228362817145174537"
        }
      ],
      "logo": [
        {
          "id": 63960606224,
          "content": "https://tpc.googlesyndication.com/simgad/4585804890802605392"
        }
      ],
      "headline": [
        {
          "id": 67311903986,
          "content": "headline-1"
        },
        {
          "id": 67311903989,
          "content": "headline-2"
        },
        {
          "id": 67311903992,
          "content": "headline-3"
        }
      ],
      "long_headline": [
        {
          "id": 67311904001,
          "content": "long-headline"
        }
      ],
      "business_name": {
        "id": 67311904004,
        "content": "Cat Pota"
      },
      "description": [
        {
          "id": 70205253686,
          "content": "description-1"
        },
        {
          "id": 70205253689,
          "content": "description-2"
        }
      ]
    }
  }
]
```


js/src/data/types.js
```diff
@@ -9,7 +9,33 @@
...
+ */
+
+/**
+ * @typedef {Object} AssetsDictionary
+ * @property {AssetEntity} [business_name] The name of merchant's business or brand.
+ * @property {AssetEntity[]} [headline] The headlines for the ad.
+ * @property {AssetEntity[]} [long_headline] The headlines for the larger ad.
+ * @property {AssetEntity[]} [description] The descriptive text for the ad to provide additional context or details.
+ * @property {AssetEntity[]} [marketing_images] The URLs of the landscape images.
+ * @property {AssetEntity[]} [square_marketing_images] The URLs of the square images.
```
eason9487, line 28: These property names are incorrect and have been fixed in PR #1873. Ref: https://github.com/woocommerce/google-listings-and-ads/pull/1873#discussion_r1099983959

js/src/data/types.js
```diff
@@ -9,7 +9,33 @@
...
+ */
+
+/**
+ * @typedef {Object} AssetEntity
+ * @property {number} id The ID of the asset.
+ * @property {string} content The content of the asset.
+ */
+
+/**
+ * @typedef {Object} AssetsDictionary
```
puntope, line 22: 💅  As an Idea we can follow the naming in the backend. `Asset`, `AssetGroup` and `AssetGroupAssets`

js/src/constants.js
```diff
@@ -79,3 +79,25 @@ export const CATEGORY_CONDITION_SELECT_TYPES = {
 };
 
 export const CATEGORIES_TO_SHOW_IN_TOOLTIP = 5;
+
+// Assets of Performance Max Campaign related
+export const ASSET_KEY = {
```
puntope, line 84: Since this constants are a set of keys I wonder if we should rename them as ASSET_KEYS, ASSET_GROUP_KEYS, and ASSET_FORM_KEYS

js/src/components/paid-ads/validateAssetGroup.test.js
```diff
@@ -120,57 +128,44 @@ describe( 'validateAssetGroup', () => {
...
-				const limit = specs[ i ].maxCharacterCount;
-				return i.toString().repeat( limit + 1 );
-			} );
+			values[ ASSET_FORM_KEY.DISPLAY_URL_PATH ] = Array.from(
+				{ length },
+				( _, i ) => {
+					const limit = specs[ i ].maxCharacterCount;
+					return i.toString().repeat( limit + 1 );
+				}
+			);
```
puntope: I was hard to understand and read the logic under this block. I would write it in a more verbose way.

puntope (approved):
✅  LGTM

Checked the code as well as the behaviour.

Left some 💅  comments

js/src/data/types.js
eason9487: Originally I had tried to use the naming of `Asset` and `AssetGroup`, but considering that there are many variables, docs or filenames in the frontend that use/include the same naming, they imply more abstract structures or different things like components. About `AssetGroupAsset`, it's the name that describes the link between an `Asset` unit and an `AssetGroup` unit at the database layer of Goodle Ads. But there is no such concept of data link in the frontend, because when getting the data from backend API, `Asset` and `AssetGroup` are already linked together as a result.

The purpose here is to explicitly indicate that a structure belongs to the data obtained by the backend API, so I finally added `Entity` and `Dictionary` to distinguish them.

js/src/constants.js
eason9487: The plural is usually easier to be deduced as an array type, while the singular is usually not deducible as a specific type, so I think the singular is still more appropriate for this case. Sometimes I would use the naming like `FOO_BAR_DICT` (DICT for the abbr of dictionary) but it is also often too wordy.

js/src/components/paid-ads/validateAssetGroup.test.js
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/1872/commits/80e1c00c1e98ad4c06bf3a6f0e66d2f91fe5a456.

js/src/constants.js
puntope: I understand your thoughts. But for me, it was a bit confusing `ASSET_FORM_KEY` in singular. I thought was some kind of API KEY or something like that. Was unexpected to see a set of keys inside. 

Maybe FIELDS instead of KEY would make it clearer.

js/src/data/types.js
puntope: So its not this estructure the one we get on `get_assets_by_asset_group_ids` when we call `ads/campaigns/asset-groups`? 

Another way of seeing it. Having that consistency between BE and FE could help to understand connections between both sides of the implementation. As always naming is a very subjective thing.

eason9487: > So its not this estructure the one we get on `get_assets_by_asset_group_ids` when we call `ads/campaigns/asset-groups`?

It is, but the type relies on the backend APIs rather than the actual or internal implementation behind the APIs.

> Another way of seeing it. Having that consistency between BE and FE could help to understand connections between both sides of the implementation.

When possible, it would be nice to have the same naming between backend and frontend. However, the naming styles (`final_url` and `finalUrl`) between them are different by default, and the targets and layers of abstraction are different as well. Therefore, it's usually not that easier to make them consistent.

Here it uses `AssetsDictionary` instead of other names I thought of while developing, e.g. `Assets`, `AssetsData`, `AssetsObject`, `AssetEntities`, `AssestEntity`, because I thought it's the least confusing one with the other names.

PR #1873, opened 2023-02-08
PMax Assets: Support the portrait marketing images

eason9487:

This PR implements the "Assets form" part of 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) in #1787.

- Support the portrait marketing image and adjust the maximum number of each marketing image type to make the sum of the maximum limits 20.

The limits are adjusted to:
- MARKETING_IMAGE: min 1, max 8
- SQUARE_MARKETING_IMAGE: min 1, max 8
- PORTRAIT_MARKETING_IMAGE: min 0, max 4
   - Image size: (4:5) 960 x 1200 recommended; 480 x 600 min

#### 📷 Added portrait marketing image asset and adjusted maximum limits

image

#### 📷 Validation of the portrait marketing image asset

image

1. Go to the campaign creation or editing page.
2. Check if the limits of each marketing image type match the proposed solution.
3. Check if the UI of the portrait marketing image asset work well.
4. Click the "Save changes" button to see if the validation of all marketing image types works correctly.

### Additional details:

- Reason for adjustment: pcTzPl-12o-p2#comment-2228
- Adopted solution for the present: pcTzPl-12o-p2#comment-2247


js/src/data/types.js
```diff
@@ -7,6 +7,7 @@
  * @property {string[]} description The descriptive text for the ad to provide additional context or details.
  * @property {string[]} marketing_images The URLs of the landscape images.
  * @property {string[]} square_marketing_images The URLs of the square images.
```
jorgemd24: Not introduced in this PR, and probably it was updated when I started to use the field types from the Google Ads library.

`marketing_images` and `square_marketing_images` should be in singular: `marketing_image` and `square_marketing_image`

js/src/data/types.js
```diff
@@ -26,6 +27,7 @@
  * @property {AssetEntity[]} [description] The descriptive text for the ad to provide additional context or details.
  * @property {AssetEntity[]} [marketing_images] The URLs of the landscape images.
  * @property {AssetEntity[]} [square_marketing_images] The URLs of the square images.
```
jorgemd24: See previous message.

jorgemd24 (approved):
LGTM 👍 

I tested locally and can see that the validation is working and the image gets cropped to the correct size. 

Other things/opinions beyond this PR:

- I think the delete button looks too big compared with the preview images, what do you think? Or comparing with the Figma file, it seems that the preview images should be bigger. 

image

- When I am in the editing asset section and I refresh the page, I get back to step one, shouldn't we get back to the editing asset section?
- Do you think we need the button "Scan for assets"? After I choose the Final URL, the only option possible is to scan for the assets, so I would expect to do this automatically.

js/src/data/types.js
eason9487: Thanks for reminding me. Updated in https://github.com/woocommerce/google-listings-and-ads/pull/1873/commits/3736430c3ddee5f406e4700c64b4fc2d0a7a961a.

js/src/data/types.js
eason9487: Thanks for catching this. Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1873/commits/3736430c3ddee5f406e4700c64b4fc2d0a7a961a.

eason9487:
@jorgemd24, thanks for the review.

> * I think the delete button looks too big compared with the preview images, what do you think? Or comparing with the Figma file, it seems that the preview images should be bigger.
> image

Could you provide the WP and WC versions you used for testing? From the screenshot, it seems that there were many styles that were overwritten or not applied to buttons and images. This might be a compatible issue or some styles were not loaded correctly for some reason.

The styles on my local env look like this:

image

> * When I am in the editing asset section and I refresh the page, I get back to step one, shouldn't we get back to the editing asset section?

I guess you were on the campaign creation page. Only the campaign editing page will stay on the same step after refreshing the page. The creation page doesn't resume the current step because it doesn't keep the entered form values on the first step so it requires the user to re-fill from the first step.

> * Do you think we need the button "Scan for assets"? After I choose the Final URL, the only option possible is to scan for the assets, so I would expect to do this automatically.

If I understood it right, it tended to align with the same Card UI pattern of other similar features in GLA, so I think it should be fine. I will forward it to the designer to see if it needs to be adjusted.

jorgemd24:
> Could you provide the WP and WC versions you used for testing? From the screenshot, it seems that there were many styles that were overwritten or not applied to buttons and images. This might be a compatible issue or some styles were not loaded correctly for some reason.

WP version: 6.1.1
WC version: 7.3.0

Let me know if you can reproduce it with those versions, otherwise, I will look deeper into the issue.

> I guess you were on the campaign creation page. Only the campaign editing page will stay on the same step after refreshing the page. The creation page doesn't resume the current step because it doesn't keep the entered form values on the first step so it requires the user to re-fill from the first step.

It happens when I click on the editing campaign button, see this video:

https://user-images.githubusercontent.com/2488994/217787824-8c7f6639-4e07-4bd1-bf47-1879df610ec7.mp4

> If I understood it right, it tended to align with the same Card UI pattern of other similar features in GLA, so I think it should be fine.

👍

jorgemd24:
The conflict with the styling is with AutomateWoo 5.6.6 when I disable AW loads correctly.

eason9487:
@jorgemd24, thanks for the seconde round and the more infomation!

> It happens when I click on the editing campaign button, [...]

You are right, it's a bug. I previously tested it with another "Edit" button on the campaign table, which is for demo purpose. 🤦‍♂️  I opened PR #1877 for this.

> The conflict with the styling is with AutomateWoo 5.6.6 when I disable AW loads correctly.

Nice finding! I guess it relates to #1391.

eason9487:
Hi @jorgemd24, I got the update on this question.

> > * Do you think we need the button "Scan for assets"? After I choose the Final URL, the only option possible is to scan for the assets, so I would expect to do this automatically.
> 
> If I understood it right, it tended to align with the same Card UI pattern of other similar features in GLA, so I think it should be fine. I will forward it to the designer to see if it needs to be adjusted.

After checking with the designer, the decision is to stay with the current UI, and here are the reasons copied from the conversation:

- Actually the reason we did the button was to be consistent about the pattern. We use a similar pattern for connecting accounts. We want to share feedback about what exactly is happening after users select a page.
- Since we only have 1 flow for the MVP version (change URL + import assets), I’m thinking to keep things as is for now, or simplify the wording and use “Select”. B/c in the future, changing the URL won't always mean “scan for assets”.

jorgemd24:
Thanks so much for the explanation @eason9487 !

PR #1874, opened 2023-02-08
PMax Assets: Integrate with APIs to load and save assets

eason9487:

This PR implements parts of 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) and a part of 📌 [Integrating with backend APIs](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-integrating-apis) in #1787.

- Load existing asset groups or suggested assets
- When creating a new campaign, update the assets into the empty asset group of the newly created campaign
- When editing a campaign, create an asset group if needed, and update the assets into the existing or newly created asset group
- The Final URL of the non-empty asset group is not allowed to change 

#### 📹 Update a campaign without an existing asset group

https://user-images.githubusercontent.com/17420811/217511157-9d795612-96f1-430c-b8aa-7c50270d30b6.mp4

This video includes:

- Create a new asset group first if needed
- Avoid re-creating another new asset group if the subsequent asset group update is failed
- Update the assets for the campaign
- Enter the same editing page to see if the update is in effect

#### 📹 Update a campaign with an existing asset group

https://user-images.githubusercontent.com/17420811/217512174-a4ff1e16-6ce7-4fa9-abaa-c128e2ce3d34.mp4

This video includes:

- The Final URL of the non-empty asset group is not allowed to change
- Except for the Final URL, update all assets into the existing asset group
- Enter the same editing page to see if the update is in effect

#### 📹 Create a campaign with assets

https://user-images.githubusercontent.com/17420811/217514975-345883b1-fdcc-4ba0-a6c5-11b9f4c91551.mp4

This video includes:

- Avoid re-creating another new campaign if the subsequent asset group update is failed
- Update the assets into the empty asset group of the newly created campaign
- Enter the editing page of the same campaign to see if the assets are saved
- Start creating another campaign and loading the suggested assets with the same Final URL from the Google Ads end

#### 💡 Additional notes before testing

- Google Ads side will check if the domain in the Final URL matches the claimed domain of the linked Merchant Center account.
   image
   Therefore, it may need to make the local env have the same domain by switching the local env to be accessible publicly or replacing the domain temporarily. The way I used:
   - Replace [this line](https://github.com/woocommerce/google-listings-and-ads/blob/2f448130ce83497e46102a63e28464505f30e146/js/src/components/paid-ads/convertToAssetGroupUpdateBody.js#L86) with:
      ```js
      final_url: values.final_url.replace( 'your local domain', 'GMC domain' ),
      ```
   - Add the following at [this line](https://github.com/woocommerce/google-listings-and-ads/blob/2f448130ce83497e46102a63e28464505f30e146/src/Ads/AssetSuggestionsService.php#L191):
      ```php
      $final_url = str_replace( 'your local domain', 'GMC domain', $final_url );
      ```
   - Add the following at the end of **google-listings-and-ads.php** if got the "cURL error 60" issue with `wp_remote_get`:
      ```php
      add_filter( 'https_ssl_verify', '__return_false' );
      add_filter( 'https_local_ssl_verify', '__return_false' );
      ```
- Please note that the images may not be loaded properly the first few times after saving new asset images to Google Ads end. Looks like my network or Google's CDN takes a little more time to load the images correctly.

#### Test creation page

2. Select a Final URL that has never been used for any asset groups before.
3. Fill in all asset fields but enter at least an invalid content to text assets, for example, emoji chars.
4. After clicking "Saving changes", it should result in a submission failure.
5. Remove the invalid chars entered in step 3.
6. Click "Saving changes" again to see if the assets are saved.
7. After back to the Dashboard page, it should be no duplicate campaign even if the submission failure occurred intentionally during steps 3-4.
8. Go to the campaign creation page again.
9. Select the same Final URL as step 2 and load the suggested assets.
10. Check if the suggested assets are the same as the previously created ones.

#### Test editing page

1. Create a new campaign by skipping adding assets.
2. Go to the editing page of the newly created campaign. (Currently, the asset group of this campaign is empty.)
3. Fill in all asset fields.
4. Click "Saving changes" again to see if the assets are saved.
5. Go to Google Ads admin page
   1. Navigate to the assets management tab of the same campaign
   2. Remove all asset groups
      
6. Reload GLA web page and go to the editing page of the same campaign. (Currently, this campaign has no asset groups.)
7. Add at least an invalid content to text assets, for example, emoji chars.
8. After clicking "Saving changes", it should result in a submission failure.
9. Remove the invalid chars entered in step 7.
10. Click "Saving changes" again to see if the assets are saved.
11. Switch back to Google Ads admin page and reload it.
    - There should be only one asset group linked to the same campaign.
    - It should be no other (empty) asset groups campaign even if the submission failure occurred intentionally during steps 7-8.


js/src/components/paid-ads/asset-group/asset-group.js
```diff
@@ -57,7 +55,10 @@ export default function AssetGroup( { campaign } ) {
 			/>
 			<AssetGroupSection />
 			<StepContentFooter>
-				{ ( isCreation || isEmptyAssetGroup ) && (
+				{ ( isCreation || adapter.isEmptyAssetEntityGroup ) && (
```
puntope, line 58: It's a bit confuse now the difference between `AssetEntityGroup` and `AssetGroup` it's referring to the same?

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Converts the assets form values into the request body based on an existing asset entity group
+ * for updating that existing asset group.
+ *
+ * @param  {AssetEntityGroup} assetGroup The existing asset entity group as the base data for comparing the differences.
+ * @param  {AssetGroupFormValues} values The assets form values to be converted.
+ * @return {AssetEntityGroupUpdateBody} The request body for updating the asset group.
+ */
+export default function convertToAssetGroupUpdateBody( assetGroup, values ) {
```
puntope, line 111: It's declared in the docs, but still `formValues` would be more clear than just `values`.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
```diff
@@ -0,0 +1,91 @@
...
+		const deletions = assetEntities.map( ( assetEntity ) => ( {
+			...assetEntity,
+			// Setting the `content` to null indicates the asset deletion operation.
+			content: null,
+			field_type: key,
+		} ) );
+		operations.push( ...deletions );
+	}
+
+	Object.values( ASSET_KEY ).forEach( ( key ) => {
```
puntope, line 58: Same as my comment in previous PRs. It was confusing that we want to iterate over the ASSET_KEYs but name is still singular.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
```diff
@@ -0,0 +1,91 @@
...
+		const contents = [ values[ key ] ].flat().filter( Boolean );
```
puntope, line 61: I also would recommend to change here the `values` and `contents` generic for something more meaningful.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
```diff
@@ -0,0 +1,91 @@
...
+		const assetEntities = [ assets[ key ] ].flat().filter( Boolean );
+		let entityIndex = 0;
+
+		contents.forEach( ( content ) => {
```
puntope, line 65: I as having a little bit of a trouble understanding this function. In concrete when we should push a deletion operation and when a creation.

puntope (commented):
I'm still reviewing this. However I found something to keep an eye on 👁️ 

When I removed the assets from Google Ads. If I go back to the dashboard and click edit (no refreshing the page) it still shows the assets. Only when I refresh they disappear.

js/src/data/actions.js
```diff
@@ -862,6 +870,75 @@ export function* deleteAdsCampaign( id ) {
...
+export function* updateCampaignAssetGroup( assetGroupId, body ) {
+	try {
+		yield apiFetch( {
+			path: `${ API_NAMESPACE }/ads/campaigns/asset-groups/${ assetGroupId }`,
+			method: 'PUT',
+			data: body,
+		} );
+
+		return {
+			type: TYPES.UPDATE_CAMPAIGN_ASSET_GROUP,
```
puntope, line 926: We are using here `TYPES.UPDATE_CAMPAIGN_ASSET_GROUP,` but seems that is not implemented in the reducer?

js/src/components/paid-ads/asset-group/asset-group.js
eason9487: No, they refer to different data sources in the `CampaignAssetsForm` component.

- The state or prop named including `/AssetGroup/i` means that it relates to the form values which contain the structure of `AssetGroupFormValues` type.
- The state or prop named including `/AssetEntityGroup/i` means that it relates to the asset group data which refers to the `AssetEntityGroup` type and is obtained from the backend API.

For example, `assetGroupErrors` is calculated from the form values and it tells the validation results of the form.

Using `isEmptyAssetEntityGroup` as another example, it's calculated from the `assetEntityGroup` data and it tells whether the obtained `assetEntityGroup` data is empty or not. Its state won't be changed along with the form value updates.

If they were named `assetGroupErrors` and `isEmptyAssetGroup`, then it would be more confusing because it couldn't be distinguished they refer to which data source via their name. Furthermore, in the `CampaignAssetsForm` component, there is another variable named `emptyAssetGroup`, which refers to the empty asset group in the form values, and naming `isEmptyAssetGroup` would be more easily confused with it.

Therefore, for the components like `CampaignAssetsForm` having both data sources inside or on their props, the mid-name `Entity` is used to help distinguish the sources.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: I would agree with it if the type of value was just declared `Object`.

However, just because JSDoc has explicitly pointed out the type and use of `values`, a relatively short name was used for this small function. Otherwise, if no matter how clear the JSDos it's, the parameter name should still very align with the explicit stuff, then the next question might be "why `formValues` but not `assetsFormValues` or `assetGroupFormValues`?"

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: Unfortunately, I'm not pretty sure why seeing `Object.values( ASSET_KEY )` is confusing. The only purpose of the `Object.values` function is to accept an object type parameter and return an array of a given object's own enumerable property values.

> It was confusing that we want to iterate over the ASSET_KEYs but name is still singular.

In other words by the same token, `Object.values( ASSET_KEYS )` could be still considered confusing because "why it needs to iterate over an array?"

And It's just the reason why `ASSET_KEYS` might be inferred as the array type and may try to `ASSET_KEYS.forEach()` first then get the "{}.forEach is not a function" error.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: Could you explain more thoughts? Especially the `contents` are the array of `content` values which is the needed name for the API request body, and the primary purpose of this function is to generate the part of mutation operations in the request body. I don't pretty understand why it's considered generic.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: Added code comment in https://github.com/woocommerce/google-listings-and-ads/pull/1874/commits/50bffc816f65f5b03fe13ae553b96be9c92a5bb6. Do let me know if the explanation needs adjustments.

js/src/data/actions.js
eason9487: The API response doesn't have the updated asset group, so no mutations can be done in the reducer. I originally thought it might be good to still signal this update action to **wp-data** even if it changes nothing.

In addition, it could be used in the `shouldInvalidate` of `getCampaignAssetGroups` resolver if someday it needs to invalidate the state rather than calling `invalidateResolution` [here](https://github.com/woocommerce/google-listings-and-ads/blob/50bffc816f65f5b03fe13ae553b96be9c92a5bb6/js/src/pages/edit-paid-ads-campaign/index.js#L118), but it can be added when needed. Do you prefer removing this action type?

eason9487:

> When I removed the assets from Google Ads. If I go back to the dashboard and click edit (no refreshing the page) it still shows the assets. Only when I refresh they disappear.

This happens in a lot of apps as not many apps' runtime states would need to be continuously and instantly synchronized with the server data. GLA also doesn't implement such a mechanism but keeps showing with the runtime states that are fetched and cached via **wp-data** unless a state is signaled invalidated or the webpage is refreshed.

For example, the campaign states in the program table on Dashboard, the states of the connected accounts on Settings, and so on are the same. If load related GLA pages first and then delete campaigns from Google Ads admin page or disconnect accounts from Google admin page, the states on GLA won't reflect that action data changes after switching back and forth within GLA pages, instead, the states show the outdated states.

puntope:
> @puntope, thanks for the review.
> 
> > When I removed the assets from Google Ads. If I go back to the dashboard and click edit (no refreshing the page) it still shows the assets. Only when I refresh they disappear.
> 
> This happens in a lot of apps as not many apps' runtime states would need to be continuously and instantly synchronized with the server data. GLA also doesn't implement such a mechanism but keeps showing with the runtime states that are fetched and cached via **wp-data** unless a state is signaled invalidated or the webpage is refreshed.
> 
> For example, the campaign states in the program table on Dashboard, the states of the connected accounts on Settings, and so on are the same. If load related GLA pages first and then delete campaigns from Google Ads admin page or disconnect accounts from Google admin page, the states on GLA won't reflect that action data changes after switching back and forth within GLA pages, instead, the states show the outdated states.

A ok, if it's expected and we are aware of it. Fine

js/src/data/actions.js
puntope: not any preference. Seems like working. I just was curious why, yes it was confusing for me as you can see, I was looking for the `UPDATE_CAMPAIGN_ASSET_GROUP ` in reducer as we (almost) always do.  Just curious. In this case is it wrong to just return?

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
puntope: In general my troubles come when I read the function (maybe because I'm reading it in GITHUB file changes page)

I read first what content/values variables are holding. Then I follow reading and at some point my mind is like "Wait a moment, what `content` was, and what `values` had? Values was the form ones or the previous assets?" So I go up again and need to confirm what that variable was holding. 

Again, maybe is because I was reading that here in GH instead of my IDE and at night (slow brain) but still I think `content` and `values` are a bit meaningless in the diff logic context. In comparisons I usually like to use `previousValues` / `newValues` or something equivalent. But in previous discussions seems you disliked this approach.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
puntope: I don't know how to explain it in other words. For me seems obvious that we are iterating over a set of (plural) keys and not just a single Key. Not a blocker tho since it's a naming issue. But I still disagree ASSET_KEY is correct

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
puntope: > "why formValues but not assetsFormValues or assetGroupFormValues?"

Yes, that would be ideal 

assetGroup --> existingAssetGroup or currentAssetGroup
values --> assetsGroupFormValues 

I don't think is small function (50 lines).

js/src/components/paid-ads/asset-group/asset-group.js
puntope: Ok, understood. Thanks for the further explanation. It's of course good we have different naming between both cases. 

One issue I see on this approach is that is hard to know at glance the "Entity" is referring to data and without "Entity" is referring to the Form Values

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
puntope: I was playing a bit with the new version. And considering all the previous comments and discussions. Seems like the only part we discuss is about this function. 

the new comments are nice, they clarified a bunch. However I think it would be more useful to add the steps in line with the code to facilitate understanding for the code newcomers in the future.

```javascript
/**
 * Differences the assets form values against the existing asset entity group and
 * returns the updating operations of the asset group.
 *
 * @param  {AssetEntityGroup} currentAssetEntityGroup The existing asset entity group as the base data for comparing the differences.
 * @param  {AssetGroupFormValues} formValues The assets form values to be compared.
 * @return {AssetOperations[]} The creation and deletion operations for updating the asset group.
 */
export function diffAssetOperations( currentAssetEntityGroup, formValues ) {
	const operations = [];

	function pushDeletions( key, ...assetEntities ) {
		const deletions = assetEntities.map( ( assetEntity ) => ( {
			...assetEntity,
			// Setting the `content` to null indicates the asset deletion operation.
			content: null,
			field_type: key,
		} ) );
		operations.push( ...deletions );
	}

	// The assets of Google Ads are immutable once created. Therefore if it wants to update
	// a content of an existing asset, the processing will be made by two operations:
	// 1. Delete that existing asset.
	// 2. Create a new asset.
	//
	// In the general case, the creation order of the contents in an asset is also the order
	// of the asset's contents after fetching the asset group. Thus, this loop compares the
	// contents of form values and the contents of the assets in the existing entity group by
	// the same key and one by one in order, and then differences out the operations.

	// Loops all possible asset keys one by one
	Object.values( ASSET_KEY ).forEach( ( key ) => {
		const currentKeyFormContents = [ formValues[ key ] ]
			.flat()
			.filter( Boolean );
		const currentKeyAssetEntities = [
			currentAssetEntityGroup.assets[ key ],
		]
			.flat()
			.filter( Boolean );
		let entityIndex = 0;

		// For the current Assets Key. Compares the contents of the form values and the contents current assets for that key.
		currentKeyFormContents.forEach( ( content ) => {
			do {
				const assetEntity = currentKeyAssetEntities[ entityIndex ];

				// If a content of the form value is the same as the current asset content and with same index position.
				// this means there is no  operation needed and we should continue with the next form asset content.
				if ( content === assetEntity?.content ) {
					break;
				}

				// it it's different order or content, we should delete that assetEntity
				if ( assetEntity ) {
					pushDeletions( key, assetEntity );
				}

				entityIndex += 1;
			} while ( entityIndex < currentKeyAssetEntities.length );

			// if at this point entityIndex is bigger than the current asset entities it means there is a new form content asset to be added as entity
			if ( entityIndex >= currentKeyAssetEntities.length ) {
				operations.push( {
					// Setting the `id` to null indicates the asset creation operation.
					id: null,
					content,
					field_type: key,
				} );
			}

			entityIndex += 1;
		} );

		// At this point. If an asset content is not found in the contents of the form value, this means it's a deletion
		pushDeletions( key, ...currentKeyAssetEntities.slice( entityIndex ) );
	} );

	return operations;
}
```

p.s The good thing is that after discussion and changes I now understand fully (I guess) the logic of this function. I found really a gotcha the fact that the order of the assets is relevant

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
```diff
@@ -0,0 +1,450 @@
...
+			values[ ASSET_FORM_KEY.FINAL_URL ] = 'https://example/belt';
+			values[ ASSET_FORM_KEY.DISPLAY_URL_PATH ] = [ 'belt', 'specials' ];
+
+			const body = convertToAssetGroupUpdateBody( assetGroup, values );
+
+			expect( body ).toHaveProperty( 'final_url' );
+			expect( body ).toHaveProperty( 'path1' );
+			expect( body ).toHaveProperty( 'path2' );
+			expect( body ).toHaveProperty( 'assets' );
+		} );
```
puntope: I think we are covering this with the further next 2.

js/src/components/paid-ads/asset-group/asset-group.js
eason9487: > One issue I see on this approach is that is hard to know at glance the "Entity" is referring to data and without "Entity" is referring to the Form Values

I won't extremely chase finding a name that let most devs know what it's at a glance for those variables that originally have complex relations or contexts.

Names like `assetGroupFormValuesErrors`, `emptyAssetGroupFormValues` and `isEmptyAssetEntityGroup` might make it but I don't really want a wordy naming just for making it understandable at a glance but hurt the readability once devs understood the meanings through the JSDoc or the context.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
```diff
@@ -0,0 +1,450 @@
...
+
+	let assetGroup;
+	let values;
+
+	beforeEach( () => {
+		assetGroup = structuredClone( filledAssetGroup );
+		values = structuredClone( filledValues );
+	} );
+
+	test( 'in each operation, one of `id` and `content` must be null and the other must not be null', () => {
```
puntope, line 267: 💅  "On each" instead of "in each"

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: > Yes, that would be ideal
> assetGroup --> existingAssetGroup or currentAssetGroup
> values --> assetsGroupFormValues

Sorry, but it's too wordy for me.

Things like this are probably very subjective, so I don't really have a good idea to make all devs happy with the naming. In this kind of case, I usually think if it could be understood after a few seconds (or a few minutes against those very complex things), then the naming is good enough.

> I don't think is small function (50 lines).

Not sure what I missed but this function should have only 11 lines of code.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: Thanks for sharing your thoughts.

> For me seems obvious that we are iterating over a set of (plural) keys and not just a single Key.

Sorry, I'm still unable to understand why making it look like "iterating over a set of keys" could be less confusing. But it's not a set of keys, right? And `Object.values` doesn't even handle a set:
```js
Object.values(new Set([1, 2, 3]))
// returns `[]`
```

None of these names could let us intuitively and immediately know that the variable is an object type when just seeing `ASSET_KEYS` or `ASSET_KEY`.

> But I still disagree ASSET_KEY is correct

I didn't think `ASSET_KEY` is the correct naming either. Both namings work for me, and `ASSET_KEY` is relatively a bit more acceptable naming for me. In the meanwhile, there are roughly dozens of object variables not named with plural if search the codebase with `const \w*(?<![Dd]ata)(?<![sS]) = \{([^\}]+(,$)){2}`. 😂 Many of them might be a bit weird if named with plural.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
```diff
@@ -0,0 +1,450 @@
...
+		// If this test failed, the modified code might be problematic or
+		// need adjustment for corresponding tests.
```
puntope: 💅  What does this mean?  🤔

I guess this applies to all the tests.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
```diff
@@ -0,0 +1,450 @@
...
+		],
+		[ ASSET_FORM_KEY.HEADLINE ]: [
+			'headline 1',
+			'headline 2',
+			'headline 3',
+		],
+		[ ASSET_FORM_KEY.LONG_HEADLINE ]: [ 'long headline 1' ],
+		[ ASSET_FORM_KEY.DESCRIPTION ]: [ 'description 1' ],
+		[ ASSET_FORM_KEY.CALL_TO_ACTION_SELECTION ]: 'sign_up',
+	};
```
puntope: 💅  Wonder if we could use this two as global CONSTANTS at the top. i.e. `DUMMY_ASSET_GROUP` and `DUMMY_FILLED_VALUES` shared between `convertToAssetGroupUpdateBody` and `diffAssetOperations` tests.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
```diff
@@ -0,0 +1,450 @@
...
+				} else {
+					asset = { content };
+				}
+			} else if ( content === null ) {
+				if ( isMultiple ) {
+					asset = asset.filter( ( item ) => item.id !== id );
+				} else {
+					asset = { content: '' };
+				}
+			}
```
puntope: 💅  WDYT of adding a bit of commenting here to explain a bit this nested conditionals?

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
```diff
@@ -0,0 +1,450 @@
...
+	describe( 'should get the same asset contents after applying the diff operations to the asset group', () => {
+		test( 'no differences', () => {
+			const operations = diffAssetOperations( assetGroup, values );
+
+			expect( operations ).toHaveLength( 0 );
+		} );
+
+		test( 'create assets', () => {
+			delete assetGroup.assets[ ASSET_KEY.BUSINESS_NAME ];
+			delete assetGroup.assets[ ASSET_KEY.CALL_TO_ACTION_SELECTION ];
```
puntope, line 300: ❓   Why we are deleting this if we are checking asset creation?

puntope (approved):
After testing the behaviour looks good on me.

We discussed in dev call and since the comments here are about code readability I would mark it as approve. However I still think we should refactor them a bit. Especially the diffOperations function. But could be done in further PRs

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: Sorry, I hardly follow the reason that if any variable names are considered confusing just because the naming is hard to read in GH's file change page.

> [...] but still I think `content` and `values` are a bit meaningless in the diff logic context. In comparisons I usually like to use `previousValues` / `newValues` or something equivalent. But in previous discussions seems you disliked this approach.

It has no bearing on whether I like the approach or not, but just its context is not what you mentioned.

There is only one instance of the form values in this function, so it doesn't have the relative concept or transient status like previous, current, or new. From the perspective of this function, it only needs two different data sources and returns the diffs. Using the naming with a relative concept like `previousValues`, `newValues` would make it very confusing instead.

js/src/data/actions.js
eason9487: > Just curious. In this case is it wrong to just return?

It will work well. Like the `requestPhoneVerificationCode`, although it's not just `return;`, the return doesn't contain the `type` property and it works well. Recuder ignores it as the `switch` is mismatched.

puntope:
Btw. Other thing I saw while testing https://github.com/woocommerce/google-listings-and-ads/pull/1884

- You create a new campaing
- load the assets with an emoji to make it fail
- Launch the campaign 


Then skip assets is disabled. I guess because the Asset’s  Group is created in Ads even tho it failed. Not sure if it's intended

js/src/components/paid-ads/convertToAssetGroupUpdateBody.js
eason9487: Thanks for the suggestion. Added a few more code comments in https://github.com/woocommerce/google-listings-and-ads/pull/1874/commits/c0662c7474916968a8952f14bf61f7e4c4b7e18e. I kept the logic bullets because they describe the concept but do not map to the actual lines of code one by one.

js/src/data/actions.js
eason9487: After actual testing, the `yield` or `return` actions without `type` property won't be passed to the reducer.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
eason9487: The target of this test is to examine whether the returned body has the required properties regardless of the properties' value; The targets of the next two are to examine whether the value of properties matched the expectations.

Although they appear to overlap, when the tests fail, the cause can be located more quickly. For example, if the `final_url` is miswritten to `finalUrl`,
- without this test, the failure will be the following, and whether it's caused by mismatched value or the incorrect property will need further confirming.
   image
- with this test, the failure will plus the following, and the cause is clearly pointed out that the property name is mismatched.
   image

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
eason9487: This test case verifies the contractual rules with the backend API, so if it fails, it will need to be confirmed carefully. I rephrased the code comment to make the reason clearer. See https://github.com/woocommerce/google-listings-and-ads/pull/1874/commits/c155aec7ec2d9a87a77d55c8709cc591ab8638c9.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
eason9487: They are slightly different, and the tests focus on different target functions, so changing to shared constants will cause the test code to require some extra processing for the sake of sharing, which will make the tests slightly out of focus.

However, I found some redundancies and an inconsistent default value when I revisited the code. Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/1874/commits/3e178c737de23fadf1702d99fbc7d99ac626b913.

js/src/components/paid-ads/convertToAssetGroupUpdateBody.test.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/1874/commits/6ce832a6541fa6e446324fc066200a145d6dc4b3.

eason9487:
@puntope, thanks for the second round of code reviews and your consideration of those naming concerns.

> Btw. Other thing I saw while testing #1884
> 
> * You create a new campaing
> * load the assets with an emoji to make it fail
> * Launch the campaign
> 
> Then skip assets is disabled. I guess because the Asset’s Group is created in Ads even tho it failed. Not sure if it's intended

Thanks for catching this. It's a bug caused by syncing the `submitter` behind the update of `submission` state in the `AdaptiveForm` component. The bug should be fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1874/commits/2edffa3d19568fe186639062e553a471938b3055.

Video for demonstrating the fix:

https://user-images.githubusercontent.com/17420811/219301301-8f798672-ee5d-409c-81bb-88193ba6f95b.mp4

PR #1877, opened 2023-02-09
PMax Assets: Stay on the same step after refreshing the campaign editing page

eason9487:

In the code review of https://github.com/woocommerce/google-listings-and-ads/pull/1873#pullrequestreview-1288947337 and https://github.com/woocommerce/google-listings-and-ads/pull/1873#issuecomment-1423966440, @jorgemd24 found that it will back to the first step after refreshing the campaign editing page on the second step.

This PR fixes the bug, and also reverts d3dccb2dbf83af32b23535ff92d1a3bb68a24bbb because the dashboard page was changed to not add the new assets column and that commitment is only a demo.

1. Go to the campaign editing step.
2. Refresh the page on each step and check if it stays on the same step.
3. Check if other relevant navigation functions on the page work correctly.

### DIFF
--- a/js/src/dashboard/all-programs-table-card/index.js
+++ b/js/src/dashboard/all-programs-table-card/index.js
@@ -3,7 +3,6 @@
  */
 import { __, sprintf } from '@wordpress/i18n';
 import { getQuery, onQueryChange } from '@woocommerce/navigation';
-import { Link } from '@woocommerce/components';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


jorgemd24 (commented):
Thanks for working on this. I tested locally and I see that page is staying on the same step, however, it seems that the button "Skip adding assets" does not work.


https://user-images.githubusercontent.com/2488994/217882982-3aaa28fc-dfc4-4739-a265-f21840db1899.mp4

eason9487:
Thanks for the review, @jorgemd24

> however, it seems that the button "Skip adding assets" does not work.

Yes, this bug was introduced in #1863 and will be addressed along with #1872 (18c600abc3b98a4905ab5c49620d03cc194d382f).

jorgemd24 (approved):
Thanks for the clarification 👍

PR #1879, opened 2023-02-10
Add the tours to wp-data and the `useTour` hook

eason9487:

In order to reuse the finished implementation of accessing the tour API for PMax Assets project, this PR picked the shared parts from #1752 and fixes some related code review comments commented by myself.

The shared parts are the tours in the wp-data store and the `useTour` hook. Three commits were partially cherry-picked from #1752:

- 90f73b6d4673f185d823196ad69590c507a2b723
- 62094907daf43c95353e8baaabfbe033fd70d919
- 52916f9cd2c912718b5ccdd745ab31950fd54004

The related fixes and adjustments:

- Fix the uncaught error occurred when querying the non-existing tour in the `getTour` resolver.
- Remove the invalidator of the `getTour` resolver.
- Fix incorrect type definitions of the `useTour` hook and the `receiveTour` action.
- Add the `null` type to the returning value of `getTour` selector and related hook.
- In addition, it adds the `setTourChecked` to the return value of `useTour` hook.

💡 The first three commits in this PR have been reviewed in #1752, if the reviewers tend to check the changes made in this PR, please refer to [this link](https://github.com/woocommerce/google-listings-and-ads/pull/1879/files/da4044640a544a85946b6d4cd6b313226c643b50..c06bcda0fa9d8761b387607553fe94735e789eca).

💡 As this PR only extracted the shared code, it would be easier to test it along with the campaign assets tour in #1880.


jorgemd24:
While I was testing https://github.com/woocommerce/google-listings-and-ads/pull/1880 I felt that is unnecessary to wait until the tour is updated to close the Tour, we could just close the Tour and in the background, we do the request. 

Besides that, what would happen if the Tour API fails? How the user can close the tour pop-up? 

WDYT @puntope? 

See full comment here: https://github.com/woocommerce/google-listings-and-ads/pull/1880#pullrequestreview-1297677428
See the code related:  https://github.com/woocommerce/google-listings-and-ads/pull/1879/files#diff-0c2b3d99302fcbdbf2b14d17ea7a100fca8caa8e5696e565ac2d5ad8e64e0656R30-R54

puntope:
> WDYT @puntope?

Good catch there @jorgemd24 

I will modify the code this week for making the adjustments

puntope:
@jorgemd24 @eason9487 

I did a little refactor here https://github.com/woocommerce/google-listings-and-ads/pull/1879/commits/86286dacf07151346483cd90021549ee05dbe4aa

So now the tour closes firsts and then updates the DB. To avoid complexity I simplified the exposed properties. Now we receive `tour` object, the `showTour` and `closeTour`.  If in the future we need to open tour programatically we can do further refactoring.

Let me know how this looks for you 👌 

PS. @eason9487 if this approach is ok for you you will need to update `setTourChecked(true)` to just `closeTour()`  in your PR https://github.com/woocommerce/google-listings-and-ads/pull/1880

js/src/data/actions.js
```diff
@@ -1120,3 +1120,44 @@ export function* receiveStoreCategories( storeCategories ) {
 		storeCategories,
 	};
 }
+
+/**
+ * Action to receive the Tours.
+ *
+ * @param {Object} tour The tour to receive.
```
jorgemd24: Can we use the typedef defined here instead of Object?  https://github.com/woocommerce/google-listings-and-ads/blob/c71374945deb32e28db5c2fdce99b82927a75146/js/src/data/selectors.js#L44

js/src/data/actions.js
```diff
@@ -1120,3 +1120,44 @@ export function* receiveStoreCategories( storeCategories ) {
...
+	return {
+		type: TYPES.RECEIVE_TOUR,
+		tour,
+	};
+}
+
+/**
+ * Updates/Inserts a Tour action
+ *
+ * @param {Object} tour The tour to update in the state.
```
jorgemd24: See previous comment.

js/src/data/reducer.js
```diff
@@ -444,6 +445,16 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
 
+		case TYPES.RECEIVE_TOUR: {
+			const { tour } = action;
+			return setIn( state, [ 'tours', tour.id ], tour );
+		}
+
+		case TYPES.UPSERT_TOUR: {
+			const { tour } = action;
+			return setIn( state, [ 'tours', tour.id ], tour );
+		}
```
jorgemd24: This code seems that has been duplicated unnecesarily, can we do something like this: 

```
case TYPES.RECEIVE_TOUR:
case TYPES.UPSERT_TOUR: {
    const { tour } = action;
	return setIn( state, [ 'tours', tour.id ], tour );
}

```
?

js/src/hooks/useTour.js
```diff
@@ -0,0 +1,54 @@
...
+ * @property {boolean} showTour Indicates if the tour should be shown based on it's checked prop
+ */
+
+/**
+ * This hook returns a tour status based on a Tour ID, a setter for the tour and a property indicating if the tour should be shown.
+ *
+ * @param {string} tourId Tour ID to get
+ * @return {TourHook} The tour Hook
+ */
+const useTour = ( tourId ) => {
```
jorgemd24, line 30: Do we have any test for this hook?

js/src/hooks/useTour.js
```diff
@@ -0,0 +1,54 @@
...
+			return upsertTour( nextTour );
+		}
+
+		return Promise.resolve( { tour: nextTour } );
+	}, [ checked, setTourClosed, upsertTour, tourId ] );
+
+	return {
+		tour,
+		closeTour,
+		showTour,
```
jorgemd24: `closeTour` is a function and `showTour` is a boolean, the first time that I read the code I thought that `closeTour` and `showTour` were both functions, can we change `showTour` for something like `isTourShown` or something like that?

js/src/data/resolvers.js
```diff
@@ -462,3 +463,35 @@ export function* getStoreCategories() {
 		);
 	}
 }
+
+/**
+ * Resolver for getting the Tours.
```
jorgemd24: Can we get multiple Tours or the Tour?

jorgemd24 (approved):
Thanks, @puntope and @eason9487 for working on this.

I tested the PR along with https://github.com/woocommerce/google-listings-and-ads/pull/1880 and it is working well.

I left some comments about the JSDocs, variable names, some code duplication and tests.

js/src/hooks/useTour.js
```diff
@@ -0,0 +1,54 @@
...
+	const closeTour = useCallback( () => {
+		const nextTour = { id: tourId, checked: true };
+
+		if ( ! checked ) {
+			setTourClosed( true );
+			return upsertTour( nextTour );
+		}
+
+		return Promise.resolve( { tour: nextTour } );
+	}, [ checked, setTourClosed, upsertTour, tourId ] );
```
eason9487: The exposed `showTour` won't be reflected the changes of the hook parameter `tourId`. In general, it's considered an issue when implementing a custom hook.

For example, in the following video:
1. Initially, the `showTour` of both tours is `true`.
2. Close the **test-1** tour.
3. The `showTour` of **test-1** tour is changed to `false`.
4. After switching to **test-2** tour, the `showTour` is `false` but we did not close it.

https://user-images.githubusercontent.com/17420811/219339728-ed0d8326-9bcd-4ecc-a845-356942092ba3.mp4

---

The code for testing the issue. I pasted it at the first line within `Dashboard` function in file: js/src/dashboard/index.js
```js
const [ id, setId ] = useState( 'test-1' );
const { showTour, closeTour } = useTour( id );
return (
	<div>
		<div>{ `Current ID: ${ id }` }</div>
		<div>{ `Current showTour: ${ showTour }` }</div>
		<div>
			<button onClick={ () => closeTour() }>
				Close current tour
			</button>
		</div>
		<button onClick={ () => setId( 'test-1' ) }>
			Set ID to test-1
		</button>
		<button onClick={ () => setId( 'test-2' ) }>
			Set ID to test-2
		</button>
	</div>
);
```

eason9487: The `setTourClosed` in the deps of this `useCallback` could be omitted.

js/src/hooks/useTour.js
puntope: For me no problem

eason9487 (commented):
@puntope, thanks for proposing the approach for https://github.com/woocommerce/google-listings-and-ads/pull/1879#issuecomment-1431244337.

> I did a little refactor here [86286da](https://github.com/woocommerce/google-listings-and-ads/commit/86286dacf07151346483cd90021549ee05dbe4aa)
> 
> So now the tour closes firsts and then updates the DB. To avoid complexity I simplified the exposed properties. Now we receive `tour` object, the `showTour` and `closeTour`. If in the future we need to open tour programatically we can do further refactoring.

I'm not pretty sure if the approach in 86286dacf07151346483cd90021549ee05dbe4aa is a good idea since it seems to bring more technical design concerns. For example:
- The exposed hook interface of `useTour` is confusing. Originally, the `showTour` seems to start with a verb was a bit improper but there was a JSDoc for that so it's fine. However, now adding the `closeTour` make the overall interface easier to confuse.
- The exposed `showTour` won't be reflected the changes of the hook parameter `tourId`. Although it could be solved by adding an observer such as `useEffect` but it might make the implementation of state handling more complex.
- The concept of whether to "show" or "close" a tour in the `useTour` hook and also mix the internal state `checked` in its JSDoc. This makes the purpose of the `useTour` blur. In practice, it is not the `useTour` that determines whether the corresponding popup of a tour is shown and closed, but the components that use it. The `useTour` primarily provides the interface to get and update the `checked` state.

If we want to refine it more fully at this moment, I'd like to propose an alternative solution. The code was committed to `update/1879-use-tour-alternative` branch can be viewed in https://github.com/woocommerce/google-listings-and-ads/commit/9c79cba9b7c9c5481d46f0b83d0f6dd54dddd837. The idea of this approach:

- Makes the interface clearer.
- The immediate update is handled by the `upsertTour` action instead of `useTour`. If needed in the future, we could still add `upsertingClientStoreFirst` to `useTour` parameters, and the `upsertTour` default will remain in the same processing mode as the other wp-data actions. This gives both sides more flexibility.
- The exposed state `tourChecked` reflect the change of the `tourId` parameter.
- Supports toggling the state.
- `setTourChecked` doesn't return the promise instance since the state is updated immediately when using this hook, so asynchronous things about the API should be hidden from the hook users.
- The `upsertTour` action does not throw errors, the UI still shows the toast message when an API error occurs but the users of `upsertTour` don't need to catch it and then ignore it.
- Rephrase useTour's JSDoc to make it appear to focus on managing the checked state only.

---

For testing the proposed solution with things like the sync of `checked` state after switching the tour ID. (Paste it at the first line within Dashboard function in file: js/src/dashboard/index.js)
```js
const [ id, setId ] = useState( 'test-1' );
const { tourChecked, setTourChecked } = useTour( id );
return (
	<div>
		<div>{ `Current ID: ${ id }` }</div>
		<div>{ `Current checked: ${ tourChecked }` }</div>
		<div>
			<button onClick={ () => setTourChecked( ! tourChecked ) }>
				{ `Toggle checked to ${ ! tourChecked }` }
			</button>
		</div>
		<button onClick={ () => setId( 'test-1' ) }>
			Set ID to test-1
		</button>
		<button onClick={ () => setId( 'test-2' ) }>
			Set ID to test-2
		</button>
	</div>
);
```

puntope:
@eason9487  Thanks for the new approach. It definitely improves my previous implementation..

Only thing to change is the checked condition


https://user-images.githubusercontent.com/5908855/219398984-af8b384c-ba6d-4aca-b946-f8bc08eaec44.mov


At this moment it will show the tour when is resolving.

I guess it should be:

`const tourChecked = ! tour.hasFinishedResolution || Boolean( checked );`

puntope:
Also, I would omit the `upsertingClientStoreFirst` since we don't needed (it always true) and as far as I know there is no perspective of using that `upsertingClientStoreFirst = false.`. So for me is unnecessary complexity at this moment.

jorgemd24:
Thanks @eason9487, I tested using the new approach with two Tours and It works perfectly.

eason9487:
Thanks for your consideration, @puntope

> Only thing to change is the checked condition [...]
> I guess it should be:
> `const tourChecked = ! tour.hasFinishedResolution || Boolean( checked );`

Yes, it should be. 👍 Thanks for catching it.

I have added your condition fix to the `update/1879-use-tour-alternative` branch.

> Also, I would omit the `upsertingClientStoreFirst` since we don't needed (it always true) and as far as I know there is no perspective of using that `upsertingClientStoreFirst = false.`. So for me is unnecessary complexity at this moment.

The `upsertingClientStoreFirst` defaults to `false` because all wp-data actions that have the same processing of asking API to mutate server data and updating the wp-data store are following the same flow: Issues the API request first, and then notifying the reducer to update the wp-data store.

If the default flow of `upsertTour` is reversed, then it will become a very special case among actions at the same level of abstraction. This may reduce the complexity of the code implementation a bit, but also increases the complexity of consistency at the same level of abstraction.

However, understanding and maintaining consistency for the abstraction levels are usually more time-consuming.

For example, how files in a repo should be sorted and placed is also the management of abstraction levels, and GLA repo has clearly been highly inconsistent for a long time. Each time a new file is added, it takes longer and longer to think about where to place it, and it will take more and more time to reorganize the whole repo to improve consistency in the future.

The levels of abstraction for managing state and data have always been a relatively critical core of client apps. Therefore, I would recommend prioritizing managing the complexity of abstraction levels over the complexity of the code fragments.

---

If the solution elaborated in `update/1879-use-tour-alternative` works for you as well, I would like to continue to organize the solution and the following adjustments for this PR:
1. Adjust to adopt the solution in `update/1879-use-tour-alternative`.
2. Address Jorge's code review comments and request a new round of code reviews from him.
3. Merge once getting approval.
4. If it works for you as well, I would sync all changes made in this PR to [#1752](https://github.com/woocommerce/google-listings-and-ads/pull/1752). (Not including the needed adjustments that are out of `useTour` and wp-data store)

In this way, it could facilitate and smooth the subsequent merging process of #1880 and the shared breach of PMax Assets.

puntope:
> The `upsertingClientStoreFirst` defaults to `false` because all wp-data actions that have the same processing of asking API to mutate server data and updating the wp-data store are following the same flow: Issues the API request first, and then notifying the reducer to update the wp-data store.

Not sure if I follow on this. What I meant is that `upsertingClientStoreFirst` is only being used in useTour. There is no case of false in the code. So why not omit `upsertingClientStoreFirst` prop?

puntope:
> If the solution elaborated in update/1879-use-tour-alternative works for you as well, I would like to continue to organize the solution and the following adjustments for this PR:

For me all good the steps you mention. Works for me

eason9487:
Hi @puntope, I just realized that I didn't seem to fully understand the direction you meant.

> Not sure if I follow on this. What I meant is that `upsertingClientStoreFirst` is only being used in useTour. There is no case of false in the code. So why not omit `upsertingClientStoreFirst` prop?

Could you clarify more about what you meant by how was `upsertingClientStoreFirst` expected to be omitted from where?
1. Remove the default value `false`.
   ```diff
   - export function* upsertTour( tour, upsertingClientStoreFirst = false ) {
   + export function* upsertTour( tour, upsertingClientStoreFirst ) {
   
   ```
2. Reverse the default value `false` to `true` and omit it from the caller.
   ```diff
   - export function* upsertTour( tour, upsertingClientStoreFirst = false ) {
   + export function* upsertTour( tour, upsertingClientStoreFirst = true ) {
   
   - upsertTour( { id: tourId, checked: nextChecked }, true );
   + upsertTour( { id: tourId, checked: nextChecked } );
   ```
3. Remove `upsertingClientStoreFirst` completely and just make `upsertTour` update to wp-data store first.
   ```diff
   - upsertTour( { id: tourId, checked: nextChecked }, true );
   + upsertTour( { id: tourId, checked: nextChecked } );
   ```
   ```js
   export function* upsertTour( tour ) {
     try {
       yield {
         type: TYPES.UPSERT_TOUR,
         tour,
       };
   
       return apiFetch( {
         path: `${ API_NAMESPACE }/tours`,
         method: REQUEST_ACTIONS.POST,
         data: tour,
       } );
     } catch ( error ) {
       // ... omitted
     }
   ```
4. None of the above. Could you provide me a sample/pseudo code?

js/src/data/actions.js
eason9487: Replaced in https://github.com/woocommerce/google-listings-and-ads/pull/1879/commits/fce415882f4a25d7dcded71467405bd7767c30c7.

js/src/data/actions.js
eason9487: Replaced in https://github.com/woocommerce/google-listings-and-ads/pull/1879/commits/fce415882f4a25d7dcded71467405bd7767c30c7.

js/src/data/reducer.js
eason9487: Merged in https://github.com/woocommerce/google-listings-and-ads/pull/1879/commits/847542648871da5613aae830b03fdb233fe5b673.

js/src/hooks/useTour.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/1879/commits/526f81874efb1fd8bde2ecf08cd253d5a5a28de0.

js/src/hooks/useTour.js
eason9487: > `closeTour` is a function and `showTour` is a boolean, the first time that I read the code I thought that `closeTour` and `showTour` were both functions, can we change `showTour` for something like `isTourShown` or something like that?

These return values no longer exist after changing to use the alternative solution. Ref: https://github.com/woocommerce/google-listings-and-ads/pull/1879/commits/2aa7988850f990b2480251e42d8c111772bb71fe

js/src/data/resolvers.js
eason9487: No. Adjusted the wording in https://github.com/woocommerce/google-listings-and-ads/pull/1879/commits/f3d7c5b58bd367b277b36e91a9da559388cbc20e.

js/src/hooks/useTour.js
eason9487: This problem no longer exists after changing to use the alternative solution. Ref: https://github.com/woocommerce/google-listings-and-ads/commit/2aa7988850f990b2480251e42d8c111772bb71fe.

eason9487: This suggestion is no longer applicable after changing to use the alternative solution. Ref: https://github.com/woocommerce/google-listings-and-ads/commit/2aa7988850f990b2480251e42d8c111772bb71fe.

eason9487:
Hi @jorgemd24,

As per the proposed follow-up steps in https://github.com/woocommerce/google-listings-and-ads/pull/1879#issuecomment-1434094218, I have changed this PR to use the alternative solution and addressed all review comments. In addition, PR #1880 was updated accordingly. Could you help with a new round of code reviews?

puntope:
Hey @eason9487 

What I meant is the option 3.

eason9487:
Hi @puntope,

> > 3\. Remove `upsertingClientStoreFirst` completely and just make `upsertTour` update to wp-data store first.
>
> What I meant is the option 3.

> Not sure if I follow on this. What I meant is that `upsertingClientStoreFirst` is only being used in useTour. There is no case of false in the code. So why not omit `upsertingClientStoreFirst` prop?

Yes, there is no case of calling with `false` in the codebase. But the reasons of why I thought it should not be omitted and should default to `false` have already been explained in the first part of https://github.com/woocommerce/google-listings-and-ads/pull/1879#issuecomment-1434094218. Could you provide thoughts if those reasons didn't make any valid points, or if any reasons are unclear?

puntope:
> Could you provide thoughts if those reasons didn't make any valid points, or if any reasons are unclear?

In brief... I agree about that `upsertTour` acts differently than other functions, but thats a fact inherent to the requirements. Instead of adding that extra prop `upsertingClientStoreFirst` I would add just a useful comment explaining why. 

With `upsertingClientStoreFirst` 

- upsertTour has still different behaviour and still capable of updating first wp.data as opposite to other functions. 
- we have an extra condition block to deal with and an extra argument to send (increase of complexity) 
- That extra argument makes the dev (at least me) to think "Oh whats this argument for..., does it means we have a tour somewhere updating first the DB and then the wp.data, why, where?". 

Without `upsertingClientStoreFirst` and adding just a comment

- upsertTour has still different behaviour and still capable of updating first wp.data as opposite to other functions. 
- simpler code, no extra argument that could create problems or at least potential confusion.

jorgemd24:
Thanks, @eason9487 for addressing the comments and adding the tests. I tested it locally this PR and PR #1880 and it is working well.

Just to add my two cents about the above discussion, I think it is worth keeping the argument because it will allow us to be more flexible, leaving the decision to update first the store or not to the developer. Also, similar logic could be implemented in cases where the API request is not critical.

It is also true that removing the argument will reduce the complexity of that function however we could add a comment in the `upsertTour` function adding more details if needed, or put a link to this discussion so the developer has more context why it was done like that.

eason9487:
@puntope, thanks for the explanation.

> With `upsertingClientStoreFirst`
> 
> * upsertTour has still different behaviour and still capable of updating first wp.data as opposite to other functions.

Sorry, I'm not sure if this makes a valid reason. Given the premise that this requirement is handled within wp-data, `upsertTour` must be able to reverse the common process, which means it must have at least one different process from other actions. Because of this, the `upsertingClientStoreFirst` was designed to be the same process by default, which is what was mentioned before: maintaining consistency at the same level of abstraction.

> [...] Instead of adding that extra prop `upsertingClientStoreFirst` I would add just a useful comment explaining why.

> * we have an extra condition block to deal with and an extra argument to send (increase of complexity)

I agree that adding code comments can help explain the reason for the special use case.

However, even if the code complexity is reduced a little bit (one parameter and one condition block with 5 lines of code), the comments themselves are still part of the complexity.

Considering that `upsertingClientStoreFirst` involves at most 10 lines of code (and they are all very concise code statements), I don't think the difference in complexity is significant when compared to code comments.

> * That extra argument makes the dev (at least me) to think "Oh whats this argument for..., does it means we have a tour somewhere updating first the DB and then the wp.data, why, where?".

In fact, similar cases happen all the time. For example:

1. Why set defaults for `page` and `perPage`? Does it mean somewhere uses this hook without specifying them?
   https://github.com/woocommerce/google-listings-and-ads/blob/c94b6ab1df5abc182e09df62421b269ff33df1db/js/src/hooks/useMappingRules.js#L14
2. Why set defaults for `selected` and `allCategories`? Does it mean somewhere uses this hook and its internal function without specifying them?
   https://github.com/woocommerce/google-listings-and-ads/blob/7d74e7877482ce785524a0d1d087da9a0b436bc2/js/src/hooks/useCategories.js#L33
   https://github.com/woocommerce/google-listings-and-ads/blob/7d74e7877482ce785524a0d1d087da9a0b436bc2/js/src/hooks/useCategories.js#L77
3. Why set defaults for three of the props? Does it mean somewhere uses this component without specifying them?
   https://github.com/woocommerce/google-listings-and-ads/blob/aa86f0cc0c91e63c08e2e941c870e906c89341b8/js/src/attribute-mapping/attribute-mapping-category-control.js#L24-L29
4. Why set default for the two `sources`?
   https://github.com/woocommerce/google-listings-and-ads/blob/1c8116cea9b8976c6135be94a4ffed5b2dc4c09f/js/src/attribute-mapping/attribute-mapping-field-sources-control.js#L18-L19
   https://github.com/woocommerce/google-listings-and-ads/blob/6f9b309595e4f30036e8a1ddff2298cb3b923f4e/js/src/attribute-mapping/attribute-mapping-source-type-selector.js#L41-L42
5. Why return `status`? Does it mean somewhere needs the status at that time?
   https://github.com/woocommerce/google-listings-and-ads/blob/c94b6ab1df5abc182e09df62421b269ff33df1db/js/src/data/controls.js#L17-L29

All these codes were reviewed by me and I didn't think they are totally useless nor had a relatively strong opinion on whether to remove the defaults. As it only took a few seconds to consider it could at least be a precaution such as breaking the app into a blank screen or it's simple/common enough to keep staying there. In addition, this PR happened to use the `status` from `fetchWithHeaders`.

Otherwise, if it's still important for you to avoid extra thinking about these reasons, then I would recommend re-checking such cases to make the codebase align consistent with "no unnecessary things unless they are actually used somewhere in the runtime".

Back to the case of `upsertingClientStoreFirst`, its existence is even more specific and particular than the above examples. It exists to notify or prompt devs more explicitly to think about why it exists, and I originally thought its existence was easy to understand because dealing with abstraction is always present during programming, and it's like fundamental consciousness. Having had the discussion till here, I would be happy to add a code comment explaining why `upsertTour` has the `upsertingClientStoreFirst` parameter.

---

If this is still a blocker for you, do any of the following options work for you?

1. As mentioned above, add a code comment to `upsertTour` and also refer to the relevant discussion in this PR.
2. Change back to a previous version as the following and call `receiveTour( nextTour )` ahead of `upsertTour` in `useTour`.
   https://github.com/woocommerce/google-listings-and-ads/blob/c06bcda0fa9d8761b387607553fe94735e789eca/js/src/data/actions.js#L1141-L1152
3. Change to use example 3 in [this comment](https://github.com/woocommerce/google-listings-and-ads/pull/1879#issuecomment-1434414851) and also rename `upsertTour` to `upsertTourToClientStoreFirst`.
4. Maybe you could branch from this PR and create a PR to implement your idea.

puntope:
> If this is still a blocker for you, do any of the following options work for you?

For me this was never a blocker. The code works and as far as I remember I approved this PR 🤔 (I see I didn't)  Sorry if this  was confusing. Also seems that @jorgemd24 is also agree with your implementation. So I guess you can go ahead with this PR like it's. Even tho I disagree this is a healthy flexibility. (I don't see a chance in `upsertTours` that we need to use this `upsertTourToClientStoreFirst`). 

> In fact, similar cases happen all the time

In regards your last arguments about defaults. I don't see the similarity with the case. In our case we are adding a new property that we are not using. The `page` and `perPage` are necessary arguments, the defaults are just for safety. 

The key difference for me is:

- If you remove `page` and `perPage` you cannot make it work.
- If you remove `upsertTourToClientStoreFirst` you can make it work. Since it's always true.

puntope (approved):

eason9487:
Hi @puntope, thanks for clarifying it's never a blocker.

> Even tho I disagree this is a healthy flexibility. (I don't see a chance in `upsertTours` that we need to use this `upsertTourToClientStoreFirst`).

My primary thoughts were focusing on maintaining the consistency of abstraction levels, and the existence of `upsertTourToClientStoreFirst` is not about whether it will be used or not but whether it brings out the special thing within the action function.

The flexibility was a minor point as whether it would be actually flexible will depend on future requirements.

> In regards your last arguments about defaults. I don't see the similarity with the case. In our case we are adding a new property that we are not using. The `page` and `perPage` are necessary arguments, the defaults are just for safety. 

I would like to clarify my comment was not an argument, but an analogy. My statement was that if similar questions are raised, then perhaps those "default values" should not exist either.

Still, actually, I meant the "unused default value" of those parameters, not the parameters; And I meant they are similar, not the same, to bring up similar questions like:

- "Oh, what's this default value for?"
- "Does it mean somewhere uses this default value?"
- "Why? Where?"

At least I did check if these default values are actually used somewhere while doing code reviews but didn't find them. And as I said, they were/are OK to me.

PR #1880, opened 2023-02-10
PMax Assets: Add a tour to the Dashboard page for introducing the campaign assets

eason9487:

This PR implements 📌 [Programs table on the Dashboard page](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-dashboard-programs) in #1787.

- Add a tour to the Dashboard page for introducing this new feature.

:warning: Steps to merge this PR (@eason9487):

- Wait for #1879 being merged onto `develop`
- Merge `develop` onto `feature/pmax-assets`
- Rebase this PR onto `feature/pmax-assets` and change the merge base to `feature/pmax-assets`
- Merge this PR

### Screenshots:

https://user-images.githubusercontent.com/17420811/218076585-3868c312-ca34-46aa-9c2c-96c8cf940072.mp4

💡 Please note that the UI layout and style are not that similar to the design in Figma because the `TourKit` component imported from `@woocommerce/components` doesn't open up the custom renderer, and I think it should be fine leaving this tour aligning with the default layout and style as possible.

1. Go to the Dashboard.
2. If there is at least one campaign, the tour should be shown on the top of the edit button of the first one.
3. Click the X button on the tour, it should be closed after a moment.
4. Refresh the page.
5. The tour should not be shown.
6. To retry, the tour flag can be reset by running the following code via the Console panel in the DevTool:
   ```js
   wp.data.dispatch( 'wc/gla' ).upsertTour( { id: 'dashboard-feature--campaign-assets', checked: false } )
   ``` 

### Changelog entry

>

### DIFF
--- a/js/src/css/shared/_gutenberg-components.scss
+++ b/js/src/css/shared/_gutenberg-components.scss
@@ -41,6 +41,16 @@
 			}
 		}
 	}
+
+	// Fix the obvious styles conflicts caused by WooCommerce Payments since approximately version 2.2.0.
+	// Ref: https://github.com/Automattic/woocommerce-payments/blob/2.2.0/client/payment-details/summary/style.scss#L3-L9
+	.components-card {
+		margin-bottom: 0;
+
+		&__header {
+			font-size: inherit;
+		}
+	}
 }
 
 // hack to fix radio button selected style bug caused by woocommerce-admin.
--- a/js/src/dashboard/all-programs-table-card/campaign-assets-tour.js
+++ b/js/src/dashboard/all-programs-table-card/campaign-assets-tour.js
@@ -1,0 +1,79 @@
+/**
+ * External dependencies
+ */
+import { __, _x } from '@wordpress/i18n';
+import { TourKit, Pill } from '@woocommerce/components';
+
+/**
+ * Internal dependencies
+ */
+import useTour from '.~/hooks/useTour';
+import './campaign-assets-tour.scss';
+
+const TOUR_ID = 'dashboard-feature--campaign-assets';
+
+/**
+ * Renders the tour for notifying the new feature of campaign assets
+ * if its flag is not yet set to hidden.
+ *
+ * @param {Object} props React props
+ * @param {string} props.referenceElementCssSelector The CSS selector to find the first DOM to render this tour nearby.
+ */
+export default function CampaignAssetsTour( { referenceElementCssSelector } ) {
+	const { showTour, setTourChecked } = useTour( TOUR_ID );
+
+	if ( ! showTour ) {
+		return null;
+	}
+
+	const config = {
+		steps: [
+			{
+				referenceElements: {
+					desktop: referenceElementCssSelector,
+				},
+				meta: {
+					heading: (
+						<>
+							{ __(
+								'🎉 Add creative assets',
+								'google-listings-and-ads'
+							) }
+							<Pill>
+								{ _x(
+									'New',
+									'A highlighting label behind the heading of the new feature',
+									'google-listings-and-ads'
+								) }
+							</Pill>
+						</>
+					),
+					descriptions: {
+						desktop: (
+							<>
+								{ __(
+									'Boost your ads performance by adding creative assets to your campaign.',
+									'google-listings-and-ads'
+								) }
+								<br />
+								<br />
+								{ __(
+									'Edit your campaign to explore this new feature.',
+									'google-listings-and-ads'
+								) }
+							</>
+						),
+					},
+				},
+			},
+		],
+		options: {
+			classNames: 'gla-campaign-assets-tour',
+			effects: { overlay: false },
+		},
+		placement: 'top',
+		closeHandler: () => setTourChecked( true ),
+	};
+
+	return <TourKit config={ config } />;
+}
--- a/js/src/dashboard/all-programs-table-card/campaign-assets-tour.js
+++ b/js/src/dashboard/all-programs-table-card/campaign-assets-tour.js
@@ -3,6 +3,7 @@
  */
 import { __, _x } from '@wordpress/i18n';
 import { TourKit, Pill } from '@woocommerce/components';
+import GridiconTrending from 'gridicons/dist/trending';
 
 /**
  * Internal dependencies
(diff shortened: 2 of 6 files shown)


jorgemd24 (commented):
Thanks, @eason9487 for working on this.

> Please note that the UI layout and style are not that similar to the design in Figma because the TourKit component imported from @woocommerce/components doesn't open up the custom renderer, and I think it should be fine leaving this tour aligning with the default layout and style as possible.

Is it needed to keep this margin at the bottom?

image

> Click the X button on the tour, it should be closed after a moment.

Do we really need to wait until the tour is updated to close the Tour? If I click the button to close the tour, I  would expect the pop-up to close immediately. Also in case the updating fails, there is no way to dismiss the Tour.


https://user-images.githubusercontent.com/2488994/218760407-78784dce-9ab8-4c77-a0f6-0cabf41d2978.mp4

eason9487:
@jorgemd24, thanks for the review.

> Is it needed to keep this margin at the bottom?

Thanks for finding this. It shouldn't have that margin. But I couldn't reproduce the problematic layout. Could you provide the versions of WP and WC while testing?

> Do we really need to wait until the tour is updated to close the Tour? If I click the button to close the tour, I would expect the pop-up to close immediately. Also in case the updating fails, there is no way to dismiss the Tour.

I don't think it's really needed and agree that it would be a bit nicer to close the popup immediately.

However, it's a bit out of this PR's scope. It would be more suitable to move this discussion to #1879 and loop in @puntope, the author of the shared codes picked from #1752, since the processing of updating the `showTour` state was added in #1879.

jorgemd24:
> Thanks for finding this. It shouldn't have that margin. But I couldn't reproduce the problematic layout. Could you provide the versions of WP and WC while testing?

It seems that is related to a plugin conflict, this time the conflict is with WC Payments ( I used the latest version 5.4.0). I think this line could be the reason: https://github.com/Automattic/woocommerce-payments/blob/b66af093d123bd7a285f7ac17b4f64e472fc7c26/client/payment-details/summary/style.scss#L8

I used WP: 6.1.1 and WC: 7.3.0

Can we overwrite that style only for the GL&A plugin? 

> I don't think it's really needed and agree that it would be a bit nicer to close the popup immediately.

> However, it's a bit out of this PR's scope. It would be more suitable to move this discussion to https://github.com/woocommerce/google-listings-and-ads/pull/1879 and loop in @puntope, the author of the shared codes picked from https://github.com/woocommerce/google-listings-and-ads/pull/1752, since the processing of updating the showTour state was added in https://github.com/woocommerce/google-listings-and-ads/pull/1879.

Thanks for your comments, I will move this discussion there.

jorgemd24 (approved):
@eason9487, I moved the discussion here: https://github.com/woocommerce/google-listings-and-ads/pull/1879#issuecomment-1431244337

The Tour is working, the only two things are:
-  The styling (that is a conflict issue with WC Payments), let me know if you need help reproducing this. 
-  the process to update the tour (we moved the discussion). 

The rest functionality is working as expected  ✅

eason9487:
@jorgemd24, thanks for providing more info and even accurately locating the root cause!

> > Thanks for finding this. It shouldn't have that margin. But I couldn't reproduce the problematic layout. Could you provide the versions of WP and WC while testing?
> 
> It seems that is related to a plugin conflict, this time the conflict is with WC Payments ( I used the latest version 5.4.0). I think this line could be the reason: https://github.com/Automattic/woocommerce-payments/blob/b66af093d123bd7a285f7ac17b4f64e472fc7c26/client/payment-details/summary/style.scss#L8
> 
> I used WP: 6.1.1 and WC: 7.3.0
> 
> Can we overwrite that style only for the GL&A plugin?

Yes, the conflict has been fixed in https://github.com/woocommerce/google-listings-and-ads/pull/1880/commits/0be0fff4288e081a2ae117f413467a0ff212e987.

jorgemd24:
@eason9487 Thanks again! It is fixed now 👍

eason9487:
Appended https://github.com/woocommerce/google-listings-and-ads/pull/1880/commits/8a7f785fd19845331e124e8e23b590ab63844fa7 for updating the copy. Ref: fqR0EHi63lWahRcVTKCcba-fi-5299%3A224027

image

jorgemd24:
Thanks for updating this. Just a slight difference, not sure if it is a problem though.

In Figma, the title and the `X` are aligned however in the PR they are not. See:

image

image

eason9487:
Hi @jorgemd24

> Just a slight difference, not sure if it is a problem though.
> 
> In Figma, the title and the `X` are aligned however in the PR they are not.

The layout and style are not that similar to the design in Figma because the `TourKit` component imported from `@woocommerce/components` doesn't open up the custom renderer. Ref:

- `@automattic/tour-kit` offers [the `config.renderers`](https://github.com/Automattic/wp-calypso/blob/%40automattic/tour-kit%401.1.0/packages/tour-kit/README.md#usage) for library users to custom the rendering.
- However, [the `TourKit` component of `@woocommerce/components` was written in such a way](https://github.com/woocommerce/woocommerce/blob/7.1.0/packages/js/components/src/tour-kit/index.tsx#L41-L43) that the library users must use only its built-in renderer.

Even if it's still possible to override the default styles with GLA's CSS, it's sometimes getting unstable with version iterations. So in GLA only heavily used or important UI may purposely override the styles, while the rest is kept the default styles of the core components.

Further, the different point you mentioned will further involve the need to hack the rendered HTML structure by CSS or even JS in order to adjust it.

Therefore, I think it should be fine leaving this tour aligning with the default layout and style.

jorgemd24:
Ok, thanks for your explanation 🙇‍♂️

eason9487:
As mentioned in PR description.

> ⚠️ Steps to merge this PR:
> 
> * ✅ Wait for [Add the tours to wp-data and the `useTour` hook #1879](https://github.com/woocommerce/google-listings-and-ads/pull/1879) being merged onto `develop`
> * ✅ Merge `develop` onto `feature/pmax-assets`
> * ✅ Rebase this PR onto `feature/pmax-assets` and change the merge base to `feature/pmax-assets`
> * Merge this PR

The first three steps were processed. I'll be merging this PR.

PR #1886, opened 2023-02-14
Remove unnecessary filter param from woocommerce_loop_add_to_cart_link

### LINKED ISSUE #8020: Apply filter woocommerce_loop_add_to_cart_link is called with 2 params instead of 3
## Describe the bug

While we were testing Google Listings & Ads and the Products Block (Beta) a fatal error was displayed, the issue is caused because the filter `woocommerce_loop_add_to_cart_link` is applied with two parameters instead of three.

See how WooCommerce applies the filter: https://github.com/woocommerce/woocommerce/blob/7.2.2/plugins/woocommerce/templates/loop/add-to-cart.php#L35

See WooCommerce Blocks: 
https://github.com/woocommerce/woocommerce-blocks/blob/ffe341c2b77b9d55dddc5908a6d8317f0ba03244/src/BlockTypes/ProductButton.php#L111

See how GL&A use the filter: https://github.com/woocommerce/google-listings-and-ads/blob/d4b2dfd8751ebadb096f2a974de0202d08c048d0/src/Google/GlobalSiteTag.php#L153

Fatal error:

`Uncaught ArgumentCountError: Too few arguments to function Automattic\WooCommerce\GoogleListingsAndAds\Google\GlobalSiteTag::Automattic\WooCommerce\GoogleListingsAndAds\Google\{closure}(), 2 passed in /var/www/html/wp-includes/class-wp-hook.php on line 308 and exactly 3 expected`

The filter should be aligned with how WooCommerce applies the filter. 

## To reproduce

1. Install Google Listings & Ads plugin.
2. Create a page with the Products Block (Beta)

image

5. See error: `Uncaught ArgumentCountError: Too few arguments to function Automattic\WooCommerce\GoogleListingsAndAds\Google\GlobalSiteTag::Automattic\WooCommerce\GoogleListingsAndAds\Google\{closure}(), 2 passed in /var/www/html/wp-includes/class-wp-hook.php on line 308 and exactly 3 expected`

## Expected behavior

Not fatal errors.

Aljullu:

Closes https://github.com/woocommerce/woocommerce-blocks/issues/8020. (Heads-up that issue will also be closed by https://github.com/woocommerce/woocommerce-blocks/pull/8422, but I think it makes sense to improve that part of the codebase in GL&A too just to be extra-cautious)

When hooking into `woocommerce_loop_add_to_cart_link`, GL&A was declaring the three params of the function, but in fact, it was only using the second one (`$product`).

In WC Blocks, the Products was incorrectly only passing two filter arguments, so that caused an issue because GL&A expects it to have three. While this will be fixed in WC Blocks as mentioned, it's also a good idea to only declare the params which are actually used. This PR removes the `$args` param which was unused.

1. With a block theme (ie: TT3), go to Appearance > Editor and edit the Product Catalog template.
2. Remove the default placeholder block and in its place add the _Product (Beta)_ block.
3. In the frontend, load the Shop page.
4. Verify that the script `gla-gtag-events` and the variable `glaGtagData` contain a list of products with their name and price.

See video:

https://user-images.githubusercontent.com/2488994/218812249-ac16566e-7232-4318-ac08-0f68a40e2b49.mov

### Changelog entry

> Tweak - Remove unnecessary woocommerce_loop_add_to_cart_link filter param

### DIFF
--- a/src/Google/GlobalSiteTag.php
+++ b/src/Google/GlobalSiteTag.php
@@ -150,12 +150,12 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 		// Add product data for any add_to_cart link.
 		add_filter(
 			'woocommerce_loop_add_to_cart_link',
-			function ( $link, $product, $args ) {
+			function ( $link, $product ) {
 				$this->add_product_data( $product );
 				return $link;
 			},
 			10,
-			3
+			2
 		);
 
 		// Add display name for an available variation.


jorgemd24:
Thanks, @Aljullu for this PR!

It could be tested by loading the shop page and checking that the script `gla-gtag-events` and the variable `glaGtagData` contain a list of products with their name and price.

See video:

https://user-images.githubusercontent.com/2488994/218812249-ac16566e-7232-4318-ac08-0f68a40e2b49.mov

.

We had a discussion about this subject previously, see this comment: https://github.com/woocommerce/google-listings-and-ads/issues/1630#issuecomment-1240293554 and we think that we should encourage the developers to apply the filter correctly as it's run in WooCommerce.

On other hand, you are right that this PR will help to prevent this issue in cases where the filter is called incorrectly so I am happy to go ahead with this PR too.

WDYT?

Aljullu:
> It could be tested by loading the shop page and checking that the script `gla-gtag-events` and the variable `glaGtagData` contain a list of products with their name and price.

Thanks, @jorgemd24! I updated the PR with your testing steps.

> We had a discussion about this subject previously, see this comment: https://github.com/woocommerce/google-listings-and-ads/issues/1630#issuecomment-1240293554 and we think that we should encourage the developers to apply the filter correctly as it's run in WooCommerce.

I definitely agree that this needs to be fixed in WC Blocks. But I think it's also a good practice to only declare the filter arguments which are actually used, instead of declaring all of them. I don't have a strong opinion, though, so happy to close this PR if needed.

jorgemd24 (approved):
Thanks, @Aljullu , it working as expected. LGTM 👍 

Do you have access rights to merge it or would you like me to do it?

Aljullu:
> Do you have access rights to merge it or would you like me to do it?

I don't have access, so feel free to go ahead and merge yourself. Thanks for the review! :slightly_smiling_face:

PR #1798, opened 2022-12-01
Fix certain inbox notifications were shown before setup completed

ianlin:
### Changes proposed in this Pull Request:

Project: pcTzPl-1bL-p2

This PR fixes two inbox notifications that appear before GLA setup completed.

- **Reconnect WordPress Notification**

This notification is sent when we detect Jetpack connection is not ready and prompt them to reconnect their WordPress account. However it should only be shown when the user has gone through the onboarding and complete the setup.

- **Before / After Campaign Migration Notification**

This two notifications were implemented when we were trying to migrate from Smart Shopping Campaigns to Performance Max Campaigns. This happened in July to September 2022 so the notifications are not necessary to show now.

Closes #1620.
Closes #1756.

### Detailed test instructions:

**Reconnect WordPress Notification**

1. Disconnect all accounts from GLA's settings page.
2. If you have other plugins using Jetpack, disconnect the Jetpack connection from MyJetpack manually.
3. Remove the notifications that were added in the database already
    ```sql
    DELETE FROM wp_wc_admin_notes WHERE name = 'gla-reconnect-wordpress';
    ```
4. Run the Action Scheduler task `wc_gla_cron_daily_notes` manually.
    image
5. The notification `gla-reconnect-wordpress` is not added to the database table `wp_wc_admin_notes`.
6. Go to `WooCommerce > Home` and check the inbox message, the Reconnect WordPress notification is not shown.
7. Go through the GLA onboarding and complete the setup.
8. Disconnect Jetpack from connection test page: `/wp-admin/admin.php?page=connection-test-admin-page`.
9. Run the Action Scheduler task `wc_gla_cron_daily_notes` manually again.
10. The notification `gla-reconnect-wordpress` is added to the database table `wp_wc_admin_notes`.
11. Go to `WooCommerce > Home` and check the inbox message, the Reconnect WordPress notification is shown.
12. Reconnect WordPress account from GLA's settings page.
13. Run the Action Scheduler task `wc_gla_cron_daily_notes` manually again.
14. The notification is not shown in `WooCommerce -> Home`.

**Before / After Campaign Migration Notification**

1. Remove both notifications from the database
    ```sql
    DELETE FROM wp_wc_admin_notes WHERE name = 'gla-before-campaign-migration';
    DELETE FROM wp_wc_admin_notes WHERE name = 'gla-after-campaign-migration';
    ```
2. Disconnect all accounts from GLA's settings page.
3. Run the Action Scheduler task `wc_gla_cron_daily_notes` manually.
[...]

PR #1889, opened 2023-02-15
Remove PMax Migration Banners

jorgemd24:

PMax Migration banners were still showing even though the migration ended in September last year. 

This PR removes the PMax migration banner and it is aligned with the following PR https://github.com/woocommerce/google-listings-and-ads/pull/1798 that removed the inbox notifications about the migration. 

#### Dashboard
image

#### Reports
image

1. Go to the dashboard page, and the banner should not be displayed.
2. Go to the reports page, and the banner should not be displayed.

If you would like to test that banners were showing in develop branch follow these steps:

 1. Remove the following items from the local storage (in case they were dismissed previously): 
-  `localStorage.removeItem('gla-is-before-migration-notice-dismissed')`
-  `localStorage.removeItem('gla-is-after-migration-notice-dismissed')`
- `localStorage.removeItem('gla-is-reports-migration-notice-dismissed')`
2. Go to the dashboard or the reports page and the banner should be displayed. 

### Changelog entry

> Tweak - Remove unnecessary PMax migration banners


puntope:
👀  Are we using `js/src/utils/getConversionCampaignStatusNotice.js`?  if not we can consider remove it and the attached tests.

puntope:
If we end up deleting `js/src/utils/getConversionCampaignStatusNotice.js` as suggested here https://github.com/woocommerce/google-listings-and-ads/pull/1889#issuecomment-1432792151

Also we might remove `CAMPAIGN_ADS_MIGRATION` constant.

puntope (commented):
@jorgemd24 Thanks for working on this.

I don't see anymore the notice. However, there are some residual files and a constant we might consider for deletion. (check comments)

jorgemd24:
Thanks @puntope for catching this.

At this point, all campaigns should be converted so I think it is not necessary to keep the hook.

I removed the hook and the constant.

puntope (approved):
✅  LGTM 

Good job

PR #1894, opened 2023-02-20
PMax Assets: Change the max number of image assets to be the shared max across the same group

eason9487:

This PR implements a part of Assets form in 📌 [Assets form and related components](https://github.com/woocommerce/google-listings-and-ads/issues/1787#fe-misc) in #1787.

This PR changes the implementation to share the maximum number of image assets across all types of **marketing** image assets.
- Add a new prop `props.reachedMaxNumberTip` to ImagesSelector component.
- Change the default of `maxNumberOfImages` prop to -1 for ImagesSelector since 0 will be a valid max value.
- Change the max number of image assets to be the shared number across the same group of types.
- Add the tip to the add button in the asset images selector when the selected number of images reaches the shared max.
- Replace the hard-coded number in `getImageSharedMaxHelpContent` with the actual shared max value.

### Screenshots:

image

https://user-images.githubusercontent.com/17420811/220082073-9fcc422c-3819-440b-a4a7-2bd0dd942b52.mp4

💡  Considering the max limit of 20 is not that easy to test, it would be easier by changing the number in the following line to a smaller one:

https://github.com/woocommerce/google-listings-and-ads/blob/3ca5e71b58a863bfab520bd8143493ad3709374e/js/src/components/paid-ads/assetSpecs.js#L147

2. Manage the assets of "Landscape", "Square" and "Portrait" images to see if the shared max number work correctly.
3. When reaching the max number, it should show a tip floating on the "Add image" button after hovering on the disabled button.

### Additional details:

Detailed context:
- Main discussion thread: pcTzPl-12o-p2#comment-2268
- Adopted solution: pcTzPl-12o-p2#comment-2281

### DIFF
--- a/js/src/components/paid-ads/assetSpecs.js
+++ b/js/src/components/paid-ads/assetSpecs.js
@@ -9,6 +9,12 @@ import { Fragment, createInterpolateElement } from '@wordpress/element';
  */
 import { ASSET_FORM_KEY } from '.~/constants';
 
+/**
+ * @typedef {import('.~/components/types.js').AssetGroupFormValues} AssetGroupFormValues
+ */
+
+const sharedMaxSymbol = Symbol( 'sharedMax' );
+
 const ASSET_DISPLAY_URL_PATH_SPECS = [
 	{
 		maxCharacterCount: 15,
@@ -32,7 +38,6 @@ const ASSET_MARKETING_IMAGE_SPECS = [
 	{
 		key: ASSET_FORM_KEY.MARKETING_IMAGE,
 		min: 1,
-		max: 8,
 		imageConfig: {
 			minWidth: 600,
 			minHeight: 314,
@@ -58,7 +63,6 @@ const ASSET_MARKETING_IMAGE_SPECS = [
 	{
 		key: ASSET_FORM_KEY.SQUARE_MARKETING_IMAGE,
 		min: 1,
-		max: 8,
 		imageConfig: {
 			minWidth: 300,
 			minHeight: 300,
@@ -84,7 +88,6 @@ const ASSET_MARKETING_IMAGE_SPECS = [
 	{
 		key: ASSET_FORM_KEY.PORTRAIT_MARKETING_IMAGE,
 		min: 0,
-		max: 4,
 		imageConfig: {
 			minWidth: 480,
 			minHeight: 600,
@@ -139,6 +142,11 @@ const ASSET_LOGO_SPECS = [
 ];
 
 // The max number of asset images is shared across different types of asset images.
+// Therefore, the shared max numbers are set to each group array via the Symbol,
+// and the groups are shown in the structure of the below `ASSET_IMAGE_SPECS_GROUPS`.
+ASSET_MARKETING_IMAGE_SPECS[ sharedMaxSymbol ] = 20;
+ASSET_LOGO_SPECS[ sharedMaxSymbol ] = 5;
+
 const ASSET_IMAGE_SPECS_GROUPS = [
 	ASSET_MARKETING_IMAGE_SPECS,
 	ASSET_LOGO_SPECS,
@@ -371,17 +379,40 @@ const ASSET_TEXT_SPECS = [
 		);
 	}
 
+	/**
+	 * Gets the current maximum number of this type of asset images.
+	 *
+	 * @param {number} sharedMax The shared maximum number of the grouped types of image assets.
+	 * @param {Object[]} specs The image specs in the same group.
+	 * @param {AssetGroupFormValues} values The assets form values to be calculated for the occupied number of asset images.
+	 * @return {number} The current maximum number of this type of asset images.
+	 */
+	function getMax( sharedMax, specs, values ) {
+		return specs.reduce( ( remaining, spec ) => {
+			if ( spec.key === this.key ) {
+				return remaining;
+			}
+
+			const occupied = Math.max( spec.min, values[ spec.key ].length );
+			return remaining - occupied;
+		}, sharedMax );
+	}
+
 	ASSET_IMAGE_SPECS_GROUPS.forEach( ( specs ) => {
 		// Currently, the PMax Assets feature in this extension doesn't offer managing the landscape_logo
 		// asset but only the logo asset. So the logo asset shares the total number of images by itself.
 		// To avoid confusing extension users, the UI and wording are shown the shared max concept when
 		// the number of manageable images in the same group is > 1.
 		const shownAsSharedMax = specs.length > 1;
+		const sharedMax = specs[ sharedMaxSymbol ];
+
 		const help = getImageHelpContent( specs, shownAsSharedMax );
 
 		specs.forEach( ( spec ) => {
 			spec.subheading = getSubheading( spec, shownAsSharedMax );
 			spec.help = help;
+
+			spec.getMax = getMax.bind( spec, sharedMax, specs );
 		} );
 	} );
 
--- a/js/src/components/paid-ads/assetSpecs.js
+++ b/js/src/components/paid-ads/assetSpecs.js
@@ -116,7 +116,6 @@ const ASSET_LOGO_SPECS = [
 	{
 		key: ASSET_FORM_KEY.LOGO,
 		min: 1,
-		max: 5,
 		imageConfig: {
 			minWidth: 128,
 			minHeight: 128,
@@ -417,6 +416,12 @@ const ASSET_TEXT_SPECS = [
 		);
 
 		specs.forEach( ( spec ) => {
+			// Currently, the logo asset is not shown as shared max on UI, so it still needs to
+			// set `spec.max` for generating its subheading.
+			if ( ! shownAsSharedMax && Number.isInteger( sharedMax ) ) {
+				spec.max = sharedMax;
+			}
+
 			spec.subheading = getSubheading( spec, shownAsSharedMax );
 			spec.help = help;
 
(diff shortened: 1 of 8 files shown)


puntope:
https://user-images.githubusercontent.com/5908855/220559751-dcf20af3-5e24-4e44-b75d-34cedea8915b.mov

Not sure if this is the targeted behaviour but I expected the tips to disappear.  Didn't go deeper on it, but seems for me that when the mouse goes down the tip doesn't disappear correctly. 

Update: I see you're using `<AppTooltip>` so maybe there is an issue there. Not sure.

puntope:
This is more an UX issue I'd say...but there we go.

- Lets have a limit of 4  in ASSET_MARKETING_IMAGE_SPECS
- You have 1 pic in Portrait, 2 in landscape and 1 in square. (so total 4)
- Buttons are disabled and tip says "You have reached the maximum of 4 total ad images. Please remove some images to continue uploading." Ok. Makes sense
- Then I remove 1 image in Portrait. The button in Portrait is enabled again. Ok.
- However, the buttons in Landscape and Square are still disabled. I guess because a minimum of one Portrait is required so doesn't makes sense to enable Landscape and Square buttons. Thats ok still
- The problem for me is the message in the tip is  still "You have reached the maximum of 4 total ad images. Please remove some images to continue uploading.". Which is not true since I have 3 images. 

Not sure if my point is clear. I hope it's. Otherwise I'd try to rephrase it a bit. 

P.S. Maybe a quick fix would be to add. "Notice that this maximum includes the required minimum images in your assets." Or something like that maybe better written by UX team. I think you can get the point.

puntope:
Another issue I see:

- Let's have a limit of 2 for ASSET_LOGO_SPECS
- The limit works but the copy in Logo shows "Up to 5"

Screenshot 2023-02-22 at 12 19 08

js/src/components/paid-ads/assetSpecs.js
```diff
@@ -9,6 +9,12 @@ import { Fragment, createInterpolateElement } from '@wordpress/element';
  */
 import { ASSET_FORM_KEY } from '.~/constants';
 
+/**
+ * @typedef {import('.~/components/types.js').AssetGroupFormValues} AssetGroupFormValues
+ */
+
+const sharedMaxSymbol = Symbol( 'sharedMax' );
```
puntope, line 16: ❓  Whats the advantage of the symbol here?

puntope: I see you mentioned something [here](https://github.com/woocommerce/google-listings-and-ads/pull/1894/files#diff-151d3ef49a9966c58a144e85b30c5fcea9eaa4f6d93ff302c7031c89a5d05dc2R145)

But not sure yet why we cannot use a traditional   ` const SHARED_MAX_KEY = 'SHARED_MAX_KEY'`

eason9487: Generally, one of `Symbol` uses is hidden from any mechanisms other code will typically use to access the object. That explicitly indicates a form of weak encapsulation or a weak form of information hiding.

Here the concept and implementation of the shared max numbers are encapsulated within the `assetSpecs` module, and they are not expected to be accessed directly from outside as the exposed constant is `ASSET_IMAGE_SPECS` and the overall of its array items don't have the same context of the shared max numbers.

Thus, the `Symbol( 'sharedMax' )` is a good fit.

eason9487:

> Not sure if this is the targeted behaviour but I expected the tips to disappear. Didn't go deeper on it, but seems for me that when the mouse goes down the tip doesn't disappear correctly.
> 
> Update: I see you're using `<AppTooltip>` so maybe there is an issue there. Not sure.

`Tooltip` component [won't handle the `onMouseLeave` event for a disabled child](https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4019.17.0/packages/components/src/tooltip/index.js#L193-L201) to hide the popover. Although `AppTooltip` wraps a div as the passed child to `Tooltip` and `onMouseLeave` handler can be bound to the wrapper div, a disabled button still won't be triggered the `onMouseLeave` event.

I guess it could be better by workarounds in GLA:
1. Make the wrapper div have a bigger floating placeholder so that the bound `onMouseLeave` handler will have a bigger chance to be triggered.
2. Add another event handler to `window` and remove the whole `<AppTooltip>` when it fulfills somewhat state.

But it probably won't be handled in this PR, or even it would be nicer to be addressed from upstream.

---

> Another issue I see:
> 
> * Let's have a limit of 2 for ASSET_LOGO_SPECS
> * The limit works but the copy in Logo shows "Up to 5"

From the perspective of the `getSubheading` function, it only knows the concept of `shownAsSharedMax`, and all things about max and shared max should still work well if someday the `landscape_logo` asset is added to `ASSET_LOGO_SPECS`.

Nevertheless, it will be nicer to make it a single source of truth and also address the potential problem you found. Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/1894/commits/f523303887c644e0276dcdeb869e3c4c6ba44ec1.

---

> This is more an UX issue I'd say...but there we go.
> 
> * Lets have a limit of 4  in ASSET_MARKETING_IMAGE_SPECS
> * [...]
> * Then I remove 1 image in Portrait. The button in Portrait is enabled again. Ok.
> * However, the buttons in Landscape and Square are still disabled. I guess because a minimum of one Portrait is required so doesn't makes sense to enable Landscape and Square buttons. Thats ok still
> * The problem for me is the message in the tip is  still "You have reached the maximum of 4 total ad images. Please remove some images to continue uploading.". Which is not true since I have 3 images.
> [...]
> P.S. Maybe a quick fix would be to add. "Notice that this maximum includes the required minimum images in your assets." Or something like that maybe better written by UX team.

I couldn't reproduce the case with the same steps but I guess you were referring to the following situation, right?


This is a good catch, thanks! I will forward it to the designer to see if they want to adjust the wording.

puntope (approved):
✅  LGTM

- Pending to verify the copy with UX
- Pending to improve the AppTooltip

MatthiasReinholz:
> > This is more an UX issue I'd say...but there we go.
> > 
> > * Lets have a limit of 4  in ASSET_MARKETING_IMAGE_SPECS
> > * [...]
> > * Then I remove 1 image in Portrait. The button in Portrait is enabled again. Ok.
> > * However, the buttons in Landscape and Square are still disabled. I guess because a minimum of one Portrait is required so doesn't makes sense to enable Landscape and Square buttons. Thats ok still
> > * The problem for me is the message in the tip is  still "You have reached the maximum of 4 total ad images. Please remove some images to continue uploading.". Which is not true since I have 3 images.
> >   [...]
> >   P.S. Maybe a quick fix would be to add. "Notice that this maximum includes the required minimum images in your assets." Or something like that maybe better written by UX team.
> 
> I couldn't reproduce the case with the same steps but I guess you were referring to the following situation, right? 
> 
> This is a good catch, thanks! I will forward it to the designer to see if they want to adjust the wording.

@eason9487 I just spoke with Nikki about this issue.
What about this solution:
- portrait images upload is disabled until at least one landscape + one square image is uploaded
- the tooltip for the portrait images upload button mentions that you can only upload a portrait image once you have uploaded at least one landscape + one square image

eason9487:
Hi @MatthiasReinholz, thanks for the idea.

> I just spoke with Nikki about this issue. What about this solution:
> 
> * portrait images upload is disabled until at least one landscape + one square image is uploaded
> * the tooltip for the portrait images upload button mentions that you can only upload a portrait image once you have uploaded at least one landscape + one square image

Unfortunately, this approach probably won't go well if I understood it right.

Take the same limit of total 4 images as premise and proceed with the possible steps:
1. Portrait button is disabled
1. Select 2 landscape images
1. Select 1 square image
1. Portrait button is enabled
1. Select 1 portrait image
1. All buttons are disabled
1. Unselect 1 square image

At this moment, the selected images are 2 landscape and 1 portrait. What will be the next action to be done via the code?
1. Enable all buttons
   - It allows to select invalid combinations like 2 landscape and 2 portrait images
   - After trying to submit, the user is told they did the wrong selection
   - It may frustrating the user as the UI allows they select the invalid combinations and then unfairly claims they did the wrong thing
2. Enable landscape and square buttons and also unselect 1 portrait image
   - Auto-unselect may surprise the user
   - It may frustrating the user as later they need to select again the portrait image they had previously selected
3. Enable square button
   - This goes into the same state as the screenshot in [this comment](https://github.com/woocommerce/google-listings-and-ads/pull/1894#issuecomment-1439874287)

---

Another possible steps:
1. Portrait button is disabled
1. Select 3 landscape images

At this moment, what will be the next action to be done via the code?
1. No action
   - It allows to select the invalid combination - 4 landscape images
2. Disable square button
   - This goes into a similar state as the screenshot in [this comment](https://github.com/woocommerce/google-listings-and-ads/pull/1894#issuecomment-1439874287)

---

The primary thing that might confuse users is that the total number of image in the popover of the disabled buttons may not be the same as the actual total number selected. Considering that the possibility of users entering this state is basically low, even after they entered, there is only an extremely low chance that they don't know how to continue to complete the assets form. Because
- There is the text "At least 1 required" on the square field
- After try to submit, the page will scrolls the user to the position of the square field, and there are a "1 issue" label and a "Add at least 1 square image" error message on it
- The user can still change to any other combinations
- The current implementation won't let the user select an invalid combination

Therefore, based on the current implementation, unless there is still time for performing a larger scope of UI redesign, I would still recommend adjusting the wording in the popover of the disabled buttons.

MatthiasReinholz:
> Therefore, based on the current implementation, unless there is still time for performing a larger scope of UI redesign, I would still recommend adjusting the wording in the popover of the disabled buttons.

@eason9487 thanks for the explanations!

Can you make the tooltip show text that will be different based on the condition?
As long as there is not at least one landscape and one square image, we could always make the tooltip say something like "You need to upload at least one square image and one landscape image to proceed."

eason9487:
Hi @MatthiasReinholz,

> Can you make the tooltip show text that will be different based on the condition?
> As long as there is not at least one landscape and one square image, we could always make the tooltip say something like "You need to upload at least one square image and one landscape image to proceed."

Yes, it can be conditional. Nikki has provided the new wording. I'll adjust it later.

PR #1896, opened 2023-02-20
PMax Assets - Update Google Ads library v12

jorgemd24:

Closes part of #1780.

This PR updates the branch  [feature/pmax-assets](https://github.com/woocommerce/google-listings-and-ads/tree/feature/pmax-assets)  with the new Google Ads library version (v12). See commit: [4c44730](https://github.com/woocommerce/google-listings-and-ads/commit/4c4473062a8befcac4a38b7e4987c356c9eedadb) 

Also, it merges the branch `develop` into  `feature/pmax-assets` so we include the rest of the changes.

### Creating campaign
1. Go to GLA-> Dashboard
2. Create a campaign
3. Choose a final URL and add some assets.
4. Click "Launch Paid Campaign"
5. If your local domain matches the Merchant center URL, the campaign should be created successfully with the assets.

### Editing campaign
1. Choose a campaign with an existing asset group
2. Click on "Edit" the campaign, go to the assets section.
3. Edit some assets.
4. Click Save changes
5. The assets should be saved.

### Changelog entry

>


mikkamp (approved):
I've had some trouble testing this PR.

First of all I tried testing with several images over 1MB although the total never reached more than 10MB. Somehow as soon as I surpass 1MB I get the following exception:

```
Automattic\WooCommerce\GoogleListingsAndAds\Vendor\GuzzleHttp\Exception\ClientException: Client error: `POST https://localhost:5500/google/google-ads/v12/customers/123456789/googleAds:mutate` resulted in a `413 Payload Too Large` response:
{"statusCode":413,"error":"Request Entity Too Large","message":"Payload content length greater than maximum allowed: 104 (truncated...)
```

Problem however is that the Google Gax library tries to convert the exception data, but the error contains a basic string `Request Entity Too Large`. There have been some recent [fixes here](https://github.com/googleapis/gax-php/blob/v1.19.0/src/ApiException.php#L310) however that sets `basicMessage` to null, and when the create function is called it's expected to be a string.

```
[21-Feb-2023 15:30:13 UTC] PHP Fatal error:  Uncaught TypeError: Google\ApiCore\ApiException::create(): Argument #1 ($basicMessage) must be of type string, null given, called in google-listings-and-ads/vendor/google/gax/src/ApiException.php on line 194 and defined in google-listings-and-ads/vendor/google/gax/src/ApiException.php:242
Stack trace:
#0 google-listings-and-ads/vendor/google/gax/src/ApiException.php(194): Google\ApiCore\ApiException::create()
#1 google-listings-and-ads/vendor/google/gax/src/ApiException.php(315): Google\ApiCore\ApiException::createFromRestApiResponse()
#2 google-listings-and-ads/vendor/google/gax/src/Transport/RestTransport.php(168): Google\ApiCore\ApiException::createFromRequestException()
#3 google-listings-and-ads/vendor/guzzlehttp/promises/src/Promise.php(204): Google\ApiCore\Transport\RestTransport->Google\ApiCore\Transport\{closure}()
#4 google-listings-and-ads/vendor/guzzlehttp/promises/src/Promise.php(153): Automattic\WooCommerce\GoogleListingsAndAds\Vendor\GuzzleHttp\Promise\Promise::callHandler()
```

After removing some images, the next one I discovered was:
> Final url and shopping merchant url does not have the same domain.

That was easy enough to resolve but the next error was:
> Error editing asset group: Marketing image asset for a valid asset group is not enough.

Took me a while to realize that because I removed larger images I was no longer sending any images in the size `marketing_image`. Once that was added again the request worked.

So while it doesn't look like the Ads library v12 is changing anything or causing any errors. There is a slight behaviour difference in the Google Gax library which upgraded from 1.13.0 > 1.19.0. Both versions struggle with the error that's thrown, but it results in the wrong error being reported in the dashboard:
image

I finally realized why I'm unable to use more than 1MB, it's because I'm also using this proxy for testing, and it also needs the payload maxBytes set: https://github.com/woocommerce/google-listings-and-ads/blob/2.3.10/tests/proxy/index.js#L55-L57

So while I think this PR can get merged, I just wanted to document the struggles when testing. I am worried about error handling from our middleware. It's expecting to get the error in a certain format because it sent a request to the Ads API, but if the error actually comes from the middleware failing then it's not handled as expected. Maybe we should log a follow up issue to see if we can resolve that. We do already have a [custom error_handler](https://github.com/woocommerce/google-listings-and-ads/blob/2.3.10/src/Internal/DependencyManagement/GoogleServiceProvider.php#L196) so maybe the missing piece is to ensure it's an error the Ads library can handle if that's where the request came from. Probably also something to cover in unit tests, as the different error exception format doesn't seem to be covered in our unit tests nor in the Google Gax unit tests.

jorgemd24:
Thanks @mikkamp for testing this PR and the elaborated details.

>Problem however is that the Google Gax library tries to convert the exception data, but the error contains a basic string Request Entity Too Large. There have been some recent [fixes here](https://github.com/googleapis/gax-php/blob/v1.19.0/src/ApiException.php#L310) however that sets basicMessage to null, and when the create function is called it's expected to be a string.

When I tried the maximum payload using the Google Ads library v11, the message was not `null`, it threw the first letter from the Error Response, in our case the letter `R`  and I was handling the error in the following way:

https://github.com/woocommerce/google-listings-and-ads/blob/b6c6e786a2d898f44018ebdc327d6523d7765b1e/src/API/Google/AdsAssetGroup.php#L370-L372

image


> After removing some images, the next one I discovered was:
> Final url and shopping merchant url does not have the same domain.

It was mentioned in the PR description but maybe was not clear enough. 

> Error editing asset group: Marketing image asset for a valid asset group is not enough.
> Took me a while to realize that because I removed larger images I was no longer sending any images in the size marketing_image. Once that was added again the request worked.

Didn't the front end prevent updating the asset group if the minimum image quantity was not filled correctly? 

> I finally realized why I'm unable to use more than 1MB, it's because I'm also using this proxy for testing, and it also needs the payload maxBytes set: https://github.com/woocommerce/google-listings-and-ads/blob/2.3.10/tests/proxy/index.js#L55-L57

Ah! I did not test it using that proxy. I will update the maxBytes value there.

> We do already have a [custom error_handler](https://github.com/woocommerce/google-listings-and-ads/blob/2.3.10/src/Internal/DependencyManagement/GoogleServiceProvider.php#L196) so maybe the missing piece is to ensure it's an error the Ads library can handle if that's where the request came from.

We could check if the response contains the header [request-id](https://developers.google.com/google-ads/api/rest/auth#request_id) if so we throw the ApiException otherwise we could throw a ProxyException.  WDYT?

mikkamp:
> When I tried the maximum payload using the Google Ads library v11, the message was not null, it threw the first letter from the Error Response, in our case the letter R and I was handling the error in the following way:

Yes that's correct, because with V11 it was using [Google Gax 1.13.0](https://github.com/googleapis/gax-php/blob/v1.13.0/src/ApiException.php#L302) which didn't have the null fallback but instead incorrectly uses an array index on a string. It pretty much does the following:
```php
$foo = 'Request entity too large'';
$bar = $foo['error'];
var_dump( $bar );
```
Which on PHP 7.4 throws a warning and then just results in the first letter from the string being assigned to `$bar`. However I was testing it with PHP 8.0 which turns that into a fatal error (meaning it never reaches the code where you handle the entity too large error). They fixed that fatal in [Google Gax 1.19.0](https://github.com/googleapis/gax-php/blob/v1.19.0/src/ApiException.php#L310) by setting the message to null, but that's an incorrect fix since it results in a [type mismatch in a later function call](https://github.com/googleapis/gax-php/blob/v1.19.0/src/ApiException.php#L243). Probably something we should report in their code as well so it can be fixed there.

> It was mentioned in the PR description but maybe was not clear enough.

Yes I saw it, but figured I'd continue as is so I could also test error handling.

> Didn't the front end prevent updating the asset group if the minimum image quantity was not filled correctly?

Yes it did, but I cloned the request into postman to play around with different combinations and better understand how the request was handled.

> We could check if the response contains the header [request-id](https://developers.google.com/google-ads/api/rest/auth#request_id) if so we throw the ApiException otherwise we could throw a ProxyException. WDYT?

Yes something along those lines would be good. I've seen it in the past where error messages only returned the first letter, but we've never got round to resolving it properly. Now that it generates a fatal on PHP 8 it's a must to fix though, and I'd like the scenario covered in a unit test, in case there are further changes in the Google Gax library further down the line.

jorgemd24:
Thanks for the explanation @mikkamp,

>Yes something along those lines would be good. I've seen it in the past where error messages only returned the first letter, but we've never got round to resolving it properly. Now that it generates a fatal on PHP 8 it's a must to fix though, and I'd like the scenario covered in a unit test, in case there are further changes in the Google Gax library further down the line.

I opened an [issue](https://github.com/woocommerce/google-listings-and-ads/issues/1902) that includes a possible solution. If my assumptions are correct, I will be happy to implement it and cover the scenario with some tests.

PR #1901, opened 2023-02-23
UX enhancement: Make the popover of the tooltip can be closed properly

eason9487:

It's a follow-up PR to solve the UX concern in https://github.com/woocommerce/google-listings-and-ads/pull/1894#issuecomment-1439586973:

> https://user-images.githubusercontent.com/5908855/220559751-dcf20af3-5e24-4e44-b75d-34cedea8915b.mov
> 
> Not sure if this is the targeted behaviour but I expected the tips to disappear. Didn't go deeper on it, but seems for me that when the mouse goes down the tip doesn't disappear correctly.

This PR implements one of the possible workarounds in GLA:

- ~Add a placeholder to `AppTooltip` for making the popover easier to be closed.~
- Forward the `disabled` prop of the actual child to the wrapper `<div>` in order to make `Tooltip` handle the close of tooltip properly.
- Extra tweak: Replace the CSS class prefix "app-" with "gla-" for the `AppTooltip` component as the `app-` might be too general to avoid a collision.

### Screenshots:

https://user-images.githubusercontent.com/17420811/220883326-183c5638-9ea8-436a-9fa8-45f2a1529a37.mp4

💡 Considering the max limit of 20 is not that easy to test, it would be easier by changing the number in the following line to a smaller one:

https://github.com/woocommerce/google-listings-and-ads/blob/3ca5e71b58a863bfab520bd8143493ad3709374e/js/src/components/paid-ads/assetSpecs.js#L147

1. Go to step 2 of the campaign creation page.
2. Add the "Landscape", "Square" and "Portrait" images till reaching the max limit number to make the "Add image" buttons disabled.
3. Hover over a disabled button to view the popup.
4. Move out of the disabled button to see if it's easier to close the popup than before.
5. Repeat steps 3-4 but move the mouse out of the disabled button from different directions.

### Changelog entry

> Tweak - Make the popover of the tooltip can be closed properly.

### DIFF
--- a/js/src/components/app-tooltip/index.js
+++ b/js/src/components/app-tooltip/index.js
@@ -32,7 +32,7 @@ const AppTooltip = ( props ) => {
 			This inline-block div is needed for the tooltip to show up correctly.
 			If we use span, the tooltip will not wrap ToggleControl nicely.
 			*/ }
-			<div className="app-tooltip__children-container">{ children }</div>
+			<div className="gla-tooltip__children-container">{ children }</div>
 		</Tooltip>
 	);
 };
--- a/js/src/components/app-tooltip/index.js
+++ b/js/src/components/app-tooltip/index.js
@@ -1,9 +1,8 @@
 /**
  * External dependencies
  */
-import classnames from 'classnames';
 import { Tooltip } from 'extracted/@wordpress/components';
-import { Children, isValidElement } from '@wordpress/element';
+import { Children } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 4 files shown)


puntope:
My only concern is that if you pass the mouse fast it also opens the tooltip without closing. See my mouse in video. I think is again something with Tooltip and not with the implementation in PMAX


https://user-images.githubusercontent.com/5908855/220922733-eb582454-15eb-4a35-ab1f-e05700a3a9b8.mov

eason9487:

> My only concern is that if you pass the mouse fast it also opens the tooltip without closing. See my mouse in video. I think is again something with Tooltip and not with the implementation in PMAX

Originally, I had tried to create [the placeholder](https://github.com/woocommerce/google-listings-and-ads/blob/3b4e41cd7e5f1135a09ebdafe0ce7365145fd268/js/src/components/app-tooltip/index.scss#L12) without `:has(.components-tooltip)`. In this way, it should not have this issue, but the placeholder would cover the child before the popup shows up, so I thought it might be a little bit too early and aggressive to cover the child.

After revisiting the source code, I found I misunderstood it. Fortunately, `Tooltip` already has [the same placeholder (event catcher)](https://github.com/WordPress/gutenberg/blob/%40wordpress/components%4019.17.0/packages/components/src/tooltip/style.scss#L30-L39) to handle the close of popover for the disabled child.

So I changed to use the solution that forwards the `disabled` prop to `Tooltip` in https://github.com/woocommerce/google-listings-and-ads/pull/1901/commits/d32d1d72fc1283d67fc540cecdd0907553203b30. It should make popover close properly and also won't have the issue when passing over the button fast.

Could you help with a new round of code reviews?

puntope (approved):
Thanks for the adjustments. It work like a charm now 

✅  Approved

PR #1903, opened 2023-02-24
PMax Assets: Prevent the edited asset values from being reset after toggling asset fields

eason9487:

This PR fixes a bug and it relates to #1787.

#### Problem

Edited images or texts will be reset back to the fetched asset suggestion after collapsing and expanding the `AssetField`.

https://user-images.githubusercontent.com/17420811/221122703-b55d9203-8f33-41bb-9574-295c1c41c708.mp4

#### Solution

This PR changes to hide/show the children via the `hidden` attribute on the wrapper instead of whether to render the children to the DOM tree. In this way, the children won't be unmounted and remounted so their states can be kept.

### Screenshots:

https://user-images.githubusercontent.com/17420811/221123913-dfeae92f-1145-4936-98e0-afdf2d3da0a9.mp4

2. After assets are loaded on UI, change asset images and texts.
3. Collapse and expand asset fields to see if the edited values are kept.
4. Check if the toggle function of all asset fields works well.

### DIFF
--- a/js/src/components/paid-ads/asset-group/asset-field.js
+++ b/js/src/components/paid-ads/asset-group/asset-field.js
@@ -138,8 +138,8 @@ function AssetField(
 					/>
 				</div>
 			</header>
-			<div className="gla-asset-field__content">
-				{ shouldExpand && children }
+			<div className="gla-asset-field__content" hidden={ ! shouldExpand }>
+				{ children }
 			</div>
 		</div>
 	);
--- a/js/src/components/paid-ads/asset-group/asset-field.scss
+++ b/js/src/components/paid-ads/asset-group/asset-field.scss
@@ -93,6 +93,10 @@
 	}
 
 	&__content {
+		&[hidden] {
+			display: none !important;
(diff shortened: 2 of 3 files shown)


jorgemd24 (approved):
Thanks @eason9487, this PR is fixing the issue of resetting the asset fields.

Not related to this PR but while testing it I saw that some images are not appropriately cropped using the media selector, for example, the picture of "Polo"  from the sample products.

See video:

https://user-images.githubusercontent.com/2488994/221201426-b4af064a-e24a-4cee-8363-4bc5a761cf1a.mp4

.

See payload when cropping the image:


image

I tried with other images and they were cropped ok.

eason9487:
@jorgemd24, thanks for the review and for finding the issue.

> Not related to this PR but while testing it I saw that some images are not appropriately cropped using the media selector, for example, the picture of "Polo" from the sample products.

It seems to be a bug in the `imgAreaSelect` jQuery plugin. I will create a PR for it.

PR #1905, opened 2023-02-24
Update maximum payload size in the test proxy

jorgemd24:

This PR increase the maximum payload size in the test proxy so we are consistent with a similar change done in WCS. See PR [2025](https://github.com/Automattic/woocommerce-connect-server/pull/2025).

1. Checkout `develop` branch.
2. Download this image, it is around 4.8MB. 

3. Send a POST request to `http://YOUR_LOCAL_TEST_PROXY/google/google-ads/v11/customers/123456789/assets:mutate`,  use the previous image as payload. You can use Postman to make the request:

image

4. You should get 413 error - Payload content length greater than maximum allowed: 1048576
5. Checkout this branch and restart the server.
6. You will get the following error which means that the request has not been rejected because of the payload size.
```
{
    "statusCode": 401,
    "error": "Unauthorized",
    "message": "Missing authentication"
}
```

### Changelog entry

> Dev - Increase maximum payload size in the test proxy.

### DIFF
--- a/tests/proxy/index.js
+++ b/tests/proxy/index.js
@@ -54,6 +54,7 @@ const init = async () => {
 			},
 			payload: {
 				parse: false,
+				maxBytes: 40 * 1024 * 1024,
 			},
 		},
 	} );


eason9487 (approved):
LGTM.

PR #1907, opened 2023-02-27
PMax Assets: Prevent the edited asset values from being reset after switching between steps

eason9487:

This PR fixes a bug and it relates to #1787.

#### Problem

Edited images or texts will be reset back to the fetched asset suggestion after switching between steps.

https://user-images.githubusercontent.com/17420811/221531053-b2918a0b-ed88-49ce-873f-7ddffb8420d6.mp4

#### Solution

Distinguish whether it should use `values` or `baseAssetGroup` as the initial values.

### Screenshots:

https://user-images.githubusercontent.com/17420811/221531090-6432d08d-fe75-49c6-b618-0e5b68931f93.mp4

2. After assets are loaded on UI, change asset images and texts.
3. Switch to step 1 and back to step 2.
4. Check if the edited values are kept.
5. Select another final URL and see if the loaded assets are presented on the UI.

### DIFF
--- a/js/src/components/paid-ads/asset-group/asset-group-card.js
+++ b/js/src/components/paid-ads/asset-group/asset-group-card.js
@@ -96,10 +96,24 @@ export default function AssetGroupCard() {
 		firstErrorRef.current = ref;
 	}
 
+	// Currently, because WC's `Form` component can not properly set the multiple values
+	// synchronously. So, it needs to distinguish whether the `baseAssetGroup` is already
+	// set when mounting and not being cleaned. If yes, it means the `values` are also
+	// available and include the changes made by the user. Otherwise, it should need to
+	// wait for the `baseAssetGroup` to be set from the `AssetGroupSection`. So that the
+	// edited values in `AssetField` won't be reset after remounting the `AssetGroupCard`.
+	const isSelectedAssetGroupInitiallyRef = useRef( isSelectedFinalUrl );
+	if ( ! isSelectedFinalUrl ) {
+		isSelectedAssetGroupInitiallyRef.current = isSelectedFinalUrl;
+	}
+	const initialValues = isSelectedAssetGroupInitiallyRef.current
+		? values
+		: baseAssetGroup;
+
 	return (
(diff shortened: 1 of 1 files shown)


js/src/components/paid-ads/asset-group/asset-group-card.js
```diff
@@ -96,10 +96,24 @@ export default function AssetGroupCard() {
 		firstErrorRef.current = ref;
 	}
 
+	// Currently, because WC's `Form` component can not properly set the multiple values
+	// synchronously. So, it needs to distinguish whether the `baseAssetGroup` is already
+	// set when mounting and not being cleaned. If yes, it means the `values` are also
+	// available and include the changes made by the user. Otherwise, it should need to
+	// wait for the `baseAssetGroup` to be set from the `AssetGroupSection`. So that the
+	// edited values in `AssetField` won't be reset after remounting the `AssetGroupCard`.
+	const isSelectedAssetGroupInitiallyRef = useRef( isSelectedFinalUrl );
```
puntope, line 118: 💅  I found this explanations a bit confusing. It took me a while to understand the relation between `isSelectedFinalUrl` if `baseAssetGroup` is set or not.

On my understanding 

-  if `assetGroup` is set, means `values` are also available. And we use them as current values. 
- `assetGroup` is set when `isSelectedFinalUrl` is true when the component mounts.  
- We save this in a ref `isSelectedAssetGroupInitiallyRef ` for preventing it to be override in further component re-renders. 
- In case `assetGroup` is not initially set we load `baseAssetGroup` as current values

puntope (approved):
✅  LGTM 

Tested behaviour locally. Left some comments regarding your comment in the code.

js/src/components/paid-ads/asset-group/asset-group-card.js
eason9487: Rewrote the code comment in https://github.com/woocommerce/google-listings-and-ads/pull/1907/commits/eb61adf03ddc5b14f3f36c07b30023ee70cfb882. Hope this explains more clearly.

PR #1909, opened 2023-02-28
PMax Assets: Prevent the button styles in `ImagesSelector` from overriding by AutomateWoo

eason9487:

It's a follow-up and relates to #1787.

In https://github.com/woocommerce/google-listings-and-ads/pull/1873#pullrequestreview-1288947337 and https://github.com/woocommerce/google-listings-and-ads/pull/1873#issuecomment-1423979040:
> I think the delete button looks too big compared with the preview images, what do you think? Or comparing with the Figma file, it seems that the preview images should be bigger.

> The conflict with the styling is with AutomateWoo 5.6.6 when I disable AW loads correctly.

This PR fixes the style conflict that results in style problems in the `ImagesSelector`:

In the screenshot, the style problems are:
1. Incorrect image button size.
2. The border of Image buttons disappears.
3. The X button should show up when hovering over the image button.
4. Incorrect background color of X button.

1. Install and activate AutomateWoo.
2. Go to step 2 of the campaign creation page.
3. Select some images to see if the style looks good and is similar to the design in Figma.

### DIFF
--- a/js/src/components/paid-ads/asset-group/images-selector.scss
+++ b/js/src/components/paid-ads/asset-group/images-selector.scss
@@ -14,44 +14,47 @@
 		display: flex;
 	}
 
-	&__replace-image-button {
-		justify-content: center;
-		width: 100px;
-		height: 100px;
-		padding: 0;
-		border: 1px solid $gray-200;
-		border-radius: 2px;
-		overflow: hidden;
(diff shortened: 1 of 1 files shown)


puntope (approved):
Looking good. Now it looks nicer and matching the Figma

PR #1912, opened 2023-03-01
Add support for the Assets of Performance Max Campaigns

eason9487:
### Changes proposed in this Pull Request:

Closes #1780 
Closes #1787 

This PR merges the feature branch `feature/pmax-assets` that implements stage 1 of the PMax Assets feature to the main branch `develop`.

💡 This PR will be added additional commits that only include the updates of the [tracking document](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/README.md). All other code changes have been reviewed and tested in the respective PRs. Therefore, this PR should not require extra rounds of code reviews.

### 📌 Before merging (@eason9487)

- [x] Update [tracking document](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/README.md) and commit to this PR.

### Changelog entry

> Add - Support for the Assets of Performance Max campaigns.

Issue #1833, opened 2022-12-30
Gradually externalize `@wordpress/components` and finally opt-in all imports via DEWP


eason9487:
With the ability to selectively externalize specific imports, introduced in #1762, we can start gradually externalizing `@wordpress/components`, and finally opt-in all imports via DEWP.

This issue is used to track this process and illustrate the approach.

## Progress of `@wordpress/components` imports to be externalized

- [x] __experimentalInputControl - [#2054](https://github.com/woocommerce/google-listings-and-ads/pull/2054)
- [x] Button - [#1834](https://github.com/woocommerce/google-listings-and-ads/pull/1834)
   - `@import "node_modules/@wordpress/components/src/button/style";` in *_gutenberg-components.scss* should be removed together.
- [x] Card - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] CardBody - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] CardDivider - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] CardFooter - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] CardHeader - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] CheckboxControl - [#2054](https://github.com/woocommerce/google-listings-and-ads/pull/2054)
- [x] Dashicon - [#2037](https://github.com/woocommerce/google-listings-and-ads/pull/2037)
- [x] Dropdown - [#2042](https://github.com/woocommerce/google-listings-and-ads/pull/2042)
- [x] ExternalLink
- [x] Flex - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] FlexBlock - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] FlexItem - [#2036](https://github.com/woocommerce/google-listings-and-ads/pull/2036)
- [x] Icon - [#2037](https://github.com/woocommerce/google-listings-and-ads/pull/2037)
- [x] KeyboardShortcuts - [#1914](https://github.com/woocommerce/google-listings-and-ads/pull/1914)
- [x] Modal - [#2042](https://github.com/woocommerce/google-listings-and-ads/pull/2042)
- [x] NavigableMenu - [#2042](https://github.com/woocommerce/google-listings-and-ads/pull/2042)
[...]

PR #1914, opened 2023-03-02
Fix the blank Product Feed page after completing the onboarding flow

eason9487:

While testing the PMax Assets feature (#1912), I found that the Product Feed is a blank page after completing the onboarding flow. The problem is the `KeyboardShortcuts` calls to an undefined function `isAppleOS`.

image

This problem could be simply fixed by externalizing `KeyboardShortcuts`, which is also a part of #1833 - Gradually externalize `@wordpress/components`. (See the Additional details section for more details)

Additional details

### Screenshots:

image

1. Set up a WP version < 6.1.
2. Go to the Product Feed page with the opened submission success guild: `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed&guide=submission-success`
3. The page and guild should be rendered.

#### The root cause

In #1779:

> Moreover, to avoid conflict between WC and WP packages and different `react` versions, we also have to bump all `@woocommerce/*` packages to align with WC 6.9, and all `@wordpress/*` packages need to be aligned with the corresponding versions used in `@woocommerce/components` 10.3.0. Otherwise, the JS unit test would encounter a lot of "Invalid hook call" errors due to multiple `react` versions.

the version of all `@wordpress/*` packages was aligned to the corresponding versions in WC 6.9 rather than the actual versions in WP 5.9. Another weird thing is the `@wordpress/*` packages [used by WC 6.9](https://github.com/woocommerce/woocommerce/blob/6.9.4/packages/js/components/package.json) doesn't match [its min WP version](https://github.com/woocommerce/woocommerce/blob/6.9.4/plugins/woocommerce/woocommerce.php#L11) - 5.8.

The inconsistency of the versions between GLA, WC and WP, although most of the components still functioned properly, this problem occurred after all.

In WP 5.9.5, the provided core packages are
- `@wordpress/components` 19.2.3 for `KeyboardShortcuts` component
- `@wordpress/compose` 5.0.7 for `useKeyboardShortcut` hook, and there is a function `isAppleOS` within the hook
- `@wordpress/keycodes` 3.2.4` but it doesn't expose `isAppleOS` function

However, GLA depends on
- `@wordpress/components` 19.17.0 for `KeyboardShortcuts`
- `@wordpress/compose` 5.20.0 for `useKeyboardShortcut`, and it depends on `isAppleOS` exported from `@wordpress/keycodes`
- DEWPed `@wordpress/keycodes`

This causes the bundled `KeyboardShortcuts` to call the `isAppleOS` via WP core but it's not yet exposed in WP 5.9.x.

### Changelog entry

> Fix - The blank Product Feed page after completing the onboarding flow.
> Dev - Externalize the KeyboardShortcuts component.

### DIFF
--- a/js/src/external-components/wordpress/guide/index.js
+++ b/js/src/external-components/wordpress/guide/index.js
@@ -11,8 +11,8 @@
 import classnames from 'classnames';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
-import { Modal, KeyboardShortcuts } from '@wordpress/components';
-import { Button } from 'extracted/@wordpress/components';
+import { Modal } from '@wordpress/components';
+import { Button, KeyboardShortcuts } from 'extracted/@wordpress/components';
 
 /**
  * Internal dependencies


jorgemd24 (approved):
@eason9487 Thanks for fixing this, I could reproduce the issue with WP < v6.1 and this PR fix the issue.

✅  Tested with WP 5.9
✅  Tested with WP 6.0 & 6.1
✅  Tested with WP 6.2-beta3

Issue #864, opened 2021-07-07
Add test coverage report


nima-karimi:
Add a test [coverage](https://phpunit.readthedocs.io/en/7.5/code-coverage-analysis.html) report for unit tests and include a check for Travis.

Some useful links:
- https://phpunit.readthedocs.io/en/7.5/code-coverage-analysis.html
- https://tomasvotruba.com/blog/2019/09/02/how-to-speedup-test-coverage-on-travis-by-95-percent/
- https://dev.to/acho_arnold/php-test-coverage-with-travis-ci-and-code-climate-using-clover-eih
- https://docs.travis-ci.com/user/coveralls/

tomalec:
For the front-end tests/jest we can start with just `--coverage`, but I :+1: Coveralls

mikkamp:
After updating to phpunit 9.5 in PR #1867, as long as we have a PHP version with xdebug installed then a coverage report can be produced as follows.

1. Add the following section to `phpunit.xml.dist`
```xml
	<coverage>
		<include>
			<directory suffix=".php">./src</directory>
			<directory suffix=".php">./views</directory>
			<file>google-listings-and-ads.php</file>
		</include>
	</coverage>
```

2. Setup tests: `bin/install-wp-tests.sh <db_name> <db_user> <db_pass>`
3. Run the tests with a coverage report: `XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-text`

Example output:
![image](https://user-images.githubusercontent.com/11388669/215832210-df1c16d2-9125-4d81-a2df-fb3ebeeb10bf.png)

mikkamp:
Update: A codecov report has been added in #1915
So far this only covers PHP, we also do unit testing in JS which we could include in the coverage report.

PR #1915, opened 2023-03-06
Add PHP Code coverage report as a GitHub action

mikkamp:

Resolves part of #864 

This PR adds a code coverage report as a PR comment. Since we run multiple unit tests in GitHub actions the coverage report is only generated once when the tests are run with the latest WP version and for PHP 8.0.

I've added the coverage using a flag `php-unit-tests`, the intention is to also include the coverage report for the JS unit tests, but that will be part of a followup PR.

> **Note**
There is currently no report stored for the develop branch. So it's only producing a coverage report of the current PR and isn't able to compare the difference between the two. Once this PR is merged to develop it will generate the report for develop which will allow the comparison to appear in subsequent PR's.

1. View the code coverage report here in the [PR comment](https://github.com/woocommerce/google-listings-and-ads/pull/1915#issuecomment-1483222517) and confirm it contains the right test results (no comparison data available)
2. Confirm that in [the workflow runs](https://github.com/woocommerce/google-listings-and-ads/actions/runs/4532661162/jobs/7984348011?pr=1915) the coverage report is only generated for PHP 8.0 and WP latest, we should see skipped tasks that differ:
image

### Changelog entry
* Dev - Add PHP Code coverage report as GitHub action.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,28 @@
+codecov:
+    require_ci_to_pass: yes
+
+coverage:
+    precision: 1
+    round: nearest
+    range: '50...80'
+
+    status:
+        project:
+            default:
+                target: auto
+        patch:
+            default:
+                target: auto
+
+ignore:
+    - 'bin/*'
+    - 'tests/*'
+    - 'vendor/*'
+    - 'node_modules/*'
+
+comment:
+    layout: 'reach, diff, flags, files'
+    behavior: default
+    require_changes: false
+    require_base: no
+    require_head: yes
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,14 +6,6 @@ coverage:
     round: nearest
     range: '50...80'
 
-    status:
-        project:
-            default:
-                target: auto
-        patch:
-            default:
-                target: auto
-
 ignore:
     - 'bin/*'
     - 'tests/*'
@@ -26,3 +18,4 @@ comment:
     require_changes: false
     require_base: no
     require_head: yes
+    show_carryforward_flags: false
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -14,8 +14,3 @@ ignore:
 
 comment:
     layout: 'reach, diff, flags, files'
-    behavior: default
-    require_changes: false
-    require_base: no
-    require_head: yes
-    show_carryforward_flags: false
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -17,4 +17,11 @@
 			<directory suffix=".php">./tests/integration</directory>
 		</testsuite>
 	</testsuites>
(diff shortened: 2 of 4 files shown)


phpunit.xml.dist
```diff
@@ -17,4 +17,11 @@
 			<directory suffix=".php">./tests/integration</directory>
 		</testsuite>
 	</testsuites>
+    <coverage includeUncoveredFiles="true">
```
puntope: 💅  [Seems like this is the default value](https://docs.phpunit.de/en/9.5/code-coverage-analysis.html#:~:text=includeUncoveredFiles%3D%22true%22%20(default)%20means%20that%20all%20files%20are%20included%20in%20the%20code%20coverage%20report%20even%20if%20not%20a%20single%20line%20of%20code%20of%20such%20a%20file%20is%20executed), so no need for adding this .

.codecov.yml
```diff
@@ -0,0 +1,21 @@
...
+    - 'tests/*'
+    - 'vendor/*'
+    - 'node_modules/*'
+
+comment:
+    layout: 'reach, diff, flags, files'
+    behavior: default
+    require_changes: false
+    require_base: no
+    require_head: yes
```
puntope: 💅 This is default to yes. So we can consider omit it.

.codecov.yml
```diff
@@ -0,0 +1,21 @@
...
+ignore:
+    - 'bin/*'
+    - 'tests/*'
+    - 'vendor/*'
+    - 'node_modules/*'
+
+comment:
+    layout: 'reach, diff, flags, files'
+    behavior: default
+    require_changes: false
```
puntope: These three seems for me like default options

.codecov.yml
```diff
@@ -0,0 +1,21 @@
...
+    require_base: no
+    require_head: yes
+    show_carryforward_flags: false
```
puntope: This I think is also false by default

.codecov.yml
mikkamp: I thought so too, but apparently the default looked different without adding these settings. For example the graph (`reach`) is not displayed by default. If you have a look at the edited link of the codecov comment you can see the history of what the comment looked like. I ran it quite a few times to get it to contain the sections I wanted.

I can take another look to see if there are additional sections we can omit but I tried to copy the defaults from the WooCommerce core `yml` file.

mikkamp:
I went ahead and removed more of those defaults as you suggested. I still left the comment.layout since I wasn't getting the `reach` section included initially. The `files` section is still not included, but that relies on comparing with a parent branch, so will only start showing up after we've merged to develop. It's ready for another round of review.

puntope (approved):
✅  LGTM

Issue #1922, opened 2023-03-10
Return the URL to setup ads account in MCM dashboard if it is not already set up


nima-karimi:
Based on the discussion here: https://github.com/woocommerce/woocommerce/pull/37044#discussion_r1129068032

Currently, return the following static URL to WooCommerce's Multichannel Marketing dashboard to allow users to create a new campaign: `admin.php?page=wc-admin&path=/google/dashboard&subpath=/campaigns/create`

This URL will be used on the marketing dashboard in the following section (PR https://github.com/woocommerce/woocommerce/pull/37044):

<img width="1035" alt="image" src="https://user-images.githubusercontent.com/73110514/224313190-4e5378fb-6704-44eb-867c-cb2065c370c3.png">

However, if the ads account has not been set up in GLA, the user cannot visit that URL. This means that we are returning a broken URL.

To resolve this, we can a check in the `MultichannelMarketing\GLAChannel::generate_campaign_types` method to see if the ads account has been set up and connected before returning the URL to create a campaign page. If it is not set up, we need to return the URL to the ads setup page.

martynmjones:
Resolved in #1926 as any attempt to access `/google/dashboard` will be redirected to onboarding if the account isn't set up.

PR #1926, opened 2023-03-20
Add redirects between dashboard and start

### LINKED ISSUE #1668: Consistent URL to redirect to from external location
## User story

If we want to redirect a user to a page within GLA we currently only have the option of redirecting them to:

- `?page=wc-admin&path=/google/dashboard`
- `?page=wc-admin&path=/google/start`

Which page to redirect them to depends on the state of onboarding, however if we are redirecting form an external source we don't know the onboarding state. This is relevant to cases where we want to provide a landing URL from a [remote inbox notification](https://github.com/woocommerce/woocommerce-admin/tree/main/src/RemoteInboxNotifications#readme).

So we would either need one consistent landing page, or alternatively have the `/google/dashboard` link automatically redirect to `/google/start` if onboarding hasn't completed yet. Currently when you visit the wrong URL you are shown an access denied message:

image

One consistent URL to link to without getting any error pages.

One alternative suggestion was to link to `?page=wc-admin&path=/marketing` but this is less than ideal since it can show many extensions in this list.

martynmjones:

Closes #1668, #1922

If a merchant attempts to access the Dashboard before they have completed setup then they'll get a `Sorry, you are not allowed to access this page.` error message. 

Screenshot 2023-03-20 at 18 55 52

This PR adds a redirect on the backend so that if a merchant **has not** completed setup they'll be redirected to the GetStarted page when attempting to access the Dashboard.

If the merchant **has** completed setup and tries to access the GetStarted page they'll be redirected to the Dashboard.

1. Checkout branch on a test site that has not yet completed GLA setup
2. Activate Google Listings & Ads on it's own and confirm you're redirected to onboarding
3. Visit `admin.php?page=wc-admin&path=/google/dashboard`
4. Confirm that you are redirected to the GetStarted page
5. Complete setup
6. Visit `admin.php?page=wc-admin&path=/google/start`
3. Confirm that you are redirected to the Dashboard

### Changelog entry

> Tweak - Redirect users between Dashboard and Get Started pages as required


src/Menu/GetStarted.php
```diff
@@ -65,4 +95,31 @@ protected function register_navigation_pages(): void {
...
+
+	/**
+	 * Redirect to start page and exit.
+	 *
+	 * @return void
+	 */
+	private function redirect_to_dashboard(): void {
+		wp_safe_redirect( admin_url( 'admin.php?page=wc-admin&path=' . rawurlencode( Dashboard::$path ) ) );
+		exit;
+	}
```
puntope: This two functions are 90% equal than the ones defined in the Dashboard redirection. The only thing changing is the referred path.  So maybe we can refactor them as something like:

`redirect_to_path( $path )`

src/Menu/Dashboard.php
```diff
@@ -70,12 +100,39 @@ protected function register_navigation_pages(): void {
...
+	}
+
+	/**
+	 * Redirect to start page and exit.
+	 *
+	 * @return void
+	 */
+	private function redirect_to_start_page(): void {
+		wp_safe_redirect( admin_url( 'admin.php?page=wc-admin&path=' . rawurlencode( GetStarted::$path ) ) );
+		exit;
```
puntope: See https://github.com/woocommerce/google-listings-and-ads/pull/1926/files#r1143240315

puntope:
The code works. But I'm wondering about the implementation:

Checking `ActivationRedirect.php` I see some similar logic  about what we are doing, so what do you thing about moving the logic there? 

That way we can:

- Code reduction `is_onboarding_page` and `is_current_wc_admin_page` can be merged in one unique function.
- Avoid code duplication in  `maybe_redirect_to_dashboard ` and `maybe_redirect_to_start_page`
- Avoid injecting WP class inside the Menu items. 
- Avoid adding a new method in MenuFixTrait
- Separation of concerns between menu items and redirections.

Maybe we can rename that as Redirects.php  and include there both. Current on_activation redirect. + the new redirects between get_started and dashboard. 

What do you think?

martynmjones:
Hey @puntope 👋 Thanks for the review and suggestion. I think that's a far cleaner approach and will be a lot easier to work with down the line.

I've reworked this PR based on your recommendation so we're using a `Redirect` class to manage the activation redirect and redirecting merchants to the correct path depending on the status of the account setup.

src/Admin/Redirect.php
```diff
@@ -81,62 +85,90 @@ public function activate(): void {
...
+	protected function maybe_redirect_after_activation(): bool {
 		// Do not redirect if setup is already complete
 		if ( $this->merchant_center->is_setup_complete() ) {
 			$this->options->update( self::OPTION, 'no' );
 			return false;
 		}
 
 		// if we are on the get started page don't redirect again
-		if ( $this->is_onboarding_page() && 'yes' === $this->options->get( self::OPTION, 'yes' ) ) {
+		if ( $this->is_current_wc_admin_page( self::PATHS['get_started'] ) && 'yes' === $this->options->get( self::OPTION, 'yes' ) ) {
```
puntope: 💅  I guess` 'yes' === $this->options->get( self::OPTION, 'yes' )` is not necessary since at this point the option is always YES.

puntope (approved):
✅  LGTM 

Thanks for the changes and further adjustments. 

💅  I left a comment regarding one condition that seems unnecessary. Another thing I'm missing are the unit tests.

martynmjones:
> I left a comment regarding one condition that seems unnecessary. Another thing I'm missing are the unit tests.

Nice catch! I've [removed the condition](https://github.com/woocommerce/google-listings-and-ads/pull/1926/commits/d77a926a6ea595277ffc49c09d8df2cf9db31482) and [added unit tests](https://github.com/woocommerce/google-listings-and-ads/pull/1926/commits/0f1c34fd9d106501cda85772b823412c21990665) if you could please take one last look 🙌

puntope (approved):
Thanks for adding tests.

Just noticing setMethodsExcept is deprecated so it might create problems in future versions of PHPUnit https://github.com/sebastianbergmann/phpunit/pull/3687

martynmjones:
Thanks for the review and link @puntope 👍 I've updated the tests to avoid needing to do that when phpunit is updated.

PR #1927, opened 2023-03-27
Fix attributes in variants for attribute mapping

puntope:

This PR Fixes reported bugs getting taxonomy and custom attributes based sources in Attribute Mapping on Variations. 

**Testing Taxonomy based Attributes** 

1. Create a product variant with a Taxonomy based variations, like Color.
2. Create an attribute mapping rule as (target Color, source Product Color,  Apply to ALL)
3. Save
4. Sync your product (using connection test is ok)
5. See the Color attribute is being loaded in Merchant.
6. Create a simple product.
7. Add an attribute Color (or several colors) to it
8. Sync that product
9. See the Color attribute (first one) is being loaded in Merchant.

**Testing Custom based Attributes** 

Use this snippet to create a custom attribute: 

```php
add_filter(
		'woocommerce_gla_attribute_mapping_sources_custom_attributes',
		function( $values ) {
			return array_merge( $values, ['material']);
		}
	);
```

1. Create a product variant with a Custom attribute named "material" based variations.
2. Create an attribute mapping rule as (target Material, source material,  Apply to ALL)
3. Save
4. Sync your product (using connection test is ok)
5. See the Material attribute is being loaded in Merchant.
6. Create a simple product.
7. Add an attribute material
8. Sync that product
9. See the Material attribute is being loaded in Merchant.

### Changelog entry

> Fix - Bug in Attribute Mapping  - with Taxonomy based rules not being applied in variations.

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1093,6 +1093,11 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			$attribute_value = $product->get_meta( $attribute_name );
 		}
 
-		return is_scalar( $attribute_value ) ? $attribute_value : '';
+		// We only support scalar values.
+		if ( ! is_scalar( $attribute_value ) ) {
+			return '';
+		}
+
+		return explode( ' ' . WC_DELIMITER . ' ', $attribute_value )[0];
 	}
 }
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1098,6 +1098,6 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			return '';
 		}
 
-		return explode( ' ' . WC_DELIMITER . ' ', $attribute_value )[0];
+		return explode( ' ' . WC_DELIMITER . ' ', strval( $attribute_value ) )[0];
 	}
 }
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1104,6 +1104,6 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			return '';
 		}
 
-		return explode( ' ' . WC_DELIMITER . ' ', strval( $attribute_value ) )[0];
+		return explode( ' ' . WC_DELIMITER . ' ', (string) $attribute_value )[0];
 	}
 }
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1019,12 +1019,14 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 		$product = $this->get_wc_product();
 
 		if ( $product->is_type( 'variation' ) ) {
-			return $product->get_attribute( $taxonomy );
+			$value = $product->get_attribute( $taxonomy );
+			return empty( $value ) ? null : $value;
+
 		}
 
 		$values = get_the_terms( $product->get_id(), $taxonomy );
 
-		if ( ! $values ) {
+		if ( empty( $values ) || is_wp_error( $values ) ) {
 			return null;
 		}
 
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -954,7 +954,7 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 	 * Get a source value for attribute mapping
 	 *
 	 * @param string $source The source to get the value
-	 * @return string|null The source value for this product
+	 * @return string The source value for this product
 	 */
 	protected function get_source( string $source ) {
 		$source_type = null;
@@ -1026,8 +1026,8 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			$values = wp_list_pluck( $values, 'name' );
 		}
 
-		if ( empty( $values ) || is_wp_error( $values ) || empty( $values[0] ) ) {
-			return null;
+		if ( empty( $values ) || is_wp_error( $values ) ) {
+			return '';
 		}
 
 		return $values[0];
@@ -1053,7 +1053,7 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			return $product->$getter();
 		}
 
-		return null;
+		return '';
 	}
 
 	/**
@@ -1087,14 +1087,12 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 	protected function get_custom_attribute( $attribute_name ) {
 		$product = $this->get_wc_product();
 
-		$attribute_values = $product->get_attribute( $attribute_name );
+		$attribute_value = $product->get_attribute( $attribute_name );
 
-		if ( ! $attribute_values ) {
-			$attribute_values = $product->get_meta( $attribute_name );
+		if ( ! $attribute_value ) {
+			$attribute_value = $product->get_meta( $attribute_name );
 		}
 
-		$attribute_values = explode( ', ', $attribute_values );
-
-		return empty( $attribute_values[0] ) ? null : $attribute_values[0];
+		return is_scalar( $attribute_value ) ? $attribute_value : '';
 	}
 }
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1019,18 +1019,18 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 		$product = $this->get_wc_product();
 
 		if ( $product->is_type( 'variation' ) ) {
-			$value = $product->get_attribute( $taxonomy );
-			return empty( $value ) ? null : $value;
-
+			$values = $product->get_attribute( $taxonomy );
+			$values = explode(', ', $values);
+		} else {
+			$values = wc_get_product_terms( $product->get_id(), $taxonomy );
+			$values = wp_list_pluck( $values, 'name' );
 		}
 
-		$values = get_the_terms( $product->get_id(), $taxonomy );
-
-		if ( empty( $values ) || is_wp_error( $values ) ) {
+		if ( empty( $values ) || is_wp_error( $values ) || empty( $values[0] ) ) {
 			return null;
 		}
 
-		return wp_list_pluck( $values, 'name' )[0];
+		return $values[0];
 	}
 
 	/**
@@ -1082,17 +1082,19 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 	 * Gets a custom attribute from a product
 	 *
 	 * @param string $attribute_name - The attribute name to get.
-	 * @return string The attribute value
+	 * @return string|null The attribute value or null if no value is found
 	 */
 	protected function get_custom_attribute( $attribute_name ) {
 		$product = $this->get_wc_product();
 
-		$attribute = $product->get_attribute( $attribute_name );
+		$attribute_values = $product->get_attribute( $attribute_name );
 
-		if ( ! $attribute ) {
-			$attribute = $product->get_meta( $attribute_name );
+		if ( ! $attribute_values ) {
+			$attribute_values = $product->get_meta( $attribute_name );
 		}
 
-		return $attribute;
+		$attribute_values = explode(', ', $attribute_values);
+
+		return empty( $attribute_values[0] ) ? null : $attribute_values[0];
 	}
 }
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1020,6 +1020,12 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 
 		if ( $product->is_type( 'variation' ) ) {
 			$values = $product->get_attribute( $taxonomy );
+
+			if ( ! $values ) {
+				$parent = wc_get_product( $product->get_parent_id() );
+				$values = $parent->get_attribute( $taxonomy );
+			}
+
 			$values = explode( ', ', $values );
 		} else {
 			$values = wc_get_product_terms( $product->get_id(), $taxonomy );
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1020,7 +1020,7 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 
 		if ( $product->is_type( 'variation' ) ) {
 			$values = $product->get_attribute( $taxonomy );
-			$values = explode(', ', $values);
+			$values = explode( ', ', $values );
 		} else {
 			$values = wc_get_product_terms( $product->get_id(), $taxonomy );
 			$values = wp_list_pluck( $values, 'name' );
@@ -1093,7 +1093,7 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			$attribute_values = $product->get_meta( $attribute_name );
 		}
 
-		$attribute_values = explode(', ', $attribute_values);
+		$attribute_values = explode( ', ', $attribute_values );
 
 		return empty( $attribute_values[0] ) ? null : $attribute_values[0];
 	}
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1104,6 +1104,8 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			return '';
 		}
 
-		return explode( ' ' . WC_DELIMITER . ' ', (string) $attribute_value )[0];
+		$values = explode( WC_DELIMITER, (string) $attribute_value );
+		$values = array_filter( array_map( 'trim', $values ) );
+		return empty( $values ) ? '' : $values[0];
 	}
 }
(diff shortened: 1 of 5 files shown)


puntope:
```php
add_filter(
		'woocommerce_gla_product_attribute_value_color',
		function( $source, $product ) {
			if ( ! $source && $product->get_type( 'variation') ) {
				return $product->get_attribute( 'pa_color' );
			}

			return $source;
		}, 10, 2
	);

	add_filter(
		'woocommerce_gla_product_attribute_value_size',
		function( $source, $product ) {
			if ( ! $source && $product->get_type( 'variation') ) {
				return $product->get_attribute( 'pa_size' );
			}

			return $source;
		}, 10, 2
	);
```

This Snippets should fix also the issue in the meanwhile the PR is approved and released.

mikkamp:
Do we have a unit test that covers this scenario? I feel like it's something which could have easily been caught in testing.

If a variation is just another custom post type then I'm curious why this doesn't work and why it wasn't found in previous testing. Are there any changes that prevent this from no longer working?

puntope:
I created this issue to follow up with more tests. https://github.com/woocommerce/google-listings-and-ads/issues/1928

src/Product/WCProductAdapter.php
```diff
@@ -1017,7 +1017,12 @@ protected function rule_match_conditions( array $rule ): bool {
 	 */
 	protected function get_product_taxonomy( $taxonomy ) {
 		$product = $this->get_wc_product();
-		$values  = get_the_terms( $product->get_id(), $taxonomy );
+
+		if ( $product->is_type( 'variation' ) ) {
+			return $product->get_attribute( $taxonomy );
```
mikkamp: Before we were using `wp_list_pluck( $values, 'name' )[0]` which isn't very efficient because it would be faster to get the first array element and then fetch the name. But aside from the optimization that indicates that we are only getting a single attribute value.

The [get_attribute function](https://woocommerce.github.io/code-reference/classes/WC-Product.html#method_get_attribute) implodes multiple values separated by a comma. If it's our intention to only return the first attribute value then it would be nice to keep this consistent.

It seems the reason this is failing in the first place for a `variation`, is because WC assigns the term to the parent product, for easy filtering in the catalog.

src/Product/WCProductAdapter.php
```diff
@@ -1017,7 +1017,12 @@ protected function rule_match_conditions( array $rule ): bool {
...
+		}
+
+		$values = get_the_terms( $product->get_id(), $taxonomy );
```
mikkamp: `get_the_terms` can also return a `WP_Error` which isn't handled gracefully and will most likely result in a PHP error. We should probably include a `is_wp_error` check.

I'm also wondering if we should instead use `wc_get_product_terms`, because it's what WC uses. Also it handles caching slightly differently, which means that WC can correctly invalidate that cache (which it can't do the same with `get_the_terms`). Just need to note that it will return an empty array instead of false so the return value needs to be checked differently.

src/Product/WCProductAdapter.php
```diff
@@ -1070,4 +1075,22 @@ protected function format_attribute( $value, $attribute_id ) {
...
+	/**
+	 * Gets a custom attribute from a product
+	 *
+	 * @param string $attribute_name - The attribute name to get.
+	 * @return string The attribute value
+	 */
+	protected function get_custom_attribute( $attribute_name ) {
+		$product = $this->get_wc_product();
+
+		$attribute = $product->get_attribute( $attribute_name );
```
mikkamp: This again merges all the product attributes into a comma separated list, I'd assume we'd also want just a singular value in this case.

src/Product/WCProductAdapter.php
```diff
@@ -1070,4 +1075,22 @@ protected function format_attribute( $value, $attribute_id ) {
...
+
+		if ( ! $attribute ) {
+			$attribute = $product->get_meta( $attribute_name );
```
mikkamp: The function `get_meta` doesn't guarantee a string return type. If the data is serialized within the database then it will call `maybe_unserialize` which could potentially return the data as any type. We should handle that gracefully as we don't want it to break with a fatal if it happens to return an array.

src/Product/WCProductAdapter.php
```diff
@@ -1070,4 +1075,22 @@ protected function format_attribute( $value, $attribute_id ) {
...
 		return $value;
 	}
+
+	/**
+	 * Gets a custom attribute from a product
+	 *
+	 * @param string $attribute_name - The attribute name to get.
+	 * @return string The attribute value
+	 */
+	protected function get_custom_attribute( $attribute_name ) {
```
mikkamp, line 1093: It makes sense to open up this function to also support "custom attributes" but I think we need to clarify this a little more in the documentation: https://github.com/woocommerce/google-listings-and-ads/wiki/Attribute-Mapping-Feature

Previously we only supported "global attributes" (which we called taxonomies) or "meta data" (which we called custom attributes). Now we are including support for what WC actually calls a "custom attribute" but it also means if we have a product with "color" set as both custom attribute and as meta data then even with a code snippet you can't force it to only pick from a specific source. I don't think that's a bad thing, but lets have that clearly included in the documentation so we are clear on how this is expected to work.

mikkamp (commented):
I've added some more comments on edge cases that we might encounter. I personally have the feeling that since a mapping rule is set once and then potentially applied to a whole range of different products, we should also ensure that these edge cases are handled because we don't know what the product might look like. Under the same premise I have the feeling since we provided the customer that reported the issue with a temporary code snippet, there is less of an urge to get this PR released as soon as possible. Hence I think it will be the right time to add those unit tests so we can not only be confident in mapping a variety of products but the unit tests also set a standard of how we expect the mapping to work.

I think we should even create a separate set of test for mapping various product types, so it doesn't have to be part of the WCProductAdapter tests.

sukafia:
@puntope and @mikkamp a friendly reminder that this fix still hasn't been released.

puntope:
> @puntope and @mikkamp a friendly reminder that this fix still hasn't been released.

Howdy @sukafia 

We want to add some tests and support for edge cases before merging. I plan to start with it next week. However, for the issues coming in, the snippet might be used.

puntope:
Howdy @mikkamp I added a bunch of tests and applied almost all of your suggested changes. Now the coverage reached almost 98%

Screenshot 2023-05-02 at 18 00 06

src/Product/WCProductAdapter.php
puntope: This one I'm not sure what you mean. As far as I understand we only support strings, int and boolean. Serialized data or other non scalar types should be avoided I guess.

src/Product/WCProductAdapter.php
puntope: Done

src/Product/WCProductAdapter.php
puntope: Ok, done

src/Product/WCProductAdapter.php
puntope: Done

src/Product/WCProductAdapter.php
```diff
@@ -1070,4 +1077,24 @@ protected function format_attribute( $value, $attribute_id ) {
...
+		$product = $this->get_wc_product();
+
+		$attribute_values = $product->get_attribute( $attribute_name );
+
+		if ( ! $attribute_values ) {
+			$attribute_values = $product->get_meta( $attribute_name );
+		}
+
+		$attribute_values = explode( ', ', $attribute_values );
```
mikkamp: `get_attribute` only implodes with a comma when we pass it a taxonomy (global attribute) but for custom attributes it uses `WC_DELIMITER` surrounded by spaces. So if we assign multiple custom attributes (to a simple product) it gets the values as `Blue | Green | Red`, which means this explode does nothing. Since we know it's always a taxonomy here would `get_attributes` be a better function to use and then fetch the name from the slug?

If it's not an attribute we try to fetch it from meta data instead which could be a string with any delimiter. If we want to default to using a comma as a delimiter it would be better to explode using just a comma and then trim spaces.

However one other thing that doesn't seem to be taken into account here is that `get_meta` can return a mixed type, it could be an array if it was unserialized, which for newer PHP versions would result in a fatal when we try to explode:

```
FATAL ERROR Uncaught TypeError: explode(): Argument #2 ($string) must be of type string, array given
```

I previously mentioned that before [here](https://github.com/woocommerce/google-listings-and-ads/pull/1927#discussion_r1152864538). To clarify, we can't control what kind of meta data is added / set per product. For example the data from yoast is saved as meta data `wpseo_global_identifier_values` and is a serialized array. While we don't necessarily need to support non strings we do need to ensure it doesn't result in a fatal if it can't be interpreted.

tests/Tools/HelperTrait/ProductTrait.php
```diff
@@ -472,4 +472,52 @@ protected function generate_mock_image_attachment( int $product_id, string $imag
...
+			[
+				'wc_product'     => $product,
+				'mapping_rules'  => $rules,
+				'gla_attributes' => [],
+				'targetCountry'  => 'US',
+			]
+		);
+	}
+
+	protected function generate_attribute_mapping_adapted_product_variant( $rules ) {
```
mikkamp, line 526: I'm seeing here a helper function to test product variation, but this always creates the same type of variable product (using global attributes).

This means we are missing the scenario where a variable product is created with custom attributes and what kind of results that has when we try to set mapping rules for this type of product.

We also don't get a variation where the attribute is set to `Any`.

src/Product/WCProductAdapter.php
```diff
@@ -1017,13 +1017,20 @@ protected function rule_match_conditions( array $rule ): bool {
 	 */
 	protected function get_product_taxonomy( $taxonomy ) {
 		$product = $this->get_wc_product();
-		$values  = get_the_terms( $product->get_id(), $taxonomy );
 
-		if ( ! $values ) {
+		if ( $product->is_type( 'variation' ) ) {
+			$values = $product->get_attribute( $taxonomy );
+			$values = explode( ', ', $values );
```
mikkamp, line 1029: I think I was partially wrong in my [previous comment](https://github.com/woocommerce/google-listings-and-ads/pull/1927#discussion_r1152837087). The product type `variation` has a different [`get_attribute` function](https://github.com/woocommerce/woocommerce/blob/7.6.1/plugins/woocommerce/includes/class-wc-product-variation.php#L135), which only returns a singular value, so the explode here isn't useful.

mikkamp (commented):
Thanks for the changes, although I think we are still missing some edge cases which the tests aren't covering. I've outlined those below and in the comments.

We still end up with an inconsistency between a simple product and a variation. If I use a global attribute (ex: pa_color) and then set multiple values (ex: Red + Green) for a simple product it picks the first value and sets the color as red.
When creating a variation we can either assign a specific value or we can select any (ex: Any Color) which would be equivalent to what we did for the simple product. However in such a scenario `get_attribute` returns `""` which means we don't set any value.

I'm also missing in tests where we set different types of product meta (like an array) and then try to map from that custom meta. A test like that would have caught the fatal error with explode.

sukafia:
We have a new issue: 6198523-zd-woothemes that I believe to be related. They're mapping brand attributes but it's not working:



The current code snippet only targets size and color, would we need the following code?: 


```
add_filter(
		'woocommerce_gla_product_attribute_value_brand',
		function( $source, $product ) {
			if ( ! $source && $product->get_type( 'variation') ) {
				return $product->get_attribute( 'pa_brand' );
			}

			return $source;
		}, 10, 2
	);
```

puntope:
> We have a new issue: 6198523-zd-woothemes that I believe to be related. They're mapping brand attributes but it's not working:

Yes, this has an impact on all variations using taxonomies sources. So also their custom `pa_brand` won't work. The snippet should fix the issue for now.

src/Product/WCProductAdapter.php
puntope: So these custom attributes are never taxonomies. These are the attributes saved on the fly in the product or any other custom field (like the ones you can set by checking Screen -> Custom fields) and should be loaded by the user using a filter.  In the last ones (custom fields) they are not returned by `get_attribute` that's why `get_meta` was in place.

Screenshot 2023-05-04 at 17 29 54

Screenshot 2023-05-04 at 17 54 42

Screenshot 2023-05-04 at 17 54 20

Example: 
```
add_filter(
		'woocommerce_gla_attribute_mapping_sources_custom_attributes',
		function( $values ) {
			return array_merge( $values, ['isbn', 'logo']);
		}
	);
```

The recommendation in the placeholder is to separate those by "|" which is transformed by WC into " | ". So we can try to do so in our code and also check for the meta being a scalar in order to prevent errors. You can see the adapted tests and the refactor in the last commits.

tests/Tools/HelperTrait/ProductTrait.php
puntope: I was checking this in detail. Seems like when setting "Any" in the product variant WC saves the attribute at product variant level as "".  Yes, it's kinda accessible from the parent.

I did some fixes in here https://github.com/woocommerce/google-listings-and-ads/pull/1927/commits/fe3ebb26a5f6baabdedc4dee8a15fd0e94a33551 and adding also some tests for it.

src/Product/WCProductAdapter.php
puntope: After changes in https://github.com/woocommerce/google-listings-and-ads/commit/fe3ebb26a5f6baabdedc4dee8a15fd0e94a33551  this is not applicable anymore.

puntope:
Howdy @mikkamp  I was checking your new suggestions and I did new tweaks in this PR. This is ready for another round of reviews.

src/Product/WCProductAdapter.php
```diff
@@ -1070,4 +1083,27 @@ protected function format_attribute( $value, $attribute_id ) {
...
+		if ( ! $attribute_value ) {
+			$attribute_value = $product->get_meta( $attribute_name );
+		}
+
+		// We only support scalar values.
+		if ( ! is_scalar( $attribute_value ) ) {
+			return '';
+		}
+
+		return explode( ' ' . WC_DELIMITER . ' ', strval( $attribute_value ) )[0];
```
mikkamp: Just a small nitpick. `strval` is still around because it's useful for when we want to convert a whole array to strings with `array_map`. But if we are just type casting an individual variable then it's preferred to use `(string)`
```suggestion
		return explode( ' ' . WC_DELIMITER . ' ', (string) $attribute_value )[0];
```

mikkamp: Since we are sharing this function for both `get_attribute` and `get_meta`. Only in the case of `get_attribute` will it add the surrounding spaces around `WC_DELIMITER`. Would it be better to handle that consistently?

```suggestion
		$values = explode( WC_DELIMITER, (string) $attribute_value );
		$values = array_filter( array_map( 'trim', $values ) );
		return empty( $values ) ? '' : $values[0];
```

mikkamp (approved):
When running the unit tests locally they pass when I run all the tests together, but when I filter to run just the tests in this class I'm getting 2 failures:
`vendor/bin/phpunit --filter=AttributeMappingWCProductAdapterTest`

Also when I run just one test individually:
`vendor/bin/phpunit --filter=AttributeMappingWCProductAdapterTest::test_maps_rules_product_fields_sku`

```
There were 2 failures:

1) Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Product\AttributeMapping\AttributeMappingWCProductAdapterTest::test_maps_rules_product_fields_sku
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'DUMMY SKU'
+''

2) Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Product\AttributeMapping\AttributeMappingWCProductAdapterTest::test_maps_rules_product_taxonomies
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'0'
+'21'
```

It seems the first one is because when it creates the product it triggers the WC rule that all SKU's must be unique so the product is created with an empty SKU.

```
WC_Data_Exception: Invalid or duplicated SKU.
```

In the second test failure it seems to be that after creating several test products it ends up with the attribute names being changed (didn't track down exactly where that was). But TIL that `--filter` is called after it loads data providers so it still runs everything with `@dataProvider` which means a whole bunch of test products get created before the filtered test is started.

So to filter a specific test it's better to use the specific file and then filter for a function:
`vendor/bin/phpunit tests/Unit/Product/AttributeMapping/AttributeMappingWCProductAdapterTest.php --filter=test_maps_rules_product_taxonomies`

After running that the individual tests pass.

The only part I'm still missing in the tests is single / multiple values set for custom attributes (not through product meta). I can't seem to find a helper like `WC_Helper_Product::create_product_attribute_object` but it can be easily replicated by doing something like this (maybe worth it to create a helper function):

```php
		$attribute = new \WC_Product_Attribute();
		$attribute->set_name( 'Custom Color' );
		$attribute->set_options( 'Red | Blue | Green' );
		$attribute->set_position( 1 );
		$attribute->set_visible( true );
		$attribute->set_variation( true );

		$attributes   = $variable->get_attributes();
		$attributes[] = $attribute;
		$variable->set_attributes( $attributes );
```

Besides that everything seems to be working as expected (and covered in the unit tests) so I'll go ahead and mark this PR as approved already.

Issue #1229, opened 2022-02-02
Shipping Integration - Store and return mapped Google locationIds


nima-karimi:
As part of #1228

To map the shipping rates for regions of a country such as a state, province, etc. to Merchant Center we need to map the stored WooCommerce IDs for those regions to a list of corresponding locationIds supported by Google.

To provide some background (more info available on #1228):

- `locationIds` is a property of the [LocationIdSet](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#LocationIdSet) object.  All location IDs in this list must be of the same type (state, etc.).
- The [LocationIdSet](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#LocationIdSet) object is set in the `locations` property of the [Headers](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Headers) class and it specifies the list of locations to which a shipping rate applies.
- The `Headers` object is used to create a table rate and it's set as part of a [Table](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#Table) property (`mainTable`) in the [RateGroup](https://developers.google.com/shopping-content/reference/rest/v2.1/shippingsettings#RateGroup) object.

Google provides the location IDs in a [downloadable CSV file](https://developers.google.com/adwords/api/docs/appendix/geotargeting) on the AdWords API documentation. This file is regularly updated.

At the moment, Merchant Center only supports state-level granularity for a handful of countries:

- **US**: postal code, state
- **AU**: postal code, territory
- **FR**: postal code, region
- **JP**: prefecture (shipping cost only)
- **All others**: country-level

WooCommerce stores the regions using the [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) format as key and their translated name as the value in the [states.php](https://github.com/woocommerce/woocommerce/blob/9d08ac8ac3723c897e08ac2ff8a6d950f0e26b5b/plugins/woocommerce/i18n/states.php) file, whereas Google stores their name, ID, and country-code in the provided [CSV file](https://developers.google.com/adwords/api/docs/appendix/geotargeting). Google API requires each location’s respective ID in the `locationIds` property when submitting the shipping settings. WooCommerce, for example, has the following entry for the United States:

```PHP
'US' => array( // United States.
        'AL' => __( 'Alabama', 'woocommerce' ),
        'AK' => __( 'Alaska', 'woocommerce' ),
[...]

PR #1939, opened 2023-04-04
Fix the fatal errors caused by adding any US Armed Forces location to WC Shipping setting

### LINKED ISSUE #1910: Review return value of $google_id in `src/Shipping/ZoneLocationsParser.php` - Null value causes fatal error in some cases
5997199-zen

Merchant entered some countries in their shipping zone: https://d.pr/i/ZUcZp2

Return value null from:

https://github.com/woocommerce/google-listings-and-ads/blob/592e0cb62bf6e7301506fd4285b141d052308986/src/Shipping/ZoneLocationsParser.php#L77

is causing a fatal error on step 3 of onboarding screen: 

```
2023-02-28T08:52:42+00:00 CRITICAL Uncaught TypeError: Argument 1 passed to Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ShippingLocation::__construct() must be of the type int, null given, called in /home2/a2knozmd/public_html/wp-content/plugins/google-listings-and-ads/src/Shipping/ZoneLocationsParser.php on line 78 and defined in /home2/a2knozmd/public_html/wp-content/plugins/google-listings-and-ads/src/Shipping/ShippingLocation.php:48
Stack trace:
#0 /home2/a2knozmd/public_html/wp-content/plugins/google-listings-and-ads/src/Shipping/ZoneLocationsParser.php(78): Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ShippingLocation->__construct(NULL, 'US', 'AA', NULL)
#1 /home2/a2knozmd/public_html/wp-content/plugins/google-listings-and-ads/src/Shipping/ShippingZone.php(144): Automattic\WooCommerce\GoogleListingsAndAds\Shipping\ZoneLocationsParser->parse(Object(WC_Shipping_Zone))
#2 /home2/a2knozmd/public_html/wp-content/ in /home2/a2knozmd/public_html/wp-content/plugins/google-listings-and-ads/src/Shipping/ShippingLocation.php on line 48
```

I am assuming the null value comes from here: 

https://github.com/woocommerce/google-listings-and-ads/blob/592e0cb62bf6e7301506fd4285b141d052308986/src/Google/GoogleHelper.php#L1335

There are other functions as well on the same file: `find_country_id_by_code()`  `find_subdivision_code_by_id()` that may return a null value for some countries.

I am not sure which country code is causing this issue here but I was able to replicate it on my test site by copying the same countries present in merchant's shipping zone.

eason9487:

Closes #1910 

- Fix a fatal error by checking `null` before creating `ShippingLocation` instance in the ZoneLocationsParser.php.
- Add a new case to ZoneLocationsParserTest.php to test `null` value returned by `find_subdivision_id_by_code`.

1. Go to the admin page: WooCommerce -> Settings -> Shipping -> Shipping zones
2. Add a shipping zone that includes 1-3 of the following locations and at least a shipping method:
   - Armed Forces (AA), United States (US)
   - Armed Forces (AE), United States (US)
   - Armed Forces (AP), United States (US)
   - For example:
   image
3. Make a request to the endpoint: **GET** `/wp-json/wc/gla/mc/target_audience/suggestions`
   - The request can be issued by running a script via the Console tab of browser DevTool:
      ```js
      wp.apiFetch({path: 'wc/gla/mc/target_audience/suggestions'}).then(console.log)
      ```
4. Check if the response is successful without fatal errors.
5. Check if the `countries` property of the response body doesn't include `"US"`.
6. Add a supported US location to the same shipping zone. For example, `California, United States (US)`.
7. Make another request and check if the `countries` property includes `"US"`.

### Additional details:

In #1229, the source of subdivisions was:

> Google provides the location IDs in a [downloadable CSV file](https://developers.google.com/adwords/api/docs/appendix/geotargeting) on the AdWords API documentation. This file is regularly updated.

After searching for `"US","State",Active` in the raw content of [the latest CSV file (2023-03-28)](https://developers.google.com/static/google-ads/api/reference/data/geo/geotargets-2023-03-28.csv), there are 51 locations, and they are the same set as the `COUNTRY_SUBDIVISIONS['US']` in the GoogleHelper.php file.

Thus, this PR doesn't need to update GoogleHelper.php.

### Changelog entry

> Fix - The fatal errors caused by adding any US Armed Forces location to WooCommerce Shipping setting

### DIFF
--- a/src/Shipping/ZoneLocationsParser.php
+++ b/src/Shipping/ZoneLocationsParser.php
@@ -74,8 +74,11 @@ class ZoneLocationsParser implements Service {
 
 					// Only add the state if the regional shipping is supported for the country.
 					if ( $this->google_helper->does_country_support_regional_shipping( $country ) ) {
-						$google_id                    = $this->google_helper->find_subdivision_id_by_code( $state, $country );
-						$locations[ $location->code ] = new ShippingLocation( $google_id, $country, $state, $region );
+						$google_id = $this->google_helper->find_subdivision_id_by_code( $state, $country );
+
+						if ( ! is_null( $google_id ) ) {
+							$locations[ $location->code ] = new ShippingLocation( $google_id, $country, $state, $region );
+						}
 					} else {
 						$google_id             = $this->google_helper->find_country_id_by_code( $country );
 						$locations[ $country ] = new ShippingLocation( $google_id, $country, null, $region );
--- a/tests/Unit/Shipping/ZoneLocationsParserTest.php
+++ b/tests/Unit/Shipping/ZoneLocationsParserTest.php
@@ -51,6 +51,36 @@ class ZoneLocationsParserTest extends UnitTest {
 		$this->assertEquals( 'NV', $parsed_locations[0]->get_state() );
 	}
 
+	public function test_returns_state_locations_if_subdivision_shipping_unsupported() {
+		$zone_locations = [
+			(object) [
+				'code' => 'US:AA',
+				'type' => 'state',
+			],
+		];
+
+		$zone = $this->createMock( WC_Shipping_Zone::class );
+		$zone->expects( $this->any() )
+			->method( 'get_zone_locations' )
+			->willReturn( $zone_locations );
+
+		$this->google_helper->expects( $this->any() )
+			->method( 'is_country_supported' )
+			->with( 'US' )
+			->willReturn( true );
+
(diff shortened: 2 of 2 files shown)


martynmjones (approved):
Hey @eason9487,

This fix resolves the fatal error, we're getting the expected response from the endpoint, and tests are passing so LGTM ✅

Thanks for working on this issue 🎉

PR #1940, opened 2023-04-06
Prevent creating assets for non PMax campaigns

### LINKED ISSUE #1931: Prevent creating assets for non PMax campaigns
## User story
When we create campaign a new campaign it will always be created as a PMax campaign. However we don't have a filter for other campaign types. So it could be possible a different campaign type is created directly in the Google Ads Dashboard.

We currently don't filter any of the campaign types so even if it's not a PMax campaign it will be visible in the list and contain an edit button. When proceeding to edit it will allow you to configure assets which don't work well.

In this case we will probably just want to remove the edit button for any campaigns that are not PMax. We also need to ensure the boost your campaign guide does not show up for any non PMax campaigns:
image

The endpoint to list campaigns already returns the `type` field which is set to `performance_max` for PMax Campaigns. So we should be able to implement a fix without needing any changes in the API calls.

eason9487:

Closes #1931 

To prevent creating assets for nonPMax campaigns, this PR

- Disable the "Edit" button for non-pmax campaigns on the Dashboard page.
- Only show the campaign assets tour when there is at least one pmax campaign.
- Redirect back to the Dashboard page when entering campaign editing page with a non-pmax campaign.

### Screenshots:

image

image

#### Reset tour flag

Reset tour flag by running the following code via the Console panel in the DevTool:

```js
wp.data.dispatch( 'wc/gla' ).upsertTour( { id: 'dashboard-feature--campaign-assets', checked: false } )
```

Or by running SQL:
```sql
DELETE FROM wp_options WHERE option_name = 'gla_tours';
```

#### Create a non-pmax campaign

1. Go to https://ads.google.com/
1. Create a new campaign
   image
1. Select "Standard Shopping campaign"
   image
1. Finish the rest steps.

#### Testing

1. Go to the Dashboard page.
2. Check if the "Edit" button of the non-pmax campaign is disabled.
3. Find out the ID of the non-pmax campaign and go to its editing page directly: `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Fcampaigns%2Fedit&programId={non-pmax-campaign-id}`
4. Check if the page redirects back to the Dashboard page.
5. Filter out all pmax campaigns to see if the tour is not shown when there are only non-pmax campaigns.
   - It would be easier to filter out by locally changing [this line of code](https://github.com/woocommerce/google-listings-and-ads/blob/1031dcd617cc258acb443883bfd504e97e8cafab/js/src/data/resolvers.js#L209) with:
   ```js
   adsCampaigns: campaigns.filter( c => c.type ==='shopping' ).map( adaptAdsCampaign ),
   ```
6. Switch the order of the campaign data to see if the tour shows on the first pmax campaign
   - It would be easier to filter out by locally sorting at [this line of code](https://github.com/woocommerce/google-listings-and-ads/blob/1031dcd617cc258acb443883bfd504e97e8cafab/js/src/data/resolvers.js#L205) with:
   ```js
   campaigns.sort( ( a, b ) => a.type ==='shopping' ? -1 : 1 )
   ```
### Changelog entry

> Fix - Prevent creating assets for non-Performance Max campaigns.

### DIFF
--- a/js/src/constants.js
+++ b/js/src/constants.js
@@ -70,6 +70,9 @@ export const CATEGORY_CONDITION_SELECT_TYPES = {
 
 export const CATEGORIES_TO_SHOW_IN_TOOLTIP = 5;
 
+// Campaign related
+export const CAMPAIGN_TYPE_PMAX = 'performance_max';
+
 // Assets of Performance Max Campaign related
 export const ASSET_KEY = {
 	BUSINESS_NAME: 'business_name',
--- a/js/src/dashboard/all-programs-table-card/edit-program-button/index.js
+++ b/js/src/dashboard/all-programs-table-card/edit-program-button/index.js
@@ -12,12 +12,16 @@ import EditProgramPromptModal from './edit-program-prompt-modal';
 import AppButtonModalTrigger from '.~/components/app-button-modal-trigger';
 
 const EditProgramButton = ( props ) => {
-	const { className, programId } = props;
+	const { className, programId, ...buttonProps } = props;
 
 	return (
 		<AppButtonModalTrigger
 			button={
-				<AppButton isLink className={ classnames( className ) }>
+				<AppButton
+					{ ...buttonProps }
+					isLink
+					className={ classnames( className ) }
+				>
 					{ __( 'Edit', 'google-listings-and-ads' ) }
 				</AppButton>
 			}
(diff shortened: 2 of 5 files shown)


puntope (approved):
✅  LGTM

- Edit button disabled on standard campaigns and enabled on PMAX 
- Redirect to Dashboard works when entering in the standard campaign
- Tour doesn't show if no PMAX campaigns
- Tour shows in the first PMAx campaign.

💅  One thing I miss are some unit tests.

eason9487:
Hi @puntope, thanks for the review. Added some basic tests and this PR-related test for the `AllProgramsTableCard` component by https://github.com/woocommerce/google-listings-and-ads/pull/1940/commits/d1099f50bf572d0386c3155b31dbca076415fc3f.

PR #1937, opened 2023-04-03
Unit test support for PHP 8.2

mikkamp:
### Changes proposed in this Pull Request:

This PR prepares support for running our unit tests in PHP 8.2. The main change in this PR is to get rid of the deprecation messages for creating dynamic properties. In our unit tests we've been following that pattern a lot by declaring them as properties within the doc block, but not as actual properties in the class. This PR changes that to move them from the docblock to actual declarations.

In addition to that one of the library packages `google/protobuf` was also generating a substantial amount of warnings. In the latest version they get around that by declaring `#[\AllowDynamicProperties]` in the Message base class. So I've upgraded the library to take advantage of this change:

`composer update google/protobuf --with-all-dependencies`
```
- Upgrading google/protobuf (v3.21.12 => v3.22.2)
```

In short the changes in this PR bring down the errors from 500+, to just 5 errors left when running the unit tests.

Two of the errors are coming from the same error that prevents us from running tests on PHP 8.1, see: https://github.com/woocommerce/google-listings-and-ads/issues/1758#issuecomment-1415897005
It has already been fixed but we are still waiting on a [2.13.2 release](https://github.com/googleapis/google-api-php-client/releases) to take advantage of the fix.

##### Update
2.13.2 has been released so I went ahead and updated the google apiclient package which reduces the error count to just 3.
`composer update google/apiclient --with-all-dependencies`
```
- Upgrading firebase/php-jwt (v6.3.2 => v6.4.0)
- Upgrading google/apiclient (v2.13.0 => v2.13.2)
- Upgrading google/apiclient-services (v0.286.0 => v0.295.0)
- Upgrading google/auth (v1.25.0 => v1.26.0)
- Upgrading guzzlehttp/psr7 (2.4.3 => 2.4.4)
- Upgrading phpseclib/phpseclib (3.0.18 => 3.0.19)
- Upgrading psr/http-client (1.0.1 => 1.0.2)
- Upgrading psr/http-factory (1.0.1 => 1.0.2)
- Upgrading psr/http-message (1.0.1 => 1.1)
```

The other three errors are coming from WooCommerce core, which has already been fixed, but at the moment the [PR is still a draft](https://github.com/woocommerce/woocommerce/pull/37063) and we would have to wait till it's included within a release.

Luckily enough all our code (which is covered by unit tests) did conform to PHP 8.2 standards, so we didn't need any changes there.

### Detailed test instructions:
1. Run `composer install` to upgrade composer packages
[...]

PR #1941, opened 2023-04-06
Report the Ads account currency in Woo marketing dashboard

nima-karimi:

Previously the current store currency was used for the ad campaigns when returning them to WooCommerce multichannel marketing dashboard. This was inaccurate because when the user changes their store currency, the amount should still be displayed in the currency of the ads account, not the store's currency.

This PR changes the MCM channel code to use the ads account currency instead of WooCommerce currency. I've also removed the dependency on the WC service since it's no longer used and modified the unit tests accordingly.

Related to the issue reported in internal testing: pb22l9-1WT-p2#comment-2066

Ensure you have the latest WooCommerce version and enable the new multichannel marketing feature in the Advanced section of WC settings (if you're using v 7.6+ it's enabled by default).

1. Connect a Google Ads account and create a campaign
2. Confirm that the currency displayed on the Campaigns card of the new marketing dashboard is the same as the ads account
3. Change your store's currency
4. Confirm step 2 again (the currency should not change)

### Changelog entry

> Fix - Use Ads account currency in the WooCommerce marketing dashboard

### DIFF
--- a/src/Internal/DependencyManagement/CoreServiceProvider.php
+++ b/src/Internal/DependencyManagement/CoreServiceProvider.php
@@ -411,7 +411,7 @@ class CoreServiceProvider extends AbstractServiceProvider {
 		$this->share_with_tags( AttributeMappingHelper::class );
 
 		if ( defined( 'WC_MCM_EXISTS' ) ) {
-			$this->share_with_tags( GLAChannel::class, MerchantCenterService::class, AdsCampaign::class, Ads::class, MerchantStatuses::class, ProductSyncStats::class, WC::class );
+			$this->share_with_tags( GLAChannel::class, MerchantCenterService::class, AdsCampaign::class, Ads::class, MerchantStatuses::class, ProductSyncStats::class );
 			$this->share_with_tags( MarketingChannelRegistrar::class, GLAChannel::class, WC::class );
 		}
 	}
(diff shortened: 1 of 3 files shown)


eason9487 (approved):
Thanks for the fix. Testing well. LGTM.

The only thing is that looks like the `use` and `@property` of `WC` could be removed from [GLAChannelTest.php](https://github.com/woocommerce/google-listings-and-ads/blob/a07bd5f97243b6be15e950a04ff46a8bfa7c8e12/tests/Unit/MultichannelMarketing/GLAChannelTest.php) as well. The recently merged #1937 changed almost all PHP unit tests:

> The main change in this PR is to get rid of the deprecation messages for creating dynamic properties. In our unit tests we've been following that pattern a lot by declaring them as properties within the doc block, but not as actual properties in the class.

It would be nice to sync the latest `develop` branch to this PR and remove the following lines of code before merging:

https://github.com/woocommerce/google-listings-and-ads/blob/354bd4cf8f4a1418610872976a6f488e44247c95/tests/Unit/MultichannelMarketing/GLAChannelTest.php#L16

https://github.com/woocommerce/google-listings-and-ads/blob/354bd4cf8f4a1418610872976a6f488e44247c95/tests/Unit/MultichannelMarketing/GLAChannelTest.php#L48-L49

Issue #1943, opened 2023-04-08
Php error


betayazilim:
Hi there,

I use latest WP and Woo.
Php 8.1

Keep getting this error message.
`Warning: Undefined array key 0 in /srv/htdocs/wp-content/plugins/google-listings-and-ads/src/Product/WCProductAdapter.php on line 1026`

puntope:
Hello @bsahins. Thanks for contacting us.

I wasn't able to reproduce the mentioned error. Potentially it could be some conflicts with other plugins as happened [here](https://github.com/woocommerce/google-listings-and-ads/issues/1935). 

I would advise you to [contact WooCommerce support](https://woocommerce.com/my-account/create-a-ticket/) for further help and investigation. Or you can also get [some tips about how to deal with conflicts here](https://woocommerce.com/document/how-to-get-help/)

PR #1945, opened 2023-04-11
Refactor to get the first element independently of its key

puntope:

Related to #1943 

Even tho this is likely not happening because Wp_Terms are ordered starting from 0. I added this fix to be sure we return the first element of the array independently of the key. 

That way we can be sure we can prevent the warning of `Warning: Undefined array key 0`

1. Create  a new attribute mapping rule with a taxonomy Color
2. See the attribute mapping rule working as normally. 

ℹ️  For not waiting the 30 minutes in order to sync the rules you can sync a specific product via connection test. 

### Changelog entry

> Fix - Prevent potential warning when accessing undefined index.

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1023,7 +1023,9 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			return null;
 		}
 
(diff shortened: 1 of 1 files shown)


puntope:
Fixed here https://github.com/woocommerce/google-listings-and-ads/pull/1927

PR #1950, opened 2023-04-17
[SEMGREP] - Prevent Semgrep warnings in add_query vars and in Scripts loader

puntope:

This PR prevents two issues in SEMGREP to warn in the output:

- Potential Unsafe usage (non escaped) of remove_query_arg() or add_query_arg(), which could lead to XSS           
- Argument of a function used in an include/require, could lead to File Inclusion

1. ⚠️ **Verify you can connect to Wordpress.com from the Extension Setup  in Step 1.** 
2. Verify that running SEMGREP against the extension doesn't generate this warnings anymore

### Additional details:

For running semgrep:

- Install semgrep
- clone rules --> `git clone git@github.com:Automattic/wpscan-semgrep-rules.git`
- cd wpscan-semgrep-rules
- download last release google-listings-and-ads zip somewhere and unzip it.
- Run --> semgrep --no-git-ignore --text -c audit {UNZIPPED_GLA_PATH} 
- (You can also run it agains your locally cloned repo, but that would include noise in regards vendor files)

### Changelog entry

> Dev - Fix SEMGREP warnings

### DIFF
--- a/src/Assets/ScriptWithBuiltDependenciesAsset.php
+++ b/src/Assets/ScriptWithBuiltDependenciesAsset.php
@@ -60,7 +60,7 @@ class ScriptWithBuiltDependenciesAsset extends ScriptAsset {
 				return $fallback;
 			}
 			// Reason of exclusion: These files are being loaded manually in the function call with no user input
-			// nosemgrep audit.php.lang.security.file.inclusion-arg
+			// nosemgrep: audit.php.lang.security.file.inclusion-arg
 			return new DependencyArray( include $build_dependency_path );
 		} catch ( Throwable $e ) {
 			do_action( 'woocommerce_gla_exception', $e, __METHOD__ );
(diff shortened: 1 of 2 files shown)


src/Assets/ScriptWithBuiltDependenciesAsset.php
```diff
@@ -59,7 +59,8 @@ protected function get_dependency_data( string $build_dependency_path, Dependenc
 			if ( ! is_readable( $build_dependency_path ) ) {
 				return $fallback;
 			}
-
+			// Reason of exclusion: These files are being loaded manually in the function call with no user input
+			// nosemgrep audit.php.lang.security.file.inclusion-arg
```
mikkamp: According to [the documentation](https://semgrep.dev/docs/ignoring-files-folders-code/) it should always have a colon after `nosemgrep` to specify a specific rule.

```suggestion
			// nosemgrep: audit.php.lang.security.file.inclusion-arg
```

Without the colon it just skips all rules. For example if I add `// nosemgrep foobar` I'd expect it to still detect the warning, and only skip a rule called `foobar` but it skips all checks for that block.

mikkamp (approved):
Thanks for the changes. I can confirm the Jetpack Connect URL remains the same regardless of adding `esc_url`. All parts of the URL are already properly escaped so it doesn't really do anything but `esc_url` is safe to call multiple times, so that's ok.

Just left one comment adding a colon to the ignore comment, it's important to add (in other PR's as well) since without it it disables all rules.

I'll go ahead and approve the PR already because once we add the colon it should all be good.

PR #1952, opened 2023-04-19
Prefix Psr Container Package

### LINKED ISSUE #1938: Please namespace the `psr/container` dependency
### Describe the bug:
On WordPress 6.1.1, installing 2.4.2 (both as an upgrade and from scratch) trigger a post-install error.

The fatal error shown is:
> Fatal error: Declaration of Automattic\WooCommerce\GoogleListingsAndAds\Container::has($id) must be compatible with Psr\Container\ContainerInterface::has(string $id): bool in /code/wp-content/plugins/google-listings-and-ads/src/Container.php on line 104 Notice: Function is_embed was called incorrectly. Conditional query tags do not work before the query is run. Before then, they always return false. Please see Debugging in WordPress for more information. (This message was added in version 3.1.0.) in /code/wp-includes/functions.php on line 5835 Notice: Function is_search was called incorrectly. Conditional query tags do not work before the query is run. Before then, they always return false. Please see Debugging in WordPress for more information. (This message was added in version 3.1.0.) in /code/wp-includes/functions.php on line 5835

### Steps to reproduce:
1. Install 2.4.2 via /wp-admin/ alongside another plugin using `psr/container` > 2.0.0 (such as https://github.com/publishpress/PublishPress-Revisions/releases/tag/v3.1.13)

### Expected behavior:
Plugin update runs smoothly.

### Actual behavior:
Depending on plugin load order, a fatal error prevents site from being used.

### Additional details:
As @mikkamp noted below, this problem is caused by multiple plugins loading `psr/container` without namespacing them, and the breaking change of `psr/container` 2.0 causes this crit error.

mikkamp:

This PR prefixes the Psr\Container package to ensure that we don't conflict with other plugins using version 2.x of the same package. I also updated the container packages to the latest compatible versions:

`composer update league/container --with-all-dependencies`
```
- Upgrading psr/container (1.1.1 => 1.1.2)
```

I did experiment with updating the container package to 2.x and league container 4.x however that requires some substantial changes, so I left it as is. In the newer version we can no longer share providers using just strings we need to provide an instance to the container.

Closes #1938 

1. Checkout the regular develop branch
2. Install the [PublishPress Revisions](https://wordpress.org/plugins/revisionary/) plugin
3. Notice a fatal error as soon as the plugin is activated
4. Check out this branch `fix/1938-prefix-psr-container-package` and re-run `composer install`
5. Activate the two plugins together and confirm that this time there are no errors

### Changelog entry
* Fix - Prefix Psr\Container package to prevent conflicts with other plugins.

### DIFF
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -17,6 +17,11 @@ declare( strict_types=1 );
  */
 $packages = [
 	[
+		'namespace' => 'Psr\\Container',
+		'package'   => 'psr/container',
+		'strict'    => false,
+	],
+	[
 		'namespace' => 'League\\Container',
 		'package'   => 'league/container',
 		'strict'    => false,
@@ -74,6 +79,9 @@ $dependencies = [
 		'google/apiclient',
 		'google/auth',
 		'google/gax',
+	],
+	'psr/container'    => [
+		'league/container',
 	],
 ];
 
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,10 @@
 		"google/apiclient": "^2.13",
 		"google/apiclient-services": "~0.286",
 		"googleads/google-ads-php": "^17.1",
-		"league/container": "^3.3",
+		"league/container": "^3.4",
 		"league/iso3166": "^4.1",
 		"phpseclib/bcmath_compat": "^2.0",
-		"psr/container": "^1.0",
+		"psr/container": "^1.1",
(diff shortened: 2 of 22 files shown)


puntope (approved):
✅  LGTM

- I verified the fatal error shown in develop when installing Revisionary plugin 
- I verified this is not happening anymore when using this PR
- I checked in the codebase for `Psr\Container` to see if there were some of them missing to replace.
- I did an Onboarding setup to verify the GLA plugin is working

PR #1954, opened 2023-04-20
Run PHP unit tests GH workflow with L-2 versions

puntope:

Implementing the new action get-releases-version in GLA. This action allows getting automatically the latest releases from WordPress dynamically.

### Screenshots:

Screenshot 2023-04-21 at 16 21 17

1. [Check the action fetching correctly the WP releases](https://github.com/woocommerce/google-listings-and-ads/pull/1954/checks?sha=d1afbd0acceb771eb1069ffe2b4360c0a05d3f79)

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -22,8 +22,21 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  GetMatrix:
+    name: Get WP version Matrix
+    runs-on: ubuntu-latest
+    outputs:
+      wp-matrix: ${{ steps.wp.outputs.matrix }}
+    steps:
+      - name: Get Release versions from Wordpress
+        id: wp
+        uses: woocommerce/grow/get-plugin-releases@actions-v1.5.0-pre
+        with:
+          slug: wordpress
+          
   UnitTests:
     name: PHP unit tests - PHP ${{ matrix.php }}, WP ${{ matrix.wp-version }}
+    needs: GetMatrix
     runs-on: ubuntu-latest
     env:
       WP_CORE_DIR: "/tmp/wordpress/src"
@@ -31,7 +44,7 @@ jobs:
     strategy:
       matrix:
         php: [8.0]
-        wp-version: ["6.0", "6.1", latest]
+        wp-version: ${{ needs.GetMatrix.outputs.wp-matrix }}
         include:
           - php: 7.4
             wp-version: latest
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -44,7 +44,7 @@ jobs:
     strategy:
       matrix:
         php: [8.0]
-        wp-version: ${{ needs.GetMatrix.outputs.wp-matrix }}
+        wp-version: ${{ fromJson(needs.GetMatrix.outputs.wp-matrix) }}
         include:
           - php: 7.4
             wp-version: latest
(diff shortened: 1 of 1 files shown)


mikkamp (commented):
This looks really good, thanks for the effort.

I just noticed though that there is a warning when it's run:
image

Maybe this PR is helpful for finding an alternative: https://github.com/woocommerce/grow/pull/48

I was also wondering how it would work if both WP and WC versions are retrieved maybe it would help to set them in environment variables like is done here: https://github.com/woocommerce/automatewoo/blob/develop/.github/workflows/php-unit-tests.yml#L70

puntope:
> This looks really good, thanks for the effort.
> 
> I just noticed though that there is a warning when it's run: image
> 
> Maybe this PR is helpful for finding an alternative: [woocommerce/grow#48](https://github.com/woocommerce/grow/pull/48)
> 
> I was also wondering how it would work if both WP and WC versions are retrieved maybe it would help to set them in environment variables like is done here: https://github.com/woocommerce/automatewoo/blob/develop/.github/workflows/php-unit-tests.yml#L70

Thanks, @mikkamp 

[According to GitHub 
](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#:~:text=Action%20authors%20who%20are%20using%20the%20toolkit%20should%20update%20the%20%40actions/core%20package%20to%20v1.10.0%20or%20greater%20to%20get%20the%20updated%20saveState%20and%20setOutput%20functions.)

> Action authors who are using the toolkit should update the @actions/core package to v1.10.0 or greater to get the updated saveState and setOutput functions.

So I would suggest handling this in a separate PR for updating the actions/core from 1.8 to 1.10 and test the change over all the JS actions that would be impacted by this. 

WDYT?

mikkamp:
> So I would suggest handling this in a separate PR for updating the actions/core from 1.8 to 1.10 and test the change over all the JS actions that would be impacted by this.

I'm not quite sure how that would work. The quote you mentioned seems to be part of the section "if you are using self hosted runners". I don't think that's the case, from my understanding the following was the paragraph that applied:

> Starting today runner version 2.298.2 will begin to warn you if you use the save-state or set-output commands via stdout. We are monitoring telemetry for the usage of these commands and plan to fully disable them on 31st May 2023. Starting 1st June 2023 workflows using save-state or set-output commands via stdout will fail with an error.

So it will start failing in about 2 weeks, so probably better to address sooner than later. From what I can see here it outputs the runner version:
image

puntope:
> > So I would suggest handling this in a separate PR for updating the actions/core from 1.8 to 1.10 and test the change over all the JS actions that would be impacted by this.
> 
> I'm not quite sure how that would work. The quote you mentioned seems to be part of the section "if you are using self hosted runners". I don't think that's the case, from my understanding the following was the paragraph that applied:
> 
> > Starting today runner version 2.298.2 will begin to warn you if you use the save-state or set-output commands via stdout. We are monitoring telemetry for the usage of these commands and plan to fully disable them on 31st May 2023. Starting 1st June 2023 workflows using save-state or set-output commands via stdout will fail with an error.
> 
> So it will start failing in about 2 weeks, so probably better to address sooner than later. From what I can see here it outputs the runner version: image

As far as I understand in the JS actions we use `setOutput` call via `core/actions` package.  This call seems to be responsible for giving the output from the JS to the action output itself.  

See for example --> https://github.com/woocommerce/grow/blob/trunk/packages/js/github-actions/actions/get-release-notes/src/get-release-notes.js#L39

As you can see[ here in this test](https://github.com/woocommerce/google-listings-and-ads/actions/runs/5016244293/jobs/8993540243), updating actions/core to 1.10.0 doesn't show a warning anymore.

mikkamp (approved):
Great happy to use whatever will resolve this issue. I'll go ahead and mark this PR as approved so we can move forward with the solution. I'll leave it up to you to decide how you want to address it, seems like there are still some followup points to resolve in the other PR.

PR #1962, opened 2023-05-10
Filters for manually mapping product IDs

layoutd:

This PR adds two filters that permit manual mapping of the WooCommerce product ID to the Merchant Center product ID (`woocommerce_gla_get_google_product_offer_id`),and vice versa (`woocommerce_gla_get_wc_product_id`). The functionality should allow merchants with existing catalogs on Merchant Center to specify which WooCommerce products correspond to which existing MC products, which will likely not have IDs that follow the GLA structure (`gla_1234` for product 1234).

**Note:** I've been poring over the code to see if there are any places where `product_id` or `mc_id` or `slug` are used in a context of WooCommerce product ID <=> Merchant Center product ID. Nothing has stuck out to me, but it would be nice to have another set of eyes, especially regarding reporting and Ads.

**WC to MC**
1. Fast test: 
    1. Add a quick filter to map Woo IDs to custom MC IDs (the actual product IDs don't matter):
	```
	add_filter( 'woocommerce_gla_get_google_product_offer_id', function( $mc_id, $woo_id ) {
		$woo_to_mc = [
			10 => 'test_id_1',
			20 => 'test_id_2',
			30 => 'gla_5',
		];
		if( is_numeric( $woo_id ) && isset( $woo_to_mc[$woo_id] ) ) {
			$mc_id = $woo_to_mc[$woo_id];
		}
		return $mc_id;
	}, 10, 2 );
	```

    2. Test some of the returned MC IDs using `WCProductAdapter::get_google_product_offer_id`:
	```
	add_action( 'init', function(){
		// Enable WP_DEBUG and WP_DEBUG_LOG
		foreach( [ 10, 20, 30, 40 ] as $id ) {
			$mapped_id = Automattic\WooCommerce\GoogleListingsAndAds\Product\WCProductAdapter::get_google_product_offer_id( 'gla', $id );
			error_log( "$id mapped to $mapped_id"  );
		}
		// Log should read:
		// 10 mapped to test_id_1
		// 20 mapped to test_id_2
		// 30 mapped to gla_5
		// 40 mapped to gla_40
	});
	```
2. Or, longer test:
    1. Add the filter listed in _1.i_, but using real product IDs
    2. Connect to a Merchant Center account and wait for product sync or force sync from the Test Connection page.
    3. Confirm that the mapped products have been created in Merchant Center with their custom IDs.

**MC to WC**
1. Add a filter to map custom MC IDs to Woo IDs (the product IDs don't matter):
	```
	add_filter( 'woocommerce_gla_get_wc_product_id', function( $mc_product_id, $prefix ) {
		$mc_to_woo = array_flip( [
			10 => 'test_id_1',
			20 => 'test_id_2',
			30 => 'gla_5',
		] );
		if ( isset( $mc_to_woo[$mc_product_id] ) ) {
			$mc_product_id = $prefix . $mc_to_woo[$mc_product_id];
		}
		return $mc_product_id;
	}, 10, 2 );
	```
    2. Test some of the returned Woo IDs using `ProductHelper::get_wc_product_id`:
    ```
	add_action( 'init', function() {
		// Enable WP_DEBUG and WP_DEBUG_LOG
		$product_helper = woogle_get_container()->get( Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductHelper::class );
		foreach ( [ 'test_id_1', 'test_id_2', 'gla_5', 'gla_40' ]  as $id ) {
			error_log( "$id mapped to " . $product_helper->get_wc_product_id( $id ) );
		}
		// Log should read:
		// test_id_1 mapped to 10
		// test_id_2 mapped to 20
		// gla_5 mapped to 30
		// gla_40 mapped to 40
	});
    ```

### Changelog entry

> Add - Filters for manually mapping product IDs.


src/Product/ProductHelper.php
```diff
@@ -254,6 +254,21 @@ public function get_synced_google_product_ids( WC_Product $product ): ?array {
...
+		 * product 5678 to Merchant Center product gla_1234.
+		 *
+		 * @param string $mc_product_id Merchant Center product ID.
+		 * @param string $prefix        The slug prefixed to generated Merchant Center product IDs.
+		 * @since x.x.x
+		 *
+		 * @return string Merchant Center product ID as normally generated by the plugin (e.g., gla_1234).
+		 */
+		$mc_product_id = apply_filters( 'woocommerce_gla_get_wc_product_id', $mc_product_id, $this->get_slug() . '_' );
+
```
mikkamp: The other filter is a mapping of `WC Product ID` > `Merchant Center ID`. So for simplicity I'd expect this to be the exact opposite, but instead it's mapping `Merchant Center ID` > `GLA_ID`. Wouldn't it make more sense to apply the filter later. Something like this:

```php
		$wc_product_id = 0;
		$pattern       = '/' . preg_quote( $this->get_slug(), '/' ) . '_(\d+)$/';
		if ( preg_match( $pattern, $mc_product_id, $matches ) ) {
			$wc_product_id = (int) $matches[1];
		}

		return (int) apply_filters( 'woocommerce_gla_get_wc_product_id', $wc_product_id, $mc_product_id );
```

The name of the filter also implies we are filtering a WC Product ID, so it would be a better match. And we no longer need to pass the prefix and append it in the filter we can just do a straight forward mapping.

mikkamp (commented):
Thanks for the changes, the filtering / mapping based on the test code is working. I just left one comment about switching the filter around.

I also tested the long way and can confirm that the mapping works fine to submit a product with a mapped ID. However the reverse mapping isn't fully working. In the reports it does because there it uses the short ID. However for the Product Feed page it gets the full Merchant Center ID. Ex: `online:en:GB:custom_2628_ID`
So it isn't able to match that to a mapped ID and leaves out the product issues for the mapped product.

I had to add the following to the mapping function to make sure it matches the right one in the array:

```php
$parts = explode( ':', $mc_id );
$mc_id = end( $parts );
```

Do you think that difference should be mentioned in the filter or should we handle that within `get_wc_product_id`? I'd say the latter is probably friendlier. If we do go with the latter some unit tests to cover it might be helpful to ensure we aren't breaking anything.

src/Product/ProductHelper.php
layoutd: Yeah, I went back and forth on this one. The symmetry of WC > MC and MC > WC seemed nicer to me, but I was also concerned with allowing a possible custom mapping to another GLA-generated ID (i.e. make product 500 sync to product 55, so 500 > gla_55) and somehow landed on the decision that it would be better to place it earlier. Thanks for calling it out, updated in 6f69d09d4ce82cd37898536760b240a73b6b3570

mikkamp: I actually think being able to map product 55 to 500 is an advantage. While probably not common, we could end up with a situation where someone migrates to a new site while exporting/importing products. Yet they still want to keep their original campaigns / products in MC.

layoutd: > I actually think being able to map product 55 to 500 is an advantage. 

Agreed, and it's still possible with the latest changes, just more logical altogether.

layoutd:
@mikkamp Thank for taking a look, I suspected your discerning eye would catch something I'd missed! I've updated the `woocommerce_gla_get_wc_product_id` filter parameters so that it accepts the simple MC product ID, and the `ProductHelper::get_wc_product_id()` method.
I also added some tests for both filters 👍🏻

mikkamp (approved):
Thanks for all the changes, the code looks good to me. 

I can confirm that it maps the ID when syncing:
```
2023-05-16T11:50:45+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\WCProductAdapter::map_product_categories Product category (ID: 2628): ["Uncategorized"].
2023-05-16T11:50:46+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer::update_by_batch_requests Submitted 1 products:
[{"woocommerce_id":2628,"google_id":"online:en:GB:custom_2628_ID"}]
```

And when I get the list of product issues it populates the DB table so querying the mapped product returns some results:
`SELECT * FROM wp_gla_merchant_issues WHERE product_id=2628;`

layoutd:
Thanks for taking a look, @mikkamp 👍🏻

PR #1977, opened 2023-06-08
Prefix Psr Http Message Package

### LINKED ISSUE #1976: Fatal error when using GLA 2.4.7 and flexible-shipping-ups plugin
### Describe the bug:

Reported in p1686202399910239-slack-C7U3Y3VMY and originally from the [support forum](https://wordpress.org/support/topic/plugin-causes-the-cart-and-checkout-to-completely-crash-after-todays-update/).

GLA 2.4.7 uses `psr/http-message` `2.0` while the [flexible-shipping-ups](https://wordpress.org/plugins/flexible-shipping-ups/) plugin uses `1.1`. In `2.0` it adds function return types in the class interface and in GLA we didn't add prefix to `psr/http-message` library so the fatal error happened.

Here's the error:

```
2023-06-07T11:48:32+00:00 CRITICAL Declaration of UpsFreeVendor\GuzzleHttp\Psr7\Uri::getScheme() must be compatible with Psr\Http\Message\UriInterface::getScheme(): string in /home/795922.cloudwaysapps.com/axqxgxgnqn/public_html/wp-content/plugins/flexible-shipping-ups/vendor_prefixed/guzzlehttp/psr7/src/Uri.php on line 346
```

### Steps to reproduce:

~Note that I personally couldn't reproduce the error as I did not set up the UPS plugin properly. But from the log the root cause is obvious so we should add the prefix to `psr/http-message` library.~

1. Install GLA 2.4.7 and [flexible-shipping-ups](https://wordpress.org/plugins/flexible-shipping-ups/) plugin.
2. Go to `wp-admin/admin.php?page=wc-settings&tab=shipping`, add `UPS Live Rates` shipping method to the shipping zone you're testing.
3. Go to cart and checkout page, notice the fatal error.

### Expected behavior:
There shouldn't be fatal error.

### Actual behavior:

The fatal error appears and the site crashes.

ianlin:

This PR prefixes the `Psr\Http\Message` package to ensure that we don't conflict with other plugins using version < 2 of the same package.

Closes #1976.

1. Install GLA 2.4.7 and [flexible-shipping-ups](https://wordpress.org/plugins/flexible-shipping-ups/) plugin.
2. Go to `wp-admin/admin.php?page=wc-settings&tab=shipping`, add `UPS Live Rates` shipping method to the shipping zone you're testing.
3. Go to cart and checkout page, notice the fatal error. 
4. Checkout this branch `fix/1976-prefix-psr-http-message-package`, remove the `vendor/` directory, and run `composer install`
5. Go to the cart/checkout page again and confirm there are no errors.

### Changelog entry

> Fix - Prefix Psr\Http\Message package to prevent conflicts with other plugins.

### DIFF
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -88,7 +88,7 @@ $dependencies = [
 	'psr/container'    => [
 		'league/container',
 	],
-	'psr/http-message'    => [
+	'psr/http-message' => [
 		'firebase/php-jwt',
 		'google/apiclient',
 		'google/auth',
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -22,6 +22,11 @@ $packages = [
 		'strict'    => false,
 	],
 	[
+		'namespace' => 'Psr\\Http\\Message',
+		'package'   => 'psr/http-factory',
+		'strict'    => false,
+	],
+	[
 		'namespace' => 'Psr\\Container',
 		'package'   => 'psr/container',
 		'strict'    => false,
@@ -89,14 +94,16 @@ $dependencies = [
 		'league/container',
 	],
 	'psr/http-message' => [
-		'firebase/php-jwt',
 		'google/apiclient',
 		'google/auth',
 		'google/gax',
 		'guzzlehttp/guzzle',
 		'guzzlehttp/psr7',
 		'psr/http-client',
-		'psr/http-factory',
+	],
+	'psr/http-factory' => [
+		'firebase/php-jwt',
+		'guzzlehttp/psr7',
 	],
 ];
 
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -17,6 +17,11 @@ declare( strict_types=1 );
  */
 $packages = [
 	[
+		'namespace' => 'Psr\\Http\\Message',
+		'package'   => 'psr/http-message',
+		'strict'    => false,
+	],
+	[
 		'namespace' => 'Psr\\Container',
 		'package'   => 'psr/container',
 		'strict'    => false,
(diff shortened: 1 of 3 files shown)


ianlin:
Oops forgot to test the linting locally, will updated shortly 😅

bin/prefix-vendor-namespace.php
```diff
@@ -83,6 +88,16 @@
...
 	],
+	'psr/http-message' => [
+		'firebase/php-jwt',
+		'google/apiclient',
+		'google/auth',
+		'google/gax',
+		'guzzlehttp/guzzle',
+		'guzzlehttp/psr7',
+		'psr/http-client',
+		'psr/http-factory',
```
mikkamp: I think we need to prefix both `psr/http-message` and `psr/http-factory`. But I don't think this belongs as a dependency. 

We aren't using any of the factory classes directly so it doesn't seem to break anything. But since they share a namespace it's causing some replacements which could cause failures (if there are other plugins using those factory classes).

For example in `vendor/guzzlehttp/psr7/src/HttpFactory.php` we are prefixing all instances of `Psr\Http\Message`. But for the class `Psr\Http\Message\RequestFactoryInterface`, the class can be found in `vendor/psr/http-factory/src/RequestFactoryInterface.php` but doesn't use a prefixed namespace. Same thing for other factory interfaces.

mikkamp (commented):
Thanks for the changes I can confirm that the fatal no longer occurs after prefixing the package.

As mentioned though I think we need to move `psr/http-factory` from a dependency to a package we are prefixing since it shares a namespace with `psr/http-message`.

ianlin:
Thanks @mikkamp, did not notice `psr/http-factory` shares the same namespace with `psr/http-message`. Updated in 
c9e3d71.

mikkamp (approved):
Thanks for the extra changes, it's working great now.

PR #1989, opened 2023-06-15
Enable unit testing for PHP 8.1

### LINKED ISSUE #1758: Update phpunit for all tests and add testing for PHP 8.1
## User story

Previously we've stuck with phpunit 7.5.20 because we needed to remain compatible with older WordPress / WooCommerce tests. We use the `yoast/phpunit-polyfills` package to still be able to take advantage of newer functions introduced for unit testing.

However for testing with PHP 8.0 we use a [workaround and upgrade to phpunit 9.5](https://github.com/woocommerce/google-listings-and-ads/blob/2.2.0/.github/workflows/php-unit-tests.yml#L54-L56), with the limitation that it was unable to test with a WP version older than 5.9. Now that WP 6.1 is released, if we support WP versions L-2 we can run unit testing for 5.9 - 6.1 without needing to check older versions. This means we can upgrade the phpunit for all tests and no longer need the workaround.

Doing so will also enable us to setup unit testing for PHP 8.1, which could help identify issues like the one reported in #1745 

## Technical
- [x] Update the phpunit composer package
- [x] Remove workaround for PHP 8.0
- [ ] Add phpunit testing for PHP 8.1
- [x] Update all documentation referring to unit testing

mikkamp:

We had one [pending point](https://github.com/woocommerce/google-listings-and-ads/issues/1758#issuecomment-1415897005) to enable unit testing for PHP 8.1. The package has been updated so this PR enables unit testing for PHP 8.1 in GitHub actions.

> **Note**
Enabling it for PHP 8.2 is still too premature as there are several warning still coming from WooCommerce:
```
There were 3 errors:

1) Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers\MerchantCenter\ContactInformationControllerTest::test_register_route
Creation of dynamic property WC_Email_New_Order::$email_type is deprecated

2) Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Shipping\SyncerHooksTest::test_deleting_method_from_shipping_zone_schedules_update_job
Creation of dynamic property WC_Shipping_Free_Shipping::$ignore_discounts is deprecated

3) Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Shipping\SyncerHooksTest::test_updating_method_options_schedules_update_job
Creation of dynamic property WC_Shipping_Free_Shipping::$ignore_discounts is deprecated
```

Closes #1758 

1. Confirm unit tests in this PR runs for PHP versions 7.4, 8.0 (various WP versions), and PHP 8.1 

### Changelog entry
* Dev - Enable unit testing for PHP 8.1.

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -49,6 +49,8 @@ jobs:
         include:
           - php: 7.4
             wp-version: ${{ needs.GetMatrix.outputs.latest-wp-version }}
+          - php: 8.1
+            wp-version: ${{ needs.GetMatrix.outputs.latest-wp-version }}
 
     steps:
       - name: Checkout repository


eason9487 (approved):
LGTM.

Issue #1993, opened 2023-06-20
Improve the error handling and prompts in the onboarding


eason9487:
## Goals

During the onboarding

- Prompts for explicit errors such as incorrect input formatting or missing required fields to reduce confusion for users
- Make error messages more specific as to which part of the error occurred

The main scope is the Onboarding flow and the Edit Free Listings page. (The editing page basically reuses most of the code)

### 📌 <a href="#tasks-error-handling" id="tasks-error-handling">Deal with technical error handling</a>

```[tasklist]
- [x] https://github.com/woocommerce/google-listings-and-ads/issues/489
- [x] https://github.com/woocommerce/google-listings-and-ads/issues/1331
- [x] https://github.com/woocommerce/google-listings-and-ads/issues/1246
```

These tasks are more about technical adjustments. It expects to use the current UI and will not change significantly.

### 📌 <a href="#tasks-display-validation-errors" id="tasks-display-validation-errors">Display form validation errors</a>

```[tasklist]
- [x] https://github.com/woocommerce/google-listings-and-ads/issues/634
```

This is the top priority in this project as there were a lot of user reports related to it.

To display the form/data validation errors, the main points of the change are:

1. Change the timing of triggering form validation and showing errors.
   - Don't disable the "Continue" or "Compete" buttons even if the form validation doesn't pass.
   - The form validation errors are not displayed until the user clicks the above button.
   - After displaying the validation errors, the errors will continue to be updated with user input.
2. Similar to the Asset Group management page. The error messages are grouped by `Card` UI and are displayed below the UI layout they belong to.
   - For example:
      ![image](https://github.com/woocommerce/google-listings-and-ads/assets/17420811/854af97f-047f-46d8-b65c-d0cf6263dfdc)
3. The # issues pill UI won't be added because the forms other than Asset Group don't have the collapse and expanding interaction.
4. These changes won't be applied to the modals of shipping setup as the modal should be clear enough and no user seems to have reported a problem with this.
   - Shipping modal:
      ![image](https://github.com/woocommerce/google-listings-and-ads/assets/17420811/2df5c4d5-7d33-4095-aea4-3229f3073ffb)
5. To be consistent, the event tracking of the above buttons will still be triggered only when the step can be proceeded to the next one or be completed.

### 📌 <a href="#tasks-clarify-api-errors" id="tasks-clarify-api-errors">Clarify API error messages</a>

```[tasklist]
[...]

PR #1598, opened 2022-07-14
Propagate errors for saveSettings action

jorgemd24:
### Changes proposed in this Pull Request:

This is a follow-up of https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1109566842 , https://github.com/woocommerce/google-listings-and-ads/issues/1331#issuecomment-1110084066 & https://github.com/woocommerce/google-listings-and-ads/pull/1586

This PR removes the try and catch from `saveSettings` action, allowing the error to bubble up and be handled by the consumer. if `saveSettings` were failing, the consumer was not able to tell if `saveSettings` throw an exception because the error was already caught in the action. Using this PR, `saveSettings` will throw an error and the error can be handled. 

I found four places where `saveSettings` was used:

- `handleSubmitCallback` - Step 3 (Configure your product listings).

Before this PR the `saveSettings` could fail and allow to continue to step 4 without saving rates, times or taxes **types**. With this PR if `saveSettings` fails it will not allow to continue as the data has not been saved correctly. 

https://github.com/woocommerce/google-listings-and-ads/blob/0614caa5a6227e5e41b3ae15b3cc7df6442deb57/js/src/setup-mc/setup-stepper/setup-free-listings/index.js#L91-L97

Before this PR:

https://user-images.githubusercontent.com/2488994/179293546-3cc15421-71f5-4dc6-be62-d0a951651e4b.mp4

After this PR

https://user-images.githubusercontent.com/2488994/179300591-b86882ce-7c0f-49b4-8bfe-abe0a7fb08ce.mov

- `handleChangeCallback` - Step 4 (Confirm store requirements). Before this PR, checking the store requirements could fail and allow the user to complete the process. 

https://github.com/woocommerce/google-listings-and-ads/blob/0614caa5a6227e5e41b3ae15b3cc7df6442deb57/js/src/setup-mc/setup-stepper/store-requirements/index.js#L41-L43

Before this PR:

https://user-images.githubusercontent.com/2488994/179303214-78a70c2b-ce41-4ac4-84de-f4759748047f.mp4

After this PR:

https://user-images.githubusercontent.com/2488994/179303657-e02266ab-f34b-4ed3-b019-b5da98c8342a.mp4

- `useAutoSaveSettingsEffect`,    this PR adds the `try` and `catch` in the autosaving function.

[...]

PR #1994, opened 2023-06-21
Avoid continuing to sync settings with GMC after shipping time API failed

### LINKED ISSUE #489: Consider different way to handle errors for upsert & delete shipping rates/times
_This is a follow-up from https://github.com/woocommerce/google-listings-and-ads/pull/468#discussion_r613085411_

### Describe the bug:
Currently, all fetch, upsert, delete calls for shipping rates and times handles the errors internally in [`.~/data/actions.js`](https://github.com/woocommerce/google-listings-and-ads/blob/trunk/js/src/data/actions.js)

That makes, aggregating those errors or handling them in a use-case-specific way, tricky. I see two problems here:
1. Due to `try...catch` the promise for the rejected request is resolved as fulfilled.
	That could be somewhat mitigated in the consuming code by inspecting the value with which the promise was resolved. Probably doable, but looks fragile and hacky.
2. The notice is already rendered by the action, so there is nothing the consuming code could do to prevent, block or avoid it.

### Expected behavior:
When I call `upsertShippingRates` or similar action, I'd like to be able to know whether it was successful or not and to act/affect UI upon the potential errors in the way I specify for my use-case.

### Actual behavior:
Accessing the information about the failure is cumbersome, there is no way to avoid notification being rendered.

### Steps to reproduce:
1. Call `upsertAction = upsertShippingRate( { countryCode: 'PL', currency: 'PLN', rate: 'bogus' } );`
2. Try acting on rejections either with `try ... catch` or `upsertAction.catch(...)`

### Additional details:
1. While saving a number of different shipping rates & times, settings, and target audience, we would like to aggregate the information about all the changes, then show messages like:
	- "All changes were saved successfully."
	- "There was an error trying to save shipping rates for Poland and Australia. Other changes have been saved."
	- "Your changes were not saved. Please try again."
2. The new solution may be explored in a new action that would be useful in the above case anyway. An action to save a bulk of `AggregatesShippingRate`s. So we would limit the risk of regression issues in onboarding MC Setup.

### LINKED ISSUE #1331: Edit Free Listings: Success notification is shown when there is an API error and saving is not successful

eason9487:

Closes #1331
Closes #489

This PR solves two tasks of 📌 [Deal with technical error handling](https://github.com/woocommerce/google-listings-and-ads/issues/1993#tasks-error-handling) in #1993.

- Fixes the issue of continuing to sync settings with GMC after shipping time API failed, like the following screenshot.
- Extra change: Update JSDoc of shipping rates related actions and hooks to indicate they may throw an error.

image

#### Summary for #1331

In order to solve #1331, it needs to:
- Throw out errors from shipping rate actions ([#1313](https://github.com/woocommerce/google-listings-and-ads/pull/1313))
- Throw out errors from the `saveTargetAudience` action ([#1586](https://github.com/woocommerce/google-listings-and-ads/pull/1586))
- Throw out errors from the `saveSettings ` action ([#1598](https://github.com/woocommerce/google-listings-and-ads/pull/1598))
- Hande multiple errors for failed API calls ([#1602](https://github.com/woocommerce/google-listings-and-ads/pull/1602))
- This PR is the last piece - Throw out errors from shipping time actions

#### Summary for #489

In order to handle errors from the outside of **wp-data** actions

- [#1313](https://github.com/woocommerce/google-listings-and-ads/pull/1313) removed the `try...catch` block from a shipping rate action
- This PR removes the `try...catch` blocks from shipping time actions

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/10fb6bd3-6418-4eaf-8acf-05ad865ed84c

To trigger API failures, one of the easier ways is to change the API endpoint(s) in actions.js:

https://github.com/woocommerce/google-listings-and-ads/blob/f1775c2696ee930595ea9a33ba116c4234c90a1c/js/src/data/actions.js#L265

https://github.com/woocommerce/google-listings-and-ads/blob/f1775c2696ee930595ea9a33ba116c4234c90a1c/js/src/data/actions.js#L287

and rebuild the client side JS files.

2. Make some changes to the shipping time setting.
3. Check if an error message is displayed via a notification at the bottom left of the page.
4. Go to the free listings editing page.
5. Make some changes to the shipping time setting and save changes.
6. Check if an error message is displayed via a notification at the bottom left of the page and if there is no success notification at the same time.

### Changelog entry

> Fix - Avoid continuing to save settings to Google Merchant Center after the shipping time save failed on the Edit Free Listings page.

### DIFF
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -254,63 +254,45 @@ export function* fetchShippingTimes() {
  * Updates or inserts given aggregated shipping rate.
  *
  * @param {AggregatedShippingTime} shippingTime
+ * @throws Will throw an error if the request failed.
  */
 export function* upsertShippingTimes( shippingTime ) {
 	const { countryCodes, time } = shippingTime;
 
-	try {
-		yield apiFetch( {
-			path: `${ API_NAMESPACE }/mc/shipping/times/batch`,
-			method: 'POST',
-			data: {
-				country_codes: countryCodes,
-				time,
-			},
-		} );
+	yield apiFetch( {
+		path: `${ API_NAMESPACE }/mc/shipping/times/batch`,
+		method: 'POST',
+		data: {
+			country_codes: countryCodes,
+			time,
+		},
+	} );
 
-		return {
-			type: TYPES.UPSERT_SHIPPING_TIMES,
-			shippingTime,
-		};
-	} catch ( error ) {
-		yield handleFetchError(
-			error,
-			__(
-				'There was an error trying to add / update shipping times. Please try again later.',
-				'google-listings-and-ads'
-			)
-		);
-	}
+	return {
+		type: TYPES.UPSERT_SHIPPING_TIMES,
+		shippingTime,
+	};
 }
 
 /**
  * Deletes shipping times associated with given country codes.
  *
  * @param {Array<CountryCode>} countryCodes
+ * @throws Will throw an error if the request failed.
  */
 export function* deleteShippingTimes( countryCodes ) {
-	try {
-		yield apiFetch( {
-			path: `${ API_NAMESPACE }/mc/shipping/times/batch`,
-			method: 'DELETE',
-			data: {
-				country_codes: countryCodes,
-			},
-		} );
+	yield apiFetch( {
+		path: `${ API_NAMESPACE }/mc/shipping/times/batch`,
+		method: 'DELETE',
+		data: {
+			country_codes: countryCodes,
+		},
+	} );
 
-		return {
-			type: TYPES.DELETE_SHIPPING_TIMES,
-			countryCodes,
-		};
-	} catch ( error ) {
-		yield handleFetchError(
-			error,
-			__(
-				'There was an error trying to delete shipping times. Please try again later.',
-				'google-listings-and-ads'
-			)
-		);
-	}
+	return {
+		type: TYPES.DELETE_SHIPPING_TIMES,
+		countryCodes,
+	};
 }
 
 export function* fetchSettings() {
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -158,6 +158,7 @@ export function* fetchShippingRates() {
  *
  * @param {Array<ShippingRate>} shippingRates Shipping rates to be upserted.
  * @return {Object} Action object to update shipping rates.
+ * @throws Will throw an error if the request failed.
  */
 export function* upsertShippingRates( shippingRates ) {
 	const data = yield apiFetch( {
@@ -186,6 +187,7 @@ export function* upsertShippingRates( shippingRates ) {
  *
  * @param {Array<string>} ids IDs of shiping rates to be deleted.
  * @return {Object} Action object to delete shipping rates.
+ * @throws Will throw an error if the request failed.
  */
 export function* deleteShippingRates( ids ) {
 	yield apiFetch( {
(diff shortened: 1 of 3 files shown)


puntope (approved):
LGTM ✅  I can see the error prompts being triggered when the API fails.

The rest of the data is saved correctly. I tested locally in Edit and also on New Free Listings in the Onboarding. 

One thing I noticed is that in Onboarding, in the step if creating Free Listings. It says "There was an error saving your Shipping times" However it still allows me to continue with the form and finish the onboarding. Then, when I finish and it redirects me to Dashboard, it shows an error "An error occurred while retrieving your product feed. Please try again later."

Just pointing this out since I didn't see anything related in the task presented in https://github.com/woocommerce/google-listings-and-ads/issues/1993

puntope:
p.s Another thing I missed was the presence of some unit tests testing the exception flow for the API calls.

eason9487:

> One thing I noticed is that in Onboarding, in the step if creating Free Listings. It says "There was an error saving your Shipping times" However it still allows me to continue with the form and finish the onboarding. Then, when I finish and it redirects me to Dashboard, it shows an error "An error occurred while retrieving your product feed. Please try again later."
> 
> Just pointing this out since I didn't see anything related in the task presented in #1993

This issue has probably existed for a while, but it doesn't appear to have been reported. I guess it's mainly because this part has a low probability of API errors or is easily fixed by users themselves.

Fixing this issue might involve changing the current mechanism:
1. Data is stored automatically when `onBlur` or `onChange` event is triggered.
2. There is no clear way for users to retry the API request after an error occurs.

I opened a follow-up issue #1997. #1993 may change the way for disabling "Continue" buttons, so #1997 may be considered later on, but it's still unclear.

> p.s Another thing I missed was the presence of some unit tests testing the exception flow for the API calls.

Taking into account the complexity of the onboarding flow and the current implementation, a large number of modules must be mocked if the unit test is to be added directly, the scope of which highlights that the current implementation is not easy to write tests.

If we want to make the implementation more testable, the workload has reached a scale that can be an independent project, so I won't add tests to the high-level component units in this project when working on it for now.

PR #2000, opened 2023-06-26
Replace undefined $border-width-focus var

puntope:

Remove an undefined CSS variable that once existed in a Gutenberg package.

When working on setting the npm engines in the repo I found an issue with an undefined CSS variable: $border-width-focus. 

Apparently, this variable once existed in the Gutenberg code but doesn't seem to any more. I think it was [renamed to](https://github.com/WordPress/gutenberg/blob/b3ea2b6a80ee92f26148899455669c53ae2f49de/packages/base-styles/_variables.scss#L87) $border-width-focus-fallback and its recommended to use `--wp-admin-border-width-focus`.

1. See the error compiling the app with `npm run start` (might need to remove `package-lock.json` first, and run `npm i`).
2. Checkout this PR and see the error not happening
3. See Nav Tabs in GLA plugin looking good 

### Additional details:

ℹ️  I was thinking to replace it with `$border-width-focus-fallback` or just `2px` but since it's the written recommendation I guess it's better `--wp-admin-border-width-focus` 

### Changelog entry

> Fix - Remove deprecated $border-width-focus variable.

### DIFF
--- a/js/src/components/app-tab-nav/index.scss
+++ b/js/src/components/app-tab-nav/index.scss
@@ -29,12 +29,12 @@
 		}
 
 		&:focus:not(:disabled) {
-			box-shadow: inset 0 $border-width-focus var(--wp-admin-theme-color);
+			box-shadow: inset 0 var(--wp-admin-border-width-focus) var(--wp-admin-theme-color);
 		}
 
 		&.is-active {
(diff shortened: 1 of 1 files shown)


ianlin (approved):
Thanks for fixing this, I could reproduce the error and the error is gone using this PR, LGTM.

PR #411, opened 2021-04-01
Add batch upsert/delete actions and reducer for shipping rates/times to communicate with batch APIs

eason9487:
### Changes proposed in this Pull Request:

This PR implemented a part of issue #358 and depends on #410. It needs to change the shared data lib that corresponding to others' work, so it's presented separately to facilitate collaboration.

### Detailed test instructions:

1. Head to the MC onboarding flow: `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-mc`
2. The upsert/deletion of shipping rates and times should work as before

❗  Please note that the migrations of the consuming side aren't included in this PR, so the UI may still appear bouncy problems and too many connection errors.

PR #2003, opened 2023-06-28
Add the built-in WP REST API validation to `CountryCodeTrait` to avoid errors when clearing all audience countries in the onboarding flow

### LINKED ISSUE #1855: Audience Auto-save shows error when selector is empty.
### Describe the bug:

When clearing the audience selector. The auto-save runs with empty target countries and shows an error in the screen that could be confusing for the users. This is getter ing worse if you select All countries right after.

### Steps to reproduce:

1. Go to onboarding Step 2
2. Clear the country selector 
3. See error
4. Choose all countries 
5. See error

### Expected behavior:
- If selector is empty. Auto-save shouldn't run. Or it should show a more custom error like "Please select at least one country"
- If selected All countires after empty selector. Error shouldn't show. 

### Actual behavior:
- Autosaver runs and an error appears. 

### Additional details:

https://user-images.githubusercontent.com/5908855/212833050-70209d2f-9c14-4e18-a2f0-eaf76632833e.mov

eason9487:

This PR solves a task of 📌 [Clarify API error messages](https://github.com/woocommerce/google-listings-and-ads/issues/1993#tasks-clarify-api-errors) in #1993

Closes #1855

Originally, clearing the recipient country did not cause API errors. After #1327, the POST `/mc/target_audience` API was also validated by [this check](https://github.com/woocommerce/google-listings-and-ads/blob/3cc6936fc2938952f120f8c0478fccfaf2ba9eac/src/API/Site/Controllers/CountryCodeTrait.php#L52-L54), which causes the auto-saving behavior in the onboarding flow to get errors.

- Add the built-in validation of WP REST API for the country codes in `CountryCodeTrait`.
- Change to use the parameter's schema to validate if the country codes array is not empty for needed Controllers:
   - `BudgetRecommendationController`
   - `CampaignController`
- Add or adjust tests for the related Controllers:
   - Add basic tests for the `TargetAudienceController` and `ShippingTimeBatchController`.
   - Add the test case of empty country codes for the `BudgetRecommendationController`
   - Remove an unused variable from the `test_create_campaign_empty_targeted_locations` in CampaignControllerTest.php

📌 Please take a look at the **Additional details** for contexts and other solutions.

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/16973d6c-388c-4081-a4c5-25905a88a1ca

2. Remove all audience countries.
3. Check if there is no API error.

#### About the validation mechanism of WP REST API

When extending the WP REST API, [the mechanism of `validate_callback` and `sanitize_callback`](https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#api) is somewhat surprising. In summary, the findings are:
1. If the `sanitize_callback` is not specified but `type` is specified, both built-in `validate_callback` and `sanitize_callback` will be applied no matter whether the `validate_callback` is a custom one.
2. If both `validate_callback` and `sanitize_callback` are specified with custom callbacks, the built-in schema validation won't be applied.
3. The built-in `validate_callback` may be applied two times if not specify both callbacks.
4. When the built-in `validate_callback` is skipped, the `required` specified in `args` will still work. This makes the validation options of `args` work or not, resulting in more unexpected inconsistencies.
5. Looks like there is no way to specify in `args` that when a custom `validate_callback` is specified, still uses the built-in schema validation.

This can be a bit confusing to use because the main goal of `sanitize_callback` doesn't seem to be data validation, but whether the built-in schema validation works in the end depends on `sanitize_callback` and not just `validate_callback`.

#### Why the current solution: Add the built-in validation to `CountryCodeTrait`

When dealing with this issue, I have thought about other possible solutions, and the reasons for choosing the current one so far are:
- The ability to use core validation that has been validated for a long time can prevent GLA from doing similar validation additionally.
- There are some GLA APIs that use validation schemas such as `minItems` and `uniqueItems` for country codes, but these have not really worked in the past. This solution will allow them to work together as well.
   - [BatchSchemaTrait](https://github.com/woocommerce/google-listings-and-ads/blob/3cc6936fc2938952f120f8c0478fccfaf2ba9eac/src/API/Site/Controllers/BatchSchemaTrait.php#L40-L45)
   - [ShippingRateSuggestionsController](https://github.com/woocommerce/google-listings-and-ads/blob/3cc6936fc2938952f120f8c0478fccfaf2ba9eac/src/API/Site/Controllers/MerchantCenter/ShippingRateSuggestionsController.php#L68-L73)
- There are a few similar implementations in the WP core.
   - [class-wp-rest-posts-controller.php](https://github.com/WordPress/wordpress-develop/blob/6.2.2/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L2977-L2980)
   - [class-wp-rest-menus-controller.php](https://github.com/WordPress/wordpress-develop/blob/6.2.2/src/wp-includes/rest-api/endpoints/class-wp-rest-menus-controller.php#L538-L542)
   - [class-wp-rest-themes-controller.php](https://github.com/WordPress/wordpress-develop/blob/6.2.2/src/wp-includes/rest-api/endpoints/class-wp-rest-themes-controller.php#L648-L652)

But I still have some doubts about whether doing so within `CountryCodeTrait` is too subtle and becomes inconsistent at a higher level concept.

#### Other possible solutions

1. Add the third parameter to [`CountryCodeTrait::validate_country_codes`](https://github.com/woocommerce/google-listings-and-ads/blob/3cc6936fc2938952f120f8c0478fccfaf2ba9eac/src/API/Site/Controllers/CountryCodeTrait.php#L47) and to all related callers.
2. Similar to the first one but don't add the parameter to callers. Instead, add a new method such as `get_country_code_and_allow_empty_validate_callback` to `CountryCodeTrait`.
3. Wrap [the `validate_callback` in `TargetAudienceController`](https://github.com/woocommerce/google-listings-and-ads/blob/3cc6936fc2938952f120f8c0478fccfaf2ba9eac/src/API/Site/Controllers/MerchantCenter/TargetAudienceController.php#L253) with an additional process: When `countries` is found to be empty, skip calling to the validation function returned by `get_country_code_validate_callback`.
4. Remove the custom `sanitize_callback` from all `CountryCodeTrait::validate_country_codes`-related callers so that the built-in `validate_callback` will be applied.
5. Add a new API dedicated to removing the audience countries and change the original POST `/mc/target_audience` to create audience countries instead of a PUT like the one defined in the RESTful API.
6. Any others?

### Changelog entry

> Fix - Avoid errors when clearing all audience countries in the onboarding flow.

### DIFF
--- a/src/API/Site/Controllers/CountryCodeTrait.php
+++ b/src/API/Site/Controllers/CountryCodeTrait.php
@@ -49,10 +49,10 @@ trait CountryCodeTrait {
 	 * @throws OutOfBoundsException When the country code cannot be found.
 	 */
 	protected function validate_country_codes( bool $check_supported_country, $countries, $request, $param ) {
-		$valid = rest_validate_request_arg( $countries, $request, $param );
+		$validation_result = rest_validate_request_arg( $countries, $request, $param );
 
-		if ( true !== $valid ) {
-			return $valid;
+		if ( true !== $validation_result ) {
+			return $validation_result;
 		}
 
 		try {
--- a/src/API/Site/Controllers/CountryCodeTrait.php
+++ b/src/API/Site/Controllers/CountryCodeTrait.php
@@ -39,6 +39,11 @@ trait CountryCodeTrait {
 	 * Validate that a country or a list of countries is valid and supported,
 	 * and also validate the data by the built-in validation of WP REST API with parameter’s schema.
 	 *
+	 * Since this extension's all API endpoints that use this validation function specify both
+	 * `validate_callback` and `sanitize_callback`, this makes the built-in schema validation
+	 * in WP REST API not to be applied. Therefore, this function calls `rest_validate_request_arg`
+	 * first, so that the API endpoints can still benefit from the built-in schema validation.
+	 *
 	 * @param bool    $check_supported_country  Whether to check the country is supported.
 	 * @param mixed   $countries                An individual string or an array of strings.
 	 * @param Request $request                  The request to validate.
--- a/src/API/Site/Controllers/CountryCodeTrait.php
+++ b/src/API/Site/Controllers/CountryCodeTrait.php
@@ -7,6 +7,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Exception\WPErrorTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Google\GoogleHelperAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\ISO3166Awareness;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\League\ISO3166\Exception\OutOfBoundsException;
+use WP_REST_Request as Request;
 use Exception;
 use Throwable;
 
@@ -35,16 +36,25 @@ trait CountryCodeTrait {
 	}
 
 	/**
-	 * Validate that a country or a list of countries is valid and supported.
+	 * Validate that a country or a list of countries is valid and supported,
+	 * and also validate the data by the built-in validation of WP REST API with parameter’s schema.
 	 *
-	 * @param mixed $countries                An individual string or an array of strings.
-	 * @param bool  $check_supported_country  Whether to check the country is supported.
+	 * @param bool    $check_supported_country  Whether to check the country is supported.
+	 * @param mixed   $countries                An individual string or an array of strings.
+	 * @param Request $request                  The request to validate.
+	 * @param string  $param                    The parameter name, used in error messages.
 	 *
 	 * @return mixed
 	 * @throws Exception            When the country is not supported.
 	 * @throws OutOfBoundsException When the country code cannot be found.
 	 */
-	protected function validate_country_codes( $countries, bool $check_supported_country ) {
+	protected function validate_country_codes( bool $check_supported_country, $countries, $request, $param ) {
+		$valid = rest_validate_request_arg( $countries, $request, $param );
+
+		if ( true !== $valid ) {
+			return $valid;
+		}
+
 		try {
 			// This is used for individual strings and an array of strings.
 			$countries = (array) $countries;
@@ -91,24 +101,26 @@ trait CountryCodeTrait {
 	}
 
 	/**
-	 * Get a callable function for validating that a provided country code is recognized.
+	 * Get a callable function for validating that a provided country code is recognized
+	 * and fulfilled the given parameter's schema.
 	 *
 	 * @return callable
 	 */
 	protected function get_country_code_validate_callback(): callable {
-		return function( $value ) {
-			return $this->validate_country_codes( $value, false );
+		return function( ...$args ) {
+			return $this->validate_country_codes( false, ...$args );
 		};
 	}
 
 	/**
-	 * Get a callable function for validating that a provided country code is recognized and supported.
+	 * Get a callable function for validating that a provided country code is recognized, supported,
+	 * and fulfilled the given parameter's schema..
 	 *
 	 * @return callable
 	 */
 	protected function get_supported_country_code_validate_callback(): callable {
-		return function( $value ) {
-			return $this->validate_country_codes( $value, true );
+		return function( ...$args ) {
+			return $this->validate_country_codes( true, ...$args );
 		};
 	}
 }
--- a/src/API/Site/Controllers/CountryCodeTrait.php
+++ b/src/API/Site/Controllers/CountryCodeTrait.php
@@ -59,10 +59,6 @@ trait CountryCodeTrait {
 			// This is used for individual strings and an array of strings.
 			$countries = (array) $countries;
 
-			if ( empty( $countries ) ) {
-				throw new Exception( __( 'No countries provided.', 'google-listings-and-ads' ) );
-			}
-
 			foreach ( $countries as $country ) {
 				$this->validate_country_code( $country );
 				if ( $check_supported_country ) {
--- a/tests/Unit/API/Site/Controllers/MerchantCenter/ShippingTimeBatchControllerTest.php
+++ b/tests/Unit/API/Site/Controllers/MerchantCenter/ShippingTimeBatchControllerTest.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers\MerchantCenter;
+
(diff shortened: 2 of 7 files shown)


src/API/Site/Controllers/CountryCodeTrait.php
```diff
@@ -35,24 +36,29 @@ protected function validate_country_code( string $country ): void {
...
 	 * @throws Exception            When the country is not supported.
 	 * @throws OutOfBoundsException When the country code cannot be found.
 	 */
-	protected function validate_country_codes( $countries, bool $check_supported_country ) {
+	protected function validate_country_codes( bool $check_supported_country, $countries, $request, $param ) {
+		$valid = rest_validate_request_arg( $countries, $request, $param );
+
+		if ( true !== $valid ) {
+			return $valid;
+		}
```
puntope: I'd suggest renaming this as `$validation_result`, `$validation` or something like that. Since it can return a WP_Error object, hence is not "valid".

src/API/Site/Controllers/CountryCodeTrait.php
```diff
@@ -35,24 +36,29 @@ protected function validate_country_code( string $country ): void {
...
-	 * Validate that a country or a list of countries is valid and supported.
+	 * Validate that a country or a list of countries is valid and supported,
+	 * and also validate the data by the built-in validation of WP REST API with parameter’s schema.
 	 *
-	 * @param mixed $countries                An individual string or an array of strings.
-	 * @param bool  $check_supported_country  Whether to check the country is supported.
+	 * @param bool    $check_supported_country  Whether to check the country is supported.
+	 * @param mixed   $countries                An individual string or an array of strings.
+	 * @param Request $request                  The request to validate.
+	 * @param string  $param                    The parameter name, used in error messages.
```
puntope, line 50: Maybe we can add somewhere a short explanation of your findings of why `rest_validate_request_arg` is not being called automatically.

tests/Unit/API/Site/Controllers/MerchantCenter/ShippingTimeBatchControllerTest.php
```diff
@@ -0,0 +1,109 @@
+<?php
```
puntope, line 1: Maybe We can add some tests for:

- Countries with wrong format return Error
- Countries not supported can be added without errors.

tests/Unit/API/Site/Controllers/MerchantCenter/TargetAudienceControllerTest.php
```diff
@@ -0,0 +1,104 @@
+<?php
```
puntope, line 1: Same as https://github.com/woocommerce/google-listings-and-ads/pull/2003/files#r1245074785

puntope:
Can we test if in `ShippingRateController` the validation for `country` also works?

Same for `ShippingRateSuggestionsController`, at least would be nice to include the `country_codes` test.

puntope:
> If the sanitize_callback is not specified but type is specified, both built-in validate_callback and sanitize_callback will be applied no matter whether the validate_callback is a custom one.

Checking here, seems like it runs `rest_parse_request_arg` as the default sanitize callback when a type is added. 
https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api/class-wp-rest-request.php#L821

That [rest_validate_request_arg](https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api.php#L1362) runs the validation.  

> When the built-in validate_callback is skipped, the required specified in args will still work. This makes the validation options of args work or not, resulting in more unexpected inconsistencies.

Seems like the [dispatcher](https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api/class-wp-rest-server.php#L1017) is calling `has_valid()` function which checks the required params.

puntope:
Hi @eason9487 , the PR looks good to me. I tested it locally and I cannot see anymore the early error when no countries are selected. 

I left some checks I did after your findings. And some suggestions in regards test and some variable naming. Since the variable naming and the tests are not blockers, I will approve this in advance.

6. Any others? 

To say something one simple solution might be to just not call the endpoint if country codes are empty in the JS side and keep all of them as required (minItems 1) in the backend.

puntope (approved):

src/API/Site/Controllers/CountryCodeTrait.php
eason9487: Renamed in https://github.com/woocommerce/google-listings-and-ads/pull/2003/commits/bdfb3651bf14b1066336b92839e2b4b10eff8498.

src/API/Site/Controllers/CountryCodeTrait.php
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/2003/commits/4eeffe896c0dc87a8ddeff94e2941e1eabad6f28.

eason9487:

> Can we test if in `ShippingRateController` the validation for `country` also works?
> Same for `ShippingRateSuggestionsController`, at least would be nice to include the `country_codes` test.

`ShippingRateController` specifies the `country` as a string type and it only indicates the `required` option, so it's not affected by this PR and doesn't have a validation schema that will work. ~In addition, the POST `mc/shipping/rates` API and all `mc/shipping/rates/{country_code}` APIs have not been in use after #411.~

About `ShippingRateSuggestionsController`, I just recalled that its API endpoint `mc/shipping/rates/suggestions` has not been in use for some time after the use was removed by #1616  Perhaps removing it would be a bit better than writing tests for it.

> Maybe We can add some tests for: [...]
> Quoted from https://github.com/woocommerce/google-listings-and-ads/pull/2003#discussion_r1245074785 and https://github.com/woocommerce/google-listings-and-ads/pull/2003#discussion_r1245077778

Considering several controllers use `CountryCodeTrait` to get the validation callback for country codes, it would be a bit repetitive to add the relevant tests for each of them. So I'm going to add tests for `CountryCodeTrait` instead, but the code changes were already more than expected and beyond this PR scope. I will open a separate PR for it.


> > 6\. Any others?
> 
> To say something one simple solution might be to just not call the endpoint if country codes are empty in the JS side and keep all of them as required (minItems 1) in the backend.

The intention of calling this API even if the countries array is empty on the UI side is to always remember the entered data during the onboarding flow, so that the user can continue with the previously entered data if they interrupt or refresh the onboarding flow. So I would think this solution would conflict with the purpose.

eason9487:
Sorry, @puntope. A thing I mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/2003#issuecomment-1612941617 is incorrect. The POST `mc/shipping/rates` API and most `mc/shipping/rates/{country_code}` APIs are still being used by `ShippingRateBatchController` internally. I have also added a strikethrough for the incorrect description.

PR #2004, opened 2023-06-30
Add PHP tests for `CountryCodeTrait` and two related API controllers

eason9487:

It's a follow-up PR of https://github.com/woocommerce/google-listings-and-ads/pull/2003#issuecomment-1612941617

- Add test cases for asserting failed requests with invalid country codes.
   - `TargetAudienceController`
   - `ShippingTimeBatchController`
- Add tests for the `CountryCodeTrait`

P.S. I didn't add test cases for "requests should be successful with unsupported country codes" as they seem to be missing data validations from the APIs rather than missing these test cases.

1. Check if the PHP unit tests can pass.
2. Check if the test cases fit with the test purpose.

### Additional details:

My original idea was to cover the test objectives more comprehensively by:

1. Test all cases for `CountryCodeTrait`
2. Assert whether the `args` of the API endpoints in the related controllers like `TargetAudienceController` has specified `minItems` and use `CountryCodeTrait::get_country_code_validate_callback` as its callable `validate_callback`.

Because the invalid/unsupported country codes will already be covered by `CountryCodeTraitTest`, and the schema validations are already tested by WP core. Therefore, this way should be able to ensure that the validations in these API controllers will work well.

Unfortunately, after some time I still couldn't figure out how to verify the second point robustly. Several of my attempts have been relatively indirect and have too much relied on internal implementation. Perhaps the concept of PHPUnit doesn't advocate this but should focus on testing the I/O of public methods.

In view of the above and I have already spent too much time on this follow-up work, I have instead added similar test cases for these API controllers and also kept the newly added `CountryCodeTraitTest`.

### DIFF
--- a/tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
+++ b/tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
@@ -1,0 +1,248 @@
+<?php
+// This namespace must be identical to CountryCodeTrait for replacing and
+// mocking the global function rest_validate_request_arg.
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers {
+	use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers\CountryCodeTraitTest;
+
+	/**
+	 * Replace and mock the global function rest_validate_request_arg.
+	 */
+	function rest_validate_request_arg() {
+		$mocked_callback = CountryCodeTraitTest::$rest_validate_request_arg_callback;
+
+		if ( is_null( $mocked_callback ) ) {
+			return \rest_validate_request_arg( ...func_get_args() );
+		}
+
+		return $mocked_callback( ...func_get_args() );
+	}
+};
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers {
+	use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\CountryCodeTrait;
+	use Automattic\WooCommerce\GoogleListingsAndAds\Google\GoogleHelper;
+	use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\League\ISO3166\ISO3166DataProvider;
+	use PHPUnit\Framework\MockObject\MockObject;
+	use PHPUnit\Framework\TestCase;
+	use Exception;
+	use WP_Error;
+	use WP_REST_Request as Request;
+
+	/**
+	 * Class CountryCodeTraitTest
+	 *
+	 * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers
+	 * @phpcs:disable Squiz.Classes.ClassDeclaration.SpaceBeforeKeyword
+	 * @phpcs:disable Squiz.Classes.ClassDeclaration.SpaceBeforeCloseBrace
+	 */
+	class CountryCodeTraitTest extends TestCase {
+
+		/** @var callable $rest_validate_request_arg_callback */
+		public static $rest_validate_request_arg_callback = null;
+
+		/** @var MockObject|ISO3166DataProvider $iso_provider */
+		protected $iso_provider;
+
+		/** @var MockObject|GoogleHelper $google_helper */
+		protected $google_helper;
+
+		/** @var bool $country_supported */
+		protected $country_supported;
+
+		/**
+		 * Runs before each test is executed.
+		 */
+		public function setUp(): void {
+			parent::setUp();
+
+			$this->iso_provider  = $this->createMock( ISO3166DataProvider::class );
+			$this->google_helper = $this->createMock( GoogleHelper::class );
+
+			// phpcs:ignore WordPress.Classes.ClassInstantiation.MissingParenthesis
+			$this->trait = new class {
+				use CountryCodeTrait {
+					get_country_code_sanitize_callback as public;
+					get_country_code_validate_callback as public;
+					get_supported_country_code_validate_callback as public;
+				}
+			};
+
+			$this->trait->set_iso3166_provider( $this->iso_provider );
+			$this->trait->set_google_helper_object( $this->google_helper );
+
+			$this->country_supported = true;
+
+			$this->google_helper->method( 'is_country_supported' )
+				->willReturnCallback(
+					function () {
+						return $this->country_supported;
+					}
+				);
+		}
+
+		/**
+		 * Runs after each test is executed.
+		 */
+		public function tearDown(): void {
+			self::$rest_validate_request_arg_callback = null;
+		}
+
+		/**
+		 * Test all get_*_callback methods return a callable.
+		 */
+		public function test_all_get_callbacks_return_callable() {
+			$this->assertIsCallable( $this->trait->get_country_code_sanitize_callback() );
+			$this->assertIsCallable( $this->trait->get_country_code_validate_callback() );
+			$this->assertIsCallable( $this->trait->get_supported_country_code_validate_callback() );
+		}
+
+		/**
+		 * Test the get_country_code_validate_callback and get_supported_country_code_validate_callback
+		 * methods will apply arguments to rest_validate_request_arg_callback and return early if invalid.
+		 */
+		public function test_all_get_validate_callbacks_with_rest_validate_request_arg() {
+			$called_count = 0;
+
+			self::$rest_validate_request_arg_callback = function () use ( &$called_count ) {
+				// When it returns a value other than `true`, it means the validation did not pass,
+				// and here we also make it return a called count.
+				$called_count++;
+				return [ $called_count, func_get_args() ];
+			};
+
+			$args = [ 'US', new Request(), 'country_code' ];
+
+			$callback = $this->trait->get_country_code_validate_callback();
+			$results  = $callback( ...$args );
+
+			$this->assertEquals( [ 1, $args ], $results );
+
+			$callback = $this->trait->get_supported_country_code_validate_callback();
+			$results  = $callback( ...$args );
+
+			$this->assertEquals( [ 2, $args ], $results );
+		}
+
+		/**
+		 * Test a sanitized country code with a string.
+		 */
+		public function test_get_country_code_sanitize_callback_with_string() {
+			$callback = $this->trait->get_country_code_sanitize_callback();
+			$result   = $callback( 'us' );
+
+			$this->assertEquals( 'US', $result );
+		}
+
+		/**
+		 * Test sanitized country codes with an array.
+		 */
+		public function test_get_country_code_sanitize_callback_with_array() {
+			$callback = $this->trait->get_country_code_sanitize_callback();
+			$results  = $callback( [ 'us', 'gb', 'jp' ] );
+
+			$this->assertEquals( [ 'US', 'GB', 'JP' ], $results );
+		}
+
+		/**
+		 * Test a valid country code with a string.
+		 */
+		public function test_get_country_code_validate_callback_with_string() {
+			$callback = $this->trait->get_country_code_validate_callback();
+			$result   = $callback( 'US', new Request(), 'country_code' );
+
+			$this->assertTrue( $result );
+		}
+
+		/**
+		 * Test valid country codes with an array.
+		 */
+		public function test_get_country_code_validate_callback_with_array() {
+			$callback = $this->trait->get_country_code_validate_callback();
+			$result   = $callback( [ 'US', 'GB' ], new Request(), 'country_codes' );
+
+			$this->assertTrue( $result );
+		}
+
+		/**
+		 * Test valid country codes with unsupported countries.
+		 *
+		 * It determines as valid since this method only checks if the countries match ISO 3166
+		 * but won't check if Google Merchant Center supports the countries.
+		 */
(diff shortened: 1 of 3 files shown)


tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		/**
+		 * Test the get_country_code_validate_callback and get_supported_country_code_validate_callback
+		 * methods will apply arguments to rest_validate_request_arg_callback and return early if invalid.
+		 */
+		public function test_all_get_validate_callbacks_with_rest_validate_request_arg() {
+			$called_count = 0;
+
+			self::$rest_validate_request_arg_callback = function () use ( &$called_count ) {
```
martynmjones: I'm unsure how beneficial it is to replace the global `rest_validate_request_arg` so that we can monitor how many times it's called here. It makes it harder to understand and feels quite heavy-handed for what we're trying to test. Had you considered adding a wrapper for `rest_validate_request_arg` to `CountryCodeTrait` instead?

If we were using a wrapper then when setting up `$this->trait` in the test we could modify how that functions so that we can set the desired return values by adding something like:

```php
$this->trait = new class {
  use CountryCodeTrait {
	  get_country_code_sanitize_callback as public;
	  get_country_code_validate_callback as public;
	  get_supported_country_code_validate_callback as public;
  }
  
  private $rest_validate_request_arg_response;
  
  public function rest_validate_request_arg() {
	  return $this->rest_validate_request_arg_response;
  }
  
  public function set_rest_validate_request_arg_response( $value ) {
	  $this->rest_validate_request_arg_response = $value;
  }
};
```

The test could then be simplified a lot to something along the lines of:

```php
public function test_all_get_validate_callbacks_with_rest_validate_request_arg() {
	$this->trait->set_rest_validate_request_arg_response( new WP_Error( 'invalid', 'Invalid' ) );
	$args = [ 123, new Request(), 'country_codes' ];

	$this->assertInstanceOf( WP_Error::class, $this->trait->get_country_code_validate_callback()( ...$args ) );
	$this->assertInstanceOf( WP_Error::class, $this->trait->get_supported_country_code_validate_callback()( ...$args ) );
}
```
⬆️ _Examples haven't been tested._

This would avoid the need for two namespace declarations in the file and we wouldn't need to replace the global function.

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		/**
+		 * Test a sanitized country code with a string.
+		 */
+		public function test_get_country_code_sanitize_callback_with_string() {
+			$callback = $this->trait->get_country_code_sanitize_callback();
+			$result   = $callback( 'us' );
+
+			$this->assertEquals( 'US', $result );
```
martynmjones: Instead of setting `$callback` and `$result` for each test we could reduce that to a single line (_where possible..._). It looks like this would be applicable to quite a few of the tests and make the file a bit shorter and tidier.

```suggestion
			$this->assertEquals( 'US', $this->trait->get_country_code_sanitize_callback()( 'us' ) );
```

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
```diff
@@ -0,0 +1,248 @@
...
+
+		if ( is_null( $mocked_callback ) ) {
+			return \rest_validate_request_arg( ...func_get_args() );
+		}
+
+		return $mocked_callback( ...func_get_args() );
+	}
+};
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\Site\Controllers {
```
martynmjones: If we adjust `test_all_get_validate_callbacks_with_rest_validate_request_arg` as [suggested](https://github.com/woocommerce/google-listings-and-ads/pull/2004#discussion_r1250897333) then we could drop the curly braces for the namespace which is specified as [incorrect in the WordPress coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#namespace-declarations) (_Although I appreciate we're in tests at the moment so its not as big of an issue._)

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		/**
+		 * Test valid country codes with unsupported countries.
+		 *
+		 * It determines as valid since this method only checks if the countries match ISO 3166
+		 * but won't check if Google Merchant Center supports the countries.
+		 */
+		public function test_get_country_code_validate_callback_with_unsupported_country() {
+			$this->country_supported = false;
```
martynmjones: The value of `$this->country_supported` has no bearing on this test.

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
```diff
@@ -0,0 +1,248 @@
...
+
+			$this->assertEquals( [ 'US', 'GB', 'JP' ], $results );
+		}
+
+		/**
+		 * Test a valid country code with a string.
+		 */
+		public function test_get_country_code_validate_callback_with_string() {
+			$callback = $this->trait->get_country_code_validate_callback();
+			$result   = $callback( 'US', new Request(), 'country_code' );
```
martynmjones: Should this be `country_codes`? I know this doesn't have an impact on the test but for clarity it would be good to ensure the naming is correct throughout.

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
```diff
@@ -0,0 +1,248 @@
...
+			$this->assertEquals( 'gla_invalid_country', $result->get_error_code() );
+			$this->assertEquals( 'Country is invalid', $result->get_error_message() );
+		}
+
+		/**
+		 * Test a valid country code with a string.
+		 */
+		public function test_get_supported_country_code_validate_callback_with_string() {
+			$callback = $this->trait->get_supported_country_code_validate_callback();
+			$result   = $callback( 'US', new Request(), 'country_code' );
```
martynmjones: [As above](https://github.com/woocommerce/google-listings-and-ads/pull/2004#discussion_r1251137816)

martynmjones (commented):
Hey @eason9487, Thanks for your work expanding the GLA tests!

I've gone through the code and added a few comments which mainly surround how we're working with `rest_validate_request_arg`. Let me know if anything I've mentioned isn't clear 👍

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
eason9487: I only apply the suggestion to this test case. If the rest test cases were applied, a single line would be too complicated and even might have to scroll the editor to see the whole line of code, which reduces the readability too much.

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
eason9487: Indeed, this line doesn't affect the test results if removed it, but it's there to meet the objective and semantics of this test case.

Certainly, it's possible to show that this line is required to further check if the mocked `google_helper` does return `false` in this test by the following code. But I would consider this to be a bit too involved in the actual implementation detail and a bit redundant.

```php
$reflector     = new \ReflectionClass( get_class( $this->trait ) );
$google_helper = $reflector->getProperty( 'google_helper' );
$google_helper->setAccessible( true );

$this->assertFalse( $google_helper->getValue( $this->trait )->is_country_supported( 'CN' ) );
$this->assertFalse( $google_helper->getValue( $this->trait )->is_country_supported( 'KP' ) );
```

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
eason9487: Probably not. `validate_country_codes` also accepts a `$countries` as a string and this test case is testing it with a string code.

https://github.com/woocommerce/google-listings-and-ads/blob/575d90ee207c4e08400fae8a598355ec69198122/src/API/Site/Controllers/CountryCodeTrait.php#L48

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
eason9487: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2004#discussion_r1251453403. This test case is also testing it with a string.

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
eason9487: > I'm unsure how beneficial it is to replace the global `rest_validate_request_arg` so that we can monitor how many times it's called here.

This testing goal is to verify that the two callbacks do call the global `rest_validate_request_arg` to further ensure that the built-in schema validation of WP REST API is applied.

It's indeed very complicated and I got stuck with this part the longest. But PHPUnit seems not to provide an explicit way to mock a global function, and I couldn't find a better way to achieve this goal, so I ended up with the current solution.

> It makes it harder to understand and feels quite heavy-handed for what we're trying to test. Had you considered adding a wrapper for `rest_validate_request_arg` to `CountryCodeTrait` instead?
> 
> If we were using a wrapper then when setting up `$this->trait` in the test we could modify how that functions so that we can set the desired return values by adding something like: [...]
> This would avoid the need for two namespace declarations in the file and we wouldn't need to replace the global function.

Sorry, I can't understand from the provided code example how this approach is going to achieve the goals mentioned above. I tried to change some tests with it, but it didn't even be able to mock the result of calling `rest_validate_request_arg`.

Could you explain the solution a bit more?

eason9487:
Hi @martynmjones, thanks for the review.

I replied to all the review comments and it's ready for a new round of code reviews.

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
martynmjones: > Sorry, I can't understand from the provided code example how this approach is going to achieve the goals mentioned above. I tried to change some tests with it, but it didn't even be able to mock the result of calling rest_validate_request_arg.

Sorry if what I was saying wasn't clear! The point I was trying to make is if we were to add a method to `CountryCodeTrait`:

```php
public function rest_validate_request_arg() {
	return rest_validate_request_arg( ...func_get_args() );
}
```

Then we could easily test that other methods are invoking that wrapper rather than calling the global function directly which we can't easily mock.

Granted all that would confirm in the tests is that the wrapper is being called rather than the global function but we could then add a test for the `rest_validate_request_arg` response along the lines of:

```php
public function test_invalid_rest_validate_request_arg_response() {
	$request = new WP_REST_Request( 'GET', '/' );
	$request->set_attributes(
		[
			'args' => [
				'country_codes' => [
					'type' => 'string',
				],
			],
		],
	);

	$args = [ 123, $request, 'country_codes' ];
	$this->assertEquals( 'rest_invalid_type', $this->trait->rest_validate_request_arg( $args ) );
}
```

That somewhat blurs the lines between unit test and integration test but given some of our tests already rely on libraries and it's not uncommon I think that would still be preferable to the separate namespace declarations.

Alternatively, `test_all_get_validate_callbacks_with_rest_validate_request_arg` could rely on the response of `rest_validate_request_arg` for the test like in the above example rather than adding a wrapper. What do you think?

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
martynmjones: Thanks for the clarification and apologies, I thought we were using `country_codes` as the parameter for all endpoints but I can see that's not the case 👍

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
martynmjones: Right! I understand, thanks for the explanation.

martynmjones:
Hey @eason9487, thanks for getting back to the points I raised and for the additional context.

I left another [comment](https://github.com/woocommerce/google-listings-and-ads/pull/2004#discussion_r1251734197) about the use of `rest_validate_request_arg` but aside from that everything looks good.

tests/Unit/API/Site/Controllers/CountryCodeTraitTest.php
eason9487: Thank you for the idea.

On the balance of robustness, complexity and WP coding standards, I agree it's feasible to indirectly verify "Whether the global `rest_validate_request_arg` is called."

Since the main purpose is not to mock the call result of `rest_validate_request_arg`, I omitted the part about creating the wrapper method in `CountryCodeTrait`.

I adjusted it by two commits:
1. https://github.com/woocommerce/google-listings-and-ads/pull/2004/commits/c6f320fc05e3b75707f012dece2c73499aa7b306 for the main change
2. https://github.com/woocommerce/google-listings-and-ads/pull/2004/commits/18c7d9599fe5ba7510cb7f35a03634e9358b8a2b for indentation adjustment

I believe it would be easier to review the changes.

eason9487:
Hi @martynmjones, thanks for the detailed explanation!

The multiple namespaces in the same file have been addressed. Could you take another look at this PR?

martynmjones (approved):
Hey @eason9487, many thanks for all your work on this and for the changes 🙌 

The tests are a lot cleaner and easier to read now and all are passing so LGTM ✅

Issue #1725, opened 2022-10-12
Detailed error when unable to complete site verification


mikkamp:
### Describe the bug:
When we perform the site verification, we do so by inserting a meta tag on the website. The site must be public at this point otherwise Google won't be able to see the meta tag and the verification fails.

When we perform this verification we do so through an API insert call, but when this fails the error message is vague. The response from the API is:

> The necessary verification token could not be found on your site.

However this error needs to be retrieved from the log file. Since we do an insert call in the API the UI just returns a generic error:

> Unable to insert site verification.

![image](https://user-images.githubusercontent.com/11388669/195287571-5cc1ccf9-7018-4619-aab9-b3b088e46fc3.png)

We should be giving the user some actionable steps to try and resolve it, or at least clarify what's going on. This can be done by changing the error message to something like:

> The verification token can not be found on your site, please ensure your site is public and not being cached.

### Steps to reproduce:
1. Use a coming soon plugin, or use a hosting option to prevent the site from being public. 
2. Go through the onboarding
3. Confirm the verification fails when connecting the Merchant Account

PR #1258, opened 2022-02-18
Improve Ads error messages returned by the API

mikkamp:
### Changes proposed in this Pull Request:

Resolves the backend part of #1246 

This PR handles the backend error messages when we catch an ApiException from the Ads API.

The following changes have been added to this PR:

- Add a helper function for extracting the errors from an ApiException object
- Return detailed error messages containing a list of errors as well as relevant data such as the campaign ID
- Pass on the detailed messages in the CampaignController
- Add a detailed error for report requests
- Add a detailed error when requesting a list of Ad accounts

Before this PR if the request for retrieving account ID's failed we wouldn't map the error code (causing a 200 response to be returned). This scenario seems to have gone untested as it causes a failure when trying to add an Ad account

image

This PR fixes the behaviour so it correctly maps the error code, which results in the error being handled correctly. However the user is unable to continue since it stays in the loading state.

image

### Unit tests
Explicit testing for detailed error messages was added in the AdsCampaignTest. However there are no unit tests for Proxy or AdsReport, so I didn't add anything there. That's something which can be done in a followup PR although it would cover a lot more code.

#### Campaign errors
1. Go to WooCommerce > Settings > General and set the number of decimals to 3
2. Confirm we have an Ads account connected with some existing campaigns
3. Edit a campaign and try to set the budget amount to 0.001
4. Confirm that the UI responds the same as previously with a default error message
image
5. Open the network tab in the browser and confirm the request status remains the same and the response has additional details about the error:

```json
{
	"message": "Error editing campaign: A money amount was less than the minimum CPC for currency.",
	"errors": {
		"MONEY_AMOUNT_LESS_THAN_CURRENCY_MINIMUM_CPC": "A money amount was less than the minimum CPC for currency.",
		"INVALID_ARGUMENT": "Request contains an invalid argument."
	},
	"id": 1234567890
}
```

The error used to contain just the message:
```json
{
	"message": "Error editing campaign: Request contains an invalid argument."
}
```

#### Get accounts error
1. Disconnect any connected Ads account
2. Simulate an error by adding the following line in the file `src/API/Google/Proxy.php` within the `try` block of the function `get_ads_account_ids`
```php
throw new ApiException( 'invalid', 3, 'INVALID_ARGUMENT' );
```
3. Go Marketing > Google Listings & Ads > Dashboard and add a campaign to connect our Ad account
4. Confirm the error appears for the existing ad accounts
image

#### Report errors 
[...]

PR #2007, opened 2023-07-04
Adjust the infrastructure to uniform the error messages returned from Google Shopping Content service

eason9487:

This PR is the infrastructure part of #1725 and relates to #1993.

Before making the error message clearer to users, a lot of fundamental changes might be required. Prior to making more changes, I would like to start solution reviews with this PR to make sure this direction is feasible.

Previously, #1258 has pointed out the way to unify GLA error messages for `ApiException` errors thrown from Google Ads APIs. On the Google Shopping Content side, an exception instance may be `Google\Exception` or `Google\Service\Exception`.

The idea is to add the support of these exceptions to the `ExceptionTrait`, renamed from `ApiExceptionTrait`, and then transform the uniformed error messages to `ExceptionWithResponseData` on the error catcher sides.

To do so, this PR:

- Add `get_exception_errors` to `ApiExceptionTrait` as a more general method for getting errors from an exception.
- Rename `ApiExceptionTrait` to `ExceptionTrait` and change all callers to use the general method for getting errors.
- Add the support of `GoogleServiceException` type to `ExceptionTrait::get_exception_errors`.
- Adjust the methods of `MerchantTrait` to mock the possible exceptions.
- Change the relevant getter and updater of Merchant Center account to throw an ExceptionWithResponseData when failed.
- Remove an unused file *src/Exception/MerchantApiException.php*

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/c5c03aff-b272-4f6a-bc1d-5f2bf9a422ff

Some other error messages I tested:

1. Disconnect a connected Google Merchant Center account (via the Connection Test page).
2. Go to step 1 of the onboarding flow and connect to the same Google Merchant Center account with an invalid site URL. This can be achieved by changing this line of code locally
   https://github.com/woocommerce/google-listings-and-ads/blob/d0209ea221ba6b8de741e2273ada9c708f9fad76/src/MerchantCenter/AccountService.php#L462
3. It should show a clearer error message in the notification UI than before. Previously, the message was "Unable to update Merchant Center account."
4. Inspect the response body of the `gla/mc/accounts/switch-url` API via Console tab of DevTool. It should be in a format like:
   ```json
   {
       "message": "Unable to update Merchant Center account: [websiteUrl] URL starts with an invalid scheme",
       "errors": {
           "invalid": "[websiteUrl] URL starts with an invalid scheme"
       }
   }
   ```

### Additional details:

About the data structure and format of the `errors` field in the `ExceptionWithResponseData`, it's obviously not very consistent across the different data sources. Since the frontend side is not yet using it but only the `message` field, I think it would be appropriate to continue its optimization when the failed response body returned by GLA APIs is more consistent. The current tuning won't focus much on this part.

### Changelog entry

> Tweak - Make the error message clearer for errors that occur in getting or updating a Google Merchant Center account.


src/API/Google/ExceptionTrait.php
```diff
@@ -44,6 +46,36 @@ protected function has_api_exception_error( ApiException $exception, string $err
...
+	 *
+	 * @link https://github.com/googleads/google-ads-php/blob/v19.1.0/src/Google/Ads/GoogleAds/V13/Services/Gapic/CustomerServiceGapicClient.php#L359-L361
+	 * @link https://github.com/googleapis/google-api-php-client/blob/v2.15.0/src/Http/REST.php#L119-L135
+	 * @link https://github.com/googleapis/google-api-php-client/blob/v2.15.0/src/Service/Resource.php#L86-L175
+	 *
+	 * @param ApiException|GoogleServiceException|Exception $exception Exception to check.
+	 *
+	 * @return array
+	 */
+	protected function get_exception_errors( Exception $exception ): array {
```
puntope, line 65: In the docs it says --> `ApiException|GoogleServiceException|Exception `

However, in the type it says only Exception.

puntope:
> Add get_exception_errors to ApiExceptionTrait as a more general method for getting errors from an exception.

One thing I noticed is that we changed `get_api_exception_errors` to `get_exception_errors` However, we are still doing  `catch ( ApiException $e )`.  So it won't be much difference from using `get_api_exception_errors`

src/API/Google/ExceptionTrait.php
```diff
@@ -82,6 +114,23 @@ protected function get_api_exception_errors( ApiException $exception ): array {
...
+	 *
+	 * @param GoogleServiceException $exception Exception to check.
+	 *
+	 * @return array
+	 */
+	private function get_google_service_exception_errors( GoogleServiceException $exception ): array {
+		$errors = [];
+
+		foreach ( $exception->getErrors() as $error ) {
+			$errors[ $error['reason'] ] = $error['message'];
```
puntope: How we can be sure `$error['reason'` and `$error['message']` are defined?  

I would add some kind of [isset](https://www.php.net/manual/en/function.isset.php) to be sure is defined, and return a generic otherwise.

puntope (commented):
Hi @eason9487 

I reviewed the code. My local tests were working as described in the PR-

I left some comments regarding the code tho.

src/API/Google/ExceptionTrait.php
eason9487: Support for union type declarations was added in PHP 8.0 but GLA needs to be compatible with PHP >= 7.4. Considering both `ApiException` and `GoogleServiceException` extend the `Exception` class, so I thought it should be okay to declare `Exception` here and be a bit better than no type declaration.

However, I'm not that familiar with PHP, does this sound reasonable?

src/API/Google/ExceptionTrait.php
eason9487: From Google's source code and doc, I think `$error['reason']` and `$error['message']` should always exist:

- Only [`Google\Http\REST::decodeHttpResponse`](https://github.com/googleapis/google-api-php-client/blob/v2.15.0/src/Http/REST.php#L121-L135) may create `GoogleServiceException` instances along with calling [`getResponseErrors`](https://github.com/googleapis/google-api-php-client/blob/v2.15.0/src/Http/REST.php#L176-L185)
- [`GoogleServiceException::getErrors`](https://github.com/googleapis/google-api-php-client/blob/v2.15.0/src/Service/Exception.php#L54-L72) provides an example of the returned errors 
- [Google doc](https://developers.google.com/shopping-content/reference/rest/v2.1/Errors) shows the errors returned by Google Shopping Content service

Do you think it's still better to add `isset` checks?

BTW, while tracing the above things, I noticed that `$exception->getErrors()` might return `null`, so I added the null check in https://github.com/woocommerce/google-listings-and-ads/pull/2007/commits/8e46c2ae48e2e9d383fa86c5a6073ce7b637ef98.

eason9487:
Hi @puntope, thanks for the review.

> > Add `get_exception_errors` to `ApiExceptionTrait` as a more general method for getting errors from an exception.
> 
> One thing I noticed is that we changed `get_api_exception_errors` to `get_exception_errors` However, we are still doing `catch ( ApiException $e )`. So it won't be much difference from using `get_api_exception_errors`

The purpose is to make the handler for exceptions thrown from Google Ads and Google Shopping Content service unified. I believe it's relatively appropriate for caller catchers to still use `ApiException`. Because it explicitly indicates that the catcher only wants to handle `ApiException` rather than other exceptions, and the catching target of a `catch {}` block is usually very dependent on all the exceptions that can occur within its `try {}` block.

Could you help with a new round of code reviews?

src/API/Google/ExceptionTrait.php
puntope: Yes, I didn't get into that compatibility problem. 

What do you think about removing the type in the param, tho? 

`protected function get_exception_errors( $exception ): array `

src/API/Google/ExceptionTrait.php
puntope: I'm always into checking 3rd party inputs... even if it's from Google and it's documented.  Even if actually is always defined, what will happen if we update the GLA library or the API and they change this? Seems to me that It's going to be missed since we are not going to test the error messages as frequently. 

Since it's not a big change and it's definitely not harmful to the code I would suggest adding it. What do you think?

src/API/Google/ExceptionTrait.php
eason9487: > Yes, I didn't get into that compatibility problem.
> 
> What do you think about removing the type in the param, tho?
> 
> `protected function get_exception_errors( $exception ): array `

If I understood it right, `mixed` is assumed when no type is declared, which means it would be broader than expected types. Moreover, most PHP files have `declare( strict_types=1 );` in this codebase, I assumed the point of enabling strict types is also to catch type errors earlier. So I considered it better to declare type when possible.

Do you think it's better not to declare the type in this use case?

puntope:
> The purpose is to make the handler for exceptions thrown from Google Ads and Google Shopping Content service unified. I believe it's relatively appropriate for caller catchers to still use ApiException. Because it explicitly indicates that the catcher only wants to handle ApiException rather than other exceptions, and the catching target of a catch {} block is usually very dependent on all the exceptions that can occur within its try {} block.

From my understanding the big deal in this PR was that added `get_google_service_exception_errors` and the changes in the usage of `ExceptionWithResponseData` instead of the `MerchantApiException` generics.  Yes, I believe they are making the experience with the error awareness and also the code better. 

What I meant with `get_api_exception_errors` change... is that knowing is an ApiException we know also for sure it will call `get_api_exception_errors`.  So it doesn't make any difference to call `get_exception_errors` on those cases. Nothing against that change. It was just an observation and also to be sure I didn't miss any reason for it.  I wonder if it would be more convenient to switch to Exception instead of ApiException since we are now handling them with `get_exception_errors` that are able to manage more ranges of Exceptions. Otherwise, why do we need to call `get_exception_errors` instead of `get_api_exception_errors`?

puntope (commented):
Hi @eason9487 , thanks for your time answering and addressing my comments. Especially your catch in the compatibility for the types in PHP. :til: 

I left some more discussions tho to be resolved before approval.

src/API/Google/ExceptionTrait.php
eason9487: > I'm always into checking 3rd party inputs... even if it's from Google and it's documented. Even if actually is always defined, what will happen if we update the GLA library or the API and they change this? Seems to me that It's going to be missed since we are not going to test the error messages as frequently.

I'm hard to agree that it's reasonable to "**always** do data structure checking to 3rd-party inputs", rather it's a bit overkill.

Fault tolerance is usually used to allow the application to recover from errors such as retry when the network is unstable, or not to result in a critical problem like system crash.

However, when the data structure provided by a 3rd-party is suddenly not presented in its promised format, the application usually doesn't work, and even fault tolerance for missing or breaking changed data would hardly change the result of the unavailable application.

> Since it's not a big change and it's definitely not harmful to the code I would suggest adding it.

Although I don't think it's necessary, I also don't strongly disagree it. Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2007/commits/8fd35cc10b76ce1a8b7561cf5ffb3239a4a85644.

eason9487:
Hi @puntope, thanks for the second round review. Could you take another look at this PR?

> I wonder if it would be more convenient to switch to Exception instead of ApiException since we are now handling them with `get_exception_errors` that are able to manage more ranges of Exceptions.

Switching to `Exception` instead of `ApiException` involves a wider re-examination range of error handling. Because the capability of `get_exception_errors` to handle `Exception` doesn't mean that all callers can immediately call the method with `Exception` without revisiting its error handling one by one.

This could indeed be an enhancement project, but I think it would be more appropriate to make up for it when there is extra time during the Cooldown weeks.

> Otherwise, why do we need to call `get_exception_errors` instead of `get_api_exception_errors`?

I thought the reason was already mentioned in the previous reply - The purpose is to make the handler for exceptions thrown from Google Ads and Google Shopping Content service unified.

Or, do you mean using the same method as the only accessible method to handle different exception types and getting the same data structure returned is not a good idea or an anti-pattern in PHP?

In addition, there are two classes of exceptions that may be thrown from the Google Shipping Content library: `Google\Exception` and `Google\Service\Exception`. The former one simply extends `Exception` without adding anything. If an exception caught by the `catch {}` block should be always passed to a method that receives the **exact** same class type of parameter, then `ExceptionTrait::get_exception_errors` would have to be split into `get_google_exception_errors` and `get_google_service _exception_errors`, and then all the caller sides would also be split into `catch ( GoogleException $e )` and `catch ( GoogleServiceException $e )`, which in order to make them 1-1 mapping.

Is that a relatively good practice?

src/API/Google/ExceptionTrait.php
puntope: > it better to declare type when possible.

Yes, I agree with this, the issue for me was that I thought Exception was another Google-based Exception alias excluding ApiException and GoogleServiceException but I checked again and seems to be a generic PHP Exception. So you're right and the Exception type is the most correct. I wonder if we should add just Exception as a type in the docs since we are already handling the interfaces ( via $exception instanceof  ) and ApiException and GoogleServiceException inherit from Exception. BUt this is not anymore a blocker for me.

BTW, Strict typing is only defined for scalar type declarations. So it doesn't apply here. ([More info](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict)) 

In regards to `mixed`, you're right. It's equivalent to "object|resource|array|string|float|int|bool|null"

eason9487: > BTW, Strict typing is only defined for scalar type declarations. So it doesn't apply here. ([More info](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict))

Thanks for the correction!

puntope:
> Or, do you mean using the same method as the only accessible method to handle different exception types and getting the same data structure returned is not a good idea or an anti-pattern in PHP?

Yes, I think that was the baseline of my comment. Definitely not an anti-pattern. As mentioned was just an observation since for me is actually the same. I was wondering if it was just a preference to use the generic name or we were missing to change ApiException to Exception. 

```php
function do_things( $exception ) { 
    if ($exception == 'exceptionA') {
        doFoo( $exception );
    }

    if ($exception == 'exceptionB') {
        doBar( $exception );
    }
}
```

```php
$exception = 'exceptionA';

doFoo( $exception ) is the same as do_things( $exception )
```

However,

```php
$exception =  'exceptionA' | 'exceptionB'; // we don't know 

doFoo( $exception ) is not necessarily the same as do_things( $exception )

```

puntope (approved):
✅  This is approved

My main concern was the `isset` the rest are just discussions but not blockers. 

The Exception case was also clarified

PR #2011, opened 2023-07-11
Add Tip component when assets are imported

puntope:

This PR adds a component when assets are imported from the URL, based on the Figma project

Figma - fqR0EHi63lWahRcVTKCcba-fi-5568%3A228053
Project - P2 - pcTzPl-1ue-p2

### Screenshots:

Implementation:

Screenshot 2023-07-11 at 13 13 07

Figma:

Screenshot 2023-07-11 at 14 09 14

1. Create a new Paid Campaign
2. In step 2 select a URL
3. When the assets are imported, see the Tip appearing 
4. See the Tip matching Figma content and design
5. Select another URL
6. See the TIp disappears
7. Choose an URL
8. When the assets are imported, see the Tip appearing 
9. Co to a previously created Campaign with Assets
10. See the TIp not appearing

### Additional details:

 - I added a new `hasImportedAssets`  prop in the adapter. This prop is updated every time the Final URL changes. It iterates over the Assets and if any of them is filled, then it sets the prop as TRUE.  
 - I added some tests for `<AssetsGroupSection>`. I wanted to add also test covering the resetForm functionality. However, I didn't find any test created yet for that functionality, so I suggest approaching it in future PRs.    

### Changelog entry

> Tweak - Add Tip with information with Campaign assets are imported

### DIFF
--- a/js/src/components/paid-ads/asset-group/asset-group-section.scss
+++ b/js/src/components/paid-ads/asset-group/asset-group-section.scss
@@ -11,11 +11,8 @@
 		background: #f0f6fc;
 		border: $border-width solid #c5d9ed;
 		line-height: $gla-line-height-medium;
+		color: $gray-900;
 
-		> p {
-			color: $gray-900;
-			font-size: $default-font-size;
-		}
 		> svg {
 			fill: #007cba;
 			align-self: initial;
--- a/js/src/components/paid-ads/asset-group/asset-group-section.scss
+++ b/js/src/components/paid-ads/asset-group/asset-group-section.scss
@@ -5,4 +5,20 @@
 		font-size: $default-font-size;
 		color: $gray-700;
 	}
+
+	.components-tip {
+		padding: $grid-unit-15 $grid-unit-20;
+		background: #F0F6FC;
+		border: $border-width solid #C5D9ED;
+		line-height: $gla-line-height-medium;
+
+		> p {
+			color: $gray-900;
+			font-size: $default-font-size;
+		}
+		> svg {
+			fill: #007CBA;
+			align-self: initial;
+		}
+	}
 }
--- a/js/src/components/paid-ads/asset-group/asset-group-section.scss
+++ b/js/src/components/paid-ads/asset-group/asset-group-section.scss
@@ -8,8 +8,8 @@
 
 	.components-tip {
 		padding: $grid-unit-15 $grid-unit-20;
-		background: #F0F6FC;
-		border: $border-width solid #C5D9ED;
+		background: #f0f6fc;
+		border: $border-width solid #c5d9ed;
 		line-height: $gla-line-height-medium;
 
 		> p {
@@ -17,7 +17,7 @@
 			font-size: $default-font-size;
 		}
 		> svg {
-			fill: #007CBA;
+			fill: #007cba;
 			align-self: initial;
 		}
 	}
--- a/js/src/components/paid-ads/asset-group/asset-group-section.test.js
+++ b/js/src/components/paid-ads/asset-group/asset-group-section.test.js
@@ -18,7 +18,7 @@ jest.mock( '.~/components/adaptive-form', () => ( {
  * External dependencies
  */
 import '@testing-library/jest-dom';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 
 /**
  * Internal dependencies
@@ -32,14 +32,16 @@ jest.mock( '.~/components/paid-ads/asset-group/asset-group-card', () =>
 
 describe( 'AssetGroupSection', () => {
 	test( 'Component renders', () => {
-		const { queryByText } = render( <AssetGroupSection /> );
-		expect( queryByText( /Add additional assets/i ) ).toBeInTheDocument();
+		render( <AssetGroupSection /> );
+		expect(
+			screen.queryByText( /Add additional assets/i )
+		).toBeInTheDocument();
 	} );
 
 	test( 'Component not showing Tip if there are no imported assets', () => {
-		const { queryByText } = render( <AssetGroupSection /> );
+		render( <AssetGroupSection /> );
 		expect(
-			queryByText(
+			screen.queryByText(
 				'We auto-populated assets directly from your Final URL. We encourage you to edit or add more in order to best showcase your business.'
 			)
 		).not.toBeInTheDocument();
@@ -56,9 +58,9 @@ describe( 'AssetGroupSection', () => {
 				},
 			};
 		} );
-		const { queryByText } = render( <AssetGroupSection /> );
+		render( <AssetGroupSection /> );
 		expect(
-			queryByText(
+			screen.queryByText(
 				'We auto-populated assets directly from your Final URL. We encourage you to edit or add more in order to best showcase your business.'
 			)
 		).toBeInTheDocument();
--- a/js/src/components/paid-ads/asset-group/asset-group-section.test.js
+++ b/js/src/components/paid-ads/asset-group/asset-group-section.test.js
@@ -1,0 +1,66 @@
+jest.mock( '.~/components/adaptive-form', () => ( {
+	useAdaptiveFormContext: jest
+		.fn()
+		.mockName( 'useAdaptiveFormContext' )
+		.mockImplementation( () => {
+			return {
+				adapter: {
+					baseAssetGroup: { final_url: 'https://example.com' },
+					hasImportedAssets: false,
+					isEmptyAssetEntityGroup: false,
+					resetAssetGroup: jest.fn(),
+				},
+			};
+		} ),
+} ) );
+
+/**
+ * External dependencies
+ */
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+
+/**
+ * Internal dependencies
+ */
+import AssetGroupSection from '.~/components/paid-ads/asset-group/asset-group-section';
+import { useAdaptiveFormContext } from '.~/components/adaptive-form';
+
+jest.mock( '.~/components/paid-ads/asset-group/asset-group-card', () =>
+	jest.fn( ( props ) => <div { ...props } /> ).mockName( 'AssetGroupCard' )
+);
+
+describe( 'AssetGroupSection', () => {
+	test( 'Component renders', () => {
+		const { queryByText } = render( <AssetGroupSection /> );
+		expect( queryByText( /Add additional assets/i ) ).toBeInTheDocument();
+	} );
+
+	test( 'Component not showing Tip if there are no imported assets', () => {
+		const { queryByText } = render( <AssetGroupSection /> );
+		expect(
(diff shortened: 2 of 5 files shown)


js/src/components/paid-ads/campaign-assets-form.js
```diff
@@ -90,13 +91,20 @@ export default function CampaignAssetsForm( {
 			baseAssetGroup,
 			validationRequestCount,
 			assetGroupErrors,
+			hasImportedAssets,
```
eason9487, line 99: As the related code to determine `hasImportedAssets` is very specific to business logic, I would recommend adding some code comment for the "why". For example, the code comment could explain:
- Why it uses `assetGroup` rather than `nextAssetGroup` in the following `if` statement
- Why it needs a separate state rather than `hasImportedAssets: Boolean( baseAssetGroup[ ASSET_FORM_KEY.FINAL_URL ] ),`

js/src/components/paid-ads/asset-group/asset-group-section.scss
```diff
@@ -5,4 +5,20 @@
...
 	}
+
+	.components-tip {
+		padding: $grid-unit-15 $grid-unit-20;
+		background: #f0f6fc;
+		border: $border-width solid #c5d9ed;
+		line-height: $gla-line-height-medium;
+
+		> p {
+			color: $gray-900;
```
eason9487: This `color` style could be placed in the upper `.components-tip`, just like the `line-height` style.

js/src/components/paid-ads/asset-group/asset-group-section.scss
```diff
@@ -5,4 +5,20 @@
...
+			font-size: $default-font-size;
```
eason9487: Looks like this `font-size` could be omitted or also moved to the upper `.components-tip`.

js/src/components/paid-ads/asset-group/asset-group-section.test.js
```diff
@@ -0,0 +1,66 @@
...
+import { useAdaptiveFormContext } from '.~/components/adaptive-form';
+
+jest.mock( '.~/components/paid-ads/asset-group/asset-group-card', () =>
+	jest.fn( ( props ) => <div { ...props } /> ).mockName( 'AssetGroupCard' )
+);
+
+describe( 'AssetGroupSection', () => {
+	test( 'Component renders', () => {
+		const { queryByText } = render( <AssetGroupSection /> );
+		expect( queryByText( /Add additional assets/i ) ).toBeInTheDocument();
```
eason9487: Assertions checking for existence would be a bit better to use `getBy...` in terms of maintainability because it will throw a helpful error message if no element is found.

#### `queryBy...`

It only shows it received a `null`:

image

#### `getBy...`

It prints out the whole document so the dev can see what's rendered and maybe why the query failed to find what they were looking for:

image

js/src/components/paid-ads/asset-group/asset-group-section.test.js
eason9487, line 1: It's recommended to import `screen` from `@testing-library/react` and call all queries from its APIs, e.g. `screen.queryByText( ... )`. The benefit is it no longer needs to keep the `render` call destructure up-to-date as the test changed. Thus, it could improve maintainability a bit.

eason9487 (commented):
Thanks for adding the tip component. Except for a question mentioned in this summary comment, the rest are not blockers.

---

There is a close button ("x") on the design document but the implementation doesn't have it.

image

I would like to be sure it would be in a separate PR or not included in this PR for some other reason?

puntope:
> There is a close button ("x") on the design document but the implementation doesn't have it.

Since all our Tip components (before this one) were without a close button, I discussed with Nikki to confirm if this was a strict requirement or if it was just an update in their Desing Collection. 

Since it's a very tiny component not very intrusive, and it only shows when the Final URL changes and has imported assets. We decided to NOT add the close button for now. 

Supposedly this is an update that will go over the Tip component, we can update it when WordPress updates the component adding the close button, or create our own in the future if we find that closing the Tip is a "must have"

I wanted to ask tho, if you have any other thoughts about this so that we should add the close button and logic.

js/src/components/paid-ads/campaign-assets-form.js
puntope: I'm ok adding documentation. However, I find it self-descriptive `hasImportedAssets` clearly indicates (IMO at least) if yes/no assets were imported. 

Before adding the docs, I would love to clarify some points in case the logic behind my decisions is wrong or there are other more straightforward ways to achieve the same result. 

> Why it uses `assetGroup` rather than `nextAssetGroup` in the following `if` statement
 
This is because `nextAssetGroup` has a callback to `initialAssetGroup` that might have present assets on it, on the other hand, `assetGroup` are only the new assets imported. 

> Why it needs a separate state rather than hasImportedAssets: Boolean( baseAssetGroup[ ASSET_FORM_KEY.FINAL_URL ] ),

From my understanding, the presence of `ASSET_FORM_KEY.FINAL_URL` in `baseAssetGroup` only indicates that a new Final URL was loaded. So it would be lexically incorrect. Yes, It's true that right now in 100% of the cases, there will be imported assets since `BUSSINESS_NAME` is always set as `get_bloginfo( 'name' )`. However, I thought this might change in the future ( for example, if we won't import directly the Blog name)  then `ASSET_FORM_KEY.FINAL_URL` might be filled but might be that nothing was imported... 
 
 
 WDYT?

js/src/components/paid-ads/asset-group/asset-group-section.scss
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2011/commits/625025acbf7e1e9c1c2f0704d332544c0fc7ff21

js/src/components/paid-ads/asset-group/asset-group-section.scss
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2011/commits/625025acbf7e1e9c1c2f0704d332544c0fc7ff21

js/src/components/paid-ads/asset-group/asset-group-section.test.js
puntope: Interesting. Good to know. As reference --> IT appears here https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2011/commits/86ee1bbc6a892e778ab9eb353f76286161f80085

js/src/components/paid-ads/asset-group/asset-group-section.test.js
puntope: > Assertions checking for existence would be a bit better to use getBy...

I disagree with this one. Using `getByText` will throw an exception before calling `.toBeInTheDocument()`. So for example the test "Component not showing Tip if there are no imported assets" will break. 

Besides that, the error provided in the current implementation "Value must be an HTML Element or SVG element but NULL received", seems to me more concise and clearer in terms of checking existence.

puntope:
Howdy @eason9487, thanks for your time reviewing this.

I added some explanations and discussions over your comments, as well as some adjustments. Could you please give me more feedback? Thanks in advance

js/src/components/paid-ads/campaign-assets-form.js
eason9487: > However, I find it self-descriptive `hasImportedAssets` clearly indicates (IMO at least) if yes/no assets were imported.

If without looking further into the design document and just inferring the business logic from the implementation, I would **guess** the rules for whether to display the `Tip` might be all of the following conditions are met:
1. There must be an action first that the user imports the suggested assets by selecting a Final URL via the UI.
2. At least one of the imported suggested assets is applied to the form's default values.
3. When editing assets that have been previously saved, assets got from Google Ads API are not considered as "using imported assets" because it doesn't meet the first rule: There must be an importing action first.

But these were only my **guess**, and I was still unsure what actual rules have to be fulfilled. IMO, the naming of `hasImportedAssets` alone doesn't clearly self-describe the above rules. The questions of "why" mentioned above were also questions I asked myself during my inference when the first round of code reviews.

If it seems to exist a totally equivalent state but was not used, there must be a particular reason why they were not used, otherwise, I would wonder if the developer missed a state that could be directly reused.

> > Why it needs a separate state rather than hasImportedAssets: Boolean( baseAssetGroup[ ASSET_FORM_KEY.FINAL_URL ] ),

> From my understanding, the presence of `ASSET_FORM_KEY.FINAL_URL` in `baseAssetGroup` only indicates that a new Final URL was loaded. So it would be lexically incorrect. [...] However, I thought this might change in the future ( for example, if we won't import directly the Blog name) then `ASSET_FORM_KEY.FINAL_URL` might be filled but might be that nothing was imported...

This is indeed different from my original guess, though they eventually can be achieved by the same implementation as the current one.

> WDYT?

From the design documents, I was/am not actually sure what the exact business logic is, but the current implementation and what the `Tip` is trying to convey seem reasonable.

Unless there were something missing, I think the current implementation is fine. Maybe just add the business logic as you understood it and why it checks all assets regardless of the backend implementation.

js/src/components/paid-ads/asset-group/asset-group-section.test.js
eason9487: It's a suggestion got from the author of `@testing-library`: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

> I disagree with this one. Using `getByText` will throw an exception before calling `.toBeInTheDocument()`. So for example the test "Component not showing Tip if there are no imported assets" will break.

I would say the break of the test doesn't matter because the test eventually failed regardless of which one.

> Besides that, the error provided in the current implementation "Value must be an HTML Element or SVG element but NULL received", seems to me more concise and clearer in terms of checking existence.

Proceeding to call `.toBeInTheDocument()` and showing it got a `null` doesn't actually help debugging any further. Often it's still necessary to add `screen.debug()` temporarily to print out the rendered DOM tree to find the cause, and this is exactly the thrown result of `getBy...` will print out.

eason9487 (approved):
Thanks for the adjustment, and for the confirmation of the close button in https://github.com/woocommerce/google-listings-and-ads/pull/2011#issuecomment-1633629319.

> I wanted to ask tho, if you have any other thoughts about this so that we should add the close button and logic.

No, I'm ok with the current implementation.

I was mainly wanting to confirm the inconsistency between the implementation and design document was decided not to implement or would be added later.

js/src/components/paid-ads/campaign-assets-form.js
puntope: Ok, I will add the mentioned explanations if you think they will help

PR #2012, opened 2023-07-12
Provide more detailed error reasons when unable to complete site verification

### LINKED ISSUE #1725: Detailed error when unable to complete site verification

eason9487:

Closes #1725

This PR solves a task of 📌 [Clarify API error messages](https://github.com/woocommerce/google-listings-and-ads/issues/1993#tasks-clarify-api-errors) in #1993.

In #2007, the handler for `GoogleServiceException` has been added. It also adjusted the errors that may occur when connecting to a Google Merchant Center account. This is the pre-process that must be passed through before doing the site verification process.

This PR is based on the same direction as #2007 to adjust the error handling for site verification methods:
- `SiteVerification::get_token`: retrieve a site meta token
- `SiteVerification::insert`: verify a site token

1. Disconnect Google Merchant Center account (via the Connection Test page) if there is a connected one.
1. Go to step 1 of the onboarding flow
1. `SiteVerification::get_token`
   1. Since this error is easier to be tested in an intentional way, it's recommended to tweak some local code below, such as the invalid `type` or forced to throw a `GoogleServiceException` error.
      https://github.com/woocommerce/google-listings-and-ads/blob/2cab1cd369be2b7f5cfb7945617ba8160bc828bc/src/API/Google/SiteVerification.php#L101-L118
   1. Connect to a Google Merchant Center account.
   1. It should show a clearer error message in the notification UI than before. Previously, the message was "Unable to retrieve site verification token."
1. `SiteVerification::insert`
   1. Use a coming soon plugin, or use a hosting option to prevent the site from being public.
      - I tested with [this one](https://wordpress.org/plugins/minimal-coming-soon-maintenance-mode/) and ticked the "Enable Maintenance Mode" option on its setting page.
   1. Connect to a new Google Merchant Center account, or to the existing account with the same site URL.
   1. It should show a clearer error message in the notification UI than before. Previously, the message was "Unable to insert site verification."

### Additional details:

💡 There are still a few error handling that can continue to be improved along the same direction as #2007 and this PR, but they are outside the scope of #1725. I will open a follow-up issue ~once this PR gets merged~. Created: #2013

### Changelog entry

> Tweak - Provide more detailed error reasons when unable to complete site verification for the Google Merchant Center account being connected in the onboarding flow.

### DIFF
--- a/src/API/Google/SiteVerification.php
+++ b/src/API/Google/SiteVerification.php
@@ -3,6 +3,7 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\Exception\ExceptionWithResponseData;
 use Automattic\WooCommerce\GoogleListingsAndAds\Internal\ContainerAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Internal\Interfaces\ContainerAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
@@ -27,6 +28,7 @@ defined( 'ABSPATH' ) || exit;
 class SiteVerification implements ContainerAwareInterface, OptionsAwareInterface {
 
 	use ContainerAwareTrait;
+	use ExceptionTrait;
 	use OptionsAwareTrait;
 	use PluginHelper;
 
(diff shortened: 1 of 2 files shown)


martynmjones (approved):
Hey @eason9487, thanks for making the error messages more detailed 👍 

Tested locally and can confirm that the `Unable to retrieve site verification token` and `Unable to insert site verification` error messages include more information about the specific issue so LGTM ✅

Issue #634, opened 2021-05-19
Expose errors to allow setup to complete


mikkamp:
### Describe the bug:

Followup from https://github.com/woocommerce/google-listings-and-ads/issues/530#issuecomment-829222356

We have error messages for validating the form, see: https://github.com/woocommerce/google-listings-and-ads/blob/0.5.6/js/src/setup-mc/setup-stepper/setup-free-listings/checkErrors.js#L29

But we only disable the Complete Setup button if there is at least one error: https://github.com/woocommerce/google-listings-and-ads/blob/0.5.6/js/src/setup-mc/setup-stepper/setup-free-listings/index.js#L128

I don't see the actual error messages being shown anywhere, so I have to guess that I'm missing a country to add a shipping rate. Can we add some kind of indication why the "Complete Setup" button is disabled to guide them to completing the setup.

### Expected behavior:
To be shown the error why the "Complete Setup" button is disabled, so I can know what to fix.

### Actual behavior:
"Complete Setup" button is disabled without showing errors.

### Steps to reproduce:
1. Go through the onboarding
2. At step 3 setup shipping for all countries except one
3. Notice that the "Complete Setup" button is disabled but we aren't shown what step to take to resolve it

tijosh:
35355092-hc could benefit from this error logging

mikkamp:
Some additional details from a scenario where the error message would be helpful to indicate that the store address is missing street.

![image](https://user-images.githubusercontent.com/11388669/208656067-4a2e9c0b-d850-41d0-ab38-fd09fc2b9414.png)

Especially since the user has to manually confirm policy compliance checks it looks like the disabled continue button is related to the policy compliance checks, but in this case it's because of the store address. Highlighting the exact error would prevent confusion.

Brianmitchtay:
7973235-hc could be benefitted from this error message displaying as well

sukafia:
6192158-zen didn't enter the store address fully and so couldn't complete the setup. (They had no way of knowing that was the issue)

sukafia:
6350049-zen

sukafia:
6350049-zen

Brianmitchtay:
Seeing another case of this in 6485161-zen

I'm not sure if we even have a list of the specific requirements available to HEs or users to know what is required here? Can we add this to the documentation? It's pretty frustrating for the user and for the HE to not have an easy way to solve this.

[...]

PR #2019, opened 2023-07-20
Show validation errors on step 2 of the onboarding flow when unable to continue

eason9487:

This is a part of implementation for #634 and for 📌 [Display form validation errors](https://github.com/woocommerce/google-listings-and-ads/issues/1993#tasks-display-validation-errors) in #1993.

To show validation errors on step 2 of the onboarding flow when unable to continue, this PR:

#### Extract the shareable component and adjust the layouts of shipping time components

- Add a new component `ValidationErrors` for showing form validation error messages, and replace the errors rendering in `AssetGroupCard` with it.
- Move the Card layouts from `ShippingTimeSection` into `ShippingTimeSetup` and remove unneeded codes.

#### Fix and adjust validator `js/src/components/free-listings/configure-product-listings/checkErrors.js`

- Fix incorrect keys and rephrase a few messages.
- Update the validation of tax rate data to check if the store country code is `US`.

#### Add the mechanism of showing validation errors to the product listings form

- Request showing validation errors when clicking the submit of product listings form if any data is invalid.
- Add a function to render validation errors to the product listings form.
- Add the rendering of validation errors to:
   - `ChooseAudienceSection`
   - `FlatShippingRatesInputCards`
   - `EstimatedShippingRatesCard`
   - `OfferFreeShippingCard`
   - `MinimumOrderCard`
   - `ShippingTimeSetup`
   - `TaxRate`

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/4c977398-fb9b-4abc-9ef6-fc551cca362b

Some free listings data are unable to clean via UI. It might need to run the following SQL to clean them from database.

```sql
DELETE FROM wp_options WHERE option_name = 'gla_target_audience';
DELETE FROM wp_options WHERE option_name = 'gla_merchant_center';
```

2. Before clicking the "Continue" button, the UI should not show any validation errors even if there are invalid data.
3. Check if the "Continue" button is always enabled no matter whether the data are valid or not.
4. With some invalid data, click the "Continue" button to see if the validation errors are shown in the corresponding places.
5. Change data to view if the validation errors are shown or hidden accordingly.
6. Make all data valid and click the "Continue" button to check if it goes to step 3.

### Additional details:

💡 The implementation for step 3 will be done by subsequent PR(s)


js/src/components/validation-errors/validation-errors.js
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	return (
+		<ul className="gla-validation-errors">
+			{ messages.map( ( message ) => (
+				<li key={ message }>{ message }</li>
+			) ) }
+		</ul>
+	);
+}
```
puntope: This function was a bit hard for me to read, ideally, I would expect to only allow an array.

If that's not possible I think using another naming instead of `messages` for the local variable would make it easier to understand. At least for me, the fact that the local variable `messages` is the same as the argument `massages` an right after, we assigned it to `rawMessages` wich is the name override for `messages` made it hard to understand at first sight.

A potential way to do it might be:

```jsx
export default function ValidationErrors( { messages } ) {
        let messagesList = messages;  

        if ( ! messages?.length ) {
		return null;
	}
	
        if ( ! Array.isArray( messages ) ) {
		messagesList = [ messages ];
	}
	

	return (
		<ul className="gla-validation-errors">
			{ messagesList.map( ( message ) => (
				<li key={ message }>{ message }</li>
			) ) }
		</ul>
	);
}
```

puntope:
> DELETE FROM wp_options WHERE option_name = 'gla_target_audience';
DELETE FROM wp_options WHERE option_name = 'gla_merchant_center';

I was able to perform the test without this. Might you explain why is this needed? Does the user (or us programmatically) handle this after the user upgrades with this feature?

puntope:
Is there any place where I can check the new validation error texts? Or there are no specific requirements?

puntope:
When I was testing the errors I discover some inconsistency with the Audience Selector and the rest of inputs.


https://github.com/woocommerce/google-listings-and-ads/assets/5908855/af27482e-c0f3-484b-b72e-726cae4c9425

As you can see in the video, the inputs are not showing validation errors until the validation error in the Audience selector is solved.

puntope (commented):
I check the requirements for this feature:

✅  Don’t disable the “Continue” or “Compete” buttons even if the form validation doesn’t pass.

✅  The form validation errors are not displayed until the user clicks the above button.

⚠️  I was unable to find a reference for the texts in the `errors.shipping_country_rates` and `errors. shipping_country_times`

⚠️  Some inconsistencies in the error renderization (see comments)

💅  A minor suggestion for the function `ValidationErrors()`

js/src/components/validation-errors/validation-errors.js
eason9487: Renamed in https://github.com/woocommerce/google-listings-and-ads/pull/2019/commits/10df591df6bdf0c0bc8ac68e399776c7cab7544a.

eason9487:
Hi @puntope, thanks for the review.

This PR is ready for the second round of code reviews. Could you help with it?

> > DELETE FROM wp_options WHERE option_name = 'gla_target_audience';
> > DELETE FROM wp_options WHERE option_name = 'gla_merchant_center';
> 
> I was able to perform the test without this. Might you explain why is this needed? Does the user (or us programmatically) handle this after the user upgrades with this feature?

The selected **Location** option value ("Selected countries only" or "All countries") is stored in the `gla_target_audience`. The selected **Shipping rates** and **Tax rate (required for U.S. only)** option values are stored in the `gla_merchant_center`. These values are unable to be cleared via the UI. Deleting them from DB is a way to test step 2 as in the case of the first time entering the onboarding flow.

image

P.S. **Location** and **Shipping rates** will be filled in automatically when empty, so this PR didn't add validation errors for them but it's still a bit better to double-confirm that the automatic file in processing works.

> Is there any place where I can check the new validation error texts? Or there are no specific requirements?

No. Except for PMax Assets and phone verification, no previous development or design documents have specified the texts for validation errors.

The main reason for adjusting some texts was to avoid the following situation, which happens to be a line break at the last word.

image

> When I was testing the errors I discover some inconsistency with the Audience Selector and the rest of inputs. [...] As you can see in the video, the inputs are not showing validation errors until the validation error in the Audience selector is solved.

This is expected, as each field focuses on different validation errors. When no audience country is selected, the relevant validation errors are already displayed in the most relevant position. It would be a bit redundant to display the same error on both Shipping rates and Shipping times.

However, I noticed an issue from your video: after clearing a value equal to `0` in Shipping rates or Shipping times and moving out the focus, the `AppInputNumberControl` should be backfilled with 0 again but it didn't.

I tried a couple of different approaches and finally realized that there was only one way to solve the issue first with a relatively small amount of changes, while the rest of the approaches involved too many changes. Please view the fix in https://github.com/woocommerce/google-listings-and-ads/pull/2019/commits/a8f9bf90a5b864956a8fe784582907b8aa850dd1.

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/4d5237d3-f759-417b-a499-00b6af1de3c9

js/src/components/app-input-number-control/index.js
```diff
@@ -36,6 +41,17 @@ const AppInputNumberControl = ( props ) => {
 	const numberSettings = useStoreNumberSettings( settings );
 	const numberFormat = useNumberFormat( settings );
 
+	// The `InputControl` in `@wordpress/components` doesn't take into account the context
+	// in which the `value` of a controlled component is updated via `onBlur`, resulting in
+	// the incoming `value` prop may not syncing to the <input> element's display value.
+	// Therefore, a synchronization needle is used to make the value synchronized via the
+	// onChange callback.
```
puntope: Thanks for the changes. They look good. The only thing is that "sync Needle" naming, seems to me a kinda personal concept name? . I was checking around the internet and I didn't find anything regarding "synchronisation needle". So I would suggest using something more descriptive like `refreshInput`

puntope (approved):
✅  LGTM

Thanks for the adjustments, explanation and extra Unit tests.

I left one last 💅  comment in regards to that "syncNeedle" name.

js/src/components/app-input-number-control/index.js
eason9487: The naming was inspired by the reset needle.

resetrouter

Changed to adopt [an official approach and naming](https://legacy.reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate) in https://github.com/woocommerce/google-listings-and-ads/pull/2019/commits/f763449a5b47efffe9b3ddd3a9a12dc711563a56.

puntope: Much better, thanks for this and for the reference.

PR #2025, opened 2023-07-26
Show validation errors on step 3 of the onboarding flow when unable to continue

### LINKED ISSUE #634: Expose errors to allow setup to complete

eason9487:

Closes #634 

This is the last part for #634 and for 📌 [Display form validation errors](https://github.com/woocommerce/google-listings-and-ads/issues/1993#tasks-display-validation-errors) in #1993.

To show validation errors on step 3 of the onboarding flow when unable to continue, this PR:

#### Adjust/add validator

- Adjust the error messages in the pre-launch checklist validator to a centralized one.
- Expose the missing required fields of store address from `useStoreAddress` hook and add new util to map the missing fields of store address to corresponding error messages.

#### Add the mechanism for showing validation errors

- Request showing validation errors when clicking the submit of store requirements form if any data is not ready.
- Add a function to render validation errors to the store requirements form.
- Add the rendering of validation errors for the phone number verification.
- Add the rendering of missing required fields for the store address setup.
- Add the rendering of validation errors to `PreLaunchChecklist`.

#### Event tracking

- Add event tracking to log how many times and what fields the store address is having validation errors.
- Update `src/Tracking/README.md`.

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/335d836b-7f79-483a-ade4-5e00493ebc44

The country of WC store address data is unable to clean via WC Admin Settings. It might need to run the following SQL to clean it from database.

```sql
DELETE FROM wp_options WHERE option_name = 'woocommerce_default_country';
```

💡 WC/GLA will still use `US` as the fallback country code if the `woocommerce_default_country` is not available. Therefore, the case of missing country filed will not be able to test in practice.

1. Start the onboarding flow with a newly created Google Merchant Center account. So that it can test the verification of phone number later.
2. Proceed the flow until step 3
3. Before clicking the "Continue" button, the UI should not show any validation errors even if there is unverified phone number, invalid store address or unchecked pre-launch checklist.
4. With some invalid data, click the "Continue" button to see if the validation errors are shown in the corresponding places.
5. Update the data to view if the validation errors are shown or hidden accordingly.
6. Make all data valid and click the "Continue" button to check if it goes to step 4.

### Additional details:

The validation error message for phone number is placed in a different location from the other Card UI because the content of phone number card is more varied and complex, and there is no relatively suitable location at the bottom to display the validation error.

image

image

### Changelog entry

> Update - Show validation errors on steps 2 and 3 of the onboarding flow when unable to continue.

### DIFF
--- a/js/src/components/contact-information/mapStoreAddressErrors.js
+++ b/js/src/components/contact-information/mapStoreAddressErrors.js
@@ -1,0 +1,54 @@
+/**
+ * External dependencies
+ */
+import { __, _x, sprintf } from '@wordpress/i18n';
+
+/**
+ * @typedef {import('.~/hooks/types.js').StoreAddress} StoreAddress
+ */
+
+/**
+ * Maps the missing fields of store address to corresponding error messages.
+ *
+ * @param {StoreAddress} storeAddress [description]
+ * @return {string[]} Error massages of store address.
+ */
+export default function mapStoreAddressErrors( storeAddress ) {
+	// The sources of the possible field names:
+	// - https://github.com/woocommerce/google-listings-and-ads/blob/2.5.0/src/API/Google/Settings.php#L322-L339
+	// - https://github.com/woocommerce/woocommerce/blob/7.9.0/plugins/woocommerce/includes/class-wc-countries.php#L841-L1654
+	const fieldNameDict = {
+		address_1: _x(
+			'address line',
+			'The field name of the address line in store address',
+			'google-listings-and-ads'
+		),
+		city: _x(
+			'city',
+			'The field name of the city in store address',
+			'google-listings-and-ads'
+		),
+		country: _x(
+			'country/state',
+			'The field name of the country in store address',
+			'google-listings-and-ads'
+		),
+		postcode: _x(
+			'postcode/zip',
+			'The field name of the postcode in store address',
+			'google-listings-and-ads'
+		),
+	};
+
+	return storeAddress.missingRequiredFields.map( ( field ) => {
+		const fieldName = fieldNameDict[ field ] || field;
+		return sprintf(
+			// translators: %s: The missing field name of store address.
+			__(
+				'The %s of store address is required.',
+				'google-listings-and-ads'
+			),
+			fieldName
+		);
+	} );
+}
--- a/js/src/components/contact-information/mapStoreAddressErrors.js
+++ b/js/src/components/contact-information/mapStoreAddressErrors.js
@@ -14,9 +14,10 @@ import { __, _x, sprintf } from '@wordpress/i18n';
  * @return {string[]} Error massages of store address.
  */
 export default function mapStoreAddressErrors( storeAddress ) {
-	// The sources of the possible field names:
-	// - https://github.com/woocommerce/google-listings-and-ads/blob/2.5.0/src/API/Google/Settings.php#L322-L339
-	// - https://github.com/woocommerce/woocommerce/blob/7.9.0/plugins/woocommerce/includes/class-wc-countries.php#L841-L1654
+	// The possible fields to be mapped are defined in the first file,
+	// and their specs come from the second file:
+	// 1. https://github.com/woocommerce/google-listings-and-ads/blob/2.5.0/src/API/Google/Settings.php#L322-L339
+	// 2. https://github.com/woocommerce/woocommerce/blob/7.9.0/plugins/woocommerce/includes/class-wc-countries.php#L841-L1654
 	const fieldNameDict = {
 		address_1: _x(
 			'address line',
--- a/js/src/components/contact-information/phone-number-card/phone-number-card.js
+++ b/js/src/components/contact-information/phone-number-card/phone-number-card.js
@@ -12,6 +12,7 @@ import { Spinner } from '@woocommerce/components';
 import AccountCard, { APPEARANCE } from '.~/components/account-card';
 import AppButton from '.~/components/app-button';
 import AppSpinner from '.~/components/app-spinner';
+import ValidationErrors from '.~/components/validation-errors';
 import VerifyPhoneNumberContent from './verify-phone-number-content';
 import EditPhoneNumberContent from './edit-phone-number-content';
 import './phone-number-card.scss';
@@ -27,7 +28,11 @@ const basePhoneNumberCardProps = {
  * @typedef { import(".~/hooks/useGoogleMCPhoneNumber").PhoneNumber } PhoneNumber
  */
 
-function EditPhoneNumberCard( { phoneNumber, onPhoneNumberVerified } ) {
+function EditPhoneNumberCard( {
+	phoneNumber,
+	showValidation,
+	onPhoneNumberVerified,
+} ) {
 	const { loaded, data } = phoneNumber;
 	const [ verifying, setVerifying ] = useState( false );
 	const [ unverifiedPhoneNumber, setUnverifiedPhoneNumber ] = useState(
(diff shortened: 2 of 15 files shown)


js/src/components/contact-information/phone-number-card/phone-number-card.js
```diff
@@ -27,7 +28,11 @@ const basePhoneNumberCardProps = {
  * @typedef { import(".~/hooks/useGoogleMCPhoneNumber").PhoneNumber } PhoneNumber
  */
 
-function EditPhoneNumberCard( { phoneNumber, onPhoneNumberVerified } ) {
+function EditPhoneNumberCard( {
```
puntope, line 31: Not related really, but it would be nice to take advantage of the touches in this component to move  `<EditPhoneNumberCard>` to its own file.

puntope:
Not really related to this PR, as it happened before as well. But we can maybe consider this edge case in this or future revisions:

- Have a correct store address with no errors.
- Rest of the fields have no errors
- Change the address in WC, for example removing the ZIP code.
- DONT click on refresh
- Click continue (Validation pass, even when there is no ZIP code)
- Go to GLA Settings
- See the error in the address

I would expect to do an extra sanity check when clicking on continue to show the error.

js/src/components/contact-information/mapStoreAddressErrors.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Maps the missing fields of store address to corresponding error messages.
+ *
+ * @param {StoreAddress} storeAddress [description]
+ * @return {string[]} Error massages of store address.
+ */
+export default function mapStoreAddressErrors( storeAddress ) {
+	// The sources of the possible field names:
+	// - https://github.com/woocommerce/google-listings-and-ads/blob/2.5.0/src/API/Google/Settings.php#L322-L339
+	// - https://github.com/woocommerce/woocommerce/blob/7.9.0/plugins/woocommerce/includes/class-wc-countries.php#L841-L1654
```
puntope: 💅  I was, at first sight, a bit confused by this comment.

Maybe we can say "The fields to be validated are defined in these 2 files:"

js/src/components/contact-information/store-address-card.js
```diff
@@ -29,24 +32,62 @@ import './store-address-card.scss';
...
+	const path = getPath();
 	const { subpath } = getQuery();
-	const editButton = (
+
+	const refetchedCallbackRef = useRef( null );
+
+	if ( loaded && refetchedCallbackRef.current ) {
+		refetchedCallbackRef.current( data );
+		refetchedCallbackRef.current = null;
+	}
```
puntope, line 66: Can we add some comments explaining why is this code for, I mean the need for recreating the callback instead of just doing:

 ```jsx
 if ( loaded ) {
		const eventProps = {
				path,
				subpath,
				country_code: storeAddress.countryCode,
				missing_fields: storeAddress.missingRequiredFields.join( ',' ),
			};

			recordEvent( 'gla_wc_store_address_validation', eventProps );
 }
 ```

puntope:
Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2025#issuecomment-1653306696 

I couldn't test it well since I have no other phone number, but I wonder what will happen in this edge case:

- Have an MC account with all correct (including verified phone number)
- Go to Google MC and change the phone number 
- Go to Edit the previous MC account 

In any case, seems unrelated to this PR since (in case it's really an issue)  it's happened previously also. So just pointing out my concerns here.

puntope (commented):
Hi @eason9487,  thanks for the work on this. I can see a big improvement now in the Onboarding experience.

Overall looking great and working as expected, just before merging I wanted to verify/discuss my comment here https://github.com/woocommerce/google-listings-and-ads/pull/2025#issuecomment-1653306696

In case you think is a blocker before approval.

The rest are 💅  code suggestions mainly.

js/src/components/contact-information/phone-number-card/phone-number-card.js
eason9487: I lean toward keeping it as it is.

js/src/components/contact-information/mapStoreAddressErrors.js
eason9487: I didn't take the suggestions because they're not actually being validated here. Adjusted it for more detail in https://github.com/woocommerce/google-listings-and-ads/pull/2025/commits/023914a9ea6455e577c81e9c0819f471a1045bd0.

js/src/components/contact-information/store-address-card.js
eason9487: The code comments for this event tracking were already added above. I didn't add more comments to explain why the implementation was made in this way as it would be probably only about how React's component and hook work together.

https://github.com/woocommerce/google-listings-and-ads/blob/023914a9ea6455e577c81e9c0819f471a1045bd0/js/src/components/contact-information/store-address-card.js#L49-L56

https://github.com/woocommerce/google-listings-and-ads/blob/023914a9ea6455e577c81e9c0819f471a1045bd0/js/src/components/contact-information/store-address-card.js#L36-L38

eason9487:
Hi @puntope, thanks for the review. Could you help with a new round of code reviews?

> Not really related to this PR, as it happened before as well. But we can maybe consider this edge case in this or future revisions:
> 
> * Have a correct store address with no errors.
> * Rest of the fields have no errors
> * Change the address in WC, for example removing the ZIP code.
> * DONT click on refresh
> * Click continue (Validation pass, even when there is no ZIP code)
> * Go to GLA Settings
> * See the error in the address
> 
> I would expect to do an extra sanity check when clicking on continue to show the error.

Currently, users were reporting problems because they were not prompted for the missing address fields and didn't know why they couldn't continue. This edge case can happen but in practice users are hardly likely to do so.

Even if it does happen, both GLA Settings and Google Merchant Center will show the address as incorrect, which is enough to allow users to correct the problem further. So I think it's only worthwhile to deal with such a case if it's been reported a few times in the future.

> I couldn't test it well since I have no other phone number, but I wonder what will happen in this edge case:
> 
> * Have an MC account with all correct (including verified phone number)
> * Go to Google MC and change the phone number
> * Go to Edit the previous MC account
> 
> In any case, seems unrelated to this PR since (in case it's really an issue) it's happened previously also. So just pointing out my concerns here.

Google Merchant Center's requirement is to have a verified phone number, changing to another verified one still meets the requirement.

Furthermore, Google Merchant Center doesn't allow for a verified phone number to be cleared, nor allow for an unverified phone number to be saved, so there shouldn't be any issues except the phone number showing differently on GLA UI for that moment.

puntope (approved):

PR #2026, opened 2023-07-26
Add task using `TaskLists::add_task` API

### LINKED ISSUE #1704: Migrate JS `woocommerce_admin_onboarding_task_list` to PHP `TaskLists::add_task()`
### Describe the bug:

For some time WooCommerce produces a warning:
```
woocommerce_admin_onboarding_task_list is deprecated and will be removed from @woocommerce/data in version 2.10.0. Please use TaskLists::add_task() instead.
```
WooCommerce 7.0.0 [uses `@woocommerce/data@4.1.0`](https://github.com/woocommerce/woocommerce/blob/7.0.0-beta.2/packages/js/data/package.json#L3) and [WC 6.7.0 uses 4.0.0](https://github.com/woocommerce/woocommerce/blob/6.7.0/packages/js/data/package.json#L3) so we are already far behind.

I think we should migrate it, as it may stop working at any moment.

### Steps to reproduce:

1. Install GLA
2. Start onboarding
3. Open browser console

### Expected behavior:

No warnings should be there

### Actual behavior:

There is one, as mentioned above.

tomalec:

- Add task using `TaskLists::add_task` API, `woocommerce_admin_onboarding_task_list` is deprecated.
- Remove `add_woocommerce_extended_task_list_item` & `remove_woocommerce_extended_task_list_item` hooks.
- [Add npm script to generate hook's docs.](https://github.com/woocommerce/google-listings-and-ads/pull/2026/commits/d64ae3621daa16a2d7d6d201a76b000780b819ea) 
   Re-generate those docs.
- [Add tests for CompleteSetupTask](https://github.com/woocommerce/google-listings-and-ads/pull/2026/commits/2cc3cd79049e2d22cf2b41be35835da4a604029d)

### Screenshots:
image

1. With incomplete onboarding, go to **WooCommerce > Home** `/wp-admin/admin.php?page=wc-admin`
2. You should see a dismissable, incomplete task to "Set up Google Listings & Ads (20 minutes)"
3. Click on the task to get redirected to the GLA start page
4. Complete the onboarding
5. go back to **WooCommerce > Home** `/wp-admin/admin.php?page=wc-admin`
6. You should see the complete setup GLA task that, when clicked, redirects to the GLA dashboard.

### Additional details:

1. The `register` method is being called a few times for some reason. I don't know how to make sure the callback is called only once in https://github.com/woocommerce/google-listings-and-ads/pull/2026/files#diff-b45e7e0b771cc2f644e3a4805a651b28213c4f6464b615c7c26e947275a3d52fR27-R39
   Currently, it does not introduce any problem, just does unnecessary runs, but it could lead to unexpected quirks in the future.
   Tests are checking it was called [at least once](https://github.com/woocommerce/google-listings-and-ads/pull/2026/files#diff-0fcabbd1197a594303df3dea2b0873564aeabc0b42f585d8d5db8daba6a9d89fR45).

### Changelog entry

> Tweak - Use the latest API to add an item to the WC tasks list.
> Break - Remove `add_woocommerce_extended_task_list_item` and `remove_woocommerce_extended_task_list_item` hooks.


tests/Unit/TaskList/CompleteSetupTaskTest.php
```diff
@@ -0,0 +1,99 @@
...
+	protected $task;
+
+	/**
+	 * Runs before each test is executed.
+	 */
+	public function setUp(): void {
+		parent::setUp();
+		// Mock tasks list.
+		TaskLists::clear_lists();
+		TaskLists::add_list( [ 'id' => 'extended' ] );
```
mikkamp: All the WP hooks have already been called before the unit tests starts. So at this point it's already triggered the `register` code and `init` code which adds the task to the list.

So what we are really doing here is clearing the WC task list with our custom task added. And then we try to add it again in `test_register` but we shouldn't trigger the `init` hook a second time as that calls all kinds of code we don't want to run a second time.

My suggestion would be to remove these lines of code and leave TaskLists as is (so we are working with a non mocked version of TaskLists).

tests/Unit/TaskList/CompleteSetupTaskTest.php
```diff
@@ -0,0 +1,99 @@
...
+	 */
+	public function test_register() {
+		$this->task->register();
+		do_action( 'init', 1 );
+
+		$this->assertGreaterThan( 0, did_action( 'add_woocommerce_extended_task_list_item' ) );
+
+		$this->assertNotNull( TaskLists::get_list( 'extended' )->get_task( $this->task->get_id() ) );
+		$this->assertEquals( $this->task, TaskLists::get_list( 'extended' )->get_task( $this->task->get_id() ) );
+	}
```
mikkamp: As mentioned in the comment above we shouldn't register and call init a second time here. Instead we should just test that it was registered and the hook was called.
The task is never going to be an identical object so at most I'd compare the ID or just leave it at assertNotNull since we are already fetching a specific ID so if that returns something it should always be the right one.

```suggestion
	public function test_register() {
		$this->assertEquals( 1, did_action( 'add_woocommerce_extended_task_list_item' ) );
		$this->assertNotNull( TaskLists::get_list( 'extended' )->get_task( $this->task->get_id() ) );
		$this->assertEquals(
			$this->task->get_id(),
			TaskLists::get_list( 'extended' )->get_task( $this->task->get_id() )->get_id()
		);
	}
```

tests/Unit/TaskList/CompleteSetupTaskTest.php
```diff
@@ -0,0 +1,99 @@
...
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class CompleteSetupTaskTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\TaskList
+ */
+class CompleteSetupTaskTest extends UnitTest {
+
+	/** @var MerchantCenterService $merchant_center */
```
mikkamp: This is a mocked object so we usually show that as:
```suggestion
	/** @var MerchantCenterService|MockObject $merchant_center */
```

mikkamp (commented):
Thanks for the changes. It's working well and I'm seeing the task being added.
I just left some comments on how to fix the tests as the WP hooks shouldn't be triggered twice.

src/TaskList/CompleteSetupTask.php
```diff
@@ -0,0 +1,111 @@
...
+	public function register(): void {
+		add_action(
+			'init',
+			function() {
+				$list_id = 'extended';
+
+				$this->task_list = TaskLists::get_list( $list_id );
+				TaskLists::add_task( $list_id, $this );
+
+				do_action( 'add_woocommerce_extended_task_list_item', $this->get_id() );
```
mikkamp: Seems a previous comment I made about this hook got lost. But I was questioning if we still need to call this hook, we don't use it anywhere in our code and neither does WC need it to add the task to the list. It seems convenient for testing to confirm the function was called, but as long as we are testing the task got added to the list then I don't think that's necessary to test.

tomalec: I just kept it as it was to minimize the set of changes, but if you think we don't need that hook. I'm always for YAGNI and less code :)

tests/Unit/TaskList/CompleteSetupTaskTest.php
tomalec: Thanks for the explanation :)

Implemented in https://github.com/woocommerce/google-listings-and-ads/pull/2026/commits/e9d7c2178f24d21356b5b3349d5604b1aeecae53

But that means we will not be asserting the actual timing of the registration.

Are we fine with that?

tests/Unit/TaskList/CompleteSetupTaskTest.php
tomalec: Maybe we should not create any instance but test the behavior of the already registered one 
https://github.com/woocommerce/google-listings-and-ads/pull/2026/commits/1c7b5fba988d0265e99b7dae5122b768e6462153

tests/Unit/TaskList/CompleteSetupTaskTest.php
tomalec: Also, CodeCov complains about no coverage for the `register` method https://app.codecov.io/gh/woocommerce/google-listings-and-ads/pull/2026/blob/src/TaskList/CompleteSetupTask.php#L27

tomalec:
Thank you @mikkamp  for the great suggestions. I changed the tests a bit, so everything is passing. However, Codecov still complains about the lack of coverage for `register`. 
Would you mind another round?

mikkamp (approved):
Thanks that looks a lot simpler.

I can understand the desire to get 100% unit test coverage. Although this case might be a bit tricky since the register function would be better covered in an integration test. PHPunit looks for  a line of code in the tests which triggers the "covered" code, but that's only happening (triggered by bootstrap.php) before it starts the actual tests. So at most you could add `@codeCoverageIgnore` to the register function (but we haven't done that much in other cases).

Also note that if register / init would not be called then this part of the test would fail:
```php
		// Fetch the task from the global list.
		$this->task = TaskLists::get_list( 'extended' )->get_task( 'gla_complete_setup' );
		$this->task->set_merchant_center_object( $this->merchant_center );
```
Because `$this->task` would be false if the task is not found, so the function `set_merchant_center_object` can't be called. So we are testing the result of register, but just not proving it in the code coverage. One improvement we could do is `assertInstanceOf` to confirm that `$this->task` is the expected object, as that will help with troubleshooting if it ever fails.

Either way I'm going to approve the PR as the actual code changes performs as expected.

tomalec:
Thanks for the clarifications. I agree it's more of an integration test, but I was pretty surprised that Codecov reported as "uncovered" lines that were actually executed. It's a pity it was covered before Codecov checks, but I think as long as it's actually tested, we're safe.


> Also note that if register / init would not be called then this part of the test would fail:

Yep, that's why I put it there. There is a small downside that it's in `setup` not `test`, so troubleshooting could be a bit less straightforward. I've added a failure message in `setUp` to make it clearer https://github.com/woocommerce/google-listings-and-ads/pull/2026/commits/37332943e19eca5585ed71798d55d0f1c3017715
Also, I think the `assertInstanceOf` can help https://github.com/woocommerce/google-listings-and-ads/pull/2026/commits/5565e313b501c2e0d9d7fd2e4d565b37fd027ffe.

PR #2028, opened 2023-07-28
Add To Cart Tracking support for Products (Beta) Block

### LINKED ISSUE #2020: Confirm compatibility with Blocks
### Describe the issue:
As mentioned in https://github.com/woocommerce/woocommerce-blocks/issues/9212 there could potentially be some compatibility issues with blockified content. In particular with the tracking code.

### Tasks:
- [x] The `add_to_cart` hook is only used for analytics tracking and already has an alternative for blocks: https://github.com/woocommerce/google-listings-and-ads/blob/2.5.0/js/src/gtag-events/index.js#L18-L25
Double check it though to make sure there wasn't any changes since we use the hook prefix `experimental__woocommerce_blocks`, in case that's no longer experimental.

- [x] The Order Confirmation page, at the time of creating this wasn't fully "blockified" yet and was using a separate WP page which would call the hook. Recheck this to make sure it's still called with a fully blockified cart/checkout process.

- [x] We should also double check the following warning:
> ⚠️ Order confirmation page - works ok for now, in most cases (sometimes it doesn't fire, but I don't think that's blocks specific) however be careful when the order confirmation page is blockified, we should ensure the action still fires.

Uncover any details on which scenarios it doesn't fire.

puntope:

Closes #2020 

This PR adds a fix for triggering the Add To Cart event when it takes place on Products (Beta) Block.

1. Install Google Analytics Debugger extension
2. Install GLA, complete the setup and connect Google Ads as well
3. Install a theme with Blocks  support, for example Storefront
4. Create a new page (Page 1), and insert the block "All Products"
5. Crate a new page (Page 2), and insert the block "Products (beta)"
6. Go to Shop page in the store, click on Add to cart, see the 'add_to_cart' event is logged once by Google Analytics Debugger
7. Go to "Page 1", click on Add to cart, see the 'add_to_cart' event is logged once by Google Analytics Debugger
8. Go to "Page 2", click on Add to cart, see the 'add_to_cart' event is logged once by Google Analytics Debugger  

### Changelog entry

> Fix - Fix support for "add_to_cart" event in Products (Beta) block

### DIFF
--- a/js/src/gtag-events/index.js
+++ b/js/src/gtag-events/index.js
@@ -75,6 +75,23 @@ document.defaultView.addEventListener( 'DOMContentLoaded', function () {
 			button.addEventListener( 'click', addToCartClick );
 		} );
 
+	/**
+	 * Fix for Products (Beta) block
+	 *
+	 * Products (Beta) block doesn't trigger addAction events. Also it's not being queried by the previous query selector
+	 * because we added :not( .wc-block-components-product-button__button ) to prevent tracking duplicates with
+	 * other blocks that yes trigger addAction events.
+	 *
+	 * So the fix is to query again specifically the add to cart button in Products (Beta) block
+	 */
+	document
+		.querySelectorAll(
+			'[data-block-name="woocommerce/product-button"] > .add_to_cart_button:not( .product_type_variable ):not( .product_type_grouped )'
+		)
+		.forEach( ( button ) => {
+			button.addEventListener( 'click', addToCartClick );
+		} );
+
 	document
 		.querySelectorAll( '.single_add_to_cart_button' )
 		.forEach( ( button ) => {


js/src/gtag-events/index.js
```diff
@@ -75,6 +75,23 @@ document.defaultView.addEventListener( 'DOMContentLoaded', function () {
 			button.addEventListener( 'click', addToCartClick );
 		} );
 
+	/**
+	 * Fix for Products (Beta) block
+	 *
+	 * Products (Beta) block doesn't trigger addAction events. Also it's not being queried by the previous query selector
+	 * because we added :not( .wc-block-components-product-button__button ) to prevent tracking duplicates with
+	 * other blocks that yes trigger addAction events.
```
jorgemd24, line 83: Perhaps we can mention that Products (Beta) is rendered in the backend, which means the `addAction(experimental__woocommerce_blocks--cart-add-item ...)` won't work.

I think the "yes" could be omitted here: `other blocks that ->yes<- trigger addAction events`

jorgemd24 (approved):
Thanks, @puntope, for fixing this, it works as expected 👍 . It took me some time to figure out why the hook `experimental__woocommerce_blocks--cart-add-item` doesn't work with "Products Beta."

I noticed that in Products Beta, the "Add to the cart" button is rendered in the backend:

https://github.com/woocommerce/woocommerce-blocks/blob/e4e57d465419aa7901d171754bf6ef2fe3ce3dae/src/BlockTypes/ProductButton.php#L40-L117

On the other hand, "All products" are rendered using React, and the hook `experimental__woocommerce_blocks--cart-add-item` is attached in the click event:

https://github.com/woocommerce/woocommerce-blocks/blob/e4e57d465419aa7901d171754bf6ef2fe3ce3dae/assets/js/atomic/blocks/product-elements/button/block.tsx#L91-L95

So, do you think that a similar issue could be affecting the google-analytics extension?

https://github.com/woocommerce/woocommerce-google-analytics-integration/blob/730da93ae7820a7a8db4c719d2b7b5f2a2c46b7f/assets/js/src/actions.js#L114-L118

puntope:
> So, do you think that a similar issue could be affecting the google-analytics extension?

Yes, Google Analytics has the same issue. And all the plugins relying on that hook using Products (Beta) but this PR only achieves the fix for GLA

Issue #1599, opened 2022-07-18
Implement new E2E environment once its ready in Woo platform


tomalec:
## ~User~developer story
While [trying to add `admin-e2e-tests`](https://github.com/woocommerce/google-listings-and-ads/pull/1593), I encountered some problems with `e2e-environment` while trying to fix them I got notified that [`e2e-environment` is actually deprecated](https://github.com/woocommerce/woocommerce/pull/33828#issuecomment-1184729665), and there is a new solution being baked: https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/e2e, but still not ready for extensions yet.

Once, something ready and stable for extension use will appear, we'd better implement it, as probably by that time, our dependencies will get pretty far behind.

### Is your feature request related to a problem? 

We cannot update Jest to `28`, due to `e2e-environment` dependencies problems

#### How to reproduce the problem

See https://github.com/woocommerce/woocommerce/pull/33852, https://github.com/woocommerce/woocommerce/pull/33828, https://github.com/woocommerce/woocommerce/pull/33827

### Describe the solution you'd like 

Wait for an E2E solution with clear setup instructions to be prepared, for the Woo extensions. Then just follow the instructions to implement it.

### Describe alternatives you've considered 

1. Revive `e2e-environment` by trying to push updates and fixes there
2. Try adopting Playwright solution today

## Unknowns

- When the new solution would be released for external use?
- We would have to balance the costs of additional effort for early adopting & contributing to the platform and working with stale, legacy, non-supported `e2e-environment`

findingsimple:
> https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/e2e, but still not ready for extensions yet.

I can't find that now

Update - [Found this thought](https://github.com/woocommerce/woocommerce/tree/fddee1a237ae0c79a96513aa8ba24c398a520ed8/plugins/woocommerce/e2e)

~@tomalec is there an internal project you can point me to, I can't seem to find anything~

Ok I think I found some recently closed ones that are likely related

pcr5Cj-3bw-p2

tomalec:
But so far, it's all for `woocommerce/woocommerce` nothing stable for extensions

PR #2030, opened 2023-07-31
Setup wp-env for E2E tests

mikkamp:

Our main goal is to convert E2E tests from Puppeteer to PlayWright, as well as remove any dependencies for the WooCommerce E2E packages (as they will be deprecated).

This PR is intended to setup a wp-env site for running the E2E tests against, which previously was done using a docker image as part of [`@woocommerce/e2e-environment`](https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/e2e-environment) package. A followup PR will address converting tests to PlayWright.

Resolves part of #1599 

1. Checkout this PR and run `nvm use && npm ci`
2. Start the `wp-env` site `npm run wp-env start`
3. Once completed confirm the site is running at http://localhost:8889
4. Run the E2E tests `npm run test:e2e` and confirm it uses the site `http://localhost:8889`
5. Expect only some of the tests to pass (some will still fail but that will be fixed when converting to PlayWright)

### Additional details:

Some of the tests still fail, because the related products section has switched to blocks and also because some of the tests have been flaking. Since this is the reason we are switching to PlayWright, the failing tests will be addressed there.

It's still possible to get the previous test site up and running with `npm run docker:up`, all that will be removed once the full dependency on the WooCommerce packages is no longer needed. This is still being used by GitHub actions on release branches.

The command to add a ready page was added because the `wc-e2e` script waits explicitly till this page is available. Once we convert to PlayWright it should be possible to remove this.

### Changelog entry
- Dev - Setup wp-env for E2E tests.

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,0 +1,12 @@
+{
+    "phpVersion": "8.0",
+    "plugins": [
+		"https://downloads.wordpress.org/plugin/woocommerce.latest-stable.zip",
+		"https://github.com/WP-API/Basic-Auth/archive/master.zip",
+        "."
+    ],
+	"lifecycleScripts": {
+        "afterStart": "./tests/e2e/bin/test-env-setup.sh",
+        "afterClean": "./tests/e2e/bin/test-env-setup.sh"
+    }
+}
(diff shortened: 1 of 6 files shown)


martynmjones (approved):
Hey @mikkamp,

The environment was properly set up when testing locally and the `test:e2e` script is using the correct URL so LGTM ✅

PR #2041, opened 2023-08-03
Convert E2E tests from Puppeteer to Playwright

### LINKED ISSUE #1599: Implement new E2E environment once its ready in Woo platform

### LINKED ISSUE #2002: Upgrade to Node 20
## Details
This issue is for upgrading the Node version since 16 (current version) has reached the end of it’s [security support period](https://href.li/?https://endoflife.date/nodejs). Since we rely on WP and WC components we should probably follow along with the Gutenberg issue:

https://github.com/WordPress/gutenberg/issues/48588

Their current plan is to jump to Node 20.

This issue will also be helpful to collect any blockers which might be preventing us from upgrading the Node version.

Node 16 (current version) has reached the end of it’s [security support period]

Update to Node 20

### Why blocked

With the current dependency tree, there are some packages do not / will not support Node.js v20, for example:

1. `@es-joy/jsdoccomment@0.20.1`, a child dependency of `@wordpress/scripts@24.6.0`, [doesn't support v18](https://github.com/es-joy/jsdoccomment/blob/v0.20.1/package.json#L51).
2. This extension currently uses `@woocommerce/components@10.3.0`, but `@woocommerce/components@12.0.0` (the current L-2 WC) [only supports v16](https://github.com/woocommerce/woocommerce/blob/7.8.2/packages/js/components/package.json#L8).

The first example blocks this extension to use v18 for now, and the second one may force it to downgrade to v16 in the future.

### How to check

Since packages don't always accurately reflect the scope of Node.js support for children dependencies, confirming that these blocks have been removed usually requires actually upgrading the dependencies and testing the installation with Node.js v18 to see if it completes successfully.

Until then, another way to get a quick overview of package's Node.js support is `npm view <packageName> engines`, for example:

```bash
> npm view "@woocommerce/components@>=12" engines
@woocommerce/components@12.0.0 { node: '^16.13.1', pnpm: '^7.13.3' }
@woocommerce/components@12.1.0 { node: '^16.14.1', pnpm: '^8.6.5' }
```

## Acceptance criteria

* [ ]  Node engines set to >= 20
* [ ] .nvmrc set to v20
* [ ] App compiles without errors with Node 20
* [ ] JS tests run correctly without errors with Node 20

mikkamp:

This PR converts the E2E tests from Puppeteer to Playwright and removes any dependencies on the `@woocommerce/e2e-*` packages. This is a followup from #2030

> **Note**
Until we resolve issue #2002 we are still using Node 14, which means the latest compatible version of Playwright was 1.33. Ideally this will be updated once we upgrade Node.

Overview of the changes to convert tests:
1. Conversion ID
The test Conversion ID used to be set through the Connection Test page. This has now been replaced with a custom plugin `tests/e2e/bin/test-data.php` which adds some API endpoints to set this data.

2. Site Settings
The payment gateway COD used to be enabled through an API call at the start of the test, this is now done in the `wp-env` start up script.

3. API requests
All API requests use the [basic auth plugin](https://github.com/WP-API/Basic-Auth) to send requests to the API. We use this for both WordPress and WooCommerce endpoints, see `tests/e2e/utils/api.js`

4. Avoiding `waitForSelector`
In our previous tests we often used `waitForSelector`, while this is still compatible with Playwright, I've replaced it with other [Locators](https://playwright.dev/docs/locators). This was done to simplify the tests and reduce the need for utility functions.

5. Mocking API requests
Instead of using a utility for mocking and spying on API requests, it's been replaced with the `route.fulfill`. We use this to mock connections to Jetpack or Google.

Closes #1599

1. Checkout this PR and install packages: `nvm use && npm ci`
2. Start the `wp-env` environment: `npm run wp-env:up`
3. Install the Chromium browser for Playwright: `npx playwright install chromium`
4. Start the E2E tests: `npm run test:e2e`
5. Confirm that all the tests pass, and a test report is available in tests/e2e/test-result
6. Check that we can run the tests in debug mode and step through them in the browser: `npm run test:e2e-dev`

### Additional details:

Currently the E2E workflow is set to run only on release branches or manually. I've triggered a manual run, and all the tests passed: https://github.com/woocommerce/google-listings-and-ads/actions/runs/5753482975/job/15596778163

### Changelog entry
* Dev - Convert E2E tests from Puppeteer to Playwright.


tests/e2e/bin/test-data.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
puntope, line 1: I'm aware this is a test file but would be nice to document a bit the functions.

tests/e2e/config/playwright.config.js
```diff
@@ -0,0 +1,69 @@
+const { defineConfig, devices } = require( '@playwright/test' );
+const { url } = require( './default.json' );
+
+module.exports = defineConfig( {
+	testDir: '../specs',
+
+	/* Maximum time one test can run for. */
```
puntope: 💅 

`/* Maximum time in milliseconds one test can run for. */`

tests/e2e/config/playwright.config.js
```diff
@@ -0,0 +1,69 @@
...
+	timeout: 120 * 1000,
+
+	expect: {
+		/**
+		 * Maximum time expect() should wait for the condition to be met.
```
puntope: `Maximum time, in milliseconds, expect() should wait for the condition to be met.`

tests/e2e/config/playwright.config.js
```diff
@@ -0,0 +1,69 @@
...
+		 * For example in `await expect(locator).toHaveText();`
+		 */
+		timeout: 20 * 1000,
+	},
+
+	/* Fail the build on CI if you accidentally left test.only in the source code. */
+	forbidOnly: !! process.env.CI,
```
puntope: 💅  I don't understand well this comment.

I will suggest the one appearing below on the web:
https://playwright.dev/docs/test-configuration#:~:text=Whether%20to%20exit%20with%20an%20error%20if%20any%20tests%20are%20marked%20as%20test.only.%20Useful%20on%20CI.

Whether to exit with an error if any tests are marked as "test.only"

tests/e2e/config/playwright.config.js
```diff
@@ -0,0 +1,69 @@
...
+			{
+				outputFolder: '../test-results/playwright-report',
+				open: 'never',
+			},
+		],
+	],
+
+	/* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+	use: {
+		/* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
```
puntope: /* Maximum time in milliseconds each action such as `click()` can take. Defaults to 0 (no limit). */

tests/e2e/global-setup.js
```diff
@@ -0,0 +1,81 @@
...
+	// Clear out the previous save states
+	try {
+		fs.unlinkSync( process.env.ADMINSTATE );
+		console.log( 'Admin state file deleted successfully.' );
+	} catch ( err ) {
+		if ( err.code === 'ENOENT' ) {
+			console.log( 'Admin state file does not exist.' );
+		} else {
+			console.log( 'Admin state file could not be deleted: ' + err );
+		}
```
puntope: Maybe `console.error` might be more accurate here.

tests/e2e/global-setup.js
```diff
@@ -0,0 +1,81 @@
...
+				.context()
+				.storageState( { path: process.env.ADMINSTATE } );
+			console.log( 'Logged-in as admin successfully.' );
+			adminLoggedIn = true;
+			break;
+		} catch ( e ) {
+			console.log(
+				`Admin log-in failed, Retrying... ${ i }/${ adminRetries }`
+			);
+			console.log( e );
```
puntope: Maybe we can use `console.warn.` Also, I don't understand well why we need to retry 5 times?

puntope:
One test failed for me in local

  ✘  7 [chromium] › gtag-events/gtag-events.test.js:74:2 › GTag events › Add to cart event is sent from a block shop page (2.0m)

Screenshot 2023-08-07 at 20 05 27

puntope:
Same with the rest of Add to cart events... 

I will try to check tomorrow why.

tests/e2e/config/playwright.config.js
mikkamp: Thanks, this is mostly copying and pasting from other examples. I think we can just remove this setting all together.

tests/e2e/global-setup.js
mikkamp: These are the parts that we need to do the same as WC core, but it isn't reusable to the same extent it was when we had the e2e packages. So it's basically copied code from [here](https://github.com/woocommerce/woocommerce/blob/7.9.0/plugins/woocommerce/tests/e2e-pw/global-setup.js#L53) with the comment further down:

> This step was failing occasionally, and globalsetup doesn't retry, so make it retry

I've had it fail in one of the test runs I did in GitHub Actions, so it must be a bit flakey at times (which is resolved with the retry).

Maybe I should add some references where the code is copied from to help explain. I can change the to console.warn / error where appropriate though.

mikkamp:
> One test failed for me in local

As mentioned I'd suggest to run the tests from PR #2043 as I did improve a few tests that were still a bit flakey. If they still fail there then I'd suggest to run them with `npm run test:e2e-dev` and then step through them to see where they fail. It should show the selectors it's matched while doing so.

mikkamp:
@puntope I went ahead and fixed those suggestions you made. Regarding running the tests maybe you can try with the following commands to start with a clean slate (I just ran that locally and all the tests pass):

```
npm run wp-env destroy
npm run wp-env:up
npm run test:e2e
```

puntope:
Thanks for the changes @mikkamp Unfortunately it is still stuck in the `waitForRequest`  call. (Even if I merged https://github.com/woocommerce/google-listings-and-ads/pull/2043


https://github.com/woocommerce/google-listings-and-ads/assets/5908855/3a9814c1-3245-40a5-b978-7dcb60c47a09

puntope:
Some conflicts to be resolved in `tests/e2e/docker/initialize.sh`

But ready to go. The issue was on my side likely.

puntope (approved):

PR #2043, opened 2023-08-04
Allow E2E testing with Release Candidates

mikkamp:

As a followup to PR #2041 this PR makes it easier to install Release Candidate versions of both WordPress and WooCommerce. The documentation has been updated to describe how the tests can be run with different versions of WordPress and WooCommerce.

Additionally several selectors in the test were updated as WooCommerce 8.0 introduces blocks for the standard checkout.

I also added workflow inputs to allow the E2E workflow to be run with release candidate versions of WP and/or WC. I [ran the test successfully here](https://github.com/woocommerce/google-listings-and-ads/actions/runs/5763656994/job/15625948620) with WP `6.3-RC3` and WC `8.0.0-rc.1`.

1. Checkout this PR and install packages: `nvm use && npm ci`
2. Start the `wp-env` environment: `npm run wp-env:up` (if you previously had this running you might need to run `npm run wp-env destroy`)
3. Install the Chromium browser for Playwright: `npx playwright install chromium`
4. Update to WP RC:
```
npm run -- wp-env run tests-cli -- wp core update --version=6.3-RC3
npm run -- wp-env run tests-cli -- wp core update-db
```
5. Update to WC RC:
```
npm run -- wp-env run tests-cli -- wp plugin update woocommerce --version=8.0.0-rc.1
npm run -- wp-env run tests-cli -- wp wc update
```
6. Start the E2E tests: `npm run test:e2e`
7. Confirm that all the tests pass

### Changelog entry
* Dev - Allow E2E testing with Release Candidates.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -8,6 +8,11 @@ on:
     branches:
       - "release/**"
   workflow_dispatch:
+    inputs:
+      wp-rc-version:
+        description: 'WordPress version for Release Candidate (ex. 6.3-RC3)'
+      wc-rc-version:
+        description: 'WooCommerce version for Release Candidate (ex. 8.0.0-rc.1)'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -42,6 +47,18 @@ jobs:
 
       - name: Start wp-env container
         run: npm run wp-env:up
+
+      - name: Install WP release candidate (optional)
+        if: github.event.inputs.wp-rc-version != ''
+        run: |
+          npm run -- wp-env run tests-cli -- wp core update --version=${{ github.event.inputs.wp-rc-version }}
+          npm run -- wp-env run tests-cli -- wp core update-db
+
+      - name: Install WC release candidate (optional)
+        if: github.event.inputs.wc-rc-version != ''
+        run: |
+          npm run -- wp-env run tests-cli -- wp plugin update woocommerce --version=${{ github.event.inputs.wc-rc-version }}
+          npm run -- wp-env run tests-cli -- wp wc update
 
       - name: Download and install Chromium browser.
         run: npx playwright install chromium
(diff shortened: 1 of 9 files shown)


martynmjones (approved):
Hey @mikkamp, thanks for adding an easy way to test with RCs.

Tested the branch locally and getting environments with specific versions set up and running tests against them is easy and works well.

Although not related directly to the changes here (_I think_) I am getting a test failure.

Screenshot 2023-08-09 at 09 17 08

It looks like the [endpoint being called](https://github.com/woocommerce/google-listings-and-ads/blob/4bedd7f3e2a38481d412888b5862071084e80853/tests/e2e/utils/api.js#L55) isn't available as the `test-data.php` plugin gets deactivated due to an invalid header.

Screenshot 2023-08-09 at 09 15 42

All others tests are passing though and no issues settings release candidates so LGTM ✅

mikkamp:
> It looks like the [endpoint being called](https://github.com/woocommerce/google-listings-and-ads/blob/4bedd7f3e2a38481d412888b5862071084e80853/tests/e2e/utils/api.js#L55) isn't available as the test-data.php plugin gets deactivated due to an invalid header.

Interesting error, I'm not able to replicate that and it's the first time I come across it. From a quick search on Google I see mentions of this happening when the plugin is activated too quick after installing, activating it a second time usually works. Do you still get that error after trying again?

One thing we could try is add a [mapping in wp-env](https://github.com/woocommerce/woocommerce/blob/8.0.0/plugins/woocommerce/.wp-env.json#L14) and then [activate it in the install script](https://github.com/woocommerce/woocommerce/blob/8.0.0/plugins/woocommerce/tests/e2e-pw/bin/test-env-setup.sh#L13-L14) (which is what WC does with a custom plugin). Should I add a commit for that change, or is that something you can test locally (since I'm not getting the same error here)?
I'd also recommend to run `npm run wp:env destroy` to start with a completely clean wp-env site.

PR #2046, opened 2023-08-07
Fetch WC L-1 versions

puntope:

This PR fetch WooCommerce L-1 versions for our tests using https://github.com/woocommerce/grow/tree/trunk/packages/github-actions/actions/get-plugin-releases

Verify the WC version is properly fetched and the test passing
https://github.com/woocommerce/google-listings-and-ads/actions/runs/5787000158/job/15682944873?pr=2046

### Changelog entry

> Dev - Fetch WooCommerce L-1 versions for our tests

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -60,17 +60,17 @@ jobs:
       matrix:
         php: [8.0]
         wp-version: ${{ fromJson(needs.GetWPMatrix.outputs.wp-versions) }}
-        wc-version: ${{ fromJson(needs.GetWCMatrix.outputs.latest-wc-version) }}
+        wc-version: [latest]
         include:
           - php: 7.4
             wp-version: ${{ needs.GetWPMatrix.outputs.latest-wp-version }}
-            wc-version: ${{ fromJson(needs.GetWCMatrix.outputs.previous-wc-version) }}
+            wc-version: ${{ needs.GetWPMatrix.outputs.previous-wc-version }}
           - php: 8.1
             wp-version: ${{ needs.GetWPMatrix.outputs.latest-wp-version }}
-            wc-version: ${{ fromJson(needs.GetWCMatrix.outputs.previous-wc-version) }}
+            wc-version: [latest]
           - php: 8.2
             wp-version: ${{ needs.GetWPMatrix.outputs.latest-wp-version }}
-            wc-version: ${{ fromJson(needs.GetWCMatrix.outputs.latest-wc-version) }}
+            wc-version: [latest]
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -42,7 +42,7 @@ jobs:
           slug: woocommerce
           
   UnitTests:
-    name: PHP unit tests - PHP ${{ matrix.php }}, WP ${{ matrix.wp-version || 'latest' }}, WC ${{ matrix.wc-version || 'latest' }}
+    name: PHP unit tests - PHP ${{ matrix.php }}, WP ${{ matrix.wp-version || 'latest' }}, WC ${{ matrix.wc-versions || 'latest' }}
     needs: GetMatrix
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


tomalec:
Why do we switch to WC L-1?

IMHO, the opposite setup could be desired WC L-2, but WP L-1 - as that is what WC supports.

puntope:
> Why do we switch to WC L-1?
> 
> IMHO, the opposite setup could be desired WC L-2, but WP L-1 - as that is what WC supports.

We didn't do any switch. Actually, we increased from L-0 to L-1, if you check the old version it was fetching the latest WC version. 

In regards to the policies, I checked https://woogrowp2.wordpress.com/2023/07/31/l-1-policy/ but I didn't see any accepted answer yet with the final consensus, so I lean to keep it as it was until we decide.

.github/workflows/php-unit-tests.yml
```diff
@@ -35,24 +35,42 @@ jobs:
         with:
           slug: wordpress
           
+  GetWCMatrix:
```
mikkamp: Same as comment https://github.com/woocommerce/automatewoo/pull/1527/files#r1286879075
Does it make sense to merge this as one job?

.github/workflows/php-unit-tests.yml
```diff
@@ -35,24 +35,42 @@ jobs:
...
       matrix:
         php: [8.0]
-        wp-version: ${{ fromJson(needs.GetMatrix.outputs.wp-versions) }}
+        wp-version: ${{ fromJson(needs.GetWPMatrix.outputs.wp-versions) }}
+        wc-version: [ latest ]
         include:
           - php: 7.4
-            wp-version: ${{ needs.GetMatrix.outputs.latest-wp-version }}
+            wp-version: ${{ needs.GetWPMatrix.outputs.latest-wp-version }}
+            wc-version: ${{ needs.GetWCMatrix.outputs.previous-wc-version }}
```
mikkamp: In the interest of not testing all combinations (to save billing minutes), on other extensions we make this test do the oldest supported version of everything. So in that case wouldn't this need to be PHP 7.4, WP L-2, WC L-1.
Since this is the only test that uses an older version of WC I wouldn't be against changing this to WC L-2.

.github/workflows/php-unit-tests.yml
```diff
@@ -67,9 +85,9 @@ jobs:
         uses: woocommerce/grow/prepare-mysql@actions-v1
 
       - name: Install WP tests
-        run: ./bin/install-wp-tests.sh wordpress_test root root localhost ${{ matrix.wp-version }}
+        run: ./bin/install-wp-tests.sh wordpress_test root root localhost ${{ matrix.wp-version }} ${{ matrix.wc-version }}
 
-      - if: matrix.wp-version == needs.GetMatrix.outputs.latest-wp-version && matrix.php == 8.0
+      - if: matrix.wp-version == needs.GetWPMatrix.outputs.latest-wp-version && matrix.php == 8.0
```
mikkamp: We might as well add an additional check to make sure WC is at latest version to ensure the coverage report is only generated on one run. Right now it will still run only once but if we tweak the matrix later on then it would be easy to miss:

```suggestion
      - if: matrix.wp-version == needs.GetWPMatrix.outputs.latest-wp-version && matrix.wc-version == needs.GetWCMatrix.outputs.latest-wc-version && matrix.php == 8.0
```

mikkamp (approved):
Thanks for the changes, this is working as expected. I left a few comments but none of them are blocking.

I do agree with Tomek that we might as well bump this to WC L-2 even if we aren't testing every single combination.

.github/workflows/php-unit-tests.yml
puntope: Done ✅

.github/workflows/php-unit-tests.yml
puntope: This is done as well, as we are now changing the matrix I check PHP 8.0 and WC latest

.github/workflows/php-unit-tests.yml
puntope: I changed the matrix, now we follow the same as Automatewoo

puntope:
Hi @tomalec and @mikkamp  I did some changes and change to WC - WP L-2 

This is ready for another round

.github/workflows/php-unit-tests.yml
mikkamp: The suggestion to check all three was in case we would test more combinations of WP as well. But not a big deal since as you mentioned it is cover how it is now.

mikkamp (approved):
Thanks for the changes, they are working well.

The only minor comment I have left is that none of the combinations test WP L-1. Maybe for variety we could pass that to the PHP 8.1 test.

tomalec (approved):

PR #2052, opened 2023-08-08
Add filter to be able to build custom shipping method rate handers.

matt-h:

Allows other plugins to be able to extend the capabilities to handle shipping method rates.

This adds a `woocommerce_gla_handle_shipping_method_to_rates` filter to the `shipping_method_to_rates` function for the unhandled shipping type method so that other plugins can implement handlers for those shipping methods.

### DIFF
--- a/src/Shipping/ZoneMethodsParser.php
+++ b/src/Shipping/ZoneMethodsParser.php
@@ -96,8 +96,17 @@ class ZoneMethodsParser implements Service {
 				$shipping_rates[] = $shipping_rate;
 				break;
 			default:
-				// We don't support other shipping methods.
-				return [];
+				/**
+				 * Filter the shipping rates for a shipping method that is not supported.
+				 *
+				 * @param ShippingRate[] $shipping_rates The shipping rates.
+				 * @param object|WC_Shipping_Method $method The shipping method.
+				 */
+				return apply_filters(
+					'woocommerce_gla_handle_shipping_method_to_rates',
+					$shipping_rates,
+					$method
+				);
 		}
 
 		return $shipping_rates;


src/Shipping/ZoneMethodsParser.php
```diff
@@ -96,8 +96,19 @@ protected function shipping_method_to_rates( object $method ): array {
...
+				 *
+				 * @param ShippingRate[] $shipping_rates The shipping rates.
+				 * @param object|WC_Shipping_Method $method The shipping method.
+				 * @param ZoneMethodsParser $this The ZoneMethodsParser instance.
+				 */
+				return apply_filters(
+					'woocommerce_gla_handle_shipping_method_to_rates',
+					$shipping_rates,
+					$method,
+					$this
```
puntope: ❓  Why the need of sending `ZoneMethodsParser` instance as an argument?

puntope:
Thanks for the contribution. Overall looking good for me, just have a question regarding the need of sending the class instance as a parameter.

matt-h:
@puntope I didn't need it specifically, was just trying err on the side of passing enough data to solve all potential use cases to filters, even if they aren't needed right away.

Feel free to remove that parameter.

puntope:
> @puntope I didn't need it specifically, was just trying err on the side of passing enough data to solve all potential use cases to filters, even if they aren't needed right away.
> 

Yes, I think is better to keep it as simpler as possible.

> Feel free to remove that parameter.

You can change it yourself in your side and I can then approve the PR. 👌

matt-h:
@puntope I have removed it, should be all set.

PR #2060, opened 2023-08-17
Add action for Hook Documentation Generation

puntope:

This PR creates an Action for generating PHP Hooks documentation when PHP files change.

It will commit a README.md in src/Hooks/README.md

### Screenshots

Screenshot 2023-08-17 at 15 04 53

Screenshot 2023-08-17 at 15 03 52

1. Verify how the [README.md ](https://github.com/woocommerce/google-listings-and-ads/blob/0c8537f7e8d3f1b6d7ac3d651fe61d4e38b16e63/src/Hooks/README.md) has been generated in src/HOOKS/README.md 
2. Check [in the file diff ](https://github.com/woocommerce/google-listings-and-ads/pull/2060/files#diff-712190684cdb42bf9516f3e2ebe570a0fe642f133b7797dd03c31899195b3241) the links working and it's most likely as it was before. 
3. See the workflow is triggered on release branches. https://github.com/woocommerce/google-listings-and-ads/pull/2064

### Changelog entry

> Dev - Add Action for Hooks Documentation Generator

### DIFF
--- a/.github/workflows/php-hook-documentation.yml
+++ b/.github/workflows/php-hook-documentation.yml
@@ -1,0 +1,47 @@
+name: PHP Hook Documentation Generator
+
+on:
+  pull_request:
+    paths:
+      - "**.php"
+      - .github/workflows/php-hook-documentation.yml
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  HookDocumentation:
+    name: Hook Documentation Generator
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+        with:
+          # Checks out a branch instead of a commit in detached HEAD state
+          ref: ${{ github.head_ref }}
+
+        # This generates the documentation string. The `id` property is used to reference the output in the next step.
+      - name: Generate hook documentation
+        id: generate-hook-docs
+        uses: woocommerce/grow/hook-documentation@actions-v1
+        with:
+          debug-output: yes
+          source-directories: src/,templates/
+
+      - name: Commit hook documentation
+        shell: bash
+        # Use the github-actions bot account to commit.
+        # https://api.github.com/users/github-actions%5Bbot%5D
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+          echo "${{ steps.generate-hook-docs.outputs.hook-docs }}" > src/Hooks/README.md
+          git add src/Hooks/README.md
+          if git diff --cached --quiet; then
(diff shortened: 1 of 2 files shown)


puntope:
At first, this Workflow was triggered when a push event happened with some PHP files.

After discussion with the team, now we changed the trigger to only run on release branches. 

You can see the trigger in another release branch here: https://github.com/woocommerce/google-listings-and-ads/pull/2064

.github/workflows/php-hook-documentation.yml
```diff
@@ -0,0 +1,57 @@
+name: PHP Hook Documentation Generator
+
+on:
```
mikkamp, line 3: The push and pull_request (with the PHP filter), look fine to me. But I was wondering if it would do any harm to also allow us to run this manually, like [here](https://github.com/woocommerce/google-listings-and-ads/blob/2.5.2/.github/workflows/e2e-tests.yml#L10). It's not going to effect our automated flows, and will make it easier if we ever want to run it manually.

.github/workflows/php-hook-documentation.yml
```diff
@@ -0,0 +1,57 @@
...
+          # Checks out a branch instead of a commit in detached HEAD state
+          ref: ${{ github.head_ref }}
+
+        # This generates the documentation string. The `id` property is used to reference the output in the next step.
+      - name: Generate hook documentation
+        id: generate-hook-docs
+        uses: woocommerce/grow/hook-documentation@actions-v1
+        with:
+          debug-output: yes
+          source-directories: src/,views/
```
mikkamp: What about the main plugin file `google-listings-and-ads.php`? Is this strictly a list of directories or does it support single files as well?

It so happens we don't have any hooks there now, but considering we want to use it for other extensions the script should be flexible enough to include the main plugin file. If that's not possible then can we log an issue in the Grow repo that this should be fixed?

mikkamp (approved):
Thanks for the changes they seem to be working well as far as I can see. The workflow is triggered and commits a new version of `src/Hooks/README.md` when a release branch is update.

As far as I understand we also want to maintain only one script within our grow repo. So the script `bin/HooksDocsGenerator.php` can be removed.

I just left a few minor comments but this should be pretty much good to go, so I'll mark the PR as approved.

.github/workflows/php-hook-documentation.yml
puntope: Seems like its possible as internally it uses [Finder()->path()](https://symfony.com/doc/current/components/finder.html#path)

You can see it here in a test when I added `gla_test_action` inside the main plugin file.

Screenshot 2023-08-18 at 17 03 01

https://github.com/woocommerce/google-listings-and-ads/actions/runs/5902917550/job/16011805311

.github/workflows/php-hook-documentation.yml
puntope: I implemented this suggestions

PR #2073, opened 2023-08-25
Update trigger method in Hooks Generator Workflow

puntope:

This is a follow-up PR after a comment here https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/293#discussion_r1301459977 i

In brief, the workflow doesn't trigger correctly when woorlease creates the release branch.

This PR updates the trigger following the suggested and already approved way to trigger the workflow. 

Now were trigger the workflow when changelog.txt is pushed, as this is the last commit woorelease does. 

Checking the changes. We need to test it directly in the release after merge.

### DIFF
--- a/.github/workflows/php-hook-documentation.yml
+++ b/.github/workflows/php-hook-documentation.yml
@@ -5,16 +5,7 @@ on:
     branches:
       - "release/**"
     paths:
-      - "**.php"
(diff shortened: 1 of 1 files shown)


eason9487 (approved):
LGTM.

PR #2074, opened 2023-08-28
Fix Taxonomy Attribute Mapping for Product Variations

### LINKED ISSUE #2071: Product tags not working with Google attributes on variable products
### Describe the bug:

The attribute mapper supports mapping tags to Google product attributes. For instance, mapping tags as brand attributes. However, mapping tags to product attributes only works with simple products, the tags' taxonomy data isn't passed for variable products.

### Steps to reproduce:

1. Create a variable product with multiple variants and use tags to specify the brand (e.g. Woo)
2. Create a brand attribute rule and assign product tags

3. Allow the product to sync to Google Merchant Center and notice the brand isn't synced. (This causes the product to be flagged as missing brand attribute)
4. Repeat the above step for a simple product and notice the brand attribute is correctly synced.

### Expected behavior:

Mapping tags to Google product attributes should work for both simple and variable products.

### Actual behavior:

It only works for simple products. 

### Affected users

6632964-zd-a8c
6666355-zd-a8c

puntope:

Closes #2071 

- Adjusts `WCProductAdapter::get_product_taxonomy` method to return Parent terms as a fallback for taxonomies like product_tag, product_cat etc... 
- Adding some tests covering this edge case

1. Follow the use case in #2071 and see it solved.

### Changelog entry

> Fix - Fix Taxonomy Attribute Mapping for Product Variations.

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -1139,8 +1139,9 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 
 	/**
 	 * Get a taxonomy term names from a product using
-	 * @param $product_id int The product ID to get the taxonomy term
-	 * @param $taxonomy string The taxonomy to get.
+	 *
+	 * @param int    $product_id - The product ID to get the taxonomy term
+	 * @param string $taxonomy - The taxonomy to get.
 	 * @return string[] An array of term names.
 	 */
 	protected function get_taxonomy_term_names( $product_id, $taxonomy ) {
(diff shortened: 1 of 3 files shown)


martynmjones (approved):
Hey @puntope, thanks for the quick fix 🙌 

Tested locally and confirmed that the parent products terms will be used for variations as a fallback and the data successfully syncs with Merchant Center. So LGTM ✅

sukafia:
@eason9487 I see the fix is ready. Is there a timeframe for rolling out the update?

eason9487:
Hi @sukafia, I added this extension to the release schedule and expect to release it next Tuesday.

Issue #2070, opened 2023-08-24
Add additional E2E tests


ianlin:
(co-author: @jorgemd24)

Project: pcTzPl-1Kl-p2

## Goals

To reduce the amount of [manual smoke tests](<https://github.com/woocommerce/google-listings-and-ads/wiki/Smoke-tests>) we need to run for compatibility testing.

So far the E2E test covers the first step of onboarding and connecting / disconnecting both a WordPress and a Google account. We can start by adding the test to connect to Merchant Account and then move on to the next steps of onboarding.

For testing the main page we can follow the list from [manual smoke tests](<https://github.com/woocommerce/google-listings-and-ads/wiki/Smoke-tests>).

## Tests to add

We will assign priorities to the tasks below:

🔴 means high priority<br>🟢 means low priority

### 📌 [Onboarding Step 1 - Set up accounts](<#onboarding-step-1>)

```[tasklist]
- [x] 🔴 Connecting an existing Merchant Account (including all the steps such as claiming / overriding a URL etc).
- [x] 🔴 Create a new Merchant Account.
- [x] 🟢 Verify FAQ can expand.
- [x] 🟢 Verify the links on the page can be clicked and can open the correct page.
```

### 📌 [Onboarding Step 2 - Configure product listings](<#onboarding-step-2>)

```[tasklist]
- [x] 🔴 The default language is preselected
- [x] 🔴 Select countries US and UK and both countries will appear in the country select box
- [x] 🔴 Checking the "All countries" option hides the country select textbox
- [x] 🔴 If US is not selected, the card `Tax rate (required for U.S. only)` at the bottom is not displayed.
- [x] 🔴 Test when "My shipping settings are simple”
- [x] 🔴 Test when "My shipping settings are complex”
- [x] 🔴 Test "Recommended” shipping rates
- [x] 🟢 The "Continue" button is disabled if no shipping time is entered on a simple shipping setup
- [x] 🟢 The headings `Configure your product listings` etc are shown
- [x] 🟢 Verify the links on the page can be clicked and can open the correct page.`
```

### 📌 [Onboarding Step 3 - Confirm store requirements](<#onboarding-step-3>)

```[tasklist]
- [x] 🟢 Phone verification
- [x] 🟢 Pre-Launch Checklist
- [x] 🟢 The "Continue" button is disabled until the Pre-Launch Checklist is checked and the Store address is valid
- [x] 🟢 Confirm links, copies, etc are shown in the UI
```

[...]

PR #2069, opened 2023-08-23
Override vulnerability packages: `xmlhttprequest-ssl` and `ws`

eason9487:
### Changes proposed in this Pull Request:

To resolve two critical vulnerability issues, this PR:
- Uses Node.js v16 as the target version.
   - In order to use npm's [overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides), it required npm >= 8.
   - It also needs to bump prettier packages, and all the related linting errors were fixed by `npm run lint:js -- --fix`
- Overrides `xmlhttprequest-ssl` < 1.6.2 dependency to 1.6.2.
   - Should close two critical vulnerability issues.
   ```
   └─┬ @woocommerce/components@10.3.0
     └─┬ @automattic/tour-kit@1.1.1
       └─┬ @automattic/components@2.0.1
         └─┬ @automattic/data-stores@3.0.1
           └─┬ @automattic/happychat-connection@1.0.0
             └─┬ socket.io-client@2.3.0
               └─┬ engine.io-client@3.4.4
                 └── xmlhttprequest-ssl@1.5.5
   ```
- By the way, it overrides `ws` < 6.2.2 dependency to 6.2.2.
   - Should close a moderate vulnerability issue.
   ```
   └─┬ @woocommerce/components@10.3.0
     └─┬ @automattic/tour-kit@1.1.1
       └─┬ @automattic/components@2.0.1
         └─┬ @automattic/data-stores@3.0.1
           └─┬ @automattic/happychat-connection@1.0.0
             └─┬ socket.io-client@2.3.0
               └─┬ engine.io-client@3.4.4
                 └── ws@6.1.4
   ```

### Detailed test instructions:

1. `npm ci`
2. `npm run test:js` to see if the test can pass
3. `npm run lint:js` to see if the linting can pass
4. `npm ls xmlhttprequest-ssl` to check if all this package's versions are >= 1.6.2, which has been patched.
5. `npm ls ws` to check if all this package's versions are >= 6.2.2, which has been patched.

### Additional details:

[...]

PR #2080, opened 2023-08-29
E2E tests: Dashboard - Edit Free Listings

jorgemd24:

Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/2070

This PR aims to create the base for testing the dashboard and editing the free listings. To access the dashboard page, merchants must log in with their Google account, merchant centre, and so on; otherwise, they'll be redirected to the Getting Started page.[ Since the redirection occurs on the server ](https://github.com/woocommerce/google-listings-and-ads/blob/d722ba48999c6656b24743367fca98027fe24c94/src/Admin/Redirect.php#L90-L111 )and can't be easily mocked through requests, so I opted to simulate the onboarded state by changing the values returned from the database. This was addressed by introducing a filter, as seen in this commit: https://github.com/woocommerce/google-listings-and-ads/commit/68d76b3a8b41f08e9f5c396143893b098419bc73. This enables us to attach an action that emulates the onboarding state. Initially, I attempted to store options in the database similarly to how it's done for the conversion id: https://github.com/woocommerce/google-listings-and-ads/blob/551b3851a7426922173b36c0da1e63e19706619b/tests/e2e/bin/test-data.php#L19-L36

However, this approach would cause issues, as it would lead to all tests being in the onboarded state, preventing us from testing scenarios like the getting started page.

This PR covers the testing of the following aspects:

- The dashboard page is rendered correctly.
- Tests the presence of both Free Listings and Edit buttons on the dashboard page.
- Verifying that clicking the Edit button triggers the display of a modal containing two options: "Continue to Edit" and "Don't Edit".
- Checking the checkbox for recommended shipping rates.
- Entering shipping times.
- Selecting tax rates.
- Testing the functionality of the "Save changes" button.- Asserting the requests generated during the process of saving changes.

1. Follow these steps to set up the e2e env: https://github.com/woocommerce/google-listings-and-ads#e2e-testing
2. Run `npm run test:e2e ./tests/e2e/specs/dashboard/edit-free-listings.test.js` or `npm run test:e2e` to run all tests.
3. Check that all tests ✅  

### Additional details:

While working on this, I experienced the problem described here: https://github.com/woocommerce/google-listings-and-ads/pull/2043#pullrequestreview-1569007864. Specifically, the test-data plugin failed to load correctly, leading to 404 errors in certain tests or preventing the successful simulation of the onboarded state:

image

As a temporary solution, I had to stop the containers and restart them. Additionally, on occasions, I had to manually activate the plugin by visiting http://localhost:8889/wp-admin/plugins.php?plugin_status=all. We will need to look deeper into this possible issue. 

While there are numerous other scenarios that could be tested, the primary objective of this PR is to create the base for the tests for the dashboard page.

### Changelog entry

>Dev - E2E - Dashboard - Edit Free Listings


jorgemd24:
The JS linting is complaining about this:

```
/home/runner/work/google-listings-and-ads/google-listings-and-ads/tests/e2e/specs/dashboard/edit-free-listings.test.js
  56:31  error  Replace `·` with `⏎↹↹↹`  prettier/prettier
  72:28  error  Replace `·` with `⏎↹↹↹`  prettier/prettier
```

However, when I run `npm run lint:js` locally it is all good.  Not sure if it is related to a Carriage Return and some tabs? 

It seems related to this: https://stackoverflow.com/questions/67702186/prettier-ask-me-to-replace-with, I will have a look at it tomorrow.

eason9487:
> The JS linting is complaining about this:
> 
> ```
> /home/runner/work/google-listings-and-ads/google-listings-and-ads/tests/e2e/specs/dashboard/edit-free-listings.test.js
>   56:31  error  Replace `·` with `⏎↹↹↹`  prettier/prettier
>   72:28  error  Replace `·` with `⏎↹↹↹`  prettier/prettier
> ```
> 
> However, when I run `npm run lint:js` locally it is all good. Not sure if it is related to a Carriage Return and some tabs?

The prettier packages were updated in #2069. Rebasing onto the current `develop` should resolve the inconsistency.

tests/e2e/utils/pages/dashboard.js
```diff
@@ -0,0 +1,131 @@
...
+	async closePage() {
+		await this.page.close();
+	}
+
+	/**
+	 * Mock all requests related to external accounts such as Merchant Center, Google, etc.
+	 */
+	async mockRequests() {
+		// Mock Reports Programs
+		this.fulfillMCReportProgram( {
```
ianlin: It'd probably be better to add `await` in this call and also other calls below. Otherwise this function `mockRequests` would return void directly without waiting these calls to be finished.

tests/e2e/utils/pages/edit-free-listings.js
```diff
@@ -0,0 +1,90 @@
...
+		} );
+	}
+
+	/**
+	 * Click the Save Changes button.
+	 * @returns {Promise<void>}
+	 */
+	async clickSaveChanges() {
+		const saveChangesButton = await this.getSaveChangesButton();
+		saveChangesButton.click();
```
ianlin: Same as adding `await`.

mikkamp:
> This was addressed by introducing a filter, as seen in this commit: https://github.com/woocommerce/google-listings-and-ads/commit/68d76b3a8b41f08e9f5c396143893b098419bc73. This enables us to attach an action that emulates the onboarding state. Initially, I attempted to store options in the database similarly to how it's done for the conversion id
> However, this approach would cause issues, as it would lead to all tests being in the onboarded state, preventing us from testing scenarios like the getting started page.

I have a question about this approach. For the conversion ID it's possible to both set and unset the conversion ID, which in the case of the GTag tests it's set through `test.beforeAll` and unset in `test.afterAll` so it will only be in that state for a certain set of tests. Can't we apply the same here and enable it only for a certain set of tests, since it uses an API request it's very quick to set/unset data in a specific state?
My reasoning why I prefer not to use the filter approach is because it will end up in the automatically generated hook document, but it's not a filter that should be used outside of E2E testing. Also as we build out the tests we are going to need many more "test" states (such as MC account connected, Ads account connected, etc...), so we'll have to add multiple filters which are intended for testing only.

ianlin (approved):
Thanks for adding the tests, especially the shared util classes so other tests can use them. I've tested it and the tests are working as expected. Just added a `async` `await` suggestion in the comment.

mikkamp:
> While working on this, I experienced the problem described here: https://github.com/woocommerce/google-listings-and-ads/pull/2043#pullrequestreview-1569007864. Specifically, the test-data plugin failed to load correctly, leading to 404 errors in certain tests or preventing the successful simulation of the onboarded state

Since you are able to reproduce this at least some of the time could you try to see if some tweaks would work? One such suggestion would be to map the plugin to a separate folder instead of having it directly in `wp-content/plugins`, so if we change:
```
		"wp-content/plugins/test-data.php": "./tests/e2e/bin/test-data.php"
```

```
		"wp-content/plugins/test-data/test-data.php": "./tests/e2e/bin/test-data.php"
```

Another suggestion would be to clear cache. As `get_plugins` caches the list of plugins and if it's missing from the list it outputs the message that the header is missing. Maybe running `wp cache delete plugins plugins` before activating the plugin.

jorgemd24:
@mikkamp, thanks for your comment:

> I have a question about this approach. For the conversion ID it's possible to both set and unset the conversion ID, which in the case of the GTag tests it's set through test.beforeAll and unset in test.afterAll so it will only be in that state for a certain set of tests. Can't we apply the same here and enable it only for a certain set of tests, since it uses an API request it's very quick to set/unset data in a specific state?

That was my first idea. However, the issue I encountered was that Playwright uses multiple workers to execute tests. This means that, for instance, the `get-started.test.js` test might run concurrently with `edit-free-listings.test.js`. Therefore, when `edit-free-listings.test.js` updates the database to simulate the onboarded merchant, it causes `get-started.test.js` to fail since the getting started page won't load instead it will load the dashboard or the product feed.

I think that if you make another test that needs the ADS_CONVERSION_ACTION option to be blank, you may encounter the problem because `gtag-events.test.js` will set it in the DB that is shared for all tests.

>My reasoning why I prefer not to use the filter approach is that it will end up in the automatically generated hook document, but it's not a filter that should be used outside of E2E testing. Also as we build out the tests we are going to need many more "test" states (such as MC account connected, Ads account connected, etc...), so we'll have to add multiple filters which are intended for testing only.

I agree that currently, this filter is only used for E2E tests. However, I don't see needing multiple filters since the extensions mainly use REST endpoints, which can be easily simulated with Playwright. For example, when dealing with MC accounts and other accounts, we're using this approach: https://github.com/woocommerce/google-listings-and-ads/pull/2080/files#diff-f64b9525d898297dabb242cc89a61158db06de7159c4ec987ca62f6bb790e143R56-R74.

> Since you are able to reproduce this at least some of the time could you try to see if some tweaks would work? One such suggestion would be to map the plugin to a separate folder instead of having it directly in wp-content/plugins, so if we change:

Thanks for all your suggestions! The issue is intermittent but I will try with those suggestions to see if it fixes it.

mikkamp:
> That was my first idea. However, the issue I encountered was that Playwright uses multiple workers to execute tests.

Thanks for the clarification, I forgot about that part and was originally contemplating turning that off like is done in [Bookings when run for CI](https://github.com/woocommerce/woocommerce-bookings/blob/2.0.2/tests/e2e/playwright.config.js#L31).
Do you think it will allow us to simplify tests if we don't run them in parallel? So we can set a state, run a series of tests and then reset the state? We use that same model for our unit tests as well.

Having a closer look now at the filter it seems you only do so when a query parameter is being set and not persistently, which seems like a workable solution and not too intrusive. I'm just wondering if that technique will eventually contribute to more code to handle certain states while supporting parallel tests, and whether just turning off parallel would simplify things.

jorgemd24:
Thanks @ianlin for your time reviewing this PR. I have updated the below points in this PR:

- Adjusted the number of workers to 1 (https://github.com/woocommerce/google-listings-and-ads/pull/2080/commits/d68f4d7a8e0f33cac9ac1f58e9fd98b1d624a42c) as it was decided in the dev call peeuvX-Te-p2#comment-1093.
 (cc: @mikkamp ).
- Removed filter (https://github.com/woocommerce/google-listings-and-ads/pull/2080/commits/0664818c35c60bd2de6f3f0e36fddb11abf0ff36 ) and add an endpoint to simulate the onboarded status (https://github.com/woocommerce/google-listings-and-ads/pull/2080/commits/d3b3448f9b05ee3777cec7bde29d8353b2966402 ). 
- Added missing `awaits` (https://github.com/woocommerce/google-listings-and-ads/pull/2080/commits/4b4f063f282e941e25e1158d53c3d3f652e5923d)
- Fixed the JS Linting error (https://github.com/woocommerce/google-listings-and-ads/pull/2080/commits/62f62705d8805e9eff5b7df22271f583a3308485) (thanks @eason9487 )

Would you mind to take a look? Thanks!

tests/e2e/specs/dashboard/edit-free-listings.test.js
```diff
@@ -0,0 +1,95 @@
...
+		await expect( dontEditButton ).toBeEnabled();
+	} );
+
+	test( 'Continue to edit Free listings', async () => {
+		await dashboardPage.clickContinueToEditButton();
+		await page.waitForLoadState( 'domcontentloaded' );
+	} );
+
+	test( 'Check recommended shipping settings', async () => {
+		editFreeListingsPage = new EditFreeListingsPage( page );
```
ianlin: While the test is configured to run serial, I think it'd still be better to move this line to `beforeAll()` like `dashboardPage`.

tests/e2e/utils/pages/edit-free-listings.js
```diff
@@ -0,0 +1,96 @@
...
+			.check();
+	}
+
+	/**
+	 * Register the requests when the save button is clicked.
+	 *
+	 * @return {Promise<import('@playwright/test').Request[]>} The requests.
+	 */
+	registerSavingRequests() {
+		const targetAudienteRequest = this.page.waitForRequest(
```
ianlin: Typo: should be "Audience".

tests/e2e/utils/pages/edit-free-listings.js
```diff
@@ -0,0 +1,96 @@
...
+			( request ) =>
+				request.url().includes( '/gla/mc/settings' ) &&
+				request.method() === 'POST' &&
+				request.postDataJSON().shipping_rate === 'automatic' &&
+				request.postDataJSON().shipping_time === 'flat'
+		);
+
+		const syncRequest = this.page.waitForRequest(
+			( request ) =>
+				request.url().includes( '/gla/mc/settings/sync' ) &&
```
ianlin, line 104: ❔We didn't mock this request `/gla/mc/settings/sync` and by running `npm run test:e2e-dev tests/e2e/specs/dashboard/edit-free-listings.test.js` I saw this request returned `500` server error with the message `Please reconnect your Jetpack account.`. Although the test still passed since we only check the response from `gla/mc/settings` API, do you think it'd be a good idea to mock `/gla/mc/settings/sync` API so it wouldn't respond the error?

ianlin (approved):
Thanks for the extra changes. I tested it again and everything passed as expected. Just left some non-blocking questions and suggestions in the comment below.

Note that during the review I also encountered this problem https://github.com/woocommerce/google-listings-and-ads/pull/2043#pullrequestreview-1569007864. I had to restart the container to make it work again. I tried to delete the plugins cache using wp cli in the container but still didn't work so I think those potential fixes @mikkamp suggested also require restarting the container so it makes it a bit tricky to verify if the fixes work. I'll keep trying it when developing other e2e tests.

tests/e2e/utils/pages/edit-free-listings.js
jorgemd24: At first, I didn't mock the response because I believed there wasn't much to test. But now I realize we can actually verify the message in the snackbar. See here: https://github.com/woocommerce/google-listings-and-ads/pull/2080/commits/5a9c0e2d1c3e4fd4a7036a79a514a5b2bed12b7d

jorgemd24:
Thanks @ianlin for your suggestions! I addressed all of them.

> Note that during the review I also encountered this problem https://github.com/woocommerce/google-listings-and-ads/pull/2043#pullrequestreview-1569007864. I had to restart the container to make it work again. I tried to delete the plugins cache using wp cli in the container but still didn't work so I think those potential fixes @mikkamp suggested also require restarting the container so it makes it a bit tricky to verify if the fixes work. I'll keep trying it when developing other e2e tests.


Since yesterday I have been working with this tweak:

> Another suggestion would be to clear cache. As get_plugins caches the list of plugins and if it's missing from the list it outputs the message that the header is missing. Maybe running wp cache delete plugins plugins before activating the plugin.

 It looks like it's solving the problem. Once I confirm this, I'll implement the tweak.

ianlin (approved):
Thanks for the extra changes! I've tested again and it worked perfectly, LGTM.

PR #2087, opened 2023-09-07
WooCommerce Subscriptions compatibility: Fix the visible issue of the product data tab and meta box for some unsupported product types

### LINKED ISSUE #2086: WooCommerce Subscriptions compatibility: Product UI groups are still shown for some unsupported product types
### Describe the bug:

When WooCommerce Subscriptions is activated, the "Google Listings and Ads" tab and "Channel visibility" meta box are still shown on the product editing page for some unsupported product types:
- Grouped product
- External/Affiliate product

### Steps to reproduce:

1. Install and activate WooCommerce Subscriptions.
2. Go to edit a product.
3. Change the product type to "Grouped product".

### Expected behavior:

The "Google Listings and Ads" tab and "Channel visibility" meta box should be hidden when selecting the Grouped product or External/Affiliate product type.

### Actual behavior:

The "Google Listings and Ads" tab and "Channel visibility" meta box are still shown when selecting the Grouped product or External/Affiliate product type.

### Additional details:

This is mainly caused by the additional processing added in WooCommerce Subscriptions, which overrides the associated processing in WooCommerce core.
- WooCommerce core: https://github.com/woocommerce/woocommerce/blob/8.0.3/plugins/woocommerce/client/legacy/js/admin/meta-boxes-product.js#L204-L243
- WooCommerce Subscriptions: https://github.com/Automattic/woocommerce-subscriptions-core/blob/6.2.0/assets/js/admin/admin.js#L51

Relevant follow-up issue: https://github.com/Automattic/woocommerce-subscriptions-core/issues/127

eason9487:

Closes #2086

- Remove the controls' show/hide CSS classes of the product editing page from PHP side.
- Control the show or hide of the product data tab and meta box added to the product editing page by JS.
- Merge the custom-inputs script into the product-attributes script.

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/f8ef39ee-48ec-4161-b5b7-1dc8a2205794

1. Install and activate WooCommerce Subscriptions and WooCommerce Product Bundles.
2. Restart the client build script (`npm start`) if it's running.
3. Go to edit a product.
4. Select each product type to see if the "Google Listings and Ads" tab and "Channel visibility" meta box are visible for supported product types and hidden for unsupported ones.
   - Supported product types:
      https://github.com/woocommerce/google-listings-and-ads/blob/f325aa1cebee24ffd95985c3dd719233e045e44a/js/src/product-attributes/index.js#L11-L18
5. Create a product with the "Simple subscription" type and go to its editing page.
6. After the page is initialized, check if the UIs mentioned in step 3 are hidden.

### Changelog entry

> Fix - WooCommerce Subscriptions compatibility: Fix the visible issue of the "Google Listings and Ads" tab and "Channel visibility" meta box for some unsupported product types.


src/Integration/WooCommerceSubscriptions.php
eason9487, line 1: This integration class doesn't actually participate in any runtime processing anymore now, but I'm not sure if it would be better to remove it entirely.

martynmjones (approved):
Hey @eason9487, thanks for your work on this issue 🙌 

Tested the branch locally with Subscriptions and Product Bundles. The GLA tab and channel visibility meta box now only appear if the product type is simple, variable, or bundle. So LGTM ✅

src/Integration/WooCommerceSubscriptions.php
martynmjones: What would be the downside of removing the file entirely?

js/src/product-attributes/index.js
```diff
@@ -1,4 +1,45 @@
...
 import './index.scss';
+
+const applicableProductTypes = new Set( [
+	// Simple product
+	'simple',
+	// Variable product
+	'variable',
+	// Product bundle from WooCommerce Product Bundles
+	'bundle',
+] );
```
mikkamp: Just have one additional comment about this list. Previously we had the PHP filter `woocommerce_gla_supported_product_types` which allows an integration such as Bundles to add themselves to the list. We added this filter not only for the internal integrations we support, but also if there is a plugin which wants to add support for syncing their "specific" product type. With this list it's not possible to extend the supported product types. Can we pass this list dynamically to the JS so it's run through the `woocommerce_gla_supported_product_types` filter?

src/Integration/WooCommerceSubscriptions.php
eason9487: Originally I guess it might be ok to keep it as a reminder that this extension doesn't support the product types of WooCommerce Subscriptions. After second thoughts, I think it should be fine to remove it because the current way only uses an inclusive list of the supported product types. Other product types will naturally not be supported.

Removed in https://github.com/woocommerce/google-listings-and-ads/pull/2087/commits/d013eaad48069cd1110292807606541b0ca42a25

js/src/product-attributes/index.js
eason9487: Thanks for the suggestion!

Since the original execution order of the related processing is:
1. [AssetsHandler->add_many](https://github.com/woocommerce/google-listings-and-ads/blob/338300fc25a2b79acccd3b35a5e6f8f51175c609/src/Admin/Admin.php#L74-L75)
2. [AssetsHandler->register](https://github.com/woocommerce/google-listings-and-ads/blob/338300fc25a2b79acccd3b35a5e6f8f51175c609/src/Infrastructure/GoogleListingsAndAdsPlugin.php#L102-L105)
3. [IntegrationInitializer->register](https://github.com/woocommerce/google-listings-and-ads/blob/338300fc25a2b79acccd3b35a5e6f8f51175c609/src/Integration/IntegrationInitializer.php#L43-L46) and [WooCommerceProductBundles->init_product_types](https://github.com/woocommerce/google-listings-and-ads/blob/338300fc25a2b79acccd3b35a5e6f8f51175c609/src/Integration/WooCommerceProductBundles.php#L137-L143)
4. [AssetsHandler->enqueue_many](https://github.com/woocommerce/google-listings-and-ads/blob/338300fc25a2b79acccd3b35a5e6f8f51175c609/src/Admin/Admin.php#L77-L85)

On the first call, the inline script `glaData` is created along with the script file registration. However, the third call has not yet been made, which results in the supported product types not containing the `'bundle'`.

Therefore, I adjusted the execution order as follows in https://github.com/woocommerce/google-listings-and-ads/pull/2087/commits/d13cdf6b239f32d5261385857740834684aad0d2, so that the script registration will be called after the `IntegrationInitializer->register`.

1. [IntegrationInitializer->register](https://github.com/woocommerce/google-listings-and-ads/blob/d13cdf6b239f32d5261385857740834684aad0d2/src/Integration/IntegrationInitializer.php#L43-L46) and [WooCommerceProductBundles->init_product_types](https://github.com/woocommerce/google-listings-and-ads/blob/d13cdf6b239f32d5261385857740834684aad0d2/src/Integration/WooCommerceProductBundles.php#L137-L143)
2. [AssetsHandler->register_many](https://github.com/woocommerce/google-listings-and-ads/blob/d13cdf6b239f32d5261385857740834684aad0d2/src/Admin/Admin.php#L86): Combines the former `AssetsHandler->add_many` and `AssetsHandler->register`
3. [AssetsHandler->enqueue_many](https://github.com/woocommerce/google-listings-and-ads/blob/d13cdf6b239f32d5261385857740834684aad0d2/src/Admin/Admin.php#L87)

And removed the former calls of `AssetsHandler->add_many` and `AssetsHandler->register`.

eason9487:
Hi @mikkamp, thanks for the suggestion.

As there are quite a number of changes made for https://github.com/woocommerce/google-listings-and-ads/pull/2087#discussion_r1319665596, could you take another look at this PR?

src/Assets/AssetsHandler.php
```diff
@@ -4,14 +4,13 @@
...
 use Automattic\WooCommerce\GoogleListingsAndAds\Exception\InvalidAsset;
-use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Registerable;
 
 /**
  * Class AssetsHandler
  *
  * @package Automattic\WooCommerce\GoogleListingsAndAds\Assets
  */
-final class AssetsHandler implements Registerable, AssetsHandlerInterface {
+final class AssetsHandler implements AssetsHandlerInterface {
```
mikkamp, line 13: I was a bit confused about the renaming of the add > register functions. If I recall correctly that name was previously avoided because it implemnted the Registerable interface. But I guess since you removed that here there is no issue with using the function name `register`.

mikkamp (commented):
Thanks for all the changes. I've taken some time to look over it and so far it looks good.

But I just wanted to note that previously we would call:

`Asset::register` right after `init` hook. Now it's called directly whenever we call `AssetHandler::register` for the individual `Asset`. In the case of the collection of admin assets that should be fine because we call `register` on the hook `admin_enqueue_scripts`.

But I'd like to test that a little further also on the frontend assets and ensure there aren't any other adverse affects from changing those timings. I'll continue testing on Monday.

src/Assets/AssetsHandler.php
eason9487: If I understood it right, only when a class implements the `Service` and `Registerable` interfaces and is tagged with its interfaces via a `AbstractServiceProvider` method like [share_with_tags](https://github.com/woocommerce/google-listings-and-ads/blob/d13cdf6b239f32d5261385857740834684aad0d2/src/Internal/DependencyManagement/AbstractServiceProvider.php#L71-L78) will be [treated as a service and registered](https://github.com/woocommerce/google-listings-and-ads/blob/d13cdf6b239f32d5261385857740834684aad0d2/src/Infrastructure/GoogleListingsAndAdsPlugin.php#L135-L140) in `GoogleListingsAndAdsPlugin`.

The previous `AssetsHandler` didn't implement `Service` and its interfaces were not tagged, so I think it should be fine to remove `Registerable` and use the register method in this way. So that it also pairs and aligns with the `enqueue` method in the perspective of `wp_register_script` and `wp_enqueue_script`.

mikkamp: Thanks for the clarification. Yes that's correct that it needs to be both a `Service` and `Registerable`. The tagging isn't important as it will still fetch the classes even if it's not tagged.

I looked at all the assets further and they all seem to be registered/enqueued at the right time even after these changes. So that all looks good.

src/Admin/Admin.php
```diff
@@ -152,13 +155,19 @@ protected function get_assets(): array {
...
+			"{$this->get_root_dir()}/js/build/product-attributes.asset.php",
+			new BuiltScriptDependencyArray(
+				[
+					'dependencies' => [],
+					'version'      => (string) filemtime( "{$this->get_root_dir()}/js/build/product-attributes.js" ),
+				]
+			),
 			$product_condition
-		) );
+		) )->add_inline_script( ...$gla_data_inline_script_args );
```
mikkamp: Based on the conditions we won't ever be on a WC Admin registered page and a (traditional) product edit page, so it's ok to pass the data to two separate scripts. I wonder though if this will always remain the case, like with the new product editing experience. Because in that case we would be setting `glaData` twice on the page.

If `applicableProductTypes` is all that's needed then why wouldn't we just pass that separately so we'd never be able to conflict:

```suggestion
		) )->add_inline_script(
			[
				'glaProductData',
				[
					'applicableProductTypes' => ProductSyncer::get_supported_product_types(),
				],
			]
		);
```

mikkamp (approved):
Thanks for all the changes. I've tested the assets and I can confirm that they are all loaded correctly with their associated data.

I just left one comment though about the data possibly conflicting. It's not a major blocker as it's not an issue right now. So I'll go ahead and approve this PR.

src/Admin/Admin.php
eason9487: Thanks for the suggestion.

It's uncertain if the new Product Edit page will have the case of setting `glaData` twice. Indeed, it will be better to avoid possible conflicts in advance. Separated the inline script withe different variable names in https://github.com/woocommerce/google-listings-and-ads/pull/2087/commits/0468a73eff2460467a569d0f3c29b385499b1560.

Issue #1250, opened 2022-02-16
Performance improvements - Limiting the number of classes loaded by the container


jorgemd24:
## User story
This is a follow-up from [#1244(comment) ](https://github.com/woocommerce/google-listings-and-ads/pull/1244#pullrequestreview-882686847) and the issue #1205.

### Is your feature request related to a problem? 
After the investigation of #1205, we detected that we could limit the number of classes loaded by the container on every request. In PR #1244 we did an improvement related to the loading of the NoteInitializer and the GoogleAdsClient classes and we saw a significant improvement in the performance.

We could continue doing an investigation to limit the number of classes loaded by the container depending on the request. For example, it seems that for some fronted requests, it would not be necessary to call the method [maybe_register_services](https://github.com/woocommerce/google-listings-and-ads/blob/1.11.1/src/Infrastructure/GoogleListingsAndAdsPlugin.php#L113) and consequently improve the performance. There are other cases where the classes are only needed for scheduled jobs, we could think to load these classes only when the scheduled jobs are executed.

#### How to reproduce the problem
Debugging the function [$services = $this->container->get( Service::class );](https://github.com/woocommerce/google-listings-and-ads/blob/de788f4effbf63d14c28ef343277de0ea167a728/src/Infrastructure/GoogleListingsAndAdsPlugin.php#L126) and checking the number of services loaded and the time that it takes to finalize that call. 

### Describe the solution you'd like 
In the function where we are registering the classes, i.e maybe_register_services, we could load the services depending on the type of the request:
 - Scheduled_Services -> $services = $this->container->get( Scheduled_Services::class );
 - Rest_Services -> $services = $this->container->get( Rest_Services::class );
 - Guest_Services -> $services = $this->container->get( Guest_Services::class );

In this way, we have more control of the classes that are loaded.

## Technical
We will need to do a deeper investigation of how we can split these services and how we could limit the number of classes loaded depending on the type of request.

mikkamp:
This issue has recently surfaced again with the following profiling:
![image (1)](https://github.com/woocommerce/google-listings-and-ads/assets/11388669/21102fa4-4615-4d33-968e-c564c4249fd4)
*Note: this is taken from a site where GLA hasn't even been setup/onboarded yet*

[...]

PR #2099, opened 2023-09-18
Performance issue:  Move GoogleAdsFailures::init to error_handler middleware

jorgemd24:

Closes part of #1250

The `GoogleAdsClient:init` is an expensive operation that is currently loaded whenever the Google Ads client is created. It gets triggered by the [GoogleServiceProvider](https://github.com/woocommerce/google-listings-and-ads/blob/9895a35a1439a57fbdb48f9694986bcb823fce98/src/Internal/DependencyManagement/GoogleServiceProvider.php#L154-L163) and is executed each time there's an admin page load or a request to `wp-json`.

See the full details of this issue: https://github.com/woocommerce/google-listings-and-ads/issues/1250#issuecomment-1723187436

In this PR, we relocate `GoogleAdsFailures::init` from the GoogleAdsClient constructor to the `error_handler` middleware. This change means that GoogleAdsFailure will only be loaded when there is a request made to the `google-ads` API.

Before this PR:

image

After this PR:

image

Shifting `GoogleAdsFailure:init` to the `error_handler` will help in cutting down the time allocated for the Google protobufs functions. However, the GL&A container is still consuming roughly 4.88% of the total loading process, therefore still some room for improvement. 

#### General Testing
1. Connect your Ads account and test that you can create, edit, and delete ads campaigns.

#### Simulate an Ads API error
1. Change this line: https://github.com/woocommerce/google-listings-and-ads/blob/9546add794d96e5ea571fdc1b84e058ba72a6449/src/API/Google/AdsCampaign.php#L268

To:  `$campaign_fields['status'] = 999999`

2. Go to the dashboard page, and change the status of one of your campaigns. 
3. You should see the following error:

image

#### Simulate a partial failure error.

1. Keep the change that you made in the previous testing (campaign_status = 99999).
2. Change this:

https://github.com/woocommerce/google-listings-and-ads/blob/9546add794d96e5ea571fdc1b84e058ba72a6449/src/API/Google/AdsCampaign.php#L557-L560

With:

```
		$responses = $this->client->getGoogleAdsServiceClient()->mutate(
			$this->options->get_ads_id(),
			$operations,
			[
				'partialFailure' => true,
			]
		);
```
3. You should see the following error:

image

It shows an invalid campaign ID, because the current code is not expecting partial failures.

4. If you comment out `GoogleAdsFailures::init();` and try to change the campaign status, you will get the following error:

image

### Changelog entry

> Fix - Performance issue with GoogleAdsFailures::init

### DIFF
--- a/src/Internal/DependencyManagement/GoogleServiceProvider.php
+++ b/src/Internal/DependencyManagement/GoogleServiceProvider.php
@@ -46,6 +46,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\League\Container\Definiti
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Psr\Container\ContainerInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Psr\Http\Message\RequestInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Psr\Http\Message\ResponseInterface;
+use Google\Ads\GoogleAds\Util\V14\GoogleAdsFailures;
 use Jetpack_Options;
 
 defined( 'ABSPATH' ) || exit;
@@ -192,6 +193,14 @@ class GoogleServiceProvider extends AbstractServiceProvider {
 				return $handler( $request, $options )->then(
 					function ( ResponseInterface $response ) use ( $request ) {
 						$code = $response->getStatusCode();
+
+						$path = $request->getUri()->getPath();
+
+						// Partial Failures come back with a status code of 200, so it's necessary to call GoogleAdsFailures:init every time..
+						if ( strpos( $path, 'google-ads' ) !== false ) {
+							GoogleAdsFailures::init();
+						}
+
 						if ( $code < 400 ) {
 							return $response;
 						}
--- a/src/Internal/DependencyManagement/GoogleServiceProvider.php
+++ b/src/Internal/DependencyManagement/GoogleServiceProvider.php
@@ -196,7 +196,7 @@ class GoogleServiceProvider extends AbstractServiceProvider {
 
 						$path = $request->getUri()->getPath();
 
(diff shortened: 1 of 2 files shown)


src/Internal/DependencyManagement/GoogleServiceProvider.php
```diff
@@ -192,6 +193,14 @@ protected function error_handler(): callable {
 				return $handler( $request, $options )->then(
 					function ( ResponseInterface $response ) use ( $request ) {
 						$code = $response->getStatusCode();
+
+						$path = $request->getUri()->getPath();
+
+						// Partial Failures come back with a status code of 200, so it's necessary to call GoogleAdsFailures:init every time.
+						if ( strpos( $path, 'google-ads' ) !== false ) {
+							GoogleAdsFailures::init();
```
mikkamp, line 201: I was wondering what happens if we ever send two requests (one after the other). However it seems that the init function only initializes a new instance if it can't find a previously initialized one.

mikkamp (approved):
Thanks for the changes. I've tested failures and partial failures and I get the expected response regardless of when it happens.

Just wanted to note that the before and after screenshots in the PR are identical, but I've seen the ones in https://github.com/woocommerce/google-listings-and-ads/issues/1250#issuecomment-1723241468.

It's a shame that we can't initialize only on an error response, but it's still a great improvement to only initialize when there has actually been a request to the Google Ads API.

src/Internal/DependencyManagement/GoogleServiceProvider.php
jorgemd24: Yes that's correct.

jorgemd24:
Thanks @mikkamp for reviewing this PR.

> Just wanted to note that the before and after screenshots in the PR are identical, but I've seen the ones in https://github.com/woocommerce/google-listings-and-ads/issues/1250#issuecomment-1723241468.

I updated the description with the right images.

Thanks again!

PR #2106, opened 2023-09-20
Add section to privacy policy guide

martynmjones:

Adds a section to the privacy policy guide directing merchants to the Google Ads privacy documentation.

### Screenshots:

Screenshot 2023-09-20 at 18 32 37

1. Checkout `add/privacy-policy-content`
2. Go to `wp-admin > Settings > Privacy > Policy Guide`
3. Check `Google Listings & Ads` section and confirm links work

### Changelog entry

> Add - Privacy policy guide section

### DIFF
--- a/src/Admin/Admin.php
+++ b/src/Admin/Admin.php
@@ -269,12 +269,17 @@ class Admin implements Service, Registerable, Conditional, OptionsAwareInterface
 	 * @return void
 	 */
 	public function privacy_policy() {
-		$content = '<p class="privacy-policy-tutorial">' . sprintf(
+		$policy_text = sprintf(
 			/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
-			esc_html__( 'By using this extension, you may be storing personal data or sharing data with an external service. %1$sLearn more about what data is collected by Google and what you may want to include in your privacy policy.%2$s.', 'google-listings-and-ads' ),
+			esc_html__( 'By using this extension, you may be storing personal data or sharing data with an external service. %1$sLearn more about what data is collected by Google and what you may want to include in your privacy policy%2$s.', 'google-listings-and-ads' ),
 			'<a href="https://support.google.com/adspolicy/answer/54817" target="_blank">',
 			'</a>'
-		) . '</p>';
+		);
+
+		// As the extension doesn't offer suggested privacy policy text, the button to copy it is hidden.
+		$content = '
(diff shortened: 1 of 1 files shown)


src/Admin/Admin.php
```diff
@@ -261,6 +263,22 @@ protected function enableReports(): bool {
 		return apply_filters( 'woocommerce_gla_enable_reports', true );
 	}
 
+	/**
+	 * Add suggested privacy policy content
+	 *
+	 * @return void
+	 */
+	public function privacy_policy() {
+		$content = '<p class="privacy-policy-tutorial">' . sprintf(
```
ianlin: I believe that adding the class `privacy-policy-tutorial` might cause the `Copy suggested policy text to clipboard` button to not function correctly. According to the [wp_add_privacy_policy_content](https://developer.wordpress.org/reference/functions/wp_add_privacy_policy_content/#description) doc:

> Any content contained within HTML elements that have the .privacy-policy-tutorial CSS class applied will be omitted from the clipboard when the section content is copied.

ianlin (commented):
Thanks for adding this, the privacy policy was displayed correctly in my local. However I found the `Copy suggested policy text to clipboard` button didn't work as expected, more details in the comment below.

src/Admin/Admin.php
```diff
@@ -261,6 +263,22 @@ protected function enableReports(): bool {
...
+			/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
+			esc_html__( 'By using this extension, you may be storing personal data or sharing data with an external service. %1$sLearn more about what data is collected by Google and what you may want to include in your privacy policy.%2$s.', 'google-listings-and-ads' ),
```
ianlin: There's an extra full stop at the end of the paragraph.

Screenshot 2023-09-21 at 15 28 53

martynmjones:
Hey @ianlin, thanks for the review!

As we're only providing direction for merchants to visit Google to get updated privacy information the button doesn't make sense as there's no suggested policy text to copy. So in e2c97c7 I've hidden the button with CSS so that we don't have a redundant button sitting beneath the text.

If you think this approach makes sense then I'll do the same for Google Analytics Integration so this is ready for a second look 👀.

Thanks!

ianlin (approved):
Thanks for the change. Yeah it makes more sense to not display the copy button, LGTM.

PR #2110, opened 2023-09-26
Remove AttributeMapping new feature inbox notification

puntope:
This PR removes AttributeMapping's new Feature inbox notification

1. Install GLA
2. Go to `wp_options` and set `gla_install_timestamp` as "1669161400"
3. Go to `wc_admin_notes` table and be sure `gla-attribute-mapping-new-feature` is not there, delete it if it's there
4. Go to Woocommerce - Status - Scheduled Actions and run `wc_gla_cron_daily_notes`
5. Go to WooCommerce - Home and verify that "Our latest improvement to the Google Listings & Ads extension: Custom Attribute Mapping" is NOT appearing in the Inbox Notification
6. Verify there is no residual code related to AttributeMapping Inbox Notification

### Additional details:

p1695687924258069-slack-C0187GK80Q1

Also after discussing with @mikkamp we decided to remove this fully, as the feature was created a long time ago.

### Changelog entry

> Fix - Remove AttributeMapping new feature inbox notification


src/Notes/AttributeMappingNewFeature.php
```diff
@@ -1,73 +0,0 @@
...
-				'google-listings-and-ads'
-			)
-		);
-
-		$note->set_name( $this->get_name() );
-		$note->set_source( $this->get_slug() );
-		$note->add_action(
-			'learn-more-attribute-mapping',
-			__( 'Learn more', 'google-listings-and-ads' ),
-			$this->get_attribute_mapping_url()
```
tomalec, line 56: I think we should now remove unused `get_attribute_mapping_url` 
https://github.com/woocommerce/google-listings-and-ads/blob/fix/attribute-mapping-inbox-timestamp/src/PluginHelper.php#L203-L211

tomalec: BTW, I wonder by Codecov points to `get_documentation_url` instead :thinking: 

Anyway, removing `get_attribute_mapping_url` should make Codecov complain less about reduced coverage.

tomalec (approved):
I reviewed the code and tested it locally. Works fine.

My only comment is, that we should remove a bit of dead code. But I approve it in advance :)

puntope:
Hello @tomalec Thanks for the review. I applied the suggested change and this is ready for another round.

puntope:
> Hello @tomalec Thanks for the review. I applied the suggested change and this is ready for another round.

I saw you approved it in advance. So I merged it

PR #2111, opened 2023-09-26
Add missing tracking to onboarding when changing steps and reduce their confusing properties

### LINKED ISSUE #1467: Confusing properties of `gla_setup_mc` track event
_(This is a follow-up issue from https://github.com/woocommerce/google-listings-and-ads/pull/1408#issuecomment-1090434772)_
### Describe the problem:

When I was moving around the event tracking docs I noticed that `gla_setup_mc` is rather confusing. Its properties are described as:
> - `target` Button ID
> - `trigger` Action (e.g. `click`)

I see a few problems here:
1. `trigger` is always set to `click` - do we really need such a property, that always has a static value?
2. `target` property contains an ID of the button that **triggered** the action, not the target of the merchant's flow. (This may relate to DOM `event.target`, but this is a technical implementation detail, and I'm not sure if that matters for people analyzing data)

So when a merchant clicks a "Continue" button on step 1 and goes to step two, we fire `gla_setup_mc { trigger: 'click', target: 'step1'}`, whereas when I'd be to describe in plain English, what an event just happened I'd say:
> An event in `MC setup` just happened: the merchant `trigger`ed it by clicking/tapping/activating the "Continue" button on `step 1` and is being `target`ted to `step 2`

So, `gla_setup_mc { trigger: 'step1-continue', target 'step2' }` - that is more-or-less what I did in Pinterest in a similar use-case https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/TRACKING.md#wcadmin_pfw_setup

-----

I guess it may be problematic to change the properties of the already observed events, as that may corrupt the data reports.
However, I doubt `trigger` which was always `click` was used for anything meaningful.
Also, I suppose providing the actual target of user flow, may be a piece of useful information.

//cc @jconroy

### LINKED ISSUE #2104: Missing tracking when changing onboarding steps
### Describe the issue:
We have two different types of onboarding within the extension:
1. Merchant Center
2. Ads

Both these onboarding flows consists of several steps which must be completed. In the Merchant Center onboarding we added tracking for the event [`gla_setup_mc`](https://github.com/woocommerce/google-listings-and-ads/tree/develop/src/Tracking#gla_setup_mc) when we switch from one step to the next. When onboarding through the Ads flow I'd expect a similar tracking event for [`gla_setup_ads`](https://github.com/woocommerce/google-listings-and-ads/tree/develop/src/Tracking#gla_setup_ads). The event name is used, but it seems to only be sent when we click on a back button. And not when we move through steps within the onboarding.

One way to find the usage of the tracking is to search for `setStep` which shows that we add a call to recordEvent in MC onboarding but this same event is not there for Ads onboarding. I'd expect this to be consistent. I'm also wondering why the tracking is left out in some cases for MC onboarding.

### Steps to reproduce:
1. Disconnect Ads account if previously connected 
2. Track requests in the browser dev tools (filter for requests to pixel.wp.com) 
3. Go to Dashboard to and click on "Add paid campaign"
4. Go through the onboarding steps for creating an initial campaign
5. Confirm that we don't see any tracking sent for moving between steps 

### Expected behavior:
Tracking to be consistent amongst onboarding stepper, and occur for each step.

eason9487:

Closes #2104
Closes #1467

This PR adds the missing tracking to extension onboarding (Merchant Center) onboarding and ads onboarding when changing steps, and it also adds new properties and marks confusing properties as deprecated to reduce confusion in the future.

For ads onboarding:
- Add `gla_setup_ads` tracking to record step changes.
- Update the `gla_setup_ads` tracking in `SetupAdsTopBar` to align with the new tracking properties.

For extension onboarding:
- Add `gla_setup_mc` tracking to record step changes when clicking on the stepper.
- Adjust the `gla_setup_mc` tracking to reduce confusion in the future and also align with the `gla_setup_ads` tracking properties.

1. Enable tracking debug tool
   - Open the Console tab of the browser DevTool.
   - Change the filter level to "All levels".
   - Run `localStorage.setItem( 'debug', 'wc-admin:*' )` and refresh the page to make it work.
   - It should be able to view tracking via the Console tab. Ref:
      
2. Test extension onboarding
   1. Go to the Get Started page.
   2. Click on both "Start listing products" buttons to see if the new tracking properties are correct and the old properties stay the same.
   3. Click on the back button at the top-left of the onboarding flow to check the tracking properties.
   4. Go to the onboarding flow.
   5. Changing steps via the "Continue" buttons and the stepper to check the new tracking and properties.
3. Test ads onboarding
   1. Go to the ads onboarding flow.
   2. Changing steps via the "Continue" buttons and the stepper to check the new tracking and properties.
   3. Click on the back button at the top-left to check the tracking properties.

### Changelog entry

> Add - The missing tracking to onboarding when changing steps.
> Tweak - The properties of the gla_setup_mc and gla_setup_ads tracking events to reduce their confusion.

### DIFF
--- a/js/src/get-started-page/get-started-with-video-card/index.js
+++ b/js/src/get-started-page/get-started-with-video-card/index.js
@@ -17,7 +17,7 @@ import { getSetupMCUrl } from '.~/utils/urls';
 import './index.scss';
 
 /**
- * @fires gla_setup_mc with `{ target: 'set_up_free_listings', trigger: 'click', context: 'get-started-with-video' }`.
+ * @fires gla_setup_mc with `{ triggered_by: 'start-onboarding-button', action: 'go-to-onboarding', context: 'get-started-with-video', target: 'set_up_free_listings', trigger: 'click' }`.
  * @fires gla_documentation_link_click with `{ context: 'get-started-with-video', linkId: 'wp-terms-of-service', href: 'https://wordpress.com/tos/' }`.
  */
 const GetStartedWithVideoCard = () => {
@@ -67,9 +67,12 @@ const GetStartedWithVideoCard = () => {
 					href={ getSetupMCUrl() }
 					eventName="gla_setup_mc"
 					eventProps={ {
+						triggered_by: 'start-onboarding-button',
+						action: 'go-to-onboarding',
+						context: 'get-started-with-video',
+						// 'target' and 'trigger' were deprecated and can be removed after Q1 2024.
 						target: 'set_up_free_listings',
 						trigger: 'click',
-						context: 'get-started-with-video',
 					} }
 				>
 					{ __(
--- a/js/src/setup-ads/ads-stepper/index.js
+++ b/js/src/setup-ads/ads-stepper/index.js
@@ -4,6 +4,7 @@
 import { Stepper } from '@woocommerce/components';
 import { __ } from '@wordpress/i18n';
 import { useState } from '@wordpress/element';
+import { recordEvent } from '@woocommerce/tracks';
 
 /**
  * Internal dependencies
@@ -12,24 +13,43 @@ import SetupAccounts from './setup-accounts';
 import AdsCampaign from '.~/components/paid-ads/ads-campaign';
 import SetupBilling from './setup-billing';
 
-const AdsStepper = ( props ) => {
-	const { formProps } = props;
+/**
+ * @param {Object} props React props
+ * @param {Object} props.formProps Form props forwarded from `Form` component.
+ * @fires gla_setup_ads with `{ triggered_by: 'step1-continue-button' | 'step2-continue-button' , action: 'go-to-step2' | 'go-to-step3' }`.
+ * @fires gla_setup_ads with `{ triggered_by: 'stepper-button', action: 'go-to-step1' | 'go-to-step2' }`.
+ */
+const AdsStepper = ( { formProps } ) => {
 	const [ step, setStep ] = useState( '1' );
 
 	// Allow the users to go backward only, not forward.
 	// Users can only go forward by clicking on the Continue button.
 	const handleStepClick = ( value ) => {
 		if ( value < step ) {
+			recordEvent( 'gla_setup_ads', {
+				triggered_by: 'stepper-button',
+				action: `go-to-step${ value }`,
+			} );
 			setStep( value );
 		}
 	};
 
+	const continueStep = ( to ) => {
+		const from = step;
+
+		recordEvent( 'gla_setup_ads', {
+			triggered_by: `step${ from }-continue-button`,
+			action: `go-to-step${ to }`,
+		} );
+		setStep( to );
+	};
+
 	const handleSetupAccountsContinue = () => {
-		setStep( '2' );
+		continueStep( '2' );
 	};
 
 	const handleCreateCampaignContinue = () => {
-		setStep( '3' );
+		continueStep( '3' );
 	};
 
 	return (
--- a/js/src/setup-ads/ads-stepper/index.js
+++ b/js/src/setup-ads/ads-stepper/index.js
@@ -17,7 +17,7 @@ import SetupBilling from './setup-billing';
  * @param {Object} props React props
  * @param {Object} props.formProps Form props forwarded from `Form` component.
  * @fires gla_setup_ads with `{ triggered_by: 'step1-continue-button' | 'step2-continue-button' , action: 'go-to-step2' | 'go-to-step3' }`.
- * @fires gla_setup_ads with `{ triggered_by: 'stepper-button', action: 'go-to-step1' | 'go-to-step2' }`.
+ * @fires gla_setup_ads with `{ triggered_by: 'stepper-step1-button' | 'stepper-step2-button', action: 'go-to-step1' | 'go-to-step2' }`.
  */
 const AdsStepper = ( { formProps } ) => {
 	const [ step, setStep ] = useState( '1' );
@@ -27,7 +27,7 @@ const AdsStepper = ( { formProps } ) => {
 	const handleStepClick = ( value ) => {
 		if ( value < step ) {
 			recordEvent( 'gla_setup_ads', {
-				triggered_by: 'stepper-button',
+				triggered_by: `stepper-step${ value }-button`,
 				action: `go-to-step${ value }`,
 			} );
 			setStep( value );
--- a/js/src/setup-ads/ads-stepper/index.js
+++ b/js/src/setup-ads/ads-stepper/index.js
@@ -34,6 +34,11 @@ const AdsStepper = ( { formProps } ) => {
 		}
 	};
 
+	/**
+	 * Handles "onContinue" callback to set the current step and record event tracking.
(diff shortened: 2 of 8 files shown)


js/src/get-started-page/get-started-with-video-card/index.js
```diff
@@ -67,9 +67,12 @@ const GetStartedWithVideoCard = () => {
 					href={ getSetupMCUrl() }
 					eventName="gla_setup_mc"
 					eventProps={ {
+						triggered_by: 'start-onboarding-button',
+						action: 'go-to-onboarding',
+						context: 'get-started-with-video',
+						// 'target' and 'trigger' were deprecated and can be removed after Q1 2024.
```
puntope, line 73: I see in other tracking actions added in this PR we don't use any more `trigger` and `target`. So, if they are not necessary, why not to remove them now?

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -12,24 +13,43 @@ import SetupAccounts from './setup-accounts';
...
 		}
 	};
 
+	const continueStep = ( to ) => {
+		const from = step;
+
+		recordEvent( 'gla_setup_ads', {
+			triggered_by: `step${ from }-continue-button`,
+			action: `go-to-step${ to }`,
+		} );
```
puntope, line 48: I guess this is to avoid a race condition between the current `step` and the `step` after being updated by `setStep( to );`.

If is like this, I would add a comment clarifying so.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -12,24 +13,43 @@ import SetupAccounts from './setup-accounts';
...
 		if ( value < step ) {
+			recordEvent( 'gla_setup_ads', {
+				triggered_by: 'stepper-button',
+				action: `go-to-step${ value }`,
+			} );
 			setStep( value );
 		}
 	};
 
+	const continueStep = ( to ) => {
```
puntope, line 42: At least in the new created functions would be nice to add JSDOC comments

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -12,24 +13,43 @@ import SetupAccounts from './setup-accounts';
...
+ */
+const AdsStepper = ( { formProps } ) => {
 	const [ step, setStep ] = useState( '1' );
 
 	// Allow the users to go backward only, not forward.
 	// Users can only go forward by clicking on the Continue button.
 	const handleStepClick = ( value ) => {
 		if ( value < step ) {
+			recordEvent( 'gla_setup_ads', {
+				triggered_by: 'stepper-button',
```
puntope: Since we are adding `step1-continue-button `, `step2-continue-button `...etc I wonder if it would be better to add the origin step here. i.e `step1-stepper-button`

puntope:
Would be nice to add some unit tests covering these events

js/src/utils/recordEvent.js
```diff
@@ -76,9 +76,11 @@ export const recordTablePageEvent = ( context, page, direction ) => {
  * Setup Merchant Center
  *
  * @event gla_setup_mc
- * @property {string} target Button ID
- * @property {string} trigger Action (e.g. `click`)
- * @property {string} context Indicates which CTA is clicked
+ * @property {string} triggered_by Button name
```
puntope:
```suggestion
 * @property {string} triggered_by Indicates which button triggered this event
```

js/src/utils/recordEvent.js
```diff
@@ -107,6 +109,16 @@ export const recordTablePageEvent = ( context, page, direction ) => {
  * @property {string} href Link's URL
  */
 
+/**
+ * Triggered on events during ads onboarding
+ *
+ * @event gla_setup_ads
+ * @property {string} triggered_by Button name
```
puntope:
```suggestion
 * @property {string} triggered_by Indicates which button triggered this event
```

js/src/get-started-page/get-started-with-video-card/index.js
eason9487: As it will take some time for a larger number of user-end extensions to be updated to the new version, for tracking events that have already recorded `trigger` and `target`, keeping them for a period is to reduce the degree of data interruption due to event property adjustments when analyzing tracking data. It also follows the suggestion in https://github.com/woocommerce/google-listings-and-ads/issues/2104#issuecomment-1727532869.

js/src/setup-ads/ads-stepper/index.js
eason9487: The `from` variable is to pair with the `to` variable so that the semantic would be a bit more readable than:

```js
recordEvent( 'gla_setup_ads', {
	triggered_by: `step${ step }-continue-button`,
	action: `go-to-step${ to }`,
} );
```

js/src/setup-ads/ads-stepper/index.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/2111/commits/465a8e24c9dc6cecc32de918c2952faf8e6a38e7.

js/src/setup-ads/ads-stepper/index.js
eason9487: Tweaked in https://github.com/woocommerce/google-listings-and-ads/pull/2111/commits/6ed73ac53ae0935ad0b3182f8113ea63770c972e.

I guess which one would be better will depend on the users of tracking data, the goal here is to reduce the confusion of the data itself.

The `stepN-continue-button` naming means the continue button in step N. But the stepper buttons exist in all steps, so I named it `stepper-stepN-button`.

eason9487:

This PR is ready for the second round of code reviews. Could you help with it?

js/src/setup-ads/ads-stepper/index.js
puntope: Oh, I see, It was confusing for me why we did that. As the step variable is the current step.

puntope (approved):
✅  LGTM

Thanks for further adjustments.

mikkamp:
@eason9487 I was testing these changes and I noticed that for the regular Ads onboarding and if the ads account was connected during MC onboarding, then the stepper events work fine.

However when we create an additional campaign we also do this in a 2 step process:
1. Create paid campaign
2. Optimize your campaign

Going from step 1 > 2 we have a continue button without any tracking. Is that a completely different stepper within the code or can we get some similar tracking setup there? Let me know if I should create an additional issue.

eason9487:
Hi @mikkamp

> However when we create an additional campaign we also do this in a 2 step process:
> 
> 1. Create paid campaign
> 2. Optimize your campaign
> 
> Going from step 1 > 2 we have a continue button without any tracking. Is that a completely different stepper within the code or can we get some similar tracking setup there?

Yes, these event tracking codes are not within the `<Stepper>` and `<AdsCampaign>` components, they are each different instances on each page.

> Let me know if I should create an additional issue.

As the event tracking to be added is similar, I think opening a PR and referencing this comment and #2104 would be fine. I will do so later.

PR #2114, opened 2023-09-27
Add QIT workflow

tomalec:
:warning: depends on https://github.com/woocommerce/grow/pull/81

1. Change the `build` workflow so it publish an artifact and can be called by other workflows.
2. Add a manual workflow, that builds the extension and run selected QIT tests for that package.

### Screenshots:

image

1. Fork the repo
2. Setup secrets. Links to secrets are at Pe98dy-Ir-p2 
3. Merge this branch to the main one /`trunk`
4. Run the manual workflow at  https://github.com/woocommerce/google-listings-and-ads/actions/workflows/run-qit.yml

### Additional details:

1. Before margin we need to remove those lines 
https://github.com/woocommerce/google-listings-and-ads/pull/2114/files#diff-0550f940c28a3fbd24937bad01914ede672b538aff10de709e999759ccf4f9dfR41-R44

### Changelog entry

> Dev - Add manual QIT workflow

### DIFF
--- a/.github/workflows/run-qit.yml
+++ b/.github/workflows/run-qit.yml
@@ -42,6 +42,7 @@ on:
   push:
     branches:
       - dev/qit-workflow
+      - dev/qit-workflow-alternative
 
 jobs:
   build:
@@ -50,17 +51,25 @@ jobs:
     secrets: inherit
   qit-tests:
     name: Run QIT Tests
-    # Update it with more stable path once merged.
-    uses: woocommerce/grow/.github/workflows/run-qit-extension.yml@add/qit-workflows
+    runs-on: ubuntu-20.04
     needs: build
-    secrets: inherit
-    with:
-      version: local
-      # Conditional statements are here to allow testing on push triggers, without manual input. To be removed before merging.
-      test-activation: ${{ inputs.test-activation || true }}
-      test-security: ${{ inputs.test-security || true }}
-      test-phpstan: ${{ inputs.test-phpstan || true }}
-      test-api: ${{ inputs.test-api || true }}
-      test-e2e: ${{ inputs.test-e2e || false }}
-      extension: 'google-listings-and-ads'
-      options: ${{ inputs.options }}
+    steps:
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: google-listings-and-ads.zip
+      - name: Delegate QIT Tests
+        # Update it with more stable path once merged.
+        uses: woocommerce/grow/packages/github-actions/actions/run-qit-extension@add/qit-workflows-alternative
+        with:
+          qit-partner-user: ${{ secrets.QIT_PARTNER_USER }}
+          qit-partner-secret: ${{ secrets.QIT_PARTNER_SECRET }}
+          version: local
+          extension: 'google-listings-and-ads'
+          # Conditional statements are here to allow testing on push triggers, without manual input. To be removed before merging.
+          test-activation: ${{ inputs.test-activation || true }}
+          test-security: ${{ inputs.test-security || true }}
+          test-phpstan: ${{ inputs.test-phpstan || true }}
+          test-api: ${{ inputs.test-api || true }}
+          test-e2e: ${{ inputs.test-e2e || false }}
+          options: ${{ inputs.options }}
--- a/.github/workflows/run-qit.yml
+++ b/.github/workflows/run-qit.yml
@@ -52,6 +52,7 @@ jobs:
     name: Run QIT Tests
     # Update it with more stable path once merged.
     uses: woocommerce/grow/.github/workflows/run-qit-extension.yml@add/qit-workflows
+    needs: build
     secrets: inherit
     with:
       version: local
--- a/.github/workflows/run-qit.yml
+++ b/.github/workflows/run-qit.yml
@@ -60,7 +60,7 @@ jobs:
       test-activation: ${{ inputs.test-activation || true }}
       test-security: ${{ inputs.test-security || true }}
       test-phpstan: ${{ inputs.test-phpstan || true }}
-      test-api: ${{ inputs.test-api || false }}
+      test-api: ${{ inputs.test-api || true }}
       test-e2e: ${{ inputs.test-e2e || false }}
       extension: 'google-listings-and-ads'
       options: ${{ inputs.options }}
--- a/.github/workflows/run-qit.yml
+++ b/.github/workflows/run-qit.yml
@@ -1,0 +1,65 @@
+name: Run QIT
+
+# **What it does**: Runs a suite of QIT tests for the extension.
+# **Why we have it**: To be able to check QIT compatibility at once. For example, to test a specific branch, or upcoming release.
+
+on:
+  workflow_dispatch:
+    inputs:
+      # Configure which tests to run.
+      test-activation:
+        description: 'Should activation be tested?'
+        required: true
+        default: true
+        type: boolean
+      test-security:
+        description: 'Should security be tested?'
+        required: true
+        default: true
+        type: boolean
+      test-phpstan:
+        description: 'Should phpstan be tested?'
+        required: true
+        default: true
+        type: boolean
+      test-api:
+        description: 'Should API be tested?'
+        required: true
+        default: true
+        type: boolean
+      test-e2e:
+        description: 'Should E2E be tested? (takes a lot of time)'
+        required: true
+        default: false
+        type: boolean
+
+      # Advanced customization.
+      options:
+        description: 'Additional options for `qit` command, like `--optional_features=hpos`.'
+        required: false
+
+  # Used to test this PR, to be removed after before merging.
+  push:
+    branches:
+      - dev/qit-workflow
+
+jobs:
+  build:
+    name: Build extension
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
+  qit-tests:
+    name: Run QIT Tests
+    # Update it with more stable path once merged.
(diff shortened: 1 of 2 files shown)


.github/workflows/run-qit.yml
```diff
@@ -0,0 +1,66 @@
...
+      test-api:
+        description: 'Should API be tested?'
+        required: true
+        default: true
+        type: boolean
+      test-e2e:
+        description: 'Should E2E be tested? (takes a lot of time)'
+        required: true
+        default: false
+        type: boolean
```
puntope, line 38: Are not the e2e tests already in another workflow?

.github/workflows/run-qit.yml
```diff
@@ -0,0 +1,66 @@
...
+    uses: woocommerce/grow/.github/workflows/run-qit-extension.yml@add/qit-workflows
+    needs: build
+    secrets: inherit
+    with:
+      version: local
+      # Conditional statements are here to allow testing on push triggers, without manual input. To be removed before merging.
+      test-activation: ${{ inputs.test-activation || true }}
+      test-security: ${{ inputs.test-security || true }}
+      test-phpstan: ${{ inputs.test-phpstan || true }}
+      test-api: ${{ inputs.test-api || false }}
```
puntope: Test API is true by default in the inputs. Shouldn't be this consistent with it, as the rest of the parameters?

.github/workflows/run-qit.yml
```diff
@@ -0,0 +1,66 @@
...
+  # Used to test this PR, to be removed after before merging.
+  push:
+    branches:
+      - dev/qit-workflow
+
+jobs:
+  build:
+    name: Build extension
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
```
puntope, line 54: If we generated the artifact? Why do we need to run again the build? Shouldn't we download the artefact instead?

puntope:
Screenshot 2023-09-29 at 11 58 02

I can see the artefact generated... however, when I download it, it generates a zip named `google-listing-and-ads.zip.zip ` so, it's double-zipped. 

Screenshot 2023-09-29 at 11 58 14

https://github.com/actions/upload-artifact#zipped-artifact-downloads

Not sure if this will be an issue for downloading the artefact... or we should maybe add the directory instead.

puntope:
I tried to run QIT workflow in the fork, and followed the steps... but I don't see it.

Screenshot 2023-09-29 at 11 59 48

What I'm doing wrong?

puntope:
> I tried to run QIT workflow in the fork, and followed the steps... but I don't see it.
> 
> Screenshot 2023-09-29 at 11 59 48
> What I'm doing wrong?

Ok, I solved it by merging the branch to develop instead of trunk.

puntope:
I added QIT_PARTNER_SECRET and QIT_PARTNER_USER in my forked repo secrets

But I got this error

Screenshot 2023-09-29 at 12 15 46

.github/workflows/run-qit.yml
tomalec: These are QIT E2E tests, not local GLA E2Es.

`qit run:e2e` run the suite of E2E tests to check if installing GLA does not break WC core behavior.

[`.github/workflows/e2e-tests.yml`](https://github.com/woocommerce/google-listings-and-ads/blob/develop/.github/workflows/e2e-tests.yml) (if that's the workflow you mean) - tests GLA-specific behavior

.github/workflows/run-qit.yml
tomalec: > If we generated the artifact

This is the only job in this workflow that generates an artifact.

-----

I do see we have a redundancy in general. As for example, for release PR, we run the build multiple times:
- once in general, build for bundle size on  PR trigger - https://github.com/woocommerce/google-listings-and-ads/blob/trunk/.github/workflows/build.yml#L3-L9
- once running local E2E on release PR trigger - https://github.com/woocommerce/google-listings-and-ads/blob/trunk/.github/workflows/e2e-tests.yml#L4-L46
- once here, we would also call it for release PRs.

However, to solve that, we would need to restructure our existing workflows, which is out of the scope of this PR.

We would need to make workflow interact with each other more, fetching their artifacts, controlling concurrency, etc. But I think we could benefit from taking one step back, and doing a more overall change.

:bulb: 
Suppose we redesign our workflows to distinguish reusable, composable blocks ("what") from actual flow definitions ("when _ then _"). Something like:


> reusable blocks: // What **work** should be done (technically **work**flows, jobs, & actions)
> 🏗 build package
> 👁 apply bundle watch
> 🧪 run local E2E tests
> 🆏 run QIT
> 🚀 release `gha-dev-build`
> 📖 deploy storybook
> 
> 
> work**flows**: // When what should happen - the actual flow
> - On Release PR: 🏗 , (👁, 🧪, 🆏)
> - On regular PR 🏗 , 👁
> - On push to develop 🏗 , (👁, 🧪, 🆏, 📖, 🚀)

- 
It would be easier to avoid redundancy, parallelize, and IMHO reason & maintain.
As we start to reason from the flow perspective, 
Q: "What is the process that we automate?",  
A: "release" / "regular feature development" / "develop branch consolidation" / "compatibility checks"
Q: "When do we do that?"
A: "for `release/` branch" / "for any PR" / "for pushes to `develop`" / "manual call"
Q: "What do we do?"
A: "This then that"

Instead of what we have currently:
```
run build, bundle watch, and conditional `gha-dev-build` publish on pushes and all pull requests
run storybook on pushes to develop & trunk
run build & E2E on pushes and PRs to `release/*` & manual calls
run build & QIT on manual calls, pushes to develop, and release PRs
```
When we reason only from the perspective of job, that's done. 
Q: "What do we do?"
A: "Build the package to watch its size & optionally publish it" / "Build the package to run E2E tests" / ..
Q: "When do we do that?"
A: "On releases, but also on regular pushes, but then we do only part of it. We also do the thing when we do that other thing..." :confused: 
Q: "What is the process that we automate?"
A: ":thinking: ... work?"

But that's a discussion and project on its own :)

tomalec:
> I added QIT_PARTNER_SECRET and QIT_PARTNER_USER in my forked repo secrets
>
> But I got this error

Could you expand the command run?

Also, did you use the CLI credentials from Pe98dy-Ir-p2 ?

tomalec:
@puntope Thank you for the review :) would you mind another round?

puntope:
> Also, did you use the CLI credentials from Pe98dy-Ir-p2 ?

YEs, I used the "For CLI and GitHub Workflows" secret in the mentioned P2

puntope:
> Could you expand the command run?

Screenshot 2023-10-02 at 13 06 05

Screenshot 2023-10-02 at 13 05 11

This is with everything expanded, for some reason is not getting the credentials. This is how I set the action secrets

 
Screenshot 2023-10-02 at 13 07 15

I take as granted they should be like this as the credentials is set as inherit, so I checked the GLA credentials section to see the names.

.github/workflows/run-qit.yml
puntope: Hey @tomalec sorry maybe I expressed wrong my question.

I see this PR generates an artifact in the build workflow.

Why is this for? 

My assumption was that it was needed for the new Qit workflow... but I see we are running the full build again instead of downloading the artifact.

tomalec: `woocommerce/grow/.github/workflows/run-qit-extension.yml@add/qit-workflows` 
does not run the build, it downloads the artifact :confused: 

https://github.com/woocommerce/grow/pull/81/files#diff-f653f2a856040f6c20d52799af46664b5dc76e12be91f8443c4f04f22bdf052dR78-R82

The workflow here `.github/workflows/run-qit.yml`:

0. On manual trigger
1. Runs the build workflow (`.github/workflows/build.yml`)
2. Runs the QIT which uses the artifact (`woocommerce/grow/.github/workflows/run-qit-extension.yml`)

Then we also have a case when the build is run on other triggers `.github/workflows/build.yml`:

0. On PRs and pushes to trunk/develop
1. Run the build workflow

tomalec:
I updated the PR, to use the latest version of grow action (instead of a workflow), so it is ready for another round.

puntope:
Hi @tomalec I tested again but got this error

Screenshot 2024-01-31 at 11 20 34

.github/workflows/run-qit.yml
```diff
@@ -0,0 +1,85 @@
...
+  qit-tests:
+    name: Run QIT Tests
+    runs-on: ubuntu-20.04
+    needs: build
+    steps:
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: google-listings-and-ads.zip
+      - name: Delegate QIT Tests
```
puntope: 💅  Even if it's true that we delegate the QIT Test to the Grow Repo... I would just call this `Run QIT Tests` as it's the final goal of what we are doing. The same way we do for other workflows. We don't call them Delegate Get Woocommerce Matrix or Delegate Run PHP tests.

puntope (approved):
Hi @tomalec. This seems like working ✅  so I'm approving it in advance taking these points into consideration. 

- I left one 💅 suggestion  about naming
- ⚠️  We still need to remove the lines mentioned in the PR description before merging
- 💅  We might wait for https://github.com/woocommerce/grow/pull/81 to be merged and update the branch reference

puntope:
p.s Notice the phpstan checks are failing tho...

Screenshot 2024-01-31 at 11 53 01

tomalec:
Thank you for the review :)

>p.s Notice the phpstan checks are failing tho...

This is kinda expected, as GLA fails to pass those.

1. I pushed your suggestion.
2. I'll update the references, now that it is reviewed.
3. Yep, let's wait.

PR #2096, opened 2023-09-14
[WC 8.1] - Fix Selector and Target for Add To Cart 

puntope:
### Changes proposed in this Pull Request:

Closes #2095 

This PR fixes 2 bugs.

- "Added" doesn't match anymore as a valid selector when the product is added to cart
- Change `target` to `currentTarget` Avoiding the click event being triggered in children elements (span) that doesn't contain the data-product-id attribute, and hence, GLA ID is set as undefined. 

See https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

### Screenshots
https://github.com/woocommerce/google-listings-and-ads/actions/runs/6186698347/job/16794930005

### Detailed test instructions:

1. Run `npm run wp-env:up` (be sure you're using WC 8.1, you can use `npm run -- wp-env run tests-cli -- wp core update --version=8.1.0`) 
2. Run `npm run test:e2e  gtag-events/gtag-events.test.js`
3. Verify that test are passing
4. Install WC 8.1 in your site and also TT2.
5. Go to Shop
6. Click on several products on the text inside the button (so, clicking in the span element inside the button) And Verify gla ID is loaded correctly 

### Changelog entry

> Fix - Tweak E2E tests for WC 8.1
> Fix - Adjust target on click events preventing GLA ID to be Undefined

Issue #2095, opened 2023-09-14
E2E Tests failing with WooCommerce 8.1.0


martynmjones:
### Describe the bug:
A few of the E2E tests are failing when run against WooCommerce `8.1.0`. There is not an issue with any previous version.

![Screenshot 2023-09-14 at 12 49 39](https://github.com/woocommerce/google-listings-and-ads/assets/40762232/21ce4bfa-6f63-4f6c-b0ad-1c0250481f18)

### Steps to reproduce:
1. Checkout develop
2. `npm run wp-env:up`
3. `npm run test:e2e`
4. See errors

puntope:
Checking

puntope:
I tested the events in the store itself GLA 2.5.5 + WC 8.1.0 + TT2 and it tracking seems working. 

Seems like problem is the class "added" which is not anymore there

puntope:
Another issue I'm still investigating is that sometimes GLA Id is undefined. THis also happens in the Shop. Not only in the tests

puntope:
Hi @martynmjones 

I created a PR for this https://github.com/woocommerce/google-listings-and-ads/pull/2096

PR #2088, opened 2023-09-07
E2E tests: Onboarding Step 2 - Configure product listings

ianlin:
### Changes proposed in this Pull Request:

Closes part of #2070, this PR adds E2E tests for [📌 Onboarding Step 2 - Configure product listings][1].

### Detailed test instructions:

1. Follow these steps to set up the e2e env: https://github.com/woocommerce/google-listings-and-ads#e2e-testing
2. Run `npm run test:e2e -- ./tests/e2e/specs/setup-mc/step-2-product-listings.test.js`, the test should pass.
3. Or, run `npm run test:e2e-dev -- ./tests/e2e/specs/setup-mc/step-2-product-listings.test.js`, check the test in the chromium step by step.
4. Run `npm run test:e2e` to make sure all tests are passed.

### Changelog entry

> Dev - E2E - Onboarding Step 2 - Configure product listings

[1]: https://github.com/woocommerce/google-listings-and-ads/issues/2070#onboarding-step-2

PR #2122, opened 2023-10-04
Fix E2E tests - Track gtag event on specific page

jorgemd24:

As of now, the `trackGtagEvent` function only checks for the event itself. Yet, there are scenarios where the same event may carry different properties, and within a brief timeframe, the same event can be triggered with varying properties. Since we don't perform an assertion to wait for a request with specific data, these tests could potentially fail.

https://github.com/woocommerce/google-listings-and-ads/blob/78582d6ffab31b5bf3dfcf97e69fcbf623f0b932/tests/e2e/utils/track-event.js#L15-L21

This PR tweaks the `trackGtagEvent` function. It will now wait for a specific event and also provides the capability to verify if the "ecomm_pagetype" data is sent.

1. Follow these steps to set up the e2e env: https://github.com/woocommerce/google-listings-and-ads#e2e-testing
2. Run `npm run test:e2e -- ./tests/e2e/specs/gtag-events/gtag-events.test.js` and see that all tests passes.

### Changelog entry

> Dev - E2E tests - Track gtag event on specific page

### DIFF
--- a/tests/e2e/utils/track-event.js
+++ b/tests/e2e/utils/track-event.js
@@ -17,13 +17,13 @@ export function trackGtagEvent( page, eventName, urlPath = null ) {
 	const eventPath = '/pagead/';
 	return page.waitForRequest( ( request ) => {
 		const url = request.url();
-		const match = encodeURIComponent( 'event=' + eventName );
+		const match = 'event=' + eventName;
 		const origin = new URL( page.url() ).origin;
 		const params = new URL( url ).searchParams;
 
 		return (
 			url.includes( eventPath ) &&
-			url.includes( match ) &&
+			params.get( 'data' ).includes( match ) &&
 			( urlPath
 				? params
 						.get( 'url' )
--- a/tests/e2e/utils/track-event.js
+++ b/tests/e2e/utils/track-event.js
@@ -10,18 +10,26 @@
  *
  * @param {Page} page
  * @param {string} eventName Event name to match.
- * @param  {boolean} ecommPagetype Whether to check for ecomm_pagetype in the URL.
+ * @param {string|null} URLpath The URL path where the event should be triggered.
  * @return {Promise<Request>} Matching request.
  */
-export function trackGtagEvent( page, eventName, ecommPagetype = true ) {
+export function trackGtagEvent( page, eventName, URLpath = null ) {
 	const eventPath = '/pagead/';
 	return page.waitForRequest( ( request ) => {
 		const url = request.url();
 		const match = encodeURIComponent( 'event=' + eventName );
+		const origin = new URL( page.url() ).origin;
+
 		return (
 			url.includes( eventPath ) &&
 			url.includes( match ) &&
-			( ecommPagetype ? url.includes( 'ecomm_pagetype' ) : true )
+			( URLpath
+				? url.includes(
+						`url=${ encodeURIComponent(
+							`${ origin }/${ URLpath }`
+						) }`
+				  )
+				: true )
 		);
 	} );
 }
--- a/tests/e2e/utils/track-event.js
+++ b/tests/e2e/utils/track-event.js
@@ -10,25 +10,24 @@
  *
  * @param {Page} page
  * @param {string} eventName Event name to match.
- * @param {string|null} URLpath The URL path where the event should be triggered.
+ * @param {string|null} urlPath The starting path to match where the event should be triggered.
  * @return {Promise<Request>} Matching request.
  */
-export function trackGtagEvent( page, eventName, URLpath = null ) {
+export function trackGtagEvent( page, eventName, urlPath = null ) {
 	const eventPath = '/pagead/';
 	return page.waitForRequest( ( request ) => {
 		const url = request.url();
 		const match = encodeURIComponent( 'event=' + eventName );
 		const origin = new URL( page.url() ).origin;
+		const params = new URL( url ).searchParams;
 
 		return (
 			url.includes( eventPath ) &&
 			url.includes( match ) &&
-			( URLpath
-				? url.includes(
-						`url=${ encodeURIComponent(
-							`${ origin }/${ URLpath }`
-						) }`
-				  )
+			( urlPath
+				? params
+						.get( 'url' )
+						.includes( `${ `${ origin }/${ urlPath }` }` )
 				: true )
 		);
 	} );
--- a/tests/e2e/utils/track-event.js
+++ b/tests/e2e/utils/track-event.js
@@ -10,14 +10,19 @@
  *
  * @param {Page} page
  * @param {string} eventName Event name to match.
+ * @param  {boolean} ecommPagetype Whether to check for ecomm_pagetype in the URL.
  * @return {Promise<Request>} Matching request.
  */
-export function trackGtagEvent( page, eventName ) {
+export function trackGtagEvent( page, eventName, ecommPagetype = true ) {
 	const eventPath = '/pagead/';
 	return page.waitForRequest( ( request ) => {
 		const url = request.url();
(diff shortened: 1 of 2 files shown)


mikkamp:
> Yet, there are scenarios where the same event may carry different properties, and within a brief timeframe, the same event can be triggered with varying properties.

Can you clarify which scenario this was observed? Is this where we listen for events across various page views or is it for a single page view?
Sending the same event with different parameters on a single page sounds like something isn't being tracked correctly.

jorgemd24:
Hi @mikkamp, it is because of a timing issue in the E2E tests:

https://github.com/woocommerce/google-listings-and-ads/blob/78582d6ffab31b5bf3dfcf97e69fcbf623f0b932/tests/e2e/specs/gtag-events/gtag-events.test.js#L152-L155

When the `singleProductAddToCart` function is called, it calls the `page_view` event without including `ecomm_pagetype`, now I realized that we changed from `networkidle` to `domcontentloaded`, which means that the waitForRequest function could monitor the incorrect request. See here: https://github.com/woocommerce/google-listings-and-ads/pull/2088/files#diff-ad23d05dc6460c47ba77086291640c8373da27bae0b441f8cf9465032e6079e2L34. 

So basically the issue is that `waitForRequest ` was expecting to catch the request in the cart page instead of product page. 


We opted against using `networkidle` because it's discouraged in the documentation and in general it increases the execution time:

`'networkidle' - DISCOURAGED wait until there are no network connections for at least 500 ms. Don't use this method for testing, rely on web assertions to assess readiness instead.
`

However, I'm not sure in this particular scenario, considering what you mentioned earlier about having only one type of event request per page. In this case, it might be worth considering using 'networkidle.'

mikkamp:
Thanks for the explanation, that's helpful to understand the scenario.

I think I'd agree though that in this case it's almost good to have waitForRequest fail as it highlighted the issue that it's not receiving the right event.

I can understand not wanting to use `networkidle`, is it maybe better to switch to waiting till something unique on the cart / product page loads? So we can be certain we are hooking into the event tracking at the right time?

Just to clarify I think the solution in this PR just makes it slightly less transparent what we are actually testing, because of the change it's not apparent that we want to catch the tracking event from the second page load and avoid the one from the first page. In this case they happen to have different parameters, but in theory there might be scenarios where the event is exactly identical.

jorgemd24:
> I can understand not wanting to use networkidle, is it maybe better to switch to waiting till something unique on the cart / product page loads? So we can be certain we are hooking into the event tracking at the right time?

Yes, I will try to look for something "unique" to see if we can avoid the `networkidle`.

> Just to clarify I think the solution in this PR just makes it slightly less transparent what we are actually testing, because of the change it's not apparent that we want to catch the tracking event from the second page load and avoid the one from the first page. In this case they happen to have different parameters, but in theory there might be scenarios where the event is exactly identical.

Yep, when I initially created the PR, I thought it was more of a race condition problem on the same page. However, later on, I realized it was due to the switch from 'networkidle' to 'domcontentloaded.'

jorgemd24:
@mikkamp, I've made some adjustments to the PR to make it clearer on which page the event should be triggered. Let me know what you think. Thanks!

tests/e2e/utils/track-event.js
```diff
@@ -10,14 +10,27 @@
  *
  * @param {Page} page
  * @param {string} eventName Event name to match.
+ * @param {string|null} URLpath The URL path where the event should be triggered.
```
mikkamp: I don't think we need to over engineer this function but maybe add a better description that it matches the starting part of the path.

For example if we go to checkout: `https://domain.test/checkout/`
When we complete checkout we are redirected to: `https://domain.test/checkout/order-received/<orderId>/`

Both of those would match if URLpath is set to `checkout`

tests/e2e/utils/track-event.js
```diff
@@ -10,14 +10,27 @@
...
  * @return {Promise<Request>} Matching request.
  */
-export function trackGtagEvent( page, eventName ) {
+export function trackGtagEvent( page, eventName, URLpath = null ) {
```
mikkamp: Not sure what the recommended format is for naming, but wouldn't `urlPath` or `URLPath` be the better choice?

tests/e2e/utils/track-event.js
```diff
@@ -10,14 +10,27 @@
...
+			url.includes( eventPath ) &&
+			url.includes( match ) &&
+			( URLpath
+				? url.includes(
+						`url=${ encodeURIComponent(
+							`${ origin }/${ URLpath }`
+						) }`
+				  )
+				: true )
+		);
```
mikkamp: I'm not an expert in JS code but it takes me a while to grok what's being compared here.
I wonder if it would be simpler to convert the request URL to a [URL object](https://developer.mozilla.org/en-US/docs/Web/API/URL) so we can check if the event `searchParams` is set and matches the eventName.

That would also allow for more accurate matching with the exact URL. Looking at the payload of a request I see:
image

So using the `url.includes` might not be very foolproof as it could easily match the referrer URL which would be set to the previous page visited.

mikkamp (commented):
When switching to this branch I do get the test to work correctly now:
> GTag events › Cart page view event is sent from the cart page

Although for some reason the following two tests still fail locally (although that seems related to something outside of this PR):
> GTag events › Add to cart event is sent from related product on single product page
> GTag events › Add to cart event is sent from the shop page

As outlined in the comments I think we might want to improve the matching a bit more so we don't run into problems later if we use this function to track other scenarios.

tests/e2e/utils/track-event.js
jorgemd24: Thanks, @mikkamp, for the suggestion. I don't believe that scenario would occur because I added the prefix `url=` within `url.includes`. Nonetheless, I've made the switch to using URL parameters for the sake of clarity.

jorgemd24:
@mikkamp Thanks for your suggestions, I addressed all of them.

> Although for some reason the following two tests still fail locally (although that seems related to something outside of this PR):

mmmm, that's weird for me all tests are passing:

image

Could you describe the errors that you are finding?

tests/e2e/utils/track-event.js
mikkamp: Ah, thanks for clarifying. I seem to have missed the `url=` prefix you added.

Since we now extract the params does it also mean we can change `eventName === params.get( 'event' )`

mikkamp:
image

Seems the event data isn't submitting the correct product ID.

tests/e2e/utils/track-event.js
jorgemd24: I don't think so because the `event` is not a parameter, it is included in the `data` parameter.

image

We could do something like: `params.get( 'data' ).includes(match)` if you believe it's a slightly more clear compared to the current implementation.

jorgemd24:
It seems related to this: https://github.com/woocommerce/google-listings-and-ads/issues/2095 but it was already fixed here: https://github.com/woocommerce/google-listings-and-ads/pull/2096

tests/e2e/utils/track-event.js
mikkamp: Ok, that makes sense. I do think the last suggestion you made does make it quite explicit what we are expecting to match (which might help with troubleshooting).

jorgemd24: Yep that sounds right. I made the change here: https://github.com/woocommerce/google-listings-and-ads/pull/2122/commits/6557bd19f84999acae74ccc7c4074bf2b51f2a1b

mikkamp (approved):
Thanks for the final changes and pointing to the related PR.
It helps if I rebuild the `gtag-events` script and don't load an older copy. With a fresh copy everything passes.

PR #2123, opened 2023-10-04
Add E2E tests for Setup Google Ads - Step 3

jorgemd24:

Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/2070, this PR adds E2E tests for https://github.com/woocommerce/google-listings-and-ads/issues/2070#gla-dashboard.

- Follow these steps to set up the e2e env: https://github.com/woocommerce/google-listings-and-ads#e2e-testing
- Run `npm run test:e2e -- ./tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js`, the test should pass.
- Or, run `npm run test:e2e-dev -- ./tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js`, check the test in the chromium step by step.
- Run `npm run test:e2e` to make sure all tests are passed.

### Additional details:
- I cherry-pick a commit https://github.com/woocommerce/google-listings-and-ads/pull/2123/commits/65d7479cc5e87105ae783ef76d840268ca506ba0 from https://github.com/woocommerce/google-listings-and-ads/pull/2126 which fixes an issue with the auto-refresh of billing status.
- The auto-refresh occurs every 30 seconds, which means that when we're testing if the billing is set up, the E2E tests might take a bit longer. An alternative approach would be to return to step 2 and then proceed to step 3 to trigger the auto-refresh. However, I believe this wouldn't accurately simulate the real-world scenario.

### Changelog entry

> Dev - E2E - Setup Google Ads Step 3 - Setup billing data.

### DIFF
--- a/tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
+++ b/tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
@@ -16,7 +16,7 @@ import {
 	getFAQPanelTitle,
 	getFAQPanelRow,
 	checkFAQExpandable,
-	checkAdsPopup,
+	checkBillingAdsPopup,
 } from '../../utils/page';
 
 const ADS_ACCOUNTS = [
@@ -388,7 +388,7 @@ test.describe( 'Set up Ads account', () => {
 
 			// eslint-disable-next-line jest/expect-expect
 			test( 'should open a popup when clicking set up billing button', async () => {
-				await checkAdsPopup( page );
+				await checkBillingAdsPopup( page );
 			} );
 		} );
 		test.describe( 'Billing status is approved', () => {
--- a/tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
+++ b/tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
@@ -406,13 +406,20 @@ test.describe( 'Set up Ads account', () => {
 					id: ADS_ACCOUNTS[ 1 ],
 					billing_url: null,
 				} );
+
+				// Simulate a bit of delay when creating the Ads campaign so we have enough time to test the content in the page before the redirect.
+				await page.route(
+					/\/wc\/gla\/ads\/campaigns\b/,
+					async ( route ) => {
+						await new Promise( ( f ) => setTimeout( f, 500 ) );
+						await route.continue();
+					}
+				);
 			} );
 			test( 'It should say that the billing is setup', async () => {
 				//Every 30s the page will check if the billing status is approved and it will trigger the campaign creation.
 				await setupBudgetPage.awaitForBillingStatusRequest();
-				await setupBudgetPage.awaitForCampaignCreationRequest( budget, [
-					'US',
-				] );
+				await setupBudgetPage.mockCampaignCreation( budget, [ 'US' ] );
 
 				await expect(
 					page.getByText(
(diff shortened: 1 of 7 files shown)


tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -358,4 +364,68 @@ test.describe( 'Set up Ads account', () => {
...
+				).toBeVisible();
+			} );
+
+			// eslint-disable-next-line jest/expect-expect
+			test( 'should open a popup when clicking set up billing button', async () => {
+				await checkBillingAdsPopup( page );
+			} );
+		} );
+		test.describe( 'Billing status is approved', () => {
+			test.beforeAll( async () => {
```
eason9487, line 400: Would it be better to also mock requests for a campaign creation?

eason9487 (commented):
Thanks for the work.

Ran E2E tests locally and all tests passed.

The code looks good. Before approving it, I would like to know if the following test case is needed and another question in the code review comment.

> * An alternative approach would be to return to step 2 and then proceed to step 3 to trigger the auto-refresh. However, I believe this wouldn't accurately simulate the real-world scenario.

This could be a separate test case for [another scenario](https://github.com/woocommerce/google-listings-and-ads/blob/ad01b350d267317252b8c9b9b47f50d5bee1d444/js/src/setup-ads/ads-stepper/setup-billing/index.js#L60) with these steps:
1. In step 1 of Setup Google Ads, a merchant connected a Google Ads account that already has an approved billing status.
2. After they proceed to step 3 of Setup Google Ads, the page should show a `<BillingSavedCard>` rather than the `<BillingSetupCard>` with auto-refresh check.
3. They have to click the "Launch paid campaign" button themself to complete Setup Google Ads.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
jorgemd24: My initial goal was to ensure that the request to create the campaign was executed, but I believe it's also beneficial to simulate the response and verify the redirection to the dashboard. I've included some tests to assess the modal after the creation and to validate the redirect URL.

jorgemd24:
Thanks @eason9487 for your time reviewing this PR! 

>Before approving it, I would like to know if the following test case is needed and another question in the code review comment.

I don't believe it's necessary. That was just an alternate way to check if the `<BillingSavedCard>` would appear after setting the billing data, without having to wait for the 30-second interval.


>This could be a separate test case for [another scenario](https://github.com/woocommerce/google-listings-and-ads/blob/ad01b350d267317252b8c9b9b47f50d5bee1d444/js/src/setup-ads/ads-stepper/setup-billing/index.js#L60) with these steps:

>In step 1 of Setup Google Ads, a merchant connected a Google Ads account that already has an approved billing status.
>After they proceed to step 3 of Setup Google Ads, the page should show a <BillingSavedCard> rather than the <BillingSetupCard> with auto-refresh check.
>They have to click the "Launch paid campaign" button themself to complete Setup Google Ads.

Yeah, that's a good idea. I've included that scenario, and this PR is now ready for a second round!

tests/e2e/utils/pages/setup-ads/setup-budget.js
```diff
@@ -145,4 +145,78 @@ export default class SetupBudget extends MockRequests {
...
+	}
+
+	/**
+	 * Mock the campaign creation process.
+	 *
+	 * @param {string} budget The campaign budget.
+	 * @param {Array}  targetLocations The targeted locations.
+	 * @return {Promise<void>}
+	 */
+	async mockCampaignCreation( budget, targetLocations ) {
```
eason9487: This function also mocks the request to signal the completion of Setup Google Ads, this request is only issued in this setup flow. It would be more accurate to include it in its name.

eason9487 (approved):
Thanks for the adjustments. LGTM.

Left a comment about a method naming but it's not a blocker.

PR #2126, opened 2023-10-05
Fix the auto-refresh processing of billing status in the ads onboarding flow

### LINKED ISSUE #2124: Ads - Setup Billing not working when it auto-refreshes
### Describe the bug:

When setting up the billings for the Ads Accounts, the `checkStatusAndCompleteSetup` callback fails because the event is not set.

The error happens in the following line because the event is undefined:

https://github.com/woocommerce/google-listings-and-ads/blob/f0b030444f851db1310ca9acffe9bc3a4f21bb67/js/src/components/adaptive-form/adaptive-form.js#L212

### Steps to reproduce:

1. Add the following filter: 

```
add_filter(
	'woocommerce_gla_ads_billing_setup_status',
	function( $status ) {
		return 'unknown';
	}
);
```

2. Click Add paid campaign
3. Fill all the data from step 1 to step 3.
4. In step 3, you'll notice it prompts you to configure the billing.
5. Change the previous filter to return "approved"
6. Wait a bit and see the error.  

### Video

https://github.com/woocommerce/google-listings-and-ads/assets/2488994/141b08c0-3e89-48e1-95a8-cfe5bb762141

### Expected behavior:
It should allow us to add the paid campaign.

### Actual behavior:
It does not allow us  add the paid campaign.

### Additional details:
I tried changing  `event.currentTarget` to `event?.currentTarget` and it seems that everything works ok however I did not have enough time to see if it will affect other parts of the code or to investigate the root cause of this issue.

eason9487:

Closes #2124 

To fix the issue, the wrapped `handleSubmit` in `AdaptiveForm` component is changed to allow it to be called without a corresponding event. As the `submitter` is unknown in this case, the value is kept as `null` to align with [its type in JSDoc](https://github.com/woocommerce/google-listings-and-ads/blob/60e1782084eb74ff6d7613ea3341136ea8ac3aae/js/src/components/adaptive-form/adaptive-form-context.js#L25).

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/f61a2b3d-3e84-4223-a6cd-99d0f2ab3f2c

Follow the reproduction steps in #2124 to see if the ads onboarding flow can be completed.

### Additional details:

In the demo video, there is another issue that it may create two campaigns after completing ads onboarding. I will fix it by a subsequence PR.

### Changelog entry

> Fix - The auto-refresh processing of billing status in the Google Ads onboarding flow.

### DIFF
--- a/js/src/components/adaptive-form/adaptive-form.js
+++ b/js/src/components/adaptive-form/adaptive-form.js
@@ -209,7 +209,7 @@ function AdaptiveForm( { onSubmit, extendAdapter, children, ...props }, ref ) {
 				// Keep the target element for identifying which one triggered the submission when
 				// there are multiple submit buttons.
 				formContext.handleSubmit = function ( event ) {
-					adapterRef.current.submitter = event.currentTarget;
+					adapterRef.current.submitter = event?.currentTarget || null;
 					return handleSubmit.call( this, event );
 				};
 
--- a/js/src/components/adaptive-form/adaptive-form.test.js
+++ b/js/src/components/adaptive-form/adaptive-form.test.js
@@ -116,16 +116,25 @@ describe( 'AdaptiveForm', () => {
 							<button onClick={ formContext.handleSubmit }>
 								B
 							</button>
+							<button
+								onClick={ () => {
+									// To simulate that `handleSubmit` is called without passing an `event`.
+									formContext.handleSubmit();
+								} }
+							>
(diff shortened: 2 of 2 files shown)


jorgemd24 (approved):
@eason9487 I can confirm that the issue has been fixed and thanks for covering this scenario in the tests. LGTM 👍

PR #2127, opened 2023-10-05
Fix the issue that may create two campaigns after completing ads onboarding

eason9487:

While working on #2126, found another issue that may create two campaigns after completing ads onboarding. The root cause is [the reference changes of `onStatusApproved`](https://github.com/woocommerce/google-listings-and-ads/blob/60e1782084eb74ff6d7613ea3341136ea8ac3aae/js/src/components/paid-ads/billing-card/useAutoCheckBillingStatusEffect.js#L55) in `useAutoCheckBillingStatusEffect` hook, and it leads to calling `onStatusApproved` multiple times before redirecting out of the ads onboarding page.

This PR fixes it by ensuring the `onSetupComplete` callback of `BillingSetupCard` is only called once after getting an approved status for billing.

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/037ca89b-1f95-4f12-966c-5068e2e661c9

1. `git checkout 3dd7efe8b8da3ed7d00b9ff17a559a5a43e0fb16`
2. `npm run env -- wp-scripts test-unit-js billing-setup-card` to see the failure of this test.
3. `git checkout ee841eba98df32ebc50d3661df529f06f3ea0d3c`
4. Rerun the above test to check if the test can pass.
5. `git checkout 3235e1496fc9a0bfd4ff7dc8ebd6506384935f9b`
6. Rerun the above test to see the failure of the second test case.
7. `git checkout fix/finish-onboarding-ads-duplicate-campaigns`
8. Rerun the above test to check if the second test case can pass.
9. Follow the reproduction steps in #2124 to see if it creates only a new campaign after finishing the ads onboarding flow 

### Changelog entry

> Fix - Avoid creating two campaigns after completing the Google Ads onboarding.

### DIFF
--- a/js/src/components/paid-ads/billing-card/billing-setup-card.test.js
+++ b/js/src/components/paid-ads/billing-card/billing-setup-card.test.js
@@ -1,0 +1,96 @@
+/**
+ * External dependencies
+ */
+import { render, act } from '@testing-library/react';
+import apiFetch from '@wordpress/api-fetch';
+
+/**
+ * Internal dependencies
+ */
+import BillingSetupCard from './billing-setup-card';
+import useWindowFocus from '.~/hooks/useWindowFocus';
+
+jest.mock( '.~/hooks/useGoogleAdsAccount', () =>
+	jest.fn().mockName( 'useGoogleAdsAccount' ).mockReturnValue( {} )
+);
+
+jest.mock( '.~/hooks/useDispatchCoreNotices', () =>
+	jest
+		.fn()
+		.mockName( 'useDispatchCoreNotices' )
+		.mockReturnValue( { createNotice: () => {} } )
+);
+
+jest.mock( '.~/hooks/useWindowFocus', () =>
+	jest.fn().mockName( 'useWindowFocus' ).mockReturnValue( true )
+);
+
+jest.mock( '@wordpress/api-fetch', () => {
+	const impl = jest.fn().mockName( '@wordpress/api-fetch' );
+	impl.use = jest.fn().mockName( 'apiFetch.use' );
+	return impl;
+} );
+
+describe( 'BillingSetupCard', () => {
+	let fetchBillingStatus;
+
+	beforeEach( () => {
+		fetchBillingStatus = jest
+			.fn()
+			.mockResolvedValue( { status: 'approved' } );
+
+		apiFetch.mockImplementation( ( { path } ) => {
+			switch ( path ) {
+				case '/wc/gla/ads/accounts':
+					return Promise.resolve( {} );
+
+				case '/wc/gla/ads/billing-status':
+					return fetchBillingStatus();
+
+				default:
+					throw new Error( `No mocked apiFetch for path: ${ path }` );
+			}
+		} );
+	} );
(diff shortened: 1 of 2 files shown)


js/src/components/paid-ads/billing-card/useAutoCheckBillingStatusEffect.js
```diff
@@ -27,19 +27,29 @@ const completeGoogleAdsAccountSetup = () => {
...
+	onStatusApprovedRef.current = onStatusApproved;
 
 	const checkStatusAndCompleteSetup = useCallback( async () => {
+		const prevStatus = prevStatusRef.current;
 		const billingStatus = await apiFetch( {
 			path: '/wc/gla/ads/billing-status',
 		} );
 
-		if ( billingStatus.status !== GOOGLE_ADS_BILLING_STATUS.APPROVED ) {
+		prevStatusRef.current = billingStatus.status;
```
jorgemd24, line 41: Perhaps we could include a brief comment explaining why it's necessary to store the previous status. This would help us to undersand the code in the fure.

js/src/components/paid-ads/billing-card/billing-setup-card.test.js
```diff
@@ -0,0 +1,114 @@
...
+		} );
+
+		expect( fetchBillingStatus ).toHaveBeenCalledTimes( 1 );
+		expect( onSetupComplete ).toHaveBeenCalledTimes( 1 );
+
+		await act( async () => {
+			jest.runOnlyPendingTimers();
+		} );
+
+		expect( fetchBillingStatus ).toHaveBeenCalledTimes( 2 );
```
jorgemd24: I would assume that once the billing status is approved, there's no need to call fetchBillingStatus again since it becomes unnecessary. WDYT?

jorgemd24 (approved):
Thanks @eason9487 for working on this. LGTM 👍 

I left a couple of comments but they are not blockers.

js/src/components/paid-ads/billing-card/useAutoCheckBillingStatusEffect.js
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/2127/commits/93eeba40ca59a10fd200473c558758623c7629ed.

js/src/components/paid-ads/billing-card/billing-setup-card.test.js
eason9487: The `fetchBillingStatus` being called again is the internal implementation logic of `useAutoCheckBillingStatusEffect` hook, and yes, it's not needed for meeting the same business logic. However, the hook naming does not only scope the need for `checkStatusAndCompleteSetup` but also any possible changes of billing status, although there has been no that kind of need from the beginning till now.

From a test case perspective, it's not required to assert it. Removed by https://github.com/woocommerce/google-listings-and-ads/pull/2127/commits/f8e81083cee0b2c8dfacb0dea2652308cfe7a6f9.

PR #2128, opened 2023-10-05
Prevent Prefix Vendor to be added twice

puntope:

This PR fixes a bug when running `prefix-vendor-namespace.php` script on `composer install` when running a second time.

1. Remove vendor folder
2. Run `composer install`
3. Go to `vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php` 
4. See `$options['sink'] = \Automattic\WooCommerce\GoogleListingsAndAds\Vendor\GuzzleHttp\Psr7\Utils::tryFopen('php://temp', 'w+');` as been prefixed.
5. Run again `composer install` (without removing vendor)
6. See line in step 4 double prefixed
7. Checkout this PR
8. Repeat steps 1-5
9. Verify the mentioned line is not double prefixed
10. Search for "\Automattic\WooCommerce\GoogleListingsAndAds\Vendor" to verify everything is fine in terms of replacement.
11. Do a full onboarding to verify everything works

### Additional details:

This only impacts in `prefix_string` function, namespace and uses are not impacted by this bug. 

### Changelog entry

> Dev - Prevent Prefix Vendor to be added twice

### DIFF
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -296,8 +296,10 @@ function prefix_string( &$contents, $search ) {
 	global $namespace_prefix;
 
 	$quoted   = preg_quote( $search, '#' );
+	$namespace_quoted = preg_quote( "\\" . $namespace_prefix . "\\", '#' );
+
 	$contents = preg_replace(
-		"#({$quoted})#m",
+		"#((?<!{$namespace_quoted}){$quoted})#m", // prevent to add the namespace twice
 		"{$namespace_prefix}\\\\\$1",
 		$contents
 	);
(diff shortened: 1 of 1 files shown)


mikkamp (approved):
Thanks for catching that. We used to have a check at the beginning of `process_file` which would prevent double replacements: https://github.com/woocommerce/google-listings-and-ads/blob/97a9a424e026fb467a8cb1a8d2b0ab70712ca702/bin/prefix-vendor-namespace.php#L134-L137

But after we refactored to strict replacing that check wasn't relevant in the same way. The changes here work as expected. I'm no longer able to reproduce double replacements and the extension continues functioning even after multiple calls to `composer install`.

PR #2137, opened 2023-10-13
Add tracking for campaign count

### LINKED ISSUE #2130: Campaign Count Tracking
## User story
For opted in users, the WC Tracker data keeps track of when the MC onboarding has been completed and when the Ads onboarding has been completed.
However it's possible to onboard an Ads account without creating a campaign or create a campaign directly in the Google Ads dashboard. So it would be helpful to include campaign count in the tracking data.

We can add to the tracking data here: https://github.com/woocommerce/google-listings-and-ads/blob/2.5.8/src/Tracking/TrackerSnapshot.php#L87-L101

We can keep the campaign count at zero if an account has not connected yet.

### Account ID
In addition to the campaign count we'd like to also track the connected Ads Account ID.

mikkamp:
This PR adds tracking for campaign count and the ads customer ID.

Since the campaign count is retrieved from the Ads API the result is cached for 12 hours to reduce any delays. Since we also fetch a list of campaigns when viewing the dashboard, we use this opportunity to cache the count so when the tracker snapshot is sent it can use a cached value.

The cached values are cleared upon a disconnect, when a campaign is created, or if the transient expires.

Closes #2130 

1. Test on a site with MC and Ads accounts onboarded 
2. Use a code snippet to output a tracking snapshot to a page:
```php
add_action(
	'admin_init',
	function() {
		if ( ! empty( $_GET['wctracker'] ) ) {
			$tracks = apply_filters( 'woocommerce_tracker_data', [] );
			echo '<pre>' . json_encode( $tracks, JSON_PRETTY_PRINT ) . '</pre>';
			exit;
		}
	}
);
```
3. Visit the page: `https://domain.test/wp-admin/?wctracker=1` (swap for your admin domain)
4. Confirm the snapshot data has the correct values for `ads_customer_id` and `ads_campaign_count`
5. Disconnect the ads account and recheck the snapshot data
6. Confirm the values are both 0
7. Confirm that the transient `_transient_gla_ads_campaign_count` is cleared up (after the disconnect)
8. Reconnect and view the list of campaigns on the dashboard page
9. Check the DB (or transient manager) to confirm that the transient `_transient_gla_ads_campaign_count` has been populated

### Changelog entry
* Tweak - Add tracking for campaign count.

### DIFF
--- a/src/API/Google/AdsCampaign.php
+++ b/src/API/Google/AdsCampaign.php
@@ -14,6 +14,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Google\Ads\GoogleAdsClient;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
 use Google\Ads\GoogleAds\Util\FieldMasks;
 use Google\Ads\GoogleAds\Util\V14\ResourceNames;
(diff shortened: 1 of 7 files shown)


martynmjones (approved):
Hey @mikkamp, thanks for your work to improve tracking.

Tested this branch and confirmed that the number of campaigns is cached and accurately reported when tracking is enabled so LGTM ✅

PR #2144, opened 2023-11-06
Update docs domain name to Woo.com

ianlin:

Update the old WooCommerce.com → the new Woo.com in several places, especially for documentation links.

Project: peeuvX-13n-p2

1. Confirm all new URLs actually go to a page.

### Additional details:

- Did not change the `developer.woocommerce.com` and `api-vipgo.woocommerce.com` URLs
- Added a TODO (87e5872f28866c08d5d19e9123255a553822877d) for updating `developer.woocommerce.com` when the migration is done
- Also changed the anchor [#get-500-in-free-ad-credits](https://woo.com/document/google-listings-and-ads/#get-500-in-free-ad-credits) to [#google-performance-max-campaigns](https://woo.com/document/google-listings-and-ads/#google-performance-max-campaigns) because there is no anchor for the former, and the latter one is the parent section of it.

### Changelog entry

> Doc - Use new Woo.com domain.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -34,7 +34,7 @@ This Code of Conduct applies both within project spaces and in public spaces whe
 
 ## Enforcement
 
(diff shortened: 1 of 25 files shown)


mikkamp (approved):
Thanks for the changes, they look good, I confirmed the links are working and not redirecting to other locations. I don't find any other instances which should be changed.

Issue #2147, opened 2023-11-06
GLA inaccessible when WooCommerce Beta Tester plugin is active


martynmjones:
### Describe the bug:

When the [WooCommerce Beta Tester](https://woo.com/products/woocommerce-beta-tester/) plugin is active on a site, Google Listings & Ads stops working and instead an error is displayed reading `Sorry, you are not allowed to access this page.`

<img width="1007" alt="Screenshot 2023-11-06 at 19 34 57" src="https://github.com/woocommerce/google-listings-and-ads/assets/40762232/7769b1de-8f3c-44c7-b423-1a67305741e3">

### Steps to reproduce:
1. Install GLA and [WooCommerce Beta Tester](https://woo.com/products/woocommerce-beta-tester/)
2. Activate both
3. Go to `Marketing > Google Listings & Ads`

eason9487:
The root cause is the same as https://github.com/woocommerce/google-listings-and-ads/issues/1059#issuecomment-966791638. WooCommerce Beta Tester has [added `'wc-admin-app'` as its dependency since version 7.8.0](https://github.com/woocommerce/woocommerce/blob/7.8.0/plugins/woocommerce-beta-tester/woocommerce-beta-tester.php#L91).

Given that this is the second time the same issue has happened, even modifying the WooCommerce Beta Tester won't completely avoid it in the future.

I believe any other extension does the same thing may cause it again, so this needs to be investigated further to see if there is a better way to ensure that GLA is loaded before `'wc-admin-app'`, or that the [`addFilter` callback](https://github.com/woocommerce/google-listings-and-ads/blob/2.5.11/js/src/index.js#L55-L172) is invoked correctly even if GLA is loaded after `'wc-admin-app'`.

Ref: https://github.com/woocommerce/woocommerce/pull/38113

eason9487:
## Possible solutions

After exploring this issue further,  I suppose there are currently three ways to (temporarily) solve it.

### 1. Quick fix from the extension side

Set the `$priority` for [the `admin_enqueue_scripts` action](https://github.com/woocommerce/google-listings-and-ads/blob/2.7.0/src/Admin/Admin.php#L77-L90) to be less than the default of `10`, so that it can be output and run earlier than the `'wc-admin-app'` script.

This solution may not work if any script that depends on `'wc-admin-app'` is enqueued with a higher priority than this extension.

### 2. Quick fix from the WooCommerce Core side

[...]

PR #2151, opened 2023-11-10
Product Block Editor: Add the fundamental support and map `Text` and `Integer` attribute inputs to currently available generic blocks

eason9487:

To be compatible with Product Block Editor, this PR adds fundamental support:

- Extract the processing of getting the visible and hidden product types in `AttributesForm` as a method.
- Convert the `init_input` method in `AttributesForm`  to a static method.
- Add methods to `Admin\Input\Input` for getting block config.
- Add a new class `AttributesBlock` to register blocks to Product Block Editor for product attributes.

Also, it turns a few attribute inputs to support Product Block Editor first:

- Map `Text` and `Integer` inputs to currently available generic blocks.
- Add the block attribute 'min' to the **multipack** input.

Last, it adds some PHP tests and removes a few unused PHP classes or method:

- Remove the unused method `is_custom_value` from `SelectWithTextInput`.
- Add PHP unit tests for
   - the `Admin\Input\Input` class and its inheritance classes.
   - the attribute input classes under `Admin\Product\Attributes\Input`.
   - the `AttributesForm` class.
   - the `AttributesBlock` class.

#### 📷 Simple product

image

#### 📷 Variation product

image

#### 📷 The `min` attribute of  the multipack input

image

#### 📌 <a href="#prepare-test-environment" id="prepare-test-environment">Prepare test environment</a>

1. Set up [WooCommerce 8.3.0](https://github.com/woocommerce/woocommerce/releases/tag/8.3.0)
2. Install [WooCommerce Beta Tester](https://woo.com/products/woocommerce-beta-tester/)
3. Go to WooCommerce > Settings > Advanced > Features. Enable the **New product editor**
   
4. Go to Tools > WCA Test Helper > Features. Make sure the **product-block-editor** and **product-variation-management** options are enabled.
   
5. Deactivate WooCommerce Beta Tester as it's incompatible with GLA for now. See #2147

#### 📌 Test the simple product type

1. Go to edit a simple product, e.g. the **Beanie with Logo** imported from WC's sample products
2. Switch to the **Organization** tab and find the **Attributes** section.
3. Check if there is the **Google Listings & Ads** section under the **Attributes** section.
4. Within the **Google Listings & Ads** section, operate the text and number fields and change their values.
5. Click on the **Update** button at the top-right of the page.
6. Refresh the page or switch back to the classical product editor to see if the changed values are saved.

#### 📌 Test the variable and variation product type

1. Go to edit a variable product, e.g. the **Hoodie** imported from WC's sample products
2. Switch to the **Organization** tab and find the **Attributes** section.
3. Check if there is the **Google Listings & Ads** section under the **Attributes** section.
4. Within the **Google Listings & Ads** section, there should be only a brand field for now. Operate the brand field and change its value.
5. Click on the **Update** button at the top-right of the page.
6. Refresh the page or switch back to the classical product editor to see if the changed value is saved.
7. Back to edit the same variable product with Product Block Editor.
8. Switch to the **Variations** tab and find the **Variations** section.
9. Click on one of the Edit links within the **Variations** section to go to edit a variation product.
10. With the **General** tab, find the **Image** section.
11. Check if there is the **Google Listings & Ads** section under the **Image** section.
12. Repeat the similar test to verify the fields for this variation product.

#### 📌 Test the related filters

1. Continue to use the above variation product editing page.
2. Filters the list of applicable product types for each attribute - `"woocommerce_gla_attribute_applicable_product_types_{$attribute_id}"`
   1. Confirm there is no **Brand** field.
   2. Add a filter to the **google-listings-and-ads.php** file:
      ```php
      add_filter(
      	"woocommerce_gla_attribute_applicable_product_types_brand",
      	function ( array $applicable_types ) {
      		$applicable_types[] = 'variation';
      
      		return $applicable_types;
      	}
      );
      ```
   3. Refresh the page.
   4. Check if the **Brand** field appears.
3. Filters the list of product types to hide for each attribute - `"woocommerce_gla_attribute_hidden_product_types_{$attribute_id}"`
   1. Confirm there is the **Size** field.
   2. Add a filter to the **google-listings-and-ads.php** file:
      ```php
      add_filter(
      	"woocommerce_gla_attribute_hidden_product_types_size",
      	function ( array $applicable_types ) {
      		$applicable_types[] = 'variation';
      
      		return $applicable_types;
      	}
      );
      
      ```
   3. Refresh the page.
   4. Check if the **Size** field disappears.

### Additional details:

The generic number block (`'woocommerce/product-number-field'`) doesn't support verifying/limiting for entering an integer. If it's not supported afterward, then we may have to convert it to a custom block.


src/Admin/Product/Attributes/AttributesBlock.php
```diff
@@ -0,0 +1,200 @@
...
+		);
+	}
+
+	/**
+	 * Add this extension's section block after the given reference section block.
+	 *
+	 * @param BlockInterface $reference_section The reference section block to add this extension's section block
+	 */
+	private function add_section( BlockInterface $reference_section ): BlockInterface {
+		$group = $reference_section->get_parent();
```
jorgemd24, line 102: Would be nice to have the type reference here: `/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Group */`

src/Admin/Product/Attributes/AttributesBlock.php
```diff
@@ -0,0 +1,200 @@
...
+			}
+		);
+	}
+
+	/**
+	 * Add this extension's section block after the given reference section block.
+	 *
+	 * @param BlockInterface $reference_section The reference section block to add this extension's section block
+	 */
+	private function add_section( BlockInterface $reference_section ): BlockInterface {
```
jorgemd24, line 101: The return type should be: `SectionInterface` 

https://github.com/woocommerce/woocommerce/blob/633bf40f13263cd80c6d803b44c2b0d40422b7e1/plugins/woocommerce/src/Internal/Features/ProductBlockEditor/ProductTemplates/Group.php#L48

src/Admin/Product/Attributes/AttributesBlock.php
```diff
@@ -0,0 +1,200 @@
...
+
+		add_action(
+			"woocommerce_block_template_area_{$template_area}_after_add_block_{$block_id}",
+			function ( BlockInterface $attributes_section ) {
+				// Please note that the simple and variable product types use the same product block template 'simple-product'.
+				if ( 'simple-product' !== $attributes_section->get_root_template()->get_id() ) {
+					return;
+				}
+
+				$section = $this->add_section( $attributes_section );
```
jorgemd24, line 75: Initially this part confused me, as it seems that the method `add_section` would add the `$attributes_section`. However, it turns out the 'add' method retrieves the group from this attribute section and subsequently adds a new section. I'm thinking, perhaps renaming this function to something like 'add_section_to_group' could better clarify its purpose.

src/Admin/Product/Attributes/AttributesBlock.php
```diff
@@ -0,0 +1,200 @@
...
+				$this->add_blocks( $section );
```
jorgemd24, line 76: Similar to my earlier comment, this could be `add_blocks_to_section`

src/Admin/Product/Attributes/AttributesBlock.php
```diff
@@ -0,0 +1,200 @@
...
+		// https://github.com/woocommerce/woocommerce/blob/8.3.0/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/ProductTemplates/AbstractProductFormTemplate.php#L16
+		$template_area = 'product-form';
+
+		// https://github.com/woocommerce/woocommerce/blob/8.3.0/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/ProductTemplates/SimpleProductTemplate.php#L361
+		$block_id = 'product-attributes-section';
+
+		add_action(
+			"woocommerce_block_template_area_{$template_area}_after_add_block_{$block_id}",
+			function ( BlockInterface $attributes_section ) {
+				// Please note that the simple and variable product types use the same product block template 'simple-product'.
```
jorgemd24, line 70: Do we want to keep the type generic using `BlockInterface` ? Or can we use `Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Section`?

src/Admin/Product/Attributes/AttributesForm.php
```diff
@@ -91,13 +79,44 @@ public function get_view_data(): array {
 		return $view_data;
 	}
 
+	/**
+	 * Get the hidden and visible types of an attribute's applicable product types.
+	 *
+	 * @param string $attribute_type An attribute class extending AttributeInterface
```
jorgemd24: Maybe An attribute classname ....

src/Admin/Product/Attributes/AttributesForm.php
```diff
@@ -91,13 +79,44 @@ public function get_view_data(): array {
...
+		 * This filter is documented in AttributeManager::map_attribute_types
+		 *
+		 * @see AttributeManager::map_attribute_types
+		 */
+		$applicable_product_types = apply_filters( "woocommerce_gla_attribute_applicable_product_types_{$attribute_id}", $applicable_product_types, $attribute_type );
+
+		/**
+		 * Filters the list of product types to hide the attribute for.
+		 */
+		$hidden_product_types = apply_filters( "woocommerce_gla_attribute_hidden_product_types_{$attribute_id}", [] );
```
jorgemd24, line 103: If I use the filter in the following way I get an error:

```
add_filter(
	'woocommerce_gla_attribute_hidden_product_types_gtin',
	function ( array $applicable_types ) {
		return [ 'simple', 'variation' ];
	}
);
```

image

jorgemd24 (commented):
Thanks @eason9487 for working on this, I ran some tests with simple products, and things went smoothly, except for the issue I mentioned in the comment about encountering an error with the filter.

Regarding the variation you mentioned:

> Within the Google Listings & Ads section, there should be only a brand field for now. Operate the brand field and change its value.

All I'm seeing is a blank section.

image

The filters didn't seem to affect anything when I tested them with the variations.

I ran out of time today, but I am happy to debug this issue deeper if you are not able to replicate it.

src/Admin/Product/Attributes/AttributesBlock.php
eason9487: This line will be removed by PR #2179. I added related type references to that PR in https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/ac4abe9ff386b535481d720cb193631ddfbc2af8.

src/Admin/Product/Attributes/AttributesBlock.php
eason9487: Thanks! This method will be removed by [this commit in PR #2179](https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/5cf49b154ac52a6adb4f0d8a842ad9adb5117d24#diff-2b96a65cd604d35139576ee14972bfb745506fe4a781d276e9c721d20a18fa59L167). To avoid future code conflicts, I would skip this adjustment.

src/Admin/Product/Attributes/AttributesBlock.php
eason9487: Thanks for the suggestion.

This method call will be inlined by [this commit in PR #2179](https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/5cf49b154ac52a6adb4f0d8a842ad9adb5117d24#diff-2b96a65cd604d35139576ee14972bfb745506fe4a781d276e9c721d20a18fa59R110-R128). Instead of using `$this` to indirectly add group and section to template, the changed code fragment now calls to the literal callee, which I believe should not cause similar confusion as here.

src/Admin/Product/Attributes/AttributesBlock.php
eason9487: [In this commit of PR #2179](https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/5cf49b154ac52a6adb4f0d8a842ad9adb5117d24#diff-2b96a65cd604d35139576ee14972bfb745506fe4a781d276e9c721d20a18fa59R130), this line will be changed to `$this->add_product_attribute_blocks( $product_attributes_section );`, which should be able to clarify the purpose that it adds product attribute blocks to a section.

src/Admin/Product/Attributes/AttributesBlock.php
eason9487: In [the definition of this hook](https://github.com/woocommerce/woocommerce/blob/8.3.0/plugins/woocommerce/src/Internal/Admin/BlockTemplates/BlockContainerTrait.php#L255-L267), there are several possible incoming instances, and each of them is a class that implements `BlockInterface`.

image

Although in runtime it passes in a `Section` instance, I think it also makes sense to use the more generic type `BlockInterface` here. Like when I adjusted their insertion position in #2179, the type reference here didn't need to be changed.

src/Admin/Product/Attributes/AttributesForm.php
eason9487: Sorry, I don't understand this comment. Could you explain a bit more?

src/Admin/Product/Attributes/AttributesForm.php
eason9487: Thanks for catching this! Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/2151/files/0408038b4607fb9e4779f33b38848cf2920398ef..466d83e5d874c1b7cd1daeba45aedaadbca5f8fa.

eason9487:
Hi @jorgemd24, thanks for the review.

> Regarding the variation you mentioned:
> 
> > Within the Google Listings & Ads section, there should be only a brand field for now. Operate the brand field and change its value.
> 
> All I'm seeing is a blank section.
> [...]
> The filters didn't seem to affect anything when I tested them with the variations.

Could you check if WooCommerce Brands is activated? If it is, the `Brand` attribute input will be dynamically converted to use `SelectWithTextInput`, which is not yet converted in this PR. Attribute inputs that are not yet converted will be skipped, so the GL&A section in a variable product will be blank.

https://github.com/woocommerce/google-listings-and-ads/blob/46830dce8d762c8cd1f1068917be291a86415d83/src/Admin/Product/Attributes/AttributesBlock.php#L130-L133

If this is not the cause, then I can't reproduce it, so I may need your help to find the cause.

src/Admin/Product/Attributes/AttributesBlock.php
jorgemd24: Thanks, yes that will remove the ambiguity!

src/Admin/Product/Attributes/AttributesForm.php
jorgemd24: My bad, I should have clarified this better 😅. What I meant was in the PHP doc, when we mention 'An attribute class,' strictly speaking, it refers to the actual name of that class. So, my suggestion was more along the lines of asking whether we should update this to: 'The name of the Attribute class extending AttributeInterface.'

jorgemd24 (approved):
Thanks @eason9487,  for fixing the issue and your comments. The reason why the brands field is not displayed was because the SelectInput doesn't have a block name, therefore the code is skipping this field. For this test, I added the blockname `'woocommerce/product-text-field'` and I could test the behaviour. I see that you already work on this here https://github.com/woocommerce/google-listings-and-ads/pull/2178/files#diff-eb72915300551a54c7e5237203c3878b9c0bcedb253bb0d0624e45890e7b so I am approving this PR as everything works nos smoothly.

Thanks again!

src/Admin/Product/Attributes/AttributesForm.php
eason9487: Thanks for the suggestion! Updated all related PHP docs for the `AttributesForm` class by https://github.com/woocommerce/google-listings-and-ads/pull/2151/commits/9ff6a970a6bb6a102e2976e840fafe24ae92c027.

Issue #2146, opened 2023-11-06
Google API Project


jorgemd24:
## User story

The full details of the project can be found in the following links:

pbIy4N-3Lx-p2  <br>pcTzPl-1PW-p2<br>pcTzPl-1PW-p2

## Shared branch for development

[feature/google-api-project](<https://github.com/woocommerce/google-listings-and-ads/tree/feature/google-api-project>)

### WPCOM - GLA

- [X] Allow to connect WPCOM App without Jetpack (the plugin) **Stage 1**
- [ ] Once `automattic/jetpack-connection` is updated and includes all the functions required to connect the WPCOM App, update the package and remove the duplicated code from the previous step. \*\* This is depending on Jetpack \*\*
- [X] Add WC REST API filters to transform the data before sending the response.  **Stage 1**
- [X] Restrict API access if the channel visibility is `dont-sync-and-show`, for products and coupons. **Stage 1**
- [X] Restrict API access if coupons are allowed for specific emails. **Stage 1**
- [X] Expose Attribute values. **Stage 1**
- [X] Expose `gla_target_audience` option in `wc/v3/settings/general` **Stage 1**
- [X] Expose shipping and the rest of the onboarding data in `wc/v3/settings/general` **Stage 1**
- [X] Expose `offer_id` in deleting notification products.
- [X] Add Endpoint to get partner integration status  **Stage 1**
- [X] Add Endpoint to revoke WPCOM access  **Stage 1**
- [X] Add Event tracking (`gla_enable_product_sync_click` & `gla_disable_product_sync_click`) **Stage 1** - [#2424](<https://github.com/woocommerce/google-listings-and-ads/issues/2424>)

### Granting/Onboarding flow.

- [X] Decide the onboarding flow for the new connections. **Stage 1**
- [X] Create Auth URL **Stage 1**
- [X] Grant Access to the App in the Settings Page  **Stage 1**
- [X] Add a section to the Connection Test page to retrieve the integration status. **Stage 1**
- [X] Add a section to the Connection Test page to revoke the WPCOM access. **Stage 1**

### Notifications

- [X] Notify Google when a product is created/updated/deleted. **Stage 1**
- [X]  Notify Google when a coupon is created/updated/deleted. **Stage 1**
- [X]  Notify Google when a shipping zone is created/updated/deleted. **Stage 1**
- [X] Validate merchant center setup when triggering notifications **Stage 1**

### Migration

[...]

PR #2152, opened 2023-11-10
Enable filtering of WC REST API responses based on the gla_syncable query parameter

jorgemd24:

Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/2146

This PR allows the filtering of private metadata (metadata prefixed with `_`), excluding metadata with the prefix `_wc_gla`, when a WC REST API request is made using a specific parameter, `gla_syncable`.

Additionally, this PR ensures that only the products the merchant intends to sync are returned. In this case, the required metadata setting is: `_wc_gla_visibility` = `sync-and-show`.

The WC REST API endpoints that need to be filter are the following:

| Endpoint  | Filters |
| ------------- | ------------- |
| `/wp-json/wc/v3/products`  | private metadata + _wc_gla_visibility = sync-and-show  |
| `/wp-json/wc/v3/products/<product_id>`l  | private metadata + _wc_gla_visibility = sync-and-show  |
| `/wp-json/wc/v3/products/<product_id>/variations`  | private metadata  |
| `/wp-json/wc/v3/products/<product_id>/variations/<variation_id>`l  | private metadata  |
| `/wp-json/wc/v3/coupons`l  | private metadata + _wc_gla_visibility = sync-and-show + email restrictions |
| `/wp-json/wc/v3/coupons/<coupon_id>`l  | private metadata + _wc_gla_visibility = sync-and-show + email restrictions  |

Filtering is unnecessary for shipping zone endpoints as they do not contain metadata.

- See here for more context: pcTzPl-1Qd-p2#comment-2663
- Here a list of all endponts: pcTzPl-1Qd-p2#comment-2650

1. Make requests to the endpoints mentioned above and add the `gla_syncable` param, for example: `/wp-json/wc/v3/products?gla_syncable=1`
2. See that the data is filtered correctly.
3. If you remove the query param: `gla_syncable` it should not filter any data. 

### Additional details:

- Variants are special cases since they don't rely on `_wc_gla_visibility` and instead depend on the parent attribute. Hence, if we aim to avoid fetching a variant that the parent product doesn't have the `gla_visibility`, we'd have to check its parent, meaning an additional query. I'm unsure if it's necessary to go to that far, so I'll seek feedback in the project thread.
- Coupons with email restrictions should also be filtered:

image

### Changelog entry

> Add - Filter WC REST API responses for gla_syncable param


src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,209 @@
...
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Integration
+ */
+class WPCOMProxy implements Service, Registerable {
+
+	/**
+	 * The Post types to be filtered.
+	 *
+	 * @var array
+	 */
+	public static $post_types_to_be_filter = [
```
martynmjones: Should this variable be called `$post_types_to_filter`?

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+
+	/**
+	 * Register the filters for a specific object type.
+	 *
+	 * @param string $object_type The object type.
+	 */
+	protected function register_object_types_filter( string $object_type ): void {
+		add_filter( 'woocommerce_rest_prepare_' . $object_type . '_object', [ $this, 'check_item_is_syncable' ], 9, 3 );
```
martynmjones: For consistency it would be nice if this was written over multiple lines as the two filters added below are.

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,209 @@
...
+
+	/**
+	 * Whether the data should be filtered.
+	 *
+	 * @param WP_REST_Request $request The request object.
+	 *
+	 * @return bool
+	 */
+	protected function should_filter_data( WP_REST_Request $request ): bool {
+		return $request->get_param( 'gla_syncable' ) !== null;
```
martynmjones: I appreciate this won't be widely used, however, I'm curious why we're only relying on the presence of the parameter and not paying attention to the value?

If we send a request to both:

- `/wp-json/wc/v3/products?gla_syncable=true`
- `/wp-json/wc/v3/products?gla_syncable=false`

I would expect different results but at the moment we're going to get the same filtered results for both.

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,209 @@
...
+			],
+			403
+		);
+	}
+
+	/**
+	 * Query items with specific args for example where _wc_gla_visibility is set to sync-and-show.
+	 *
+	 * @param array           $args The query args.
+	 * @param WP_REST_Request $request The request object.
```
martynmjones: The description for `$args` and `$request` are not aligned here.

martynmjones (commented):
Hey @jorgemd24, thanks for your work adding this filter!

It's working as expected but I've left a few comments asking for clarification and on a few formatting points.

puntope:
> Additionally, this PR ensures that only the products the merchant intends to sync are returned. In this case, the required metadata setting is: _wc_gla_visibility = sync-and-show.

I don't see the synced products when I call `/wp-json/wc/v3/products`

Steps:

- Fresh installation
- Do full onboarding 
- Checkout PR
- Call `/wp-json/wc/v3/products?gla_syncable=1` 

Result:

Empty array 

- Go to the product 
- Save
- Call `/wp-json/wc/v3/products?gla_syncable=1` 

Result: 
Product shows


https://github.com/woocommerce/google-listings-and-ads/assets/5908855/cd2f2fce-1135-4f94-b1b3-fdc3ed8eb2a4

When I do Syn all products with Connection Tests it works as well.

src/Integration/WPCOMProxy.php
jorgemd24: Initially, my intention was only to check the presence because we include the parameter only for the GL&A app, making the value irrelevant. However, I understand your point that you would expect different results with the above requests so I adjusted the code. I didn't use "true" or "false" because in WPCOM the boolean gets sent as "1" or "0"

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,215 @@
...
+
+	/**
+	 * Register the filters for a specific object type.
+	 *
+	 * @param string $object_type The object type.
+	 */
+	protected function register_object_types_filter( string $object_type ): void {
+		add_filter(
+			'woocommerce_rest_prepare_' . $object_type . '_object',
+			[ $this, 'check_item_is_syncable' ],
```
puntope: I would recommend to add a consistent `filter_` name to `check_item_is_syncable` as it's also a filter

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,215 @@
...
+	/**
+	 * Check if a single item is syncable.
+	 *
+	 * @param WP_REST_Response $response The response object.
+	 * @param mixed            $item     The item.
+	 * @param WP_REST_Request  $request  The request object.
+	 *
+	 * @return WP_REST_Response The response object updated.
+	 */
+	public function check_item_is_syncable( $response, $item, WP_REST_Request $request ): WP_REST_Response {
```
puntope: More than just a check (boolean return) I find this more as a filtering of the response. 

Maybe "filter_response_by_syncable_item" ?? Just an idea, the point is that check asses a true/false statement

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,215 @@
...
+			return $response;
+		}
+
+		foreach ( $data['meta_data'] as $key => $meta ) {
+			if ( str_starts_with( $meta->key, '_' ) && ! str_starts_with( $meta->key, '_wc_gla' ) ) {
+				unset( $data['meta_data'][ $key ] );
+			}
+		}
+
+		$data['meta_data'] = array_values( $data['meta_data'] );
```
puntope: Why we are just returning the values?

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,215 @@
...
+			return $args;
+		}
+
+		$args['meta_query'] = [ ...$args['meta_query'] ?? [], ...$post_type_filters['meta_query'] ];
+
+		return $args;
+	}
+
+	/**
+	 * Filter the metadata of an object.
```
puntope: I think it would be more accurate something like:

"Filter the response metadata returning only those prefixed with _wc_gla"

src/Integration/WPCOMProxy.php
```diff
@@ -0,0 +1,215 @@
...
+			'products',
+			'coupons',
+		];
+		preg_match( $pattern, $route, $matches );
+
+		if ( ! isset( $matches['id'] ) || ! isset( $matches['resource'] ) || ! in_array( $matches['resource'], $protected_resources, true ) ) {
+			return $response;
+		}
+
+		$meta_data = $response->get_data()['meta_data'] ?? [];
```
puntope, line 139: I would  return $response in case `$meta_data === []` because otherwise `$meta->key `will throw a fatal if it happens.

src/Integration/WPCOMProxy.php
jorgemd24: Thanks for catching this, I believe it was there because I used `array_filter` to return values and I wanted to reset the indexes. However, it's no longer necessary. I removed here: https://github.com/woocommerce/google-listings-and-ads/pull/2152/commits/5ec640eed1f815a9e83cbfe0a4832b6e9627bce1

jorgemd24:
@martynmjones Thanks for your time reviewing this PR! I addressed all your comments and I replied the question about the `gla_syncable` value.

@puntope 

> I don't see the synced products when I call /wp-json/wc/v3/products

> Steps:

> Fresh installation
> Do full onboarding
Checkout PR
Call /wp-json/wc/v3/products?gla_syncable=1

Is it because the `_wc_gla_visibility` metadata hasn't been added to the products yet? Perhaps the AS takes some time to update the metadata for those products. If that's the case, it's something we could include in the Migration Project.

While investigating the above, I realised that we could use the below constants, so I made changes to the code:

https://github.com/woocommerce/google-listings-and-ads/blob/b84999bfefe337177b41f5bd028c2bdc182042b5/src/Value/ChannelVisibility.php#L17-L18

puntope (approved):
I tested again this PR and I get the expected results in all the endpoints (products & coupons)

Also, no more inconsistencies from my side in the code. I approve this, so if @martynmjones  doesn't have any issue to address this can be merged.

martynmjones:
> Also, no more inconsistencies from my side in the code. I approve this, so if @martynmjones doesn't have any issue to address this can be merged.

No issues for me. Thanks for the changes @jorgemd24

PR #2153, opened 2023-11-13
Tweak Budget recommendations logic

puntope:

This PR changes the logic in budget recommendations.

- Adapt Budget Recommendation Controller for using `daily_budget` (single column) instead of a range
- Adapt Budget Recommendation Frontend for using `daily_budget` (single column) instead of a range
- Migrate the database tables
- Adapt tests

⚠️  The Budget Recommendation File is not final. We are waiting for a final version with all the currencies. YOu an use the one provided as a sample.

⚠️  The tracking requisite is not implemented in this PR

Possible setup for testing:

- All countries - USD
- Spain and France - Euro

### Screenshots:

Screenshot 2023-11-13 at 11 26 48

1. Change the Applicable version to `2.5.11` in `src/DB/Migration/Migration20231109T1653383133.php`
2. Go to the Database and change `gla_file_version` and `gla_db_version` to `2.5.10` in `wp_options` table to refresh the data and migrate the columns. 
3. Re-activate the plugin.
4. Verify that the sample Budget Recommendations has been loaded in `wp_gla_budget_recommendations ` and contains. the right values and columns 
5. Go to the Onboarding, select Spain and France as the Audience, and Euro as the currency
6. COmplete the setup and create an Ads Campaign
7. In the budget notice see the recommended budget as the highest of both countries in EUR 
8. Introduce a budget less than this value and see the message "With a lower budget than your competitor..."
9. Change the budget to equal or more than the suggested one
10. See the message mentioned in step 8 disappear.

### Additional details:

PT --> pcTzPl-1Uo-p2

### Changelog entry

> Tweak - Use a single daily budget instead of a range


puntope:
cc: @MatthiasReinholz  this is a partial PR with the logic as we are waiting still for Google Budget Final Recommendations

src/DB/Table.php
```diff
@@ -127,6 +127,28 @@ protected function get_collation(): string {
...
+	 *
+	 * @return bool True if the column exists on the table or False if not.
+	 *
+	 * @since x.x.x
+	 */
+	public function has_column( string $column_name ): bool {
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
+		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		$this->wpdb->get_results(
+			$this->wpdb->prepare( "SHOW COLUMNS FROM `{$this->get_sql_safe_name()}` LIKE %s", [ '%' . $this->wpdb->esc_like( $column_name ) . '%' ] )
```
jorgemd24: Why do we check if the column name contains `$column_name` rather than checking if the table has exactly that column name? For instance, if we have a table with columns named `daily_budget_low` and `daily_budget_high`, and we want to confirm the existence of the column `daily_budget`, this function would return `true` which is incorrect.

src/Internal/DependencyManagement/DBServiceProvider.php
```diff
@@ -100,6 +101,7 @@ public function register() {
 		$this->share_migration( MigrationVersion141::class, MerchantIssueTable::class );
 		$this->share_migration( Migration20211228T1640692399::class, ShippingRateTable::class, OptionsInterface::class );
 		$this->share_with_tags( Migration20220524T1653383133::class, BudgetRecommendationTable::class );
+		$this->share_migration( Migration20231109T1653383133::class, BudgetRecommendationTable::class );
```
jorgemd24, line 104: Do you think that it's necessary to keep the previous Migration20220524T1653383133?

src/DB/Migration/Migration20231109T1653383133.php
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Apply the migrations.
+	 *
+	 * @return void
+	 */
+	public function apply(): void {
+		if ( $this->budget_rate_table->exists() && $this->budget_rate_table->has_column( 'daily_budget_low' ) ) {
+			$this->wpdb->query( "ALTER TABLE `{$this->wpdb->_escape( $this->budget_rate_table->get_name() )}` RENAME COLUMN `daily_budget_low` TO `daily_budget`" );    // phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+
+		}
```
jorgemd24: This actually does not work. When the table is installed using [dbDelta](https://github.com/woocommerce/google-listings-and-ads/blob/f7b95dc90cf28a40f90f5f1b82c5ebb92351e11e/src/DB/Table.php#L45), it updates the table, which means the table already has a column named `daily_budget` at that point. Therefore, the renaming process will fail, and the `daily_budget_low` will still exist in the table.

jorgemd24 (commented):
Thanks @puntope for updating the budgets, I get the expected results however I've left a few comments about the code that I think need to be addressed.

src/DB/Table.php
puntope: Thanks for catching this, I updated the method so I use a strict where clause

src/DB/Migration/Migration20231109T1653383133.php
puntope: Thanks for this catch, I update the apply() method so I drop the old coulmns.

src/Internal/DependencyManagement/DBServiceProvider.php
puntope: I prefer to keep it for the history or for the other versions. Seems not harmful to keep it-

puntope:
Hi @jorgemd24 Thanks for the review. I applied the changes and this is ready for a new round.

src/DB/Migration/Migration20231109T1653383133.php
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+
+	/**
+	 * Returns the version to apply this migration for.
+	 *
+	 * @return string A version number. For example: 1.4.1
+	 */
+	public function get_applicable_version(): string {
+		return '2.5.11';
```
jorgemd24: We should include a heads up in the release spreadsheet for the next porter  so it updates it with the right version

data/budget-recommendations.csv
```diff
@@ -1,4231 +1,279 @@
...
-ZAR,TN,15,25
-ZAR,LB,15,40
-ZAR,IN,15,60
-ZAR,PY,15,45
-ZAR,UZ,15,25
-ZAR,EC,15,45
-ZAR,MM,15,20
-ZAR,GE,15,20
\ No newline at end of file
+Country,Budget,Currency
```
jorgemd24, line 1: I see that the updated  CSV (pcTzPl-1Uo-p2#comment-2733) with the correct currency is now available, with the columns: `budget_local_currency` and `budget_usd`. Are you planning to remove one of the columns in the CSV, or will you be adjusting the logic here?

jorgemd24 (commented):
Hi @puntope, I have just left a comment regards the new CSV format, I am not sure how you will handle it. Apart from that, everything is working as expected.

puntope:
Hi @jorgemd24 thanks again for your review and valuable comments. I made some tweaks in the column name of the new file and loaded it so now the values are with the right column names and file format.

See the discussion here pcTzPl-1Uo-p2#comment-2741 (if they require changes I will do so in a new PR)

I also changed the testing version for a placeholder 'x.x.x' in the migration file. Once merged I will add the Warning in the Release Spreadsheet.

jorgemd24 (approved):
Thanks @puntope for the changes! LGTM!

PR #2155, opened 2023-11-14
Remove unused flush_rewrite_rules

### LINKED ISSUE #2154: Flushing rewrite rules
### Describe the bug:
The rewrite rules get flushed upon plugin activation and plugin deactivation.

When plugins get activated with `wp --skip-plugins --skip-themes "$WP_CMD_KIND" "activate" "$NAME"` (like is done on Atomic).
GLA will get activated without a dependency like WooCommerce. This will flush the rewrite rules, WITHOUT the WooCommerce related ones, ending up with products returning a 404 and an empty shop page.

In theory this shouldn't be the case because upon activation we check dependencies with the [plugin validator](https://github.com/woocommerce/google-listings-and-ads/blob/2.5.11/src/Infrastructure/GoogleListingsAndAdsPlugin.php#L51-L56), and if that fails we delay activation till a time we get a request where the plugin validator succeeds. Which should in theory mean WooCommerce is active when we flush the rewrite rules.

However it seems there are still random circumstances which are being observed where the rewrite rules are being flushed without the WooCommerce ones being included.

Since we don't need to flush the rewrite rules as we aren't registering any custom pages/endpoints we decided to remove this code. See discussion: p1699456066588239-slack-C046L2P9BEK

### Steps to reproduce:
As described above this occurs randomly so we don't have exact steps to reproduce.

mikkamp:

This PR removes the `flush_rewrite_rules` as we are not registering any custom pages/endpoints that require the rules to be flushed. See details in #2154 on why it's problematic.

Closes #2154 

1. Install a plugin like https://wordpress.org/plugins/rewrite-rules-inspector/ to inspect the registered rewrite rules
2. Activate and deactivate GLA (before testing this branch) and compare the rewrite rules to make sure there are no changes when flushing upon activation/deactivation (copy output from Tools > Rewrite Rules to a file and use a `diff` tool to compare changes)
3. Install this PR and test activating / deactivating the extension
4. Ensure there are no errors / warnings (rewrite rules should again remain the same)

### Changelog entry
* Tweak - Remove rewrite rules flush.

### DIFF
--- a/src/Infrastructure/GoogleListingsAndAdsPlugin.php
+++ b/src/Infrastructure/GoogleListingsAndAdsPlugin.php
@@ -62,8 +62,6 @@ final class GoogleListingsAndAdsPlugin implements Plugin {
 				$service->activate();
 			}
 		}
-
-		flush_rewrite_rules();
 	}
 
 	/**
@@ -79,8 +77,6 @@ final class GoogleListingsAndAdsPlugin implements Plugin {
 				$service->deactivate();
 			}
 		}
-
-		flush_rewrite_rules();
 	}
 
 	/**


martynmjones (approved):
Hey @mikkamp, thanks for addressing the activation issue on Atomic/ WooExpress.

Tested the branch locally and am not able to find any difference between the stored rewrite rules before/ after activation and reviewing the codebase I can't find any potential problems with removing `flush_rewrite_rules` so LGTM ✅

PR #2156, opened 2023-11-15
Add node ^18 and npm ^9 as supported engines for dependabot

eason9487:

Add node ^18 and npm ^9 as supported engines to facilitate dependabot upgrades npm package by opening PRs.

This cannot be tested from our side. The adjustment was made mainly according to the message on the Dependabot alerts page. A few alerts were unable to update npm packages due to:

```
Dependabot uses Node.js v18.18.2
 and NPM 9.6.5
. Due to the engine-strict setting, the update will not succeed.
```


mikkamp (approved):
Looks good to me. For AW we were a bit stricter with version: https://github.com/woocommerce/automatewoo/pull/1563
And we also did a followup PR to explain why we added the exception in the readme. Should we do the same here?

eason9487:
@mikkamp, thanks for the review and suggestion.

> And we also did a followup PR to explain why we added the exception in the readme. Should we do the same here?

Added in https://github.com/woocommerce/google-listings-and-ads/pull/2156/commits/8912f4a887425316bc164e3c6454791a362e746d.

PR #2161, opened 2023-11-21
Fix E2E gtag events for a blockified confirmation page.

jorgemd24:

~~Currently, there's a fatal error due to this issue: https://github.com/woocommerce/woocommerce-blocks/issues/11851 and a "blockified" confirmation page. This problem is caused because the hook `woocommerce_before_thankyou` is being called with incorrect parameters.~~

Additionally, our E2E tests are failing. Check here: https://github.com/woocommerce/google-listings-and-ads/actions/runs/6942346039/job/18885109983?pr=2160. The blockified confirmation page uses a different class for the order status.

This PR ~~introduces a temporary solution for the mentioned issue~~ and resolves the E2E test problem.

~~Fatal Error in the confirmation page.~~ This has been fixed in WC 8.3.1 p1700588910958219-slack-C01DT6U03HC
~~1.  Use WC 8.3~~
~~2. Use a theme that supports blocks, for example twenty twenty-two.~~
~~3. Blockify your confirmation page. See instructions here: https://woo.com/document/cart-checkout-blocks-status/ (Search for "Replacing the order confirmation page" )~~
~~4. Buy a product and complete the order, you shouldn't get any fatal error.~~
~~5. Checkout develop and repeat the steps and you will get a fatal error.~~

### E2E Tests

1. Check that E2E are ✅ . 

### Additional details:
- I added `await page.waitForTimeout( 3000 );` because it was clicking the button too quickly making the tests fail.

### Changelog entry

> Fix - E2E gtag evens tests.

### DIFF
--- a/tests/e2e/utils/customer.js
+++ b/tests/e2e/utils/customer.js
@@ -100,6 +100,9 @@ export async function checkout( page ) {
 		await page.locator( '#billing-state input' ).fill( user.statename );
 	}
 
+	//TODO: See if there's an alternative method to click the button without relying on waitForTimeout.
+	await page.waitForTimeout( 3000 );
+
 	await page.locator( 'text=Place order' ).click();
 
 	await expect(
--- a/tests/e2e/utils/customer.js
+++ b/tests/e2e/utils/customer.js
@@ -106,6 +106,6 @@ export async function checkout( page ) {
 	await page.locator( 'text=Place order' ).click();
 
 	await expect(
-		page.locator( '.woocommerce-thankyou-order-received' )
+		page.locator( '.wc-block-order-confirmation-status' )
 	).toContainText( 'order has been received' );
 }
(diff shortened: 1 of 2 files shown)


tests/e2e/utils/customer.js
```diff
@@ -100,9 +100,12 @@ export async function checkout( page ) {
 		await page.locator( '#billing-state input' ).fill( user.statename );
 	}
 
+	//TODO: See if there's an alternative method to click the button without relying on waitForTimeout.
+	await page.waitForTimeout( 3000 );
```
eason9487, line 104: I commented out this line and ran the related tests several times, and all attempts can pass. 

> * I added `await page.waitForTimeout( 3000 );` because it was clicking the button too quickly making the tests fail.

Maybe an alternative method is to wait for the button to be enabled:

```suggestion
	await expect( page.locator( 'text=Place order' ) ).toBeEnabled();
```

eason9487 (approved):
Thanks for fixing the E2E tests. LGTM.

Left an alternative method to avoid clicking the button too quickly and it's not a blocker for this PR.

tests/e2e/utils/customer.js
jorgemd24: Thanks for the suggestion! Locally the E2E tests failed for me, even with the suggestion that you mentioned. For now, I will keep it as it is. If I find the root cause I will change it.

PR #2164, opened 2023-11-28
Product Block Editor: Add the fundamental setup for the custom product blocks and map `Select` and `BooleanSelect` inputs to the custom select field block

eason9487:

To be compatible with Product Block Editor, this PR adds fundamental support:

- Adjust the `can_enqueue` of `BaseAsset` as it should allow `$enqueue_condition_callback` to be null.
- Adjust the `AttributesBlock` to register the custom blocks and the related assets.
- Adjust the frontend config to build the custom blocks.
- Add the structure of custom blocks.
- Update the `@woocommerce/dependency-extraction-webpack-plugin` package to get the support of the newer WC packages.

To show how to develop and use the custom product blocks, this PR:

- Implement the custom block of the product select field.
- Map `Select` and `BooleanSelect` inputs to the custom block of the select field.

#### 📷 Simple product

image

#### 📷 Variable product

image

#### 📷 Variation product

image

1. Please refer to [the test preparation in PR 2151](https://github.com/woocommerce/google-listings-and-ads/pull/2151#prepare-test-environment).
1. `npm install`
1. `npm start`

#### 📌 Test the simple product type

1. Go to edit a simple product, e.g. the **Beanie with Logo** imported from WC's sample products
1. Switch to the **Organization** tab and find the **Google Listings & Ads** section.
1. Operate the selection fields and change their selected option.
1. Click on the **Update** button at the top-right of the page.
1. Refresh the page or switch back to the classical product editor to see if the changed options are saved.

#### 📌 Test the variable and variation product type

1. Go to edit a variable product, e.g. the **Hoodie** imported from WC's sample products
1. Switch to the **Organization** tab and find the **Google Listings & Ads** section.
1. Within the **Google Listings & Ads** section, there should be two fields. Operate the selection field and change its selected option.
1. Click on the **Update** button at the top-right of the page.
1. Refresh the page or switch back to the classical product editor to see if the changed option is saved.
1. Back to edit the same variable product with Product Block Editor.
1. Switch to the **Variations** tab and find the **Variations** section.
1. Click on one of the Edit links within the **Variations** section to go to edit a variation product.
1. With the **General** tab, find the **Google Listings & Ads** section.
1. Repeat the similar test to verify the fields for this variation product.

#### 📌 Test the related filters

1. Go to edit a simple product.
1. Filters the options for each attribute - `"woocommerce_gla_product_attribute_value_options_{$attribute_id}"`
   1. Confirm the **Condition** field has four options: Default, New, Refurbished, and Used.
   1. Add a filter to the **google-listings-and-ads.php** file:
      ```php
      add_filter(
      	'woocommerce_gla_product_attribute_value_options_condition',
      	function ( array $value_options ) {
      		$value_options[ 'broken' ] = 'Broken';
      		return $value_options;
      	}
      );
      ```
   1. Refresh the page.
   1. Check if the **Condition** has the fifth option: Broken.
      image

### Additional details:

With WooCommerce 8.3.0, the tooltip is not aligned well, and the generic product blocks don't support the `tooltip` attribute.

image

These issues will be resolved in WooCommerce 8.4.0, so this PR won't take care of them.

image


js/src/blocks/README.md
```diff
@@ -0,0 +1,74 @@
...
+### Directory structure of source code
+
+```
+/js/src/blocks/            # The root directory of custom blocks
+├── product-select-field/  # A custom block
+│   ├── block.json         # The metadata file of this block and also the canonical way to register this block with both PHP and JavaScript side
+│   └── edit.js            # The component to work with the `edit` function when registering this block, and it's the interface for how this block is going to be rendered within the Product Block Editor
+├── another-field/         # Another custom block
+│   ├── block.json
+│   └── edit.js
```
puntope, line 20: 💅  Thanks for documenting the structure. I'd suggest adding the Custom Blocks in their folder (ie custom_blocks, fields... ) so they are not at the same level as the components folder.

src/Assets/BaseAsset.php
```diff
@@ -130,6 +130,10 @@ public function get_uri(): string {
 	 * @return bool
 	 */
 	public function can_enqueue(): bool {
+		if ( is_null( $this->enqueue_condition_callback ) ) {
```
puntope, line 133: 💅  Can we cover this by test unit as well?

tests/data/existing-block/block.json
```diff
@@ -0,0 +1 @@
+{}
```
puntope, line 1: ❓ Why is this file for?

js/src/blocks/components/label.js
```diff
@@ -0,0 +1,35 @@
+// Copied from WooCommerce core and simplified to leave only the needed parts.
```
puntope, line 1: 💅  I assume it will be done in future PRs. I expected to have some tests for the JS part as well

js/src/blocks/product-select-field/block.json
```diff
@@ -0,0 +1,37 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 2,
+	"name": "google-listings-and-ads/product-select-field",
+	"title": "Product select control",
+	"textdomain": "google-listings-and-ads",
+	"attributes": {
+		"_templateBlockHideConditions": {
+			"type": "array"
+		},
```
puntope, line 10: ❓  Why is this for? When I delete it I don't see any difference at first look

js/src/blocks/product-select-field/block.json
```diff
@@ -0,0 +1,37 @@
...
+		"label": {
+			"type": "string"
+		},
+		"tooltip": {
+			"type": "string"
+		},
+		"property": {
```
puntope, line 17: ❓  Same for these "property" and "context". I don't see clear references out there but seems to be an essential part of the config.

puntope (approved):
✅  LGTM. Thanks for working on this. 

- I tested changing values in Products Single, Variable and Variations. They worked and they persisted after refreshing. 
- I tested adding custom values in the fields and also updating the values after.

In regards the code I didn't see any blocker. I left some 💅  and some questions. But as they are not blockers I approve this in advance.

js/src/blocks/README.md
eason9487: Since at most there will only be 3 or 4 custom blocks, I lead toward not using a deeper directory structure.

src/Assets/BaseAsset.php
eason9487: Added in https://github.com/woocommerce/google-listings-and-ads/pull/2164/commits/e4a743e4e7717b111df589d4401921daed53cd49.

tests/data/existing-block/block.json
eason9487: For testing the registration of custom blocks without actually building client files.

https://github.com/woocommerce/google-listings-and-ads/blob/2eaee14e70377efa1164ff02b376b670d2737511/tests/Unit/Admin/Product/Attributes/AttributesBlockTest.php#L210-L235

js/src/blocks/components/label.js
eason9487: Only using jest to test this component or custom blocks would be fragile as they involve a lot of Woo core mechanisms and codes. I planned to test them via E2E tests after all required development settled down.

js/src/blocks/product-select-field/block.json
eason9487: `__experimentalRole` is an undocumented attribute property, and is commonly set in WooCommerce core, but it seems a bit paradoxical. Although I did find [the use of its source code](https://github.com/WordPress/gutenberg/blob/125c130818ebb83c50618722858e8315d90e76b8/packages/blocks/src/store/selectors.js#L796-L815), it's not clear what it's used for, only knew that: without this property, it may lead to a custom block in editing-disabled mode.

As the product block editor in Woo core is still under development, this issue may be/has been resolved. However, it would be safer to keep it for the time being.

js/src/blocks/product-select-field/block.json
eason9487: The `property` value will be passed into the corresponding block and be used with the below `useProductEntityProp` to access a property or meta of product data. The naming is to align with all generic product blocks in Woo core, e.g. [number](https://github.com/woocommerce/woocommerce/blob/8.3.0/packages/js/product-editor/src/blocks/generic/number/block.json#L15), [checkbox](https://github.com/woocommerce/woocommerce/blob/8.3.0/packages/js/product-editor/src/blocks/generic/checkbox/block.json#L18), [radio](https://github.com/woocommerce/woocommerce/blob/8.3.0/packages/js/product-editor/src/blocks/generic/radio/block.json#L17) and so on.

https://github.com/woocommerce/google-listings-and-ads/blob/2eaee14e70377efa1164ff02b376b670d2737511/js/src/blocks/product-select-field/edit.js#L15-L17

Not sure if the "context" refers to `useContext`. If yes, I added its use by [this commit in PR #2178](https://github.com/woocommerce/google-listings-and-ads/pull/2178/commits/57e1128a06d05b078075e166ae99c699163226c9#diff-a274946a182787bcb5c604c81e92a8901771ec0aaaf10e7444b2f1208ee84503R39-R44)

js/src/blocks/product-select-field/block.json
puntope: Thanks for the explanation. I was referring to 

```
"_templateBlockHideConditions": {
			"type": "array"
		},
```

eason9487: Sorry, I got it wrong.

In order to make a custom block compatible with [the conditionally hiding feature](https://github.com/woocommerce/woocommerce/blob/8.4.0/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/ProductTemplates/README.md#conditionally-hiding-a-block-in-product-editor-templates), it requires the `_templateBlockHideConditions` attribute when registering the custom block on both PHP and JS sides.

On JS side, a custom block can use [registerProductEditorBlockType](https://github.com/woocommerce/woocommerce/blob/8.4.0/packages/js/product-editor/src/utils/register-product-editor-block-type.ts#L58) to [inject that attribute](https://github.com/woocommerce/woocommerce/blob/8.4.0/packages/js/block-templates/src/utils/register-woo-block-type.ts#L92-L123), but on PHP side, there is no corresponding way to do so. Therefore, a custom block still needs to do a similar injection as [BlockRegistry::register_block](https://github.com/woocommerce/woocommerce/blob/8.4.0/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/BlockRegistry.php#L210-L213) when calling Gutenberg `register_block_type`.

The good news is Woo 8.5.0 will have the core support by [PR 41973 - Template API: Registration of custom block types](https://github.com/woocommerce/woocommerce/pull/41937), so the `_templateBlockHideConditions` attribute will be removed in a subsequent PR.

PR #2179, opened 2023-12-12
Product Block Editor: Preprocessing for adding the channel visibility

eason9487:

To be compatible with Product Block Editor, it also needs to add the channel visibility block.

Since product attributes and channel visibility are different groupings in terms of code and UI presentation, it would be better to separate them into different sections in the Product Block Editor as well. Therefore, before adding the block, it requires some preprocessing:

- Move and rename `Admin\Product\Attributes\AttributesBlock` class to `Admin\ProductBlocksService` class.
   - This way, when the block corresponding to `ChannelVisibilityMetaBox` is added to it afterward, the code structure should make more sense.
- Change to create a dedicated group in the Product Block Editor for this extension.
- Add a new section to this extension's group in the Product Block Editor for channel visibility setup.
- Clean up the temporary codes that prevent unmapped inputs from being added to the Product Block Editor.

#### 📷 Simple product

image

#### 📷 Variable product

image

#### 📷 Variation product

image

1. Please refer to [the test preparation in PR 2151](https://github.com/woocommerce/google-listings-and-ads/pull/2151#prepare-test-environment) with **Woo 8.4.0**.
   - Currently, it can be tested with [8.4.0-rc.1](https://github.com/woocommerce/woocommerce/releases/tag/8.4.0-rc.1)
1. `npm install`
1. `npm start`

#### 📌 Test the dedicated group for GL&A

1. Check if there is a **Google Listings & Ads** tab, which is that dedicated group.
1. Check if there are **Channel visibility** and **Product attributes** sections.
   - Currently, the Channel visibility section is empty.
   - The Product attributes section should have attribute fields that correspond to the current product's type.
1. Repeat the above test steps with variable product and variation product types.

### DIFF
--- a/src/Admin/ProductBlocksService.php
+++ b/src/Admin/ProductBlocksService.php
@@ -100,6 +100,7 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 		add_action(
 			"woocommerce_block_template_area_{$template_area}_after_add_block_{$block_id}",
 			function ( BlockInterface $general_group ) {
+				/** @var Automattic\WooCommerce\Admin\Features\ProductBlockEditor\ProductTemplates\ProductFormTemplateInterface */
 				$template = $general_group->get_root_template();
 
 				// Please note that the simple and variable product types use the same product block template 'simple-product'.
@@ -107,6 +108,7 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 					return;
 				}
 
+				/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Group */
 				$group = $template->add_group(
 					[
 						'id'         => 'google-listings-and-ads-group',
@@ -117,6 +119,7 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 					]
 				);
 
+				/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Section */
 				$channel_visibility_section = $group->add_section(
 					[
 						'id'         => 'google-listings-and-ads-channel-visibility-section',
@@ -127,6 +130,7 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 					]
 				);
 
+				/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Section */
 				$product_attributes_section = $group->add_section(
 					[
 						'id'         => 'google-listings-and-ads-product-attributes-section',
@@ -203,6 +207,7 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 				// When editing a simple or variable product, its product type on the frontend side can be
 				// changed dynamically. So, it needs to use the ProductTemplates API `add_hide_condition`
 				// to conditionally hide attributes.
+				/** @var BlockInterface */
 				$block = $section->add_block( $input->get_block_config() );
 				$block->add_hide_condition( $this->get_hide_condition( $attribute_type ) );
 			}
--- a/src/Admin/ProductBlocksService.php
+++ b/src/Admin/ProductBlocksService.php
@@ -191,11 +191,6 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 			$input_type = call_user_func( [ $attribute_type, 'get_input_type' ] );
 			$input      = AttributesForm::init_input( new $input_type(), new $attribute_type() );
 
-			// TODO: Remove this check after all attribute inputs have specified a block name
-			if ( '' === $input->get_block_config()['blockName'] ) {
-				continue;
-			}
-
 			if ( $is_variation_template ) {
 				// When editing a variation, its product type on the frontend side won't be changed dynamically.
 				// In addition, the property of `editedProduct.type` doesn't exist in the variation product.
--- a/src/Admin/ProductBlocksService.php
+++ b/src/Admin/ProductBlocksService.php
@@ -17,6 +17,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\Attributes\AttributeManager;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\BuiltScriptDependencyArray;
 use Automattic\WooCommerce\Admin\BlockTemplates\BlockInterface;
+use Automattic\WooCommerce\Admin\Features\ProductBlockEditor\ProductTemplates\SectionInterface;
 use Automattic\WooCommerce\Admin\PageController;
 
 defined( 'ABSPATH' ) || exit;
@@ -120,7 +121,7 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 					]
 				);
 
-				/** @var Automattic\WooCommerce\Admin\Features\ProductBlockEditor\ProductTemplates\SectionInterface */
+				/** @var SectionInterface */
 				$channel_visibility_section = $group->add_section(
 					[
 						'id'         => 'google-listings-and-ads-channel-visibility-section',
@@ -131,7 +132,7 @@ class ProductBlocksService implements Service, Registerable, Conditional {
 					]
 				);
 
-				/** @var Automattic\WooCommerce\Admin\Features\ProductBlockEditor\ProductTemplates\SectionInterface */
+				/** @var SectionInterface */
 				$product_attributes_section = $group->add_section(
 					[
 						'id'         => 'google-listings-and-ads-product-attributes-section',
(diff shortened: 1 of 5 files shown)


puntope:
Not introduced in this PR, but since the Service is a Conditional service, what do you think of moving the conditionals from `register()` to the `is_needed()` function?

```php
		return is_admin() && version_compare( WC_VERSION, '8.5', '>=' ) && $this->merchant_center->is_setup_complete();
	}
```

src/Admin/ProductBlocksService.php
```diff
@@ -91,34 +93,55 @@ function () {
...
 		$template_area = 'product-form';
 
-		// https://github.com/woocommerce/woocommerce/blob/8.3.0/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/ProductTemplates/SimpleProductTemplate.php#L361
-		$block_id = 'product-attributes-section';
+		// https://github.com/woocommerce/woocommerce/blob/8.3.0/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/ProductTemplates/SimpleProductTemplate.php#L18
+		// https://github.com/woocommerce/woocommerce/blob/8.3.0/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/ProductTemplates/ProductVariationTemplate.php#L18
+		$block_id = 'general';
 
 		add_action(
 			"woocommerce_block_template_area_{$template_area}_after_add_block_{$block_id}",
```
puntope, line 103: Also not introduced in this PR but seems that we can just do:

`woocommerce_block_template_area_product-attributes-section_after_add_block_general`

src/Admin/ProductBlocksService.php
```diff
@@ -91,34 +93,55 @@ function () {
...
+						'id'         => 'google-listings-and-ads-group',
+						'order'      => 100,
+						'attributes' => [
+							'title' => __( 'Google Listings & Ads', 'google-listings-and-ads' ),
+						],
+					]
+				);
+
+				/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Section */
+				$channel_visibility_section = $group->add_section(
```
puntope, line 125: I think it should be `SectionInterface`, however, the variable `$channel_visibility_section` is not being used so we can omit it.

src/Admin/ProductBlocksService.php
```diff
@@ -91,34 +93,55 @@ function () {
...
+			function ( BlockInterface $general_group ) {
+				/** @var Automattic\WooCommerce\Admin\Features\ProductBlockEditor\ProductTemplates\ProductFormTemplateInterface */
+				$template = $general_group->get_root_template();
 
-		add_action(
-			"woocommerce_block_template_area_{$template_area}_after_add_block_{$block_id}",
-			function ( BlockInterface $images_section ) {
-				if ( ! $this->is_variation_template( $images_section ) ) {
+				// Please note that the simple and variable product types use the same product block template 'simple-product'.
+				if ( 'simple-product' !== $template->get_id() && ! $this->is_variation_template( $general_group ) ) {
```
puntope, line 107: 💅  Maybe we can warp as well `'simple-product' !== $template->get_id()` into `$this->is_simple_template( $general_group )`

src/Admin/ProductBlocksService.php
```diff
@@ -91,34 +93,55 @@ function () {
...
 					return;
 				}
 
-				$section = $this->add_section( $images_section );
-				$this->add_blocks( $section );
+				/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Group */
```
puntope: Shouldn't be `GroupInterface` instead of `Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Group`

src/Admin/ProductBlocksService.php
```diff
@@ -91,34 +93,55 @@ function () {
...
+					[
+						'id'         => 'google-listings-and-ads-group',
+						'order'      => 100,
+						'attributes' => [
+							'title' => __( 'Google Listings & Ads', 'google-listings-and-ads' ),
+						],
+					]
+				);
+
+				/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Section */
```
puntope, line 122: Shouldn't this be `SectionInterface`

src/Admin/ProductBlocksService.php
```diff
@@ -91,34 +93,55 @@ function () {
...
+					[
+						'id'         => 'google-listings-and-ads-channel-visibility-section',
+						'order'      => 1,
+						'attributes' => [
+							'title' => __( 'Channel visibility', 'google-listings-and-ads' ),
+						],
+					]
+				);
+
+				/** @var Automattic\WooCommerce\Internal\Admin\Features\ProductBlockEditor\ProductTemplates\Section */
```
puntope: I think it should be `SectionInterface`

puntope:
I see you added some hints for the IDE `@var BlockInterface` for example. 

They seem unneeded for me as knowing that is coming from BlocmInterface the IDE is smart enough to know whats type of these variables.

src/Admin/ProductBlocksService.php
eason9487: The use of PHP string interpolation here is to indicate that the action name is dynamically composited in Woo core.

src/Admin/ProductBlocksService.php
eason9487: Changed in https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/666cc9dcef530943e109981f5647af95e9a1a942.

> however, the variable `$channel_visibility_section` is not being used so we can omit it.

Sorry, the `$channel_visibility_section` will be used to add the channel visibility block later, and I should have added it in a subsequent PR but didn't notice. Since it was known the variable was going to be used, I didn't delete it.

src/Admin/ProductBlocksService.php
eason9487: Changed in https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/666cc9dcef530943e109981f5647af95e9a1a942.

src/Admin/ProductBlocksService.php
eason9487: Changed in https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/666cc9dcef530943e109981f5647af95e9a1a942.

eason9487:
Thanks for the code review, @puntope! This PR is ready for the new round. Could you help with it?

> Not introduced in this PR, but since the Service is a Conditional service, what do you think of moving the conditionals from `register()` to the `is_needed()` function?
> 
> ```
> public static function is_needed(): bool {
> 		return is_admin() && version_compare( WC_VERSION, '8.5', '>=' ) && $this->merchant_center->is_setup_complete();
> 	}
> ```

The `is_needed` method in the `Conditional` interface is static, so the `$this->merchant_center` won't be available. Only moving the parts that can be run within the static method doesn't seem to help much. Therefore, I prefer to keep the current way.

> I see you added some hints for the IDE `@var BlockInterface` for example.
> 
> They seem unneeded for me as knowing that is coming from BlocmInterface the IDE is smart enough to know whats type of these variables.

It came from this code review suggestion: https://github.com/woocommerce/google-listings-and-ads/pull/2151#discussion_r1424353087.

src/Admin/ProductBlocksService.php
puntope: Thanks for the changes, in regards `Automattic\WooCommerce\Admin\Features\ProductBlockEditor\ProductTemplates\SectionInterface` which is used twice.

maybe we can do `use Automattic\WooCommerce\Admin\Features\ProductBlockEditor\ProductTemplates\SectionInterface;` 

and then only do `SectionInterface` without writing all the path.

puntope:
> The `is_needed` method in the `Conditional` interface is static, so the `$this->merchant_center` won't be available. Only moving the parts that can be run within the static method doesn't seem to help much. Therefore, I prefer to keep the current way.
> 
> >

You are right, didn't notice it was static.

puntope:
> It came from this code review suggestion: https://github.com/woocommerce/google-listings-and-ads/pull/2151#discussion_r1424353087.

At least for my ID is not necessary, bu ok to keep it if for others is.

puntope (approved):
LGTM ✅  Thanks for the changes @eason9487 

I left a couple of comments but approved in advance

src/Admin/ProductBlocksService.php
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2179/commits/34ca9e8c11055feb9f41f055376c51a69591d91f. I also adjusted the `add_product_attribute_blocks` method as the `add_block` method is defined in `SectionInterface` rather than `BlockInterface`.

Issue #2182, opened 2023-12-15
PHP 8.3 Compatibility


mikkamp:
### Describe the issue:
The unit tests can be run with PHP 8.3 which confirms that the code (at least the part that's covered by unit tests) within the extension is compatible.

However one of the composer packages isn't fully compatible and we get the following deprecation notice:
```
PHP Deprecated:  Calling get_class() without arguments is deprecated in google-listings-and-ads/vendor/google/apiclient/src/Http/REST.php on line 58
```

There is already an opened issue for this and we are just waiting for the following PR to be resolved: https://github.com/googleapis/google-api-php-client/pull/2509

### Steps to reproduce:
1. Run the extension on a site with PHP 8.3 and WP debug logging on
2. View a page that sends a request to the Content API, like the Product Statuses page
3. Observe the deprecation notice in debug.log or on screen if configured to show notices there

### Expected behavior:
No deprecation messages.

PR #2183, opened 2023-12-15
Include PHP 8.3 in tested versions for PHPunit

mikkamp:

This PR changes the tested versions for PHPunit.

I switched it to running the default unit tests on PHP 8.2 and then included an additional test for the lowest (7.4) supported and highest (8.3) supported versions.

> [!Note]
> Although the unit tests run successfully with PHP 8.3 we are still waiting on #2182 to be resolved before we can mark the extension as fully compatible.

1. Check the workflow runs in this PR and confirm it has a successful unit test for PHP versions 7.4, 8.2, 8.3

### Changelog entry
* Dev - Include PHP 8.3 in tested versions for PHPunit.


puntope (approved):
✅  LGTM

PR #2184, opened 2023-12-15
Fix item price in purchase event

### LINKED ISSUE #2181: item price is incorrect in purchase event
### Steps to reproduce:

Assuming that there is a connection between GA4 and WooCommerce already.

1. Open GA4 Debug View.
2. Place an order with one (or multiple products). Each product has quantity > 1.
3. Complete the order. 
4. in GA4 Debug View, a purchase event should appear, and we can check the item price

### Expected behavior:

the price of one unit of each product.

### Actual behavior:

the returned price is subtotal, which is multiplied by the quantity already.

### Additional details:

By using subtotal for item price, the GA4 revenue report will be wrong. Because the revenue = item price of one unit * quantity.

jorgemd24:

Closes https://github.com/woocommerce/google-listings-and-ads/issues/2181 .

The purchase event on the Order confirmation page was displaying the wrong price for each item. Instead of showing the individual item's price, it was reflecting the total amount for that line.

For example, if we have an item priced at 3€ and we purchase 2 units:

#### Before:

image

#### After:

image

1. Open an incognito window.
1. Purchase a product and add more than 1 unit to the cart.
2. Finish the order.
3. Use the Google Analytics debugger extension, GA4 Debug View, or inspect the code on the Order confirmation page.
4. Check that the purchase event -> items -> price aligns with the product price.

### Additional details:

@mikkamp suggested that a nice way to test these scenarios by including `wp_print_inline_script_tag` into the [WP Proxy](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Proxies/WP.php) class, enabling us to test various outcomes. I'll create a follow-up issue to focus on this.

### Changelog entry

> Fix - item price in purchase event.

### DIFF
--- a/src/Google/GlobalSiteTag.php
+++ b/src/Google/GlobalSiteTag.php
@@ -302,7 +302,7 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 			$product_id   = $item->get_product_id();
 			$product_name = $item->get_name();
 			$quantity     = $item->get_quantity();
-			$price        = $item->get_subtotal();
+			$price        = $order->get_item_total( $item );
 			$item_info [] = sprintf(
 				'{
 				id: "gla_%s",


puntope (approved):
Thx @jorgemd24 LGTM

I tested adding 4 units and it showed only the price per unit.

Screenshot 2023-12-15 at 16 06 31

Also, I tested with decimals as I saw there was a round param in that new function.

Screenshot 2023-12-15 at 16 12 11

PR #2190, opened 2023-12-27
API Pull - Notify when product changes

puntope:

Closes part of #2146

This PR adds logic for calling the Notification Partner endpoint when a Product changes. 

In detail this PR:

- Add a Service NotificationsService. This service includes the topics, the API call to the partner notifications service in WPCOM, and a public method to notify() to perform the call.
- Tweak the hooks in SyncerHooks to call the notify() method 
- Logs when an error happens in the API Call 
- Tests for NotificationsService and tweak tests for SyncerHooks

1. Checkout this PR
2. Be sure you are proxied in WPCOM and the branch you are is `trunk`
3. Authorize your APP and change the partner ID in the WPCOM `sites-woocommerce-partner-notifications.php` in the PARTNERS variable (this is because we don't have yet the final Partner APP)   
4. Create a product
5. Go to WooCommerce - Status - Logs  -- and see the error log
6. Update the product
7. Go to WooCommerce - Status - Logs  -- and see the error log
8. Delete the product 
9. Go to WooCommerce - Status - Logs  -- and see the error log

### Additional details:

- The Topics are not matching the topics in WPCOM Notification PArtners, this is due the new documentation for the endpoint and it's pending to update them in the WPCOM side. So in this PR we only need to test the call and also is a good chance to test the errors.
- I removed  `woocommerce_delete_product_variation` and `woocommerce_trash_product_variation` as they are duplicated with `trashed_post` and `deleted_post`
- After rebase PHPCS triggered some issues in `src/Infrastructure/GoogleListingsAndAdsPlugin.php` but they are not introduced here. So they are out Scope this PR

### Changelog entry

> Add - Notify when product changes


src/Google/NotificationsService.php
```diff
@@ -0,0 +1,114 @@
...
+
+	// List of Topics to be used.
+	public const TOPIC_PRODUCT_CREATED   = 'product.created';
+	public const TOPIC_PRODUCT_DELETED   = 'product.deleted';
+	public const TOPIC_PRODUCT_UPDATED   = 'product.updated';
+	public const  TOPIC_COUPON_CREATED   = 'coupon.created';
+	public const  TOPIC_COUPON_DELETED   = 'coupon.deleted';
+	public const  TOPIC_COUPON_UPDATED   = 'coupon.updated';
+	public const  TOPIC_SHIPPING_SAVED   = 'action.woocommerce_after_shipping_zone_object_save';
+	public const  TOPIC_SHIPPING_DELETED = 'action.woocommerce_delete_shipping_zone';
```
jorgemd24: Constants are misaligned.

src/Google/NotificationsService.php
```diff
@@ -0,0 +1,114 @@
...
+
+		if ( is_wp_error( $response ) ) {
+			$this->notification_error( $item_id, $topic, $response->get_error_message() );
+			return false;
+		}
+
+		if ( wp_remote_retrieve_response_code( $response ) >= 400 ) {
+			$this->notification_error( $item_id, $topic, wp_remote_retrieve_body( $response ) );
+			return false;
+		}
```
jorgemd24: What about:

```
if(is_wp_error( $response ) || wp_remote_retrieve_response_code( $response ) >= 400){
    $error = is_wp_error( $response ) ? $response->get_error_message() : wp_remote_retrieve_body( $response );
    $this->notification_error( $item_id, $topic, $error );
    return false;
}
```

src/Google/NotificationsService.php
```diff
@@ -0,0 +1,114 @@
...
+class NotificationsService implements Service {
+
+	// List of Topics to be used.
+	public const TOPIC_PRODUCT_CREATED   = 'product.created';
+	public const TOPIC_PRODUCT_DELETED   = 'product.deleted';
+	public const TOPIC_PRODUCT_UPDATED   = 'product.updated';
+	public const  TOPIC_COUPON_CREATED   = 'coupon.created';
+	public const  TOPIC_COUPON_DELETED   = 'coupon.deleted';
+	public const  TOPIC_COUPON_UPDATED   = 'coupon.updated';
+	public const  TOPIC_SHIPPING_SAVED   = 'action.woocommerce_after_shipping_zone_object_save';
```
jorgemd24: Just curious, why the topics for shipping zones are so different from the rest?

src/Product/SyncerHooks.php
```diff
@@ -82,39 +88,45 @@ class SyncerHooks implements Service, Registerable {
...
+		$this->notifications_service = $notifications_service;
+		$this->wc                    = $wc;
 	}
 
 	/**
 	 * Register a service.
 	 */
 	public function register(): void {
 		// only register the hooks if Merchant Center is connected and ready for syncing data.
+		// TODO: Potentially change this after API Pull is implemented as we don't need MC to be connected for the API Pull
```
jorgemd24, line 116: I think we will still need the MC to be connected, but instead of pushing the data, they will pull the data.

tests/Unit/Google/NotificationsServiceTest.php
```diff
@@ -0,0 +1,102 @@
...
+			$this->assertEquals( did_action( 'woocommerce_gla_error' ), 1 );
+		}
+
+		/**
+		 * Mocks the service
+		 * @return NotificationsService
+		 */
+		public function get_mock() {
+			return $this->getMockBuilder( NotificationsService::class)
+				->setMethods( [ 'do_request' ] )
```
jorgemd24: `setMethods` seems deprecated.

image

jorgemd24 (commented):
Thanks, @puntope, for tackling this. It's running smoothly. I've dropped a few comments on the code and below are some doubts that I have:

I've noticed that when updating a variation, you receive the topic "product.updated" along with the variation's ID. I don't think it's an issue as you can call the WC REST API wc/v3/products/variation_id without explicitly stating it's a variation ID.

However, when creating a product, even as just a draft without any data, it triggers the notification endpoint. Is this expected behavior? It appears that with the pushing mechanism, we were bypassing these scenarios:

`
2023-12-28T13:23:04+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\WCProductAdapter::map_product_categories Product category (ID: 933): ["Uncategorized"].
2023-12-28T13:23:04+00:00 DEBUG Automattic\WooCommerce\GoogleListingsAndAds\Product\BatchProductHelper::validate_and_generate_update_request_entries Skipping product (ID: 953) because it does not pass validation: {"woocommerce_id":953,"errors":["[description] This value should not be blank.","[imageLink] This value should not be blank.","[gtin] This value is not valid."]}
2023-12-28T13:23:06+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\Google\NotificationsService::notification_error Error sending notification for Item ID 1020 with topic product.created. {"code":"wpcom_partner_notifications_proxy_error","message":"Invalid x-woocommerce-topic","data":{"status":400}}`


Another question that comes to my mind is, will we be notifying every product, regardless of its channel visibility? Or do you plan to implement some filtering in a following PR?

src/Google/NotificationsService.php
puntope: This is because of the Google Endpoint specification. They are using https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#webhooks as reference.

`action.woocommerce_after_shipping_zone_object_save` means they are referring to `woocommerce_after_shipping_zone_object_save` action in WC.

puntope: Maybe there is not a resource named `shipping` (yes they are `product` and `coupon`) or if it exists maybe they didn't know about it. In any case, the expect that one. `action.woocommerce_after_shipping_zone_object_save`

src/Product/SyncerHooks.php
puntope: Mmm, I thought that authorizing the App and WPCOM Google can actually get data by themselves right?

puntope:
> However, when creating a product, even as just a draft without any data, it triggers the notification endpoint. Is this expected behavior? It appears that with the pushing mechanism, we were bypassing these scenarios:

I also thought we should avoid notifying drafts. However, when running the tests I discovered the current batch syncer (MC push based) yes does actually load them I guess, I thought was expected and Google will handle that on their side (one advantage of doing Pull approach) . WHat do you think about this?

puntope:
> Another question that comes to my mind is, will we be notifying every product, regardless of its channel visibility? Or do you plan to implement some filtering in a following PR?

I assume we will have filters on Google side avoiding access to data without channel visibility set as On. But yes we might implement that maybe.

One question about it is: what happens if a merchant wants to remove the product from the Google App without deleting it?  In case we only sync products with Channel Visibility as On, it would be impossible for the merchant to remove it from MC as setting channel visibility as Off will stop the notifications and hence Google won't be able to get the data anymore.

I am thinking about some options here:

1. We can maybe implement some kind of action when the product changes and call `product.delete` in case channel visibility is set as Off and `product.create` in case is set as On
2. We can rely on Google and on the WPCOM endpoints you created to be sure they don't load Channel Visibility "Off" products. 
3. Any other option you suggest.

> 1. We can maybe implement some kind of action when the product changes and call product.delete in case channel visibility is set as Off and product.create in case is set as On

This is how is done in current implementation.

puntope:
Hi @jorgemd24 I applied most of the suggestions and answered your questions. 

In regards to the discussion about filtering the "not ready" products, I opened some discussions

pcTzPl-1OX-p2#comment-2903
pcTzPl-1OX-p2#comment-2902

In my point of view, relying on the partner to handle all of that may be a valid solution, as well as doing it ourselves. So I prefer to wait for some input before spending time on the implementation. So in any case, it would be done in a future PR if that happens.

If you agree, this is ready for another round.

src/Product/SyncerHooks.php
jorgemd24: Yes, they will have access to the data however, from what I gather, they still require the MC ID, and the account needs to be connected.

jorgemd24 (approved):
Thanks @puntope for your comments! I'm aligned with the idea that ideally, the partner should manage those scenarios. When you look at the webhooks, they essentially send the data, and it's up to the consumer to decide how to handle it. However, I'm unsure, for instance, why we'd need to notify the partner if they prefer not to share certain products with Google. Nevertheless, with the filter we've put in place, when they request a product that isn't syncable, the response will be empty.

>One question about it is: what happens if a merchant wants to remove the product from the Google App without deleting it? In case we only sync products with Channel Visibility as On, it would be impossible for the merchant to remove it from MC as setting channel visibility as Off will stop the notifications and hence Google won't be able to get the data anymore.

> I am thinking about some options here:

>1.We can maybe implement some kind of action when the product changes and call product.delete in case channel visibility is set as Off and product.create in case is set as On
>2.We can rely on Google and on the WPCOM endpoints you created to be sure they don't load Channel Visibility "Off" products.

For option 1,  let's see the opinions shared in the internal discussion you initiated 👍 

Regarding option 2, in theory, when they use the WPCOM endpoints, if the channel visibility is switched off, they shouldn't receive the data. Hence, it should work as well. However, that merchant will receive unnecessary requests.

PR #2198, opened 2024-01-15
Include connected accounts in tracks from the backend.

mikkamp:

This PR adds any connected accounts to tracks which we send from the backend. The accounts are only included if they are connected (non-zero).

There aren't a lot of tracks being sent from the backend to test with, so I added the following line to `ConnectionTest::render_admin_page`:

```php
do_action( 'woocommerce_gla_url_switch_success', [] );
```

1. Make the change (suggested above) to add the manual track
2. Browse a site with both MC and Ads accounts connected
3. Turn on browser developer tools and browse to the hidden connection page:  `https://domain.test/wp-admin/admin.php?page=connection-test-admin-page`
4. In the browser developer tools filter network requests to `pixel.wp.com` 
5. Look for a request for `t.gif` and check the payload
image
6. Confirm that `gla_version`, `gla_ads_id`, `gla_merchant_id` all reflect the right values

### Changelog entry
* Add - Include connected accounts in tracks from the backend.

### DIFF
--- a/src/Tracking/Tracks.php
+++ b/src/Tracking/Tracks.php
@@ -3,6 +3,8 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Tracking;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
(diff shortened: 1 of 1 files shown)


martynmjones (approved):
Hey @mikkamp, thanks for the tracks adjustment.

Confirmed that `gla_version`, `gla_ads_id`, and `gla_merchant_id` values are all included in the payload and values are correct so LGTM ✅

PR #2199, opened 2024-01-17
Add plugin version, Google Merchant Center account ID, and Google Ads account ID to all frontend tracking events

eason9487:
### Changes proposed in this Pull Request:

Project thread: pcTzPl-20H-p2

This PR implements the JS part of the project. The target is to attach the following properties when sending all JS tracking events:

- Plugin Version
- Google Merchant Center account ID if connected
- Google Ads account ID if connected

To achieve the above target, this PR:

- On the PHP side, add base tracking properties to the inline script of outputting the global JS variable `glaData`.
- Add a general state to this extension's wp-data store for accessing the prefetched data (`glaData`) so that the account IDs in frontend cache can be synced after connecting/disconnecting an account.
- To reduce confusion with the `recordEvent` function imported from `@woocommerce/tracks`
   - Rename the **js/src/utils/recordEvent.js** file to **tracks.js**.
   - In the **tracks.js**, add new functions to wrap the `recordEvent` and `queueRecordEvent` to attach the base tracking properties.
   - Replace all `recordEvent` calls with `recordGlaEvent`.
   - Replace all `queueRecordEvent` calls with `queueRecordGlaEvent`.
- Update tracking README.

💡 The `wcadmin_page_view` event is sent from Woo core, it will be handled by another PR

#### 📷 Base tracking properties

image

#### 📹 Attach Google Merchant Center account ID if connected

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/e4d6180f-de45-440a-b593-1b3702fd850e

#### 📹 Attach Google Ads account ID if connected

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/ccb4371e-62f8-4f4f-9913-f0105fe1d0cd

#### Queue a tracking event

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/a5eafe7b-a3d1-4efe-bbcf-a15811e8d94a

### Detailed test instructions:

💡 Apart from the `wcadmin_page_view` event, all tracking events should have the base properties.

1. Enable tracking debug tool
   - Open the Console tab of the browser DevTool.
   - Change the filter level to "All levels".
   - Run `localStorage.setItem( 'debug', 'wc-admin:*' )` and refresh the page to make it work.
   - It should be able to view tracking via the Console tab. Ref:
      
2. Disconnect Google Merchant Center and Google Ads accounts if already connected.
3. Go to the first step of onboarding flow.
4. Check if all tracking events have the `gla_version` property.
[...]

PR #2201, opened 2024-01-18
Send the related tracking event with the account ID to be connected when connecting to an existing Google Merchant Center or Google Ads account

eason9487:

It's a follow-up PR for https://github.com/woocommerce/google-listings-and-ads/pull/2199#issuecomment-1895859258.

This PR adds the account ID to be connected to the properties of the `gla_mc_account_connect_button_click` and `gla_ads_account_connect_button_click ` events.

#### 📹 Connect Google Merchant Center account

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/ffb7cda6-411f-4aa5-9a76-9475e52b8d46

#### 📹 Connect Google Ads account

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/0ce40437-db4b-4022-ac13-898a28d144f4

1. Enable tracking debug tool
   - Open the Console tab of the browser DevTool.
   - Change the filter level to "All levels".
   - Run `localStorage.setItem( 'debug', 'wc-admin:*' )` and refresh the page to make it work.
   - It should be able to view tracking via the Console tab. Ref:
      
2. Disconnect Google Merchant Center and Google Ads accounts if already connected.
3. Go to the first step of onboarding flow.
4. Connect to an existing Google Merchant Center account and check if the `gla_mc_account_connect_button_click` event properties have the `id` and if its value is correct.
5. Proceed to the last step of onboarding flow.
6. Connect to an existing Google Ads account and check if the `gla_ads_account_connect_button_click` event properties have the `id` and if its value is correct.

### Changelog entry

> Add - Send the related tracking event with the account ID to be connected when connecting to an existing Google Merchant Center or Google Ads account.

### DIFF
--- a/js/src/components/google-ads-account-card/connect-ads/index.js
+++ b/js/src/components/google-ads-account-card/connect-ads/index.js
@@ -25,6 +25,7 @@ import './index.scss';
  * Clicking on the button to connect an existing Google Ads account.
  *
  * @event gla_ads_account_connect_button_click
+ * @property {number} id The account ID to be connected.
  */
 
 /**
@@ -127,6 +128,7 @@ const ConnectAds = ( props ) => {
 							isSecondary
 							disabled={ ! value }
 							eventName="gla_ads_account_connect_button_click"
+							eventProps={ { id: Number( value ) } }
 							onClick={ handleConnectClick }
 						>
 							{ __( 'Connect', 'google-listings-and-ads' ) }
--- a/js/src/components/google-mc-account-card/connect-mc/index.js
+++ b/js/src/components/google-mc-account-card/connect-mc/index.js
@@ -26,6 +26,7 @@ import './index.scss';
  * Clicking on the button to connect an existing Google Merchant Center account.
  *
  * @event gla_mc_account_connect_button_click
+ * @property {number} id The account ID to be connected.
  */
 
 /**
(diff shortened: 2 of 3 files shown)


jorgemd24 (commented):
Hi, @eason9487, thanks for this adjustment. I followed the steps that you mentioned in the PR and everything works well, however, I have a doubt before approving this PR, why do we use `id` as an event property instead of their respective names, for example: `gla_ads_id ` and `gla_mc_id `?

eason9487:
Thanks for the review, @jorgemd24

> why do we use `id` as an event property instead of their respective names, for example: `gla_ads_id ` and `gla_mc_id`?

The `gla_mc_id` property is only included when there is a connected Google Merchant Center account, and the same goes for `gla_ads_id` and Google Ads account.

At the moment the "Connect" button is clicked, the account is not yet connected, so `id` is used to indicate the account ID to be connected and to distinguish it from `gla_mc_id` or `gla_ads_id`.

jorgemd24 (approved):
Thanks @eason9487 for the clarification, so it's just a "way" to distinguish between connected and not connected. I guess we will be able to tell which type of `id` is (ads or mc) by checking the event name. I am approving the PR as it works as expected and the doubt that I had about the event property name has been clarified.

eason9487:
> so it's just a "way" to distinguish between connected and not connected. I guess we will be able to tell which type of `id` is (ads or mc) by checking the event name.

Yes, like many other events, they may have the same event property name, but each property's meaning needs to be interpreted along with its event name or somewhat context.

PR #2202, opened 2024-01-18
Scheduled notifications using meta handler

puntope:

This PR adds some work for scheduling the Notifications in batches. 
The strategy of this PR is to set up some product metas and then filter them in a Notification Job.

For that this PR:

- Add single Product Notifcation Job for Create, Delete, Update 
- Adapt `ProductHelper.php` with some helper functions regarding Notifications.
- Adapt `ProductMetaHandler.php` creating some metas for the notifications.
- Create `NotificationStatus.php` to hold the Statuses for the Notification metas
- Tweak `ProductRepository.php` adding some functions to filter the products by Notifications meta.
- Update `SyncerHooks.php` in order to trigger the Notification Jobs.
- Tweak `NotificationsService.php` adding some functions and dependencies.

The way this works is that on Update Product we check if the Notifications are enabled. (For now using a filter). If so, instead of sync to the Merchant we updated some product metas marking a product as PENDING_CREATE, PENDING_DELETE or PENDING_UPDATE. Then we schedule the Jobs that filter those products based on the metas in batches. Finally, we process those items and update their metas to CREATED, UPDATED, DELETED.

The advantages of this strategy are:

- It impacts less in other jobs as it doesn't need to tweak the Action Scheduler Abstract Classes. 
- More efficient and clean in terms of code and collisions. For example, if a product gets pending update in the middle of a schedule it can be notified in the next schedule
- Can be easily refactored to use recurrent jobs and be even more efficient

No need to test in detail the PR. Just offering me some feedback about the strategy. Final PR will come afterwards.

Here are some test to see how it works

1. Create a Product
2. See in Action Scheduler a new job created --> `gla/jobs/notifications//process_item` with one item (the product ID just created) and product.created as topic
3. Run this action
4. Go to Woocommerce - Status - Logs and see the Notify debug Log created  with the product.create topic
5. Update the product 
6. See in Action Scheduler a new job created --> `gla/jobs/notifications/process_item` with one item (the product ID just created) and product.update as topic
7.  Run this action
8. Go to Woocommerce - Status - Logs and see the Notify debug Log created with the product.update topic 
9. Send the product to trash
10. 6. See in Action Scheduler a new job created --> `gla/jobs/notifications/process_item` with one item (the product ID just created) and product.delete as topic
11. Run the action
12. Go to Woocommerce - Status - Logs and see the Notify debug Log created  with the product.delete topic
13. Untrash the product
14. See in Action Scheduler a new job created --> `gla/jobs/notifications/process_item` with one item (the product ID just restored) and product.create as topic
15. Run this action
16. Go to Woocommerce - Status - Logs and see the Notify debug Log created  with the product.create topic
17. Set Dont sync and show in the product
18. See in Action Scheduler a new job created --> `gla/jobs/notifications/process_item` with one item (the product ID just restored) and product.delete as topic
19. Run this action
20. Go to Woocommerce - Status - Logs and see the Notify debug Log created  with the product.delete topic
21. Update the product
22. Nothing is Scheduled (because is set as don't sync and show)
23. Set Sync and Show
24. See in Action Scheduler a new job created --> `gla/jobs/notifications/process_item` with one item (the product ID just restored) and product.create as topic
25. Run this action
26. Go to Woocommerce - Status - Logs and see the Notify debug Log created  with the product.create topic
27. Try to update several times to see there are not several updates happening 
28. Save the product as draft
29. See in Action Scheduler a new job created --> `gla/jobs/notifications/process_item` with one item (the product ID just restored) and product.delete as topic
30. Run the action
31. Go to Woocommerce - Status - Logs and see the Notify debug Log created  with the product.delete topic
32. Save the product as publish
33. See in Action Scheduler a new job created --> `gla/jobs/notifications//process_item` with one item (the product ID just created) and product.created as topic

### Changelog entry

>


mikkamp:
Thanks for presenting these two strategies as separate PR's it's helpful to see how it works. Personally I like the initial approach in this PR more, but I think we should do some changes to support the benefits of both PR's, while reducing the cons of both approaches.

For example these are some of my preferences:
1. In PR #2203 Notification Job inherits from `AbstractProductSyncerBatchedJob`, eventually we are going to reach a point where we will no longer maintain (completely remove) the code for Product Syncing, having a dependency like this in place means we are going to need to do some refactoring later on. So based on that I much prefer the solution in this PR where a Notification Job is completely different.
2. What I don't like so much in this PR is that `AbstractNotificationJob` inherits from `AbstractBatchedActionSchedulerJob`. With product syncing we had either a batch of products to be synced or a trigger such as product create/update which created a batch of 1 to sync. However as far as I can see from the notifications we will always get triggers for a single product. So my intuition says we should be inheriting from `AbstractActionSchedulerJob` we might needs some more overrides in the `AbstractNotificationJob` class to support that single scheduling but that shouldn't be an issue as it won't interfere with any of the existing jobs. The advantages of changing this would be less confusion on keeping track/debugging how notifications get sent, as well as less log (scheduled actions) entries for a single notification. For example we won't need a scheduled action to create a batch of 1 (`gla/jobs/notifications/product.create/create_batch`) and another scheduled action to send the notification (`gla/jobs/notifications/product.create/process_item`), just the latter should be enough.
Taking a closer look it seems that we still need a batch because we mark each variation in a variable product to send a separate notification, it was my understanding that we were just going to send 1 notification for the parent product, if we want we can still include which variation was altered as a notification prop, but we shouldn't have to send a whole batch of notifications for each variation separately (which is what it seems we are doing in this PR).
3. Related to point 2, if we switch to just scheduling single actions for each notification, it would become much easier to just pass the props to a single scheduled action (since it doesn't need to be included in the batch), so we wouldn't need a separate notification class for each type of notification. Having said that though, I don't think it's a real issue to have a separate class for each type as long as they share all the necessary code in the abstract class.

I'm also not super clear on the notification meta which is set for the product. As far as I understand this is to prevent multiple updates from being scheduled shortly after each other. But looking at the code it seems that this status is never completely cleared. So if I have an updated trigger then the product meta will get set to PENDING_UPDATED, and then later once the notification is sent it will be changed to UPDATED, but what happens after that? Does it stay the status UPDATED, so if we legitimately update the product a second time it won't be able to send an update notification? Similar if I update a product and then right away delete it, based on the code it looks like it would block the delete notification from sending because an update notification was still pending.

Another thing I noticed is the way it's setup in the SyncerHooks. Even though we don't trigger `schedule` if a notification for the product is already pending, we can still get double notifications:

1. Product update hook is triggered for product 123
2. Product 123 is marked as PENDING_UPDATED
3. We call `$this->update_products_notification_job->schedule()`
4. It schedules a task for `gla/jobs/notifications/product.update/create_batch`
5. The action `create_batch` is run and queries all products with the meta PENDING_UPDATED (will return 123)
6. It schedules a task for `gla/jobs/notifications/product.update/process_item` with arg 123 (in this scenario we will have to imagine this task takes a long time to complete/run so by the time it creates the next batch it won't have updated the product meta yet, or potentially fails and is retried)
7. A second product is updated which triggers an update hook for product 345
8. Product 345 is marked as PENDING_UPDATED
9. We call `$this->update_products_notification_job->schedule()`
10. It schedules a task for `gla/jobs/notifications/product.update/create_batch`
11. The action `create_batch` is run and queries all products with the meta PENDING_UPDATED (will return 123 and 345)
12. It schedules a task for `gla/jobs/notifications/product.update/process_item` for both the products 123 and 345

If scheduled tasks are all executed in consecutive order (and each task runs without failures or retries), then this wouldn't happen, but that's not something we can guarantee. Hence I think this would be another argument why dropping the batches would be a good idea. At the time when we call `$this->update_products_notification_job->schedule()` we know exactly which product we want to have the notification scheduled for so why not do it directly at that point. The batching and querying made sense for product syncing, but in my opinion it's unnecessary overkill here.

Let me know what you think about these suggestions, happy to discuss them further.

jorgemd24:
> So based on that I much prefer the solution in this PR where a Notification Job is completely different.

Agree with this, I would prefer to keep this independent from the rest of the code, one for the reason that @mikkamp mentioned earlier, that eventually we will remove all the code related to the Product Syncing and second I think the other solution could cause issues with the current implementations creating more friction in the migration process. 

> Some disadvantages
> It requires to create one single job for each topic.

I don't really see this as a disadvantages. It actually would helps us break down the logic and makes it easier to read.

> Taking a closer look it seems that we still need a batch because we mark each variation in a variable product to send a separate notification, it was my understanding that we were just going to send 1 notification for the parent product, if we want we can still include which variation was altered as a notification prop, but we shouldn't have to send a whole batch of notifications for each variation separately (which is what it seems we are doing in this PR)

Yeah, I think so too. Sending all the variations seems unnecessary. A more efficient approach would be to include the variation_id so that Google can fetch the specific item. But for now, I guess we can just send the parent id, and Google will have to fetch the complete product.

I played around with `AbstractActionSchedulerJob`, and I think we could use a single job to handle all the notifications. For instance, we could create a job that takes the item id and the topic as parameters:

`$this->notification_job->schedule_notification( $product_id , $this->notifications_service::TOPIC_X );`

And then the `process_items` method could look like the below:

```
	protected function process_items( array $items ) {
		foreach ( $items as $item ) {
			$this->notifications_service->notify( $item['item_id'], $item['topic'] );
		}
	}
```

This way, we could use the same job for different topics, regardless if it's a product, coupon, shipping zone, and so on.

Another benefit of using `AbstractActionSchedulerJob` is that we can check if a job with the same topic and the same item_id is scheduled or has been scheduled before so we can avoid duplicate notifications.

puntope:
Thanks, @mikkamp and @jorgemd24 For all the suggestions. I also think the approach followed in this PR is cleaner. So I will close the other one and continue working in this direction. 

> I'm also not super clear on the notification meta which is set for the product. As far as I understand this is to prevent multiple updates from being scheduled shortly after each other. But looking at the code it seems that this status is never completely cleared. So if I have an updated trigger then the product meta will get set to PENDING_UPDATED, and then later once the notification is sent it will be changed to UPDATED, but what happens after that? Does it stay the status UPDATED,

Yes, the metas are for reducing duplications and inconsistencies, for example, if a product has not states "CREATED", "UPDATED" or "PENDING_UPDATE"  we don't trigger "PENDING_DELETE"   because the product was not notified creation yet. So google would receive `product.delete` notification before `product.create`.

The status UPDATED allows you to call "deletes" and "updates" in further changes.

puntope:
Regarding batches, if we are not going to introduce batches we don't really need jobs... we can just call the notify function I guess. And reduce quite a lot the logic. 

However, I thought a product with let's say 20 variants... would be too much (20 requests to WPCOM) That's why I introduced batched jobs.

puntope:
>  played around with AbstractActionSchedulerJob, and I think we could use a single job to handle all the notifications. For instance, we could create a job that takes the item id and the topic as parameters:

Another alternative, if using batches, is to have a single NotificationJob and depending on the state, it calls different topics. For example, if it has PENDING_UPDATE... it will call topic `product.update`

puntope:
> Yeah, I think so too. Sending all the variations seems unnecessary. A more efficient approach would be to include the variation_id so that Google can fetch the specific item. But for now, I guess we can just send the parent id, and Google will have to fetch the complete product.

If it's ok to send just the parent, then I guess we don't need jobs.

jorgemd24:
> Regarding batches, if we are not going to introduce batches we don't really need jobs... we can just call the notify function I guess. And reduce quite a lot the logic.

> If it's ok to send just the parent, then I guess we don't need jobs.

I think the benefit of having the job is that the request will happen in the background and not during updating/creating/deleting the products. Also, what would happen if you do a bulk update of products without having a job? Would we make all the requests at once?

puntope:
> I think the benefit of having the job is that the request will happen in the background and not during updating/creating/deleting the products. 

Since it is a single request. I still don't see the need for the job.

> Also, what would happen if you do a bulk update of products without having a job? Would we make all the requests at once?

For the bulk update, the way WordPress works is that it will trigger the hook one time for each product. So, at the end, for 1000 products, it will trigger 1000 jobs. 

The good thing about using jobs in batches is that we can omit the scheduled job call, and just update the metas. Then we can have a single recurring batched job notifying products in the background based on those metas.

jorgemd24:
> For the bulk update, the way WordPress works is that it will trigger the hook one time for each product. So, at the end, for 1000 products, it will trigger 1000 jobs.

From my understanding, that's similar to how it works now:

image

We made a single request/job for each product updated. Instead, if we don't use jobs, those 1000 requests will be executed during the bulk update. With jobs, these requests can be spread out over an interval since the AS will execute them when possible.

puntope:
Hi @jorgemd24 and @mikkamp I updated this PR adding a job scheduler without batches.

Note: After discussing with @mikkamp non published products are not notified (except for deletion if they go from publish to draft for example)

puntope:
Notice there is a problem with PHP 7.3 I'm not sure whats going on and I'm investigating why tests are failing there

mikkamp:
> Notice there is a problem with PHP 7.3 I'm not sure whats going on and I'm investigating why tests are failing there

I assume you mean for PHP 7.4 as that's the minimum supported version. As for the reason it's failing, it's having a problem parsing the constructor function of the NotificationsService class. Only in PHP 8.0+ are you allowed to have a trailing comma after the last parameter: https://github.com/woocommerce/google-listings-and-ads/pull/2202/files#diff-cd15bec1b37fc3ce5e4e98e98f3883bf14138327286512ddaf10f513613bf456R63

Removing the comma should be sufficient to get it to pass.

src/Google/NotificationsService.php
```diff
@@ -34,12 +38,34 @@ class NotificationsService implements Service {
...
-		$blog_id     = Jetpack_Options::get_option( 'id' );
-		$this->route = "https://public-api.wordpress.com/wpcom/v2/sites/{$blog_id}/partners/google/notifications";
+	public function __construct(
+		ProductRepository $product_repository,
+		ProductHelper $product_helper,
+	) {
+		$this->product_repository = $product_repository;
+		$this->product_helper     = $product_helper;
+		$blog_id                  = Jetpack_Options::get_option( 'id' );
+		$this->route              = "https://public-api.wordpress.com/wpcom/v2/sites/{$blog_id}/partners/google/notifications";
```
mikkamp: Not a big deal, but I think `route` might not be an accurate description. A route would usually describe a path to the controller. For example `sites/{$blog_id}/partners/google/notifications`. Since this is the full URL I'd describe it more as a WP.com Notification Endpoint, or Notification URL. Do you think `notification_url` would be a better describing alternative?

src/Google/NotificationsService.php
```diff
@@ -107,4 +141,74 @@ protected function do_request( $args ) {
...
+
+	/**
+	 * Find a product matching the provided status
+	 *
+	 * @param int    $product_id The product Id to filter
+	 * @param string $topic The product topic to filter
+	 *
+	 * @return int|null
+	 */
+	public function filter_product( $product_id, $topic ) {
```
mikkamp: I'm still not understanding the full purpose of this function. Could you help clarify why it's still needed?

From my perspective this is a left over from working through notifications in batches. With a single item (product) a query seems overkill. Let me explain what I'd be expecting:

We have one function to check if a product notification should be sent, which passes either the product ID or the product object. Based on this product object we check directly if:
- Channel Visibility is set to sync
- it has the right status for a notification (published)
- is a syncable product
- is the right product type
- we call a filter with product object so extensions (or our compatibility code) can prevent this notification from being scheduled/sent

Once we have a function like this we can call it both before the notification is scheduled (in the SyncerHooks class). And again just before the scheduled task is run in case there has been any changes to the product in the mean time.

That would be very different from the current behaviour where we do the following:
1. In SyncerHooks we set the status of a product through a meta key, then we schedule a task for the notification (at this stage we do some preliminary checks to see if the product should send the notification but this doesn't include a filter).
2. When the scheduled notification is run for a specific product we run it through a `product_filter` which is done by querying products with a particular notification status, at this stage we also allow a filter to modify the query in case an extension wants to prevent the notification from being sent.
3. Only if the query returns a valid product (the same one we queried for) do we send the scheduled notification.

In my opinion the current method in this PR is limited in the following ways:
- We can't hook into the notification early enough to prevent it from being scheduled (we can only prevent it from being sent)
- Querying is intended for when we want a set of products with specific criteria (which is needed for batch product handling). When we already know which product we want to work with we can check the same criteria on the product object directly (which will use less resources).
- We need to set a NotificationStatus as a product meta, but if we eliminate the query and use ActionScheduler to check if we already have a scheduled action with the same item ID and notification topic. Then my question is do we really need this extra product meta? Action Scheduler can handle all the meta needed for the notification (which would also make notifications for shipping methods easier because we don't have the option to store meta on the item).

src/Jobs/Notifications/ProductNotificationJob.php
```diff
@@ -0,0 +1,84 @@
...
+	defined( 'ABSPATH' ) || exit;
+
+	/**
+	 * Class ProductNotificationJob
+	 *
+	 * Class for all the Product Notifications Jobs
+	 *
+	 * @package Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications
+	 */
+class ProductNotificationJob extends AbstractActionSchedulerJob implements JobInterface {
```
mikkamp: Not sure if PHPCS is catching this as the task fails with some errors, but this file uses a bit of mixed alignment with some parts being intended a bit to far.

src/Google/NotificationsService.php
```diff
@@ -34,12 +38,34 @@ class NotificationsService implements Service {
...
+	 *
+	 * @param ProductRepository $product_repository
+	 * @param ProductHelper     $product_helper
 	 */
-	public function __construct() {
-		$blog_id     = Jetpack_Options::get_option( 'id' );
-		$this->route = "https://public-api.wordpress.com/wpcom/v2/sites/{$blog_id}/partners/google/notifications";
+	public function __construct(
+		ProductRepository $product_repository,
+		ProductHelper $product_helper,
```
mikkamp: I'm wondering if these dependencies make sense. I've already suggested why we might not need a query in other comments.

But if the `notify` function is agnostic of what type of notification it is then shouldn't the calling class handle any specifics about the notification type? Setting the "after notification status" can easily be handled by the `Product NotificationJob` which knows that it's working with a product.

Otherwise eventually we are going to need to keep increasing the dependencies for this service to have product helpers, shipping rate helpers, etc...

mikkamp (commented):
Thanks for refactoring the PR, it's already starting to look a lot cleaner.

I left some comments about specific parts which I feel we could simplify even further (and would help with reducing maintenance burden). I also think we can have a bit better separation between which class handles what, which will make unit testing a bit easier to solely test the strict responsibility of each class/service. Let me know what you think about those comments and if there is anything else I should elaborate further.

src/Google/NotificationsService.php
mikkamp: Another alternative approach would be to make `https://public-api.wordpress.com/wpcom/v2/` a constant or just a protected/private property defined in the class. And then have route as a property set in the constructor to `sites/{$blog_id}/partners/google/notifications`

Then the `get_route` function can be renamed to `get_notification_url` and return a combination of the URL and the route.

mikkamp:
> If it's ok to send just the parent, then I guess we don't need jobs.

For variable products we might need to debate a bit more on whether we still include the variation ID as a property. I don't know if we want to give that a name like "child item".

For example if we have:
```
Variable Product ID = 10
- Variation ID = 11
- Variation ID = 12
- Variation ID = 13
```

If we catch an update to the variable product we just send item ID 10. If we get an update to just variation ID 12, we send an update for item ID 10 with child item 12. The client attached to the notification can then pull ID 10 and decide whether it wants to update all variations or just 12.

I think we'd then need to decide whether we want to have it amend an already scheduled task if we have a notification for an additional variation, or if we find it acceptable to send another notification.

For the update it doesn't matter too much if the child item ID(s) are sent, but it does make a big difference for a delete notification. For example if we delete variation 13, we can't just convert that into a delete for product 10. So either we need to change that to an update item 10, where the client attached to the notification can pull ID 10 and refetch all variations, and then decide if some disappeared or not. Or we'll have to include the child item in the delete so only the right variation gets deleted.

I think both approaches would be feasible, we just need to keep in mind who has the responsibility to handle what part.

src/Product/ProductHelper.php
```diff
@@ -332,6 +334,81 @@ public function is_product_synced( WC_Product $product ): bool {
...
+	public function should_trigger_delete_notification( WC_Product $product ): bool {
+		return ! $this->is_ready_to_notify( $product ) && $this->has_notified_creation( $product );
+	}
+
+	/**
+	 * @param WC_Product $product
+	 *
+	 * @return bool
+	 */
+	public function has_notified_creation( WC_Product $product ): bool {
```
jorgemd24, line 404: I got a bit confused with all the statuses but what I understand is that these statuses indicate whether the product has been created/synced, helping us, for example, to avoid sending an update/delete topic if the product has not been notified yet.

Also, I believe we need to clarify that the creation/delete topics are related to the MC status. Sometimes, I found myself mixing the concepts of creation/delete with creating/deleting products in WC, whereas in the MC case, it could be just changing the channel visibility from "syncable" to "not syncable." or vice-versa. 

Moreover, I'm trying to understand why we need to track this status. For instance:

- If the product goes from not syncable to syncable, from my understanding, we could just send the create notification and rely on checking if there are no more jobs with the creation topic for the product id. I'd say the same for update and delete. However, it would be helpful to know the logic Google has in place. For example, if they receive an update notification for a product that is not created in the MC, will they create the product? Or will it be refused? If this logic is handled by Google, we could simplify the notification system significantly. Usually webhooks/notications are responsible for alerting about changes, and it's up to the consumer to handle the various scenarios.

- Another scenario involves an existing product that is already synced with the MC; the created metadata does not exist because the product was already synced before we started using notifications. If the merchant updates the product, we will send the create notification, which is incorrect because the product was already created in the MC. A similar case would happen when removing a product.

src/Product/ProductHelper.php
```diff
@@ -5,11 +5,13 @@
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Exception\InvalidValue;
 use Automattic\WooCommerce\GoogleListingsAndAds\Google\GoogleProductService;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\NotificationsService;
```
jorgemd24: NotificationsService seems unused in this file

src/Google/NotificationsService.php
```diff
@@ -107,4 +141,74 @@ protected function do_request( $args ) {
...
+		return $query_results[0] ?? null;
+	}
+
+	/**
+	 * Set the notification status for a product.
+	 *
+	 * @param int    $product_id
+	 * @param string $status
+	 */
+	public function set_status( $product_id, $status ) {
```
jorgemd24: In this file we have a mix of functions having the arguments types and return types, should we not follow the same strategy to keep consistency in the file?

src/Google/NotificationsService.php
```diff
@@ -107,4 +141,74 @@ protected function do_request( $args ) {
...
+	 */
+	public function is_enabled(): bool {
+		return apply_filters( 'woocommerce_gla_notifications_enabled', true );
+	}
+
+	/**
+	 * Find a product matching the provided status
+	 *
+	 * @param int    $product_id The product Id to filter
+	 * @param string $topic The product topic to filter
```
jorgemd24: We say that the params can be only int or string but in reality can be null too.

src/Jobs/Notifications/ProductNotificationJob.php
```diff
@@ -0,0 +1,84 @@
...
+	protected function process_items( array $args ) {
+		$item  = $args[0] ?? null;
+		$topic = $args[1] ?? null;
+
+		$item_id = $this->notifications_service->filter_product( $item, $topic );
+
+		if ( ! is_null( $item_id ) && ! is_null( $topic ) ) {
+			$this->notifications_service->notify( $item, $topic );
+		}
+	}
```
jorgemd24: For me sounds more logical that if an item or topic is null, we should just `return` instead of calling `$this->notifications_service->filter_product` with `null` parameters.

src/Google/NotificationsService.php
```diff
@@ -107,4 +141,74 @@ protected function do_request( $args ) {
...
+	/**
+	 * Set the notification status for a product.
+	 *
+	 * @param int    $product_id
+	 * @param string $status
+	 */
+	public function set_status( $product_id, $status ) {
+		$product = $this->product_helper->get_wc_product( $product_id );
+		$this->product_helper->set_notification_status( $product, $status );
+	}
```
jorgemd24: Agree with @mikkamp here, that I think it makes more sense to update the status in the ProductNotificationJob so we keep the NotificationService agnostic.

jorgemd24 (commented):
Thanks, @puntope, for all the work! I've added some comments, discussing the statuses and other scenarios that might occur. Let me know if anything isn't clear!

src/Google/NotificationsService.php
puntope: Hi @mikkamp. The only reason for this function is to allow the user to hook a filter (as we discussed) 

The function calls `$this->product_repository->find_notification_products` that internally query by ids and calls `execute_woocommerce_query` and `prepare_query_args`. This is how we allow the user to filter the query args i sync strategy. 

If this is not needed I agree to remove this function.

> We need to set a NotificationStatus as a product meta.

The reason of this metas is to know if a product notified `product.create`  or other notifications so we can handle the updates/creates/deletes.

src/Product/ProductHelper.php
puntope: > I got a bit confused with all the statuses but what I understand is that these statuses indicate whether the product has been created/synced, helping us, for example, to avoid sending an update/delete topic if the product has not been notified yet.

Correct. For example. Imagine this scenario:

- Creating a product with DONT SYNC AND SHOW won't trigger `product.create` notification. 
- Then the user updates the product and set SYNC AND SHOW. The notification should be `product.create` not `product.update` for doing this is necessary the metas and some helper functions.

puntope: > Another scenario involves an existing product that is already synced with the MC; the created metadata does not exist because the product was already synced before we started using notifications. If the merchant updates the product, we will send the create notification, which is incorrect because the product was already created in the MC. A similar case would happen when removing a product

I assume we don't take in consideration the previous statuses of (PUSH based) MC for sending the Notifications

src/Google/NotificationsService.php
puntope: Adjusted here. https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/89400673971356a0d34b2ca3d74f2b329e1b5ee8

src/Jobs/Notifications/ProductNotificationJob.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/b3b0735c09a3a9464f54eaa1024fcf0c481dc0d2

src/Google/NotificationsService.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/commit/b3b0735c09a3a9464f54eaa1024fcf0c481dc0d2

src/Google/NotificationsService.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/ba7ed8fb588c6dc432dd64e0502cb4a27b3d77f7

src/Product/ProductHelper.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/3d9b4bc5eb5ef92e2d03fdc67b78dd378172dcae

src/Google/NotificationsService.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/89400673971356a0d34b2ca3d74f2b329e1b5ee8

src/Jobs/Notifications/ProductNotificationJob.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/8ee54d1c1572bf351dfe69ff3315f6e0d24dbc06

src/Google/NotificationsService.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/545cd22decc425db15325df7cc96914209008517

src/Google/NotificationsService.php
puntope: I remove tho the `filter_product` function. I would appreciate some indications about where to include the filter for the users.

puntope:
> > If it's ok to send just the parent, then I guess we don't need jobs.
> 
> For variable products we might need to debate a bit more on whether we still include the variation ID as a property. I don't know if we want to give that a name like "child item".
> 
> For example if we have:
> 
> ```
> Variable Product ID = 10
> - Variation ID = 11
> - Variation ID = 12
> - Variation ID = 13
> ```
> 
> If we catch an update to the variable product we just send item ID 10. If we get an update to just variation ID 12, we send an update for item ID 10 with child item 12. The client attached to the notification can then pull ID 10 and decide whether it wants to update all variations or just 12.
> 
> I think we'd then need to decide whether we want to have it amend an already scheduled task if we have a notification for an additional variation, or if we find it acceptable to send another notification.
> 
> For the update it doesn't matter too much if the child item ID(s) are sent, but it does make a big difference for a delete notification. For example if we delete variation 13, we can't just convert that into a delete for product 10. So either we need to change that to an update item 10, where the client attached to the notification can pull ID 10 and refetch all variations, and then decide if some disappeared or not. Or we'll have to include the child item in the delete so only the right variation gets deleted.
> 
> I think both approaches would be feasible, we just need to keep in mind who has the responsibility to handle what part.

I will move this discussion into a P2 then and work on that part in further PRs

src/Product/ProductHelper.php
puntope: I will move that question to the P2

puntope:
Hi @mikkamp  and @jorgemd24  I addressed all the comments and posted some of the discussions in P2. So they can be addressed in further PRs once they are discussed.

pcTzPl-1OX-p2#comment-3036

p.s The tests in regards the Notifications Jobs will be done also in a future PR once we agreed on the structure of the Jobs.

src/Google/NotificationsService.php
mikkamp: Sorry for any confusing. When I said we'd like to retain a filter my intention was to still allow overriding when a notification gets sent through a WP hook (filter). The querying and filtering of only eligible products for notifications isn't what I meant to retain.

As for where to place the hook, I believe this should be called twice:
1. Just before being scheduled
2. Just before the scheduled action is run (in case the product changed in the meantime)

Right now you already have `is_ready_to_notify` which gets checked before scheduling, I think that's the right place to add a hook, and also make sure it checks all the conditions like:
- Channel Visibility is set to sync
- it has the right status for a notification (published)
- is a syncable product
- is the right product type
- call filter at the end so extensions (or our compatibility code) can prevent this notification from being scheduled/sent

We'd just then need to repeat the call when the action is run and if `is_ready_to_notify` returns false it would prevent the notification from being sent.

mikkamp:
Thanks for all the refactoring it's definitely going in the right direction. I had an initial look and added some more clarification. But I didn't have a chance to fully test the code yet.

src/Google/NotificationsService.php
puntope: Hi @mikkamp thanks for the insightful clarifications. I added some useful hooks here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/99f5b9ede4459201090e16b51d6a790add75a03b

src/Google/NotificationsService.php
```diff
@@ -95,7 +115,7 @@ private function notification_error( int $item_id, string $topic, string $error
 	 * @param array $args
 	 * @return array|\WP_Error
 	 */
-	protected function do_request( $args ) {
+	protected function do_request( array $args ) {
```
jorgemd24: Missing return type

src/Jobs/Notifications/ProductNotificationJob.php
```diff
@@ -0,0 +1,134 @@
...
+		if ( ! isset( $args[0] ) || ! isset( $args[1] ) ) {
+			return;
+		}
+
+		$item  = $args[0];
+		$topic = $args[1];
+
+		if ( $this->notifications_service->notify( $item, $topic ) ) {
+			$this->set_status( $item, $this->get_after_notification_status( $topic ) );
+
```
jorgemd24: Unnecessary extra space

src/Product/ProductHelper.php
```diff
@@ -332,6 +333,93 @@ public function is_product_synced( WC_Product $product ): bool {
...
+	 * @param WC_Product $product
+	 *
+	 * @return bool
+	 */
+	public function should_trigger_delete_notification( WC_Product $product ): bool {
+		return ! $this->is_ready_to_notify( $product ) && $this->has_notified_creation( $product );
+	}
+
+	/**
+	 * @param WC_Product $product
```
jorgemd24, line 400: Can we add a PHPDoc explaining this function?

src/Product/ProductHelper.php
jorgemd24: > I assume we don't take in consideration the previous statuses of (PUSH based) MC for sending the Notifications

Imagine a merchant using the plugin and having products already synced in the MC before the notification system was implemented. Then, they start using the new version of the plugin, which includes the notification system. For that merchant, we'll end up sending the wrong notifications because the initial metadata isn't set.

Another potential concern is that if one of the topic's metadata fields isn't set, we'll begin sending incorrect notifications, leading to an out-of-sync situation. I believe relying on previous notification states seems a bit risky.

src/Product/ProductHelper.php
```diff
@@ -332,6 +333,93 @@ public function is_product_synced( WC_Product $product ): bool {
...
+
+		/**
+		 * Allow users to filter if a product is ready to notify.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param bool $value The current filter value.
+		 * @param WC_Product $product The product for the notification.
+		 */
+		return apply_filters( 'woocommerce_gla_is_ready_to_notify', $is_ready, $product );
```
jorgemd24, line 357: I thought I could use this filter to stop notifications for a specific product, say 123, but instead of stopping them, it's triggering the `product.delete` notification for the product in the MC. Is this expected behavior?

src/Jobs/Notifications/ProductNotificationJob.php
```diff
@@ -0,0 +1,134 @@
...
+	 * @param array $args
+	 */
+	public function schedule( array $args = [] ) {
+		/**
+		 * Allow users to disable the notification job schedule.
+		 *
+		 * @since x.x.x
+		 *
+		 * @param bool $value The current filter value. By default, it is the result of `$this->can_schedule` function.
+		 * @param array $args The arguments for the schedule function.
```
jorgemd24: Can we add that `$args` is composed of `item_id` and `topic`?

jorgemd24 (commented):
Thanks, @puntope, for the improvements! Overall, it's working well, but I still have some concerns about relying on previous notification updates as It could lead to easily going out of sync and sending incorrect notifications, such as deleting products in the MC. Also see my comment about existing merchants using the GL&A plugin. 

Additionally, I left a comment about the `woocommerce_gla_is_ready_to_notify` filter because I'm not sure if it's producing the expected result.

I am happy to approve this PR so we can set the initial code, and then we can address the other comments in further PRs. Let me know what you think.

puntope:
> Imagine a merchant using the plugin and having products already synced in the MC before the notification system was implemented. Then, they start using the new version of the plugin, which includes the notification system. For that merchant, we'll end up sending the wrong notifications because the initial metadata isn't set.

> Another potential concern is that if one of the topic's metadata fields isn't set, we'll begin sending incorrect notifications, leading to an out-of-sync situation. I believe relying on previous notification states seems a bit risky.

Thanks @jorgemd24  for the helpful insights I moved the discussion in the p2 pcTzPl-1OX-p2#comment-3036

I [added some code ](https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/63152097da3ea670dc202fcfd61cff6c234c9f63)for now, supporting already synced products in MC. So, in case after discussion we decide we don't, I will remove it.

src/Product/ProductHelper.php
puntope: I believe yes. If a product has notified creation and is not anymore ready for notify, it should be deleted. 

```
public function should_trigger_delete_notification( WC_Product $product ): bool {
		return ! $this->is_ready_to_notify( $product ) && $this->has_notified_creation( $product );
	}
```

src/Jobs/Notifications/ProductNotificationJob.php
puntope: Added here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/677e4ec93e058379e99d00c99211ad8570af7615

src/Google/NotificationsService.php
puntope: Added here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/08efcce68547d6bedbfa2efce5b68cee68f1856b

src/Product/ProductHelper.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/cbb00a26b3b3decaef9b8cf8396fc433651818d2

src/Jobs/Notifications/ProductNotificationJob.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/d93c7b730cc687922a1763acd9644eaee47fd131

puntope:
Thanks @jorgemd24 for the reviews. I adjusted the code based on your comments and answered your questions.

puntope:
I added some tests for the Job here https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/65dff810293346c5396e16b4a8541b6f44add6df

tests/Unit/Google/NotificationsServiceTest.php
```diff
@@ -5,12 +5,13 @@
 
 	use Automattic\WooCommerce\GoogleListingsAndAds\Google\NotificationsService;
 	use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
+	use WC_Helper_Product;
```
jorgemd24: Unused WC_Helper_Product

src/Jobs/Notifications/ProductNotificationJob.php
```diff
@@ -0,0 +1,133 @@
...
+		return 'notifications/products';
+	}
+
+
+	/**
+	 * Logic when processing the items
+	 *
+	 * @param array $args Arguments with the item id and the topic
+	 */
+	protected function process_items( array $args ) {
```
jorgemd24: Missing return types, I haven't checked all the files specifically, but I see a mix of functions with types and return types, and others without.

src/Product/ProductHelper.php
jorgemd24: I was imagining a scenario where I have my product in the MC, and I want to keep it and manage it myself. So, I would use this filter to stop the notifications.  WDYT?

jorgemd24 (approved):
> I [added some code ](https://github.com/woocommerce/google-listings-and-ads/pull/2202/commits/63152097da3ea670dc202fcfd61cff6c234c9f63)for now, supporting already synced products in MC. So, in case after discussion we decide we don't, I will remove it.

Thanks, @puntope, for the adjustments. However, now when I trash a product and then restore it, I get the following:

- When I trash the product, I get the delete notification as expected.
- However, when I restore it, I get the update notification instead of the create notification. This happens because the product is still "synced" and still has the `gla_google_ids` in the database.

For now, I think it's better to leave it as it was until we decide what to do with existing products.

Another concern that comes to mind is what happens for a store with a high volume of jobs. The job will take some time to be executed, and the status can be incorrect. Similar to the earlier scenario, the following will happen:

- Trash the product.
- Now the status will be "pending_delete" -> eventually we will send a delete notification.
- Restore the product while the status is still "pending_delete".
- You get a "product.update" notification job. I would expect a create notification.


So, I think this PR is a good initial approach. However, I have my concerns about the status and the metadata. I'm approving this PR so we can address these discussions in a different PR based on the discussion that are in the P2.

puntope:
> However, when I restore it, I get the update notification instead of the create notification. This happens because the product is still "synced" and still has the gla_google_ids in the database.

Thanks @jorgemd24 I missed that case. If we follow the approach of considering the current MC as a parameter for the notifications.  Then on deletion, we should also delete those metas. Or at least tweak the functions for doing it in another way. 

> Another concern that comes to mind is what happens for a store with a high volume of jobs. The job will take some time to be executed, and the status can be incorrect. Similar to the earlier scenario, the following will happen:

That's right, If the notification was under PENDING_DELETE it didn't send the notification for delete yet.  In the previous approach, in case the product changes again it would change the status to PENDING_UPDATE (because delete was never called). However, now it doesn't work that way as we process the items right away, so I tweaked the function to not use PENDING_DELETE as status for the `has_notified_creation`. We can continue tweaking those functions once the discussion is finished.

puntope:
> I was imagining a scenario where I have my product in the MC, and I want to keep it and manage it myself. So, I would use this filter to stop the notifications. WDYT?

For that right now, you can use DONT_SYNC_AND_SHOW and create the product yourself in MC

src/Jobs/Notifications/ProductNotificationJob.php
puntope: : void is the default return type so seems not necessary. However I added them if you think is more convinient

mikkamp (approved):
I'll go ahead and approve this as we can continue tweaking in followup PR's to get the status right.

The only comment I have is to also call the `is_ready_to_notify` function from `ProductNotificationJob::process_items`, and stop early if it's false. Since it could take a while before the notification is actually sent.

Also can we just get the PHPCS checks to pass so we end up with a bit cleaner code:
image

PR #2205, opened 2024-01-18
[R-1] Enable users to seamlessly set up conversion tracking, without having to set up merchant center first or requiring campaign creation

### LINKED ISSUE #2215: [R-1] Enable users to seamlessly setup conversion tracking, without requiring merchant center setup or campaign creation
# Overview 

The goal of this requirement is to move the Google Ads Account Connection process from Step 4 to Step 1 of the onboarding workflow, as a required step above the Merchant Center account setup so that conversion tracking can be set up without requiring a Merchant Center connection.

Design Reference: [Figma Link](https://www.figma.com/file/fODAWLBsbvd2MVmdh1Hgk3/Google-Listings-%26-Ads---Alternative-Conversion-Tracking-Flow?type=design&node-id=616-11376&mode=design&t=04zEOEnkTYBWOUia-0)

## Requirements Breakdown

- [ ] Add Google Ads Account Connection to Step 1 of the Onboarding flow
- [ ] Update Step 4 of the Onboarding flow

### Add Google Ads Account Connection to Step 1 of the Onboarding flow

PR: [Allow conversion tracking without merchant account](https://github.com/woocommerce/google-listings-and-ads/pull/2205)

- Add the Google Ads Account Connection Module from Step 4 to Step 1 of onboarding flow (this will also trigger ads conversion tracking setup)
  - Google Ads Account Connection Module should be above the Google Merchant Center Connection Module 
  - Google Ads Account setup should be required in this step; i.e., users should not be able to move forwards in the onboarding flow without setting up a Google Ads account (or other accounts connected in this page)  
  - If users set up Google Ads first, then the link between Merchant Center and Google Ads should happen upon MC creation. 
- Enable Ads Conversion Tracking Setup Upon Ads Account Creation (previously upon “Complete Setup” after billing & campaign creation)
- Users should be able to clearly understand what each step is doing and what is required depending on their journey (i.e., conversion tracking setup only, full ads setup, free listings only) 
  - Google Ads: “Required to set up conversion measurement and create campaigns. You can also create or connect an account later.”
  - Google Merchant Center: Required to sync products and list on Google 
- Once the Google ads account is created and connected and conversion tracking is enabled, the user should be informed of that state. Users who would like to set up Conversion Tracking only would stop at this step  
- E2E tests
  - Update or add E2E tests in tests/e2e/specs/setup-mc/
  - Fix existing  E2E test if failing
- Add backward compatibility if needed. This change should not break existing functionality. 

### Update Step 4 of the Onboarding flow

- Implement copy changes to step title, subtitle, ads benefits card, and sidebars as listed in the [Figma](https://www.figma.com/file/fODAWLBsbvd2MVmdh1Hgk3/Google-Listings-%26-Ads---Alternative-Conversion-Tracking-Flow?type=design&node-id=616-11376&mode=design&t=bIsFdqe2i9xl5SBA-0)
- Remove product sync completion card 
- Since the user has already connected an ads account in Step 1:
  - Remove the Google Ads Account Connection Module from step 4
  - Campaign setup steps should be shown by default
  - The user will still be given an option at the bottom of the page to “Skip Campaign Setup, for now
- E2E tests
  - Update or add E2E tests in tests/e2e/specs/setup-mc/
  - Fix existing  E2E test if failing
- Add backward compatibility if needed. This change should not break existing functionality.
```[tasklist]
### Sub-tasks
- [ ] https://github.com/woocommerce/google-listings-and-ads/issues/2262
- [ ] https://github.com/woocommerce/google-listings-and-ads/pull/2308
```

ankitrox:

Make improvements to ads conversion tracking experience, including the ability to set-up conversion tracking without needing to complete full onboarding flow

Users should be able to connect to Google ads account in step-1 of the onboarding flow.

Closes #2215.

#### Current Issue

In the current onboarding journey, user has to go through the Merchant Account center setup in order to connect to the ads account which will eventually setup the conversion tracking. This creates a dependency between ads and merchant account which should not be the case and user should be able to setup conversion tracking without having to setup the merchant account.

#### Changes in the PR

1. Moved the Google Ads card component from step 4 of the onboarding journey to step 1 of the onboarding journey.
2. Placed the Google Ads card component after Google account card and before Merchant Center card component.
3. Google Ads and Merchant account cannot be connected till jetpack and Google accounts are connected.
4. However; Google ads and MC account card are decoupled from each other; meaning ads or merchant account can be created before/after the other one.
5. Users would require all accounts to be connected in step-1 of the onboarding journey in order to proceed to the next step.

### Screenshots:

Screenshot 2024-02-05 at 10 57 58 PM

1. Start onboarding journey with Marketing > Google Listing & Ads > Start listing products
2. In the step-1 of the onboarding journey, you should be able to view 4 cards
    a. Jetpack
    b. Google
    c. Google Ads
    d. Google Merchant Center
3. Until Jetpack and Google accounts are not connected, Ads and MC cards should stay in disabled state. User should not be able to connect to ads or MC if the Jetpack and Google accounts are not connected.
4. Once Jetpack and Google accounts are connected, Google ads and MC cards should be enabled. If there are existing ads account, you should be able to see dropdown with those accounts. If no existing accounts, there should be "Create Account" button, which would create an ads account.
5. Once the Ads account is successfully connected, there would be a message in ads card component stating "Conversion measurement has been set up. You can create a campaign later.". Please refer following screenshot.
6. You should be able to connect to MC account before Ads account and vice versa.

### Changelog entry

> Add - Google Ads account invitation acceptance step to the connection process.
> Update - Move the Google Ads account connection process from step 4 to step 1 of the onboarding flow.
> Update - Enable users to seamlessly set up conversion tracking, without having to set up merchant center first or requiring campaign creation.


src/Ads/AccountService.php
```diff
@@ -160,10 +160,6 @@ public function setup_account(): array {
 						$this->check_billing_status( $account );
 						break;
 
-					case 'link_merchant':
```
joemcgill, line 163: Rather than removing this step entirely, I think we still want to check to see if a Merchant Account is already registered, and if so, try to link them during the Ads setup. However, since we are giving the option to create an Ads account prior to the MC connection, we shouldn't throw an exception if an MC id does not exist so the step will be marked as done.

We will also need to add a step to the `MerchantCenter\AccountService::setup_account_steps` method to try to link the Ads and MC accounts so that this process happens when connecting an MC account as well.

I've set this all up in #2223 with some updated PHPUnit tests.

joemcgill (commented):
Thanks @ankitrox. I think this is getting close. The main thing that needs updating is the API handling for linking the Ads and MC accounts and updating PHPUnit and E2E tests. I've taken a pass at getting most of this working in #2223. Feel free to merge those into this branch if you'd like and handle additional test updates here.

js/src/components/google-mc-account-card/google-mc-account-card.js
```diff
@@ -3,19 +3,26 @@
...
 	if ( ! hasFinishedResolution ) {
 		return <SpinnerCard />;
 	}
 
-	if ( ! isPreconditionReady ) {
+	if (
+		! isPreconditionReady ||
+		! googleAdsAccount ||
+		googleAdsAccount.status === GOOGLE_ADS_ACCOUNT_STATUS.DISCONNECTED
+	) {
```
joemcgill: We want to allow for these connections to be made in either order, so these changes to this file need to be removed.

joemcgill (dismissed):
The updated E2E tests are looking good, though I've noticed that they seem a bit flaky when I run them locally—sometimes stalling out when trying to find the connect account button, though I don't see any reason for that. Perhaps it's something that was already happening when the Google Ads tests were in step 4.

The main things that I think still need to be adjusted:

1. The MC modal should be able to be connected before the Ads account, so I've suggested a change to that component inline so that it is enabled even when the Ads account is not connected.
2. According to the [Figma designs](https://www.figma.com/file/fODAWLBsbvd2MVmdh1Hgk3/Google-Listings-%26-Ads---Alternative-Conversion-Tracking-Flow?type=design&node-id=616-11376&mode=design&t=xwaqZBv3fDi4KBxn-0) and the note about "Users should be able to clearly understand what each step is doing" in the requirements, the copy on the Ads and MC cards should be updated with the following:
  - Ads: “Required to set up conversion measurement and create campaigns. You can also create or connect an account later.”
  - MC: "Required to sync products and list on Google."
3. In addition to the previous content changes, according to the Figma design, after the Ads connection is made, there should be a green success message that reads "Conversion measurement has been set up. You can create a campaign later". Conversion tracking is handled as a part of the Ads connection process ([here](https://github.com/woocommerce/google-listings-and-ads/blob/4a821a7a6b4a78179e0bef3d6897475c4371a746/src/Ads/AccountService.php#L276)).

src/API/Google/Merchant.php
```diff
@@ -316,6 +316,10 @@ public function link_ads_id( int $ads_id ): bool {
 		$account   = $this->get_account();
 		$ads_links = $account->getAdsLinks();
 
+		if ( ! is_array( $ads_links ) ) {
```
joemcgill: @mikkamp while working on this feature, we ran into a bug here because it looks like there are cases where the `Google\Service\ShoppingContent\Account` object returned when you run `$this->get_accounts()` above can contain a `null` value for the `$adsLinks` property, which then throws an error below. 

This may have happened due to the fact that I've created multiple MC accounts under the same domain while testing. Either way, adding this bit of code fixes the issue here, but unsure if a more general fix is needed. Let me know if you have any questions or need more info.

mikkamp: Great catch @joemcgill although would you be able to provide a little more detail on how you are getting this issue. I'm not able to reproduce this locally, both with a MC account that never linked an Ads account and an account where previous links have been removed/cancelled I'm still unable to get it to return null from `getAdsLinks`.

What's happening here is I'm getting a REST response without the `adsLinks` field included, since the Account object is based off a [Model](https://github.com/googleapis/google-api-php-client/blob/v2.15.3/src/Model.php), when it calls mapTypes it doesn't set the `adsLinks` property (mapTypes is called from the constructor when the [REST class creates the object](https://github.com/googleapis/google-api-php-client/blob/v2.15.3/src/Http/REST.php#L144)). So when it calls `getAdsLinks` it calls the magic `__get` function from the Model class. Which as we [can see here](https://github.com/googleapis/google-api-php-client/blob/v2.15.3/src/Model.php#L70) will return an empty array based on the property `adsLinksDataType`.

It's no harm to have another check to set it to an empty array, but I'm just curious why that wouldn't have been reported before, as we have plenty of MC accounts linked to an Ads account with all kinds of configurations.

mikkamp: Also I saw the latest code was changed to doing a typecast:
```php
$ads_links = (array) $account->getAdsLinks();
```

If we are going to expect the function to return `null` in some cases then IMHO it would be better to use the null coalescing operator so if it does return an unexpected value it can catch and log it as an exception.
```diff
- $ads_links = (array) $account->getAdsLinks();
+ $ads_links = $account->getAdsLinks() ?? [];
```

joemcgill:
This PR should not be merged until after #2247 is addressed, but as this is starting to become a large PR, I would appreciate feedback from @mikkamp on the current approach.

src/API/Google/Merchant.php
```diff
@@ -314,7 +314,7 @@ public function update_account( Account $account ): Account {
 	 */
 	public function link_ads_id( int $ads_id ): bool {
 		$account   = $this->get_account();
-		$ads_links = $account->getAdsLinks();
+		$ads_links = (array) $account->getAdsLinks();
```
mikkamp: I'd suggest to use the null coalescing operator to be a little more explicit that we allow a list of links or a null value. Typecasting does pretty much the same but it will also hide any other unexpected values as an array. If the function isn't returning a list of links like we expect, it should throw an error early.

```suggestion
		$ads_links = $account->getAdsLinks() ?? [];
```

src/Ads/AccountService.php
```diff
@@ -252,17 +252,18 @@ private function check_billing_status( array $account ) {
...
-			throw new Exception( 'A Merchant Center account must be connected' );
-		}
-
 		if ( ! $this->options->get_ads_id() ) {
 			throw new Exception( 'An Ads account must be connected' );
 		}
 
+		if ( ! $this->options->get_merchant_id() ) {
+			// Return early if the merchant account hasn't been created yet.
+			return;
```
mikkamp: If we return early here then the AdsAccountState will mark the step `link_merchant` as completed. I see that the MerchantAccountState does the opposite of this for the new step `link_ads`. And the only thing they rely on is the ID of the opposite account to allow the linking to happen.

Maybe it's a little early and out of scope for this PR but eventually we should only link if the state of the other account has reached a certain step, for example step `link_ads` can only be completed if ads steps `set_id` and `access_account` have been completed.

So this makes me wonder if we should make this part a bit more flexible and define a fourth state:
```php
	public const STEP_SKIPPED = 2;
```

We'll need to change several comparisons from `$step === STEP_DONE` to `$step >= STEP_DONE` but it would give us some more flexibility into knowing which steps were completed and which ones were skipped/optional.

src/MerchantCenter/AccountService.php
```diff
@@ -469,6 +473,26 @@ private function maybe_add_merchant_center_url( int $merchant_id ) {
...
+	 * @throws Exception When the merchant account hasn't been set yet.
+	 */
+	private function link_ads_account() {
+		if ( ! $this->options->get_merchant_id() ) {
+			throw new Exception( 'A Merchant Center account must be connected' );
+		}
+
+		if ( ! $this->options->get_ads_id() ) {
+			// Return early if the ads account hasn't been created yet.
+			return;
```
mikkamp: Same as previous comment, do we need to mark this as STEP_SKIPPED?

src/MerchantCenter/AccountService.php
```diff
@@ -469,6 +473,26 @@ private function maybe_add_merchant_center_url( int $merchant_id ) {
...
+		}
+
+		// Create link for Merchant and accept it in Ads.
+		$this->container->get( Merchant::class )->link_ads_id( $this->options->get_ads_id() );
+		$this->container->get( Ads::class )->accept_merchant_link( $this->options->get_merchant_id() );
```
mikkamp, line 501: For easy reference we add any class that's fetched from the container in the comment at the top of the class. This helps us keep track of the list of dependencies (which aren't explicitly passed in the constructor). Can we add the `Ads` class to that list?

```php
 * Container used to access:
 * - CleanupSyncedProducts
 * - Merchant
 * - MerchantAccountState
 * - MerchantCenterService
 * - MerchantIssueTable
 * - MerchantStatuses
 * - Middleware
 * - SiteVerification
 * - ShippingRateTable
 * - ShippingTimeTable
```

tests/Unit/MerchantCenter/AccountServiceTest.php
```diff
@@ -591,6 +598,77 @@ public function test_switch_url() {
...
+
+		$this->state->expects( $this->once() )
+			->method( 'get' )
+			->willReturn(
+				[
+					'link_ads' => [ 'status' => MerchantAccountState::STEP_PENDING ],
+				]
+			);
+
+		$this->assertEquals( self::TEST_ACCOUNT_DATA, $this->account->setup_account( self::TEST_ACCOUNT_ID ) );
```
mikkamp: In this test we call `setup_account` with the step `link_ads` marked as STEP_PENDING, but we don't actually have any expectations to confirm it returned early or performed the actions anyway. Can we either check that the step `link_ads` has a new status? Or if we expect it to return early can we confirm that the functions `link_ads_id` and `accept_merchant_link` are never called?

```php
$this->merchant->expects( $this->never() )->method( 'link_ads_id' );
$this->ads->expects( $this->never() )->method( 'accept_merchant_link' );
```

tests/Unit/Ads/AccountServiceTest.php
```diff
@@ -382,10 +386,7 @@ public function test_setup_account_step_link_merchant_no_merchant_id() {
 				]
 			);
 
-		$this->expectException( Exception::class );
-		$this->expectExceptionMessage( 'A Merchant Center account must be connected' );
-
-		$this->account->setup_account();
+		$this->assertEquals( [ 'id' => self::TEST_ACCOUNT_ID ], $this->account->setup_account() );
```
mikkamp, line 486: Same as previous comment can we check for a status change, or that the functions for linking were never called.

mikkamp (commented):
Thanks for all the changes here. I've only gone over the PHP changes and excluded the scenario for new accounts where the billing status can't be completed.

In general everything is working as expected, I just left some comments about improvements we could make to help us give a better reflection of the actual state of each connected account.

js/src/setup-mc/setup-stepper/setup-accounts/index.js
```diff
@@ -129,6 +134,9 @@ const SetupAccounts = ( props ) => {
 				<VerticalGapLayout size="large">
 					<WPComAccountCard jetpack={ jetpack } />
 					<GoogleAccountCard disabled={ isGoogleAccountDisabled } />
+					<GoogleAdsAccountCard
+						disabled={ isGoogleAccountDisabled }
```
eason9487: The `disabled` is not a prop of `GoogleAdsAccountCard` and won't work. It should not need to be added.

js/src/components/google-mc-account-card/non-connected.js
```diff
@@ -17,7 +17,7 @@ const NonConnected = () => {
 		return <SpinnerCard />;
 	}
 
-	if ( existingAccounts.length > 0 ) {
+	if ( existingAccounts && existingAccounts.length > 0 ) {
```
eason9487: Wonder why this change is needed.

js/src/components/google-ads-account-card/google-ads-account-card.js
```diff
@@ -26,6 +31,18 @@ export default function GoogleAdsAccountCard() {
 	}
 
 	return (
-		<ConnectedGoogleAdsAccountCard googleAdsAccount={ googleAdsAccount } />
+		<ConnectedGoogleAdsAccountCard googleAdsAccount={ googleAdsAccount }>
+			{ googleAdsAccount.status ===
+				GOOGLE_ADS_ACCOUNT_STATUS.CONNECTED && (
+				<Notice status="success" isDismissible={ false }>
+					<p>
```
eason9487: This `<p>` layer could be omitted so that it won't need to reset `p`'s margin in the *js/src/components/account-card/index.scss* file.

js/src/components/google-ads-account-card/google-ads-account-card.js
```diff
@@ -14,7 +19,7 @@ export default function GoogleAdsAccountCard() {
 	const { googleAdsAccount } = useGoogleAdsAccount();
 
 	if ( ! google || ! googleAdsAccount ) {
-		return <SpinnerCard />;
+		return <NonConnected />;
```
eason9487: Returning `<NonConnected />` here is semantically confusing because the conditions are to check if the data are not yet fetched.

It's recommended to check in a similar way to [useGoogleMCAccount](https://github.com/woocommerce/google-listings-and-ads/blob/8597a8c34bd2e8342ffee3a63e959f4681578666/js/src/hooks/useGoogleMCAccount.js#L41) and [GoogleMCAccountCard](https://github.com/woocommerce/google-listings-and-ads/blob/8597a8c34bd2e8342ffee3a63e959f4681578666/js/src/components/google-mc-account-card/google-mc-account-card.js#L11-L16) and still return `<SpinnerCard />`

js/src/components/google-ads-account-card/non-connected.js
```diff
@@ -6,28 +6,35 @@ import { useState } from '@wordpress/element';
...
-import useExistingGoogleAdsAccounts from '.~/hooks/useExistingGoogleAdsAccounts';
 import SpinnerCard from '.~/components/spinner-card';
+import CreateAccount from './create-account';
+import useGoogleAccountCheck from '.~/hooks/useGoogleAccountCheck';
 import ConnectAds from './connect-ads';
 
 const NonConnected = () => {
-	const { existingAccounts } = useExistingGoogleAdsAccounts();
+	const { google, existingAccounts, hasFinishedResolution } =
+		useGoogleAccountCheck();
```
eason9487: Curious about why there is a need to combine `useGoogleAccount` and `useExistingGoogleAdsAccounts` into a new hook with a Google account status check?

Is it possible to use only the existing `useGoogleAccount` and `useExistingGoogleAdsAccounts` with [the Google status check below](https://github.com/woocommerce/google-listings-and-ads/blob/8597a8c34bd2e8342ffee3a63e959f4681578666/js/src/components/google-ads-account-card/non-connected.js#L32) to accomplish the same result?

js/src/components/google-ads-account-card/create-account.js
```diff
@@ -49,26 +51,38 @@ const ClaimTermsAndCreateAccountButton = () => {
...
 
 	return (
 		<AccountCard
+			disabled={ disabled }
 			appearance={ APPEARANCE.GOOGLE_ADS }
 			alignIcon="top"
-			indicator={ <ClaimTermsAndCreateAccountButton /> }
+			indicator={
+				<ClaimTermsAndCreateAccountButton disabled={ disabled } />
+			}
```
eason9487: Considering the `ClaimTermsAndCreateAccountButton` is an internal component of `CreateAccount` and there is no need to render the button when the incoming `disabled` prop is true, using `indicator={ disabled ? null : <ClaimTermsAndCreateAccountButton /> }` should get the same UI result, right?

If so, the button would not need to be aware of the statuses of Google account and `disabled`.

https://github.com/woocommerce/google-listings-and-ads/blob/8597a8c34bd2e8342ffee3a63e959f4681578666/js/src/components/google-ads-account-card/create-account.js#L54-L56

https://github.com/woocommerce/google-listings-and-ads/blob/8597a8c34bd2e8342ffee3a63e959f4681578666/js/src/components/google-ads-account-card/create-account.js#L62

js/src/components/google-mc-account-card/disabled-card.js
```diff
@@ -8,16 +8,22 @@ import { __ } from '@wordpress/i18n';
...
 			indicator={
-				<AppButton isSecondary disabled>
-					{ __( 'Create account', 'google-listings-and-ads' ) }
-				</AppButton>
+				google &&
+				google.active === 'yes' && (
+					<AppButton isSecondary disabled>
+						{ __( 'Create account', 'google-listings-and-ads' ) }
+					</AppButton>
+				)
```
eason9487: There is no case for the `DisabledCard` component to display this button with the disabled state, so this `indicator` could be removed entirely.

js/src/components/google-mc-account-card/create-account-button.js
```diff
@@ -23,7 +23,9 @@ const CreateAccountButton = ( props ) => {
 	const { onCreateAccount = noop, onClick = noop, ...rest } = props;
 	const [ activeModal, setActiveModal ] = useState( MODALS.NONE );
 	const { data: existingAccounts } = useExistingGoogleMCAccounts();
-	const matchingDomainAccount = getMatchingDomainAccount( existingAccounts );
+	const matchingDomainAccount = getMatchingDomainAccount(
+		existingAccounts ?? []
+	);
```
eason9487: Wonder why this change is needed.

eason9487 (commented):
Thanks for the work.

I've reviewed the frontend changes and left some suggestions or questions. In addition, there are three other issues:

1. When entering the onboarding, it will [initialize with the unfinished step](https://github.com/woocommerce/google-listings-and-ads/blob/8597a8c34bd2e8342ffee3a63e959f4681578666/js/src/setup-mc/setup-stepper/index.js#L34) fetched from [the `mc/setup` API](https://github.com/woocommerce/google-listings-and-ads/blob/8597a8c34bd2e8342ffee3a63e959f4681578666/src/API/Site/Controllers/MerchantCenter/AccountController.php#L102-L111). Although this is not a common operation step, it may still result in jumping straight to step 2, 3 or 4. Steps to cause a problem:
   1. Proceed the onboarding to step 2, 3 or 4.
   2. Back to step 1 and disconnect Google Ads account by clicking the "Or, connect to a different Google Ads account" button.
   3. Refresh the webpage.

   https://github.com/woocommerce/google-listings-and-ads/assets/17420811/1e592a72-76bf-4695-a6a1-b1d46d19b656
2. Some E2E tests failed. (I haven't checked the changes to the E2E test yet as I would like to clarify the code review questions first.)
   ```
   3 failed
     [chromium] › setup-mc/step-1-accounts.test.js:48:2 › Set up accounts › should see accounts step header, "Connect your WordPress.com account" & connect button 
     [chromium] › setup-mc/step-3-confirm-store-requirements.test.js:508:5 › Confirm store requirements › Pre-Launch Checklist › When the policy checks are compliant and settings are true › should see the heading of next step after clicking "Continue" 
     [chromium] › setup-mc/step-4-complete-campaign.test.js:131:2 › Complete your campaign › should see the heading and the texts below 
   59 skipped
   23 passed (1.6m)
   ```
3. Looks like this copy change was missed.
   image

src/API/Google/Merchant.php
ankitrox: This has been addressed.

src/Ads/AccountService.php
ankitrox: @mikkamp I think as of now we can skip adding a new flag

```php
public const STEP_SKIPPED = 2;
```

Introducing the new flag would require some checks to perform at several places and also make changes to several places including unit tests, it will also open some possibility to introduce any possible issue.

Instead, we can call `link_merchant_account` conditionally if the `ads_id` and `merchant_id` is present. This change is introduced for both AccountService.php for `Ads` and `Merchant` classes.

src/MerchantCenter/AccountService.php
ankitrox: Added the `Ads` class in docblock.

src/MerchantCenter/AccountService.php
ankitrox: Responded [here](https://github.com/woocommerce/google-listings-and-ads/pull/2205#discussion_r1500782920)

tests/Unit/Ads/AccountServiceTest.php
ankitrox: This is passing now. Explained [here](https://github.com/woocommerce/google-listings-and-ads/pull/2205#discussion_r1500782920)

mikkamp: As clarified in https://github.com/woocommerce/google-listings-and-ads/pull/2205#discussion_r1489567034

It's not so much about the test passing or not. It's that the assertions in the test do not fully check the expected behaviour. For example:

1. We mock `get_ads_id` to always return a test ID
2. We mock `get_merchant_id` to always return a 0
3. We mock the state with the step `link_merchant` to be pending
4. We check that `setup_account` returns the matching test ID we mocked in step 1

However what we don't test in this scenario, is if it actually attempted to link or not. If we look at the actual function:
```php
	private function link_merchant_account() {
		if ( ! $this->options->get_ads_id() ) {
			throw new Exception( 'An Ads account must be connected' );
		}

		// Create link for Merchant and accept it in Ads.
		$this->container->get( Merchant::class )->link_ads_id( $this->options->get_ads_id() );
		$this->container->get( Ads::class )->accept_merchant_link( $this->options->get_merchant_id() );
	}
```
With the input we provided of a test Ads ID and a 0 Merchant ID, it would try to link them. The call to the function gets prevented higher up, but we aren't actually confirming in this test, that's what happened. So we need to confirm that those functions were never called, which we can do with something like:

```php
$this->merchant->expects( $this->never() )->method( 'link_ads_id' );
$this->ads->expects( $this->never() )->method( 'accept_merchant_link' );
```

Or alternatively we need to check for a status change, which might be harder to test.

Same goes for both the tests:
- test_setup_account_step_link_merchant_no_ads_id
- test_setup_account_step_link_merchant_no_merchant_id

If they aren't going to throw an exception like they previously did then they need to be testing something valid, not just passing a test.

src/Ads/AccountService.php
mikkamp: Thanks for clarifying, only working with the available STEP statuses sounds like a good approach.

tests/Unit/Ads/AccountServiceTest.php
ankitrox: @mikkamp I agree with your point. I have made the changes to assert that the `link_ads_id ` and `accept_merchant_link ` methods are never called when return early.

mikkamp: Thanks for taking a look @ankitrox 
I tried testing those changes and it seems there is still some room for improvement.

I'm wondering why the following code was removed:
```php
		$this->state->expects( $this->once() )
			->method( 'get' )
			->willReturn(
				[
					'link_ads' => [ 'status' => MerchantAccountState::STEP_PENDING ],
				]
			);
```

If we don't mock that state then it just returns an empty state array, so it won't perform any step. In a scenario like that it's obvious that it won't call any functions to link, but we do need it to actually handle the expected step, otherwise the test once again becomes invalid.
All the comments about these tests apply to both `Ads/AccountServiceTest.php` and `MerchantCenter/AccountServiceTest.php`, so they'll need to be refactored in both.

Also when running the tests I get the following two warnings:
```
Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in google-listings-and-ads/src/Ads/AccountService.php on line 175

Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in google-listings-and-ads/src/MerchantCenter/AccountService.php on line 343
```

If that code is intended to continue the `foreach` loop then `continue 2` must be used (otherwise it's just breaking out of the `switch` statement and will still set the step to completed). Although to avoid confusion I'd suggest to add a comment there to explain what the expected behaviour is. Also would like to point out that setting the step to STEP_PENDING without calling `$this->state->update( $state )` does nothing as the state won't be stored for the next time it calls these steps (this is especially a problem if it's the last step it's handling).

joemcgill: @ankitrox I looked at the updates you made to the `Ads\AccountServiceTest::test_setup_account_step_link_merchant_no_merchant_id` method and they looked pretty good to me. For consistency, I went ahead and made the same changes to the other account linking tests in both Account Service test suites in [961d2d9](https://github.com/woocommerce/google-listings-and-ads/pull/2205/commits/961d2d96e1ca8918488b08a3f8755b2c1644969f) and also made sure in each case that we are confirming that the method that links the accounts is never called if one of the account IDs is not present.

Now that we're also not calling the `link_merchant_account` and `link_ads_account` methods from the account setup steps if both IDs are not present, I think we can also reinstate exceptions in those methods whenever both are not present, which is how the original method used to work. I went ahead and made that update [b71e497](https://github.com/woocommerce/google-listings-and-ads/pull/2205/commits/b71e49725d15239a409329d25987e8f82c223d59). What do you think?

ankitrox:
@mikkamp Addressed your feedback in the PR and also responded to the questions. Requesting you to please re-review the changes. Thanks

CC: @joemcgill

src/Ads/AccountService.php
```diff
@@ -252,20 +267,20 @@ private function check_billing_status( array $account ) {
...
 			throw new Exception( 'An Ads account must be connected' );
 		}
 
+		$mc_state = $this->container->get( MerchantAccountState::class );
+
 		// Create link for Merchant and accept it in Ads.
 		$this->container->get( Merchant::class )->link_ads_id( $this->options->get_ads_id() );
 		$this->container->get( Ads::class )->accept_merchant_link( $this->options->get_merchant_id() );
+
+		$mc_state->complete_step( 'link_ads' );
```
mikkamp, line 359: Minor comments: Can we get the MerchantAccountState added to the top as a dependency? Can we also combine these two state lines together, if one of the linking steps fails we don't modify the state, so no need to fetch the dependency yet.

Just to note this code relies on the MerchantAccountState already being initialized, which is fine in this case because we always initialize before storing the merchant ID.

src/MerchantCenter/AccountService.php
```diff
@@ -469,6 +489,25 @@ private function maybe_add_merchant_center_url( int $merchant_id ) {
...
+			throw new Exception( 'A Merchant Center account must be connected' );
+		}
+
+		$ads_state = $this->container->get( AdsAccountState::class );
+
+		// Create link for Merchant and accept it in Ads.
+		$this->container->get( Merchant::class )->link_ads_id( $this->options->get_ads_id() );
+		$this->container->get( Ads::class )->accept_merchant_link( $this->options->get_merchant_id() );
+
+		$ads_state->complete_step( 'link_merchant' );
```
mikkamp, line 503: Same minor comments here, can we add the dependencies at the top and keep the two state lines together?

src/Ads/AccountService.php
```diff
@@ -160,14 +161,28 @@ public function setup_account(): array {
...
+							 * Mark the step as pending, save the state and
+							 * continue the foreach loop with `continue 2`.
+							 *
+							 * Marking the step as pending will make sure to run this step next time.
+							 */
+							$state[ $name ]['status'] = AdsAccountState::STEP_PENDING;
+							$this->state->update( $state );
+							continue 2;
+						}
+						break;
```
mikkamp: This pattern is still confusing, which is made obvious by `test_setup_account_step_link_merchant_no_ads_id`. When we don't have a Merchant ID we want to hit the else condition and set the state as pending, but when we are setting up an ads account without an ads ID, that's a clear error scenario. Since we already have that being thrown within `link_merchant_account`, I'd suggest to refactor this code to make it flow better (so we need less comments to explain what's happening):

```suggestion
					case 'link_merchant':
						// Continue to next step if the MC account is not connected yet.
						if ( ! $this->options->get_merchant_id() ) {
							// Save step as pending and continue the foreach loop with `continue 2`.
							$state[ $name ]['status'] = AdsAccountState::STEP_PENDING;
							$this->state->update( $state );
							continue 2;
						}

						$this->link_merchant_account();
						break;
```

I've also removed the empty check since `get_merchant_id` will always return an `int`. As a followup to this we'll need to revert the test `test_setup_account_step_link_merchant_no_ads_id`

tests/Unit/Ads/AccountServiceTest.php
```diff
@@ -357,35 +361,43 @@ public function test_setup_account_step_link_merchant_no_ads_id() {
...
-		$this->expectExceptionMessage( 'An Ads account must be connected' );
+		$this->state->expects( $this->once() )
+			->method( 'update' )
+			->with( $ads_account_state );
 
-		$this->account->setup_account();
+		$this->middleware->expects( $this->never() )
+			->method( 'link_ads_account' );
+
+		$this->assertEquals( [ 'id' => 0 ], $this->account->setup_account() );
```
mikkamp: As suggested in the comment about refactoring, we can get this test to revert again to the way it was before this PR where it expects an error:

```php
	public function test_setup_account_step_link_merchant_no_ads_id() {
		$this->options->expects( $this->any() )
			->method( 'get_ads_id' )
			->willReturn( 0 );

		$this->options->expects( $this->any() )
			->method( 'get_merchant_id' )
			->willReturn( self::TEST_MERCHANT_ID );

		$this->state->expects( $this->once() )
			->method( 'get' )
			->willReturn(
				[
					'link_merchant' => [ 'status' => AdsAccountState::STEP_PENDING ],
				]
			);

		$this->expectException( Exception::class );
		$this->expectExceptionMessage( 'An Ads account must be connected' );

		$this->account->setup_account();
	}
```

src/MerchantCenter/AccountService.php
```diff
@@ -330,6 +333,23 @@ private function setup_account_steps() {
...
+							 * Mark the step as pending, save the state and
+							 * continue the foreach loop with `continue 2`.
+							 *
+							 * Marking the step as pending will make sure to run this step next time.
+							 */
+							$state[ $name ]['status'] = MerchantAccountState::STEP_PENDING;
+							$this->state->update( $state );
+							continue 2;
+						}
+						break;
```
mikkamp: Same as the previous suggestion, we can simplify this code to make it flow better and still have it return an error if the merchant account isn't set:

```suggestion
					case 'link_ads':
						// Continue to next step if Ads account is not connected yet.
						if ( ! $this->options->get_ads_id() ) {
							// Save step as pending and continue the foreach loop with `continue 2`.
							$state[ $name ]['status'] = MerchantAccountState::STEP_PENDING;
							$this->state->update( $state );
							continue 2;
						}

						$this->link_ads_account();
						break;
```

js/src/setup-mc/setup-stepper/setup-accounts/index.js
asvinb: Removed @eason9487

js/src/components/google-mc-account-card/non-connected.js
asvinb: Reverted the change 👍

js/src/components/google-ads-account-card/google-ads-account-card.js
asvinb: 👍

js/src/components/google-ads-account-card/google-ads-account-card.js
asvinb: Updated @eason9487

js/src/components/google-ads-account-card/non-connected.js
asvinb: Sure, updated @eason9487

js/src/components/google-ads-account-card/create-account.js
asvinb: 👍

js/src/components/google-mc-account-card/disabled-card.js
asvinb: 👍

js/src/components/google-mc-account-card/create-account-button.js
asvinb: Reverted the change, unless there was something specific that I might be missing @ankitrox ?

tests/Unit/MerchantCenter/AccountServiceTest.php
```diff
@@ -591,6 +598,88 @@ public function test_switch_url() {
...
+
+		$this->state->expects( $this->once() )
+			->method( 'update' )
+			->with( $mc_account_state );
+
+		$this->middleware->expects( $this->never() )
+			->method( 'link_ads_account' );
+
+		$this->assertEquals( [ 'id' => 0 ], $this->account->setup_account( self::TEST_ACCOUNT_ID ) );
+	}
```
mikkamp: If we simplify the code to return an exception then this test can also be simplified and made more logical:

```suggestion
	public function test_setup_account_step_link_ads_without_mc() {
		$this->options->expects( $this->any() )
			->method( 'get_ads_id' )
			->willReturn( self::TEST_ADS_ID );

		$this->options->expects( $this->any() )
			->method( 'get_merchant_id' )
			->willReturn( 0 );

		$this->state->expects( $this->once() )
			->method( 'get' )
			->willReturn(
				[
					'link_ads' => [ 'status' => MerchantAccountState::STEP_PENDING ],
				]
			);

		$this->expectException( Exception::class );
		$this->expectExceptionMessage( 'A Merchant Center account must be connected' );

		$this->account->setup_account( self::TEST_ACCOUNT_ID );
	}
```

tests/Unit/MerchantCenter/AccountServiceTest.php
```diff
@@ -591,6 +598,88 @@ public function test_switch_url() {
 		$this->assertEquals( self::TEST_ACCOUNT_DATA, $this->account->switch_url( self::TEST_ACCOUNT_ID ) );
 	}
 
+	public function test_setup_account_step_link_ads() {
```
mikkamp, line 607: This test now fails, and it's the same for the `Ads/AccountServiceTest::test_setup_account_step_link_merchant`.

Because Ads/AccountService updates MerchantAccountState, and MerchantCenter/AccountService updates AdsAccountState. Those dependencies were pulled from the container class. Which is precisely why we have the comment at the top of the file so when we write the unit test we can see which dependencies are required. In this case we need to add each of those state dependencies to the unit tests.

It's also something which is indicated in the unit test run here: https://github.com/woocommerce/google-listings-and-ads/actions/runs/8261130029/job/22597836016?pr=2205

src/Ads/AccountService.php
```diff
@@ -15,6 +15,7 @@
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\MerchantAccountState;
```
mikkamp: Just a minor comment, but we like to keep these in alphabetical order.

src/MerchantCenter/AccountService.php
```diff
@@ -20,6 +21,7 @@
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\AdsAccountState;
```
mikkamp: Just a minor comment, but we like to keep these in alphabetical order.

mikkamp (commented):
Thanks @ankitrox for all the changes, there are some good improvements in there. Some of the state handling code is still a little confusing though so I offered suggestions on how we can simplify that flow, so the flow/unit tests can make a little more sense.

It would also be helpful to check the workflow runs for any failing unit tests as it would point out some of the issues. For the rest the comments are just minor little things to keep consistent.

Note: I've again only reviewed the PHP parts of the code.

tests/Unit/MerchantCenter/AccountServiceTest.php
ankitrox: This is fixed in [this change](https://github.com/woocommerce/google-listings-and-ads/pull/2205/commits/50a1cb245e34c271761a7745239541fe2adf014e).

ankitrox:
> Thanks @ankitrox for all the changes, there are some good improvements in there. Some of the state handling code is still a little confusing though so I offered suggestions on how we can simplify that flow, so the flow/unit tests can make a little more sense.
> 
> It would also be helpful to check the workflow runs for any failing unit tests as it would point out some of the issues. For the rest the comments are just minor little things to keep consistent.
> 
> Note: I've again only reviewed the PHP parts of the code.

@mikkamp Thank you very much for your review. I have accepted your suggestions in PR and also made the changes as advised.

mikkamp (approved):
I had a look and the PHP part is flowing much better now. I just added one last commit https://github.com/woocommerce/google-listings-and-ads/pull/2205/commits/9911323046463d174184e9af531e064f1173c5b9 because I think doubling up with a mocked state might give someone trouble in the future even though it doesn't cause a failing test now.

So I'm going to go ahead and approve the PHP part of the code.

js/src/components/google-ads-account-card/google-ads-account-card.js
```diff
@@ -10,22 +16,37 @@ import AuthorizeAds from './authorize-ads';
...
+	const {
+		google,
+		scope,
+		isResolving: isResolvingGoogleAccount,
+	} = useGoogleAccount();
+	const { googleAdsAccount, isResolving: isResolvingGoogleAdsAccount } =
+		useGoogleAdsAccount();
 
-	if ( ! google || ! googleAdsAccount ) {
+	if ( isResolvingGoogleAccount || isResolvingGoogleAdsAccount ) {
```
eason9487: When it's possible, it would be appreciated if it could avoid using `isResolving`.

js/src/components/google-ads-account-card/create-account.js
```diff
@@ -58,17 +58,26 @@ const ClaimTermsAndCreateAccountButton = () => {
 };
 
 const CreateAccount = ( props ) => {
-	const { allowShowExisting, onShowExisting } = props;
+	const { allowShowExisting, onShowExisting, disabled } = props;
 
 	return (
 		<AccountCard
+			disabled={ disabled }
```
eason9487: This disabled status doesn't work.

image

Adding a `DisabledCard` to separate out this status would be easier than handling it here. Ref: 
https://github.com/woocommerce/google-listings-and-ads/pull/2270#discussion_r1521346100

js/src/components/google-ads-account-card/create-account.js
```diff
@@ -58,17 +58,26 @@ const ClaimTermsAndCreateAccountButton = () => {
...
 			appearance={ APPEARANCE.GOOGLE_ADS }
 			alignIcon="top"
-			indicator={ <ClaimTermsAndCreateAccountButton /> }
+			indicator={
+				disabled ? null : (
+					<ClaimTermsAndCreateAccountButton disabled={ disabled } />
```
eason9487: If it still wants to handle the disabled status here, then the `disabled` prop for `ClaimTermsAndCreateAccountButton` is not needed.

js/src/components/google-ads-account-card/google-ads-account-card.js
eason9487: An uncaught error occurs and results in a blank page when disconnecting Google Ads account. I believe it's also related to status checks like this one.

```
Uncaught TypeError: Cannot read properties of null (reading 'status')
    at GoogleAdsAccountCard (google-ads-account-card.js:41:23)
    at renderWithHooks (react-dom.js?ver=18.2.0:16315:20)
    at updateFunctionComponent (react-dom.js?ver=18.2.0:19598:22)
    at beginWork (react-dom.js?ver=18.2.0:21611:18)
    at HTMLUnknownElement.callCallback (react-dom.js?ver=18.2.0:4151:16)
    at Object.invokeGuardedCallbackDev (react-dom.js?ver=18.2.0:4200:18)
    at invokeGuardedCallback (react-dom.js?ver=18.2.0:4264:33)
    at beginWork$1 (react-dom.js?ver=18.2.0:27461:9)
    at performUnitOfWork (react-dom.js?ver=18.2.0:26567:14)
    at workLoopSync (react-dom.js?ver=18.2.0:26476:7)
```

js/src/components/google-ads-account-card/non-connected.js
```diff
@@ -6,28 +6,37 @@ import { useState } from '@wordpress/element';
...
 import CreateAccount from './create-account';
+import useGoogleAccount from '.~/hooks/useGoogleAccount';
 import useExistingGoogleAdsAccounts from '.~/hooks/useExistingGoogleAdsAccounts';
-import SpinnerCard from '.~/components/spinner-card';
 import ConnectAds from './connect-ads';
 
 const NonConnected = () => {
-	const { existingAccounts } = useExistingGoogleAdsAccounts();
+	const { google, hasFinishedResolution } = useGoogleAccount();
+	const { existingAccounts, isResolving } = useExistingGoogleAdsAccounts();
```
eason9487: When it's possible, it would be appreciated if it could avoid using `isResolving`.

js/src/components/google-ads-account-card/non-connected.js
```diff
@@ -6,28 +6,37 @@ import { useState } from '@wordpress/element';
...
 		setIgnoreExisting( false );
 	};
 
-	if ( existingAccounts.length === 0 || ignoreExisting ) {
+	if ( ! hasFinishedResolution || isResolving ) {
+		return <SpinnerCard />;
+	}
+
+	if (
+		! existingAccounts ||
```
eason9487: This check won't work well due the the use of `isResolving` above. The concern can be inspected by adding `debugger;` at line 62:

https://github.com/woocommerce/google-listings-and-ads/blob/9911323046463d174184e9af531e064f1173c5b9/js/src/components/google-ads-account-card/create-account.js#L60-L62

The video shows the `<CreateAccount>` component is initially rendered at some points, but it turns out there are some existing accounts. This issue is similar to the example in the main comment of https://github.com/woocommerce/google-listings-and-ads/pull/2270#pullrequestreview-1927576354.

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/113799c3-8e6a-43dc-ada0-4b25eddc742c

js/src/setup-mc/setup-stepper/setup-paid-ads/setup-paid-ads.js
```diff
@@ -163,16 +161,14 @@ export default function SetupPaidAds() {
...
+					'Create a campaign to advertise your products',
 					'google-listings-and-ads'
 				) }
 				description={ __(
-					'As soon as your products are approved, your listings and ads will be live. In the meantime, let’s set up your ads.',
+					'You’re ready to set up a Performance Max campaign to drive more sales with ads. Your products will be included in the campaign after they’re approved.',
 					'google-listings-and-ads'
 				) }
 			/>
-			<ProductFeedStatusSection />
```
eason9487, line 174: There were some codes used by the `ProductFeedStatusSection` component could be removed from the codebase since they are no longer in use.

The `useSyncableProductsCalculation` hook used in the `GetStartedPage` component can be removed as well because it triggers the calculation for the number of syncable products and later gets the number from `ProductFeedStatusSection`.

eason9487 (commented):
Thanks for the updates.

I've reviewed the frontend changes for the second round but there are a few blockers. There are also other issues need to be addressed:

1. After connecting a Google Merchant Center account first without a connected Google Ads account, the UI still asks for connecting Google Merchant Center.
   
2. The three issues in [my first round](https://github.com/woocommerce/google-listings-and-ads/pull/2205#pullrequestreview-1887765969) are not addressed.
   1. See the vidoe

      https://github.com/woocommerce/google-listings-and-ads/assets/17420811/177051c5-5ac6-4a75-a6c4-1636e8a0a055
   2. Apart from the GTag tests, [some E2E tests still failed](https://github.com/woocommerce/google-listings-and-ads/actions/runs/8277992316/job/22649450574).
   3. Doesn't the copy need updating?

joemcgill:
Thanks for re-reviewing, @eason9487. Given how much of the related JS code is being refactored in #2270, we plan to address this feedback as part of that PR rather than making additional changes here.

js/src/components/google-ads-account-card/create-account.js
asvinb: @eason9487 The `DisabledCard` component added in the R-1.3 PR: https://github.com/woocommerce/google-listings-and-ads/pull/2270/files#diff-cdcc611dbe20591dfed6c352e68b20da4d4c7ead088168c10af0d0b879d5bc9eR1

js/src/components/google-ads-account-card/create-account.js
asvinb: @eason9487 This has been taken care of here:
https://github.com/woocommerce/google-listings-and-ads/pull/2270/files#diff-71265b178122038f740e17dd2ceba431a4d3828e9d6843a29ca73478fbc124eeR136-R138

js/src/components/google-ads-account-card/google-ads-account-card.js
asvinb: Yes, that's fixed here:
https://github.com/woocommerce/google-listings-and-ads/pull/2270/files#diff-bd272f1d34592b741f7c8eaf92dede6d3d90942964c73defd66d1ffe2c5b538bR37-R41

asvinb: @eason9487 I tried on the R1.3 PR and I could not reproduce. Can you kindly check on your end please?
https://github.com/woocommerce/google-listings-and-ads/pull/2270

Thanks!

js/src/components/google-ads-account-card/non-connected.js
asvinb: Yes, this has been removed here:
https://github.com/woocommerce/google-listings-and-ads/pull/2270/files#diff-e24415381364935e7ba38003c1248099254e1666922243c7f400ca86f2b41acbR17
😁

asvinb:
@eason9487 

1. I was able to connect my Merchant account without connecting the Ads account in the R1.3 [PR](https://github.com/woocommerce/google-listings-and-ads/pull/2270). See screenshot: image
2. For the other [3 issues you mentioned](https://github.com/woocommerce/google-listings-and-ads/pull/2205#pullrequestreview-1935734483):
    1. @ankitrox Can you [advise here please](https://github.com/woocommerce/google-listings-and-ads/pull/2205#pullrequestreview-1935734483)?
    2. @ankitrox Can you [advise here please](https://github.com/woocommerce/google-listings-and-ads/pull/2205#pullrequestreview-1935734483)?
    3. @joemcgill / @ankitrox  Can you [please advise](https://github.com/woocommerce/google-listings-and-ads/pull/2205#pullrequestreview-1935734483) about the copy change?

js/src/components/google-ads-account-card/non-connected.js
asvinb: @eason9487 This has been fixed here: https://github.com/woocommerce/google-listings-and-ads/pull/2270/files#diff-e24415381364935e7ba38003c1248099254e1666922243c7f400ca86f2b41acbR26-R33

js/src/components/google-ads-account-card/google-ads-account-card.js
eason9487: Commenting on the issue in this PR is because it happens with this PR, but it doesn't mean the issue also happens on the subsequent PR.

Since the merge base of this PR is `develop`, even though #2270 will not have the same issue, it doesn't mean that this PR with that issue is suitable to be merged directly to `develop` without fixing it.

js/src/setup-mc/setup-stepper/setup-paid-ads/setup-paid-ads.js
asvinb: @eason9487 Code has been removed in the following commit:
https://github.com/woocommerce/google-listings-and-ads/pull/2270/commits/1b79f1c0b21d94c37d144e7b8cf173c71a3f9a4e

eason9487:
Hi @asvinb 

> 1. I was able to connect my Merchant account without connecting the Ads account in the R1.3 [PR](https://github.com/woocommerce/google-listings-and-ads/pull/2270).

I guess you are referring to the first issue in https://github.com/woocommerce/google-listings-and-ads/pull/2205#pullrequestreview-1935734483. Please refer to this reply: https://github.com/woocommerce/google-listings-and-ads/pull/2205#discussion_r1526037608

eason9487:
Hi @joemcgill and @asvinb

> Given how much of the related JS code is being refactored in #2270, we plan to address this feedback as part of that PR rather than making additional changes here.

PRs merged into `develop` may be released with our weekly routine work, so it would be inappropriate to merge a PR without fixing known issues. In addition, each PR merged directly into `develop` is expected to work properly so that there will be less confusion for future developers when tracking the history of code changes.

To fix this PR's issues by PR #2270, may I ask you to branch off a new feature branch from the current `develop` and set it as the merge base of this PR?

joemcgill:
@eason9487

> PRs merged into develop may be released with our weekly routine work, so it would be inappropriate to merge a PR without fixing known issues. In addition, each PR merged directly into develop is expected to work properly so that there will be less confusion for future developers when tracking the history of code changes.

I agree with you completely, and was not suggesting that we merge this PR as is. Now that the API in this branch is approved, we've refreshed the branch for #2270 with those changes, and are finalizing changes to the UI code there so that you don't need to review stale code in this branch that has been updated or replaced in the other PR. Once the PR for #2270 is approved, our plan is to merge those changes into this branch so the whole feature can be completed and merged into `develop` with all feedback addressed (essentially, using this branch as the feature branch).

joemcgill (dismissed):
Marking this PR as blocked until all feedback has been addressed and merged from #2270, to avoid confusion.

joemcgill:
Now that #2270 is approved an merged, everything in this branch should be ready for final review/merge into `develop`. E2E tests for this whole branch have been updated in #2308. 

cc: @MatthiasReinholz

js/src/hooks/useUpsertAdsAccount.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useCallback, useState } from '@wordpress/element'; // Add this line
```
eason9487: This irrelevant code comment should be removed.
Ref: https://github.com/woocommerce/google-listings-and-ads/pull/2270#discussion_r1568525890

eason9487 (changes requested):

js/src/hooks/useUpsertAdsAccount.js
joemcgill: Handled in [65f240f](https://github.com/woocommerce/google-listings-and-ads/pull/2205/commits/65f240f0c4a504af0a116b0b7cfe098b9d3f60b4).

eason9487 (approved):

eason9487:
Change the merge base to the new feature branch `feature/seamlessly-conversion-tracking` as there will be other PRs to be added. Continuing to stack more PRs onto this PR is not ideal.

I will be merging this PR. The subsequent PRs will also use the `feature/seamlessly-conversion-tracking` as their merge base.

PR #2214, opened 2024-01-23
Product Block Editor: Add a custom block for editing channel visibility

eason9487:

To be compatible with Product Block Editor, this PR adds a custom block for editing channel visibility.

#### 📹 Channel visibility is disabled when the product is hidden from the product catalog

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/38fe3f30-12ff-4329-b23e-fb9ef2dfe852

1. Please refer to [the test preparation in PR 2151](https://github.com/woocommerce/google-listings-and-ads/pull/2151#prepare-test-environment) with **Woo 8.5.1**.
1. `npm install`
1. `npm start`

#### 📌 Test with simple product and disabled state

1. Change the selected channel visibility option and click the **Update** button at the top-right of the page.
1. Switch to the **Organization** tab.
1. Tick the "Hide in product catalog" option in the **Product catalog** section.
1. Click the **Update** button.
1. See if the channel visibility selection is disabled.
1. Switch to the **Organization** tab.
1. Uncheck the "Hide in product catalog" option.
1. Click the **Update** button.
1. See if the channel visibility selection is changed from disabled to enabled after updating.

💡 It must update the product data first because the `is_visible` value in the wp-data store won't be recalculated after changing the product's `catalog_visibility ` (the "Hide in product catalog" option) in the wp-data store. Instead, the `is_visible` value is only updated after the product update issues API request and syncs the response to the wp-data store.

#### 📌 Test with variable and variation products

1. Go to edit a variable product, e.g. the **Hoodie** imported from WC's sample products.
1. Change the selected channel visibility option and click the **Update** button.
1. Switch to the **Variations** tab.
1. Click one of the **Edit** links within the **Variations** section to edit a variation product.
1. Check if the channel visibility section is not shown.

#### 📌 Test the `has-errors` status

1. Go to edit a simple product.
1. Switch to the **Pricing** tab.
1. Clear the list price and sale price.
1. Enter an invalid GTIN such as `aaaa`.
1. Select "Sync and show" for the channel visibility.
1. Click the **Update** button.
1. Open the Connection Test: `/wp-admin/admin.php?page=connection-test-admin-page`.
1. Use the **Sync Product** to sync the above product
   - You might need to make the [MerchantCenterService::is_ready_for_syncing](https://github.com/woocommerce/google-listings-and-ads/blob/20169c47a3cfa2ef9c588ea69684688d4195c677/src/MerchantCenter/MerchantCenterService.php#L106) always return `true` if using a local test site.
1. Back to the product editor and refresh the webpage.
1. Check if it shows **Issues**.
   image

### DIFF
--- a/src/Admin/Product/ChannelVisibilityBlock.php
+++ b/src/Admin/Product/ChannelVisibilityBlock.php
@@ -1,0 +1,134 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Admin\Product;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer;
+use Automattic\WooCommerce\GoogleListingsAndAds\Value\ChannelVisibility;
+use Automattic\WooCommerce\GoogleListingsAndAds\Value\SyncStatus;
+use WP_REST_Request as Request;
+use WP_REST_Response as Response;
+use WC_Product;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class ChannelVisibilityBlock
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Admin\Product
+ */
+class ChannelVisibilityBlock {
+
+	public const PROPERTY = 'google_listings_and_ads__channel_visibility';
+
+	/**
+	 * @var ProductHelper
+	 */
+	protected $product_helper;
+
+	/**
+	 * ChannelVisibilityBlock constructor.
+	 *
+	 * @param ProductHelper $product_helper
+	 */
+	public function __construct( ProductHelper $product_helper ) {
+		$this->product_helper = $product_helper;
+	}
+
+	/**
+	 * Register hooks for querying and updating product via REST APIs.
+	 */
+	public function register_hooks(): void {
+		// https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-products-v2-controller.php#L182-L192
+		add_filter( 'woocommerce_rest_prepare_product_object', [ $this, 'prepare_data' ], 10, 2 );
+
+		// https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php#L200-L207
+		// https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php#L247-L254
+		add_filter( 'woocommerce_rest_insert_product_object', [ $this, 'update_data' ], 10, 2 );
+	}
+
+	/**
+	 * Get channel visibility data from the given product and add it to the given response.
+	 *
+	 * @param Response   $response Response to be added channel visibility data.
+	 * @param WC_Product $product WooCommerce product to get data.
+	 *
+	 * @return Response
+	 */
+	public function prepare_data( Response $response, WC_Product $product ): Response {
+		$response->data[ self::PROPERTY ] = [
+			'is_visible'         => $product->is_visible(),
+			'channel_visibility' => $this->product_helper->get_channel_visibility( $product ),
+			'sync_status'        => $this->product_helper->get_sync_status( $product ),
+			'issues'             => $this->product_helper->get_validation_errors( $product ),
+		];
+
+		return $response;
+	}
+
+	/**
+	 * Get channel visibility data from the given request and update it to the given product.
+	 *
+	 * @param WC_Product $product WooCommerce product to be updated.
+	 * @param Request    $request Response to get the channel visibility data.
+	 */
+	public function update_data( WC_Product $product, Request $request ): void {
+		if ( ! in_array( $product->get_type(), $this->get_visible_product_types(), true ) ) {
+			return;
+		}
+
+		$params = $request->get_params();
+
+		if ( ! isset( $params[ self::PROPERTY ] ) ) {
+			return;
+		}
+
+		$channel_visibility = $params[ self::PROPERTY ]['channel_visibility'];
+
+		if ( $channel_visibility !== $this->product_helper->get_channel_visibility( $product ) ) {
+			$this->product_helper->update_channel_visibility( $product, $channel_visibility );
+		}
+	}
+
+	/**
+	 * Return the visible product types to control the hidden condition of the channel visibility block
+	 * in the Product Block Editor.
+	 *
+	 * @return array
+	 */
+	public function get_visible_product_types(): array {
+		return array_diff( ProductSyncer::get_supported_product_types(), [ 'variation' ] );
+	}
+
+	/**
+	 * Return the config used for the input's block within the Product Block Editor.
+	 *
+	 * @return array
+	 */
+	public function get_block_config(): array {
+		$options = [];
+
+		foreach ( ChannelVisibility::get_value_options() as $key => $value ) {
+			$options[] = [
+				'label' => $value,
+				'value' => $key,
+			];
+		}
+
+		$attributes = [
+			'property'          => self::PROPERTY,
+			'options'           => $options,
+			'valueOfSync'       => ChannelVisibility::SYNC_AND_SHOW,
+			'valueOfDontSync'   => ChannelVisibility::DONT_SYNC_AND_SHOW,
+			'statusOfSynced'    => SyncStatus::SYNCED,
+			'statusOfHasErrors' => SyncStatus::HAS_ERRORS,
+		];
+
+		return [
+			'id'         => 'google-listings-and-ads-product-channel-visibility',
+			'blockName'  => 'google-listings-and-ads/product-channel-visibility',
+			'attributes' => $attributes,
+		];
+	}
+}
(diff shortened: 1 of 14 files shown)


src/Admin/Product/ChannelVisibilityBlock.php
```diff
@@ -0,0 +1,134 @@
...
+	/**
+	 * Register hooks for querying and updating product via REST APIs.
+	 */
+	public function register_hooks(): void {
+		// https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-products-v2-controller.php#L182-L192
+		add_filter( 'woocommerce_rest_prepare_product_object', [ $this, 'prepare_data' ], 10, 2 );
+
+		// https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php#L200-L207
+		// https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php#L247-L254
+		add_filter( 'woocommerce_rest_insert_product_object', [ $this, 'update_data' ], 10, 2 );
```
jorgemd24: Shouldn't this be `add_action` instead of `add_filter`?

src/Admin/Product/ChannelVisibilityBlock.php
```diff
@@ -0,0 +1,134 @@
...
+
+	/**
+	 * Get channel visibility data from the given product and add it to the given response.
+	 *
+	 * @param Response   $response Response to be added channel visibility data.
+	 * @param WC_Product $product WooCommerce product to get data.
+	 *
+	 * @return Response
+	 */
+	public function prepare_data( Response $response, WC_Product $product ): Response {
```
jorgemd24: In the filter, it mentions that the second argument will have the type of `WC_Data` (`WC_Product` extends WC_Data). So, perhaps it's safer to set the type to `WC_Data`, or are we confident that it will always be WC_Product?

src/Admin/Product/ChannelVisibilityBlock.php
```diff
@@ -0,0 +1,134 @@
...
+		return $response;
+	}
+
+	/**
+	 * Get channel visibility data from the given request and update it to the given product.
+	 *
+	 * @param WC_Product $product WooCommerce product to be updated.
+	 * @param Request    $request Response to get the channel visibility data.
+	 */
+	public function update_data( WC_Product $product, Request $request ): void {
```
jorgemd24: Similar concern to my previous comment about the `WC_Product` type.

src/Admin/ProductBlocksService.php
```diff
@@ -67,13 +72,16 @@ class ProductBlocksService implements Service, Registerable, Conditional {
...
 	 * @param AttributeManager       $attribute_manager
 	 * @param MerchantCenterService  $merchant_center
 	 */
-	public function __construct( AssetsHandlerInterface $assets_handler, AttributeManager $attribute_manager, MerchantCenterService $merchant_center ) {
+	public function __construct( AssetsHandlerInterface $assets_handler, ProductHelper $product_helper, AttributeManager $attribute_manager, MerchantCenterService $merchant_center ) {
 		$this->assets_handler    = $assets_handler;
 		$this->attribute_manager = $attribute_manager;
 		$this->merchant_center   = $merchant_center;
+
+		$this->channel_visibility_block = new ChannelVisibilityBlock( $product_helper );
```
jorgemd24: What's the reason to initiate the Class `ChannelVisibilityBlock` here instead of Core ServiceProvider?

For example:

`$this->share_with_tags( ChannelVisibilityBlock::class, ProductHelper::class );`

And then you pass the ChannelVisibilityBlock as a param here:

https://github.com/woocommerce/google-listings-and-ads/blob/20169c47a3cfa2ef9c588ea69684688d4195c677/src/Internal/DependencyManagement/CoreServiceProvider.php#L324

tests/Unit/Value/ChannelVisibilityTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
+declare( strict_types=1 );
```
jorgemd24, line 2: Thanks for covering this file with tests!

jorgemd24 (commented):
Thanks @eason9487 for working on this, It works nicely with the steps that you mention in the PR description. However, I've left a few comments about the code that I'd like to discuss before giving approval. Thanks!

src/Admin/Product/ChannelVisibilityBlock.php
eason9487: Thanks for catching this! Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/2214/commits/aa28eda518ead1a3cc1368220eb521e15a5332ed.

src/Admin/Product/ChannelVisibilityBlock.php
eason9487: The key call stacks should show the `$product` is always an instance of `WC_Product`:

- https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-products-v2-controller.php#L149-L151
- https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/wc-product-functions.php#L64-L74

Also, the method of `$product->is_visible()` call is implemented in [`WC_Product`](https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L1584), but not in the `WC_Data`. I think it would be clearer than using the abstract class `WC_Data` to explicitly indicate that this method expects the incoming `$product` to be an instance of `WC_Product`, and also ensures that subsequent calls are valid.

WDYT?

src/Admin/Product/ChannelVisibilityBlock.php
eason9487: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2214#discussion_r1464321716, the key call stacks are:

- https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php#L232-L238
- https://github.com/woocommerce/woocommerce/blob/8.5.0/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-products-controller.php#L377-L389

The subsequent method calls against to `ProductHelper` also require the incoming `$product` to be an instance of `WC_Product`, and some methods of its internal calls like `$product->get_parent_id()` are only implemented in `WC_Product` as well.

src/Admin/ProductBlocksService.php
eason9487: > What's the reason to initiate the Class `ChannelVisibilityBlock` here instead of Core ServiceProvider?

Since I'm still not quite sure what a "service" is in the `league/container` package and what kind of class should belong to "service". When I was adding the `ChannelVisibilityBlock` class, I thought it might not be a "service", so I used this approach.

> And then you pass the ChannelVisibilityBlock as a param here: [...]

Thanks for the suggestion. Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2214/commits/177e5ab911aed0798d711055321cbd1da59ea1c6.

eason9487:
Hi @jorgemd24, thanks for the review! This PR is ready for a new round of code reviews. Could you take a look?

src/Admin/ProductBlocksService.php
jorgemd24: > Since I'm still not quite sure what a "service" is in the league/container package and what kind of class should belong to "service". When I was adding the ChannelVisibilityBlock class, I thought it might not be a "service", so I used this approach.

Here there is a nice explanation: https://carlalexander.ca/dependency-injection-wordpress/  (section: On service classes and singletons). Another advantage of having the class `ChannelVisibilityBlock` as a dependency is that it's easier to test since you can simply mock the dependencies you pass to `ProductBlocksService`.

src/Admin/Product/ChannelVisibilityBlock.php
jorgemd24: Yes, from what I can tell, it seems like it will always be `WC_Product`. However, when I check examples in other plugins calling this filter, they also expect `WC_Data`.

See here: 

- https://github.com/woocommerce/woocommerce-brands/blob/develop/includes/class-wc-brands.php#L845
- https://github.com/woocommerce/woocommerce-brands/blob/f82f270efb374b746883a3b30374583c5fb9d9c5/includes/class-wc-brands.php#L873

jorgemd24 (approved):
Thanks @eason9487 for the adjustments! LGTM. I've left a comment about the types in the hooks callback, but since I can't simulate that scenario and considering the explanations given, I believe it's safe to approve the PR. If we want to be extra cautious, we could consider something like this:

```
if ( ! $product instanceof WC_Product ) {
	return null;
}
```

src/Admin/Product/ChannelVisibilityBlock.php
eason9487: Thanks for the suggestion mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/2214#pullrequestreview-1842430278, adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2214/commits/34abdb217c324689c8c537402ef1e2acc09e46db.

src/Admin/Product/ChannelVisibilityBlock.php
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2214/commits/34abdb217c324689c8c537402ef1e2acc09e46db.

PR #2217, opened 2024-01-25
Generate coverage report with xdebug

### LINKED ISSUE #2208: Calling phpdbg produces an incorrect coverage report for anonymous functions
### Describe the bug:

In our unit tests we use phpdbg for generating a coverage report, [see here](https://github.com/woocommerce/google-listings-and-ads/blob/2.5.15/.github/workflows/php-unit-tests.yml#L89).

However it struggles showing coverage for anonymous functions, which is quite a common pattern we use in our REST API controllers.

I can reproduce this locally by first running the unit tests with phpdbg:
`phpdbg -qrr vendor/bin/phpunit --coverage-html=tests/php-coverage/phpdbg`

When we check the report for the `ProductVisibilityController` we see it shows the function `get_update_callback` is 100% covered by unit testing:
image

When we look at the highlighting we can see it highlights returning the anonymous function as covered code, but it misses highlighting the contents of the anonymous function:
image
We can see the [same in the results on codecov.io](https://app.codecov.io/gh/woocommerce/google-listings-and-ads/pull/2207/blob/src/API/Site/Controllers/MerchantCenter/ProductVisibilityController.php#L75)

If I rerun the same test using xdebug instead:
`XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-html=tests/php-coverage/xdebug`

When we check the same report for the `ProductVisibilityController` we see it shows the function `get_update_callback` is only covered 8.7% by unit testing:
image

When we look at the highlighting we now see a correct picture because the code within the anonymous function is not covered:
image

This difference makes the coverage of the `ProductVisibilityController` drop from 83.33% to 63.22%. In some controllers this difference is even bigger.

### Steps to reproduce:
1. Setup unit tests `bin/install-wp-tests.sh <db_name> <db_user> <db_pass>`
2. Run the tests with phpdbg `phpdbg -qrr vendor/bin/phpunit --coverage-html=tests/php-coverage/phpdbg`
3. Run the tests with xdebug `XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-html=tests/php-coverage/xdebug`

### Expected behavior:
I would expect the code coverage to include anonymous functions so we can get an accurate report.

### Actual behavior:
The anonymous functions are being marked as 100% covered which is over inflating the coverage reports and not giving an accurate picture when changes are made within these functions.

### Additional details:
Running locally I don't see much difference in time when running the tests, they both take about a minute to run locally. According to [this report](https://github.com/sebastianbergmann/php-code-coverage/issues/945) phpdbg is getting removed for PHPUnit 10 anyway. PCOV seems to be the recommended solution but we'll have to check if that needs additional packages installed to run.

mikkamp:

This PR runs the coverage report using xdebug instead of phpdbg. Two changes were required to make this work:
- Setup PHP passing the coverage mode
- Run the unit tests with xdebug mode set

In addition to this change I've also set the codecov status checks to be informational only. This prevents the tests from being marked as failures when we don't reach the target coverage amount. Even though target is set to `auto` it still happens quite often that the checks on the PR fail because we don't submit enough unit tests. This will allow us to use the report as a guide instead of a requirement.

> [!Note]
To calculate the threshold / target it compares with the last submitted codecov report for the develop branch. We use an anonymous method (without token) to upload the codecov report since our repo is public. Unfortunately that means it sometimes hits API limits and prevents the codecov report from uploading. So it doesn't always have the latest to compare with, causing the status check to fail more often then expected. Setting it to informational prevents it from being a blocker. The other alternative would be to submit with a token so we have a more flexible limit.

Closes #2208 

1. View the report for the ProductVisibilityController and see that the contents of the anonymous function is now marked red as not covered: https://app.codecov.io/gh/woocommerce/google-listings-and-ads/pull/2217/blob/src/API/Site/Controllers/MerchantCenter/ProductVisibilityController.php#L75
2. Similar to a controller that is covered, we can now see that all the lines in the anonymous function are covered: https://app.codecov.io/gh/woocommerce/google-listings-and-ads/pull/2217/blob/src/API/Site/Controllers/Ads/CampaignController.php#L122
3. Confirm the codecov checks on the PR now pass even though the target has not been reached:
image

### Changelog entry
* Dev - Generate coverage report with xdebug.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,6 +5,13 @@ coverage:
     precision: 1
     round: nearest
     range: '50...80'
+    status:
+        project:
+            default:
+                informational: true
+        patch:
+            default:
+                informational: true
 
 ignore:
     - 'bin/*'
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -40,7 +40,7 @@ jobs:
         uses: woocommerce/grow/get-plugin-releases@actions-v1
         with:
           slug: woocommerce
-          
+
   UnitTests:
     name: PHP unit tests - PHP ${{ matrix.php }}, WP ${{ matrix.wp-version || 'latest' }}, WC ${{ matrix.wc-versions || 'latest' }}
     needs: GetMatrix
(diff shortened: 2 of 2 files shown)


puntope (approved):
✅  LGTM

PR #2227, opened 2024-01-30
[R-1.1] Move useGoogleAccountCheck to its own file

asvinb:

Update the `useGoogleAccountCheck` hook to live in its own file and return additional properties. Remove the line from the original PR where the rule of hooks was disabled within the hook.


joemcgill (approved):
I had the same thought. Thanks, @asvinb!

Issue #1194, opened 2022-01-14
Cleanup products which are not found in Merchant Center


mikkamp:
We've seen a recent Uptick in API errors  which is logged in #1181 
A lot of these errors are from the API methods productstatuses.get and products.delete and are classified as a notFound error.

This can happen when for example products have been deleted in the Merchant Center but the sites still has those products marked as synced. We store the Google ID's in the database so we need to update those ID's as soon as we know that a product no longer exists in the Merchant Center.

When we run the product sync we have a handler which handles all delete errors, and if it detects a notFound error it will call `remove_google_id` to update the ID's in the database. See: https://github.com/woocommerce/google-listings-and-ads/blob/1.10.0/src/Product/ProductSyncer.php#L299-L312

However when we call productstatuses.get we create batch requests for all ID's we have stored in the local database. If we get a notFound error we just skip to the next product, see: https://github.com/woocommerce/google-listings-and-ads/blob/1.10.0/src/MerchantCenter/MerchantStatuses.php#L318-L333
A suggestion was to call the same cleanup routine to ensure that the product is marked as unsynced since it no longer exists in the database. So when we call the product sync the next time round we won't try to delete it again.

Another suggestion:
> I think adding a cleanup or sync job that goes through all products and checks if they still exist in MC would be a good idea to decrease those `notFound` errors.

We will need to evaluate whether this will increase or decrease the current amount of requests or additional API errors.

mikkamp:
I created a [branch](https://github.com/woocommerce/google-listings-and-ads/tree/fix/1194-cleanup-not-found-products) with a scheduled cleanup task to remove any products that are not found when we retrieve the Merchant Status. It is working to cleanup the products, but after further testing it seems we might not want to add this cleanup task.

Google products can take up to 30 minutes to become available since it takes a while for them to be processed. So if we sync a product to Google and then request a product status update (using the stored ID) we will get a notFound error because it's not available yet. If we repeat the same request a little later we do get the product status, but with the proposed change we would have already cleaned up the product (marked as unsynced).

The problem stems from the fact that we use the following logic:
1. Retrieve all Google ID's of synced products
2. Send a [ProductStatuses batch](https://developers.google.com/shopping-content/reference/rest/v2.1/productstatuses/custombatch) request for a 1000 products at a time
3. Handle responses and retrieve the issues for each product and store them in our issues table
4. Repeat from step 2 for the next 1000 products

[...]

PR #2229, opened 2024-01-30
Get product stats using the Merchant Center report query

jorgemd24:

This PR explores a new method to gather product stats independently of the MC API ID. The past two approaches are:

1. The first approach (GLA v1.0.0) involved using product status to list all the products in the merchant center. However, this required fetching all the products in the MC, with a maximum of 250 results per request. For larger customers, this meant making multiple requests and fetching products that might not even be synced through the GLA.

2. The current approach involves using the MC API IDs created when submitting the products, stored in the metadata with the key `_wc_gla_google_ids`. Instead of fetching all the products, this approach fetches the products using the MC API ID, making it more efficient than the first approach. However, it may return multiple 404 errors if the product has been deleted in the MC and the GLA extension is not aware of it.

3. The new approach: As we're migrating to the API approach, we can't continue using the `_wc_gla_google_ids` metadata since we won't be submitting the product. Instead, Google will create the product for us, so we won't know the MC API ID. However, we do know the `offerID` that the product will have, which by default (or at least as it is now) is the WC product ID. We'll use the `offerIDs` from the "syncable" products to query the Merchant Center Reports, which allows us to retrieve a lot more results (the default page size is 1k) in one request and be independent of the MC API ID. 

It is a bit difficult to test the results as you will need products in MC with different statuses. I have been able to test the "Pending", "Not Synced", and "Disapproved" and for the "Expiring" I changed the code to simulate an expiring product.  

1.  Go to the Product Feed - Overview
2. Check the stats and compare the numbers with the "develop" branch.
 

### Additional details:
- The plan is to remove the `gla_` prefix for the `offerID` for newly synced products to the MC.
- If the WC product has `_wc_gla_google_ids`, Google will need to keep the `ID` and the `offerID` saved in the metadata (confirmation on how this will work is still needed).
- I've introduced the new static function `get_offer_id` as a replacement for `get_google_product_offer_id.` Since we're not using the slug, we could simply remove it from the slug argument in` get_google_product_offer_id`, but I'm unsure if this change might affect other plugins using this function for other purposes. Maybe we could mark it as deprecated.
- In the next PR, I'll include tests for this approach.
- It is still pending to test the query with a thousand products to see the performance.

### More context:

- pcTzPl-1Rz-p2#comment-2990
- https://github.com/woocommerce/google-listings-and-ads/issues/1194

### Changelog entry

>


src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+				]
+			);
+
+			/** @var SearchResponse $response  */
+			$response = $query
+			->set_client( $this->service, $this->options->get_merchant_id() )
+			->get_results();
+
+			if ( $response->count() ) {
+				$results = [ ...$results, ...$response->getResults() ];
```
mikkamp: Since we still support PHP lower than 8.1, this is risky as if either of those is an associative array this will fail, https://php.watch/versions/8.1/spread-operator-string-array-keys

src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+	}
+
+	/**
+	 * Get product statuses.
+	 *
+	 * @return array List of product statuses.
+	 * @throws GoogleException If the search call fails.
+	 */
+	public function get_product_statuses(): array {
+		$sync_ready_products_ids = $this->product_repository->find_sync_ready_products()->get_product_ids();
```
mikkamp: This is failing on my test site with a lot of products, as it runs out of memory. I don't have a lot of memory allocated with 128MB but I left it that way purposely to catch heavy sessions. Either way large sites are going to hit this limit even if it's increased, because `find_sync_ready_products` will be loading all the product objects and then extracting the IDS with `get_product_ids`

I temporarily added a limit of a 100 here to continue testing.

src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+			MCStatus::EXPIRING    => 0,
+			MCStatus::PENDING     => 0,
+			MCStatus::DISAPPROVED => 0,
+		];
+
+		/** @var $row ReportRow  */
+		foreach ( $rows as $row ) {
+			/** @var ProductView $product_view  */
+			$product_view    = $row->getProductView();
+			$experation_date = $product_view->getExpirationDate();
```
mikkamp: Just a typo here:
```suggestion
			$expiration_date = $product_view->getExpirationDate();
```
Same with all the later references where the variable is used.

src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+	}
+
+	/**
+	 * Calculate statistics for a list of report rows.
+	 *
+	 * @param ReportRow[] $rows List of report rows.
+	 *
+	 * @return array List of statistics.
+	 */
+	public function calculate_statuses( array $rows ): array {
```
mikkamp: Not the fault of this function, but this fails if `tests/proxy` is running as it returns mocked responses for the MC reports. Would be nice to change the proxy to be more explicit about which reports it filters.

I can fix that by simply adding:
```js
		if ( body.query.includes( 'ProductView' ) ) {
			return false;
		}
```
Would be nice to add that permanently.

src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+				case 'ELIGIBLE':
+				case 'ELIGIBLE_LIMITED':
+					++$statistics['active'];
+					break;
+				case 'PENDING':
+					++$statistics[ MCStatus::PENDING ];
+					break;
+				case 'NOT_ELIGIBLE_OR_DISAPPROVED':
+					++$statistics[ MCStatus::DISAPPROVED ];
+					break;
```
mikkamp: It seems a bit odd that we use the MCStatus class here for some, which match up with some internal statuses but not all, and also differ from the aggregated statuses. Is it still worthwhile to keep this as is, or are we doing that only to remain compatible with the old code to fetch the statuses.

Seems we might be better off using a new/refactored class, or not using it at all, so when we remove the old code it's easier to keep this working.

src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+		if ( count( $sync_ready_products_ids ) === 0 ) {
+			return $this->calculate_statuses( [] );
+		}
+
+		$offer_ids = array_map(
+			function ( $item ) {
+				return WCProductAdapter::get_offer_id( $item );
+			},
+			$sync_ready_products_ids
+		);
```
mikkamp: Seems like the function is unnecessary here, I would have gone for something like this:
```suggestion
		$offer_ids = array_map(
			[ WCProductAdapter::class, 'get_offer_id' ],
			$sync_ready_products_ids
		);
```

src/Product/WCProductAdapter.php
```diff
@@ -260,6 +260,39 @@ protected function map_wc_product_id(): WCProductAdapter {
...
+	 *
+	 * @param int $product_id The product ID.
+	 * @return string The offer ID
+	 * @since x.x.x
+	 * @see https://developers.google.com/shopping-content/reference/rest/v2.1/products#Product.FIELDS-table
+	 */
+	public static function get_offer_id( int $product_id ): string {
+		/** @var ProductHelper $product_helper  */
+		$product_helper = woogle_get_container()->get( ProductHelper::class );
+		$product        = wc_get_product( $product_id );
```
mikkamp: This becomes a problem because we run `array_map` currently for ALL syncable products. Since the current implementation already has loaded the product objects in `find_sync_ready_products` then we'll be getting them from the cache here so it won't use additional memory. But we need to avoid situations where we load all the product objects without doing so in batches.

src/Product/WCProductAdapter.php
```diff
@@ -260,6 +260,39 @@ protected function map_wc_product_id(): WCProductAdapter {
...
 
+	/**
+	 * Get the slug for the product.
+	 *
+	 * @param int $product_id The product ID.
+	 * @return string The offer ID
+	 * @since x.x.x
+	 * @see https://developers.google.com/shopping-content/reference/rest/v2.1/products#Product.FIELDS-table
+	 */
+	public static function get_offer_id( int $product_id ): string {
```
mikkamp: Curious why we need this to be a static function of WCProductAdapter. Seems like it would be better placed within ProductHelper. MerchantReport already has access to ProductHelper so then we wouldn't need to make it a static function.

src/API/Google/Query/MerchantProductViewReportQuery.php
```diff
@@ -0,0 +1,57 @@
...
+
+	/**
+	 * Set the initial columns for this query.
+	 */
+	protected function set_initial_columns() {
+		$this->columns(
+			[
+				'id'              => 'product_view.id',
+				'offer_id'        => 'product_view.offer_id',
+				'experation_date' => 'product_view.expiration_date',
```
mikkamp: Typo here, since the column key is just to prevent duplicate columns it doesn't actually get sent in the query, but would be nice to correct it either way.

src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php
```diff
@@ -102,15 +113,17 @@ protected function get_product_statistics_refresh_callback(): callable {
 	 * Get the overall product status statistics array.
 	 *
 	 * @param Request $request
-	 * @param bool    $force_refresh True to force a refresh of the product status statistics.
 	 *
 	 * @return Response
 	 */
-	protected function get_product_status_stats( Request $request, bool $force_refresh = false ): Response {
+	protected function get_product_status_stats( Request $request ): Response {
```
mikkamp: The endpoint `mc/product-statistics/refresh` still exists and calls this function with `$force_refresh = true`. Are we removing that endpoint now? Could we still benefit from caching if this is a "heavy" query for sites with lots of synced products.

src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php
```diff
@@ -125,13 +138,7 @@ protected function get_product_status_stats( Request $request, bool $force_refre
 	 */
 	protected function get_schema_properties(): array {
 		return [
-			'timestamp'      => [
```
mikkamp: Considering that we were previously storing the statistics in a transient, isn't this timestamp still useful to know when to trigger a hard refresh?

src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+			MCStatus::DISAPPROVED => 0,
+		];
+
+		/** @var $row ReportRow  */
+		foreach ( $rows as $row ) {
+			/** @var ProductView $product_view  */
+			$product_view    = $row->getProductView();
+			$experation_date = $product_view->getExpirationDate();
+
+			$formatted_expiration_date = DateTime::createFromFormat( 'Y-m-d', "{$experation_date->getYear()}-{$experation_date->getMonth()}-{$experation_date->getDay()}" );
```
mikkamp: Does Google have a helper function to convert this date? If not we might want to create one to reuse. Also if we are going to only use this date for doing a comparison, then wouldn't it be easier to convert it to a timestamp so the offset can use some basic math with the DAY_IN_SECONDS constant?

src/API/Google/MerchantReport.php
```diff
@@ -44,12 +59,114 @@ class MerchantReport implements OptionsAwareInterface {
...
+			if ( $response->count() ) {
+				$results = [ ...$results, ...$response->getResults() ];
+			}
+
+			$next_page_token = $response->getNextPageToken();
+
+		} while ( $next_page_token );
+
+		$statistics               = $this->calculate_statuses( $results );
+		$statistics['not_synced'] = count( $sync_ready_products_ids ) - count( $results );
```
mikkamp: Just a question about calculating these statistics. As far as I can see this is just counting any product type we support (simple and variation). In the previous version we aggregated variations to count as 1. Otherwise for a single dissapproved variable product with a 100 variations we'll get a 100 added to the disapproved section. Are we planning to aggregate those here as well?

src/Product/WCProductAdapter.php
```diff
@@ -260,6 +260,39 @@ protected function map_wc_product_id(): WCProductAdapter {
...
+		/** @var ProductHelper $product_helper  */
+		$product_helper = woogle_get_container()->get( ProductHelper::class );
+		$product        = wc_get_product( $product_id );
+		// Check if the product has the google_ids (mc ids) metadata set in the DB. If it has the google_ids we respect the offerID set in the metadata.
+		$has_mc_ids = $product_helper->is_product_synced( $product );
+
+		if ( $has_mc_ids ) {
+			$mc_ids = $product_helper->get_synced_google_product_ids( $product );
+			return $product_helper->get_offer_id_from_mc_id( reset( $mc_ids ) );
```
mikkamp: `reset` returns `false` if the array is empty, but `get_offer_id_from_mc_id` forces it's parameter to be a string. Do we need some additional checking to confirm we are getting the expected data?

mikkamp (commented):
Thanks for the PR @jorgemd24, it's working after I limit the amount of products it pulls statistics for.

Considering that this is a relatively "heavy" query what would you think about turning this into a background job like is done in `SyncableProductsCountController`. That way we could prevent it from trying to do everything in one request.

Also in my opinion predicting the offerID is still a little prone to failure if we no longer have access to accurate sync data in the WP database. Which could happen if they reconnect the MC account to another site, or they start from a product import. Since Google will be pulling the product data, wouldn't it be more foolproof to consider the MC as "source of truth" about what products exist (are synced) and what it's offer ID is. If we would change it to a batched request anyway. We could just query the MC for all products (without a filter of offerIds), and then go through them one by one and see if we find a matching WC product ID. The existing function `get_wc_product_id( string $mc_product_id )` would be good enough. We can then update the syncing status of that particular product and set any meta data as required. If we don't find a matching WC product then we can just ignore it and exclude it from the count.

jorgemd24:
Hi @mikkamp thanks for your comments.

I made some adjustments to the PR. I realized that we can still use most of the existing functions to calculate and update the metadata for the products. Now what we are doing is converting the aggregated status to the MC status, so we can reuse the functions that were already used.

I've moved the Merchant Report Query request to the AS job. This way, it can run in the background if the transient is not set. We might want to discuss what's more convenient: scheduling the job when we make a request to `mc/product-statistics` and the transient is not set, or leaving an endpoint for the client to schedule the job when needed similar to what's done in `SyncableProductsCountController`. 

Additionally, I think that while the job is running, the client could display something like "Loading Stats" or a similar message. Currently, it just shows N/A if the transient is not set.

I also bumped up my PHP memory limit to 128MB and set the limit to 1000 in the report query. It seemed to work well with more than 2k products, but I'm interested to hear about your results.

I have not added/updated the MCStatus tests as I would like to define the direction first.

src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php
```diff
@@ -35,18 +36,26 @@ class ProductStatisticsController extends BaseOptionsController {
...
 	/**
 	 * ProductStatisticsController constructor.
 	 *
 	 * @param RESTServer       $server
 	 * @param MerchantStatuses $merchant_statuses
 	 * @param ProductSyncStats $sync_stats
+	 * @param MerchantReport   $merchant_report
 	 */
-	public function __construct( RESTServer $server, MerchantStatuses $merchant_statuses, ProductSyncStats $sync_stats ) {
+	public function __construct( RESTServer $server, MerchantStatuses $merchant_statuses, ProductSyncStats $sync_stats, MerchantReport $merchant_report ) {
```
mikkamp: It looks like this might have been missed after the refactoring. But I don't see any location where the MerchantReport dependency is used.

src/Jobs/UpdateMerchantProductStatuses.php
```diff
@@ -0,0 +1,109 @@
...
+		$next_page_token = null;
+
+		do {
+			$response = $this->merchant_report->get_product_view_report( $next_page_token );
+
+			$this->merchant_statuses->process_product_report( $response->getResults() );
+
+			$next_page_token = $response->getNextPageToken();
+
+		} while ( $next_page_token );
```
mikkamp: I'm not sure if this loop is fully taking advantage of it being an async process. Yes we moved the job to a background process, but this loop means it will still process all the results in a single scheduled task. So it's not distributing the resource load.

Since after the refactor we revert back to loading WP_Post objects and updating meta in bulk we don't run into resource issues, but in theory even those resource limits can be reached for extreme sites.

What do you think about removing this loop and instead calling `$this->schedule` with the `next_page_token` as an argument (and keep doing so until there are no more `next_page_token`s? We might need to refactor `$this->merchant_statuses->update_product_stats` to support a partial update and then when there are no more `next_page_token`s left we set some status that everything is completed and calculate the final totals.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -106,7 +112,21 @@ public function __construct() {
...
+		}
+
+		// if not transient, return empty array and scheduled the job to update the statuses.
+		if ( ! $force_refresh && null === $this->mc_statuses ) {
+			// Schedule job to update the statuses.
+			$this->container->get( UpdateMerchantProductStatuses::class )->schedule();
+			return [
+				'timestamp'  => $this->cache_created_time->getTimestamp(),
+				'statistics' => [],
+			];
```
mikkamp, line 125: I found this a bit confusing, when I first saw the API response. Do you think we should set a flag that `schedule` has been called and we are waiting for updated results?

Another reason to consider adding a flag is when we force the refresh. If I call the endpoint with `force_refresh = true` but the transient hasn't expired yet, I get a response with counted statistics. But then if I call the endpoint repeatedly without force refresh I don't know at what point I transition between the old statistics and the scheduled job being completed and returning the new set of statistics. How does the UI then reflect to the user that it's waiting for new statistics and finish the polling to show the final statistics?

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -572,32 +586,112 @@ protected function refresh_presync_product_issues(): void {
...
+				$mc_product_status = MCStatus::EXPIRING;
+			}
+
+			$wc_product = $product_helper->get_wc_product_by_wp_post( $wc_product_id );
+			if ( ! $wc_product || 'product' !== substr( $wc_product->post_type, 0, 7 ) ) {
+				// Should never reach here since the products IDS are retrieved from postmeta.
+				do_action(
+					'woocommerce_gla_debug_message',
+					sprintf( 'Merchant Center product %s not found in this WooCommerce store.', $mc_product_id ),
+					__METHOD__ . ' in remove_invalid_statuses()',
```
mikkamp: The extra string is intended to be used when it's being called from an anonymous function, or to identify where in the function it's being called from. I think in this case `__METHOD__` should be enough.

src/MerchantCenter/ShoppingContentDateTrait.php
```diff
@@ -0,0 +1,28 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\Date as ShoppingContentDate;
+use Datetime;
```
mikkamp: Should be the right case:
```suggestion
use DateTime;
```

src/MerchantCenter/ShoppingContentDateTrait.php
```diff
@@ -0,0 +1,28 @@
...
+
+	/**
+	 * Convert ShoppingContentDate to DateTime.
+	 *
+	 * @param ShoppingContentDate $date The Google date.
+	 *
+	 * @return DateTime|false The date converted or false if the date is invalid.
+	 */
+	protected function convert_shopping_content_date( ShoppingContentDate $date ) {
+		return DateTime::createFromFormat( 'Y-m-d|', "{$date->getYear()}-{$date->getMonth()}-{$date->getDay()}" );
```
mikkamp: Doesn't seem like it would be a bother in converting the date but is `|` a typo, or does it have a specific purpose?

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -21,6 +21,11 @@
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\ChannelVisibility;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\MCStatus;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\ProductStatus as GoogleProductStatus;
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\ReportRow;
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\ProductView;
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\Date as ShoppingContentDate;
```
mikkamp: If the ShoppingContentDateTrait already sets this mapping it doesn't seem like it's necessary to include this here again. But see my other comment, on where this code should all be located under API/Google

src/API/Google/MerchantReport.php
```diff
@@ -52,6 +54,28 @@ public function __construct( ShoppingContent $service, ProductHelper $product_he
...
 
+	/**
+	 * Get ProductView Query response.
+	 *
+	 * @param string|null $next_page_token The next page token.
+	 * @return SearchResponse The response from the search call.
+	 *
+	 * @throws GoogleException If the search call fails.
+	 */
+	public function get_product_view_report( $next_page_token = null ): SearchResponse {
```
mikkamp: I have a bit of a mixed feeling about returning a SearchResponse from this function. It shifts the responsibility of understanding the objects returned in the response to the consuming classes. In all other uses we translate the results and pass it back as an associative array / internal object. As far as I recall the reason we did that was to abstract the Google APIs so classes don't necessarily need to know about all the objects.

For example in MerchantStatuses we need to tell it:
- what a ProductView is
- how to get data from a ReportRow
- convert from an aggregated status to a MC status
- convert the expired date to something it can work with

I think those responsibilities should all be part of MerchantReport which is a child of API/Google. So in short I think we should have this function return something like:

```php
[
	'statuses' => [
		[
			'product_id' => (int) $convertedByProductHelper,
			'status' => (MCStatus) $convertedFromAggregateStatus,
			'expired' => (DateTime) $convertedExpiryDate,
			...
		],
		...
	],
	'next_page' => $token,
];
```

The same goes with throwing a `GoogleException`, the `UpdateMerchantProductStatuses` job doesn't really know how to deal with that type of exception so it will just bubble up and be recorded as a generic exception for the scheduled action. We should catch it here, so we can log the exception and then throw an internal/generic exception with a formatted error message.
Most of that can be copied from `get_report_data`.

mikkamp (commented):
Thanks for the refactoring, I'm able to run it now without reaching out of memory issues.

I left some comments as I went through the code, but I think it's important to consider batching the call per page (1000 rows at a time). For now I think it's ok to stick with it using the WP_Post objects and bulk meta updates to prevent it from loading the full WC_Product objects (which consume more memory). But I personally think it's a bit of a maintenance burden to not work with WC_Product objects/CRUD classes. If anyone (or WC) decides to work with a different storage system for products our code will fail and we need to adjust to remain compatible. We knew that when we implemented this code, but we opted for speed over compatibility since we needed it to complete in one request.

Since we are moving it to async I think we'd be missing an opportunity if we don't take full advantage of batching the async process. Which means in the future we can:
- Swap out the code to work with CRUD classes
- Use a filter to control the batch size of a 1000 (a simple snippet can fix a site that is reaching it's memory limits)

The filter for the batch size is actually not a bad idea to add now already.

src/MerchantCenter/ShoppingContentDateTrait.php
jorgemd24: The `|` resets the time to 00:00:00, see this comment: https://github.com/php/doc-en/issues/671#issuecomment-1692909428.

I added it because each time I called this function, it would return the time depending on when it was executed.

mikkamp: Nice, I learned something new. To be honest I haven't used createFromFormat before, but seems like a good use case here. My solution would have been to just use the [regular constructor](https://www.php.net/manual/en/datetime.construct.php) and then blanking out time values. Something like this:
```suggestion
		return new DateTime( "{$date->getYear()}-{$date->getMonth()}-{$date->getDay()}T00:00:00+00:00" );
```

src/Jobs/UpdateMerchantProductStatuses.php
jorgemd24: Yes, I have the same feeling and I tried that approach too. The reason I hesitated to proceed was that when there are too many pending jobs, completing the full job would take even longer. It's not just one job that needs to be completed, but multiple. I'm open to exploring this in a different PR. Also, as we discussed, we could check the AS queue health to determine the best approach here.

src/MerchantCenter/MerchantStatuses.php
jorgemd24: That makes sense, now we're checking if the job is scheduled, and if it is, we won't return stats until the transient is refresh. One idea to inform the client that the data is refreshing could be to send an is_scheduled flag. Then we can use that flag to inform the user, or we could use the `scheduled_sync` to indicate the process.

src/API/Google/MerchantReport.php
jorgemd24: The primary reason I didn't format the response was that we were looping through the results twice - first to format them and then again to calculate the status. I agree with you that for consistency with the rest of the methods, it makes sense to return the data formatted. I've made the change here: https://github.com/woocommerce/google-listings-and-ads/pull/2229/commits/0e90b7d4aacac335b97840e6591375afdbaf0fbb

jorgemd24:
Thanks @mikkamp for your comments!

> For now I think it's ok to stick with it using the WP_Post objects and bulk meta updates to prevent it from loading the full WC_Product objects (which consume more memory). But I personally think it's a bit of a maintenance burden to not work with WC_Product objects/CRUD classes. If anyone (or WC) decides to work with a different storage system for products our code will fail and we need to adjust to remain compatible.

Yes, that makes sense. Now that we have a clearer strategy for fetching the product stats, we can start looking into those improvements.

> The filter for the batch size is actually not a bad idea to add now already.

I added a filter here: https://github.com/woocommerce/google-listings-and-ads/pull/2229/commits/1cf01782233e1750132eda45986af4e9ca09aa64

Also, I addressed your suggestions and replied to your questions! Do you mind to have another look?

Thanks!

src/API/Google/MerchantReport.php
mikkamp: That makes sense. Although in this case I think the intermediate data we store is relatively small (id, status, date), so shouldn't be much of a bottle neck. So in that case I think going with consistency / better separation is the preferred choice.

src/MerchantCenter/MerchantStatuses.php
mikkamp: IMHO the flag is cleaner, as it's a little more self documenting when looking at the schema of the API response. Which can be helpful for the UI to know whether to show the statistics or not.

src/Jobs/UpdateMerchantProductStatuses.php
mikkamp: That sounds like a good plan. Can we log all those details in a followup issue, so we can keep track of all things we want to improve eventually?

src/MerchantCenter/ShoppingContentDateTrait.php
```diff
@@ -0,0 +1,28 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter;
```
mikkamp: Can we move this trait as well and make it part of `Automattic\WooCommerce\GoogleListingsAndAds\API\Google`, since the classes in MerchantCenter only have to deal with a DateTime object.

src/API/Google/MerchantReport.php
```diff
@@ -52,6 +56,89 @@ public function __construct( ShoppingContent $service, ProductHelper $product_he
...
+			throw new Exception( __( 'Unable to retrieve Product View Report.', 'google-listings-and-ads' ), $e->getCode() );
+		}
+	}
+
+	/**
+	 * Convert the product view aggregated status to the MC status.
+	 *
+	 * @param string $status The status of the product.
+	 *
+	 * @return array The MC status.
```
mikkamp: Can we just make a small clarification that this is "The aggregated status of the product".
Also I think the return status was intended to be a string here.

src/API/Google/MerchantReport.php
```diff
@@ -52,6 +56,89 @@ public function __construct( ShoppingContent $service, ProductHelper $product_he
 		$this->product_helper = $product_helper;
 	}
 
+	/**
+	 * Get ProductView Query response.
+	 *
+	 * @param string|null $next_page_token The next page token.
+	 * @return array Array of products along with their statuses.
```
mikkamp: Not sure what the best description is here, but it's also including the next_page token.

src/API/Google/MerchantReport.php
```diff
@@ -52,6 +56,89 @@ public function __construct( ShoppingContent $service, ProductHelper $product_he
...
+				]
+			);
+
+			$response = $query
+			->set_client( $this->service, $this->options->get_merchant_id() )
+			->get_results();
+
+			$rows = $response->getResults();
+
+			foreach ( $rows as $row ) {
```
mikkamp: Not a big deal, but we might as well shorten this to one line since $rows isn't referenced anywhere else.

src/API/Google/MerchantReport.php
```diff
@@ -52,6 +56,89 @@ public function __construct( ShoppingContent $service, ProductHelper $product_he
...
+				];
+
+			}
+
+			$product_view_data['next_page'] = $response->getNextPageToken();
+
+			return $product_view_data;
+		} catch ( GoogleException $e ) {
+			do_action( 'woocommerce_gla_mc_client_exception', $e, __METHOD__ );
+			throw new Exception( __( 'Unable to retrieve Product View Report.', 'google-listings-and-ads' ), $e->getCode() );
```
mikkamp: Just wondering, since this is going to be recorded as a failure in a Scheduled Action scheduler job, do you think it will be useful to include both `$e->getMessage` and `$e->getCode`? Otherwise they have to look up the matching timestamp to see more details in the log.
```suggestion
			throw new Exception( __( 'Unable to retrieve Product View Report: ', 'google-listings-and-ads' ) . $e->getMessage(), $e->getCode() );
```

Or format the error with sprintf if we want to allow for greater translation/RTL support.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -572,32 +583,86 @@ protected function refresh_presync_product_issues(): void {
...
 	}
 
+	/**
+	 * Update the product status statistics.
+	 */
+	public function update_product_stats() {
+		$this->mc_statuses = [];
+		// Update each product's mc_status and then update the global statistics.
+		$this->update_product_mc_statuses();
+		$this->update_mc_statuses();
```
mikkamp: The naming of these two functions makes me have to look up what each one does. Since they are just internal functions would a more descriptive name help here?

```suggestion
		$this->update_products_meta_with_mc_status();
		$this->update_mc_status_statistics();
```

src/Product/ProductRepository.php
```diff
@@ -267,9 +267,9 @@ public function find_mc_not_synced_product_ids( int $limit = -1, int $offset = 0
 			'type'       => $types,
 			'meta_query' => [
 				[
-					'key'     => ProductMetaHandler::KEY_SYNC_STATUS,
-					'compare' => '!=',
-					'value'   => SyncStatus::SYNCED,
+					'key'     => ProductMetaHandler::KEY_MC_STATUS,
+					'compare' => '=',
+					'value'   => MCStatus::NOT_SYNCED,
```
mikkamp, line 272: I'm wondering if we are getting the right amount of "unsynced" products here.

Firstly it seems that `KEY_SYNC_STATUS` is no longer used after this PR, should we remove it?

Secondly based on PR #1728 it seems we are expecting to fetch "syncable but unsynced" products. In `update_product_mc_statuses` we set the MC status to `NOT_SYNCED` for all products that aren't found in MC. But we only have the meta values for all products that previously had a MC status meta key.

What happens if we have existing products, never synced, and then we install GLA. At what point do they get the initial MC status set?

Previously that was handled by triggering the job UpdateAllProducts which would walk through all the "syncable" products and then mark them as "pending" for their initial status. But if we don't trigger that job then it seems like we'd get 0 for unsynced products after initial install as they don't get a MC status set until something in a product is updated and we trigger the syncer hooks.

mikkamp (commented):
Thanks @jorgemd24 for all the changes. It looks pretty much good to go, the only part I think we should still settle on is how the unsynced count is handled. The rest of the comments I left are just minor things I noticed.

src/Product/ProductRepository.php
mikkamp: Ah looking closer at the details I see `update_product_mc_statuses` works with both a combination of:
- `$all_product_ids` all products
- `$current_product_statuses` all products which have a MC status previously set

So that means the MC status will be set for all products, but the meta `ProductMetaHandler::KEY_VISIBILITY` isn't going to be set for those products. Getting syncable products is expensive that's why we have a background job to get the count. Do we need to combine that somehow instead of querying here?

mikkamp:
Also forgot to mention, but as we discussed in slack, should we have this PR target a separate feature branch? Or is it easier to have it all in one and pull out the changes we want to release later?

src/Product/ProductRepository.php
jorgemd24: Mmmm yes, when I checked the code I saw that `MCStatus::NOT_SYNCED` was set when the product was not found in the MC but I didn't realize about `ProductMetaHandler::KEY_VISIBILITY`.

> Getting syncable products is expensive that's why we have a background job to get the count. Do we need to combine that somehow instead of querying here?

Do you mean fetching the count of syncable products from `OptionsInterface::SYNCABLE_PRODUCTS_COUNT` and then performing a simple calculation like `not_synced = OptionsInterface::SYNCABLE_PRODUCTS_COUNT -synced_products`? However, I think this would need updating the `OptionsInterface::SYNCABLE_PRODUCTS_COUNT` more frequently, as it's currently set only during user onboarding.

jorgemd24: > Firstly it seems that KEY_SYNC_STATUS is no longer used after this PR, should we remove it?

I see that the key is used here: 

https://github.com/woocommerce/google-listings-and-ads/blob/56d20b558f2fc24764c22f1a5c8dfadddbdd5b19/src/Product/SyncerHooks.php#L310

So if we plan to release this earlier, before the pulling method I guess it could affect when duplicating a product.

jorgemd24:
> Also forgot to mention, but as we discussed in slack, should we have this PR target a separate feature branch? Or is it easier to have it all in one and pull out the changes we want to release later?

I've been thinking about this because the current branch already contains the changes made in notifications, etc., since I merged `feature/google-api-project` ( https://github.com/woocommerce/google-listings-and-ads/pull/2229/commits/fb619db6d8d8935504f6b714914d30c47432fa69) into it. So, I'm considering either reverting that commit and targeting the branch to a new one based off "develop," or creating a fresh branch and cherry-picking these commits. Future PRs related to this section could then reference the new branch. WDYT?

src/Product/ProductRepository.php
mikkamp: > So if we plan to release this earlier, before the pulling method I guess it could affect when duplicating a product.

Makes sense, I guess we should leave it to remain backwards compatible. Probably good to leave a note somewhere why we are keeping it around.

mikkamp: > Do you mean fetching the count of syncable products from OptionsInterface::SYNCABLE_PRODUCTS_COUNT and then performing a simple calculation like not_synced = OptionsInterface::SYNCABLE_PRODUCTS_COUNT -synced_products? However, I think this would need updating the OptionsInterface::SYNCABLE_PRODUCTS_COUNT more frequently, as it's currently set only during user onboarding.

That's what I was thinking only on the first round, and then hoping that eventually we have meta updated so we can do a better calculation quicker. But I actually don't have a good solution there as it seems messy.

I'm currently wondering though, why do we make it difficult on ourselves to have this show "syncable but unsynced". Does the merchant even know that we are counting it like this? What do you think the effect would be if we would put just all unsynced products in there? I don't think syncable is a very clear state to the merchant as technically a product they would normally sync but is currently out of stock is temporarily classified as unsyncable until it comes back in stock. If we have it show just all unsynced products it would be much easier to explain what the count is there. At most we might need to clarify in the docs. We now have this in the docs:

> **Not synced**. These products do not appear across Google or the Google Shopping tab as free listings. They are queued for submission or may be ineligible or excluded from the product feed.

That actually sounds like it's describing "all unsynced" anyways, so might as well match that.

So in simple terms that means querying for `MCStatus::NOT_SYNCED` and removing the `ProductMetaHandler::KEY_VISIBILITY` condition.

mikkamp:
> So, I'm considering either reverting that commit and targeting the branch to a new one based off "develop," or creating a fresh branch and cherry-picking these commits. Future PRs related to this section could then reference the new branch. WDYT?

Either one would work for me. Reverting the merge commit would keep all the history of this PR around, but either way that would still be retained in the PR comments. Could also keep it simple and close this PR and replace it with a fresh one.

src/Product/ProductRepository.php
jorgemd24: > I'm currently wondering though, why do we make it difficult on ourselves to have this show "syncable but unsynced". Does the merchant even know that we are counting it like this? What do you think the effect would be if we would put just all unsynced products in there?

In my opinion, from the merchant's perspective and without checking the documentation, I would say that "Not synced" means all products that are not in the Merchant Center regardless of their visibility. 

In the follow-up issue that I am creating, I will add the item to update the documentation.

jorgemd24:
@mikkamp, thanks for your suggestions and discussions. I added your suggestions and per our https://github.com/woocommerce/google-listings-and-ads/pull/2229#discussion_r1482697054, I removed the  `ProductMetaHandler::KEY_VISIBILITY` from the non-syncable query.

As I mentioned in the other comment,  I will create a follow-up issue that includes all the tasks we discussed during this PR.

Thanks!

src/API/Google/MerchantReport.php
```diff
@@ -52,6 +56,87 @@ public function __construct( ShoppingContent $service, ProductHelper $product_he
...
+
+			foreach ( $response->getResults() as $row ) {
+
+				/** @var ProductView $product_view  */
+				$product_view = $row->getProductView();
+
+				$wc_product_id     = $this->product_helper->get_wc_product_id( $product_view->getId() );
+				$mc_product_status = $this->convert_aggregated_status_to_mc_status( $product_view->getAggregatedDestinationStatus() );
+
+				// Skip if the product id does not exist
```
mikkamp, line 95: Not a big deal, but technically we've only tried extracting the product ID from the string, we haven't actually checked if it exists.

This makes me wonder though, we are extracting from `$product_view->getId()`, why is both the `offer_id` and `id` included in the query MerchantProductViewReportQuery then? Not a big deal to leave it there, but was just curious. Seems like `getOfferId()` would return a much shorter string.

mikkamp (approved):
Thanks for all the refactoring it looks good to go from my side. I just left one question about the offer_id vs. id field, but it still works regardless. 

I'll let you figure out from here which PR is going to handle the tests and how you are going to merge it to a branch.

PR #2231, opened 2024-01-31
Prevent Google API requests when disconnected

### LINKED ISSUE #2224: Continued requests after Google account is marked as disconnected
### Describe the bug:
When monitoring the logs it was observed that there were still sites which were sending a daily request, even though the connection was marked as expired. The only request that we have which runs daily is `wc_gla_cron_daily_notes` which triggers both `ReviewAfterClicks` and `ReviewAfterConversions`. For the first one we can see it correctly calls `$this->merchant_center->is_connected()` however for the latter it only calls `$this->options->get_ads_id()`.

Only checking for an Ads ID is a problem, because it doesn't confirm that a Google connection is available. Instead we should call `AdsService::is_connected` which does take into consideration whether the Google account is connected.

### Steps to reproduce:
1. To be able to test this scenario a local instance of WCS is needed with an expired token (older than 6 months)
2. Go to WooCommerce > Status > Scheduled Actions and filter for pending
3. Manually run the action `wc_gla_cron_daily_notes`
4. Observe the server logs and confirm a 401 response was returned
5. Repeat the action `wc_gla_cron_daily_notes`
6. Confirm that again the server logs a request and returns a 401

### Expected behavior:
I would expect the option `GOOGLE_CONNECTED` to be false after the first request. Which should prevent any request from being sent.

### Actual behavior:
Only `$this->options->get_ads_id()` is checked to determine whether it should send a request to get Ad metrics. `AdsService::is_connected` should be called instead.

mikkamp:

This PR resolves the issue where the notification scheduler (which runs daily) still sends requests to the Google API even though it's marked as disconnected. The correct behaviour is to prevent any requests from being sent since a disconnected state means we are not able to retrieve the right tokens for the request.

For the ReviewAfterConversions notification it was only checking the presence of an Ads ID, I've expanded this to check if it's connected before retrieving conversions.

For the SetupCouponSharing notification I tweaked the behaviour since it throws an error when not connected. The exception was getting caught higher up in the NoteInitializer class, but I changed it so it would get caught directly when a connection is not established, which is the same way it's handled in SetupCampaign.

In addition to this I also checked all the other notifications to make sure none of them sent any requests, here is the overview of the status:
- CompleteSetup (does not send any requests to Google API)
- ContactInformation (checks `is_connected` before retrieving contact information)
- ReconnectWordPress (sends a request to WCS to check Google connection status, does not send request to Google API)
- ReviewAfterClicks (checks `is_connected` before retrieving clicks)
- SetupCampaign (catches an error when statuses are retrieved but google not connected)
- SetupCampaignTwoWeeks  (catches an error when statuses are retrieved but google not connected)

Closes #2224 

##### Testing ReviewAfterConversions
1. If using an existing site, clear out any previously sent notifications`DROP FROM wp_wc_admin_notes WHERE name LIKE 'gla-%'`
2. Clear the Ads metrics transient: `_transient_gla_ads_metrics`
3. Complete MC onboarding, Complete Ads onboarding, disconnect Google
4. Go to WooCommerce > Status > Scheduled Actions and filter for pending
5. Run the task `wc_gla_cron_daily_notes`
6. If connected to a local WCS the requests to the API can be checked. There should only be one request sent by the notifications (from ReconnectWordPress) to `/google/connection/google-mc`. Before the changes in this PR we would have also seen a request to `/google/google-ads/v14/customers/123456789/googleAds:search` which would have been logged as a failure:
```
2024-01-31T09:00:05+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\Internal\DependencyManagement\GoogleServiceProvider::handle_unauthorized_error Client error: `POST https://localhost:5000/google/google-ads/v14/customers/123456789/googleAds:search` resulted in a `401 Unauthorized` response:
{"statusCode":401,"error":{"message":"Site is not connected to Google account for google-ads"},"message":"Site is not co (truncated...)

2024-01-31T09:00:05+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\Notes\NoteInitializer::add_notes Please reconnect your Google account.
```

##### Testing SetupCouponSharing
1. If using an existing site, clear out any previously sent notifications`DROP FROM wp_wc_admin_notes WHERE name LIKE 'gla-%'`
2. Clear the statuses transient: `_transient_gla_mc_statuses`
3. Set the store country to a promotion supported country (like US)
4. Create a coupon (without setting it to Sync and Show)
5. Complete MC onboarding, disconnect Ads, disconnect Google
6. Go to WooCommerce > Status > Scheduled Actions and filter for pending
7. Run the task `wc_gla_cron_daily_notes`
8. Confirm that there are no more entries in WooCommerce > Status > Logs for the caught exception. Previously the disconnected error would be caught by the NoteInitializer and logged:
```
2024-01-31T09:00:05+00:00 ERROR Automattic\WooCommerce\GoogleListingsAndAds\Notes\NoteInitializer::add_notes Please reconnect your Google account.
```

### Additional details:
I've added two unit tests for the notifications ReviewAfterConversions and SetupCouponSharing.

### Changelog entry
* Fix - Prevent notifications from sending request to Google API when disconnected.

### DIFF
--- a/src/Notes/ReviewAfterConversions.php
+++ b/src/Notes/ReviewAfterConversions.php
@@ -4,10 +4,10 @@ declare( strict_types=1 );
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Notes;
 
 use Automattic\WooCommerce\Admin\Notes\Note as NoteEntry;
+use Automattic\WooCommerce\GoogleListingsAndAds\Ads\AdsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Ads\AdsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\MerchantMetrics;
 use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\Utilities;
-use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterAwareInterface;
-use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
 use Exception;
@@ -21,10 +21,10 @@ defined( 'ABSPATH' ) || exit;
  *
  * @package Automattic\WooCommerce\GoogleListingsAndAds\Notes
  */
-class ReviewAfterConversions extends AbstractNote implements MerchantCenterAwareInterface {
+class ReviewAfterConversions extends AbstractNote implements AdsAwareInterface {
 
+	use AdsAwareTrait;
 	use LeaveReviewActionTrait;
-	use MerchantCenterAwareTrait;
 	use PluginHelper;
 	use Utilities;
 
@@ -111,6 +111,10 @@ class ReviewAfterConversions extends AbstractNote implements MerchantCenterAware
 	 */
 	public function should_be_added(): bool {
 		if ( $this->has_been_added() ) {
+			return false;
+		}
+
+		if ( ! $this->ads_service->is_connected() ) {
 			return false;
 		}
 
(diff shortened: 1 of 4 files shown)


martynmjones (approved):
Hey @mikkamp, thanks for the quick fix!

I'm no longer seeing requests sent to the Google API using this branch instead the prompt to reconnect the account is logged and all tests are passing so it looks good to merge ✅

PR #2232, opened 2024-01-31
API PULL - Check notification statuses on processing

puntope:

This is a follow-up for https://github.com/woocommerce/google-listings-and-ads/pull/2202 

See https://github.com/woocommerce/google-listings-and-ads/pull/2202#pullrequestreview-1850621965 and https://github.com/woocommerce/google-listings-and-ads/pull/2202#pullrequestreview-1849507552

When a product changes the user may change it again before the job gets processed. That might cause some undesired notifications. This PR checks again the statuses to be sure this doesn't happen. 

1. Create a product and see `gla/jobs/notifications/products/process_item` action with `product.create`.
2. Run it and see the Notification Create happened in the WooCommerce - GLA - Logs.
3. Trash the product and fast do "undo" to untrash the product. 
4. see 2 actions  `gla/jobs/notifications/products/process_item` with `product.delete` and `gla/jobs/notifications/products/process_item` action with `product.update`.
5. Run both
6. see ONLY Notification Updated happened in the WooCommerce - GLA - Logs.

### Changelog entry

>

### DIFF
--- a/src/Jobs/Notifications/ProductNotificationJob.php
+++ b/src/Jobs/Notifications/ProductNotificationJob.php
@@ -75,7 +75,7 @@ class ProductNotificationJob extends AbstractActionSchedulerJob implements JobIn
 		$item  = $args[0];
 		$topic = $args[1];
 
-		if ( $this->notifications_service->notify( $item, $topic ) ) {
+		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $item, $topic ) ) {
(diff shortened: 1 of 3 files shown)


jorgemd24 (approved):
Hi @puntope, thanks for working on this, it is fixing the issues mentioned in the previous PRs. While I'm not entirely fan of scheduling jobs that may not perform any action, I understand that, for now, it seems like the most viable solution. Debugging might become more difficult, because you might assume that a scheduled job triggers the notification, but the actual outcome could be different.

PR #2233, opened 2024-02-01
Only sync selected categories as product type

### LINKED ISSUE #2230: Google product type: Parent category gets synced when only sub-category is selected
### Describe the bug:

The selected Woo Product Category is synced as [Google product type](https://support.google.com/merchants/answer/6324406?hl=en). However, when you assign a sub-category to a product, both the assigned sub-category and its parent category get synced as the product type - instead of just the assigned sub-category. This causes GMC/Google Ads to assign incorrect product type (the parent category) to the product, instead of the sub-category. 

Screenshot of sub-category assigned on store: 
Screenshot of parent and child category synced to GMC: 

### Steps to reproduce:

1. Assign a sub-category to a product
2. Check in GMC and you'll see both the sub-category and parent category get synced.

### Expected behavior:

Only the assigned sub-category should be synced. 

### Additional details:

- Reported in 7600082-zen

- Discussed on Slack: p1706088275307189-slack-C0410KV3YLW

mikkamp:

This PR changes the behaviour to only sync the selected categories as product type. Previously it would sync parent categories, both included as the parent of a child category as well as the parent individually even if the parent wasn't selected directly.

This change switches from using `wc_get_product_cat_ids` to using `wc_get_product_term_ids` as the first function will include a list of ancestors (regardless of whether they are selected or not). The latter function will return a list of all the selected categories. Which means we can influence which parent categories get synced individually by selecting them or not.

I altered the unit test to match the expected behaviour as well as testing some more edge cases to ensure they are synced correctly:
- test for unselected parent not included
- test 2 levels of ancestors
- test parent and child are included individually when both are selected
- test unrelated category is not included

Closes #2230

1. Setup a site with the extension and onboarded to Merchant Center 
2. Create some categories with hierarchy
3. Create a test product where only the child category is selected
image
4. Save the product and wait for it to sync, and then confirm the final attributes in MC, the parent category should not be included as an individual line
image
5. Create a test product where both the parent and child categories are selected
image
6. Save the product and wait for it to sync, and then confirm the final attributes in MC, the parent category should be included as an individual line and also as part of the child line
image

### Changelog entry
* Fix - Only sync selected categories as product type.

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -194,8 +194,11 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 	 */
 	protected function map_product_categories() {
 		// set product type using merchants defined product categories
-		$base_product_id            = $this->is_variation() ? $this->parent_wc_product->get_id() : $this->wc_product->get_id();
-		$this->product_category_ids = wc_get_product_cat_ids( $base_product_id );
+		$base_product_id = $this->is_variation() ? $this->parent_wc_product->get_id() : $this->wc_product->get_id();
+
+		// Fetch only selected term ids without parents.
+		$this->product_category_ids = wc_get_product_term_ids( $base_product_id, 'product_cat' );
+
 		if ( ! empty( $this->product_category_ids ) ) {
 			$google_product_types = self::convert_product_types( $this->product_category_ids );
 			do_action(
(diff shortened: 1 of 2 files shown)


martynmjones (approved):
Hey @mikkamp, thanks for the fix.

Tested syncing products with different category selections to confirm it will sync with:

- Only the parent category
- Parent and child categories
- Child categories only

LGTM ✅

PR #2237, opened 2024-02-05
Product Block Editor: Add a custom block for prompting user to complete onboarding

eason9487:

With the original product editor, this extension shows a prompt if the user has not yet completed onboarding.

image

To have a consistent prompt in the Product Block Editor, this PR:

- Change to let the `ChannelVisibilityBlock` handle its own registration as it has different conditions than `ProductBlocksService`.
- Change `ProductBlocksService` to still add a group for this extension and register custom blocks when the onboarding is not yet completed.
- Add a new custom block for prompting the user to complete the onboarding.
- Show the onboarding prompt in the Product Block Editor when it's not yet completed.
- Extra changes:
   - Fix the integrity checks for `woocommerce-grow-jsdoc` and `woocommerce-grow-storybook`.
   - Fix the deprecations of using `${var}` in a string for the `ProductBlocksService` class.

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/dc8d1530-aece-4b3b-96c1-8818babe7a55

1. Please refer to [the test preparation in PR 2151](https://github.com/woocommerce/google-listings-and-ads/pull/2151#prepare-test-environment) with **Woo 8.5.1**.
   - 💡 Please note that the current implementation doesn't work with Woo >= 8.6 due to breaking changes. This will be fixed by another PR.
1. `npm install`
1. `npm start`
1. To control the onboarding status, locally changing this return value would be easier to test.
   https://github.com/woocommerce/google-listings-and-ads/blob/1ef20671a867ec921393c809e58f2e88e06cec9e/src/MerchantCenter/MerchantCenterService.php#L75-L77

#### 📌 Test the onboarding prompt

1. Set the onboarding status to not completed.
1. Check if it shows a prompt as in the following screenshot.
   image
1. Click the "Get Started" button to see if it brings you to this extension's Get Started page.
1. Repeat this test with variable and variation products.

### DIFF
--- a/src/Admin/Product/ChannelVisibilityBlock.php
+++ b/src/Admin/Product/ChannelVisibilityBlock.php
@@ -3,7 +3,9 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Admin\Product;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Registerable;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
+use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\ChannelVisibility;
@@ -20,7 +22,7 @@ defined( 'ABSPATH' ) || exit;
  *
  * @package Automattic\WooCommerce\GoogleListingsAndAds\Admin\Product
  */
-class ChannelVisibilityBlock implements Service {
+class ChannelVisibilityBlock implements Service, Registerable {
 
 	public const PROPERTY = 'google_listings_and_ads__channel_visibility';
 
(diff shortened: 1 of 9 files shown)


src/Admin/Product/ChannelVisibilityBlock.php
```diff
@@ -29,19 +31,30 @@ class ChannelVisibilityBlock implements Service {
...
 	}
 
 	/**
 	 * Register hooks for querying and updating product via REST APIs.
 	 */
-	public function register_hooks(): void {
+	public function register(): void {
+		if ( ! $this->merchant_center->is_setup_complete() ) {
+			return;
+		}
```
puntope, line 56: 💅  Why not make this controller conditional as well?

tests/Unit/Admin/ProductBlocksServiceTest.php
```diff
@@ -193,13 +193,35 @@ public function test_register_merchant_center_setup_is_not_complete() {
...
+				[
+					'id'         => 'google-listings-and-ads-product-onboarding-prompt',
+					'blockName'  => 'google-listings-and-ads/product-onboarding-prompt',
+					'attributes' => [ 'startUrl' => 'http://example.org/wp-admin/admin.php?page=wc-admin&path=/google/start' ],
+				]
+			);
+
 		$this->variation_anchor_group->get_root_template()
-			->expects( $this->exactly( 0 ) )
+			->expects( $this->exactly( 1 ) )
```
puntope, line 211: 💅  Another simple version for `$this->exactly( 1 )`  and `$this->exactly( 0 )` that we normally use is `$this->once()` and `$this->never()`. Not against using `$this->exactly()` just mentioning it in case you didn't know.

puntope (commented):
Thanks for the changes @eason9487 ✅ LGTM

- I tested the `product-onboarding-prompt` and I can see it when the onboarding is not completed. 
- When it's completed then it shows the GLA blocks instead.

❓Not introduced in this PR but it seems that `ChannelVisibilityBlock::update_data` and `ChannelVisibilityBlock::prepare_data` have the wrong type in the params. Shouldn't be just `WC_Product` instead of `WC_Data` ?

Actually. We return if `$product` is not `WC_Product`  

```php 
if ( ! $product instanceof WC_Product || ! in_array( $product->get_type(), $this->get_visible_product_types(), true ) ) {
			return;
		}
```

puntope (approved):

src/Admin/Product/ChannelVisibilityBlock.php
eason9487: The `Conditional::is_needed` method is static but the `MerchantCenterService::is_setup_complete` method is not static, so this service cannot be registered conditionally.

tests/Unit/Admin/ProductBlocksServiceTest.php
eason9487: Thanks for the information. I chose `exactly` because it's relatively effortless to maintain.

eason9487:

> ❓Not introduced in this PR but it seems that `ChannelVisibilityBlock::update_data` and `ChannelVisibilityBlock::prepare_data` have the wrong type in the params. Shouldn't be just `WC_Product` instead of `WC_Data` ?
> 
> Actually. We return if `$product` is not `WC_Product`
> [...]

The code reviews in #2214 did discuss this, please refer to https://github.com/woocommerce/google-listings-and-ads/pull/2214#discussion_r1463518319 and https://github.com/woocommerce/google-listings-and-ads/pull/2214#discussion_r1463533278.

eason9487:
Rebased this PR onto the current `feature/support-product-block-editor` branch (4f2e536756c4d27ecc717a0e763b9f685297566c) and also removed the integrity checksum fix (d75ee2818cc650a6f72b2ca76bb25d18845d8e43) in favor of the same fix merged in #2241.

PR #2252, opened 2024-02-08
Add Shipping notifications

puntope:

This PR implement Notifications for Shipping Updates.
As discussed for now we don't send Item IDs and just send a generic "shipping.update" method with no Item ID

1. Checkout this PR, complete onboarding and select "Shipping Rates --> Automatically sync my store’s shipping settings to Google."
2. Go to WooCommerce -- Settings -- Shipping and update something 
3. See the Job `gla/jobs/notifications/shipping/process_item` appearing
4. Run it and see the notification `shipping.update` happening 

### Additional details:

⚠️  The topic needs to be enabled in the Proxy, so all the notifications will fail 

### Changelog entry

>

### DIFF
--- a/src/Google/NotificationsService.php
+++ b/src/Google/NotificationsService.php
@@ -48,7 +48,7 @@ class NotificationsService implements Service {
 	 * https://public-api.wordpress.com/wpcom/v2/sites/{site}/partners/google/notifications
 	 *
 	 * @param string   $topic The topic to use in the notification.
-	 * @param int|null $item_id Tme item ID to notify. It can be null for topics that doesn't need Item ID
+	 * @param int|null $item_id The item ID to notify. It can be null for topics that doesn't need Item ID
 	 * @return bool True is the notification is successful. False otherwise.
 	 */
 	public function notify( string $topic, $item_id = null ): bool {
--- a/src/Google/NotificationsService.php
+++ b/src/Google/NotificationsService.php
@@ -25,8 +25,7 @@ class NotificationsService implements Service {
 	public const TOPIC_COUPON_CREATED   = 'coupon.create';
 	public const TOPIC_COUPON_DELETED   = 'coupon.delete';
 	public const TOPIC_COUPON_UPDATED   = 'coupon.update';
-	public const TOPIC_SHIPPING_SAVED   = 'action.woocommerce_after_shipping_zone_object_save';
-	public const TOPIC_SHIPPING_DELETED = 'action.woocommerce_delete_shipping_zone';
+	public const TOPIC_SHIPPING_UPDATED = 'shipping.update';
 
 	/**
 	 * The url to send the notification
@@ -48,11 +47,11 @@ class NotificationsService implements Service {
 	 * Calls the Notification endpoint in WPCOM.
 	 * https://public-api.wordpress.com/wpcom/v2/sites/{site}/partners/google/notifications
 	 *
-	 * @param int    $item_id
-	 * @param string $topic
+	 * @param string   $topic The topic to use in the notification.
+	 * @param int|null $item_id Tme item ID to notify. It can be null for topics that doesn't need Item ID
 	 * @return bool True is the notification is successful. False otherwise.
 	 */
-	public function notify( int $item_id, string $topic ): bool {
+	public function notify( string $topic, $item_id = null ): bool {
 		/**
 		 * Allow users to disable the notification request.
 		 *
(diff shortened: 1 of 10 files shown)


src/Google/NotificationsService.php
```diff
@@ -48,11 +47,11 @@ public function __construct() {
 	 * Calls the Notification endpoint in WPCOM.
 	 * https://public-api.wordpress.com/wpcom/v2/sites/{site}/partners/google/notifications
 	 *
-	 * @param int    $item_id
-	 * @param string $topic
+	 * @param string   $topic The topic to use in the notification.
+	 * @param int|null $item_id Tme item ID to notify. It can be null for topics that doesn't need Item ID
```
jorgemd24: Typo Tme -> The

src/Shipping/SyncerHooks.php
```diff
@@ -45,17 +47,30 @@ class SyncerHooks implements Service, Registerable {
 	 */
 	protected $update_shipping_job;
 
+	/**
+	 * @var NotificationsService $notifications_service
+	 */
+	protected $notifiations_service;
```
jorgemd24: Typo: $notifiations_service -> $notifications_service

src/Shipping/SyncerHooks.php
```diff
@@ -128,7 +143,13 @@ protected function handle_update_shipping_settings() {
 		if ( $this->already_scheduled ) {
 			return;
 		}
-		$this->update_shipping_job->schedule();
+
+		if ( $this->notifiations_service->is_enabled() ) {
```
jorgemd24: WDYT if we also add this checking in the method `can_schedule()`?

jorgemd24 (commented):
Thanks @puntope for adding this new notification. I left a couple of minor comments about the code.

> As discussed for now we don't send Item IDs and just send a generic "shipping.update" method with no Item ID

Can you point me to that discussion? What I see is that without that ID, Google would have to make multiple requests to figure out which Shipping Zone has been modified.

When I send a request to `wc/v3/shipping/zones`, it only gives me the Shipping Zone name without details about the shipping zone methods. So, if a shipping zone method gets updated, Google would have to make separate calls to `wc/v3/shipping/zones/ZONE_ID/methods` for each zone.

Additionally, what do you think in adding tests for the ShippingNotification Job?

I'm guessing there'll be a follow-up PR for create/delete notifications, is that correct?

puntope:
> Additionally, what do you think in adding tests for the ShippingNotification Job?

Added here https://github.com/woocommerce/google-listings-and-ads/pull/2252/commits/a98ca65d0fd1b65d8e49821fe9cb168fb00c9102

src/Shipping/SyncerHooks.php
puntope: Adjusted here  https://github.com/woocommerce/google-listings-and-ads/pull/2252/commits/34e13297686e07a5d33a1d21f5d4618eb0e49a4f

src/Shipping/SyncerHooks.php
puntope: Fixed here https://github.com/woocommerce/google-listings-and-ads/pull/2252/commits/72f00f8af290befd45248dbfa5e4a087ef2d4de1

src/Google/NotificationsService.php
puntope: Fixed here https://github.com/woocommerce/google-listings-and-ads/pull/2252/commits/72f00f8af290befd45248dbfa5e4a087ef2d4de1

puntope:
> Can you point me to that discussion? What I see is that without that ID, Google would have to make multiple requests to figure out which Shipping Zone has been modified.

This was discussed in the dev call as we decided to operate on ourselves and wait to google for feedback. (instead of asking Google what to do beforehand). In my 1:1 with Mik we decided to just send `shipping.update` and rely on Google to fetch all the shipping and all the methods by themselves.

I will open a thread in P2 to discuss further. We can add them in future PRs

I created tho some example PR with that approach. https://github.com/woocommerce/google-listings-and-ads/pull/2256
However, it comes with more questions like:

- Shall we also add notifications for methods?
- How do we handle shipping class updates?
- Do we support more options than flat or free shipping? 

I suggest to go for now with this PR and discuss further in P2. I will create the discussion tomorrow.

jorgemd24 (approved):
Thanks @puntope for the adjustments and the tests.

>However, it comes with more questions like:
>Shall we also add notifications for methods?
>How do we handle shipping class updates?
>Do we support more options than flat or free shipping?


I feel like we should at least include the Shipping Zone ID to avoid unnecessary requests. I don't really see the need to add notifications for methods or the shipping class. For instance, if I query `wc/v3/shipping/zones/ZONE_ID/methods`, I can retrieve all the methods for that zone ID, making it easier to tell which method has changed.

>I suggest to go for now with this PR and discuss further in P2. I will create the discussion tomorrow.

Okay let's do that!

Issue #2250, opened 2024-02-08
Followup - Enhancing the method for fetching Product Overview Stats


jorgemd24:
### Migration Project

Part of https://github.com/woocommerce/google-listings-and-ads/issues/2146

Follow up of https://github.com/woocommerce/google-listings-and-ads/pull/2229

In PR https://github.com/woocommerce/google-listings-and-ads/pull/2229, we shifted the calculation from the product statistics to a background job. This opens up more options for enhancing performance when fetching statuses from the MC. Here's a list of potential tasks that could be implemented or explored to improve overall performance. Also it includes some pending tasks due to the changes made in this PR. 

#### Tasks

- [x] Schedule multiple jobs to fetch the product statistics using the page token. See https://github.com/woocommerce/google-listings-and-ads/pull/2229#discussion_r1479851013 
- [x] Explore checking the AS queue to determine if there are too many pending jobs. If so, attempt to execute the calculation without a job.
- [x] Increase MC status transient time. 
- [x] Explore to change WP_Post objects and bulk meta updates to WC_Product objects/CRUD classes. See https://github.com/woocommerce/google-listings-and-ads/pull/2229#pullrequestreview-1865229495
- [ ] Update the documentation regarding "Not Synced" to reflect the adjustments made in the calculation of unsynchronized products. See: https://github.com/woocommerce/google-listings-and-ads/pull/2229#discussion_r1482697054
- [x] Fix Issue with intermediate parents status.
- [x] Return error message if the `UpdateMerchantProductStatuses` job fails.
- [x] Add the UI to retry the `UpdateMerchantProductStatuses` job in case of failure.
- [x] Refactor issues
- [x] Delete stale postmeta for `mc_status` key

PR #2255, opened 2024-02-11
Create product statuses jobs for each page token

jorgemd24:

In https://github.com/woocommerce/google-listings-and-ads/pull/2229 we've added a background job that handles calculating product stats, bringing all products from MC and updating their status in WooCommerce.

This PR will allow the background job to generate multiple tasks based on the `$next_page_token`, helping us spread out resource usage more effectively.

While working on this  I noticed that we were changing the product status to "not synced" to calculate the unsynced products in MC. But, I think that's unnecessary. The actual count of unsynced products will simply be: NOT_SYNCED = WooCommerce Products ('simple', 'variable') - MC Products (with variants grouped).

In the case that we need to display the MC status, like in the Product Feed section, we can assume that if a WC product doesn't have an MC status, it's not synced in MC. I'm open to hearing other thoughts on this.

### Screenshots:

1. Change the batch size [here](https://github.com/woocommerce/google-listings-and-ads/blob/dd39acb2c1e877186678945dc5fe40936cafcb6b/src/API/Google/MerchantReport.php#L68) so you can create multiple batches depending on the number of products that you have or use the filter `woocommerce_gla_product_view_report_page_size`.
2. Make the following request to `GET gla/mc/product-statistics/refresh`
3. Go to WC -> Status -> Scheduled Actions -> Search for `gla/jobs/update_merchant_product_statuses`
4. See how multiple jobs are creating using a different page tokens.
5. Check the total stats in the Overview - Product Feed page makes sense. 

### Additional details:

- In another PR, I'll work with the WC_Product objects/CRUD classes and add the tests for MerchantStatuses.php and MerchantReport.php.

### Changelog entry

>

### DIFF
--- a/src/API/Google/MerchantReport.php
+++ b/src/API/Google/MerchantReport.php
@@ -105,7 +105,7 @@ class MerchantReport implements OptionsAwareInterface {
 
 			}
 
-			$product_view_data['next_page'] = $response->getNextPageToken();
+			$product_view_data['next_page_token'] = $response->getNextPageToken();
 
 			return $product_view_data;
 		} catch ( GoogleException $e ) {
--- a/src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php
+++ b/src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php
@@ -5,6 +5,7 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Merch
 
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\ProductSyncStats;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantStatuses;
 use WP_REST_Response as Response;
 use WP_REST_Request as Request;
@@ -27,16 +28,25 @@ class ProductStatisticsController extends BaseOptionsController {
 	 */
 	protected $merchant_statuses;
 
+	/**
+	 * Helper class to count scheduled sync jobs.
+	 *
+	 * @var ProductSyncStats
+	 */
+	protected $sync_stats;
+
 
 	/**
 	 * ProductStatisticsController constructor.
 	 *
 	 * @param RESTServer       $server
 	 * @param MerchantStatuses $merchant_statuses
+	 * @param ProductSyncStats $sync_stats
 	 */
-	public function __construct( RESTServer $server, MerchantStatuses $merchant_statuses ) {
+	public function __construct( RESTServer $server, MerchantStatuses $merchant_statuses, ProductSyncStats $sync_stats ) {
 		parent::__construct( $server );
 		$this->merchant_statuses = $merchant_statuses;
+		$this->sync_stats        = $sync_stats;
 	}
 
 	/**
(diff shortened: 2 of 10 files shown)


mikkamp:
I didn't want to do a full review of this PR, but just had a few comments.

> While working on this I noticed that we were changing the product status to "not synced" to calculate the unsynced products in MC. But, I think that's unnecessary. The actual count of unsynced products will simply be: NOT_SYNCED = WooCommerce Products ('simple', 'variable') - MC Products (with variants grouped).
> In the case that we need to display the MC status, like in the Product Feed section, we can assume that if a WC product doesn't have an MC status, it's not synced in MC. I'm open to hearing other thoughts on this.

That does seem to be an accurate calculation moving forward, although I'm wondering if that's totally backwards compatible since there might be products with the NOT_SYNCED status already set. For the sync stats calculation that seems to be fine, but do we need to consider unsetting the NOT_SYNCED status for products?

Another question, is should total WC products be done with a product type filter? Wouldn't the Merchants perceive that an excluded product type would be unsynced?

Also for intermediate data in between jobs, I see it's being stored in a transient. Given that a [transient can disappear at any time](https://developer.wordpress.org/apis/transients/), it would be safer to use an option, like is done in the UpdateSyncableProductsCount.

I noticed ProductSyncStats was being removed, if we are going to merge this PR early, then shouldn't we leave that in, since it's still used in the UI.

jorgemd24:
Thanks @mikkamp for your comments:

> since there might be products with the NOT_SYNCED status already set. For the sync stats calculation that seems to be fine, but do we need to consider unsetting the NOT_SYNCED status for products?

Can you help me to see a scenario where having the NOT_SYNCED status already set could potentially cause an issue? Or vice versa? If it is an issue, we could consider creating a job that runs once to unset the NOT_SYNCED status for products.

> Another question, is should total WC products be done with a product type filter? Wouldn't the Merchants perceive that an excluded product type would be unsynced?

Do you mean, that the total number of products should be the count of all products, regardless of their type (including variations)?

> Also for intermediate data in between jobs, I see it's being stored in a transient. Given that a [transient can disappear at any time](https://developer.wordpress.org/apis/transients/),

TIL, I thought the transient would stay until its expiration time, but it looks like it can get deleted even sooner.

> Furthermore, it is possible for the transient to not be available before the expiration time

jorgemd24:
@mikkamp and @martynmjones, I changed the PR to use an option for the intermediate data and reverted the change of removing  ProductSyncStats.

mikkamp:
> Can you help me to see a scenario where having the NOT_SYNCED status already set could potentially cause an issue? Or vice versa? If it is an issue, we could consider creating a job that runs once to unset the NOT_SYNCED status for products.

I was mainly thinking about the point where you mentioned:
> In the case that we need to display the MC status, like in the Product Feed section, we can assume that if a WC product doesn't have an MC status, it's not synced in MC.

But I see now that's for individual products so we can combine both NOT_SYNCED and not set as the same values. The other thought I was worried about is the querying, but it seems `find_mc_not_synced_product_ids` is no longer there, so I assume that means we don't have anywhere we need to fetch all products that are "not synced".

> Do you mean, that the total number of products should be the count of all products, regardless of their type (including variations)?

Yes exactly, because there could be lots of unsupported product types. Variations use a different `post_type=product_variation`. Do we need to do something special to fetch only the `post_type=product` in the query?

Thanks for switching the transient to an option, it's just a little more reliable and hard to troubleshoot if a transient ever gets cleared.

martynmjones (approved):
Hey @jorgemd24, thank you for improving the efficiency of this feature.

Aside from the comments that have already been addressed above I can't see any problems with these changes. Tested the branch in several different scenarios and have consistently had the correct stats built and the scheduled actions behave as to be expected. So LGTM ✅

jorgemd24:
Thanks @martynmjones for your time reviewing this PR!

PR #2257, opened 2024-02-12
MC Product Statuses: Replace WP functions with WC functions

jorgemd24:

In https://github.com/woocommerce/google-listings-and-ads/pull/2255, we've divided the background job for fetching product statuses into multiple tasks to distribute resource usage. Until now, we've been relying on `WP_Post` and bulk updates via raw queries to handle updating numerous products in a single request. While it sped things up, it could became less reliable as WC might change how product data is stored. For better future-proofing, it's more convenient to use WooCommerce objects to retrieve and update their data.

This PR replaces the bulk queries and the use of get_post with `wc_get_products`.  I've also updated metadata using methods available in their class. The metadata is saved using the `save_meta_data` method, so this way, we bypass calling the `woocommerce_update_product` hook and avoid triggering the syncer hooks.

At first, I experimented with batches of 1,000 products. However, in some instances, it was hitting the maximum execution time of 30 seconds. So, I opted to decrease it to 500 products. I did these tests with a memory limit of 128MB.

1.  Create a bunch of products ( I did the test using more than 3k products)
2. Reduce the PHP memory limit to 128MB
3. Delete any existing `mc_statuses` so you can check that the` mc_status` is added. You can run a query like this one: 
`DELETE FROM wp_postmeta WHERE meta_key = '_wc_gla_mc_status';`
4. Watch your server error log.
4. Call this endpoint `GET /gla/mc/product-statistics/refresh`
5. Go to WC -> Status -> Scheduled actions and see that `update_merchant_product_statuses` runs.
6. When all the jobs are completed, you should see the total results in Product View - Overview section.
7. Check that the product's metadata `_wc_gla_mc_status` is set.
8. Check that the syncer hooks have not been called. (No pending jobs trying to sync the product)

### Additional details:

- The UI will be updated in a different PR to indicate that the stats are loading while the jobs are running. 
- With these changes, the methods `ProductMetaQueryHelper::batch_update_values` and `ProductMetaQueryHelper::batch_insert_values` are not longer used in our codebase, so they could be deleted however they are public, so not sure if it is better just to deprecate them for now. 

### Changelog entry

>

### DIFF
--- a/src/API/Google/MerchantReport.php
+++ b/src/API/Google/MerchantReport.php
@@ -97,7 +97,7 @@ class MerchantReport implements OptionsAwareInterface {
 					continue;
 				}
 
-				$product_view_data['statuses'][ (int) $wc_product_id ] = [
+				$product_view_data['statuses'][ $wc_product_id ] = [
 					'product_id'      => $wc_product_id,
 					'status'          => $mc_product_status,
 					'expiration_date' => $this->convert_shopping_content_date( $product_view->getExpirationDate() ),
--- a/src/API/Google/MerchantReport.php
+++ b/src/API/Google/MerchantReport.php
@@ -65,7 +65,7 @@ class MerchantReport implements OptionsAwareInterface {
 	 * @throws Exception If the product view report data can't be retrieved.
 	 */
 	public function get_product_view_report( $next_page_token = null ): array {
-		$batch_size = apply_filters( 'woocommerce_gla_product_view_report_page_size', 1000 );
+		$batch_size = apply_filters( 'woocommerce_gla_product_view_report_page_size', 500 );
 
 		try {
 			$product_view_data = [
--- a/src/API/Google/MerchantReport.php
+++ b/src/API/Google/MerchantReport.php
@@ -97,7 +97,7 @@ class MerchantReport implements OptionsAwareInterface {
 					continue;
 				}
 
-				$product_view_data['statuses'][] = [
+				$product_view_data['statuses'][ (int) $wc_product_id ] = [
 					'product_id'      => $wc_product_id,
 					'status'          => $mc_product_status,
 					'expiration_date' => $this->convert_shopping_content_date( $product_view->getExpirationDate() ),
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -611,47 +611,27 @@ class MerchantStatuses implements Service, ContainerAwareInterface, OptionsAware
 	}
 
 	/**
-	 * Get the WC Products from the Product View statuses report.
-	 *
-	 * @param array $statuses statuses.
-	 * @see MerchantReport::get_product_view_report
-	 *
-	 * @return WC_Product[] Associative array with the key as the product ID and the value the WooCommerce Product object.
-	 */
-	protected function get_wc_products_from_product_view_statuses( array $statuses ): array {
-		$product_repository = $this->container->get( ProductRepository::class );
-		$products           = $product_repository->find_by_ids( array_column( $statuses, 'product_id' ) );
-		$map                = [];
-
-		foreach ( $products as $product ) {
-			$map[ $product->get_id() ] = $product;
-		}
-
-		return $map;
-	}
-
-	/**
 	 * Process product status statistics.
 	 *
 	 * @param array[] $statuses statuses.
 	 * @see MerchantReport::get_product_view_report
 	 */
 	public function process_product_statuses( array $statuses ): void {
+		$product_repository = $this->container->get( ProductRepository::class );
+		$products           = $product_repository->find_by_ids_as_associative_array( array_column( $statuses, 'product_id' ) );
+
 		$this->product_statuses = [
 			'products' => [],
 			'parents'  => [],
 		];
-
-		$visibility_meta_key = $this->prefix_meta_key( ProductMetaHandler::KEY_VISIBILITY );
-		$products            = $this->get_wc_products_from_product_view_statuses( $statuses );
 
 		foreach ( $statuses as $product_status ) {
 
 			$wc_product_id     = $product_status['product_id'];
 			$mc_product_status = $product_status['status'];
 
-			// Skip if the product does not exist or if the product previously found/validated.
-			if ( ! $wc_product_id || ! empty( $this->product_data_lookup[ $wc_product_id ] ) ) {
+			// Skip if the product does not exist in WooCommerce.
+			if ( ! $wc_product_id ) {
 				continue;
 			}
 
@@ -667,12 +647,6 @@ class MerchantStatuses implements Service, ContainerAwareInterface, OptionsAware
 				continue;
 			}
 
-			$this->product_data_lookup[ $wc_product_id ] = [
-				'name'       => $wc_product->get_name(),
-				'visibility' => $wc_product->get_meta( $visibility_meta_key ),
-				'parent_id'  => $wc_product->get_parent_id(),
-			];
-
 			if ( $this->product_is_expiring( $product_status['expiration_date'] ) ) {
 				$mc_product_status = MCStatus::EXPIRING;
 			}
@@ -681,13 +655,21 @@ class MerchantStatuses implements Service, ContainerAwareInterface, OptionsAware
 			$this->product_statuses['products'][ $wc_product_id ][ $mc_product_status ] = 1 + ( $this->product_statuses['products'][ $wc_product_id ][ $mc_product_status ] ?? 0 );
 
 			// Aggregate parent statuses for mc_status postmeta.
-			$wc_parent_id = $this->product_data_lookup[ $wc_product_id ]['parent_id'];
+			$wc_parent_id = $wc_product->get_parent_id();
 			if ( ! $wc_parent_id ) {
 				continue;
 			}
 			$this->product_statuses['parents'][ $wc_parent_id ][ $mc_product_status ] = 1 + ( $this->product_statuses['parents'][ $wc_parent_id ][ $mc_product_status ] ?? 0 );
 
 		}
+
+		$parent_keys     = array_values( array_keys( $this->product_statuses['parents'] ) );
+		$parent_products = $product_repository->find_by_ids_as_associative_array( $parent_keys );
+		$products        = $products + $parent_products;
+
+		// Update each product's mc_status and then update the global statistics.
+		$this->update_products_meta_with_mc_status( $products );
+		$this->update_intermediate_product_statistics( $products );
 	}
 
 	/**
@@ -701,10 +683,6 @@ class MerchantStatuses implements Service, ContainerAwareInterface, OptionsAware
 		$this->mc_statuses = [];
 
 		$this->process_product_statuses( $statuses );
-
-		// Update each product's mc_status and then update the global statistics.
-		$this->update_products_meta_with_mc_status();
-		$this->update_intermediate_product_statistics();
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


mikkamp:
I understand this is still a WIP, but just wanted to ask the question early. One of the side effects of switching to CRUD classes is that update hooks will be called if we set/update product meta. Are we going to put something in place to short circuit the syncer hooks (eventually will be notification hooks), to prevent them from being triggered?

jorgemd24:
>I understand this is still a WIP, but just wanted to ask the question early. One of the side effects of switching to CRUD classes is that update hooks will be called if we set/update product meta. Are we going to put something in place to short circuit the syncer hooks (eventually will be notification hooks), to prevent them from being triggered?

Thanks @mikkamp for the heads up! I was just in the middle of implementing something to short-circuit, but then I realized that when you use `save_meta_data`, it doesn't trigger the `woocommerce_update_product` hook. So, the syncer hooks aren't triggered either.

src/API/Google/MerchantReport.php
```diff
@@ -97,7 +97,7 @@ public function get_product_view_report( $next_page_token = null ): array {
 					continue;
 				}
 
-				$product_view_data['statuses'][] = [
+				$product_view_data['statuses'][ (int) $wc_product_id ] = [
```
mikkamp: Is the typecast here necessary? As far as I can see the function `get_wc_product_id` is required to return an `int`. And if it is necessary then my next question would be why we aren't typecasting it for all usages of this variable.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -844,33 +864,14 @@ protected function update_products_meta_with_mc_status() {
...
-
-		// Get all MC statuses.
-		$current_product_statuses = $product_meta_query_helper->get_all_values( ProductMetaHandler::KEY_MC_STATUS );
-
-		// Format: product_id=>status.
-		$to_insert = [];
-		// Format: status=>[product_ids].
-		$to_update = [];
 
+		ksort( $new_product_statuses );
```
mikkamp: From what I can see we only update product meta with this. Why does that need to be done in a particular order?

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -652,11 +668,15 @@ public function process_product_statuses( array $statuses ): void {
 			}
 
 			$this->product_data_lookup[ $wc_product_id ] = [
-				'name'       => get_the_title( $wc_product ),
-				'visibility' => get_post_meta( $wc_product_id, $visibility_meta_key ),
-				'parent_id'  => $wc_product->post_parent,
+				'name'       => $wc_product->get_name(),
+				'visibility' => $wc_product->get_meta( $visibility_meta_key ),
+				'parent_id'  => $wc_product->get_parent_id(),
```
mikkamp: I'm a bit unclear how we are expecting to use this `product_data_lookup`. The name and visibility are used when we handle product issues. However since we populated the statuses in batches this information isn't going to be pre-loaded when we loop through the product issues.

Would it be more useful to save an instance of the product objects in the class:

```php
		$this->process_product_statuses( $statuses );
		$this->update_products_meta_with_mc_status();
		$this->update_intermediate_product_statistics();
```

So when we call these three functions they all work with the same product set? I can see the update meta function reloads them from cache but that's not faster then just keeping a previous copy around. Or if we don't want to save a copy in the class, can we have `process_product_statuses` call the other two functions internally so we don't need to call the three in a row? That way it can also pass the set of product objects by reference so we don't need to store it in the class instance.

mikkamp (commented):
Thanks for all the changes here, it feels cleaner to be working with CRUD classes. I left some comments as we go through the code, I think we need to do better on how we handle the set of products we are working with to make them accessible.
I also left some comments below, but these aren't fully related to this PR so might be better handled in a followup PR.

Not directly related to this PR but I was getting the following errors in the Action Scheduler: `action failed via Async Request: Call to a member function getId() on null`
Cause was proxied requests returning invalid data, but my question is how is the API request supposed to handle the failure scenario? In my case it stayed with `loading` set to true, but had actually completed with a failure.

Again probably not related to this PR, but I was wondering about the function `update_intermediate_product_statistics` and how it handles variable products. For example if batch 1 has 1 variation (APPROVED) and batch 2 has 1 variation (DISSAPPROVED) for the same variable product. Both statuses will get a +1 count for the same variable product.

src/API/Google/MerchantReport.php
jorgemd24: It's correct, there is not need for typecasting. Removed here: https://github.com/woocommerce/google-listings-and-ads/pull/2257/commits/7b9944abc11e57aecf5814800b42de23aa438e1d

src/MerchantCenter/MerchantStatuses.php
jorgemd24: This line was left from the previous code:

https://github.com/woocommerce/google-listings-and-ads/blob/e9bc80d007c3574af4126949a67edc5519d4eb19/src/MerchantCenter/MerchantStatuses.php#L703

So I removed it here: https://github.com/woocommerce/google-listings-and-ads/pull/2257/commits/21e97e2d09ae546713af132a6e08d977b90508b4

src/MerchantCenter/MerchantStatuses.php
jorgemd24: I had the same feeling, the main reason I kept the` $product_data_lookup` as it was is because I initially intended to populate the product issues using the ProductView Report query. My plan was to reuse most of the code. However, after further investigation of the product issues, I realized that the ProductView doesn't return the details from the description or the documentation. So, I don't believe I'll be able to use the same query for that purpose.

https://github.com/woocommerce/google-listings-and-ads/blob/e9bc80d007c3574af4126949a67edc5519d4eb19/src/MerchantCenter/MerchantStatuses.php#L485-L487

One idea to retrieve the product issues could be to store the Google IDs when we retrieve the product status. Then, the product issues section could continue to function similarly to how it's currently implemented.

> Or if we don't want to save a copy in the class, can we have process_product_statuses call the other two functions internally so we don't need to call the three in a row? That way it can also pass the set of product objects by reference so we don't need to store it in the class instance.


Thanks for the suggestion! I've updated the code so that now those functions accept an array of products. While I was making these updates, I noticed that parent_id wasn't being fetched previously. So, I've adjusted the code to fetch all the parent IDs at once.

jorgemd24:
Thanks @mikkamp for your comments and suggestions.

> Not directly related to this PR but I was getting the following errors in the Action Scheduler: action failed via Async Request: Call to a member function getId() on null
Cause was proxied requests returning invalid data, 

Sorry, I've updated the proxy to avoid those errors. https://github.com/woocommerce/google-listings-and-ads/pull/2257/commits/e15c368511a3ad10c28d69bfc2b74b035c906c0c .

> but my question is how is the API request supposed to handle the failure scenario? In my case it stayed with loading set to true, but had actually completed with a failure.

One option could be to store the error in a transient let's say `gla_mc_statuses_errors` and then return this to the client. This way, the client can try to refresh the statuses, either automatically or by adding a small icon for manual retrying.

> Again probably not related to this PR, but I was wondering about the function update_intermediate_product_statistics and how it handles variable products. For example if batch 1 has 1 variation (APPROVED) and batch 2 has 1 variation (DISSAPPROVED) for the same variable product. Both statuses will get a +1 count for the same variable product.

That's a tricky situation, but I think we can resolve it by storing the parent statuses in the intermediate transient. Then, we can compare if the new status has a higher priority than the one saved in the previous batch. WDYT?

@mikkamp, when you have a chance, can you do another round? Thanks!

mikkamp:
> One option could be to store the error in a transient let's say gla_mc_statuses_errors and then return this to the client. This way, the client can try to refresh the statuses, either automatically or by adding a small icon for manual retrying.

Either a new transient, or storing the error in the transient we are already saving. So next time it can detect that the transient contains an error and can return a correct response. For UI I think it makes sense to present the user with the error message and a small refresh icon.

> That's a tricky situation, but I think we can resolve it by storing the parent statuses in the intermediate transient. Then, we can compare if the new status has a higher priority than the one saved in the previous batch. WDYT?

I agree, I was thinking about something similar, and just looping through the parent statuses at the end.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -617,32 +617,28 @@ protected function refresh_presync_product_issues(): void {
 	 * @see MerchantReport::get_product_view_report
 	 */
 	public function process_product_statuses( array $statuses ): void {
+		$product_repository = $this->container->get( ProductRepository::class );
+		$products           = $product_repository->find_by_ids_as_associative_array( array_column( $statuses, 'product_id' ) );
```
mikkamp, line 621: I like that the mapping part is moved to the ProductRepository. I was wondering why we didn't already have that and it didn't seem correctly placed handling it in the MerchantStatuses class.

src/Product/ProductRepository.php
```diff
@@ -98,6 +98,25 @@ public function find_by_ids( array $ids, array $args = [], int $limit = -1, int
...
+	 * Find and return an associative array of products with the product ID as the key.
+	 *
+	 * @param int[] $ids    Array of WooCommerce product IDs
+	 * @param array $args   Array of WooCommerce args (except 'return'), and product metadata.
+	 * @param int   $limit  Maximum number of results to retrieve or -1 for unlimited.
+	 * @param int   $offset Amount to offset product results.
+	 *
+	 * @return WC_Product[] Array of WooCommerce product objects
+	 */
+	public function find_by_ids_as_associative_array( array $ids, array $args = [], int $limit = -1, int $offset = 0 ) {
```
mikkamp: Can we force this function to have a return type of array? The function already always returns an array. I just noticed you have the code `$products + $parent_products` so I wanted to confirm that there is no way that those variables are not arrays.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -844,33 +853,23 @@ protected function update_products_meta_with_mc_status() {
...
-			} elseif ( $current_product_statuses[ $product_id ] !== $new_status ) {
-				// MC status not same as WC, update.
-				$to_update[ $new_status ][] = intval( $product_id );
+			$product = $products[ $product_id ] ?? null;
+
+			// At this point, the product should exist in WooCommerce but in the case that product is not found, log it.
+			if ( ! $product ) {
+				do_action(
+					'woocommerce_gla_debug_message',
+					sprintf( 'Merchant Center product %s not found in this WooCommerce store.', $product_id ),
```
mikkamp: This error message seems to implicate that we are outputting the MC ID, but at this time we have already mapped it to a WC ID (potentially with a filter and custom mapping). Should we reword it to something like:
```suggestion
					sprintf( 'Merchant Center product with WooCommerce ID %d is not found in this store.', $product_id ),
```

src/MerchantCenter/MerchantStatuses.php
mikkamp: > One idea to retrieve the product issues could be to store the Google IDs when we retrieve the product status. Then, the product issues section could continue to function similarly to how it's currently implemented.

Even though we can't use the ProductView Report query, we can still use the [productstatuses/list](https://developers.google.com/shopping-content/reference/rest/v2.1/productstatuses/list) endpoint, which supports pagination. That will allow us to pull the products in batches without needing to know a list of product/Google ID's up front. We can just act on whatever products the MC presents us with, and if we can map to a WC ID then we store the list of issues.

Since product issues are stored in the DB locally (until they expire), and we have the product statistics job run in batches of products. We can also consider doing something like this for each batch of UpdateMerchantProductStatuses:

- Query ProductView report to get a list of products
- Map products to WC ID's
- Add up intermediate statuses
- Fetch product statuses for the same batch of products using [productstatuses/custombatch](https://developers.google.com/shopping-content/reference/rest/v2.1/productstatuses/custombatch)
- Store issues in local DB

We might still need something to pull in account issues / add local issues to the table. And possibly decrease batch sizes if it takes more resources to pull in statuses and issues for that batch.

I think something like that will take the most advantage of running a background process, otherwise we'd have to create two to loop through the same batch of MC products. Let me know what you think about that approach.

mikkamp (approved):
Thanks for the changes, it's working as expected. I left a few minor comments, but it seems to have submitted them early. Since those are minor and the rest is reliant on follow up PR's I'll go ahead and approve this PR.

PR #2242, opened 2024-02-06
R2 – Enable Enhanced Conversions

asvinb:
### Changes proposed in this Pull Request:

Closes #2216 

_Replace this with a good description of your changes & reasoning._

### Detailed test instructions:

1. Complete the plugin onboarding process.
2. User should be redirected to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed&guide=submission-success`
3. The success modal should be displayed.
4. Refer to the requirements section in https://github.com/woocommerce/google-listings-and-ads/issues/2239 for the different screens to be displayed within the modal and also the modal content.

### Additional details:
1. There is a delay for the second screen within the success modal to show up. We must look into how we can grab the data as early as possible.
2. To create paid campaigns without setting up billing data, [refer to the following code snippet](https://github.com/woocommerce/google-listings-and-ads/wiki/Hack-for-working-locally-with-APIs#use-ads-account-without-setting-billing-data).

### Changelog entry

> Add - Enhanced Conversions Tracking panel to the welcome guide and Settings page.

PR #2258, opened 2024-02-15
(R2.1f) Google tag enhanced conversion

### LINKED ISSUE #2216: Enable Enhanced Conversions (EC) for GL&A
# Overview 

This is a tracking issue for enabling Enhanced Conversions in the plugin.

The goal of this requirement is to allow users the option to configure [Enhanced Conversions](https://support.google.com/google-ads/answer/9888656?hl=en), through Google Listings & Ads in order to improve performance of existing conversion measurement, by supplementing conversion tags with first party customer data in a privacy safe way. 

Developer Documentation: https://developers.google.com/google-ads/api/docs/conversions/enhanced-conversions/web

### R 2.1 – Develop post-onboarding pop-up modal panel to configure enhanced conversions

Success Modal

- [x] [#2239 Add a page to the `SubmissionSuccessGuide` panel to enable Enhanced Conversions](https://github.com/woocommerce/google-listings-and-ads/issues/2239)
- [x] Develop a REST API endpoint for reading ToS status (via #2242) 
- [x] Develop a REST API endpoint for reading/writing ToS Enhanced Conversions status (via #2242)
- [x] Integrate REST API endpoints with the success modal EC page (via #2242)
- [x] Update gtag and purchase conversions when EC is enabled (via #2258)

### R 2.2 – Develop a panel in the Google Listings & Ads Settings Tab to manage the current enhanced conversion activation state

Settings Panel

The behavior when this is enabled should follow the same ToS logic described in R2.1

- [x] Add panel in Google Listings & Ads Settings Tab that shows the current enhanced conversion activation state
```[tasklist]
### Updated Requirements
- [ ] Update Enhanced Conversions onboarding flow #2364
```

ankitrox:

Related to #2216.

1. If the enhanced conversion is enabled add the `{ 'allow_enhanced_conversions': true }` to the global site tag config (see [this article](https://support.google.com/google-ads/answer/13258081?visit_id=638283250727735978-1247451496&rd=2#Add_a_code_snippet&zippy=%2Cconfigure-your-conversion-page-google-tag)).
2. When the enhanced conversion is disabled, remove the above config from global site tag.
3. When the enhanced conversion is enabled, add the hashed user data on order received page so that it can be sent to analytics. The formatting for hashed data can be found in [this article](https://support.google.com/google-ads/answer/13262500?hl=en#zippy=identify-and-define-your-enhanced-conversions-variables%2Cset-up-standard-automatic-enhanced-conversions%2Cenable-enhanced-conversions-in-google-tag-manager-and-create-custom-javascript-variable:~:text=164%20standard.-,For%20hash%3A,-Use%20hex%20SHA256).
4.  When the enhanced conversion is disabled, do not add user data on order received page.

### Screenshots:

See testing screenshots in [this QA report](https://github.com/woocommerce/google-listings-and-ads/pull/2258#issuecomment-1991859193).

Steps (when enabled):

1. If the Enhanced Conversion setting is not already enabled, navigate to WP Dashboard > Marketing > Google Listing & Ads > Settings and enable it.
2. Add a product to the cart and proceed to complete the checkout process.
3. Verify that the order is successfully placed and the thank you page is displayed to the customer.
4. View the source code of the thank you page.

Expected Result:

1. The configuration 'allow_enhanced_conversions:true' should be present in the Global Site Tag within the source code of the thank you page.
2. User data should be present in encrypted form on the thank you page. As shown in the screenshot.

### Changelog entry

> Update - Set the "allow_enhanced_conversions" flag to Google tag when Enhanced Conversions is enabled.


src/Google/GlobalSiteTag.php
```diff
@@ -355,6 +356,63 @@ public function maybe_display_conversion_and_purchase_event_snippets( string $ad
...
+		if ( ! empty( $email ) ) {
+			$ec_data['email'] = $email;
+		}
+
+		// Format phone number in IE64.
+		$phone        = preg_replace( '/[^0-9]/', '', $phone );
+		$phone_length = strlen( $phone );
+		if ( $phone_length > 9 && $phone_length < 14 ) {
+			$phone            = sprintf( '%s%d', '+', $phone );
+			$ec_data['email'] = $email;
```
joemcgill: I think this was meant to be where you set ec_data for phone:

```suggestion
			$ec_data['phone_number'] = $phone;
```

src/Google/GlobalSiteTag.php
```diff
@@ -355,6 +356,63 @@ public function maybe_display_conversion_and_purchase_event_snippets( string $ad
 			join( ',', $item_info ),
 		);
 		wp_print_inline_script_tag( $purchase_page_gtag );
+
+		// Enhanced conversion data.
+		$ec_data         = [];
```
joemcgill: We should wrap all of this in an `if` statement that checks for the status of whether EC is active and only add this data when the status is 'active' to avoid unnecessary data. This should all be printed before the tracking events as well, so the user data is already set prior to the conversion or purchase events are fired.

joemcgill (changes requested):
Thanks @ankitrox. Left some feedback inline. I don't see any existing unit tests covering this `maybe_display_conversion_and_purchase_event_snippets()` function, but there are some E2E tests at `/tests/e2e/specs/gtag-events/gtag-events.test.js` that could be updated to make sure EC data is added.

src/Google/GlobalSiteTag.php
```diff
@@ -357,62 +357,64 @@ public function maybe_display_conversion_and_purchase_event_snippets( string $ad
...
+			$phone           = $customer->get_billing_phone();
+			$billing_address = $customer->get_billing_address();
+			$postcode        = $customer->get_billing_postcode();
+			$city            = $customer->get_billing_city();
+			$region          = $customer->get_billing_state();
+			$country         = $customer->get_billing_country();
+
+			// Add email in EC data.
+			if ( ! empty( $email ) ) {
+				$ec_data['email'] = $this->normalize_and_hash( $email );
```
joemcgill: Make sure all hashed data is saved using the expected keys described in [these instructions](https://support.google.com/google-ads/answer/13262500?hl=en#zippy=identify-and-define-your-enhanced-conversions-variables%2Cset-up-standard-automatic-enhanced-conversions).

For example, this should be:

```suggestion
				$ec_data['sha256_email_address'] = $this->normalize_and_hash( $email );
```

🔢 Applies here and to the other gtag values.

src/Google/GlobalSiteTag.php
```diff
@@ -578,4 +580,57 @@ private function register_js_for_fast_refresh_dev() {
...
+	 * @return bool
+	 */
+	private function is_enhanced_conversion_enabled(): bool {
+		// Check if enhanced conversion is enabled.
+		$enhanced_conversion_status = $this->options->get( OptionsInterface::ADS_ENHANCED_CONVERSION_STATUS, null );
+		return ( 'enabled' === $enhanced_conversion_status );
+	}
+
+	/**
+	 * Return the hashed data to be sent to Google ads for enhanced conversion.
```
joemcgill:
```suggestion
	 * Return the hashed data to be sent to Google Ads for enhanced conversion.
```

joemcgill (dismissed):
Thanks for the updates, @ankitrox. Left a bit of feedback that needs to be addressed. I'm looking into the E2E tests to see why those are failing as well.

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.4.3
* Theme: Storefront Version: 4.5.4
* WooCommerce - Version 8.6.1
* PHP: 8.0.22 
* Web Server: Nginx
* Browser: Chrome - Version 121
* OS: macOS Sonoma 14.3  
* Git Branch: `update/52310672-google-tag-enhanced-conversion`  


</Details>




**Test Results -** 

- When Enhance Conversation Tracking is enabled from the setting. Tag `' allow_enhanced_conversions':true }` is appearing and tag gtag("set", "user_data", ..) shows user info in encrypted form. 

- When Enhance Conversation Tracking is disabled. Above information does not appear on the order received page. 
 

| When Enhanced Conversion Tracking is Enabled |   |
|--------|--------|
| Screenshot 2024-03-12 at 8 07 16 PM | Screenshot 2024-03-12 at 8 08 38 PM |
|  | Screenshot 2024-03-12 at 8 10 45 PM | 



| When Enhanced Conversion Tracking is Disabled |   |
|--------|--------|
| Screenshot 2024-03-12 at 8 14 41 PM |  Screenshot 2024-03-12 at 8 15 23 PM

joemcgill (commented):
@mikkamp This PR is ready for an initial review. This only adds the enhanced conversion data to the gtag data during checkouts, which will be integrated into the rest of #2242. 

I've updated the testing instructions in the PR details.

We've been able to confirm this is working the way we think it's supposed to work while testing locally during QA, but I'm having trouble getting the E2E tests working and am finding them difficult to debug since the test environment doesn't seem to be set up locally the same way they are set up for the CI Workflow. If you have any advice, I can work on getting those fixed up.

src/Google/GlobalSiteTag.php
```diff
@@ -344,6 +348,66 @@ public function maybe_display_conversion_and_purchase_event_snippets( string $ad
...
+			$customer        = new WC_Customer( $order->get_customer_id() );
+			$email           = $customer->get_billing_email();
+			$fname           = $customer->get_billing_first_name();
+			$lname           = $customer->get_billing_last_name();
+			$phone           = $customer->get_billing_phone();
+			$billing_address = $customer->get_billing_address();
+			$postcode        = $customer->get_billing_postcode();
+			$city            = $customer->get_billing_city();
+			$region          = $customer->get_billing_state();
+			$country         = $customer->get_billing_country();
```
mikkamp: Can you clarify why we need to get this from the customer directly. I just tested this with a guest user and since in those cases the `customer_id` is 0 it doesn't get any details. The billing details can also be fetched directly from the order. Since the order details are about the actual purchase and not the ones stored on the customer account, I think it would be more accurate to get them from the order anyway.

src/Google/GlobalSiteTag.php
```diff
@@ -344,6 +348,66 @@ public function maybe_display_conversion_and_purchase_event_snippets( string $ad
...
+
+				// Add region.
+				if ( ! empty( $region ) ) {
+					$ec_data['address']['sha256_region'] = $this->normalize_and_hash( $region );
+				}
+			}
+
+			$purchase_user_data_gtag = sprintf( 'gtag("set", "user_data", %s)', wp_json_encode( $ec_data ) );
+
+			$this->add_inline_event_script( $purchase_user_data_gtag );
```
mikkamp, line 415: This is working, as in it's adding the snippet to the page:

```js
gtag( "set", "user_data", {
	"sha256_email_address": "2eb9d564b813d572e4ed03bb5a58459a93caa71b7b227338b682fd038b291a3d",
	"sha256_phone_number": "422ce82c6fc1724ac878042f7d055653ab5e983d186e616826a72d4384b68af8",
	"address": {
		"sha256_first_name": "96d9632f363564cc3032521409cf22a852f2032eec099ed5967c0d000cec607a",
		"sha256_last_name": "799ef92a11af918e3fb741df42934f3b568ed2d93ac1df74f1b8d41a27932a6f",
		"sha256_postal_code": "ea4ffb75adc0903ab5ff284da63b2abe9f3591415e97226a77751b93acfd5d3e",
		"sha256_country": "292c1980ba2805512acfef5d0cf8f43fba5c7b9b73a5a7afad1c37cfacad3c98",
		"sha256_street": "6008c26f4452392acb19374bc12a5ec0c360ae17356bce8b786fb128c8720951",
		"sha256_city": "11a62c23412b77477a71481aa2dc7323bcc61d076c8449076c4c58a8356c1bb1",
		"sha256_region": "18ac3e7343f016890c510e93f935261169d9e3f565436429830faf0934f4f8e4"
	}
} )
```

But when I look at the tracking data sent to Google I'm not seeing the valid data being sent. I'm following the instructions: [Validate your implementation using Chrome Developer Tools](https://support.google.com/google-ads/answer/13262500?sjid=1508041180634937993-EU#zippy=%2Cvalidate-your-implementation-using-chrome-developer-tools)

I can see the regular data about the event:
```
data: event=conversion
```

But this part I'm not seeing:

> Look for a parameter “em” with a hashed string as the value. The value should start with “tv.1~em” followed by a long string of characters. If you see the "em" parameter, this means that the enhanced conversions tag is picking up and hashing the enhanced_conversion_data object.

How can I confirm it is actually including the data? It seems the E2E tests are also just checking for the user_data being sent in the page, and not for the data being sent in the tracking request. That would make it closer to a unit test if it confirms the data is output in the page, seeing expected data appear in the tracking request like is done for the other tests would be more of an E2E test.

The other way to validate the data is to check the Ads account, but it says I have to wait 48hours for that, so I'll have to get back to you if it ever appears.

src/Google/GlobalSiteTag.php
```diff
@@ -344,6 +348,66 @@ public function maybe_display_conversion_and_purchase_event_snippets( string $ad
...
+				$phone                          = sprintf( '%s%d', '+', $phone );
+				$ec_data['sha256_phone_number'] = $this->normalize_and_hash( $phone );
+			}
+
+			// Check for required address fields.
+			if ( ! empty( $fname ) && ! empty( $lname ) && ! empty( $postcode ) && ! empty( $country ) ) {
+				$ec_data['address']['sha256_first_name']  = $this->normalize_and_hash( $fname );
+				$ec_data['address']['sha256_last_name']   = $this->normalize_and_hash( $lname );
+				$ec_data['address']['sha256_postal_code'] = $this->normalize_and_hash( $postcode );
+				$ec_data['address']['sha256_country']     = $this->normalize_and_hash( $country );
```
mikkamp: The documentation shows there is only a hashed field available for the `first_name` / `last_name`. Do the other address fields just not have updated documentation?

This documentation specifically mentions not to hash it: https://developers.google.com/google-ads/api/docs/conversions/enhanced-conversions/web#prepare-data

src/Google/GlobalSiteTag.php
```diff
@@ -583,4 +648,57 @@ private function register_js_for_fast_refresh_dev() {
...
+
+	/**
+	 * Return the hashed data to be sent to Google Ads for enhanced conversion.
+	 *
+	 * @param string $value Data that needs to be hashed.
+	 * @param string $algo  Algorithm for hashing.
+	 *
+	 * @return string
+	 */
+	private function normalize_and_hash( string $value, $algo = 'sha256' ): string {
```
mikkamp: Does this function apply all the hashing rules? Looking at this documentation: https://developers.google.com/google-ads/api/docs/conversions/enhanced-conversions/web#php

This one would be missing:
> Remove all periods (.) that precede the domain name in gmail.com and googlemail.com email addresses

mikkamp (commented):
Thanks for implementing the code, it works as outlined in the PR description. However I think it's important that it works for Guest orders as well.

Also I'm not able to actually confirm that the data gets sent with the conversion event. What else am I missing to get the data to be sent? It's something we should validate in the E2E test as well.

src/Google/GlobalSiteTag.php
ankitrox: @mikkamp Thank you for catching that. As per documentation, it seems that we need to normalize and has 

1. First name
2. Last name
3. Phone number
4. Email address

Rest of the fields have been changed to the raw format (without hash).

src/Google/GlobalSiteTag.php
ankitrox: Added the normalization to remove any periods in the email address if has domain gmail.com or googlemail.com

src/Google/GlobalSiteTag.php
ankitrox: You are right, I can't see the `em` property as mentioned in the document, but I think if its possible we should wait to check if it is being reflected in ads account after 48 hours as we did all the steps according to the documentation.

src/Google/GlobalSiteTag.php
ankitrox: Thank you. I have changed this to use `$order` object instead of customer to add the data which would be more accurate as you mentioned.

joemcgill:
@ankitrox and @mikkamp, I gave this another test and noticed a few things that got messed up during the previous merge conflict resolution, which I've resolved in [c2d73c2](https://github.com/woocommerce/google-listings-and-ads/pull/2258/commits/c2d73c25fadc8ad257a41c6635a7b8f1e947306b).

From what I can tell, all of the data is being hashed correctly now, per the docs, but I am also not able to verify the presence of an `em` param so far. I'm not sure if this is due to me testing on a local environment or if there is something else that needs to be updated before this will work, but I'm not seeing any functional errors at this point.

One of the assumptions that this implementation is based on is that Google will automatically enable enhanced conversions if the Data Terms and Conditions has been accepted and a conversion tag is sent that includes `{ 'allow_enhanced_conversions': true }` in the config along with hashed PII, so that's part of what we're trying to validate during UAT. If the code looks good, I'd suggest we merge this into the rest of the feature and continue testing there.

joemcgill (approved):

mikkamp (approved):
Thanks for the additional changes. I agree that everything is now working as intended, the code is working as expected.

The only minor comment I had left is that the function `maybe_display_conversion_and_purchase_event_snippets` is now over a 150 lines long. To add the enhanced conversion data we really only need the order object, so all that code can easily be moved to a separate function (including the call to `add_inline_event_script`). That would also allow returning early from the function `if ( ! $this->is_enhanced_conversion_enabled() )`, which would save a whole lot of indenting and make the code a bit easier to grasp (as well as unit test if we'd resort to that).

That's not a blocker for this PR though so I'll go ahead and approve it.

I did manage to get the enhanced conversion data to show up in the end, but to do so I must enable it in the Ads account (the method I choose doesn't seem to matter, but as soon as I disable it, the data is missing again).
image

With it enabled I can confirm that the hashes match and the additional address fields are also sent without hashes:

```js
gtag( "set", "user_data", {
	"sha256_email_address": "2eb9d564b813d572e4ed03bb5a58459a93caa71b7b227338b682fd038b291a3d",
	"sha256_phone_number": "422ce82c6fc1724ac878042f7d055653ab5e983d186e616826a72d4384b68af8",
	"address": {
		"sha256_first_name": "96d9632f363564cc3032521409cf22a852f2032eec099ed5967c0d000cec607a",
		"sha256_last_name": "799ef92a11af918e3fb741df42934f3b568ed2d93ac1df74f1b8d41a27932a6f",
		"postal_code": "D02 AF30",
		"country": "IE",
		"street": "Street",
		"city": "City",
		"region":"D"
	}
} )
```

```
data: event=conversion
em: tv.1~
  em.2eb9d564b813d572e4ed03bb5a58459a93caa71b7b227338b682fd038b291a3d~
  pn.422ce82c6fc1724ac878042f7d055653ab5e983d186e616826a72d4384b68af8~
  fn0.96d9632f363564cc3032521409cf22a852f2032eec099ed5967c0d000cec607a~
  ln0.799ef92a11af918e3fb741df42934f3b568ed2d93ac1df74f1b8d41a27932a6f~
  sa0.YAjCb0RSOSrLGTdLwSpewMNgrhc1a86LeG-xKMhyCVE~
  ct0.city~
  pc0.D02 AF30~
  rg0.d~
  co0.IE
```

That seems to show that we are still missing hashing the street, which the documentation isn't consistent about. I'm also curious why the hashes look very different, as far as I can see we are following the documentation to the T, so we might need to ask Google why that's happening.

While that's not a blocker for this PR, it is something that needs to be addressed in #2242 As there it's the expectation that users only accept the terms and conditions. They aren't required to enable the Enhanced Conversions in the Ad account. I also believe Google mentioned there isn't a way to automatically enable it through the API. So that needs to be discussed a bit more otherwise it still won't track anything for users that don't enable it.

joemcgill:
@mikkamp thank you for double checking the `em` data. I can confirm that I'm seeing the same behavior that you are describing where the data is only passed if I've manually enabled EC in the Ads account. I'll bring this up with folks at Google since it's not working as we were advised.

I also agree with you about the potential code cleanup for the `maybe_display_conversion_and_purchase_event_snippets()`  snippet, so moved that code to a private method for now and added hashing to the street value, which isn't specified in [this article](https://support.google.com/google-ads/answer/13262500?hl=en#zippy=identify-and-define-your-enhanced-conversions-variables%2Cset-up-standard-automatic-enhanced-conversions%2Cenable-enhanced-conversions-in-google-tag-manager-and-create-custom-javascript-variable) but is recommended [here](https://developers.google.com/google-ads/api/docs/conversions/enhanced-conversions/web#prepare-data) and I'd rather be overly cautious about PII.

I'm going to go ahead and merge this and then we can address any follow-up tasks related to the EC values not being passed in follow-up PRs.

PR #2261, opened 2024-02-19
Add - Product View Report Test

jorgemd24:

This PR introduces:

- Tests for the Product View Report.
- Removed the `product_view.offer_id` from the MerchantQuery, as we're using the MC-ID to extract the `offerID`. The product_view.id is necessary in the query and cannot be removed.
- I've updated the default placeholder for `$product_view_data` from `next_page` to `next_page_token` to maintain consistency with the rest of the code.

It looks like the report for free listings isn't covered by tests, so I'll create a follow-up issue to add tests for that section.

1. Run the tests `./vendor/bin/phpunit`

### Changelog entry

>

### DIFF
--- a/tests/Unit/API/Google/MerchantReportTest.php
+++ b/tests/Unit/API/Google/MerchantReportTest.php
@@ -15,8 +15,10 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingCo
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\SearchRequest;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\SearchResponse;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\Resource\Reports;
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Exception as GoogleException;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\ShoppingContentDateTrait;
 use DateTime;
+use Exception;
 use PHPUnit\Framework\MockObject\MockObject;
 
 defined( 'ABSPATH' ) || exit;
@@ -79,7 +81,7 @@ class MerchantReportTest extends UnitTest {
 		return $product_view;
 	}
 
-	public function test_get_free_listing_metrics() {
+	public function test_get_product_view_report() {
 		$test_merchant_id = 432;
 		$wc_product_id_1  = 882;
 		$wc_product_id_2  = 883;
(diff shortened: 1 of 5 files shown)


tests/Unit/API/Google/MerchantReportTest.php
```diff
@@ -0,0 +1,182 @@
...
+		$wc_product_id_1  = 882;
+		$wc_product_id_2  = 883;
+		$page_size        = 800;
+
+		add_filter(
+			'woocommerce_gla_product_view_report_page_size',
+			function () use ( $page_size ) {
+				return $page_size;
+			}
+		);
```
martynmjones: Could you clarify why we're using the `woocommerce_gla_product_view_report_page_size` filter here? From what I can tell it has no impact on the test itself so is the intention of this to give a more complete picture of the functionality?

tests/Unit/API/Google/MerchantReportTest.php
```diff
@@ -0,0 +1,182 @@
...
+					],
+				],
+				'next_page_token' => null,
+			],
+			$this->merchant_report->get_product_view_report()
+		);
+	}
+
+	public function test_get_product_view_report_with_exception() {
+		$this->options->method( 'get_merchant_id' )->willReturn( 432 );
```
martynmjones: We're also setting the return value for `get_merchant_id` on the options mock in `test_get_product_view_report` but beyond that we don't do anything with it so instead of repeating this had you considered creating a class property for it and then assigning it to the mock in `setUp`?

tests/Unit/API/Google/MerchantReportTest.php
```diff
@@ -0,0 +1,182 @@
...
+		$this->merchant_report = new MerchantReport( $this->shopping_client, $this->product_helper );
+		$this->merchant_report->set_options_object( $this->options );
+	}
+
+	/**
+	 * Creates a product view with the given id, status and expiration date.
+	 *
+	 * @param string        $mc_id The MC center id.
+	 * @param string        $status The status of the product.
+	 * @param DateTime|null $expiration_date The expiration date of the product.
```
martynmjones, line 72: Do we need the `$expiration_date` parameter? When we set up the product view's in `test_get_product_view_report` it's not used.

martynmjones (approved):
Hey @jorgemd24, thanks for your work on these tests.

I've left a few comments but nothing is blocking and all tests are passing locally so approving ✅

tests/Unit/API/Google/MerchantReportTest.php
jorgemd24: I use the `$expiration_date` to create a "GoogleDate", which is then used to simulate the expiration_date here:

https://github.com/woocommerce/google-listings-and-ads/pull/2261/files#diff-91e7d57ff211a17cd8684387ddd428407e3530dfca1487dd2ccc4fa022bec0eaR160

I could set a DateTime in the expiration_date response, but I believe this way allows us to simulate a more realistic scenario.

tests/Unit/API/Google/MerchantReportTest.php
jorgemd24: >  so is the intention of this to give a more complete picture of the functionality?

That's right, this way allows me to test that the filter is used to set the page size, and that the search is receiving the correct page size.

https://github.com/woocommerce/google-listings-and-ads/pull/2261/files#diff-91e7d57ff211a17cd8684387ddd428407e3530dfca1487dd2ccc4fa022bec0eaR146

tests/Unit/API/Google/MerchantReportTest.php
jorgemd24: Thanks for the suggestion, I added here: https://github.com/woocommerce/google-listings-and-ads/pull/2261/commits/6a9a0f96d577930181961b6c7fed1587a6225513

PR #2263, opened 2024-02-20
Fix Issue with intermediate parents status

jorgemd24:

Follow-up of https://github.com/woocommerce/google-listings-and-ads/pull/2257.

Now that we're calculating product statuses in batches, it's possible for variations from the same parents to come in different batches. For example:

Batch 1: Variable 1 - Parent ID 1 - DISAPPROVED
Batch 2: Variable 2 - Parent ID 1 - APPROVED

This could impact the total count because variable products are grouped by their parents. So, in the previous example, we should only count 1 DISAPPROVED product instead of 1 DISAPPROVED + 1 APPROVED.

This PR addresses the issue by storing the latest status from the parent product, ensuring it can be used for the total calculation.

1. It's a bit of a tricky situation to test, so I mainly relied on the Unit Tests to simulate the scenario. However, you can create batches of 1, each with a parent product having different variables, and check if the total count is correct. To do so identify a variable product in your store, and get a couple of variable ids with the same parent id.
2. Check that those variables are available in your Merchant Center.
3. Go to https://github.com/woocommerce/google-listings-and-ads/blob/d7ecd99a5bb010ea371b62f2578dbf5c503ac4ea/src/API/Google/Query/MerchantProductViewReportQuery.php#L22-L28 and add the following line: `$this->where( 'product_view.offer_id', ['gla_VARIABLE_ID_1', 'gla_VARIABLE_ID_2'], 'IN' ); `
4. Change the batch size to `1` in this line: https://github.com/woocommerce/google-listings-and-ads/blob/d7ecd99a5bb010ea371b62f2578dbf5c503ac4ea/src/API/Google/MerchantReport.php#L68
5. Call the following endpoint `GET wc/gla/mc/product-statistics/refresh`
6. You should get only one status. 

### Changelog entry

>

### DIFF
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -728,7 +728,7 @@ class MerchantStatuses implements Service, ContainerAwareInterface, OptionsAware
 		// If the option is set, use it to sum the total quantity.
 		$product_statistics_intermediate_data = $this->options->get( OptionsInterface::PRODUCT_STATUSES_COUNT_INTERMEDIATE_DATA );
 		if ( $product_statistics_intermediate_data ) {
-			$product_statistics = $product_statistics + ['parents' => []];
+			$product_statistics = $product_statistics + [ 'parents' => [] ];
 		}
 
 		$product_statistics_priority = [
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Thanks for addressing the problem of status priority in different batches. I tested using a small set of products and the calculation looked great. 👍

PR #2266, opened 2024-02-20
401 handling for connected Ads accounts

mikkamp:

We have a universal handler for any response with a 401 status code. This handler determines if it's either the Jetpack or the Google connection that no longer has the required permission and then handles it with a ReconnectAccount response.

Unfortunately though there is a scenario when listing Ads accounts that also throws a 401 error if the user has never signed up to Ads yet. Which is indicated by the following error message:
```json
"errorCode": {
    "authenticationError": "NOT_ADS_USER"
},
"message": "The Google account (@gmail.com user) that generated the OAuth access tokens is not associated with any Ads accounts. Create a new account, or add the Google account to an existing Ads account."
```

We previously handled this scenario by returning an empty list of existing accounts. However that is no longer possible since the error handler catches it earlier. So this PR prevents the error handler from catching the error so it can be handled correctly by the call to fetch a list of Ad accounts.

I originally wanted to set a flag in the Client class that when the flag is set we shouldn't handle 401's, however with the way the handler stack is being built, I didn't find a good way to do this. I also considered setting a header in the request, but this seemed like it would be similar to just checking the request path.

I've also added a small test to the site controller to expect the reconnect Google account exception. We are already testing for an empty list to be returned when a `NOT_ADS_USER` response is returned. ~~But unfortunately we don't have a full test that makes use of all the Guzzle handlers, so it won't catch our custom error handler preventing the error. We'd need a better integration test for that, which would be out of scope of a unit test.~~
Edit: I've found a way to mock the handlers and test them individually with various mocked requests. This PR includes unit tests to confirm the various scenarios in the `error_handler` are handled as intended.

1. Start with a site with all accounts disconnected
2. Connect a WP.com account
3. Connect a Google account (make sure it's an account that never had an Ads account connected)
4. Use the connection test page to fetch a list of customers "Get Customers from Google Ads"
5. Expect it to return an empty list "Total accounts: 0"
6. Confirm that the full exception is logged in WooCommerce > Status > Logs:
```json
{
    "message": "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https:\/\/developers.google.com\/identity\/sign-in\/web\/devconsole-project.",
    "code": 16,
    "status": "UNAUTHENTICATED",
    "details": [
        {
            "@type": "type.googleapis.com\/google.ads.googleads.v14.errors.GoogleAdsFailure",
            "errors": [
                {
                    "errorCode": {
                        "authenticationError": "NOT_ADS_USER"
                    },
                    "message": "The Google account (@gmail.com user) that generated the OAuth access tokens is not associated with any Ads accounts. Create a new account, or add the Google account to an existing Ads account."
                }
            ],
            "requestId": "duBXnLB1SNjhc5DM3tB10w"
        }
    ]
}
```

### Changelog entry
* Fix - 401 handling for connected Ads accounts.

### DIFF
--- a/tests/Unit/API/ClientTest.php
+++ b/tests/Unit/API/ClientTest.php
@@ -91,6 +91,21 @@ class ClientTest extends ContainerAwareUnitTest {
 	}
 
 	/**
+	 * Confirm that a request to listAccessibleCustomers does not return a redirect error.
+	 */
+	public function test_error_handler_list_accessible_customers() {
+		$mocked_responses = [
+			new Response( 401, [], 'error' ),
+		];
+
+		$this->expectException( RequestException::class );
+		$this->expectExceptionMessage( 'error' );
+
+		$client   = $this->mock_client_with_handler( 'error_handler', $mocked_responses );
+		$response = $client->request( 'GET', 'https://testing.local/google/google-ads/customers:listAccessibleCustomers' );
+	}
+
+	/**
 	 * Confirm that the error handler throws a generic error when the status code is higher than 400 except a 401.
 	 */
 	public function test_error_handler_generic_error_response() {
--- a/tests/Unit/API/ClientTest.php
+++ b/tests/Unit/API/ClientTest.php
@@ -3,18 +3,18 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API;
 
-use Automattic\WooCommerce\GoogleListingsAndAds\Container;
 use Automattic\WooCommerce\GoogleListingsAndAds\Exception\AccountReconnect;
-use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\ContainerAwareUnitTest;
+use Automattic\WooCommerce\GoogleListingsAndAds\Internal\DependencyManagement\GoogleServiceProvider;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\GuzzleHttp\Client;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\GuzzleHttp\Exception\RequestException;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\GuzzleHttp\Handler\MockHandler;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\GuzzleHttp\HandlerStack;
-use Automattic\WooCommerce\GoogleListingsAndAds\Internal\DependencyManagement\GoogleServiceProvider;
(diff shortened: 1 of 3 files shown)


ianlin (commented):
Thanks for handling the error. I can see the error being logged correctly after clicking `Get Customers from Google Ads`.

Then I connected to MC account and was trying to test `Setup an existing account or create a new one` with an empty Ads ID from the connection page, the response message on the connection page ask me to reconnect to Google account, which is similar to the problem that was solved in this PR. Maybe there were more cases like this that perhaps need to be skipped in the universal handler?

mikkamp:
> Then I connected to MC account and was trying to test Setup an existing account or create a new one with an empty Ads ID from the connection page, the response message on the connection page ask me to reconnect to Google account, which is similar to the problem that was solved in this PR. Maybe there were more cases like this that perhaps need to be skipped in the universal handler?

Thanks for testing @ianlin . Can you clarify on which exact request that failed. You mentioned you left the Ads ID empty, which would trigger the `create_ads_account`. This is a request which is sent to the WCS and creates a new Ad account on their behalf, and invites them to accept it, during this process there is only a request sent by the manager account which shouldn't return a permission denied error.
Currently any interaction with this newly created account is then blocked until they complete the billing step, which also has them accept the account (once they do that they will no longer be in the state which returns a `NOT_ADS_USER`). So I'm wondering in your case which request was the one that triggered the 401 response, did you happen to use a snippet to skip the billing step? If so then the response you received was correct, and the action would be to manually accept the new Ads account through the link in the email, before you could proceed further. Regular users shouldn't enter into this flow though.

ianlin:
> did you happen to use a snippet to skip the billing step

Ah you're right, I did skip the billing step. I will retest it tomorrow.

mikkamp:
> Ah you're right, I did skip the billing step. I will retest it tomorrow.

Thanks, I wasn't too happy with the coverage in unit tests, so I went ahead and found a way to mock the requests so we can check the responses or confirm the headers are set as expected. Let me know if that's all testing as expected.

tests/Unit/API/ClientTest.php
```diff
@@ -56,7 +76,28 @@ protected function invoke_handler( string $handler_function ) {
...
+	 *
+	 * @param string $handler_function Handler function name to include in stack.
+	 * @param array  $mocked_responses List of responses to return.
+	 *
+	 * @return Client Mock client.
+	 */
+	protected function mock_client_with_handler( string $handler_function, array $mocked_responses ) {
+		$mock     = new MockHandler( $mocked_responses );
+		$handlers = HandlerStack::create( $mock );
+		$handlers->push( $this->invoke_handler( 'error_handler' ) );
```
ianlin: I think it should be using the variable `$handler_function`.

ianlin (approved):
I've tested it again without the skipping billing step and also tried to accept the Ads account invitation from the email. Everything worked as expected.

Thanks for adding more unit tests. There was just one mistyped thing that need to be changed in the unit test so I'm going to approve the PR now.

tests/Unit/API/ClientTest.php
mikkamp: Nice catch, yes that's absolutely correct. Side effect of refactoring.

PR #2267, opened 2024-02-20
API PULL - Fix Notifications Body Encoding

puntope:

There is an issue when sending notifications. We are getting `invalid_body_hash`. This seems to happen on `is_jetpack_authorized_for_site` in WPCOM when trying to compare the body hash.

Seems like its failing due the encoding. This PR fix that.

1. Go to Connection test page
2. Send a notification request
3. See you get a failing error
4. Go to WooCommerce Status - Logs and get the last logs for GLA
5. See the request failed because  `invalid_body_hash`
6. Checkout this PR
7. Repeat steps 1-2
8.  See the request is successfull

### DIFF
--- a/src/Google/NotificationsService.php
+++ b/src/Google/NotificationsService.php
@@ -116,7 +116,7 @@ class NotificationsService implements Service {
 	 * @return array|\WP_Error
 	 */
 	protected function do_request( array $args ) {
-		return Client::remote_request( $args, wp_json_encode( $args['body'] ) );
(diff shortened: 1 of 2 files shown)


jorgemd24 (approved):
Hi @puntope, thanks for fixing this. The issue with the hash is now fixed! Although, I'm running into a different error: `{"code":"wpcom_partner_notifications_partner_error","message":"The caller does not have permission","data":{"status":403}}.` From what I gather, this is because Google needs to authorize the site, correct?

PR #2268, opened 2024-02-20
API PULL - Add Settings Notification

puntope:

This PR adds Notifications for Settings.

⚠️  Notice there are some pending PRs and Diffs to merge so the Notification won't be successful. Only it needs to check if it's triggered.

⚠️  Ideally the notifications only trigger when WooCommerce General settings are updated. However, the only method found that supports UI + REST + Programatically updating options was `update_option` hook. So, I added a check `woocommerce_` prefix to reduce the number of times the notifications are done. But it still triggers non-general Woocommerce Notifications.  If programmatically way is not needed yes there are some other hooks to filter only General settings.

### Screenshots:

Screenshot 2024-02-20 at 18 46 48

1. Checkout this PR
2. Update Settings via WP Admin - Woocommerce - Settings 
3. See the action `notifications/settings` is scheduled. Run it.
4. Update Settings via[ Rest API ](https://woocommerce.github.io/woocommerce-rest-api-docs/#settings)
5. See the action `notifications/settings` is scheduled. Run it.
6. Update Settings programmatically via `update_option`
7. See the action `notifications/settings` is scheduled. Run it.

### DIFF
--- a/src/Internal/DependencyManagement/JobServiceProvider.php
+++ b/src/Internal/DependencyManagement/JobServiceProvider.php
@@ -24,6 +24,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\JobInitializer;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\JobInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\JobRepository;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications\ProductNotificationJob;
+use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications\SettingsNotificationJob;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications\ShippingNotificationJob;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\ProductSyncerJobInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\ProductSyncStats;
@@ -48,6 +49,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Coupon;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
 use Automattic\WooCommerce\GoogleListingsAndAds\Shipping;
+use Automattic\WooCommerce\GoogleListingsAndAds\Settings;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -151,6 +153,15 @@ class JobServiceProvider extends AbstractServiceProvider {
 			ShippingNotificationJob::class,
 			NotificationsService::class
 		);
+
+		// Share settings notifications job
+		$this->share_action_scheduler_job(
+			SettingsNotificationJob::class,
+			NotificationsService::class
+		);
+
+		// Share settings syncer hooks
+		$this->share_with_tags( Settings\SyncerHooks::class, JobRepository::class, NotificationsService::class );
 
 		// Share shipping settings syncer job and hooks.
 		$this->share_action_scheduler_job( UpdateShippingSettings::class, MerchantCenterService::class, GoogleSettings::class );
--- a/src/Settings/SyncerHooks.php
+++ b/src/Settings/SyncerHooks.php
@@ -60,25 +60,10 @@ class SyncerHooks implements Service, Registerable {
 
 		$update_rest = function ( $option ) {
 			if ( strpos( $option, 'woocommerce_' ) === 0 ) {
-				$this->handle_update_shipping_settings();
+				$this->settings_notification_job->schedule();
 			}
 		};
 
 		add_action( 'update_option', $update_rest, 90, 1 );
-	}
-
-	/**
-	 * Handle updating of Merchant Center shipping settings.
-	 *
-	 * @return void
-	 */
-	protected function handle_update_shipping_settings() {
-		// Bail if an event is already scheduled in the current request
-		if ( $this->already_scheduled ) {
-			return;
-		}
-
-		$this->settings_notification_job->schedule();
-		$this->already_scheduled = true;
 	}
 }
--- a/src/Settings/SyncerHooks.php
+++ b/src/Settings/SyncerHooks.php
@@ -23,13 +23,6 @@ defined( 'ABSPATH' ) || exit;
 class SyncerHooks implements Service, Registerable {
 
 	/**
-	 * This property is used to avoid scheduling duplicate jobs in the same request.
-	 *
-	 * @var bool
-	 */
-	protected $already_scheduled = false;
-
-	/**
 	 * @var NotificationsService $notifications_service
 	 */
 	protected $notifications_service;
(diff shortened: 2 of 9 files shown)


src/Internal/DependencyManagement/JobServiceProvider.php
```diff
@@ -152,6 +154,15 @@ public function register(): void {
...
 		);
 
+		// Share settings notifications job
+		$this->share_action_scheduler_job(
+			SettingsNotificationJob::class,
+			NotificationsService::class
+		);
+
+		// Share settings syncer hooks
+		$this->share_with_tags( Settings\SyncerHooks::class, JobRepository::class, NotificationsService::class );
```
jorgemd24, line 164: Just curious, why do you prefer to do it like Settings\SyncerHooks instead of setting the `use Automattic\WooCommerce\GoogleListingsAndAds\Settings\SyncerHooks` as `SettingsSyncerHooks`? I am asking because I think it is first time that I see it in the codebase.

src/Settings/SyncerHooks.php
```diff
@@ -0,0 +1,84 @@
...
+
+		add_action( 'update_option', $update_rest, 90, 1 );
+	}
+
+	/**
+	 * Handle updating of Merchant Center shipping settings.
+	 *
+	 * @return void
+	 */
+	protected function handle_update_shipping_settings() {
```
jorgemd24: Should we change this to `handle_update_settings`?

src/Settings/SyncerHooks.php
```diff
@@ -0,0 +1,84 @@
...
+		// Bail if an event is already scheduled in the current request
+		if ( $this->already_scheduled ) {
+			return;
+		}
```
jorgemd24: Why do we keep track if the job is scheduled within this class rather than relying on the job's methods to check if it's running or can be scheduled?

https://github.com/woocommerce/google-listings-and-ads/blob/1af92142945aa2f5a719f1c23ec3255938620a2f/src/Jobs/AbstractActionSchedulerJob.php#L67-L69

tests/Unit/Jobs/SettingsNotificationJobTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Jobs;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\ActionScheduler\ActionScheduler;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\NotificationsService;
+use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\ActionSchedulerJobMonitor;
+use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications\SettingsNotificationJob;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
+use PHPUnit\Framework\MockObject\MockObject;
+use WC_Helper_Product;
```
jorgemd24: Unused dependency

jorgemd24 (commented):
Hi @puntope,

It is working as expected however, as you pointed out, it triggers the notification job for numerous options, which can quickly fill the AS tables. For example, simply updating the store address using the UI will schedule the job for the following options:

```
[Wed Feb 21 10:43:01.965766 2024] [php:notice] [pid 89761] [client ::1:50432] woocommerce_store_address, 
[Wed Feb 21 10:43:01.970920 2024] [php:notice] [pid 89761] [client ::1:50432] woocommerce_price_num_decimals, 
[Wed Feb 21 10:43:01.971188 2024] [php:notice] [pid 89761] [client ::1:50432] woocommerce_ces_tracks_queue, 
[Wed Feb 21 10:43:01.972507 2024] [php:notice] [pid 89761] [client ::1:50432] woocommerce_queue_flush_rewrite_rules, 
[Wed Feb 21 10:43:03.285093 2024] [php:notice] [pid 89761] [client ::1:50432] woocommerce_queue_flush_rewrite_rules, 
[Wed Feb 21 10:43:03.296233 2024] [php:notice] [pid 56806] [client ::1:50464] woocommerce_queue_flush_rewrite_rules, 
[Wed Feb 21 10:43:03.319742 2024] [php:notice] [pid 56834] [client ::1:50467] woocommerce_queue_flush_rewrite_rules, 
[Wed Feb 21 10:43:03.336927 2024] [php:notice] [pid 56803] [client ::1:50470] woocommerce_queue_flush_rewrite_rules, 
[Wed Feb 21 10:43:03.864014 2024] [php:notice] [pid 56806] [client ::1:50464] woocommerce_clear_ces_tracks_queue_for_page, 
[Wed Feb 21 10:43:05.304488 2024] [php:notice] [pid 64292] [client ::1:50634] woocommerce_ces_tracks_queue
[Wed Feb 21 10:43:05.309556 2024] [php:notice] [pid 64292] [client ::1:50634] woocommerce_clear_ces_tracks_queue_for_page
```

I was thinking, wouldn't it be better to whitelist the options we want to notify to avoid this issue? For instance, we could follow the IDs from this list:


https://github.com/woocommerce/woocommerce/blob/af03815134385c72feb7a70abc597eca57442820/plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php#L34

Also, I left some comments about the code.

src/Internal/DependencyManagement/JobServiceProvider.php
puntope: I think both ways are valid. I used this way because it's how is done with the other SyncerHooks regarding Products and Shipping.

puntope:
> It is working as expected however, as you pointed out, it triggers the notification job for numerous options, which can quickly fill the AS tables. For example, simply updating the store address using the UI will schedule the job for the following options:

The `can_schedule()` function prevents this to happen. I saw that it was not set up properly thanks to that comment. 

I did a fix for it. Can you verify again?

Regarding your suggestion. I thought in the whitelist as well. But that would not trigger if some 3rd party plugin adds some custom settings in the general tab, right?

puntope:
Hi @jorgemd24 I applied most of your suggestions and answered your concerns [here](https://github.com/woocommerce/google-listings-and-ads/pull/2268#issuecomment-1957491328).

Notice there are some issues (the same with can_schedule) with the Product and Shipping Notifications but I can fix them after refactor is done in a separate PR.

src/Settings/SyncerHooks.php
```diff
@@ -0,0 +1,69 @@
...
+ * @since x.x.x
+ */
+class SyncerHooks implements Service, Registerable {
+
+	/**
+	 * This property is used to avoid scheduling duplicate jobs in the same request.
+	 *
+	 * @var bool
+	 */
+	protected $already_scheduled = false;
```
jorgemd24: I think this can be now deleted as it is not used.

jorgemd24 (commented):
Hi @puntope!

Thanks for the fixes!

>The can_schedule() function prevents this to happen. I saw that it was not set up properly thanks to that comment.

Yeah, it reduces the number of triggered jobs. Now, every time I update the address, it will trigger the job twice, which is better than before. 

>I thought in the whitelist as well. But that would not trigger if some 3rd party plugin adds some custom settings in the general tab, right?

I don't see that as a problem because Google only requires the store address and certain settings from the response of this endpoint: `GET wc/v3/settings/general`. I don't think we need to notify about custom options.

Another issue I see with the current approach is that updating any setting will trigger the notification. For example, if you go to WC -> Settings -> Products and update the placeholder image, it will also trigger the notification. I suspect there are many other ways that any `woocommerce_` option can be updated.

jorgemd24:
Another thing that I noticed is that on the connection test page we are getting multiple warnings when the topic is not set:

image

puntope:
> Yeah, it reduces the number of triggered jobs. Now, every time I update the address, it will trigger the job twice, which is better than before. I don't see that as a problem because Google only requires the store address and certain settings from the response of this endpoint: GET wc/v3/settings/general. I don't think we need to notify about custom options.

For me, it only schedules once. But I added your suggestion.

> Another issue I see with the current approach is that updating any setting will trigger the notification. For example, if you go to WC -> Settings -> Products and update the placeholder image, it will also trigger the notification. I suspect there are many other ways that any woocommerce_ option can be updated.

I added your suggestion as seems not needed to process other settings than the original General Settings.

puntope:
Hi @jorgemd24 Thanks a lot for the review again.

I did apply your suggestions. This is ready for a new round.

jorgemd24 (approved):
Thanks, @puntope! I tested updating the general settings, and now I only see one notification, so LGTM.

PR #2271, opened 2024-02-21
Product Statuses - Handle failure of the Update Product Statuses Job

jorgemd24:

Follow-up of https://github.com/woocommerce/google-listings-and-ads/pull/2257.

Since we're calculating product statuses using AS jobs, there's a possibility that the job might fail when attempting to retrieve or update the data. Therefore, it's important to store this error so we can notify the client about any issues while updating the product statuses.

Action Scheduler will retry the job, and if it succeeds, the process will continue as usual and the error message will be overridden once all statuses are completed.

However, if the job fails permanently and the failure rate exceeds the threshold, AS will stop the job. In this case, when the client requests the data via `GET wc/gla/mc/product-statistics`, the response will contain the error message.

```
{
    "timestamp": 1708537779,
    "statistics": {},
    "scheduled_sync": 0,
    "loading": false,
    "error": "My fatal error"
}
```

1. Produce an error in purpose, for example, add the following line `throw new \Error('My fatal error');`  in https://github.com/woocommerce/google-listings-and-ads/blob/b544a507e040cadc421b94391e0b9524639d1483/src/API/Google/MerchantReport.php#L67
2. Make a request to `GET wc/gla/mc/product-statistics/refresh`
3. Go to WooCommerce -> Statuses -> Scheduled Actions and search for `gla/jobs/update_merchant_product_statuses/process_item` and see that the jobs fails.
4. Call `wc/gla/mc/product-statistics` and see that the error is present in the transient. 

### Additional details:
- As discussed here https://github.com/woocommerce/google-listings-and-ads/pull/2257#issuecomment-1953823315  the idea is for the client to be able to view the error and have the choice to manually retry the statuses.

### Changelog entry

>

### DIFF
--- a/src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php
+++ b/src/API/Site/Controllers/MerchantCenter/ProductStatisticsController.php
@@ -176,6 +176,13 @@ class ProductStatisticsController extends BaseOptionsController {
 				'context'     => [ 'view' ],
 				'readonly'    => true,
 			],
+			'error'          => [
+				'type'        => 'string',
+				'description' => __( 'Error message in case of failure', 'google-listings-and-ads' ),
+				'context'     => [ 'view' ],
+				'readonly'    => true,
+				'default'     => null,
+			],
 		];
 	}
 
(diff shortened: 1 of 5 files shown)


ianlin (approved):
Thanks for handling failures of update product statuses job. I can see the error in the response data when calling `wc/gla/mc/product-statistics` endpoint. LGTM.

jorgemd24:
Thanks @ianlin for your time reviewing this PR!

PR #2273, opened 2024-02-22
API PULL - Refactor Notification Jobs

puntope:

Refactor Notifications Jobs and Tests for reducing duplications.

After emerging the work related to coupons notifications this PR does:

- Reactor all the generic methods shared between `SettingsNotificationsJob`, `ShippingNotificationsJob`, `CouponNotificationJob`,  `ProductNotificationJob`, into an abstract `AbstractNotificatioJob` class.
- Refactor `CouponNotificationJob` and `ProductNotificationJob` (Job using Item IDs)  into an abstract `AbstractNotificatioItemJob` 
- Refactor the tests (also using abstract tests)

⚠️  Some methods and classes (mostly regarding SyncerHooks, Helpers and MetaHandlers) can be refactored too. But are out scope of the Project. This might be done after adopting the new API PULL strategy and deprecating the current MC methods completely.

`NotificationsService.php`
```php
     return true;
}
```

1. Create a product -- See `gla/jobs/notifications/products/process_item with the product and the `product.create `topic. Run it
2. Update the product -- See `gla/jobs/notifications/products/process_item with the product and the `product.update `topic. Run it
3. Delete the product -- See `gla/jobs/notifications/products/process_item with the product and the `product.delete `topic. Run it
4. Restore the product. See `gla/jobs/notifications/products/process_item with the product and the `product.create `topic
5. Create a coupon -- See `gla/jobs/notifications/coupons/process_item with the coupon and the `coupon.create `topic. Run it
6. Update the coupon -- See `gla/jobs/notifications/coupons/process_item with the coupon and the `coupon.update `topic. Run it
7. Delete the coupon -- See `gla/jobs/notifications/coupons/process_item with the coupon and the `coupon.delete `topic. Run it
8. Update the WooCommerce general Settings -- See `gla/jobs/notifications/settings/process_item without ID and the `settings.update` topic.
9.  Update/Delete/Create any WooCommerce Shipping -- See `gla/jobs/notifications/shipping/process_item without ID and the `settings.update` topic.

### DIFF
--- a/src/Jobs/Notifications/AbstractItemNotificationJob.php
+++ b/src/Jobs/Notifications/AbstractItemNotificationJob.php
@@ -24,7 +24,7 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 	 * @param array $args Arguments with the item id and the topic
 	 */
 	protected function process_items( array $args ): void {
-		if ( ! isset( $args[0] ) || ! isset( $args[1] ) ) {
+		if ( ! isset( $args['item_id'] ) || ! isset( $args['topic'] ) ) {
 			do_action(
 				'woocommerce_gla_error',
 				'Error sending the Notification. Topic and Item ID are mandatory',
@@ -33,8 +33,8 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 			return;
 		}
 
-		$item  = $args[0];
-		$topic = $args[1];
+		$item  = $args['item_id'];
+		$topic = $args['topic'];
 
 		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item ) ) {
 			$this->set_status( $item, $this->get_after_notification_status( $topic ) );
--- a/src/Jobs/Notifications/AbstractItemNotificationJob.php
+++ b/src/Jobs/Notifications/AbstractItemNotificationJob.php
@@ -1,0 +1,122 @@
+<?php
+declare(strict_types=1);
+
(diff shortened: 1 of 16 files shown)


src/Jobs/Notifications/AbstractItemNotificationJob.php
```diff
@@ -0,0 +1,122 @@
...
+	 * the Job process the item.
+	 *
+	 * @param int    $item_id
+	 * @param string $topic
+	 * @return bool
+	 */
+	protected function can_process( int $item_id, string $topic ): bool {
+		$item = $this->get_item( $item_id );
+
+		if ( str_contains( $topic, '.create' ) ) {
```
ianlin, line 84: To echo the comment https://github.com/woocommerce/google-listings-and-ads/pull/2278#discussion_r1504297176 by @martynmjones, I think it might be a bit safer to check if this is the allowed topics defined in NotificationService.php. Perhaps we could add an array `ALLOWED_TOPICS` in NotificaitonService.php?

ianlin (commented):
Thanks for the refactoring. I can see the jobs being scheduled with item_id and the topic for products and coupons, and jobs for shipping and settings don't have item_id and the topic. I left a comment below about checking the topic name a bit more strictly.

puntope:
Hi @ianlin many thanks for the review! 

The check for the topic is handled on the WPCOM side. But I added an extra check in the extension. This is ready for a new round. Thanks

ianlin (approved):
Thanks for the extra change, LGTM. 👍

PR #2277, opened 2024-02-23
Refactor product issues

jorgemd24:

This PR fetches the product issues using the same background job used to fetch the product statuses. This allows us to reuse the WC products that were fetched while updating the product statuses.

The logic remains mostly unchanged from before. The main difference is that previously, Google IDs were stored in the database and used to fetch the issues. With the new sync mechanism, since we don't have the Google IDs until we fetch all product statuses, we now use the data from the product statuses to obtain the Google IDs and fetch the issues.

1. Set your PHP memory limit low so you can catch any heavy process. For example, setting this up to 128MB.
2. Call `GET wp-json/wc/gla/mc/product-statistics/refresh`
3. Go to WooCommerce -> Status -> Action Scheduler -> Search for `gla/jobs/update_merchant_product_statuses/process_item` see how the jobs get completed successfully.
4. See how the table `wp_gla_merchant_issues` gets populated.
5. Once all jobs are completed call `GET wp-json/wc/gla/mc/product-statistics` and see that you get the right stats.

### Additional details:
 - Account and pre-sync issues are not migrated yet.
 - After all the issues logic is refactored, I will remove any legacy code.
 - I wanted to address the scenario where the job fails and some issues are inserted. I used the "created at" field, but it appears that the field will be identical for all batches. Therefore, I'll need to see alternative solutions in a follow-up PR. Any suggestions are welcome too.
 

### Changelog entry

>

### DIFF
--- a/src/DB/Table/MerchantIssueTable.php
+++ b/src/DB/Table/MerchantIssueTable.php
@@ -55,10 +55,9 @@ SQL;
 	 * Delete stale issue records.
 	 *
 	 * @param DateTime $created_before Delete all records created before this.
-	 * @param string   $compare        Comparison operator to use. Default is '<'.
 	 */
-	public function delete_stale( DateTime $created_before, string $compare = '<' ) {
-		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` {$compare} '%s'";
+	public function delete_stale( DateTime $created_before ) {
+		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` < '%s'";
 		$this->wpdb->query( $this->wpdb->prepare( $query, $created_before->format( 'Y-m-d H:i:s' ) ) ); // phpcs:ignore WordPress.DB.PreparedSQL
 	}
 
--- a/src/DB/Table/MerchantIssueTable.php
+++ b/src/DB/Table/MerchantIssueTable.php
@@ -56,10 +56,11 @@ SQL;
 	 *
 	 * @param DateTime $created_before Delete all records created before this.
 	 */
-	public function delete_stale( DateTime $created_before ) {
-		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` < '%s'";
-		$this->wpdb->query( $this->wpdb->prepare( $query, $created_before->format( 'Y-m-d H:i:s' ) ) ); // phpcs:ignore WordPress.DB.PreparedSQL
-	}
+	public function delete_stale( DateTime $created_before, string $comparison = '<') {
+		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` %s '%s'";
+		error_log( '========>'.$query );
+		$this->wpdb->query( $this->wpdb->prepare( $query, $comparison, $created_before->format( 'Y-m-d H:i:s' ) ) ); // phpcs:ignore WordPress.DB.PreparedSQL
+	}	
 
 	/**
 	 * Get the columns for the table.
--- a/src/DB/Table/MerchantIssueTable.php
+++ b/src/DB/Table/MerchantIssueTable.php
@@ -56,7 +56,7 @@ SQL;
 	 *
 	 * @param DateTime $created_before Delete all records created before this.
 	 */
-	public function delete_stale( DateTime $created_before ) {
+	public function delete_stale( DateTime $created_before ): void {
 		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` < '%s'";
 		$this->wpdb->query( $this->wpdb->prepare( $query, $created_before->format( 'Y-m-d H:i:s' ) ) ); // phpcs:ignore WordPress.DB.PreparedSQL
 	}
@@ -67,7 +67,7 @@ SQL;
 	 * @param array  $products_ids Array of product IDs to delete issues for.
 	 * @param string $source       The source of the issues. Default is 'mc'.
 	 */
-	public function delete_specific_product_issues( array $products_ids, string $source = 'mc' ) {
+	public function delete_specific_product_issues( array $products_ids, string $source = 'mc' ): void {
 		$placeholder = '(' . implode( ',', array_fill( 0, count( $products_ids ), '%d' ) ) . ')';
 		$this->wpdb->query( $this->wpdb->prepare( "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `product_id` IN {$placeholder} AND `source` = %s", array_merge( $products_ids, [ $source ] ) ) ); // phpcs:ignore WordPress.DB.PreparedSQL
 	}
--- a/src/DB/Table/MerchantIssueTable.php
+++ b/src/DB/Table/MerchantIssueTable.php
@@ -55,6 +55,7 @@ SQL;
 	 * Delete stale issue records.
 	 *
 	 * @param DateTime $created_before Delete all records created before this.
+	 * @param string   $compare         Comparison operator to use. Default is '<'.
 	 */
 	public function delete_stale( DateTime $created_before, string $compare = '<') {
 		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` {$compare} '%s'";
--- a/src/DB/Table/MerchantIssueTable.php
+++ b/src/DB/Table/MerchantIssueTable.php
@@ -57,10 +57,10 @@ SQL;
 	 * @param DateTime $created_before Delete all records created before this.
 	 * @param string   $compare         Comparison operator to use. Default is '<'.
 	 */
-	public function delete_stale( DateTime $created_before, string $compare = '<') {
+	public function delete_stale( DateTime $created_before, string $compare = '<' ) {
 		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` {$compare} '%s'";
 		$this->wpdb->query( $this->wpdb->prepare( $query, $created_before->format( 'Y-m-d H:i:s' ) ) ); // phpcs:ignore WordPress.DB.PreparedSQL
-	}	
+	}
 
 	/**
 	 * Get the columns for the table.
--- a/src/DB/Table/MerchantIssueTable.php
+++ b/src/DB/Table/MerchantIssueTable.php
@@ -56,10 +56,9 @@ SQL;
 	 *
 	 * @param DateTime $created_before Delete all records created before this.
 	 */
-	public function delete_stale( DateTime $created_before, string $comparison = '<') {
-		$query = "DELETE FROM `{$this->get_sql_safe_name()}` WHERE `created_at` %s '%s'";
-		error_log( '========>'.$query );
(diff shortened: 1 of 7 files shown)


src/MerchantCenter/MerchantStatuses.php
```diff
@@ -235,14 +258,6 @@ public function maybe_refresh_status_data( bool $force_refresh = false ): void {
 		// Update account-level issues.
 		$this->refresh_account_issues();
 
-		// Update MC product issues and tabulate statistics in batches.
```
puntope, line 238: Seems like we are removing the product issues logic. Should we update also the description? Seems like we don't update stale status-related data regarding product issues anymore.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -190,10 +191,32 @@ public function clear_cache(): void {
...
 	}
 
+	/**
+	 * Delete the stale issues from the database.
+	 *
+	 * @param string $compare The comparison operator to use for the created_at field.
+	 *
+	 * @since x.x.x
+	 */
+	protected function delete_stale_issues( string $compare = '<' ) {
```
puntope: 💅 Maybe we might add `: void `to be consistent with the rest of the class signatures.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -27,6 +27,7 @@
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use DateTime;
 use Exception;
+use WC_Product;
```
puntope, line 30: 💅  Unused `use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent\ProductStatus as GoogleProductStatus;`

src/DB/Table/MerchantIssueTable.php
```diff
@@ -55,9 +55,10 @@ public static function get_raw_name(): string {
 	 * Delete stale issue records.
 	 *
 	 * @param DateTime $created_before Delete all records created before this.
+	 * @param string   $compare         Comparison operator to use. Default is '<'.
```
puntope: 💅  Maybe we can delete the extra tabulation between $compare and the description.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -354,67 +369,71 @@ protected function fetch_issues( string $type = null, int $per_page = 0, int $pa
...
 
 	/**
-	 * Return only the valid statuses from a provided array of Google IDs. Invalid statuses:
-	 * - Aren't synced by this extension (invalid ID format), or
-	 * - Map to products no longer in WooCommerce (deleted or uploaded by a previous connection).
-	 * Also populates the product_data_lookup used in refresh_product_issues()
+	 * Get MC product issues from a list of Product View statuses.
 	 *
-	 * @param string[] $google_ids
+	 * @param [] $statuses The list of Product View statuses.
```
puntope: Shouldn't be `array[]` instead of `[]`?

puntope: Additionally. PHPDoc comment doesn't contain all the necessary @throws tags derivated from `$this->container->get`

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -354,67 +369,71 @@ protected function fetch_issues( string $type = null, int $per_page = 0, int $pa
...
+			if ( ChannelVisibility::DONT_SYNC_AND_SHOW === $wc_product->get_meta( $visibility_meta_key ) ) {
+				continue;
+			}
 
-			$this->product_data_lookup[ $wc_product_id ] = [
-				'name'       => get_the_title( $wc_product ),
-				'visibility' => get_post_meta( $wc_product_id, $visibility_meta_key ),
-				'parent_id'  => $wc_product->post_parent,
+			$product_issue_template = [
+				'product'              => html_entity_decode( $wc_product->get_name() ),
```
puntope, line 410: 💅  Maybe `product_name` may be more accurate. (However was not caused in this PR)

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -488,52 +507,11 @@ protected function refresh_custom_merchant_issues() {
 	}
 
 	/**
-	 * Retrieve all product-level issues and store them in the database.
+	 * Refresh product issues in the merchant issues table.
 	 *
-	 * @param GoogleProductStatus[] $validated_mc_statuses Product statuses of validated products.
+	 * @param [] $product_issues Array of product issues.
```
puntope: 💅  This might be `array` instead of `[]` also some throw tags missing.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -618,31 +596,25 @@ protected function refresh_presync_product_issues(): void {
...
 	 * Process product status statistics.
 	 *
-	 * @param array[] $statuses statuses.
+	 * @param array[] $product_view_statuses Product View statuses.
 	 * @see MerchantReport::get_product_view_report
 	 */
-	public function process_product_statuses( array $statuses ): void {
-		$product_repository = $this->container->get( ProductRepository::class );
-		$products           = $product_repository->find_by_ids_as_associative_array( array_column( $statuses, 'product_id' ) );
+	public function process_product_statuses( array $product_view_statuses ): void {
```
puntope, line 609: 💅  Missing throw tags

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -668,13 +640,16 @@ public function process_product_statuses( array $statuses ): void {
...
+		$parent_products           = $product_repository->find_by_ids_as_associative_array( $parent_keys );
+		$this->product_data_lookup = $this->product_data_lookup + $parent_products;
 
 		// Update each product's mc_status and then update the global statistics.
-		$this->update_products_meta_with_mc_status( $products );
-		$this->update_intermediate_product_statistics( $products );
+		$this->update_products_meta_with_mc_status();
+		$this->update_intermediate_product_statistics();
+
+		$product_issues = $this->get_product_issues( $product_view_statuses );
```
puntope, line 658: 💅  WDYT about Saving $product_issues in the class we don't need to use params here?

puntope (commented):
Thanks @jorgemd24  for working on this. I tested following the indications inn the description and worked for me.

I left some comments regarding the code.

> I wanted to address the scenario where the job fails and some issues are inserted. I used the "created at" field, but it appears that the field will be identical for all batches. Therefore, I'll need to see alternative solutions in a follow-up PR. Any suggestions are welcome too.

Could you elaborate this? I'm not sure if I understand the issue.

src/MerchantCenter/MerchantStatuses.php
jorgemd24: Yep, it would be more accurate but I'd prefer to keep it since it's tied to the column name in `wp_gla_merchant_issues`.

See: https://github.com/woocommerce/google-listings-and-ads/blob/7f3e1f57887658cadcc38f6795a36d1fb7fb1ccb/src/DB/Table/MerchantIssueTable.php#L33

src/MerchantCenter/MerchantStatuses.php
jorgemd24: I'd prefer to keep the function independent of the instance state. This way, we can insert specific product issues without relying on the state.

jorgemd24:
Thanks @puntope for the review! I addressed all your comments and replied to some of your suggestions. Regards:

> Could you elaborate this? I'm not sure if I understand the issue.

I was thinking a scenario where the job fails during inserting the issues. For instance, if the job inserts some issues but others fail, such as a query taking too long, the job itself fails, and Action Scheduler will retry it. This means it would reinsert the issues that were successfully added in the initial try. The `update_or_insert` will always insert the row because the key in the table is the `ID` and not the product_id and the issue so I decided to delete the issues for the products we were processing for that specific job, so if the job retries, it starts fresh without any stale data. See here: https://github.com/woocommerce/google-listings-and-ads/pull/2277/commits/c75eb4db0a2c9d3f4179696c25bf01ec1f354f5e

Could you please take another look at it? Thanks!

puntope:
> I decided to delete the issues for the products we were processing for that specific job, so if the job retries, it starts fresh without any stale data. 

Yes, I guess that might work. Other alternative option may be a duplicate cleanup after the AS Job batch finishes.

src/DB/Table/MerchantIssueTable.php
```diff
@@ -61,6 +61,17 @@ public function delete_stale( DateTime $created_before ) {
 		$this->wpdb->query( $this->wpdb->prepare( $query, $created_before->format( 'Y-m-d H:i:s' ) ) ); // phpcs:ignore WordPress.DB.PreparedSQL
 	}
 
+	/**
+	 * Delete product issues for specific products and source.
+	 *
+	 * @param array  $products_ids Array of product IDs to delete issues for.
+	 * @param string $source       The source of the issues. Default is 'mc'.
+	 */
+	public function delete_specific_product_issues( array $products_ids, string $source = 'mc' ) {
```
puntope: 💅  We can add `:void `as a return type in the signature to be consistent with the other functions. `delete_stale`also missed it, but was not introduced in this PR tho

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -190,10 +191,30 @@ public function clear_cache(): void {
...
+	protected function delete_product_statuses_count_intermediate_data(): void {
 		$this->options->delete( OptionsInterface::PRODUCT_STATUSES_COUNT_INTERMEDIATE_DATA );
 	}
 
+	/**
+	 * Delete the stale issues from the database.
+	 *
+	 * @since x.x.x
+	 */
+	protected function delete_stale_issues() {
```
puntope: 💅  Missed :void

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -190,10 +191,30 @@ public function clear_cache(): void {
...
+		$this->container->get( MerchantIssueTable::class )->delete_stale( $this->cache_created_time );
+	}
+
+	/**
+	 * Clear the product statuses cache and delete stale issues.
+	 *
+	 * @since x.x.x
+	 */
+	public function clear_product_statuses_cache_and_issues(): void {
```
puntope, line 212: 💅  Maybe we can add also some tests for this new logic.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -354,67 +367,74 @@ protected function fetch_issues( string $type = null, int $per_page = 0, int $pa
...
+			if ( ! $wc_product ) {
 				do_action(
 					'woocommerce_gla_debug_message',
 					sprintf( 'Merchant Center product %s not found in this WooCommerce store.', $mc_product_id ),
 					__METHOD__ . ' in remove_invalid_statuses()',
 				);
 				continue;
 			}
+			// Unsynced issues shouldn't be shown.
+			if ( ChannelVisibility::DONT_SYNC_AND_SHOW === $wc_product->get_meta( $visibility_meta_key ) ) {
```
puntope, line 405: 💅  Seems that we are not covering this case in tests.

puntope (approved):
Thanks @jorgemd24 for all the changes.

I approve this as there are no blockers anymore and it works as defined in the PR description.

I left tho some comments regarding code and test coverage.

PR #2278, opened 2024-02-23
API PULL - Add Coupons Notification

puntope:

This PR adds the logic for adding Cupon Notifications in the API PULL Project.

⚠️  It follows the same structure as Product Notifications. Hence a lot of code can be refactored reducing duplications. This will be done in further PRs.

`NotificationsService.php`
```php
     return true;
}
```

1. With the Onboarding finished, crate a coupon with the status "DRAFT" and Channel Visibility "Dont show coupon on Google"
2. Check the Action Scheduler and see there is NO action `gla/jobs/notifications/coupons/process_item`
3. Change the status to Publish but still "Dont show coupon on Google". 
4. Check the Action Scheduler and see there is NO action `gla/jobs/notifications/coupons/process_item`
5. Change the channel visibility to  "Show coupon on Google"
6. Check the Action Scheduler and see that there is an action `gla/jobs/notifications/coupons/process_item` with `coupon.create` topic.
7. Run the action 
8. Update the coupon description
9. Check the Action Scheduler and see that there is an action `gla/jobs/notifications/coupons/process_item` with `coupon.update` topic.
10. Change the channel visibility again to  "Don't Show coupon on Google"
11.  Check the Action Scheduler and see that there is an action `gla/jobs/notifications/coupons/process_item` with `coupon.delete` topic.
12. Update the coupon description
13. Check the Action Scheduler and see there is NO action `gla/jobs/notifications/coupons/process_item`
14. Change the channel visibility to  "Show coupon on Google"
15. Check the Action Scheduler and see that there is an action `gla/jobs/notifications/coupons/process_item` with `coupon.create` topic.
16. Test other scenarios to verify that the notifications are being sent correctly. 

### Changelog entry

>

### DIFF
--- a/src/Coupon/SyncerHooks.php
+++ b/src/Coupon/SyncerHooks.php
@@ -355,7 +355,7 @@ class SyncerHooks implements Service, Registerable {
 	}
 
 	/**
-	 * Schedules notifications for an updated product
+	 * Schedules notifications for an updated coupon
 	 *
 	 * @param WC_Coupon $coupon
 	 */
--- a/src/Coupon/SyncerHooks.php
+++ b/src/Coupon/SyncerHooks.php
@@ -3,14 +3,17 @@ declare(strict_types = 1);
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Coupon;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Google\DeleteCouponEntry;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\NotificationsService;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Registerable;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\JobRepository;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\DeleteCoupon;
+use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications\CouponNotificationJob;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\UpdateCoupon;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
 use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
+use Automattic\WooCommerce\GoogleListingsAndAds\Value\NotificationStatus;
 use WC_Coupon;
 defined( 'ABSPATH' ) || exit();
 
@@ -64,10 +67,21 @@ class SyncerHooks implements Service, Registerable {
 	protected $delete_coupon_job;
 
 	/**
+	 * @var CouponNotificationJob
+	 */
+	protected $coupon_notification_job;
+
+	/**
 	 *
 	 * @var MerchantCenterService
 	 */
 	protected $merchant_center;
+
+	/**
+	 * @var NotificationsService
+	 */
+	protected $notifications_service;
+
 
 	/**
 	 *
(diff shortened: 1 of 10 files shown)


src/Coupon/SyncerHooks.php
```diff
@@ -322,4 +353,22 @@ protected function set_already_scheduled_to_update( int $coupon_id ): void {
 	protected function set_already_scheduled_to_delete( int $coupon_id ): void {
 		$this->set_already_scheduled( $coupon_id, self::SCHEDULE_TYPE_DELETE );
 	}
+
+	/**
+	 * Schedules notifications for an updated product
```
martynmjones: I'm assuming this was copied over from `Product/SyncerHooks.php` but we're referencing a product here:

```suggestion
	 * Schedules notifications for an updated coupon
```

src/Jobs/Notifications/CouponNotificationJob.php
```diff
@@ -0,0 +1,184 @@
...
+			do_action(
+				'woocommerce_gla_error',
+				'Error sending Coupon Notification. Topic and Coupon ID are mandatory',
+				__METHOD__
+			);
+			return;
+		}
+
+		$item  = $args[0];
+		$topic = $args[1];
```
martynmjones: Had you considered using keys for these arguments instead of relying on them being in the correct order?

src/Jobs/Notifications/CouponNotificationJob.php
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	/**
+	 * Get the Notification Status after the notification happens
+	 *
+	 * @param string $topic
+	 * @return string
+	 */
+	protected function get_after_notification_status( string $topic ): string {
+		if ( str_contains( $topic, '.create' ) ) {
```
martynmjones, line 122: Could you clarify why we're checking if `$topic` contains `.create` or `.delete` instead of comparing the exact value to the constants?

src/Jobs/Notifications/CouponNotificationJob.php
```diff
@@ -0,0 +1,184 @@
...
+			return;
+		}
+
+		$item  = $args[0];
+		$topic = $args[1];
+
+		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item ) ) {
+			$this->set_status( $item, $this->get_after_notification_status( $topic ) );
+			$this->maybe_mark_as_unsynced( $topic, $item );
+		}
```
martynmjones, line 87: If we're not able to process the item it would be nice to throw an Exception with some basic information so that the scheduled action doesn't fail silently.

martynmjones (commented):
Hey @puntope, thanks for your work on the Coupon sync!

I'm not able to test with the sandbox so used your suggestion to edit `NotificationsService.php`.

Following your test instructions I got the expected results with actions being scheduled. I've left a few comments on the code (_I'm unsure which bits were copied over that you're intending on refactoring later_).

I did encounter one scenario that I think may be problematic. While we're scheduling immediate actions, we can't be sure they will have completed by the next time a coupon is updated.

For a coupon that has already been synced, if we update the channel visibility to `Do not show...` then a `coupon.delete` action is scheduled.

If the channel visibility is set back to `Show coupon on Google` before the delete action completes then the next action will be a `coupon.update` instead of `coupon.create` as the status hasn't yet been updated.

Screenshot 2024-02-27 at 14 01 35

When the update action is then processed, `should_trigger_update_notification` will [return false](https://github.com/woocommerce/google-listings-and-ads/blob/a73ad24eecf63bf5fef9aa303967948f1658236f/src/Coupon/CouponHelper.php#L402) meaning a coupon that should be synced might not be.

I'm unsure of how the API handles scenarios like that but does this risk causing issues where the status is out of sync and some coupons that a merchant expects to be shown on Google won't be?

src/Jobs/Notifications/CouponNotificationJob.php
puntope: This is because it is planned to be refactored in a parent class in further PRs to reuse this function.

puntope:
> I did encounter one scenario that I think may be problematic. While we're scheduling immediate actions, we can't be sure they will have completed by the next time a coupon is updated.

I was not able to reproduce this. Since when the Item updates we set back the status to PENDING_UPDATE, the delete action wont trigger Notify because the `can_process` and  `should_trigger_delete_notification` will return false. But the update yes. Because  `can_process` and  `should_trigger_update_notification` will return true.


https://github.com/woocommerce/google-listings-and-ads/assets/5908855/5bb12848-fe9a-4132-9750-2902d83124d1

Maybe I did miss some steps to reproduce the issue?

src/Jobs/Notifications/CouponNotificationJob.php
puntope: Do you mean to throw exceptions each time AS cannot process?  Wouldn't this cause excessive filling of the logs? For example when we have the case mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/2278#issuecomment-1968846588

src/Jobs/Notifications/CouponNotificationJob.php
puntope: Good suggestion @martynmjones that will make it clearer and easier to maintain. I will do as well in the refactoring for Products in further PRs if that's ok.

See https://github.com/woocommerce/google-listings-and-ads/pull/2278/commits/3819cca3173caa587ea638fa2f22a71622b59f63

puntope:
Thanks for the review @martynmjones I answered your question about the issue and this is ready for a new round.

src/Jobs/Notifications/CouponNotificationJob.php
martynmjones: My thought was that it would make things more traceable if in the Action Scheduler logs we had a bit of additional information instead of just seeing that the action ran but if you think it would be excessive it's definitely not an issue to leave it as is.

martynmjones (approved):
Hey @puntope, many thanks for the changes and clarifications!

> I was not able to reproduce this. Since when the Item updates we set back the status to PENDING_UPDATE, the delete action wont trigger Notify because the can_process and should_trigger_delete_notification will return false. But the update yes. Because can_process and should_trigger_update_notification will return true.

Apologies, I went back over it and you're right this won't cause problems so LGTM ✅

PR #2281, opened 2024-02-26
API PULL - Unsync product when delete Notification happens

puntope:

This PR will mark the product as unsynced in MC when a delete notification happens.

ℹ️  This was discussed in p1708720809845579-slack-C0410KV3YLW

`NotificationsService.php`
```php
     return true;
}
```

1. Sync a Product to Google MC
2. Delete that product
3. Go to action Scheduler and see the action `gla/jobs/notifications/products/process_item` with the product and the `product.delete` topic.
4. Run it
5. See the product is unsynced.

### DIFF
--- a/src/Jobs/Notifications/ProductNotificationJob.php
+++ b/src/Jobs/Notifications/ProductNotificationJob.php
@@ -77,6 +77,7 @@ class ProductNotificationJob extends AbstractActionSchedulerJob implements JobIn
 
 		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item ) ) {
 			$this->set_status( $item, $this->get_after_notification_status( $topic ) );
+			$this->maybe_mark_as_unsynced( $topic, $item );
 		}
 	}
 
@@ -119,6 +120,21 @@ class ProductNotificationJob extends AbstractActionSchedulerJob implements JobIn
 		} else {
 			return NotificationStatus::NOTIFICATION_UPDATED;
 		}
+	}
+
+	/**
+	 * If there is a valid Item ID and topic is a deletion topic. Mark the coupon as unsynced.
+	 *
+	 * @param string $topic
+	 * @param int    $item
+	 */
+	protected function maybe_mark_as_unsynced( string $topic, int $item ): void {
+		if ( ! str_contains( $topic, '.delete' ) ) {
+			return;
+		}
+
+		$product = $this->product_helper->get_wc_product( $item );
+		$this->product_helper->mark_as_unsynced( $product );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


ianlin (commented):
Thanks for the PR. I tried to test it but I was unsure how to verify "see the product is unsynced" from the test instruction. Here's my test step:

1. I have a product that was previously synced to MC, and I can see its meta from the postmeta table:

```
_wc_gla_visibility: sync-and-show
_wc_gla_sync_status: synced
_wc_gla_synced_at: 1708508099
_wc_gla_google_ids: a:1:{s:2:"GB";s:20:"online:en:GB:gla_873";}
_wc_gla_mc_status: disapproved
```

2. I have modified `NotificationsService.php` so the method `notify` will always return `true`
3. Then I "trashed" the product from the WC product page
4. I can see the job being scheduled and I ran it
5. Then I checked the postmeta again, here's the data after the product was trashed:

```
_wc_gla_visibility: sync-and-show
_wc_gla_mc_status: disapproved
_wc_gla_notification_status: delete
```

So what I saw was `_wc_gla_sync_status`, `_wc_gla_synced_at`, and `_wc_gla_google_ids` were removed, which is what the method `mark_as_unsynced` in `ProductHelper.php` did.

Is the three columns being deleted considered "product is unsynced"? Or maybe I miss something?

puntope:
> Is the three columns being deleted considered "product is unsynced"? Or maybe I miss something?

Hi @ianlin thanks for the review, yes that is what is expected. We just call the function `mark_as_unsynced` previously created. https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Product/ProductHelper.php#L104

Checking [the function](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Product/ProductHelper.php#L352) `is_product_synced` you can see that it only takes `synced_at` and `google_ids` in consideration

```php
/**
	 * @param WC_Product $product
	 *
	 * @return bool
	 */
	public function is_product_synced( WC_Product $product ): bool {
		$synced_at  = $this->meta_handler->get_synced_at( $product );
		$google_ids = $this->meta_handler->get_google_ids( $product );

		return ! empty( $synced_at ) && ! empty( $google_ids );
	}
```

src/Jobs/Notifications/ProductNotificationJob.php
```diff
@@ -121,6 +122,21 @@ protected function get_after_notification_status( string $topic ): string {
...
 	}
 
+	/**
+	 * If there is a valid Item ID and topic is a deletion topic. Mark the coupon as unsynced.
+	 *
+	 * @param string $topic
+	 * @param int    $item
+	 */
+	protected function maybe_mark_as_unsynced( string $topic, int $item ): void {
+		if ( ! str_contains( $topic, '.delete' ) ) {
```
ianlin, line 132: I was going to ask the same question like https://github.com/woocommerce/google-listings-and-ads/pull/2278#discussion_r1504297176 but I saw you had a reply there. Maybe we could add a TODO comment for it?

ianlin (approved):
Thanks for the reply, it worked as expected like my previous comment. Just left a comment that isn't a blocker.

PR #2286, opened 2024-02-29
Avoid the test-data plugin occasionally missing in the E2E test environment

### LINKED ISSUE #2175: Test data plugin is sometimes missing from test environment and error is displayed reading "The plugin does not have a valid header"
### Describe the bug:
`npm run wp-env:up` will setup an environment including the [test-data.php](https://github.com/woocommerce/google-listings-and-ads/blob/676d64944157e10105cf1e47e6a53bf5dcceb35d/tests/e2e/bin/test-data.php) plugin which is required for the E2E tests to run.

Occasionally, an error will be triggered on the test site reading `The plugin test-data.php has been deactivated due to an error: The plugin does not have a valid header.`. On more than one occasion I've checked the Docker container after that error was triggered and the `test-data.php` file doesn't exist.

Screenshot 2023-12-06 at 12 19 37

### Steps to reproduce:
1. Setup local environment with `npm run wp-env:up`
2. Run e2e tests

I've not yet come up with reliable reproduction steps as the issue appears to be intermittent.

### Expected behavior:
We should be able to rely on the environment being used for e2e testing.

eason9487:

Closes #2175

💡 The issue probably only happens on **MacOS**. Usually I can reproduce it after running the E2E tests 1-3 rounds.

Although I'm not fully confident this PR can fix it, after some time and validation, it seems to be working so far. The solution was mentioned by @mikkamp in https://github.com/woocommerce/google-listings-and-ads/pull/2080#issuecomment-1698679989.

This PR:

- Fix the indentation for the `.wp-env.json` file first as it was mixed with spaces and tabs.
- Make the `test-data.php` plugin have a dedicated directory and change to use [the `plugins` directive](https://github.com/WordPress/gutenberg/tree/%40wordpress/env%409.4.0/packages/env#wp-envjson) to set it up.

💡 Considering the issue seems to only happen on MacOS, it would be more appropriate to have someone who uses the same OS or can reproduce the issue to review and test it.

1. Running `npm run wp-env destroy` should not be necessary but it would be ideal to have a clean start to test it.
2. Run `npm run wp-env:up` to see if the E2E test env can be started without errors.
3. Log in http://localhost:8889/wp-admin/ and go to the **Installed Plugins** page.
4. Check if the **GLA Test Data** plugin is activated.
5. Run `npm run test:e2e` to see if all E2E tests can pass.
6. Refresh the **Installed Plugins** page to see if the **GLA Test Data** plugin is activated.
7. Repeat steps 5-6 several times to verify if the plugin missing issue no longer occurs.

#### 📌 Investigation

Although I didn't find the exact cause, the process that led to the issue is roughly as follows.

1. From the docker-compose.yml generated by `wp-env`, it's shown there are nested mounts under the `/var/www/html` directory.
   ```yml
   version: '3.7'
   services:
     # ... omitted
     tests-wordpress:
       # ... omitted
       volumes:
         - '/Users/eason/.wp-env/fa3f9473ab3c866d9e18213e3637074d/tests-WordPress:/var/www/html'
         - '/Users/eason/.wp-env/fa3f9473ab3c866d9e18213e3637074d/tests-WordPress-PHPUnit/tests/phpunit:/wordpress-phpunit'
         - 'tests-user-home:/home/eason'
         - '/Users/eason/wp-local-docker-sites/wp-test/wordpress/wp-content/plugins/google-listings-and-ads/tests/e2e/config/wp-cli.yml:/var/www/html/wp-cli.yml'
         - '/Users/eason/wp-local-docker-sites/wp-test/wordpress/wp-content/plugins/google-listings-and-ads/tests/e2e/bin/test-data.php:/var/www/html/wp-content/plugins/test-data.php'
         - '/Users/eason/.wp-env/fa3f9473ab3c866d9e18213e3637074d/master:/var/www/html/wp-content/plugins/master'
         - '/Users/eason/wp-local-docker-sites/wp-test/wordpress/wp-content/plugins/google-listings-and-ads:/var/www/html/wp-content/plugins/google-listings-and-ads'
   
   ```
2. In the container, the `test-data.php` is mounted in the beginning.
   ```bash
   root@969c943abe08:/var/www/html# mount | grep www
   grpcfuse on /var/www/html type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   grpcfuse on /var/www/html/wp-cli.yml type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   grpcfuse on /var/www/html/wp-content/plugins/test-data.php type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   grpcfuse on /var/www/html/wp-content/plugins/google-listings-and-ads type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   grpcfuse on /var/www/html/wp-content/plugins/master type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   ```
   In the container, its file size is 3.0K.
   ```bash
   root@969c943abe08:/var/www/html# ls -lh /var/www/html/wp-content/plugins/test-data.php
   -rw-r--r-- 1 eason dialout 3.0K Feb 29 07:02 /var/www/html/wp-content/plugins/test-data.php
   ```
3. Due to the nested mounts, a `test-data.php` file is created back to the host side with **0 file size**.
   
4. After a few runs of E2E test, somehow the `test-data.php` is unmounted.
   ```bash
   root@969c943abe08:/var/www/html# mount | grep www
   grpcfuse on /var/www/html type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   grpcfuse on /var/www/html/wp-content/plugins/google-listings-and-ads type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   grpcfuse on /var/www/html/wp-content/plugins/master type fuse.grpcfuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other,max_read=1048576)
   ```
   In the container, its file size is 0 now. I think the empty content file is synced from the host.
   (See the screenshot in step 3)
   ```bash
   root@969c943abe08:/var/www/html# ls -lh /var/www/html/wp-content/plugins/test-data.php
   -rwxr-xr-x 1 eason dialout 0 Feb 29 07:06 /var/www/html/wp-content/plugins/test-data.php
   ```
5. Now the issue occurs.
   - The Installed Plugins page
      
   - E2E test always failed
      

I couldn't find the same issue report, but I suspect the reason for this is similar to these past issues:

- https://github.com/docker/compose/issues/6975
- https://github.com/docker/for-mac/issues/3479
- https://github.com/docker/for-mac/issues/3976
- https://github.com/docker/for-mac/issues/2777

#### 📌 About the `wp-cli.yml` file

Please note that the issue also happens on the `wp-cli.yml` file. If I understood it right, it's a one-time setup during initialization and won't break when the `wp-cli.yml` file becomes unavailable. Therefore, this PR doesn't adjust it.

https://github.com/woocommerce/google-listings-and-ads/blob/e6c03c151b7b4eff133579dffbd3e58164bd97d6/tests/e2e/bin/test-env-setup.sh#L9-L10

### Changelog entry

> Dev - Avoid the test-data plugin occasionally missing in the E2E test environment


martynmjones (approved):
Hey @eason9487, thanks for working on this fix.

I can't confirm if it definitely resolves the issue but I've not been able to trigger the problem using this branch after running multiple tests so looks good to merge and will be interesting to see if we encounter it again.

PR #2287, opened 2024-02-29
Add support for Google Analytics for WooCommerce version 2.0.0

martynmjones:

⚠️ **Requires:** https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/322

Google Analytics for WooCommerce updates how the Global Site Tag is loaded so this PR updates the integration in GLA to support version `2.0.0` and above while maintaining backwards compatibility.

The main difference is that instead of simply adding a script to the page with a call to `gtag`, we now use `wp_add_inline_script` and the Google Analytics for WooCommerce script handle where applicable.

1. Build Google Analytics for WooCommerce from [#322](https://github.com/woocommerce/woocommerce-google-analytics-integration/pull/322)
2. Complete onboarding if needed
3. Debug a GLA test site via [TagAssistant](https://tagassistant.google.com/)
4. Confirm the GLA tag is found
5. Confirm that the GLA events are sent:
    - `page_view`
    - `view_item`
    - `purchase`
    - `conversion`

### Changelog entry

> Add - Support for Google Analytics for WooCommerce version 2.0.0 and above

### DIFF
--- a/src/Google/GlobalSiteTag.php
+++ b/src/Google/GlobalSiteTag.php
@@ -257,7 +257,7 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 
 			gtag('js', new Date());
 			gtag('set', 'developer_id.<?php echo esc_js( self::DEVELOPER_ID ); ?>', true);
-			<?php echo $this->get_gtag_config( $ads_conversion_id ); ?>
+			<?php echo esc_js( $this->get_gtag_config( $ads_conversion_id ) ); ?>
 		</script>
 
 		<?php
@@ -288,7 +288,7 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 		if ( class_exists( '\WC_Google_Gtag_JS' ) ) {
 			wp_add_inline_script(
 				'woocommerce-google-analytics-integration',
-				$inline_script
+				esc_js( $inline_script )
 			);
 		} else {
 			wp_print_inline_script_tag( $inline_script );
--- a/src/Google/GlobalSiteTag.php
+++ b/src/Google/GlobalSiteTag.php
@@ -257,7 +257,7 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 
 			gtag('js', new Date());
 			gtag('set', 'developer_id.<?php echo esc_js( self::DEVELOPER_ID ); ?>', true);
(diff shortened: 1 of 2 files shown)


src/Google/GlobalSiteTag.php
```diff
@@ -217,16 +217,24 @@ function () use ( $gtag_events ) {
...
-				'woocommerce_gtag_snippet',
-				function ( $gtag_snippet ) use ( $ads_conversion_id ) {
-					return preg_replace(
-						'~(\s)</script>~',
-						"\tgtag('config', '" . $ads_conversion_id . "', { 'groups': 'GLA', 'send_page_view': false });\n$1</script>",
-						$gtag_snippet
-					);
-				}
-			);
+			if ( version_compare( \WC_GOOGLE_ANALYTICS_INTEGRATION_VERSION, '2.0.0', '>=' ) ) {
```
tomalec, line 220: :construction: 
I don't get into this line when I run it with the new GA, as the one above returns `false`.
GLA still expect the old GA settings to be there https://github.com/woocommerce/google-listings-and-ads/blob/3a423a2d5cfca64486c74cf41905d01db1cc4c52/src/Proxies/GoogleGtagJs.php#L24:L33

The latest GA does not set `ga_gtag_enabled`

tomalec: I've made a fix in https://github.com/woocommerce/google-listings-and-ads/compare/update/global-site-tag-for-analytics-version-2...update/global-site-tag-for-analytics-version-2-enabled?expand=1

src/Google/GlobalSiteTag.php
```diff
@@ -249,16 +257,44 @@ function gtag() { dataLayer.push(arguments); }
 
 			gtag('js', new Date());
 			gtag('set', 'developer_id.<?php echo esc_js( self::DEVELOPER_ID ); ?>', true);
-			gtag('config', '<?php echo esc_js( $ads_conversion_id ); ?>', {
-				'groups': 'GLA',
-				'send_page_view': false
-			});
+			<?php echo esc_js( $this->get_gtag_config( $ads_conversion_id ) ); ?>
```
tomalec: :construction: 
Escaping JS, results in syntax errors for escaped quotes
```suggestion
			<?php echo $this->get_gtag_config( $ads_conversion_id ); ?>
```

tomalec (approved):
Reviewed the code, tested locally LGTM, thanks! :)

PR #2291, opened 2024-03-04
API Pull - Add an API to return Google WPCOM App auth URL

ianlin:

Part of #2146. 

This PR adds an API which returns an auth URL for granting access to Google WPCOM App.

The format of the auth URL:

```
https://public-api.wordpress.com/oauth2/authorize?
client_id=CLIENT_ID
&redirect_uri=PARTNER_REDIRECT_URL
&response_type=code
&blog=BLOD_ID
&scope=wc-partner-access
&state=BASE64_ENCODED_STRING
```

The content of the `state` is an URL query string where the value of its parameter `redirect_url` is being URL encoded. E.g.
`nonce=nonce-123&redirect_url=https%3A%2F%2Fmerchant-site.example.com%2Fwp-admin%2Fadmin.php%3Fpage%3Dwc-admin%26path%3D%2Fgoogle%2Fsetup-mc`

and the decoded version is:

`nonce=nonce-123&redirect_url=https://merchant-site.example.com/wp-admin/admin.php?page=wc-admin&path=/google/setup-mc`

where the `redirect_url` will be used to return back to the merchant's site by the partner.

Project: pcTzPl-1SM-p2
WPCOM App OAuth: https://developer.wordpress.com/docs/oauth2/

1. Request to `GET /wc/gla/rest-api/authorize`.
2. Check the returned auth URL has the same format from the description above.
3. Open this URL in the browser, you should see the permission granting page.
4. Make sure you can actually grant the permission and it will be redirected to woo.com.
5. After redirected to woo.com, copy the value of the `state` URL query parameter, which is a base64 encoded string.
6. Go to https://www.base64decode.org/ and decode base64 string.
7. The decoded value should look like
    ```
    nonce=nonce-123&redirect_url=https%3A%2F%2Fmerchant-site.example.com%2Fwp-admin%2Fadmin.php%3Fpage%3Dwc-admin%26path%3D%2Fgoogle%2Fsetup-mc
    ```

### Additional details:

Note that this PR hardcoded three things in the code at the moment as we will be calling a new Google API to get those data when the API is ready:

**Client ID**

WPCOM app Client ID, currently I'm using the WPCOM app I created, it should be Google's client ID in the future.

**Redirect URI**

An URL that Google will set in their WPCOM app settings. I'm currently using https://woo.com set in my WPCOM app.

**nonce**

A validation nonce that would return by the new Google API as well.

### Changelog entry

> Add - API to return Google WPCOM auth URL


src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -0,0 +1,181 @@
...
+	 * Mapping between the client page name and its path.
+	 * The first value is also used as a default,
+	 * and changing the order of keys/values may affect things below.
+	 *
+	 * @var string[]
+	 */
+	private const NEXT_PATH_MAPPING = [
+		'setup-mc'  => '/google/setup-mc',
+		'reconnect' => '/google/settings',
+	];
```
ianlin: Adding this because in GL&A plugin, both onboarding flow and the settings page will call this API so we need to know where we are redirecting the merchant back.

src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -0,0 +1,181 @@
...
+							'blog'          => $blog_id,
+							'client_id'     => $client_id,
+							'redirect_uri'  => $redirect_uri,
+							'response_type' => $response_type,
+							'scope'         => $scope,
+							'state'         => $state,
+						],
+						$auth_url
+					)
+				);
```
mikkamp: Not sure if this PR is fully ready for a review yet, as a lot of it seems preliminary. However it looks like this controller is getting a lot of responsibility to build an auth URL. Can we move that to a service class which can just have the function of building an auth URL (at least for now, until we need more there). I think it will make it easier for unit testing and allow setting contants within the class without them interfering with the controller.

The path mapping could still remain in the controller if we want to match the Google connect controller. But the controller should not care much else about how that URL is built. Especially if we have it send an external request to fetch some of the URL parts.

I'm not sure what folder services like this should end up. But maybe since it's related to the WordPress API. We could consider adding it in `src/API/WP`. The `NotificationService` seems to have been added to `src/Google` but maybe we'd want to consider having them grouped together, since they both do similar things using the blog_id and handling the URL to `public-api.wordpress.com`.

ianlin: Thanks for the suggestion @mikkamp.

> Not sure if this PR is fully ready for a review yet, as a lot of it seems preliminary

Since this PR is going to be merged into a feature branch rather than `develop` so I was thinking using the hardcoded data first for testing and revisit it when we get the information about the new Google API.

> Can we move that to a service class which can just have the function of building an auth URL

Completely agreed, why didn't I do it... 😓 Working on it.

> We could consider adding it in src/API/WP. The NotificationService seems to have been added to src/Google but maybe we'd want to consider having them grouped together, since they both do similar things using the blog_id and handling the URL to public-api.wordpress.com.

Great idea, thanks.

ianlin:
Created an `OAuthService` and added the corresponding tests in d4c0099, b3ff1e6, and ff1a516 regarding the comment https://github.com/woocommerce/google-listings-and-ads/pull/2291#discussion_r1512788358.

Moved `NotificationsService` under the folder `API/WP` in 7950c95. cc @puntope since you're the author of `NotificationsService`.

src/API/WP/OAuthService.php
```diff
@@ -0,0 +1,96 @@
...
+	 * @param string $path An URL parameter for the path within GL&A page, which will be added in the merchant redirect URL.
+	 *
+	 * @return string Auth URL.
+	 */
+	public function get_auth_url( string $path ): string {
+		$google_data = $this->get_data_from_google();
+
+		$merchant_redirect_url = urlencode_deep( admin_url( "admin.php?page=wc-admin&path={$path}" ) );
+
+		$state = base64_encode(
```
mikkamp: Just a quick comment, we should be making sure this is URL safe, maybe some helper functions like:
https://base64.guru/developers/php/examples/base64url
https://www.php.net/manual/en/function.base64-encode.php#103849

src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -0,0 +1,132 @@
...
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_authorize_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+					'args'                => $this->get_auth_params(),
+				],
+			]
+		);
+	}
+
```
jorgemd24: Shouldn't we link the schema with the endpoint? `'schema' => $this->get_api_response_schema_callback()`

src/API/WP/OAuthService.php
```diff
@@ -0,0 +1,99 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\WP;
+
+use Automattic\Jetpack\Connection\Client;
```
jorgemd24: Unused dependency.

src/API/WP/OAuthService.php
```diff
@@ -0,0 +1,99 @@
...
+	 * state=BASE64_ENCODED_STRING
+	 *
+	 * Content of state is an URL query string where the value of its parameter "redirect_url" is being URL encoded.
+	 * E.g.
+	 * nonce=nonce-123&redirect_url=https%3A%2F%2Fmerchant-site.example.com%2Fwp-admin%2Fadmin.php%3Fpage%3Dwc-admin%26path%3D%2Fgoogle%2Fsetup-mc
+	 *
+	 * where its URL decoded version is:
+	 * nonce=nonce-123&redirect_url=https://merchant-site.example.com/wp-admin/admin.php?page=wc-admin&path=/google/setup-mc
+	 *
+	 * @param string $path An URL parameter for the path within GL&A page, which will be added in the merchant redirect URL.
```
jorgemd24: I think we need to update the docs to tell that the state will be base64 url encoded.

src/API/WP/OAuthService.php
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Calls an API by Google to get required information in order to form an auth URL.
+	 *
+	 * TODO: Call an actual API by Google.
+	 * We'd probably need use WCS to communicate with the new API.
+	 *
+	 * @return array An associative array contains required information from Google.
+	 */
+	protected function get_data_from_google(): array {
```
jorgemd24, line 98: We could explain the response properties, for example for the nonce we could say that it is fetched from Google and it will be used to verify the call.

src/API/WP/OAuthService.php
```diff
@@ -0,0 +1,99 @@
...
+		return [
+			'client_id'    => '91299',
+			'redirect_uri' => 'https://woo.com',
```
jorgemd24, line 101: Also worth noting, is that if the redirect_url will be fetched from the Google side, they need to be sure that it matches with the WPCOM Client settings. The redirect URLs only work if they are added in WPCOM.

src/HelperTraits/Utilities.php
```diff
@@ -78,4 +78,46 @@ protected function gla_setup_for( $seconds ): bool {
...
+	 * @param string $data The string that will be base64 URL encoded.
+	 *
+	 * @return boolean|string
+	 */
+	protected function base64url_encode( $data ): string {
+		// First of all you should encode $data to Base64 string
+		$b64 = base64_encode( $data );
+
+		// Make sure you get a valid result, otherwise, return FALSE, as the base64_encode() function do
+		if ( $b64 === false ) {
```
jorgemd24: We say that `base64_encode` can return a boolean `false` but reading the doc it seems that always will be a string: https://www.php.net/manual/en/function.base64-encode.php

tests/Unit/API/Site/Controllers/RestAPI/AuthControllerTest.php
```diff
@@ -0,0 +1,79 @@
...
+		$expected_auth_url .= '&scope=wc-partner-access';
+		$expected_auth_url .= '&state=base64_encoded_string';
+		$expected_auth_url  = $expected_auth_url;
+
+		$this->oauth_service->expects( $this->once() )
+			->method( 'get_auth_url' )
+			->willReturn( $expected_auth_url );
+
+		$response = $this->do_request( self::ROUTE_AUTHORIZE, 'GET' );
+		$data     = $response->get_data();
```
jorgemd24: Unused variable.

src/API/WP/OAuthService.php
```diff
@@ -0,0 +1,99 @@
...
+	 * @param string $path An URL parameter for the path within GL&A page, which will be added in the merchant redirect URL.
+	 *
+	 * @return string Auth URL.
+	 */
+	public function get_auth_url( string $path ): string {
+		$google_data = $this->get_data_from_google();
+
+		$merchant_redirect_url = urlencode_deep( admin_url( "admin.php?page=wc-admin&path={$path}" ) );
+
+		$state = $this->base64url_encode(
```
jorgemd24, line 61: What happens if `base64url_encode` returns `false`?

jorgemd24 (commented):
Hi @ianlin! Thanks for working on this! It is working well and I am getting the right URL. I just left some comments about the code and some questions.

src/API/Site/Controllers/RestAPI/AuthController.php
ianlin: Updated in 94886fc.

src/API/WP/OAuthService.php
ianlin: Updated in efb5e5f.

src/API/WP/OAuthService.php
ianlin: Updated in 70442c0.

src/API/WP/OAuthService.php
ianlin: Updated in 9c70657.

src/HelperTraits/Utilities.php
ianlin: You're right, it's `base64_decode` function would return false, not `base64_encode`.

Updated in 614df20.

tests/Unit/API/Site/Controllers/RestAPI/AuthControllerTest.php
ianlin: Updated in 2b6f251.

src/API/WP/OAuthService.php
ianlin: This is not the case because `base64url_encode` won't return `false` anymore. Updated in 614df20.

src/API/WP/OAuthService.php
ianlin: Yeah, I will make sure Google is aware of this.

ianlin:
Thanks @jorgemd24 for the review and nice suggestions. It's ready for the second round.

src/HelperTraits/Utilities.php
```diff
@@ -78,4 +78,41 @@ protected function gla_setup_for( $seconds ): bool {
...
+	 * Encode data to Base64URL
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $data The string that will be base64 URL encoded.
+	 *
+	 * @return string
+	 */
+	protected function base64url_encode( $data ): string {
+		// First of all you should encode $data to Base64 string
```
jorgemd24: I think this comment could be removed, since the `base64_encode` function is self-explanatory.

src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -0,0 +1,133 @@
...
+	/**
+	 * Get the callback function for the authorization request.
+	 *
+	 * @return callable
+	 */
+	protected function get_authorize_callback(): callable {
+		return function ( Request $request ) {
+			try {
+				$next     = $request->get_param( 'next_page_name' );
+				$path     = self::NEXT_PATH_MAPPING[ $next ];
```
jorgemd24, line 77: WDYT if we move `self::NEXT_PATH_MAPPING` to the OAuthService? This way, the controller only needs to pass the parameter, and the OAuthService handles the logic to determine the correct path.

src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -0,0 +1,133 @@
...
+	/**
+	 * Mapping between the client page name and its path.
+	 * The first value is also used as a default,
+	 * and changing the order of keys/values may affect things below.
+	 *
+	 * @var string[]
+	 */
+	private const NEXT_PATH_MAPPING = [
+		'setup-mc'  => '/google/setup-mc',
+		'reconnect' => '/google/settings',
```
jorgemd24: What do you think if we change `reconnect` with `settings`? Mainly because the query param is `next_page_name` so I think it makes sense to call it 'settings'.

jorgemd24 (approved):
Thanks @ianlin for the adjustments. I left some comments about the code and a suggestion of moving `self::NEXT_PATH_MAPPING to the OAuthService` that I would like to hear your opinion! 

I am approving the PR as it works as expected 👍

src/API/Site/Controllers/RestAPI/AuthController.php
ianlin: I put `self::NEXT_PATH_MAPPING` in the controller to match other controllers like [API/Site/Controllers/Google/AccountController](https://github.com/woocommerce/google-listings-and-ads/blob/292cf662bfc6bf976ed926785bdd86cbca175631/src/API/Site/Controllers/Google/AccountController.php#L104) and [API/Site/Controllers/Jetpack/AccountController](https://github.com/woocommerce/google-listings-and-ads/blob/292cf662bfc6bf976ed926785bdd86cbca175631/src/API/Site/Controllers/Jetpack/AccountController.php#L126). I think it's fine for controllers doing such thing like request transformation. If we would move it to the service layer then it means the service layer would need to understand about the request of controller. Which to me is a little bit not that "separation of concerns".

src/API/Site/Controllers/RestAPI/AuthController.php
ianlin: Yeah, I think changing to `settings` sounds more straightforward to me. I updated in b41bad3, however in Google/Jetpack's AccountController they still use reconnect as well as some places in the frontend, I will create a follow up issue so we could do it in another PR.

PR #2294, opened 2024-03-05
Grant consent before checking the gtag events

tomalec:

Fixes tests for gtag after consent mode was added.

Adds a plugin with a PHP snippet to grant all params for consent mode, so the E2E tests will have permissions to track events regardless of the environment's region.

As we're planning further improvements to `gtag` implementation (making it closer to GA4W), right now, we'll use this workaround, then we'll add tests suite for `gtag` and consent mode, once the implementation and feature set will be in the final shape.

1. checkout, build
2. run E2Es:
```
npm run wp-env:up
npm run test:e2e
```

### Changelog entry

> Dev - Fix E2E tests for gtags consent mode

### DIFF
--- a/tests/e2e/test-snippets/test-snippets.php
+++ b/tests/e2e/test-snippets/test-snippets.php
@@ -9,8 +9,10 @@
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Snippets;
 
-// Customize/disable the gtag consent mode, to make testing easier by granting everything by default.
-// It's a hack to avoid specifying region for E2E environment, but it tests the customization of consent mode.
+/*
+ * Customize/disable the gtag consent mode, to make testing easier by granting everything by default.
+ * It's a hack to avoid specifying region for E2E environment, but it tests the customization of consent mode.
+ */
 add_filter(
     'woocommerce_gla_gtag_consent',
     function( $old_config ) {
--- a/tests/e2e/test-snippets/test-snippets.php
+++ b/tests/e2e/test-snippets/test-snippets.php
@@ -1,15 +1,16 @@
 <?php
 /**
- * Plugin name: Custom Consent Mode
- * Description: A plugin to customize/disable the gtag consent mode, to make testing esier by granting everything by default.
+ * Plugin name: E2E test snippets
+ * Description: A plugin to provide some PHP snippets used in E2E tests.
  *
  * Intended to function as a plugin while tests are running.
  * It hopefully goes without saying, this should not be used in a production environment.
- * It's a hack to avoid specifying region for E2E environment, but it tests the customization of consent mode.
  */
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Snippets;
 
+// Customize/disable the gtag consent mode, to make testing easier by granting everything by default.
+// It's a hack to avoid specifying region for E2E environment, but it tests the customization of consent mode.
 add_filter(
     'woocommerce_gla_gtag_consent',
     function( $old_config ) {
--- a/tests/e2e/test-snippets/test-snippets.php
+++ b/tests/e2e/test-snippets/test-snippets.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Plugin name: E2E test snippets
+ * Plugin name: GLA Test Snippets
  * Description: A plugin to provide some PHP snippets used in E2E tests.
  *
  * Intended to function as a plugin while tests are running.
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -2,8 +2,8 @@
 	"phpVersion": "8.0",
 	"plugins": [
 		"https://github.com/WP-API/Basic-Auth/archive/master.zip",
-		"./tests/e2e/disable-consent-mode",
 		"./tests/e2e/test-data",
+		"./tests/e2e/test-snippets",
 		"."
 	],
 	"mappings": {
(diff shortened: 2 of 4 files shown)


tests/e2e/test-snippets/test-snippets.php
```diff
@@ -0,0 +1,25 @@
...
+ * Description: A plugin to provide some PHP snippets used in E2E tests.
+ *
+ * Intended to function as a plugin while tests are running.
+ * It hopefully goes without saying, this should not be used in a production environment.
+ */
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Snippets;
+
+// Customize/disable the gtag consent mode, to make testing easier by granting everything by default.
+// It's a hack to avoid specifying region for E2E environment, but it tests the customization of consent mode.
```
puntope: 💅  minor, but maybe we can use a single block comment instead of using // for each line.

tests/e2e/test-snippets/test-snippets.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Plugin name: E2E test snippets
```
puntope: WDYT about naming it as "GLA Test Snippets" to be consistent with the "Gla Test Data" name ?

puntope (approved):
Looking good for me @tomalec 

Left some 💅 comments regarding the code.

tests/e2e/test-snippets/test-snippets.php
tomalec: The way I use comments is:
- block comments `/* */` /`/** */` - for documenting an entity (a module, class, function, etc). 
   More whitespace, and ` * line indention` to catch the attention, for important bits.
- inline comments `//` 
   - for explaining the logic in inline / in between code flow
   - to denote regions/ sections of something that happens
   
  Short and concise not to take too much attention, but flow in line by line with code
   
Like:

```js
/**
 * Checkout module
 *  a module responsible for finishing the purchase...(API/code docs)
 */

/**
 * Checkout form component... (API/code docs)
 */
export class CheckoutForm {
	/**
	 * Submit all the data and finish to create an order...(API/code docs)
	 */
	submit(){
		// Validate data. (explaining flow + section header)
		this.$('.address').validate();
		this.$('.order_notes').validate();

		// Serialize and sanitize (explaining flow + section header)
		let data = this.$('.address').serialize();
		// Server-side expect billing data to be explicitely set,
		// so we fill it using the shipping data as default. (explain "why", flow & details - could happen multi-line)
		data.billing_address = data.billing_address || data.shipping_address;
	}
}
```


Something like the convention described here https://ckeditor.com/docs/ckeditor5/latest/framework/contributing/code-style.html#comments

Problably, it's my habit from the JSDoc days, where `/**` were picked up as autogenerated API docs.

Also, inline comment takes 50% less space ;) which could be good or bad.
IMO, to describe a single function call does not require that much attention to add two more whitespace lines ;) 

-----

Do you suggest some other convention?

tests/e2e/test-snippets/test-snippets.php
tomalec: hm.. GLA is quite redundant, as we're in GLA repo, same goes for E2E ;) as that `e2e/`folder, but I' all for consistency :)

tests/e2e/test-snippets/test-snippets.php
puntope: We don't have any convention I guess. I use `//` for 1 line comments and `/**` for block comments. Thats how was intended in C and PHP 

So in this case I would do:

```
/*
* Customize/disable the gtag consent mode, to make testing easier 
* by granting everything by default.
* 
* It's a hack to avoid specifying region for E2E environment, 
* but it tests the customization of consent mode.
*/
```
https://www.php.net/manual/en/language.basic-syntax.comments.php

mikkamp: Not a strict standard we adhere to, but WordPress also recommends to use `//` only for single line comments. https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-inline-comments

Not that it doesn't have mixed opinions though: https://make.wordpress.org/core/2022/11/24/proposal-amend-the-inline-documentation-standards-for-multi-line-comments/
I'm comfortable with reading both formats.

tomalec: Thanks for all the links :)
Me and my dyslexia totally relate to the sentiment from https://make.wordpress.org/core/2022/11/24/proposal-amend-the-inline-documentation-standards-for-multi-line-comments/ especially:

> - Reduce possibility of typos and eventual parsing errors ... `/*` to `/**`
> - The de-facto JavaScript coding standard [Prettier](https://prettier.io/)
> - visually very close to the DocBlock standard which may lead to a moment of confusion when reading the code (uh, is this an incomplete DocBlock

But seeing the spiciness of some comments on the WP post and pretty strong opinions from PHP perspective (comparing JS and Prettier conventions _"The recipe instructions for making cheese on toast"_), I'll just hide my set of opinions and struggles and stick to the WP PHPCS rules for JS ;)

tomalec: FWIW, another argument for my convention is that if you select multiple lines in VSCode and press `ctrl + /` you get a nicer result IMHO, especially if you want to comment whole lines without the need select every character. (Plus for some reason `ctrl + shift + a` does not always fork wor me :shrug: )
image

I didn't find a command in VSCode to turn selected lines into
```php
/*
 * selected
 * lines
 */
```
do you know one?

PR #2298, opened 2024-03-05
Refactor Account and Presync issues

jorgemd24:

This PR fetches the account and presync issues using the same background job used to fetch the product statuses. The issues will be fetched in the first batch. The logic remains mostly unchanged from before, the main difference is the addition of a new property `loading when retrieving the issues. This indicates whether the job is scheduled or running and if the process to fetch the issues is in progress. I'll create a follow-up PR to update the UI to reflect the loading status of the issues.

1. Set your PHP memory limit low so you can catch any heavy process. For example, setting this up to 128MB.
2. Call GET `wp-json/wc/gla/mc/product-statistics/refresh`
3. Go to WooCommerce -> Status -> Action Scheduler -> Search for `gla/jobs/update_merchant_product_statuses/process_item` see how the jobs get completed successfully.
4. See how the table wp_gla_merchant_issues gets populated with presync and account issues.

### Additional details:
- I refactor `maybe_refresh_status_data` so it returns the AS job, so it is easier to determine if the process of updating the status and issues is scheduled or in progress.
- If the mc_statuses is not set for the product, we will consider that product is not synced in the MC. See here: https://github.com/woocommerce/google-listings-and-ads/pull/2255#issuecomment-1941282154

### Changelog entry

>

### DIFF
--- a/src/API/Site/Controllers/MerchantCenter/IssuesController.php
+++ b/src/API/Site/Controllers/MerchantCenter/IssuesController.php
@@ -179,6 +179,12 @@ class IssuesController extends BaseOptionsController {
 				'context'  => [ 'view' ],
 				'readonly' => true,
 			],
+			'loading'        => [
+				'type'        => 'boolean',
+				'description' => __( 'Whether the product issues are loading.', 'google-listings-and-ads' ),
+				'context'     => [ 'view' ],
+				'readonly'    => true,
+			],			
 		];
 	}
 
--- a/src/API/Site/Controllers/MerchantCenter/IssuesController.php
+++ b/src/API/Site/Controllers/MerchantCenter/IssuesController.php
@@ -11,7 +11,6 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
 use Exception;
 use WP_REST_Request as Request;
-use WP_REST_Response as Response;
 
 defined( 'ABSPATH' ) || exit;
 
--- a/src/API/Site/Controllers/MerchantCenter/IssuesController.php
+++ b/src/API/Site/Controllers/MerchantCenter/IssuesController.php
@@ -113,7 +113,7 @@ class IssuesController extends BaseOptionsController {
 	 */
 	protected function get_schema_properties(): array {
 		return [
-			'issues' => [
+			'issues'  => [
 				'type'        => 'array',
 				'description' => __( 'The issues related to the Merchant Center account.', 'google-listings-and-ads' ),
 				'context'     => [ 'view' ],
@@ -169,22 +169,22 @@ class IssuesController extends BaseOptionsController {
 					],
 				],
 			],
-			'total'  => [
+			'total'   => [
 				'type'     => 'numeric',
 				'context'  => [ 'view' ],
 				'readonly' => true,
 			],
-			'page'   => [
+			'page'    => [
 				'type'     => 'numeric',
 				'context'  => [ 'view' ],
 				'readonly' => true,
 			],
-			'loading'        => [
+			'loading' => [
 				'type'        => 'boolean',
 				'description' => __( 'Whether the product issues are loading.', 'google-listings-and-ads' ),
 				'context'     => [ 'view' ],
 				'readonly'    => true,
-			],			
+			],
 		];
 	}
 
--- a/src/DB/ProductFeedQueryHelper.php
+++ b/src/DB/ProductFeedQueryHelper.php
@@ -78,14 +78,15 @@ class ProductFeedQueryHelper implements ContainerAwareInterface, Service {
 	 * @throws Exception If the status data can't be retrieved from Google.
 	 */
 	public function get( WP_REST_Request $request ): array {
-		$this->request          = $request;
-		$products               = [];
-		$args                   = $this->prepare_query_args();
-		list( $limit, $offset ) = $this->prepare_query_pagination();
+		$this->request           = $request;
+		$products                = [];
+		$args                    = $this->prepare_query_args();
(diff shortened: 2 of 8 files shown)


src/API/Site/Controllers/MerchantCenter/IssuesController.php
```diff
@@ -113,7 +113,7 @@ protected function get_issues_read_callback(): callable {
 	 */
 	protected function get_schema_properties(): array {
 		return [
-			'issues' => [
```
puntope, line 116: 💅  `use WP_REST_Response as Response;`Is unused.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -168,13 +160,16 @@ public function get_product_statistics( bool $force_refresh = false ): array {
 	 * @throws Exception If the account state can't be retrieved from Google.
 	 */
 	public function get_issues( string $type = null, int $per_page = 0, int $page = 1, bool $force_refresh = false ): array {
```
puntope, line 162: 💅  What do you think about updating the return docs?

`* @return array With two indices, results (may be paged) and count (considers type).`

Now we are returning 'loading' as well

src/DB/ProductFeedQueryHelper.php
```diff
@@ -94,14 +95,20 @@ public function get( WP_REST_Request $request ): array {
...
-			$id     = $product->get_id();
-			$errors = $product_helper->get_validation_errors( $product );
+			$id        = $product->get_id();
+			$errors    = $product_helper->get_validation_errors( $product );
+			$mc_status = $product_helper->get_mc_status( $product ) ?: $product_helper->get_sync_status( $product );
+
+			// If the refresh_status_data_job is scheduled, we don't know the status yet as it is being refreshed.
+			if ( $refresh_status_data_job && $refresh_status_data_job->is_scheduled() ) {
+				$mc_status = '-';
+			}
```
puntope: ❓  Do we really need this chunk of code? Maybe we can keep `status`as null when is refreshing.

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -237,33 +232,30 @@ protected function check_mc_is_connected() {
...
+		// Only refresh if the current data has expired.
+		$this->mc_statuses = $this->container->get( TransientsInterface::class )->get( Transients::MC_STATUSES );
+		$job               = $this->container->get( UpdateMerchantProductStatuses::class );
 
-		// Include any custom merchant issues.
-		$this->refresh_custom_merchant_issues();
+		// If force_refresh is true or if not transient, return empty array and scheduled the job to update the statuses.
+		if ( ! $job->is_scheduled() && ( $force_refresh || ( ! $force_refresh && null === $this->mc_statuses ) ) ) {
+			// Schedule job to update the statuses.
+			$job->schedule();
```
puntope, line 255: Why we need double-checking if the job is running? 

AFAIK `schedule()` checks is_running() for the current page. 

I see you created the is_Schedule function without arguments to avoid page checks? 

```php
public function is_scheduled(): bool {
		// We set 'args' to null so it matches any arguments. This is because it's possible to have multiple instances of the job running with different page tokens
		return $this->is_running( null );
	}
```

Maybe we can just tweak `can_schedule` function and then rely on $job->schedule itself to check if the job is running. 

```php
/**
	 * Can the job be scheduled.
	 *
	 * @param array|null $args
	 *
	 * @return bool Returns true if the job can be scheduled.
	 */
	public function can_schedule( $args = [] ): bool {
		return parent::can_schedule( $args ) && $this->merchant_center->is_connected() && ! $this->is_running( null );
	}
```

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -437,6 +429,22 @@ protected function get_product_issues( array $statuses ): array {
 		return $product_issues;
 	}
 
+	/**
+	 * Refresh the account , pre-sync product validation and custom merchant issues.
+	 *
+	 * @throws Exception  If the account state can't be retrieved from Google.
+	 */
+	public function refresh_account_and_presync_issues(): void {
```
puntope, line 439: 💅  Maybe we can add `@since` param

puntope (commented):
Thanks @jorgemd24 for the refactor work. Everything works as expected. I left just some comments regarding the code.

src/MerchantCenter/MerchantStatuses.php
jorgemd24: Thanks @puntope for the suggestion, if we do it like that, the second job won't start because at the time we schedule it, the first job hasn't finished yet.

https://github.com/woocommerce/google-listings-and-ads/blob/6b7a614b36fe71f0674025eda35e388209081d62/src/Jobs/UpdateMerchantProductStatuses.php#L101

src/DB/ProductFeedQueryHelper.php
jorgemd24: I think it is necessary because if there's no `mc_status`, it'll just return 'not-synced'. And most of the time, `mc_status` won't be null.

>If the mc_statuses is not set for the product, we will consider that product is not synced in the MC. See here: https://github.com/woocommerce/google-listings-and-ads/pull/2255#issuecomment-1941282154

https://github.com/woocommerce/google-listings-and-ads/blob/6b7a614b36fe71f0674025eda35e388209081d62/src/DB/ProductFeedQueryHelper.php#L100

I switched from using `-` to `null` because it might be clearer to indicate that there's no status yet.

jorgemd24:
Thanks @puntope for your time and your suggestions! I addressed your comments and replied to your questions! Do you mind to have a second look? Thanks!

puntope (approved):
Thanks for changes and explanation @jorgemd24 Looking good to me.

PR #2301, opened 2024-03-07
Fix exception when response.json() fails

puntope:

Right now in the extension when creating Ads Account it returns 428 status sometimes. 

We don't block the Account creation in this case.

However, when testing the extension with the Jurassic Ninja site. I got a 428 status but with a wrong response body different from JSON. So when the API Fetch Hook tries to parse the JSON after the exception happens it generates another exception causing the Ads Creation to fail and get blocked.

### Screenshots:

Before:

https://github.com/woocommerce/google-listings-and-ads/assets/5908855/29ac4a74-bad6-4661-9245-5ae881a8edd3

After:

https://github.com/woocommerce/google-listings-and-ads/assets/5908855/26f7d008-c011-49c8-92f1-2c1050d0dc15

1. Set up a site in jurassic.ninja
2. Install WC and GLA (archived using this branch)
3. Check that there is no error when creating Ads Account.

### Changelog entry

> Fix - Handle parse JSON exception when Creating Ads Account

### DIFF
--- a/js/src/hooks/useApiFetchCallback.js
+++ b/js/src/hooks/useApiFetchCallback.js
@@ -173,11 +173,16 @@ const useApiFetchCallback = ( options, initialState = defaultState ) => {
 				}
 
 				const response = e;
+				let error;
 
-				const responseClone = response.clone();
-				const error = responseClone.json
-					? await responseClone.json()
-					: new Error( 'No content body in fetch response.' );
+				try {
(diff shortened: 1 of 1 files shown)


jorgemd24 (approved):
Thanks @puntope  for fixing this. I could reproduce on JN and I can confirm that this tweak fixed the issue so LGTM 👍

Issue #2251, opened 2024-02-08
Google Ads connection error displayed despite successfully connecting due to fatal error


martynmjones:
### Describe the bug:
During onboarding I encountered the `Unable to connect your Google Ads account. Please try again later.` error message caused by a `500` response from the `wp-json/wc/gla/ads/accounts` endpoint. The connection was successful and the card was updated by refreshing the onboarding page.

**Logged error for the request:**

```
Uncaught TypeError: array_merge(): Argument #1 must be of type array, null given in /srv/htdocs/wp-content/plugins/google-listings-and-ads/src/API/Google/Merchant.php:329
```

<img width="1084" alt="Screenshot 2024-02-08 at 12 31 41" src="https://github.com/woocommerce/google-listings-and-ads/assets/40762232/35d0e0ff-6f3f-4242-ad28-9b51da276bb6">

### Steps to reproduce:

Error isn't consistent and I've also been able to successfully onboard without issue. Looking through the support forums I'm not seeing reports similar to this so it's likely not a wide spread issue but will need some further investigation and better error handling.

martynmjones:
Closing as this issue will be addressed as part of https://github.com/woocommerce/google-listings-and-ads/pull/2205#discussion_r1475192067

PR #2302, opened 2024-03-08
Fix - Handle null value for getAdsLinks response

jorgemd24:

In https://github.com/woocommerce/google-listings-and-ads/issues/2251, it was mentioned that `getAdsLinks` can return a `null` value instead of the expected array, leading to a fatal error. The fix was done in https://github.com/woocommerce/google-listings-and-ads/pull/2205#discussion_r1489476259, however, it hasn't been merged into the develop branch yet. This PR brings the fix to the develop branch to prevent the fatal error.

1. I haven't been able to reproduce the issue, but the fatal error mentioned https://github.com/woocommerce/google-listings-and-ads/issues/2251 makes it clear that `getAdsLinks` returns a null value in some cases.
2. Simulate the getAdsLinks returns a `null` value.

### Changelog entry

> Fix - Fatal error when getAdsLinks response is null

### DIFF
--- a/src/API/Google/Merchant.php
+++ b/src/API/Google/Merchant.php
@@ -314,7 +314,7 @@ class Merchant implements OptionsAwareInterface {
 	 */
 	public function link_ads_id( int $ads_id ): bool {
 		$account   = $this->get_account();
-		$ads_links = $account->getAdsLinks();
+		$ads_links = $account->getAdsLinks() ?? [];
 
 		// Stop early if we already have a link setup.
 		foreach ( $ads_links as $link ) {


mikkamp (approved):
Thanks for the change. As mentioned in https://github.com/woocommerce/google-listings-and-ads/pull/2205#discussion_r1475848749 I went pretty far into debugging, and because of the `adsLinksDataType` the return value is always forced to be initialized as an array. So I'm unable to find a scenario where it is null.

Either way though, I tested this PR by forcing the links to be null using:
```
$account->adsLinks = null;
```

When doing so I get both a warning and a fatal error:
```
[08-Mar-2024 10:39:40 UTC] PHP Warning:  foreach() argument must be of type array|object, null given in google-listings-and-ads/src/API/Google/Merchant.php on line 321
[08-Mar-2024 10:42:30 UTC] PHP Fatal error:  Uncaught TypeError: array_merge(): Argument #1 must be of type array, null given in google-listings-and-ads/src/API/Google/Merchant.php:330
```

The fix in this PR resolves that issue, so it looks good to me.

PR #2306, opened 2024-03-12
Refactor Product issues and fix issues when getting stats for a fresh MC account

jorgemd24:

While I was testing the new approach with a fresh Merchant Center account I found some issues that this PR fixes:

1. After the onboarding, we call `mc/settings/sync`, which kicks off the `woocommerce_gla_mc_settings_sync hook` and schedules the `UpdateAllProducts` job. As each batch gets processed, it triggers the `woocommerce_gla_batch_updated_products` hook, clearing the merchant statuses cache for each batch. This can create a loop: stats are fetched, and then deleted when the batch is processed. So, if the user refreshes the stats, it'll redo the calculation only to delete them again.

https://github.com/woocommerce/google-listings-and-ads/blob/f4aaafbf5f4c9795451ae464e7911e37ba2998f1/src/Event/ClearProductStatsCache.php#L38-L51

This PR checks if `UpdateAllProducts` or `DeleteAllProducts` jobs is running before clearing the cache and will refresh the data when the batches are completed.

2. With a fresh MC, I noticed that the results from `MerchantProductViewReportQuery` and `get_productstatuses_batch` are returned as `null` if there is no data, instead of an empty array, this PR fixes this using `?? []`.

Also, this PR tweaks some parts of the code:

3. Remove `get_synced_google_ids` as it is not used anymore.
4. In case the job fails, it will restore the intermediate data to the previous state instead of deleting the intermediate data. So when AS retries the job, it will start with the correct data.
5. When `handle_complete_mc_statuses_fetching` gets called, we check if the intermediate data is set. If it is, we mark the process as complete. However, I reckon it's safer to return a default intermediate state just in case the data somehow gets deleted. This way, the process can always be completed.
6. I decided to remove `$this->clear_cache();` from `clear_product_statuses_cache_and_issues` because when the product stats are updated, the transient gets updated automatically. This way, the user won't even notice the process happening.

1. Disconnect all the account connections.
2. Complete the onboarding process with a new MC account.
4. Check that when all `update_all_products` jobs finish, it triggers `update_merchant_product_statuses`.
5. Once `update_merchant_product_statuses` is done, verify that the product statistics are displayed and that the account, pre-sync, and production issues are populated. (Dashboard -> Produc Feed -> Overview)
6. Initially, it may indicate that not all products are synced, as Google may require some time to process them.
7. After some time, refresh the statistics using `GET wc/gla/mc/product-statistics/refresh` and check if the stats and product issues are updated accordingly.

### Changelog entry

>


src/API/Google/MerchantReport.php
```diff
@@ -84,7 +84,9 @@ public function get_product_view_report( $next_page_token = null ): array {
 			->set_client( $this->service, $this->options->get_merchant_id() )
 			->get_results();
 
-			foreach ( $response->getResults() as $row ) {
+			$results = $response->getResults() ?? [];
+
+			foreach ( $results as $row ) {
```
puntope, line 89: I'm a bit confused about this part of the code.  First, we do get the results from the query and save it to he response with `get_results()`: 

```php
$response = $query
			->set_client( $this->service, $this->options->get_merchant_id() )
			->get_results();
```
Then we call `$response->getResults()` 

So? Are we getting the results of the results?

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -178,7 +199,14 @@ public function get_issues( string $type = null, int $per_page = 0, int $page =
 	 * @since 1.1.0
 	 */
 	public function clear_cache(): void {
-		$this->container->get( TransientsInterface::class )->delete( TransientsInterface::MC_STATUSES );
+		$update_all_products_job = $this->container->get( UpdateAllProducts::class );
+		$delete_all_products_job = $this->container->get( DeleteAllProducts::class );
+
+		// Clear the cache if we are not in the middle of updating/deleting all products. Otherwise, we might update the product stats for each individual batch.
+		// See: ClearProductStatsCache::register
+		if ( $update_all_products_job->can_schedule( null ) && $delete_all_products_job->can_schedule( null ) ) {
```
puntope, line 207: Why `can_schedule` and not `is_running`?

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -205,7 +233,6 @@ protected function delete_stale_issues(): void {
 	 * @since x.x.x
 	 */
 	public function clear_product_statuses_cache_and_issues(): void {
-		$this->clear_cache();
```
puntope, line 208: Wonder if the name of the function still valid after removing `clear_product_statuses_cache_and_issues`

src/API/Google/MerchantReport.php
jorgemd24: Yeah, I get that it's a bit confusing, but I think that's just how the Google Client works. It fetches the results from the queries first and then the rows. Maybe it would be clearer if the Google client had methods like `get_results` and `get_rows`.

See a similar example here (line 162 and 166):

https://github.com/woocommerce/google-listings-and-ads/blob/71b864b49f0740b9a7e936a0c1d3fcad5708488c/src/API/Google/MerchantReport.php#L160-L168

src/MerchantCenter/MerchantStatuses.php
jorgemd24: Because `is_running` is a protected method, I understand that maybe it would be a bit easier to read the code if we use a method name similar to `is_running`. Although I think the comment that I left explains what we are doing.

src/MerchantCenter/MerchantStatuses.php
jorgemd24: In part remains valid because we're still deleting any cached intermediate data. I was thinking of renaming it to `clear_product_statuses_intermediate_data_and_issues`. What do you think? Or do you have a better suggestion?

src/MerchantCenter/MerchantStatuses.php
puntope: More than the naming I was more thinking about the logic in can_schedule. That also checks some conditions. If MC connected for example. But if that doesn't create a problem I think we are good with can_schedule

src/MerchantCenter/MerchantStatuses.php
puntope: I would simplify with `clear_data` so we don't attach too much the name with the logic inside, maybe we lost clarity of what the function does, but with an accurate PHPDOC description, I think is ok. 

However, was just a 💅 / ❓  Not a blocker keeping it like it is

puntope:
Hi @jorgemd24 seems like is looking good. However, I cannot see the products populated (I waited a while) 🤔  I will check tomorrow morning to verify if they loaded already.

jorgemd24:
Thanks @puntope for the review! 

> Hi @jorgemd24 seems like is looking good. However, I cannot see the products populated (I waited a while) 🤔 I will check tomorrow morning to verify if they loaded already.

Do you mean that the products aren't synchronized in the MC? If that's the case, it's worth mentioning that we haven't changed any logic with the product syncer. We've only refactored how we retrieve the product stats.

puntope:
> Thanks @puntope for the review!
> 
> > Hi @jorgemd24 seems like is looking good. However, I cannot see the products populated (I waited a while) 🤔 I will check tomorrow morning to verify if they loaded already.
> 
> Do you mean that the products aren't synchronized in the MC? If that's the case, it's worth mentioning that we haven't changed any logic with the product syncer. We've only refactored how we retrieve the product stats.

Ok, thanks @jorgemd24 Let's suppose that is another issue unrelated to this PR.  

However,  I synced a product manually via Connection test page, just to check:

When I run it on this branch I got

```js
{
    "timestamp": 1710522408,
    "statistics": {},
    "scheduled_sync": 23,
    "loading": true,
    "error": null
}
```

if I checkout develop and run it again I got

```js
{
    "timestamp": 1710522512,
    "statistics": {
        "active": 0,
        "expiring": 0,
        "pending": 0,
        "disapproved": 1,
        "not_synced": 0
    },
    "scheduled_sync": 23
}
```

In the dashboard. In develop I see the product as "disapproved", in this branch the product keeps as "Not synced" and the statistics shows N/A

Develeop
Screenshot 2024-03-15 at 12 10 07


This branch
Screenshot 2024-03-15 at 12 11 08

Just wonder if this is expected. And if not, I wonder if that's related to this PR or is a separated issue.

jorgemd24:
Hi @puntope,

Thanks for your comments, regards:

```
{
    "timestamp": 1710522408,
    "statistics": {},
    "scheduled_sync": 23,
    "loading": true,
    "error": null
}
```


On that response, it appears that the job `update_merchant_product_statuses` was either running or scheduled to run (see `loading` = true). Additionally, it seems that at that moment, there were 23 jobs scheduled to sync, so it's possible that the `update_merchant_product_statuses` job hadn't been executed yet due to other jobs in the queue. Once the job is executed and finishes fetching all the stats, you should get the correct stats.

Regarding the screenshots, that' expected. I'm currently working on adding some spinning icons to indicate when the `update_merchant_product_statuses` job is scheduled or running, instead of the "N/A" currently displayed.

puntope:
Hi @jorgemd24 All the jobs mentioned are completed. But I still get the same response:

```
{
    "timestamp": 1710690602,
    "statistics": {},
    "scheduled_sync": 23,
    "loading": true,
    "error": null
}
```

However, I was able to see the stats finally in the dashboard. Let me know if that refresh stats response is expected or we should continue investigating.

jorgemd24:
Hi @puntope, 

>
```
{
    "timestamp": 1710690602,
    "statistics": {},
    "scheduled_sync": 23,
    "loading": true,
    "error": null
}
```
That response indicates that the `update_merchant_product_statuses` job was scheduled so that's why no stats should be shown. 

> However, I was able to see the stats finally in the dashboard. Let me know if that refresh stats response is expected or we should continue investigating. But it also shows 

I guess later on `update_merchant_product_statuses` job was completed and the response should look like this:

```
{
    "timestamp": 1710711097,
    "statistics": {
      "active" : 1
      "disapproved" : 2
      "expiring" : 3 
      "not_synced" : 4
      "pending" : 5
    },
    "scheduled_sync": 0,
    "loading": false,
    "error": null
}
```

If that's the case, it's expected. Keep in mind, that the UI isn't refreshing the data at the moment (at least that you refresh the page manually) because in this PR only contains the backend.

puntope (approved):
Hi @jorgemd24 Finally the issue was because I thought  calling `wc/gla/mc/product-statistics/refresh` would return the stats

Calling `wc/gla/mc/product-statistics ` return the right response

```
{
    "timestamp": 1710842811,
    "statistics": {
        "active": 0,
        "expiring": 0,
        "pending": 0,
        "disapproved": 1,
        "not_synced": 18
    },
    "scheduled_sync": 23,
    "loading": false,
    "error": null
}
```

PR #2311, opened 2024-03-13
API Pull - Grant access to google WPCOM app in settings page

ianlin:

Project: pcTzPl-1SM-p2

Part of #2146.

This PR co-authored with @puntope which it adds the granting access to Google WPCOM app flow in the settings page.
Note that the part of displaying authorisation status in the GMC card were already reviewed and tested in #2325.

The flow is:

#### Displaying "enable new product sync" banner

1. When the user enters the settings page, and if the DB option `gla_wpcom_rest_api_status` is not set, it will display a banner on the top of the page, asking merchants to enable new product sync feature.
2. User clicks on `Enable`, it will call the API `GET /wc/gla/rest-api/authorize` to get the auth URL, then redirect to it.

In the step 1, if `gla_wpcom_rest_api_status` is set, the banner will not be shown.

#### Displaying "authorisation status" in the GMC card

This was already tested and reviewed in #2325.

#### Setting authorisation status in the DB option `gla_wpcom_rest_api_status` 

1. When enters the settings page, it will check if the url includes a query parameter `google_wpcom_app_status`
2. ~Also get another query parameter `site_nonce`, which was created when generating auth URL. This query param was added by Google when they redirect back to merchant site.~
3. If yes, if will call an API `PUT /wc/gla/rest-api/authorize` to update it in the DB

Note that this query parameter will be added by the Google side, as at the end of the authorisation flow, Google will be the one who redirect the merchant back to the merchant site.

#### Displaying "enable new product sync" banner

1. Ensure the option `gla_wpcom_rest_api_status` is not set
2. Go to the settings page, you should see the banner on the top of the page displaying enable new product sync.
4. Click `Enable`, it will be redirected to the WPCOM app granting access page
    - Screenshot 2024-03-26 at 21 36 09

5. Finish the auth step, it will eventually be redirect to https://woo.com
    - Because currently we are using a test WPCOM app, in the future this will the Google's WPCOM app and they will help us to redirect back to the merchant site.
6. Set the option `gla_wpcom_rest_api_status` to `approved`, `disapproved`, or `error`, the banner will not be shown

#### Setting authorisation status in the DB option `gla_wpcom_rest_api_status`

1. Ensure the option `gla_wpcom_rest_api_status` is not set
2. Go to the settings page, click enable product sync.
3. ~After redirected to the WPCOM OAuth page, copy the value of the query parameter `state`.~
4. ~The value is encoded in base64, here's a little script to retrieve the `site_nonce` from the state value:~
    ```php
    <?php

    $b64 = 'PASTE_THE_STATE_VALUE_HERE';

    $decoded_b64 = base64_decode( $b64 );
    $url_decoded = urldecode( $decoded_b64 );
    parse_str( $url_decoded, $parsed );
    print_r( $parsed['site_nonce'] );
    ```
5. Go to the settings page, with extra query parameters `google_wpcom_app_status=approved` ~and `site_nonce=your-nonce`~. E.g.
  - `https://example.com/wp-admin/admin.php?page=wc-admin&path=/google/settings&google_wpcom_app_status=approved`
6. The banner should not be shown
7. The option `gla_wpcom_rest_api_status` in the DB is set to `approved`.
8. Test the step 2 again but with different value for the `google_wpcom_app_status` query parameter, e.g. `google_wpcom_app_status=disapproved` and `google_wpcom_app_status=error`
9. The banner should not be shown either and the the option in the DB is set to the correct value.
10. ~Test the step 2 again but with a wrong value for the `site_nonce` query parameter, the option should not be updated in the DB, and you can see the error in the DEV console and the Network tab~.

### Additional details:

1. Note that since we haven't got the detail of Google's WPCOM app, this PR is currently using a test WPCOM app. So at the end of the auth flow it will not redirect to the merchant site automatically.
2. The above lines with strikethrough about verifying `site_nonce` will be implemented in another PR.


js/src/components/google-mc-account-card/connected-google-mc-account-card.js
```diff
@@ -85,6 +97,44 @@ const ConnectedGoogleMCAccountCard = ( {
...
+			);
+		}
+
+		removeNotice( notice.id );
+	};
+
+	// Show the button if the status is "approved" and the Notification Service is not hidden.
+	const showDisconnectNotificationsButton =
+		! hideNotificationService &&
+		googleMCAccount.wpcom_rest_api_status === 'approved';
```
puntope: Since this PR was merged before some of the changes were finished. We missed to update the constant here.

js/src/components/google-mc-account-card/connected-google-mc-account-card.js
```diff
@@ -85,6 +97,44 @@ const ConnectedGoogleMCAccountCard = ( {
...
+
+	// Show the error if the status is set but is not "approved" and the Notification Service is not hidden.
+	const showErrorNotificationsNotice =
+		! hideNotificationService &&
+		googleMCAccount.wpcom_rest_api_status &&
+		googleMCAccount.wpcom_rest_api_status !== 'approved';
```
puntope: Same as https://github.com/woocommerce/google-listings-and-ads/pull/2311/files#r1539571082

js/src/components/google-mc-account-card/connected-google-mc-account-card.js
```diff
@@ -46,6 +50,14 @@ const ConnectedGoogleMCAccountCard = ( {
...
 		} );
 
+	const [
+		fetchDisableNotifications,
+		{ loading: loadingDisableNotifications },
+	] = useApiFetchCallback( {
+		path: `${ API_NAMESPACE }/rest-api/authorize`,
+		method: 'DELETE',
+	} );
+
```
puntope, line 61: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2311/files#r1539571082

I guess we need to update this to use the hook you created

js/src/components/enable-new-product-sync-button.js
```diff
@@ -0,0 +1,49 @@
...
+	const { createNotice } = useDispatchCoreNotices();
+
+	const nextPageName = glaData.mcSetupComplete ? 'settings' : 'setup-mc';
+	const query = { next_page_name: nextPageName };
+	const path = addQueryArgs( `${ API_NAMESPACE }/rest-api/authorize`, query );
+	const [ fetchRestAPIAuthorize ] = useApiFetchCallback( { path } );
+	const handleEnableClick = async () => {
+		try {
+			const d = await fetchRestAPIAuthorize();
+			window.location.href = d.auth_url;
```
puntope, line 32: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2311/files#r1539571082

Not sure if we should change the logic and use the hook here

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "google-listings-and-ads",
-	"version": "2.6.3",
```
puntope: Is this commit necessary?

puntope:
Maybe we can tweak a bit the Banner info, it lacks a bit of information for the user IMO?

Screenshot 2024-03-26 at 17 31 17

puntope (commented):
Thanks for working on this @ianlin 

I can see the DB updated when adding the query in the URL (and hence the banner + info logic)

I left some comments tho regarding the hook and some constants.

js/src/components/google-mc-account-card/connected-google-mc-account-card.js
ianlin: Hey @puntope, thanks for reviewing. I don't quite get what should be updated here? This is the same as in the PR 2325 https://github.com/woocommerce/google-listings-and-ads/pull/2325/files#diff-b7a4b2fed5fcd9b3e8df480b7fa40000976fc614ddc77831ab62970ad9702a2fR126-R128. Maybe I missed something?

js/src/components/google-mc-account-card/connected-google-mc-account-card.js
ianlin: Thanks for the suggestion. The hook `useUpdateRestAPIAuthorizeStatus` I introduced has a specific purpose which it would [check the query parameter from the URL](https://github.com/woocommerce/google-listings-and-ads/pull/2311/files#diff-a881f5d59bd2db65aafd56b0032c74cfcf652451c72a46966bb22b4bc94fb512R24) and then call an API to update the auth status based on the query param. Right now it is only used in the settings page and going forward this hook would be used in the onboarding flow as well, as we'd be adding the granting process in the onboarding flow in the next release phase soon.

The code you mentioned here also has its own purpose which is to call the DELETE auth status API when user clicks the `Disable product data fetch` button. So I think maybe the hook I introduced couldn't be shared with it here.

But I get what you mean, since deleting auth status is also some form of "update the status", so I renamed the hook to `useUpdateRestAPIAuthorizeStatusByUrlQuery` in 501c651 to make it more specifically.

js/src/components/enable-new-product-sync-button.js
ianlin: Same as above, the code here is getting the auth status by calling GET auth status API so I think the hook `useUpdateRestAPIAuthorizeStatusByUrlQuery` couldn't be used here.

package-lock.json
ianlin: I think I did not add this commit. But I merged the latest base branch and solved the conflicts so the `package-lock.json` is the latest version now.

ianlin:
Hey @puntope thanks for reviewing and the suggestions. I've added some commits to reflect your suggestion and also left comments for discussion.

src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -88,6 +107,38 @@ protected function get_authorize_callback(): callable {
...
+			}
+		};
+	}
+
+	/**
+	 * Get the callback function for the update authorize request.
+	 *
+	 * @return callable
+	 */
+	protected function get_update_authorize_callback(): callable {
```
jorgemd24, line 133: Shall we add the @since  x.x.x docs?

src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -88,6 +107,38 @@ protected function get_authorize_callback(): callable {
...
+		return function ( Request $request ) {
+			try {
+				$this->account_service->update_wpcom_api_authorization( $request['status'] );
```
jorgemd24, line 136: Shouldn't we verify if Google is the one setting the `google_wpcom_app_status` parameter when it's set in the URL? For instance, by checking the nonce or hash? I understand we don't know yet how the callback URL will look, but it might be something to consider; otherwise, anyone could set the status.

js/src/hooks/useUpdateRestAPIAuthorizeStatusByUrlQuery.js
```diff
@@ -0,0 +1,65 @@
...
+import { API_NAMESPACE } from '.~/data/constants';
+import useApiFetchCallback from '.~/hooks/useApiFetchCallback';
+
+/**
+ * A hook that calls an API to update Google WPCOM app connected status.
+ *
+ * At the end of the authorization for granting access to Google WPCOM app, Google will
+ * redirect back to the merchant site, either the settings page or onboarding setup account page.
+ * They will add a query param `google_wpcom_app_status` to the URL, we will store this status to
+ * the DB by calling an API `POST /wc/gla/rest-api/authorize`.
```
jorgemd24: I think instead of `POST` should be `PUT`

js/src/components/google-mc-account-card/google-mc-account-card.js
```diff
@@ -23,7 +23,12 @@ const GoogleMCAccountCard = () => {
 		return <NonConnected />;
 	}
 
-	return <ConnectedGoogleMCAccountCard googleMCAccount={ googleMCAccount } />;
+	return (
+		<ConnectedGoogleMCAccountCard
+			hideNotificationService
```
jorgemd24, line 28: It seems like `hideNotificationService` isn't defined. I believe its value will always be `false`. Is this intentional, or am I missing something?

src/API/Site/Controllers/RestAPI/AuthController.php
```diff
@@ -105,6 +156,23 @@ protected function get_auth_params(): array {
...
+	 * Get the query params for the update authorize request.
+	 *
+	 * @return array
+	 */
+	protected function get_update_authorize_params(): array {
+		return [
+			'status' => [
+				'description'       => __( 'The status of the merchant granting access to Google\'s WPCOM app', 'google-listings-and-ads' ),
+				'type'              => 'string',
+				'enum'              => [ 'approved', 'disapproved', 'error' ],
```
jorgemd24: What do you think about making these enums constants of the `OAuthService class`? That way, if we need to check the status later, we could simply use something like `OAuthService::APPROVED` instead of typing out the values each time.

jorgemd24 (commented):
Hi @ianlin, thanks for working on this! It's working as expected 👍  I've left some comments about the code that we could discuss further.

js/src/components/google-mc-account-card/connected-google-mc-account-card.js
puntope: I mean that instead of 'approved' we can use some constants like `GOOGLE_WPCOM_APP_CONNECTED_STATUS_APPROVED`

js/src/components/google-mc-account-card/google-mc-account-card.js
puntope: This was introduced in my PR. In the settings is false, in Onboarding is true @jorgemd24

js/src/components/google-mc-account-card/connected-google-mc-account-card.js
ianlin: Ah yeah, thanks for clarifying. Updated in 07ddeab.

src/API/Site/Controllers/RestAPI/AuthController.php
ianlin: Updated in cc23c0b.

js/src/components/google-mc-account-card/google-mc-account-card.js
ianlin: Yeah, the prop with just its name without any value is considered true.

js/src/hooks/useUpdateRestAPIAuthorizeStatusByUrlQuery.js
ianlin: Updated in 2262966.

src/API/Site/Controllers/RestAPI/AuthController.php
ianlin: Thanks for the suggestion, updated in f19b5f7.

src/API/Site/Controllers/RestAPI/AuthController.php
ianlin: Thanks for raising this @jorgemd24, that's a very good point and I think we haven't discussed this topic before. Currently a nonce/hash that will be provided by Google when we call a new API (they're still developing) before generating the auth URL, they would verify the nonce when the auth flow is redirect to their `redirect URI`.

If we'd want to do the similar thing and verify at the end of the auth flow on the merchant site, could we use [wp_create_nonce](https://developer.wordpress.org/reference/functions/wp_create_nonce/) to generate a nonce and add it to the `state` query param, and use [wp_verify_nonce](https://developer.wordpress.org/reference/functions/wp_verify_nonce/) to verify it when redirect back to the merchant site?

jorgemd24: Yes, I believe that would work. How about continuing this discussion on the P2 so we can address it in a follow-up PR if needed it?

ianlin: I think those two functions work. I added the nonce protection in commits:
- b9960ce8a07faf9cbcaf04a77fc1063ac9da199f
- 98bfad7b2a1478e041e13cba5f5a1d10ff9beec1
- 30bc13bff4dfe1bad9321eca56923a627a1b9193

ianlin:
Hey @jorgemd24 thanks for reviewing and for all the useful suggestions. I've pushed more commits to reflect your suggestions, especially adding the nonce when generating auth URL, and verify it when updating the oauth status in the DB. I also updated the test instruction at the bottom of the description.

jorgemd24 (commented):
Hi @ianlin! Thanks for the adjustments. I tested the nonce verification, but it didn't work as expected. The problem is when you "approve" the WPCOM App, you're redirected to your site to log in and authorize the app. When you do this, the token session changes, invalidating the nonce. I think the part of the nonce verification can be done in a follow-up PR so we don't block this PR. Let me know what you think. The rest of the functionality works as described in the PR. 

image

image

https://github.com/WordPress/wordpress-develop/blob/86d427315c2e18b01ca6ccfcc5876f617c90af2c/src/wp-includes/pluggable.php#L2333

ianlin:
Hi @jorgemd24, nice catch! My bad that when I was testing I did not test after the log in step so I thought it worked well.

> think the part of the nonce verification can be done in a follow-up PR so we don't block this PR

That makes complete sense, I reverted the commits regarding nonce, could you please take another quick look?

jorgemd24 (approved):
Hi @ianlin thanks for the adjustments, LGTM 👍  We will need to update the PR description so we don't mention the `site_nonce` as we will work on this on a follow-up PR.

PR #2312, opened 2024-03-14
Remove esc_js from add_inline_event_script

martynmjones:

When Google Analytics for WooCommerce version > `2.0.0` is installed, inline Javascript is being incorrectly escaped for GLA events. This PR removes the `esc_js` call so that the code is output correctly and how it previously was.

### Screenshots:

Screenshot 2024-03-14 at 16 13 24

1. Checkout `fix/escaping-inline-javascript` on a test site with Google Analytics for WooCommerce version `2.0.3` installed and active
2. View page source and look for `send_to`
3. Confirm the inline JS for GLA is correct

### Changelog entry

> Fix - Inline Javascript encoding for gtag events


Juberstine:
When is a release going to be cut that addresses this ?

martynmjones:
Hey @Juberstine,

> When is a release going to be cut that addresses this ?

There will be a new release of Google Listings & Ads (_2.6.3_) today that will include this fix.

PR #2314, opened 2024-03-15
Add require plugins header

### LINKED ISSUE #2313: Add new plugin dependency headers
Since WP 6.5 and In relation to this post: pcShBQ-1yd-p2

There is also the suggestion to add the  new header:

```
Requires Plugins: woocommerce
```

This header prevents the plugin from being installed or activated when the header is present. 

See also https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/ to learn about the new dependency header and what to expect about it.

puntope:

Closes #2313 

This PR enables the new Requires Plugins header available in WP 6.5

See --> https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/

### Screenshots:

Screenshot 2024-03-14 at 19 30 20
Screenshot 2024-03-14 at 19 30 36

1. Install WP 6.5 (RC2) --> https://wordpress.org/wordpress-6.5-RC2.zip 
2. Install Woo Commerce and GLA (checkout this PR)
3. If WC is activated, GLA can be activated
4. If WC is not activated, GLA button for activate is disabled. A notice "This plugin cannot be activated because required plugins are missing or inactive." is shown.
5. If GLA is activated and WC is activated. WC cannot be deactivated. And it shows info like "Required by: Google Listings and Ads" 

### Changelog entry

> Tweak - Add WP 6.5 Require plugins header

### DIFF
--- a/google-listings-and-ads.php
+++ b/google-listings-and-ads.php
@@ -11,7 +11,7 @@
  * Tested up to: 6.5
  * Requires PHP: 7.4
  * Requires PHP Architecture: 64 bits
- *
+ * Requires Plugins: woocommerce
  * WC requires at least: 6.9
  * WC tested up to: 8.7
  * Woo:


eason9487 (approved):
Works well. LGTM.

PR #2270, opened 2024-02-21
[R-1.3] Add Google Ads invite acceptance flow

ankitrox:
### Changes proposed in this Pull Request:

Closes #2262.

_Replace this with a good description of your changes & reasoning._

### Detailed test instructions:

**A. Account access endpoint**

1. Go to edit profile in the dashboard and create an [application password](https://make.wordpress.org/core/2020/11/05/application-passwords-integration-guide/).

2. Open POSTman tool and enter the url `https://localhost:8888/wp-json/wc/gla/ads/account-status` or if you have configured local tunnel using ngrok, then it should be `https://{$replace-with-ngrok-domain}/wp-json/wc/gla/ads/account-status`

3. In the `Authorization` tab, add username and password by selecting `basic auth` scheme. Password should be application password created in step 1. Ensure that user has permission to `manage_woocommerce`.

4. Go to step-1 of the onboarding journey and ensure that google account is created. Once that is done, create a new ads account.

5. Once new ad account is created, open POSTman and request the endpoint stated in step-2. It should return the following object.

```js
{"has_access":false,"invite_link":"https:\/\/ads.google.com\/nav\/startacceptinvite?ivid=184609397&ocid=6470691999&eivid=AVqCgUVYjpjHyQW5721GvPhkVdvNRMkJ5_AIJQ5dE_6Iz0b0_LCHn0Q"}
```

**UI**
* Setup a plugin and connect the Jetpack and Google account.
* For Google Ads account, opt to create a new one.
* A modal prompting the user to accept the terms should appear.
* After clicking the "Create account" button there should be a second popup prompting the user to click on the invite link. Copy for the modal can be found here: https://github.com/woocommerce/google-listings-and-ads/issues/2262
* Check if link to the invite is for the correct Google Ads account.
* Open a new tab with the same URL as the setup screen where you won't see the modal but instead see a different component where once again it's prompting the user to claim the account.
* Check if the links within the above component and the one in the modal are all the same.
* Click on the link (modal or non modal) and accept the invite.
* When coming back to the setup page, the status should automatically be updated within 20s (background polling of the status) and Ads account is connected.
* There should be no changes to the setup process.

PR #2316, opened 2024-03-15
Fix: jumping to wrong step in onboarding

ankitrox:

Closes # .

If the user has not accepted ads account invite, do not proceed to step 2 when revisiting the onboarding journey.

1. 
2. 
3. 

### Changelog entry

>


joemcgill:
After doing some more testing of this for #2270, I updated the approach a bit in https://github.com/woocommerce/google-listings-and-ads/pull/2320 so that we only skip to the second page once we've finished connecting the Ads and MC accounts (e.g., only when the Ads account is pending 'billing'.

I think we can close this in favor of that approach.

PR #2318, opened 2024-03-18
Update UI to handle Fetching Product Status with Action Scheduler Jobs

jorgemd24:

While we're refactoring the backend logic for the product stats and moving it to the Action Scheduler Job, we also need to update the UI to reflect these changes. This involves adjusting how the UI behaves when the job is scheduled or running and the product stats aren't available yet. Hence, we'll need to make a call to `wc/gla/mc/product-statistics/` every 30 seconds to check if the stats calculation has been completed.

https://github.com/woocommerce/google-listings-and-ads/assets/2488994/e4aff324-1074-4513-b645-f80b0deb18e9

In summary, this PR introduces the following:

- It will continuously call `wc/gla/mc/product-statistics/` every 30 seconds until the `update_merchant_product_statuses` job is finished and the stats can be retrieved.

Example response when loading:
```
{
    "timestamp": 1710711097,
    "statistics": null,
    "scheduled_sync": 0,
    "loading": true,
    "error": null
}
```
Example response when it is completed:

```
{
    "timestamp": 1710711097,
    "statistics": {
      "active" : 10
      "disapproved" : 1
      "expiring" : 2 
      "not_synced" : 3
      "pending" : 4
    },
    "scheduled_sync": 0,
    "loading": false,
    "error": null
}
```

2. Once the fetching process is completed, it will invalidate the Product Feed Query, ensuring that the Product MC Statuses are updated.

1. Make a call to `GET wp-json/wc/gla/mc/product-statistics/refresh`.
2. Navigate to `wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed` or refresh the product feed page.
3. Monitor the network requests and observe that the network request `wc/gla/mc/product-statistics` is fetched every 30 seconds until the loading status becomes false.
4. Verify that the product feed request is issued.
5. Confirm that the Overview Stats are loaded.
6. Ensure that the product feed displays the updated MC statuses.

### Additional details:

I was doubting whether it would be better to use an AppSpinner for each stats value or simply use the `SummaryListPlaceholder`. In the end, I decided to go with the placeholder because it maintains a more consistent look with what we currently have. However, I'm open to hearing other opinions.

AppSpinner:

https://github.com/woocommerce/google-listings-and-ads/assets/2488994/5c5b50c1-0e80-49e9-9771-1f000aa98621

PlaceHolder:

image

### Changelog entry

>


js/src/hooks/useMCProductStatistics.js
```diff
@@ -1,15 +1,60 @@
...
  */
 const useMCProductStatistics = () => {
-	return useAppSelectDispatch( 'getMCProductStatistics' );
+	const { second, callCount, startCountdown } = useCountdown();
+	const { invalidateResolutionForStoreSelector } = useAppDispatch();
+	const { data, hasFinishedResolution, invalidateResolution, ...rest } =
+		useAppSelectDispatch( 'getMCProductStatistics' );
+
+	const isLoading = hasFinishedResolution && data?.loading ? true : false;
+	const hasStats = hasFinishedResolution && ! data?.loading ? true : false;
```
ianlin: Maybe we could just use `isLoading` and remove `hasStats`? Or we could check if `data.statistics` had actual data.

js/src/hooks/useMCProductStatistics.js
```diff
@@ -1,15 +1,60 @@
...
+
+	useEffect( () => {
+		if ( isLoading && second === 0 ) {
+			startCountdown( 30 );
+		}
+
+		if ( isLoading && second === 0 && callCount > 0 ) {
+			invalidateResolution();
+		}
```
ianlin: We could combine this two conditions like:

```suggestion
		if ( isLoading && second === 0 ) {
			 startCountdown( 30 );
			 if ( callCount > 0 ) {
			   invalidateResolution();
			 }
		}
```

js/src/product-feed/product-statistics/status-box/sync-status.js
```diff
@@ -66,7 +67,14 @@ function getSyncResult( {
  * @return {JSX.Element} The status for the Product Sync
  */
 function SyncStatus() {
-	const { data } = useAppSelectDispatch( 'getMCProductStatistics' );
+	const { data, hasFinishedResolution } = useAppSelectDispatch(
+		'getMCProductStatistics'
+	);
+
+	if ( ! hasFinishedResolution || ( hasFinishedResolution && ! data ) ) {
```
ianlin: It could be simplified to:

```suggestion
	if ( ! hasFinishedResolution || ! data ) {
```

js/src/product-feed/product-statistics/index.js
```diff
@@ -44,10 +53,8 @@ const ProductStatistics = () => {
 				className="gla-product-statistics__summaries"
 				size={ null }
 			>
-				{ ! hasFinishedResolution && (
-					<SummaryListPlaceholder numberOfItems={ 5 } />
-				) }
-				{ hasFinishedResolution && data && (
+				{ isLoading && <SummaryListPlaceholder numberOfItems={ 5 } /> }
+				{ hasFinishedResolution && ! data?.loading && (
```
ianlin: We could probably define `hasStats` variable like above comment and check if there is actual data.

js/src/hooks/useMCProductStatistics.js
ianlin: Also probably might be helpful to add some comments, e.g. `Start counter when...` and `Clear the product statistic data...`

ianlin (commented):
Thanks for the change. It worked perfectly on my test environment, I only just needed to manually delay the job a little bit otherwise the request was too fast to complete. I left some comments that we could have some discussions on.

ianlin:
> I decided to go with the placeholder because it maintains a more consistent look with what we currently have. However, I'm open to hearing other opinions.

I think using AppSpinner would be a bit better IMO. As waiting for 30 seconds with the static summary placeholder would made me think something went wrong. But spinners spinning for 30 seconds might have the same feeling 😅.

Maybe either have something saying it is fetching the data, or reduce the interval to maybe 10 seconds might help?

js/src/hooks/useMCProductStatistics.js
jorgemd24: Thanks for the suggestion, I changed a now we are checking if `data.statistics` has data. Also, I changed the response from the API, so instead of an empty array, it returns null for the statistics. See here: https://github.com/woocommerce/google-listings-and-ads/pull/2318/commits/03ceb657f447220226c33bda30999b28a779dc89

js/src/hooks/useMCProductStatistics.js
jorgemd24: I made an update and also renamed the constant from `isLoading` to `isCalculatingStats`. I believe this change clarifies that the loading belongs to the calculation process rather than indicating that the HTTP request is in a "loading" state.

jorgemd24:
> I think using AppSpinner would be a bit better IMO. As waiting for 30 seconds with the static summary placeholder would made me think something went wrong. But spinners spinning for 30 seconds might have the same feeling

Yes, I have the same feeling, I switched it to AppSpinner and shortened the time to 15 seconds. I don't want to reduce the time too much because for a merchant with thousands of products, the jobs could take longer to complete all the calculations.

jorgemd24:
@ianlin thanks for your time reviewing this PR.

 I addressed all your suggestions and replied to your comments! Could you please take another look? Thanks!

js/src/hooks/useMCProductStatistics.js
```diff
@@ -1,15 +1,64 @@
...
+	const hasStats = hasFinishedResolution && data?.statistics ? true : false;
+
+	useEffect( () => {
+		// If the job is still processing the data, start the countdown.
+		if ( isCalculatingStats && second === 0 ) {
+			startCountdown( 15 );
+			// If the job is still processing the data, invalidate the resolution/refetch the data to get the latest status.
+			if ( callCount > 0 ) {
+				invalidateResolution();
+			}
```
ianlin: I feel like the comment would be more accurate if added something like `...the countdown has started...`.

ianlin (approved):
Thanks for the extra changes @jorgemd24. Using AppSpinner and reducing the countdown time is better regarding UX. I tested again and everything worked well, just left a non-blocker comment. LGTM.

PR #2320, opened 2024-03-19
[R-1.3] Fix onboarding flow skipping to step 2

joemcgill:
This fixes an issue where a the setup flow could load in step 2 if the Ads account is switched after initially setting up MC by checking that both Ads and MC are connected prior to moving to step 2.

This adds a new method to the AdsService to check if the Ads account is connected. It does so by making sure that there is an Ads ID and that the last incomplete step is either empty, or pending 'billing', which can optionally be skipped.

See #2270. Replaces #2316.

1. Connect both and Ads and and MC account in step one
2. Refresh the page and see that it will skip to step 2
3. Move back to step one and disconnect the Ads account
4. Refresh the page again and ensure that the setup is loading in step one

### DIFF
--- a/src/Ads/AdsService.php
+++ b/src/Ads/AdsService.php
@@ -62,4 +62,15 @@ class AdsService implements OptionsAwareInterface, Service {
 		$google_connected = boolval( $this->options->get( OptionsInterface::GOOGLE_CONNECTED, false ) );
 		return $google_connected && $this->is_setup_complete();
 	}
+
+	/**
+	 * Determine whether the Ads account is connected, even when pending billing.
+	 *
+	 * @return bool
+	 */
+	public function connected_account(): bool {
+		$id        = $this->options->get_ads_id();
+		$last_step = $this->account_state->last_incomplete_step();
+		return $id && ( $last_step === '' || $last_step === 'billing' );
+	}
 }
--- a/src/MerchantCenter/MerchantCenterService.php
+++ b/src/MerchantCenter/MerchantCenterService.php
@@ -3,6 +3,7 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\Ads\AdsService;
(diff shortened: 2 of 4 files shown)


ankitrox (approved):

joemcgill:
@mikkamp this is the PHP needed to fix the step process that could cause the setup flow to start on step 2 if the Google Ads account is disconnected after an MC account was already connected. See: https://github.com/woocommerce/google-listings-and-ads/pull/2270#issuecomment-2008467286.

src/Ads/AdsService.php
```diff
@@ -62,4 +62,15 @@ public function is_connected(): bool {
 		$google_connected = boolval( $this->options->get( OptionsInterface::GOOGLE_CONNECTED, false ) );
 		return $google_connected && $this->is_setup_complete();
 	}
+
+	/**
+	 * Determine whether the Ads account is connected, even when pending billing.
+	 *
+	 * @return bool
+	 */
+	public function connected_account(): bool {
```
mikkamp, line 71: I realize this matches the function name in `MerchantCenterService`, but all other public functions in AdsService are prefixed with `is_`. Should we make that service more consistent and call it `is_account_connected`?

mikkamp (approved):
Thanks for the changes. The code is working as expected, I'm only moving on to step 2 when the right conditions have been met.

I just left a small comment about a function name, but it's not a blocker so I'll go ahead and approve this PR.

src/Ads/AdsService.php
joemcgill: I thought about doing something like this, but wasn't sure about the history of the similarly named method in `MerchantCenterSerrvice`, so decided to stay consistent. Both service classes also have methods named `is_connected` which is a bit confusing, as I would expect both to serve similar purposes. I think leaving the two method names the same for now is probably easier to understand without doing a bigger refactor, but am open to suggestion.

I'm going to merge this into the base branch to make testing easier, but happy to continue the conversation there.

mikkamp: Yeah that's fine. As I said it's not a blocker, just something I noticed when I went through the code.

PR #2321, opened 2024-03-19
Expose OCID via REST

ankitrox:

Closes # .

Expose the ads account OCID via REST API endpoint in order to form the deeplink in FE components.

1. 
2. 
3. 

### Changelog entry

>


joemcgill:
Thanks @ankitrox. Functionally this looks good to me, but I'm not totally clear on why we need an additional endpoint for this. I think it might be better to avoid another request and instead include this value in either the `ads/connection` or `ads/account-status` endpoints that already exist. Would appreciate @asvinb's perspective.

PR #2324, opened 2024-03-19
Display error message if update_merchant_product_statuses  job throws an error

jorgemd24:

Now that the product statistics are calculated using an AS job, there's a possibility of throwing errors during the calculation process. If an error occurs, we need to notify the merchant and provide them with the option to retry the calculation.

This PR adds:
- The UI to inform the merchant about the error.
- Gives the option to re-try/refresh the stats if an error happens.
- Stops the UI of polling the results if the AS job failure rate is bigger than the threshold.

### Backend job completed with an error message:

1. Simulate an error, adding the following code:

```
  $this->mc_statuses['error']      = 'My error';
  $this->mc_statuses['statistics'] = null;
  return $this->mc_statuses;
```
After this line: https://github.com/woocommerce/google-listings-and-ads/blob/065ff6dd3ae43beb4eb5118a1d2236c230f66124/src/MerchantCenter/MerchantStatuses.php#L140

2. Go to `admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed` and check that it shows the error.

image

3. Delete the lines you inserted in step 1, open the Dev console (Netowork tab), then click "Click to retry."
4. Check that it sends a `GET wc/gla/mc/product-statistics/refresh` request.
5. The AppSpinner in the stats will be visible, and every 15 seconds, the UI will attempt to fetch the results. Once the calculation is finished, the stats should be displayed.

### AS Job - Failure rate is bigger than the threshold:

1. Add the following code `throw new \Exception( 'Error');` after this line:

https://github.com/woocommerce/google-listings-and-ads/blob/065ff6dd3ae43beb4eb5118a1d2236c230f66124/src/Jobs/UpdateMerchantProductStatuses.php#L85-L87

2. Make a request to `GET wc/gla/mc/product-statistics/refresh`
3. Go to WC -> Status -> Action Scheduler -> Search for `update_merchant_product_statuses`
4. Wait until it shows `The "update_merchant_product_statuses" job was stopped because its failure rate is above the allowed threshold.`
5. Go to `admin.php?page=wc-admin&path=%2Fgoogle%2Fproduct-feed`, and see that it shows the error:
image
6. Clicking "Click to retry" will simply display the error again if the failure rate remains high. For testing purposes, you can change the failure rate timeframe using the below filter and remove the Exception added in step 1.

```
add_filter( 'woocommerce_gla_job_failure_timeframe', function(){
	return 1;
} );
```
Maybe as an improvement, we could consider displaying a message like "Retry after X time." However, this would require additional work to calculate the remaining time based on the average timeframe and threshold and I am not sure about the frequency of this error neither.

7. See that a request is made to `GET wc/gla/mc/product-statistics/refresh` and the stats will be available after the jobs are completed. 

### Changelog entry

>

### DIFF
--- a/js/src/hooks/useMCProductStatistics.js
+++ b/js/src/hooks/useMCProductStatistics.js
@@ -8,6 +8,7 @@ import { useEffect } from '@wordpress/element';
  */
 import useAppSelectDispatch from './useAppSelectDispatch';
 import useCountdown from './useCountdown';
+import useApiFetchCallback from '.~/hooks/useApiFetchCallback';
 
 /**
  * Call `useAppSelectDispatch` with `"getMCProductStatistics"`.
@@ -21,6 +22,16 @@ const useMCProductStatistics = () => {
 		useAppSelectDispatch( 'getMCProductStatistics' );
 
 	const isLoading = hasFinishedResolution && data?.loading ? true : false;
+
+	const [ refreshProductStats ] = useApiFetchCallback( {
+		path: `/wc/gla/mc/product-statistics/refresh`,
+		method: 'GET',
+	} );
+
+	const refreshStats = async () => {
+		await refreshProductStats();
+		invalidateResolution( 'getMCProductStatistics', [] );
+	};
 
 	if ( isLoading && callCount === 0 ) {
 		startCountdown( 30 );
@@ -37,6 +48,7 @@ const useMCProductStatistics = () => {
 		data,
 		invalidateResolution,
 		hasFinishedResolution,
+		refreshStats,
 		...rest,
 	};
 };
--- a/js/src/hooks/useMCProductStatistics.js
+++ b/js/src/hooks/useMCProductStatistics.js
@@ -35,7 +35,7 @@ const useMCProductStatistics = () => {
 
 	const refreshStats = async () => {
 		await refreshProductStats();
-		invalidateResolution( 'getMCProductStatistics', [] );
+		invalidateResolution();
 	};
 
 	useEffect( () => {
--- a/js/src/hooks/useMCProductStatistics.js
+++ b/js/src/hooks/useMCProductStatistics.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { useEffect } from '@wordpress/element';
+import { useEffect, useCallback } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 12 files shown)


js/src/hooks/useMCProductStatistics.js
```diff
@@ -27,6 +28,16 @@ const useMCProductStatistics = () => {
...
 
+	const [ refreshProductStats ] = useApiFetchCallback( {
+		path: `/wc/gla/mc/product-statistics/refresh`,
+		method: 'GET',
+	} );
+
+	const refreshStats = async () => {
+		await refreshProductStats();
+		invalidateResolution();
+	};
```
ianlin: Maybe we could wrap this function by `useCallback`, as there are dependencies used in the function. But I'm not sure if it is a best practice tho.

js/src/product-feed/product-statistics/status-box/sync-product-statistics.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {string} props.error
+ * @return {JSX.Element} The status for the Product Sync
+ */
+function SyncProductStatistics( { refreshStats, error } ) {
+	return (
+		<Status
+			title={ __( 'Overview Stats:', 'google-listings-and-ads' ) }
+			icon={ <ErrorIcon size={ 24 } /> }
+			label={
+				<Button
```
ianlin: What about using [AppButton](https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/components/app-button/index.js) so we can also add the `eventName` to it for tracking.

ianlin (commented):
Thanks for handling the job error and displaying it on the UI. It worked well from my testing. I left two comments that we could have discussions on.

Also one thing that in the UI the error message is `The "update_merchant_product_statuses" job was stopped because its failure rate is above the allowed threshold.` I'm not sure whether the merchant needs to know the detail that much, especially the job name. But the message is coming from AS not sure if there's any way we could modify it.

js/src/hooks/useMCProductStatistics.js
jorgemd24: Thanks @ianlin for the suggestion, I believe it won't make much difference in terms of performance, but it won't hurt either, so I've added it here: https://github.com/woocommerce/google-listings-and-ads/pull/2324/commits/345d019019e423eb7df00d3487e1d77b9c74f137

jorgemd24:
Thanks @ianlin for your suggestions! 

> I'm not sure whether the merchant needs to know the detail that much, especially the job name. But the message is coming from AS not sure if there's any way we could modify it.

Yes, that's right, I have updated it here: https://github.com/woocommerce/google-listings-and-ads/pull/2324/commits/986c6d24d2177caeba2d902ab9970f7384479291

Could you take another look? Thanks!

ianlin (approved):
Thanks for the extra changes, LGTM. 👍

PR #2326, opened 2024-03-20
API PULL - Block syncing when notifications are enabled

puntope:

This PR disables Product and Coupon Sync when WC API is authorized. 

ℹ️  WC API is authorized  when the option `gla_wpcom_rest_api_status = approved`

Enabled WC API 

0. Set the option `gla_wpcom_rest_api_status` as "approved"
1. Go to the Connection Test page
2. Verify you see  at the end WC STAtus APi is approved.
4. Run Sync All Products --- You will get "Error submitting products to Google: Manual product sync will not run if the automatic data fetching is enabled. Please review your configuration in Google Listing and Ads settings."
5. Run Delete All Synced Products --- You will get "Error deleting products from Google: Manual product sync will not run if the automatic data fetching is enabled. Please review your configuration in Google Listing and Ads settings."
6. Run Clean up All Synced Products --- You will get "Error cleaning up products: Manual product sync will not run if the automatic data fetching is enabled. Please review your configuration in Google Listing and Ads settings."
7. For the coupons there is no test in the Connection page available. So we can only test that the jobs (delete_coupon and update_coupon) are not scheduled when the coupon is created, updated or deleted. But this was tested already in previous works.

Disabled WC API 
1. Remove the option `gla_wpcom_rest_api_status` or click on "Disconnect" in the Connection test page under "WPCOM REST API Status" section
4. Run Sync All Products successfully 
5. Run Delete All Synced Products successfully 
6. Run Clean up All Synced Products successfully 
7. For the coupons there is no test in the Connection page available. So we can only test that the jobs (delete_coupon and update_coupon) are scheduled when the coupon is created, updated or deleted. But this was tested already in previous works.


src/MerchantCenter/MerchantCenterService.php
```diff
@@ -103,7 +103,7 @@ public function is_google_connected(): bool {
 	 * @since 1.13.0
 	 * @return boolean
 	 */
-	public function is_ready_for_syncing(): bool {
+	public function is_ready(): bool {
```
jorgemd24: What do you think if we add a @since indicating that this function was renamed? Otherwise, it seems like `is_ready()` has been there since 1.13.0. However, I'm not sure if renaming the function to `is_ready` is necessary. To me, "syncing" implies that my site and the MC will be synchronized, regardless of the method used ("push" or "pull").

src/MerchantCenter/MerchantCenterService.php
```diff
@@ -115,13 +115,26 @@ public function is_ready_for_syncing(): bool {
...
 
+	/**
+	 * Whether we should run sync into MC. Only if:
+	 * - MC is ready for syncing {@see is_ready}
+	 * - Notifications Service is not enabled
+	 *
+	 * @return bool
+	 * @since x.x.x
+	 */
+	public function should_sync(): bool {
```
jorgemd24: Wouldn't it be better to say something like "should_notify"? To me, "sync" is ambiguous as I mentioned earlier; it could refer to either a push or pull method.

puntope:
Notice I didn't include testing steps for Sync Product. 

This is because of an issue in bookings https://github.com/woocommerce/woocommerce-bookings/issues/3784 

However, if bookings is not installed it should work perfectly. (Blocked when WPCOM API is authorized, working otherwise)

src/Product/SyncerHooks.php
```diff
@@ -294,6 +316,25 @@ protected function handle_delete_product( int $product_id ) {
 		}
 	}
 
+	/**
+	 * Send the notification for product deletion
```
jorgemd24: I'd phrase it as "schedule a job to send the product deletion notification" because it doesn't directly send the notification.

src/Jobs/AbstractProductSyncerJob.php
```diff
@@ -54,15 +54,6 @@ public function __construct(
...
 
-	/**
-	 * Get whether Merchant Center is connected and ready for syncing data.
-	 *
-	 * @return bool
-	 */
-	public function is_mc_ready_for_syncing(): bool {
-		return $this->merchant_center->is_ready_for_syncing();
-	}
-
```
jorgemd24, line 65: How about we leave this as it is and simply replace `$this->merchant_center->is_ready_for_syncing()` with something like `$this->merchant_center->is_ready_for_syncing() && ! $this->merchant_center->should_notify()`? We could apply the same change to the other jobs as well.

src/Product/SyncerHooks.php
```diff
@@ -120,59 +120,88 @@ public function __construct(
...
-		};
-
-		$pre_delete = function ( int $product_id ) {
-			$this->handle_pre_delete_product( $product_id );
-		};
-
-		$delete = function ( int $product_id ) {
-			$this->handle_delete_product( $product_id );
-		};
-
```
jorgemd24, line 146: What is the reason for moving these as methods instead of keeping them as anonymous functions? My understanding was that the idea behind using anonymous functions was to make it more difficult for other plugins to unhook these actions.

src/Product/ProductSyncer.php
```diff
@@ -357,10 +356,25 @@ protected function handle_delete_errors( array $invalid_products ) {
...
+				__METHOD__
+			);
+
+			throw new ProductSyncerException(
+				__(
+					'Manual product sync will not run if the automatic data fetching is enabled. Please review your configuration in Google Listing and Ads settings.',
+					'google-listings-and-ads'
+				)
+			);
+		}
```
jorgemd24: I wanted to test these errors, but I noticed that `validate_merchant_center_setup` is only triggered when the product syncers job is processing the items. So theoretically, we shouldn't encounter this error. However, I think it's a good idea to keep it just in case. While reviewing the code, I saw that for the notification part, we don't have anything similar to "validate_merchant_center_setup". We only check if the notification service is enabled, but not the MC setup. should we add something similar to the notification? Or am I missing something?

https://github.com/woocommerce/google-listings-and-ads/blob/1f3a18057ee27c08ae10261a1ab9b9adbf17a6df/src/Jobs/Notifications/AbstractNotificationJob.php#L88

src/Product/ProductSyncer.php
```diff
@@ -357,10 +356,25 @@ protected function handle_delete_errors( array $invalid_products ) {
...
 	 */
 	protected function validate_merchant_center_setup(): void {
-		if ( ! $this->merchant_center->is_ready_for_syncing() ) {
+		if ( ! $this->merchant_center->is_ready() ) {
 			do_action( 'woocommerce_gla_error', 'Cannot sync any products before setting up Google Merchant Center.', __METHOD__ );
 
 			throw new ProductSyncerException( __( 'Google Merchant Center has not been set up correctly. Please review your configuration.', 'google-listings-and-ads' ) );
 		}
+
+		if ( ! $this->merchant_center->should_sync() ) {
```
jorgemd24: Also what do you think if we add some tests for this?

jorgemd24 (commented):
Hi @puntope for working on this! It worked as expected however I left some comments about the code that we could discuss a bit more.

src/Jobs/AbstractProductSyncerJob.php
puntope: I'm towards keeping it like it's. That function seems a bit redundant and has attached a full interface for it.

src/Product/SyncerHooks.php
puntope: This is mainly for the testing.

For example for doing 

`$this->assertFalse( has_action( 'woocommerce_new_coupon', [ $this->syncer_hooks, 'update_by_id' ] ) );`
 

https://github.com/woocommerce/google-listings-and-ads/pull/2326/files#diff-afa7e4fae2bc017bdf473aacdcf84a1720a7f73a22e83979eda07d4696b6cd41R271

src/Product/ProductSyncer.php
puntope: The tests were already added in 

ProductSyncerTest - https://github.com/woocommerce/google-listings-and-ads/pull/2326/files#diff-e1a250ff800d7687a728767bcc4817c6ca40f2fa9e97d9846e4b967ff338ee1fR551

CouponSyncerTest - https://github.com/woocommerce/google-listings-and-ads/pull/2326/files#diff-19117240c30f979181c4332c987f76f2ab19a5fd4b460d3345c8e68d195a1116R149

Do you miss any case?

src/Product/ProductSyncer.php
puntope: > I wanted to test these errors, but I noticed that validate_merchant_center_setup is only triggered when the product syncers job is processing the items. So theoretically, we shouldn't encounter this error.

You can trigger it in ConnectionTest Page for example

src/MerchantCenter/MerchantCenterService.php
puntope: I changed it to should_push()

src/Product/ProductSyncer.php
puntope: >  We only check if the notification service is enabled, but not the MC setup. should we add something similar to the notification? Or am I missing something?

We can do so in a follow up. I added the item to the list.

src/MerchantCenter/MerchantCenterService.php
puntope: Adjusted (I rolled back the renaming)

src/Product/SyncerHooks.php
puntope: Adjusted

src/Jobs/AbstractProductSyncerJob.php
puntope: Besides that, all the is_mc_ready_for_syncing are doing the same in all the classes. So doesn't seem a very useful function.

Is there any reason you found for keeping it?

puntope:
Hi @jorgemd24 thanks for the review.

I applied the renaming suggestions and clarified the rest of the comments.

Can you take another look? 

Thx

package-lock.json
```diff
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "google-listings-and-ads",
-			"version": "2.6.2",
+			"version": "2.6.3",
```
jorgemd24, line 9: Not sure why the package-lock-json has been commited

tests/Unit/Product/ProductSyncerTest.php
```diff
@@ -548,6 +548,54 @@ protected function mock_google_service( $successful_products, $failed_products )
 			->willReturnCallback( $callback );
 	}
 
+	public function test_update_throws_exception_when_mc_is_blocked() {
+		$merchant_center = $this->createMock( MerchantCenterService::class );
+		$merchant_center->expects( $this->any() )
```
jorgemd24: If we want to test the exception when `should_push` returns `false`, then we should mock `is_ready_for_syncing` to return true; otherwise, `should_push` never gets called. Also, I think it's better to expect an exact number instead of any, as in this case the test would pass without the method being called.

src/Jobs/AbstractProductSyncerJob.php
jorgemd24: I don't have any specific reason; I was simply curious why we need to remove it now when we'll be removing all this code in the next stages anyway.

src/Product/ProductSyncer.php
jorgemd24: See my other comments about the tests.

jorgemd24 (approved):
Thanks @puntope for adjusting the naming, I think it's clearer now. I've left some comments about the tests, and I'm not sure why the `package-lock.json` is committed in this PR. Anyway, the PR worked as expected so I am approving this PR in advanced! Thanks!

PR #2327, opened 2024-03-21
Add E2E tests for the integration in Product Block Editor

eason9487:

- Upgrade `@playwright/test` to `^1.42.0` for having the new APIs:
   - https://playwright.dev/docs/api/class-page#page-unroute-all
   - https://playwright.dev/docs/api/class-locator#locator-press-sequentially
- Convert space indentations in `test-snippets.php` to tabs.
- Adjust the product data in `tests/e2e/config/default.json` to fit with WC v3 REST API and remove unused parts.
- Add new E2E API utils for creating variable and variation products.
- Add E2E tests

#### 📌 Checklist (@eason9487)

- [x] Revert 77022c71d358d87a942112c13f9c36b6dca25b33 before merging this PR

1. `npm install` to update `@playwright/test` package.
2. `npx playwright install chromium` to update the PlayWright browser as well.
3. `npm run dev` to build files
4. `npm run test:e2e` to see if the E2E test in headless mode can pass.
5. (Optional) `npm run test:e2e -- --ui` to inspect the E2E test in UI mode.
6. View the E2E test result on GitHub Actions: https://github.com/woocommerce/google-listings-and-ads/actions/runs/8376570038/job/22936559491?pr=2327

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - "release/**"
   pull_request:
+    branches:
+      - "release/**"
   workflow_dispatch:
     inputs:
       wp-rc-version:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - "release/**"
   pull_request:
-    branches:
-      - "release/**"
   workflow_dispatch:
     inputs:
       wp-rc-version:
--- a/tests/e2e/specs/product-editor/block-integration.test.js
+++ b/tests/e2e/specs/product-editor/block-integration.test.js
@@ -256,6 +256,38 @@ test.describe( 'Product Block Editor integration', () => {
 		await expect( panel.getByRole( 'spinbutton' ) ).toHaveCount( 1 );
 	} );
 
+	test( 'Channel visibility is disabled when hiding in product catalog', async () => {
+		await editorUtils.gotoAddProductPage();
+		await editorUtils.fillProductName();
+
+		const { selection, help } = editorUtils.getChannelVisibility();
+
+		await editorUtils.clickPluginTab();
+		await expect( selection ).toBeEnabled();
+		await expect( selection ).toHaveValue( 'sync-and-show' );
+		await expect( help ).toHaveCount( 0 );
+
+		await editorUtils.clickTab( 'Organization' );
+		await page.getByLabel( 'Hide in product catalog' ).setChecked( true );
+		await editorUtils.save();
+
+		await editorUtils.clickPluginTab();
+		await expect( selection ).toBeDisabled();
+		await expect( selection ).toHaveValue( 'dont-sync-and-show' );
+		await expect( help ).toBeVisible();
+		await expect( help ).toContainText(
+			'This product cannot be shown on any channel because it is hidden from your store catalog.'
+		);
+
+		await editorUtils.clickTab( 'Organization' );
+		await page.getByLabel( 'Hide in product catalog' ).setChecked( false );
+		await editorUtils.save();
+
+		await editorUtils.clickPluginTab();
+		await expect( selection ).toBeEnabled();
+		await expect( help ).toHaveCount( 0 );
+	} );
+
 	test.afterAll( async () => {
 		await editorUtils.toggleBlockFeature( page, false );
 		await api.clearOnboardedMerchant();
--- a/tests/e2e/specs/product-editor/block-integration.test.js
+++ b/tests/e2e/specs/product-editor/block-integration.test.js
@@ -640,6 +640,47 @@ test.describe( 'Product Block Editor integration', () => {
 		}
 	} );
 
+	test( 'Save product attributes to variation product', async () => {
+		await editorUtils.gotoEditVariableProductPage();
+		await editorUtils.gotoEditVariationProductPage();
+		await editorUtils.clickPluginTab();
+
+		/*
+		 * Since the testing of all attributes is already covered in the simple product test case above.
+		 * This case only tests a few key attributes to ensure the data saving works for variation product.
+		 */
+		const {
+			condition,
+			color,
+			multipack,
+			availabilityDate,
+			availabilityTime,
+		} = editorUtils.getAllProductAttributes();
+
+		const pairs = [
+			[ condition, 'new' ],
+			[ color, 'Cherry blossom' ],
+			[ multipack, '9999' ],
+			[ availabilityDate, '2024-02-29' ],
+			[ availabilityTime, '23:59' ],
+		];
+
+		for ( const [ attribute ] of pairs ) {
+			await expect( attribute ).toHaveValue( '' );
+		}
+
+		for ( const [ attribute, value ] of pairs ) {
+			await expect( attribute ).toHaveValue( '' );
+			await editorUtils.setAttributeValue( attribute, value );
+		}
+
+		await editorUtils.save();
+
+		for ( const [ attribute, value ] of pairs ) {
+			await expect( attribute ).toHaveValue( value );
+		}
+	} );
+
 	test.afterEach( async () => {
 		await page.unrouteAll();
 	} );
--- a/tests/e2e/specs/product-editor/block-integration.test.js
+++ b/tests/e2e/specs/product-editor/block-integration.test.js
@@ -288,6 +288,96 @@ test.describe( 'Product Block Editor integration', () => {
 		await expect( help ).toHaveCount( 0 );
 	} );
 
+	test( 'Change channel visibility and check its notice, status, and issues', async () => {
+		await editorUtils.gotoAddProductPage();
+		await editorUtils.fillProductName();
+		await editorUtils.clickPluginTab();
+
+		const issueTexts = [ 'Invalid price', 'Invalid GTIN' ];
+		const { selection, notice, status, issues } =
+			editorUtils.getChannelVisibility();
+
+		// Save another value and revert it to atomically re-fetch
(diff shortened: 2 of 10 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -5,8 +5,6 @@ on:
     branches:
       - "release/**"
   pull_request:
-    branches:
```
puntope: Does this mean we will run the e2e tests on each PR?

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+	 */
+	let page = null;
+	let editorUtils = null;
+
+	test.beforeAll( async ( { browser } ) => {
+		page = await browser.newPage();
+		editorUtils = getProductBlockEditorUtils( page );
+
+		await editorUtils.toggleBlockFeature( true );
+		await api.setOnboardedMerchant();
```
puntope, line 27: Why have this line here if we are clearing it after?

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+
+		await expect( link ).toBeVisible();
+		await link.click();
+
+		await expect( page.getByRole( 'main' ) ).toContainText(
+			/Reach millions of shoppers with product listings on Google/i
+		);
+
+		// Resume the plugin to onboarded status so that the next test can carry over.
+		await api.setOnboardedMerchant();
```
puntope, line 44: Based on the comment. Makes more sense for me to add this to the next step.

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+
+		await editorUtils.changeToStandardProduct();
+		await expect( pluginTab ).toHaveCount( 1 );
+
+		await editorUtils.changeToAffiliateProduct();
+		await expect( pluginTab ).toHaveCount( 0 );
+	} );
+
+	test( 'Check existence and availability of fields for simple product', async () => {
+		await editorUtils.gotoAddProductPage();
```
puntope, line 66: I guess at this point we are already in ProductPage?  Why not just to change the tab and save a page load?

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+
+	test( 'Prompt to Get Started when not yet finished onboarding', async () => {
+		await api.clearOnboardedMerchant();
+		await editorUtils.gotoAddProductPage();
+		await editorUtils.clickPluginTab();
+
+		const link = page.getByRole( 'link', { name: 'Get Started' } );
+
+		await expect( link ).toBeVisible();
+		await link.click();
```
puntope: I wonder if might be better option to just check the link href and save a page load (for reducing time test time)

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+
+		const pluginTab = editorUtils.getPluginTab();
+
+		await expect( pluginTab ).toHaveCount( 1 );
+
+		await editorUtils.changeToGroupedProduct();
+		await expect( pluginTab ).toHaveCount( 0 );
+
+		await editorUtils.changeToStandardProduct();
+		await expect( pluginTab ).toHaveCount( 1 );
```
puntope, line 59: This is being tested I guess in line 54 as the default product is standard. So we can omit this

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+			'This product cannot be shown on any channel because it is hidden from your store catalog.'
+		);
+
+		await editorUtils.clickTab( 'Organization' );
+		await page.getByLabel( 'Hide in product catalog' ).setChecked( false );
+		await editorUtils.save();
+
+		await editorUtils.clickPluginTab();
+		await expect( selection ).toBeEnabled();
+		await expect( help ).toHaveCount( 0 );
```
puntope, line 287: Why are we testing again this part? Maybe we can reduce execution time as 'Hide in product catalog' is false by default and is tested already.

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+	} );
+
+	test( 'Change channel visibility and check its notice, status, and issues', async () => {
+		await editorUtils.gotoAddProductPage();
+		await editorUtils.fillProductName();
```
puntope, line 292: I guess these 2 can be omitted and save a page load

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+		await refetchProductData();
+
+		await expect( notice ).toHaveCount( 0 );
+		await expect( status ).toHaveCount( 0 );
+		await expect( issues ).toHaveCount( 0 );
+	} );
+
+	test( 'Custom block: Product select with text field', async () => {
+		await editorUtils.gotoAddProductPage();
+		await editorUtils.fillProductName();
```
puntope, line 378: These 2 can be omitted and save a page load

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -0,0 +1,693 @@
...
+
+		await editorUtils.save();
+
+		for ( const [ attribute, value ] of pairs ) {
+			await expect( attribute ).toHaveValue( value );
+		}
+	} );
+
+	test.afterEach( async () => {
+		await page.unrouteAll();
```
puntope, line 684: Why we need this?

.github/workflows/e2e-tests.yml
puntope: Ok, I see you add it for remove it

puntope (commented):
Hi @eason9487 Thanks for your work in E2E test.

All is working fine. Not a strong blocker. However, I worry about the time for the tests to finish execution.

2-3mins

Screenshot 2024-03-24 at 12 31 15

I left some comments in an attempt to reduce unnecessary saves/loads. I will wait for your feedback before approve.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: Apart from the "Prompt to Get Started" case, all other cases need to be tested in the onboarded state.

If it isn't set here, then manual setup will be required when developing other cases, or the "Prompt to Get Started" case will have to be run first.

The main goal is to get the test cases as close to atomicity as possible so that test cases can be tested individually. This makes development and debugging easier and reduces false positives caused by side effects.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: It's related to https://github.com/woocommerce/google-listings-and-ads/pull/2327#discussion_r1537142428.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: This is also to make the test case as close as possible to atomicity.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: Yes, it can be reduced by about 1 ~ 2 seconds. Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2327/commits/caeefeeedff5bb47442cc9ffcb933be98935686b.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: This asserts that the tab can be changed from hidden to visible, which is not equivalent to being visible by default. Also, for the next one assert that the tab goes from visible to hidden.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: This asserts that the channel visibility status can be changed back from disabled to enabled.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: This is also to make the test case as close as possible to atomicity.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: This is also to make the test case as close as possible to atomicity.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: This is to removes all routes created with `page.route()` via the `editorUtils.mockChannelVisibility` calls in the "Change channel visibility and check its notice, status, and issues" case. In this way, the side effects between test cases can be reduced.

eason9487:
Hi @puntope, thanks for the review. Could you take another look?

> However, I worry about the time for the tests to finish execution.

The long execution time is mainly because the Product Block Editor page loading takes 6~10 seconds, which is rather uncontrollable from the plugin side.

Currently, each test has at least one atomicity that can be executed independently and has no side effects. Allowing the next case to continue where the finished states of the previous one would reduce some overall time but at the cost of making development and debugging more difficult. I'm leaning toward keeping test cases as atomic as possible.

puntope (approved):
Thanks for the clarifications @eason9487 

Sad the Product Editor loads causes those delays. I'm approving it since it doesn't seem to have any action on our side to mitigate it.

PR #2329, opened 2024-03-21
API PULL - Overview Stats - delete stale mc statuses

jorgemd24:

Since the product's MC status depends on the status in Google Merchant Center, if the product no longer exists in the MC, the local MC status will never be deleted. Hence, this PR will clear all the MC statuses, allowing us to start fresh when pulling the product MC statuses.

In the previous https://github.com/woocommerce/google-listings-and-ads/pull/2257, we refactor some code to use WC Crud classes instead of directly querying the database. However, in this scenario, I think deleting all the product's post meta `_wc_gla_mc_status` with a single query would be much more efficient than using the CRUD classes and loading all the products.

I increased the transient expiration time to 12 hours. Because currently, whenever a product is created, updated, or deleted, the transient cache is cleared because it calls the hooks (`woocommerce_gla_batch_updated_products` and `woocommerce_gla_batch_deleted_products`). Therefore, I believe it's unnecessary to remove the transient cache every hour if no changes to products have occurred.

https://github.com/woocommerce/google-listings-and-ads/blob/8ebe1ecaf902ea114ae426d2abd86aebc960b560/src/Event/ClearProductStatsCache.php#L39-L50

With the new approach of notifications, we could do something similar, once a product has been notified to Google we could clear the cache.

1. Call `wc/gla/mc/product-statistics/refresh`
2. Go to WC -> Status -> Action Scheduler and see that the `gla/jobs/update_merchant_product_statuses/process_item` is scheduled.
3. Go to `page=wc-admin&path=%2Fgoogle%2Fproduct-feed` and wait until the stats are fetched.
4. If you want to check whether the postmeta has been deleted and then reinserted, you can check if the postmeta IDs in the database have changed.

### Changelog entry

>

### DIFF
--- a/src/DB/ProductMetaQueryHelper.php
+++ b/src/DB/ProductMetaQueryHelper.php
@@ -62,6 +62,8 @@ class ProductMetaQueryHelper implements Service {
 	/**
 	 * Delete all values for a given meta_key.
 	 *
+	 * @since x.x.x
+	 *
 	 * @param string $meta_key The meta key to delete.
 	 *
 	 * @throws InvalidMeta If the meta key isn't valid.
--- a/src/DB/ProductMetaQueryHelper.php
+++ b/src/DB/ProductMetaQueryHelper.php
@@ -60,6 +60,21 @@ class ProductMetaQueryHelper implements Service {
 	}
 
 	/**
+	 * Delete all values for a given meta_key.
+	 *
+	 * @param string $meta_key The meta key to delete.
+	 *
+	 * @throws InvalidMeta If the meta key isn't valid.
+	 */
+	public function delete_all_values( string $meta_key ) {
+		self::validate_meta_key( $meta_key );
+		$meta_key = $this->prefix_meta_key( $meta_key );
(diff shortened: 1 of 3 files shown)


puntope:
> With the new approach of notifications, we could do something similar, once a product has been notified to Google we could clear the cache.

I added this to the list of TODO in https://github.com/woocommerce/google-listings-and-ads/issues/2146

src/DB/ProductMetaQueryHelper.php
```diff
@@ -59,6 +59,21 @@ public function get_all_values( string $meta_key ): array {
...
 	}
 
+	/**
+	 * Delete all values for a given meta_key.
+	 *
+	 * @param string $meta_key The meta key to delete.
+	 *
+	 * @throws InvalidMeta If the meta key isn't valid.
+	 */
+	public function delete_all_values( string $meta_key ) {
```
puntope, line 71: `@since` tag is missed

puntope (approved):
Hi @jorgemd24  thanks for working on this.

✅  LGTM the issues are refreshed and loaded as always. 

I left a minor comment about the `since`tag. But approved in advance

PR #2338, opened 2024-03-27
API Pull - Update metadata for Coupons and Product after being notified.

jorgemd24:

Once the products or coupons are notified, we can still update some metadata as we used to when we were pushing the data. This is particularly useful for coupons, as it allows us to inform the user that the coupon has been sent/notified to Google.

This PR adds the following:

- Once a coupon has been notified, it will update the sync status and sync timestamp.
- A similar approach will be applied to the products.
- I've created a new Interface called `HelperNotificationInterface` for the `ProductHelper` and `CouponHelper`. This is because the `AbstractItemNotificationJob` expects those classes to have certain methods available. By doing this, we can ensure that those methods are indeed available. The catch is that I had to remove some type hints because PHP versions earlier than 8 don't allow multiple types.

### Coupons
1. Create a new coupon and set the visibility to Show Coupon on Google.
2. Go to WC -> Status -> Action Scheduler -> search for: `gla/jobs/notifications/coupons/process_item`
3. After the job is finished and has been successfully notified, check the postmeta of the coupon you created. It should have `_wc_gla_sync_status` = `synced` and `_wc_gla_synced_at` set.

image

### Products
1. Create a new product and set the visibility to Sync and Show.
2. Go to WC -> Status -> Action Scheduler -> search for: `gla/jobs/notifications/products/process_item`
3. After the job is finished and has been successfully notified, check the postmeta of the product you created. It should have `_wc_gla_sync_status` = `synced` and `_wc_gla_synced_at` set.

### Changelog entry

>

### DIFF
--- a/src/Coupon/CouponHelper.php
+++ b/src/Coupon/CouponHelper.php
@@ -10,6 +10,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\ChannelVisibility;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\NotificationStatus;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\SyncStatus;
+use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications\HelperNotificationInterface;
 use WC_Coupon;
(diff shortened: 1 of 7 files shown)


ianlin (approved):
Thanks for the PR. I can see both coupon and product's metadata were updated. LGTM.

PR #2342, opened 2024-03-28
Add target audiences and shipping times in the settings/general endpoint. 

jorgemd24:

As part of the new mechanism for syncing merchant data with Google Merchant Center, we need to make some GLA Options accessible, such as the target audience and shipping times.

This PR adds the target audience and shipping time to the response of `wp-json/wc/v3/settings/general`

Initially, my idea was to use the filter `woocommerce_settings- . $group_id` to add a new group of settings. However, this filter can only be used for Options (which should be fine for the target audience but not for the shipping times). Additionally, the filter `woocommerce_settings-` doesn't provide the Request Object param so I decided to use the filter `rest_request_after_callbacks` which makes it easier to verify if the request is made from the proxied endpoints.

https://github.com/woocommerce/woocommerce/blob/af03815134385c72feb7a70abc597eca57442820/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-setting-options-controller.php#L79

1. Get [WC REST API credentials](https://woocommerce.github.io/woocommerce-rest-api-docs/#rest-api-keys), or alternatively, you can use your sandbox and authorize your WPCOM App in WPCOM.
2. Make a request to `GET /wp-json/wc/v3/settings/general?gla_syncable=1`. If using the WPCOM proxy, use `wpcom/v2/sites/YOUR_BLOG_ID/wc/v3/settings/general`.
3. Verify that the response contains two new fields: `gla_target_audience` and `gla_shipping_times`.
4. Check that if the parameter gla_syncable=1 is not included in the URL, for example: `GET /wp-json/wc/v3/settings/general`, it will not add those two fields. Note: If you make the request from WPCOM, it will always add the query parameter.

### Changelog entry

>

### DIFF
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -20,6 +20,8 @@ defined( 'ABSPATH' ) || exit;
  *
  * Initializes the hooks to filter the data sent to the WPCOM proxy depending on the query parameter gla_syncable.
  *
+ * @since x.x.x
+ *
  * @package Automattic\WooCommerce\GoogleListingsAndAds\Integration
  */
 class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -3,9 +3,13 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Integration;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\ShippingTimeQuery;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Registerable;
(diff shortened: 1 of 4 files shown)


src/Integration/WPCOMProxy.php
```diff
@@ -107,13 +131,55 @@ protected function register_object_types_filter( string $object_type ): void {
 	}
 
 	/**
-	 * Whether the data should be filtered.
+	 * Register the callbacks.
+	 */
+	protected function register_callbacks() {
```
puntope, line 138: Missed `@since` param

puntope: Actually since all the class is new maybe we can add it at the top of the class

puntope (approved):
LGTM ✅  Thanks @jorgemd24 

When `gla_syncable=1`
Screenshot 2024-03-29 at 11 03 32


I left a comment regarding the since tag, but approved in advance.

PR #2350, opened 2024-04-04
Add snippet to bypass plugin dependency to fix e2e tests

### LINKED ISSUE #2348: e2e tests failing
### Describe the bug:
E2E tests are all failing. Noticed while preparing this weeks release.

See https://github.com/woocommerce/google-listings-and-ads/actions/runs/8551028938/job/23429160311?pr=2347#step:10:1

martynmjones:

Closes #2348

The WooCommerce plugin dependency listed in Google Listings & Ads is causing problems with the E2E environment setup because if WooCommerce is not activated first then GLA won't be activated at all.

This PR adds a snippet to bypass the dependency so that the environment can be setup.

Ref: pcShBQ-1yd-p2#comment-2990

### Changelog entry

> Dev - Add snippet to bypass WooCommerce dependency in E2E tests

### DIFF
--- a/tests/e2e/test-snippets/test-snippets.php
+++ b/tests/e2e/test-snippets/test-snippets.php
@@ -36,3 +36,18 @@ add_filter(
 		return $value_options;
 	}
 );
+
+/**
+ * Snippet to bypass the WooCommerce dependency in Google Listings & Ads because
+ * in wp-env WooCommerce is installed in the directory woocommerce-trunk-nightly
+ */
+add_action(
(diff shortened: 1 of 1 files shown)


beaulebens:
Rather than skipping dependency checks, is there any way to manipulate the order that things are installed in wp-env?

martynmjones:
> Rather than skipping dependency checks, is there any way to manipulate the order that things are installed in wp-env?

The snippet to bypass the dependency has been removed and the E2E environment config has been updated in #2362 to fix the problem rather than work around it.

PR #2354, opened 2024-04-05
API PULL - Add offer id product deletion notification

puntope:

This PR adds offer_id when sending a `product.delete` notification.

Slack conversation: p1712250015177139-slack-C02BB3F30TG

### Screenshots:

Screenshot 2024-04-08 at 10 38 36

1. Checkout this PR and enable Notifications service. `add_filter( 'woocommerce_gla_notifications_enabled', '__return_true' );`
2. Complete onboarding and enter this URL for authorizing Notifications `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings&google_wpcom_app_status=approved`
3. Create or Update a product. Run the `gla/jobs/notifications/products/process_item` job 
4. Delete the product. Run the `gla/jobs/notifications/products/process_item` job.
5. Go to WooCommerce - Status - Logs
6. See the entry in the log with `product.delete` topic and the Offer id "gla_{product_id}".. 
7. Apply a filter to change the Offer ID: 

```php
add_filter(
			'woocommerce_gla_get_google_product_offer_id',
			function ( $value, $product_id ) {
				return "custom_{$product_id}";
			},
			10,
			2
		);
```

8. Repeat steps 3,4 and 5
9. See the entry in the log with `product.delete` topic and the Offer id "custom_{product_id}".

### DIFF
--- a/src/Jobs/Notifications/AbstractItemNotificationJob.php
+++ b/src/Jobs/Notifications/AbstractItemNotificationJob.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\NotificationStatus;
-use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications\HelperNotificationInterface;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -34,8 +33,9 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 
 		$item  = $args['item_id'];
 		$topic = $args['topic'];
+		$data  = $args['data'] ?? [];
 
-		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item ) ) {
+		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item, $data ) ) {
 			$this->set_status( $item, $this->get_after_notification_status( $topic ) );
 			$this->handle_notified( $topic, $item );
 		}
@@ -59,9 +59,9 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 	 * @return string
 	 */
 	protected function get_after_notification_status( string $topic ): string {
-		if ( str_contains( $topic, '.create' ) ) {
+		if ( $this->is_create_topic( $topic ) ) {
 			return NotificationStatus::NOTIFICATION_CREATED;
-		} elseif ( str_contains( $topic, '.delete' ) ) {
+		} elseif ( $this->is_delete_topic( $topic ) ) {
 			return NotificationStatus::NOTIFICATION_DELETED;
 		} else {
 			return NotificationStatus::NOTIFICATION_UPDATED;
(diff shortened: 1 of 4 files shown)


src/Jobs/Notifications/AbstractItemNotificationJob.php
```diff
@@ -96,15 +96,36 @@ protected function can_process( int $item_id, string $topic ): bool {
...
+	/**
+	 * If a topic is a create topic
+	 *
+	 * @param string $topic The topic to check
+	 *
+	 * @return bool
+	 */
+	protected function is_create_topic( $topic ) {
+		return str_contains( $topic, '.create' );
+	}
```
jorgemd24: Missing argument types and return types

src/Jobs/Notifications/ProductNotificationJob.php
```diff
@@ -43,6 +46,19 @@ public function __construct(
...
 	}
 
+	/**
+	 * Override Product Notification adding Offer ID for deletions.
+	 *
+	 * @param array $args Arguments with the item id and the topic.
+	 */
+	protected function process_items( $args ): void {
+		if ( isset( $args['topic'] ) && isset( $args['item_id'] ) && $this->is_delete_topic( $args['topic'] ) ) {
+			$args['data'] = [ 'offer_id' => WCProductAdapter::get_google_product_offer_id( $this->get_slug(), $args['item_id'] ) ];
```
jorgemd24, line 58: If we use `WCProductAdapter::get_google_product_offer_id`, it would mean that we'll return the offer ID `gla_product_id` even for products not previously pushed by the GL&A Plugin. Wouldn't it be preferable to get the offer ID from the postmeta `_wc_gla_google_ids` instead? If that postmeta exists, it indicates it was pushed by us; if not, it suggests the product was pulled by Google and `_wc_gla_google_ids` wouldn't exist.

jorgemd24 (commented):
Thanks @puntope for working on this. I left a concern about using `WCProductAdapter::get_google_product_offer_id`. 

The code looks good and I managed to see the `offerID`.  Just as a side note, I'm getting the error again "The caller does not have permission" again, but I suspect that's on Google's end.

src/Jobs/Notifications/ProductNotificationJob.php
puntope: The reason I used `WCProductAdapter::get_google_product_offer_id` is because when a product is deleted is there a case that the delete Notification runs after the product is completely deleted form the database. So we will lost access to those metas and the product itself as well. 

I would go for sending the OfferID and rely on Google to check if the product actually exists.

src/Jobs/Notifications/AbstractItemNotificationJob.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2354/commits/cfe9ecea0a39c00e0db98ce5ac23ea85623a0a43

puntope:
Hi @jorgemd24 I addressed your comments. This is ready for a new round.

src/Jobs/Notifications/ProductNotificationJob.php
jorgemd24: Thanks for the clarification. We could add a comment there to explain that the `offer_id` might match the real offer ID or not, depending on whether the product has been synced by us or not. Also, I think it would be helpful to discuss this with Google so they understand that the offer ID isn't always reliable. Or, we could consider renaming it to `gla_offer_id`.

jorgemd24 (approved):
Thanks @puntope  for the adjustments. I left a comment regards the `offer_id`, I think it would be helpful if we add a note so it is clear that the `offer_id`  is just a calculated/virtual id instead of the real offer id.

jorgemd24:
@puntope , I've just realized that we'll need to update the proxy as well so we can forward the `offer_id` to Google. Currently, we only send the `item_id`

PR #2356, opened 2024-04-09
API PULL - Prevent notification errors for deleted products

puntope:

When we send notifications we take as granted that the product or coupon are available in DB when handling its metas. This is not always true as the Job and some intermediate functions can run after the Product or Coupon is permanently deleted. 

When this happens, the notifications meta handler or some other functions will throw some exceptions. 

This PR fixes that ensuring that the item is available. Introducing the next changes:

- Check the item is available on Update and Create Topics. Boil in case it isn't. 
- Run the delete Notification right away (without AS job) when the product doesn't exist.
- Consider NON-existent products as already Notified products. This could trigger (very rarely, I wasn't able to reproduce it yet) unnecessarily DELETE notifications when we permanently delete a product that was never notified. But I think even if this may happen it would be better than missing a needed notification. 

1. Follow the steps in https://github.com/woocommerce/google-listings-and-ads/wiki/API-PULL-Feature-(Beta)#enabling-the-feature and https://github.com/woocommerce/google-listings-and-ads/wiki/API-PULL-Feature-(Beta)#enabling-the-feature for enabling and granting access to Notifications feature.
2. Create a Product - See the `gla/jobs/notifications/products/process_item` job is scheduled with CREATE topic. 
3. Run it. See the Notification happens in the Logs
4. Update the product - See the `gla/jobs/notifications/products/process_item` job is scheduled with UPDATE topic.
5. Run it. See the Notification happens in the Logs
6. Trash the product - See the `gla/jobs/notifications/products/process_item` job is scheduled with DELETE topic.
7. Run it. See the Notification happens in the Logs
8. Create a Product - See the `gla/jobs/notifications/products/process_item` job is scheduled with CREATE topic. 
9. Don't run it.
10. Delete the product via `wp_delete_post( {post_id}, true );`
11. Run the job from step 8. See the Notification DOESNT happen in the Logs. NO exception happens.
12. Repeat steps 2-4. Don't run the UPDATE topic Job
13. Delete the product via `wp_delete_post( {post_id}, true );`
14. Run the update job from step. See the Notification DOESNT happen in the Logs. NO exception happens.
15. Repeat steps 2-3 and 6. Don't run the DELETE topic Job
16. Delete the product via `wp_delete_post( {post_id}, true );`
17. Now run the job. See the Notification YES happens in the Logs. NO exception happens.
18. Repeat all the steps for Coupons. 

### Additional details

I considered as well having some kind of table/options to keep the Notification metas so we don't depend anymore on the item itself. But that would require a big re-structure of the Notifications so this may be discussed for phase 2.

### DIFF
--- a/src/Coupon/CouponHelper.php
+++ b/src/Coupon/CouponHelper.php
@@ -378,7 +378,7 @@ class CouponHelper implements Service, HelperNotificationInterface {
 		try {
 			$coupon = $this->get_wc_coupon( $coupon_id );
 		} catch ( InvalidValue $e ) {
-			return true; // Sent true for forcing delete notification as the product doesn't exist anymore.
+			return true; // Sent true for forcing delete notification as the coupon doesn't exist anymore.
 		}
 
 		$valid_has_notified_creation_statuses = [
--- a/src/Coupon/CouponHelper.php
+++ b/src/Coupon/CouponHelper.php
@@ -341,11 +341,17 @@ class CouponHelper implements Service, HelperNotificationInterface {
 	 * Indicates if a coupon is ready for sending Notifications.
 	 * A coupon is ready to send notifications if its sync ready and the post status is publish.
 	 *
-	 * @param WC_Coupon $coupon
+	 * @param int $coupon_id
 	 *
 	 * @return bool
 	 */
-	public function is_ready_to_notify( WC_Coupon $coupon ): bool {
+	public function is_ready_to_notify( int $coupon_id ): bool {
+		try {
+			$coupon = $this->get_wc_coupon( $coupon_id );
+		} catch ( InvalidValue $e ) {
+			return false;
+		}
+
 		$is_ready = $this->is_sync_ready( $coupon ) && $coupon->get_status() === 'publish';
 
 		/**
@@ -364,11 +370,17 @@ class CouponHelper implements Service, HelperNotificationInterface {
 	 * Indicates if a coupon was already notified about its creation.
 	 * Notice we consider synced coupons in MC as notified for creation.
 	 *
-	 * @param WC_Coupon $coupon
+	 * @param int $coupon_id
 	 *
 	 * @return bool
 	 */
-	public function has_notified_creation( WC_Coupon $coupon ): bool {
+	public function has_notified_creation( int $coupon_id ): bool {
+		try {
+			$coupon = $this->get_wc_coupon( $coupon_id );
+		} catch ( InvalidValue $e ) {
+			return true; // Sent true for forcing delete notification as the product doesn't exist anymore.
+		}
+
 		$valid_has_notified_creation_statuses = [
 			NotificationStatus::NOTIFICATION_CREATED,
 			NotificationStatus::NOTIFICATION_UPDATED,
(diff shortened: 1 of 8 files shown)


src/Coupon/CouponHelper.php
```diff
@@ -364,11 +370,17 @@ public function is_ready_to_notify( WC_Coupon $coupon ): bool {
...
+	 * @param int $coupon_id
 	 *
 	 * @return bool
 	 */
-	public function has_notified_creation( WC_Coupon $coupon ): bool {
+	public function has_notified_creation( int $coupon_id ): bool {
+		try {
+			$coupon = $this->get_wc_coupon( $coupon_id );
+		} catch ( InvalidValue $e ) {
+			return true; // Sent true for forcing delete notification as the product doesn't exist anymore.
```
ianlin: Should be `coupon` in the comment.

src/Coupon/SyncerHooks.php
```diff
@@ -299,11 +299,17 @@ protected function handle_delete_coupon( int $coupon_id ) {
 	protected function maybe_send_delete_notification( int $coupon_id ): void {
 		$coupon = $this->wc->maybe_get_coupon( $coupon_id );
 
-		if ( $coupon instanceof WC_Coupon && $this->coupon_helper->should_trigger_delete_notification( $coupon ) ) {
+		if ( is_null( $coupon ) ) {
+			// In case product is not anymore in DB we send the notification directly.
```
ianlin: Should be `coupon` in the comment.

src/Coupon/SyncerHooks.php
```diff
@@ -299,11 +299,17 @@ protected function handle_delete_coupon( int $coupon_id ) {
 	protected function maybe_send_delete_notification( int $coupon_id ): void {
 		$coupon = $this->wc->maybe_get_coupon( $coupon_id );
 
-		if ( $coupon instanceof WC_Coupon && $this->coupon_helper->should_trigger_delete_notification( $coupon ) ) {
+		if ( is_null( $coupon ) ) {
```
ianlin, line 302: I guess since we are also listening to actions `trashed_post` or `deleted_post` for triggering delete coupons function, this line would cause the `coupon.delete` notification being triggered when a product is trashed.

https://github.com/woocommerce/google-listings-and-ads/blob/61fe02dddb7fd2f941dcada20c096499dd05d1a0/src/Coupon/SyncerHooks.php#L134-L135

ianlin (commented):
Thanks for handling the case when products/coupons no longer exist before sending notifications. I noticed that when I try to delete a product, the notification `coupon.delete` would be triggered as well:

> DEBUG Automattic\WooCommerce\GoogleListingsAndAds\API\WP\NotificationsService::notify Notification - Item ID: 143 - Topic: coupon.delete - Data []

src/Coupon/SyncerHooks.php
puntope: Thanks @ianlin. That is expected.

When the item is trashed is still available in the DB and it won't be null and we schedule the deletion Notification. In case the item is not anymore in DB we notify the deletion directly.

ianlin: Yeah, I understand in this PR it triggers the deletion notifications straightaway. But my question was a bit different: when I tried to delete a **_product_**, the **_coupon_** deletion notification would be triggered as well. It did not happen prior to this PR.

eason9487:
Closing this PR since the API Pull feature has been removed. Ref: #3382

PR #2361, opened 2024-04-10
R2: Updated Enhanced Conversion flow

### LINKED ISSUE #2364: Update Enhanced Conversions onboarding flow
## Overview
This is an update to [#2216](https://github.com/woocommerce/google-listings-and-ads/issues/2216). 

The original definition for that feature were based on processing within the Google Ads which would automatically enable Enhanced Conversions on Google Ads when EC data was added to the gtag config after User Data Terms & Conditions (T&C) were enabled. After testing that implementation, we've concluded that this is not a viable option in the short term.

Instead, we now need to prompt the user to enable Enhanced Conversions directly in their GAFE Settings to enable the feature AND ensure that they have enabled the feature in the GL&A extension to be confident that the user has opted in to sharing EC data from the store with Google.

### Success modal
On completion of the setup flow, the user is presented with a success modal with two pages. For new accounts, the second page will prompt the user to set up paid accounts, but if we have determined that the user has already set up paid accounts, then the next page will be a prompt to set up Enhanced Conversions.

Image

If the user has not yet accepted Customer Data Terms & Conditions (T&C) in their Ads account (determined with a request to the Google Ads API) we prompt the user to Accept Terms & Conditions.

Image

When the user clicks the CTA, we leave the modal active and open a new window that links the user to the settings panel for accepting T&C

Image

When the user returns to the original tab after updating this setting, we update the modal to ask them to confirm that they want to enable Enhanced Conversions. This is the same screen that will be shown to the user if we had already determined that they had accepted T&C prior to showing this prompt—skipping the part of the flow described above that links them to the GAFE settings.

Image

### Settings panel
If the user has confirmed enabling EC in the success modal above, they will now see a new panel at the bottom of the settings tab that shows that EC has been enabled, and allow them to manually disable the feature. If they had accepted T&C but closed the modal without enabling, then this will show in the disabled state.

Image

If the user has closed the modal without accepting T&C, then the setting will show a button that will open a tab to the same GAFE setting used during the onboarding prompt.

Image

Once clicked, until we can confirm the user has accepted T&C in the GAFE settings, we display the following instructions to remind the user that they still need to complete accepting the terms and conditions.

Image

Once we have confirmed that T&C have been accepted, the setting will show the enable/disable toggle.

### Success modal

The success modal still begins with the same first page. This set of requirements only addresses the updates to the second page when we’re showing the prompt for enabling Enhanced Conversions.

#### Case 1: If Terms & Conditions are not accepted

When Terms & Conditions are not yet accepted for the connected Ads account, we will display a modal with instructions for how to set up Enhanced Conversions, which includes expectations of what steps they will need to take in the GAFE.

**Proposed copy**

* Title: “Optimize your conversion tracking with Enhanced Conversions”
* Body1: “Improve conversion tracking accuracy to gain deeper insights and improve campaign performance.”
* Body2: 
  - Click Enable Enhanced Conversions below to go to Google Ads Settings.
  - Turn on Enhanced Conversions, accept Terms, and then select “Google Tag” as your setup method. Click Save.
* Body3: For more information, feel free to consult our help center article (link to https://support.google.com/google-ads/answer/9888656)

**Proposed CTA**

* [Close] remove, if possible.
* [Enable Enhanced Conversions]
Include icon to designate that this will go to another site
deep link to: https://ads.google.com/aw/conversions/customersettings?ocid=12345678&eppn=enhancedconversionsaccountlevelsettings

When the user clicks the CTA to confirm they want to enable enhanced conversions, we deep link them to the settings panel in GAFE so they can confirm the Enhanced conversions setting is enabled for the Ads account.

When they return to the open modal, if they have accepted T&C, we will show a success message (described below). If they have not accepted T&C, we remain on this page unless they manually close the modal.

#### Case 2: If Terms & Conditions were previously accepted
When this modal is reached but the T&C have already been accepted (e.g., in the case of connecting a pre-existing account). We want to confirm that the user wants to enable EC on this Woo store and show the following modal content...

**Proposed Copy**
Title: “Your Enhanced Conversions are almost ready” 
Body1: “We noticed that you have already accepted Enhanced Conversions Customer Data Terms. In order to complete Enhanced Conversions setup on your store, click Confirm below”
Body2: “If you haven’t done so already please ensure that Enhanced Conversions is enabled in Google Ads Settings (add hyperlink), and that the setup method chosen is Google Tag”

**Proposed CTA**
* [Close] remove, if possible.
* [Confirm] – Enables the EC setting in the store and displays the success message

Once the user clicks the “Confirm” button, they will be presented with the success message described below (i.e., the same message shown to a user who successfully enables T&C in Case 1 above).

#### Success screen
After either case above, when we've been able to confirm that the T&C have been accepted, the modal is updated to show a message confirming that EC has been enabled on their site.

**Proposed copy**

* Title: “We’ve successfully set up Enhanced Conversions for your store!”
* Body1: “We updated your conversion tracking tags. You can manage this feature from your store's [marketing settings](/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings”)."

**Proposed CTA**
* [Close] Completes the flow and closes the flow.

### Settings panel

If we have confirmed that user has enabled Enhanced Conversions during the setup modal, and we can determine that they have accepted T&C, we will display a settings panel with a toggle showing the feature as enabled. If we determine that the user has accepted T&C but they have NOT yet confirmed that they want to enable Enhanced Conversions for their store, then we show the same toggle, but with EC disabled.

In either case, we include instructions that the user also needs to make sure EC is enabled in their Ads account for this feature to work, with a deep link to the Google Ads Settings page. This is necessary because we don’t currently have a way to see if that setting has been enabled on the Ads account via the Ads API.

When we determine that the user has not confirmed Terms & Conditions in their Ads account, we can also infer that they have not enabled Enhanced Conversions, so we show them a settings panel that instructs them to visit their Ads account to enable Enhanced Conversions and accept T&C. Clicking this CTA will open a new window that deep links to the same settings panel in the GAFE that we open from the success modal.

**Proposed Copy**
* Title: “Enhanced Conversion Tracking”
* Body1: “Improve your conversion tracking accuracy and unlock more powerful bidding.This feature works alongside your existing conversion tags, sending secure, privacy-friendly conversion data from your website to Google”
* Body2: “To properly measure your enhanced conversions, please ensure Enhanced Conversions is enabled in Google Ads Settings (add hyperlink), and that the setup method chosen is Google Tag”

**Proposed CTA**
* Option 1: [Accept Terms & Conditions]
* Option 2: [Enable] / [Disable] toggle

asvinb:

Closes #2364 .

1. 
2. 
3. 

### Changelog entry

>


js/src/hooks/useAutoCheckEnhancedConversionTOS.js
```diff
@@ -31,10 +31,7 @@ const useAutoCheckEnhancedConversionTOS = () => {
 
 	useWindowFocusCallbackIntervalEffect( checkStatus, 30 );
 
-	return {
-		startEnhancedConversionTOSPolling,
-		stopEnhancedConversionTOSPolling,
-	};
+	return polling;
```
joemcgill: We're no longer using any return value from this hook, so this could be removed. Alternately, we could return an object containing the polling status and the return value of `acceptedCustomerDataTerms()` as a convenience. So components wouldn't need to use this alongside the `acceptedCustomerDataTerms()` hook, ex:

```suggestion
	return {
		isPolling,
		acceptedCustomerDataTerms,
		hasFinishedResolution
	};
```

js/src/components/enhanced-conversion-tracking-settings/enable.js
```diff
@@ -0,0 +1,27 @@
...
+ * Internal dependencies
+ */
+import CTA from '.~/product-feed/submission-success-guide/enhanced-conversions/cta';
+
+const Enable = () => {
+	return (
+		<Fragment>
+			<p>
+				{ __(
+					'In order to enable enhanced conversions, you will need to visit the settings in your Ads account. You will be prompted to accept terms & conditions.',
```
joemcgill, line 17: Awaiting confirmation, but just flagging that this copy is likely to change to include instructions for selecting "Google Tag," similar to the Toggle component..

js/src/product-feed/submission-success-guide/enhanced-conversions/cta.js
```diff
@@ -0,0 +1,52 @@
+/**
```
joemcgill: I'd suggest that this component would be better located with the settings folder since the success modal may go away once promoting EC is no longer a priority, but the settings panel will remain.

js/src/product-feed/submission-success-guide/enhanced-conversions/cta.js
```diff
@@ -0,0 +1,52 @@
...
+import useAcceptedCustomerDataTerms from '.~/hooks/useAcceptedCustomerDataTerms';
+import EnableButton from './enable-button';
+import ConfirmButton from './confirm-button';
+import useAutoCheckEnhancedConversionTOS from '.~/hooks/useAutoCheckEnhancedConversionTOS';
+
+const CTA = ( { onConfirm = noop } ) => {
+	// We could have used the return value from useAutoCheckEnhancedConversionTOS to know
+	// if there's polling in progress. However from a UI point of view, it'll be confusing
+	// for the user when refreshing the page to see a loading spinner while polling is in progress.
+	// Hence we are showing the spinner only when the user clicks on the Enable button.
```
joemcgill: Given that we want to be able to optionally enable polling, I think it might be clearer if we moved all of this state into the `useAutoCheckEnhancedConversionTOS()` hook, and return a function that allows this consuming component to enable polling when needed. We would probably need to pass an initial polling state to the hook.

js/src/product-feed/submission-success-guide/enhanced-conversions/index.js
```diff
@@ -0,0 +1,181 @@
...
+		'Optimize your conversion tracking with Enhanced Conversions',
+		'google-listings-and-ads'
+	);
+
+	if ( acceptedCustomerDataTerms ) {
+		title = __(
+			'Your Enhanced Conversions are almost ready',
+			'google-listings-and-ads'
+		);
+	}
```
joemcgill: This could be written as a const:

```suggestion
	const title = acceptedCustomerDataTerms
		? __(
				'Your Enhanced Conversions are almost ready',
				'google-listings-and-ads'
		  )
		: __(
				'Optimize your conversion tracking with Enhanced Conversions',
				'google-listings-and-ads'
		  );
```

js/src/product-feed/submission-success-guide/enhanced-conversions/index.js
```diff
@@ -0,0 +1,181 @@
...
+
+	const getPageContentBody = () => {
+		if ( ! hasFinishedResolution ) {
+			return <AppSpinner />;
+		}
+
+		if ( ! acceptedCustomerDataTerms ) {
```
joemcgill: The copy for this has been updated so that the panel is not as long. See: https://docs.google.com/document/d/1U1Wiymz51aRsUZQKEHit-y9o1kiSdkEcBV-8e3BXJiM/edit#heading=h.27ah3dar1ze0.

js/src/product-feed/submission-success-guide/enhanced-conversions/confirm-button.js
```diff
@@ -0,0 +1,32 @@
...
+ * Internal dependencies
+ */
+import { ENHANCED_ADS_CONVERSION_STATUS } from '.~/constants';
+import { useAppDispatch } from '.~/data';
+import AppButton from '.~/components/app-button';
+
+const ConfirmButton = ( { onConfirm = noop } ) => {
+	const { updateEnhancedAdsConversionStatus } = useAppDispatch();
+
+	const handleConfirm = () => {
```
joemcgill: We're going to need to move this out to the CTA component and pass it in as part of the `onConfirm` callback that is passed to this component because the requirements have changed slightly, and we want to automatically confirm that this is enabled once the user has clicked from the T&C modal and returned, without needing to take a second step of also confirming again. This should also work in the settings panel if the user clicks the `EnableButton` and then returns to the settings page.

joemcgill (changes requested):
@asvinb this is already starting off great. I'm getting the requirements finalized and the biggest change is that we don't want to show the second "Enable" screen from the modal once the user returns from enabling EC in the GAFE. I've left some thoughts inline.

js/src/hooks/useAutoCheckEnhancedConversionTOS.js
asvinb: Sounds good! Updated @joemcgill

js/src/product-feed/submission-success-guide/enhanced-conversions/cta.js
asvinb: 👍

js/src/product-feed/submission-success-guide/enhanced-conversions/cta.js
asvinb: Nice improvement. Updated!

js/src/product-feed/submission-success-guide/enhanced-conversions/index.js
asvinb: 👍

js/src/product-feed/submission-success-guide/enhanced-conversions/index.js
asvinb: Updated @joemcgill

js/src/product-feed/submission-success-guide/enhanced-conversions/confirm-button.js
asvinb: 👍

js/src/components/enhanced-conversion-tracking-settings/enable-button.js
```diff
@@ -26,9 +27,16 @@ const EnableButton = ( { onEnable = noop } ) => {
 	};
 
 	return (
-		<AppButton isPrimary onClick={ handleOnEnableEnhancedConversions }>
-			{ __( 'Enable Enhanced Conversions', 'google-listings-and-ads' ) }
-		</AppButton>
+		<AppButton
+			icon={ externalIcon }
```
joemcgill, line 30: ✨ perfect icon choice

asvinb:
@joemcgill PR is ready for another round of review. To know whether we should skip the confirmation screen (because we are using the same components), I added a [new property in the data store](https://github.com/woocommerce/google-listings-and-ads/pull/2361/files#diff-8e875e4932c094a7da6eac9322edf0cb70fbf56fc555137bd686aa9e265a76cfR71) since we need to use that value in [cta.js](https://github.com/woocommerce/google-listings-and-ads/pull/2361/files#diff-4d5d5cd6e6e9e5b9aa9dd130f0e921c1ab104a1a00f5a8c8ed089bad7fbf132fR25) and [index.js](https://github.com/woocommerce/google-listings-and-ads/pull/2361/files#diff-e35f00421b697630fba0cb8ee3d492e36b9c7afe82a156e5d6ed1ce077fadbf2R28). I tried to have it as a state variable but unfortunately the components are rendered differently. Below are screenshots of the different screens within the modal:

Prompt user to enable EC (user has not accepted terms):
image

Confirmation (user has accepted terms but has not enabled EC in the store):
image

Success (after enabling EC):
image

Let me know if you have any questions

js/src/product-feed/submission-success-guide/enhanced-conversions/index.js
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		if (
+			acceptedCustomerDataTerms &&
+			allowEnhancedConversions !==
+				ENHANCED_ADS_CONVERSION_STATUS.ENABLED &&
+			! skipConfirmation
+		) {
+			return 'confirm';
+		}
```
joemcgill, line 175: The logic here seems incorrect to me. Seems like there could be a case where you have accepted terms, but not enabled EC, and if the `skipConfirmation` setting isn't set, you would end up with the `accept-terms` page. Is that expected?

It may be more clear to do these in the order of `loading`, `accept-terms`, `confirm`, and finally `success`. That way a simple `if ( ! acceptedCustomerDataTerms ) { ...` would result in the accept-terms page, and you would only reach success if all other cases get passed.

js/src/components/enhanced-conversion-tracking-settings/enable-button.js
```diff
@@ -0,0 +1,47 @@
...
+		updateEnhancedConversionsSkipConfirmation,
+	} = useAppDispatch();
+	const url = useGoogleAdsEnhancedConversionSettingsURL();
+
+	const handleOnEnableEnhancedConversions = () => {
+		window.open( url, '_blank' );
+		updateEnhancedAdsConversionStatus(
+			ENHANCED_ADS_CONVERSION_STATUS.PENDING
+		);
+		updateEnhancedConversionsSkipConfirmation( true );
```
joemcgill: This is only needed for the modal, so probably better to pass this in as part of the `onEnable` handler from the modal, rather than explicitly setting this here.

js/src/components/enhanced-conversion-tracking-settings/enable.js
joemcgill: No change is needed for now.

joemcgill (changes requested):
Thanks @asvinb. I started doing a new review but realized that the new success flow is not quite right. When I test the EC modal with T&C set to false, I would expect the following flow:

1. Show the accept-terms page
2. Click the accept terms CTA
3. Visit GAFE to enable EC and accept terms
4. Return to the modal (now with terms accepted)
5. Show the success page

Instead, when I returned to the modal from the GAFE, I am shown the confirm page. The confirm page should only be shown if the modal is opened with the terms already accepted. I can see that you're setting the skip confirmation state in the `EnableButton`, but it doesn't look like that state is ever reaching the data store for some reason.

I left some notes inline, but would be great to double check that this flow is working as expected.

js/src/product-feed/submission-success-guide/enhanced-conversions/index.js
asvinb: I changed the order as you suggested @joemcgill . Let me know what you think.

js/src/components/enhanced-conversion-tracking-settings/enable-button.js
asvinb: 👍 The `updateEnhancedConversionsSkipConfirmation` action has been moved. I've kept `updateEnhancedAdsConversionStatus` since it's needed when the CTA is rendered within the Settings panel.

PR #2366, opened 2024-04-11
Expose attribute mapping rule values using WCProductAdapter

jorgemd24:

This is an alternative for https://github.com/woocommerce/google-listings-and-ads/pull/2351.

Rather than moving the attribute logic from `WCProductAdapter` into the AttributeManager, we're now crating an instance of `WCProductAdapter` to retrieve all aggregated attribute values from particular products. Similar to the other PR, we'll use these attribute values to expose them in the WC REST Product endpoints. For additional context, refer to https://github.com/woocommerce/google-listings-and-ads/pull/2351.

1. Go to GLA -> Attribute -> Create some rules.
2. Make the following request `GET wc/v3/products?gla_syncable=1` or `GET wc/v3/products/YOUR_PRODUCT_ID?gla_syncable=1`
3. See that the response contains a new property named: `gla_attributes` with the correct values.

image

4. Update the products by adding a GLA Attribute. This attribute should take precedence over any mapping rule applied to the same attribute.

image

5. Remove the query parameter `gla_syncable=1` and the response should not contain `gla_attributes`.
7. Repeat the steps for a variation product using the following endpoint: `wc/v3/products/PARENT_ID/variations/VARIATION_ID?gla_syncable=1`

### Additional details:
- PHP unit tests are failing because of this issue: https://github.com/woocommerce/grow/issues/101 but you can run the tests locally.

### Changelog entry

>

### DIFF
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -10,6 +10,8 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Value\ChannelVisibility;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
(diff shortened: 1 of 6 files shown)


puntope (approved):
Thanks for adapting the PR to reduce code duplication.

It is working as expected for me. ✅

PR #2368, opened 2024-04-12
Restrict product types to be limited to only 10 when converting from categories.

matt-h:

Restrict product types to be limited to only 10 when converting from categories

If you give more than 10 then Google will give a warning of: Too many values [product type]

Pre-edited description:

~~Restrict product types to be limited to only 5 when converting from categories.~~

~~From the Google Documentation: https://support.google.com/merchants/answer/6324406?hl=en~~

~~Repeated field: Submit up to 5 times, but keep in mind that only the first value will be used to organize bidding and reporting in Google Ads Shopping campaigns.~~

~~If you give more than 5 then Google will give a warning of: Too many values [product type]~~

### Screenshots:

2024-04-12_378x194

2024-04-12_721x566

PHP Unit test added as well to make sure it is restricted to 10.

### Changelog entry

> Update - Restrict product types to be limited to only 10 when converting from categories.

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -210,6 +210,7 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 				),
 				__METHOD__
 			);
+			$google_product_types = array_slice( $google_product_types, 0, 10 );
 			$this->setProductTypes( $google_product_types );
 		}
 		return $this;
--- a/tests/Unit/Product/WCProductAdapterTest.php
+++ b/tests/Unit/Product/WCProductAdapterTest.php
@@ -571,6 +571,48 @@ DESCRIPTION;
 		$this->assertNotContains( 'Delta Category', $adapted_product->getProductTypes() );
 	}
 
+	public function test_product_types_are_limited_to_five_categories() {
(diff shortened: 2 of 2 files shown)


mikkamp:
Thanks @matt-h for the fix here. The code changes look good to me. However I did some testing and I'm not able to reproduce this warning with just 5 product types. I tested a little further and it seems the threshold lies at 10.

For example the following list submits without issues:
image

But when I add any additional ones, then it cuts off the first ones from the list and rejects them:
image

That's also when the same warning you mentioned shows up:
image

So can we set the threshold to 10 instead of just 5?

matt-h:
@mikkamp Interesting findings. I went with 5 just based on the documentation. I was testing with a product that had 11 categories assigned and it fixed it by going down to 5. I'll see if I can reproduce your results.

matt-h:
@mikkamp I was able to reproduce the same thing. Google seems to allow 10 and no more.

I've updated this Pull request to be 10 instead of 5.

mikkamp (approved):
Thanks for the changes. Everything looks good and I can confirm it allows up to 10, and doesn't submit any more.

PR #2370, opened 2024-04-15
Delete unused methods from ProductMetaQueryHelper.

jorgemd24:

While refactoring the Product Stats Overview section (already merged and released) I noticed that certain methods in ProductMetaQueryHelper are no longer use. Therefore, this PR removes those methods since they are not being used in the codebase.

- `ProductMetaQueryHelper::batch_insert_values`
- `ProductMetaQueryHelper::batch_update_values`

1. Double-check that `batch_insert_values` and `batch_update_values` are not used in the codebase.

### Changelog entry

> Tweak - Remove unused methods from ProductMetaQueryHelper


ianlin (approved):
Thanks for working on this. Tested locally using this branch and everything worked as expected. There are no more places using this two methods in the codebase.

mikkamp:
Merging this PR as the status is still the same and these functions remain unused.

Their usage has been removed since the release of [2.6.4](https://github.com/woocommerce/google-listings-and-ads/releases/tag/2.6.4) (released 26th March 2024). So even though they are public helper functions we should be safe to remove them now.

PR #2379, opened 2024-04-19
fix php warning due to iterating over null

matt-h:

Fixes this PHP warning:
```
Warning: foreach() argument must be of type array|object, null given in /wp-content/plugins/google-listings-and-ads/src/MerchantCenter/MerchantStatuses.php on line 431
```

The logic is the same, it just makes sure the results of `$mc_product_status->getItemLevelIssues()` are not empty before trying to iterate over it.

### Changelog entry
* Tweak - Confirm issues are present when retrieving product status.

### DIFF
--- a/src/MerchantCenter/MerchantStatuses.php
+++ b/src/MerchantCenter/MerchantStatuses.php
@@ -421,6 +421,11 @@ class MerchantStatuses implements Service, ContainerAwareInterface, OptionsAware
 				continue;
 			}
 
+			// Confirm there are issues for this product.
+			if ( empty( $mc_product_status->getItemLevelIssues() ) ) {
+				continue;
+			}
+
 			$product_issue_template = [
 				'product'              => html_entity_decode( $wc_product->get_name() ),
 				'product_id'           => $wc_product_id,


shameemreza:
Reported here: 8075716-zen
Thread link: https://wordpress.org/support/topic/error-log-filled-with-warning-foreach

shameemreza:
Another report here: https://wordpress.org/support/topic/version-2-6-7-update/

src/MerchantCenter/MerchantStatuses.php
```diff
@@ -428,24 +428,27 @@ protected function get_product_issues( array $statuses ): array {
 				'applicable_countries' => [],
 				'source'               => 'mc',
 			];
-			foreach ( $mc_product_status->getItemLevelIssues() as $item_level_issue ) {
-				if ( 'merchant_action' !== $item_level_issue->getResolution() ) {
-					continue;
+			$item_level_issues      = $mc_product_status->getItemLevelIssues();
+			if ( ! empty( $item_level_issues ) ) {
+				foreach ( $item_level_issues as $item_level_issue ) {
```
mikkamp: While the logic does stay the same, in this case I think it would be nicer to stick to the same style as the rest of the function where we either continue a for loop or return early. So we can avoid too many levels of indentation. The called functions will also be optimized when run so no need to store the return value in a separate variable.

So something like:
```php
// Confirm there are issues for this product.
if ( empty( $mc_product_status->getItemLevelIssues() ) {
	continue;
}

$product_issue_template = [
....
```

mikkamp (commented):
I'm not able to reproduce this error. Based on the class we are using [the data type is set to array](https://github.com/googleapis/google-api-php-client-services/blob/v0.345.0/src/ShoppingContent/ProductStatus.php#L33-L34) which should enforce it to always be an empty array when there are no issues available.

Apparently there are several reports of it going wrong, so I think we should add the extra check anyways. As mentioned in the comment I think we should do that without increasing the level of indentation.

Note: I also updated the changelog entry in this PR, as our changelog gets auto generated based on those entries.

src/MerchantCenter/MerchantStatuses.php
matt-h: Updated code to use this style of check.

mikkamp (approved):
Thanks for the change. It looks a bit simpler to me that way. As mentioned earlier I don't have a way to reproduce this beyond just mocking the returned results. But the code looks good to me.

PR #2386, opened 2024-04-25
Update Google API v16

puntope:
This PR: 

- updates the Google API to v16 
- upgrades all its dependencies
- Tweak the code based on the API changes
- Tweak the tests 
- Tweak CleanUpServices (structural changes required so)
- Fixes some bugs and deleted unused code

Please refer to the file comments to get context of all the changes. But in brief:

- Run `composer update googleads/google-ads-php --with-all-dependencies`
- Update V14 references to V16
- Replace sales_country with feed_label
- Replace Vertical with Source
- Add Request as param in some Mutation calls instead of ID+Operation
- Replace MerchantLink logic (they replaced it with ProductLink)
- Add /Client/ subfolder in the CleanupService
- Add an option to exclude deletions of certain services

```
composer update googleads/google-ads-php --with-all-dependencies

Upgrading googleads/google-ads-php (v19.2.0 => dev-legacy-v22.1.0 905fd78)
  - Upgrading google/auth (v1.34.0 => v1.37.1)
  - Upgrading google/common-protos (v4.4.0 => v4.5.0)
  - Upgrading google/gax (v1.25.0 => v1.29.1)
  - Upgrading google/longrunning (v0.2.6 => 0.4.1)
  - Upgrading google/protobuf (v3.25.0 => v3.25.3)
  - Upgrading monolog/monolog (2.9.2 => 2.9.3)
  - Upgrading symfony/deprecation-contracts (v2.5.2 => v2.5.3)
  - Upgrading symfony/polyfill-ctype (v1.28.0 => v1.29.0)
  - Upgrading symfony/polyfill-intl-normalizer (v1.28.0 => v1.29.0)
  - Upgrading symfony/polyfill-mbstring (v1.28.0 => v1.29.0)
  - Upgrading symfony/polyfill-php80 (v1.28.0 => v1.29.0)
  - Upgrading symfony/polyfill-php81 (v1.28.0 => v1.29.0)
```

```
composer update google/apiclient--with-all-dependencies
Upgrading google/apiclient (v2.15.3 => v2.16.0)
  - Upgrading google/apiclient-services (v0.331.0 => v0.350.0)
```

### Steps for testing

1. Checkout this PR
2. Run `rm -rf vendor && composer install`
3. Perform smoke testing as defined here https://github.com/woocommerce/google-listings-and-ads/wiki/Smoke-tests
4. Check there are no references to V14 in the code
5. Double check with https://developers.google.com/google-ads/api/docs/upgrade (v14 to v15 and v15 to v16) to see if we cover all the cases affecting us.
6. See[ E2E tests passing](https://github.com/woocommerce/google-listings-and-ads/actions/runs/8873455703) and [Unit Tests Passing](https://github.com/woocommerce/google-listings-and-ads/actions/runs/8873407567)

### Additional details

After merging this PR we should bump a minor version instead of a patch. So 2.7.1

Related to https://github.com/Automattic/woocommerce-connect-server/pull/2460

### Changelog entry

> Update - Update Google API to V16


bin/GoogleAdsCleanupServices.php
```diff
@@ -48,11 +48,18 @@ class GoogleAdsCleanupServices {
 	 */
 	protected $code_path = null;
 
+	/**
+	 * @var string[] List of Service to NOT remove even when usage is not found.
+	 */
+	protected $avoid_cleanup = [
```
puntope, line 54: Some services are loaded directly via `use` functions. However, sometimes we call them using `ResourceNames` helper. For that cases we add here the Service to avoid its removal.

bin/GoogleAdsCleanupServices.php
```diff
@@ -92,7 +99,10 @@ protected function remove_services() {
 		$used = array_unique(
 			array_merge(
 				$this->find_used_pattern(
-					"use Google\\\\Ads\\\\GoogleAds\\\\{$this->version}\\\\Services\\\\([A-Za-z0-9]+)ServiceClient;"
+					"use Google\\\\Ads\\\\GoogleAds\\\\{$this->version}\\\\Services\\\\Client\\\\([A-Za-z0-9]+)ServiceClient;"
```
puntope, line 102: Added the `Client` subfolder. (Thats a change in V16)

bin/GoogleAdsCleanupServices.php
```diff
@@ -282,7 +299,7 @@ protected function remove_suffix( string $suffix, string $text ): string {
 	 * @param string $name
 	 */
 	protected function remove_use_statement( string $file, string $name ) {
-		$pattern = '/\s*use\s+[\w\d\\\]+' . preg_quote( $name, '/' ) . ';/';
+		$pattern = '/\s*use\s+[\w\d\\\]+\\\{1}' . preg_quote( $name, '/' ) . ';/';
```
puntope, line 302: Fixes a bug when a service name is inside the name of another service


For example, trying to delete `FeedItem` will match and delete as well `ExtensionFeedItem`. So we add the slash to be sure we don't delete undesired services.

bin/GoogleAdsCleanupServices.php
```diff
@@ -400,10 +425,26 @@ protected function remove_directory( string $directory ) {
 		$directory = $this->file_path( $directory );
 		if ( ! is_dir( $directory ) ) {
 			$this->warning( sprintf( 'Directory does not exist: %s', $directory ) );
+
 			return;
 		}
 
-		rmdir( $directory ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_rmdir
+		if ( $this->is_empty_directory( $directory ) ) {
```
puntope, line 432: Avoids some errors trying to delete not empty directories.

src/API/Google/Ads.php
```diff
@@ -57,7 +58,7 @@ public function __construct( GoogleAdsClient $client ) {
 	 */
 	public function get_ads_accounts(): array {
 		try {
-			$customers = $this->client->getCustomerServiceClient()->listAccessibleCustomers();
+			$customers = $this->client->getCustomerServiceClient()->listAccessibleCustomers( new ListAccessibleCustomersRequest() );
```
puntope, line 61: Now the mutate calls require a Request object.

src/API/Google/Ads.php
```diff
@@ -128,22 +129,16 @@ public function get_billing_status(): string {
 	 * @throws Exception When a link is unavailable.
 	 */
 	public function accept_merchant_link( int $merchant_id ) {
-		$link = $this->get_merchant_link( $merchant_id );
-
-		if ( $link->getStatus() === MerchantCenterLinkStatus::ENABLED ) {
+		$link        = $this->get_merchant_link( $merchant_id );
```
puntope, line 132: MerchantLink was replaced by ProductLink

src/API/Google/AdsAssetGroup.php
```diff
@@ -149,23 +149,6 @@ public function create_operations( string $campaign_resource_name, string $asset
 		];
 	}
 
-	/**
-	 * Returns a set of operations to delete an asset group.
-	 *
-	 * @param string $campaign_resource_name
-	 * @return array
-	 */
-	public function delete_operations( string $campaign_resource_name ): array {
```
puntope, line 158: This was not used. So I delete it

src/API/Google/AdsAssetGroup.php
```diff
@@ -196,9 +179,9 @@ protected function asset_group_create_operation( string $campaign_resource_name,
 	protected function listing_group_create_operation(): MutateOperation {
 		$listing_group = new AssetGroupListingGroupFilter(
 			[
-				'asset_group' => $this->temporary_resource_name(),
-				'type'        => ListingGroupFilterType::UNIT_INCLUDED,
-				'vertical'    => ListingGroupFilterVertical::SHOPPING,
+				'asset_group'    => $this->temporary_resource_name(),
+				'type'           => ListingGroupFilterType::UNIT_INCLUDED,
+				'listing_source' => ListingGroupFilterListingSource::SHOPPING,
```
puntope, line 184: vertical was replaced by listing_source

src/API/Google/AdsAssetGroup.php
```diff
@@ -233,32 +216,6 @@ protected function asset_group_delete_operations( string $campaign_resource_name
 		return $operations;
 	}
 
-	/**
-	 * Returns an asset group listing group filter delete operation.
-	 *
-	 * @return MutateOperation[]
-	 */
-	protected function listing_group_delete_operations(): array {
```
puntope, line 241: This was not used so I delete it

src/API/Google/AdsCampaign.php
```diff
@@ -446,8 +447,8 @@ protected function create_operation( string $campaign_name, string $country ): M
 				'url_expansion_opt_out'     => true,
 				'shopping_setting'          => new ShoppingSetting(
 					[
-						'merchant_id'   => $this->options->get_merchant_id(),
-						'sales_country' => $country,
+						'merchant_id' => $this->options->get_merchant_id(),
```
puntope, line 450: sales_country was replaced by feed_label

src/API/Google/AdsConversionAction.php
```diff
@@ -128,7 +131,7 @@ public function get_conversion_action( $resource_name ): array {
 		try {
 			// Accept IDs too
 			if ( is_numeric( $resource_name ) ) {
-				$resource_name = ConversionActionServiceClient::conversionActionName( $this->options->get_ads_id(), $resource_name );
+				$resource_name = ConversionActionServiceClient::conversionActionName( strval( $this->options->get_ads_id() ), strval( $resource_name ) );
```
puntope, line 134: String type is required in this version

src/Google/Ads/ServiceClientFactoryTrait.php
```diff
@@ -76,12 +75,6 @@ public function getGoogleAdsClientOptions(): array {
 		return $clientOptions;
 	}
 
-	/**
-	 * @return AccountBudgetServiceClient
-	 */
-	public function getAccountBudgetServiceClient(): AccountBudgetServiceClient {
```
puntope, line 82: This was not used so I deleted it

tests/bootstrap.php
```diff
@@ -3,6 +3,7 @@
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests;
 
+use DG\BypassFinals;
```
puntope, line 6: This library is needed for bypassing final class. 

In v16 google marked all client services as Final. So when we mock them we got an error.

Please see https://github.com/googleads/google-ads-php/issues/1008

bin/GoogleAdsCleanupServices.php
```diff
@@ -312,7 +331,7 @@ protected function remove_function( string $file, string $name ) {
 		$bracket = 0;
 
 		// Parse until we find beginning of doc block.
-		$start = strrpos( $contents, '/**', ( $length - $offset ) * -1 );
+		$start = strrpos( $contents, '/**', ( $length - $offset ) * - 1 );
```
mikkamp: Not a bit deal, but how is adding the space making this code clearer?

It should be "length times a negative 1", but adding the space makes it look more like "length times and minus 1" which in my opinion is more confusing.

bin/GoogleAdsCleanupServices.php
mikkamp: Not harmful to keep this code around, but in which cases does it try to delete a not empty directory?

I don't see the message "Directory not empty" appear in the output. It's only called from `remove_enum`, my question is if it's not able to remove an unused directory for an enum, then what files are remaining there that we aren't removing? Wouldn't it be better to remove those files as well if needed?

mikkamp (approved):
Thanks for all the changes here. When testing this I was able to perform all the normal operations without any issues.

I just came across one issue which was not related so logged it in #2398 

I'm marking this PR as approved, as I just left a few small minor comments/questions.

martynmjones (approved):
Ran some additional smoke tests on this PR and all functionality appears to be working correctly after updating the library to `v16` 👍

bin/GoogleAdsCleanupServices.php
puntope: > in which cases does it try to delete a not empty directory?

Maybe it was because when I did this change it was without the changes related to the old structure. (So some enum classes were not deleted from Enum folder) 

> what files are remaining there that we aren't removing? Wouldn't it be better to remove those files as well if needed?

In my point of view not deleting a directory with some not deleted files is less risky (just a matter of space) than deleting them even when they were not deleted by the script (potential functionality class not found problem).

Since you mentioned is not harmful I will keep it just in case. But if is not necessary we can remove it as well.

PR #2393, opened 2024-05-08
Fix E2E test when using RC

puntope:
### Changes proposed in this Pull Request:

When testing E2E for WC 8.9.0-rc.1 I saw the E2E tests were failing because the workflow was not able to deactivate WooCommerce. 

After investigating. Seems like this is happening because the way we load and install WooCommerce in our wp-env. This PR updates the logic to install WC in the `test-env-setup.sh` file.

_Replace this with a good description of your changes & reasoning._

### Screenshots:

Before
Screenshot 2024-05-08 at 18 25 26

After
Screenshot 2024-05-08 at 18 25 51

### Detailed test instructions:

1. Don't checkout this PR
2. Run `npm run wp-env:up`
3. Run `npm run -- wp-env run tests-cli -- wp plugin update woocommerce --version=8.9.0-rc.1`
4. See the operation failing. 
5. Checkout this PR
6. Run `npm run wp-env:up`
7. Run `npm run -- wp-env run tests-cli -- wp plugin update woocommerce --version=8.9.0-rc.1`
8. See the operation succeeding. 
9. See workflow for E2E tests failing IN THE SETUP before this PR https://github.com/woocommerce/google-listings-and-ads/actions/runs/9001351242
10. See them correct IN THE SETUP when using this PR https://github.com/woocommerce/google-listings-and-ads/actions/runs/9003604947

### Changelog entry

> Fix - E2E test when using RC

PR #2395, opened 2024-05-09
Fix failing E2E tests

puntope:

Some tests in GLA are failing because the classes are not defined correctly. This PR addresses that.

### Screenshots:

Screenshot 2024-05-09 at 11 48 53

1. Checkout this PR
2. E2E tests pass
3. See E2E test passing in the GH WF https://github.com/woocommerce/google-listings-and-ads/actions/runs/9014174756

### Additional details:

- One of the errors is because the variable product is located on first place so the add to cart fails.
Screenshot 2024-05-09 at 11 27 09

- The other one is because of the classes used in selecting the different cards in the last step. .components-card matches a totally unrelated div in the page. So the selector won't work.
- 
Screenshot 2024-05-09 at 11 41 33

⚠️  If you test this PR use the latest WC not the RC (it will fail if you do so, fixed here https://github.com/woocommerce/google-listings-and-ads/pull/2393 )

### Changelog entry

> Fix - E2E tests

### DIFF
--- a/tests/e2e/utils/customer.js
+++ b/tests/e2e/utils/customer.js
@@ -42,8 +42,8 @@ export async function singleProductAddToCart( page, productID ) {
  */
 export async function relatedProductAddToCart( page ) {
 	const addToCart = ( await page.locator( '.related.products' ).isVisible() )
-		? '.related.products .add_to_cart_button'
-		: '.wp-block-woocommerce-related-products .add_to_cart_button';
+		? '.related.products .add_to_cart_button.product_type_simple'
+		: '.wp-block-woocommerce-related-products .add_to_cart_button.product_type_simple';
 
 	const addToCartButton = await page.locator( addToCart ).first();
 	await addToCartButton.click();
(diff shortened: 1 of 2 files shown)


tests/e2e/utils/customer.js
```diff
@@ -42,8 +42,8 @@ export async function singleProductAddToCart( page, productID ) {
  */
 export async function relatedProductAddToCart( page ) {
 	const addToCart = ( await page.locator( '.related.products' ).isVisible() )
-		? '.related.products .add_to_cart_button'
-		: '.wp-block-woocommerce-related-products .add_to_cart_button';
+		? '.related.products .add_to_cart_button.product_type_simple'
+		: '.wp-block-woocommerce-related-products .add_to_cart_button.product_type_simple';
```
jorgemd24, line 46: Since this function is made to work only for simple products, I think we should add a comment stating that only will work for simple products.

jorgemd24 (approved):
Thanks @puntope, I tested locally and reviewed the code and I can confirm that all the tests are passing now. I left a small comment about the function `relatedProductAddToCart` but I am approving in advanced.

PR #2400, opened 2024-05-13
Prevent PHP Warning when Statistics is null

### LINKED ISSUE #2398: Statistics not always available when triggering notifications
### Describe the bug:

I'm seeing the following entries show up in my debug log:
```
[13-May-2024 08:59:01 UTC] PHP Warning:  Trying to access array offset on null in wp-content/plugins/google-listings-and-ads/src/MerchantCenter/MerchantCenterService.php on line 436
```

Tracing back the code, this happens when the CampaignSetup notifications do their checks (run once a day). If we don't have the statistics cached then it will schedule to have them populated and will return the statistics as null. Meaning we can't check right away if there is at least one active product. I haven't checked the details fully but it looks like `has_account_issues` might also be affected.

### Steps to reproduce:
1. Enable debug logging to see PHP warnings 
2. Clear statistics cache
3. Trigger the scheduled action to run notification
4. Confirm that the PHP warning is generated when trying to check if a notification should be triggered

### Expected behavior:
Statistics to be available, or for it to only run the checks after statistics become available.

puntope:

Closes part of #2398 

This PR prevents a PHP warning when `statistics` is null as we try to access ['statistics']['active`] in some parts of the code without checking if it's set.

1. Enable debug logging to see PHP warnings
2. Clear statistics cache
3. Trigger the scheduled action to run notification ( `wc_gla_cron_daily_notes` )
4. Confirm that the PHP warning is not generated and note is not created

Notice that some conditions are required for this Note to be triggered. For testing this I recommend commenting off some checks under `AbstractSetupCampaign::should_be_added` or be sure you pass all those checks.

So we can focus on the problematic method `$this->merchant_center_service->has_at_least_one_synced_product()`

### Changelog entry

> Fix - Prevent PHP Warning when Statistics is null

### DIFF
--- a/src/MerchantCenter/MerchantCenterService.php
+++ b/src/MerchantCenter/MerchantCenterService.php
@@ -433,6 +433,6 @@ class MerchantCenterService implements ContainerAwareInterface, OptionsAwareInte
 	public function has_at_least_one_synced_product(): bool {
 		$statuses = $this->container->get( MerchantStatuses::class )->get_product_statistics();
 
-		return $statuses['statistics']['active'] >= 1;
+		return isset( $statuses['statistics']['active'] ) && $statuses['statistics']['active'] >= 1;
 	}
 }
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Thanks for fixing it, I no longer see the error message after running the job when statistics doesn't exist. Just one question, in the PR description you mentioned:

> Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/2398

Would it still be follow ups, or it already completely solved the issue #2398?

PR #2408, opened 2024-05-20
 Rename from Google Listings and Ads to Google for WooCommerce

puntope:

PT: pcTzPl-2i9-p2

This PR replaces all the "Google Listings and Ads" ocurrences with "Google for WooCommerce"

1.  Verify in Codebase that there are no instances left of Google Listings and Ads or Google Listings & Ads.
2. Checks (PHP Unit and JS tests) and [E2E tests](https://github.com/woocommerce/google-listings-and-ads/actions/runs/9157439287) are passing.
3. Check the changes in this PR to see if they make sense.

### Changelog entry

> Update - Update extension naming from Google Listings and Ads to Google for WooCommerce


changelog.txt
```diff
@@ -525,7 +525,7 @@
 = 1.12.0 - 2022-03-29 =
 * Add - Additional data points for tracker snapshot.
 * Add - Enables merchants to select multiple countries as their audience when creating a Google Ads campaign (Smart Shopping Campaign).
-* Add - Google Listings And Ads product attributes icon.
+* Add - Google for WooCommerce And Ads product attributes icon.
```
ianlin: The `And Ads` wasn't replaced here.

I was actually wondering should we do the rename in changelog? As the changelog records what we've done in the past and the name Google Listings & Ads would be the history as well.

ianlin (commented):
Thanks for rebranding it in the plugin code. It looks all good, just found a place that wasn't replaced correctly.

changelog.txt
puntope: Right @ianlin I restore the old changes in the changelog and just replaced the title. Can you take another look?

ianlin (approved):
Thanks for the extra change, LGTM. 👍

puntope:
Closed in favor of https://github.com/woocommerce/google-listings-and-ads/pull/2481

PR #2409, opened 2024-05-22
Add Integration Status Report for the partner integration. 

jorgemd24:

This PR adds a section to the connection test page that displays the status of the following points:

- **Jetpack Connection Health**: Indicates whether the Jetpack connection between WPCOM and the remote site is healthy. This involves a series of checks, such as communicating with the blog via XML-RPC and REST, and verifying that the blog owner exists and is a site member.
- **Last Jetpack Contact**: Shows the last time Jetpack was in contact with the remote site.
- **WC REST API Health**: Tests if WPCOM can access the WC REST API. Currently, it checks the /wc/v3/products endpoint, with the possibility of testing other endpoints in the future.
- **Google Token Health**: Displays whether there is a valid WPCOM token with the correct access scope linked to the Google WPCOM client ID and the user connected to the Jetpack token.
-  **Errors**: Shows the errors returned by WPCOM.

image

1. Apply the WPCOM Diff `D149450`
2. Go to the Connection Test Page
3. Click "Get API Pull Integration Status"
4. Check the results align with your current setup. For instance:
 -  Changing the Google Client ID in Diff `D149450`,
 -  Make your site unreachable by disconnecting Ngrok.
 -  Make WC REST API Products unreachable by changing the endpoint here: 

https://github.com/woocommerce/woocommerce/blob/42f693d8b2ca848a0e62b8ac3bed569db3995283/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-products-v2-controller.php#L64
- Throw an error in the following method: 

https://github.com/woocommerce/google-listings-and-ads/blob/d6898e0ee7be7f3fb6007551ea71cc501730e403/src/Integration/WPCOMProxy.php#L300-L302

### Changelog entry

>

### DIFF
--- a/src/ConnectionTest.php
+++ b/src/ConnectionTest.php
@@ -863,7 +863,7 @@ class ConnectionTest implements Service, Registerable {
 			$integration_status_args = [
 				'method'  => 'GET',
 				'timeout' => 30,
-				'url'     => 'https://public-api.wordpress.com/wpcom/v2/sites/' . Jetpack_Options::get_option( 'id' ) . '/wc/partners/remote-site-status?partner=google&XDEBUG_SESSION_START=XDEBUG_OMATTIC',
+				'url'     => 'https://public-api.wordpress.com/wpcom/v2/sites/' . Jetpack_Options::get_option( 'id' ) . '/wc/partners/remote-site-status?partner=google',
 				'user_id' => get_current_user_id(),
 			];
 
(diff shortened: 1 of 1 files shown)


puntope (commented):
Hi @jorgemd24 Thanks for working on this. Looking good but when I got a 400 response (without `integration_status_response`)  there is no output in the action. 

Screenshot 2024-05-23 at 11 25 46

Would be interesting to catch this scenario as well. WTDY?

jorgemd24:
Thanks @puntope for catching this! I adjusted the PR and this is ready for another round!

puntope (approved):
Thx for the updates LGTM

PR #2411, opened 2024-05-24
Implement RequestReview in the Merchant class

puntope:

This PR implementers Get Account Review Status and Request a Review in the MErchant Class instead of doing it with WooCommerce Connect Server (via Middleware)

The reason of this is in order to avoid unnecessary manager quota requests

See slack context here: p1713778408402949-slack-C02BB3F30TG 

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/5908855/a4d57af3-9d0b-411e-a394-8dbabcad80ac

1. Checkout this PR
2. Connect MC with an account  disapproved and eligible for review request
3. See the disapproved message in the product feed tab
4. See the request review notice
5. Request a review successfully and see how the status goes to "Under review"
6. Check now an approved or pending to approve account
7. See the review status message is correct in product feed tab

### Changelog entry

> Update - Implement Account Request Review Requests in the extension


src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
```diff
@@ -251,14 +254,91 @@ private function get_cached_review_status(): ?array {
 	 * @throws Exception If the get_account_review_status API call fails.
 	 */
 	private function get_review_status(): ?array {
-		$review_status = $this->get_cached_review_status();
+		$review_status = null;
```
ianlin: ❓ Are we no longer using the transient? If that's the case, we could probably remove the `get_cached_review_status` and `set_cached_review_status` methods as well. But IMO we should continue using the transient as we did before.

src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
```diff
@@ -251,14 +254,91 @@ private function get_cached_review_status(): ?array {
...
+			$response = $this->merchant->get_account_review_status();
+
+			if ( isset( $response['freeListingsProgram'] ) || isset( $response['shoppingAdsProgram'] ) ) {
+				do_action( 'woocommerce_gla_request_review_response', $response );
+				return $response;
+			}
+
+			do_action( 'woocommerce_gla_invalid_response', $response, __METHOD__ );
+			$error = $response['message'] ?? __( 'Invalid response getting account review status', 'google-listings-and-ads' );
+			throw new Exception( $error, $response->getCode() );
```
ianlin: I was wondering if this code would ever be executed. In this PR, the `merchant->get_account_review_status()` method [returns an array of responses for the statuses of freeListingsProgram and shoppingAdsProgram](https://github.com/woocommerce/google-listings-and-ads/pull/2411/files#diff-3e63bb75e657a93ffd3bad49c332e79eb7e3e321eca32a71892dae8eb1a19f73R383-R386). If an error occurs, [it throws an exception](https://github.com/woocommerce/google-listings-and-ads/pull/2411/files#diff-3e63bb75e657a93ffd3bad49c332e79eb7e3e321eca32a71892dae8eb1a19f73R389). So I believe lines 288 to 290 would never be reached.

src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
```diff
@@ -251,14 +254,91 @@ private function get_cached_review_status(): ?array {
...
+	public function get_account_review_status() {
+		try {
+			if ( ! $this->middleware->is_subaccount() ) {
+				return [];
+			}
+
+			/** @var array| Exception $response */
+			$response = $this->merchant->get_account_review_status();
+
+			if ( isset( $response['freeListingsProgram'] ) || isset( $response['shoppingAdsProgram'] ) ) {
```
ianlin: Maybe using `&&` instead of `||` as it will always include both of them in the response? Moreover, what about checking the response status code for both `$response['freeListingsProgram']` and `$response['shoppingAdsProgram']`?

ianlin (commented):
Thanks for working on this, it tested well locally, I've left some comments that we could discuss about.

I also found there was a PHP Notice while requesting to `GET /gla/mc/review`:

> Indirect modification of overloaded element of Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\S    ervice\ShoppingContent\FreeListingsProgramStatusRegionStatus has no effect in /var/www/html/wp-content/plugins/google-listings-and-ads/src/Google/RequestReviewStatuses.php on line 136"

The warning seems to be in [src/Google/RequestReviewStatuses.php:136](https://github.com/woocommerce/google-listings-and-ads/blob/88c21196813ee12dded2b79533d10dc07853c3f1/src/Google/RequestReviewStatuses.php#L136), a [quick Google search](https://stackoverflow.com/questions/74910786/indirect-modification-of-overloaded-element) suggests that we could create a copy of `$region_status['regoinCodes']` before sorting the array in-place and it worked on my end. But what I don't understand is why this PR would have this warning but without this PR it's all good.

src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
puntope: Thx for this catch @ianlin I commited that accidentally (I set it as null for testing) 

Fixed here https://github.com/woocommerce/google-listings-and-ads/pull/2411/commits/514bbe59ad8e8c324d58b0b7611064b0b3b61ed1

src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
puntope: `get_account_review_status` returns an array with both `freeListingsProgram` and `shoppingAdsProgram` we don't need both of them as we merge them afterwards in `RequestReviewStatuses::get_statuses_from_response()`

Regarding the status, the Google call will return `FreeListingsProgramStatus` or `ShoppingAdsProgramStatus` which is a Collection and doesn't contain statuses.

src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
puntope: Thats right. I tweaked it here.

https://github.com/woocommerce/google-listings-and-ads/pull/2411/commits/7d9ed7db7a5b9cab54b9fff3ad4f62fd7bc40c25

puntope:
> The warning seems to be in [src/Google/RequestReviewStatuses.php:136](https://github.com/woocommerce/google-listings-and-ads/blob/88c21196813ee12dded2b79533d10dc07853c3f1/src/Google/RequestReviewStatuses.php#L136), a [quick Google search](https://stackoverflow.com/questions/74910786/indirect-modification-of-overloaded-element) suggests that we could create a copy of `$region_status['regoinCodes']` before sorting the array in-place and it worked on my end. But what I don't understand is why this PR would have this problem but without this PR it's all good.

Thanks @ianlin Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2411/commits/c9078bd5a52b2a54c19b177298af453685afedc0

> But what I don't understand is why this PR would have this problem but without this PR it's all good.

Notice the response we got before via WCS has a JSON encode/decode process making a copy of those arrays. In the current implementation we operate with the response directly. I would bet that may be the reason.

puntope:
Hi @ianlin Thanks for your great review and for catching those issues. THis is ready for a new round.

src/API/Site/Controllers/MerchantCenter/RequestReviewController.php
ianlin: Ah that makes sense, thanks for clarifying.

src/API/Google/Merchant.php
```diff
@@ -367,4 +370,54 @@ public function has_access( string $email ): bool {
 	public function update_merchant_id( int $id ): bool {
 		return $this->options->update( OptionsInterface::MERCHANT_ID, $id );
 	}
+
+	/**
+	 * Get the review status for an MC account
+	 *
```
ianlin, line 376: Since it's a new method in this class maybe we could add `since x.x.x`.

src/API/Google/Merchant.php
```diff
@@ -367,4 +370,54 @@ public function has_access( string $email ): bool {
...
+			];
+		} catch ( GoogleException $e ) {
+			do_action( 'woocommerce_gla_mc_client_exception', $e, __METHOD__ );
+			throw new Exception( $e->getMessage(), $e->getCode() );
+		}
+	}
+
+	/**
+	 * Request a review for an MC account
+	 *
```
ianlin, line 397: Same for adding `since x.x.x`.

ianlin (approved):
Thanks for the prompt update, all good. 👍

PR #2412, opened 2024-05-27
Avoid a possible error when linking accounts due to the delay between Google APIs

eason9487:

PT: pcTzPl-2iK-p2

It's an additional fix for a bug found while testing #2215.

This PR avoids a possible error when linking accounts due to the delay between Google APIs of Google Merchant Center and Google Ads. Please refer to the **Additional details** section for the issue details.

- Add a retry mechanism to the `API\Google\Ads::get_merchant_link` method.

💡 After a few rounds of testing, I could get an invitation link on the first retry in my dev environment, so I set the number of retries to 3.

### Screenshots:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/e7e52c1d-cfaf-4ed7-88f3-53417c6259e5

1. Disconnect Google Ads account if there is a connected one.
2. Go to step 1 of the onboarding flow.
3. Connect a Google Merchant Center account first.
4. Create a new Google Ads account and claim it.
5. Check if the “Merchant link is not available to accept” error doesn't occur.

### Additional details:

Looks like after updating Google API to v16 (#2386), the invitation link may not be available in Google Ads immediately after the invitation is sent from Google Merchant Center. This may result in a “Merchant link is not available to accept” error when linking to a newly created Google Ads account.

#### The way to reproduce the issue

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/bc6b21ca-fdd7-4b8c-9238-aee4dd16fe7f

### DIFF
--- a/src/API/Google/Ads.php
+++ b/src/API/Google/Ads.php
@@ -129,7 +129,7 @@ class Ads implements OptionsAwareInterface {
 	 * @throws Exception When a link is unavailable.
 	 */
 	public function accept_merchant_link( int $merchant_id ) {
-		$link        = $this->get_merchant_link( $merchant_id );
+		$link        = $this->get_merchant_link( $merchant_id, 3 );
 		$link_status = $link->getStatus();
 		if ( $link_status === ProductLinkInvitationStatus::ACCEPTED ) {
 			return;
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Thanks for fixing the bug by adding the retry. I could reproduce the bug and this PR solved it in my local. Also confirmed in the Google Ads dashboard there were no pending invitation. 👍

PR #2419, opened 2024-05-30
Adjust event tracking for the creating and claim buttons of Google Ads account

eason9487:

PT: pcTzPl-2iK-p2

Since the `context` property of most event tracking is currently hardcoded, and there is no proper way to obtain state or data from far away components when sending event tracking, this PR tries to implement it via `@wordpress/hooks`. The goal is to avoid passing event properties in a prop drilling fashion, e.g. `context` and `step` at the `<Stepper>` layer in this PR need to be sent together in buttons several layers away.

Due to the number of files being changed, to avoid this PR getting too large, I would like to use two tracking events to show how the mechanism works, and then add the remaining event tracking in a subsequent PR.

#### 📌 Mechanism:

- Add a React hook `useEventPropertiesFilter` to handle event properties across multiple layers of components via `@wordpress/hooks`.

#### 📌 Event tracking:

- Add `context` and `step` event properties to Extension and Ads onboarding steppers for use by their subcomponents.
- Add event tracking to `ClaimAccountButton`.
- Add `context` and `step` event properties to the click event tracking in `TermsModal`.

#### 📌 Extra change:

- Reuse `ClaimAccountButton` to reduce the duplicate implementation.
- Adjust jest config and setup so that tests related to event tracking don't have to validate global event properties.

1. Open the Console tab in the browser DevTool, and run `localStorage.setItem( 'debug', 'wc-admin:*' )` to enable tracking debugging mode.
2. Disconnect Google Ads account if there is a connected one.
3. Go to step 1 of the **Extension onboarding**.
4. Create a new Google Ads account.
5. Check if the `gla_ads_account_create_button_click` event is sent with `context` and `step`
6. Click the "Claim account" button to open the invitation link in a pop-up window for the newly created Google Ads account. 
   - It doesn't need to proceed with the claim process here
7. Check if the `gla_ads_account_create_button_click` event is sent with `context` and `step`.
8. Disconnect Google Ads account if there is a connected one.
9. Go to step 1 of the **Ads onboarding**.
10. Repeat test steps 4-7

### Changelog entry

> Tweak - Adjust event tracking for the creating and claim buttons of Google Ads account.


tomalec:
At step 7. on onboardign flow, I didn't get `context` and `step` properties 
image
[Screencast from 30.05.2024 21:45:13.webm](https://github.com/woocommerce/google-listings-and-ads/assets/17435/0c3ea292-7d30-4eb5-8fda-d5279de73787)


I don't get it when I click "Claim Account" button in the modal in the main flow just after creating a new account.
Once I reload the page after account creation and click "Calim Account" the properties are set correctly.

eason9487:
Hi @tomalec, thanks for catching this.

> At step 7. on onboardign flow, I didn't get `context` and `step` properties 
> I don't get it when I click "Claim Account" button in the modal in the main flow just after creating a new account. Once I reload the page after account creation and click "Calim Account" the properties are set correctly.

Sorry, I tested each event tracking separately but missed the interaction between `useEventPropertiesFilter` instances. The issue should have been fixed in https://github.com/woocommerce/google-listings-and-ads/pull/2419/commits/2129e87052e02c7f4f3a3ed181956d21245fcc34.

Could you help with a new round?

js/src/components/google-ads-account-card/claim-account-button.js
```diff
@@ -1,33 +1,54 @@
...
+ */
+const ClaimAccountButton = ( { onClick = noop, ...restProps } ) => {
 	const { inviteLink } = useGoogleAdsAccountStatus();
+	const getEventProps = useEventPropertiesFilter( FILTER_ONBOARDING );
 
-	const handleClaimAccountClick = useCallback(
-		( event ) => {
-			const { defaultView } = event.target.ownerDocument;
-			const features = getWindowFeatures( defaultView, 600, 800 );
+	const handleClaimAccountClick = ( event ) => {
```
tomalec, line 36: :grey_question: Why do we no longer need `useCallback`? I see in both cases we re-render the ClaimAccount button dozens of times

eason9487:
Added JS tests for the `useEventPropertiesFilter` hook. https://github.com/woocommerce/google-listings-and-ads/pull/2419/commits/8dd83ab33663a2ebca9f0ecddc269ef4f683e30d

js/src/hooks/useEventPropertiesFilter.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Hook to handle event properties across multiple layers.
+ *
+ * For example, a button that is rendering a child layer requires state or context
+ * held by the parent component when recording event tracking, and they span many
+ * layers. Use this hook to avoid prop drilling.
+ *
+ * @param {string} filterName The name of the filter to use.
+ * @param {Object} extraProperties The extra event properties to be added. Only valid properties are processed.
```
tomalec: :nail_care: I find "valid" quite vague here, I think it would be nice to explain it more precisely. It's "properties defined in the `filterPropertiesMap`", right?

js/src/components/google-ads-account-card/claim-account-button.test.js
```diff
@@ -0,0 +1,94 @@
...
+		expect( windowOpen ).toHaveBeenCalledWith(
+			'https://example.com',
+			'_blank',
+			// Ignore the value of the window features.
+			expect.any( String )
+		);
+	} );
+
+	it( 'should record click events and be aware of extra event properties from filters', async () => {
+		await expectEventWithPropertiesFilter(
```
tomalec: :nail_care: 
I read `expectEventWithPropertiesFilter` as I'd read `expect(event_with_properties_filter)`, so something that starts an expectation to which I can attach matches.
But then, in this usage, "I expect the event with properties filter to do what?" 

Here, the function already applies matches and actually tests.

WDYT, of renaming the function to `expect___To___()`.  I hope such a notion could help me with the below.
Actually, I had a hard time understanding even after reading the source code, what entity we attach the expectation to: an event, a component, a filter.; and what is the behavior we expect: to be recorded, to be recorded with given properties, to be clicked, to be filtered?

tomalec: Is it that we "expect the given component, to record an event of given name, with given properties filtered by given filter, once clicked on a target"?
So something like `expect( Component ).toRecordEventWithFilteredPropertiesOnceClicked(eventName, properties, filterName, clickTarget )`
Or do I miss something?

js/src/hooks/useEventPropertiesFilter.js
```diff
@@ -0,0 +1,57 @@
...
+import { pick, uniqueId } from 'lodash';
+
+/**
+ * Internal dependencies
+ */
+import useIsEqualRefValue from './useIsEqualRefValue';
+import { NAMESPACE, hooks, filterPropertiesMap } from '.~/utils/tracks';
+
+/**
+ * Hook to handle event properties across multiple layers.
```
tomalec: :nail_care: "Handle" could mean many things. Would "propagate" fit here, and be more precise? I'm trying to build an understanding just from reading the description.

tomalec (approved):
Tested locally and reviewed the code LGTM.

Left some non-blocking comments and questions.

js/src/hooks/useEventPropertiesFilter.js
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2419/commits/1179f5b3cd616c6af5720f3309bee5869283beb6.

js/src/hooks/useEventPropertiesFilter.js
eason9487: Thanks for the suggestion! Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2419/commits/1179f5b3cd616c6af5720f3309bee5869283beb6.

js/src/components/google-ads-account-card/claim-account-button.test.js
eason9487: > Is it that we "expect the given component, to record an event of given name, with given properties filtered by given filter, once clicked on a target"?

Yes.

> WDYT, of renaming the function to `expect___To___()`. I hope such a notion could help me with the below.
> Actually, I had a hard time understanding even after reading the source code, what entity we attach the expectation to: an event, a component, a filter.; and what is the behavior we expect: to be recorded, to be recorded with given properties, to be clicked, to be filtered?

Thanks for the suggestion! Made some changes in https://github.com/woocommerce/google-listings-and-ads/pull/2419/commits/834838934266ecafbb0f27a7177431a29f34c5cb:

- Added JSDoc
- Changed the use of the `queryElement` callback and renamed it to `performAction`
   - The action of triggering an event tracking is completely moved from this function to the callers
   - This should make the use of this function more flexible and the naming does not need to carry the “OnceClicked”
- Renamed `expectEventWithPropertiesFilter` to `expectComponentToRecordEventWithFilteredProperties`

js/src/components/google-ads-account-card/claim-account-button.js
eason9487: `useCallback` is mainly used for performance optimization, as an API for custom hook, or to avoid unnecessary `useEffect` triggering. If there are no performance issues, lower code complexity is usually preferred.

PR #2420, opened 2024-05-31
Disallow tracking enabled check in the extension

puntope:

WC_Tracks disallows data tracking by itself when the user is not opted-in. This allows us to check the snapshot data in WP-CLI

This PR disables the check for `woocommerce_allow_tracking`

### Screenshots:

Screenshot 2024-05-31 at 11 26 18

1. Run `wp option update woocommerce_allow_tracking "no"`
2. Run `wp wc tracker snapshot --format=yaml`
3. See gla extension data in the output

### Changelog entry

> Fix - Show tracking snapshots in WPCLI


martynmjones (approved):
Hey @puntope, thanks for the fix!

Tested locally and confirmed that the tracker data is included in the snapshot when tracking is disabled so LGTM ✅ 

Screenshot 2024-06-05 at 12 07 54

PR #2421, opened 2024-05-31
Adjust click event tracking when connecting, disconnecting, and opening billing setup for Google Ads account

eason9487:

PT: pcTzPl-2iK-p2

The PR adds/adjusts three tracking events for for Google Ads account:

- Add `context` and `step` event properties to the click event tracking in `ConnectAds`.
- Add the `gla_ads_account_disconnect_button_click` event with `context` and `step` event properties to the click event tracking in `DisconnectAccount`.
- Adjust `context` and add `step` event properties to the click event tracking in `BillingSetupCard`.
- Extra change: Remove the unused component `GoogleAdsAccountSection`.

1. Open the Console tab in the browser DevTool, and run `localStorage.setItem( 'debug', 'wc-admin:*' )` to enable tracking debugging mode.
2. Disconnect Google Ads account if there is a connected one.
3. Go to step 1 of the **Extension onboarding**.
4. Connect an existing Google Ads account.
5. Check if the `gla_ads_account_connect_button_click` event is sent with `id`, `context` and `step`.
6. Disconnect the Google Ads account.
7. Check if the `gla_ads_account_disconnect_button_click` event is sent with `context` and `step`.
8. Create a new Google Ads account but don't set up billing, and proceed to step 4 of the onboarding flow.
9. Open the campaign creation setup.
10. Click on the "Set up billing" button to see if it can open the billing setup page in a pop-up window.
11. Check if the `gla_ads_set_up_billing_click` event is sent with `link_id`, `href`, `context` and `step`.
12. Click on the "click here instead" link to see if it can open the billing setup page in a new browser tab.
13. Check if the `gla_ads_set_up_billing_click` event is sent with `link_id`, `href`, `context` and `step`.

### Changelog entry

> Tweak - Adjust click event tracking when connecting, disconnecting, and opening billing setup for Google Ads account.


js/src/components/paid-ads/billing-card/billing-setup-card.js
```diff
@@ -58,23 +73,28 @@ const BillingSetupCard = ( { billingUrl, onSetupComplete } ) => {
...
+							}
 						) }
 					</div>
 				</div>
-				<AppButton
-					isSecondary
-					onClick={ handleClick }
-					eventName="gla_ads_set_up_billing_click"
-					eventProps={ {
-						context: 'setup-ads',
```
ianlin, line 73: It used to set the context as `setup-ads`, but right now it showed `setup-mc` when I clicked `Set up billing` button. Is it expected?

Screenshot 2024-06-03 at 16 04 51

ianlin (commented):
Thanks for adding event tracking, I started by testing the behaviour and I've left a question regarding the context `setup-ads` and `setup-mc`.

js/src/components/paid-ads/billing-card/billing-setup-card.js
eason9487: Yes, `'setup-mc'` is expected if you were doing extension onboarding. The billing setup card is also used in the Ads onboarding flow, where the value of context is expected to be `'setup-ads'`.

It's true that there's some inconsistency and confusion in the use of these context values, so I've put them down in tracks.js momentarily.

https://github.com/woocommerce/google-listings-and-ads/blob/a63df8d70c682c380a3db39378fa86de92e9f365/js/src/utils/tracks.js#L27-L40

js/src/components/google-ads-account-card/connect-ads/index.js
```diff
@@ -131,7 +136,9 @@ const ConnectAds = ( props ) => {
 							isSecondary
 							disabled={ ! value }
 							eventName="gla_ads_account_connect_button_click"
-							eventProps={ { id: Number( value ) } }
+							eventProps={ getEventProps( {
+								id: Number( value ),
```
ianlin, line 140: Just noting down what I thought while reviewing. I was thinking that the name of the Ads ID property when disconnecting account is `gla_ads_id` so why not use the same here for consistency, but it's been there before this PR so I think keeping it as is would be better for querying the tracking results.

js/src/components/google-ads-account-card/connect-ads/index.js
```diff
@@ -18,15 +18,19 @@ import Subsection from '.~/wcdl/subsection';
...
+import { FILTER_ONBOARDING } from '.~/utils/tracks';
 import './index.scss';
 
 /**
  * Clicking on the button to connect an existing Google Ads account.
  *
  * @event gla_ads_account_connect_button_click
  * @property {number} id The account ID to be connected.
+ * @property {string} [context] Indicates the place where the button is located.
+ * @property {string} [step] Indicates the step in the onboarding process.
```
ianlin, line 33: Do you think we should also add `gla_mc_id` and `gla_version` props in the JSDoc so the [tracking document](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/README.md) could be more accurate?

js/src/components/google-ads-account-card/disconnect-account.js
```diff
@@ -9,10 +9,26 @@ import { useState, useCallback } from '@wordpress/element';
...
 import AppButton from '.~/components/app-button';
+import useEventPropertiesFilter from '.~/hooks/useEventPropertiesFilter';
+import { FILTER_ONBOARDING } from '.~/utils/tracks';
 
+/**
+ * Clicking on the button to disconnect the Google Ads account.
+ *
+ * @event gla_ads_account_disconnect_button_click
+ * @property {string} [context] Indicates the place where the button is located.
+ * @property {string} [step] Indicates the step in the onboarding process.
```
ianlin, line 20: Same as above, do you think it'd be better to add more props in the doc?

ianlin (approved):
Thanks for the [quick answer](https://github.com/woocommerce/google-listings-and-ads/pull/2421#discussion_r1623998588). Tested locally and it worked as expected. Just left some comments regarding the tracking JSDoc.

js/src/components/google-ads-account-card/connect-ads/index.js
eason9487: The `gla_ads_id` is one of the global event properties on the frontend, and the `id` of this event exists before the global event property. In order to avoid unnecessary inconsistencies in tracking data, the current preference is to leave it unchanged.

js/src/components/google-ads-account-card/connect-ads/index.js
eason9487: The current `doc:tracking` script (`woocommerce-grow-jsdoc`) doesn't support global event properties, and I'm not leaning toward duplicating global properties for the docs of all frontend tracking events.

It needs more discussion about how the `woocommerce-grow-jsdoc` tool should handle global event properties. I've first listed the global event properties directly at the beginning of the tracking README file. https://github.com/woocommerce/google-listings-and-ads/pull/2421/commits/0e6b33dd25faac87265467958a9cb8367ca4e09b

js/src/components/google-ads-account-card/disconnect-account.js
eason9487: See https://github.com/woocommerce/google-listings-and-ads/pull/2421#discussion_r1624242157

js/src/components/google-ads-account-card/connect-ads/index.js
ianlin: Got it, thanks for clarifying and adding the global event props in the tracking README. 👍

PR #2423, opened 2024-06-06
Fix fatal error when fetching marketing campaigns for Marketing Channel

jorgemd24:

When loading the Marketing -> Overview section, there is a request to fetch the campaigns, but it currently fails because the campaign types are not loaded. The campaign types are only loaded if the filter `woocommerce_gla_enable_mcm` returns true.

The fatal error is the below:

```
undefined array key "google-ads" in /opt/homebrew/var/www/wp-content/plugins/google-listings-and-ads/src/MultichannelMarketing/GLAChannel.php on line 195, referer: https://civet-pleasing-internally.ngrok-free.app/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing
[Thu Jun 06 16:19:22.787147 2024] [php:error] [pid 2819] [client ::1:56339] PHP Fatal error:  Uncaught TypeError: Automattic\\WooCommerce\\Admin\\Marketing\\MarketingCampaign::__construct(): Argument #2 ($type) must be of type Automattic\\WooCommerce\\Admin\\Marketing\\MarketingCampaignType, null given, called in /opt/homebrew/var/www/wp-content/plugins/google-listings-and-ads/src/MultichannelMarketing/GLAChannel.php on line 194 and defined in /opt/homebrew/var/www/wp-content/plugins/woocommerce-dev/plugins/woocommerce/src/Admin/Marketing/MarketingCampaign.php:68\nStack trace:\n#0 /opt/homebrew/var/www/wp-content/plugins/google-listings-and-ads/src/MultichannelMarketing/GLAChannel.php(194): Automattic\\WooCommerce\\Admin\\Marketing\\MarketingCampaign->__construct('21359526875', NULL, 'Campaign 2024-0...', 'https://civet-p...', Object(Automattic\\WooCommerce\\Admin\\Marketing\\Price))\
```
### Screenshots:

image

This PR prevents the error by checking if the campaign types have been enabled before fetching the campaigns.

1. Checkout `develop`.
2. Go to Marketing -> Overview.
3. Check your PHP errors or open the console and see how the campaign request fails.
4. Checkout this branch.
5. Refresh the Marketing -> Overview page.
6. See that we prevent the fatal error.
7. Add the following filter: add_filter( 'woocommerce_gla_enable_mcm', '__return_true' );
8 Refresh and see that now the campaigns are loaded. 

### Additional details:

Tests are failing because of this issue: https://github.com/woocommerce/google-listings-and-ads/pull/2415#issuecomment-2149368962 however you can run the tests locally and see if everything is OK

### Changelog entry

> Fix - Fatal error when loading campaign in the marketing overview section.

### DIFF
--- a/src/MultichannelMarketing/GLAChannel.php
+++ b/src/MultichannelMarketing/GLAChannel.php
@@ -78,11 +78,11 @@ class GLAChannel implements MarketingChannelInterface {
 	}
 
 	/**
-	 * Determines if the marketing channel campaigns is enabled.
+	 * Determines if the marketing channel campaigns types are enabled.
 	 *
 	 * @return bool
 	 */
-	protected function is_enable_campaign_types() {
+	protected function is_enable_campaign_types(): bool {
 		return apply_filters( 'woocommerce_gla_enable_mcm', false ) === true;
 	}
 
--- a/src/MultichannelMarketing/GLAChannel.php
+++ b/src/MultichannelMarketing/GLAChannel.php
@@ -72,17 +72,17 @@ class GLAChannel implements MarketingChannelInterface {
 		$this->product_sync_stats = $product_sync_stats;
 		$this->campaign_types     = [];
 
-		if ( $this->is_enable_campaign_types() ) {
+		if ( $this->is_mcm_enabled() ) {
 			$this->campaign_types = $this->generate_campaign_types();
 		}
 	}
 
 	/**
-	 * Determines if the marketing channel campaigns types are enabled.
+	 * Determines if the multichannel marketing is enabled.
 	 *
 	 * @return bool
 	 */
-	protected function is_enable_campaign_types(): bool {
+	protected function is_mcm_enabled(): bool {
 		return apply_filters( 'woocommerce_gla_enable_mcm', false ) === true;
 	}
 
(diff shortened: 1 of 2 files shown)


src/MultichannelMarketing/GLAChannel.php
```diff
@@ -72,11 +72,20 @@ public function __construct( MerchantCenterService $merchant_center, AdsCampaign
...
 		}
 	}
 
+	/**
+	 * Determines if the marketing channel campaigns types are enabled.
+	 *
+	 * @return bool
+	 */
+	protected function is_enable_campaign_types(): bool {
+		return apply_filters( 'woocommerce_gla_enable_mcm', false ) === true;
```
ianlin: 💅 Since the filter name is `woocommerce_gla_enable_mcm`, I'd think `is_multichannel_marketing_enabled()` or just `is_mcm_enabled()` would be more suitable.

ianlin (approved):
Thanks for catching this issue and fixing it. Tested locally, I could no longer see the fatal error and the campaigns were displayed correctly. I'm marking it approved and just left a non-blocker naming suggestion.

src/MultichannelMarketing/GLAChannel.php
jorgemd24: Thanks for the suggestion! I've added here: [1968a2f](https://github.com/woocommerce/google-listings-and-ads/pull/2423/commits/1968a2f18a25f1dad2687489e1df764199d57758)

PR #2424, opened 2024-06-07
[API Pull] Auth Flow

puntope:

- Auth flow for granting access to WPCOM after enabling the product sync feature
- Tweak the banner
- Revoke access when disabling the feature

1. Go to Settings and verify you see "We will soon transition to a new and improved method for synchronizing product data with Google -- Get early access" banner 
2. Click on Get early access (see the button loading)
3. You will be redirected to auth flow
4. Complete the auth flow
5. You will be redirected back to the store
6. You see "Google has been granted access to fetch your product data." in green on the Google MC card 
7. Click on "Disable product data fetch"
8. Auth is disabled and you see the banner again

### Additional details:

- Deeper unit Tests coming on further PRs
- Consistency and more tracking events coming in further PRs
- Notice that the errors are still not handled from Google. You can see that for example removing the nonce param in `get_data_from_google` function


src/API/Google/Middleware.php
```diff
@@ -649,4 +663,44 @@ public function is_subaccount(): bool {
 		// since transients don't support booleans, we save them as 0/1 and do the conversion here
 		return boolval( $is_subaccount );
 	}
+
+	/**
+	 * Performs a request to Google Shopping Data Integration (SDI) to get required information in order to form an auth URL.
+	 *
+	 * @return array An array with the JSON response from the WCS server.
+	 * @throws NotFoundExceptionInterface  When the container was not found.
+	 * @throws ContainerExceptionInterface When an error happens while retrieving the container.
```
jorgemd24, line 576: It can also throw an Exception: "Error Authenticating Google App..."

src/API/WP/OAuthService.php
```diff
@@ -91,31 +98,65 @@ public function get_auth_url( string $path ): string {
...
+		$args = [
+			'method'  => 'DELETE',
+			'timeout' => 30,
+			'url'     => $this->get_wpcom_api_url( '/wpcom/v2/sites/' . Jetpack_Options::get_option( 'id' ) . '/wc/partners/google/revoke-token' ),
+			'user_id' => get_current_user_id(),
+		];
+
+		$response = Client::remote_request( $args );
+
+		if ( is_wp_error( $response ) ) {
```
jorgemd24: Shouldn't we check if the HTTP status code is >= 200? For example, if there are no tokens, the server will respond with status 500 - Internal Server Error. - `{"code":"wpcom_partner_token_not_associated","message":"No token found associated with the client ID and user.","data":null}`

tests/Unit/API/WP/OAuthServiceTest.php
```diff
@@ -3,9 +3,14 @@
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\API\WP;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Ads;
```
jorgemd24: Unused dependency

jorgemd24 (commented):
Thanks @puntope for working on this. I tested it along with https://github.com/Automattic/woocommerce-connect-server/pull/2485 , and it worked well except that after authorizing the WPCOM App, it seems Google is redirecting to the wrong URL.

image

However, the OAuth flow was successful, as I could see the GLA App connected to my site. I then went to the settings to disconnect the API pull, and it successfully revoked the token.

image

I left some minor comments about the code too.

ianlin:
> after authorizing the WPCOM App, it seems Google is redirecting to the wrong URL.

Had a look at it and I found they tried to redirect us to this URL:

`https://shoppingdataintegration.googleapis.com/v1/credentials/partners/WooCommerce/https%3A%2F%2Ftest-site.example.com%2Fwp-admin%2Fadmin.php%3Fpage%3Dwc-admin%26path%3D%2Fgoogle%2Fsettings%3Fnonce=XXXXX&google_wpcom_app_status=approved`

We should ask them to remove `https://shoppingdataintegration.googleapis.com/v1/credentials/partners/WooCommerce/` so they could redirect to the correct URL.

However I also noticed there were something wrong in the merchant site's URL:

The URL decoded version of the merchant site's URL is:

`https://test-site.example.com/wp-admin/admin.php?page=wc-admin&path=/google/settings?nonce=XXXXX&google_wpcom_app_status=approved`

where it added a question mark `?` before `nonce` parameter which it would cause the page not found. We should also ask them the replace that `?` with `&` so it would return to the merchant site correctly.

src/API/Google/Middleware.php
puntope: Fixed here thx. https://github.com/woocommerce/google-listings-and-ads/pull/2424/commits/5b9a462e304af83a9332114bd12271eb1c4896a4

src/API/WP/OAuthService.php
puntope: Fixed here. Thx https://github.com/woocommerce/google-listings-and-ads/pull/2424/commits/ece73b1dbf2b219230bbac33c5f607dda685a0eb

tests/Unit/API/WP/OAuthServiceTest.php
puntope: Fixed here thx. https://github.com/woocommerce/google-listings-and-ads/pull/2424/commits/436881c453d613fce326d5599a38cd33148b7850

puntope:
Thanks, @ianlin and @jorgemd24 for your suggestions and investigation. They fixed the error from Google and now seems to work. Also, I addressed the review comments. Can you test again?

jorgemd24:
@puntope, I tested it, and it is now redirecting me back to my site. Is there anything else we need to wait for, or can I proceed with the review? Also, I noticed some PHPCS errors.

puntope:
> @puntope, I tested it, and it is now redirecting me back to my site. Is there anything else we need to wait for, or can I proceed with the review? Also, I noticed some PHPCS errors.

We need to wait for a fix from google. I will ping you

puntope:
@jorgemd24 @ianlin This seems to be ready for another look. Google deployed their last changes

src/API/WP/OAuthService.php
```diff
@@ -91,31 +98,72 @@ public function get_auth_url( string $path ): string {
...
+			'method'  => 'DELETE',
+			'timeout' => 30,
+			'url'     => $this->get_wpcom_api_url( '/wpcom/v2/sites/' . Jetpack_Options::get_option( 'id' ) . '/wc/partners/google/revoke-token' ),
+			'user_id' => get_current_user_id(),
+		];
+
+		$request = Client::remote_request( $args );
+
+		if ( is_wp_error( $request ) ) {
+			throw new Exception( $request->get_error_message(), $request->get_error_code() );
```
jorgemd24: The Exception expects the second parameter to be an `int`, but wp_error accepts the code as either a string or int so if wp_error had set the code as a string it will throw a fatal.

src/API/WP/OAuthService.php
```diff
@@ -91,31 +98,72 @@ public function get_auth_url( string $path ): string {
...
+
+	/**
+	 * Calls an API by Google via WCS to get required information in order to form an auth URL.
 	 *
 	 * @return array{client_id: string, redirect_uri: string, nonce: string} An associative array contains required information that is retrived from Google.
 	 * client_id:    Google's WPCOM app client ID, will be used to form the authorization URL.
 	 * redirect_uri: A Google's URL that will be redirected to when the merchant approve the app access. Note that it needs to be matched with the Google WPCOM app client settings.
 	 * nonce:        A string returned by Google that we will put it in the auth URL and the redirect_uri. Google will use it to verify the call.
 	 */
 	protected function get_data_from_google(): array {
```
jorgemd24, line 130: `get_sdi_auth_params` can return an exception.

src/API/Google/Middleware.php
```diff
@@ -552,4 +567,45 @@ public function is_subaccount(): bool {
...
+	 * @see google-sdi in google/services inside WCS
+	 */
+	public function get_sdi_auth_params() {
+		try {
+			/** @var Client $client */
+			$client   = $this->container->get( Client::class );
+			$result   = $client->get( $this->get_sdi_auth_endpoint() );
+			$response = json_decode( $result->getBody()->getContents(), true );
+
+			if ( 200 !== $result->getStatusCode() ) {
```
jorgemd24, line 587: I simulated a response other than 200, but I don't see any error and I get redirected to: `/wp-admin/undefined`

jorgemd24 (commented):
Thanks @puntope for working on this. I tested the happy path and everything works well. However, if the response from WCS is anything other than 200, I get redirected to `wp-admin/undefined`.

jorgemd24:
Also, what do you think about setting the default value of this filter to `true`? Or do we want the merchant to enable this manually?

https://github.com/woocommerce/google-listings-and-ads/blob/ec1e862ee00e0112765b7e8bee76300dadc168b6/src/API/WP/NotificationsService.php#L176

jorgemd24:
One more thought: how will Google return any errors through the OAuth flow, and how are we going to handle them?

puntope:
> Or do we want the merchant to enable this manually?

If nothing has changed for now I think the first version requires setting the filter. But not sure if that changed after dark launch discussion

puntope:
> One more thought: how will Google return any errors through the OAuth flow, and how are we going to handle them?

The discussion between @ianlin and Google was that they would return `google_wpcom_app_status=error`. Can you confirm @ianlin ?

src/API/Google/Middleware.php
puntope: Hi @jorgemd24 How did you simulate the bad response? 

When I do so I cannot reproduce the error you mentioned.


https://github.com/woocommerce/google-listings-and-ads/assets/5908855/42093008-af4d-4e64-8c9b-6acd4bcbaa14

I did the simulation doing:

```php
$result = new Response(400 );
$response = [];
```

src/API/WP/OAuthService.php
puntope: Tweaked here. I will return just 400

https://github.com/woocommerce/google-listings-and-ads/pull/2424/commits/e42f728d08417864b8988384cc538a211fa9212e

src/API/WP/OAuthService.php
puntope: tweaked here

https://github.com/woocommerce/google-listings-and-ads/pull/2424/commits/e42f728d08417864b8988384cc538a211fa9212e

puntope:
Thanks @jorgemd24. Can you take another look ?

src/API/Google/Middleware.php
jorgemd24: Ah sorry! I realized that my simulation was not correct, because I changed this line from `if ( 200 !== $result->getStatusCode() ) {` to `if ( 200 === $result->getStatusCode() ) {` but then the Exception was thrown as 200 and that's why I was getting that result.

jorgemd24 (approved):
Thanks for the adjustments, LGTM 👍 

> If nothing has changed for now I think the first version requires setting the filter. But not sure if that changed after dark launch discussion

Okay I will ask them, for me it looks a bit redundant.

> The discussion between @ianlin and Google was that they would return google_wpcom_app_status=error. Can you confirm @ianlin ?

If so, I can work on a PR to display those errors to the user.

puntope:
> Thanks for the adjustments, LGTM 👍
> 
> > If nothing has changed for now I think the first version requires setting the filter. But not sure if that changed after dark launch discussion
> 
> Okay I will ask them, for me it looks a bit redundant.

Sounds good. @jorgemd24  Go ahead and we can remove if so. 

> 
> > The discussion between @ianlin and Google was that they would return google_wpcom_app_status=error. Can you confirm @ianlin ?
> 
> If so, I can work on a PR to display those errors to the user.

@jorgemd24 We do display a notice on the card banner and allow granting again. Is this what you mean? 

p.s You can also check the tracking topic if you are looking for issues.

ianlin:
> One more thought: how will Google return any errors through the OAuth flow, and how are we going to handle them?

> The discussion between @ianlin and Google was that they would return google_wpcom_app_status=error. Can you confirm @ianlin ?

@puntope @jorgemd24 Yeah correct, they will just set the query param to `google_wpcom_app_status=error` without what error it is.

PR #2430, opened 2024-06-13
API Pull - Revoke access if plugin is deactivated

jorgemd24:

Currently, if the plugin is deactivated, the WPCOM token is not revoked, allowing the partner to continue querying the API. This PR ensures that the WPCOM token is revoked when the plugin is deactivated.

1. Checkout this branch.
2. In WooCommerce Connect Server, checkout this PR: https://github.com/Automattic/woocommerce-connect-server/pull/2485 
3. Add the following filter in GLA extension: `add_filter( 'woocommerce_gla_notifications_enabled', '__return_true' );`
4. Go to Settings and Enable the API Pull mechanism:
image
5. The redirection of the OAuth flow [will fail](https://github.com/woocommerce/google-listings-and-ads/pull/2424#pullrequestreview-2111011892) to redirect back to your site however the App will be connected in `https://wordpress.com/me/security/connected-applications`

image

6. Once you confirmed that the app shows as connected go to Plugins->Installed Plugins and deactivate GLA
7. Go to `https://wordpress.com/me/security/connected-applications` and that the App is not connected anymore.

### Changelog entry

>


src/Proxies/Jetpack.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Proxies;
+
+use Automattic\Jetpack\Connection\Client;
+
+/**
+ * Class JP.
+ *
+ * This class provides proxy methods to wrap around Jetpack functions.
+ *
```
ianlin, line 12: Would be great to add `@since x.x.x` here.

ianlin (approved):
Thanks for adding this. Tested locally and confirm the Google WPCOM App was revoked after deactivating GL&A plugin. I just left a comment regarding adding `@since` tag.

PR #2436, opened 2024-06-18
[API Pull] Add language param in settings endpoint

puntope:

This PR adds the site language as param when requesting `GET /wp-json/wc/v3/settings/general?gla_syncable=1`

### Screenshots:

Screenshot 2024-06-18 at 19 22 09

1. Request GET /wp-json/wc/v3/settings/general?gla_syncable=1
2. See the language param appears with your site language

### DIFF
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -184,8 +184,8 @@ class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
 					];
 
 					$data[] = [
-						'id'    => 'language',
-						'label' => 'Store language',
+						'id'    => 'gla_language',
+						'label' => 'Google Listings and Ads: Store language',
 						'value' => get_locale(),
 					];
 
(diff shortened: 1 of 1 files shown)


src/Integration/WPCOMProxy.php
```diff
@@ -183,6 +183,12 @@ function ( $response, $handler, $request ) {
 						'value' => $this->shipping_time_query->get_all_shipping_times(),
 					];
 
+					$data[] = [
+						'id'    => 'language',
```
jorgemd24: Wouldn't it be better to add a prefix to this id? I'm concerned that the language is too generic and we might end up overriding this field if WC adds it in the future.

jorgemd24 (commented):
Thanks, @puntope, for working on this. It is working well, but I have doubts about the id we're using. See my comment.

src/Integration/WPCOMProxy.php
puntope: Updated here. https://github.com/woocommerce/google-listings-and-ads/pull/2436/commits/719b9a9f08ae9c7b1c46f296b4ff09b90442a2b5

puntope:
Hi @jorgemd24  Thx for the review. I addressed your suggestions. Can you take a look?

jorgemd24 (approved):
Thanks @puntope, LGTM 👍

PR #2437, opened 2024-06-20
Fix issue with inconsistent metadata format

jorgemd24:

Part of #2146 

When filtering the private metadata, we use the unset function to remove the private metadata from the response. However, this alters the final response format, changing it from an array of objects to a single object. For example:

Incorrect:
```
"meta_data": {
            "1": {
                "id": 241,
                "key": "_wc_gla_visibility",
                "value": "sync-and-show"
            },
}
```
Correct:
```
"meta_data": [
          {
                "id": 241,
                "key": "_wc_gla_visibility",
                "value": "sync-and-show"
            },
]
```

1. Choose one product and set the channel visibility to "Sync & Show". Add some private metadata to the product (private metadata = metadata starting with `_`, for example `_my_private_data`).
2. Create some WC REST API credentials and do the following request GET `wp-json/wc/v3/products/YOUR_PRODUCT_ID?gla_syncable=1`
3. Without checking this PR, see how the metadata format is incorrect.
4. Checkout this PR and see the expected response.

### Changelog entry

>

### DIFF
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -322,6 +322,8 @@ class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
 			}
 		}
 
+		$data['meta_data'] = array_values( $data['meta_data'] );
+
 		$response->set_data( $data );
 
 		return $response;


puntope (approved):
Thanks @jorgemd24 I can confirm now meta_data is an array of objects. 

Screenshot 2024-06-20 at 16 24 04

PR #2443, opened 2024-06-25
Change connection test page capabilities to manage_woocommerce

jorgemd24:

Part of #2146 

Currently, the connection test page is only available to the administrator because it is registered with the `manage_options` capability. However, all our endpoints are checking the `manage_woocommerce` capability.

https://github.com/woocommerce/google-listings-and-ads/blob/cbcff08cf1270def397a0317d55316c991c5efd2/src/API/PermissionsTrait.php#L20-L22

This makes it harder for a Shop Manager to debug any issues with the new approach as the connection test page is not available. Therefore, this PR updates the capability to `manage_woocommerce`. Since all actions available on the connection test page are also accessible through the UI for users with `manage_woocommerce` capability I don't think we are giving new privileges to the Shop Manager role.

1. Log in to wp-admin with a Shop Manager user role.
2. Add this filter `add_filter('woocommerce_gla_enable_connection_test','__return_true');` or navigate to `wp-admin/admin.php?page=connection-test-admin-page`
3. Check that you can access the Connection Test Page.

### Additional details:
Not sure why but the phpcs were failing for the account services test, so I fixed it in this PR. 

### Changelog entry

>

### DIFF
--- a/src/ConnectionTest.php
+++ b/src/ConnectionTest.php
@@ -98,7 +98,7 @@ class ConnectionTest implements Service, Registerable {
 			add_menu_page(
 				'Connection Test',
 				'Connection Test',
-				'manage_options',
+				'manage_woocommerce',
 				'connection-test-admin-page',
 				function() {
 					$this->render_admin_page();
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Thanks for the change. I can confirm that the connection page is accessible when using the shop manager account.

PR #2449, opened 2024-07-03
[API Pull] Handle exception when deactivating the plugin and the WPCOM token does not exist.

jorgemd24:

Part of #2146 .

Currently, we throw an exception if the request to revoke the token fails, such as when the WPCOM token is not found and therefore can't be deleted. This can lead to a fatal error when deactivating the plugin, preventing its deactivation. Since not all merchants will opt into the new mechanism, the scenario where the WPCOM token is missing is possible. To prevent this fatal error, I think it's better to catch the exception and log the error.

Additionally, I've removed the "Revoke WPCOM Access token" option from the Connection Test page. We now have a similar option, "WPCOM REST API Status" -> "Disconnect," which has the same purpose, making the option redundant.

image

1. Revoke any WPCOM token, you can use the connection test page if you already have one WPCOM token.
2. Deactivate the plugin.
3. See that we prevent the fatal error and log the error. 

### Changelog entry

>

### DIFF
--- a/src/API/WP/OAuthService.php
+++ b/src/API/WP/OAuthService.php
@@ -180,6 +180,11 @@ class OAuthService implements Service, OptionsAwareInterface, Deactivateable, Co
 	 * Revoke token on deactivation.
 	 */
 	public function deactivate(): void {
-		$this->revoke_wpcom_api_auth();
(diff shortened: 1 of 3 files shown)


puntope (approved):
✅ LGTM

PR #2450, opened 2024-07-03
[API Pull] Tweak Variations Notifications

puntope:

This PR tweaks the logic for notifications when handling variations. In particular:

- FIXED: We were sending create notifications for variations when the parent was not published still
- FIXED: We were only sending delete notifications for the parent when a variation was deleted.
- FIXED: SOme unhandled exceptions when we try to access. a product that is not in the database anymore. 

1. Checkout PR
2. Test that creation/update/deletion of simple product works as always
3. Create a Variable Product and generate variations for It 
4. Go to Woo - Status - Scheduled actions - That step should not schedule any notifications
5. Asign price to some of the variations and publish the product
6. Go to Woo - Status - Scheduled actions - See scheduled notifications create for the variations with price + the variable product
7. Run the actions and see if they are logged in the gla log under Woo - Status - logs
8. Edit one of the variations without price
9. It doesn't schedule or log anything as it was not created yet
10. Remove one of the variations without price
11.  It doesn't schedule or log anything as it was not created yet
12. Edit one of the variations with price 
13. Go to Woo - Status - Scheduled actions - See scheduled notifications update for the variations with price + the variable product. Run them
14. Remove the variation with price 
15. Go to Woo - Status - Scheduled actions - See scheduled notifications update for the variations with price + the variable product and one delete notification in the logs (directly without schedule) for the one you deleted with offer_id
16. Set the variable product as DONT SYNC AND SHOW
17.  Go to Woo - Status - Scheduled actions - See scheduled notifications delete for the variations with price + the variable product 
18. Before running it. Go back and set the variable product as  SYNC AND SHOW
19. Go to Woo - Status - Scheduled actions - See scheduled notifications update for the variations with price + the variable product 
20. Run the delete notifications of the step 17
21. See in the logs they are not logged (a filter prevented it, because we set is as visible in step 18)
22. Run the tasks from step 19
23. See them logged in the log
24. Try to delete products and variations in different ways to see if everything is handled properly.

### Changelog entry

>

### DIFF
--- a/src/Jobs/Notifications/AbstractItemNotificationJob.php
+++ b/src/Jobs/Notifications/AbstractItemNotificationJob.php
@@ -55,6 +55,7 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 	 *
 	 * @param int    $item_id
 	 * @param string $status
+	 * @throws InvalidValue If the given ID doesn't reference a valid product.
 	 */
 	protected function set_status( int $item_id, string $status ): void {
 		$item = $this->get_item( $item_id );
@@ -84,6 +85,7 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 	 *
 	 * @param int    $item_id
 	 * @param string $topic
+	 * @throws InvalidValue If the given ID doesn't reference a valid product.
 	 * @return bool
 	 */
 	protected function can_process( int $item_id, string $topic ): bool {
@@ -103,6 +105,7 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 	 *
 	 * @param string $topic
 	 * @param int    $item
+	 * @throws InvalidValue If the given ID doesn't reference a valid product.
 	 */
 	protected function handle_notified( string $topic, int $item ): void {
 		if ( $this->is_delete_topic( $topic ) ) {
--- a/src/Jobs/Notifications/AbstractItemNotificationJob.php
+++ b/src/Jobs/Notifications/AbstractItemNotificationJob.php
@@ -48,7 +48,6 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 				__METHOD__
 			);
 		}
-
 	}
 
 	/**
--- a/src/Jobs/Notifications/AbstractItemNotificationJob.php
+++ b/src/Jobs/Notifications/AbstractItemNotificationJob.php
@@ -3,6 +3,7 @@ declare(strict_types=1);
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Jobs\Notifications;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\Exception\InvalidValue;
 use Automattic\WooCommerce\GoogleListingsAndAds\Value\NotificationStatus;
 
 defined( 'ABSPATH' ) || exit;
@@ -35,10 +36,19 @@ abstract class AbstractItemNotificationJob extends AbstractNotificationJob {
 		$topic = $args['topic'];
 		$data  = $args['data'] ?? [];
 
-		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item, $data ) ) {
-			$this->set_status( $item, $this->get_after_notification_status( $topic ) );
-			$this->handle_notified( $topic, $item );
+		try {
+			if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item, $data ) ) {
+				$this->set_status( $item, $this->get_after_notification_status( $topic ) );
+				$this->handle_notified( $topic, $item );
+			}
+		} catch ( InvalidValue $e ) {
+			do_action(
+				'woocommerce_gla_error',
+				sprintf( 'Error sending Notification for - Item ID: %s - Topic: %s - Data %s. Product was deleted from the database before the notification was sent.', $item, $topic, json_encode( $data ) ),
+				__METHOD__
+			);
 		}
+
 	}
 
 	/**
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -398,7 +398,7 @@ class ProductHelper implements Service, HelperNotificationInterface {
 			in_array( $product->get_type(), ProductSyncer::get_supported_product_types(), true );
 
 		if ( $is_ready && $product instanceof WC_Product_Variation ) {
-			$parent = $this->maybe_swap_for_parent( $product );
+			$parent   = $this->maybe_swap_for_parent( $product );
 			$is_ready = $this->is_ready_to_notify( $parent );
 		}
 
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -765,7 +765,7 @@ class ProductHelper implements Service, HelperNotificationInterface {
 	 *
 	 * @return string The offer id
 	 */
-	public function get_offer_id( $product_id ) {
+	public function get_offer_id( int $product_id ) {
 		return WCProductAdapter::get_google_product_offer_id( $this->get_slug(), $product_id );
 	}
 }
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -397,6 +397,11 @@ class ProductHelper implements Service, HelperNotificationInterface {
 			$product->get_status() === 'publish' &&
 			in_array( $product->get_type(), ProductSyncer::get_supported_product_types(), true );
 
+		if ( $is_ready && $product instanceof WC_Product_Variation ) {
+			$parent = $this->maybe_swap_for_parent( $product );
+			$is_ready = $this->is_ready_to_notify( $parent );
+		}
+
 		/**
 		 * Allow users to filter if a product is ready to notify.
 		 *
@@ -750,5 +755,17 @@ class ProductHelper implements Service, HelperNotificationInterface {
 	public function get_categories( WC_Product $product ): array {
 		$terms = get_the_terms( $product->get_id(), 'product_cat' );
 		return ( empty( $terms ) || is_wp_error( $terms ) ) ? [] : wp_list_pluck( $terms, 'name' );
+	}
+
+	/**
+	 * Get the offer id for a product
+	 *
+	 * @since x.x.x
+	 * @param int $product_id The product id to get the offer id.
+	 *
+	 * @return string The offer id
+	 */
+	public function get_offer_id( $product_id ) {
+		return WCProductAdapter::get_google_product_offer_id( $this->get_slug(), $product_id );
 	}
 }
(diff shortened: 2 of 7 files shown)


src/Jobs/Notifications/AbstractItemNotificationJob.php
```diff
@@ -35,9 +36,17 @@ protected function process_items( array $args ): void {
...
+			if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item, $data ) ) {
+				$this->set_status( $item, $this->get_after_notification_status( $topic ) );
+				$this->handle_notified( $topic, $item );
+			}
+		} catch ( InvalidValue $e ) {
+			do_action(
+				'woocommerce_gla_error',
+				sprintf( 'Error sending Notification for - Item ID: %s - Topic: %s - Data %s. Product was deleted from the database before the notification was sent.', $item, $topic, json_encode( $data ) ),
+				__METHOD__
+			);
```
puntope, line 49: This handles the potential case in which we attempt to get an already deleted item from the database.

src/Product/ProductHelper.php
```diff
@@ -397,6 +397,11 @@ public function is_ready_to_notify( WC_Product $product ): bool {
 			$product->get_status() === 'publish' &&
 			in_array( $product->get_type(), ProductSyncer::get_supported_product_types(), true );
 
+		if ( $is_ready && $product instanceof WC_Product_Variation ) {
+			$parent   = $this->maybe_swap_for_parent( $product );
+			$is_ready = $this->is_ready_to_notify( $parent );
+		}
```
puntope, line 403: This is for checking the status of the parent also. Not just the variation

src/Product/SyncerHooks.php
```diff
@@ -285,13 +286,13 @@ protected function handle_update_product_notification( WC_Product $product ) {
...
-				]
-			);
+			$this->schedule_delete_notification( $product );
+		}
+
+		if ( $product instanceof WC_Product_Variable ) {
+			foreach ( $product->get_available_variations( 'objects' ) as $variation ) {
+				$this->handle_update_product_notification( $variation );
+			}
 		}
```
puntope: Handle also the variations when the variable changes

src/Product/SyncerHooks.php
```diff
@@ -315,24 +316,35 @@ protected function handle_delete_product( int $product_id ) {
...
-			$this->product_helper->set_notification_status( $product, NotificationStatus::NOTIFICATION_PENDING_DELETE );
-			$this->product_notification_job->schedule(
-				[
-					'item_id' => $product->get_id(),
-					'topic'   => NotificationsService::TOPIC_PRODUCT_DELETED,
-				]
-			);
+
+		if ( $product instanceof WC_Product && $this->product_helper->has_notified_creation( $product ) ) {
+			$this->notifications_service->notify( NotificationsService::TOPIC_PRODUCT_DELETED, $product_id, [ 'offer_id' => $this->product_helper->get_offer_id( $product_id ) ] );
```
puntope: For post deletions, it is better to send the notification directly.

src/Jobs/Notifications/AbstractItemNotificationJob.php
```diff
@@ -35,9 +36,17 @@ protected function process_items( array $args ): void {
 		$topic = $args['topic'];
 		$data  = $args['data'] ?? [];
 
-		if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item, $data ) ) {
-			$this->set_status( $item, $this->get_after_notification_status( $topic ) );
-			$this->handle_notified( $topic, $item );
+		try {
+			if ( $this->can_process( $item, $topic ) && $this->notifications_service->notify( $topic, $item, $data ) ) {
+				$this->set_status( $item, $this->get_after_notification_status( $topic ) );
```
jorgemd24, line 41: Should we update the PHPDocs to `set_status` and the other functions it calls to indicate that this method can throw an "InvalidValue" exception? Otherwise, it's hard to know which methods might throw the exception. From what I can see, the exception will be thrown here:

https://github.com/woocommerce/google-listings-and-ads/blob/1b922963dd28c1b2dbc39875f965228c09b736c7/src/Jobs/Notifications/ProductNotificationJob.php#L70-L72

src/Product/ProductHelper.php
```diff
@@ -751,4 +756,16 @@ public function get_categories( WC_Product $product ): array {
...
+
+	/**
+	 * Get the offer id for a product
+	 *
+	 * @since x.x.x
+	 * @param int $product_id The product id to get the offer id.
+	 *
+	 * @return string The offer id
+	 */
+	public function get_offer_id( $product_id ) {
```
jorgemd24: The rest of the file includes argument types. Should we add the type for `$product_id` as well?

src/Product/SyncerHooks.php
```diff
@@ -285,13 +286,13 @@ protected function handle_update_product_notification( WC_Product $product ) {
...
 			);
 		} elseif ( $this->product_helper->should_trigger_delete_notification( $product ) ) {
-			$this->product_helper->set_notification_status( $product, NotificationStatus::NOTIFICATION_PENDING_DELETE );
-			$this->product_notification_job->schedule(
-				[
-					'item_id' => $product->get_id(),
-					'topic'   => NotificationsService::TOPIC_PRODUCT_DELETED,
-				]
-			);
+			$this->schedule_delete_notification( $product );
```
jorgemd24, line 300: Not needed for this PR but maybe we could have similar methods for `schedule_update_notification`and `schedule_create_notification`

src/Product/SyncerHooks.php
jorgemd24: Why is it better to send the notification directly instead of scheduling the job?

jorgemd24 (commented):
Hi @puntope, thanks for fixing this. However, I found some issues:

- When I trash a simple product, we don't scheduled the DELETE topic and it is aligned with your comment here: https://github.com/woocommerce/google-listings-and-ads/pull/2450#discussion_r1664271043 but when I restore the simple product, we send an UPDATE topic instead of a CREATE topic.

>Edit one of the variations without price
>It doesn't schedule or log anything as it was not created yet
>Remove one of the variations without price
>It doesn't schedule or log anything as it was not created yet
>Edit one of the variations with price

Not sure if I'm doing this right or understanding this part, but I see notifications scheduled for all the variations with prices, as well as for the parent.

> Edit one of the variations with price
> Go to Woo - Status - Scheduled actions - See scheduled notifications update for the variations with price + the variable product. Run them

I am unsure if this is ideal because we only updated one variation. Will Google make X (number of variations + parent product) REST API requests? It seems unnecessary to make so many requests. For example, if a product has 30 variations and only 1 is updated, we'll schedule 30 + 1 jobs, and the merchant will receive 31 API requests?

Also, I left some comments about the code, and it doesn't seem like we've added tests to check which variations need to be notified.

src/Product/SyncerHooks.php
puntope: Because otherwise, it can be the case that the item is not in the database (for example, the user trash the product and deletes it from the trash before the action runs). In this cases an exception might happen or we cannot get anymore the product props.

puntope:
> * but when I restore the simple product, we send an UPDATE topic instead of a CREATE topic.
> 
> >

Good catch. Fixed here. https://github.com/woocommerce/google-listings-and-ads/pull/2450/commits/413c7c54312d0b2fa76b2b708d61016af9463540

puntope:
> I am unsure if this is ideal because we only updated one variation. Will Google make X (number of variations + parent product) REST API requests? It seems unnecessary to make so many requests. For example, if a product has 30 variations and only 1 is updated, we'll schedule 30 + 1 jobs, and the merchant will receive 31 API requests?

Seems like when we did activate the MC Push again we forgot to handle that
https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Product/SyncerHooks.php#L180

I did some tweaks for it also. Now should be trigger per variation and parent only. 

See https://github.com/woocommerce/google-listings-and-ads/pull/2450/commits/ff8468e3daadcdbb0b57214b2967b6c4e39ed284

src/Product/SyncerHooks.php
puntope: For now, we are only calling those in that place. `schedule_delete_notification` was refactored to avoid code duplication.

src/Product/ProductHelper.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2450/commits/9e041f54035c4cb8860520ac3a16e77745c3fca1

puntope:
Thanks @jorgemd24 for all your feedback. I applied some changes. Can you take another look?

src/Product/SyncerHooks.php
```diff
@@ -206,29 +206,31 @@ public function duplicate_product_exclude_meta( array $exclude_meta ): array {
...
-	protected function handle_update_products( array $products ) {
+	protected function handle_update_products( array $products, $notify = true ) {
 		$products_to_update = [];
 		$products_to_delete = [];
 
-		if ( $this->notifications_service->is_ready() ) {
-			$this->handle_update_product_notification( $products[0] );
-		}
-
 		foreach ( $products as $product ) {
```
jorgemd24, line 215: At this point, everything should be an instance of `WC_Product`. However, as an extra precaution, should we add a check like this before continuing with the loop: `if (!$product instanceof WC_Product) continue;`? I meant in these lines: https://github.com/woocommerce/google-listings-and-ads/blob/96f16b9745e4344d648c517f91d2da7c26ca83ff/src/Product/SyncerHooks.php#L213-L223

src/Product/SyncerHooks.php
jorgemd24: Thanks for the explanation. However, this could be problematic because if you delete multiple products or a variable product with many variations (say 30), we'll end up making all the requests simultaneously. From the merchant's perspective, it will take a while to delete a product with multiple variations because all the requests need to be completed. Also, we could reach the API rate limit.

As far as I can tell, we only need to notify the product ID, topic, and offer_id, so can't we just schedule the job to notify that? Or do we need the actual product?

jorgemd24 (commented):
Hi @puntope, thanks for the adjustments. It is working well and I'm getting the right results. However, I have one concern about making the request to the Notification endpoint directly instead of scheduling the job when deleting the item. [See my comment](https://github.com/woocommerce/google-listings-and-ads/pull/2450#discussion_r1666428519).

Let me know if you would prefer to address this in this PR or a follow-up PR.

Thanks!

puntope:
> As far as I can tell, we only need to notify the product ID, topic, and offer_id, so can't we just schedule the job to notify that? Or do we need the actual product?

The logic know uses get_item that needs the product from DB. We can refactor that in future iterations. Since DarkL is next week I would suggest to have it working even if the performance is not the optimal. 

p.s I also added the check for WC_Product. This is ready for a new round.

jorgemd24 (approved):
Thanks for the adjustments is looking good.

> The logic know uses get_item that needs the product from DB. We can refactor that in future iterations. Since DarkL is next week I would suggest to have it working even if the performance is not the optimal.

Yes I think we should give priority to this part as it can degrade the performance when deleting multiple products.

PR #2452, opened 2024-07-04
[API Pull] Add Track Events to the Auth API responses

jorgemd24:

Part of #2146 .

Currently, we have two tracking events, `gla_disable_product_sync_click` and `gla_enable_product_sync_click`. However, it would also be useful to track the status of the Auth flow and when merchants revoke access.

This PR adds the following server-side event tracking:

- `revoke_wpcom_api_authorization` -> When the token is revoked, this event will include the status, blog id, and any error if it exists.
- `update_wpcom_api_authorization` -> When the Auth flow is completed, this event will include the status, blog id, and any error if it exists.

1. Confirm tracking is allowed in WooCommerce > Settings > Advanced > Woo.com
2. To view the tracking, go to Tracks -> History (live tracking doesn't show the properties). Filter by your username and the following events: `wcadmin_gla_update_wpcom_api_authorization` and `wcadmin_gla_revoke_wpcom_api_auth`. It may take some time for the data to appear.
3. Alternatively you can use the following filter to see the events in the error log:

```
add_filter(
	'pre_http_request',
	function ( $pre, $args, $url ) {
		if ( str_starts_with( $url, 'https://pixel.wp.com' ) ) {
			wp_parse_str( wp_parse_url( $url, PHP_URL_QUERY ), $query_vars );
			error_log( json_encode( $query_vars, JSON_PRETTY_PRINT ) );
		}
		return $pre;
	},
	10,
	3
);
```
Ref: https://github.com/woocommerce/google-listings-and-ads/pull/2207

4. Enable the notification service: `add_filter( 'woocommerce_gla_notifications_enabled','__return_true' )`;
5. Go to GLA -> Settings -> Get Early Access -> follow the steps.
6. Revoke the access by clicking in : `Disable product data fetch`:

image

 

### Additional details:

- I included the `blog_id` because I thought it might help with debugging issues, but I can't see it in Tracks. I'm not sure why that is.

### Changelog entry

>

### DIFF
--- a/src/API/WP/OAuthService.php
+++ b/src/API/WP/OAuthService.php
@@ -167,8 +167,9 @@ class OAuthService implements Service, OptionsAwareInterface, Deactivateable, Co
 				'woocommerce_gla_track_event',
 				'revoke_wpcom_api_auth',
 				[
-					'status' => 'error',
-					'error'  => $request->get_error_message(),
+					'status'  => 'error',
+					'error'   => $request->get_error_message(),
+					'blog_id' => Jetpack_Options::get_option( 'id' ),
 				]
 			);
 
@@ -188,8 +189,9 @@ class OAuthService implements Service, OptionsAwareInterface, Deactivateable, Co
 					'woocommerce_gla_track_event',
 					'revoke_wpcom_api_auth',
 					[
-						'status' => $status,
-						'error'  => $message,
+						'status'  => $status,
+						'error'   => $message,
+						'blog_id' => Jetpack_Options::get_option( 'id' ),
 					]
 				);
 
@@ -203,7 +205,8 @@ class OAuthService implements Service, OptionsAwareInterface, Deactivateable, Co
 				'woocommerce_gla_track_event',
 				'revoke_wpcom_api_auth',
 				[
-					'status' => 200,
+					'status'  => 200,
+					'blog_id' => Jetpack_Options::get_option( 'id' ),
 				]
 			);
 
--- a/src/API/WP/OAuthService.php
+++ b/src/API/WP/OAuthService.php
@@ -164,15 +164,15 @@ class OAuthService implements Service, OptionsAwareInterface, Deactivateable, Co
 			 * When the WPCOM token has been revoked with errors.
 			 *
 			 * @event revoke_wpcom_api_authorization
-			 * @property string status The status of the request.
+			 * @property int status The status of the request.
 			 * @property string error The error message.
-			 * @property mixed blog_id The blog ID.
+			 * @property int|null blog_id The blog ID.
 			 */
 			do_action(
 				'woocommerce_gla_track_event',
 				'revoke_wpcom_api_authorization',
 				[
-					'status'  => 'error',
+					'status'  => 400,
 					'error'   => $request->get_error_message(),
 					'blog_id' => Jetpack_Options::get_option( 'id' ),
 				]
@@ -192,9 +192,9 @@ class OAuthService implements Service, OptionsAwareInterface, Deactivateable, Co
 				* When the WPCOM token has been revoked with errors.
 				*
 				* @event revoke_wpcom_api_authorization
-				* @property string status The status of the request.
+				* @property int status The status of the request.
 				* @property string error The error message.
-				* @property mixed blog_id The blog ID.
+				* @property int|null blog_id The blog ID.
 				 */
 				do_action(
 					'woocommerce_gla_track_event',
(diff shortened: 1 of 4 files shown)


puntope:
> I included the blog_id because I thought it might help with debugging issues, but I can't see it in Tracks. I'm not sure why that is.

@jorgemd24 I yes get the blog ID (in the logs using the snippets. In tracks I didn0't check)

src/API/WP/OAuthService.php
```diff
@@ -159,16 +159,57 @@ public function revoke_wpcom_api_auth(): string {
 		$request = $this->container->get( Jetpack::class )->remote_request( $args );
 
 		if ( is_wp_error( $request ) ) {
+
+			/**
+			 * @event revoke_wpcom_api_auth
+			 */
```
puntope, line 170: Can we add here the property directives? Also, I miss the event description.

src/API/WP/OAuthService.php
```diff
@@ -159,16 +159,57 @@ public function revoke_wpcom_api_auth(): string {
...
 			if ( ! $status || $status !== 200 ) {
-				$data = json_decode( $body, true );
-				throw new Exception( $data['message'] ?? 'Error revoking access to WPCOM.', $status );
+				$data    = json_decode( $body, true );
+				$message = $data['message'] ?? 'Error revoking access to WPCOM.';
+
+				/**
+				 * @event revoke_wpcom_api_auth
+				 */
+				do_action(
```
puntope, line 199: Same as https://github.com/woocommerce/google-listings-and-ads/pull/2452/files#r1666846943

src/API/WP/OAuthService.php
jorgemd24: Thanks I add them here: [fb8b49c](https://github.com/woocommerce/google-listings-and-ads/pull/2452/commits/fb8b49cf7b43daf59cac9f48b00d162d432c4943)

jorgemd24:
Thanks @puntope for the review!

> I yes get the blog ID (in the logs using the snippets. In tracks I didn0't check)

Yes, the `blog_id` is being sent to Tracks, and it shows in the logs, but I can't see it in Tracks. I'll need to investigate further to see if there are any restrictions on sending this parameter to Tracks.

Can you take another look to this PR?

src/API/WP/OAuthService.php
```diff
@@ -159,16 +159,73 @@ public function revoke_wpcom_api_auth(): string {
...
 
 		if ( is_wp_error( $request ) ) {
+
+			/**
+			 * When the WPCOM token has been revoked with errors.
+			 *
+			 * @event revoke_wpcom_api_authorization
+			 * @property string status The status of the request.
+			 * @property string error The error message.
+			 * @property mixed blog_id The blog ID.
```
puntope: Instead of mixed WDYT about `number|null` ?

src/API/WP/OAuthService.php
```diff
@@ -159,16 +159,73 @@ public function revoke_wpcom_api_auth(): string {
...
-				throw new Exception( $data['message'] ?? 'Error revoking access to WPCOM.', $status );
+				$data    = json_decode( $body, true );
+				$message = $data['message'] ?? 'Error revoking access to WPCOM.';
+
+				/**
+				*
+				* When the WPCOM token has been revoked with errors.
+				*
+				* @event revoke_wpcom_api_authorization
+				* @property string status The status of the request.
```
puntope: Status is number in this case right?

src/API/WP/OAuthService.php
puntope: Also, maybe we can consider sending 400 or 500 for example or get the error code to have consistency between both `woocommerce_gla_track_event` tracks (in some is a number, the request status,  and in this one the string "error" )

puntope: error seems to be optional

puntope (commented):
Thanks @jorgemd24 for adding the docs. I left some additional comments

jorgemd24:
Hi @puntope,

I addressed your suggestions. Can u have another look? Thanks

src/API/WP/OAuthService.php
```diff
@@ -159,16 +159,72 @@ public function revoke_wpcom_api_auth(): string {
...
+				);
+
+				throw new Exception( $message, $status );
 			}
 
+			/**
+			* When the WPCOM token has been revoked successfully.
+			*
+			* @event revoke_wpcom_api_authorization
+			* @property string status The status of the request.
```
puntope: Missing to refactor the status, is also int here.

puntope (approved):
Thx @jorgemd24 approved in advance. However, I miss some doc tweak in one of the `revoke_wpcom_api_authorization `

PR #2455, opened 2024-07-07
[API Pull] Ensure empty GLA attributes are returned as an object instead of an empty array in the responses.

jorgemd24:

Part of #2146 

This PR ensures that empty GLA attributes are returned as an empty object instead of an empty array, maintaining consistency in the responses.

1. Remove any attribute mapping set.
2. Make a request to `GET wp-json/wc/v3/products/PRODUCT_ID?gla_syncable=1`
3. Check that `gla_attributes` is returned as an empty object instead of an empty array.

### Changelog entry

>


puntope (approved):
LGTM ✅  I can see the empty array is now an empty object

Issue #1887, opened 2023-02-14
Dashboard campaigns Row per page not working


puntope:
### Describe the bug:

While testing PMAX campaigns I realised I had more than 25 campaigns but paginator still shows 25 per page and showing my 35 campaigns. 

### Steps to reproduce:

1. Create 26 campaigns
2. See the dashboard showing 26 campaigns
3. See rows per page showing 25 

### Expected behavior:

- It only shows 25 campaigns and navigation for switching between pages
- The selector allows to change campaigns per page

### Actual behavior:
- It shows all the campaigns
- The rows per page selector is not working

### Screenshots

https://user-images.githubusercontent.com/5908855/218848838-4cdc3648-f3c7-4b4c-a8b4-9f27bd8f94a6.mov

puntope:
After some investigation it seems like we need to implement the Google API pagination. Seems like it's not an easy fix since Google API Pagination works with PageToken links instead of OFFSET param. But in overview the steps to make it happen could be:

Backend:

- Accept PageSize and PageToken in the `/ads/campaigns` endpoint
- Grab the results from Google API using that params
- Return the campaigns and also the NextPageToken

Frontend:

- Add pageSize and  pageToken in he API call
- Modify the response handler to take the campaigns data and the nextPageToken
- Map the pageTokens with page numbers for the forwards and backwards navigation
- Modify the TableCard and add a custom paginator similar to AttributeMapping rules or ProductFeed Issues.

jorgemd24:
While working on this PR https://github.com/woocommerce/google-listings-and-ads/pull/2472, I was also considering this issue. However, a downside of using pagination is that the [pagination component](https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/components/src/pagination) requires the total number of results across all pages. Currently[, Google doesn't offer an equivalen](https://groups.google.com/g/adwords-api/c/yg_Nxf4L6UE)t of `COUNT` as in MySQL; you have to fetch all the results. Thus, the benefit of pagination would be lost if we still have to retrieve all the campaigns.

mikkamp:
[...]

PR #2472, opened 2024-07-16
Enable page size in campaigns endpoint

jorgemd24:

Part of pcTzPl-2nS-p2

This PR allows setting the per_page parameter in the gla/ads/campaigns endpoint. Previously, we returned all the results, but now it is possible to limit the number of results.

1. Make a request to `GET gla/ads/campaigns?per_page=X` and check that the number of results matches the per_page parameter.
2. Make a request to `GET gla/ads/campaigns` and ensure that you get all the results.

### Additional details:

I had to replace [iterateAllElements()](https://googleapis.github.io/gax-php/master/Google/ApiCore/PagedListResponse.html#method_iterateAllElements) with `getPage()->getIterator()` to avoid iterating through all the pages and retrieving all the results at once.

### Changelog entry

> Update - Enable Page Size Parameter in Campaigns Endpoint

### DIFF
--- a/src/API/Google/AdsCampaign.php
+++ b/src/API/Google/AdsCampaign.php
@@ -99,15 +99,16 @@ class AdsCampaign implements ContainerAwareInterface, OptionsAwareInterface {
 	/**
 	 * Returns a list of campaigns with targeted locations retrieved from campaign criterion.
 	 *
-	 * @param bool $exclude_removed Exclude removed campaigns (default true).
-	 * @param bool $fetch_criterion Combine the campaign data with criterion data (default true).
+	 * @param bool  $exclude_removed Exclude removed campaigns (default true).
+	 * @param bool  $fetch_criterion Combine the campaign data with criterion data (default true).
+	 * @param array $args Additional arguments.
 	 *
 	 * @return array
 	 * @throws ExceptionWithResponseData When an ApiException is caught.
 	 */
-	public function get_campaigns( bool $exclude_removed = true, bool $fetch_criterion = true ): array {
+	public function get_campaigns( bool $exclude_removed = true, bool $fetch_criterion = true, $args = [] ): array {
 		try {
-			$query = ( new AdsCampaignQuery() )->set_client( $this->client, $this->options->get_ads_id() );
+			$query = ( new AdsCampaignQuery( $args ) )->set_client( $this->client, $this->options->get_ads_id() );
 
 			if ( $exclude_removed ) {
 				$query->where( 'campaign.status', 'REMOVED', '!=' );
@@ -117,7 +118,7 @@ class AdsCampaign implements ContainerAwareInterface, OptionsAwareInterface {
 			$campaign_results    = $query->get_results();
 			$converted_campaigns = [];
 
-			foreach ( $campaign_results->iterateAllElements() as $row ) {
+			foreach ( $campaign_results->getPage()->getIterator()  as $row ) {
 				++$campaign_count;
 				$campaign                               = $this->convert_campaign( $row );
 				$converted_campaigns[ $campaign['id'] ] = $campaign;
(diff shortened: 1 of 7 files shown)


src/API/Google/AdsCampaign.php
```diff
@@ -99,15 +99,16 @@ public function __construct( GoogleAdsClient $client, AdsCampaignBudget $budget,
 	/**
 	 * Returns a list of campaigns with targeted locations retrieved from campaign criterion.
 	 *
-	 * @param bool $exclude_removed Exclude removed campaigns (default true).
-	 * @param bool $fetch_criterion Combine the campaign data with criterion data (default true).
+	 * @param bool  $exclude_removed Exclude removed campaigns (default true).
+	 * @param bool  $fetch_criterion Combine the campaign data with criterion data (default true).
+	 * @param array $args Additional arguments.
```
puntope: WDYT of being more concrete here?

Maybe 

 @param array $args Arguments for the Ads Campaign Query.

tests/Unit/API/Google/AdsCampaignTest.php
```diff
@@ -136,6 +136,66 @@ public function test_get_campaigns() {
...
+				'geo_target_constant' => 'geoTargetConstants/2158',
+			],
+			[
+				'campaign_id'         => 5678901234,
+				'geo_target_constant' => 'geoTargetConstants/2344',
+			],
+			[
+				'campaign_id'         => 5678901234,
+				'geo_target_constant' => 'geoTargetConstants/2826',
+			],
```
puntope, line 152: Its just a test, but the ids are duplicated.

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
```diff
@@ -152,6 +152,67 @@ public function test_get_campaigns_converted_names() {
 		$this->assertEquals( 200, $response->get_status() );
 	}
 
+	public function test_get_campaigns_with_args() {
```
puntope, line 177: In the test added here where are we testing another page_size? For example page_size=1 to see if we don't get all the campaigns and just 1

puntope (commented):
Thanks for the changes @jorgemd24 

I left some minor comments but the per_page works as expected.

One thing I miss is the page parameter to get the results for a specific page. Not sure if its for an other PR.

jorgemd24:
Thanks @puntope for reviewing this PR! I addressed your suggestions.

> One thing I miss is the page parameter to get the results for a specific page. Not sure if its for an other PR.

Initially, I planned not to set the next page token in the endpoint response for the reasons mentioned in this comment: https://github.com/woocommerce/google-listings-and-ads/issues/1887#issuecomment-2231712265. However, after @mikkamp's comment, I realized it is possible to get the total number of results in the first query. Therefore, I implemented the following headers:

- `X-WP-Total` -> Total number across all pages.
- `X-WP-TotalPages` -> Total pages.
- `X-GLA-NextPageToken` -> The token to get the next page. (I used the GLA prefix to avoid future conflicts with this header name)

With this, we can fetch the next page and store the page token in the client to navigate backwards. However, it seems that it is not possible to get the forward page tokens until you access the page, (for example jumping from page 1 to page 5) so we need to see if it is feasible to implement this in the front end and the back end. 

I found this example that can be useful: https://developers.google.com/google-ads/api/samples/navigate-search-result-pages-caching-tokens but for now, I will leave that implementation/investigation for a different PR as the mobile team is interested in limiting the result to 1 to check if the merchant is active and I think this PR will help us to move forward to implement the pagination in the campaign table.

src/API/Google/AdsCampaign.php
```diff
@@ -99,15 +99,17 @@ public function __construct( GoogleAdsClient $client, AdsCampaignBudget $budget,
...
+	 * @param bool  $exclude_removed Exclude removed campaigns (default true).
+	 * @param bool  $fetch_criterion Combine the campaign data with criterion data (default true).
+	 * @param array $args Arguments for the Ads Campaign Query for example: per_page for limiting the number of results.
+	 * @param bool  $return_pagination_params Whether to return pagination params (default false).
 	 *
 	 * @return array
 	 * @throws ExceptionWithResponseData When an ApiException is caught.
 	 */
-	public function get_campaigns( bool $exclude_removed = true, bool $fetch_criterion = true ): array {
+	public function get_campaigns( bool $exclude_removed = true, bool $fetch_criterion = true, $args = [], $return_pagination_params = false ): array {
```
puntope, line 110: What's the value of NOT returning pagination params? Even if they are not needed I assume they are not harmful. So maybe we can simplify this method by returning them always.

jorgemd24: @puntope, the reason for not always returning the pagination params is that we would need to refactor other parts of the code and tests that depend on this method:

- https://github.com/woocommerce/google-listings-and-ads/blob/6ff59162df0fc9c79e1f0c1af8267a764fe143a9/src/API/Google/AdsCampaign.php#L405
- https://github.com/woocommerce/google-listings-and-ads/blob/6ff59162df0fc9c79e1f0c1af8267a764fe143a9/src/MultichannelMarketing/GLAChannel.php#L210
- https://github.com/woocommerce/google-listings-and-ads/blob/6ff59162df0fc9c79e1f0c1af8267a764fe143a9/tests/Unit/API/Google/AdsCampaignTest.php#L95
- ..etc

Do you think it’s worth refactoring those parts even if they don't require pagination?

puntope (approved):
Thx. @jorgemd24 I think we can go forward with this and evaluate the refactoring you mentioned

PR #2485, opened 2024-07-31
Enable labeling of Ads campaigns

jorgemd24:

Part of pcTzPl-2nS-p2

We would like to track where campaigns are created from by labeling them with their source of creation. For example, use `wc-gla` for the web version, and for the mobile versions, something like `wc-gla-android` or `wc-gla-ios`.

This PR allows the creation of a new campaign with a specific label. Note that I haven't updated the frontend to send the label query parameter yet, as I would first like to define the label names we'll be using.

 1. Make the following call `POST wp-json/wc/gla/ads/campaigns` with this body: 
```
{
    "amount": 0.01,
    "targeted_locations": ["ES"],
    "label": "wc-gla"
}
```
3. The campaign should be created successfully.
4. Go to https://ads.google.com/
5. Filter your campaign by label:

image

You should see the campaign that you just created.
6. Alternatively, you can customize the campaigns table to display the labels:

image

image

image

### Changelog entry

> Add - Enable labeling of Ads campaigns

### DIFF
--- a/src/API/Google/AdsCampaignLabel.php
+++ b/src/API/Google/AdsCampaignLabel.php
@@ -5,7 +5,6 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query\AdsCampaignLabelQuery;
 use Automattic\WooCommerce\GoogleListingsAndAds\Google\Ads\GoogleAdsClient;
-use Automattic\WooCommerce\GoogleListingsAndAds\Exception\ExceptionWithResponseData;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Google\Ads\GoogleAds\Util\V16\ResourceNames;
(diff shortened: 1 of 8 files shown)


src/API/Google/AdsCampaignLabel.php
```diff
@@ -0,0 +1,162 @@
...
+		$query->set_client( $this->client, $this->options->get_ads_id() );
+		$query->where( 'label.name', $name, '=' );
+		$label_results = $query->get_results();
+		$label_id      = null;
+
+		foreach ( $label_results->iterateAllElements() as $row ) {
+			return $row->getLabel()->getId();
+		}
+
+		return $label_id;
```
martynmjones: I'm curious why you decided to create the `$label_id` variable instead of just returning `null` after the loop?

martynmjones (approved):
Hey @jorgemd24, thanks for adding the label functionality 🙌 

Confirmed that the label is created and assigned to the campaign.

Screenshot 2024-08-01 at 19 30 11

I left one question but there aren't any issues so LGTM ✅

src/API/Google/AdsCampaignLabel.php
jorgemd24: Yep that's true, it is unnecessary, I removed it here [d30851d](https://github.com/woocommerce/google-listings-and-ads/pull/2485/commits/d30851d08749e68bdbab6ecfa508d66c7c95a735)

PR #2505, opened 2024-08-06
Display body message if error status is 406 during account creation

martynmjones:
For merchants who have created more than `20` Ads accounts via the connect server we want to display a more detailed error message and direct them to contact support for help. This PR adjusts the error handling so that if the server returns `406` status then we display the error message as it's received.

### Screenshots:

Screenshot 2024-08-06 at 21 05 33

---

When reviewing this PR, please review https://github.com/Automattic/woocommerce-connect-server/pull/2554 at the same time.

---

1. Run a local instance of the connect server using the branch in https://github.com/Automattic/woocommerce-connect-server/pull/2554
1. Connect a local GLA test site to that connect server
1. Update the local connect server config and set [the google ads account limit](https://github.com/Automattic/woocommerce-connect-server/blob/da9eb1558182f7a87a64f2cd72db01c610e9b9e4/config/default.js#L223) to `0` 
1. On your local GLA test site, go through onboarding and attempt to create a new Google Ads account
1. Confirm the error message reads `Error creating account: Account creation limit reached. Contact support for help.`

### Changelog entry

> Tweak - Display additional context in error message when Google Ads account limit reached

### DIFF
--- a/js/src/hooks/useUpsertAdsAccount.js
+++ b/js/src/hooks/useUpsertAdsAccount.js
@@ -52,12 +52,13 @@ const useUpsertAdsAccount = () => {
 			// and only display error message and exit this function for non-428 error.
 			if ( e.status !== 428 ) {
 				const body = await e.json();
-				const message = e.status === 406 ?
-					body.message :
-					__(
-						'Unable to create Google Ads account. Please try again later.',
-						'google-listings-and-ads'
-					);
+				const message =
+					e.status === 406
+						? body.message
+						: __(
+								'Unable to create Google Ads account. Please try again later.',
+								'google-listings-and-ads'
+						  );
 
 				createNotice( 'error', message );
 			}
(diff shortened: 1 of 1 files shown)


js/src/hooks/useUpsertAdsAccount.js
```diff
@@ -51,13 +51,16 @@ const useUpsertAdsAccount = () => {
...
-					'error',
-					__(
-						'Unable to create Google Ads account. Please try again later.',
-						'google-listings-and-ads'
-					)
-				);
+				const body = await e.json();
+				const message =
+					e.status === 406
+						? body.message
```
mikkamp: If we are certain that a 406 will only return one type of error then wouldn't it be better to use a text string here even if it's a duplicate of the server message? Otherwise this error message can't be translated.

mikkamp (approved):
Thanks for changing the error message. I can confirm it's being shown when attempting to create an account:

image

I checked the response code and confirmed it's `406 Not Acceptable`. I just left one small comment about translation but it's not a big issue if we don't decide to do that.

I've checked this with https://github.com/Automattic/woocommerce-connect-server/pull/2554 and left a review for that PR as well.

js/src/hooks/useUpsertAdsAccount.js
martynmjones: Thanks for the suggestion Mik. I don't imagine we'd rely on a `406` for any other error down the line so I've made the change in f772909.

PR #2511, opened 2024-08-07
Update Block validation to support error context

### LINKED ISSUE #2507: Snackbar error message showing as "undefined" with WC `9.2.0-rc.1`
### Describe the bug:
Using the new product editor we get an "undefined" error message if an input on the Google for WooCommerce tab has an invalid value.

### Steps to reproduce:
1. Install WooCommerce `9.2.0-rc.1`
2. Enable the new product editor
3. Edit a product and click on the `Google for WooCommerce` tab
4. Enter an invalid value for `Availability Date`
5. Click `Save draft` or `Publish`
6. See error

Screenshot 2024-08-07 at 14 42 06

### Additional details:
Surfaced by failing E2E test during compat testing. See #2506

The `Multipack` input also has some validation issues and if we attempt to save a product after entering a decimal value in the field we get an empty Snackback and no error under the field:

Screenshot 2024-08-07 at 21 58 27

**Related changes in core**
- https://github.com/woocommerce/woocommerce/pull/49242
- https://github.com/woocommerce/woocommerce/pull/49246

martynmjones:

Closes #2507 

Updates Block validation to return an object containing the error message and the clientId as context to support the new link displayed in the snackbar.

Screenshot 2024-08-07 at 16 09 11

Additionally, with WooCommerce `9.2.0-rc.1` the `Multipack` field validation changes so if a decimal value was entered a blank error message was displayed:

Screenshot 2024-08-07 at 21 58 27

As it's not possible to either set a default validation error message or add custom rules outside of the supported few, the field has been updated to use a text field instead of number. A [pattern is then supplied with a custom error message](https://github.com/woocommerce/google-listings-and-ads/blob/681a07c3c107ccc33014680fa9ad64a3e3812737/src/Admin/Input/Integer.php#L23) to prevent users from entering non numeric values, decimal values, or negative values.

Screenshot 2024-08-08 at 12 06 32

1. `npm run wp-env:up`
1. `npm run -- wp-env run tests-cli -- wp plugin update woocommerce --version=9.2.0-rc.1`
1. `npm run -- wp-env run tests-cli -- wp wc update`
1. `npm run test:e2e`
1. Confirm tests pass
1. Enable the new product editor in `WooCommerce > Settings > Advanced > Features`
1. Edit a product and test different values in the `Multipack` field under the `Google for WooCommerce` tab
1. Confirm only whole numbers are accepted input

### Additional
⚠️  Please merge once approved as this PR is required before GLA can be released with the `9.2.0` compat bump.

### Changelog entry

> Update - Block validation to support error context

### DIFF
--- a/js/src/blocks/product-date-time-field/edit.js
+++ b/js/src/blocks/product-date-time-field/edit.js
@@ -32,7 +32,7 @@ async function resolveValidationMessage( inputRef, context ) {
 
 		return {
 			message: input.validationMessage,
-			context
+			context,
 		};
 	}
 }
@@ -43,6 +43,7 @@ async function resolveValidationMessage( inputRef, context ) {
  * @param {Object} props React props.
  * @param {ProductBasicAttributes} props.attributes
  * @param {ProductEditorBlockContext} props.context
+ * @param {string} props.clientId
  */
 export default function Edit( { attributes, context, clientId } ) {
 	const { property } = attributes;
--- a/js/src/blocks/product-date-time-field/edit.js
+++ b/js/src/blocks/product-date-time-field/edit.js
@@ -10,6 +10,7 @@ import {
 	useValidation,
 } from '@woocommerce/product-editor';
 import { Flex, FlexBlock } from '@wordpress/components';
+import { isWcVersion } from '@woocommerce/settings'; // eslint-disable-line import/no-unresolved
 
 /**
  * Internal dependencies
@@ -21,11 +22,18 @@ import styles from './editor.module.scss';
  * @typedef {import('../types.js').ProductBasicAttributes} ProductBasicAttributes
  */
 
-async function resolveValidationMessage( inputRef ) {
+async function resolveValidationMessage( inputRef, context ) {
 	const input = inputRef.current;
 
 	if ( ! input.validity.valid ) {
-		return input.validationMessage;
+		if ( isWcVersion( '9.2.0', '>' ) ) {
+			return input.validationMessage;
+		}
+
+		return {
+			message: input.validationMessage,
+			context
+		};
 	}
 }
 
@@ -36,7 +44,7 @@ async function resolveValidationMessage( inputRef ) {
  * @param {ProductBasicAttributes} props.attributes
  * @param {ProductEditorBlockContext} props.context
  */
-export default function Edit( { attributes, context } ) {
+export default function Edit( { attributes, context, clientId } ) {
 	const { property } = attributes;
 	const blockProps = useWooBlockProps( attributes );
 	const [ value, setValue ] = useProductEntityProp( property, {
@@ -76,11 +84,11 @@ export default function Edit( { attributes, context } ) {
 	};
 
 	const dateValidation = useValidation( `${ property }-date`, () =>
-		resolveValidationMessage( dateInputRef )
+		resolveValidationMessage( dateInputRef, clientId )
 	);
 
 	const timeValidation = useValidation( `${ property }-time`, () =>
-		resolveValidationMessage( timeInputRef )
+		resolveValidationMessage( timeInputRef, clientId )
 	);
 
 	return (
--- a/js/src/blocks/product-date-time-field/edit.js
+++ b/js/src/blocks/product-date-time-field/edit.js
@@ -26,7 +26,8 @@ async function resolveValidationMessage( inputRef, context ) {
 	const input = inputRef.current;
 
 	if ( ! input.validity.valid ) {
-		if ( isWcVersion( '9.2.0', '>' ) ) {
+		// compatibility-code "WC < 9.2"
+		if ( isWcVersion( '9.2.0', '<' ) ) {
 			return input.validationMessage;
 		}
 
--- a/src/Admin/Input/Integer.php
+++ b/src/Admin/Input/Integer.php
@@ -20,6 +20,12 @@ class Integer extends Input {
 		// the text field block to work around it.
 		parent::__construct( 'integer', 'woocommerce/product-text-field' );
 
-		$this->set_block_attribute( 'type', [ 'value' => 'number' ] );
+		$this->set_block_attribute(
+			'pattern',
+			[
+				'value'   => '\d+',
+				'message' => __( 'Please enter a valid value.', 'woocommerce' ),
+			]
+		);
 	}
 }
--- a/src/Admin/Input/Integer.php
+++ b/src/Admin/Input/Integer.php
@@ -24,6 +24,9 @@ class Integer extends Input {
 			'pattern',
 			[
 				'value'   => '\d+',
+				// This error message is used in core so to avoid needing to translate
+				// the string a second time WooCommerce is supplied as the text domain.
+				// phpcs:ignore WordPress.WP.I18n.TextDomainMismatch
 				'message' => __( 'Please enter a valid value.', 'woocommerce' ),
 			]
 		);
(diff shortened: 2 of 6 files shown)


js/src/blocks/product-date-time-field/edit.js
```diff
@@ -21,11 +22,18 @@ import styles from './editor.module.scss';
  * @typedef {import('../types.js').ProductBasicAttributes} ProductBasicAttributes
  */
 
-async function resolveValidationMessage( inputRef ) {
+async function resolveValidationMessage( inputRef, context ) {
 	const input = inputRef.current;
 
 	if ( ! input.validity.valid ) {
-		return input.validationMessage;
+		if ( isWcVersion( '9.2.0', '>' ) ) {
```
eason9487: Looks like the comparison should be `<`, and suggest adding a `compatibility-code` comment.

```suggestion
		// compatibility-code "WC < 9.2"
		if ( isWcVersion( '9.2.0', '<' ) ) {
```

Otherwise, with WC < 9.2, it will cause an error after entering an invalid value.

https://github.com/user-attachments/assets/6adddec1-0479-423c-8db7-4ffdd69cd9b7

tests/e2e/utils/product-editor.js
```diff
@@ -577,17 +577,15 @@ export function getProductBlockEditorUtils( page ) {
...
-				.filter( { hasText: new RegExp( message ) } );
-
-			const failureNoticeDismissButton =
-				failureNotice.getByRole( 'button' );
+			const failureNotice = page.locator(
+				'.components-snackbar__content'
+			);
 
 			await expect( failureNotice ).toBeVisible();
+			await expect( failureNotice ).toHaveText( message );
```
eason9487: This assertion failed with WC < 9.2. How about making it also compatible?

```suggestion
			await expect( failureNotice ).toContainText( message );
```

tests/e2e/utils/product-editor.js
```diff
@@ -577,17 +577,15 @@ export function getProductBlockEditorUtils( page ) {
...
 
-			// Dismiss the notice.
-			await failureNoticeDismissButton.click();
+			// // Dismiss the notice.
+			await failureNotice.click();
```
eason9487: This action failed with WC < 9.2. How about making it also compatible?

```suggestion
			// Dismiss the notice.
			await failureNotice
				.getByRole( 'button' ) // compatibility-code "WC < 9.2" -- Dismiss by its inner close button.
				.or( failureNotice )
				.click();

```

src/Admin/Input/Integer.php
```diff
@@ -20,6 +20,15 @@ public function __construct() {
 		// the text field block to work around it.
 		parent::__construct( 'integer', 'woocommerce/product-text-field' );
 
-		$this->set_block_attribute( 'type', [ 'value' => 'number' ] );
+		$this->set_block_attribute(
+			'pattern',
+			[
+				'value'   => '\d+',
```
eason9487: Maybe the pattern could be `[0-9]\d?` to avoid saving a value like '000123'.

src/Admin/Input/Integer.php
```diff
@@ -20,6 +20,15 @@ public function __construct() {
...
+				// This error message is used in core so to avoid needing to translate
+				// the string a second time WooCommerce is supplied as the text domain.
+				// phpcs:ignore WordPress.WP.I18n.TextDomainMismatch
+				'message' => __( 'Please enter a valid value.', 'woocommerce' ),
```
eason9487: I couldn't find the same text in WooCommerce.

Is it supposed to be "Invalid value for the field."? Or perhaps leave it unspecified and let it fall back to the default error message in WooCommerce.

eason9487 (commented):
Thanks for fixing the compatibility!

There is a blocker that the comparison parameter of calling `isWcVersion` may not be correct. Other comments are optional.

src/Admin/Input/Integer.php
jorgemd24: Not sure about this. If I use `[0-9]\d`, it will only match two digits and won't work for single-digit numbers. If I switch to `[0-9]\d+`, it will match any number of digits but won't prevent cases like `000123`. To handle that, I think the pattern needs to be something like `^(0|[1-9]\d*)$`. WDYT?

eason9487: Sorry, it should be `[0-9]\d*`.

> To handle that, I think the pattern needs to be something like `^(0|[1-9]\d*)$`. WDYT?

`^(?:` and `)$` impliedly wrap up the pattern of `<input>`, so it will match the entire value rather than matching a substring.

src/Admin/Input/Integer.php
jorgemd24: I couldn't find that text either, so I removed the message and switched to the fallback.: [e40b122](https://github.com/woocommerce/google-listings-and-ads/pull/2511/commits/e40b1225afb555a74e8a959de3cd008c578824fc)

tests/e2e/utils/product-editor.js
jorgemd24: Thanks for the suggestion I've updated it here: [51e3679](https://github.com/woocommerce/google-listings-and-ads/pull/2511/commits/51e36797ff3f70569e16539409aed4990bd5120d)

tests/e2e/utils/product-editor.js
jorgemd24: I tried using the or method, but if both elements are present, it throws a "strict mode violation." Using nth to select one element worked for some tests but not for others. So, I decided to take a different approach, but you can adjust it if needed. [51e3679](https://github.com/woocommerce/google-listings-and-ads/pull/2511/commits/51e36797ff3f70569e16539409aed4990bd5120d)

src/Admin/Input/Integer.php
jorgemd24: mmm, I tried using `[0-9]\d*` but I got a similar result to `0-9]\d+`, we are still able to type `00123`.

eason9487: Sorry again 😂. I just tested with `\d|[1-9]\d*` and hope it's correct this time.

eason9487: I was too focused on the `^` and `$` parts. Just noticed the pattern `0|[1-9]\d*` you commented in https://github.com/woocommerce/google-listings-and-ads/pull/2511#discussion_r1715375981 should work as well.

tests/e2e/utils/product-editor.js
eason9487: > I tried using the or method, but if both elements are present, it throws a "strict mode violation."

I later realized it had to be changed to this but I missed updating to the comment before submitting review.

```js
// Dismiss the notice.
await failureNotice
	.or( failureNotice.getByRole( 'button' ) )
	.last()
	.click();
```

> So, I decided to take a different approach, but you can adjust it if needed. [51e3679](https://github.com/woocommerce/google-listings-and-ads/pull/2511/commits/51e36797ff3f70569e16539409aed4990bd5120d)

Your method is better and clearer, thanks!

src/Admin/Input/Integer.php
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2511/commits/ea781f60c49d9b37ec68c3d4e98b1ad6cf56085e with more assertions for invalid values.

eason9487 (approved):
The E2E tests related to product editor integration passed for the latest change. 

E2E test result with:
- WC 9.2.0-rc.1: https://github.com/woocommerce/google-listings-and-ads/actions/runs/10380513588/job/28740467547
- WC 9.1.4: https://github.com/woocommerce/google-listings-and-ads/actions/runs/10380933469/job/28742220894
   - Not sure why a few gtag-events tests keep failing but it should be irrelavant.

PR #2512, opened 2024-08-08
Update WPORG header image

puntope:

This PR updates the banners with the new ones provided by Google

1. Check that the banners are the same size as the other 2 and match the one In https://woocommerce.com/products/google-listings-and-ads/

### Changelog entry

> Tweak - Update WPORG plugin page header image


mikkamp (approved):
Thanks for the changes. I can confirm the sizes match and the change looks good to me.

Just wanted to note that WooRelease does not update these asset images. So either we'll need to commit to svn after this PR is merged. Or leave a note for the porter that this needs to be done in combination with the release.

puntope:
> Thanks for the changes. I can confirm the sizes match and the change looks good to me.
> 
> Just wanted to note that WooRelease does not update these asset images. So either we'll need to commit to svn after this PR is merged. Or leave a note for the porter that this needs to be done in combination with the release.

Thx @mikkamp I added a not to myself for doing this tomorrow

PR #2514, opened 2024-08-08
Label campaigns for the web version and the WC Mobile app

jorgemd24:

Part of pcTzPl-2nS-p2

This PR will automatically label campaigns created via the WC Mobile App by checking the user agent headers. The labels used will be `wc-android` or `wc-ios`. If the campaign isn't created through the WC Mobile App, it will be labeled as `wc-web`.

1. Download the WC Mobile App and connect it to one of your local sites. Then, navigate to WC Admin -> Google for WooCommerce and create a new campaign. Alternatively,  simulate this process by sending a request to `POST /wp-json/wc/gla/ads/campaigns` with the `User Agent` header set to something like: `Mozilla/5.0 (iPhone; CPU iPhone OS 17_5_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 wc-ios/19.7.1`. Set the below body:

```
{
    "amount": 0.01,
    "targeted_locations": ["ES"],
}
```

2. Follow the steps mentioned in this PR to check if the label has been set: https://github.com/woocommerce/google-listings-and-ads/pull/2485. It has to be `wc-ios` or `wc-android`.
3. Repeat the process on a laptop, and the campaign label should be `wc-web`

### Changelog entry

> Tweak - Label campaigns for the web version and the WooCommerce Mobile app.


js/src/data/actions.js
```diff
@@ -805,6 +805,7 @@ export function* createAdsCampaign( amount, countryCodes ) {
 			data: {
 				amount,
 				targeted_locations: countryCodes,
+				label: 'wc-web',
```
mikkamp: I'm assuming this means we are defaulting the label to `wc-web` even on mobile, and then only overwriting the value once we process the request in PHP. It also means that when testing for mobile if I view the requests it's sending a request with `wc-web` but the final campaign gets a different label.

Is that in preparation for the mobile eventually calling the API directly? I was imagining that in this implementation we'd fetch the user agent in JS and then populate this label accordingly. Fine either approach though if it's better to detect the user agent in PHP.

src/API/Site/Controllers/Ads/CampaignController.php
```diff
@@ -152,6 +154,14 @@ protected function create_campaign_callback(): callable {
 					);
 				}
 
+				if ( $this->is_wc_mobile_app( $request ) ) {
+					if ( $this->is_wc_ios( $request ) ) {
+						$fields['label'] = 'wc-ios';
+					} else {
+						$fields['label'] = 'wc-android';
+					}
+				}
```
mikkamp: PHP will do some optimization but if we follow the code path this will technically fetch the user agent header and then check if it contains the string `wc-ios` twice. If we aren't going to optimize it so it only needs to compare the header once, then wouldn't it look slightly cleaner to check it like this:

```suggestion
				if ( $this->is_wc_ios( $request ) ) {
					$fields['label'] = 'wc-ios';
				} elseif ( $this->is_wc_android( $request ) ) {
					$fields['label'] = 'wc-android';
				}
```

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
```diff
@@ -332,6 +332,99 @@ public function test_create_campaign() {
...
+			'targeted_locations' => [ 'US', 'GB', 'TW' ],
+		];
+
+		$matcher = $this->exactly( 2 );
+		$this->ads_campaign->expects( $matcher )
+			->method( 'create_campaign' )
+			->willReturnCallback(
+				function ( $data ) use ( $campaign_data, $matcher, $expected ) {
+					$invokation = $matcher->getInvocationCount();
+					if ( $invokation === 1 ) {
```
mikkamp: Not a big deal, but I would spell the variable with a `c`, or reduce the need for storing it in a variable.

```suggestion
					if ( 1 === $matcher->getInvocationCount() ) {
```

mikkamp (approved):
Thanks for getting this added, and including some unit tests.

I tested by alternating the user agent and can confirm it gets the right label when creating the campaign. I left a few smaller comments, but none of them are blocking so I'll go ahead and approve the PR.

js/src/data/actions.js
jorgemd24: Thanks for the suggestion! My initial thought was to detect the header directly on the server, but I like the idea of moving the logic to the client. It keeps the API agnostic to the labels and lets the client handle setting them. If the mobile team wants to use the API to create campaigns, they can simply add the label parameter in the body.

jorgemd24:
Hi @mikkamp thanks for the review! I liked your suggestion about detecting the user agent in the client, (see my [comment](https://github.com/woocommerce/google-listings-and-ads/pull/2514#discussion_r1712027164)) so I went ahead and I moved the logic to the client. Can you have another look? Thanks.

js/src/data/test/createAdsCampaign.test.js
```diff
@@ -0,0 +1,92 @@
...
+			method: 'POST',
+			data: {
+				amount: 100,
+				targeted_locations: [ 'ES' ],
+				label: 'wc-ios',
+			},
+		} );
+	} );
+
+	it( 'If the user agent is set to wc-ios the label should be wc-android', async () => {
```
mikkamp: This test title seems to have a copy paste error in it.

js/src/utils/isMobileApp.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Check if the WC app is running on iOS.
+ *
+ * @return {boolean} Whether the WC app is running on iOS.
+ *
+ */
+const isWCIos = () => {
+	return window.navigator.userAgent.includes( 'wc-ios' );
```
mikkamp: Don't think this should be a problem, but in the PHP solution a non case sensitive solution was used, this one is case sensitive. If that's acceptable then ok to leave as is.

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
```diff
@@ -332,6 +332,47 @@ public function test_create_campaign() {
...
+			'created_campaign',
+			[
+				'id'                 => self::TEST_CAMPAIGN_ID,
+				'status'             => 'enabled',
+				'name'               => 'New Campaign',
+				'amount'             => 20,
+				'country'            => self::BASE_COUNTRY,
+				'targeted_locations' => 'US,GB,TW',
+			]
+		);
```
mikkamp: Just looking at this track here. Wouldn't it be useful to also pass the source? Or is it going to still include the user agent string so it can be extracted from there?

mikkamp (approved):
Thanks for the changes, it looks like a much simpler implementation, and continues to work as expected.

I just left a few comments / thoughts, but none of them are blocking.

jorgemd24:
Thanks @mikkamp for the review! I've addressed all your comments.

> Just looking at this track here. Wouldn't it be useful to also pass the source? Or is it going to still include the user agent string so it can be extracted from there?

I think it is good to add the source in the event tracking so I've added it here: [aa6185e](https://github.com/woocommerce/google-listings-and-ads/pull/2514/commits/aa6185e93a520469e2303062ef005a1b200c952e)

PR #2515, opened 2024-08-08
Fix Issue disconnecting wpcom api

jorgemd24:

Closes # .

This PR addresses an issue when attempting to disconnect all accounts on the settings page. The problem occurs when the user hasn't granted access to the new Pull mechanism, but we still try to revoke the wpcom token. Since there’s no token to revoke, WPCOM responds with an error.

I’m wondering if we should adjust this behavior in WPCOM so that if there’s no token to revoke, it simply returns without throwing an error. I’d be happy to hear thoughts on this.

1. Check out the develop branch and open the dev console.
2. Navigate to the settings page.
3. Do not grant access to the new API Pull mechanism.
4. Click "Disconnect all accounts."
5. See the error in the dev console from the REST API authorization endpoint, because there is not aWPCOM token to revoke.
6. Refresh the page and go through the onboarding process again.
7. Switch to this branch.
8. Repeat the steps and notice that you can now disconnect all your accounts without any issues.

### Changelog entry

> Fix - Disconnecting all accounts when WPCOM connection is not granted.

### DIFF
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -523,13 +523,8 @@ export function* disconnectAllAccounts() {
 			type: TYPES.DISCONNECT_ACCOUNTS_ALL,
 		};
 	} catch ( error ) {
-		if (
-			error.errors[
-				`${ API_NAMESPACE }/rest-api/authorize`
-			]?.message?.includes(
-				'No token found associated with the client ID and user'
-			)
-		) {
+		// Skip any error related to revoking WPCOM token.
+		if ( error.errors[ `${ API_NAMESPACE }/rest-api/authorize` ] ) {
 			return {
 				type: TYPES.DISCONNECT_ACCOUNTS_ALL,
 			};
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -523,6 +523,18 @@ export function* disconnectAllAccounts() {
 			type: TYPES.DISCONNECT_ACCOUNTS_ALL,
 		};
 	} catch ( error ) {
+		if (
+			error.errors[
+				`${ API_NAMESPACE }/rest-api/authorize`
+			]?.message?.includes(
+				'No token found associated with the client ID and user'
+			)
+		) {
+			return {
+				type: TYPES.DISCONNECT_ACCOUNTS_ALL,
+			};
+		}
+
 		handleApiError(
 			error,
 			__(
--- a/src/MerchantCenter/AccountService.php
+++ b/src/MerchantCenter/AccountService.php
@@ -623,13 +623,4 @@ class AccountService implements OptionsAwareInterface, Service {
 	public function delete_wpcom_api_auth_nonce(): bool {
 		return $this->options->delete( OptionsInterface::GOOGLE_WPCOM_AUTH_NONCE );
 	}
-
-	/**
-	 * Check if the WPCOM API is approved.
-	 *
-	 * @return bool
-	 */
-	public function is_wpcom_api_approved(): bool {
-		return $this->options->get( OptionsInterface::WPCOM_REST_API_STATUS ) === 'approved';
-	}
 }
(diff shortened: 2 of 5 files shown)


mikkamp:
> I’m wondering if we should adjust this behavior in WPCOM so that if there’s no token to revoke, it simply returns without throwing an error. I’d be happy to hear thoughts on this.

I can't find any documentation about revoking a token, but based on the [documentation about validating a token](https://developer.wordpress.com/docs/oauth2/#6-validating-tokens) it seems it's common to return an error when the token is not valid (or possibly doesn't exist).

So I'd suggest to keep expecting an error, and just handle it accordingly in the client so they can still proceed.

src/MerchantCenter/AccountService.php
```diff
@@ -623,4 +623,13 @@ public function update_wpcom_api_authorization( string $status, string $nonce ):
 	public function delete_wpcom_api_auth_nonce(): bool {
 		return $this->options->delete( OptionsInterface::GOOGLE_WPCOM_AUTH_NONCE );
 	}
+
+	/**
+	 * Check if the WPCOM API is approved.
+	 *
+	 * @return bool
+	 */
+	public function is_wpcom_api_approved(): bool {
```
puntope: We do have already this method in [Options::is_wpcom_api_authorized](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Options/Options.php#L215)

I don't mind one of another but it might need a little refactor. Maybe moving the check inside $this->oauth_service->revoke_wpcom_api_auth())

jorgemd24:
> I can't find any documentation about revoking a token, but based on the [documentation about validating a token](https://developer.wordpress.com/docs/oauth2/#6-validating-tokens) it seems it's common to return an error when the token is not valid (or possibly doesn't exist).

Thanks for your input! We implemented the endpoint in WPCOM for revoking the GFW token, so it’s great to know we're aligned with the other responses in WPCOM.

jorgemd24:
Hi @puntope, thanks for the suggestion! I’ve updated the PR to handle the error on the client side instead of the server. I believe it’s better to attempt revoking the token regardless of the WPCOM Auth status in the DB (just to be safe), and as @mikkamp mentioned, this keeps us aligned with the other endpoints. Could you take another look? Thanks!

js/src/data/actions.js
```diff
@@ -523,6 +523,18 @@ export function* disconnectAllAccounts() {
 			type: TYPES.DISCONNECT_ACCOUNTS_ALL,
 		};
 	} catch ( error ) {
+		if (
+			error.errors[
+				`${ API_NAMESPACE }/rest-api/authorize`
+			]?.message?.includes(
+				'No token found associated with the client ID and user'
+			)
```
puntope: Why not just` if ( error.errors['${ API_NAMESPACE }/rest-api/authorize'] ) ` It seems more generic and catches a wide range of present and future `rest-api/authorize` errors. I wonder if we also change the message to other one in WPCOM then this won't work anymore.

jorgemd24: My intention was to only skip errors related to a missing token. The revoke endpoint in WPCOM can throw errors when the client is invalid or missing JP Headers. Should we skip those too?

puntope: What I would avoid is to attach the error message to the condition to trigger the error. So if we are going to catch all errors (I think we want to) I will just check if `error.errors[${ API_NAMESPACE }/rest-api/authorize]` I assume that if the key is present in the errors array. it means some error happened in the auth endpoint.

jorgemd24: Thanks for the discussion, let's go ahead and skip the errors for now, and if they cause any issues later, we can always make adjustments. I've made the updates here: [007190f](https://github.com/woocommerce/google-listings-and-ads/pull/2515/commits/007190f26e4851e5350b8aab64cc0a2bf0f40c96)

puntope (approved):
Approved!

PR #2516, opened 2024-08-09
Streamline onboarding: Hide WordPress.com account connection setting from the onboarding if already connected

### LINKED ISSUE #2487: Onboarding: Hide WP.com connection card if already connected
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2458

When a user is going through the plugin setup process, they are shown a list of accounts that need to be connected, the first of which is the required connection to WP.com. However, whenever the store is already connected to WP.com, this account card is unnecessary and can be hidden.

## Acceptance Criteria

- [ ] When the store is not connected to a WP.com account, the current `WPComAccountCard` component should be shown in a disconnected state.
- [ ] When the WP.com account is already connected, the `WPComAccountCard` should be hidden from the "setup your accounts screen".
- [ ] When the `WPComAccountCard` is hidden, the `GoogleAccountCard` should be enabled.
- [ ] On completion of onboarding, the linked accounts on the plugin settings page should still show the connected `WPComAccountCard` when connected.

## Implementation Brief

To test the WP.com card in a disconnected state see https://github.com/woocommerce/google-listings-and-ads/issues/2487#issuecomment-2270527832 and https://github.com/woocommerce/google-listings-and-ads/issues/2487#issuecomment-2276122847

The WP.com status is determined in the `SetupAccounts` component defined in `js/src/setup-mc/setup-stepper/setup-accounts/index.js` by using the `useJetpackAccount()` hook. The `jetpack` value is passed to the `WPComAccountCard` component once loaded to determine whether it is shown in a connected or disconnected state.

If the WP.com state is not connected (i.e. `jetpack?.active !== 'yes'`) then the `WPComAccountCard` should be shown so the user can set up their connection and so they see the WP.com card in the connected state once they've completed the setup.

If the initial WP.com state is already connected (i.e. `jetpack?.active === 'yes'`) then the `WPComAccountCard` should not be rendered at all.

No changes should be made to the `WPComAccountCard` component itself, since it needs to be able to show both a connected and disconnected state for other parts of the application.

## Test Coverage
The E2E tests in `tests/e2e/specs/setup-mc/step-1-accounts.test.js` should be updated to show:

1. The `WPComAccountCard` is visible and disconnected initially.
2. The `WPComAccountCard` is hidden in any tests where the WP.com connection is mocked in the `test.beforeAll()` method.

dsawardekar:

This PR improves the plugin setup process by conditionally hiding the WP.com account connection card when the store is already connected to WP.com.

Closes: https://github.com/woocommerce/google-listings-and-ads/issues/2487

### Screenshots:

Disconnected:
image

Connected:
image

Settings:
image

1. Switch to this branch, and rebuild the plugin with `npm run build`
2. With Jetpack not connected, go to the onboarding flow.
  - You should see the WP.com / Jetpack Connection Card
3. Connect Jetpack, and then refresh the page to restart the onboarding steps
  - The WP.com / Jetpack Connection Card should be hidden
4. Verify that the WP.com card is still shown on the settings tab.
  
You can use the below code in an mu-plugin to temporarily test this by toggling the active state to 'yes' / 'no'.

```php
add_filter( 'rest_pre_echo_response', function( $result, $server, $request ) {
	if ( '/wc/gla/jetpack/connected' === $request->get_route() ) {
		$result['active'] = 'no';
	}

	return $result;
}, 10, 3 );
```

### Additional details:

Update - Hides WP.com card if already connected

To run the e2e tests for the onboarding process you can use,

```bash
$ npm run test:e2e -- tests/e2e/specs/setup-mc/step-1-accounts.test.js'
```

This PR also includes some refactoring of the E2E test playwright utilities to avoid relying on nth-child locations for account cards, now that the WP.com account card will not be displayed in some scenarios.

### Changelog entry

> Update - Hides WordPress.com account connection setting from the onboarding flow if already connected.

### DIFF
--- a/js/src/setup-mc/setup-stepper/setup-accounts/index.js
+++ b/js/src/setup-mc/setup-stepper/setup-accounts/index.js
@@ -101,17 +101,15 @@ const SetupAccounts = ( props ) => {
 	 * and the whole page would display this AppSpinner, which is not desired.
 	 */
 	const isLoadingJetpack = ! jetpack;
-	const isJetpackInactive = jetpack?.active !== 'yes';
+	const isJetpackActive = jetpack?.active === 'yes';
 
-	const isLoadingGoogle = jetpack?.active === 'yes' && ! google;
+	const isLoadingGoogle = isJetpackActive && ! google;
 	const isLoadingGoogleMCAccount =
 		google?.active === 'yes' && scope.gmcRequired && ! googleMCAccount;
 
 	if ( isLoadingJetpack || isLoadingGoogle || isLoadingGoogleMCAccount ) {
 		return <AppSpinner />;
 	}
-
-	const isGoogleAccountDisabled = jetpack?.active !== 'yes';
 
 	// Ads is ready when we have a connection and verified and verified access.
 	// Billing is not required, and the 'link_merchant' step will be resolved
@@ -157,10 +155,10 @@ const SetupAccounts = ( props ) => {
 				) }
 			>
 				<VerticalGapLayout size="large">
-					{ isJetpackInactive && (
+					{ ! isJetpackActive && (
 						<WPComAccountCard jetpack={ jetpack } />
 					) }
-					<GoogleAccountCard disabled={ isGoogleAccountDisabled } />
+					<GoogleAccountCard disabled={ ! isJetpackActive } />
 					<GoogleAdsAccountCard />
 				</VerticalGapLayout>
 			</Section>
--- a/js/src/setup-mc/setup-stepper/setup-accounts/index.js
+++ b/js/src/setup-mc/setup-stepper/setup-accounts/index.js
@@ -148,27 +148,22 @@ const SetupAccounts = ( props ) => {
 					'google-listings-and-ads'
 				) }
 			/>
-			{ isJetpackInactive && (
-				<Section
-					className="gla-wp-google-accounts-section"
-					title={ __(
-						'Connect accounts',
-						'google-listings-and-ads'
-					) }
-					description={ __(
-						'The following accounts are required to use the Google for WooCommerce plugin.',
-						'google-listings-and-ads'
-					) }
-				>
-					<VerticalGapLayout size="large">
+			<Section
+				className="gla-wp-google-accounts-section"
+				title={ __( 'Connect accounts', 'google-listings-and-ads' ) }
+				description={ __(
+					'The following accounts are required to use the Google for WooCommerce plugin.',
+					'google-listings-and-ads'
+				) }
+			>
+				<VerticalGapLayout size="large">
+					{ isJetpackInactive && (
 						<WPComAccountCard jetpack={ jetpack } />
-						<GoogleAccountCard
-							disabled={ isGoogleAccountDisabled }
-						/>
-						<GoogleAdsAccountCard />
-					</VerticalGapLayout>
-				</Section>
-			) }
+					) }
+					<GoogleAccountCard disabled={ isGoogleAccountDisabled } />
+					<GoogleAdsAccountCard />
+				</VerticalGapLayout>
+			</Section>
 			<Section
 				className="gla-google-mc-account-section"
 				description={ <GoogleMCDisclaimer /> }
--- a/js/src/setup-mc/setup-stepper/setup-accounts/index.js
+++ b/js/src/setup-mc/setup-stepper/setup-accounts/index.js
@@ -101,6 +101,8 @@ const SetupAccounts = ( props ) => {
 	 * and the whole page would display this AppSpinner, which is not desired.
 	 */
 	const isLoadingJetpack = ! jetpack;
+	const isJetpackInactive = jetpack?.active !== 'yes';
+
 	const isLoadingGoogle = jetpack?.active === 'yes' && ! google;
 	const isLoadingGoogleMCAccount =
 		google?.active === 'yes' && scope.gmcRequired && ! googleMCAccount;
@@ -146,20 +148,27 @@ const SetupAccounts = ( props ) => {
 					'google-listings-and-ads'
 				) }
 			/>
-			<Section
-				className="gla-wp-google-accounts-section"
-				title={ __( 'Connect accounts', 'google-listings-and-ads' ) }
-				description={ __(
-					'The following accounts are required to use the Google for WooCommerce plugin.',
-					'google-listings-and-ads'
-				) }
-			>
-				<VerticalGapLayout size="large">
-					<WPComAccountCard jetpack={ jetpack } />
-					<GoogleAccountCard disabled={ isGoogleAccountDisabled } />
-					<GoogleAdsAccountCard />
-				</VerticalGapLayout>
-			</Section>
+			{ isJetpackInactive && (
+				<Section
+					className="gla-wp-google-accounts-section"
+					title={ __(
+						'Connect accounts',
+						'google-listings-and-ads'
+					) }
+					description={ __(
+						'The following accounts are required to use the Google for WooCommerce plugin.',
+						'google-listings-and-ads'
+					) }
+				>
+					<VerticalGapLayout size="large">
+						<WPComAccountCard jetpack={ jetpack } />
+						<GoogleAccountCard
+							disabled={ isGoogleAccountDisabled }
+						/>
+						<GoogleAdsAccountCard />
+					</VerticalGapLayout>
+				</Section>
+			) }
 			<Section
 				className="gla-google-mc-account-section"
 				description={ <GoogleMCDisclaimer /> }
--- a/tests/e2e/specs/setup-mc/step-1-accounts.test.js
+++ b/tests/e2e/specs/setup-mc/step-1-accounts.test.js
@@ -61,6 +61,7 @@ test.describe( 'Set up accounts', () => {
 		const wpAccountCard = setUpAccountsPage.getWPAccountCard();
 		await expect( wpAccountCard ).toBeEnabled();
 		await expect( wpAccountCard ).toContainText( 'WordPress.com' );
+		await expect( wpAccountCard.getByRole( 'button' ) ).toBeEnabled();
 
 		const googleAccountCard = setUpAccountsPage.getGoogleAccountCard();
 		await expect( googleAccountCard.getByRole( 'button' ) ).toBeDisabled();
@@ -139,10 +140,6 @@ test.describe( 'Set up accounts', () => {
 					'Connect the accounts required to use Google for WooCommerce.'
 				)
 			).toBeVisible();
-
-			await expect(
-				page.getByRole( 'button', { name: 'Connect' } ).first()
-			).toBeEnabled();
 
 			const wpAccountCard = setUpAccountsPage.getWPAccountCard();
 			await expect( wpAccountCard ).not.toBeVisible();
--- a/tests/e2e/specs/setup-mc/step-1-accounts.test.js
+++ b/tests/e2e/specs/setup-mc/step-1-accounts.test.js
@@ -151,7 +151,6 @@ test.describe( 'Set up accounts', () => {
 			const wpAccountCard = setUpAccountsPage.getWPAccountCard();
 			await expect( wpAccountCard ).not.toBeVisible();
 		} );
-
 	} );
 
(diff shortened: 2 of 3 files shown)


js/src/setup-mc/setup-stepper/setup-accounts/index.js
```diff
@@ -101,6 +101,8 @@ const SetupAccounts = ( props ) => {
 	 * and the whole page would display this AppSpinner, which is not desired.
 	 */
 	const isLoadingJetpack = ! jetpack;
+	const isJetpackInactive = jetpack?.active !== 'yes';
```
joemcgill: Nit: There are a couple other places below where we are checking for the `jetpack.active` value (e.g., `isLoadingGoogle` and `isGoogleAccountDisabled`. Let's consolidate all of this logic by updating those references to make use of the new variable you're adding here. For readability, I'd also suggest making the variable `isJetpackActive` rather than inactive, to avoid the double negative when checking for `! isJetpackInactive`, ex:

```js
const isJetpackActive = jetpack?.active === 'yes';

// ...etc.

{ ! isJetPackActive && ( <WPComAccountCard /> ) }
```

joemcgill (changes requested):
This looks right to me. One small code quality nitpick, but it looks like it's functioning correctly. 👍🏻

js/src/setup-mc/setup-stepper/setup-accounts/index.js
dsawardekar: Updated

dsawardekar:
@joemcgill I've refactored the code per the feedback. Also confirmed that settings still show the card, AC is also updated to reflect this. Back to you for another review.

joemcgill (commented):
Thanks @dsawardekar. The updates all look good to me. I tried running the e2e tests locally with these changes and noticed that this would require quite a few updates to the existing test suite, since many of the existing playwright tests depend on finding HTML elements in a particular order in the screen.

I went ahead and opened a [follow-up PR](https://github.com/woocommerce/google-listings-and-ads/pull/2521) that addresses those issues so you can review how these were previously set up.

Once you've taken a look, let's merge that PR into this branch prior to passing to QA for review.

dsawardekar:
Thanks, @joemcgill . I've merged the PR, and updated the AC to remove the non-mocking based instructions.

joemcgill (approved):
Thanks, @dsawardekar. Looks good to me.

ankitguptaindia:
@joemcgill, I have reviewed the PR and found it to be functioning as described ✅. However, I came across a use case where I would appreciate your feedback: 

**Current Behavior:** When a user completes the onboarding flow, the WordPress.com card is hidden if Jetpack is connected, and the status is displayed as "connected" on the plugin’s Settings tab.

**Use Case:** The button on the Settings screen labelled "Disconnect from all accounts" suggests that clicking it will disconnect all accounts, including the WordPress.com account. Furthermore, when the user clicks this button, a confirmation message (e.g., "I understand that I am disconnecting any WordPress.com account...") appears, reinforcing that the WordPress.com account will be disconnected.

Given this, a user might expect the WordPress.com card to reappear on the onboarding screen when they restart the plugin setup. However, with the current implementation, the WordPress.com card does not appear on the first screen of the onboarding steps after all accounts have been disconnected.

This discrepancy between the expected behaviour and the label/button details could cause some confusion for users.


https://github.com/user-attachments/assets/cb41c87d-0a6c-4235-9adc-77c488ddcb1a

<Details>
  <summary>Disconnection Screens</summary>
  
  
image


image

  

</Details>

---

**Additional Note:** I tested the same flow with the stable build and found that, when the user disconnects all accounts and restarts the onboarding journey, the WordPress.com card appears on the onboarding screen with a  "CONNECTED" status. 

<Details>
<summary>Screenshot</summary>

image


</Details>

joemcgill:
Thanks @ankitguptaindia! Re:

> Use Case: The button on the Settings screen labelled "Disconnect from all accounts" suggests that clicking it will disconnect all accounts, including the WordPress.com account. Furthermore, when the user clicks this button, a confirmation message (e.g., "I understand that I am disconnecting any WordPress.com account...") appears, reinforcing that the WordPress.com account will be disconnected.

This behavior is something that both @dsawardekar and I noticed as well. This seems to be the current behavior in the `develop` branch and not something related to the changes made here. I'm going to create a new issue to address this.

joemcgill:
@eason9487 this one is ready for your review.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -133,14 +133,44 @@ test.describe( 'Set up accounts', () => {
...
+
+			await expect(
+				page.getByText(
+					'Connect the accounts required to use Google for WooCommerce.'
+				)
+			).toBeVisible();
+
+			await expect(
+				page.getByRole( 'button', { name: 'Connect' } ).first()
+			).toBeEnabled();
```
eason9487: It would be a bit clearer to assert this button by specifying the Google account card.

https://github.com/woocommerce/google-listings-and-ads/blob/9e606c1b80d164f770ed7d5c29e8272a6ae0d62f/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L180-L184

The `.first()` in the code reference could be omitted.

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
```diff
@@ -69,31 +69,15 @@ export default class SetUpAccountsPage extends MockRequests {
...
 	/**
 	 * Get Jetpack description row.
 	 *
 	 * @return {import('@playwright/test').Locator} Get Jetpack description row.
 	 */
 	getJetpackDescriptionRow() {
-		return this.getCardDescriptionClass().first();
+		return this.getWPAccountCard().locator(
+			'.gla-account-card__description'
+		);
```
eason9487: `getJetpackDescriptionRow` could be removed as it's no longer used.

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
```diff
@@ -120,7 +108,9 @@ export default class SetUpAccountsPage extends MockRequests {
 	 * @return {import('@playwright/test').Locator} Get Google Ads title.
 	 */
 	getAdsTitleRow() {
-		return this.getCardTitleClass().nth( 2 );
+		return this.getGoogleAdsAccountCard().locator(
+			'.gla-account-card__title'
+		);
```
eason9487: `getAdsTitleRow` could be removed as it's no longer used.

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
```diff
@@ -177,22 +167,13 @@ export default class SetUpAccountsPage extends MockRequests {
...
-	}
-
 	/**
 	 * Get Jetpack connected label.
 	 *
 	 * @return {import('@playwright/test').Locator} Get Jetpack connected label.
 	 */
 	getJetpackConnectedLabel() {
-		return this.getConnectedLabelClass().first();
+		return this.getWPAccountCard().locator( '.gla-connected-icon-label' );
```
eason9487: `getJetpackConnectedLabel` is not in use as well.

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
```diff
@@ -201,7 +182,9 @@ export default class SetUpAccountsPage extends MockRequests {
 	 * @return {import('@playwright/test').Locator} Get Google connected label.
 	 */
 	getGoogleConnectedLabel() {
-		return this.getConnectedLabelClass().nth( 1 );
+		return this.getGoogleAccountCard().locator(
+			'.gla-connected-icon-label'
+		);
```
eason9487: `getGoogleConnectedLabel` is not in use as well.

js/src/setup-mc/setup-stepper/setup-accounts/index.js
```diff
@@ -101,16 +101,16 @@ const SetupAccounts = ( props ) => {
...
+
+	const isLoadingGoogle = isJetpackActive && ! google;
 	const isLoadingGoogleMCAccount =
 		google?.active === 'yes' && scope.gmcRequired && ! googleMCAccount;
 
 	if ( isLoadingJetpack || isLoadingGoogle || isLoadingGoogleMCAccount ) {
 		return <AppSpinner />;
 	}
 
-	const isGoogleAccountDisabled = jetpack?.active !== 'yes';
```
eason9487, line 112: Nice improvement! Thanks!

eason9487 (commented):
Thanks for the work and the change in `SetupAccounts` looks good.

Left a few suggestions for the e2e test. It would be great if unused methods could be removed.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
eason9487: This assertion and the two above it seem to be copied from:

https://github.com/woocommerce/google-listings-and-ads/blob/9e606c1b80d164f770ed7d5c29e8272a6ae0d62f/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L51-L63

Assuming the original intention of this assertion placed here is not to check the connect button in *Google account card*, considering the same assertion will be performed by another test case, then I would suggest removing this assertion.

joemcgill:
Re: https://github.com/woocommerce/google-listings-and-ads/pull/2516#issuecomment-2291417223

> > Use Case: The button on the Settings screen labelled "Disconnect from all accounts" suggests that clicking it will disconnect all accounts, including the WordPress.com account. Furthermore, when the user clicks this button, a confirmation message (e.g., "I understand that I am disconnecting any WordPress.com account...") appears, reinforcing that the WordPress.com account will be disconnected.
>
> This behavior is something that both @dsawardekar and I noticed as well. This seems to be the current behavior in the develop branch and not something related to the changes made here. I'm going to create a new issue to address this.

I'm no longer seeing this behavior today and the WP.com connection is being disconnected correctly when clicking "Disconnect all accounts" so this may have been due to a local configuration issue.

dsawardekar:
@joemcgill PR updated, back to you for CR. 

RE: Disconnection

I'm still seeing the disconnection issue. It appears to disconnect correctly, but shows an error on the FE. But if you refresh the page, everything except the Jetpack connection has disconnected. I'm currently using the jetpack REST override to reset the state. If there is a better way, please let me know.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -58,10 +58,6 @@ test.describe( 'Set up accounts', () => {
 			)
 		).toBeVisible();
 
-		await expect(
-			page.getByRole( 'button', { name: 'Connect' } ).first()
-		).toBeEnabled();
-
 		const wpAccountCard = setUpAccountsPage.getWPAccountCard();
 		await expect( wpAccountCard ).toBeEnabled();
 		await expect( wpAccountCard ).toContainText( 'WordPress.com' );
```
joemcgill, line 63: After removing the earlier assertion about the button being enabled, we should add it back here using the new `getWPAccountCard()` helper:

```suggestion
		await expect( wpAccountCard ).toContainText( 'WordPress.com' );
		await expect( wpAccountCard.getByRole( 'button' ) ).toBeEnabled();
```

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -133,24 +129,52 @@ test.describe( 'Set up accounts', () => {
...
+			await expect(
+				page.getByText(
+					'Connect the accounts required to use Google for WooCommerce.'
+				)
+			).toBeVisible();
 
-			await expect( jetpackDescriptionRow ).toContainText(
+			await expect(
+				page.getByRole( 'button', { name: 'Connect' } ).first()
+			).toBeEnabled();
```
joemcgill: Let's remove this assertion once the once the previous suggestion is updated.
```suggestion
```

joemcgill (approved):
Thanks for the updates

eason9487 (approved):
Thanks for the update for e2e tests! LGTM.

PR #2518, opened 2024-08-12
Fix error when Google Merchant Center account is undefined

jorgemd24:

Closes https://wordpress.org/support/topic/typeerror-cannot-read-properties-of-undefined-2/

This PR fixes an issue where `googleMCAccount` is `undefined` when checking if `notification_service_enabled` is active. This can happen if the Google account isn’t connected or is missing scopes, leading to the following error:

`Cannot read properties of undefined (reading ‘notification_service_enabled’)`

https://github.com/woocommerce/google-listings-and-ads/blob/90e9747c463df9007e72637452ea5eb327400ab8/js/src/hooks/useGoogleMCAccount.js#L37-L47

1. Go here: https://github.com/woocommerce/google-listings-and-ads/blob/90e9747c463df9007e72637452ea5eb327400ab8/src/API/Site/Controllers/Google/AccountController.php#L168 and set the `active` to `no`
2. Go to the Settings page.
3. Check that you don't see the following error:

image

### Changelog entry

> Fix - Error when Google Merchant Center account is undefined while checking the notification service enabled property.

### DIFF
--- a/js/src/components/enable-new-product-sync-notice.js
+++ b/js/src/components/enable-new-product-sync-notice.js
@@ -26,8 +26,8 @@ const EnableNewProductSyncNotice = () => {
 	// Do not render if already switch to new product sync.
 	if (
 		! hasGoogleMCAccountFinishedResolution ||
-		! googleMCAccount.notification_service_enabled ||
-		googleMCAccount.wpcom_rest_api_status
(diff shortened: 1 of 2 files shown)


puntope (approved):
LGTM

PR #2519, opened 2024-08-13
Set empty props as objects in WPCOM Proxy responses

puntope:

Context: p1723540799285679/1722414955.111429-slack-C02BB3F30TG

In this PR we convert empty array properties into empty objects. 

### Screenshots:
Screenshot 2024-08-13 at 18 47 37

1. Hardcode an empty prop like `$data['test'] = []`; in the first line of the new `prepare_data()` function to simulate we get an empty array in a prop.
2. See the prop is returned as {} 
3. Rest of the props still working as before.

### Changelog entry

> Fix - Return empty array props as empty objects in WCOM Proxy responses

### DIFF
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -158,7 +158,7 @@ class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
 		add_filter(
 			'rest_request_after_callbacks',
 			/**
-			 * Add the Google Listings and Ads settings to the settings/general response.
+			 * Add the Google for WooCommerce and Ads settings to the settings/general response.
 			 *
 			 * @param WP_REST_Response|WP_HTTP_Response|WP_Error|mixed $response The response object.
 			 * @param mixed                                             $handler  The handler.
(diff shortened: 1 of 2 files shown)


src/Integration/WPCOMProxy.php
```diff
@@ -169,36 +169,59 @@ function ( $response, $handler, $request ) {
 					return $response;
 				}
 
+				$data = $response->get_data();
```
jorgemd24, line 172: The `$response` can be one of these types: `WP_REST_Response`, `WP_HTTP_Response`, `WP_Error`, or `mixed`. If `$response` isn't a `WP_REST_Response`, it will throw an error because `get_data` or `set_data` might no be defined.

jorgemd24 (commented):
Thanks @puntope for working on this. I left a comment about calling `get_data` and `set_data` without checking the $response type.

puntope:
Thx @jorgemd24 I updated the code so now it returns early if the response is not WP_REST_Response. Can you take another look?

jorgemd24 (approved):
Thanks @puntope for the adjustment. LGTM 👍

PR #2520, opened 2024-08-13
Add Shipping Max time column in gla_shipping_times Table

jorgemd24:

Part of pcTzPl-2qP-p2

This PR updates the DB to support minimum and maximum shipping times. We'll keep the existing `time` column (which will be used as min time) for backwards compatibility and add a new column called `max_time`, which will be populated with the current values from the `time` column for existing users.

Once the UI is set up to handle max and min times, we’ll update the backend to manage these new columns accordingly.

1. Edit your Free Listing to have different countries and shipping times. 
2. Update the Migration class to your current GLA version: https://github.com/woocommerce/google-listings-and-ads/pull/2520/files#diff-898d6eb44e5ad0a4ef4f1b8ac7af29f6d561f5ddfb30b1ab79398dd81a629df8R46 for example: `2.8.1`
3. Adjust the `gla_db_version` option to an older version, such as 2.8.0, so that the migration class is triggered.: `UPDATE wp_options SET option_value = '2.8.0' WHERE option_name = 'gla_db_version';`
4. Refresh the GFW page.
5. Check the DB and see that there is a new column called `max_time` and is populated with the values used in `time`.
6. To simulate a new installation run the following queries: 

```
DROP TABLE wp_gla_shipping_times;
DELETE FROM wp_options WHERE `option_name` = 'gla_db_version';
```
7. Refresh and see that a new column is added. 

### Additional details:
I didn’t include tests since this code will only run once.

### Changelog entry

>

### DIFF
--- a/src/DB/Migration/Migration20240813T1653383133.php
+++ b/src/DB/Migration/Migration20240813T1653383133.php
@@ -1,0 +1,62 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\DB\Migration;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\DB\Table\ShippingTimeTable;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class Migration20240813T1653383133
+ *
+ * Migration class to enable min and max time shippings.
+ *
+ * @see pcTzPl-2qP
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\DB\Migration
+ *
+ * @since x.x.x
+ */
+class Migration20240813T1653383133 extends AbstractMigration {
+
+	/**
+	 * @var ShippingTimeTable
+	 */
+	protected $shipping_time_table;
+
+	/**
+	 * Migration constructor.
+	 *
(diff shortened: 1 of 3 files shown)


src/DB/Migration/Migration20240813T1653383133.php
```diff
@@ -0,0 +1,62 @@
...
+		return 'x.x.x';
+	}
+
+	/**
+	 * Apply the migrations.
+	 *
+	 * @return void
+	 */
+	public function apply(): void {
+		if ( $this->shipping_time_table->exists() && ! $this->shipping_time_table->has_column( 'max_time' ) ) {
```
jorgemd24, line 55: Since the table installation happens earlier (even if the table already exists) and is handled by the delta function, the table should already be updated, so this code likely won't run. However, I think it's a good idea to keep it as a precaution.

https://github.com/woocommerce/google-listings-and-ads/blob/9f669d440f01309cac77c94d57f4c6b203729ae7/src/DB/Table.php#L45

https://developer.wordpress.org/reference/functions/dbdelta/

puntope (approved):
LGTM

Screenshot 2024-08-14 at 10 47 53

I see the new max_time created and populated correctly.

PR #2526, opened 2024-08-14
Adjust auto-sync shipping option

jorgemd24:

Part of pcTzPl-2qP-p2

This PR includes the following changes:

1. Removes the "Recommended" label from the auto-sync option.
2. By default, selects the manual option ("My shipping settings are simple. I can manually estimate flat shipping rates.") if no shipping settings were previously stored, such as during onboarding.
3. Hides the auto-sync option during onboarding if the merchant has no shipping methods set up.

### Onboarding:

https://github.com/user-attachments/assets/e8676c18-944f-4e7f-9f49-a724e9656a4f

### Editing Page (the auto-sync option is always displayed, regardless of shipping settings):

https://github.com/user-attachments/assets/78c6ece8-2869-4161-b30d-a91b00f484fb

1. Disconnect all your accounts and remove any shipping method in WC.
2. Go to GFW onboarding.
3. When setting up the Shipping Rates, see that the `auto-sync` option is not available.
4. In WC add a new shipping method.
5. Refresh the onboarding page.
6. See that the `auto-sync` option is available.
7. See that always the option: `My shipping settings are simple. I can manually estimate flat shipping rates` is checked.
8. Finish the onboarding.
9. Go to Edit Free Listings and see that the `auto-sync` option is always displayed, regardless of shipping settings

### Additional details:
- Local Pickup isn't currently recognized as a valid shipping method, but it looks like we're not syncing it to Google either. So, I think it's fine if Local Pickup isn't counted as a valid method in this case.
- While working on this, I discovered a bug. If you use a comma as the decimal separator, the shipping rate won't be recognized as valid because `is_numeric` will return false. I'll create the issue.

image

https://github.com/woocommerce/google-listings-and-ads/blob/fe41b5be9da361607780bb913a7a4c4e553fbe3c/src/Shipping/ZoneMethodsParser.php#L128-L131

### Changelog entry

>

### DIFF
--- a/js/src/components/shipping-rate-section/shipping-rate-section.js
+++ b/js/src/components/shipping-rate-section/shipping-rate-section.js
@@ -32,9 +32,7 @@ const ShippingRateSection = () => {
 	const hideAutomatticShippingRate =
 		! settings?.shipping_rates_count &&
 		hasFinishedResolution &&
-		mcSetup?.status === 'incomplete'
-			? true
-			: false;
+		mcSetup?.status === 'incomplete';
 
 	return (
 		<Section
--- a/js/src/components/shipping-rate-section/shipping-rate-section.js
+++ b/js/src/components/shipping-rate-section/shipping-rate-section.js
@@ -64,14 +64,9 @@ const ShippingRateSection = () => {
 							{ ! hideAutomatticShippingRate && (
 								<AppRadioContentControl
 									{ ...inputProps }
-									label={ createInterpolateElement(
-										__(
-											'Automatically sync my store’s shipping settings to Google.',
-											'google-listings-and-ads'
-										),
-										{
-											strong: <strong></strong>,
-										}
+									label={ __(
+										'Automatically sync my store’s shipping settings to Google.',
+										'google-listings-and-ads'
 									) }
 									value="automatic"
 									collapsible
--- a/js/src/components/shipping-rate-section/shipping-rate-section.js
+++ b/js/src/components/shipping-rate-section/shipping-rate-section.js
@@ -14,6 +14,8 @@ import RadioHelperText from '.~/wcdl/radio-helper-text';
 import AppDocumentationLink from '.~/components/app-documentation-link';
 import VerticalGapLayout from '.~/components/vertical-gap-layout';
 import FlatShippingRatesInputCards from './flat-shipping-rates-input-cards';
+import useSettings from '.~/components/free-listings/configure-product-listings/useSettings';
+import useMCSetup from '.~/hooks/useMCSetup';
 
 /**
  * @fires gla_documentation_link_click with `{ context: 'setup-mc-shipping', link_id: 'shipping-read-more', href: 'https://support.google.com/merchants/answer/7050921' }`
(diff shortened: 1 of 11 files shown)


src/API/Site/Controllers/MerchantCenter/SettingsController.php
```diff
@@ -71,6 +90,9 @@ protected function get_settings_endpoint_edit_callback(): callable {
 			}
 
 			foreach ( $schema as $key => $property ) {
+				if ( ! in_array( 'edit', $property['context'] ?? [], true ) ) {
```
jorgemd24, line 93: I added this so it skips only view properties; otherwise, it would save them as options, like the `shipping_rates_count` property.

js/src/components/shipping-rate-section/shipping-rate-section.js
```diff
@@ -22,8 +24,18 @@ import FlatShippingRatesInputCards from './flat-shipping-rates-input-cards';
...
+	const { hasFinishedResolution, data: mcSetup } = useMCSetup();
 	const inputProps = getInputProps( 'shipping_rate' );
 
+	// Hide the automatic shipping rate option if there are no shipping rates and the merchant is onboarding.
+	const hideAutomatticShippingRate =
+		! settings?.shipping_rates_count &&
+		hasFinishedResolution &&
+		mcSetup?.status === 'incomplete'
+			? true
+			: false;
```
puntope: This can be simplified as just `mcSetup?.status === 'incomplete'`

js/src/components/shipping-rate-section/shipping-rate-section.js
```diff
@@ -51,27 +63,29 @@ const ShippingRateSection = () => {
...
+									{ ...inputProps }
+									label={ createInterpolateElement(
+										__(
+											'Automatically sync my store’s shipping settings to Google.',
+											'google-listings-and-ads'
+										),
+										{
+											strong: <strong></strong>,
+										}
 									) }
```
puntope: Since we dont use anymore <strong> tag we can omit  createInterpolateElement and just use 

```js
label={
					__(
											'Automatically sync my store’s shipping settings to Google.',
											'google-listings-and-ads'
										)
}
								
```

puntope:
@jorgemd24 I set a shipping method (flat) but still, I don't see the auto-sync option. If I add free shipping yes it works.



https://github.com/user-attachments/assets/4f787242-26e2-4948-b6da-76926442502f

jorgemd24:
Thanks @puntope for the review and your suggestions! 

>  I set a shipping method (flat) but still, I don't see the auto-sync option. If I add free shipping yes it works.

Could it be that the issue is related to using a comma instead of a dot for decimal pricing? I encountered a similar problem while working on this #2527, Could you share a screenshot of how the flat rate is set up?

Also, could you provide a screenshot of the response from the settings endpoint?

image

Thanks!

puntope:
> Could it be that the issue is related to using a comma instead of a dot for decimal pricing? I encountered a similar problem while working on this https://github.com/woocommerce/google-listings-and-ads/pull/2527, Could you share a screenshot of how the flat rate is set up?

Hi @jorgemd24 I tested it again. Seems like it's because the price field was empty. If I add a value (0, -1, 0.0, 0,0) it works


> Also, could you provide a screenshot of the response from the settings endpoint?



```
{
    "shipping_rate": "flat",
    "shipping_time": "flat",
    "tax_rate": "destination",
    "website_live": true,
    "checkout_process_secure": true,
    "payment_methods_visible": true,
    "refund_tos_visible": true,
    "contact_info_visible": true,
    "shipping_rates_count": 0
}
```

jorgemd24:
> Seems like it's because the price field was empty

Hi @puntope, you need to have valid shipping methods for the auto sync option to be displayed; otherwise, it won't show up. For example, something like this should work:

image

The original idea was to show the `auto-sync` option only if there are valid shipping methods. The logic for syncing shipping to MC was already implemented, and this PR is simply leveraging that code.

puntope (approved):
LGTM @jorgemd24 Thanks for clarifying the issue with the empty shipping method.

PR #2527, opened 2024-08-14
Fix issue with comma separators for Shipping Rates

### LINKED ISSUE #2470: Using a decimal separator other than dot may cause shipping data not to be synced
### Describe the bug:

If the decimal separator in WooCommerce core currency options is not `.` and shipping data is obtained from WooCommerce core, the shipping data may not be synced to Google Merchant Center.

This issue occurs in all relevant fields:

- Single shipping cost
- Shipping class costs
- Free shipping threshold

### Steps to reproduce:

1. Go to **WooCommerce > Settings > General** and set the decimal separator in the currency options to `,`.
   image
2. Go to **WooCommerce > Settings > Shipping > Shipping zones** and set up a zone with:
   - Free shipping and minimum order amount, e.g. `999,99`
      image
   - Flat rate with single cost or class costs.
      image
3. Go to edit free listings.
4. Select the **Recommended:  Automatically sync my store’s shipping settings to Google** option for the shipping rates setting.
5. Add the above configured zone to the Audience setting.
6. Save changes.
7. Wait a moment for the settings to sync to Google Merchant Center.
8. Go to Google Merchant Center and navigate to **Shipping and returns** page.
9. No shipping data has been synchronized to **Shipping policies**.

image

### Expected behavior:

The shipping costs should be parsed according to the currency options set in the WooCommerce core and synced correctly to Google Merchant Center.

### Actual behavior:

The shipping costs may not be parsed and omitted to sync to Google Merchant Center when the decimal separator in WooCommerce core currency options is not using `.`.

### Additional details:

How WooCommerce core handles amount formatting: https://github.com/woocommerce/woocommerce/blob/9.0.2/plugins/woocommerce/includes/shipping/flat-rate/class-wc-shipping-flat-rate.php#L108-L115

jorgemd24:

While working on [PR #2526](https://github.com/woocommerce/google-listings-and-ads/pull/2526), I noticed that if the flat rate cost is entered with a comma separator, it's considered invalid because the following function returns false:

https://github.com/woocommerce/google-listings-and-ads/blob/40a068a287c385e9ee4fd658ec332e2b80d018fc/src/Shipping/ZoneMethodsParser.php#L130-L133

This PR addresses the issue by handling the value similarly to how WooCommerce does it here:

https://github.com/woocommerce/woocommerce/blob/91afd71b96e55338affb9284d22df0c38297ee09/plugins/woocommerce/includes/shipping/flat-rate/class-wc-shipping-flat-rate.php#L85-L112

Specifically, it replaces the decimal separators with a `.`.

### Flat Rate
1. Create a Shipping Zone for some countries, and create a flat rate method, when you introduce the cost set this to `,`
2. Call the following endpoint: `GET mc/shipping/rates/suggestions?country_codes[]=YOUR_COUNTRY`
3. See that the response contains the rate that you created.  Check out develop and see that the rate is not returned.

### Flat Rate with "No Shipping Class Cost" and Free Shipping
- Set the "No Shipping Class Cost" using comma decimals.
- Create a Free Shipping method with a minimum requirement. The decimal values don't matter since only [integer numbers](https://github.com/woocommerce/woocommerce/blob/91afd71b96e55338affb9284d22df0c38297ee09/plugins/woocommerce/includes/shipping/free-shipping/class-wc-shipping-free-shipping.php#L30) are allowed.
- The expected result should be the flat rate cost plus the "No Shipping Class Cost," as the [other classes are ignored](https://github.com/woocommerce/google-listings-and-ads/blob/6310f9c00372b348ac248822146bf44f2ce524ac/src/Shipping/ShippingSuggestionService.php#L68-L70).

image

image

image

### Check in MC

Follow the steps mentioned here: #2470 and check that the shipping rates sync correctly, whether you use commas or dots as the separator.

### Additional details:

I noticed that if you use "Select All Countries," save, and then refresh the page, it appears that WooCommerce isn’t saving the option, and GFW doesn’t detect any Shipping Zone either. I didn’t have time to investigate further this week, but I’ll look into it next week to see if there’s an issue with WooCommerce.

image

### Changelog entry

> Fix - issue with comma separators for Shipping Rates.


src/Shipping/ZoneMethodsParser.php
```diff
@@ -123,7 +123,9 @@ protected function get_flat_rate_method_rate( object $method ): ?float {
 		$rate = null;
 
 		$flat_cost = 0;
-		$cost      = $method->get_option( 'cost' );
+		$locale    = localeconv();
+		$decimals  = [ wc_get_price_decimal_separator(), $locale['decimal_point'], $locale['mon_decimal_point'], ',' ];
+		$cost      = str_replace( $decimals, '.', (string) $method->get_option( 'cost' ) );
```
mikkamp: This looks like a great candidate for a helper function, both to clarify what is being done here as well as making it reusable. It also seems like WC already has [wc_format_decimal](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-formatting-functions.html#function_wc_format_decimal). Should we add that to the WC proxy class, so we can simplify testing as a bonus?

mikkamp (commented):
Thanks for the change. It works as expected, but as mentioned in the comment I think we should move this to a helper function, it would also make it easier to add a test for it in the ZoneMethodsParserTest.

Also probably not a big deal since in most currencies shipping costs won't surpass a thousand. But it would still ignore shipping costs with a thousand separator. I suppose since that's semi-broken in WooCommerce as well (allowed to enter but doesn't show up correctly as a shipping cost), we can ignore this.

mikkamp:
Can we also link issue #2470 so we don't work on this twice. This is also another reason why we might need to move it to a helper, since the issue describes multiple locations where the number is parsed incorrectly.

src/Shipping/ZoneMethodsParser.php
jorgemd24: I tried using the `wc_format_decimal` function, but there is an issue. If your decimal separator is `.` and you use `wc_format_decimal('2,4', '')`, it returns `24.00`, which is incorrect. The issue is that `wc_format_decimal` doesn't specifically replace the comma; it just uses the decimal settings stored in WooCommerce and the local environment:

https://github.com/woocommerce/woocommerce/blob/91afd71b96e55338affb9284d22df0c38297ee09/plugins/woocommerce/includes/wc-formatting-functions.php#L292

On the other hand, the Shipping Flat Rate function explicitly replaces the comma:

https://github.com/woocommerce/woocommerce/blob/91afd71b96e55338affb9284d22df0c38297ee09/plugins/woocommerce/includes/shipping/flat-rate/class-wc-shipping-flat-rate.php#L86

jorgemd24:
Thanks, @mikkamp, for the suggestion! I moved the logic to a helper function and updated the other part of the code to address all the points mentioned in issue #2470. I realized there's no need to handle decimal places for the minimum value in Free Shipping Methods since it's converted to an integer anyway, so the main issue was with the flat rate and the classes.

https://github.com/woocommerce/woocommerce/blob/91afd71b96e55338affb9284d22df0c38297ee09/plugins/woocommerce/includes/shipping/free-shipping/class-wc-shipping-free-shipping.php#L30

Do you mind to have a second look? Thanks!

src/Shipping/ZoneMethodsParser.php
mikkamp: Thanks for clarifying that I failed to see the difference of the extra `,` being added to the array.

My question though, is that the right way to handle it? Because if we enter a shipping rate of `2,399.99`, then it will be evaluated as `2.399.99`, which I don't think is correct either, and won't pass an `is_numeric` test. If I have a store setup with `.` as the decimal separator and ',' as the thousand separator, then I don't expect it to be replacing `,` for decimals. So in that case I'd see 2,4 being replaced with 24 as the "more correct" way of handling it.

jorgemd24: Thanks for the discussion. It's a tricky situation because the issue arises from the flexibility users have when setting prices in the modal. It allows any characters as decimal or thousand separators.

The solution I proposed aligns best with how the current WC Shipping Flat Rate handles these cases, although it was not ideal. However, I’ve adjusted the function to strip out any thousand separators and convert the string into a valid decimal number. For example, 1.234,01 will be converted to 1234.01. The thousand separator is more of a formatting issue and doesn’t affect syncing rates or operating with the numbers itself. If we want to format the number, we can use the string returned from the function and apply any desired decimal or thousand separators.

jorgemd24:
The PHP units are failing because it seems that there is an issue with the latest WC and mapping the  attributes and the sku, but it doesn't seem related to this PR. Will look into this next week if no one has done it before.

mikkamp:
> The PHP units are failing because it seems that there is an issue with the latest WC and mapping the attributes and the sku, but it doesn't seem related to this PR. Will look into this next week if no one has done it before.

Fixed in #2559

src/Shipping/ZoneMethodsParser.php
mikkamp: Thanks for the clarification, it is indeed tricky.

Although I don't totally agree with the validation of the modal as if I leave it to the default with no thousand separator then I'm only allowed to use a `.` and a comma is not allowed.
image

If I do have both symbols included then it accepts any combination as long as they are either numbers or a thousand separator, and a single decimal separator. So entering `1,,23.45,,99` is totally acceptable even though it won't make sense. In your case that will be converted to 12345.99

That's fine for now as generally we will expect sane values being entered, but I think in the long run it's the modal that should do a bit better validation there.

mikkamp (approved):
Thanks for the additional changes, I understand the inconsistency so I see we can't account for everything.

The shipping rates are now syncing correctly. There is one error I'm still getting:
```
action failed via Async Request: { "error": { "code": 400, "message": "[services[1].rateGroups[0].singleValue] Precision for price 12.156 exceeds maximum precision of 2 decimal digits.", "errors": [ { "message": "[services[1].rateGroups[0].singleValue] Precision for price 12.156 exceeds maximum precision of 2 decimal digits.", "domain": "global", "reason": "invalid" } ] } }
```

I think regardless of the decimal setting in WC we need to always ensure it doesn't exceed 2 places. Note that the modal doesn't allow this many decimal places, but if you right click the edit button to open in a new tab, it goes to the old input form where more decimal places can be added. Most 3PD shipping rates don't support the modal so we might not always have that validation in place.

Either way though since all the expected shipping rates are syncing I'll go ahead and approve the PR and we can just add a small tweak to always round to a max of 2 decimals.

jorgemd24:
Thanks @mikkamp for fixing the issue with the unit tests and for catching this:

> There is one error I'm still getting:

> action failed via Async Request: { "error": { "code": 400, "message": "[services[1].rateGroups[0].singleValue] Precision for price 12.156 exceeds maximum precision of 2 decimal digits.", "errors": [ { "message": "[services[1].rateGroups[0].singleValue] Precision for price 12.156 exceeds maximum precision of 2 decimal digits.", "domain": "global", "reason": "invalid" } ] } }

I'll add the tweak in a follow-up PR.

PR #2544, opened 2024-08-19
Migrate jest tests to use Node.js 20 and make components compatible with upgraded packages

### LINKED ISSUE #1098: Replace deprecated keyCode atttribute in Verification Code Control.
## Summary

Checking into [verification-code-control.js](https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/components/contact-information/phone-number-card/verification-code-control.js)

I saw a chunk of code using the deprecated attribute `keyCode`.  It should be replaced by `code` attribute 

Check here the deprecation notice: 
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

## Technical

Notice the constants at the beginning

```
const KEY_CODE_LEFT = 37;
const KEY_CODE_RIGHT = 39;
const KEY_CODE_BACKSPACE = 8;
```

Should be also replaced by the correspondent key code. Check -> https://keycode.info/ in order to know which key should be used.

Would be interesting to search over the codebase for more cases like this.

eason9487:

Closes #1098

To make this repo use Node.js 20, this PR migrates jest tests along with two rule fixes for eslint:

- Adjust the settings so that jest can run:
   - Fix that `is-plain-obj@4.1.0` doesn't provide the CommonJS build.
   - Force `uuid` to resolve with the CommonJS entry point as jest doesn't support `package.json.exports`.
   - Fix `@woocommerce/components` still using incompatible `@woocommerce/currency`.
   - Ignore that `qrcode.react` module is incorrectly listed in dev dependencies of `@automattic/components`.
   - Fix the React versioning conflicts between `@wordpress/*` and `@woocommerce/*`.
- Update timer mocks to deprecate the `timers` option.
- Migrate test cases to `@testing-library/user-event` v14.
   - Invoke `userEvent.setup()` in the test itself and before the component is rendered.
   - Adding `async` and `await` as all APIs are asynchronous.
   - No need to be wrapped in `act`.
   - Replace the simulation of pressing buttons with the `keyboard` API.
   - `clear()` is changed to only support editable elements only.
- Prevent form submission during testing.
- The ARIA role of `` is considered as `presentation` if its `alt` is an empty string.
- Update snapshots to reflect the format changes.
- Fix `CampaignNameCell` tests by replacing `fireEvent` with `userEvent`.
- Resolve eslint errors for the `testing-library/no-wait-for-multiple-assertions` rule.
- Resolve eslint errors for the `testing-library/no-unnecessary-act` rule.
- Ignore the deprecation warnings of `defaultProps` from React by calling `console.error()`.
- Remove the `structuredClone` polyfill no longer needed.

In addition, it also needs a few adjustments to components or their tests to be compatible with upgraded `@wordpress/components` and `@woocommerce/components`:

- Prevent the `<Dashboard>` tests from calling `window.fetch`.
- Adjust `<Stepper>` related tests to wait for stepper content to be rendered.
- Make `<AppTooltip>` compatible with the `placement` prop introduced by WordPress 6.4 to replace `position`.
- Rewrite the replacer of `JSON.stringify` in `getReportKey` to ensure it returns the same key regardless of the query keys' order.
- Replace deprecated `event.keyCode` with `event.code` for `<VerificationCodeControl>`.
- Avoid accidentally using the event object to reset the asset group in `<CampaignAssetsForm>`.

1. Set Node.js to v20
2. Run `npm install` to see if it can finish without errors
3. Check if the following scripts work well
   - `npm run lint:js`
   - `npm run test:js`
4. Go to edit the phone number of contact information in the Setting page
   - In the verification code inputs, check if Arrow-Left, Arrow-Right and Backspace keyboards can move the input cursor as expected
5. Go to the Product Feed page
   - Hover to the edit icon on the Product Feed table
   - Check if the tooltip is shown and centered above the icon
   image
6. View the e2e test result on GitHub Actions: https://github.com/woocommerce/google-listings-and-ads/actions/runs/10452211003/job/28940184227
7. Do some smoke testing

### Changelog entry

> Tweak - Make the Tooltip use the new placement prop when WordPress >= 6.4.
> Tweak - Replace deprecated event.keyCode with event.code for the verification code inputs in the contact information setting.
> Dev - Migrate jest tests to use Node.js 20.
> Dev - Avoid accidentally using the event object to reset the asset group values in the CampaignAssetsForm component.
> Dev - Rewrite the replacer of JSON.stringify in getReportKey to ensure it returns the same key regardless of the query keys' order.


js/src/tests/dependencies/woocommerce/settings.js
```diff
@@ -1,3 +1,19 @@
...
 	return {};
 }
+
+export function getSetting( key, backup ) {
+	return global.wcSettings[ key ] || backup;
+}
+
+// Jest is currently running tests with npm packages bundled with WordPress 6.4.
+export function isWpVersion( version, operator ) {
+	if ( version === '6.4' && operator === '<' ) {
```
puntope, line 14: Not sure if this is correct, I expected some logic for comparing versions

js/src/data/utils.js
```diff
@@ -85,6 +85,18 @@ export function getReportQuery( category, type, query, dateReference ) {
 	return reportQuery;
 }
 
+function replacer( key, value ) {
```
puntope, line 99: Can we document what is the purpose of this function?

js/src/components/app-tooltip/index.js
```diff
@@ -3,12 +3,34 @@
...
+// It's an inverse map of the following reference and contains only the values used in this repo.
+// Ref: https://github.com/WordPress/gutenberg/blob/wp/6.4/packages/components/src/popover/utils.ts#L17-L71
+const PLACEMENT_TO_POSITION = {
+	'top-start': 'top right',
+	top: 'top center',
+};
+
+// compatibility-code "WP < 6.4" -- `position` prop in Tooltip is deprecated since WP 6.4
+function mapPlacementToPosition( props ) {
+	if ( isWpVersion( '6.4', '<' ) ) {
```
puntope, line 22: This will be always false

js/src/tests/dependencies/woocommerce/settings.js
eason9487: This mock function is used for jest testing only and therefore doesn't implement the actual comparison logic. Add a more detailed description in https://github.com/woocommerce/google-listings-and-ads/pull/2544/commits/c9a259fadc079c01303d7edc7c5535163783f6f5.

js/src/data/utils.js
eason9487: Sure. Added in https://github.com/woocommerce/google-listings-and-ads/pull/2544/commits/69fc1b485b03de3c2e7cf5e7162013bd8f3fb588.

js/src/components/app-tooltip/index.js
eason9487: The result will depend on the actual version of installed WordPress, not the one mocked in `js/src/tests/dependencies/woocommerce/settings.js`.

eason9487:
Hi @puntope, thanks for the review. This PR is ready for a new round. Could you take a look?

js/src/tests/dependencies/woocommerce/settings.js
puntope: Ah gotcha, I believe because the big amount of files I missed this was a mock for tests.

puntope (approved):
✅  LGTM

Tested the next:

✅  Adjust the settings so that jest can run -- When these changes are not in place the tests return errors. 

- Fix that is-plain-obj@4.1.0 doesn't provide the CommonJS build.
- Force uuid to resolve with the CommonJS entry point as jest doesn't support package.json.exports.
- Fix @woocommerce/components still using incompatible @woocommerce/currency.
- Ignore that qrcode.react module is incorrectly listed in dev dependencies of @automattic/components.
- Fix the React versioning conflicts between @wordpress/* and @woocommerce/*.

✅  Update timer mocks to deprecate the timers option. 
✅  Migrate test cases to @testing-library/user-event v14.
✅  Smoke test Dashboard
✅  Smoke test <AppTooltip> 
✅  Smoke test reports 
✅  Smoke test phone verification code keyboard functionality.

PR #2547, opened 2024-08-20
Use remote-site-status to check the WPCOM Auth status

puntope:

Context: p1724064639654099/1722414955.111429-slack-C02BB3F30TG

When authorizing Google to fetch data from the Shop as well as generating a token we set an option `gla_wpcom_rest_api_status` as `approved`. If the token is revoked outside the extension, the option will still be `approved` and the notifications will be sent. 

This PR handles that and checks the status of the token using the WPCOM REST API.

1.  Disable fetch data from  Settings or disconnect it from the Connection Test page
2. Grant access from scratch still works.
3. You can send notifications in the Connection test page for example
4. Now disconnect WPCOM Access from Connection Test page
5. You CANNOT send notifications anymore
6. Set `gla_wpcom_rest_api_status` as `approved` to simulate a granted access that has no token.
7. You CANNOT send notifications anymore
8. Go to settings. You see the warning notice in the MC Card
9. Grant access again
10. Now You can send notifications again in the Connection test page 

### Additional details:

ℹ️  I was thinking about removing `gla_wpcom_rest_api_status`  and only relying on the transient + WPCOM REST API. status endpoint for considering the WPCOM API as ready. I would appreciate thoughts on this. 

### Changelog entry

> Tweak - Use remote-site-status to check the WPCOM Auth status

### DIFF
--- a/src/API/WP/NotificationsService.php
+++ b/src/API/WP/NotificationsService.php
@@ -171,10 +171,11 @@ class NotificationsService implements Service, OptionsAwareInterface {
 	 * If the Notifications are ready
 	 * This happens when the WPCOM API is Authorized and the feature is enabled.
 	 *
+	 * @param bool $with_health_check If true. Performs a remote request to WPCOM API to get the status.
 	 * @return bool
 	 */
-	public function is_ready(): bool {
-		return $this->options->is_wpcom_api_authorized() && $this->is_enabled() && $this->merchant_center->is_ready_for_syncing() && $this->account_service->is_wpcom_api_status_healthy();
+	public function is_ready( bool $with_health_check = true ): bool {
+		return $this->options->is_wpcom_api_authorized() && $this->is_enabled() && $this->merchant_center->is_ready_for_syncing() && ( $with_health_check === false || $this->account_service->is_wpcom_api_status_healthy() );
 	}
 
 	/**
--- a/src/API/WP/NotificationsService.php
+++ b/src/API/WP/NotificationsService.php
@@ -72,7 +72,7 @@ class NotificationsService implements Service, OptionsAwareInterface {
 	 * Class constructor
 	 *
 	 * @param MerchantCenterService $merchant_center
-	 * @param AccountService $account_service
+	 * @param AccountService        $account_service
 	 */
 	public function __construct( MerchantCenterService $merchant_center, AccountService $account_service ) {
 		$blog_id                = Jetpack_Options::get_option( 'id' );
--- a/src/API/WP/NotificationsService.php
+++ b/src/API/WP/NotificationsService.php
@@ -5,6 +5,7 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\API\WP;
 
 use Automattic\Jetpack\Connection\Client;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
+use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\AccountService;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
@@ -59,15 +60,24 @@ class NotificationsService implements Service, OptionsAwareInterface {
 	 */
 	public MerchantCenterService $merchant_center;
 
+	/**
+	 * The AccountService service
+	 *
+	 * @var AccountService $account_service
+	 */
+	public AccountService $account_service;
+
 
 	/**
 	 * Class constructor
 	 *
 	 * @param MerchantCenterService $merchant_center
+	 * @param AccountService $account_service
 	 */
-	public function __construct( MerchantCenterService $merchant_center ) {
+	public function __construct( MerchantCenterService $merchant_center, AccountService $account_service ) {
 		$blog_id                = Jetpack_Options::get_option( 'id' );
 		$this->merchant_center  = $merchant_center;
+		$this->account_service  = $account_service;
 		$this->notification_url = "https://public-api.wordpress.com/wpcom/v2/sites/{$blog_id}/partners/google/notifications";
 	}
 
@@ -164,7 +174,7 @@ class NotificationsService implements Service, OptionsAwareInterface {
 	 * @return bool
 	 */
 	public function is_ready(): bool {
-		return $this->options->is_wpcom_api_authorized() && $this->is_enabled() && $this->merchant_center->is_ready_for_syncing();
+		return $this->options->is_wpcom_api_authorized() && $this->is_enabled() && $this->merchant_center->is_ready_for_syncing() && $this->account_service->is_wpcom_api_status_healthy();
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


src/MerchantCenter/AccountService.php
```diff
@@ -623,4 +632,51 @@ public function update_wpcom_api_authorization( string $status, string $nonce ):
...
+				'url'     => 'https://public-api.wordpress.com/wpcom/v2/sites/' . Jetpack_Options::get_option( 'id' ) . '/wc/partners/google/remote-site-status',
+				'user_id' => get_current_user_id(),
+			];
+
+			$integration_remote_request_response = Client::remote_request( $integration_status_args, null );
+
+			if ( is_wp_error( $integration_remote_request_response ) ) {
+				$this->delete_wpcom_api_status_transient();
+				return false;
+			}
```
jorgemd24: If the token isn't healthy or if the user disconnect Jetpack, does that mean we'll still make a request every time , even though we know the token isn't valid? Shouldn't we store that status in the transient as well?

I also noticed that the `SyncerHooks` class is registered every time the plugin is initialized, which means we could end up calling this endpoint repeatedly without end.

https://github.com/woocommerce/google-listings-and-ads/blob/e8e51f30affac3f3eda6cb5d179067311e4b15f0/src/Infrastructure/GoogleListingsAndAdsPlugin.php#L91

https://github.com/woocommerce/google-listings-and-ads/blob/e8e51f30affac3f3eda6cb5d179067311e4b15f0/src/Settings/SyncerHooks.php#L82-L85

tests/Unit/API/WP/NotificationsServiceTest.php
```diff
@@ -5,6 +5,8 @@
 
 use Automattic\WooCommerce\Admin\RemoteInboxNotifications\TransformerService;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\WP\NotificationsService;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\WP\OAuthService;
```
jorgemd24: unused dependency

src/MerchantCenter/AccountService.php
```diff
@@ -623,4 +632,51 @@ public function update_wpcom_api_authorization( string $status, string $nonce ):
...
+
+			$status = json_decode( wp_remote_retrieve_body( $integration_remote_request_response ), true ) ?? [];
+
+			if ( isset( $status['error'] ) ) {
```
jorgemd24: The response from the remote status endpoints returns `errors` instead of `error`, should it be `error`?

jorgemd24 (commented):
Thanks, @puntope, for working on this. It works as you described in the PR, but I’ve left some concerns about how we’re handling the errors returned by the remote status endpoint.

puntope:
Thanks for your useful review @jorgemd24 I addressed your comments so now I believe this is ready for a new round

src/MerchantCenter/AccountService.php
```diff
@@ -227,6 +229,11 @@ public function get_connected_status(): array {
 		$id                    = $this->options->get_merchant_id();
 		$wpcom_rest_api_status = $this->options->get( OptionsInterface::WPCOM_REST_API_STATUS );
 
+		// If token is revoked outside the extension. Set the status as error to force the merchant to grant access again.
+		if ( $wpcom_rest_api_status === 'approved' && ! $this->is_wpcom_api_status_healthy() ) {
```
jorgemd24, line 233: What do you think if we update the `WPCOM_REST_API_STATUS` accordingly? Or Maybe is better just to delete the option so then the user can start the onboarding again.

src/API/WP/NotificationsService.php
```diff
@@ -161,10 +171,11 @@ public function get_notification_url(): string {
...
 	 * This happens when the WPCOM API is Authorized and the feature is enabled.
 	 *
+	 * @param bool $with_health_check If true. Performs a remote request to WPCOM API to get the status.
 	 * @return bool
 	 */
-	public function is_ready(): bool {
-		return $this->options->is_wpcom_api_authorized() && $this->is_enabled() && $this->merchant_center->is_ready_for_syncing();
+	public function is_ready( bool $with_health_check = true ): bool {
+		return $this->options->is_wpcom_api_authorized() && $this->is_enabled() && $this->merchant_center->is_ready_for_syncing() && ( $with_health_check === false || $this->account_service->is_wpcom_api_status_healthy() );
 	}
```
jorgemd24, line 179: Should we add a test for this to ensure that when `with_health_check` is set to false, `is_wpcom_api_status_healthy` is never called?

src/MerchantCenter/AccountService.php
```diff
@@ -623,4 +632,45 @@ public function update_wpcom_api_authorization( string $status, string $nonce ):
...
+			];
+
+			$integration_remote_request_response = Client::remote_request( $integration_status_args, null );
+
+			if ( is_wp_error( $integration_remote_request_response ) ) {
+				$this->delete_wpcom_api_status_transient();
+				return false;
+			}
+
+			$status = json_decode( wp_remote_retrieve_body( $integration_remote_request_response ), true ) ?? [];
```
jorgemd24: If `json_decode` returns `null`, we currently set it to an empty array, which means we'll keep trying to get the status repeatedly. Instead, shouldn't we set the transient to something like `['is_healthy' => false]` to avoid querying the endpoint?

src/MerchantCenter/AccountService.php
jorgemd24: Thanks for the changes, but I still see this issue. If the response is `wp_error`, we will keep querying the endpoint. Wouldn't it be better to update the transient with the status error instead?

jorgemd24 (commented):
Thanks for the changes, @puntope! However I’m still seeing some possible issues—see my comments. Thanks!

puntope:
Hi @jorgemd24 for the review. I updated again the code with your suggestions:

- New tests here: https://github.com/woocommerce/google-listings-and-ads/pull/2547/files#diff-f34e5f2ef622a3c8104b654486ebfe6d8e6343a5c0a0c17437ff42d77c6d2fe3R226-R236
- Set the Option as error if the status is not healthy https://github.com/woocommerce/google-listings-and-ads/pull/2547/files#diff-c3732a2e44ad6802f55b880fa6023ec164cc9862b1520e96f9c3198918fc8d3bR235
- Set the status as ['is_healthy' => false] in case is wp_error or json_decode is null. https://github.com/woocommerce/google-listings-and-ads/pull/2547/files#diff-c3732a2e44ad6802f55b880fa6023ec164cc9862b1520e96f9c3198918fc8d3bR666-R670

Can we do a new round of review?

Thanks

jorgemd24 (commented):
Thanks for the adjustments, @puntope! One last thing: if the token isn't valid, shouldn't we reset the `gla_wpcom_rest_api_status` option? Currently, if I disconnect the GFW app, it doesn't send the notification, but the status still shows as "approved."

image

puntope:
>  if the token isn't valid, shouldn't we reset the gla_wpcom_rest_api_status option? Currently, if I disconnect the GFW app, it doesn't send the notification, but the status still shows as "approved."

Hi @jorgemd24 Thanks for the review. Can you elaborate the steps for this use case? As far as I understand, When the token is invalid we set the transient and option as "error". (See `AccountService::get_connected_status` and `AccuntService::is_wpcom_api_status_healthy`)

jorgemd24 (approved):
> Hi @jorgemd24 Thanks for the review. Can you elaborate the steps for this use case? As far as I understand, When the token is invalid we set the transient and option as "error". (See AccountService::get_connected_status and AccuntService::is_wpcom_api_status_healthy)

Thanks for the clarification! I think this happened because I had the transient set to `healthy.` Then, I disconnected the Google App and refreshed the settings page, but since the transient was still set to "healthy" it appeared as if it was approved. But now I realize this was just a coincidence when I was doing the review and that the issue is temporary. So, I think we can move forward with this PR!

PR #2551, opened 2024-08-20
Streamline campaign setup: Remove ads audience field from paid ads setup during onboarding

asvinb:
### Changes proposed in this Pull Request:

Closes #2501 

* Removed the `AudienceSection` from the `PaidAdsSetupSections` component.
* The `countryCodes` prop in the `paidAds` state variable is now initialized with the values of `targetAudience` which itself contains the values from the second step of the onboarding process.

### Detailed test instructions:

1. Go through the onboarding flow.
2. Note the countries selected in Step 2 in the **Audience** section (**Location**). You can choose specific countries or all countries.
3. On the fourth step of the flow, there should not longer be the "Ads Audience" section. (Refer to the [screenshot in the ticket description](https://github.com/woocommerce/google-listings-and-ads/issues/2501))
4. Create a new paid ad.
5. Go to the dashboard and edit the ad which was just created.
6. Check if the countries displayed match those which were selected in Step 2 of the onboarding process.

### Changelog entry

> Update - Remove ads audience field from paid ads setup during onboarding.

Issue #2535, opened 2024-08-16
Consolidate the ad creation step in the Ads Setup flow with the one used in Onboarding


joemcgill:
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2460

When a merchant is setting up a new ads campaign from the plugin dashboard, they enter the "Set up paid campaign" flow, which includes individual steps for connecting a Google Ads account, configuring a campaign, and completing their billing setup.

<img width="1084" alt="image" src="https://github.com/user-attachments/assets/8a5d15ee-dc60-400e-8ca4-a12b913dcfdf">

Currently, the second step is where a merchant configures their campaign, including setting their targeted audience(s) and their daily budget.

We have a similar form on the last step of onboarding which is being simplified as part of #2458. 

<img width="913" alt="image" src="https://github.com/user-attachments/assets/a8c20d88-87a1-42c8-a7f6-de6269062ace">

Maintaining two different components for this same action is redundant so we will consolidate on a single component that is used in both flows. However, the form displayed in the "Set up paid campaign" flow is also used when editing current campaigns, so the ability to set your audience will need to be optionally shown if we're editing a campaign.

 

## Acceptance Criteria

- [ ] When a user is on the "Create your paid campaign" step of the "Set up paid ads" flow, they will see the same form used on the "Create a campaign" step of the "Get started..." onboarding flow.
- [ ] The merchant can set their daily budget for their campaign
- [ ] If a merchant hasn't set up billing for their Ads account, they will not be able to complete the form and the button will be disabled.
- [ ] If a merchant hasn't set up billing, show the billing setup component
- [ ] A user should be able to successfully set up billing from this form
- [ ] Once billing is set up, the completion button should be enabled and the clicking should successfully create a campaign
- [ ] When editing a campaign, the user should no longer see target audience fields, following #2501.
- [ ] ~When editing a campaign, the user should be able to modify the target audiences~ (no longer applicable)

## Implementation Brief

The PR should get merged into the [feature/2460-simplify-paid-ads-setup](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2460-simplify-paid-ads-setup) branch.

[...]

Issue #2503, opened 2024-08-06
Campaign Creation: Add minimum budget limit


joemcgill:
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2459

As a follow-up to #2502, even when we initialize the `BudgetSection` component with the recommended daily budget amount, merchants can edit the value to a lower value. For example, a merchant could set a campaign for 1 cent per day. 

We already have budget recommendations for each country, so we could take a % of that recommendation and set it up as a minimum. For example, if the budget recommendation for US is $15/day, the minimum possible budget should be $5/day (~30% of recommended budget).

If a user tries to input a number lower than the minimum, we should display a validation error with an explanation of what needs to change.

_Example mockup_

<img width="315" alt="image" src="https://github.com/user-attachments/assets/2f853bd6-4268-4620-b0ae-09461d583edb">

## Acceptance Criteria

- [ ] A minimum possible value for the Daily Average Cost will be 30% of the recommended daily budget
- [ ] If the Daily Average Cost input is set to a value lower than the minimum possible budget, a validation warning is shown reading "Please make sure daily average cost is greater than %d." where %d is replaced with the minimum.
- [ ] If the Daily Average Cost input is set to a value lower than the minimum possible budget, the validation error is shown
- [ ] If the Daily Average Cost input is set to a value lower than the minimum possible budget, the "Complete setup" button is disabled.

## Implementation Brief
The component where this is displayed is the `BudgetSection` component, but the validation logic will need to be applied in the `CampaignAssetsForm` component by updating the validation logic of the `validateCampaign` function ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/a55258332f0c46faaf1834d28e30a6a44a691d49/js/src/components/paid-ads/validateCampaign.js)).

## Test Coverage
E2E tests for step 4 of the setup process is updated to cover examples where the minimum value is set and when a value less than the minimum is set.

joemcgill:
Consolidating into #2502.

joemcgill:
Actually, no, setting a minimum budget limit will be a separate task from setting the initial value, which is what will be handled in #2502. Reopening this to define requirements for setting the minimum limit.

eclarke1:
Adding back the custom labels as this was not appearing in our 'Detailed Epic View'

eclarke1:
@joemcgill  please could you re-add the size and Eng Focus labels (once the AC/IB has been added of course)

joemcgill:
[...]

PR #2552, opened 2024-08-22
Streamline campaign setup: Change the campaign setup and creation to use the recommended budget as the initial value

### LINKED ISSUE #2502: Campaign Creation: Update Budget Setup Card

Image

For the Budget card, we want to explain to users that: 
1. They will be charged by clicks on ads and can pause the campaign whenever they want.
2. We recommend running the campaign at least for 1 month because it uses AI to optimize the campaign to reach better results, so it takes time to learn the signals of what works for their business.

Also, the warning message of “With a budget lower than your competitor range, your campaign may not get noticeable results.” appears from the start, before putting any number. Merchants might not connect that warning message to the actual budget they’re putting afterwards. 

To set proper expectations with users we will automatically start with a the input set at the minimum recommended value and hide the initial budget warning from this card. 

We'll also update the copy in the blue box to read something like this:

>  “We recommend running campaigns at least 1 month so it can learn to optimize for your business. Tip: Most merchants targeting United States (US) set a daily budget of $15 USD.”

The recommendation amount will still be location-based as it is currently.

 

## Acceptance Criteria

- [ ] The budget warning message is no longer shown when the budget card first loads.
- [ ] The Daily Average Cost input should be set to the minimum recommended value initially instead of 0.00
- [ ] When the Daily Average Cost input is changed, the budget warning should be shown only if the number is lower than the recommended daily average.
- [ ] The copy in the blue box should read `We recommend running campaigns at least 1 month so it can learn to optimize for your business. Tip: Most merchants targeting United States (US) set a daily budget of $%d USD.`, where %d is replaced by the recommended daily budget.
- [ ] When multiple target locations are selected, the location with the highest recommended daily budget is used (no change from current behavior)

## Implementation Brief
The budget warning is already wired up so that it won't show if the daily average cost is higher than the recommended daily average, so the main thing to solve here is to ensure the initial form value is set to the recommended value.

The `BudgetRecommendations` component makes use of the `useFetchBudgetRecommendationEffect()` hook to fetch the current recommendations. This will need to be moved up to the `PaidAdsSetupSections` component. Currently the default amount for campaigns is set to `0` [here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js#L33-L38). This can be set using the `useFetchBudgetRecommendationEffect()` value.

**Note for testing:** when this screen is loaded, the last value set gets stored in session storage in a `gla-onboarding-paid-ads` key. This will need to be cleared in order to see changes.

The budget recommendations copy for both the warning and the tip (blue box) can be edited in `js/src/components/paid-ads/budget-section/budget-recommendation/index.js`.

## Test Coverage

E2E tests for `tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js` need to be updated to confirm the behavior of the initial value, the visibility of the budget warning, and any tests confirming the copy of the budget tip needs to be updated.

ankitrox:

Closes #2502  .

### Screenshots:

Screenshot 2024-08-22 at 9 40 27 PM

1. Proceed with the onboarding flow and complete first 3 steps of the onboarding flow.
2. On the 4th step, notice that `Daily average cost` input will have the pre-populated value with highest budget in the recommended budgets. For example:
  * If UAE (14 USD), Hong Kong (6 USD)and Vietnam (5 USD) are selected, the input should be populated with 14.00
3. Go through the setup ads flow/create a new paid campaign from the dashboard, the budget input should be pre populated with the highest budget value.
4. The screenshot shows the ability of changing/updating countries but this is no longer the case now.

### Changelog entry

> Update - Change the campaign setup and creation to use the recommended budget as the initial value and adjust its description.


tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
```diff
@@ -266,16 +284,11 @@ test.describe( 'Complete your campaign', () => {
 							textContent
 						);
 
-					const responsePromise =
```
joemcgill: Why do we no longer need to wait for this response after the country setting is changed? It looks like when the country values change, the text that reads "Tip: Most merchants targeting similar countries set a daily budget of %d USD" does change, so we should probably ensure that is still being tested.

js/src/components/paid-ads/budget-section/budget-recommendation/index.js
```diff
@@ -50,17 +49,19 @@ function toRecommendationRange( isMultiple, ...values ) {
...
 
 	if ( ! data ) {
 		return null;
 	}
 
 	const { currency, recommendations } = data;
-	const { daily_budget: dailyBudget, country } =
-		getHighestBudget( recommendations );
+	const { daily_budget: dailyBudget, country } = getHighestBudget(
+		recommendations.filter( ( recommendation ) =>
```
joemcgill: I was able to trigger a case where `recommendations` was initially undefined, which caused this to throw a fatal error:

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at BudgetRecommendation (index.js:61:19)
    at renderWithHooks (react-dom.development.js:15486:18)
    at mountIndeterminateComponent (react-dom.development.js:20098:13)
    at beginWork (react-dom.development.js:21621:16)
    at HTMLUnknownElement.callCallback (react-dom.development.js:4164:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
    at invokeGuardedCallback (react-dom.development.js:4277:31)
    at beginWork$1 (react-dom.development.js:27485:7)
    at performUnitOfWork (react-dom.development.js:26591:12)
    at workLoopSync (react-dom.development.js:26500:5)
```

To do so, I changed the selected audience countries in step 2 and then this triggered when I got to step 4.

tests/e2e/utils/pages/setup-ads/setup-budget.js
```diff
@@ -12,6 +12,15 @@ export default class SetupBudget extends MockRequests {
 		this.page = page;
 	}
 
+	/**
+	 * Get budget recommendation tip section.
+	 *
+	 * @return {import('@playwright/test').Locator} The budget recommendation text row.
+	 */
+	getBudgetRecommendationTip() {
+		return this.page.locator( '.components-tip p' );
```
joemcgill: If there are multiple `Tip` components on this screen, then this will fail. Better to target the `.gla-budget-recommendation` class for the locator, since the `toContainText()` assertion will find any text inside that node.

```suggestion
		return this.page.locator( '.gla-budget-recommendation > .components-tip' );
```

joemcgill (changes requested):
This is looking really good, but issues I've noted a few issues where this could be improved.

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
ankitrox: @joemcgill This is no longer require because we are fulfilling the budget recommendation [here](https://github.com/woocommerce/google-listings-and-ads/pull/2552/files#diff-3b03eaec7b7b793a56912cfe86e6d384ea90da173ddd1f9855ba1fa5cd80270bR87-R104), so we no longer have to wait to capture the budget recommendation response.

Those assertions will still be tested with the new fulfilment.

tests/e2e/utils/pages/setup-ads/setup-budget.js
joemcgill: Addressed in [103ae13](https://github.com/woocommerce/google-listings-and-ads/pull/2552/commits/103ae13b1474ed672ee9c492561923df3d47e9ca)

ankitrox:
Thank you @joemcgill for your feedback.

I've addressed your comments and responded to one of your comments related to removing the response promise [here](https://github.com/woocommerce/google-listings-and-ads/pull/2552#discussion_r1731898598).

Please let me know if there's anything else you want me to look into. Over to you for another round of review. 

Thanks again!

joemcgill (commented):
@ankitrox I'm still seeing some unexpected behavior with these changes. If you go back to a previous step after after the budget recommendations have been displayed and then return to the last step, the budget recommendations are no longer showing. I've added a video to show what I'm seeing. Can you give this another look?

https://github.com/user-attachments/assets/96a72295-67e8-41fd-b4c8-e592252f975a

Could you also provide more context for [this question about the e2e test changes](https://github.com/woocommerce/google-listings-and-ads/pull/2552#discussion_r1731898598)?

ankitrox:
@joemcgill 
Thanks for the video. It helped me understand the issue well. 

The issue was happening because `PaidAdsSetupSections` component was getting rendered before `recommendationData` is resolved.

There was also an issue of amount getting changed to `dailyBudget` even if user has entered some value previously which gets stored in sessionStorage. This has been fixed now.

js/src/components/paid-ads/budget-section/budget-recommendation/index.js
joemcgill: This appears to be fixed.

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
joemcgill: @ankitrox I'm still unsure why we need to remove this. Could you clarify?

joemcgill (approved):
Thanks @ankitrox. This looks like it's working better now. I still have a question about one of the E2E changes that you made, which would be good to clarify, but I'm going to approve and send to @ankitguptaindia for QA.

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
ankitrox: @joemcgill Sorry for missing this one.

Initially when we were removing country from (or updating anything in `Select country/s` textbox), a request to budget recommendation endpoint was triggered which we were awaiting to get fulfilled in `setupBudgetPage.registerBudgetRecommendationResponse()`

Now, as per our new implementation, that request will no longer trigger because we are using the countries that have been selected in step 2 which acts as a superset for step 4 data.

Please let me know if this helps or in case of any other query.

ankitrox:
Hi @joemcgill 

Apologies for missing the comment previously.

I have responded to the query [here](https://github.com/woocommerce/google-listings-and-ads/pull/2552#discussion_r1731898598)

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.1
* WooCommerce - Version 9.2.3 
* PHP: 8.3
* Web Server: Nginx
* Browser: Chrome - Version 127
* OS: macOS Sonoma 14.6.1 

</Details>





**Test Results -** 

Changes are working as expected. tested all related test scenarios and all are working well. 

https://github.com/user-attachments/assets/b2a82a7d-ded2-45bc-9fde-9f3c52dabfc6

ankitguptaindia (approved):

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -93,6 +93,8 @@ const BudgetSection = ( { formProps, disabled = false, children } ) => {
 							<BudgetRecommendation
 								countryCodes={ countryCodes }
 								dailyAverageCost={ amount }
+								currency={ currency }
```
eason9487: `BudgetRecommendation` component doesn't use the `currency` prop.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -11,6 +11,7 @@ import { Form } from '@woocommerce/components';
 import useGoogleAdsAccount from '.~/hooks/useGoogleAdsAccount';
 import useTargetAudienceFinalCountryCodes from '.~/hooks/useTargetAudienceFinalCountryCodes';
 import useGoogleAdsAccountBillingStatus from '.~/hooks/useGoogleAdsAccountBillingStatus';
+import useFetchBudgetRecommendationEffect from '.~/components/paid-ads/budget-section/budget-recommendation/useFetchBudgetRecommendationEffect';
```
eason9487: What I understand is that the same logic of setting the default budget will then be applied to all ad creation forms, so moving the `useFetchBudgetRecommendationEffect` hook to the `.~/hooks` directory would be more appropriate.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -78,6 +96,14 @@ export default function PaidAdsSetupSections( { onStatesReceived } ) {
 	const { hasGoogleAdsConnection } = useGoogleAdsAccount();
 	const { data: targetAudience } = useTargetAudienceFinalCountryCodes();
 	const { billingStatus } = useGoogleAdsAccountBillingStatus();
+	const { data: recommendationData } =
+		useFetchBudgetRecommendationEffect( targetAudience );
```
eason9487: It may cause API to fail with a 400 bad request error.

image

Steps:

1. Proceed to onboarding step 4
2. Refresh webpage

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -127,12 +153,27 @@ export default function PaidAdsSetupSections( { onStatesReceived } ) {
...
+		if ( dailyBudget !== undefined ) {
+			// If the amount is already set in session, use that one.
+			const sessionData = clientSession.getCampaign();
+			const { amount: sessionAmount } = sessionData;
+
+			setPaidAds( ( currentPaidAds ) => {
+				currentPaidAds = {
+					...currentPaidAds,
+					amount: sessionAmount || dailyBudget,
+					recommendationData,
```
eason9487: Wonder why `recommendationData` needs to be a part of the `paidAds` state?

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -143,6 +184,7 @@ export default function PaidAdsSetupSections( { onStatesReceived } ) {
 	const initialValues = {
 		countryCodes: paidAds.countryCodes,
 		amount: paidAds.amount,
+		recommendationData: paidAds.recommendationData,
```
eason9487: The `recommendationData` is not used as the form value, so passing it around this way is not ideal.

js/src/components/paid-ads/budget-section/budget-recommendation/index.js
```diff
@@ -50,17 +49,19 @@ function toRecommendationRange( isMultiple, ...values ) {
...
+			countryCodes.includes( recommendation.country )
+		)
```
eason9487: I'm curious about in which case it needs this filter?

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -35,8 +36,25 @@ const defaultPaidAds = {
...
+ * then the budget recommendation should be (20 USD).
+ */
+function getHighestBudget( recommendations ) {
+	return recommendations.reduce( ( defender, challenger ) => {
+		if ( challenger.daily_budget > defender.daily_budget ) {
+			return challenger;
+		}
+		return defender;
+	} );
+}
```
eason9487: This function is completely the same as:

https://github.com/woocommerce/google-listings-and-ads/blob/096c0866ad7f6b310410077f0aa4c9e57079fc13/js/src/components/paid-ads/budget-section/budget-recommendation/index.js#L15-L29

Suggest considering whether it can be integrated into the `useFetchBudgetRecommendationEffect` hook or extracted as a util to eliminate duplication.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -127,12 +153,27 @@ export default function PaidAdsSetupSections( { onStatesReceived } ) {
...
+
+			setPaidAds( ( currentPaidAds ) => {
+				currentPaidAds = {
+					...currentPaidAds,
+					amount: sessionAmount || dailyBudget,
+					recommendationData,
+				};
+				return resolveInitialPaidAds( currentPaidAds, targetAudience );
+			} );
+		}
```
eason9487: If there is an amount restored from client session, the `currentPaidAds.amount` is already that value when the first time running this useEffect, so it may not need to read from client session again.

https://github.com/woocommerce/google-listings-and-ads/blob/096c0866ad7f6b310410077f0aa4c9e57079fc13/js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js#L112-L115

Additionally, moving this logic into the `resolveInitialPaidAds` function would be better as it intends to resolve the initial values.

tests/e2e/utils/pages/setup-ads/setup-budget.js
```diff
@@ -12,6 +12,17 @@ export default class SetupBudget extends MockRequests {
 		this.page = page;
 	}
 
+	/**
+	 * Get budget recommendation tip section.
+	 *
+	 * @return {import('@playwright/test').Locator} The budget recommendation text row.
```
eason9487: This JSDoc description is not quite consistent.

eason9487 (changes requested):
Left some suggestions, questions, and issues that need to be resolved.

In addition, if I understand correctly, post-onboarding can still select audience countries when creating a campaign via the `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Fcampaigns%2Fcreate` path.

However, there are now two issues with this page:

- The recommended budget block is not shown
- There is no API triggered to get the recommended budget after changing the audience countries.

image

eason9487:
The #2551 seems to have some code conflicts with this PR and even some of the logic might need to be rewritten afterward, so it's recommended to defer adjustment of this PR and rebase this PR onto #2551 after it's approved and merged.

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -55,6 +57,14 @@ export default function AdsCampaign( {
 				'google-listings-and-ads'
 		  );
 
+	// Set the amount from client session if it is available.
+	useEffect( () => {
```
asvinb: @ankitrox  Wondering if we can set the amount in the initialValues prop for the forms directly instead of setting the initial form value in the child components?

- https://github.com/woocommerce/google-listings-and-ads/blob/24044c53e4a1e84116086291f02bef8e26af7372/js/src/pages/create-paid-ads-campaign/index.js#L132
- https://github.com/woocommerce/google-listings-and-ads/blob/24044c53e4a1e84116086291f02bef8e26af7372/js/src/setup-ads/setup-ads-form.js#L32

asvinb (changes requested):
@ankitrox I left a comment about setting the initial values. Can you let me know what you think please?
Also there'll be additional work once #2551 is merged.

js/src/components/paid-ads/ads-campaign.js
ankitrox: @asvinb Thank you.

This component is not using the `<Form>` component like `PaidAdsSetupSections` component, thus we cannot pass the initial value, also I have updated this side effect to run when the recommendation is loaded and budget section will be mounted when the data is loaded, till then spinner card will be shown.

asvinb: @ankitrox I see `AdsCampaign` being used in 3 places:
* `js/src/pages/create-paid-ads-campaign/index.js` - the initial values are set via `CampaignAssetsForm`
* `js/src/pages/edit-paid-ads-campaign/index.js` - the initial values are set via `CampaignAssetsForm`. Here we don't need to worry about using grabbing `amount` from `localStorage` since it's for editing a campaign.
* `js/src/setup-ads/ads-stepper/index.js` - The `AdsStepper` component is used by `SetupAdsForm` which also makes use of `CampaignAssetsForm` to set the initial values.

So that leaves only `js/src/pages/create-paid-ads-campaign/index.js` and `js/src/setup-ads/ads-stepper/index.js` where we can potentially set the amount to the value from localStorage. Maybe we can even create a hook for it to reduce the code duplication.

Let me know what you think.

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -55,6 +61,24 @@ export default function AdsCampaign( {
 				'google-listings-and-ads'
 		  );
 
+	const { data: budgetData, loading } = useFetchBudgetRecommendationEffect(
+		formContext.values.countryCodes
+	);
+
+	const { country = '', daily_budget: dailyBudget } = getHighestBudget(
+		budgetData?.recommendations || []
```
asvinb: @ankitrox We already have a check here https://github.com/woocommerce/google-listings-and-ads/pull/2552/files#diff-3bdecec9939ad884905f627b3905c8b138f73d843890bb6d41799694c57f2ab8R10 in case `recommendations` is falsy. Shall we return an empty array in the `getHighestBudget` function instead of `null`?

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -91,13 +115,23 @@ export default function AdsCampaign( {
...
 				formProps={ formContext }
 			/>
-			<BudgetSection
-				formProps={ formContext }
-				disabled={ disabledBudgetSection }
-				countryCodes={ formContext.values.countryCodes }
-			>
-				<CampaignPreviewCard />
-			</BudgetSection>
+			{ ! loading && (
```
asvinb: @ankitrox This behaviour is different from what we currently have, i.e having a spinner and loading state while the data is being loaded. Is there a particular reason why we need to wait for the recommendations?

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -91,13 +115,23 @@ export default function AdsCampaign( {
...
+				<BudgetSection
+					formProps={ formContext }
+					disabled={ disabledBudgetSection }
+					countryCodes={ formContext.values.countryCodes }
+					country={ country }
```
asvinb: @ankitrox We are passing those new props `country`, `dailyBudget` and `isMultiple` to `BudgetSection`. However those are being then passed to `BudgetRecommendation`. Can't we get those values directly in `BudgetRecommendation`?

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -91,13 +115,23 @@ export default function AdsCampaign( {
...
+					dailyBudget={ dailyBudget }
+					isMultiple={ multipleRecommendations }
+				>
+					<CampaignPreviewCard />
+				</BudgetSection>
+			) }
+			{ loading && (
```
asvinb: @ankitrox See my comment above.

js/src/components/paid-ads/budget-section/budget-recommendation/index.js
```diff
@@ -50,21 +33,19 @@ function toRecommendationRange( isMultiple, ...values ) {
...
 const BudgetRecommendation = ( props ) => {
-	const { countryCodes, dailyAverageCost = Infinity } = props;
-	const { data } = useFetchBudgetRecommendationEffect( countryCodes );
-	const map = useCountryKeyNameMap();
-
-	if ( ! data ) {
-		return null;
-	}
+	const {
+		isMultiple,
```
asvinb: I see the `countryCodes` prop was removed but it's still being passed to `BudgetRecommendation` in `js/src/components/paid-ads/budget-section/index.js`

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -29,15 +29,21 @@ const nonInteractableProps = {
...
+ * @param {string} props.country Country code.
+ * @param {Array<CountryCode>} props.countryCodes Country codes to fetch budget recommendations for.
+ * @param {number} props.dailyBudget Daily budget.
  * @param {boolean} [props.disabled=false] Whether display the Card in disabled style.
+ * @param {boolean} [props.isMultiple=false] Whether the campaign is targeting multiple countries.
  * @param {JSX.Element} [props.children] Extra content to be rendered under the card of budget inputs.
  */
 const BudgetSection = ( {
 	formProps,
+	country,
```
asvinb: @ankitrox See my previous comment about passing those props.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -76,6 +78,13 @@ export default function PaidAdsSetupSections( {
 	const isBillingCompleted =
 		billingStatus?.status === GOOGLE_ADS_BILLING_STATUS.APPROVED;
 
+	const { data: budgetData, loading } =
+		useFetchBudgetRecommendationEffect( countryCodes );
+	const { country = '', daily_budget: dailyBudget } = getHighestBudget(
+		budgetData?.recommendations || []
+	);
+	const multipleRecommendations = budgetData?.recommendations.length > 1;
```
asvinb: @ankitrox The code is duplicated here. Can we move it to a hook which can be reused elsewhere?

asvinb (changes requested):
@ankitrox I left some more comments. Can you kindly let me know about the suggested approach about setting the default value for amount via the `initialValues` prop of `CampaignAssetsForm`.
Also you need to check when there's no recommended budget to not display the recommendation. This is what I can see on my side while testing:
image

js/src/components/paid-ads/ads-campaign.js
ankitrox: @asvinb I've created the new hook `useBudgetRecommendationData`, but kept the `useEffect`in same component because if we move up to `SetupAdsForm`  component as well as for `CreatePaidAdsCampaign`, it will create the network error issue for `useFetchBudgetRecommendationEffect` because country codes can be undefined or blank and since these are hooks, we won't be able to call them conditionally.

ankitrox:
@asvinb I've created the new hook `useBudgetRecommendationData`, but kept the `useEffect`in same component because if we move up to `SetupAdsForm`  component as well as for `CreatePaidAdsCampaign`, it will create the network error issue for `useFetchBudgetRecommendationEffect` because country codes can be undefined or blank and since these are hooks, we won't be able to call them conditionally.

js/src/hooks/useBudgetRecommendationData.js
```diff
@@ -0,0 +1,45 @@
...
+	const [ country, setCountry ] = useState( '' );
+	const [ dailyBudget, setDailyBudget ] = useState( 0 );
+	const [ multipleRecommendations, setMultipleRecommendations ] =
+		useState( false );
+	const [ recommendations, setRecommendations ] = useState( [] );
+
+	useEffect( () => {
+		if ( ! loading ) {
+			const {
+				country: budgetCountries = '',
```
asvinb: @ankitrox This will cause an error in the app when `getHighestBudget` returns `null` as you have defined here: https://github.com/woocommerce/google-listings-and-ads/pull/2552/files#diff-3bdecec9939ad884905f627b3905c8b138f73d843890bb6d41799694c57f2ab8R10

js/src/hooks/useBudgetRecommendationData.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+import useFetchBudgetRecommendationEffect from './useFetchBudgetRecommendationEffect';
+import getHighestBudget from '.~/utils/getHighestBudget';
+
+const useBudgetRecommendationData = ( countryCodes ) => {
+	const { data, loading } =
+		useFetchBudgetRecommendationEffect( countryCodes );
+
+	const [ country, setCountry ] = useState( '' );
+	const [ dailyBudget, setDailyBudget ] = useState( 0 );
```
asvinb: @ankitrox Instead of defining all those state variables, can we not define only one state variable which is an object having all those properties?

js/src/hooks/useBudgetRecommendationData.js
```diff
@@ -0,0 +1,45 @@
...
+	useEffect( () => {
+		if ( ! loading ) {
+			const {
+				country: budgetCountries = '',
+				daily_budget: recommendedDailyBudget,
+			} = getHighestBudget( data?.recommendations );
+
+			setCountry( budgetCountries );
+			setDailyBudget( recommendedDailyBudget );
+			setMultipleRecommendations( data?.recommendations.length > 1 );
```
asvinb: @ankitrox Do we need to store that in a state variable? We can have a variable where `data?.recommendations.length > 1` and return the variable.

js/src/setup-ads/setup-ads-form.js
```diff
@@ -67,14 +68,19 @@ const SetupAdsForm = () => {
...
+	const handleChange = ( _, values, isValid ) => {
 		const args = [ initialValues, values ].map(
 			( { countryCodes, ...v } ) => {
 				v.countrySet = new Set( countryCodes );
 				return v;
 			}
 		);
 
+		// Set the amount in session storage.
+		if ( isValid && _?.name === 'amount' ) {
```
asvinb: @ankitrox Why not use `values.amount`?

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
```diff
@@ -108,14 +124,18 @@ export default function PaidAdsSetupSections( {
...
 	};
 
 	return (
 		<Form
 			initialValues={ initialValues }
 			onChange={ ( _, values, isValid ) => {
 				setPaidAds( { ...paidAds, ...values, isValid } );
+
+				if ( isValid ) {
+					clientSession.setCampaign( values );
```
asvinb: @ankitrox The campaign is already being updated in local storage here: 
https://github.com/woocommerce/google-listings-and-ads/pull/2552/files#diff-6b9f6824ae1208e70b059bf81217131f2d1c42fad1664514c2280500e8e1a7b6R114

asvinb (dismissed):
@ankitrox I left some more comments. Can you check them out please?
Most importantly I think we can reduce a lot of the complexity around `countryCodes`. Right now, we need to fetch the budget recommendation data once the country code changes and the country codes are tied to the form. In https://github.com/woocommerce/google-listings-and-ads/issues/2535 we are planning to remove `countryCodes` altogether, instead relying on the country codes from `useTargetAudienceFinalCountryCodes`. Here is what I suggest:

1. Remove the Audience field everywhere in the current PR (as opposed to removing them in #2535)
2. Have the `initialValues` for `CampaignAssetsForm` components set with the recommended budget or data from local storage.

The `useBudgetRecommendationData` won't need to accept any parameters since it can get the list of targeted countries via the `useTargetAudienceFinalCountryCodes`

What do you think @joemcgill ?

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -1,8 +1,9 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
joemcgill: Let's avoid disabling this rule for the whole file.

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -55,6 +58,21 @@ export default function AdsCampaign( {
...
+		recommendations,
+		loading,
+	} = useBudgetRecommendationData( formContext.values.countryCodes );
+
+	useEffect( () => {
+		if ( ! loading ) {
+			const { amount } = clientSession.getCampaign() || {};
+			formContext.setValue( 'amount', amount || dailyBudget );
+		}
+	}, [ loading ] );
```
joemcgill: This is why the `eslint-disable` rule was added above. Looks like this is meant to only run once to set the initial value after we've finished loading the `dailyBudget` value. However, this causes two problems. 1) I think only running this when `loading` changes runs the risk of a race condition where this runs before the `dailyBudget` value is updated. 2) If you add those dependencies back, each time you update the formContext value, it will cause the `formContext` value to change and cause a loop. To avoid this, we can update this to also only set the form value if it hasn't already been set. Here's an example that seems to be working for me locally. It assumes that the starting amount will always be 0 or falsey, so once it's set to a positive value, we can assume we've already set the value and not rerun the setter.

```js
	useEffect( () => {
		const setFormValue = ( value ) => {
			formContext.setValue( 'amount', value );
		};

		const { amount } = formContext.values;

		if ( ! loading && ! amount ) {
			const { savedAmount } = clientSession.getCampaign() || {};
			setFormValue( savedAmount || dailyBudget );
		}
	}, [ dailyBudget, formContext, loading ] );
```

joemcgill (dismissed):
One change, but otherwise it looks good. I see some additional feedback from @asvin's CR which look like they haven't been resolved. Can we close the loop on those?

joemcgill:
Realized I reviewed this prematurely because things have been updated in #2637 to resolve merge conflicts with #2623. Dismissed my review.

joemcgill (approved):
@asvinb and @ankitrox I've tested these changes after reviewing and merging in the updates resolving the conflicts after #2535 was integrated into this branch. This is looking pretty good to me, but wanted you both to have a chance to have another look and resolve any remaining conversations before sending back for another QA pass.

js/src/setup-ads/setup-ads-form.js
asvinb: Outdated with latest changes.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
asvinb: Outdated with latest changes.

js/src/hooks/useBudgetRecommendationData.js
asvinb: Outdated with latest changes.

js/src/hooks/useBudgetRecommendationData.js
asvinb: Outdated with latest changes.

js/src/hooks/useBudgetRecommendationData.js
asvinb: Outdated with latest changes.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
asvinb: Outdated with latest changes.

js/src/components/paid-ads/budget-section/index.js
asvinb: Outdated with latest changes.

js/src/components/paid-ads/ads-campaign.js
asvinb: Outdated comment. The amount is now set in `PaidAdsSetupForm`

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
asvinb: Outdated comment.

tests/e2e/utils/pages/setup-ads/setup-budget.js
asvinb: Looks like it has been updated. 👍

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
asvinb: Moved to `js/src/utils/getHighestBudget.js` 👍

js/src/components/paid-ads/budget-section/budget-recommendation/index.js
asvinb: Outdated comment with latest changes.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js
asvinb: The `useFetchBudgetRecommendation` hook is now being used which avoids this issue.

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
asvinb: @ankitrox Can you check @joemcgill 's comment please? Not sure about this one.

asvinb:
Thanks @joemcgill . @ankitguptaindia Can you please QA? thanks!

ankitguptaindia:
QA Status: Need feedback/remidiation ⚠️ 
---


Hi @asvinb,

**Point-1:**

The functionality works as expected when the user completes the steps continuously without navigating back in the onboarding flow. However, I noticed an issue when the user goes back after entering a lower value in the "Daily Average Cost" field.

If the user enters an amount lower than the minimum required (for example, 4 USD) and then goes back to the previous step, upon returning to the final step of the onboarding flow, the lower amount (e.g., 4 USD) remains in the "Daily Average Cost" field.

when the same steps are followed while creating a campaign from the dashboard, the "Daily Average Cost" field is automatically updated to the minimum required value (e.g., 15 USD).

Could you please check this once? 

**Campaign creation from Onboarding Flow:** 

https://github.com/user-attachments/assets/2a4decaf-82e3-4b1c-a77e-2173bf2e0375


**Campaign creation from Onboarding vs. Dashboard:** 

https://github.com/user-attachments/assets/305ed52d-7afd-427c-9646-b022fd3e79bf

----

@ankitrox 

**Point-2** 

Could you please suggest what happens when the merchant selects 'All Countries' in the Audience field during the second step? and what value will the Daily Average Cost input display in the Daily Avg Cost field?

image


_(Please  let me know if we are handing this case in any other PR)_

asvinb:
@ankitguptaindia Nice catch for [#1](https://github.com/woocommerce/google-listings-and-ads/pull/2552#issuecomment-2391707293). I just pushed an updated. Can you kindly check?
As for the second point, correct me if am wrong @ankitrox , but if you select all countries, the minimum budget will be the highest recommended amount among all the countries.

ankitguptaindia:
Staus of issues reported here in https://github.com/woocommerce/google-listings-and-ads/pull/2552#issuecomment-2391707293

@asvinb 


> @ankitguptaindia Nice catch for https://github.com/woocommerce/google-listings-and-ads/pull/2552#issuecomment-2391707293. I just pushed an updated. Can you kindly check?

When creating a campaign from the dashboard, if the user inputs a value higher than the required amount (e.g., if the required value is 17 and the user enters 200) and navigates back to the first screen, the budget is automatically reduced to 17.

However, when creating a campaign from the onboarding flow and following the same steps, the budget is not reduced to 17, meaning the system doesn’t enforce the minimum value.

Can we apply the same logic to both places?

- When the user inputs a value lower than the required minimum and navigates back and forth between steps, the value should be adjusted to meet the minimum requirement.
- When the user inputs a value higher than the required amount and navigates back and forth between steps, the value should remain as entered and not be reduced to the required minimum.













Campaign Creation from Dashboard screen:
 
https://github.com/user-attachments/assets/34b2eb2a-0d46-4e1c-b434-c803da997b00 

Campaign Creation from the Onboarding screen: 

https://github.com/user-attachments/assets/e0e249bb-a540-4662-ba7a-dc0693a57bd0

asvinb:
@ankitguptaindia The issue you mentioned [here](https://github.com/woocommerce/google-listings-and-ads/pull/2552#issuecomment-2399966908) should be fixed now. One thing to note that is if you input a lower number than the recommended budget, you move forward and then come back, you'll still see the lower number. It's fine since in #2503 , the user will not be able to go to the next step if the amount he enters is lower than the recommended budget.

ankitguptaindia:
Staus of issues reported here in https://github.com/woocommerce/google-listings-and-ads/pull/2552#issuecomment-2399966908

>@ankitguptaindia The issue you mentioned https://github.com/woocommerce/google-listings-and-ads/pull/2552#issuecomment-2399966908 should be fixed now. One thing to note that is if you input a lower number than the recommended budget, you move forward and then come back, you'll still see the lower number. It's fine since in https://github.com/woocommerce/google-listings-and-ads/issues/2503 , the user will not be able to go to the next step if the amount he enters is lower than the recommended budget.

Thanks, @asvinb reported use cases have been addressed and working as expected now.

js/src/hooks/useFetchBudgetRecommendationEffect.js
eason9487, line 1: This file could be removed as it's no longer in use.

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -125,15 +129,21 @@ export default function SetupPaidAds() {
...
+		return <AppSpinner />;
+	}
+
 	return (
 		<CampaignAssetsForm
 			initialCampaign={ paidAds }
 			onChange={ ( _, values ) => {
-				clientSession.setCampaign( { ...values } );
+				if ( values.amount >= highestDailyBudget ) {
+					clientSession.setCampaign( { ...values } );
```
eason9487: Recreating the object is redundant.

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -29,7 +29,7 @@ const nonInteractableProps = {
  *
  * @param {Object} props React props.
  * @param {Object} props.formProps Form props forwarded from `Form` component.
- * @param {Array<CountryCode>|undefined} props.countryCodes Country codes to fetch budget recommendations for.
+ * @param {Array<CountryCode>} props.countryCodes Country codes to fetch budget recommendations for.
```
eason9487: After this change, the optional chaining operator is not required.

https://github.com/woocommerce/google-listings-and-ads/blob/e712a2c83778fc55eb9e808a9de20ac295f1fa3a/js/src/components/paid-ads/budget-section/index.js#L91

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -52,17 +52,6 @@ const BudgetSection = ( {
...
 
-	/**
-	 * In addition to the initial value setting during initialization, when `disabled` changes
-	 * - from false to true, then clear filled amount to `undefined` for showing a blank <input>.
-	 * - from true to false, then reset amount to the initial value passed from the consumer side.
-	 */
-	const initialAmountRef = useRef( amount );
-	useEffect( () => {
-		const nextAmount = disabled ? undefined : initialAmountRef.current;
-		setValueRef.current( 'amount', nextAmount );
```
eason9487, line 63: `setValueRef` should be removed along with this deletion.

https://github.com/woocommerce/google-listings-and-ads/blob/e712a2c83778fc55eb9e808a9de20ac295f1fa3a/js/src/components/paid-ads/budget-section/index.js#L49-L53

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -52,17 +52,6 @@ const BudgetSection = ( {
...
-	}, [ disabled ] );
```
eason9487, line 64: Although the change to no longer use the `disabled` prop of `BudgetSection` is not made by this PR, the related codes within `BudgetSection` could be removed by the way.

js/src/data/selectors.js
```diff
@@ -406,3 +411,16 @@ export const getTour = ( state, tourId ) => {
 export const getGoogleAdsAccountStatus = ( state ) => {
 	return state.ads.accountStatus;
 };
+
+/**
+ * Retrieves ad budget recommendations for provided country codes.
+ * If no recommendations are found, it returns `null`.
+ *
+ * @param {Object} state The state
+ * @param {Array<CountryCode>} [countryCodes] - An array of country code strings used to generate a unique key.
```
eason9487: In terms of parameter description, it should be more important to describe the use of the country codes for querying relevant data than mention it's for generating a unique key in internal implementation.

js/src/data/utils.js
```diff
@@ -190,6 +194,20 @@ export function mapReportFieldsToPerformance(
...
+ * Generates a unique key (slug) from an array of country codes.
+ *
+ * This function sorts the array of country codes alphabetically,
+ * joins them into a single string with underscore (`_`), and converts
+ * the result to lowercase.
+ *
+ * @param {Array<CountryCode>} countryCodes - An array of country code strings.
+ * @return {string} A underscore-separated, lowercase string representing the sorted country codes.
+ */
+export function getCountryCodesKey( countryCodes = [] ) {
```
eason9487: The `countryCodes` in JSDoc is inconsistent with the code on whether it is optional.

js/src/setup-ads/setup-ads-form.js
```diff
@@ -27,9 +29,11 @@ const SetupAdsForm = () => {
 	const [ handleSetupComplete, isSubmitting ] = useAdsSetupCompleteCallback();
 	const adminUrl = useAdminUrl();
 	const { data: countryCodes } = useTargetAudienceFinalCountryCodes();
+	const { highestDailyBudget, hasFinishedResolution } =
+		useFetchBudgetRecommendation( countryCodes );
 
 	const initialValues = {
-		amount: 0,
+		amount: highestDailyBudget,
```
eason9487: The default budget value may be incorrectly set due to the different currencies between WooCommerce store settings and the existing Google Ads account.

Steps to reproduce:

1. Have an existing Google Ads account with *A* currency, e.g. USD.
2. Disconnect Google Ads account if it's connected.
3. Set WooCommerce store settings to use *B* currency, e.g. JPY.
4. Go to the Ads setup flow (post-onboarding)
   1. Connect to the Google Ads account mentioned in step 1.
      
   2. Continue to step 2 and view the inconsistent currencies and incorrect initial budget.

eason9487 (changes requested):

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: This is no longer applicable since we moved this logic to `BudgetSection`

js/src/hooks/useFetchBudgetRecommendationEffect.js
asvinb: File removed.

js/src/components/paid-ads/budget-section/index.js
asvinb: I've updated the JSDocs since `countryCodes` can be `undefined` while the data is being resolved.

js/src/components/paid-ads/budget-section/index.js
asvinb: 👍

js/src/components/paid-ads/budget-section/index.js
asvinb: 👍

js/src/data/selectors.js
asvinb: JSDocs updated.

js/src/data/utils.js
asvinb: Updated.

js/src/setup-ads/setup-ads-form.js
asvinb: Thanks for catching that @eason9487 . We've moved the logic to fetch the budget recommendation to `BudgetSection` now and should solve the issue.

asvinb:
@ankitguptaindia Can you please check if you can still reproduce the issues [mentioned here](https://github.com/woocommerce/google-listings-and-ads/pull/2552#discussion_r1804597699) after the latest updates please?
Thanks!

asvinb:
@eason9487 Can you review proposed changes please? We are now calling the `useFetchBudgetRecommendation` hook in the `BudgetSection` component (https://github.com/woocommerce/google-listings-and-ads/pull/2552/files#diff-930c930d2c6b54b942dd7cecad526e73a68845aece1d8cb09885f1aa8c43a1e9R55) which is displayed once a Google Ads account is connected. Furthermore, we've invalidated the state for for `useFetchBudgetRecommendation` whenever a Google Ads account is disconnected/connected.

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -32,36 +35,56 @@ const nonInteractableProps = {
...
+
+		if ( values.amount >= highestDailyBudget ) {
+			clientSession.setCampaign( values );
+		}
+	}, [ values, highestDailyBudget, context, hasFinishedResolution ] );
+
+	if ( ! initialAmountRef.current && ! amount && hasFinishedResolution ) {
+		let clientSessionAmount = 0;
+		if ( context === 'setup-mc' ) {
+			( { amount: clientSessionAmount } = clientSession.getCampaign() );
```
eason9487: An error occurs when entering step 4.

https://github.com/user-attachments/assets/4a850c78-85f2-49e6-b940-c30650b4fbea

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -32,36 +35,56 @@ const nonInteractableProps = {
...
+		}
+
+		initialAmountRef.current = true;
+		setValue(
+			'amount',
+			Math.max( clientSessionAmount, highestDailyBudget )
+		);
```
eason9487: This processing causes a warning.

image

js/src/components/paid-ads/budget-section/index.js
```diff
@@ -32,36 +35,56 @@ const nonInteractableProps = {
...
-	const { getInputProps, setValue, values } = formProps;
+	const { getInputProps, values, setValue } = formProps;
 	const { amount } = values;
 	const { googleAdsAccount } = useGoogleAdsAccount();
+	const {
+		data: budgetRecommendationData,
+		highestDailyBudget,
+		highestDailyBudgetCountryCode,
+		hasFinishedResolution,
+	} = useFetchBudgetRecommendation( countryCodes );
```
eason9487: One of the solutions I suggested for resolving the issue of https://github.com/woocommerce/google-listings-and-ads/pull/2552#discussion_r1804597699 is to adjust the timing of issuing the recommended budget API request, and invalidate the corresponding states in the wp-data store after disconnecting Google Ads account.

The current implementation is in this direction but it's not ideal. This implementation might only be considered when there is no relatively suitable and easier one.

Suggestions:

1. Revert these commits: https://github.com/woocommerce/google-listings-and-ads/compare/48aeec3..f4631e9
2. Adjust the structure of components under `js/src/setup-ads` to a similar one as [js/src/setup-mc/setup-stepper/setup-paid-ads.js](https://github.com/woocommerce/google-listings-and-ads/blob/48aeec39d95edc688bece8ad072ba7d37a4aa5fc/js/src/setup-mc/setup-stepper/setup-paid-ads.js)
   - Move `CampaignAssetsForm` and all related codes from `SetupAdsForm` to `SetupPaidAds`
   - Move `SetupAdsTopBar` and `AdsStepper` from `SetupAdsForm` to `SetupAds`
   - After the above changes, `SetupAdsForm` component should be empty and can be removed
3. It should only need to invalidate the recommended budget states after disconnecting Google Ads account
   https://github.com/woocommerce/google-listings-and-ads/blob/f4631e95b90458f0e07b7cc278c90624797a75b7/js/src/data/resolvers.js#L565-L570

eason9487 (changes requested):
The [e2e testing](https://github.com/woocommerce/google-listings-and-ads/actions/runs/11438516731/job/31820265811) couldn't pass.

js/src/components/paid-ads/budget-section/index.js
asvinb: Thanks @eason9487 . I think your suggestions make sense. Updated.

js/src/components/paid-ads/budget-section/index.js
asvinb: Should not longer be the case now.

js/src/components/paid-ads/budget-section/index.js
asvinb: No longer applicable.

eason9487 (approved):

PR #2558, opened 2024-08-22
Streamline campaign setup: Change to use banner to present the ad credit offer during onboarding

### LINKED ISSUE #2499: Campaign Creation: Add coupon banner to the top of the page

Image

In order to bring more prominence to the $500 ads credit that is available for new accounts, we'll make the offer more prominent on the campaign creation step of the onboarding flow.

## Acceptance Criteria

- [x] When an Ads account is eligible for a free Ads credit, the `FreeAdCredit` component will be displayed at the bottom of the `PaidAdsFeatureSection` Card in the Campaign Creation step of the onboarding flow.
- [x] When an Ads account is not eligible, the component will not be shown.
- [x] The existing text in that card that shows the $500 credit is removed.

## Implementation Brief
A component for this banner is already part of the `SetupAccounts` step for the ads setup stepper (js/src/setup-ads/ads-stepper/setup-accounts/index.js). This component can be moved to the top-level `js/src/components` directory and used in both the current `SetupAccounts` component and imported into the `SetupPaidAds` component in `js/src/setup-mc/setup-stepper/setup-paid-ads/setup-paid-ads.js`. This should use the same `useFreeAdCredit()` hook to determine whether it is displayed.

## Test Coverage

1. E2E tests in `tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js` should be updated to add a test that confirms the banner is shown
2. Ensure E2E tests for the ads stepper are updated if needed.

asvinb:

Closes #2499 

### Screenshots:
image

1. Go through the onboarding flow.
2. When connecting accounts on the first step, create a new Google Ads account from the UI. This is the only way for the user to be "eligible" to see the banner.
3. On the last step of the onboarding process, the banner should be visible as per the screenshot in the ticket [description](https://github.com/woocommerce/google-listings-and-ads/issues/2499#issue-2451602170) .
4. The previous text for the $500 credit that was below "Set a daily budget" should no longer appear.

### Changelog entry

> Tweak - Change to use a banner to present the ad credit offer during onboarding.


tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
```diff
@@ -140,6 +140,27 @@ test.describe( 'Complete your campaign', () => {
...
+			await completeCampaign.goto();
+
+			await expect(
+				page.getByText(
+					'Spend $500 to get $500 in Google Ads credits!'
+				)
+			).toBeVisible();
+		} );
+	} );
+
```
ankitrox: @asvinb While this seems to be passing, it looks like those are not working as expected. For the first test `shoud not see the Free Ad Credit section if the account is not eligible`, page is not loading at all and there is always a spinner on the page, thus it is passing the assertion that text is not existing.

This can be better validated with the `--ui` option in e2e tests.

```
npm run test:e2e -- step-4-complete-campaign.test.js --grep "Free Ad Credit" --ui
```

tests/e2e/utils/mock-requests.js
```diff
@@ -541,6 +541,23 @@ export default class MockRequests {
...
+	async mockAdsSubAccountConnected( id = 123456 ) {
+		await this.fulfillAdsConnection( {
+			id,
+			sub_account: true,
+			created_timestamp: Date.now(),
+			currency: 'TWD',
+			symbol: 'NT$',
+			status: 'connected',
+		} );
+	}
```
ankitrox: I think this won't be required if we accept the `args` in `mockAdsAccountConnected` and merge them with default args inside it.

@joemcgill Thoughts?

ankitrox (dismissed):

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
asvinb: @ankitrox I made a small tweak to the PR and added 
`await completeCampaign.goto();`
I checked again and there's indeed a spinner but as you hover on the "timeline", you can see the loaded page:
image

Additionally, to make sure, I added:
```js
await expect(
	page.getByRole( 'heading', {
		name: 'Create a campaign to advertise your products',
	} )
).toBeVisible();
```

just before the "Spend $500 to get $500 in Google Ads credits!" test and the test passes. Can you kindly check and let me know please?

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js
```diff
@@ -103,32 +108,36 @@ export default function PaidAdsFeaturesSection( {
...
+								<FeatureList
+									hideBudgetContent={ hideBudgetContent }
+								/>
+							</FlexBlock>
+							<FlexItem>
+								<CampaignPreview />
+							</FlexItem>
+						</Flex>
+
+						{ hasFreeAdCredit && <FreeAdCredit /> }
```
joemcgill, line 115: Rather than using the `hasFreeAdCredit` from the `useFreeAdCredit()` hook, for now we should use the `hideBudgetContent` property, which is what determines whether the $500 credit content is shown in the `FeatureList` component defined above and that line should be removed because it's now redundant.

https://github.com/woocommerce/google-listings-and-ads/blob/02537c218b8c0be21c1d5d62f5bb39f3dce71474/js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js#L32-L40

tests/e2e/utils/mock-requests.js
joemcgill: I agree that we don't need a separate utility to mock a sub-account, but the bigger issue with this mock is that the `/wc/gla/ads/connection` endpoint rarely if ever returns the `sub_account` or `created_timestamp` values, from what I can tell.

https://github.com/woocommerce/google-listings-and-ads/blob/a3131d30776a4479692025f95abc9f6c66e70181/src/Ads/AccountService.php#L76-L94

Those values are only set during the setup process, otherwise they don't exist.

https://github.com/woocommerce/google-listings-and-ads/blob/02537c218b8c0be21c1d5d62f5bb39f3dce71474/src/Ads/AccountService.php#L152-L161

I can see that these values were originally an requirement for the logic in the `useFreeAdCredit()` hook, but something must have changed at some point. Rather than using that hook, lets use the same logic that is already in use to display the $500 ad credit in this card (explained in my other comment).

joemcgill (changes requested):
Thanks all. I've left some feedback inline, but the main issue is that this isn't displaying on the fourth step when testing locally. 

I'll get some clarity about whether the `useFreeAdCredit()` hook needs to be adjusted, but if so, we'll handle that as a separate task. For now, the new banner should display any time the current "Claim $500 in ads credit..." text was previously shown and that text should be removed.

tests/e2e/utils/mock-requests.js
asvinb: 👍

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js
asvinb: Sounds good @joemcgill

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js
```diff
@@ -103,32 +94,34 @@ export default function PaidAdsFeaturesSection( {
...
+									) }
+								</div>
+								<FeatureList />
+							</FlexBlock>
+							<FlexItem>
+								<CampaignPreview />
+							</FlexItem>
+						</Flex>
+
+						{ ! hideBudgetContent && <FreeAdCredit /> }
```
joemcgill: This correctly retains the current behavior for the $500 credit, which meets the expectations for this PR. However, the current behavior as implemented actually doesn't actually make sense, because the `hideBudgetContent` prop is only `true` when this component is rendered by the `SetupPaidAds` component when there is no Google Ads connection ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/db9d822f614f8860110d07d8d14c8014bd829e1a/js/src/setup-mc/setup-stepper/setup-paid-ads/setup-paid-ads.js#L173)).

Now that Google Ads has moved to the first step of the onboarding flow, this will always be true. I'll create a new ticket to address this logic and we can move this forward in the meantime.

joemcgill (approved):
This is working as expected. I left a comment inline about needing to address the requirements for when this should be displayed in a follow-up task.

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.1
* Theme active on store: Storefront Version: 4.6.0 / Twenty Twenty-Four Version: 1.2
* WooCommerce - Version 9.2.3 
* PHP: 8.3
* Web Server: Nginx
* Browser: Chrome - Version 127
* OS: macOS Sonoma 14.6.1 

</Details>


**Test Results -** Ad coupon banner appears as described in the description.  ✅ 

image

ankitguptaindia (approved):

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js
eason9487: This condition should be controlled by the `useFreeAdCredit` hook. Otherwise, it won't meet the acceptance criteria in #2499.

The reason for the previous use of the `hideBudgetContent` was to comply with the requirement at that time: "Before the Google account is connected we can show the Boost card but we’ll want to limit the buttons and mentioning of Ads audience and budget." but not consider whether the account is eligible.

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
```diff
@@ -443,4 +443,28 @@ test.describe( 'Complete your campaign', () => {
...
+		test( 'should not see the Free Ad Credit section if the account is not eligible', async () => {
+			await setupAdsAccountPage.mockAdsAccountDisconnected();
+			await completeCampaign.goto();
+
+			await expect(
+				page.getByText(
+					'Spend $500 to get $500 in Google Ads credits!'
+				)
+			).not.toBeVisible();
+		} );
```
eason9487: This test case doesn't fulfill its test purpose:

- The Google Ads account should have been connected in this step. Mocking it with disconnected status is not ideal.
- It incorrectly passes the assertion while the page is still in loading state
   image

eason9487 (changes requested):
Thanks for the work. There are two issues need to be resolved.

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
asvinb: 👍

asvinb:
@joemcgill I made a few updates following @eason9487 's comments. Can you kindly check please?

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
```diff
@@ -443,4 +443,39 @@ test.describe( 'Complete your campaign', () => {
...
+			await setupAdsAccountPage.mockAdsAccountConnected();
+		} );
+
+		test( 'should not see the Free Ad Credit section if the account is not eligible', async () => {
+			await completeCampaign.goto();
+
+			// Check we are on the correct page.
+			await expect(
+				page.getByText( 'Create a campaign to advertise your products' )
+			).toBeVisible();
```
eason9487, line 369: I'm a little earlier viewing the changes.

Although the above mocking account status is correct, this test case still happens to pass. This check doesn't guarantee that the Google Ads account status has been fetched.



For example, in the screenshot, the audiences field is still disabled. From its implementation, it can be observed that the account status has not been fetched yet.

https://github.com/woocommerce/google-listings-and-ads/blob/90fd7e771ed1158b1b4a2d54981a27cf2b6b1710/js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-setup-sections.js#L158-L166

https://github.com/woocommerce/google-listings-and-ads/blob/90fd7e771ed1158b1b4a2d54981a27cf2b6b1710/js/src/hooks/useGoogleAdsAccount.js#L45-L48

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js
joemcgill: The suggestion to use the `useFreeAdCredit()` was a part of the initial implementation brief, but is not part of the acceptance criteria. This was initially implemented using that hook, but we found that doing so didn't result in the expected visual behavior ([see related conversation](https://github.com/woocommerce/google-listings-and-ads/pull/2558#pullrequestreview-2261690968)).

I do think we need to clarify the correct behavior for determining when `useFreeAdCredit()` should return true, but would like to handle that in a separate issue. I'm fine with changing this PR back to using the hook in the meantime, which seems to be done in [90fd7e7](https://github.com/woocommerce/google-listings-and-ads/pull/2558/commits/90fd7e771ed1158b1b4a2d54981a27cf2b6b1710).

joemcgill:
@eason9487 I've opened #2599 to clarify and address the requirements for the `useFreeAdCredit()` hook.

js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js
eason9487: > The suggestion to use the `useFreeAdCredit()` was a part of the initial implementation brief, but is not part of the acceptance criteria.

It's probably not about whether or not `useFreeAdCredit` was used. I meant that the actual results did not meet this acceptance criteria in #2499:

> * When an Ads account is not eligible, the component will not be shown.

With the 24c93287e63381abdf77640e6e071418407d14c5 revision, the banner is shown after connecting to an account that has been created more than 60 days ago.

asvinb: 👍

tests/e2e/specs/setup-mc/step-4-complete-campaign.test.js
asvinb: @eason9487 I added the following
https://github.com/woocommerce/google-listings-and-ads/pull/2558/commits/9a612d3d0ba0e4506880f16da88f1bb8c3f26904#diff-3b03eaec7b7b793a56912cfe86e6d384ea90da173ddd1f9855ba1fa5cd80270bR384

and this is how it looks on my end while testing:
image

Let me know what you think.

eason9487 (approved):
Thanks for the further update! LGTM.

PR #2560, opened 2024-08-23
Return settings prop as object in the REST API

puntope:

Ref: p1724235667349019/1722414955.111429-slack-C02BB3F30TG

Reverts https://github.com/woocommerce/google-listings-and-ads/pull/2519

This PR fixes an error introduced in https://github.com/woocommerce/google-listings-and-ads/pull/2519 

At first, I understood that it was expected to get all empty props as an empty object. However, what was expected was only to get the Settings prop as an object.  Rest of the props should be as before.

This PR reverts https://github.com/woocommerce/google-listings-and-ads/pull/2519  and makes only settings to be returned as an object.

### Screenshots:

Screenshot 2024-08-23 at 11 34 23
Screenshot 2024-08-23 at 11 34 46
Screenshot 2024-08-23 at 11 35 45
Screenshot 2024-08-23 at 11 35 31

1.  Checkout this PR and 
2. Request `wp-json/wc/v3/products?gla_syncable=1` 
3. `attributes` field is shown as an array (when is empty and when is filled)
4. `gla_attributes` field is shown as an object (when is empty and when is filled)
5. ~~Request `wp-json/wc/v3/settings/general?gla_syncable=1`~~ 
6. ~~`settings` field is shown as an object (when is empty and when is filled)~~
5. Add the following filter which will return empty settings for the flat shipping rate:

```
add_filter(
	'woocommerce_shipping_instance_form_fields_flat_rate',
	function ( $defaults ) {
		return [];
	},
	10,
	1
);
```
6. Make a request to `v3/shipping/zones/YOUR_SHIPPING_ZONE/methods?gla_syncable=1` which includes a Flat Rate method.
7. See the response:

### Before:

image

### After

image

### Changelog entry

>


jorgemd24:
Thanks, @puntope, for fixing this! However, ~~I noticed there was some confusion—the issue wasn’t actually with the `wp-json/wc/v3/settings/general` endpoint but with the settings in `shipping/zones/X/methods`. See this comment: p1724749602301079/1722414955.111429-slack-C02BB3F30TG.~~ I just realized the testing steps hadn’t been updated and still mentioned the settings endpoint, so I’ve updated them.  I’ve updated the PR to fix the issue specifically for the shipping zone method endpoint, rather than for all settings in every response. We’re not sure if other endpoints with the setting key will return an object.

### Before:

image

### After:

image

I’ll go ahead and merge this PR for you.

PR #2568, opened 2024-08-27
Limits steps in onboarding based on previously completed state

### LINKED ISSUE #2493: Onboarding: Remove “Confirm store requirements” step if possible
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2458

Image

Once the Pre-Launch checklist is removed from the `StoreRequirements` page, the only thing this page is for is verifying store contact info (i.e., phone and address). In cases where we already have a verified phone number and address, we don't need to include this step in the onboarding.

If we can already determine that we have a verified phone number and an address associated with the store, don't show this step during onboarding and move directly to Campaign creation after Configuring Product listings (or whatever step comes before confirming store requirements). These values can be modified in the settings screen after onboarding if needed.

## Acceptance Criteria

- [ ] When the onboarding app loads, the Confirm Store Requirements step is not included if the phone number is verified and the MC address is the same as the store address.
- [ ] If the phone number isn't verified, or if the address isn't the same in MC and WC, include the step and show the `ContactInformation` card with both phone and address cards visible.

## Implementation Brief

To check the phone and address info before setting up the setup stepper, we'll need to use the `useStoreAddress()` hook in `js/src/setup-mc/setup-stepper/saved-setup-stepper.js`. We know we have a valid phone number when both there is a `phone_number` value and  `phone_verification_status` is `"verified"`. We know the address is valid when we have a `mc_address ` value and the `is_mc_address_different` value is `false`. 

When all those conditions are met, we can update the `SavedSetupStepper` to remove the `store_requirements` step from the stepper ([reference](https://github.com/woocommerce/google-listings-and-ads/blob/trunk/js/src/setup-mc/setup-stepper/saved-setup-stepper.js#L212-L224)). We'll also need to update the `handleSetupListingsContinue()` callback so that it calls `continueStep( stepNameKeyMap.paid_ads )` instead of `continueStep( stepNameKeyMap.store_requirements )` if the store requirements are not being included.

## Test Coverage

Update the E2E tests for step 2 of the onboarding (tests/e2e/specs/setup-mc/step-2-product-listings.test.js) to include two scenarios for the continue button:

1. The current scenario that confirms that clicking continue results in showing the Confirm store requirements page if we don't have a verified phone and address.
2. A new scenario that confirms that clicking continue shows the Campaign Setup step if we already have a verified phone and address.

dsawardekar:

Closes #2493 

This PR hides the confirm registration step if the user has sufficient data filled in already. If the user was on the confirm registration step previously, they are sent to the next step.

### Screenshots:

With incomplete store requirements: (Confirm store requirements is present)
image

With complete store requirements: (Confirm store requirements is absent)
image

1. Start the onboarding process with incomplete store data.
2. You should see the Step no. 3 Confirm Store Requirements
3. Reset your env, and delete some store data (eg:- Delete City under WooCommerce Store Address)
4. Restart the onboarding process
5. You should not see the Step no. 3

### Additional details:

Update: Hides the Confirm Store Requirements step when feasible

### Changelog entry

Update: Hides the Confirm Store Requirements step when feasible

### DIFF
--- a/js/src/setup-mc/setup-stepper/index.js
+++ b/js/src/setup-mc/setup-stepper/index.js
@@ -9,14 +9,28 @@ import { getHistory, getNewPath } from '@woocommerce/navigation';
 import AppSpinner from '.~/components/app-spinner';
 import SavedSetupStepper from './saved-setup-stepper';
 import useMCSetup from '.~/hooks/useMCSetup';
+import useStoreAddress from '.~/hooks/useStoreAddress';
+import useGoogleMCPhoneNumber from '.~/hooks/useGoogleMCPhoneNumber';
 import stepNameKeyMap from './stepNameKeyMap';
 
 const SetupStepper = () => {
 	const { hasFinishedResolution, data: mcSetup } = useMCSetup();
+	const { data: address, loaded: addressLoaded } = useStoreAddress();
+	const { data: phone, loaded: phoneLoaded } = useGoogleMCPhoneNumber();
 
-	if ( ! hasFinishedResolution && ! mcSetup ) {
-		return <AppSpinner />;
-	}
+	const hasValidPhoneNumber =
+		phoneLoaded && phone?.isValid && phone?.isVerified;
+
+	const hasValidAddress =
+		addressLoaded &&
+		address?.isAddressFilled &&
+		! address?.isMCAddressDifferent;
+
+	const hasConfirmedStoreRequirements =
+		hasValidPhoneNumber && hasValidAddress;
+
+	const hasLoaded =
+		hasFinishedResolution && mcSetup && addressLoaded && phoneLoaded;
 
 	if ( hasFinishedResolution && ! mcSetup ) {
 		// this means error occurred, we just need to return null here,
@@ -24,14 +38,31 @@ const SetupStepper = () => {
 		return null;
 	}
 
-	const { status, step } = mcSetup;
+	if ( ! hasLoaded ) {
+		return <AppSpinner />;
+	}
+
+	const { status } = mcSetup;
+	let { step } = mcSetup;
+
+	// If the user has already completed the store requirements, but is currently still on the
+	// store requirements step, we should skip the store requirements step and go to the paid ads step.
+	// else they will get stuck on a non-existent step #3
+	if ( step === 'store_requirements' && hasConfirmedStoreRequirements ) {
+		step = 'paid_ads';
+	}
 
 	if ( status === 'complete' ) {
 		getHistory().replace( getNewPath( {}, '/google/dashboard' ) );
 		return null;
 	}
 
-	return <SavedSetupStepper savedStep={ stepNameKeyMap[ step ] } />;
+	return (
+		<SavedSetupStepper
+			savedStep={ stepNameKeyMap[ step ] }
+			hasConfirmedStoreRequirements={ hasConfirmedStoreRequirements }
+		/>
+	);
 };
 
 export default SetupStepper;
--- a/js/src/setup-mc/setup-stepper/index.js
+++ b/js/src/setup-mc/setup-stepper/index.js
@@ -9,28 +9,14 @@ import { getHistory, getNewPath } from '@woocommerce/navigation';
 import AppSpinner from '.~/components/app-spinner';
 import SavedSetupStepper from './saved-setup-stepper';
 import useMCSetup from '.~/hooks/useMCSetup';
-import useStoreAddress from '.~/hooks/useStoreAddress';
-import useGoogleMCPhoneNumber from '.~/hooks/useGoogleMCPhoneNumber';
 import stepNameKeyMap from './stepNameKeyMap';
 
 const SetupStepper = () => {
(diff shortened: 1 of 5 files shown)


dsawardekar:
@joemcgill This is working but I needed to move the implementation to the parent container component. Having the logic inside the SavedSetupStepper was causing a broken UX like below when the user already had completed store requirements.

Screenshot from 2024-08-21 16-31-16

Can you review and let me know if this looks good to proceed on the e2e tests? 

Thanks!

js/src/setup-mc/setup-stepper/index.js
```diff
@@ -9,29 +9,60 @@ import { getHistory, getNewPath } from '@woocommerce/navigation';
...
+
+	const { status } = mcSetup;
+	let { step } = mcSetup;
+
+	// If the user has already completed the store requirements, but is currently still on the
+	// store requirements step, we should skip the store requirements step and go to the paid ads step.
+	// else they will get stuck on a non-existent step #3
+	if ( step === 'store_requirements' && hasConfirmedStoreRequirements ) {
+		step = 'paid_ads';
+	}
```
joemcgill: Smart. Rather than mutating the `step` variable here, it may be better to do something like this:

```js
const { status, step } = mcSetup;

const currentStep = ( step === 'store_requirements' && hasConfirmedStoreRequirements ) ?  'paid_ads' : step;
```

joemcgill (commented):
@dsawardekar I think this approach makes sense. Particularly since it's the `SetupStepper` that manages the loading state prior to rendering the `SavedSetupStepper`.

Once you've updated the E2E tests, feel free to send this back for review.

dsawardekar:
@joemcgill I've added the e2e tests for this, but I need assistance with mocking the email and mc account address. When I try to set mockMCSetup to use 'complete', it redirects to the Home page, and skips the onboarding flow completely.

image


Test File:
https://github.com/woocommerce/google-listings-and-ads/blob/31a3ba5017434a3d4f6995f89839dd64546eb888/tests/e2e/specs/setup-mc/step-3-hide-store-requirements.test.js

Can you please take a look? Thanks!

js/src/setup-mc/setup-stepper/index.js
```diff
@@ -43,14 +43,15 @@ const SetupStepper = () => {
 	}
 
 	const { status } = mcSetup;
-	let { step } = mcSetup;
+	const { step } = mcSetup;
```
joemcgill: These can be combined:

```js
const { status, step } = mcSetup;
```

tests/e2e/specs/setup-mc/step-3-hide-store-requirements.test.js
joemcgill, line 1: Rather than creating a whole new spec here, I think we should test for these scenarios in either the existing `tests/e2e/specs/setup-mc/step-3-confirm-store-requirements.test.js` spec, or test for this scenario in `tests/e2e/specs/setup-mc/step-2-product-listings.test.js` whenever the continue button is clicked with the address info already correctly mocked. I think that should look something like this:

```js
productListingsPage.mockContactInformation( {
	phone_number: '+15555555',
	phone_verification_status: 'verified',
	mc_address: {
		street_address: '556 Woo St.',
		locality: 'City',
		region: 'California',
		postal_code: '90210',
		country: 'US',
	},
	wc_address: {
		street_address: '556 Woo St.',
		locality: 'City',
		region: 'California',
		postal_code: '90210',
		country: 'US',
	},
	is_mc_address_different: false,
	wc_address_errors: [],
} );
```

One tricky part here is that the step is only removed when the stepper is first loaded, so we'll need a beforeAll step that reloads the page after the state is correctly mocked.

I'm going to take a look tomorrow to see if I can get the state for this set up properly, since it's a bit tricky.

joemcgill (commented):
Thanks @dsawardekar. I've added a bit of initial feedback. I'm going to see if I can get these e2e tests working, but in the mean time can you address the other feedback?

joemcgill (commented):
@dsawardekar while testing this again, I've realized that part of the problem that has made the E2E difficult to debug is that these updates won't work when an MC account is not connected when the onboarding flow is loaded, which is a pretty significant problem.

If you test this with all accounts disconnected, you'll see an error message pop up when the following hooks are called:

https://github.com/woocommerce/google-listings-and-ads/blob/abc86c3a2ce2a6813c6da38d7707b6b70ff1934f/js/src/setup-mc/setup-stepper/index.js#L18-L19

Sorry I didn't catch this when you initially suggested moving these hooks here. I assume I was testing by just refreshing the page after already completing step 1 of the onboarding. I'm curious if we're able to initially show step 3 and then hide it once someone connects their MC account in step 1. If not, then we'll likely need to rethink this issue.

dsawardekar:
@joemcgill That might work, but the UX would be less than ideal. Maybe we need to consider folding the things that in that extra dissapearing step into sections inside Step 2 / 3. Then if the user has sufficient data, we could hide that section.

eason9487:
Closing this as an alternative has been implemented and merged. See https://github.com/woocommerce/google-listings-and-ads/issues/2493#issuecomment-2391705651.

Issue #2534, opened 2024-08-16
Remove the accounts connection page from the Ads Setup flow


joemcgill:
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2460

When a merchant is setting up a new ads campaign from the plugin dashboard, they enter the "Set up paid campaign" flow, which includes individual steps for connecting a Google Ads account, configuring a campaign, and completing their billing setup.

<img width="1146" alt="image" src="https://github.com/user-attachments/assets/2b933370-56ba-4860-b8cb-26e278584d2c">

Now that connecting an Ads account is a requirement of the initial onboarding process, we will remove the first step of this flow entirely, as it's no longer necessary. We'll still need to allow for an Ads account to be connected in this flow if one was disconnected from the settings screen ([see this comment](https://github.com/woocommerce/google-listings-and-ads/issues/2534#issuecomment-2306716676)), but that flow will be consolidated into step 2 as a follow up to #2535.

## Acceptance Criteria

- [ ] The "Setup your accounts" step is removed from the "Set up paid campaign" flow.
- [ ] The flow can still be completed to successfully to add a campaign if Ads is connected.

Out of scope: addressing the use case where an Ads account was disconnected after onboarding.

## Implementation Brief
The PR should get merged into the [feature/2460-simplify-paid-ads-setup](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2460-simplify-paid-ads-setup) branch.

The component that defines the Set up paid campaign flow is the `AdsStepper` located at `/js/src/setup-ads/ads-stepper/index.js`. [The first object](https://github.com/woocommerce/google-listings-and-ads/blob/2.8.3/js/src/setup-ads/ads-stepper/index.js#L72-L85) in the array passed to the `steps` prop can be removed and any unused code cleaned up. The keys associated with each remaining step objects and the `handleCreateCampaignContinue()` function will need to be updated.

The `SetupAccounts` component defined in `js/src/setup-ads/ads-stepper/setup-accounts/index.js` can be deleted, along with any dependencies that are no longer in use after it has been removed.

## Test Coverage
- The jest tests in `js/src/setup-ads/ads-stepper/index.test.js` should not be updated in this PR, and can be updated as a separate follow up task when #2536 is completed.
- The E2E tests in `tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js` cover this flow, but should be refactored since they focus on setting up a Google Ads account.
[...]

Issue #2460, opened 2024-07-08
Post Onboarding Overview


eclarke1:
## Objective

In order to help merchants enhance product visibility and attract potential customers, we will improve the user experience for working with Ads after finishing the onboarding process.

## Background

Currently, during the initial onboarding flow, merchants can complete the onboarding process without setting up billing and adding any Ads campaigns. In those cases, merchants must go through another multi-step user flow to set up Ads for the first time. Similar to the Onboarding Improvements project, we can streamline this flow to make creating a campaign easier and avoid unnecessary steps:

**Step 1: Shows connected accounts**. Now that setting up an Ads account is a requirement of onboarding, this is unnecessary and can be removed.

![Image](https://github.com/user-attachments/assets/ddb11926-dedb-40a9-8e9c-999c164be1af)

**Step 2: Create a campaign.** This page is similar to the final page of the onboarding flow, but by using the exact same UI we can provide the same streamlined experience regardless of whether the user is setting up a campaign during onboarding or after.

**Step 3: Set up billing.** This is always shown, even if the user has already set up billing on their ads account. This is unnecessary and can be removed if not needed.

![Image](https://github.com/user-attachments/assets/9d98dd06-5ac5-473d-bd58-7bb6e202c98e)

In addition to streamlining campaign creation after onboarding, there is a general lack of information about the value of setting up ads or about the current status of the ads account or previously added campaigns on the dashboard.

![Image](https://github.com/user-attachments/assets/bad7744a-feec-407e-9cac-af792c9bc59d)

## Overview

To improve the user experience of working with Ads campaigns after initial onboarding we can make the following improvements.

### Streamline the Setup Ads flow

> [!Warning]
> This section has been implemented as part of #2459 instead.

### **Communicate the value of Google Ads after onboarding**

There are three tabs that are part of the Google for WooCommerce settings page where we will can highlight the value proposition for setting up ads:

* the dashboard tab  
* the reports tab  
* the product feed tab (the first tab a new user sees after onboarding)

On the dashboard tab, we will swap the position of the performance cards for free listings and paid campaigns and rename these to “Google Ads” and “Free Listings (Limited Visibility)”. If no paid ads campaigns have been set up, we will show a similar value proposition as is shown during onboarding, including promoting the $500 Google Ads credit.

![Image](https://github.com/user-attachments/assets/4b4410f8-0787-451b-894f-53a09101f068)
*Current Paid campaigns card*

[...]

PR #2577, opened 2024-09-04
Streamline campaign setup: Temporarily remove the billing setup step from the Ads-onboarding

### LINKED ISSUE #2536: Remove the billing setup page from the Ads setup flow

image

In #2535, we are replacing the second step with the same component used during onboarding, which includes the ability to set up billing if needed. That makes the "Setup up billing" step redundant and can be removed.

## Acceptance Criteria

- [ ] The "Setup up billing" step is no longer included in the `AdsStepper`.
- [ ] When billing is already set up, completing the "Create your paid campaign" step should result in the Setup Paid Ads flow being completed successfully with a new campaign. 

## Implementation Brief

The `AdsStepper` component for the "Set up paid campaign" flow is in `js/src/setup-ads/ads-stepper/index.js`. The billing step can be removed from the array [passed to the `steps` prop here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/setup-ads/ads-stepper/index.js#L100-L105).

The `handleSubmit` prop that is passed via formProps to the `SetupBilling` component will need to be moved to the component responsible for rendering the ads campaign setup (e.g. `AdsCampaign`) so that step can complete the form when finished.

For testing, billing status can be mocked using the following filter in an MU plugin:

```php
function gla_filter_ads_billing_setup_status( $status ) {
	return 'approved';
}
add_filter( 'woocommerce_gla_ads_billing_setup_status', 'gla_filter_ads_billing_setup_status' );
```

## Test Coverage
- Jest tests will be updated as a follow-up task at the completion of #2534, #2535, and this issue.
- E2E tests related to the billing step can be removed [from here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js#L392-L536).
  - Tests related to completion of the campaign should be retained as a standalone test. ([current example](https://github.com/woocommerce/google-listings-and-ads/blob/develop/tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js#L473-L498))

kt-12:

Closes #2536 .

Billing setup page should be removed and also `Launch Paid Campaign` should be possible from second step.

1. delete any residual ``DELETE FROM `wp_options` WHERE `option_name` = 'gla_ads_setup_completed_at';``
2. Click on `Add paid campaign`
Screenshot 2024-09-04 at 17 55 11
3.  Click on continue on `Set up your accounts` page
Screenshot 2024-09-04 at 17 55 23
4.  You should be able to see disabled `Launch Paid Campaign` button which enables on adding `Daily Average Cost`
Screenshot 2024-09-04 at 17-56-39 Setup Google Ads ‹ Google for WooCommerce ‹ Marketing ‹ WooCommerce ‹ google-woo — WooCommerce


kt-12:
@joemcgill Based on your comment on issue, I have not modified Jest tests here.

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -25,19 +25,21 @@ import FaqsSection from './faqs-section';
...
- * @param {() => void} props.onContinue Callback called once continue button is clicked.
+ * @param {Object} props.formProps Form props forwarded from `Form` component.
+ * @param {Function} props.formProps.isSubmitting Callback to display loading till Ads campaign is created.
+ * @param {Function} props.formProps.handleSubmit Callback to handle campaign creation.
  * @param {'create-ads'|'edit-ads'|'setup-ads'} props.trackingContext A context indicating which page this component is used on. This will be the value of `context` in the track event properties.
  */
 export default function AdsCampaign( {
 	campaign,
-	onContinue,
+	formProps: { isSubmitting, handleSubmit },
```
asvinb: @kt-12 Instead of passing an object, how about we keep the `onContinue` prop and add a new `isLoading` prop which will show the loading state?
It feels that `AdsCampaign` is like "tied" to the form.

```suggestion
	onContinue,
	isLoading,
```

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -92,17 +87,11 @@ const AdsStepper = ( { formProps } ) => {
 					content: (
 						<AdsCampaign
 							trackingContext="setup-ads"
-							onContinue={ handleCreateCampaignContinue }
+							formProps={ formProps }
```
asvinb:
```suggestion
onContinue={formProps.handleSubmit}
isLoading={formProps.isSubmitting}
```

asvinb (changes requested):
@kt-12 Thanks for the PR! Works well. I suggested a couple of improvements though. Can you kindly let me know what you think?
Also, let's update the tracking comments here:
https://github.com/woocommerce/google-listings-and-ads/blob/5b14c071fc51a11aad28a9fd825cba70eb5611bd/js/src/setup-ads/ads-stepper/index.js#L24-L25

It should be:
```js
 * @fires gla_setup_ads with `{ triggered_by: 'step1-continue-button', action: 'go-to-step2' }`.
 * @fires gla_setup_ads with `{ triggered_by: 'stepper-step1-button', action: 'go-to-step1'}`.
```

js/src/components/paid-ads/ads-campaign.js
kt-12: Makes sense. Fixed.

js/src/setup-ads/ads-stepper/index.js
kt-12: Addressed.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -92,17 +87,16 @@ const AdsStepper = ( { formProps } ) => {
 					content: (
 						<AdsCampaign
 							trackingContext="setup-ads"
-							onContinue={ handleCreateCampaignContinue }
+							onContinue={ formProps.handleSubmit }
+							isLoading={ formProps.isSubmitting }
+							submitButtonText={ __(
+								'Launch paid campaign', 
+								'google-listings-and-ads' 
+							) }
```
kt-12: `AdsCampaign` component is used in 2 flow. One setting up our first paid campaign flow (here). And another for the setting up other paid campaing.

Next is once the first campaign set up you see a button in the right hand side of dashboard- `Add paid campaign` , 
Screenshot 2024-09-06 at 18 01 35


once you click that you see a different flow where `AdsCampaign` is the first component in the flow.

Screenshot 2024-09-06 at 18-05-51 Dashboard ‹ Google for WooCommerce ‹ Marketing ‹ WooCommerce ‹ google-woo — WooCommerce

`submitButtonText` was introduced to make change to button text in a way it doesn't effect the other flow.

@joemcgill

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -25,24 +25,29 @@ import FaqsSection from './faqs-section';
...
 	campaign,
 	onContinue,
+	isLoading,
+	submitButtonText,
 	trackingContext,
 } ) {
 	const isCreation = ! campaign;
 	const formContext = useAdaptiveFormContext();
 	const { isValidForm } = formContext;
+	const continueButtonText = submitButtonText || __( 'Continue', 'google-listings-and-ads' );
```
asvinb: @kt-12 How about we have the default value set for the parameter here: https://github.com/woocommerce/google-listings-and-ads/pull/2577/files#diff-7679c27bf3a373cf135b63e0095de040d052c742d825f0896246656c0c0b53f0R44
instead of having a conditional here?

asvinb (changes requested):
@kt-12 Can you kindly check my comment about the default value for the `submitButtonText` parameter please?

js/src/components/paid-ads/ads-campaign.js
kt-12: Almost missed this one, Sure I'll update the defaults.

kt-12:
> @kt-12 Can you kindly check my comment about the default value for the `submitButtonText` parameter please?

Done. I have addressed other open review comments too.

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -25,19 +25,23 @@ import FaqsSection from './faqs-section';
...
- * so it expects a `CampaignAssetsForm` to existing in its parents.
+ * Please note that this component relies on a CampaignAssetsForm's context and custom adapter,
+ * so it expects a `CampaignAssetsForm` to exist in its parents.
  *
  * @fires gla_documentation_link_click with `{ context: 'create-ads' | 'edit-ads' | 'setup-ads', link_id: 'see-what-ads-look-like', href: 'https://support.google.com/google-ads/answer/6275294' }`
  *
  * @param {Object} props React props.
  * @param {Campaign} [props.campaign] Campaign data to be edited. If not provided, this component will show campaign creation UI.
  * @param {() => void} props.onContinue Callback called once continue button is clicked.
+ * @param {() => void} props.isLoading Callback called once continue button is clicked.
```
asvinb: @kt-12 This is incorrect. `isLoading` is a `boolean`. Can you update please?

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -10,7 +10,6 @@ import { useState } from '@wordpress/element';
  */
 import SetupAccounts from './setup-accounts';
 import AdsCampaign from '.~/components/paid-ads/ads-campaign';
-import SetupBilling from './setup-billing';
```
asvinb, line 13: @kt-12 Looks like the `SetupBilling` component is no more used in the codebase. Can you remove the components and associated unused functions/components please?

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -280,11 +279,15 @@ test.describe( 'Set up Ads account', () => {
 	} );
 
 	test.describe( 'Create your paid campaign', () => {
+		const launchPaidCampaignButton = page.getByRole( 'button', {
```
asvinb: @kt-12 Let's move this within a function in `tests/e2e/utils/pages/setup-ads/setup-budget.js`

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -280,11 +279,15 @@ test.describe( 'Set up Ads account', () => {
...
+			name: 'Launch paid campaign',
+		} );
+
 		test.beforeAll( async () => {
 			setupBudgetPage = new SetupBudgetPage( page );
 		} );
 
-		test( 'Continue to create paid campaign', async () => {
+		test( 'Continue to create paid campaign and create ads', async () => {
```
asvinb:
```suggestion
		test( 'Continue to create paid ad campaign', async () => {
```

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -370,169 +371,57 @@ test.describe( 'Set up Ads account', () => {
...
-						name: "You've set up a paid Performance Max Campaign!",
-					} )
-				).toBeVisible();
-
-				await expect(
-					page.getByRole( 'button', {
-						name: 'Create another campaign',
-					} )
-				).toBeEnabled();
+		test( 'Paid Campaign can be created', async () => {
```
asvinb: @kt-12 This test has no assertion.

asvinb (changes requested):
@kt-12 I left some more comments. Can you check them out please?
Also why we have this error message for the PR?
 Check failure on line 90 in js/src/setup-ads/ads-stepper/index.test.js

js/src/components/paid-ads/ads-campaign.js
kt-12: `@param {() => void}` to `@param {() => boolean}` Does the text seems alright to you?

js/src/setup-ads/ads-stepper/index.js
kt-12: Removed.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
kt-12: Moved as a function in `setup-budget.js`

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
kt-12: I have moved content of this to the next test. As paid campaign created can be determined by success message.

js/src/components/paid-ads/ads-campaign.js
kt-12: Updated.

asvinb:
@kt-12 Can you please fix the linting and unit tests errors please?

kt-12:
> @kt-12 Can you please fix the linting and unit tests errors please.

Unit test changes are not handled in this ticket. It is handled in a separate ticket 2579, which merges 2534, 2536 and 2535

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -25,19 +25,23 @@ import FaqsSection from './faqs-section';
...
- * so it expects a `CampaignAssetsForm` to existing in its parents.
+ * Please note that this component relies on a CampaignAssetsForm's context and custom adapter,
+ * so it expects a `CampaignAssetsForm` to exist in its parents.
  *
  * @fires gla_documentation_link_click with `{ context: 'create-ads' | 'edit-ads' | 'setup-ads', link_id: 'see-what-ads-look-like', href: 'https://support.google.com/google-ads/answer/6275294' }`
  *
  * @param {Object} props React props.
  * @param {Campaign} [props.campaign] Campaign data to be edited. If not provided, this component will show campaign creation UI.
  * @param {() => void} props.onContinue Callback called once continue button is clicked.
+ * @param {() => boolean} props.isLoading Callback called once continue button is clicked.
```
asvinb: @kt-12 What you currently have right now means the following. You're saying that the `isLoading` prop is a function that takes no arguments (represented by ()) and returns a boolean value. This is incorrect since `isLoading` is a `boolean` prop. Furthermore it's not a callback and looks like it's optional. So the JSDoc should be updated to the following:
```suggestion
 * @param {boolean} [props.isLoading] If true, the Continue button will display a loading spinner .
```

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -25,19 +25,23 @@ import FaqsSection from './faqs-section';
...
+ * Please note that this component relies on a CampaignAssetsForm's context and custom adapter,
+ * so it expects a `CampaignAssetsForm` to exist in its parents.
  *
  * @fires gla_documentation_link_click with `{ context: 'create-ads' | 'edit-ads' | 'setup-ads', link_id: 'see-what-ads-look-like', href: 'https://support.google.com/google-ads/answer/6275294' }`
  *
  * @param {Object} props React props.
  * @param {Campaign} [props.campaign] Campaign data to be edited. If not provided, this component will show campaign creation UI.
  * @param {() => void} props.onContinue Callback called once continue button is clicked.
+ * @param {boolean} [props.isLoading] If true, the Continue button will display a loading spinner .
+ * @param {string} props.submitButtonText Text to display on submit button.
```
asvinb: @kt-12 
```suggestion
 * @param {string} [props.submitButtonText] Text to display on submit button.
```
The `submitButtonText` prop is optional.

asvinb (approved):
@kt-12 I left a comment. Feel free to send the ticket to QA once the comment has been addressed.
Thanks!

ankitguptaindia:
Thanks, @kt-12, for the PR!

### QA Status: ⚠️ Needs feedback/remediation. 

#### I’ve encountered an issue where a merchant can create a paid campaign without setting up a billing address if creating a campaign from the Dashboard.

**Use Case:** 
During the onboarding process, it’s not mandatory to create a paid campaign or add billing details. Merchants can complete onboarding by creating free listings instead. In such cases, since billing information isn’t added during onboarding, once the flow is completed, the merchant can navigate to the Plugin Dashboard, click on "Add Paid Campaign," and successfully launch a paid campaign without having any billing information.

 **Can we make this conditional?**:
If billing info wasn’t added during the onboarding steps, the system should direct the merchant to the billing setup page during the ads setup flow. If billing information is already available, it should bypass the billing setup page.


**Screencast for correct flow:** 


https://github.com/user-attachments/assets/27295d50-d30a-4434-947c-fdce5a898206



**Next Ste:-** Moving back to in-progress. 

**_Note: Please let me know if we have another task to handle this scenario._**

joemcgill:
Just seeing this QA feedback and wanted to provide some additional context:

> Can we make this conditional?:
If billing info wasn’t added during the onboarding steps, the system should direct the merchant to the billing setup page during the ads setup flow. If billing information is already available, it should bypass the billing setup page.

This PR is a sub-task of the larger work defined in #2460, in which we are streamlining the campaign creation flow from the dashboard so that it can usually all be completed in one step. A critical part of that effort will be handled in #2535, in which the same form used in the onboarding flow for campaign creation will be used in this flow as well. Once that issue is complete, the budget setup will be included in the campaign creation form, the same way that it works now in the onboarding flow.

For the purposes of this PR, if we can easily disable the CTA button when we don't have an approved billing status, then I think that would be fine, but conditionally showing the 3rd step is counter to the purpose of this issue.

.wp-env.override.json
```diff
@@ -0,0 +1,3 @@
+{
```
asvinb: @kt-12 Please remove the file from the commit.

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -105,7 +119,7 @@ export default function AdsCampaign( {
 			<StepContentFooter>
 				<AppButton
 					isPrimary
-					disabled={ ! isValidForm }
+					disabled={ ! enableContinue }
```
asvinb: @kt-12 My understanding is that the Continue button should be disabled if:

1. the form is not valid (using the `isValidForm` variable)
2. if billing is not set up/approved.

How about we add the following condition then?
```js
disabled={
	! isValidForm ||
	billingStatus?.status !==
		GOOGLE_ADS_BILLING_STATUS.APPROVED
}
```

There is no need to show the spinner while the budget status is being resolved.

asvinb (changes requested):

js/src/components/paid-ads/ads-campaign.js
kt-12: Button will be disabled till we have status so we don't need the spinner, totally make sense to me. Thanks @asvinb .

joemcgill (approved):
This looks like it's working now when billing is not set up properly.

js/src/components/paid-ads/ads-campaign.js
```diff
@@ -101,10 +108,16 @@ export default function AdsCampaign( {
 			<StepContentFooter>
 				<AppButton
 					isPrimary
-					disabled={ ! isValidForm }
+					disabled={
+						! isValidForm ||
+						( trackingContext === 'setup-ads' &&
```
asvinb: @kt-12 Why we need to check for `trackingContext` 🤔 ? Irrespective of the context, shouldn't the button be disabled if the form is not valid and billing has not been approved?

kt-12: This form is used in a different place too where billing is not involved. [ref](https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1747864917)

js/src/components/paid-ads/ads-campaign.js
asvinb: I see you've updated the comment with the correct syntax.

js/src/components/paid-ads/ads-campaign.js
kt-12: Since this will be revisited in 2535 with a holistic view of new billing scenarios, I'll close this with the current condition.

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* WooCommerce - Version 9.3.2
* PHP: 8.3
* Web Server: Nginx
* OS: macOS Sonoma 14.6.1 

</Details>


✅ When Billing info is added from the onboarding steps: 


https://github.com/user-attachments/assets/941b5fc3-9a06-4d16-97a0-2860656f0e97




✅ When Billing info is not added: 


https://github.com/user-attachments/assets/22f1b7f8-0e29-4613-93b8-e3eb00db3962

ankitguptaindia (approved):

js/src/components/paid-ads/ads-campaign.js
asvinb: Sounds good @kt-12 .

asvinb:
@eason9487 This is ready for review please.
Thanks!

eason9487:
Hi @kt-12 and @asvinb, I haven't looked into the code changes as the JS unit testing and e2e testing didn't pass.

- https://github.com/woocommerce/google-listings-and-ads/actions/runs/10940178407/job/30372010381
- https://github.com/woocommerce/google-listings-and-ads/actions/runs/11010342771/job/30572044380

Could you help with ensuring a PR passes all testing before passing its ticket to Woo CR stage?

kt-12:
Hi @eason9487, Test changes were not done in this ticket as it will be handled in #2579, where we have changes from #2534, #2535, #2536 is merged (see first AC #2536 ).

eason9487:
Hi @kt-12

> Test changes were not done in this ticket as it will be handled in #2579, where we have changes from #2534, #2535, #2536 is merged (see first AC #2536 ).

Not sure if I missed anything but they don't seem to include postponing the update to e2e testing

kt-12:
@joemcgill @asvinb I have updated e2e to support new flow. E2E is currently failing at form submit. While most of the flow worked fine clicking on `Launch Paid Campaign` is working differently in e2e than when tested manually. 

Few observation:
- Click of button gives an alert 'You have unsaved campaign data. Are you sure you want to leave?', which related to incomplete form.
- URL on the button click is a bit different than the one we do manually.
- I am also getting following error in API call made 
```
{
    "message": "Please reconnect your Jetpack account.",
    "status": 401,
    "code": "JETPACK_DISCONNECTED"
}
```

I tried recconecting to jetpack again just before the click of button but it filed too.

Curl extracted from E2E :- 
```
curl 'http://localhost:8889/wp-json/wc/gla/ads/campaigns?_locale=user' \
  -H 'Accept: application/json, */*;q=0.1' \
  -H 'Accept-Language: en-US' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Cookie: wordpress_test_cookie=WP%20Cookie%20check; wordpress_logged_in_23778236db82f19306f247e20a353a99=admin%7C1727504916%7CbP9l9wZZLAEhjXSgSnYIKLoqAFkUkIChMNtDV3Qxaw9%7Cb8843a1de830928d1d4215d5c310740c13de29e72c8e30e81285da9d6d5ad988; tk_ai=woo%3AxSUpIUAfZ0IfZUDBI8ZCAVpW; wp-settings-time-1=1727332131' \
  -H 'Origin: http://localhost:8889' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://localhost:8889/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.6312.4 Safari/537.36' \
  -H 'X-WP-Nonce: 431d70c3b8' \
  -H 'sec-ch-ua: "Chromium";v="123", "Not:A-Brand";v="8"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "Windows"' \
  --data-raw '{"amount":1,"targeted_locations":["US"],"label":"wc-web"}'
  ````

**Comment by kt-12** [comment]
> Hi @kt-12
> 
> > Test changes were not done in this ticket as it will be handled in #2579, where we have changes from #2534, #2535, #2536 is merged (see first AC #2536 ).
> 
> Not sure if I missed anything but they don't seem to include postponing the update to e2e testing

Hi @eason9487, 

Sorry for the initial confusion. I have fixed e2e for this task and it did run without any failure [(run)](https://github.com/woocommerce/google-listings-and-ads/actions/runs/11077820826/job/30783906059). Jest changes will still be handled separately as we see some overlap within other sister tasks. I feel this is good for WooTeam review now, adding this back to Woo CR. Thanks.

File: js/src/components/paid-ads/ads-campaign.js [hunk 15]
```diff
@@ -101,10 +108,16 @@ export default function AdsCampaign( {
...
+							billingStatus?.status !==
+								GOOGLE_ADS_BILLING_STATUS.APPROVED )
+					}
+					loading={ isLoading }
 					onClick={ onContinue }
 				>
-					{ __( 'Continue', 'google-listings-and-ads' ) }
+					{ submitButtonText }
```
eason9487: Considering this button now requires 5 props to render and its only internal dependency is on one of those conditions, its degree of customization makes it unsuitable to be part of this component. It's better to externalize it as a single prop such as `continueButton` to composite this component.

```jsx
<StepContentFooter>
	{ continueButton }
</StepContentFooter>

```

js/src/setup-ads/ads-stepper/setup-billing/billing-saved-card/index.js
```diff
@@ -1,79 +0,0 @@
...
-import { __ } from '@wordpress/i18n';
-import GridiconCreditCard from 'gridicons/dist/credit-card';
-import { createInterpolateElement } from '@wordpress/element';
-
-/**
- * Internal dependencies
- */
-import toAccountText from '.~/utils/toAccountText';
-import AppSpinner from '.~/components/app-spinner';
-import TitleButtonLayout from '.~/components/title-button-layout';
```
eason9487, line 13: The files within `js/src/components/title-button-layout` could be removed as well.

tests/e2e/utils/pages/setup-ads/setup-budget.js
```diff
@@ -84,6 +84,18 @@ export default class SetupBudget extends MockRequests {
 		);
 	}
 
+	/**
+	 * Get the Launch paid cmpaign button.
+	 *
+	 * @return {import('@playwright/test').Locator} Launch paid cmpaign button.
```
eason9487: Two typos: `cmpaign`

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -281,10 +281,13 @@ test.describe( 'Set up Ads account', () => {
 
 	test.describe( 'Create your paid campaign', () => {
 		test.beforeAll( async () => {
-			setupBudgetPage = new SetupBudgetPage( page );
+			// `Launch Paid Campaign` button is only enabled when billing status is approved.
```
eason9487: This code comment is a bit imprecise as the billing status is not the only condition determining whether the button is enabled.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -387,152 +390,45 @@ test.describe( 'Set up Ads account', () => {
...
+		test( 'It should show the campaign creation success message', async () => {
+			// Mock the campaign creation request.
+			await setupBudgetPage.mockCampaignCreation( budget, {} );
 
-				await expect(
-					page.getByText(
-						'Great! You already have billing information saved for this'
-					)
-				).toBeVisible();
+			await setupBudgetPage.getLaunchPaidCampaignButton().click();
```
eason9487: Wonder why the test was changed to not wait for the campaign creation request? (`setupBudgetPage.mockCampaignCreationAndAdsSetupCompletion`)

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -387,152 +390,45 @@ test.describe( 'Set up Ads account', () => {
...
-			test( 'It should say that the billing is setup', async () => {
-				//Every 30s the page will check if the billing status is approved and it will trigger the campaign creation.
-				await setupBudgetPage.awaitForBillingStatusRequest();
-				await setupBudgetPage.mockCampaignCreationAndAdsSetupCompletion(
-					budget,
-					[ 'US' ]
-				);
+		test( 'It should show the campaign creation success message', async () => {
+			// Mock the campaign creation request.
+			await setupBudgetPage.mockCampaignCreation( budget, {} );
```
eason9487: Although it won't cause issues, the second parameter should be an array.

eason9487 (commented):
Thanks for the work. The main concern is whether there is a better way to composite the continue button for the `AdsCampaign` component rather than passing 5 props.

js/src/components/paid-ads/ads-campaign.js
joemcgill: @eason9487 I believe this continue button is only temporarily going to need to handle this many conditions until #2535 is completed, to ensure that the current flow could still be tested properly. I'd like to avoid making any additional modifications to the `AdsCampaign` component in this PR, which is really only meant to remove the billing step because it will be handled during step 3 once we've consolidated the form from the onboarding flow.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
kt-12: `setupBudgetPage.mockCampaignCreationAndAdsSetupCompletion` was waiting indefinitely. We only needed the campaign to created to get the button clickable, hence the split.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
kt-12: Fixed.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
kt-12: Removed the comment. Changing it was making it even more hard to understand.

tests/e2e/utils/pages/setup-ads/setup-budget.js
kt-12: Fixed.

joemcgill:
I'm updating this PR to point to the new consolidated feature branch `feature/2459-campaign-creation-flow`. We'll need to resolve the merge conflicts and then send it back for review.

joemcgill:
@eason9487 besides the changes requested for the continue button, I think this is ready for another look.

tests/e2e/utils/pages/setup-ads/setup-budget.js
```diff
@@ -75,6 +75,48 @@ export default class SetupBudget extends MockRequests {
...
+	 */
+	registerBudgetRecommendationResponse() {
+		return this.page.waitForResponse(
+			( response ) =>
+				response
+					.url()
+					.includes( '/gla/ads/campaigns/budget-recommendation' ) &&
+				response.status() === 200
+		);
+	}
```
eason9487: Looks like these two methods that had been removed from #2551 were added back when resolving code conflicts.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
eason9487: I don't see waiting indefinitely if it's used similarly to https://github.com/woocommerce/google-listings-and-ads/blob/3d38d913c5ff4b2bce72e177a07a1188ead68dd8/tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js#L527-L535

Without that check, the "It should show the campaign creation success message" test case can still pass after skipping the campaign creation by replacing the following line with `return Promise.resolve()`:

https://github.com/woocommerce/google-listings-and-ads/blob/fdd3aae49f0520e5d94c0f15581025618a034d15/js/src/hooks/useAdsSetupCompleteCallback.js#L39

This makes it not fulfill the purpose of the test case it describes.

tests/e2e/utils/pages/setup-ads/setup-budget.js
```diff
@@ -149,13 +191,13 @@ export default class SetupBudget extends MockRequests {
...
-	 * Mock the campaign creation process and the Ads setup completion.
+	 * Mock the campaign creation process.
 	 *
-	 * @param {string} budget The campaign budget.
+	 * @param {string} budget          The campaign budget.
 	 * @param {Array}  targetLocations The targeted locations.
 	 * @return {Promise<void>}
 	 */
-	async mockCampaignCreationAndAdsSetupCompletion( budget, targetLocations ) {
+	async mockCampaignCreation( budget, targetLocations ) {
```
eason9487: Referring to https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1782514032, although I don't think this method needs to be extracted, the name of this method doesn't match its implementation because it still includes the mocking of Ads setup completion.

https://github.com/woocommerce/google-listings-and-ads/blob/fdd3aae49f0520e5d94c0f15581025618a034d15/tests/e2e/utils/pages/setup-ads/setup-budget.js#L202-L207

js/src/components/paid-ads/ads-campaign.js
eason9487: https://github.com/woocommerce/google-listings-and-ads/blob/34c53d7d4d37fbfdf46b00aa36195c2e09710eb4/js/src/components/paid-ads/ads-campaign/ads-campaign.js#L103-L117

https://github.com/woocommerce/google-listings-and-ads/blob/34c53d7d4d37fbfdf46b00aa36195c2e09710eb4/js/src/components/paid-ads/ads-campaign/ads-campaign.js#L176-L182

From the current revision of `AdsCampaign` in #2623, I couldn't figure out how this button only *temporarily* needs to handle 5 props, because both PRs currently increase the complexity and conditional branching of this button to varying degrees.

All of these changes convinced me that both PRs should try to see if it's possible to move this button out of the `AdsCampaign` component sooner rather than having to resolve more code conflicts and re-test all button scenarios when merging them with each other.

tests/e2e/utils/pages/setup-ads/setup-budget.js
kt-12: Removed.

kt-12:
@eason9487 https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1782514032

Thank you, you were correct. awating after button is clicked did solve waiting indefinitely issue. I have also removed the newly introduced function.

tests/e2e/utils/pages/setup-ads/setup-budget.js
kt-12: Removed the function. Test working with early function when await was moved to after the button click.

kt-12:
@eason9487 we have addressed your review comments and also moved continue button out of AdsCampaign. Moving it back to you for CR.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -102,13 +103,11 @@ export default function AdsCampaign( {
...
-						disabled={ ! isValidForm }
-						onClick={ onContinue }
-					>
-						{ __( 'Continue', 'google-listings-and-ads' ) }
-					</AppButton>
+					<ContinueButton
+						onContinue={ onContinue }
+						setupAdsFlow={ trackingContext === 'setup-ads' }
+						isLoading={ isLoading }
+					/>
```
eason9487: Thanks for the further update! However, it doesn't solve the concern in https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1780557466.

To externalize the button, the changes may be like:
- In `AdsCampaign`
   - After externalization, `AdsCampaign` won't have the `loading` and `onContinue` props
   - The continue button is given via a prop or a render prop
   - For example
      ```jsx
      export default function AdsCampaign( {
      	campaign,
      	continueButton,
      	trackingContext,
      } ) {
      	// ... omitted
      
      	return (
      		<StepContent>
      			// ... omitted
      			<StepContentFooter>
      				<StepContentActions>
      					{ typeof continueButton === 'function'
      						? continueButton( formContext )
      						: continueButton }
      				</StepContentActions>
      				<PaidAdsFaqsPanel />
      			</StepContentFooter>
      		</StepContent>
      	);
      }
      ```
- In `AdsStepper`
   - The [ContinueButton](https://github.com/woocommerce/google-listings-and-ads/blob/fe7bab660d13bab427fa28a3b37e7f51569d5865/js/src/components/paid-ads/ads-campaign/continue-button.js) can be inlined and passed via the `continueButton` prop
   - For example
      ```jsx
      const AdsStepper = ( { formProps } ) => {
      	// ... omitted
      
      	const isDisabledLaunch = ! formProps.isValidForm; // and also the conditions in https://github.com/woocommerce/google-listings-and-ads/blob/fe7bab660d13bab427fa28a3b37e7f51569d5865/js/src/components/paid-ads/ads-campaign/continue-button.js#L34-L37
      
      	return (
      		<Stepper
      			// ... omitted
      			steps={ [
      				// ... omitted
      				{
      					key: '2',
      					label: __(
      						'Create your paid campaign',
      						'google-listings-and-ads'
      					),
      					content: (
      						<AdsCampaign
      							trackingContext="setup-ads"
      							continueButton={
      								<AppButton
      									isPrimary
      									text={ __(
      										'Launch paid campaign',
      										'google-listings-and-ads'
      									) }
      									disabled={ isDisabledLaunch }
      									loading={ formProps.isSubmitting }
      									onClick={ formProps.handleSubmit }
      								/>
      							}
      						/>
      					),
      					onClick: handleStepClick,
      				},
      			] }
      		/>
      	);
      };
      ```
- In `CreatePaidAdsCampaign` and `EditPaidAdsCampaign`
   - Similar to `AdsStepper`, change to pass the button via the `continueButton` prop
   - For example
      ```jsx
      const SetupAdsForm = () => {
      	// ... omitted
      
      	// This can be a component shared between CreatePaidAdsCampaign and EditPaidAdsCampaign
      	const ContinueButton = ( formProps ) => {
      		return (
      			<AppButton
      				isPrimary
      				text={ __( 'Continue', 'google-listings-and-ads' ) }
      				disabled={ ! formProps.isValidForm }
      				onClick={ () => handleContinueClick( STEP.ASSET_GROUP ) }
      			/>
      		);
      	};
      
      	return (
      		<>
      			// ... omitted
      			<CampaignAssetsForm
      				initialCampaign={ {
      					amount: 0,
      					countryCodes: initialCountryCodes,
      				} }
      				onSubmit={ handleSubmit }
      			>
      				<Stepper
      					currentStep={ step }
      					steps={ [
      						{
      							key: STEP.CAMPAIGN,
      							label: __(
      								'Create paid campaign',
      								'google-listings-and-ads'
      							),
      							content: (
      								<AdsCampaign
      									trackingContext={ eventContext }
      									continueButton={ ContinueButton }
      								/>
      							),
      							onClick: handleStepperClick,
      						},
      						// ... omitted
      					] }
      				/>
      			</CampaignAssetsForm>
      		</>
      	);
      };
      ```

In this way, the logic of this button will be split up at a higher level for each consumer component, further avoiding the need to add/pass on different contextual conditions to the `AdsCampaign` and centralize all the conditions with the same button.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -57,6 +57,10 @@ const AdsStepper = ( { formProps } ) => {
 		continueStep( '2' );
 	};
 
+	// @todo: Add check for billing status once billing setup is moved to step 2.
```
joemcgill, line 60: Checking billing status cannot be done here, it must instead be done after an Ads account is connected in Step 1. The billing status check will be handled in `AdsCampaign` after #2535 is completed, so we can ignore this check for now.

joemcgill:
@eason9487 I've added a todo concerning the need to check for billing status, which should be handled after #2623 is updated, since the new consolidated form will be responsible for checking the billing status.

js/src/components/paid-ads/continue-button.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+import AppButton from '.~/components/app-button';
+
+/**
+ * Renders Continue button on paid ad campaign create and edit page.
+ *
+ * @param {Object} props Props
+ * @param {Object} props.formProps Form props forwarded from `Form` component.
+ * @param {Function} props.handleContinueClick Function to handle the continue button click.
+ * @return {import(".~/components/paid-ads/continue-button").default} The button.
```
eason9487: The return type is not the `ContinueButton` function itself.

js/src/components/paid-ads/continue-button.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+const ContinueButton = ( { formProps, handleContinueClick } ) => {
```
eason9487: The props naming convention of an event handler usually starts with `on` followed by the name of the event. A name starting with `handle` is used for an event handler set to the corresponding prop starting with `on`.

The `handleContinueClick` is directly forwarded to the button, so the naming would be better to use `onClick`.

js/src/components/paid-ads/continue-button.js
```diff
@@ -0,0 +1,30 @@
...
+	return (
+		<AppButton
+			isPrimary
+			text={ __( 'Continue', 'google-listings-and-ads' ) }
+			disabled={ ! formProps.isValidForm }
+			onClick={ () => handleContinueClick() }
```
eason9487: It doesn't need to wrap an additional arrow function.

js/src/components/paid-ads/continue-button.js
kt-12: I have changed it to `import(".~/components/app-button").default`

eason9487: It's still incorrect. The return type is `JSX.Element` or the `@return` tag can be omitted.

js/src/components/paid-ads/continue-button.js
kt-12: Understood. Changed the name to onClick.

js/src/components/paid-ads/continue-button.js
kt-12: Fixed. onClick={ onClick }

kt-12:
@eason9487 I have addressed your review changes.  Moved it back to you for CR.

js/src/components/paid-ads/continue-button.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+import AppButton from '.~/components/app-button';
+
+/**
+ * Renders Continue button on paid ad campaign create and edit page.
+ *
+ * @param {Object} props Props
+ * @param {Object} props.formProps Form props forwarded from `Form` component.
+ * @param {Function} props.onClick Function to handle the continue button click.
+ * @return {import(".~/components/app-button").default} The button.
```
eason9487: Related to https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1792721347. The return type is not the `AppButton` function either. The return type is `JSX.Element` or the `@return` tag can be omitted.

js/src/pages/create-paid-ads-campaign/index.js
```diff
@@ -146,9 +147,16 @@ const CreatePaidAdsCampaign = () => {
 							content: (
 								<AdsCampaign
 									trackingContext={ eventContext }
-									onContinue={ () =>
-										handleContinueClick( STEP.ASSET_GROUP )
-									}
+									continueButton={ ( props ) => (
+										<ContinueButton
+											{ ...props }
+											handleContinueClick={ () =>
```
eason9487: The `handleContinueClick` should be updated accordingly after [renaming](https://github.com/woocommerce/google-listings-and-ads/pull/2577/commits/2564a42648754191c155cb8d53e2ad9ac3df25a9#diff-6b724af416790c2e621567319e6734b68b00681465f87e9e0e710dda87aa6870R19).

js/src/pages/edit-paid-ads-campaign/index.js
```diff
@@ -197,9 +198,16 @@ const EditPaidAdsCampaign = () => {
 								<AdsCampaign
 									campaign={ campaign }
 									trackingContext={ eventContext }
-									onContinue={ () =>
-										handleContinueClick( STEP.ASSET_GROUP )
-									}
+									continueButton={ ( props ) => (
+										<ContinueButton
+											{ ...props }
+											handleContinueClick={ () =>
```
eason9487: The `handleContinueClick` should be updated accordingly after [renaming](https://github.com/woocommerce/google-listings-and-ads/pull/2577/commits/2564a42648754191c155cb8d53e2ad9ac3df25a9#diff-6b724af416790c2e621567319e6734b68b00681465f87e9e0e710dda87aa6870R19).

eason9487 (changes requested):

js/src/pages/create-paid-ads-campaign/index.js
kt-12: My mistake, I did make the changes but the push didn't go through earlier. Sorry, for that.
 Now it should reflect to onClick.

js/src/pages/edit-paid-ads-campaign/index.js
kt-12: Now it should reflect to onClick.

js/src/components/paid-ads/continue-button.js
kt-12: I have changed it to JSX.Element now. 

Return type was a bit confusing for me here. I was referring to `help-icon-button.js` which had the element import statement as return type. Some places even has `@return {AppButton}` as return type while other have JSX.Element.

eason9487 (approved):

PR #2594, opened 2024-09-08
Update UI to support minimum and maximum shipping times.

jorgemd24:

Part of pcTzPl-2qP-p2

This PR introduces a UI to manage minimum and maximum shipping times, and it updates the input to include increase and decrease buttons (like a stepper input). Additionally, we're considering showing "Same Day" instead of 0 to make it clearer for merchants.

image

Since mixing a string like "Same Day" with the actual integer values for shipping times could make the code more complex, I opted to use the placeholder attribute to display "Same Day" instead of 0 days to keep things simple. However, I'm open to other suggestions or approaches.

For more context see this comment here: pcTzPl-2qP-p2#comment-3652

#### Other Steeper Inputs in WC

https://github.com/woocommerce/woocommerce/blob/64dcafe29b52cff2dee1d56f5e06caad09a0c470/packages/js/product-editor/src/components/number-control/number-control.tsx#L166-L222

1. Follow the steps mentioned in this PR to update the database: https://github.com/woocommerce/google-listings-and-ads/pull/2520
2. Navigate to Dashboard -> Free Listings -> Edit
3. Test the new input buttons by increasing or decreasing the values. If the value is 0, it should display "Same Day".
4. Verify that the data is being stored correctly.
5. Add and modify shipping times as needed.
6. Disconnect all your accounts.
7. Do the onboarding, and see if everything works correctly.

### Additional details:

I haven't updated the add and edit modals to use the stepper component yet to make reviewing this PR easier. Also, I want to ask for confirmation from the designers to see if we want to implement it there as well.

### Changelog entry

>

### DIFF
--- a/js/src/components/free-listings/configure-product-listings/checkErrors.js
+++ b/js/src/components/free-listings/configure-product-listings/checkErrors.js
@@ -98,7 +98,7 @@ const checkErrors = (
 	if (
 		values.shipping_time === 'flat' &&
 		( shippingTimes.length < finalCountryCodes.length ||
-			shippingTimes.some( ( el ) => el.time < 0 ) )
+			shippingTimes.some( ( el ) => el.time < 0 || el.maxTime < 0 ) )
 	) {
 		errors.shipping_country_times = __(
 			'Please specify estimated shipping times for all the countries, and the time cannot be less than 0.',
--- a/js/src/components/free-listings/configure-product-listings/checkErrors.js
+++ b/js/src/components/free-listings/configure-product-listings/checkErrors.js
@@ -106,6 +106,16 @@ const checkErrors = (
 		);
 	}
 
+	if (
+		values.shipping_time === 'flat' &&
+		shippingTimes.some( ( el ) => el.time > el.maxTime )
+	) {
+		errors.shipping_country_times = __(
+			'The minimum shipping time must not be less than the maximum shipping time.',
+			'google-listings-and-ads'
+		);
+	}
+
 	/**
 	 * Check tax rate (required for U.S. only).
 	 */
--- a/js/src/components/free-listings/configure-product-listings/checkErrors.js
+++ b/js/src/components/free-listings/configure-product-listings/checkErrors.js
@@ -111,7 +111,7 @@ const checkErrors = (
 		shippingTimes.some( ( el ) => el.time > el.maxTime )
 	) {
 		errors.shipping_country_times = __(
-			'The minimum shipping time must not be less than the maximum shipping time.',
+			'The minimum shipping time must not be more than the maximum shipping time.',
 			'google-listings-and-ads'
 		);
 	}
--- a/js/src/components/free-listings/configure-product-listings/checkErrors.test.js
+++ b/js/src/components/free-listings/configure-product-listings/checkErrors.test.js
@@ -419,7 +419,7 @@ describe( 'checkErrors', () => {
 				expect( errors ).toHaveProperty( 'shipping_country_times' );
 				expect( errors.shipping_country_times ).toMatchSnapshot();
 			} );
-			
+
 			it( `When minimum max_time is < 0, should not pass`, () => {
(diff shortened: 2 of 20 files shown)


js/src/components/free-listings/configure-product-listings/checkErrors.js
```diff
@@ -98,14 +98,24 @@ const checkErrors = (
...
 			'google-listings-and-ads'
 		);
 	}
 
+	if (
+		values.shipping_time === 'flat' &&
+		shippingTimes.some( ( el ) => el.time > el.maxTime )
+	) {
+		errors.shipping_country_times = __(
+			'The minimum shipping time must not be less than the maximum shipping time.',
```
puntope: I believe this should be "The minimum shipping time must not be more than the maximum shipping time."

js/src/components/free-listings/configure-product-listings/checkErrors.test.js
```diff
@@ -400,7 +401,7 @@ describe( 'checkErrors', () => {
 			} );
 
 			it( `When there are any shipping times is < 0, should not pass`, () => {
```
puntope: Can we add separate assertion for `time` and `max_time` is < 0

puntope:
It's handled by the error checker but I think it would be even better if we add some checks on the onChange method to verify that `time` is not bigger than `max_time`.

puntope:
> I haven't updated the add and edit modals to use the stepper component yet to make reviewing this PR easier. Also, I want to ask for confirmation from the designers to see if we want to implement it there as well.

Potentially this will be handled separately in other PR. And maybe not setting the stepper... But at least we want to set also the placeholder "Same day" instead of 0? 

Screenshot 2024-09-09 at 12 08 53

jorgemd24:
Thanks @puntope for looking at this, I've addressed your comments, could you please take another look?

Regards these points:

> It's handled by the error checker but I think it would be even better if we add some checks on the onChange method to verify that time is not bigger than max_time.

I think to keep consistency with the rest of the form, we could use this change to all fields or leave it as it is. I'm not sure what was the reason to show the validation only when clicking "continue," but if we think it's worth changing, we could explore it in a separate PR.

> Potentially this will be handled separately in other PR. And maybe not setting the stepper... But at least we want to set also the placeholder "Same day" instead of 0?

Thanks for the suggestion! I'll discuss this with the designers to see what's best for this form, for now I think we can keep it as it is and then I can adjust it.

puntope:
Thanks @jorgemd24 I found a new issue when setting `time`or `max_time` as "Same day"


https://github.com/user-attachments/assets/8bc60164-bb9a-4028-91f3-0514e1f4b25e

jorgemd24:
Hi @puntope, thanks for reviewing this! The issue was linked to a change I made in the other PR, where `0` was being treated as an invalid value. I've removed that check since the schema already requires those fields. [e4686c5](https://github.com/woocommerce/google-listings-and-ads/pull/2594/commits/e4686c559b8a86c4214371a8974acf1cce565aca)

I also found a small issue with the handleBlur function and the original value, which I've fixed here: [f218b47](https://github.com/woocommerce/google-listings-and-ads/pull/2594/commits/f218b47483fb73194100987a4b6bb0607f97c2c5)

Could you please take another look? Thanks!

PS: The PHP tests failed because of this issue: pb22l9-3cx-p2#comment-3657 but they are OK locally.

puntope (approved):
Thx @jorgemd24 for the tweaks. Now it's working as expected.

Just one thing about the UI. I see the minus symbol in the right. When normally the standard is to place it in the left. 

Screenshot 2024-09-16 at 18 25 47


I believe maybe we can discuss with UI. But not a blocker.

Screenshot 2024-09-16 at 18 23 14

Screenshot 2024-09-16 at 18 21 12

jorgemd24:
Thanks @puntope for the review

> Just one thing about the UI. I see the minus symbol in the right. When normally the standard is to place it in the left.

I understand your point, however, the new version of the product editor uses the minus symbol on the right, so the plan is to maintain that style for consistency.

image

Issue #2579, opened 2024-09-04
Update Jest tests for Set up paid ads flow


joemcgill:
>[!CAUTION]
> This is blocked by #2534, #2535, and #2536.

As a follow-up to the changes to streamline the paid ads flow defined in #2460, The related Jest tests will need to be updated once changes from the following issues are completed:

- https://github.com/woocommerce/google-listings-and-ads/issues/2534
- https://github.com/woocommerce/google-listings-and-ads/issues/2535
- https://github.com/woocommerce/google-listings-and-ads/issues/2536

## Acceptance criteria

- JS Unit tests related to the updated Ads Setup flow are passing (ex: https://github.com/woocommerce/google-listings-and-ads/actions/runs/10706073441)

## Implementation details
This work should branch from the [feature/2460-simplify-paid-ads-setup](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2460-simplify-paid-ads-setup) branch once each of the related issues listed in the overview of this ticket have been merged.

There is an AdsStepper test at `js/src/setup-ads/ads-stepper/index.test.js` that is currently failing.

asvinb:
@joemcgill @kt-12 The failing test was fixed in the following PR:
https://github.com/woocommerce/google-listings-and-ads/pull/2623/files#diff-c10ae52d5079fec4d79e7e3a503fd239e7c34f2f23567d3ed563dc818cf6cee0R42

I don't think there is anything else to do here. Furthermore, all tests pass on the [feature/2459-campaign-creation-flow](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2459-campaign-creation-flow) branch.

joemcgill:
Closing this as redundant now that this has been addressed in #2595 and #2632.

PR #2595, opened 2024-09-09
Streamline campaign setup: When the accounts have been connected before, skip accounts setup step during the Ads-onboarding

### LINKED ISSUE #2534: Remove the accounts connection page from the Ads Setup flow

kt-12:

Closes #2534.

### Screenshots:

After removal of accounts setup page.
Screenshot 2024-09-09 at 09 11 26

1. Delete `gla_ads_setup_completed_at` options. 
2. Go to dashboard, under `Performance (Paid Campaigns)` section click `Add paid campaign`  
3. You should not see the `Setup accounts page` as shown in the screenshot. Proceed with the other steps and see if you are able to create first paid campaign.

### Changelog entry

> Update - When the accounts have been connected before, skip accounts setup step during the Ads-onboarding.


asvinb (changes requested):
@kt-12 Can you please do the following:
* Change the target branch for your PR to [feature/2460-simplify-paid-ads-setup](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2460-simplify-paid-ads-setup) as per the IB.
* Fix the linting errors.
* Review the E2E tests. For e.g there's empty tests for `Set up your accounts page` in `tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js`

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,12 +28,18 @@ import {
  */
 const AdsStepper = ( { formProps } ) => {
 	const [ step, setStep ] = useState( '1' );
+	const { hasFinishedResolution, hasGoogleAdsConnection } =
```
asvinb: @kt-12 As per your comment here: https://github.com/woocommerce/google-listings-and-ads/issues/2534#issuecomment-2353038593, there's no page reload. Instead the `AdsStepper` component gets re rendered because `hasGoogleAdsConnection` has been modified once you connect a Google Ads account. Instead of using the `useGoogleAdsAccount` hook, which will cause the component to render again whenever there's a change, I _think_ we can use the `glaData.initialWpData.adsId` global variable which is static and set upon page load. So upon page load, if `glaData.initialWpData.adsId` is not `null` then it means that there's a connected Google Ads account. Even when you connect an Ads account, `glaData.initialWpData.adsId` will still be `null`. So you can have something like this:
```js
const hasGoogleAdsId = glaData.initialWpData.adsId;
```

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,12 +28,18 @@ import {
...
+		useGoogleAdsAccount();
 
 	useEventPropertiesFilter( FILTER_ONBOARDING, {
 		context: CONTEXT_ADS_ONBOARDING,
 		step,
 	} );
 
+	if ( ! hasFinishedResolution && ! hasGoogleAdsConnection ) {
```
asvinb: If we use `glaData.initialWpData.adsId`, there won't be the need to show a spinner since the data is already available.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -55,21 +62,38 @@ const AdsStepper = ( { formProps } ) => {
 	};
 
 	const handleSetupAccountsContinue = () => {
-		continueStep( '2' );
+		continueStep( hasGoogleAdsConnection ? '1' : '2' );
```
asvinb:
```suggestion
		continueStep( hasGoogleAdsConnection '2' );
```
This will always be `2` since the accounts step will always be displayed as the first step.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -55,21 +62,38 @@ const AdsStepper = ( { formProps } ) => {
...
 	};
 
 	const handleCreateCampaignContinue = () => {
-		continueStep( '3' );
+		continueStep( hasGoogleAdsConnection ? '2' : '3' );
```
asvinb:
```suggestion
		continueStep( hasGoogleAdsId ? '2' : '3' );
```

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -55,21 +62,38 @@ const AdsStepper = ( { formProps } ) => {
...
 
-	return (
-		// This Stepper with this class name
-		// should be refactored into separate shared component.
-		// It is also used in the Setup MC flow.
-		<Stepper
-			className="gla-setup-stepper"
-			currentStep={ step }
-			steps={ [
+	const getSteps = () => {
```
asvinb: @kt-12 Instead of adding the accounts step afterwards, L96, it's better if we remove it from the steps. It's easier.
```js
let steps = [
	{
		key: '1',
		label: __( 'Set up your accounts', 'google-listings-and-ads' ),
		content: (
			<SetupAccounts onContinue={ handleSetupAccountsContinue } />
		),
		onClick: handleStepClick,
	},
	{
		key: '2',
		label: __( 'Create your paid campaign', 'google-listings-and-ads' ),
		content: (
			<AdsCampaign
				trackingContext="setup-ads"
				onContinue={ handleCreateCampaignContinue }
			/>
		),
		onClick: handleStepClick,
	},
	{
		key: '3',
		label: __( 'Set up billing', 'google-listings-and-ads' ),
		content: <SetupBilling formProps={ formProps } />,
		onClick: handleStepClick,
	},
];

if ( hasGoogleAdsId ) {
	// Remove first step if there's an Ads account connected.
	steps.shift();

	steps = steps.map( ( singleStep ) => {
		return {
			...singleStep,
			key: ( parseInt( singleStep.key, 10 ) - 1 ).toString(),
		};
	}, [] );
}
```

then to update the `key` property of each step, you can use the `map` function.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -83,27 +107,21 @@ const AdsStepper = ( { formProps } ) => {
...
+	};
+
+	return (
+		// This Stepper with this class name
+		// should be refactored into separate shared component.
+		// It is also used in the Setup MC flow.
+		<Stepper
+			className="gla-setup-stepper"
+			currentStep={ step }
+			steps={ getSteps() }
```
asvinb:
```suggestion
			steps={ steps }
```

asvinb (changes requested):

js/src/setup-ads/ads-stepper/index.js
kt-12: This works. Thanks.

js/src/setup-ads/ads-stepper/index.js
kt-12: Initially I was concerned about value not being available when we manually reload the page, as this was the case when I was using `hasGoogleAdsConnection`. But with use of `glaData.initialWpData.adsId` worked totally fine.

js/src/setup-ads/ads-stepper/index.js
kt-12: Fixed.

js/src/setup-ads/ads-stepper/index.js
kt-12: Corrected.

js/src/setup-ads/ads-stepper/index.js
kt-12: This was new to me, but it defiantly did the trick.

asvinb (approved):
Changes LGTM @kt-12

ankitguptaindia:
#
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* WooCommerce - Version 9.3.2
* PHP: 8.3
* Web Server: Nginx
* OS: macOS Sonoma 14.6.1 

</Details>




**Test Results -** Tested the functionally of the new changes. Working as described in scope. ✅ 

https://github.com/user-attachments/assets/5f6ea4de-789d-4d37-968a-cd6c4b924582


**Next Step-** Ready to Code Review(Woo)

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,6 +28,7 @@ import {
  */
 const AdsStepper = ( { formProps } ) => {
 	const [ step, setStep ] = useState( '1' );
+	const hasGoogleAdsId = glaData.initialWpData.adsId;
```
eason9487: The `glaData.initialWpData` is not suitable for this context because its content is only used to initialize the wp-data store and once a Google Ads account is connected, it remains in the state before the connection and doesn't correspond to the actual state. This also makes the code less readable and less maintainable because it implies such runtime state inconsistencies.

To avoid the problem in https://github.com/woocommerce/google-listings-and-ads/pull/2595#discussion_r1766579925, using use something like this can retain the initial state of `hasGoogleAdsConnection`:

```js
const { googleAdsAccount, hasGoogleAdsConnection } = useGoogleAdsAccount();
const initHasAdsConnectionRef = useRef( null );

if ( ! googleAdsAccount ) {
	return null; // or a spinner
}

if ( initHasAdsConnectionRef.current === null ) {
	initHasAdsConnectionRef.current = hasGoogleAdsConnection;
}
```

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -59,51 +61,57 @@ const AdsStepper = ( { formProps } ) => {
...
+	if ( hasGoogleAdsId ) {
+		// Remove first step if there's an Ads account connected.
+		steps.shift();
+
+		steps = steps.map( ( singleStep ) => {
+			return {
+				...singleStep,
+				key: ( parseInt( singleStep.key, 10 ) - 1 ).toString(),
+			};
+		}, [] );
```
eason9487: The callback of `.map()` doesn't use the `[]` as `this` value within its context and it won't work when using an arrow function.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -59,51 +61,57 @@ const AdsStepper = ( { formProps } ) => {
...
+	];
+
+	if ( hasGoogleAdsId ) {
+		// Remove first step if there's an Ads account connected.
+		steps.shift();
+
+		steps = steps.map( ( singleStep ) => {
+			return {
+				...singleStep,
+				key: ( parseInt( singleStep.key, 10 ) - 1 ).toString(),
```
eason9487: How about using the callback's `index` parameter to get the key string?

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -59,51 +61,57 @@ const AdsStepper = ( { formProps } ) => {
...
+		{
+			key: '3',
+			label: __( 'Set up billing', 'google-listings-and-ads' ),
+			content: <SetupBilling formProps={ formProps } />,
+			onClick: handleStepClick,
+		},
+	];
+
+	if ( hasGoogleAdsId ) {
+		// Remove first step if there's an Ads account connected.
```
eason9487: It would be a bit clearer to mention that it's determined by the initial connection state.

eason9487 (changes requested):
In addition to the comments left, there are two other issues:

- The merge base should be the `feature/2460-simplify-paid-ads-setup` branch rather than `develop`.
- The e2e testing failed: https://github.com/woocommerce/google-listings-and-ads/actions/runs/11029983509/job/30633596790

js/src/setup-ads/ads-stepper/index.js
kt-12: It does make sense. I have updated the code to reflect this.

js/src/setup-ads/ads-stepper/index.js
kt-12: Fixed.

js/src/setup-ads/ads-stepper/index.js
kt-12: index was better. Updated.

js/src/setup-ads/ads-stepper/index.js
kt-12: Updated comment to reflect this.

kt-12:
@eason9487 I have addressed your review comment. I have also fixed [E2E](https://github.com/woocommerce/google-listings-and-ads/actions/runs/11102033157/job/30840946684) to reflect the current flow.

joemcgill:
Sorry for the branch name confusion. We've consolidated feature/2460-simplify-paid-ads-setup into feature/2459-campaign-creation-flow. This should be correct now.

ankitguptaindia (approved):
Changes look good and are working as expected. The accounts connection page has been removed from the Paid Ads setup screen. Both the paid ads campaign creation and the edit paid campaign functions are working fine. ✅

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,12 +28,22 @@ import {
...
 		step,
 	} );
 
+	if ( ! googleAdsAccount ) {
+		return null;
+	}
+
+	if ( initHasAdsConnectionRef.current === null ) {
+		initHasAdsConnectionRef.current = hasGoogleAdsConnection;
+	}
```
eason9487: Just noticed this check is not enough to skip the first step. Without the following checks, the Google Ads account claiming and connection with Google Merchant Center account could be skipped.

https://github.com/woocommerce/google-listings-and-ads/blob/bfec77368693e28cd95e7e41ba5d224899b52b4b/js/src/setup-ads/ads-stepper/setup-accounts/index.js#L35-L41

Steps to reproduce:

https://github.com/user-attachments/assets/c7728938-200b-406f-b7ca-94884d6ab8d6

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -215,7 +215,7 @@ test.describe( 'Set up Ads account', () => {
 
 			await setupAdsAccounts.mockAdsStatusClaimed();
 
-			await page.reload();
+			await page.waitForTimeout( 30000 );
```
eason9487: How about using another trick to speed up this test case?

```suggestion
			await page.dispatchEvent( 'body', 'blur' );
			await page.dispatchEvent( 'body', 'focus' );
```

js/src/setup-ads/ads-stepper/index.js
joemcgill: Good catch. We hadn't considered the claim state here.

joemcgill:
@eason9487 I've addressed both concerns and was able to validate that the step remains visible whenever the account needs to be claimed. E2E tests are still passing after applying your suggestion. Can you look again?

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,12 +29,25 @@ import {
  */
 const AdsStepper = ( { formProps } ) => {
 	const [ step, setStep ] = useState( '1' );
+	const { googleAdsAccount, hasGoogleAdsConnection } = useGoogleAdsAccount();
+	const { hasAccess } = useGoogleAdsAccountStatus();
+	const initHasAdsConnectionRef = useRef( null );
+
+	const isGoogleAdsReady = hasGoogleAdsConnection && hasAccess;
```
eason9487: Checking `hasAccess` only may still skip creating the conversion action of Google Ads account and connection to Google Merchant Center account. The `step` in `ads/account-status` API response may be stalled in `'conversion_action'` status.



Steps to reproduce:
1. Select to create new Google Ads account
2. Open the pop-up window for claiming account but don't proceed to claim
3. Close the webpage of Ads-onboarding flow
4. Claim account via the pop-up window
5. Open new webpage to visit the Ads-onboarding flow again
6. The flow starts from the paid campaign setup step

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,12 +29,25 @@ import {
...
 
 	useEventPropertiesFilter( FILTER_ONBOARDING, {
 		context: CONTEXT_ADS_ONBOARDING,
 		step,
 	} );
 
+	if ( ! googleAdsAccount ) {
+		return null;
+	}
```
eason9487: `useGoogleAdsAccount` and `useGoogleAdsAccountStatus` are concurrent asynchronous requests. Only checking `googleAdsAccount` can't guarantee that the required data have been obtained to get the correct value of `initHasAdsConnectionRef.current`.

When `useGoogleAdsAccount` responds first and a Google ads account has been already fully connected, three steps will flash for a moment.

https://github.com/user-attachments/assets/cd74daae-b863-4268-bf91-6b1c6b56ecf2

Also, when a Google Ads account is claimed but not yet fully connected such as https://github.com/woocommerce/google-listings-and-ads/pull/2595#discussion_r1789871915, three steps will flash for a moment as well.

https://github.com/user-attachments/assets/625291cc-3aa0-445a-b6a6-25a3470cb0c6

---

The issues can be simulated by adding delays to the API request of `useGoogleAdsAccountStatus`.

```js
import { __unstableAwaitPromise } from '@wordpress/data-controls';

export function* getGoogleAdsAccountStatus() {
	// Add this at the position needed a delay
	yield __unstableAwaitPromise( new Promise( ( resolve ) => setTimeout( resolve, 3000 ) ) );

	yield fetchGoogleAdsAccountStatus();
}
```

joemcgill (approved):
@kt-12, I've made a few adjustments in [64051b6](https://github.com/woocommerce/google-listings-and-ads/pull/2595/commits/64051b6d7471970e601472ef6949980995ca5655), which fixes some JSLint issues but more importantly checks the `step` value from the `ads/account-status` request rather than the `step` being used in this component to control the `currentStep` of the `Stepper` being rendered.

I think this looks good now, but would like to see if @ankitguptaindia can give this another test and confirm that the issues @eason9487 shared [here](https://github.com/woocommerce/google-listings-and-ads/pull/2595#discussion_r1789871915) can no longer be reproduced when a new account is created and claimed while this screen is closed.

ankitguptaindia:
> I think this looks good now, but would like to see if @ankitguptaindia can give this another test and confirm that the issues @eason9487 shared https://github.com/woocommerce/google-listings-and-ads/pull/2595#discussion_r1789871915 can no longer be reproduced when a new account is created and claimed while this screen is closed.


@joemcgill Tested the reported issue and performed the cases when a new account is created and claimed while this screen is closed. It is working as expected and not able to reproduce the issue now. Tested others that are related to the main changes in this PR.

joemcgill:
This needs a refresh after #2577 was merged to the feature branch

kt-12:
@joemcgill I have handled merge conflict. This is good for WooReview.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,11 +30,45 @@ import {
...
+		step: adsAccountSetupStep,
+	} = useGoogleAdsAccountStatus();
+
 	useEventPropertiesFilter( FILTER_ONBOARDING, {
 		context: CONTEXT_ADS_ONBOARDING,
 		step,
 	} );
 
+	if (
+		! hasResolvedGoogleAccount ||
```
eason9487: Is it necessary to check this condition?

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,11 +30,45 @@ import {
...
+		! hasResolvedGoogleAdsAccount ||
+		! hasResolvedAdsAccountStatus ||
+		googleAdsAccount === null
+	) {
+		return null;
+	}
```
eason9487: Before this PR, it only flashes the loading spinner within the Google Ads Card:

https://github.com/user-attachments/assets/1b91bacb-bfb5-4ebd-aac1-dd06f57eacc9

Checking `hasFinishedResolution` states here leads to almost the entire page flashing momentarily during the account connection.

https://github.com/user-attachments/assets/de8d2727-e0db-464d-8d8b-420438ffe7b8

Could it be kept as close to the original as possible?

ankitguptaindia:
@kt-12 When we click on the ‘Add paid campaign’ button, it takes a few seconds to show campaign creation fields (the page looks blank for a moment). 
Please refer- https://somup.com/cZ62DhHtJN 

_Please let me know if we are fixing this issue on another PR._

js/src/setup-ads/ads-stepper/index.js
kt-12: Fair point, yes we don't need this check. Other check here can handled google account check.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -27,11 +29,42 @@ import {
...
 		step,
 	} );
 
+	if (
+		! hasResolvedGoogleAdsAccount ||
+		! hasResolvedAdsAccountStatus ||
+		googleAdsAccount === null
+	) {
+		return null;
+	}
```
joemcgill: What if we also check the `initHasAdsConnectionRef` here too, so that we only return an empty page if we don't already know that the state of whether the connection page should be shown?

```suggestion
	if (
		initHasAdsConnectionRef.current === null &&
		! ( hasResolvedGoogleAdsAccount && hasResolvedAdsAccountStatus )
	) {
		return null;
	}
```

kt-12: Make sense, I have made this change (slightly different way though ). 


@joemcgill @ankitguptaindia we will still have delays if the Google ads account is not connected. I have replaced the blank screen with a spinner instead for better UX. 

This changes also resolves @eason9487  review finding (https://github.com/woocommerce/google-listings-and-ads/pull/2595#discussion_r1796558994) , see video below.

https://github.com/user-attachments/assets/8bb7983c-9e07-4e07-86c4-93082acf82a6

kt-12:
@joemcgill this is good for review. I have moved final fixes to jest here so with this merging we will also close 2579.

joemcgill (approved):
I've tested several scenarios and the card refresh looks much better now. Thanks, @kt-12

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -26,12 +29,41 @@ import SetupPaidAds from './setup-paid-ads';
...
 
 	useEventPropertiesFilter( FILTER_ONBOARDING, {
 		context: CONTEXT_ADS_ONBOARDING,
 		step,
 	} );
 
+	if ( initHasAdsConnectionRef.current === null ) {
+		if (
+			! ( hasResolvedGoogleAdsAccount && hasResolvedAdsAccountStatus ) ||
+			googleAdsAccount === null // Catch errors retrieving accounts.
```
eason9487: After moving here, it seems no need to check `googleAdsAccount`. Beyond that, `googleAdsAccount` being `null` may also mean that the data has not been retrieved yet, not necessarily that there are errors.

js/src/setup-ads/ads-stepper/index.test.js
eason9487, line 1: Since #2579 has been closed and ideally the jest tests can be run independently, it would be better to make [the second test case](https://github.com/woocommerce/google-listings-and-ads/blob/4bb48b19992bf8795782eaf4e73755e5c1a8f946/js/src/setup-ads/ads-stepper/index.test.js#L55-L74) also pass when it's run with `.only`.

image

Adding `.only` to the following line like `it.only( 'Should record events ....` would make it easier to determine it:

https://github.com/woocommerce/google-listings-and-ads/blob/4bb48b19992bf8795782eaf4e73755e5c1a8f946/js/src/setup-ads/ads-stepper/index.test.js#L55

kt-12: Fixed. Tests are now working independently. 

Screenshot 2024-10-18 at 08 09 00

js/src/setup-ads/ads-stepper/index.js
kt-12: Makes sense. Removed this check here.

js/src/setup-ads/ads-stepper/index.test.js
```diff
@@ -53,6 +57,10 @@ describe( 'AdsStepper', () => {
 
 			render( <AdsStepper /> );
 
+			await waitFor( () => {
+				expect( screen ).toBeDefined();
+			} );
```
eason9487: Related to https://github.com/woocommerce/google-listings-and-ads/pull/2595#discussion_r1805794312.

Before this PR, the `AdsStepper` could render the stepper buttons directly without waiting for data to be loaded. Now it indeed needs to wait for something before it can start testing with the full rendering, but waiting for `screen` is not ideal because the `screen` object is already defined before the testing starts.

Therefore, waiting for an already defined variable to be defined makes the code look questionable. In addition, this test case currently can pass independently just because calling `waitFor` happens to trigger the component to update to full rendering.

These facts can be inspected by:

```js
it.only( 'Should record events after clicking step navigation buttons', async () => {
	// Inspect whether the `screen` is already defined.
	console.debug( screen );

	const user = userEvent.setup();

	render( <AdsStepper /> );

	// Waiting for empty can make it pass as well.
	await waitFor( () => {} );

	const step1 = screen.getByRole( 'button', { name: /accounts/ } );

	await continueToStep2();
	recordEvent.mockClear();
	expect( recordEvent ).toHaveBeenCalledTimes( 0 );

	await user.click( step1 );

	expect( recordEvent ).toHaveBeenCalledTimes( 1 );
} );
```

A possible and more appropriate way to solve this concern is to change to use the [`findBy`](https://testing-library.com/docs/dom-testing-library/api-async/#findby-queries) to query [the button](https://github.com/woocommerce/google-listings-and-ads/blob/95cabc973a9fcc554a260b627ec1f63b1e103771/js/src/setup-ads/ads-stepper/index.test.js#L64) instead of `waitFor` and `getByRole`.

kt-12: Thank you @eason9487. findByRole did the trick.

eason9487 (approved):

PR #2601, opened 2024-09-11
Consolidate accounts onboarding: Create the basic combo card for connecting Google account

ankitrox:
### Changes proposed in this Pull Request:

Closes #2566  .

_Replace this with a good description of your changes & reasoning._

### Screenshots:

Screenshot 2024-09-11 at 5 53 28 PM

### Detailed test instructions:

1. Disconnect all Google accounts if already connected in the plugin.

2. Start the onboarding process again.

3. There should be Google card visible in the first step.

4. The connect button would be disabled if Jetpack is not connected or Terms and conditions are not accepted.

5. Clicking the "Connect" button initiates the current flow to connect a Google account and grant permissions

6. After connecting the Google account successfully, the connected version of Google card should be visible.

### Changelog entry

> Update - Introduce new UI for Google accounts setup during the onboarding.

Issue #2597, opened 2024-09-09
Onboarding: Connect to existing MC account in Google Accounts Card


joemcgill:
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2509

After connecting a Google account, we will provide the UI to allow the merchant to connect to an existing MC account.

<img width="695" alt="image" src="https://github.com/user-attachments/assets/27b40daf-7b07-4c44-8834-25161b2c15a0">

## Acceptance Criteria

- [ ] When the Google account has 1 or more existing MC accounts, show the MC connection UI
- [ ] If only one account exists, it should automatically be selected.
- [ ] If multiple accouns exists, the user should be able to select which to connect.
- [ ] The user can click the "connect" button to connect the selected MC account.
- [ ] If there is no Ads account and one is being created, the card description should only reference Ads (see designs)

## Implementation Brief
* Using the `GoogleComboAccountCard` created in #2566.
* Check if the Google connected account has an Merchant Center account connected.
	* This is done by checking `googleMCAccount.status === connected` via the `useGoogleMCAccount` hook.
* Create `connect-mc/index.js` in the same folder as `GoogleComboAccountCard` which will house the `ConnectMC` component.
* It should follow the logic contained in `js/src/components/google-mc-account-card/connect-mc/index.js` to display the data (not the UI) and styled as per the designs.
	* Use the existing components, for e.g `Section.Card.Body` for the layout.
	* Rendering a select if there are multiple options or text if there's only one option is handled in #2593.
	* Clicking the "Connect" button should follow the same logic as `handleConnectMC` in `js/src/components/google-mc-account-card/connect-mc/index.js`. 
* If there's no connected Google Merchant Center account and the user has at least one MC account, render `ConnectMC` within `GoogleComboAccountCard`.
	* To check if the user has at least a MC account, use the `useExistingGoogleMCAccounts` hook.
* Additionally, if there is no Ads account, `GoogleComboAccountCard` should automatically create one.
  * The logic for this should be handled in #2567.
  * The logic for updating the account creation text "You don’t have Google Ads account..." should be made in the same component that adds the "You don’t have Merchant Center nor Google Ads accounts..." text from #2567.
  * Evaluate whether there are any race conditions when creating/connecting either MC or Ads accounts. In this case, one possible solution would be to disable the MC connection section while the Ads account is being created. This is highly unlikely but something to look for during implementation.

## Test Coverage
[...]

Issue #2596, opened 2024-09-09
Onboarding: Connect to existing Ads account in Google Accounts Card


joemcgill:
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2509

After connecting a Google account, we will provide the UI to allow the merchant to connenct to an existing Ads account.

<img width="697" alt="image" src="https://github.com/user-attachments/assets/eef071cc-c8f2-4dea-89e2-ec6afb527e2e">

## Acceptance Criteria

- [x] When the Google account has 1 or more existing Ads accounts, show the Ads connection UI
- [x] If only one account exists, it should automatically be selected.
- [x] If multiple accouns exists, the user should be able to select which to connect.
- [x] The user can click the "connect" button to connect the selected Ads account.
- [x] If there is no MC account and one is being created, the card description should only reference MC (see designs)

## Implementation Brief
* Using the `GoogleComboAccountCard` created in #2566.
* Check if the Google connected account has an Ads account connected.
	* This should check the `hasGoogleAdsConnection` prop returned by the `useGoogleAdsAccount` hook, but only after the Google account is connected, otherwise that property won't exist.
* Create `connect-ads/index.js` in the same folder as `GoogleComboAccountCard` which will house the `ConnectAds` component.
* It should follow the logic contained in `js/src/components/google-ads-account-card/connect-ads/index.js` to display the data (not the UI) and styled as per the designs.
	* Use the existing components, for e.g `Section.Card.Body` for the layout.
	* Rendering a select if there are multiple options or text if there's only one option is handled in #2593.
	* Clicking the "Connect" button should follow the same logic as `handleConnectClick` in `js/src/components/google-ads-account-card/connect-ads/index.js`. 
* If there's no connected Google Ads account and the user has at least one Ad account, render `ConnectAds` within `GoogleComboAccountCard`.
	* To check if the user has at least an Ad account, use the `useExistingGoogleAdsAccounts` hook.
* Additionally, if there is no Merchant Center account, `GoogleComboAccountCard` should automatically create one.
	* The logic for creating that account should be handled in #2567.
	* Update the description shown in `GoogleComboAccountCard` as per the designs indicating to the user that an account is being created for them.
	* The above should only happen when there is at least an Ad account but no MC account.
* The implementation for the "Or, create a new Google Ads account" button is handled in https://github.com/woocommerce/google-listings-and-ads/issues/2603
[...]

PR #2608, opened 2024-09-12
Consolidate accounts onboarding: Automatically preselect a Google Ads account when there is only one, as well as adjust the UI presentation

### LINKED ISSUE #2593: Update `AppSelectControl` when only one account is available
Related to of https://github.com/woocommerce/google-listings-and-ads/issues/2509

To provide a consistent user experience in all places where a user might be selecting a Google account to connect from a list of existing accounts, we'll update `AppSelectControl ` that is used in both `MerchantCenterSelectControl` and `AdsAccountSelectControl` so that any time there is only one account to select from, that account will immediately shown and the select/dropdown UI will be hidden.

**Screenshot:**
Image

## Acceptance Criteria

- [ ] When either `MerchantCenterSelectControl` or `AdsAccountSelectControl` components are rendered with only one account, no dropdown icon is shown and the account is automatically selected.
- [ ] When either `MerchantCenterSelectControl` or `AdsAccountSelectControl` components are rendered with more than one account, the current `SelectControl` behavior is retained (i.e., users can select the account from a dropdown, and if a value is already set, that value will be selected).

## Implementation Brief

Per [this suggestion](https://github.com/woocommerce/google-listings-and-ads/issues/2592#issuecomment-2337242693), The current `AppSelectControl` component should be updated to be able to show a simplified UI whenever there is only one account to connect. The simplified UI should be controllable by a new `selectSingleValue` prop that defaults to `false`.

The `MerchantCenterSelectControl` and `AdsAccountSelectControl` components will both be updated to pass `selectSingleValue:true`.

The `AdsAccountSelectControl` is currently located in `/js/src/components/google-ads-account-card/connect-ads/ads-account-select-control.js`, as part of the `GoogleAdsAccountCard` component. Since this will be reused, it should be moved to `js/src/components` to be consistent with the location of the `MerchantCenterSelectControl`.

A nice to have would be to make these two controls consistent in that if no accounts are passed to the `AdsAccountSelectControl` component, then it will make use of the `useExistingGoogleAdsAccounts()` hook to automatically populate the dropdown, similar to the way `MerchantCenterSelectControl` handles this [here](https://github.com/woocommerce/google-listings-and-ads/blob/e9b9774d7dff2c72da3a9b577feae95207ea93de/js/src/components/merchant-center-select-control/index.js#L24), as this consistency should make reusing this control in the new `GoogleComboAccountsCard` component easier to implement.

## Test Coverage
- Existing Jest tests for either component should be updated to pass if needed.

## Definition Questions
Is there a reason that the logic between `MerchantCenterSelectControl` and `AdsAccountSelectControl` is not consistent? For example, the MC version handles fetching the existing MC accounts, sorting and setting the initial value ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/e9b9774d7dff2c72da3a9b577feae95207ea93de/js/src/components/merchant-center-select-control/index.js#L24)) while the Ads version requires the account fetching logic to happen by the consuming component.
 - Answered [here](https://github.com/woocommerce/google-listings-and-ads/issues/2593#issuecomment-2337273077)

dsawardekar:

This PR updates the AppSelectControl to show a single selection when only one choice is present.

Closes #2593

### Screenshots:

Multi-select:

image

Single-select:
image

1. Reset your environment, and start a fresh onboarding 
2. On Step 1, connect your Google account
3. In the Merchant and Ads dropdowns you should see a pre-selected option if you have only one choice
4. If you have multiple choices, the dropdown should still show

### Technical decisions:
* To maintain consistency with `MerchantCenterSelectControl`, the `Select one` label has been removed from `AdsAccountSelectControl`: https://github.com/woocommerce/google-listings-and-ads/pull/2608/files#diff-3c83d48212438511a41629a8f6b02ac7666ca664d0c25a734497a1416856d115L17
* The first option within `AdsAccountSelectControl` is set as the default value if none is provided (same behaviour as `MerchantCenterSelectControl`).
* Labels have been updated to `Connect to an existing account`

### Changelog entry

> Update - Automatically preselect a Google Ads account when there is only one, as well as adjust the UI presentation.


dsawardekar:
@joemcgill I have implemented the changes, but this needs a UX revisit due to the way the components are structured. The Label for the AppSelectControl is outside, and the options are only computed in the parent AdsAccountSelectControl, etc. The label will still say 'Select an existing account' even when we have preselected the option.

https://github.com/woocommerce/google-listings-and-ads/blob/update/2593-update-app-select-control-single-select/js/src/components/google-ads-account-card/connect-ads/index.js#L96

cc @asvinb

asvinb:
@dsawardekar @joemcgill In that case maybe we need to rename the label to what we have in the [new designs](https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7293-237837&node-type=frame&t=clG1ewJeSswzRor6-0) which is more generic and it prompts the user to connect an account instead of selecting an account:
> Connect to existing Google Ads account

joemcgill:
> The Label for the AppSelectControl is outside, and the options are only computed in the parent AdsAccountSelectControl, etc. The label will still say 'Select an existing account' even when we have preselected the option.

Thanks for flagging this @dsawardekar. When we use this control in the new `GoogleComboAccountsCard` in #2596 or #2597, we will use a more appropriate label. In the current `GoogleAdsAccountCard` and `GoogleMCAccountCard` components that share these controls, the should still use the current behavior, since we're not setting `selectSingleValue` to `true` by default.

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +16,29 @@ import './index.scss';
  * it will be added to the container div's `className`,
  * so that you can further control its style.
  *
- * @param {*} props
+ * @param {*} props The component props.
+ * @param {boolean} [props.selectSingleValue=false] Whether the select should show only one value.
  */
 const AppSelectControl = ( props ) => {
-	const { className, ...rest } = props;
+	const { className, options, selectSingleValue, ...rest } = props;
```
joemcgill: Need to set the default value for `selectSingleValue` here.
```suggestion
	const { className, options, selectSingleValue = false, ...rest } = props;
```

joemcgill (dismissed):
Small nitpick, but looks good.

joemcgill:
> In the current `GoogleAdsAccountCard` and `GoogleMCAccountCard` components that share these controls, the should still use the current behavior, since we're not setting selectSingleValue to true by default.

After testing this PR, I think I misunderstood the concern. Since this sets `selectSingleValue` to `true` in both the `MerchantCenterSelectControl` and `AdsAccountSelectControl` components, this _will_ affect the current places where those controls are used. We could also add the ability to set the `selectSingleValue` prop on those components to be able to retain the current behavior for the `GoogleAdsAccountCard` and `GoogleMCAccountCard` components, but I think I prefer @asvinb's suggestion that we update those components so the text reads "Connect to an existing account." rather than "Select and existing account".

Example:

image

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +16,34 @@ import './index.scss';
...
+		};
+	}
 
 	return (
-		<div className={ classNames( 'app-select-control', className ) }>
-			<SelectControl { ...rest } />
+		<div
+			className={ classNames( 'app-select-control', className, {
+				'app-select-control--has-single-value': showSingleValue,
+			} ) }
```
ankitrox: It's better to prefix the class names with `gla` or `g4w`

js/src/components/app-select-control/index.scss
ankitrox, line 1: If this file is not being used anywhere else, it would be better to rename it to `app-select-control.scss`

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,53 @@
...
+	const { existingAccounts: accounts = [] } = useExistingGoogleAdsAccounts();
+
+	const options = accounts.map( ( acc ) => ( {
+		value: acc.id,
+		label: sprintf(
+			// translators: 1: account name, 2: account ID.
+			__( '%1$s (%2$s)', 'google-listings-and-ads' ),
+			acc.name,
+			acc.id
+		),
```
ankitrox:
```suggestion
		label: `${ acc.name } (${ acc.id })`,
```

Can we simplify like this?

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,53 @@
...
+	} ) );
+
+	useEffect( () => {
+		// Triggers the onChange event in order to pre-select the initial value
+		if ( value === undefined ) {
```
ankitrox:
```suggestion
		if ( value === undefined && options.length > 0 ) {
```

We may need to handle this as well.

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,53 @@
...
+			onChange( options[ 0 ]?.value );
+		}
+	}, [ options, onChange, value ] );
```
ankitrox:
```suggestion
	}, [ options.length, onChange, value ] );
```

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,53 @@
...
+
```
ankitrox: Should we return null if `options.length` is zero?

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +16,34 @@ import './index.scss';
...
+			readOnly: true,
+		};
+	}
 
 	return (
-		<div className={ classNames( 'app-select-control', className ) }>
-			<SelectControl { ...rest } />
+		<div
+			className={ classNames( 'app-select-control', className, {
+				'app-select-control--has-single-value': showSingleValue,
```
ankitrox:
```suggestion
				'gla_app-select-control--single-value-readonly': showSingleValue,
```

Can we have more descriptive class name like this?

ankitrox (changes requested):
Thanks for the excellent work here. I have tested the functionality and it seems to be working fine. I have added few comments that needs to be addressed or may be needs to looked into.

js/src/components/app-select-control/index.js
asvinb: @ankitrox I see there are other parts where the `app-` prefix it used. For e.g most of the `js/src/components/app-*` components. Let's keep it as is.
Examples: https://github.com/search?q=repo%3Awoocommerce%2Fgoogle-listings-and-ads+.app-+language%3ASCSS&type=code&l=SCSS

js/src/components/app-select-control/index.js
asvinb: @ankitrox See my comment below for the prefix to be used. Also adding `readonly` to the class name does not have any semantic meaning. 👍

js/src/components/ads-account-select-control/index.js
asvinb: No. Consider the situation where the options are still being resolved. If we hide the select, there'll be a big layout shift when the data is resolved.

js/src/components/ads-account-select-control/index.js
asvinb: I think just `options` is fine and the recommended way. Because in any change, the side effect will be triggered because `options` has changed.

js/src/components/ads-account-select-control/index.js
asvinb: @ankitrox on L39, there's already an optional chaining operator to handle it. Also it's consistent to what we already have in `js/src/components/merchant-center-select-control/index.js`

js/src/components/app-select-control/index.scss
asvinb: @ankitrox This is consistent with how the styles live for the other `js/src/components/app-*` components.

js/src/components/ads-account-select-control/index.js
asvinb: Sure thing @ankitrox . Updated!

js/src/components/app-select-control/index.scss
ankitrox: Please check Eason's comment [here](https://github.com/woocommerce/google-listings-and-ads/pull/2601#discussion_r1762502466) which suggests the same thing.

js/src/components/ads-account-select-control/index.js
ankitrox: If our component is only concerned with `options.length`, we should add the same dependency so that we can avoid the re-renders.

js/src/components/app-select-control/index.js
ankitrox: Please check [this comment](https://github.com/woocommerce/google-listings-and-ads/pull/2601#discussion_r1762492941) from Eason.

ankitrox (approved):
@asvinb I have added only couple of comments, but you can decide whether to address those. I am approving this PR as other changes are looking fine to me.

Thanks

js/src/components/app-select-control/index.js
asvinb: I understand what you're saying @ankitrox , but if you look at the current codebase, there are components which have the`gla-*` prefix class names and other specific components com which have the `app-*` prefix. Clearly there is a reason to why this is so. In this particular case, we are editing an existing component which already uses the `app-*` prefix, so I would be reluctant to change it. If you look at https://github.com/woocommerce/google-listings-and-ads/blob/24044c53e4a1e84116086291f02bef8e26af7372/js/src/components/shipping-rate-section/minimum-order-card/minimum-order-card.scss#L3, you'll notice that it's using both, gla and app prefix. I think the base components used within the app are prefixed with `app-*`. @eason9487 Can you shed some light please, about why we have 2 different CSS prefix, i.e `gla-*` and `app-*`? Thanks!

js/src/components/app-select-control/index.scss
asvinb: [Please see my comment](https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1768848945) @ankitrox It's an existing file which we are editing.

js/src/components/ads-account-select-control/index.js
asvinb: @ankitrox Here we need the value of the first option, not only the length.

js/src/components/app-select-control/index.js
eason9487: Early components often came with an `app-` prefix, but that is unnecessary and can be omitted for new components as it becomes a bit verbose.

Regarding the CSS class naming, the main reason is to reduce the chance of CSS naming collisions. `app-` has significantly higher generality, so `gla-` or `g4w-` would be more appropriate. The current `app-` uses are almostly carry-over from earlier times, and newer code would be better to avoid using it as a prefix.

If it involves changing existing code, it is not necessary to modify it. However, if the change (scope) is minor, it would be appreciated if it could be adjusted in the meantime.

ankitguptaindia:
QA Status: ⚠️ Needs feedback
---



>In the Merchant and Ads dropdowns you should see a pre-selected option if you have only one choice
If you have multiple choices, the dropdown should still show


@asvinb 
As per the requirement, if a Google account has only one Ads or Merchant Center (MC) account, it should automatically pre-select that account. However, if multiple accounts exist, the dropdown should display the 'Select One' option by default. 

**Currently, it shows the first Ads and MC account in the dropdown as autoselected even when multiple sub-accounts are present in the Google account.**

Please refer to the screencast below:

https://github.com/user-attachments/assets/14e900ae-e927-437a-934f-ff070c4eaeaf

asvinb:
@ankitguptaindia I added a note in the "Technical decisions":
> To maintain consistency with MerchantCenterSelectControl, the Select one label has been removed from AdsAccountSelectControl: https://github.com/woocommerce/google-listings-and-ads/pull/2608/files#diff-3c83d48212438511a41629a8f6b02ac7666ca664d0c25a734497a1416856d115L17

@joemcgill What do you think? It's mostly making the `MerchantCenterSelectControl` and `AdsAccountSelectControl` consistent and `MerchantCenterSelectControl` did not have the "Select one" option.

joemcgill:
I think making these consistent sounds reasonable to me. So if none are already selected the first one in the list becomes the default, since that’s how the MC field already worked?

asvinb:
Yes @joemcgill that's how the MC field works. @ankitguptaindia Can you confirm please?

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* WooCommerce - Version 9.3.3
* PHP: 8.3
* Web Server: Nginx
* Browser: Chrome - Version 129
* OS: macOS Sonoma 14.6.1 

</Details>




**Test Results** - The PR is functioning as expected.

- ✅ When no Google/MC accounts are available, it prompts to create new accounts:

  image

- ✅ When only one Google/MC account is associated with the Google Account, it should automatically select that account. No drop-down should appear.

image

- ✅ When multiple Google/MC accounts are available, it should select the first account and display a drop-down to choose other accounts.

  image

- ✅ When a Google account has multiple MC accounts but only a single Ad account, it should show a drop-down for MC but none for Ads.

  image


https://github.com/user-attachments/assets/b8e23f67-309f-4fd7-8e61-a68bec071c3e

ankitguptaindia (approved):

js/src/components/app-select-control/index.js
asvinb: Thanks for the clarification @eason9487 . I think that helps a lot. For the scope of this ticket, I suggest we keep it as is. We can have a separate ticket to adjust the class name. 👍

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +16,34 @@ import './index.scss';
  * it will be added to the container div's `className`,
  * so that you can further control its style.
  *
- * @param {*} props
+ * @param {*} props The component props.
+ * @param {boolean} [props.selectSingleValue=false] Whether the select should show only one value.
```
eason9487: Is it possible to move them to this component?

- https://github.com/woocommerce/google-listings-and-ads/blob/f0e12a7f5ed9654928115093311ea1df0cd811de/js/src/components/ads-account-select-control/index.js#L30-L35
- https://github.com/woocommerce/google-listings-and-ads/blob/f0e12a7f5ed9654928115093311ea1df0cd811de/js/src/components/merchant-center-select-control/index.js#L41-L46

Otherwise, the prop name and description could be adjusted because it looks like it starts with a verb but the prop doesn't involve any selection action.

The prop description could be adjusted as well because it in fact changes the presentation and interaction of the UI.

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,47 @@
...
+} ) => {
+	const { existingAccounts: accounts = [] } = useExistingGoogleAdsAccounts();
+
+	const options = accounts.map( ( acc ) => ( {
+		value: acc.id,
+		label: `${ acc.name } (${ acc.id })`,
+	} ) );
+
+	useEffect( () => {
+		// Triggers the onChange event in order to pre-select the initial value
```
eason9487: The code comment says it "pre-select the initial value" which is not the actual behavior of this `useEffect`. Whenever the any of `options`, `onChange` or `value` is changed and the `value` is undefined, it will perform an effect.

Suggest either adjusting this code comment to align with the implementation or vice versa.

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,47 @@
...
+		if ( value === undefined ) {
+			onChange( options[ 0 ]?.value );
+		}
+	}, [ options, onChange, value ] );
```
eason9487: Although the current use of this component won't have issues in the runtime, from the perspective of component design, I'm concerned about its potential issue of infinite rendering loop.

```jsx
function TestComponent() {
	const [ _, setValue ] = useState( 0 );

	return (
		<AdsAccountSelectControl
			onChange={ () => {
				// Update some states that always trigger a top-down re-rendering
				setValue( ( v ) => v + 1 );
			} }
		/>
	);
}
```

eason9487 (commented):
Thanks for the work. Left a few comments about the possibility of simplifying code and improving consistency.

js/src/components/app-select-control/index.js
asvinb: @eason9487 I've moved the logic to the `AppSelectControl` component and also updated the prop name and description. Let me know if that works.

js/src/components/ads-account-select-control/index.js
asvinb: Updated @eason9487

js/src/components/ads-account-select-control/index.js
asvinb: @eason9487 The way we have it, we'll always need to pass on the value prop which should not cause the infinite rendering.

joemcgill:
Just noticed that this PR should be pointing at the `feature/2509-consolidate-google-account-cards` branch. I've update the base branch and resolved conflicts in [dc9b152](https://github.com/woocommerce/google-listings-and-ads/pull/2608/commits/dc9b152f03ff5bc477a1220087061db2b239ad4c)

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +17,54 @@ import './index.scss';
...
+	const hasSingleValueStyle = autoSelectFirstOption && options?.length === 1;
+
+	useEffect( () => {
+		// Triggers the onChange event to set the initial value for the select
+		if (
+			autoSelectFirstOption &&
+			options?.length > 0 &&
+			value === undefined
+		) {
+			onChange( options[ 0 ].value );
```
eason9487, line 43: This may lead to an error:

image

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +17,54 @@ import './index.scss';
...
+		}
+	}, [ autoSelectFirstOption, onChange, options, value ] );
```
eason9487: Continuing with https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1780819380, it still has the possibility to run into infinite rendering loop.

```jsx
function TestComponent() {
	const [ , setValue ] = useState( 0 );
	const options = [
		{ value: undefined, label: 'Please select one' },
		{ value: 'cat', label: 'Cat' },
		{ value: 'dog', label: 'Dog' },
	];

	return (
		// Or <AdsAccountSelectControl
		<AppSelectControl
			options={ options }
			onChange={ () => {
				// Update some states that always trigger a top-down re-rendering
				setValue( ( v ) => v + 1 );
			} }
			autoSelectFirstOption
		/>
	);
}
```

> The way we have it, we'll always need to pass on the value prop which should not cause the infinite rendering.

As I have mentioned in my previous comment, this is a concern about a potential issue from the perspective of component design. How these components should be used is not clearly documented.

- https://github.com/woocommerce/google-listings-and-ads/blob/dc9b152f03ff5bc477a1220087061db2b239ad4c/js/src/components/app-select-control/index.js#L13-L23
- https://github.com/woocommerce/google-listings-and-ads/blob/dc9b152f03ff5bc477a1220087061db2b239ad4c/js/src/components/ads-account-select-control/index.js#L7-L13

The JSDoc of `AdsAccountSelectControl` says its `props.value` can be optional and what happens if it's not defined. The above infinite rendering example doesn't violate the JSDoc specification for the use of props but encounters the serious issue.

To eliminate the issue as well as make it clear how the component is supposed to be used, I think it is possible to converge its implementation and/or documentation toward one of the following directions:
1. When `props.autoSelectFirstOption` is true, it tries to automatically select the first option after mounted. This may mention some requirements and how it triggers `onChange`, for example:
   - `props.value` and `props.onChange` are required and paired as the use of a [controlled-component](https://legacy.reactjs.org/docs/forms.html#controlled-components)
   - Whether to trigger `props.onChange` only when it requires asking the consumer side to update the `value` state of the controlled component
   - After mounted, it only triggers `onChange` once whenever all required props are ready
   - and so on
2. When `props.autoSelectFirstOption` is true, it always tries to automatically select the first option whenever any of the related props are changed.
   - The implementation should avoid the possibility of infinite rendering loop
   - JSDoc indicates whether it works with controlled-component/uncontrolled-component

The current implementation and [JSDoc](https://github.com/woocommerce/google-listings-and-ads/blob/dc9b152f03ff5bc477a1220087061db2b239ad4c/js/src/components/app-select-control/index.js#L21) are close to direction 2 but a [code comment](https://github.com/woocommerce/google-listings-and-ads/blob/dc9b152f03ff5bc477a1220087061db2b239ad4c/js/src/components/app-select-control/index.js#L35) is close to direction 1. However, direction 1 seems to be more suitable in terms of practicality.

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +17,54 @@ import './index.scss';
...
+		options,
+		onChange,
+		value,
+		autoSelectFirstOption = false,
+		...rest
+	} = props;
+	const hasSingleValueStyle = autoSelectFirstOption && options?.length === 1;
+
+	useEffect( () => {
+		// Triggers the onChange event to set the initial value for the select
```
eason9487: This code comment is still not pretty aligned with the actual implementation. Please refer to https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1782282468.

js/src/components/app-select-control/index.js
```diff
@@ -16,14 +17,54 @@ import './index.scss';
  * it will be added to the container div's `className`,
  * so that you can further control its style.
  *
- * @param {*} props
+ * @param {*} props The component props.
+ * @param {boolean} [props.autoSelectFirstOption=false] Whether the automatically select the first option.
```
eason9487: It would be nicer to indicate that when there is only one option, it will also change the UI presentation and will not allow interacting with the `select` control.

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @param {Object} props The component props
+ * @param {string} [props.value] The selected value. IF no value is defined, then the first option is selected and onChange function is triggered.
+ * @param {Function} [props.onChange] Callback when the select value changes.
+ * @return {JSX.Element} An enhanced AppSelectControl component.
+ */
+const AdsAccountSelectControl = ( {
+	value,
+	onChange = () => {},
```
eason9487: There seems no need to destructure these two props as they are just directly forwarded to `AppSelectControl` and whether the `onChange` needs a fallback will depend on the possible changes in https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1782282468.

js/src/components/merchant-center-select-control/index.js
eason9487, line 1: There is the same suggestion as https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1782303692.

https://github.com/woocommerce/google-listings-and-ads/blob/dc9b152f03ff5bc477a1220087061db2b239ad4c/js/src/components/merchant-center-select-control/index.js#L19-L20

js/src/components/app-select-control/index.js
asvinb: @eason9487 Added fallback for `onChange`.

js/src/components/app-select-control/index.js
asvinb: @eason9487 I made the updates, let me know what you think.

js/src/components/app-select-control/index.js
asvinb: I removed the comment in favour of props comment at the top.

js/src/components/app-select-control/index.js
asvinb: 👍

js/src/components/ads-account-select-control/index.js
asvinb: 👍

js/src/components/merchant-center-select-control/index.js
asvinb: 👍

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,28 @@
...
+import useExistingGoogleAdsAccounts from '.~/hooks/useExistingGoogleAdsAccounts';
+
+/**
+ * @param {Object} props The component props
+ * @return {JSX.Element} An enhanced AppSelectControl component.
+ */
+const AdsAccountSelectControl = ( props ) => {
+	const { existingAccounts: accounts = [] } = useExistingGoogleAdsAccounts();
+
+	const options = accounts.map( ( acc ) => ( {
```
eason9487: Before the accounts are fetched, the returned `existingAccounts` is `null` and won't fall back to `[]`, which leads to a "Cannot read properties of null (reading 'map')" error.

Although the current use case won't have the issue because it has a pre-check in [NonConnected](https://github.com/woocommerce/google-listings-and-ads/blob/8c43ac67f7bf2047098eb0de5d3433bf2d5547cb/js/src/components/google-ads-account-card/non-connected.js#L26-L28) plus [ConnectAds](https://github.com/woocommerce/google-listings-and-ads/blob/8c43ac67f7bf2047098eb0de5d3433bf2d5547cb/js/src/components/google-ads-account-card/connect-ads/index.js), considering this component is under the `.~/components` directory, it's expected it could be used without having a pre-check.

js/src/components/merchant-center-select-control/index.js
```diff
@@ -12,15 +11,9 @@ import AppSelectControl from '.~/components/app-select-control';
...
  * @return {JSX.Element} An enhanced AppSelectControl component.
  */
-const MerchantCenterSelectControl = ( {
-	value,
-	onChange = () => {},
-	...props
-} ) => {
+const MerchantCenterSelectControl = ( props ) => {
 	const { data: existingAccounts = [] } = useExistingGoogleMCAccounts();
 	const options = existingAccounts.map( ( acc ) => {
```
eason9487: There is the same issue as https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1787274032 but not introduced by this PR. It would be appreciated if it could be fixed in the meantime.

js/src/components/app-select-control/index.js
```diff
@@ -12,18 +14,72 @@ import './index.scss';
...
- * If you provide `className` via props,
- * it will be added to the container div's `className`,
- * so that you can further control its style.
- *
- * @param {*} props
+ * @param {*} props The component props.
+ * @param {Array} props.options Array of options for the select dropdown. Each option should be an object containing `label` and `value` properties.
+ * @param {string} [props.className] Additional classname to further control the style of the component.
+ * @param {Function} [props.onChange=noop] Callback function triggered when the selected value changes. Receives the new value as an argument.
+ * @param {string} [props.value] The selected value. If no value is defined, the first option is selected and `onChange` is triggered when `autoSelectFirstOption` is true.
```
eason9487: How about adjusting this description to the following?

```suggestion
 * @param {string} [props.value] The currently selected value. This component should be used as a controlled component. A special case is that after mounting, when `autoSelectFirstOption` is true and `value` is undefined, it tries to call back `onChange` once to select the first option so that the `value` can be consistent with the `<select>` element's own value.
```

js/src/components/app-select-control/index.js
```diff
@@ -12,18 +14,72 @@ import './index.scss';
...
+ * @param {boolean} [props.autoSelectFirstOption=false] If true, automatically triggers the onChange callback with the first option as value when no value is provided. If only one option is available, the select is also disabled to prevent user interaction.
```
eason9487: The `select` is not disabled nor rendered in a disabled state. How about adjusting this description to the following?

```suggestion
 * @param {boolean} [props.autoSelectFirstOption=false] If true, automatically triggers the onChange callback with the first option as value when no value is provided. If only one option is available, the select control is also changed to non-interactive.
```

js/src/components/app-select-control/index.js
```diff
@@ -12,18 +14,72 @@ import './index.scss';
...
+
+	useEffect( () => {
+		if (
+			autoSelectFirstOptionRef.current &&
+			optionsRef.current?.length > 0 &&
+			valueRef.current === undefined
+		) {
+			onChangeRef.current( optionsRef.current[ 0 ].value );
+		}
+	}, [] );
```
eason9487: This handling only supports usage where the `options` data is prepared before the first rendering. Usages such as `MerchantCenterSelectControl` and `AdsAccountSelectControl` may not achieve the expected results.

Simulate the issue:
(It needs to resolve the "Cannot read properties of null (reading 'map')" errors in `MerchantCenterSelectControl` and `AdsAccountSelectControl` components first.)

```jsx
function TestComponent() {
	const [ mcId, setMcId ] = useState();
	const [ adId, setAdId ] = useState();

	return (
		<div style={ { width: '300px', margin: '24px' } }>
			<p>Merchant onChange value: { mcId ?? 'undefined' }</p>
			<MerchantCenterSelectControl value={ mcId } onChange={ setMcId } />
			<p>Ads onChange value: { adId ?? 'undefined' }</p>
			<AdsAccountSelectControl value={ adId } onChange={ setAdId } />
		</div>
	);
}
```

Although the `<select>` elements selected the first option, the value states in the consumer component are still `undefined`.

image

eason9487: Using one `useRef` should fulfill the needs without the mentioned issues.

For example:

```js
const shouldAutoSelectOnceRef = useRef( autoSelectFirstOption === true );

useEffect( () => {
	if ( ! shouldAutoSelectOnceRef.current ) {
		return;
	}
	if ( value === undefined && options.length ) {
		shouldAutoSelectOnceRef.current = false;
		onChange( options[ 0 ].value );
	}
}, [ value, options, onChange ] );
```

eason9487 (commented):
Found two new issues and it would be better to address them to avoid potential issues in the future.

js/src/components/ads-account-select-control/index.js
asvinb: 👍 Updated @eason9487

js/src/components/merchant-center-select-control/index.js
asvinb: 👍

js/src/components/app-select-control/index.js
asvinb: 👍

js/src/components/app-select-control/index.js
asvinb: Thanks for that. I tried something like that before and for whatever reason I was getting an infinite loop. Thanks again!

js/src/components/ads-account-select-control/index.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+import AppSelectControl from '.~/components/app-select-control';
+import useExistingGoogleAdsAccounts from '.~/hooks/useExistingGoogleAdsAccounts';
+
+/**
+ * @param {Object} props The component props
+ * @return {JSX.Element} An enhanced AppSelectControl component.
+ */
+const AdsAccountSelectControl = ( props ) => {
+	const { existingAccounts: accounts = [] } = useExistingGoogleAdsAccounts();
```
eason9487: In https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1787274032:
> Before the accounts are fetched, the returned `existingAccounts` is `null` and won't fall back to `[]`, [...]

The ineffective fallback can be removed to minimize unnecessary logic branches.

js/src/components/merchant-center-select-control/index.js
```diff
@@ -12,17 +11,11 @@ import AppSelectControl from '.~/components/app-select-control';
...
- * @param {Function} [props.onChange] Callback when the select value changes.
  * @return {JSX.Element} An enhanced AppSelectControl component.
  */
-const MerchantCenterSelectControl = ( {
-	value,
-	onChange = () => {},
-	...props
-} ) => {
+const MerchantCenterSelectControl = ( props ) => {
 	const { data: existingAccounts = [] } = useExistingGoogleMCAccounts();
```
eason9487: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2608#discussion_r1789750320

js/src/components/app-select-control/index.js
```diff
@@ -12,18 +14,60 @@ import './index.scss';
 /**
  * Renders a `@wordpress/components`'s `SelectControl` with margin-bottom removed.
  *
- * If you provide `className` via props,
- * it will be added to the container div's `className`,
- * so that you can further control its style.
- *
- * @param {*} props
+ * @param {*} props The component props.
+ * @param {Array} props.options Array of options for the select dropdown. Each option should be an object containing `label` and `value` properties.
```
eason9487: The JSDoc of `props.options` was not updated accordingly.

Ref: https://github.com/woocommerce/google-listings-and-ads/pull/2608/commits/08b16e067128362d3053e50fb1e8c7c7ebf9f204#diff-52e887fd2c4ad94dd408eb3141de68f1a16cd7c24e7ab6122f62bb7b52747844R27

js/src/components/ads-account-select-control/index.js
asvinb: 👍

js/src/components/merchant-center-select-control/index.js
asvinb: 👍

js/src/components/app-select-control/index.js
asvinb: Updated @eason9487

eason9487 (approved):

PR #2609, opened 2024-09-16
Update Add/Edit Shipping Time Modals to support minimum and maximum shipping times stepper

jorgemd24:

Part of https://github.com/woocommerce/google-listings-and-ads/pull/2594

Since we're now working with min and max shipping times, the "Edit" and "Add" modals need to be updated to use the stepper component when creating or editing shipping times. This PR introduces a new component, `MinMaxShippingTimes`, which includes steppers for both the minimum and maximum shipping times.

image

image

The errors in the modals are displayed within the form instead of beneath the stepper component to prevent issues like this:

image

1. Follow the steps mentioned in this PR to update the database: https://github.com/woocommerce/google-listings-and-ads/pull/2520
2. Navigate to Dashboard -> Free Listings -> Edit
3. Add and edit shipping times using the modals, and verify that the min and max times are updated correctly.

### Additional details:

In this commit [8652d24](https://github.com/woocommerce/google-listings-and-ads/pull/2609/commits/8652d2421db60f69107ae69c8c6355440493dbfa) I updated the `onBlur` function to align with `onIncrement`. However, I'm unsure what's best—whether we should allow users to enter any value and display errors in the form, or restrict input to values within the allowed range, like preventing negative numbers. Any advice would be appreciated.

### Changelog entry

>


martynmjones (commented):
Hey @jorgemd24, thanks for implementing the stepper for shipping.

All is working properly although I noticed a few useability issues which we may want to address before releasing.

1. The modal size changes when an error is displayed or hidden. If a merchant is using the `+` and `-` to set the value then the content jumps and the mouse may move off the button when the error status changes.

https://github.com/user-attachments/assets/c5047242-94fe-4c11-b117-f421739d7e68

2. Invalid fields aren't highlighted when attempting to save changes. If a merchant is shipping to a number of countries and has different settings for each then it would be quite difficult to spot where the issue is.

Screenshot 2024-09-19 at 11 22 05

3. Locations are immediately grouped if the min/max values match an existing field. at the moment merchants would need to enter an identical value and then focus on a different element before they are grouped.

    As it happens when using the steppers we potentially have a situation where we want to use the `+` button to get to `4` but it's grouped before that can happen.

    What do you think of adding a small delay before grouping so that it would be possible to use the stepper to get to a higher number?

https://github.com/user-attachments/assets/75c6a3fa-8c69-4932-9fe8-986b366f45bd

jorgemd24:
Thanks @martynmjones for your helpful comments! I’ve addressed points 1 and 3.

Regards the second one: 

> Invalid fields aren't highlighted when attempting to save changes. If a merchant is shipping to a number of countries and has different settings for each then it would be quite difficult to spot where the issue is.

I agree that it's not easy to see which one is incorrect, but this behaviour hasn't changed compared to the one in the `develop` branch. 

image

I believe the issue is that we're treating `shipping_times` as a group of inputs rather than individual inputs, which makes it harder to display errors for each input separately. I'll create an issue for this since I think it goes beyond the scope of this PR.

martynmjones (approved):
Many thanks for making those adjustments @jorgemd24! Tested the updated version and both the issues I mentioned previously have been resolved. 

> I believe the issue is that we're treating shipping_times as a group of inputs rather than individual inputs, which makes it harder to display errors for each input separately. I'll create an issue for this since I think it goes beyond the scope of this PR.

Perfect! Thank you.

PR #2621, opened 2024-09-20
Prioritise core WooCommerce GTIN field in product adapter

### LINKED ISSUE #2614: [GTIN] Prioritise core WooCommerce GTIN field when syncing
WooCommerce 9.2 introduced a global unique ID field to products that will replace the `GTIN` field included in Google for WooCommerce. To support the existing data being migrated to the core field we will first need to update our product adapter to prioritise the core field and then fallback to existing Google for WooCommerce data.

Screenshot 2024-09-17 at 20 33 45

## Acceptance criteria
GTIN is automatically populated when syncing products using the following priority:

1. WC Core field
1. Mapping Rule
1. Fallback to GLA value

## Additional

pcTzPl-2p2-p2

martynmjones:

Closes #2614

Adjusts `WCProductAdapter.php` to map the GTIN field from WooCommerce core if it's available. This is checked after the attribute mapping rules and Google for WooCommerce attributes as we want to prioritise the core field.

We return early if the product does not have a `get_global_unique_id` method to maintain backwards compatibility.

### Screenshots:

Screenshot 2024-09-20 at 12 20 42

1. Checkout `update/2614-prioritise-core-gtin-field-in-product-adapter`
2. Set up an attribute mapping rule for GTIN to some field
3. Create a product and enter different valid GTIN values (_for example 00000001, 00000002, 00000003_) in
    - The core `GTIN` field on the `Inventory` tab
    - The `GTIN` field on the `Google for WooCommerce` tab
    - The field set via the attribute mapping rule created in step 2
4. Go to `wp-admin/admin.php?page=connection-test-admin-page` and sync the product to Google
5. View the product in Merchant Center and confirm the GTIN number matches the value entered in the `Inventory` tab
6. Delete the GTIN value from the `Inventory` tab
7. Re-sync the product and then confirm the value in Merchant Center now matches the value from the `Google for WooCommerce` tab
8. Delete the GTIN value from the `Google for WooCommerce` tab
9. Re-sync the product and then confirm the value in Merchant Center now matches the value from the attribute mapping rule

### Changelog entry

> Update - Product adapter to map GTIN value from WooCommerce core field if it's available

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -109,6 +109,7 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 		$this->map_woocommerce_product();
 		$this->map_attribute_mapping_rules( $mapping_rules );
 		$this->map_gla_attributes( $gla_attributes );
+		$this->map_gtin();
 
 		// Allow users to override the product's attributes using a WordPress filter.
 		$this->override_attributes();
@@ -922,6 +923,28 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 		// Size
 		if ( ! empty( $attributes['size'] ) ) {
 			$this->setSizes( [ $attributes['size'] ] );
+		}
+
+		return $this;
+	}
+
+	/**
+	 * Map the WooCommerce core global unique ID (GTIN) value if it's available.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return $this
+	 */
+	protected function map_gtin(): WCProductAdapter {
+		// compatibility-code "WC < 9.2" -- Core global unique ID field was added in 9.2
+		if ( ! method_exists( $this->wc_product, 'get_global_unique_id' ) ) {
+			return $this;
+		}
+
+		$global_unique_id = $this->wc_product->get_global_unique_id();
+
+		if ( ! empty( $global_unique_id ) ) {
+			$this->setGtin( $global_unique_id );
 		}
 
 		return $this;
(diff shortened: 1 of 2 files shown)


src/Product/WCProductAdapter.php
```diff
@@ -927,6 +928,28 @@ protected function map_gla_attributes( array $attributes ): WCProductAdapter {
 		return $this;
 	}
 
+	/**
+	 * Map the WooCommerce core global unique ID (GTIN) value if it's available.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return $this
```
puntope, line 936: I believe this should be `WCProductAdapter` ?

src/Product/WCProductAdapter.php
```diff
@@ -109,6 +109,7 @@ public function mapTypes( $properties ) {
 		$this->map_woocommerce_product();
 		$this->map_attribute_mapping_rules( $mapping_rules );
 		$this->map_gla_attributes( $gla_attributes );
+		$this->map_gtin();
```
puntope, line 112: Reading this, I wanted to point out that it won't be allowed to override the GTIN via Google for WooCommerce tab or Attribute Mapping rules. Just wanted to be sure that this is expected.

puntope (approved):
Just a little comment regarding docs. But approved in advance

Screenshot 2024-09-23 at 16 59 12
Screenshot 2024-09-23 at 16 58 56

puntope:
p.s I also tested with a non valid GTIN 

Screenshot 2024-09-23 at 17 06 09

src/Product/WCProductAdapter.php
martynmjones: Throughout Google for WooCommerce we're documenting returns as `$this` in a few places and other methods in this class do the same. So while I agree and normally would have gone with `WCProductAdapter` as it was only a small change being made I opted to go for consistency.

PR #2622, opened 2024-09-20
Hide `GTIN` field for new installs make readonly for existing installs

### LINKED ISSUE #2615: [GTIN] Hide the Global Trade Item Number field for new users and make read-only for existing users
As we will be prioritising the global unique ID field in core (See #2614) when syncing product data, the `GTIN` field in Google for WooCommerce will need to be hidden for new users of the extension and made read-only for existing users. The tooltip will also need to be updated to direct existing users to the core `Inventory` tab to popular the field.

Screenshot 2024-09-17 at 20 54 30

## Acceptance criteria

1. For new merchants, the `GTIN` field is automatically hidden in both the classic and new product editor
2. For existing merchants, the `GTIN` field is made read-only and the tooltip directs them to the `Inventory` tab

## Additional

pcTzPl-2p2-p2

martynmjones:

Adds restrictions to the `GTIN` field in Google for WooCommerce depending on the version of WooCommerce installed and what version of Google for WooCommerce was first installed.

The following conditions apply:

1. If WooCommerce `9.2.0` or higher is installed and Google for WooCommerce is installed for the first time the `GTIN` field in G4W is hidden
2. If WooCommerce `9.2.0` or higher is installed and Google for WooCommerce had already been installed on a website then the `GTIN` field in G4W is read-only
3. If any version of WooCommerce lower than `9.2.0` is installed then the `GTIN` field in G4W continues to function as it previously did

Additionally, a new option is added called `INSTALL_VERSION` for tracking what version of Google for WooCommerce was first installed on a website 43a76c3.

Closes #2615

### Screenshots:

Screenshot 2024-09-20 at 17 46 09

1. Checkout `update/2615-hide-gtin-for-new-users-make-readonly-for-exisitng` on a site with WC `9.1.0` installed that has already onboarded with Google for WooCommerce
4. Add or edit a product and confirm that the `GTIN` field is available on the `Google for WooCommerce` tab
5. Update WooCommerce to version `9.2.0` or higher
6. Add or edit a product and confirm that the `GTIN` field is read-only and the tooltip directs merchants to the `Inventory` tab
7. Install the extension from this branch on a new test site and go through onboarding
    - Alternatively, run the following WP Cli command to simulate the first install
    `wp option update gla_install_version 2.8.4`
8. Confirm that the `GTIN` field is not displayed on the `Google for WooCommerce` tab when editing a product

### Changelog entry

> Update - Restrict the GTIN field based on the version of WooCommerce installed and the initial version of G4W installed


src/Admin/Input/Input.php
```diff
@@ -142,6 +158,34 @@ public function set_value( $value ): InputInterface {
...
+	public function set_disabled( $value = false ) {
+		$this->is_disabled = $value;
+
+		return $this;
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function is_disabled(): bool {
```
puntope, line 185: WDYT of adding `get_` prefix here?

src/Admin/Input/Input.php
```diff
@@ -142,6 +158,34 @@ public function set_value( $value ): InputInterface {
...
+	/**
+	 * @param bool $value
+	 *
+	 * @return InputInterface
+	 */
+	public function set_disabled( $value = false ) {
+		$this->is_disabled = $value;
+
+		return $this;
+	}
```
puntope: WDYT of adding the types in the arguments and signature fore these 2?

public function set_readonly(bool $value = false ): InputInterface

puntope: Another concern I have is that `disabled` is the right name to use for a hidden field. As in HTML disabled inputs are still readable. So I would suggest `is_hidden` / `set_hidden` but not a blocker tho, just following the naming for PR name and  issue description:

> ...  GTIN field in Google for WooCommerce will need to be hidden for new users of the extension and made read-only for existing users...

src/Admin/Product/Attributes/Input/GTINInput.php
```diff
@@ -16,6 +17,13 @@
  */
 class GTINInput extends Text {
 
+	/**
+	 * Initial install version to disable the GTIN field for.
+	 *
+	 * @var string
+	 */
+	private $hidden_from = '2.8.5';
```
puntope: COuld we add the keyword `version` somewhere in this field, also should this be a constant instead?

src/HelperTraits/Utilities.php
```diff
@@ -68,6 +68,22 @@ protected function gla_setup_for( $seconds ): bool {
...
+	 */
+	protected function gla_installed_after( $date ): bool {
+		$gla_installed = $this->options->get( OptionsInterface::INSTALL_TIMESTAMP, false );
+
+		if ( false === $gla_installed ) {
+			return false;
+		}
+
+		return ( $gla_installed >= strtotime( $date ) );
+	}
```
puntope: Where are we using this?

src/Admin/Input/Input.php
```diff
@@ -60,6 +73,9 @@ class Input extends Form implements InputInterface {
 	public function __construct( string $type, string $block_name ) {
 		$this->type       = $type;
 		$this->block_name = $block_name;
+
+		$this->set_options_object( woogle_get_container()->get( OptionsInterface::class ) );
```
puntope: Do we need to instantiate the container in all the inputs? Maybe we can just do this in GTIN?

src/Admin/Input/Input.php
```diff
@@ -51,6 +51,16 @@ class Input extends Form implements InputInterface {
 	 */
 	protected $value;
 
+	/**
+	 * @var bool
+	 */
+	protected $is_readonly = false;
```
mikkamp, line 57: Not really necessary since we don't do it in most other classes. But in theory we support PHP 7.4+ which means we can declare type for class properties. Just looks odd to set the type only in the doc-block and not actually restrict it in the class.

src/Admin/Input/Input.php
```diff
@@ -142,6 +152,34 @@ public function set_value( $value ): InputInterface {
 		return $this;
 	}
 
+	/**
+	 * @param bool $value
+	 *
+	 * @return InputInterface
+	 */
+	public function set_readonly( $value ): InputInterface {
```
mikkamp: Similar comment here, if we are stating in the doc-block that `$value` should be a bool why don't we restrict it by setting the type? That's supported well before PHP 7.4

src/Admin/Product/Attributes/Input/GTINInput.php
```diff
@@ -24,5 +35,28 @@ public function __construct() {
 
 		$this->set_label( __( 'Global Trade Item Number (GTIN)', 'google-listings-and-ads' ) );
 		$this->set_description( __( 'Global Trade Item Number (GTIN) for your item. These identifiers include UPC (in North America), EAN (in Europe), JAN (in Japan), and ISBN (for books)', 'google-listings-and-ads' ) );
+		$this->set_options_object( woogle_get_container()->get( OptionsInterface::class ) );
```
mikkamp: This line is a little confusing to me. In other classes we use the `OptionsAwareTrait` which will automatically set the Options object within the class so we don't need to set it for each class individually. However for an input field it's not shared within the container class so it doesn't automatically set this object.

Since this class differs and it only needs to access options once, my suggestion would be to not use the `OptionsAwareTrait` and get the options object only once when needed. For example:
```php
		$options         = woogle_get_container()->get( OptionsInterface::class );
		$initial_version = $options->get( OptionsInterface::INSTALL_VERSION, false );
```

It would make it harder to mock, but I don't see us testing with a mocked options object for this class.

mikkamp (approved):
Thanks for all the changes here. I have done the testing and I can confirm that the field is shown as normal for WC < 9.2. For newer versions of WC it's shown as a read only field:
image

And with the install later than 2.8.5 it's completely removed:
image

Should we keep a note that we need to tweak that version, I notice it's already been bumped from 2.8.4 > 2.8.5 but by the time we release this we might be at a newer version.

I went through the code and added a few opinionated comments, but these aren't strictly necessary to merge this PR, so I've gone ahead and approved anyway.

puntope:
> Should we keep a note that we need to tweak that version, I notice it's already been bumped from 2.8.4 > 2.8.5 but by the time we release this we might be at a newer version.

Correct @mikkamp Thanks for catching that. Added here https://github.com/woocommerce/google-listings-and-ads/issues/2648

src/Admin/Input/Input.php
puntope: The reason I didn't do it is because the rest of the files (and other files like this one) have not the type set neither. So I bet for consistency in regards this topic.

src/Admin/Input/Input.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2622/commits/fa973aa57ac7587350543cfcc2e094d2ab95bda1

src/Admin/Product/Attributes/Input/GTINInput.php
puntope: Adjusted here https://github.com/woocommerce/google-listings-and-ads/pull/2622/commits/a2c365411adea67c72d8b0c274631318b090b3c5

PR #2575, opened 2024-09-03
Consolidate Ad Creation

asvinb:
### Changes proposed in this Pull Request:

Closes #2535 

_Replace this with a good description of your changes & reasoning._

### Detailed test instructions:

1. 
2. 
3. 

### Changelog entry

>

PR #2531, opened 2024-08-15
Streamline onboarding: Move FAQs to the bottom of pages

asvinb:
### Changes proposed in this Pull Request:

Closes #2488

- Add new `StepContentActions` component which replaces `StepContentFooter`.
- Add a new `StepContentFooter` component which uses the existing `Section` and `VerticalGapLayout` components.
- Added the new `xlarge` size to `VerticalGapLayout` that uses the `--large-gap` CSS property for spacing.

### Screenshots:

**Onboarding flow:**
- image
- image

**Editing a paid campaign**
image

**Optimizing your campaign**
image

### Detailed test instructions:

* Go through the onboarding flow.
* The FAQ section should be after the CTA for the following pages:
	* During the onboarding process, the first step ("Set up your accounts") and fourth step ("Create a campaign")
	* When editing a paid campaign (Go to the dashboard and "Edit" any paid campaign which has been previously created)
	* When optimizing a paid campaign.
		* Click on the "Add paid campaign" from the dashboard.
		* Fill in the form in the first step and go to the next one (Optimize your campaign) to see the FAQ.
* The spacing between the CTA and the FAQ panel should be 40px which is the value of the `--large-gap` CSS property as per the AC.

### Additional details:
On the first step of the onboarding process, there is extra space between the CTA and the section above because of the long text as per the screenshot:
image

### Changelog entry

> Tweak - Move FAQs to the bottom of pages.

PR #2623, opened 2024-09-24
Streamline campaign setup: Consolidate the campaign setup UI in the onboarding with the one in the Ads-onboarding and add the billing setup back

### LINKED ISSUE #2535: Consolidate the ad creation step in the Ads Setup flow with the one used in Onboarding

asvinb:

Closes #2535  .

1. Go through the onboarding process and ensure a paid campaign can be created.
	* When billing is not set up, the "Launch Paid Campaign" button should be disabled.
2. Ensure the skip paid ads modal is displayed when the user clicks the skip button during the onboarding flow.
3. Go through the setup ads flow (/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-ads) and ensure a paid campaign can be created if billing is setup.
4. Ensure a paid campaign can be added via the "Add a paid campaign" button on the dashboard.
5. Ensure a paid campaign can be edited via the "Edit" links on the dashboard.
	* The budget recommendation should be for the selected countries for the paid campaign.
6. Test when billing is not setup flow, ie. the user is prompted to setup billing.
7. Users should not be able to submit/continue any flow if billing is not set up and approved.
8. Users are no longer able to view the targeted countries.

Extra issues to test:
* https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1794919412
* https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1794924425

### Changelog entry

> Update - Consolidate the campaign setup UI in the onboarding flow with the one in the Ads-onboarding flow.
> Update - Merge the billing setup into the campaign setup for the Ads-onboarding flow.


asvinb:
@joemcgill Can you kindly review PR please?
I've merged the campaign creation flow branch in this one and set it as the target branch where it'll simplify a lot of our work in the campaign creation flow tasks.
The other PR: https://github.com/woocommerce/google-listings-and-ads/pull/2575 should no longer be relevant now.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,182 @@
...
+			eventName: 'gla_onboarding_complete_with_paid_ads_button_click',
+			eventProps: {
+				budget: paidAds.amount,
+				audiences: countryCodes?.join( ',' ),
+			},
+		};
+	}
+
+	const description =
+		headerDescription ||
```
joemcgill: `headerDescription` seems to be a required property, should it default to `null` or `''` to allow it to be empty? What if someone intentionally passes a falsey value to that prop? Won't this fallback kick in? Similarly, we don't seem to handle any default fallback for the `headerTitle` prop here. Is that intentional?

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,182 @@
...
+				onStatesReceived={ handleOnStatesReceived }
+				campaign={ campaign }
+				countryCodes={ countryCodes }
+				showCampaignPreviewCard={
+					trackingContext === 'setup-ads' ||
+					trackingContext === 'create-ads'
+				}
+			/>
+
+			<FaqsSection />
```
joemcgill: This `FaqsSection` component still doesn't include the improvements from #[#2531](https://github.com/woocommerce/google-listings-and-ads/pull/2531). Can we cherry-pick those fixes into this branch as well and make sure this doesn't cause a regression when all is finalized?

js/src/components/paid-ads/billing-card/billing-card.js
```diff
@@ -15,20 +16,32 @@ import fallbackBillingUrl from './fallbackBillingUrl';
...
+		if ( billingStatus.status === PENDING && ! showNotice ) {
+			setShowNotice( true );
+		}
+	}, [ billingStatus, showNotice ] );
+
 	if ( ! hasFinishedResolution ) {
 		return <SpinnerCard />;
 	}
 
+	if ( billingStatus.status === APPROVED && ! showNotice ) {
```
joemcgill: Nit: This could be combined with the if below and use a tertiary to show the success component or `null` based on the value of `showNotice`, which could improve readability and make the relationship between these two logic branches more obvious. Not a blocker.

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
```diff
@@ -225,7 +226,15 @@ const SavedSetupStepper = ( { savedStep } ) => {
 				{
 					key: stepNameKeyMap.paid_ads,
 					label: __( 'Create a campaign', 'google-listings-and-ads' ),
-					content: <SetupPaidAds />,
+					content: (
+						<CampaignAssetsForm
```
joemcgill: Curious if the `SetupPaidAds` component (setup-mc/setup-paid-ads) could include the `CampaignAssetsForm` so this change wouldn't be needed?

joemcgill (changes requested):
This is looking really good so far. I've left some inline feedback. 

Additionally, I noticed that with these changes that when I create a new campaign from after editing an existing campaign that the initial value gets filled from the session storage value in `gla-onboarding-paid-ads`.

Screenshot 2024-09-26 at 10 50 57 AM


Path: `page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Fcampaigns%2Fcreate`

This may already get handled in the new minimum budget work we have in progress, but ideally session storage would not be a factor when creating a campaign other than your first one.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: @joemcgill I noticed the changes were merged to the [feature/2458-streamline-onboarding](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2458-streamline-onboarding) branch. I only merged the campaign creation one. If we want to have those changes, then it's better we merge the feature/2458-streamline-onboarding branch?

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: @joemcgill I removed the `headerDescription` prop and instead conditionally set the description based on the `isOnboardingFlow` prop. Because it's only during the onboarding flow that the description is different.

js/src/components/paid-ads/billing-card/billing-card.js
asvinb: Nice. Updated!

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
asvinb: I didn't think of that 🤣 . Definitely that works!

js/src/components/paid-ads/ads-campaign/ads-campaign.js
joemcgill: The way I originally expected us to sequence this is for all of the 2458 work to be merged to develop and then pulled back into this branch to resolve merge conflicts before merging this work as well. However, we want to hold all of this work so it ships together. If we already had PRs set up for these feature branches, I think it would look like:

* feature/2458-streamline-onboarding > develop (not merged until everything is ready)
  * feature/2590-consolidate-google-account-cards > feature/2458
  * feature/2459-campaign-creation > feature/2458
    * feature/2460-simplify-paid-ads-setup > feature/2459 (these are getting combined)
* feature/2460-google-ads-value-prop (post onboarding) > develop (refreshed after 2458 is merged)

So I think there are two ways of managing this, we could either merge feature/2458 into  feature/2459 and to update/2535 now, or wait and merge feature/2458 into feature/2459 after the feature branch is ready and resolve merge conflicts at that point. Even so, since this moves and replaces some of the changed files, we'll have to be careful about applying those changes here during that merge. I think either strategy is fine.

asvinb: @joemcgill 
The best way forward would be to merge #2458 into #2459 and then we merge everything in 2535.
I have a draft PR set up for it:
https://github.com/woocommerce/google-listings-and-ads/pull/2630

I have another PR ready to merge the updated 2459 into 2535 which we can review afterwards:
https://github.com/woocommerce/google-listings-and-ads/pull/2627

So right now, I think if we can review and merge https://github.com/woocommerce/google-listings-and-ads/pull/2630 and then the campaign creation branch would be the target branch for all future PRs.
Let me know what you think.

asvinb:
@joemcgill I've made some updates where we are loading the amount from client session only during onboarding. Can you take a look and let me know what you think.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
joemcgill: This sounds right to me. To make things a bit clearer, I'll create draft PRs for the feature branches because 2590 and 2459 should both me opened against 2458, that way we'll be able to see the conflicts that will need to be resolved in an upstream branch.

joemcgill: @asvinb I believe I've gotten all the feature branches updated and set up so that this should be ready to finish up. I've created the following draft PRs for the feature branches:

- #2631
- #2632
- #2634

In doing so, I realized that we somehow had a duplicate feature branch for #2509 with a typo in the name (feature/2590... instead of feature/2509...) so I've updating all existing PRs to use the correct feature branch and deleted feature/2590...

I've merged all of the upstream changes from your merge conflict PRs here as well and fixed up an additional merge conflict that remained, so we should be in good shape now, but let me know if you have any issues.

js/src/components/paid-ads/ads-campaign/paid-ads-setup-sections.js
```diff
@@ -54,22 +61,35 @@ function resolveInitialPaidAds( paidAds ) {
...
  * @param {Array<CountryCode>|undefined} props.countryCodes Country codes for the campaign.
+ * @param {Campaign} [props.campaign] Campaign data to be edited. If not provided, this component will show campaign creation UI.
+ * @param {boolean} [props.showCampaignPreviewCard=false] Whether to show the campaign preview card.
+ * @param {boolean} [props.loadCampaignFromClientSession=false] Whether to load the campaign data from the client session.
  */
 export default function PaidAdsSetupSections( {
 	onStatesReceived,
 	countryCodes,
+	campaign,
+	loadCampaignFromClientSession,
```
joemcgill: I think this is a clever way of handling this. If adding a prop for controlling this ends up being undesirable, an alternate option might be to clear the clientSession data whenever the a campaign is created, and update this component to only save the clientSession data if `isCreation` is true, but I don't see this as a blocker.

joemcgill (approved):
I think these updates are looking pretty good and are testing well for me locally. Left one suggestion, but marking as approved and ready for another QA pass.

One note for QA that should be kept in mind is that when this PR is tested with billing not approved, you'll see the billing setup card on this form as well as a duplicate billing step when creating your first campaigns from the dashboard, until #2536 is completed.

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* WooCommerce - Version 9.3.3
* PHP: 8.3
* Web Server: Nginx
* OS: macOS Sonoma 14.6.1 

</Details>





**Test Results -** Followed the testing instructions, acceptance criteria, and tested all possible ways to create, edit, and manage paid ad campaigns. All tests were passed when billing was not set, with billing set, and other use cases. 


**Next Step-** Ready to Code Review(Woo)

ankitguptaindia (approved):

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,189 @@
...
+							paidAds={ paidAds }
+						/>
+					) }
+
+					<AppButton
+						isPrimary
+						disabled={ disabledComplete }
+						onClick={ handleCompleteClick }
+						loading={ completing === ACTION_COMPLETE }
+						{ ...continueButtonProps }
```
eason9487: The suggestion is the same as https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1780557466 and https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1789658751. Otherwise, this button would become more complicated after merging #2577.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,189 @@
...
+						<SkipButton
+							text={ __(
+								'Skip paid ads creation',
+								'google-listings-and-ads'
+							) }
+							onSkipCreatePaidAds={ handleSkipCreateAds }
+							completing={ completing }
+							paidAds={ paidAds }
+						/>
+					) }
```
eason9487: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1791369972. Suggest externalizing this button as a `skipButton` prop or render prop. In addition, the entire [`SkipButton` component](https://github.com/woocommerce/google-listings-and-ads/blob/34c53d7d4d37fbfdf46b00aa36195c2e09710eb4/js/src/components/paid-ads/ads-campaign/skip-button.js) could be kept inline in the  `SetupPaidAds` component such as the implementation before this PR.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,189 @@
...
+			<PaidAdsSetupSections
+				onStatesReceived={ handleOnStatesReceived }
+				campaign={ campaign }
+				countryCodes={ countryCodes }
+				loadCampaignFromClientSession={ isOnboardingFlow }
+				showCampaignPreviewCard={
+					trackingContext === 'setup-ads' ||
+					trackingContext === 'create-ads'
+				}
+			/>
```
eason9487: After adding `PaidAdsSetupSections` to this component, it leads to nested forms. It's not ideal as all sub-components interacting with `Form` or `AdaptiveForm` component APIs become confused and less maintainability.



The form-related logic in `PaidAdsSetupSections` would be better to split and lift to `SetupPaidAds`, and then determine if the rest of its implementation could be achieved with [the original way](https://github.com/woocommerce/google-listings-and-ads/blob/3d38d913c5ff4b2bce72e177a07a1188ead68dd8/js/src/components/paid-ads/ads-campaign.js#L95-L101). Perhaps eventually the `PaidAdsSetupSections` component can be fully removed.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,189 @@
...
+
+			<PaidAdsSetupSections
+				onStatesReceived={ handleOnStatesReceived }
+				campaign={ campaign }
+				countryCodes={ countryCodes }
+				loadCampaignFromClientSession={ isOnboardingFlow }
+				showCampaignPreviewCard={
+					trackingContext === 'setup-ads' ||
+					trackingContext === 'create-ads'
+				}
```
eason9487: As per the first acceptance criteria:

> When a user is on the "Create your paid campaign" step of the "Set up paid ads" flow, they will see the same form used on the "Create a campaign" step of the "Get started..." onboarding flow.

I would like to confirm what the scope of the “same form” is, after all, they are currently presented differently in the campaign preview.

js/src/components/paid-ads/ads-campaign/paid-ads-setup-sections.js
eason9487: When the `campaign` is provided and `loadCampaignFromClientSession` is true, they functionally conflict. Because their JSDocs both say that their campaign data source will be adopted, yet one of them will be ignored.

If adjustments are made in the direction suggested in https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1791508147, this problem should also disappear.

js/src/components/paid-ads/billing-card/billing-card.js
```diff
@@ -15,21 +16,33 @@ import fallbackBillingUrl from './fallbackBillingUrl';
...
+	const [ showNotice, setShowNotice ] = useState( false );
 	const { billingStatus, hasFinishedResolution } =
 		useGoogleAdsAccountBillingStatus();
 
+	useEffect( () => {
+		// only show notice if we are in pending state before
+		if ( billingStatus.status === PENDING && ! showNotice ) {
+			setShowNotice( true );
+		}
+	}, [ billingStatus, showNotice ] );
```
eason9487: The billing card is not shown after refreshing the webpage at step 4 of the onboarding flow.

image

js/src/components/paid-ads/billing-card/billing-card.js
```diff
@@ -15,21 +16,33 @@ import fallbackBillingUrl from './fallbackBillingUrl';
...
+
 	if ( ! hasFinishedResolution ) {
 		return <SpinnerCard />;
 	}
 
 	if ( billingStatus.status === APPROVED ) {
+		if ( ! showNotice ) {
+			return null;
+		}
```
eason9487: This makes the appearance and disappearance of the last loading spinner even more inappropriate.

### Refresh the webpage at step 4 of the onboarding flow

https://github.com/user-attachments/assets/86113cc6-4c2b-4d4d-89b4-22de7436a4ab

### Enter the campaign creation page

https://github.com/user-attachments/assets/3f7fbb47-7339-434b-9750-cac8dd6f1f9e

eason9487 (changes requested):

js/src/components/paid-ads/ads-campaign/ads-campaign.js
joemcgill: The "same form" means that we will use the same component for both forms so that the requirements of each input section are consistent. The main difference is that the `PaidAdsFeaturesSection` should only be shown during onboarding.

eason9487: Thanks for the explanation.

I would suggest rephrasing it a little bit to avoid having “When a user …, they will see …” wording that confuses its goal of just having to share components in development. Or, clarify the goal is to have the same UI as before the change.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: `continueButton` prop added.
https://github.com/woocommerce/google-listings-and-ads/pull/2623/files#diff-e44a2cf9e5c408b79fc5c34e07b13864077ddae579701d8834647bca62494ccdR52

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: `skipButton` prop added @eason9487 
https://github.com/woocommerce/google-listings-and-ads/pull/2623/files#diff-e44a2cf9e5c408b79fc5c34e07b13864077ddae579701d8834647bca62494ccdR51

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: Some of the logic has been ported @eason9487 and simplified as well.

js/src/components/paid-ads/ads-campaign/paid-ads-setup-sections.js
asvinb: This has been removed and no longer applicable.

js/src/components/paid-ads/billing-card/billing-card.js
asvinb: Removed the logic.

js/src/components/paid-ads/billing-card/billing-card.js
asvinb: @eason9487 We are now showing the `BillingCard` component only during onboarding or the setup Ads flow.

ankitguptaindia (commented):
✅  Verified the changes from a user perspective and confirmed that creating, editing, and managing paid ad campaigns from all desired screens is working without any errors.

✅  When billing is not set up, the merchant is unable to complete the setup process, as expected.

✅ All cases defined in Acceptance Criteria works as expected.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,113 @@
...
+import PaidAdsFeaturesSection from './paid-ads-features-section';
+import PaidAdsSetupSections from './paid-ads-setup-sections';
+import useTargetAudienceFinalCountryCodes from '.~/hooks/useTargetAudienceFinalCountryCodes';
+
+/**
+ * @typedef {import('.~/components/adaptive-form/adaptive-form-context').AdaptiveFormContext} AdaptiveFormContext
+ */
+
+/**
+ * @typedef {import('.~/components/paid-ads/ads-campaign/paid-ads-setup-sections').PaidAdsData} PaidAdsData
```
eason9487: This file doesn't have the `PaidAdsData` type definition.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,113 @@
...
+ * so it expects a `CampaignAssetsForm` to existing in its parents.
+ *
+ * @fires gla_documentation_link_click with `{ context: 'create-ads' | 'edit-ads' | 'setup-ads', link_id: 'see-what-ads-look-like', href: 'https://support.google.com/google-ads/answer/6275294' }`
+ * @param {Object} props React props.
+ * @param {string} props.headerTitle The title of the step.
+ * @param {boolean} [props.isOnboardingFlow=false] Whether this component is used in onboarding flow.
+ * @param {'create-ads'|'edit-ads'|'setup-ads'} props.trackingContext A context indicating which page this component is used on. This will be the value of `context` in the track event properties.
+ * @param {JSX|(PaidAdsData) =>JSX} [props.skipButton] A React element or function to render the "Skip" button. If a function is passed, it receives the paid ads data and returns the button element.
+ * @param {JSX|(AdaptiveFormContext, PaidAdsData) =>JSX} [props.continueButton] A React element or function to render the "Continue" button. If a function is passed, it receives the form context and paid ads data and returns the button element. It handles submission logic in the form.
```
eason9487: There are some issues in these JSDoc:
1. `JSX` is a type namespace rather than a type.
2. Due to the limitation of the `npm run doc:tracking` script, the function type needs to be placed ahead of another one. For example, swap `string | () => boolean` to `() => boolean | string`. Otherwise, the script can't parse it.
   ```
   ERROR: Unable to parse a tag's type expression for source file js/src/components/paid-ads/ads-campaign/ads-campaign.js in line 29 with tag title "param" and text "{JSX|(module:components/paid-ads/ads-campaign/ads-campaign~PaidAdsData) =>JSX} [props .skipButton] A React element or function to render the "Skip" button. If a function is passed, it receives the paid ads data and returns the button element.": Invalid type expression "JSX|(module:components/paid-ads/ads-campaign/ads-campaign~PaidAdsData) =>JSX": Expected "!", "$", "'", "*", ".", "...", "0", "?", "@", "Function", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "false", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "null", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "true", "try", "typeof", "undefined", "var", "void", "while", "with", "yield", "{", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter, or [1-9] but "(" found.
   ```
3. Their implementation doesn't pass a `PaidAdsData` as a function parameter.
4. The `props.continueButton` doesn't actually handle submission logic in the form in this component, and it's not always expected that the consumer side will use it to submit a form.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+export default function AdsCampaign( {
+	headerTitle,
+	isOnboardingFlow = false,
+	trackingContext,
+	skipButton,
+	continueButton,
+} ) {
+	const formContext = useAdaptiveFormContext();
+	const { data: countryCodes } = useTargetAudienceFinalCountryCodes();
```
eason9487, line 53: This `countryCodes` may lead to incorrect budget recommendation text and amount when editing an existing campaign having different target audience countries than the current ones in the free listings setup.

For example, given a campaign was created with one country code but is being edited after the free listings setup has been updated to have two country codes, the recommendation text will be incorrect:
- Expected: Most merchants targeting **Taiwan** set a daily budget of **12** USD
- Received: Most merchants targeting **similar countries** set a daily budget of **17** USD

https://github.com/user-attachments/assets/582ed01c-aadd-4553-8479-81b3a56a00b6

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -0,0 +1,113 @@
...
+			/>
+
+			{ isOnboardingFlow && <PaidAdsFeaturesSection /> }
+
+			<PaidAdsSetupSections
+				countryCodes={ countryCodes }
+				showCampaignPreviewCard={
+					trackingContext === 'setup-ads' ||
+					trackingContext === 'create-ads'
+				}
```
eason9487: Before this PR, the preview card is shown as well when editing campaign.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
eason9487, line 1: Now that the `AdsCampaign` component has become more flexible and simple to use, however, there are still a few factors that can be polished.

- Render logic depending on a prop named `trackingContext` seems inappropriate.
- `isOnboardingFlow` and `trackingContext` could be consolidated and lifted to a higher abstraction level.
- `PaidAdsSetupSections` is now purer and contains only **one** section.
- Even if there is no need to render a `BillingCard` when creating or editing campaign, a redundant API request will still be sent to get the billing status.

Suggestions to further simplify the logic and divide component responsibilities:

- Remove `isOnboardingFlow` prop and rename `trackingContext` to `context`.
   - `context` accepts one more possible string value, such as `'setup-mc'`. to indicate it's being used in onboarding
   - Declare render condition variables within the components, for example:
      ```js
      const isOnboardingFlow = context === 'setup-mc';
      const showBillingCard = context === 'setup-mc' || context === 'setup-ads';
      ```
- Remove `PaidAdsSetupSections` to reduce the complexity of a layer of components.
   - Move the API request of getting billing status and `SpinnerCard` back to the `BillingCard` component
   - Moe the rest to the `AdsCampaign` component and use the above render condition variables, for example:
      ```jsx
      <BudgetSection
      	formProps={ formContext }
      	countryCodes={ /* value for this prop needs to consider another comment */ }
      >
      	{ showBillingCard && <BillingCard /> }
      	{ ! isOnboardingFlow && <CampaignPreviewCard /> }
      </BudgetSection>
      ```

js/src/components/paid-ads/ads-campaign/continue-button.js
```diff
@@ -0,0 +1,38 @@
...
+	const { billingStatus } = useGoogleAdsAccountBillingStatus();
+	const isBillingCompleted =
+		billingStatus?.status === GOOGLE_ADS_BILLING_STATUS.APPROVED;
+
+	return (
+		<AppButton
+			isPrimary
+			text={ __( 'Continue', 'google-listings-and-ads' ) }
+			onClick={ onClick }
+			disabled={ ! formProps.isValidForm || ! isBillingCompleted }
```
eason9487: As per https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1792488819, suggest moving the billing status check out from this button, and maybe using the same [button implementation](https://github.com/woocommerce/google-listings-and-ads/blob/6b85d4dd65b4bb3504dfec0545d9a242ca16fc78/js/src/components/paid-ads/continue-button.js) for this PR to reduce the difficulty when merging each PR together.

js/src/pages/create-paid-ads-campaign/index.js
```diff
@@ -133,6 +135,7 @@ const CreatePaidAdsCampaign = () => {
 					countryCodes: initialCountryCodes,
 				} }
 				onSubmit={ handleSubmit }
+				validate={ validateCampaign }
```
eason9487: This appears to be redundant since `CampaignAssetsForm` already uses the same validator.

js/src/pages/edit-paid-ads-campaign/index.js
```diff
@@ -183,6 +185,7 @@ const EditPaidAdsCampaign = () => {
 				} }
 				assetEntityGroup={ assetEntityGroup }
 				onSubmit={ handleSubmit }
+				validate={ validateCampaign }
```
eason9487: This appears to be redundant since `CampaignAssetsForm` already uses the same validator.

js/src/setup-mc/setup-stepper/skip-button.js
```diff
@@ -0,0 +1,88 @@
...
+			google_ads_account_status: googleAdsAccount?.status,
+			billing_method_status: billingStatus?.status || 'unknown',
+			campaign_form_validation: paidAds?.isValid ? 'valid' : 'invalid',
+		};
+		recordGlaEvent( 'gla_onboarding_complete_button_click', eventProps );
+
+		onSkipCreatePaidAds();
+	};
+
+	const isDisabled = disabled || ! hasGoogleAdsConnection;
```
eason9487: The connection check of Google Ads account could be omitted as it should have been checked at step 1.

js/src/setup-mc/setup-stepper/skip-button.js
```diff
@@ -0,0 +1,88 @@
...
+
+	return (
+		<>
+			<AppButton
+				isTertiary
+				data-action={ ACTION_SKIP }
```
eason9487: This `data-action` is no longer in use.

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -0,0 +1,190 @@
...
+			await finishOnboardingSetup( onBeforeFinish );
+		};
+
+		return (
+			<AppButton
+				isPrimary
+				disabled={ disabled }
+				onClick={ handleCompleteClick }
+				loading={ completing === ACTION_COMPLETE }
+				data-action="ACTION_COMPLETE"
```
eason9487: This `data-action` is no longer in use. (Despite this, its value is incorrect.)

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -0,0 +1,190 @@
...
+ * @typedef {Object} PaidAdsData
+ * @property {number|undefined} amount Daily average cost of the paid ads campaign.
+ * @property {boolean} isValid Whether the campaign data are valid values.
+ * @property {boolean} isReady Whether the campaign data and the billing setting are ready for completing the paid ads setup.
+ */
+
+const defaultPaidAds = {
+	amount: 0,
+	isValid: false,
+	isReady: false,
```
eason9487: `isReady` is no longer available.

js/src/setup-ads/ads-stepper/index.test.js
```diff
@@ -25,6 +13,17 @@ import SetupAccounts from './setup-accounts';
...
+jest.mock( '@woocommerce/tracks', () => {
+	return {
+		recordEvent: jest.fn().mockName( 'recordEvent' ),
+	};
+} );
+jest.mock( './setup-accounts', () => jest.fn().mockName( 'SetupAccounts' ) );
+jest.mock( '.~/components/paid-ads/ads-campaign', () =>
+	jest.fn().mockName( 'AdsCampaign' )
+);
+jest.mock( './setup-billing', () => jest.fn().mockName( 'SetupBilling' ) );
```
eason9487: Why does it need to move these codes down here?

js/src/setup-ads/ads-stepper/index.js
eason9487, line 1: Since #2577 has been merged, the todo https://github.com/woocommerce/google-listings-and-ads/pull/2577#discussion_r1792488819 item can be handled here.

js/src/setup-ads/setup-ads-form.js
```diff
@@ -87,6 +88,7 @@ const SetupAdsForm = () => {
 			initialCampaign={ initialValues }
 			onChange={ handleChange }
 			onSubmit={ handleSubmit }
+			validate={ validateCampaign }
```
eason9487: This appears to be redundant since `CampaignAssetsForm` already uses the same validator.

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -0,0 +1,190 @@
...
+	};
+
+	useEffect( () => {
+		clientSession.setCampaign( paidAds );
+	}, [ paidAds ] );
+
+	return (
+		<CampaignAssetsForm
+			initialCampaign={ paidAds }
+			validate={ validateCampaign }
```
eason9487: This appears to be redundant since `CampaignAssetsForm` already uses the same validator.

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -0,0 +1,190 @@
...
+		);
+	};
+
+	const createContinueButton = ( formContext ) => {
+		const { isValidForm } = formContext;
+
+		const disabled =
+			completing === ACTION_SKIP ||
+			! paidAds.isValid ||
+			! isValidForm ||
```
eason9487: These two values are equivalent.

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -0,0 +1,190 @@
...
+				eventProps={ {
+					budget: paidAds.amount,
+					audiences: countryCodes?.join( ',' ),
+				} }
+			/>
+		);
+	};
+
+	useEffect( () => {
+		clientSession.setCampaign( paidAds );
```
eason9487: This can be handled in the `onChange` callback of `CampaignAssetsForm`.

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -0,0 +1,190 @@
...
+	}, [ paidAds ] );
+
+	return (
+		<CampaignAssetsForm
+			initialCampaign={ paidAds }
+			validate={ validateCampaign }
+			onChange={ ( _, values, isValid ) => {
+				setPaidAds( { ...paidAds, ...values, isValid } );
```
eason9487: Previously, the form component was under several layers of this component, so it required a more cumbersome way to retrieve the form data and validation state.

Now this component is wrapped by `CampaignAssetsForm`, so all handling related to using `paidAds` and calling `setPaidAds` should be able to be removed.

- The `paidAds.amount` value can be replaced with `formContext.values.amount`
- The `paidAds.isValid` state can be replaced with `formContext.isValidForm`

js/src/pages/create-paid-ads-campaign/index.js
```diff
@@ -133,6 +135,7 @@ const CreatePaidAdsCampaign = () => {
 					countryCodes: initialCountryCodes,
```
eason9487: The audience section has been completely removed from campaign creation, so it would be better to decouple the `countryCodes` from the form component.

js/src/pages/edit-paid-ads-campaign/index.js
eason9487, line 1: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1796405650.

js/src/setup-ads/setup-ads-form.js
eason9487, line 1: Similar to https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1796405650.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -409,107 +385,67 @@ test.describe( 'Set up Ads account', () => {
...
-						'In order to launch your paid campaign, your billing information is required. You will be billed directly by Google and only pay when someone clicks on your ad.'
+						'You do not have billing information set up in your Google Ads account. Once you have set up billing, you can start running ads.'
 					)
 				).toBeVisible();
-
-				await expect(
-					page.getByRole( 'link', {
-						name: 'click here instead',
-					} )
-				).toBeVisible();
```
eason9487: Is it necessary to remove this assertion?

eason9487 (changes requested):

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: 👍

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: JSDoc updated @eason9487 . Thanks for catching that.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: Updated!

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: Thanks for the suggestions @eason9487 . Makes sense. Updated!

js/src/pages/create-paid-ads-campaign/index.js
asvinb: 👍

js/src/pages/edit-paid-ads-campaign/index.js
asvinb: 👍

js/src/setup-mc/setup-stepper/skip-button.js
asvinb: 👍

js/src/setup-mc/setup-stepper/skip-button.js
asvinb: 👍

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: 👍

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: 👍

js/src/setup-ads/ads-stepper/index.test.js
asvinb: Restored. I see the position of the mocks are quite inconsistent across the codebase. Sometimes it's after/before the dependencies comment.

js/src/setup-ads/setup-ads-form.js
asvinb: 👍

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: 👍

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: 👍 Using `isValidForm` now

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: 👍

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: Sounds good. Updated.

js/src/pages/create-paid-ads-campaign/index.js
asvinb: 👍

js/src/pages/edit-paid-ads-campaign/index.js
asvinb: 👍

js/src/setup-ads/setup-ads-form.js
asvinb: 👍

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
asvinb: Not sure why it was removed in first place. I restored it 👍

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: @eason9487 This should be fixed now. We are now passing the `displayCountries` property of a campaign.
https://github.com/woocommerce/google-listings-and-ads/pull/2623/files#diff-e44a2cf9e5c408b79fc5c34e07b13864077ddae579701d8834647bca62494ccdR99

js/src/components/paid-ads/ads-campaign/continue-button.js
asvinb: 👍

js/src/setup-ads/ads-stepper/index.js
asvinb: 👍

asvinb:
@ankitguptaindia Can you kindly test the changes please? I also added items to double check in the QAB.

Extra issues to test:

https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1794919412
https://github.com/woocommerce/google-listings-and-ads/pull/2623#discussion_r1794924425

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* Theme active on store:  Twenty Twenty-Four Version: 1.2
* WooCommerce - Version 9.3.3
* PHP: 8.3
* Web Server: Nginx
* Browser: Chrome - Version 129
* OS: macOS Sonoma 14.6.1 

</Details>



**Test Results -**  Tested the use cases described here https://github.com/woocommerce/google-listings-and-ads/pull/2623#issuecomment-2407721963 and paid ad creation flow after the recent changes. New paid ad campaign and editing ad campaign flow and other related use cases are working fine now. 


https://github.com/user-attachments/assets/c8f174e5-b8d6-4e69-8770-c109f8fc0767

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -30,81 +36,86 @@ import PaidAdsFaqsPanel from '../faqs-panel';
...
- * @param {JSX.Element|Function} props.continueButton Continue button component.
- * @param {'create-ads'|'edit-ads'|'setup-ads'} props.trackingContext A context indicating which page this component is used on. This will be the value of `context` in the track event properties.
+ * @param {Campaign} [props.campaign] Campaign data to be edited. The displayCountries property will be used to fetch budget recommendation data.
+ * @param {string} props.headerTitle The title of the step.
+ * @param {'create-ads'|'edit-ads'|'setup-ads'|'setup-mc'} props.context A context indicating which page this component is used on. This will be the value of `context` in the track event properties.
+ * @param {(AdaptiveFormContext) => JSX.Element|JSX.Element} [props.skipButton] A React element or function to render the "Skip" button. If a function is passed, it receives the form context and returns the button element.
+ * @param {(AdaptiveFormContext) => JSX.Element|JSX.Element} [props.continueButton] A React element or function to render the "Continue" button. If a function is passed, it receives the form context and returns the button element.
```
eason9487: Fix the syntax for the callback parameters:

```suggestion
 * @param {(formContext: AdaptiveFormContext) => JSX.Element | JSX.Element} [props.skipButton] A React element or function to render the "Skip" button. If a function is passed, it receives the form context and returns the button element.
 * @param {(formContext: AdaptiveFormContext) => JSX.Element | JSX.Element} [props.continueButton] A React element or function to render the "Continue" button. If a function is passed, it receives the form context and returns the button element.
```

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -21,12 +21,13 @@ import {
 
 /**
  * @param {Object} props React props
- * @param {Object} props.formProps Form props forwarded from `Form` component.
+ * @param {Object} [props.isSubmitting] Where the form ads setup has been submitted via the useAdsSetupCompleteCallback hook.
```
eason9487: - Should it be optional?
- The "Where" seems a typo.
- It's more precise to describe this state as "is currently being submitted" rather than "has been submitted".

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -90,19 +87,17 @@ const AdsStepper = ( { formProps } ) => {
...
-									onClick={ formProps.handleSubmit }
+							headerTitle={ __(
+								'Create your paid campaign',
+								'google-listings-and-ads'
+							) }
+							context="setup-ads"
+							continueButton={ ( formContext ) => (
+								<LaunchPaidCampaignButton
+									formProps={ formContext }
+									isLoading={ isSubmitting }
```
eason9487: This button is not shared with others, so it may not needed as a standalone component.

js/src/setup-ads/setup-ads-form.js
```diff
@@ -68,17 +67,14 @@ const SetupAdsForm = () => {
...
 	const handleChange = ( _, values ) => {
-		const args = [ initialValues, values ].map(
-			( { countryCodes, ...v } ) => {
-				v.countrySet = new Set( countryCodes );
-				return v;
-			}
-		);
+		const args = [ initialValues, values ].map( ( v ) => {
+			return v;
+		} );
```
eason9487: This array mapping becomes redundant.

js/src/setup-mc/setup-stepper/setup-paid-ads.js
```diff
@@ -0,0 +1,157 @@
...
+ * @property {number} budget The budget for the campaign
+ * @property {string} audiences The targeted audiences for the campaign
+ */
+
+/**
+ *
+ * @typedef {Object} PaidAdsData
+ * @property {number|undefined} amount Daily average cost of the paid ads campaign.
+ * @property {boolean} isValid Whether the campaign data are valid values.
+ */
```
eason9487: This type is no longer in use.

eason9487 (changes requested):
Other issues:
- After submitting to create or edit a campaign, an uncaught error occurs.
   image
- The `CampaignFormValues` type no longer has the `countryCodes` property.
   https://github.com/woocommerce/google-listings-and-ads/blob/d7f0d5182d0b7c746d9f03f327da8b5656bbd7e5/js/src/components/types.js#L6-L7

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: Updated @eason9487

js/src/setup-ads/ads-stepper/index.js
asvinb: * No it's not optional (where it's used currently)
* Fixed typo
* Updated comment.

js/src/setup-ads/ads-stepper/index.js
asvinb: I moved it within `AdsStepper`

js/src/setup-ads/setup-ads-form.js
asvinb: 👍

js/src/setup-mc/setup-stepper/setup-paid-ads.js
asvinb: Removed @eason9487

asvinb:
@eason9487 The uncaught error has been fixed. Can you kindly check again please?
Thanks!

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -12,21 +12,27 @@ import SetupAccounts from './setup-accounts';
...
 	FILTER_ONBOARDING,
 	CONTEXT_ADS_ONBOARDING,
 } from '.~/utils/tracks';
 
+const { APPROVED } = GOOGLE_ADS_BILLING_STATUS;
+
 /**
  * @param {Object} props React props
- * @param {Object} props.formProps Form props forwarded from `Form` component.
+ * @param {Object} props.isSubmitting When the form in the parent component, i.e SetupAdsForm, is currently being submitted via the useAdsSetupCompleteCallback hook.
```
eason9487: The type should be boolean.

js/src/setup-ads/ads-stepper/index.js
```diff
@@ -12,21 +12,27 @@ import SetupAccounts from './setup-accounts';
...
  * @fires gla_setup_ads with `{ triggered_by: 'step1-continue-button', action: 'go-to-step2' }`.
  * @fires gla_setup_ads with `{ triggered_by: 'stepper-step1-button', action: 'go-to-step1'}`.
  */
-const AdsStepper = ( { formProps } ) => {
+const AdsStepper = ( { isSubmitting } ) => {
 	const [ step, setStep ] = useState( '1' );
+	const { billingStatus } = useGoogleAdsAccountBillingStatus();
```
eason9487: Without a connected Google Ads account, entering the Ads Setup flow will cause critical errors.

image

If the reason for making the continue button a standalone component (`LaunchPaidCampaignButton`) was to avoid these errors, then I would suggest using a similar structure as `SetupPaidAds` to isolate the following scope.

https://github.com/woocommerce/google-listings-and-ads/blob/6b89299a76d15200e5d88e514fe15b0aa90c75f5/js/src/setup-ads/ads-stepper/index.js#L94-L115

For example:

```jsx
function ExampleComponent() {
	const { billingStatus } = useGoogleAdsAccountBillingStatus();

	return (
		<AdsCampaign
			// ... omitted
			continueButton={ ( formContext ) => (
				<AppButton
					// ... omitted
				/>
			) }
		/>
	);
}
```

This would create a more consistent level of abstraction within `AdsStepper`, and the continue button implementation could still be placed together with `AdsCampaign` for better readability.

js/src/components/paid-ads/asset-group/asset-group.js
```diff
@@ -62,15 +63,17 @@ export default function AssetGroup( { campaign } ) {
 	const isCreation = ! campaign;
 	const { isValidForm, handleSubmit, adapter, values } =
 		useAdaptiveFormContext();
+	const { data: countryCodes } = useTargetAudienceFinalCountryCodes();
 	const { isValidAssetGroup, isSubmitting, isSubmitted, submitter } = adapter;
 	const currentAction = submitter?.dataset.action;
+	const audiences = isCreation ? countryCodes : campaign.displayCountries;
```
eason9487: This variable is only used within the scope of `recordSubmissionClickEvent` function. How about moving the variable declaration into the function?

eason9487 (changes requested):

js/src/setup-ads/ads-stepper/index.js
asvinb: 🤦 Didn't notice that. Thanks!

js/src/setup-ads/ads-stepper/index.js
asvinb: Thanks for the suggestion @eason9487 . I've added another `SetupPaidAds` component where we now have `AdsCampaign` along with the billing status check. I've kept the `SetupPaidAds` name to be consistent.

js/src/components/paid-ads/asset-group/asset-group.js
asvinb: 👍

eason9487 (approved):
Thanks for the work. LGTM.

PR #2626, opened 2024-09-27
[API PULL] Fix tests for Notifying delete products 

puntope:

There is a test failing in regards 

```
Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Product\SyncerHooksTest::test_delete_created_product_calls_notify_directly
Expectation failed for method name is "notify" when invoked 1 time(s).
Method was expected to be called 1 times, actually called 0 times.
```

This seems to happen because the metadata is empty after the code calls the delete hook. To solve it. I moved the notifications to the pre-delete.

1.  Tests are passing
2.  Create a simple product. It will schedule a create notification. See it works as expected and shows in the logs after the action is completed.
3. Trash the simple product. See directly the delete notification in the logs.  

### Changelog entry

> Tweak - Logic for Delete notifications.

### DIFF
--- a/src/Product/SyncerHooks.php
+++ b/src/Product/SyncerHooks.php
@@ -324,8 +324,6 @@ class SyncerHooks implements Service, Registerable {
 	 */
 	protected function maybe_send_delete_notification( int $product_id ) {
 		$product = $this->wc->maybe_get_product( $product_id );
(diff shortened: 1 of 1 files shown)


jorgemd24 (approved):
Thanks @puntope for fixing this! I tested it locally, and everything is working well.

PR #2583, opened 2024-09-05
Streamline campaign setup: Adjust the minimum average daily cost of a campaign to 30% of the highest recommended value among audience countries

dsawardekar:
### Changes proposed in this Pull Request:

Closes #2503

This PR updates the campaign flow to include validation of minimum daily limits. An error message is shown if the amount is less than the daily limit.

### Screenshots:

With an insufficient amount:
image

With a sufficient amount:
image

### Detailed test instructions:

1. On the final step of the onboarding flow, Click create a campaign
2. You should see the Set your budget section
3. In the Daily avg cost, enter 1 < value < 5
4. You should see the above error below the input (red styling TBD)

### Additional details:

Update - Enforce the daily limit of 30% in setup your budget section

### Changelog entry

> Update - Adjust the minimum average daily cost of a campaign to 30% of the highest recommended value among audience countries.

PR #2637, opened 2024-10-02
Streamline campaign setup: Copy the `usFetchBudgetRecommendation` hook

### LINKED ISSUE #2502: Campaign Creation: Update Budget Setup Card

asvinb:

Closes #2502  .

### Additional details:
* Copied the `usFetchBudgetRecommendation` hook from https://github.com/woocommerce/google-listings-and-ads/pull/2583 along with dependencies.

### Changelog entry

> changelog: none


joemcgill (approved):
These changes look good to me. Let's go ahead and merge this and re review the original PR for #2502.

PR #2618, opened 2024-09-18
Consolidate accounts onboarding: Automatically create Google Merchant Center or Google Ads accounts when the connected Google account doesn't have a respective existing one

ankitrox:
### Changes proposed in this Pull Request:

Closes #2567  .

_Replace this with a good description of your changes & reasoning._

## Detailed test instructions

1. Create a new site and setup the GLA plugin (Make sure you are testing on the feature branch if this is not merged).

2. Go to `Marketing > Google for Woocommerce` and start onboarding process.

3. Connect to the fresh Google account. This account should not have any existing merchant center accounts or Google ads account.

4. As soon as user sign-in flow is completed and user lands back to onboarding screen, there should be the card status changed to the one which should read `You don’t have Merchant Center nor Google Ads accounts, so we’re creating them for you.` as per the [Figma designs](https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7237-22545&node-type=frame&t=RDgyblprY6tuzPTD-0).

5. Once Merchant Center account and Ads account is created, card state will be changed [this](https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7363-4521&node-type=frame&t=RDgyblprY6tuzPTD-0). Which will display Merchant Center ID and Google Ads ID (for 1-2 seconds ads ID may be seen as zero because ads request gets resolved later).

6. Note that there would be some error notices popping up regarding ads account connection and ads connection status, but this can be ignored for this issue and will be handled in #2582

#### Use custom plugin to test with existing account.

As it is very inconvenient to test with the fresh account everytime; especially in the scenario where Google may not allow more accounts to be created and asks for new mobile number (where we are out of phone numbers). Following method can help us to test with existing Google account.

1. Following plugin mimics the scenario such that for the first request to MC or Ads, it mocks the request such that there are no MC and Ads account.

```php
<?php
/**
 * Plugin name: 2567 Test
 * Description: A plugin to provide REST API helper utilities.
 */

add_filter('rest_pre_dispatch', function($result, $server, $request) {
    $method = $request->get_method();
    $route  = $request->get_route();

    if ( 'POST' === $method ) {
        switch ($route) {
            case '/wc/gla/ads/accounts':
                $result = [
                    'message' => 'Account must be accepted before completing setup.',
                    'has_access' => false,
                    'step' => 'account_access',
[...]

PR #2639, opened 2024-10-03
Consolidate accounts onboarding: Set up Google Merchant Center account via the combo card

### LINKED ISSUE #2597: Onboarding: Connect to existing MC account in Google Accounts Card

### LINKED ISSUE #2604: Onboarding: Create new MC account when there are existing accounts in Google Accounts Card
>[!Caution]
>Depends on #2597 

While connecting the store to Merchant Center, a merchant may want to create a new account, even if they already have existing accounts associated with their Google account.

To support this, we will add a link below the account selector to kick off account creation.

Image

Clicking this link should show the same "Are you sure" modal that is shown today, which lets them initiate account creation or cancel.

Image

Once they confirm they want a new account, the account selector will replaced with a message that reads:

> **Creating a new Google Merchant Center account**
> This may take a few minutes, please wait a moment...

And the Connect button will be replaced by a loading spinner with the word "Creating..."

If the merchant needs to reclaim their URL, we'll display the reclaim UI in the card as shown:

Image

Once the new account is created and reclaimed, the card will show the new account as connected

Image

## Background

In the current `GoogleMCAccountCard`, when there are already MC accounts associated with the Google account, we show a link below the account selector that allows them to create a new account instead of selecting an existing one.

Image

Clicking this link shows a modal to confirm they want to create an account.

Image

Once they've confirmed they want a new account and have accepted terms and conditions (not applicable, since they now accept terms and conditions when connecting their Google account), we show this UI while the account is being connected.

Image

After the new account is created, the user is shown the UI to reclaim their URL if applicable:

Image

## Acceptance Criteria

- [ ] When the MC account selector panel is visible, the user can see the link to create a new account
- [ ] Clicking the create link, displays the "Are you sure..." modal
- [ ] If the user cancels, the modal closes and the account selector is still shown
- [ ] If the user confirms they want to create a new account, the panel is replaced by the UI showing that the account is being connected
- [ ] Once the account is created, if the URL needs to be reclaimed, the reclaim UI is shown.
- [ ] Once the account is created and reclaimed, the plugin finishes setting up the account and shows MC as connected
- [ ] Once an account is connected, the create new link is replaced with a link to disconnect and connect a different account

## Implementation Brief
* Using the `ConnectMC` component created in #2597,
	* Add a new callback when clicking the "Or, create a new Merchant Center account". The `WarningModal` from `js/src/components/google-mc-account-card/warning-modal/index.js` should be rendered, prompting the user's confirmation.
	* If the user confirms, start the creation of a new Google MC account using the `handleCreateAccount` function from the `useCreateMCAccount` hook.
		* Inspiration can be taken from `js/src/components/google-mc-account-card/create-account.js`
		* While the creation is in progress,
			* the dropdown to select an existing account is replaced by a message (refer to the AC)
			* The "Connect" button label changes to "Creating..." with a loading spinner.
		* Appropriate states should be added to the component to cater for the above.
	* Once an account has been created, the merchant might need to reclaim the URL.
		* Render the `ReclaimUrlCard` if the URL needs to be reclaimed.
		* Use the following code `js/src/components/google-mc-account-card/connect-mc/index.js` as inspiration.
	* Once the account has been created and URL claimed (if needed),
		* Set the newly created account to be connected by calling the `handleConnectMC` function via the `useConnectMCAccount` hook.
			* Refer to the implementation in `js/src/components/google-mc-account-card/connect-mc/index.js` on how to pass the new Google MC ID to `useConnectMCAccount`
			* To get the newly created MC ID, check the result of the `useCreateMCAccount` hook.
	* Once we have a connected the MC account, display the accounts dropdown again along with the connected state button.
	* Within the connected state,
		* the "Or, create a new Merchant Center account" button should be replaced by "Or, connect to a different Google Merchant Center account".
			* Refer to the implementation for `handleSwitch` in `js/src/components/google-mc-account-card/connected-google-mc-account-card.js`
				* Look into the possibility of extracting the button into a separate component to reduce code duplication which can be used both, in `js/src/components/google-mc-account-card/connected-google-mc-account-card.js` and current component.

## Test Coverage
* Relevant E2E tests should be added to confirm what's in the AC.

## Definition Questions

- Confirm the UI for the "Creating" state here: https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba?node-id=7702-4791#946046858

asvinb:

Closes #2597  .

1. Go through the onboarding process and make sure you have at least one MC account.
2. The card to connect your MC account should be displayed where you can connect a MC account
3. The user should also be able to create a new account or disconnect their MC account.
4. Designs: https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7447-22280&node-type=frame&t=tKJkit0BKc6985PH-0

### Additional details:
* There are no designs to reclaim a URL and existing components have been used as is.


asvinb:
@joemcgill Can you review PR please? The failing unit tests are from the parent branch.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,130 @@
...
+	const { googleMCAccount, hasFinishedResolution, isPreconditionReady } =
+		useGoogleMCAccount();
+	const [ value, setValue ] = useState();
+	const [ handleConnectMC, resultConnectMC ] = useConnectMCAccount( value );
+	const [ handleCreateAccount, resultCreateAccount ] = useCreateMCAccount();
+
+	// MC is ready when we have a connection.
+	// The `link_ads` step will be resolved when the Ads account is connected
+	// since these can be connected in any order.
+	const isConnected =
```
joemcgill: This should to be cast as a bool, otherwise `isConnected` gets set as the value of `googleMCAccount.id`.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,130 @@
...
+	) {
+		return (
+			<CreatingCard
+				retryAfter={ resultCreateAccount.error?.retry_after }
+				onRetry={ handleCreateAccount }
+			/>
+		);
+	}
+
+	if ( ! hasFinishedResolution ) {
```
joemcgill: Could this be moved to the top of the component as an early return?

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,130 @@
...
+		( resultCreateAccount.loading ||
+			resultCreateAccount.response?.status === 503 )
+	) {
+		return (
+			<CreatingCard
+				retryAfter={ resultCreateAccount.error?.retry_after }
+				onRetry={ handleCreateAccount }
+			/>
+		);
+	}
```
joemcgill: This duplicates a lot of logic directly from `js/src/components/google-mc-account-card/connect-mc/index.js`. Curious if we could abstract some of this into a shared component and avoid the redundant `isConnected` check in the process?

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -109,20 +110,24 @@ const ConnectedGoogleComboAccountCard = () => {
...
+						isCreatingAdsAccount={ isCreatingAdsAccount }
+						isCreatingMCAccount={ isCreatingMCAccount }
+						accountsCreated={ accountsCreated }
+					/>
+				}
+				helper={ getHelper() }
+				indicator={ getIndicator() }
+			/>
+
+			<ConnectMC />
```
joemcgill: Rather than making this a sibling component of the `AccountCard` above, I would have expected `ConnectMC` to return a `Section.Card.Body` component that is passed as a child of the `AccountCard` above.

joemcgill (changes requested):
Thanks @asvinb This is looking pretty good, but I've got a few questions/suggestions inline.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: Updated https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-10d40a2e822051c07a4f91bc799e14e8d7ddba3c48a6771ab6d9085d69b12c06R42

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: Sure thing. Moved to the top of the component @joemcgill

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: @joemcgill I've introduced a new `AccountConnectionStatus` where we have this logic.
https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-398198aee87feb6f89983540431ac7e76d24a4d36793dd440d8758a287e190bbR1

However am not sure if there's an easier way to have the conditions to render that component:
https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-adf83b7ae2454ff433c2f50eb0f115c58b3a97776516764ef0ec88707353d177R45-R51

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
asvinb: 👍

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -26,73 +26,46 @@ import SpinnerCard from '.~/components/spinner-card';
...
-					resultCreateAccount.reset();
-				} }
-			/>
-		);
-	}
+	// MC is ready when we have a connection and preconditions are met.
+	// The `link_ads` step will be resolved when the Ads account is connected
+	// since these can be connected in any order.
+	const isConnected =
+		isPreconditionReady &&
```
joemcgill: Unsure that `isPreconditionReady` is necessary, since we're also checking for values from the `googleMCAccount`, which will always be undefined whenever `isPreconditionReady` is `false`.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -69,10 +74,10 @@ const ConnectedGoogleComboAccountCard = () => {
...
 
 	const isGoogleMCAccountConnected =
-		googleMCAccount?.id ||
-		googleMCAccount?.status === GOOGLE_MC_ACCOUNT_STATUS.CONNECTED ||
-		( googleMCAccount?.status === GOOGLE_MC_ACCOUNT_STATUS.INCOMPLETE &&
-			[ 'link_ads', 'claim' ].includes( googleMCAccount?.step ) );
+		isPreconditionReady &&
+		( googleMCAccount?.status === GOOGLE_MC_ACCOUNT_STATUS.CONNECTED ||
+			( googleMCAccount?.status === GOOGLE_MC_ACCOUNT_STATUS.INCOMPLETE &&
+				googleMCAccount?.step === 'link_ads' ) );
```
joemcgill: Given that we're using this check (or a derivative) in a number of places, we should consider adding it to the `useGoogleMCAccount` hook for consistency.

js/src/components/google-combo-account-card/account-creation-description.js
```diff
@@ -19,12 +18,18 @@ import AccountInfo from './account-info';
...
+ * @param {boolean} props.hasExistingMCAccounts Whether there are existing MC accounts.
+ * @param {boolean} props.isGoogleAdsAccountConnected Whether Google Ads account is connected.
+ * @param {boolean} props.isGoogleMCAccountConnected Whether Merchant Center account is connected.
  */
 const AccountCreationDescription = ( {
 	accountsCreated,
 	isCreatingBothAccounts,
 	isCreatingMCAccount,
 	isCreatingAdsAccount,
+	hasExistingMCAccounts,
```
joemcgill: I don't see a requirement to update the description in this card when manually creating an account. Can we avoid this state and handle any UI related to the manual creation to the ConnectMC component (or a child)?

js/src/components/google-combo-account-card/account-creation-description.js
```diff
@@ -19,12 +18,18 @@ import AccountInfo from './account-info';
...
+	isGoogleAdsAccountConnected,
+	isGoogleMCAccountConnected,
```
joemcgill: Can we infer these values from the `googleMCAccount.id` and `googleAdsAccount.id` values? I'm not sure the preconditions are necessary in order to show the ID that is associated.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -110,24 +115,29 @@ const ConnectedGoogleComboAccountCard = () => {
...
+					}
+					accountsCreated={ accountsCreated }
+					isGoogleMCAccountConnected={ isGoogleMCAccountConnected }
+					isGoogleAdsAccountConnected={ isGoogleAdsAccountConnected }
+				/>
+			}
+			helper={ getHelper() }
+			indicator={ getIndicator() }
+		>
+			<ConnectMC
```
joemcgill: Moving this to a child, while still returning an AccountCard, results in nested `AccountCard` components, which I think is undesirable. 

image

Instead, I thought it would return a `Section.Card.Body` that rendered inside the outer `AccountCard`. However, there are several sub-components that also return `AccountCard` components of `AccountConnectionStatus` that also return Card components, so I think we need to consider whether the best path forward is to implement the whole Google combo card as a `Card` component or as a set of cards that are visually contained together.

joemcgill (commented):
I've been looking over the structure of all of this and noticed a few issues, but most importantly think we need to think about the overall structure for the combo card component that we're producing here.

js/src/components/google-combo-account-card/account-creation-description.js
asvinb: Unfortunately no @joemcgill . For e.g if you connect a MC account and you need to claim the URL, we cannot solely rely on the ID. This is an example of the object returned:
```json
{
    "id": 111111111,
    "status": "incomplete",
    "notification_service_enabled": true,
    "wpcom_rest_api_status": null,
    "step": "claim"
}
```

Technically we cannot show the MC Account has connected when we are at the `claim` step, hence why we need to have all those conditions.

js/src/components/google-combo-account-card/account-creation-description.js
asvinb: That's true @joemcgill . I was looking at https://github.com/woocommerce/google-listings-and-ads/issues/2604 and I think 90% is already done in the current PR. We just needed to add the description which I did. Let me know if you want me to revert. 👍

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
asvinb: Sure. I've added a new `isConnected` property to the `useGoogleMCAccount` hook.
https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-0650a58e854ed2990fc06bb82d5000550962a6df508b21f2a7c29bdec2721ac8R65

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
asvinb: I think it's better to have a set of cards that are visually container together where you have a list of `AccountCard` components where the first one is the Google card where you have all the info displayed and the other cards are for MC and Ads. What do you think?

js/src/components/google-combo-account-card/account-creation-description.js
joemcgill: Creating the new account from #2604 here is great, but when that account is being created, there isn't a need to update the description in the top card.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: While I think it would be great to compose the entire `GoogleComboAccountCard` so that it always rendered a single `AccountCard` component with the MC and Ads subsections returning `<Section.Card.Body>` components, this would require a bigger refactor of the existing `GoogleMCAccountCard` and possibly the `GoogleAdsAccountCard` components in order to share components like the `ReclaimUrlCard` and `SwitchUrlCard` that are repurposed in this design, so this does feel like more of a grouping of `AccountCard` components rather than a single card.

joemcgill:
@asvinb I'm setting this back to in-progress so the changes from #2618 can be incorporated.

js/src/components/google-combo-account-card/account-details.js
```diff
@@ -17,13 +17,14 @@ import useGoogleMCAccount from '.~/hooks/useGoogleMCAccount';
 const AccountDetails = () => {
 	const { google } = useGoogleAccount();
 	const { googleAdsAccount } = useGoogleAdsAccount();
-	const { googleMCAccount } = useGoogleMCAccount();
+	const { googleMCAccount, isReady: isGoogleMCConnected } =
```
joemcgill: Nitpick: destructuring `isReady: isGoogleMCConnected` is a bit confusing since this hook also returns a `hasGoogleMCConnection` prop. 

Suggest:
```suggestion
	const { googleMCAccount, isReady: isGoogleMCReady } =
```

js/src/components/google-combo-account-card/connect-account-card.js
```diff
@@ -0,0 +1,52 @@
...
+			title={ title }
+			helper={ helperText }
+			disabled={ disabled }
+		>
+			<Section.Card.Body className="gla-google-combo-connect-account-card__body">
+				{ body }
+			</Section.Card.Body>
+			<Section.Card.Footer className="gla-google-combo-connect-account-card__footer">
+				{ footer }
+			</Section.Card.Footer>
```
joemcgill: Is a footer section a requirement for this component, or should it only render the `Section.Card.Footer` if a `footer` prop is passed?

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,97 @@
...
+import useConnectMCAccount from '.~/hooks/useConnectMCAccount';
+
+/**
+ * Clicking on the "Switch account" button to select a different Google Merchant Center account to connect.
+ *
+ * @event gla_mc_account_switch_account_button_click
+ * @property {string} context (`switch-url`|`reclaim-url`) - indicate the button is clicked from which step.
+ */
+
+const ConnectMC = ( { createMCAccount, resultCreateMCAccount } ) => {
```
joemcgill: Missing docblock

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,97 @@
...
+	const {
+		googleMCAccount,
+		hasFinishedResolution,
+		isPreconditionReady,
+		isReady: isGoogleMCConnected,
```
joemcgill: Same nit about isReady as before

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,97 @@
...
+
+	if ( ! hasFinishedResolution ) {
+		return <SpinnerCard />;
+	}
+
+	if (
+		! isGoogleMCConnected &&
+		( [ 403, 409 ].includes( resultConnectMC.response?.status ) ||
+			[ 403, 503 ].includes( resultCreateMCAccount.response?.status ) ||
+			resultCreateMCAccount.loading )
```
joemcgill: This same logic is repeated in the `ConnectMC` component [here](https://github.com/woocommerce/google-listings-and-ads/blob/c13c524410cd90962d22e47a886030e947f254ca/js/src/components/google-mc-account-card/connect-mc/index.js#L46-L48). Could we abstract this to a utility or a hook?

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,97 @@
...
+			setAccountID( googleMCAccount.id );
+		}
+	}, [ googleMCAccount, isGoogleMCConnected ] );
+
+	if ( ! hasFinishedResolution ) {
+		return <SpinnerCard />;
+	}
+
+	if (
+		! isGoogleMCConnected &&
```
joemcgill: Is the `isGoogleMCConnected` check necessary for this?

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,97 @@
...
+			}
+			footer={
+				<ConnectMCFooter
+					isConnected={ isGoogleMCConnected }
+					resultConnectMC={ resultConnectMC }
+					resultCreateAccount={ resultCreateMCAccount }
+					handleCreateAccount={ createMCAccount }
+				/>
+			}
+			disabled={ ! isPreconditionReady }
```
joemcgill: I'm not sure it's possible for the `isPreconditionReady` to be `false` here? I can't think of a way that this would be visible if the Google account weren't connected.

joemcgill (commented):
I'm not all the way through the review, but wanted to leave what I have so far. No real blockers and worked as expected when testing.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: Sounds good. Removed @joemcgill

js/src/components/google-combo-account-card/account-details.js
asvinb: 👍 . Updated

js/src/components/google-combo-account-card/connect-account-card.js
asvinb: The footer will be displayed for both, `ConnectMC` and `ConnectAds`. It will house the buttons to either connect/create a new account.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: Thanks. Added.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: 👍

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: I've extracted it in a utility function, `hasAccountConnectionIssue` in :
`js/src/components/google-mc-account-card/connect-mc/utils.js`

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: Yes, because if we have a connected account, we don't want to show the reclaim card. The response for creating the account will stay the same.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: Makes sense @joemcgill . I removed it.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
asvinb: @joemcgill We synced in Slack about that. We'll have a group of cars, so it'll be siblings.

js/src/components/google-combo-account-card/connect-mc/connect-mc-body.js
```diff
@@ -0,0 +1,78 @@
...
+			{ isConnecting && (
+				<LoadingLabel
+					text={ __( 'Connecting…', 'google-listings-and-ads' ) }
+				/>
+			) }
+
+			{ ! isConnected && ! isConnecting && (
+				<AppButton
+					isSecondary
+					disabled={ ! value }
```
joemcgill: In what circumstance would this be visible and disabled?

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
```diff
@@ -0,0 +1,105 @@
...
+	} = useGoogleMCAccount();
+	const [ accountID, setAccountID ] = useState();
+	const [ handleConnectMC, resultConnectMC ] =
+		useConnectMCAccount( accountID );
+
+	useEffect( () => {
+		if ( isGoogleMCReady ) {
+			setAccountID( googleMCAccount.id );
+		}
+	}, [ googleMCAccount, isGoogleMCReady ] );
```
joemcgill: Not sure I understand why we need to save the MC account separately as component state here rather than using `googleMcaccount.id` directly?

js/src/components/google-combo-account-card/connect-mc/merchant-center-select.js
```diff
@@ -0,0 +1,67 @@
...
+const MerchantCenterSelect = ( { isConnected, ...rest } ) => {
+	const { data: existingAccounts } = useExistingGoogleMCAccounts();
+	const { googleMCAccount } = useGoogleMCAccount();
+	const domain = new URL( getSetting( 'homeUrl' ) ).host;
+
+	const accountIdExists = existingAccounts?.some(
+		( existingAccount ) => existingAccount.id === googleMCAccount.id
+	);
+
+	if ( ! accountIdExists && isConnected ) {
```
joemcgill: This is a clever workaround. I assume this is for a newly created account that has not yet shown up in the API response for existing accounts? If so, adding that context to the inline doc would be helpful.

js/src/components/google-mc-account-card/connect-mc/account-connection-status.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {Object} props.resultConnectMC - The result object for connecting the MC account.
+ * @param {Object} props.resultCreateAccount - The result object for creating a new account.
+ * @param {Function} props.onRetry - The function to call when the user wants to retry account creation.
+ */
+const AccountConnectionStatus = ( {
+	resultConnectMC,
+	resultCreateAccount,
+	onRetry,
+} ) => {
+	if ( resultConnectMC.response?.status === 409 ) {
```
joemcgill: Not a blocker, but I think using raw response status codes like this is not very readable. Curious if we could create a utility to map these to more human friendly, or at least add inline comments that makes it easy to understand what a 409 indicates.

joemcgill (approved):
Left a few questions, but no blockers. Marking this as approved and ready for QA/WooCR

js/src/components/google-combo-account-card/connect-mc/connect-mc-body.js
asvinb: @joemcgill This will be visible only when we need to connect a new account. I removed the `disabled` prop since we will always have a value set.

js/src/components/google-combo-account-card/connect-mc/connect-mc.js
asvinb: @joemcgill We need to save the value set from the select in case the user needs to connect to a different acccount.

js/src/components/google-combo-account-card/connect-mc/merchant-center-select.js
asvinb: @joemcgill I added it at the start of the component where we explain why we have this intermediary component:
https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-38d86396eccec4e8e7e5b37e8de9cf25e28fb0a2536261feb312adfe9c6689aeR15-R24

js/src/components/google-mc-account-card/connect-mc/account-connection-status.js
asvinb: I added some inline comments, just describing what are the error statuses. More details can be found in the utils function: https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-289f2d84147885af5105af45f5b616e4057a4490f0fe42b5cf487ca942fdfcefR6-R7

ankitguptaindia:
**QA Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* WooCommerce - Version 9.3.3
* PHP: 8.3
* Web Server: Nginx
* Browser: Chrome - Version 130
* OS: macOS Sonoma 14.6.1 

</Details>

**Test Results:** Tested multiple scenarios related to MC account creation and connection with the store, including cases where Google accounts have multiple MC accounts, a single MC account, connection when the URL is already claimed, connection when the URL doesn’t need to be claimed, and creating a new MC account from the options. All cases were found to be working.



<Details>
  <summary>Functional Demo /  Screencast </summary>



**Connection when Google account has only single MC account:** 

https://github.com/user-attachments/assets/a39bd371-880d-47f1-88e4-f66fcbfa6562


**Creating New account when Google Account has single MC account:** 

https://github.com/user-attachments/assets/07319e39-ae88-484d-8d71-fb880fc4c618


**Creating account when Google Accounts has multiple MC accounts:**


https://github.com/user-attachments/assets/d1c36861-b11e-4cfc-bc24-6900d637c6eb


**Connection with existing MC account when Google has multiple MC accounts:** 

https://github.com/user-attachments/assets/ea0ade86-2e1d-4198-8d60-7552eb7e2cc5


</Details>

ankitguptaindia (approved):

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -39,7 +56,7 @@ const ConnectedGoogleComboAccountCard = () => {
 			<AccountCard
 				appearance={ APPEARANCE.GOOGLE }
 				alignIcon="top"
-				className="gla-google-combo-account-card--connected"
+				className="gla-google-combo-account-card gla-google-combo-account-card--connected"
```
asvinb: Adding the `gla-google-combo-account-card` to be consistent with the other cards.

js/src/components/google-combo-account-card/connect-mc/merchant-center-select.js
```diff
@@ -0,0 +1,68 @@
...
+ * which is not in the list of existing accounts. We then fake the connected select by manually providing
+ * the connected ID.
+ *
+ * @param {Object} props
+ * @param {boolean} props.isConnected Whether the Merchant Center account is connected.
+ */
+const MerchantCenterSelect = ( { isConnected, ...rest } ) => {
+	const { data: existingAccounts } = useExistingGoogleMCAccounts();
+	const { googleMCAccount } = useGoogleMCAccount();
+	const domain = new URL( getSetting( 'homeUrl' ) ).host;
```
eason9487: How about moving this variable into the below `if` block?

js/src/data/resolvers.js
```diff
@@ -102,6 +102,10 @@ export function* getExistingGoogleMCAccounts() {
 	yield fetchExistingGoogleMCAccounts();
 }
 
+getExistingGoogleMCAccounts.shouldInvalidate = ( action ) => {
+	return action.type === TYPES.RECEIVE_ACCOUNTS_GOOGLE_MC;
+};
```
eason9487: Not sure if this invalidation is trying to refetch the existing accounts so that it can have the newly created account. If so, after the following findings, maybe this could be removed since it usually won't work.

https://github.com/woocommerce/google-listings-and-ads/blob/8fba3e4ecd1d8b12f365706545d47f66976a1770/js/src/components/google-combo-account-card/connect-mc/merchant-center-select.js#L20-L22

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -248,6 +263,9 @@ test.describe( 'Set up accounts', () => {
 				await setUpAccountsPage.mockMCConnected();
 				await setUpAccountsPage.mockAdsAccountConnected();
 
+				await setUpAccountsPage.mockMCHasAccounts();
+				await setUpAccountsPage.mockMCConnected();
```
eason9487: Duplicated `await setUpAccountsPage.mockMCConnected();`

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+				const continueButton = setUpAccountsPage.getContinueButton();
+				await expect( continueButton ).toBeDisabled();
+			} );
+
+			test( 'click "Create account" button should see the modal of confirmation of creating account', async () => {
+				// Click the create account button
+				const createAccountButton =
+					setUpAccountsPage.getMCCreateAccountButtonFromPage();
+				await createAccountButton.click();
+				await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
```
eason9487: Why it needs to wait for load state?

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+						// Mock Merchant Center as connected with ID 12345
+						setUpAccountsPage.mockMCConnected( 12345 ),
+					] );
+
+					const createAccountButtonFromModal =
+						setUpAccountsPage.getMCCreateAccountButtonFromModal();
+					await createAccountButtonFromModal.click();
+					await page.waitForLoadState(
+						LOAD_STATE.DOM_CONTENT_LOADED
+					);
```
eason9487: Why it needs to wait for load state?

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
 		} );
 	} );
 
+	test.describe( 'Connect Merchant Center account', () => {
```
eason9487, line 294: There are 7 `await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );` within this block. Suggest rechecking if each one is needed.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+				// Mock Google Ads as connected.
+				setUpAccountsPage.mockAdsAccountConnected(),
+				setUpAccountsPage.mockAdsStatusClaimed(),
+
+				// Mock merchant center as not connected.
+				setUpAccountsPage.mockMCNotConnected(),
+			] );
+		} );
+
+		test.describe( 'Merchant Center has no connected accounts', () => {
```
eason9487: Suggest rephrasing the description to mention that the target of this test group is account creating.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+							setUpAccountsPage.getMCDescriptionRow();
+						await expect( mcDescriptionRow ).toContainText(
+							'Merchant Center ID: 12345'
+						);
+					} );
+				} );
+			} );
+		} );
+
+		test.describe( 'Merchant Center has existing accounts', () => {
```
eason9487: Suggest rephrasing the description to mention that the target of this test group is account connecting.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+							setUpAccountsPage.mockJetpackConnected(),
+
+							// Mock google as connected.
+							setUpAccountsPage.mockGoogleConnected(
+								'google@example.com'
+							),
+							setUpAccountsPage.mockAdsAccountsResponse(
+								ADS_ACCOUNTS
+							),
+							setUpAccountsPage.mockMCHasAccounts(),
```
eason9487: These mocks seem to be redundant as the upper `test.beforeAll` blocks have set the same mocks and they won't invalidated when entering here.

https://github.com/woocommerce/google-listings-and-ads/blob/8fba3e4ecd1d8b12f365706545d47f66976a1770/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L297-L304

https://github.com/woocommerce/google-listings-and-ads/blob/8fba3e4ecd1d8b12f365706545d47f66976a1770/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L315-L318

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+		test.describe( 'Merchant Center has existing accounts', () => {
+			test.beforeAll( async () => {
+				await Promise.all( [
+					// Mock Jetpack as connected.
+					setUpAccountsPage.mockJetpackConnected(),
+
+					// Mock google as connected.
+					setUpAccountsPage.mockGoogleConnected(
+						'google@example.com'
+					),
```
eason9487: These mocks seem to be redundant as the upper `test.beforeAll` block has set the same mocks and they won't invalidated when entering here.

https://github.com/woocommerce/google-listings-and-ads/blob/8fba3e4ecd1d8b12f365706545d47f66976a1770/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L297-L304

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
```diff
@@ -204,7 +212,9 @@ export default class SetUpAccountsPage extends MockRequests {
 	 * @return {import('@playwright/test').Locator} Get select MC accounts select element.
 	 */
 	getMCAccountsSelect() {
-		return this.page.locator( 'select[id*="inspector-select-control"]' );
+		return this.getMCAccountCard().locator(
+			'select[id*="inspector-select-control"]'
+		);
```
eason9487:
```suggestion
		return this.getMCAccountCard().getByRole( 'combobox' );
```

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+						setUpAccountsPage.mockJetpackConnected(),
+
+						// Mock google as connected.
+						setUpAccountsPage.mockGoogleConnected(
+							'google@example.com'
+						),
+
+						// Mock Google Ads as connected.
+						setUpAccountsPage.mockAdsAccountConnected(),
+						setUpAccountsPage.mockAdsStatusClaimed(),
```
eason9487: These mocks seem to be redundant.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +294,361 @@ test.describe( 'Set up accounts', () => {
...
+						setUpAccountsPage.getModalPrimaryButton();
+					await expect( modalPrimaryButton ).toContainText(
+						'Create account'
+					);
+					await expect( modalPrimaryButton ).toBeDisabled();
+
+					// Select the checkbox, the button should be enabled.
+					await modalCheckbox.click();
+					await expect( modalPrimaryButton ).toBeEnabled();
+				} );
```
eason9487: Looks like the same test is already in place.

https://github.com/woocommerce/google-listings-and-ads/blob/8fba3e4ecd1d8b12f365706545d47f66976a1770/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L342-L364

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
```diff
@@ -244,11 +256,13 @@ export default class SetUpAccountsPage extends MockRequests {
...
-				hasText: 'Google',
-			} ),
-		} ).first();
+		return this.page
+			.locator( '.components-card-body', {
+				has: this.page.locator( '.gla-account-card__title', {
+					hasText: 'Google',
+				} ),
+			} )
+			.first();
```
eason9487: May add a CSS class for this as well like the `getMCAccountCard` method below.

eason9487 (commented):
Most of the original components under the `js/src/components/google-mc-account-card` directory are only used for the Onboarding accounts step, except for `ConnectedGoogleMCAccountCard` which is shared by the Setting page. Given this, extracting some code into shared components or functions and creating another `ConnectedMC` component would make for unnecessary redundant code and reduce maintainability.

This PR should be able to be done based on the original existing code and minimize code changes to optimally leverage codes that have been run for 2-3 years.

Suggest exploring possibilities in this direction:

- Adjust the `google-mc-account-card/connect-mc/index.js` to fulfill the new UI and functionality rather than creating `google-combo-account-card/connect-mc/connect-mc.js`
- Revert the extraction of `AccountConnectionStatus` and `hasAccountConnectionIssue`
- `ConnectedGoogleMCAccountCard` will only be used on the Settings page and its `hideAccountSwitch` prop will no longer be needed
  - Remove code related to `hideAccountSwitch`
  - Rename to specify that this component is dedicated to the Settings page

js/src/components/google-combo-account-card/connect-mc/merchant-center-select.js
asvinb: Sure thing! Updated @eason9487

js/src/data/resolvers.js
asvinb: 👍

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Thanks! Removed.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Turns out we don't need it. Removed.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Reviewed and removed unnecessary wait for load state.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Reviewed and removed unnecessary wait for load state.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Description updated @eason9487

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: 👍

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Mocks updated.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: 👍

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
asvinb: 👍

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: 👍 Removed.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Test removed @eason9487

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
asvinb: 👍

asvinb:
@eason9487 I've updated the PR to use the existing `ConnectMC` component. Let me know what you think.

js/src/components/google-mc-account-card/connect-mc/index.js
```diff
@@ -37,12 +36,30 @@ import './index.scss';
  */
 
 /**
+ * ConnectMC component.
+ *
+ * This component renders Merchant Center connection section.
```
eason9487: Suggest replacing "section" with "card".

js/src/components/google-mc-account-card/connect-mc/index.js
```diff
@@ -84,54 +101,69 @@ const ConnectMC = () => {
...
+				'google-listings-and-ads'
+			) }
+			helper={ __(
+				'Required to sync products so they show on Google.',
+				'google-listings-and-ads'
+			) }
+			alignIndicator="toDetail"
+			indicator={ getIndicator() }
+			detail={
+				<MerchantCenterSelect
```
eason9487, line 154: After using the `detail` prop to render the select control, the related styling can be removed.

https://github.com/woocommerce/google-listings-and-ads/blob/9a368f15b081c26de323def85ee4d043e2a656e0/js/src/components/google-mc-account-card/connect-mc/index.scss#L2-L4

js/src/components/google-mc-account-card/connect-mc/index.js
```diff
@@ -84,54 +101,69 @@ const ConnectMC = () => {
...
-				</CreateAccountButton>
-			</Section.Card.Footer>
-		</AccountCard>
+			className={ classnames( 'gla-connect-mc-card', className ) }
+			title={ __(
+				'Connect to existing Merchant Center account',
+				'google-listings-and-ads'
+			) }
+			helper={ __(
+				'Required to sync products so they show on Google.',
```
eason9487, line 148: There is a different wording. Maybe need to confirm which one should be used.

image

js/src/components/google-mc-account-card/connect-mc/actions.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param {boolean}  props.isConnected Whether the Merchant Center account is connected.
+ * @param {Object}   props.resultConnectMC The result of the connection request, used to handle loading state.
+ * @param {Object}   props.resultCreateAccount The result of the create account request.
+ * @param {Function} props.handleCreateAccount Callback function for creating a new Merchant Center account.
+ */
+const Actions = ( {
+	isConnected,
+	resultConnectMC,
+	resultCreateAccount,
+	handleCreateAccount,
```
eason9487: Prefer `onCreateAccount`.

js/src/components/google-mc-account-card/disconnect-account-button.js
```diff
@@ -0,0 +1,90 @@
...
+				'Disconnecting your Google Merchant Center account, please wait…',
+				'google-listings-and-ads'
+			)
+		);
+
+		try {
+			await fetchGoogleMCDisconnect();
+			invalidateResolution( 'getExistingGoogleMCAccounts', [] );
+			invalidateResolution( 'getGoogleMCAccount', [] );
+			onDisconnect();
```
eason9487: How about renaming `onDisconnect` to `onDisconnected` to more clearly convey that it's called after the disconnection is completed?

js/src/components/google-mc-account-card/disconnect-account-button.js
```diff
@@ -0,0 +1,90 @@
...
+				)
+			);
+		}
+
+		removeNotice( notice.id );
+	};
+
+	return (
+		<AppButton
+			isLink
```
eason9487: This prop could be removed.

js/src/components/google-mc-account-card/disconnect-account-button.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Renders a button to disconnect the current connected Google Merchant Center account.
+ *
+ * * @fires gla_mc_account_connect_different_account_button_click
+ */
+const DisconnectAccountButton = ( { onDisconnect = noop, ...restProps } ) => {
+	const { createNotice, removeNotice } = useDispatchCoreNotices();
+	const { invalidateResolution } = useAppDispatch();
+
+	const [ fetchGoogleMCDisconnect, { loading: loadingGoogleMCDisconnect } ] =
```
eason9487: I understood these names were moved from the existing code. However, these names were originally less than ideal. It would be appreciated if they could be renamed to something like:
- From `fetchGoogleMCDisconnect` to `disconnect`, `disconnectAccount`, or `disconnectGoogleMCAccount`
- From `loadingGoogleMCDisconnect` to `isDisconnecting`, `isDisconnectingAccount`, or `isDisconnectingGoogleMCAccount`

js/src/components/google-mc-account-card/non-connected.js
```diff
@@ -1,27 +0,0 @@
...
-
-	if ( ! hasFinishedResolution ) {
-		return <SpinnerCard />;
-	}
-
-	if ( existingAccounts.length > 0 ) {
-		return <ConnectMC />;
-	}
-
-	return <CreateAccount onSwitchAccount={ invalidateResolution } />;
```
eason9487, line 24: Two related components can be removed together:

- `js/src/components/google-mc-account-card/create-account-card.js`
- `js/src/components/google-mc-account-card/create-account.js`

js/src/components/google-mc-account-card/google-mc-account-card.js
```diff
@@ -1,38 +0,0 @@
...
-const GoogleMCAccountCard = () => {
-	const { hasFinishedResolution, isPreconditionReady, googleMCAccount } =
-		useGoogleMCAccount();
-
-	if ( ! hasFinishedResolution ) {
-		return <SpinnerCard />;
-	}
-
-	if ( ! isPreconditionReady ) {
-		return <DisabledCard />;
```
eason9487, line 19: `js/src/components/google-mc-account-card/disabled-card.js` can be removed as well.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -276,27 +291,284 @@ test.describe( 'Set up accounts', () => {
...
+						// Mock merchant center as not connected.
+						setUpAccountsPage.mockMCNotConnected(),
+
+						// Mock merchant center has accounts
+						setUpAccountsPage.mockMCHasAccounts(),
+					] );
+
+					await setUpAccountsPage.goto();
+				} );
+			} );
```
eason9487: This test group should be removed as it doesn't have any tests.

js/src/components/google-mc-account-card/connect-mc/index.js
asvinb: Update @eason9487

js/src/components/google-mc-account-card/connect-mc/index.js
asvinb: File removed.

js/src/components/google-mc-account-card/connect-mc/index.js
asvinb: I checked the designs and the wording is as per the designs:
https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7361-3088&node-type=frame&t=PAZXnb7jo194rJ9P-0

js/src/components/google-mc-account-card/connect-mc/actions.js
asvinb: 👍 Updated.

js/src/components/google-mc-account-card/disconnect-account-button.js
asvinb: `onDisconnect` renamed to `onDisconnected` 👍

js/src/components/google-mc-account-card/disconnect-account-button.js
asvinb: 👍

js/src/components/google-mc-account-card/disconnect-account-button.js
asvinb: - `fetchGoogleMCDisconnect` renamed to `disconnectGoogleMCAccount`
- `loadingGoogleMCDisconnect` renamed to `isDisconnectingGoogleMCAccount`

js/src/components/google-mc-account-card/non-connected.js
asvinb: Files removed.

js/src/components/google-mc-account-card/google-mc-account-card.js
asvinb: 👍

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: 👍

js/src/components/google-mc-account-card/connect-mc/index.js
```diff
@@ -37,101 +35,137 @@ import './index.scss';
...
-					resultConnectMC.error?.website_url ||
-					resultCreateAccount.error?.website_url
-				}
-				onSwitchAccount={ () => {
-					resultConnectMC.reset();
-					resultCreateAccount.reset();
-				} }
-			/>
-		);
+	if ( ! isGoogleMCReady ) {
```
eason9487, line 63: Wondering in which scenario this condition is required?

js/src/components/google-mc-account-card/connect-mc/index.js
eason9487: The screenshot was also taken from the design document: https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7448-22449&node-type=frame&t=NwWu4XTpGolsJr8R-0

I knew the current wording existed in the design document as well. That's why I said there is a different wording rather than the wording is incorrect. The question is to confirm which one should be used.

js/src/components/google-mc-account-card/connect-mc/index.js
asvinb: Consider the following:
* The user is editing the accounts and the `ConnectMC` component is rendered.
* User creates a new account and is prompted to claim the URL.
* User claims the URL.
* The responses from `resultConnectMC` and `resultCreateAccount` will still persist, i.e they have not been reset and the claim component will be rendered although the account has been successfully claimed.
* Thus, to prevent the claim components from being rendered, we make sure the account is connected because `resultConnectMC` and `resultCreateAccount` still contains the original response.

js/src/components/google-mc-account-card/connect-mc/index.js
asvinb: Asked for clarifications.

js/src/components/google-mc-account-card/connect-mc/index.js
eason9487: Thanks for the explanation! It makes sense.

eason9487 (approved):
Thanks for the work. LGTM.

There is only one https://github.com/woocommerce/google-listings-and-ads/pull/2639#discussion_r1828827878 left to be confirmed for now. Perhaps it needs a bit of tweaking and I believe that even then, it won't need another review. I will be approving in advance.

joemcgill:
I confirmed with @fblascogarma today that the current wording, "Required to sync products so they show on Google." is preferred. Thanks all!

Issue #2605, opened 2024-09-11
Onboarding: Support editing connected accounts in GoogleComboAccountsCard


joemcgill:
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2509

>[!Caution]
>Depends on #2596 and #2597.

While connecting accounts to the store, a merchant may need to edit one or more of the connected accounts being shown in the `GoogleComboAccountsCard`.

To achieve this, it is important to understand the three main visual states that the `GoogleComboAccountCard` supports:

1. **Compact** – In this state, all three accounts are represented in the top card body section. This state is used when automatically creating new accounts and after all accounts have been successfully connected.

![Image](https://github.com/user-attachments/assets/6a4ee4e5-65db-4f94-a452-33901f1370fc)

2. **Semi-expanded** – In this state, the top card body section represents multiple connected accounts (e.g., Google and Google Ads) but one account is represented in it's own section to allow for more complex interactions, like selecting from an existing account to connect.

![Image](https://github.com/user-attachments/assets/a0bae85b-57d9-4168-8a11-bf259528a01a)

3. **Fully-expanded** – This state is used when both the Ads and MC account need to be shown in their own sections, or when the user needs to edit any accounts that were already connected

> [!Warning]
> When fully expanded, the top card will include a "Cancel" button (not pictured)

![Image](https://github.com/user-attachments/assets/63bfa882-56d6-4ebc-9259-5446a8d9434c)

In either of the first two scenarios, the top section of the combo card should include an "Edit" link, which will switch the `GoogleComboAccountsCard` to the third, Fully-expanded state. 

In this state, the top section only represents the Google Account, and includes a link to connect a different Google Account. Below, the Ads and MC accounts will be shown in their own panels where accounts can be disconnected, connected, created, etc.

## Acceptance Criteria

- [ ] When multiple accounts are represented in the top section of the card, an "edit" button is shown
- [ ] Clicking the "edit" button causes the card to be shown in the fully-expanded state
- [ ] From the edited state the connected Google account can be changed (in scope of this ticket)
- [ ] From the edited state, Ads and MC account can be connected, disconnected, or created (if already implemented, otherwise out of scope)
- [ ] (Update: Nov 7) When the card is in it's edit state, a cancel button is shown to exit the edit state

## Implementation Brief
* Using the `GoogleComboAccountsCard` component,
	* Add a state variable to indicate whether the user clicked on the "Edit" button.
	* If this is the case,
[...]

Issue #2603, opened 2024-09-11
Onboarding: Create new Ads account when there are existing accounts in Google Accounts Card


joemcgill:
>[!Caution]
>Blocked by #2596  

Part of https://github.com/woocommerce/google-listings-and-ads/issues/2509

While connecting the store to Google Ads, a merchant may want to create a new account, even if they already have an existing Ads account associated with their Google account.

To support this, we will add a link below the account selector to kick off account creation.

![Image](https://github.com/user-attachments/assets/9d7595d6-67f6-4df9-8f16-e761a5baf9ad)

Clicking this link should show an "Are you sure" modal, which lets them initiate account creation or cancel (screenshot shows an example of the Merchant Center modal, copy for the Ads modal TBD).

![Image](https://github.com/user-attachments/assets/ea9bec77-3f89-40be-9f19-5436e3663333)

Once they confirm they want a new account, the account selector will replaced with a message that reads:

> **Creating a new Google Ads account**
> This may take a few minutes, please wait a moment...

And the Connect button will be replaced by a loading spinner with the word "Creating..."

Once the account is created, we'll display the Claim you account UI:

![Image](https://github.com/user-attachments/assets/562690d4-5953-421d-9696-b5cf3974e1da)

Once the new account is created and reclaimed, the card will show the new account as connected

![Image](https://github.com/user-attachments/assets/e24311c9-4f41-40a5-85eb-e1688251c9c1)

## Acceptance Criteria

- [x] When the Ads account selector panel is visible, the user can see the link to create a new account
- [x] Clicking the create link, displays the "Are you sure..." modal
- [x] If the user cancels, the modal closes and the account selector is still shown
- [x] If the user confirms they want to create a new account, the panel is replaced by the UI showing that the account is being created
- [x] Once the account is created, the UI to claim the account will be shown.
- [x] Once the account is created and claimed, the plugin finishes setting up the account and shows the Ads account as connected.
- [x] Once an account is connected, the create new link is replaced with a link to disconnect and connect a different account

## Implementation Brief
* Create the `ConfirmCreateModal` component in the same folder as the `ConnectAds` component created in #2596.
	* The modal should be similar to `js/src/components/google-mc-account-card/warning-modal/index.js` with updated copy.
	* Copy:
```
Title: Create Google Ads Account

⚠️ Are you sure you want to create a new Google Ads account?

You already have another Ads account associated with this Google account.

[...]

Issue #2567, opened 2024-08-23
Onboarding: Automatically kick off MC & Ads accounts creation


joemcgill:
As part of #2509, when a merchant connects a Google account that doesn't already have associated Google Merchant Center or Google Ads accounts, we'll streamline the process by creating the accounts for them and then prompt them to claim them.

![Image](https://github.com/user-attachments/assets/9fb3247c-4c98-4905-9392-38749b5fac95)

**Account creation copy:**

> You don’t have Merchant Center nor Google Ads accounts, so we’re creating them for you.
>
> _Merchant Center is required to sync products so they show on Google. Google Ads is required to set up conversion measurement for your store._

Once the accounts are created and accepted, the Google card will be shown as connected, and all three accounts will be displayed on the card as shown in the following screenshot.

![Image](https://github.com/user-attachments/assets/10ef4fbd-b1b6-46bc-a9d8-fce19d4ff2b0)

> [!NOTE]
> Handling the Ads account acceptance from this flow will be handled in a follow-up issue

## Acceptance Criteria

- [ ] Once a Google account is connected, the account creation copy is shown along with a spinner labeled "Creating..."
- [ ] An API request is made to the `/wc/gla/ads/accounts` to create an Ads account
- [ ] An API request is made to the `/wc/gla/mc/accounts` to create an MC account
- [ ] Once both accounts are created, a green "Connected" label (not pictured) should be shown and connected accounts should be displayed.

## Implementation Brief

> [!NOTE]
> This work should be branched from [feature/2566-google-combo-card](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2566-google-combo-card) (if not already merged) and a PR should be created against the [feature/2590-consolidate-google-account-cards](https://github.com/woocommerce/google-listings-and-ads/tree/feature/2590-consolidate-google-account-cards) branch.

[...]

PR #2640, opened 2024-10-04
Consolidate accounts onboarding: Connect to existing Google Ads account via the combo card

### LINKED ISSUE #2596: Onboarding: Connect to existing Ads account in Google Accounts Card

ankitrox:

Closes #2596  .

### Screenshots:

Screenshot 2024-10-04 at 10 37 36 PM

There are mainly two scenarios to test for this issue.

### Scenario 1
Both Google Ads and MC accounts are present for the Google account.

### Scenario 2
Google Ads exists, but no MC account is associated with the Google account.

Install [tweak extension](https://tweak-extension.com/docs/intro) in chromw browser to test both the scenarios.

#### Rules for scenario 1

**Rule 1:**

URL: https://domain-name.ngrok.app/index.php?rest_route=%2Fwc%2Fgla%2Fmc%2Fconnection&_locale=user  
Method: GET  
Delay: 2000ms  
Response: 
```js
{
  "id": 1234567,
  "currency": "USD",
  "symbol": "$",
  "status": "connected",
  "step": "",
}
```

**Rule 2:**

URL: https://domain-name.ngrok.app/index.php?rest_route=%2Fwc%2Fgla%2Fads%2Faccounts&_locale=user  
Method: GET  
Delay: 2000ms  
Response: 
```js
[
  {
    "id": 1234567,
    "subaccount": true,
    "domain": "https://domain-name.ngrok.app"
  }
]
```

#### Rules for scenario 2

**Rule 1:**

URL: https://domain-name.ngrok.app/index.php?rest_route=%2Fwc%2Fgla%2Fmc%2Fconnection&_locale=user  
Method: GET  
Delay: 2000ms  
Response: 
```js
{
  "id": 0,
  "currency": "USD",
  "symbol": "$",
  "status": "disconnected",
  "step": "",
}
```

**Rule 2:**

URL: https://domain-name.ngrok.app/index.php?rest_route=%2Fwc%2Fgla%2Fads%2Faccounts&_locale=user  
Method: GET  
Delay: 2000ms  
Response: 
```js
[]
```

### Steps to test

1. Only enable the scenario related rules in extension. Enable extension by clicking play button.

2. Go to Step 1 of the onboarding, make sure to connect with Google account which already have existing ads accounts associated.

3. Once Google account is connected, you should see `Connect to existing Google Ads account` card with the ads accounts in dropdown.

4. If there is only one account, the dropdown won't be able to expand.

5. Clicking the connect button will trigger connection process and will connect to that ads account. You should be able to see ads account ID in card description once connected successfully.

6. For scenario 2, you should be able to see that merchant account is being created. It will always be in creation status because we are mocking that state via extension.

7. Once the ads account is connected, the dropdown won't be able to expand. It will show the `Connected` label beside dropdown.

8. If ads account is connected, text in footer should be `Or, connect to a different Google Ads account`.


js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,84 @@
...
+	handleConnectClick,
+	isLoading,
+	setValue,
+	value,
+} ) => {
+	const initialValueRef = useRef( googleAdsAccount.id );
+	const getEventProps = useEventPropertiesFilter( FILTER_ONBOARDING );
+
+	useEffect( () => {
+		setValue( initialValueRef.current );
```
asvinb: @ankitrox If an Ads account is not connected, `googleAdsAccount.id` will return `0`, hence on mount, `value` would be `0`. Which means that we'll end with the "Connect" button being disabled.
image

I suggest we move this logic to `ConnectAds` where we set the value initially.

```js
if ( googleAdsAccount?.id ) {
	initialValueRef.current = googleAdsAccount.id;
}

useEffect( () => {
	if ( initialValueRef.current ) {
		setValue( initialValueRef.current );
	}
}, [ initialValueRef ] );
```

js/src/components/google-combo-account-card/account-creation-description/index.js
```diff
@@ -106,15 +106,15 @@ const AccountCreationDescription = ( {
 							'Merchant Center ID: %s',
 							'google-listings-and-ads'
 						),
-						googleMCAccount.id
+						googleMCAccount?.id
```
asvinb: Just FYI, this will updated in https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-8c885f578f99827b77511ccb568a741365e07f9fd3eecea8bc62ac67253ac3a3R126

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,84 @@
...
+	}, [ initialValueRef, setValue ] );
+
+	const ConnectCTA = () => {
+		if ( isConnected && initialValueRef.current === Number( value ) ) {
```
asvinb: @ankitrox I feel `isConnected` along is enough to display the `ConnectedIconLabel` component. Any idea why you want to check for the id as well?

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,84 @@
...
+	value,
+} ) => {
+	const initialValueRef = useRef( googleAdsAccount.id );
+	const getEventProps = useEventPropertiesFilter( FILTER_ONBOARDING );
+
+	useEffect( () => {
+		setValue( initialValueRef.current );
+	}, [ initialValueRef, setValue ] );
+
+	const ConnectCTA = () => {
```
asvinb: @ankitrox Do you think we can have `ConnectCTA` defined outside of the component body?

js/src/components/google-combo-account-card/connect-ads/connect-ads-footer.js
```diff
@@ -0,0 +1,27 @@
...
+import AppButton from '.~/components/app-button';
+
+/**
+ *
+ * @param {Object} props Props.
+ * @param {boolean} props.isLoading Whether the footer is in a loading state.
+ * @param {Function} props.onCreateNew Callback to handle the new ads account creation.
+ * @return {JSX.Element} Footer component.
+ */
+const ConnectAdsFooter = ( { isLoading, onCreateNew = () => {} } ) => (
```
asvinb:
```suggestion
const ConnectAdsFooter = ( { isLoading, onCreateNew = noop } ) => (
```
@ankitrox Are we actively using those props? If no, I suggest we remove them for now.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,124 @@
...
+
+	const isConnected =
+		googleAdsAccount?.status === 'connected' ||
+		( googleAdsAccount?.status === 'incomplete' &&
+			googleAdsAccount?.step === 'link_merchant' );
+
+	const [ value, setValue ] = useState();
+	const [ isLoading, setLoading ] = useState( false );
+	const [ fetchConnectAdsAccount ] = useApiFetchCallback( {
+		path: `/wc/gla/ads/accounts`,
```
asvinb:
```suggestion
		path: '/wc/gla/ads/accounts',
```

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,124 @@
...
+				'Connect to existing Google Ads account',
+				'google-listings-and-ads'
+			) }
+			helperText={ __(
+				'Required to set up conversion measurement for your store.',
+				'google-listings-and-ads'
+			) }
+			body={
+				<ConnectAdsBody
+					{ ...{
```
asvinb: @ankitrox I've not seen the pattern used in the codebase. Just to be consistent, can you list out the props please?

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -11,6 +11,7 @@ import AppSpinner from '../app-spinner';
 import useAutoCreateAdsMCAccounts from '../../hooks/useAutoCreateAdsMCAccounts';
 import LoadingLabel from '../loading-label/loading-label';
 import AccountCreationDescription from './account-creation-description';
+import { ConnectAds } from './connect-ads';
```
asvinb:
```suggestion
import ConnectAds from './connect-ads';
```

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -388,6 +391,133 @@ test.describe( 'Set up accounts', () => {
...
 
+	test.describe( 'Google Ads card', () => {
+		test.beforeAll( async () => {
+			// Mock Jetpack as connected
+			await setUpAccountsPage.mockJetpackConnected(
+				'Test user',
+				'jetpack@example.com'
+			);
+
+			await setUpAccountsPage.fulfillMCAccounts( [
```
asvinb:
```suggestion
			await setUpAccountsPage.mockMCHasAccounts()
```
@ankitrox Can we replace with the mock function we already have?

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,124 @@
...
+ * @event gla_ads_account_connect_button_click
+ * @property {number} id The account ID to be connected.
+ * @property {string} [context] Indicates the place where the button is located.
+ * @property {string} [step] Indicates the step in the onboarding process.
+ */
+
+/**
+ * ConnectAds component renders an account card to connect to an existing Google Ads account.
+ *
+ * @fires gla_ads_account_connect_button_click when "Connect" button is clicked.
```
asvinb: @ankitrox Do you think we should move that to `ConnectAdsBody` where we have the "Connect" button?

asvinb (changes requested):
Thanks @ankitrox . I left a few suggestions. Can you kindly check them out please?

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
ankitrox: @asvinb It checks whether:

1. Ads account is connected.
2. The currently selected ads account is the same which is connected.

In this case, we show `Connected` label instead of `Connect` button.

js/src/components/google-combo-account-card/connect-ads/connect-cta.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {number} props.id Account ID.
+ * @param {Function} props.handleConnectClick Callback to handle the connect click.
+ * @param {string} props.value Connected account ID.
+ *
+ * @return {JSX.Element} Connect CTA component.
+ */
+const ConnectCTA = ( { isConnected, id, handleConnectClick, value } ) => {
+	const getEventProps = useEventPropertiesFilter( FILTER_ONBOARDING );
+
+	if ( isConnected && id === Number( value ) ) {
```
asvinb: @ankitrox I don't think we need to check for `id === Number( value )` here because if the account is connected, the select should be read only. 
https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7662-6154&node-type=frame&t=X7RxteEeRgucqbnq-0

Meaning the only way to connect to another Ads account will be the user clicking the "Or, connect to another" . See conversation here:
https://github.com/woocommerce/google-listings-and-ads/issues/2603#issuecomment-2372873698

Existing implementation can be found here: `js/src/components/google-ads-account-card/disconnect-account.js` Maybe we can even re use the `DisconnectAccount` component.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: @ankitrox Just checking you saw my comment. thanks!

asvinb (changes requested):

js/src/components/google-combo-account-card/connect-ads/connect-cta.js
ankitrox: @asvinb I think we will need this check. Please refer [this design](https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7662-6154&node-type=frame&t=1KlDSTB27u712REa-0) in figma. If we click the `Edit` button, it expands both Existing MC and Existing Ads account cards where we would need to show the connected accounts.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
ankitrox: Sorry, missed that one. :)

I have updated now to use `mockMCHasAccounts`.

Thanks

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
asvinb: @ankitrox Not sure if am following here. When in connected state, the select is disabled, meaning the only way to change the Ads ID is by disconnecting the Ads account.
https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7662-6154&node-type=frame&t=FZW0ZlwmMV6RKllP-0

So if we just check for `isConnected` is fine.

js/src/components/google-combo-account-card/connect-ads/connect-cta.js
asvinb: @ankitrox Not sure if you forgot to push your changes after my comment . https://github.com/woocommerce/google-listings-and-ads/pull/2640#discussion_r1799425508
We need to do the following:
* In the connected state, we should disable the select as per the [designs](https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7662-6154&node-type=frame&t=FZW0ZlwmMV6RKllP-0).
* Add the "Or, connect to a different Google Ads account" button when in connected state. Check the [conversation here](https://github.com/woocommerce/google-listings-and-ads/issues/2603#issuecomment-2372873698).

ankitrox: Screenshot 2024-10-18 at 10 18 28 AM

1. Select is in disabled state, meaning only one option is present.
2. `'Or, connect to a different Google Ads account'` text is present when in Connected state.

js/src/components/ads-account-select-control/index.js
```diff
@@ -10,8 +10,9 @@ import useExistingGoogleAdsAccounts from '.~/hooks/useExistingGoogleAdsAccounts'
  */
 const AdsAccountSelectControl = ( props ) => {
 	const { existingAccounts } = useExistingGoogleAdsAccounts();
+	const accounts = props.accounts || existingAccounts;
```
asvinb: @ankitrox The component should not receive any accounts prop. Instead the accounts are resolved via the `useExistingGoogleAdsAccounts` hook.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,74 @@
...
+	value,
+} ) => {
+	const connectedAccount = accounts.filter( ( account ) => {
+		return isConnected && account.id === value;
+	} );
+
+	return (
+		<ContentButtonLayout>
+			<AdsAccountSelectControl
+				accounts={ isConnected ? connectedAccount : accounts }
```
asvinb: @ankitrox We should not pass the `accounts` prop. Instead we should pass the connected account ID to the `value` prop of `AdsAccountSelectControl`

js/src/components/google-combo-account-card/connect-ads/connect-ads-footer.js
```diff
@@ -0,0 +1,29 @@
...
+	const text = isConnected
+		? __(
+				'Or, connect to a different Google Ads account',
+				'google-listings-and-ads'
+		  )
+		: __(
+				'Or, create a new Google Ads account',
+				'google-listings-and-ads'
+		  );
+	return <AppButton isTertiary>{ text }</AppButton>;
```
asvinb: @ankitrox Let's render `js/src/components/google-ads-account-card/disconnect-account.js` if the account is connected. If you can integrate the logic from `js/src/components/google-ads-account-card/connect-ads/index.js` for the `Or, create a new Google Ads account` button, that would be great!

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,123 @@
...
+		method: 'POST',
+		data: { id: value },
+	} );
+	const { refetchGoogleAdsAccount } = useGoogleAdsAccount();
+	const { createNotice } = useDispatchCoreNotices();
+	const { fetchGoogleAdsAccountStatus } = useAppDispatch();
+	const initialValueRef = useRef( null );
+
+	if ( googleAdsAccount?.id ) {
+		initialValueRef.current = googleAdsAccount.id;
```
asvinb: @ankitrox Any reason why we are using ref instead of state variables here please? We can set the `value` variable with the following:
```js
useEffect( () => {
	if ( isConnected ) {
		setValue( googleAdsAccount.id );
	}
}, [ googleAdsAccount, isConnected ] );
```

js/src/components/google-combo-account-card/connect-ads/connect-cta.js
```diff
@@ -0,0 +1,47 @@
...
+		return <ConnectedIconLabel />;
```
asvinb: @ankitrox It's a bit misleading to have the `ConnectIconLabel` within the `ConnectCTA` component. Let's move the `ConnectedIconLabel` to be within `ConnectAdsBody` where it's rendered only when we have a connected Google Ads account.

js/src/components/google-combo-account-card/connect-ads/connect-cta.js
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param {Object} props Props.
+ * @param {boolean} props.isConnected Whether the account is connected.
+ * @param {number} props.id Account ID.
+ * @param {Function} props.handleConnectClick Callback to handle the connect click.
+ * @param {string} props.value Connected account ID.
+ *
+ * @return {JSX.Element} Connect CTA component.
+ */
+const ConnectCTA = ( { isConnected, id, handleConnectClick, value } ) => {
```
asvinb: @ankitrox We don't need the `id` prop since `value` should be enough to connect to a new account.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+import AdsAccountSelectControl from '.~/components/ads-account-select-control';
+import ContentButtonLayout from '.~/components/content-button-layout';
+import ConnectCTA from './connect-cta';
+import LoadingLabel from '.~/components/loading-label/loading-label';
+
+/**
+ * Clicking on the button to connect an existing Google Ads account.
+ *
+ * @event gla_ads_account_connect_button_click
```
asvinb: @ankitrox Should we move this to `ConnectCTA`?

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -3,12 +3,16 @@
  */
 import { __ } from '@wordpress/i18n';
 
+/**
```
asvinb: @ankitrox Duplicate comments.

js/src/components/google-combo-account-card/connect-ads/index.js
```diff
@@ -0,0 +1,5 @@
+/**
```
asvinb: @ankitrox There are no internal dependencies here.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,74 @@
...
+		return isConnected && account.id === value;
+	} );
+
+	return (
+		<ContentButtonLayout>
+			<AdsAccountSelectControl
+				accounts={ isConnected ? connectedAccount : accounts }
+				value={ value }
+				onChange={ setValue }
+				autoSelectFirstOption={ true }
```
asvinb: @ankitrox If there's a connected account, we can disable the select as such:
https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-c71501f6145dde774344794f6079c4d6c2f7199047c34fd0eba5484b220bde42R59

js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
```diff
@@ -4,10 +4,18 @@
 		margin: 0;
 	}
 
+	.components-select-control__input {
```
asvinb: @ankitrox We won't need this if we use this approach:
https://github.com/woocommerce/google-listings-and-ads/pull/2639/files#diff-c71501f6145dde774344794f6079c4d6c2f7199047c34fd0eba5484b220bde42R59

asvinb (changes requested):

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
ankitrox: If the accounts are resolved via `useExistingGoogleAdsAccounts` , and if there are more than one ads accounts, do you want me to use `disabled`  prop for `AdsAccountSelectControl` ?

js/src/components/ads-account-select-control/index.js
ankitrox: If the accounts are resolved via `useExistingGoogleAdsAccounts` , and if there are more than one ads accounts, do you want me to use `disabled`  prop for `AdsAccountSelectControl` ?

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
ankitrox: We are already passing the `value` prop with connected account ID [here](https://github.com/woocommerce/google-listings-and-ads/pull/2640/files#diff-d9020d12bf423927b8d5168f3e91c71888755909cdafaaafc7521a8bacb5d110R54)

js/src/components/google-combo-account-card/connect-ads/connect-ads-footer.js
ankitrox: Let's handle the `Or, create a new Google Ads account` separately in #2603

js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
ankitrox: Removed it.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
ankitrox: Added it. Also, added these props in constant file so that we can use at both the places and will have a single source as object.

js/src/components/ads-account-select-control/index.js
ankitrox: Handled it as mentioned in further comments.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,68 @@
...
+				{ ...( isConnected && NON_INTERACTABLE_PROPS ) }
+			/>
+			{ isLoading ? (
+				<LoadingLabel
+					text={ __( 'Connecting…', 'google-listings-and-ads' ) }
+				/>
+			) : (
+				<>
+					{ isConnected &&
+						googleAdsAccount.id === Number( value ) && (
```
asvinb:
```suggestion

```

asvinb: @ankitrox I think `isConnected` should be enough to display the connected label.

js/src/components/google-combo-account-card/connect-account-card.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * ConnectAccountCard component renders an account card with a title, helper text, body, and footer sections.
+ *
+ * @param {Object} props Props.
+ * @param {string} props.title The title of the account card.
+ * @param {string} props.helperText The helper text for the account card.
+ * @param {JSX.Element} props.body The content for the body of the account card.
+ * @param {JSX.Element} props.footer The content for the footer of the account card.
+ * @param {string} props.className Additional class names for the account card.
```
asvinb:
```suggestion
 * @param {string} [props.className] Additional class names for the account card.
```

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,68 @@
...
+							<ConnectedIconLabel />
+						) }
+					{ ! isConnected && (
+						<ConnectCTA
+							isConnected={ isConnected }
```
asvinb: @ankitrox `ConnectCTA` does not have the `isConnected` prop anymore.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,68 @@
...
+							id={ googleAdsAccount.id }
```
asvinb: @ankitrox Same as above, `ConnectCTA` no longer has the `id` prop.

js/src/components/google-combo-account-card/connect-ads/connect-cta.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Connect CTA component.
+ *
+ * @param {Object} props Props.
+ * @param {Function} props.handleConnectClick Callback to handle the connect click.
+ * @param {string} props.value Connected account ID.
+ * @fires gla_ads_account_connect_button_click when "Connect" button is clicked.
+ * @return {JSX.Element} Connect CTA component.
+ */
+const ConnectCTA = ( { handleConnectClick, value } ) => {
```
asvinb: @ankitrox Let's update `js/src/components/google-ads-account-card/connect-ads/index.js` to use `ConnectCTA`. I think it's better we have `ConnectCTA` in the following folder: `js/src/components/google-ads-account-card/connect-ads/`

js/src/constants.js
```diff
@@ -123,3 +123,12 @@ export const GOOGLE_WPCOM_APP_CONNECTED_STATUS = {
 	ERROR: 'error',
 	DISABLED: 'disabled',
 };
+
+export const NON_INTERACTABLE_PROPS = {
```
asvinb:
```suggestion
export const NON_INTERACTABLE_SELECT_PROPS = {
```
@ankitrox Let's rename the constant to `NON_INTERACTABLE_SELECT_PROPS` because `NON_INTERACTABLE_PROPS` is too generic.

asvinb (changes requested):

js/src/components/google-ads-account-card/connect-ads/index.js
```diff
@@ -131,17 +129,10 @@ const ConnectAds = ( props ) => {
...
-							eventProps={ getEventProps( {
-								id: Number( value ),
-							} ) }
-							onClick={ handleConnectClick }
-						>
-							{ __( 'Connect', 'google-listings-and-ads' ) }
-						</AppButton>
+						<ConnectCTA
+							handleConnectClick={ handleConnectClick }
+							value={ Number( value ) }
```
asvinb: @ankitrox I don't think we need `Number` here.
```suggestion
							value={ value }
```

js/src/components/google-ads-account-card/connect-ads/connect-cta.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Connect CTA component.
+ *
+ * @param {Object} props Props.
+ * @param {Function} props.handleConnectClick Callback to handle the connect click.
+ * @param {string} props.value Connected account ID.
+ * @fires gla_ads_account_connect_button_click when "Connect" button is clicked.
+ * @return {JSX.Element} Connect CTA component.
+ */
+const ConnectCTA = ( { handleConnectClick, value } ) => {
```
asvinb: @ankitrox To be consistent with the codebase, let's rename the component to `ConnectButton` and rename the file to `connect-button.js`

asvinb: Updated in this commit: https://github.com/woocommerce/google-listings-and-ads/pull/2640/commits/8a562d15c1b5ef969841d5f18d38bb92df40ed54

js/src/components/google-ads-account-card/connect-ads/index.js
asvinb: Updated here : https://github.com/woocommerce/google-listings-and-ads/pull/2640/commits/8a562d15c1b5ef969841d5f18d38bb92df40ed54#diff-6048a51e5a593289afa602e53ec8f9cbd2bf257f6724a874ddc26fe35a937e7dR134

asvinb:
@ankitrox The changes look good. Can you please check why the E2E tests are failing?
https://github.com/woocommerce/google-listings-and-ads/actions/runs/11463080793/job/31896064948
Thanks!

asvinb (approved):

ankitrox:
@asvinb Merged changes from #2567 and made sure that [E2E passes.](https://github.com/woocommerce/google-listings-and-ads/actions/runs/11464592212)

ankitguptaindia:
QA Status: ✅ 
---

**Test Results-:**  
Tested the PR using a real Google account with only Ads accounts (no MC account). After connecting the Google account, all ad accounts were displayed, and the merchant was able to successfully connect an account from the list.




Screenshot 2024-10-23 at 9 15 44 PM


- When the Google account has 1 or more existing Ads accounts, show the Ads connection UI ✅ 

- If only one account exists, it should automatically be selected. ✅ 

- If multiple accouns exists, the user should be able to select which to connect. ✅ 

- The user can click the "connect" button to connect the selected Ads account. ✅ 

- If there is no MC account and one is being created, the card description should only reference MC ✅ 


https://github.com/user-attachments/assets/c85040a6-8348-42d4-a13a-60bac9b3de69

ankitguptaindia:
@ankitrox While performing various tests on this PR, I noticed when user clicks on the button "Or, create a new Google Ads account"  to create new ads account instead of selecting an existing Ad account, this button doesn't give any response and user can not create new ads account from ads card. 

I believe this is not part of this PR changes. Please let me know if this issue is related to correct PR changes. 

Quick Demo: https://somup.com/cZ60DQHG21

ankitrox:
@ankitguptaindia Yes, it is part of #2603

ankitguptaindia (approved):

ankitrox:
@eason9487 This is over to you for review.

eason9487:
Hi @ankitrox, please make sure all automation tests have been passed before moving a PR to Woo CR status.

ankitrox:
@eason9487 Sorry, there was a css lint issue. I've fixed it and sending this to your way for reviewing.

js/src/components/app-select-control/index.js
```diff
@@ -52,12 +54,15 @@ const AppSelectControl = ( props ) => {
...
 	const hasSingleValueStyle = autoSelectFirstOption && options?.length === 1;
-	if ( hasSingleValueStyle ) {
+	if ( hasSingleValueStyle || nonInteractive ) {
 		selectProps = {
 			...selectProps,
 			suffix: ' ',
-			tabIndex: '-1',
+			style: {
+				pointerEvents: 'none',
+			},
```
eason9487: How about using the existing CSS by moving the `nonInteractive` condition to [line 56](https://github.com/woocommerce/google-listings-and-ads/blob/60a101a25b68bfb28e37764052f4918e99bbb966/js/src/components/app-select-control/index.js#L56) instead of adding a style doing the same thing?

https://github.com/woocommerce/google-listings-and-ads/blob/60a101a25b68bfb28e37764052f4918e99bbb966/js/src/components/app-select-control/index.scss#L7-L9

The postfix `--has-single-value` could be renamed to be closer to their stylistic meaning.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -22,14 +23,19 @@ const ConnectedGoogleComboAccountCard = () => {
...
 	return (
-		<AccountCard
-			appearance={ APPEARANCE.GOOGLE }
-			alignIcon="top"
-			className="gla-google-combo-account-card--connected"
-			description={ text || <AccountDetails /> }
-			helper={ subText }
-			indicator={ <Indicator showSpinner={ Boolean( creatingWhich ) } /> }
-		/>
+		<div className="gla-google-combo-account-card gla-account-card">
```
eason9487: Does it need these two CSS classes?

js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
```diff
@@ -4,3 +4,7 @@
 		gap: 0;
 	}
 }
+
+.gla-google-combo-service-connected-icon-label {
+	width: auto;
+}
```
eason9487: Where will this style be applied?

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,117 @@
...
+import ConnectAdsFooter from './connect-ads-footer';
+import ConnectAdsBody from './connect-ads-body';
+
+/**
+ * ConnectAds component renders an account card to connect to an existing Google Ads account.
+ *
+ * @fires gla_documentation_link_click with `{ context: 'setup-ads-connect-account', link_id: 'connect-sub-account', href: 'https://support.google.com/google-ads/answer/6139186' }`
+ * @return {JSX.Element} {@link AccountCard} filled with content.
+ */
+const ConnectAds = () => {
```
eason9487, line 28: Two questions may need to be confirmed:

1. From #2605 and the design document, the expanded UI state seems to be used after the Edit button is clicked or when there is a need to represent more complex interactions. However, this component is shown by default when all accounts have already been fully connected.

   https://github.com/user-attachments/assets/faacd529-2f92-4319-b6a1-b939634c9c93
2. From the design document, the expanded UI state will be switched to the compact state after the connection is completed.
   

   But this component is still shown after that.

   https://github.com/user-attachments/assets/fc9610b2-3e4b-4840-8322-0f60b653b1e7

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,117 @@
...
+import { useAppDispatch } from '.~/data';
+import useExistingGoogleAdsAccounts from '.~/hooks/useExistingGoogleAdsAccounts';
+import ConnectAccountCard from '../connect-account-card';
+import ConnectAdsFooter from './connect-ads-footer';
+import ConnectAdsBody from './connect-ads-body';
+
+/**
+ * ConnectAds component renders an account card to connect to an existing Google Ads account.
+ *
+ * @fires gla_documentation_link_click with `{ context: 'setup-ads-connect-account', link_id: 'connect-sub-account', href: 'https://support.google.com/google-ads/answer/6139186' }`
```
eason9487: This component doesn't appear to fire this event tracking.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,117 @@
...
+
+	const {
+		googleAdsAccount,
+		hasFinishedResolution: hasFinishedResolutionForCurrentAccount,
+	} = useGoogleAdsAccount();
+
+	const isConnected =
+		googleAdsAccount?.status === 'connected' ||
+		( googleAdsAccount?.status === 'incomplete' &&
+			googleAdsAccount?.step === 'link_merchant' );
```
eason9487: With an account that is connected but not yet set up billing, this component still asks for connecting to an account.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,117 @@
...
+
+	const [ value, setValue ] = useState();
+	const [ isLoading, setLoading ] = useState( false );
+	const [ fetchConnectAdsAccount ] = useApiFetchCallback( {
```
eason9487: Using another name that can semantically convey that it's a function to connect a Google Ads account would be nicer.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,117 @@
...
+			body={
+				<ConnectAdsBody
+					isConnected={ isConnected }
+					handleConnectClick={ handleConnectClick }
+					isLoading={ isLoading }
+					setValue={ setValue }
+					value={ value }
+				/>
+			}
+			footer={ <ConnectAdsFooter isConnected={ isConnected } /> }
```
eason9487: It would be more robust to disable this button during connecting.

image

js/src/components/google-ads-account-card/connect-ads/connect-button.js
```diff
@@ -0,0 +1,49 @@
...
+ * Clicking on the button to connect an existing Google Ads account.
+ *
+ * @event gla_ads_account_connect_button_click
+ * @property {number} id The account ID to be connected.
+ * @property {string} [context] Indicates the place where the button is located.
+ * @property {string} [step] Indicates the step in the onboarding process.
+ */
+
+/**
+ * Connect CTA component.
```
eason9487: Should this button be considered a CTA button?

js/src/components/google-ads-account-card/connect-ads/connect-button.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param {Object} props Props.
+ * @param {Function} props.handleConnectClick Callback to handle the connect click.
+ * @param {string} props.value Connected account ID.
```
eason9487: This description is not precise.

js/src/components/google-ads-account-card/connect-ads/connect-button.js
```diff
@@ -0,0 +1,49 @@
...
+ * @fires gla_ads_account_connect_button_click when "Connect" button is clicked.
+ * @return {JSX.Element} Connect CTA component.
+ */
+const ConnectButton = ( { handleConnectClick, value } ) => {
```
eason9487: - A component prop name to receive the handler of an event would be better to start with `on`. In this case, directly forwarding the rest props or renaming it to `onClick` would be the most suitable.
- Maybe the `value` prop could be renamed to convey that it is expected to be an account ID.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,61 @@
...
+ * @param {boolean} props.isConnected Whether the account is connected.
+ * @param {Function} props.handleConnectClick Callback to handle the connect click.
+ * @param {boolean} props.isLoading Whether the card is in a loading state.
+ * @param {Function} props.setValue Callback to set the value.
+ * @param {string} props.value Ads account ID.
+ * @return {JSX.Element} Body component.
+ */
+const ConnectAdsBody = ( {
+	isConnected,
+	handleConnectClick,
```
eason9487: A component prop name to receive the handler of an event would be better to start with `on`

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
```diff
@@ -0,0 +1,61 @@
...
+	isLoading,
+	setValue,
```
eason9487: A component prop name to receive the handler of an event would be better to start with `on`

js/src/components/google-combo-account-card/connect-account-card.scss
```diff
@@ -0,0 +1,33 @@
...
+		}
+	}
+
+	.gla-google-combo-service-account-card__footer {
+		border: 0;
+		padding: 0;
+
+		.app-button,
+		.components-button.is-tertiary:hover:not(:disabled,[aria-disabled="true"]) {
+			background: none;
```
eason9487: Why does it need to invalidate the hover visual change?

js/src/components/google-ads-account-card/connect-ads/index.js
```diff
@@ -131,17 +129,10 @@ const ConnectAds = ( props ) => {
 							) }
 						/>
 					) : (
-						<AppButton
-							isSecondary
-							disabled={ ! value }
-							eventName="gla_ads_account_connect_button_click"
-							eventProps={ getEventProps( {
-								id: Number( value ),
-							} ) }
```
eason9487, line 140: It would be nicer to remove the following JSDoc together.

https://github.com/woocommerce/google-listings-and-ads/blob/b449092262e103bb8c5577871d8792674b33ee7e/js/src/components/google-ads-account-card/connect-ads/index.js#L26-L33

https://github.com/woocommerce/google-listings-and-ads/blob/b449092262e103bb8c5577871d8792674b33ee7e/js/src/components/google-ads-account-card/connect-ads/index.js#L36

eason9487 (changes requested):
There are some issues that need to be addressed or confirmed.

I didn't view the code changes related to e2e testing in detail since they were continuously pushed commits while I was reviewing.

---

## About `ConnectAccountCard` component

Beyond that, I would like to propose modifying the `AccountCard` component so that it supports the layout required by `GoogleComboAccountCard`, in place of the `ConnectAccountCard` component in this PR and the respective customized implementations in the related PRs.

### Scope to be supported by `AccountCard`

The target is to support these needs

1. It renders additional content and bottom links as well as enabling the indicator to align to additional content.
   image
2. It enables the layout of additional content to be extended to the right edge.
   image
3. The additional content it renders can also be used for this purpose, but only for layouts. The part inside the bordered area is a separate implementation, but it might be possible to try to reuse a similar implementation in `AccountCard` via the relevant PR.
   image

### The following `git diff` result shows how the `AccountCard` could be modified.

```diff
diff --git a/js/src/components/account-card/index.js b/js/src/components/account-card/index.js
index bebf7df10..a55370b42 100644
--- a/js/src/components/account-card/index.js
+++ b/js/src/components/account-card/index.js
@@ -125,6 +125,11 @@ const alignStyleName = {
 	top: `gla-account-card__styled--align-top`,
 };
 
+const indicatorAlignStyleName = {
+	...alignStyleName,
+	toDetail: 'gla-account-card__indicator--align-to-detail',
+};
+
 /**
  * Renders a Card component with account info and status.
  *
@@ -156,12 +161,16 @@ export default function AccountCard( {
 	alignIcon = 'center',
 	indicator,
 	alignIndicator = 'center',
+	detail,
+	expandedDetail = false,
+	actions,
 	children,
 	...restProps
 } ) {
 	const cardClassName = classnames(
 		'gla-account-card',
 		disabled ? 'gla-account-card--is-disabled' : false,
+		expandedDetail ? 'gla-account-card--is-expanded-detail' : false,
 		className
 	);
 
@@ -172,19 +181,15 @@ export default function AccountCard( {
 
 	const indicatorClassName = classnames(
 		'gla-account-card__indicator',
-		alignStyleName[ alignIndicator ]
+		indicatorAlignStyleName[ alignIndicator ]
 	);
 
 	return (
 		<Section.Card className={ cardClassName } { ...restProps }>
 			<Section.Card.Body>
-				<Flex gap={ 4 }>
-					{ icon && (
-						<FlexItem className={ iconClassName }>
-							{ icon }
-						</FlexItem>
-					) }
-					<FlexBlock>
+				<div className="gla-account-card__body-layout">
+					{ icon && <div className={ iconClassName }>{ icon }</div> }
+					<div className="gla-account-card__subject">
 						{ title && (
 							<Subsection.Title className="gla-account-card__title">
 								{ title }
@@ -200,13 +205,23 @@ export default function AccountCard( {
 								{ helper }
 							</div>
 						) }
-					</FlexBlock>
+					</div>
+					{ detail && (
+						<div className="gla-account-card__detail">
+							{ detail }
+						</div>
+					) }
 					{ indicator && (
-						<FlexItem className={ indicatorClassName }>
+						<div className={ indicatorClassName }>
 							{ indicator }
-						</FlexItem>
+						</div>
+					) }
+					{ actions && (
+						<div className="gla-account-card__actions">
+							{ actions }
+						</div>
 					) }
-				</Flex>
+				</div>
 			</Section.Card.Body>
 			{ children }
 		</Section.Card>
diff --git a/js/src/components/account-card/index.scss b/js/src/components/account-card/index.scss
index e33e2b5fd..4a0b96ffa 100644
--- a/js/src/components/account-card/index.scss
+++ b/js/src/components/account-card/index.scss
@@ -5,18 +5,60 @@
 		opacity: 0.5;
 	}
 
+	&--is-expanded-detail {
+		.gla-account-card__indicator {
+			grid-area: 1/3;
+		}
+
+		.gla-account-card__detail {
+			grid-area: 2/2/auto/span 2;
+		}
+	}
+
 	&__styled {
 		&--align-top {
 			align-self: flex-start;
 		}
 	}
 
+	&__body-layout {
+		display: grid;
+		grid-template-columns: auto 1fr auto;
+		align-items: center;
+	}
+
 	&__icon {
+		grid-area: 1/1/span 2;
+		margin-right: $grid-unit-20;
 		line-height: 0;
 	}
 
+	&__subject {
+		grid-area: 1/2;
+	}
+
+	&__indicator {
+		grid-area: 1/3/span 2;
+		margin-left: $grid-unit-20;
+
+		&--align-to-detail {
+			grid-area: 2/3;
+			margin-top: $grid-unit-15;
+		}
+	}
+
+	&__detail {
+		grid-area: 2/2;
+		margin-top: $grid-unit-15;
+	}
+
+	&__actions {
+		grid-area: 3/2/auto/span 2;
+		margin-top: $grid-unit-15;
+	}
+
 	&__title {
-		margin-bottom: $grid-unit-05;
+		margin: 0;
 		color: $black;
 	}
 
@@ -25,6 +67,7 @@
 		flex-direction: column;
 		align-items: flex-start;
 		gap: 1em;
+		margin-top: $grid-unit-05;
 		color: $gray-900;
 
 		> p {
@@ -65,11 +108,12 @@
 	}
 
 	@media (max-width: $break-small) {
-		.components-card__body > .components-flex {
+		&__body-layout {
+			display: flex;
 			flex-direction: column;
 			align-items: flex-start;
-			> .components-flex-block,
-			> .components-flex-item {
+
+			> div {
 				margin: $grid-unit-10;
 			}
 		}
```

### The following `git diff` result can demonstrate roughly how this solution would work with the current implementation in this PR.

```diff
diff --git a/js/src/components/google-combo-account-card/connect-ads/connect-ads.js b/js/src/components/google-combo-account-card/connect-ads/connect-ads.js
index 537526e0e..49e1a9918 100644
--- a/js/src/components/google-combo-account-card/connect-ads/connect-ads.js
+++ b/js/src/components/google-combo-account-card/connect-ads/connect-ads.js
@@ -13,9 +13,11 @@ import useDispatchCoreNotices from '.~/hooks/useDispatchCoreNotices';
 import useGoogleAdsAccount from '.~/hooks/useGoogleAdsAccount';
 import { useAppDispatch } from '.~/data';
 import useExistingGoogleAdsAccounts from '.~/hooks/useExistingGoogleAdsAccounts';
-import ConnectAccountCard from '../connect-account-card';
+import AdsAccountSelectControl from '.~/components/ads-account-select-control';
+import ConnectButton from '.~/components/google-ads-account-card/connect-ads/connect-button';
+import LoadingLabel from '.~/components/loading-label/loading-label';
+import ConnectedIconLabel from '.~/components/connected-icon-label';
 import ConnectAdsFooter from './connect-ads-footer';
-import ConnectAdsBody from './connect-ads-body';
 
 /**
  * ConnectAds component renders an account card to connect to an existing Google Ads account.
@@ -89,27 +91,47 @@ const ConnectAds = () => {
 		return null;
 	}
 
+	const getIndicator = () => {
+		if ( isLoading ) {
+			return (
+				<LoadingLabel
+					text={ __( 'Connecting…', 'google-listings-and-ads' ) }
+				/>
+			);
+		}
+		if ( isConnected ) {
+			return <ConnectedIconLabel />;
+		}
+		return (
+			<ConnectButton
+				handleConnectClick={ handleConnectClick }
+				value={ value }
+			/>
+		);
+	};
+
 	return (
-		<ConnectAccountCard
-			className="gla-google-combo-service-account-card--ads"
+		<AccountCard
+			className="gla-google-combo-account-card gla-google-combo-service-account-card--ads"
 			title={ __(
 				'Connect to existing Google Ads account',
 				'google-listings-and-ads'
 			) }
-			helperText={ __(
+			helper={ __(
 				'Required to set up conversion measurement for your store.',
 				'google-listings-and-ads'
 			) }
-			body={
-				<ConnectAdsBody
-					isConnected={ isConnected }
-					handleConnectClick={ handleConnectClick }
-					isLoading={ isLoading }
-					setValue={ setValue }
+			alignIndicator="toDetail"
+			indicator={ getIndicator() }
+			detail={
+				<AdsAccountSelectControl
 					value={ value }
+					onChange={ setValue }
+					autoSelectFirstOption
+					nonInteractive={ isConnected }
 				/>
 			}
-			footer={ <ConnectAdsFooter isConnected={ isConnected } /> }
+			actions={ <ConnectAdsFooter isConnected={ isConnected } /> }
 		/>
 	);
 };
diff --git a/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss b/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
index d68a6f461..fd8960ac3 100644
--- a/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
+++ b/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
@@ -8,3 +8,10 @@
 .gla-google-combo-service-connected-icon-label {
 	width: auto;
 }
+
+// This CSS would be better placed in index.scss file and be imported from index.js
+.gla-google-combo-account-card {
+	.gla-account-card__actions .app-button {
+		margin-left: calc(var(--main-gap) / -2);
+	}
+}
```

📌 Please note that these code changes still need to be polished.

### A sample use I temporarily did locally

image

js/src/components/app-select-control/index.js
ankitrox: @eason9487 Changed the condition to 

```js
	const isNonInteractive =
		( autoSelectFirstOption && options?.length === 1 ) || nonInteractive;
```

and changed postfix to `.app-select-control--is-non-interactive`.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
ankitrox: Removed `gla-google-combo-account-card` as it is not being used anywhere.

js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
ankitrox: Does not seem to be required. Removed.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
ankitrox: Added the condition in component so that it will only be visible when account is not connected or when account is connected, but editing.

```js
if ( isConnected && ! isEditing ) {
    return null;
}
```

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
ankitrox: Removed.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
ankitrox: Renamed it to `connectGoogleAdsAccount`.

js/src/components/google-ads-account-card/connect-ads/connect-button.js
ankitrox: Updated the doc.

js/src/components/google-ads-account-card/connect-ads/connect-button.js
ankitrox: Changed it to `Connected Google Ads account ID.`

js/src/components/google-combo-account-card/connect-account-card.scss
ankitrox: Screenshot-1

With padding, but not aligned properly.

Screenshot 2024-10-29 at 10 06 36 PM

Screenshot-2

Without padding, but aligned properly.

Screenshot 2024-10-29 at 10 08 00 PM

@eason9487 Please refer to the attached images.  If we add the hover effect, there will be some inconsistencies in padding and alignment.

js/src/components/google-ads-account-card/connect-ads/index.js
ankitrox: Removed.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
ankitrox: Updated it to use `useGoogleAdsAccountReady` hook which also takes into account about `billing` step.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
joemcgill: This makes sense to me and the isEditing prop can be incorporated in #2605.

joemcgill:
> Beyond that, I would like to propose modifying the AccountCard component so that it supports the layout required by GoogleComboAccountCard, in place of the ConnectAccountCard component in this PR and the respective customized implementations in the related PRs.

@eason9487 this is a good idea in general, but not something that we considered necessary to implement the designs for #2509. At this point, a bigger refactor of this component would also affect #2597, where we would also need to consider cases where the `AccountCard` used to connect/edit existing MC accounts makes use of other components that render `AccountCard` components that I think would result in rendering nested `Card` components.

I'd hope that we can get this new `GoogleComboAccountCard` component merged using the current structure proposed in this PR and [#2639](https://github.com/woocommerce/google-listings-and-ads/pull/2639) and consider larger structural refactors of the `AccountCard` component separately.

js/src/components/app-select-control/index.js
```diff
@@ -51,20 +53,20 @@ const AppSelectControl = ( props ) => {
...
 
-	const hasSingleValueStyle = autoSelectFirstOption && options?.length === 1;
-	if ( hasSingleValueStyle ) {
+	const isNonInteractive =
+		( autoSelectFirstOption && options?.length === 1 ) || nonInteractive;
+	if ( isNonInteractive ) {
 		selectProps = {
 			...selectProps,
 			suffix: ' ',
-			tabIndex: '-1',
```
eason9487, line 59: Setting `tabIndex` to -1 is necessary to prevent it from interacting via keyboard.

https://github.com/user-attachments/assets/d76a653e-b6c2-4a3f-ae7d-1b3a17681a39

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
eason9487: Wonder why it still needs the `.gla-account-card` CSS class?

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
```diff
@@ -0,0 +1,123 @@
...
+
+	useEffect( () => {
+		if ( isConnected ) {
+			setValue( googleAdsAccount.id );
+		}
+	}, [ googleAdsAccount, isConnected ] );
+
+	if ( isConnected && ! isEditing ) {
+		return null;
+	}
```
eason9487: How about moving this and [another](https://github.com/woocommerce/google-listings-and-ads/blob/2724b0fd25a7a5d21d671dc943135b5bd9de561e/js/src/components/google-combo-account-card/connect-ads/connect-ads.js#L84-L91) rendering controls to `ConnectedGoogleComboAccountCard`?

Because it can be expected that other cards from different scenarios may also be placed in `ConnectedGoogleComboAccountCard`, which may end up being somewhat like the following, and this makes it look like all the cards are going to be displayed at the same time.

```jsx
<div>
	<AccountCard
		// ... omitted
	/>
	<ConnectAds />
	<CreateAds />
	<ConnectMC />
	<CreateMC />
</div>
```

js/src/components/google-combo-account-card/connect-account-card.scss
eason9487: There are other element states can make the button's left padding look improperly.

- `:active`
   image
- `:focus`
   image

Adding more style overrides may not be ideal. The proposed solution in https://github.com/woocommerce/google-listings-and-ads/pull/2640#pullrequestreview-2398095640 includes a trick for aligning this button.

```diff
+// This CSS would be better placed in index.scss file and be imported from index.js
+.gla-google-combo-account-card {
+	.gla-account-card__actions .app-button {
+		margin-left: calc(var(--main-gap) / -2);
+	}
+}
```

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
eason9487: Is this missed or is there a reason to keep the naming?

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -269,8 +274,102 @@ test.describe( 'Set up accounts', () => {
 				).toBeVisible();
 
 				await expect(
-					googleAccountCard.getByText( 'Connected', { exact: true } )
```
eason9487: Looks like there's no need to replace this `googleAccountCard` locator.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -269,8 +274,102 @@ test.describe( 'Set up accounts', () => {
...
+			} );
+		} );
+	} );
+
+	test.describe( 'Google Ads card', () => {
+		test.beforeAll( async () => {
+			// Mock Jetpack as connected
+			await setUpAccountsPage.mockJetpackConnected(
+				'Test user',
+				'jetpack@example.com'
```
eason9487: This data are not the test targets and could be omitted and let it fall back to the default values.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -269,8 +274,102 @@ test.describe( 'Set up accounts', () => {
...
+						'Required to set up conversion measurement for your store.',
+						{ exact: true }
+					)
 				).toBeVisible();
+
+				await expect(
+					googleAdsAccountCard.getByRole( 'button', {
+						name: 'Connect',
+					} )
+				).toBeEnabled();
```
eason9487: Maybe delete this assertion since the next test case has the same one.

https://github.com/woocommerce/google-listings-and-ads/blob/2724b0fd25a7a5d21d671dc943135b5bd9de561e/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L347-L351

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -269,8 +274,102 @@ test.describe( 'Set up accounts', () => {
...
+			} );
+
+			test( 'should see the button as enabled when selects the account from dropdown', async () => {
+				const googleAdsAccountCard =
+					setUpAccountsPage.getGoogleAdsAccountCard();
+
+				const adsAccountDropdown =
+					googleAdsAccountCard.locator( 'select' );
+				await adsAccountDropdown.selectOption( '111111' );
```
eason9487: Now the new implementation will automatically select the first account, so perhaps this can select another one.

https://github.com/woocommerce/google-listings-and-ads/blob/2724b0fd25a7a5d21d671dc943135b5bd9de561e/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L295-L308

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -269,8 +274,102 @@ test.describe( 'Set up accounts', () => {
...
+
+				await expect(
+					googleAdsAccountCard.getByRole( 'button', {
+						name: 'Connect',
+					} )
+				).toBeEnabled();
+			} );
+
+			test( 'should send an API request to connect existing Google Ads account', async () => {
```
eason9487: This teas case doesn't assert if there is an API request is sent. It would be better to either adjust the test description or add the corresponding assertion.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -269,8 +274,102 @@ test.describe( 'Set up accounts', () => {
...
+				const once = setUpAccountsPage.fulfillTimes( 1 );
+				await once.fulfillAdsAccounts( { id: 12345 } );
+				await once.mockAdsAccountConnected();
+				await once.mockAdsStatusClaimed();
+
+				await googleAdsAccountCard
+					.getByRole( 'button', { name: 'Connect' } )
+					.click();
+
+				expect( googleAccountCard ).toContainText(
```
eason9487: It should `await` the assertion. Otherwise, it will end this test case immediately without waiting for this assertion result.

eason9487 (changes requested):

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
asvinb: @eason9487 I've removed the unused class name. I've kept the surrounding div though, so that the cards are not directly children of `gla-vertical-gap-layout__large` where there'll be a gap between the cards.

js/src/components/google-combo-account-card/connect-ads/connect-ads-body.js
asvinb: @eason9487 This is no longer applicable since we no longer have the `ConnectAdsBody` component.

js/src/components/google-combo-account-card/connect-account-card.scss
asvinb: @eason9487 I've implemented your trick. 👍

js/src/components/app-select-control/index.js
asvinb: @eason9487 Added `tabIndex`. I've also added `pointerEvents: none` to the styles directly which will avoid having a class name specifically for the non interactive mode.

js/src/components/google-combo-account-card/connect-ads/connect-ads.js
asvinb: @eason9487 I've moved the conditions to render the card in `ConnectedGoogleComboAccountCard`.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: 👍

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Sounds good!

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Assertion removed @eason9487

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Used `222222`.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: @eason9487 Test description updated.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Thanks! Updated!

js/src/components/app-select-control/index.js
```diff
@@ -51,22 +53,22 @@ const AppSelectControl = ( props ) => {
...
-	if ( hasSingleValueStyle ) {
+	const isNonInteractive =
+		( autoSelectFirstOption && options?.length === 1 ) || nonInteractive;
+	if ( isNonInteractive ) {
 		selectProps = {
 			...selectProps,
+			readOnly: true,
+			style: {
+				pointerEvents: 'none',
+			},
```
eason9487: Setting `pointer-events: none;` onto the `SelectControl` is different from to its wrapper. `SelectControl` component has the `label` prop allowing focus to the associated `<select>` element and then interacting it via keyboard.

https://github.com/user-attachments/assets/4af0962a-c6ed-429c-bdd8-f1cdf6d1e300

The original way also covers the `pointer-events: none;` to `<label>` element in order to avoid the issue. Although it won't be a problem in the scope of this PR, from the point of view of the component design and the fact that `AppSelectControl` is also a shared component, it should be reverted back to the original solution (but not including the concern in https://github.com/woocommerce/google-listings-and-ads/pull/2640#discussion_r1818466247).

asvinb: Sounds good. Updated @eason9487 https://github.com/woocommerce/google-listings-and-ads/pull/2640/commits/4cba145d913b80d949448ec3a32ee44afcc7d43a

js/src/components/app-select-control/index.js
asvinb: Reverted that change. 
https://github.com/woocommerce/google-listings-and-ads/pull/2640/commits/4cba145d913b80d949448ec3a32ee44afcc7d43a

eason9487 (approved):
Thanks for the work. LGTM.

Issue #2538, opened 2024-08-16
Show Ads value prop in Paid Campaign performance card


joemcgill:
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2460

In order to better explain why someone should set up their first PMax campaign, we'll update the Paid Campaign card in the dashboard to show a similar value prop as the one shown during onboarding.

<img width="556" alt="image" src="https://github.com/user-attachments/assets/3c2e63d9-bcb2-4cd8-a747-78960c0c4341">

## Acceptance Criteria

- [ ] Display the Paid Ads features value prop card when:
  - [ ] the connected Ads account has not completed the setup flow ([see discussion](https://github.com/woocommerce/google-listings-and-ads/pull/2641#issuecomment-2438579282)), or
  - [ ] if the Ads account does not have any existing campaigns
- [ ] If the account is eligible to see the $500 ads credit (see: #2599), include the banner.
- [ ] Clicking the Create Campaign button should open the same Create Campaign flow as the current Paid Campaign card.
- [ ] Once a campaign has been created, the Paid Campaign card should show performance metrics from the current card.

## Implementation Brief
* Create `js/src/dashboard/summary-section/paid-features.js` which will house the `PaidFeatures` component.
	* Render the component as per the ticket description.
	* The paid features are also found in `js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js`
		* Extract the features into a common component which can be used for both, `PaidAdsFeaturesSection` and `PaidFeatures`.
	* For the $500 as credit implementation, implementation details can be taken from `js/src/setup-mc/setup-stepper/setup-paid-ads/paid-ads-features-section.js`
	* Button for "Create campaign" should possibly reuse `js/src/components/paid-ads/add-paid-campaign-button.js`.
		* Necessary updates should be make to the componen so that it can be used across different components.
		* Possibly mode the component to `js/src/component`
* Using `js/src/dashboard/summary-section/index.js`,
	* Add `PaidFeatures` to `SummarySection` if
		* `adsSetupComplete` is `true`
		* and there are no campaign.
			* To check whether there are no campaigns, use the `useAdsCampaigns`
* We will deprecate the `useFreeAdCredit()` while implementing this, see #2599. Instead of any places where that hook is currently in use should be updated so that the Ads Credit banner is always shown when there are no existing campaigns. 

## Test Coverage
[...]

PR #2641, opened 2024-10-10
Improve post-onboarding: Show promotion for Google Ads campaign on the Product Feed page

### LINKED ISSUE #2539: Show promo for setting up Google Ads on the product feed tabs

Once a merchant has approved products in their product feed, we want to make it easy for them to start running ads for those products by adding a promo for creating their first campaign on the product feed overview card.

image

## Acceptance Criteria

- [ ] Show the create campaign promo/notice in the overview card when the store has approved products, but has not yet created any campaigns.
- [ ] Clicking on the Create Campaign button enters the campaign creation flow.
- [ ] If there are no approved products, or the connected Ads account already has campaigns running, the notice should not be shown.

## Implementation Brief
* Using `js/src/product-feed/product-statistics`,
	* Create the `create-campaign-notice.js` which will house the `CreateCampaignNotice` component.
* Within `CreateCampaignNotice`,
	* Use the `useMCProductStatistics` hook to determine whether there are approved products.
	* Check if there are any campaigns via the `useAdsCampaigns` hook (see [this update](https://github.com/woocommerce/google-listings-and-ads/issues/2539#issuecomment-2397847396) to ensure that hook always checks for campaigns when a Google Ads account is connected).
	* If there are no approved products or there are campaigns, render `null`.
	* If there are approved products but no campaigns, render the notice as per the ticket description.
		* Make use of the `Notice` component with custom class name for styling.
	* Clicking the "Create Campaign" button should redirect the user to create a campaign.
		* Use the `js/src/components/paid-ads/add-paid-campaign-button.js` component, but pass a new context to the eventProps called `product-statistics-card`.
* Update `js/src/product-feed/product-statistics/index.js` to include `CreateCampaignNotice`

## Test Coverage
* Appropriate E2E tests should be added to test when the notice is displayed/not displayed.

## Definition Questions
* In the AC, there is mention of approved products but not specifically mentioning how many products. Does that mean that if the user has at least one approved product and has not running campaigns, we should show the notice? Or should all the products be approved? A: This should apply as long as at least one product is approved.
* @michaeleleder mentioned that this Notice style is an existing "inline notice" but I don't already see a component for this style in this repo. Should it be pulled from elsewhere, or should we create a custom `AppNotice` and style for this project?

kt-12:

Closes #2539 

####  When there is Product but no campaign:  ( Show the notice )
no-campaign
no-campaign-yes-notice

#### When there is Product and a campaign:  ( Don't show the notice )
campaign
when-there-is-campaign

1. Create a Simple Product, sync to merchant account.
2. Visit Dashboard > Product Feed
3. If there is one active ( approved or partially approved) product and no campaign, notice will be shown. Check screenshots.
4. Click on the button and add campaign. Once added revisit Product Feed page and this section should disappear.

### Additional details:
Further more this ticket address a bug in `useAdsCampaign` https://github.com/woocommerce/google-listings-and-ads/issues/2539#issuecomment-2397847396, which required certain E2E to be adjusted. 
 

### Changelog entry

> Add - Show promotion for Google Ads campaign on the Product Feed page.


js/src/product-feed/product-statistics/create-campaign-notice.js
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Internal dependencies
+ */
+import AddPaidCampaignButton from '.~/components/paid-ads/add-paid-campaign-button';
+import useMCProductStatistics from '.~/hooks/useMCProductStatistics';
+import useAdsCampaigns from '.~/hooks/useAdsCampaigns';
+
+const CreateCampaignNotice = () => {
+	const { hasFinishedResolution: productsResolution, data: products } =
```
joemcgill: Renaming the `hasFinishedResolution` prop here is ok, but for readability, let's stick with the `hasFinished....` prefix
```suggestion
	const { hasFinishedResolution: hasMCProductStatsFinishedResolution, data: products } =
```

js/src/product-feed/product-statistics/create-campaign-notice.js
```diff
@@ -0,0 +1,58 @@
...
+				</p>
+				<AddPaidCampaignButton
+					eventProps={ {
+						context: 'product-feed-overview-promotion',
+					} }
+					children={ __(
+						'Create Campaign',
+						'google-listings-and-ads'
+					) }
+				/>
```
joemcgill: Instead of passing the text as the `children` prop, you can do it like this:

```suggestion
				<AddPaidCampaignButton
					eventProps={ {
						context: 'product-feed-overview-promotion',
					} }
				>
					{ __( 'Create Campaign', 'google-listings-and-ads' ) }
				</AddPaidCampaignButton>
```

joemcgill (commented):
@kt-12 I think I figured out some of the E2E issues and pushed a couple of commits. I've left a few other bits of feedback inline.

js/src/product-feed/product-statistics/create-campaign-notice.js
```diff
@@ -0,0 +1,58 @@
...
+
+const CreateCampaignNotice = () => {
+	const { hasFinishedResolution: productsResolution, data: products } =
+		useMCProductStatistics();
+	const { loaded: campaignsLoaded, data: campaigns } = useAdsCampaigns();
+
+	const isLoading =
+		! productsResolution || products?.loading || ! campaignsLoaded;
+
+	const inValidData =
```
asvinb:
```suggestion
	const invalidData =
```

js/src/product-feed/product-statistics/create-campaign-notice.js
```diff
@@ -0,0 +1,58 @@
...
+		( productsResolution && ! products ) ||
+		( campaignsLoaded && ! campaigns );
+
+	if (
+		isLoading ||
```
asvinb: @kt-12 We have more or less the same conditions when using `isLoading` in `inValidData`. Can we simplify it please?
```suggestion
if (
		isLoading ||
		! campaigns ||
		campaigns?.length > 0 ||
		! products ||
		products?.statistics?.active === 0
	) {
```

js/src/product-feed/product-statistics/create-campaign-notice.js
```diff
@@ -0,0 +1,58 @@
...
+		inValidData ||
+		products?.statistics?.active === 0 ||
+		campaigns?.length > 0
+	) {
+		return null;
+	}
+
+	return (
+		<Notice className="gla-create-campaign-notice" isDismissible={ false }>
```
asvinb: @kt-12 Looks like we are not using this class name for any styling purposes. Can we remove it? Within your tests, you can use `hasText` to target that notice.

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
```diff
@@ -499,6 +511,20 @@ test.describe( 'Set up Ads account', () => {
 	} );
 
 	test.describe( 'Create Ads with billing data already setup', () => {
+		test.beforeAll( async () => {
+			// Reset the campaigns set in the previous tests.
+			await setupBudgetPage.fulfillAdsCampaignsRequest( [] );
+
+			// Mock budget recommendations.
+			await setupBudgetPage.fulfillRequest(
```
asvinb: @kt-12 Why we need to set the budget recommendation ? 🤔

asvinb (changes requested):
@kt-12 I left a couple of comments. Can you kindly check them out please? Also, the unit tests are failing. Can you take a look?

tests/e2e/specs/add-paid-campaigns/add-paid-campaigns.test.js
joemcgill: Without budget recommendations, the BudgetRecommendation component was throwing an error when the internal getHighestBudget() function runs recommendations.reduce() because recommendations is undefined. I couldn't reproduce this condition so am not sure it's a realistic scenario, since so much of the state is getting mocked previous to this running. Mocking this as well resolves the issue.

js/src/product-feed/product-statistics/create-campaign-notice.js
kt-12: I removed alias, this was the only  `hasFinishedResolution` in that file.

js/src/product-feed/product-statistics/create-campaign-notice.js
kt-12: Removed the variable as your next comment made sense to me

js/src/product-feed/product-statistics/create-campaign-notice.js
kt-12: Done.

joemcgill (approved):
@kt-12 and @asvinb, I was reviewing this again it it mostly looks good to me, except that the styling of the notice didn't match the designs. I've pushed another commit, [199675d](https://github.com/woocommerce/google-listings-and-ads/pull/2641/commits/199675d6c77e4e41b73a0aaa83713fa2e918ae02), which is better.

**Before**
Screenshot 2024-10-21 at 5 58 37 PM

**After**
Screenshot 2024-10-21 at 6 33 29 PM

For now, I've used `gla-ads-inline-notice` as the `className` for handing this styling as there doesn't seem to be any existing components. This was mentioned in the definition questions in the original issue but wasn't addressed, so I've implemented it specifically at this component level, rather than as a larger abstraction.

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* Theme active on store:  Twenty Twenty-Four Version: 1.2
* WooCommerce - Version 9.3.2
* PHP: 8.3
* Web Server: Nginx
* OS: macOS Sonoma 14.6.1 

</Details>



**Test Results -** 

When No Paid Campaign available -

image


When Paid Campaign Available - 

image


Paid campaign creation flow from Notice button: 

https://github.com/user-attachments/assets/15a66c7e-d27e-4bb8-bf5f-75517245bfac

ankitguptaindia (approved):

js/src/product-feed/product-statistics/create-campaign-notice/index.js
```diff
@@ -0,0 +1,54 @@
...
+	const isLoading =
+		! hasFinishedResolution || products?.loading || ! campaignsLoaded;
+
+	if (
+		isLoading ||
+		! products ||
+		products?.statistics?.active === 0 ||
+		! campaigns ||
+		campaigns?.length > 0
+	) {
```
eason9487: The following conditions seem to be sufficient?

```js
if (
	! products?.statistics?.active ||
	! campaignsLoaded ||
	campaigns?.length > 0
) {
```

js/src/product-feed/product-statistics/create-campaign-notice/index.js
```diff
@@ -0,0 +1,54 @@
...
+		return null;
+	}
+
+	return (
+		<Notice isDismissible={ false } className="gla-ads-inline-notice">
```
eason9487: Using `Notice` requires certain style adjustments/overrides but it doesn't seem necessary. Could this component be implemented with layout components like `Flex`, `FlexBlock` and `FlexItem`, and fewer styling?

js/src/product-feed/product-statistics/create-campaign-notice/index.js
```diff
@@ -0,0 +1,54 @@
...
+						'google-listings-and-ads'
+					) }
+				</p>
+				<AddPaidCampaignButton
+					isSmall={ false }
+					eventProps={ {
+						context: 'product-feed-overview-promotion',
+					} }
+				>
+					{ __( 'Create Campaign', 'google-listings-and-ads' ) }{ ' ' }
```
eason9487: `{ ' ' }` seems unnecessary.

js/src/product-feed/product-statistics/create-campaign-notice/index.scss
```diff
@@ -0,0 +1,13 @@
+.gla-ads-inline-notice {
+	margin: $grid-unit-20;
+	border-left: none;
+	padding: 0.75rem 1rem;
```
eason9487: Prefer `$grid-unit-*`.

js/src/product-feed/product-statistics/index.js
```diff
@@ -133,6 +134,7 @@ const ProductStatistics = () => {
 					</SummaryList>
 				) }
 			</CardBody>
+			<CreateCampaignNotice />
```
eason9487: Looks like it should be placed in the below `<CardFooter>`? That way there's no divider.

image

js/src/hooks/useAdsCampaigns.js
```diff
@@ -30,19 +30,15 @@ const selectorName = 'getAdsCampaigns';
...
 
 	return useSelect(
 		( select ) => {
-			// TODO: ideally adsSetupComplete should be retrieved from API endpoint
-			// and then put into wp-data.
-			// With that in place, then we don't need to depend on glaData
-			// which requires force reload using window.location.href.
-			const { adsSetupComplete } = glaData;
-
-			if ( ! adsSetupComplete ) {
```
eason9487: After this change, when a Google Ads account is created via Ads Setup flow but is not claimed, it will cause errors on the Dashboard and Reports pages.

https://github.com/user-attachments/assets/383076de-064f-42df-b768-c5547e18cd05

eason9487 (changes requested):
> Further more this ticket address a bug in `useAdsCampaign` [#2539 (comment)](https://github.com/woocommerce/google-listings-and-ads/issues/2539#issuecomment-2397847396), [...]

> There is an existing issue with `useAdsCampaigns` that was discovered in [#2242 (comment)](https://github.com/woocommerce/google-listings-and-ads/pull/2242#issuecomment-2026060492), where the hook will not fetch existing Ads accounts in certain circumstances.

I wouldn't consider it's a bug. Please refer to https://github.com/woocommerce/google-listings-and-ads/pull/2242#issuecomment-2029024763. In terms of the requirement, does `useAdsCampaigns` need to be changed?

In addition, the e2e testing couldn't pass: https://github.com/woocommerce/google-listings-and-ads/actions/runs/11477389417/job/31942026350

js/src/product-feed/product-statistics/create-campaign-notice/index.js
joemcgill: Sure, it could be. The reason it was done this way is because it was described in the designs as a component that existed in the pattern library called an "inline notice", but there did not yet seem to be an implementation of such a component in this plugin. Semantically, implementing this as a new `Notice` type makes sense and the current implementation does match the design intent.

js/src/product-feed/product-statistics/index.js
joemcgill: It could have been, but doing so caused layout conflicts with the existing `Flex` container. I think changing the component to use a `Flex` wrapper instead of a `Notice` could help here, so we'll give it a try.

js/src/product-feed/product-statistics/create-campaign-notice/index.js
joemcgill: [165543c](https://github.com/woocommerce/google-listings-and-ads/pull/2641/commits/165543cc1f6e87d45b75660203e1c081476335f0) swaps this to a `Flex` which also allows it to be rendered in the `CardFooter`.

js/src/product-feed/product-statistics/create-campaign-notice/index.scss
joemcgill: Addressed in [165543c](https://github.com/woocommerce/google-listings-and-ads/pull/2641/commits/165543cc1f6e87d45b75660203e1c081476335f0)

js/src/product-feed/product-statistics/index.js
joemcgill: I think this looks better after [165543c](https://github.com/woocommerce/google-listings-and-ads/pull/2641/commits/165543cc1f6e87d45b75660203e1c081476335f0).

image

joemcgill:
@eason9487 thanks for the review, we're working through the feedback. Regarding this:

> In terms of the requirement, does `useAdsCampaigns` need to be changed?

I do think that `useAdsCampaigns` needs to be changed for this requirement. 

The purpose of this issue is to encourage merchants to create their first campaign but only if they have approved products and do not already have any ads campaigns created. To verify whether ads campaigns are running we need to check for existing campaigns on the account, which is what this hook seems to be for. However, if someone connects an existing account that already has ads and finishes onboarding by skipping creating a new ad, then the request to see if the account already has ads will never run.

This can be verified by connecting an account and creating a campaign.

image

Then from the settings screen, disconnect the account and reconnect it through the setup-ads flow and close without creating a new campaign. At this point, no more campaigns will show on the dashboard because no ads will be returned from the `useAdsCampaigns` hook, even though the account _does_ have campaigns running.

image

This does not seem like expected behavior to me.

js/src/hooks/useAdsCampaigns.js
```diff
@@ -30,19 +31,22 @@ const selectorName = 'getAdsCampaigns';
...
+			if (
+				! hasGoogleAdsConnection ||
```
kt-12: `hasGoogleAdsConnection` can be true for incomplete ads accounts(unclaimed). But we can't create or fetch campaign from those accounts, hence the next check.

eason9487:
Hi @joemcgill

> I do think that `useAdsCampaigns` needs to be changed for this requirement.

Other side effects from changing `useAdsCampaigns` will need to be reviewed one by one to see if they need to be adjusted together, otherwise there may be inconsistent presentation at a high level.

For example, the campaign list is displayed on the Dashboard page, but the Performance card on the same page still doesn't show campaign performance data.

It's also possible that a merchant has connected to a Google Ads account with existing campaigns but has not completed the Ads Setup flow for a long time. The `adsSetupComplete` will be always `false` and `ADS_SETUP_COMPLETED_AT` doesn't exist. The frontend and backend logic associated with these two values will also be affected and may need to be reviewed one by one as well.

joemcgill:
> Other side effects from changing useAdsCampaigns will need to be reviewed one by one to see if they need to be adjusted together, otherwise there may be inconsistent presentation at a high level.

That makes sense. I've reviewed the codebase and am documenting my observations here for posterity.

It looks like all of the functionality that depends on the `gla.adsSetupComplete` global, including the `useAdsCampaigns()` hook assume that there are no campaigns to display until after the Ads account has been fully set up via the onboarding process or the secondary "Set up paid ads" flow. In both cases, "fully set up" means that an ad campaign has been created (not just that the Ads account has been connected).

The only way that global is getting set is firing the `useAdsSetupCompleteCallback()` hook [here](https://github.com/woocommerce/google-listings-and-ads/blob/9dd2744500a79a46d6b31c0444bbd9024b2a8c43/js/src/setup-ads/ads-stepper/setup-paid-ads.js#L36) and [here](https://github.com/woocommerce/google-listings-and-ads/blob/9dd2744500a79a46d6b31c0444bbd9024b2a8c43/js/src/setup-mc/setup-stepper/setup-paid-ads.js#L49). Which make a `POST` request to the  ads/setup/complete endpoint, resulting in the `gla_ads_setup_completed_at` option in the DB being set. Then, on a new page load, that option is checked to set the global via an inline script set [here](https://github.com/woocommerce/google-listings-and-ads/blob/e34d260da68af46b0aad8b1635386c7159bf9b50/src/Admin/Admin.php#L139).

This means that as far as the plugin is concerned, you're not able to see and manage campaigns until you've created a new campaign after connecting the account (even if you reconnect an account you were previously using). 

That limitation should probably be resolved, IMO, but that is well beyond the scope of what we're trying to solve here or in #2538. So for consistency, I think we'll keep `useAdsCampaigns()` hook as it was prior to our changes and make our new functionality for these issues work from the same assumption—i.e., that no campaigns are known until after a new campaign is created after connecting the Ads account. 

I can log a new issue for the broader updates that we can look at coming back to.

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,141 @@
...
+			} ),
+		] );
+	}
+
+	/**
+	 * Change the value of adsSetupComplete to `true` after it is discovered.
+	 *
+	 * @return {Promise<void>}
+	 */
+	async mockAdsSetupComplete() {
```
kt-12: @eason9487 `useAdsCampaigns` mainly relies on the value of  `adsSetupComplete` to make a campaign query. I tried using `mockCampaignCreationAndAdsSetupCompletion` on this page, but it didn't work here. Finally, I went ahead modifying the global as soon as it was available.

Let me know if there is a better approach here.

js/src/product-feed/product-statistics/create-campaign-notice/index.js
kt-12: addressed.

js/src/product-feed/product-statistics/create-campaign-notice/index.scss
```diff
@@ -0,0 +1,10 @@
+.gla-ads-inline-notice {
+	background-color: #f0f6fc;
+	border-left: none;
```
eason9487: This styling is unnecessary.

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,141 @@
...
+				total: 0,
+				loading: false,
+			} ),
+
+			this.fulfillJetPackConnection( {
+				active: 'yes',
+				owner: 'yes',
+				displayName: 'John',
+				email: 'john@email.com',
+			} ),
```
eason9487: Maybe replace this with `this.mockJetpackConnected()`.

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,141 @@
...
+
+			this.mockGoogleConnected(),
+
+			this.fulfillAdsConnection( {
+				id: 1111111,
+				currency: 'USD',
+				symbol: '$',
+				status: 'connected',
+				step: '',
+			} ),
```
eason9487: Maybe replace this with `this.mockAdsAccountConnected()`.

tests/e2e/utils/pages/product-feed.js
eason9487: A more suitable solution is to set `ADS_SETUP_COMPLETED_AT` by referring to the way of setting `MC_SETUP_COMPLETED_AT`:

- https://github.com/woocommerce/google-listings-and-ads/blob/65972acffddb78cefba28ab239ef578b1054b00a/tests/e2e/test-data/test-data.php#L38-L53
- https://github.com/woocommerce/google-listings-and-ads/blob/65972acffddb78cefba28ab239ef578b1054b00a/tests/e2e/test-data/test-data.php#L72-L101
- https://github.com/woocommerce/google-listings-and-ads/blob/65972acffddb78cefba28ab239ef578b1054b00a/tests/e2e/utils/api.js#L102-L114

After then, use them like [this test file](https://github.com/woocommerce/google-listings-and-ads/blob/65972acffddb78cefba28ab239ef578b1054b00a/tests/e2e/specs/product-editor/classic-integration.test.js), and make sure to clear it in the `test.afterAll` block.

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,141 @@
...
+	 *
+	 * @return {Promise<string>} The active product value as a string.
+	 */
+	async getActiveProductValue() {
+		return this.page
+			.locator( '.woocommerce-summary__item-label span >> text=Active' )
+			.locator( '../..' )
+			.locator( '.woocommerce-summary__item-value span' )
+			.innerText();
+	}
```
eason9487: This method may not be required. Prefer `expect(locator).toHaveText()` to avoid flakiness.

https://github.com/woocommerce/google-listings-and-ads/blob/65972acffddb78cefba28ab239ef578b1054b00a/tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js#L86-L88

For example, test assertions like the above could be asserted by:

```js
await expect(
	productFeedPage.getActiveProductValueElement()
).toHaveText( /^0$/ );
```

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,141 @@
...
+			checkElement();
+		} );
+	}
+
+	/**
+	 * Get the active product value element.
+	 *
+	 * @return {Promise<import('@playwright/test').Locator>} The active product value element.
+	 */
+	async getActiveProductValueElement() {
```
eason9487: Does it need to be wrapped in a promise instance?

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,141 @@
...
+			.locator( '.woocommerce-summary__item-value span' )
+			.innerText();
+	}
+
+	/**
+	 * Get the campaign notice section.
+	 *
+	 * @return {Promise<import('@playwright/test').Locator>} The campaign notice section.
+	 */
+	async getCampaignNoticeSection() {
```
eason9487: Does it need to be wrapped in a promise instance?

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,141 @@
...
+				'You have approved products. Create a Google Ads campaign to reach more customers and drive more sales.',
+		} );
+	}
+
+	/**
+	 * Get the create campaign button in the notice section.
+	 *
+	 * @return {Promise<import('@playwright/test').Locator>} The create campaign button.
+	 */
+	async getInNoticeCreateCampaignButton() {
```
eason9487: Does it need to be wrapped in a promise instance?

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
```diff
@@ -0,0 +1,207 @@
...
+ */
+let page = null;
+
+test.describe( 'Product Feed Page', () => {
+	test.beforeAll( async ( { browser } ) => {
+		page = await browser.newPage();
+		productFeedPage = new ProductFeedPage( page );
+		setupAdsAccounts = new SetupAdsAccountsPage( page );
+		await Promise.all( [
+			setupAdsAccounts.mockAdsAccountsResponse( [] ),
```
eason9487: This mock is no longer needed.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
```diff
@@ -0,0 +1,207 @@
...
+					pending: 0,
+					disapproved: 0,
+					not_synced: 1137,
+				},
+				scheduled_sync: 0,
+				loading: false,
+			} );
+
+			await page.reload();
+			await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
```
eason9487: It doesn't seem to need reloading.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
```diff
@@ -0,0 +1,207 @@
...
+					name: 'Set up your accounts',
+				} )
+			).toBeVisible();
+
+			await productFeedPage.goto();
+			await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
+
+			await expect(
+				page.getByRole( 'heading', { level: 1, name: 'Product Feed' } )
+			).toBeVisible();
```
eason9487: Why does it need to go back here?

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
kt-12: We need it to update the statistics.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
kt-12: You are right; we can skip this.

tests/e2e/utils/pages/product-feed.js
kt-12: Done this way.

tests/e2e/utils/pages/product-feed.js
kt-12: Done.

tests/e2e/utils/pages/product-feed.js
kt-12: Done.

tests/e2e/utils/pages/product-feed.js
kt-12: Yes, we could do without promise.

tests/e2e/utils/pages/product-feed.js
kt-12: Promise removed.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
kt-12: Removed.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
kt-12: Reload was required to update the stats.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
kt-12: The page eventually crashed for some reason; all I needed was to check if `Create Campaign` was taking me to the right page and not to test the entire functionality on that page. So, going back was a good option here.

js/src/product-feed/product-statistics/create-campaign-notice/index.scss
kt-12: removed

tests/e2e/utils/pages/product-feed.js
```diff
@@ -0,0 +1,114 @@
...
+			this.mockAdsAccountConnected(),
+		] );
+	}
+
+	/**
+	 * Change the value of adsSetupComplete to `true` after it is discovered.
+	 *
+	 * @return {Promise<void>}
+	 */
+	async mockAdsSetupComplete() {
```
eason9487: This method could be removed.

tests/e2e/utils/api.js
```diff
@@ -113,6 +113,20 @@ export async function clearOnboardedMerchant() {
 	await api().delete( 'gla-test/onboarded-merchant' );
 }
 
+/**
+ * Set Ads Completed At.
+ */
+export async function setAdsCompletedAt() {
```
eason9487: Prefer `setCompletedAdsSetup`.

tests/e2e/utils/api.js
```diff
@@ -113,6 +113,20 @@ export async function clearOnboardedMerchant() {
...
+	await api().post( 'gla-test/ads-completed' );
+}
+
+/**
+ * Clear Ads Completed At.
+ */
+export async function clearAdsCompletedAt() {
```
eason9487: Prefer `clearCompletedAdsSetup`.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
eason9487: How about making the following adjustments?

- moving `productFeedPage.goto()` out from `test.beforeAll`
- calling `productFeedPage.goto()` after all mocks are set up
- moving assertions below the `productFeedPage.goto()` call
- for all reloads
   ```js
   await page.reload();
   await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
   ```
   either remove it or replace it with `await productFeedPage.goto();`

It can save some e2e testing time. For example: 

```diff
diff --git a/tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js b/tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
index 303fb0f3f..4ab4f1a4b 100644
--- a/tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
+++ b/tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
@@ -46,14 +46,9 @@ test.describe( 'Product Feed Page', () => {
 	test.describe( 'No campaign', () => {
 		test.beforeAll( async () => {
 			await productFeedPage.fulfillAdsCampaignsRequest( [] );
-			await productFeedPage.goto();
 		} );
 
 		test( 'No active product and no campaign; Do not display campaign notice', async () => {
-			await expect(
-				page.getByRole( 'heading', { level: 1, name: 'Product Feed' } )
-			).toBeVisible();
-
 			await productFeedPage.fulfillProductStatisticsRequest( {
 				timestamp: 1695011644,
 				statistics: {
@@ -67,8 +62,11 @@ test.describe( 'Product Feed Page', () => {
 				loading: false,
 			} );
 
-			await page.reload();
-			await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
+			await productFeedPage.goto();
+
+			await expect(
+				page.getByRole( 'heading', { level: 1, name: 'Product Feed' } )
+			).toBeVisible();
 
 			await expect(
 				page.getByRole( 'heading', {
@@ -103,8 +101,7 @@ test.describe( 'Product Feed Page', () => {
 				loading: false,
 			} );
 
-			await page.reload();
-			await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
+			await productFeedPage.goto();
 
 			await expect(
 				productFeedPage.getActiveProductValueElement()
@@ -129,8 +126,6 @@ test.describe( 'Product Feed Page', () => {
 					name: 'Set up your accounts',
 				} )
 			).toBeVisible();
-			await productFeedPage.goto();
-			await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
 		} );
 	} );
 
@@ -156,8 +151,7 @@ test.describe( 'Product Feed Page', () => {
 		} );
 
 		test( 'Has active product and a campaign; Do not display campaign notice', async () => {
-			await page.reload();
-			await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
+			await productFeedPage.goto();
 
 			await expect(
 				productFeedPage.getActiveProductValueElement()
@@ -185,8 +179,8 @@ test.describe( 'Product Feed Page', () => {
 				scheduled_sync: 0,
 				loading: false,
 			} );
-			await page.reload();
-			await page.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
+
+			await productFeedPage.goto();
 
 			await expect(
 				productFeedPage.getActiveProductValueElement()

```

eason9487 (approved):
Thanks for the work and adjustments.

There are some suggestions for e2e tests and I believe it won't need a new round code review for these adjustments. I'll be approving it in advance.

tests/e2e/utils/api.js
kt-12: Done.

tests/e2e/utils/pages/product-feed.js
kt-12: Done.

tests/e2e/utils/api.js
kt-12: Done.

tests/e2e/specs/product-feed/product-feed-campaign-notice.test.js
kt-12: This gave a boost. The test ran much faster after this change. Thanks, @eason9487, for this suggestion.

kt-12: Done.

Issue #2616, opened 2024-09-17
[GTIN] Migrate existing GLA Global Trade Item Number data to core


martynmjones:
The `GTIN` field in Google for WooCommerce will be retired in favour of the new global unique ID field available in core. To support this change we will need to add a migration tool to populate the core field from data available in Google for WooCommerce. Merchants will be prompted to run the migration tool rather than it running automatically.

## Acceptance criteria

1. The migration will happen using batched scheduled actions
1. Existing merchants of Google for WooCommerce will be prompted via a banner to run a migration tool
1. Existing merchants will see a prompt next to the `GTIN` field in Google for WooCommerce to run a migration tool
1. The data migrated by the tool will respect mapping rules and filters to ensure that the most relevant source of information is used
1. The migration tool will only populate the core global unique ID field if it is empty
    - Action scheduler logs should state if a product was skipped due to the core field being populated
1. The migration tool will be available via CLI
    - This is secondary to the rest of the functionality and may be added as a follow-up

## Additional

pcTzPl-2p2-p2

puntope:
Closed - https://github.com/woocommerce/google-listings-and-ads/pull/2662 https://github.com/woocommerce/google-listings-and-ads/pull/2656 https://github.com/woocommerce/google-listings-and-ads/pull/2655

PR #2645, opened 2024-10-18
[GTIN] Add GTIN migration JOB

puntope:

Closes part of #2616 as part of pcTzPl-2p2-p2.

This PR creates the logic for the Job responsible for the migration GTIN in batches for all the products.

### Screenshots:

Screenshot 2024-10-17 at 18 29 40

1. Set a simple product and add a GTIN in Google for WooCommerce tab
2. Set a simple product  and add a different GTIN in Google for WooCommerce tab and in the Inventory tab
3. Set a Variable product and set GTIN on one of the variations in Google for WooCommerce  only and in both (inventory and Google for WooCommerce) in other.
4. Go to Connection Test page and click on "Start GTIN Migration"
5. See the job  `gla/jobs/migrate_gtin/create_batch` scheduled in WooCommerce - Status -  Scheduled actions
6. Run the job and see the `gla/jobs/migrate_gtin/process_items` created with the product ids.
7. Run it
8. Go back to the products created in steps 1,2,3 and verify that:
9. For the product in step 1- the GTIN was loaded  in the inventory GTIN
10. For the product in step 2 - The GTIN is still the GTIN you set initially and is not override  by the GTIN in google for WooCommerce GTIN
11. Check the same in the variations (GTIN set in core should not be overrided in the migration).

### Additional details:

In follow-up PRs I will add:

-  the banners for starting the migration
-  tests
-  WP CLI tool

### Changelog entry

> Add - Add GTIN Migration Job

### DIFF
--- a/src/Jobs/MigrateGTIN.php
+++ b/src/Jobs/MigrateGTIN.php
@@ -1,0 +1,136 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Jobs;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\ActionScheduler\ActionSchedulerInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\Attributes\AttributeManager;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductRepository;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class MigrateGTIN
+ *
+ * Schedules GTIN migration for all the products in the store.
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Jobs
+ * @since x.x.x
+ */
+class MigrateGTIN extends AbstractBatchedActionSchedulerJob implements OptionsAwareInterface {
+	use OptionsAwareTrait;
+
+	/**
+	 * @var ProductRepository
+	 */
+	protected $product_repository;
+
+
+	/**
+	 * @var AttributeManager
+	 */
+	protected $attribute_manager;
+
+
+	/**
+	 * Job constructor.
+	 *
+	 * @param ActionSchedulerInterface  $action_scheduler
+	 * @param ActionSchedulerJobMonitor $monitor
+	 * @param ProductRepository         $product_repository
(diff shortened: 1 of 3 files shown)


src/Jobs/MigrateGTIN.php
```diff
@@ -0,0 +1,136 @@
...
+		return parent::can_schedule( $args ) && $this->is_gtin_available_in_core();
+	}
+
+	/**
+	 * Process batch items.
+	 *
+	 * @param int[] $items A single batch of WooCommerce product IDs from the get_batch() method.
+	 */
+	protected function process_items( array $items ) {
+		// todo: prevent execution if migration was completed?
```
tomalec, line 77: :question:  Is this something you are planning to add in the follow-up PRs, or was it missed?

tomalec (commented):
I run the tests locally and for simple products it goes smooth.

:question: 
I noticed a different behavior for simple products when the GTIN is already set to the same value or when it's set to a different one. Is that expected? if there are differences, it keeps the core GTIN but adds the G4W as SKU, while for the case where they're the same, it does not set SKU.
image




Plus, for variable ones, it didn't migrate the GTINs.
The process job was executed.
image
image

src/Jobs/MigrateGTIN.php
puntope: Yes. This is a todo and will be done in a further PR when setting the Banner for the merchant to run the job

puntope:
> I noticed a different behavior for simple products when the GTIN is already set to the same value or when it's set to a different one. Is that expected? if there are differences, it keeps the core GTIN but adds the G4W as SKU, while for the case where they're the same, it does not set SKU.

Hey @tomalec that's odd I was not able to reproduce this. Can you verify again or help me to reproduce this issue? 

The expected behaviour is:

- If GTIN is set in the inventory tab then migration does nothing.
- If GTIN is not set in the inventory tab then migration loads the GTIN set in Google for WooCommerce tab (if it is set)

puntope:
> Plus, for variable ones, it didn't migrate the GTINs. The process job was executed.


The variables are required to process the Variations. I just added a tweak [here](https://github.com/woocommerce/google-listings-and-ads/pull/2645/commits/d39b358d27a404f4a21cc4e57c9da56cc5261f64) because I notice I was processing only available variations (those with price set) Now I do get all the variations.

Notice the GTIN in the variable inventory is useless because we don't support GTIN in Variables. So you need to check each of the variations (they have their own GTIN fields in core part and G4W tab) .

puntope:
Hey @tomalec I addressed all your comments and questions. Can you follow up with another review?

tomalec (approved):
Thanks for changes and clarifications, now it works well for me :)

PR #2649, opened 2024-10-19
[GTIN] Prepare GTIN with the correct format before sending it to MC

puntope:

This PR prepares the GTIN for sending data to MC by removing everything but numbers.

Before this PR. When setting a GTIN in Inventory with dashes (which is an accepted format) it fails when syncing that product. This PR fixes that.

1. Add a GTIN in inventory like 0000-0000  (use the dashes)
2. Sync the product in Connection test page
3. The product is successfully synced.

### Changelog entry

> Tweak -  Prepare GTIN with the correct format before sending it to MC

### DIFF
--- a/src/Product/WCProductAdapter.php
+++ b/src/Product/WCProductAdapter.php
@@ -941,7 +941,8 @@ class WCProductAdapter extends GoogleProduct implements Validatable {
 			return $this;
 		}
 
-		$global_unique_id = $this->wc_product->get_global_unique_id();
+		// avoid dashes and other unsupported format
+		$global_unique_id = preg_replace( '/[^0-9]/', '', $this->wc_product->get_global_unique_id() );
 
 		if ( ! empty( $global_unique_id ) ) {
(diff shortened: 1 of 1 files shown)


src/Product/WCProductAdapter.php
```diff
@@ -941,7 +941,8 @@ protected function map_gtin(): WCProductAdapter {
 			return $this;
 		}
 
-		$global_unique_id = $this->wc_product->get_global_unique_id();
+		// avoid dashes and other unsupported format
+		$global_unique_id = preg_replace( '/[^0-9]/', '', $this->wc_product->get_global_unique_id() );
```
mikkamp, line 945: Unfortunate that it would look odd to use the WC function `prepare_gtin` here. But it's probably better since it's not available in all versions of WC that support GTIN.

mikkamp (approved):
Thanks for the changes here. I can confirm that the GTIN number is being synced to Google MC now (with only numbers).

image

image

Issue #2509, opened 2024-08-07
Onboarding: Consolidation of account creation/connection cards


joemcgill:
As part of the Onboarding Improvements project (#2458), we want to consolidate the functionality and UX from three separate cards that are responsible for connecting to a Google account, a Google Ads account, and a Google Merchant Center account, into a new single flow.

**Current UI:**
<img width="1056" alt="image" src="https://github.com/user-attachments/assets/47cb670c-f3e9-485b-ac9f-9297dc17d9a5">

## Proposed flow
There are 4 possible scenarios to account for:

<img width="474" alt="image" src="https://github.com/user-attachments/assets/060abe8d-19fe-47a5-a9fe-3011dfc747bc">

1. Merchant doesn’t have Merchant Center accounts nor Google Ads accounts. 
2. Merchant has one or more Merchant Center account but not an Google Ads account.
3. Merchant has one or more Merchant Center and Google Ads accounts.
4. Merchant doesn’t have Merchant Center accounts but has one or more Google Ads accounts.

## Design Overview

Design mocks are being iterated on in [this Figma file](https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=7176-108902&t=eFMkAn0y2V4OQmPR-0).

The new combined Google account connection card will begin by asking the user to accept terms & conditions for Google MC and Google Ads before starting the connection process.

![Image](https://github.com/user-attachments/assets/a36dde89-7e72-45dc-a73a-22385731475a)

Once checked, the Connect button will be enabled, allowing the user to begin by setting up their Google account and enter the current OAuth flow.

![Image](https://github.com/user-attachments/assets/8c93fe24-8099-4b61-934e-00aaec600f17)

Once all of the necessary permissions are granted, the Google combo card will see if any Ads or MC accounts already exist for that Google account. 

If no accounts exist, they will automatically be created to start to onboarding process.

![Image](https://github.com/user-attachments/assets/5fb50126-ff4a-4e6c-826a-b378d7ac1654)

If the Google account already has existing Ads or MC accounts, a new combo view will allow the user to select which accounts to connect, along with some recommendations based on info we can infer from the account status. See an example of choosing an MC account in the following screenshot.

![Image](https://github.com/user-attachments/assets/c4457836-4f68-4158-8ace-5e3db2f688f9)

## Technical Design
[...]

PR #1025, opened 2021-09-24
Separate Phone and Address settings

tomalec:
### Changes proposed in this Pull Request:

- Add placeholder mixin, (659d29d)
	to make it easier to add preloaders more subtle than spinners.
	(Copied the code from woocommerce/woocommerce-admin@9f68059/packages/style-build/abstracts/_mixins.scss#L22-L35)
- Add support for custom appearance to `AccountCard` (a7286c2)
	So we could provide any title when needed.
- Split `EditContactInformation`, (568bb3e, e836e5e)
	into `EditPhoneNumber` and `EditStoreAddress`.
- Separate preview mode of `StoreAddressCard` and `PhoneNumberCard` into separate components. (c21c74c, bc3b635, 55f63d0, 8f349e7)
	Extract shared code into  `ContactInformationPreviewCard`.
	Simplify the logic. Change "Edit" to internal link.

- This PR changes "Edit" buttons to be regular links, so they could be `ctrl + click`ed, and navigated with browser built-in features.

Addresses https://github.com/woocommerce/google-listings-and-ads/issues/962#issuecomment-925686034

### Screenshots:
loading

#### Happy path 
1. Go to the settings page [`/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings`](https://gla1.test/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings)
2. Wait for the phone number and address to load correctly
3. Check that the page looks like in the designs
2. Click the "Edit" button by the "Phone number"
3. Go back to 1., Click the "Edit" button by the "Store address"

#### Missing contact info.

This is quite hard to achieve with newly created accounts. Ideally, you should have an old account w/o a store/phone address.

##### Different store address

0. Go to Woo settings [`/wp-admin/admin.php?page=wc-settings`](https://gla1.test/wp-admin/admin.php?page=wc-settings)
1. Change the address to something else than the current one.
2. Go to the settings page [`/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings`](https://gla1.test/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings)
2. You should eventually see the warning about the missing store address.

address warning

##### Mocked missing phone number
0. Change [`/js/src/data/actions.js#L521`](https://github.com/woocommerce/google-listings-and-ads/blob/feature/962-separate-settings/js/src/data/actions.js#L521) to data: { ...data, phone_number: '' },`.
[...]

PR #2653, opened 2024-10-28
Consolidate accounts onboarding: Merge the store address setting in Step 3 into Step 1 and remove Step 3 along with the phone verification

### LINKED ISSUE #2602: Sync MC address during account connection

As a follow-up to #2493, the requirement to validate a phone number is possibly going away, in which case, we can consider moving the process of syncing the store's address with the MC account to the `GoogleComboAccountCard` when the account is first connected, if we see that there is no address associated with the MC account or if it differs from the store's address.

**Screenshots:**
> [!WARNING]  
> Screenshots are out of date with the most recent change to the acceptance criteria

Image

### Onboarding

- [ ] During onboarding, the store address information will always be shown at the bottom of the `GoogleComboAccountsCard`.
- [ ] If all of the address requirements have not been met, the missing data will be shown as validation errors on the address card, and the user will not be able to complete step 1 until this is resolved.
- [ ] Clicking the "Get data from WooCommerce Settings" button (❓ Can we use "Update store address" or something shorter?) 
- [ ] If all address requirements are met and the store address does not match what is in MC, the address info will be synced when the user clicks "Continue" to move to step 2.
- [ ] The current Step 3 of the onboarding step is removed (see #2493)

### Settings screen
- [ ] On the settings screen, the Contact section displays the current `StoreAddressCardPreview` component with the "Edit" button to open the edit flow.
- [ ] The edit flow will show the same Address Card from onboarding with validation errors visible
- [ ] The "Save" button will be disabled until validation errors have been resolved
- [ ] Clicking the "Save" button will sync address data to MC and return to the settings screen.
- [ ] All unused components related to phone verification is removed

## Implementation Brief
* Create `SyncStoreAddress` component in `js/src/components/google-combo-account-card/sync-store-address/index.js` to render the `StoreAddressCard` component based on the following conditions:
	* The MC account must be connected.
		* Check the MC status via the `useGoogleMCAccount` hook and checking the `status`
	* Use the `useStoreAddress` hook to determine if the address needs to be synced.
		* `mc_address` should not be null.
		* `is_mc_address_different` is `true` which means the address needs to be synced.
	* Update the `StoreAddressCard` component so that the `handleRefreshClick` callback uses the `updateGoogleMCContactInformation` hook, as demonstrated in [this comment] (https://github.com/woocommerce/google-listings-and-ads/issues/2602#issuecomment-2417939898) and [this followup](https://github.com/woocommerce/google-listings-and-ads/issues/2602#issuecomment-2418428983).
* Update `GoogleComboAccountsCard` to include `SyncStoreAddress` as part of the card body.
* The "Continue" button for `GoogleComboAccountsCard` should be disabled until the address has been synced, by using the above hooks.

## Test Coverage
* Appropriate E2E tests should be added to tests for the above
* PHPUnit tests related to the onboarding flow are updated

asvinb:

Closes #2602  .

### Screenshots (Onboarding):

**The store address is invalid (missing info)**
Screenshot 2024-11-01 at 10 18 22 AM

**The store address is not in sync with MC**
Screenshot 2024-11-01 at 10 19 21 AM

**The store address is in sync with MC**
(nothing is shown during onboarding in this scenario)

### Screenshots (Settings):

**The store address is not in sync with MC**
Screenshot 2024-11-01 at 10 20 58 AM

**The store address is invalid (missing info)**
Screenshot 2024-11-01 at 10 21 49 AM

**The store address is in sync with MC**
Screenshot 2024-11-01 at 10 20 23 AM

1. Go through the onboarding flow.
2. The card to sync the store address should be visible if the WooCommerce street address is different from the MC address.
3. Clicking the "Refresh to sync" button should sync the WooCommerce store street address to the MC account.
4. If the address is/has been synced, the card should not be displayed.
5. The "Continue" button should be disabled if there are connected accounts but the store address has not been synced.
6. If there is no address in WooCommerce settings, the card should display an alternate text to prompt the user to add the address in WC settings.

### Changelog entry

> Update - Merge the store address setting in Step 3 of the onboarding flow into Step 1 and remove Step 3 along with the contact phone verification.
> Update - Remove the contact phone from the Settings page.


js/src/setup-mc/setup-stepper/setup-accounts/index.js
```diff
@@ -89,6 +90,7 @@ const SetupAccounts = ( props ) => {
 		isReady: isGoogleMCReady,
 	} = useGoogleMCAccount();
 	const { hasFinishedResolution } = useGoogleAdsAccount();
+	const storeAddressSynced = useStoreAddressSynced();
```
joemcgill: This hook cannot be used here until after the MC account is connected. Otherwise it will send an API request to `/gla/mc/contact-information` which will result in a 401 error.

image

Either we need to incorporate `useGoogleMCAccount()` into the `useStoreAddressSynced()` hook so that it returns `false` early if there is no MC account, or rely on a `useState()` value that gets updated after the verification is made.

js/src/hooks/useStoreAddressSynced.js
```diff
@@ -0,0 +1,19 @@
...
+ * @return {boolean} Returns `true` if the store address matches the GMC account address, otherwise, returns `false`.
+ */
+export default function useStoreAddressSynced() {
+	const { data, loaded } = useStoreAddress();
+
+	if ( ! loaded ) {
+		return false;
+	}
+
+	return data.address && ! Boolean( data.isMCAddressDifferent );
```
joemcgill: Instead of checking `data.address` here, which is only part of the address, it's probably better to check the `data.isAddressFilled` property, which checks for all required address fields. See: https://github.com/woocommerce/google-listings-and-ads/blob/99408e1deaf2954f9325f31f2b2ee82b0147119b/js/src/hooks/useStoreAddress.js#L67

joemcgill (changes requested):
@asvinb this is looking pretty good. Only a couple of critical issues to address.

js/src/setup-mc/setup-stepper/setup-accounts/index.js
asvinb: Thanks @joemcgill . Completely missed that 🤦 . I've updated the `useStoreAddressSynced` hook. Let me know what you think.

js/src/hooks/useStoreAddressSynced.js
asvinb: With the recent changes in the hook, we are now using `! missingRequiredFields.length` directly @joemcgill

js/src/components/contact-information/store-address-card.js
```diff
@@ -66,7 +70,10 @@ const StoreAddressCard = ( { showValidation = false } ) => {
 	}
 
 	const handleRefreshClick = () => {
-		refetch();
+		setSaving( true );
+		updateGoogleMCContactInformation()
+			.then( refetch )
```
joemcgill: Adding this refetch is unnecessary, because `updateGoogleMCContactInformation()` returns the new data and updates the data store.

joemcgill (dismissed):
The functional changes look good to me, except one note inline. 

However, I noticed that the new design are slightly more streamlined, avoiding the separator between the Card description and the Address:

**Old**
image

**New (ignore the missing button)**
image

js/src/components/contact-information/store-address-card.js
asvinb: 👍

joemcgill (commented):
Thanks @asvinb.

Overall, this is working as expected for the onboarding flow. However, the need to add an additional `compactStyles` prop to render a style variation on the `StoreAddressCard` component made me curious about where else this component was used, and whether we could avoid the need to the prop entirely.

Turns out, the other place this component is used is in the `EditStoreAddress` flow that you can get to by visiting `wp-admin/admin.php?page=wc-admin&subpath=%2Fedit-store-address&path=%2Fgoogle%2Fsettings` or by clicking the Edit button on the `StoreAddressCardPreview` component shown in the settings screen after finishing onboarding.

Screenshot 2024-10-29 at 3 33 32 PM

In the `develop` branch, when you visit the `EditStoreAddress` flow, the Refresh button is used to update the component with fresh address data from the WC store that is shown in the lower `Section.Card.Body`, and the the "Save details" CTA button for that flow is what syncs the address to MC. After these changes, the refresh button does the syncing but then stays continually stuck in an updating state, see this video:

https://github.com/user-attachments/assets/4ce46dd6-3c84-4d2c-b817-86db3a5a214f

Currently in `develop` the Refresh button similarly is used to update changes to the address that have been made in the WC store that aren't yet reflected in the component state and the address is synced to Google MC when that step of the onboarding flow is completed (see [this code ref](https://github.com/woocommerce/google-listings-and-ads/blob/f662ba9f7c47b308ba29cd8c0ef66c8868b3ac9d/js/src/setup-mc/setup-stepper/store-requirements/index.js#L67-L80)).

## Suggestions

Here are my suggestions for how to address these concern. I can verify which option we want to go with, but would appreciate any ideas/feedback you have as well.

### Updating the onboard flow
First, we should apply the new compact design everywhere to simplify the implementation and remove the new `compactStyles` prop.

Next, we revert the changes to the Refresh button so they are used to update the component with local data rather than sync the data to MC for a consistent UI in all places it's used. We can sync the changes to MC when the first onboarding step is completed, as part of the callback passed to the `onClick` prop of the `AppButton` [here](https://github.com/woocommerce/google-listings-and-ads/blob/e34d260da68af46b0aad8b1635386c7159bf9b50/js/src/setup-mc/setup-stepper/setup-accounts/index.js#L172-L176).

During onboarding, we should still only show this component if we either don't have valid address info or the address info doesn't match MC. However, since we'll now sync the data automatically whenever someone finishes this step, we should only disable the Continue button if there are any validation errors for the address (e.g., `! address.isAddressFilled`) and show the validation errors in the `StoreAddresCard` similar to how it's being done in the `EditStoreAddress` flow:

```js
<StoreAddressCard
  showValidation={ ! address.isAddressFilled }
/>
```

### Alternate: Update the settings screen
An alternative suggestion would be to move the updated `StoreAddressCard` from this PR directly to the Settings screen—replacing the current `ContactInformationPreview` component [here](https://github.com/woocommerce/google-listings-and-ads/blob/0c1023cc79671a8903420627702a655718b344c4/js/src/settings/index.js#L70). With this approach, we would always show the address info (not just when it's different from MC), and the Refresh button  would be for the purpose of updating the MC data.

The `EditStoreAddress` flow will no longer be needed and could be removed. The `PhoneNumberCardPreview` and `StoreAddressCardPreview` components could likely be removed entirely as well, because I don't think they're used anywhere else.

joemcgill:
After confirming with @fblascogarma earlier this week, the approach we decided to go with for resolving [this concern](https://github.com/woocommerce/google-listings-and-ads/pull/2653#pullrequestreview-2403093334) is to update the UI on the settings screen to make use of the redesigned `StoreAddressCard` and greatly simplify the UI there to remove the unnecessary phone verification data, and the edit store address flow. Those updates were implemented in #2661 and merged to this PR for finalization.

One additional issue that we've discovered is that currently when the onboarding flow is loaded, a request is made to `mc/setup` to return the step that should be started if requirements of the setup flow have already been completed. That logic is handled in the `MerchantCenterService::get_setup_status()` method ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/ac7918c937da1cebbcb04b66601248e1236c6fff/src/MerchantCenter/MerchantCenterService.php#L219-L246)). This logic doesn't currently account for the need to check for a synced address before moving to the `product_listings` step so that will need to be resolved. I think we can add `$this->is_mc_contact_information_setup()` back to that logic, which previously existed as part of the logic to see if we could start on the `paid_ads` step.

joemcgill:
@asvinb I've updated the rest of this PR after incorporating the changes from #2661 if you want to give it another look.

@mikkamp could you review the PHP changes specific to `src/MerchantCenter/MerchantCenterService.php` and the related PHPUnit tests?

@eason9487 this will still need QA, but would appreciate an early review since there are so many changes (mostly removal of unused components).

eason9487:
Before looking into the code implementation, I would like to raise concerns about the new UI.

1. WooCommerce Settings and Google Merchant Center have different rules for required address fields, so there are often cases where required fields are missing when syncing to GMC.
   ![image](https://github.com/user-attachments/assets/2283436d-57e0-4a83-b515-fc424a80bd9d)
   The UI doesn't guide the merchant through a clear correction process. For example, after a merchant updates the address via WC Settings and switch back, they will find this card is still stuck in the same error state there until they try to refresh the webpage and see the "Refresh to sync" button.
2. This card won't be shown after the address is synced or when the address already matches at the beginning. This will make it uncertain for merchants to know exactly how to (re)update the address to GMC.
   ![image](https://github.com/user-attachments/assets/8ea3c8ff-8c57-4bdc-8754-04832b16ab92)
   And even if the merchant is told how to update the address again, the flow will be less smooth than before. For example, starting with the screenshot above, the merchant must:
   1. Open a new webpage and navigate to WooCommerce Settings
   2. Update address and save changes
   3. Switch back to the webpage of Onboarding flow
   4. Refresh the webpage
   5. Click "Refresh to sync" button
3. A similar concern about insufficient guidance also occurs on the plugin Settings page.
   ![image](https://github.com/user-attachments/assets/c6df12c7-37bc-420e-9bcb-92321e83aad5)

   https://github.com/user-attachments/assets/4adad75b-e337-45e4-ba83-2ad59b3bcf14

   For example:
   - The merchant may not know they need to refresh the webpage to see the "Refresh to sync" button after they update WC address
   - The merchant needs to refresh the page again and again if they find out they want to edit address again
   - If the merchant has to modify the address again due to something like a typo, they will find that they have to refresh the webpage over and over again

---

The following compares several scenarios: After a GMC account is connected, the operations required to reach the point where the WC address is synchronized to the GMC and continue to the next step.

| Scenarios \ Comparisons                                                                                                            | Before                                                                                                                                                                                                                                                                                                                      | After                                                                                                                                                                                                                                                                                                                                                                                 |
|------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| When the two addresses are already the same                                                                                        | 1. Click "Continue" button                                                                                                                                                                                                                                                                                                  | 1. Click "Continue" button                                                                                                                                                                                                                                                                                                                                                            |
| When the two addresses are already the same<br>but want to edit it                                                                 | 1. Click link to open WooCommerce Settings<br>2. Update and save address<br>3. Back to Onboarding webpage<br>4. Click "Refresh to sync" button<br>5. Click "Continue" button                                                                                                                                                | 1. Open new webpage<br>2. Navigate to WooCommerce Settings<br>3. Update and save address<br>4. Back to Onboarding webpage<br>5. Refresh webpage<br>6. Click "Refresh to sync" button<br>7. Click "Continue" button                                                                                                                                                                    |
| When the two addresses are different                                                                                               | 1. Click "Continue" button                                                                                                                                                                                                                                                                                                  | 1. Click "Refresh to sync" button<br>2. Click "Continue" button                                                                                                                                                                                                                                                                                                                       |
| When a required address field is missing                                                                                           | 1. Click "Continue" to be notified of the missing field<br>2. Click link to open WooCommerce Settings<br>3. Add and save address<br>4. Back to Onboarding webpage<br>5. Click "Refresh to sync" button<br>6. Click "Continue" button                                                                                        | 1. Click link to open WooCommerce Settings<br>2. Add and save address<br>3. Back to Onboarding webpage<br>4. Refresh webpage<br>5. Click "Refresh to sync" button<br>6. Click "Continue" button                                                                                                                                                                                       |
| When the two addresses are already the same<br>but want to edit an address field and later<br>find a typo in the just edited field | 1. Click link to open WooCommerce Settings<br>2. Update and save address<br>3. Back to Onboarding webpage<br>4. Click "Refresh to sync" button<br>5. Switch to WooCommerce Settings page<br>6. Correct and save address<br>7. Back to Onboarding webpage<br>8. Click "Refresh to sync" button<br>9. Click "Continue" button | 1. Open new webpage<br>2. Navigate to WooCommerce Settings<br>3. Update and save address<br>4. Back to Onboarding webpage<br>5. Refresh webpage<br>6. Click "Refresh to sync" button<br>7. Switch to WooCommerce Settings<br>8. Correct and save address<br>9. Back to Onboarding webpage<br>10. Refresh webpage<br>11. Click "Refresh to sync" button<br>12. Click "Continue" button |

That's why the original process was designed that way. It's not very convenient either, but at least the guidance is relatively clear and the operations are fewer and more consistent..

eason9487:
Without introducing wider changes, whenever the merchant updates WC address, this whole mechanism always requires an action to display the address to be synchronized and maybe also display the errors to be resolved on this plugin page.

- (Re)fetch WC address and compare if it needs to be synced or if there are errors

Replacing this action with the `updateGoogleMCContactInformation` API call or hiding the UI that triggers this action under certain conditions probably wouldn't make the whole thing any better. At least the current implementation doesn't seem to make it any easier for merchants to use.

It's recommended to confirm if the concerns in https://github.com/woocommerce/google-listings-and-ads/pull/2653#issuecomment-2454215548 are in the desired direction. I would suggest that the main direction could be to remove the phone number and merge step 3 into step 1, but to leave the original mechanism of address data (re)fetching and synchronization as unchanged as possible.

Adjustments to this mechanism might be enough by rephrasing the text of the "Refresh to sync" button, or perhaps go one step further by automatically triggering a refetch of address data when the page regains focus.

joemcgill:
Thanks for flagging your concerns @eason9487. I've asked @fblascogarma and @MatthiasReinholz to review and help provide clarification. 

For what it's worth, I did meet with @fblascogarma last week to confirm these requirements, as mentioned [in this comment](https://github.com/woocommerce/google-listings-and-ads/pull/2653#issuecomment-2452412909). The decision when we spoke was to update this component so that clicking the "Refresh to sync" button would sync the address information to MC rather than updating the UI with fresh data from the WC store.

In the interest of time, can I get the current implementation reviewed and we can make further adjustments to the acceptance criteria as part of the UAT review that will happen on this feature before it's merged to `develop`?

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.6.2
* WooCommerce - Version 9.3.3
* PHP: 8.3
* Web Server: Nginx
* Browser: Chrome - Version 130
* OS: macOS Sonoma 14.6.1 

</Details>




**Test Results -** Tested PR based on the information and current Acceptance Criteria provided in connected GH issue. Sync flow is working as expected as described in requirement. 

**Onboarding Screen Sync Flow:**  ✅ 


https://github.com/user-attachments/assets/ec09aeb5-cd60-4595-bbcf-22e03f756167



**If the store address matches with MC address:** ✅ 


https://github.com/user-attachments/assets/d5400079-12e0-4729-946e-e2b4e07c817b



**Setting Screen Sync Flow:** ✅ 

https://github.com/user-attachments/assets/0e344208-7ab1-4dc8-9f11-d54651dd3099

ankitguptaindia (approved):

js/src/components/contact-information/phone-number-card/verify-phone-number-content.js
```diff
@@ -1,231 +0,0 @@
...
-	onVerificationStateChange,
-} ) {
-	const isMounted = useIsMounted();
-	const [ method, setMethod ] = useState( verificationMethod );
-	const { second, callCount, startCountdown } = useCountdown( method );
-	const [ verification, setVerification ] = useState( null );
-	const [ verifying, setVerifying ] = useState( false );
-	const [ error, setError ] = useState( null );
-	const verificationIdRef = useRef( {} );
-	const { requestPhoneVerificationCode, verifyPhoneNumber } =
```
eason9487, line 96: The two actions and action type `VERIFIED_MC_PHONE_NUMBER` can be removed together.

https://github.com/woocommerce/google-listings-and-ads/blob/852dfde9cf4d8b6af75ddee65a3726e3eee1cde1/js/src/data/actions.js#L617-L739

joemcgill:
Change of plans here. After confirming with @fblascogarma and @MatthiasReinholz today, we're going to take a slightly different approach here. I've updated the AC in the issue, but the key changes are:

> - Show always the Store Address component, independently if a merchant already synced their store address to MC.
> - Change CTA wording and keep same functionality. Wording: “Update store address”

js/src/components/contact-information/index.js
```diff
@@ -6,21 +6,9 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import { getEditPhoneNumberUrl, getEditStoreAddressUrl } from '.~/utils/urls';
-import { useAdaptiveFormContext } from '.~/components/adaptive-form';
-import useGoogleMCPhoneNumber from '.~/hooks/useGoogleMCPhoneNumber';
```
eason9487, line 11: These can be removed together:
- `js/src/hooks/useGoogleMCPhoneNumber.js`
- `libphonenumber-js` npm dependency
- `getGoogleMCPhoneNumber` wp-data selector
   https://github.com/woocommerce/google-listings-and-ads/blob/852dfde9cf4d8b6af75ddee65a3726e3eee1cde1/js/src/data/selectors.js#L127-L148

joemcgill:
After reviewing the PR with @fblascogarma today and confirming that the updates reflect our shared understanding of the new requirements, we've got an additional change to the copy used in the address card:

When the address is valid:

> "We’re using your store address for Google verification. This information won’t be public. Edit in <link>WooCommerce settings</link> if needed and update to review the changes."

When the address is NOT valid

> "Your store address is required by Google for verification. This information won’t be public. Complete that in <link>WooCommerce settings</link> and update to review the changes."
>
>{ verification errors shown below }

I've applied those changes in d772944.

joemcgill:
@eason9487 I've updated the approach for this one based on the updated requirements. There are still some E2E tests to update tomorrow, but would be happy for you to give it another look given the number of changes.

joemcgill:
E2E tests are updated so this should be ready for another look.

js/src/components/contact-information/store-address-card.scss
```diff
@@ -6,6 +6,6 @@
 	}
 
 	.gla-account-card__description {
-		color: $gray-700;
+		color: $gray-900;
```
eason9487: The original `$gray-700` looks like the desired text color.

Design document:

image

Implementation:

image

js/src/components/contact-information/index.js
```diff
@@ -39,29 +33,16 @@ const description = (
 	</>
 );
 
-const mcTitle = __( 'Verify contact information', 'google-listings-and-ads' );
 const settingsTitle = __( 'Contact information', 'google-listings-and-ads' );
 
 /**
  * Renders a preview of contact information section,
- * or a <NoContactInformationCard> if contact informations are not saved yet.
+ * or a <NoContactInformationCard> if contact information are not saved yet.
```
eason9487:
```suggestion
 * or a notice if contact information is outdated.
```

Note: `<NoContactInformationCard>` has already been removed by #1025.

js/src/components/contact-information/index.js
```diff
@@ -39,29 +33,16 @@ const description = (
...
  */
 export function ContactInformationPreview() {
 	return (
 		<Section title={ settingsTitle } description={ description }>
 			<VerticalGapLayout size="overlap">
```
eason9487: The `VerticalGapLayout` layer could be removed.

js/src/components/contact-information/index.js
```diff
@@ -78,55 +59,3 @@ export function ContactInformationPreview() {
...
-				/>
-				<StoreAddressCard
-					showValidation={ adapter.requestedShowValidation }
-				/>
-			</VerticalGapLayout>
-		</Section>
-	);
-};
-
-export default ContactInformation;
```
eason9487, line 132: Since the default component was removed, suggest exporting `ContactInformationPreview` as the default component and renaming this file to `contact-information-preview.js`.

js/src/components/contact-information/store-address-card.js
```diff
@@ -132,38 +108,61 @@ const StoreAddressCard = ( { showValidation = false } ) => {
 			.join( ', ' );
 
 		addressContent = (
-			<div>
+			<div className="gla-store-address-card__address">
```
eason9487: Is the `gla-store-address-card__address` CSS class used anywhere?

js/src/components/contact-information/store-address-card.js
```diff
@@ -132,38 +108,61 @@ const StoreAddressCard = ( { showValidation = false } ) => {
...
+					'google-listings-and-ads'
+				),
+				{
+					link: settingsLink,
+				}
+			) }
+		</p>
+	);
+
+	const shortDescription = (
```
eason9487: The `shortDescription` doesn't seem to be relatively short, and it's even a bit longer than the `longDescription`.

js/src/components/contact-information/store-address-card.js
```diff
@@ -132,38 +108,61 @@ const StoreAddressCard = ( { showValidation = false } ) => {
...
+					link: settingsLink,
+				}
+			) }
+		</p>
+	);
+
+	const detail = (
+		<>
+			{ addressContent }
+			{ ! isAddressFilled && (
```
eason9487, line 145: Is there any reason that the `data.isAddressFilled` got from the `useStoreAddress` hook is not applicable?

js/src/utils/urls.js
```diff
@@ -86,21 +85,13 @@ export const geReportsUrl = () => {
...
-	);
-};
 export const getEditStoreAddressUrl = () => {
 	return getNewPath(
 		{ subpath: subpaths.editStoreAddress },
 		settingsPath,
 		null
 	);
 };
-
```
eason9487: Could it keep a blank line here?

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
eason9487, line 1: https://github.com/woocommerce/google-listings-and-ads/blob/53bab8422174dd6ee07558e9ae2ff0dbbacd9c2e/js/src/setup-mc/setup-stepper/saved-setup-stepper.js#L35-L36

The two `@fires` tags need to be updated to:
```
 * @fires gla_setup_mc with `{ triggered_by: 'step1-continue-button' | 'step2-continue-button', action: 'go-to-step2' | 'go-to-step3' }`.
 * @fires gla_setup_mc with `{ triggered_by: 'stepper-step1-button' | 'stepper-step2-button', action: 'go-to-step1' | 'go-to-step2' }`.
```

js/src/hooks/useStoreAddressSynced.js
```diff
@@ -0,0 +1,65 @@
...
+import { useSelect } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
+import useGoogleMCAccount from '.~/hooks/useGoogleMCAccount';
+import { STORE_KEY } from '.~/data/constants';
+
+/**
+ * @typedef {Object} StoreAddressSyncedData
```
eason9487: The `hasFinishedResolution` property is missing.

js/src/hooks/useStoreAddressSynced.js
```diff
@@ -0,0 +1,65 @@
...
+
+			return {
+				isAddressFilled: ! missingRequiredFields.length,
+				isAddressSynced: ! isMCAddressDifferent,
+				hasFinishedResolution,
+			};
+		},
+		[ hasFinishedResolutionGoogle, isReady ]
+	);
+}
```
eason9487: Considering that the only return state used for this hook is `isAddressFilled`, it's recommended to simplify the hook to:

- Only select and return the needed state.
- Rename the hook and its file, e.g. `useStoreAddressFilled`, since the synchronization state is no longer needed.

For example:

```js
export default function useStoreAddressFilled() {
	const { isReady } = useGoogleMCAccount();

	return useSelect(
		( select ) => {
			if ( ! isReady ) {
				return null;
			}

			const contact = select( STORE_KEY ).getGoogleMCContactInformation();

			if ( ! contact ) {
				return null;
			}

			return ! contact.wc_address_errors.length;
		},
		[ isReady ]
	);
}
```

js/src/setup-mc/setup-stepper/setup-accounts/index.js
```diff
@@ -89,7 +91,20 @@ const SetupAccounts = ( props ) => {
...
+
+	const handleSubmitCallback = async () => {
+		setIsSubmitting( true );
+		await updateGoogleMCContactInformation().finally( () =>
+			// Error handling is done in the action.
+			setIsSubmitting( false )
+		);
+
+		onContinue();
+	};
```
eason9487: There are four issues:

1. The `finally()` method won't handle rejection, which leads to an uncaught error. (See video below)
2. Resetting `isSubmitting` is only needed when an error occurs rather than resetting it anyway.
3. It should not call `onContinue` when an error occurs.
4. It's not recommended to mix the `await` syntax with the Promise structure.

https://github.com/user-attachments/assets/4960b5a7-6104-4272-92f0-3cf8ee75b31a

tests/e2e/utils/pages/setup-mc/step-1-set-up-accounts.js
```diff
@@ -380,4 +380,25 @@ export default class SetUpAccountsPage extends MockRequests {
...
+	 * @return {import('@playwright/test').Locator} Get store address card.
+	 */
+	getStoreAddressCard() {
+		return this.page.locator( '.gla-store-address-card' );
+	}
+
+	/**
+	 * Get store address refresh to sync button.
+	 *
+	 * @return {import('@playwright/test').Locator} Get store address refresh to sync button.
```
eason9487:
```suggestion
	 * Get the "Update store address" button on the store address card.
	 *
	 * @return {import('@playwright/test').Locator} Get the "Update store address" button on the store address card.
```

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -616,6 +616,73 @@ test.describe( 'Set up accounts', () => {
...
+			await setUpAccountsPage.goto();
+		} );
+
+		test( 'should not be shown when MC is not connected', async () => {
+			const googleAccountCard = setUpAccountsPage.getGoogleAccountCard();
+			const storeAddressCard = setUpAccountsPage.getStoreAddressCard();
+
+			// Wait for UI to render before checking for no visibility.
+			await expect( googleAccountCard ).toBeVisible();
+			await expect( storeAddressCard ).not.toBeVisible();
```
eason9487, line 770: This test case failed.

image

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -629,6 +696,10 @@ test.describe( 'Set up accounts', () => {
 			test.beforeAll( async () => {
 				await setUpAccountsPage.mockAdsAccountConnected();
 				await setUpAccountsPage.mockMCNotConnected();
+				await setUpAccountsPage.mockContactInformation( {
+					wcAddressErrors: [],
+					isMCAddressDifferent: false,
```
eason9487, line 833: These properties could be omitted as they are irrelevant to the target of this test case.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -645,6 +716,10 @@ test.describe( 'Set up accounts', () => {
 				await setUpAccountsPage.mockAdsAccountDisconnected();
 				await setUpAccountsPage.fulfillAdsAccounts( ADS_ACCOUNTS );
 				await setUpAccountsPage.mockMCConnected();
+				await setUpAccountsPage.mockContactInformation( {
+					wcAddressErrors: [],
+					isMCAddressDifferent: false,
```
eason9487: These properties could be omitted as they are irrelevant to the target of this test case.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -656,12 +731,36 @@ test.describe( 'Set up accounts', () => {
...
+			} );
+
+			test( 'should see "Continue" button disabled when the store address needs to be updated', async () => {
+				const continueButton =
+					await setUpAccountsPage.getContinueButton();
+				await expect( continueButton ).toBeDisabled();
+			} );
+		} );
+
+		test.describe( 'When all accounts are connected and store address is synced', async () => {
```
eason9487:
```suggestion
		test.describe( 'When all accounts are connected and store address is fulfilled', async () => {
```

tests/e2e/specs/setup-mc/step-2-product-listings.test.js
```diff
@@ -362,27 +362,15 @@ test.describe( 'Configure product listings', () => {
 			await productListingsPage.fillEstimatedShippingTimes( '14' );
 		} );
 
-		test( 'should see the heading of next step and request for the contact information after clicking "Continue"', async () => {
-			const requestPromise =
-				productListingsPage.registerContinueRequest();
```
eason9487, line 367: Maybe adjust the `ProductListingsPage.prototype.registerContinueRequest` method and add assertions of waiting for a POST request to the `/gla/mc/contact-information` API into the following test group as a test case, e.g. "Should send a POST request to sync contact information when clicking on "Continue" button". Or delete the method as it's no longer used.

https://github.com/woocommerce/google-listings-and-ads/blob/53bab8422174dd6ee07558e9ae2ff0dbbacd9c2e/tests/e2e/specs/setup-mc/step-1-accounts.test.js#L756

eason9487 (changes requested):
> E2E tests are updated so this should be ready for another look.

E2E test can't pass: https://github.com/woocommerce/google-listings-and-ads/actions/runs/11803624468/job/32881993947

mikkamp (commented):
For my review I'm only addressing changes in the PHP files.

Not a big deal, but I noticed that `MerchantTrait::get_valid_business_info` still sets a valid phone number and marks it as verified. This is also still checked in `ContactInformationTest::test_get_valid_contact_information`. Since we no longer validate a phone number we don't need to keep this around.

This part isn't really related to the PHP, however I noticed that loading the account information step is rather slow (in particular for me because I have a large amount of linked test accounts). On one of the requests to `ads/accounts` the request timed out. This timed out request is not visible but it leaves the user in a state where they can not reconnect to a new ads account.

image

It might not be vital to address since in this scenario I did already have an ads account connected so I could still proceed with the connected account, also a refresh cleared it up and the call didn't timeout the second time. However I just wanted to mention the behaviour.

js/src/components/contact-information/phone-number-card/verify-phone-number-content.js
joemcgill: 428738c

js/src/components/contact-information/index.js
joemcgill: 5c129a4

js/src/components/contact-information/store-address-card.scss
joemcgill: d4c0daa90

js/src/components/contact-information/index.js
joemcgill: 5c129a48

js/src/components/contact-information/index.js
joemcgill: [49f8f82](https://github.com/woocommerce/google-listings-and-ads/pull/2653/commits/49f8f8223b91b546d56e64c32c12b9e0851892fe)

js/src/components/contact-information/index.js
joemcgill: [49f8f82](https://github.com/woocommerce/google-listings-and-ads/pull/2653/commits/49f8f8223b91b546d56e64c32c12b9e0851892fe)

js/src/components/contact-information/store-address-card.js
joemcgill: Doesn't look like it, and since the address can be missing, I've removed the redundant div and added some conditional to avoid an extra empty div.

1372d3ea

js/src/components/contact-information/store-address-card.js
joemcgill: 😆 very good point. Simplified to avoid the need for these names in e49ca66a2.

js/src/components/contact-information/store-address-card.js
joemcgill: Doesn't look like it. Updated in 299fad753.

With this change, the only place using the new `useStoreAddressSynced()` hook is the `SetupAccounts` component ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/30ac53a716cca086f4ac99ac28caf36339f9ee60/js/src/setup-mc/setup-stepper/setup-accounts/index.js#L94)) since we're no longer conditionally showing this component dependent on whether the address is synced, we could eliminate this hook altogether by updating the `useStoreAddress()` hook so it uses the `useGoogleMCAccount()` instead of directly calling `getGoogleMCContactInformation` action. That way we could bail early and avoid the error that is surfaced if there is no MC account connected.

js/src/utils/urls.js
joemcgill: 👍🏻 6ad177b99

js/src/setup-mc/setup-stepper/saved-setup-stepper.js
joemcgill: c0d25ec

js/src/hooks/useStoreAddressSynced.js
joemcgill: Updated this in eeb596bda but renamed the hook `useStoreAddressReady()` since that's really its only purpose at this point (see [this comment](https://github.com/woocommerce/google-listings-and-ads/pull/2653#discussion_r1841315594)).

I'd prefer to avoid introducing a late refactor of `useStoreAddress`, but it's something that should be considered as a follow-up later.

js/src/setup-mc/setup-stepper/setup-accounts/index.js
joemcgill: Right on. I believe a1360bf83 fixes these. Tested by throwing an Error in the try block of the `updateGoogleMCContactInformation()` action.

js/src/hooks/useStoreAddressSynced.js
joemcgill: Unnecessary after applying suggestions below.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
joemcgill: Yep, looks like other test changes merged to the feature/2509... branch caused the mocked state to change. I've addressed this in [8b141d1](https://github.com/woocommerce/google-listings-and-ads/pull/2653/commits/8b141d1396256f09e1a2eab1aba72a558dc3ab1f) after refreshing this branch.

tests/e2e/specs/setup-mc/step-2-product-listings.test.js
joemcgill: Great idea. I've handled this in [ddbb900](https://github.com/woocommerce/google-listings-and-ads/pull/2653/commits/ddbb900ebc94bffc1430dfbbe2fb38630246024d).

tests/e2e/specs/setup-mc/step-1-accounts.test.js
joemcgill: ddbb900

tests/e2e/specs/setup-mc/step-1-accounts.test.js
joemcgill: ddbb900

tests/e2e/specs/setup-mc/step-1-accounts.test.js
joemcgill: ddbb900

joemcgill:
@eason9487 Thanks for the detailed review! I think I've addressed all your concerns.

@mikkamp Thanks for reviewing as well! re:

> Not a big deal, but I noticed that MerchantTrait::get_valid_business_info still sets a valid phone number and marks it as verified. This is also still checked in ContactInformationTest::test_get_valid_contact_information. Since we no longer validate a phone number we don't need to keep this around.

I've addressed this in [866e7b6](https://github.com/woocommerce/google-listings-and-ads/pull/2653/commits/866e7b6389b18c5fae80209810d512248d402509).

> This part isn't really related to the PHP, however I noticed that loading the account information step is rather slow (in particular for me because I have a large amount of linked test accounts). On one of the requests to ads/accounts the request timed out. This timed out request is not visible but it leaves the user in a state where they can not reconnect to a new ads account.

Good call out. Given that the calls to `ads/accounts` isn't related to this specific PR, I'll look into handling it in a follow-up PR before #2509 is wrapped up.

js/src/components/contact-information/index.js
eason9487: The package-lock.json file should be updated as well after removing `libphonenumber-js`.

Updated in https://github.com/woocommerce/google-listings-and-ads/pull/2653/commits/7d49d69424fb0fff8b47ac944a43e4499bd3b902.

js/src/hooks/useStoreAddressReady.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Internal dependencies
+ */
+import useGoogleMCAccount from '.~/hooks/useGoogleMCAccount';
+import { STORE_KEY } from '.~/data/constants';
+
+/**
+ * Checks if the store address is synchronized with the Merchant Center (GMC) account address.
+ *
+ * @return {boolean} Whether the store address is ready to by synced to MC.
```
eason9487: Is "by" a typo?

```suggestion
 * @return {boolean} Whether the store address is ready to be synced to MC.
```

eason9487 (approved):
Thanks for the work.

I pushed a commit for updating the package-lock.json file (ref: https://github.com/woocommerce/google-listings-and-ads/pull/2653#discussion_r1841727653), and there seems to be a typo in JSDoc.

I believe the frontend part won't need another round. I will be approving in advance.

mikkamp (approved):
Approving the PHP changes in this PR. All code for removing the phone number verification has been removed and the address validation is confirmed for an earlier step.

Note: unit tests in the PR are failing, but these are related to WP 6.7 changes which have been fixed in the main branch.

PR #2655, opened 2024-10-30
[GTIN] GTIN Migration API Controller

puntope:

### Screenshots:

Screenshot 2024-10-29 at 19 23 48

1. Request `POST /wp-json/wc/gla/gtin-migration-start` 
2. See `{"status":"success","message":"GTIN Migration successfully started."}`
3. Repeat step 1
4. See `{"status":"error","message":"GTIN Migration cannot be scheduled."}`
5. Complete the job (all the batches)
6. Go to options and see `gla_gtin_migration_started` as 1 - true.

### Additional details:

- Tests coming in other PR
- Nonce protection coming in other PR

### Changelog entry

> Add - GTIN Migration API Controller


src/Options/OptionsInterface.php
```diff
@@ -46,6 +46,7 @@ interface OptionsInterface {
 	public const UPDATE_ALL_PRODUCTS_LAST_SYNC             = 'update_all_products_last_sync';
 	public const WP_TOS_ACCEPTED                           = 'wp_tos_accepted';
 	public const WPCOM_REST_API_STATUS                     = 'wpcom_rest_api_status';
+	public const GTIN_MIGRATION_STARTED                    = 'gtin_migration_started';
```
mikkamp: Can you clarify the purpose of this option? Based on the code in this PR it's only set to true when scheduling starts, but the value is never checked. Should this be a flag for GTIN_MIGRATION_COMPLETED instead?

I understand that in PR #2656 it's used to decide when to display the migration notice. But is that total feedback we want to give on the migration? Do we want to provide a dismissable notice when it's in progress / completed?

Comparing it with the notices that appear for WC when it should do it's DB updates, I'd expect more than just a notice that disappears once the button has been clicked. In that case would it be better to save this option as GTIN_MIGRATION_STATUS with a value of `started` or `completed`?

src/API/Site/Controllers/GTINMigrationController.php
```diff
@@ -0,0 +1,100 @@
...
+	 * @return callable
+	 */
+	protected function start_migration_callback(): callable {
+		return function ( Request $request ) {
+			try {
+				if ( ! $this->job->can_schedule( [ 1 ] ) ) {
+					return new Response(
+						[
+							'status'  => 'error',
+							'message' => __( 'GTIN Migration cannot be scheduled.', 'google-listings-and-ads' ),
```
mikkamp, line 78: Just want to confirm, that although the notice will not be shown this check means we can schedule this migration multiple times with an API request as long as it isn't currently running. As far as I can see that won't destroy any data so it should be fine.

mikkamp (commented):
Thanks for the changes, it's working as expected when testing. I just left a few questions about tracking more than just the started status.

Also noticed that this triggers a `gla/jobs/update_products/process_item` for each product we updated the GTIN for. Probably necessary but just taking note that this could schedule a lot of extra tasks.

src/API/Site/Controllers/GTINMigrationController.php
puntope: I was thinking about block when it is completed as well. My main question was about if a user wanted to run again the migration after its completed. This case seems to be not very usual as the GTIN will be read-only in the extension tabs. So no chance for the user to add new GTINS to migrate. 

So do you suggest to block it after it's completed? The banner won't show tho... so I don't think its a problem to keep it as it is.

src/Options/OptionsInterface.php
puntope: > PR https://github.com/woocommerce/google-listings-and-ads/pull/2656 it's used to decide when to display the migration notice. But is that total feedback we want to give on the migration? Do we want to provide a dismissable notice when it's in progress / completed?

I was thinking also to add another banner when it's started (showing info about it) -- Your GTIN migration is currently running. Please check WooCommerce Scheduled Actions (link to the actions) to see the status. So yes that will need GTIN_MIGRATION_STATUS ( started, completed ) instead.

> Comparing it with the notices that appear for WC when it should do it's DB updates

Another option was to create a tool page with a progress bar. But that will require more work. We can start with the 2 banners (easy to implement) and if we are not convinced we can switch to the progress bar page.

src/API/Site/Controllers/GTINMigrationController.php
mikkamp: I don't suggest to block it, in case it's useful in troubleshooting to re-trigger.
Just wanted to make a note of the behaviour, just to confirm it's expected.

src/Options/OptionsInterface.php
mikkamp: Agree, no need to overthink this, as it's a run once type thing. Just the 2 banners should be sufficient.

puntope: Added a second banner here: https://github.com/woocommerce/google-listings-and-ads/pull/2656

mikkamp (approved):
Approving this PR in case it's needed as a dependency. If the changes are going to be merged through #2656 then this can be closed.

PR #2656, opened 2024-10-30
[GTIN] Migration notice

puntope:

Related to https://github.com/woocommerce/google-listings-and-ads/pull/2655

### Screenshots:

https://github.com/user-attachments/assets/fbff5990-6b4a-4e57-994b-89a3269c644c

Screenshot 2024-10-29 at 20 48 00

1. Set this in console to view the events `localStorage.setItem( 'debug', 'wc-admin:*' )`
2. Enter into Google for WooCommerce (any page)
3. See the banner 
4. Click on "click here." and the event `wcadmin_gla_gtin_migration_banner_click` 
5. See the modal opening and event  `wcadmin_gla_modal_open`
6. Close the modal and see event `wcadmin_gla_modal_closed`
7. Repeat step 4
8. Click on "Start Migration" in the modal and event `wcadmin_gla_gtin_migration_banner_migration_start`
9. See the message "GTIN Migration was successfully scheduled" and event `wcadmin_gla_gtin_migration_banner_migration_scheduled`
10. Verify the job `gla/jobs/migrate_gtin/create_batch` is scheduled. Dont run it
11. Delete the option 
12. Repeat steps 7-8
13. See the message "Unable to start GTIN Migration" and event `wcadmin_gla_gtin_migration_banner_migration_failed`

### Changelog entry

> Add - Banner for GTIN MIgration


src/Admin/Admin.php
```diff
@@ -146,6 +146,7 @@ protected function get_assets(): array {
 					'mcId'    => $this->options->get_merchant_id() ?: null,
 					'adsId'   => $this->options->get_ads_id() ?: null,
 				],
+				'gtinMigrationStarted'     => $this->options->get( OptionsInterface::GTIN_MIGRATION_STARTED, false ),
```
mikkamp: Shouldn't this take into consideration that on some sites GTIN might not be supported. Probably a low chance that an old WC version is used, but we should take that into consideration.

mikkamp (commented):
Thanks for the changes, they are working as expected. I just left a comment about checking if GTIN is supported in core. Also as we discussed we might want to give some more feedback on when the task is completed.

The code looks good to me, but we might just want to get a quick lookover by someone a bit more familiar with react to confirm the handling of the modals is done correctly.

src/Admin/Admin.php
puntope: Agree. Maybe better to call it GTINMigrationStatus  (unavailable|available|started|completed) that checks both the availability of GTIN + the GTIN_MIRATION_STATUS.

puntope:
Hi @mikkamp I added several tweaks. 

Now we do show 2 banners depending on the status of the migration.

I also refactored some code to prevent duplication. So for that, I created a Utility class to use as a trait.

Can you take another look?

ps. Notice I bumped min version to hide the GTIN to 2.8.7

src/HelperTraits/GTINMigrationUtilities.php
```diff
@@ -0,0 +1,83 @@
...
+		if ( ! $this->is_gtin_available_in_core() || $this->should_hide_gtin() ) {
+			return MigrateGTIN::GTIN_MIGRATION_UNAVAILABLE;
+		}
+
+		return $this->options()->get( OptionsInterface::GTIN_MIGRATION_STATUS, MigrateGTIN::GTIN_MIGRATION_READY );
+	}
+
+	/**
+	 *
+	 * Get the options object.
```
mikkamp, line 76: Could we get a little more explanation on why this function is here / used. Any other class that uses the `OptionsAwareTrait` doesn't have this extra helper function because the option object is already set automatically.

src/Jobs/MigrateGTIN.php
```diff
@@ -22,6 +23,12 @@
  */
 class MigrateGTIN extends AbstractBatchedActionSchedulerJob implements OptionsAwareInterface {
 	use OptionsAwareTrait;
+	use GTINMigrationUtilities;
+
+	public const GTIN_MIGRATION_COMPLETED   = 'GTIN_MIGRATION_COMPLETED';
+	public const GTIN_MIGRATION_STARTED     = 'GTIN_MIGRATION_STARTED';
+	public const GTIN_MIGRATION_READY       = 'GTIN_MIGRATION_READY';
+	public const GTIN_MIGRATION_UNAVAILABLE = 'GTIN_MIGRATION_UNAVAILABLE';
```
mikkamp: I can understand the use of the naming within the code, but for all saved options we generally use a lower case word (without prefix). The option name is already fully prefixed as `gla_gtin_migration_status`, so can we set it as values of `completed`, `started`, `ready`, `unavailable`.

js/src/components/gtin-migration-banner/gtin-migration-banner.js
```diff
@@ -0,0 +1,181 @@
...
+							}
+						) }
+					</p>
+				</AppModal>
+			) }
+			{ status === GTIN_MIGRATION_READY && (
+				<Notice isDismissible={ false }>
+					{ createInterpolateElement(
+						__(
+							"The GTIN field managed by WooCommerce in the Product's inventory section, will now be used by Google for WooCommerce. It will continue to support the previous field and any mapping rules you have setup for the GTIN field. If you would like to migrate the data <link>click here.</link>",
```
mikkamp: I realize we aren't 100% consistent on this, but isn't it recommended to keep the punctuation outside of the link?

```suggestion
							"The GTIN field managed by WooCommerce in the Product's inventory section, will now be used by Google for WooCommerce. It will continue to support the previous field and any mapping rules you have setup for the GTIN field. If you would like to migrate the data <link>click here</link>.",
```

mikkamp (commented):
Thanks for the changes, although it seems that it's fully relying on a status that is set on the page (meaning I have to reload the page to transition the banner status).

For the banner to completely clear it's not such a problem to have it stay until the page is refreshed. But for the transition between banners it looks kind of odd. When I start the migration I'm shown the new banner:
image

But if I switch to any other tab like Product Feed, or even go to another react page in WooCommerce it reverts back to the first banner (until I hard refresh the page):
image

Can we get that status updated dynamically so it keeps the status current?

src/HelperTraits/GTINMigrationUtilities.php
puntope: It failed for me sometimes on the tests because the options doesn't seem to be initialized when the input field is called

```
4) Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Admin\ProductBlocksServiceTest::test_hook_block_template
Error: Call to a member function get() on null

/var/www/html/wp-content/plugins/google-listings-and-ads/src/HelperTraits/GTINMigrationUtilities.php:51
/var/www/html/wp-content/plugins/google-listings-and-ads/src/Admin/Product/Attributes/Input/GTINInput.php:43
/var/www/html/wp-content/plugins/google-listings-and-ads/src/Admin/Product/Attributes/Input/GTINInput.php:30
/var/www/html/wp-content/plugins/google-listings-and-ads/src/Admin/ProductBlocksService.php:276
/var/www/html/wp-content/plugins/google-listings-and-ads/src/Admin/ProductBlocksService.php:209
/var/www/html/wp-content/plugins/google-listings-and-ads/tests/Unit/Admin/ProductBlocksServiceTest.php:285
```

Any other idea to fix it?

puntope:
Hey @mikkamp Thanks for the new review. I did some changes fixing the issue with the banner not changing between tabs.  I did also rename the value of the constants.

This is ready for another round. Thanks

src/API/Site/Controllers/GTINMigrationController.php
```diff
@@ -89,6 +95,22 @@ protected function start_migration_callback(): callable {
 		};
 	}
 
+	/**
+	 * Callback function for scheduling GTIN migration job.
```
mikkamp: Seems like a copy paste error, but this is not the right description. Also, would it make more sense to call the function `get_migration_status_callback`?

src/HelperTraits/GTINMigrationUtilities.php
mikkamp: That's right, classes with `OptionsAwareTrait` only get the options object initialized if they are registered in the Container class. If they are instantiated on the fly (like the input fields), then this won't get done. That's why you need to fetch it from the container.

I'm ok with the solution here, just think that we should explain that in the docblock comment so we know why this differs from other classes.

mikkamp (approved):
Thanks for the changes, the banners are working correctly now since the status is fetched and kept up to date. I just left some minor comments.

One thing that I think should still be addressed is handling duplicate GTIN numbers. The action scheduler doesn't handle that too well, because since they are done in batches, if one product fails the whole batch fails. It tries to redo the tasks until it reaches the failure threshold.

image

Anyways I'll go ahead and approve this PR if you'd like to handle the fatal in a follow up PR.

PR #2644, opened 2024-10-16
Consolidate accounts onboarding: Claim Google Ads account via the combo card

ankitrox:
### Changes proposed in this Pull Request:

Closes #2582 

_Replace this with a good description of your changes & reasoning._

### Screenshots:

Screenshot 2024-10-16 at 8 25 44 PM

Screenshot 2024-10-16 at 8 24 53 PM

### Detailed test instructions:

1. Once the MC and Ads accounts are created for the user, you should be able to see the ads accounts claim card.
2. Click on the `Claim your Google Ads account` button, it will open the pop-up window to accept the invitation.
3. If the invitation has been accepted and pop-up is closed, there would be a `Updating...` spinner in place of `Claim your Google Ads account` button and eventually claim card will go away and in place of that green colored notice will be shown as shown in screenshot.
4. If the invitation has not been accepted, claim card will stay.

PR #2660, opened 2024-10-31
Consolidate accounts onboarding: Add edit mode to the combo card to enable it to disconnect accounts

### LINKED ISSUE #2605: Onboarding: Support editing connected accounts in GoogleComboAccountsCard

asvinb:

Closes #2605 

### Screenshots:
image

Expanded mode:
image

1. Go to the onboarding flow.
2. Once there's a connected Google account, the Edit button is displayed
3. User should be able to connect to another Google account when clicking the Edit button
4. The Google Ads card and MC cards are displayed where users can edit their respective connections.


joemcgill (changes requested):
Initial thoughts after reviewing this:

First off, since you've already merged the `update/2603-create-ads-account` for testing (which I agree is a good idea), I've updated this PR to merge into that branch and merged upstream changes so it's easier to see what changes are being introduced specifically by this PR, and to make sure commits unrelated to this PR aren't merged to the feature branch before they're ready. Please double check and make sure I didn't miss anything in my merge commit, since there were conflicts.

For the `Edit` button, I think it makes more sense to pass this to the new `actions` prop that we introduced for the `ConnectAds` and `ConnectMC` components rather than as part of the `description`. 

For consistency with other actions, these could also use the `isTertiary` button style so they match the actions elsewhere in this card.

One other thing I noticed while testing this is that once you click the `Edit` action, there is no way to toggle this back off, and connecting all accounts also don't automatically toggle edit mode (nor do I think it should, since that's a bit confusing). Instead, I think we could add a second action when the card is in edit mode to "Cancel" edit mode and go back to a collapsed or semi-collapsed state.

I've taken a pass at making these changes for you to check out.

```diff
diff --git a/js/src/components/google-account-card/switch-account-button.js b/js/src/components/google-account-card/switch-account-button.js
index 73c9712bd..8e0dbb4e0 100644
--- a/js/src/components/google-account-card/switch-account-button.js
+++ b/js/src/components/google-account-card/switch-account-button.js
@@ -27,11 +27,13 @@ const SwitchAccountButton = ( {
 		'Or, connect to a different Google account',
 		'google-listings-and-ads'
 	),
+	...restProps
 } ) => {
 	const [ handleSwitch, { loading } ] = useSwitchGoogleAccount();
 
 	return (
 		<AppButton
+			{ ...restProps }
 			isLink
 			disabled={ loading }
 			text={ text }
diff --git a/js/src/components/google-combo-account-card/connected-google-combo-account-card.js b/js/src/components/google-combo-account-card/connected-google-combo-account-card.js
index 2d7d360bf..0acc9bd53 100644
--- a/js/src/components/google-combo-account-card/connected-google-combo-account-card.js
+++ b/js/src/components/google-combo-account-card/connected-google-combo-account-card.js
@@ -95,7 +95,7 @@ const ConnectedGoogleComboAccountCard = () => {
 	}
 
 	const handleEditClick = () => {
-		setEditMode( true );
+		setEditMode( ! editMode );
 	};
 
 	const hasExistingGoogleMCAccounts = existingGoogleMCAccounts?.length > 0;
@@ -118,40 +118,44 @@ const ConnectedGoogleComboAccountCard = () => {
 		( Boolean( creatingWhich ) && ! shouldClaimGoogleAdsAccount ) ||
 		( ! showConnectAds && finalizeAdsAccountCreation );
 
+	const cardActions = (
+		<div className="gla-google-combo-account-card__description-actions">
+			{ editMode ? (
+				<>
+					<SwitchAccountButton
+						isTertiary
+						text={ __(
+							'Connect to a different Google account',
+							'google-listings-and-ads'
+						) }
+					/>
+					<AppButton
+						isTertiary
+						key="cancel"
+						onClick={ handleEditClick }
+					>
+						{ __( 'Cancel', 'google-listings-and-ads' ) }
+					</AppButton>
+				</>
+			) : (
+				<AppButton
+					isTertiary
+					text={ __( 'Edit', 'google-listings-and-ads' ) }
+					onClick={ handleEditClick }
+				/>
+			) }
+		</div>
+	);
+
 	return (
 		<div>
 			<AccountCard
 				appearance={ APPEARANCE.GOOGLE }
 				alignIcon="top"
 				className="gla-google-combo-account-card gla-google-combo-account-card--connected gla-google-combo-service-account-card--google"
-				description={
-					<>
-						{ text || <AccountDetails /> }
-
-						<div className="gla-google-combo-account-card__description-actions">
-							{ ! editMode && (
-								<AppButton
-									isLink
-									text={ __(
-										'Edit',
-										'google-listings-and-ads'
-									) }
-									onClick={ handleEditClick }
-								/>
-							) }
-
-							{ editMode && (
-								<SwitchAccountButton
-									text={ __(
-										'Connect to a different Google account',
-										'google-listings-and-ads'
-									) }
-								/>
-							) }
-						</div>
-					</>
-				}
+				description={ text || <AccountDetails /> }
 				helper={ subText }
+				actions={ cardActions }
 				indicator={ <Indicator showSpinner={ showSpinner } /> }
 			>
 				<ConnectedAdsAccountsActions
diff --git a/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss b/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
index c085e78a4..89cb10ce0 100644
--- a/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
+++ b/js/src/components/google-combo-account-card/connected-google-combo-account-card.scss
@@ -20,8 +20,4 @@
 			margin-bottom: calc((var(--large-gap)) / 2);
 		}
 	}
-
-	.gla-google-combo-account-card__description-actions {
-		margin-top: $grid-unit-10;
-	}
 }
diff --git a/js/src/components/google-combo-account-card/index.scss b/js/src/components/google-combo-account-card/index.scss
index 0f59a5c38..ddc4bf5f7 100644
--- a/js/src/components/google-combo-account-card/index.scss
+++ b/js/src/components/google-combo-account-card/index.scss
@@ -1,5 +1,5 @@
 .gla-google-combo-account-card {
-	.gla-account-card__actions .app-button {
+	.gla-account-card__actions .app-button:first-child {
 		margin-left: calc(var(--main-gap) / -2);
 	}
 }
```

~Finally, it looks like E2E tests are failing in this branch (even before my upstream merge) but I've not gotten to investigate what's happening there.~

Update: ☝🏻 these are actually running fine, so was probably a problem locally or some additional flakiness we've yet to debug.

asvinb:
I forgot about the `actions` prop 🤦  I've updated the PR, can you kindly check again please?

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -800,6 +799,44 @@ test.describe( 'Set up accounts', () => {
...
 
+	test.describe( 'Edit button', () => {
+		test.beforeAll( async () => {
+			await setUpAccountsPage.mockMCConnected();
+			await setUpAccountsPage.mockAdsAccountConnected();
+			await setUpAccountsPage.mockAdsStatusClaimed();
+			await setUpAccountsPage.goto();
+
+			const editButton = setUpAccountsPage.getEditButton();
+			await editButton.click();
```
joemcgill: Let's remove the edit button click from the beforeAll hook and create new tests that confirm

1. Initial test:
  - The edit button is visible
  - The ConnectMC and ConnectAds cards are not visible

2. Edit button click
  - The "Connect to a different Google account" is visible
  - The Cancel button is visible
  - The ConnectMC and ConnectAds cards are visible

2. Cancel button click
  - The edit button is visible
  - The ConnectMC and ConnectAds cards are not visible

joemcgill (changes requested):
Left some suggestions for improving the E2E tests. Otherwise, this is looking good.

joemcgill:
I've [confirmed](https://fueled10up.slack.com/archives/C06B7R5S90A/p1731000164958139) that inclusion of the 'Cancel' button with @fblascogarma and updated the issue description accordingly.

tests/e2e/specs/setup-mc/step-1-accounts.test.js
asvinb: Thanks @joemcgill . I've updated the E2E tests. Can you please take a look?

joemcgill (approved):
E2E Changes look good! Thanks

ankitguptaindia:
@asvinb thanks for the PR. I've tested this and overall looks good. Here are few suggestions for UI/UX improvements. 

1. When accounts cards are in expended view, all three Google, Ad, MC has their own border radius. Please see the screenshot. can we make/consider this as a single card when this is in expended view? 

image


2. Google Account Section text - `Connect to a different Google account`  
Screenshot 2024-11-08 at 6 25 45 PM

When account is already connected - can we make this text to `Or, connect to a different Google account`. 
Currently it sounds like user has to still connect with Google account. 


Other Functional Use Cases are working as expected as described in AC.

joemcgill:
Thanks for the feedback, @ankitguptaindia.

1. We can't make this whole component a single card, but perhaps we can improve the styling to avoid the extra border radius (it's not as visible in my testing as the screenshot you produced)

2. Good catch on the wording. It should indeed be "Or, connect to a different Google account" per the designs.

joemcgill:
Let's wait and update this after we've resolved changes in 2603, which are causing conflicts.

joemcgill:
@ankitguptaindia QA review items have been addressed if you want to confirm.

@eason9487 I'll resolve conflicts once #2603 is resolved, but this should be ready for initial review.

ankitguptaindia:
**QA Report-** ✅ 
------------



Issues reported here in https://github.com/woocommerce/google-listings-and-ads/pull/2660#issuecomment-2464752937

Both concerns have been addressed. PR is ready for next step.

Screenshot 2024-11-12 at 3 41 55 PM

ankitguptaindia (approved):

js/src/components/google-ads-account-card/connect-ads/index.js
```diff
@@ -84,7 +84,7 @@ const ConnectAds = ( props ) => {
 			<Section.Card.Body>
 				<Subsection.Title>
 					{ __(
-						'Connect to an existing account',
+						'Or, connect to an existing account',
```
eason9487: Not sure why this wording was changed? This wording is only displayed if it's not yet connected to an existing account. The original looks more appropriate.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -66,22 +79,57 @@ const ConnectedGoogleComboAccountCard = () => {
...
 	}, [ finalizeAdsAccountCreation, upsertAdsAccount, invalidateResolution ] );
 
+	useEffect( () => {
+		setShowConversionMeasurementNotice(
+			( googleAdsAccount?.status ===
+				GOOGLE_ADS_ACCOUNT_STATUS.CONNECTED ||
+				googleAdsAccount?.step === 'link_merchant' ) &&
+				! shouldClaimGoogleAdsAccount
+		);
+	}, [ googleAdsAccount, shouldClaimGoogleAdsAccount ] );
```
eason9487: Related to https://github.com/woocommerce/google-listings-and-ads/pull/2644#discussion_r1833445209 and https://github.com/woocommerce/google-listings-and-ads/pull/2644#discussion_r1833997070.

> There is not an explicit requirement for this, but the design doc implies that after all accounts finish connecting and the combo card collapses, that this would no longer be shown.

> Also I think it's safe to say that whenever the "Cancel" button is clicked, we can hide the notice.

Although I still don't think the design document implies this logic (Ref: https://github.com/woocommerce/google-listings-and-ads/pull/2644#discussion_r1833656584), if it's determined that the quoted logic is to be implemented, then the current implementation doesn't seem to satisfy it.

https://github.com/user-attachments/assets/3ef71ac6-fcd0-498b-95dc-7373e129ab94

In the video, there are two issues:

1. If the initial state is that all accounts have finished connecting, the collapsed card still shows the notice.
2. When all accounts have finished connecting and the card is re-expanded, the notice is not shown.

Could it be clarified which logic is intended to be adopted?

js/src/components/google-account-card/switch-account-button.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {string} [props.text="Or, connect to a different Google account"] Text to display on the button
+ */
+const SwitchAccountButton = ( {
+	text = __(
+		'Or, connect to a different Google account',
+		'google-listings-and-ads'
+	),
+	...restProps
+} ) => {
+	const [ handleSwitch, { loading } ] = useSwitchGoogleAccount();
```
eason9487, line 32: It was not introduced by this PR but the `useSwitchGoogleAccount` hook needs to disconnect Google Ads account as well. Maybe the [`disconnectAllAccounts` action](https://github.com/woocommerce/google-listings-and-ads/blob/ba22a30b21d62392c95855eeaed1aa2c812a4cac/js/src/data/actions.js#L511-L538) could do it.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -66,22 +79,57 @@ const ConnectedGoogleComboAccountCard = () => {
...
+
 	if ( ! hasDetermined ) {
 		return <SpinnerCard />;
 	}
 
-	// @TODO: edit mode implementation in 2605
-	const editMode = false;
-	const shouldClaimGoogleAdsAccount = Boolean(
-		googleAdsAccount?.id && hasAccess === false
+	const onCancelClick = () => {
```
eason9487: Prefer to use `handleCancelClick` to conform to the naming convention.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -66,22 +79,57 @@ const ConnectedGoogleComboAccountCard = () => {
...
+		setShowConversionMeasurementNotice( false );
+		setEditMode( false );
+	};
+
+	const onEditClick = () => {
```
eason9487: Prefer to use `handleEditClick` to conform to the naming convention.

eason9487 (changes requested):

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: [660ea2a](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/660ea2a9e620841c728d04c8ce1560064ce65398)

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: [660ea2a](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/660ea2a9e620841c728d04c8ce1560064ce65398)

js/src/components/google-ads-account-card/connect-ads/index.js
joemcgill: Good catch. This was applied to the wrong component in addition to the `ConnectedGoogleComboAccountCard` component in [0f0aa16](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/0f0aa16b8cd7f7f4cfe87bfafc36437900b92933). Reverted in [df93210](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/df93210dfebcad82fb82de5286c44e1714be22da).

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: This was removed in #2644 but those upstream changes have not yet been merged back to this branch. I plan to handle this as part of resolving merge conflicts before asking for another review.

js/src/components/google-account-card/switch-account-button.js
joemcgill: Good catch! I opted to use an approach consistent to how the MC account is currently disconnected in [3b9dbbb](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/3b9dbbbf19d60ab999d7c5d59ca91d589724def2). I believe the `disconnectAllAccounts` action will also cause the WP.org account to be disconnected, which we don't want.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: This is no longer present after resolving upstream merge conflicts.

joemcgill:
@eason9487 This has been refreshed and is ready for another look.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -32,15 +35,16 @@ import './connected-google-combo-account-card.scss';
  * It will also kickoff Ads and Merchant Center account creation if the user does not have accounts.
  */
 const ConnectedGoogleComboAccountCard = () => {
+	const [ editMode, setEditMode ] = useState( false );
+	const { hasDetermined, creatingWhich } = useAutoCreateAdsMCAccounts();
```
eason9487: The `createMCAccount` function should be passed in as a parameter.

image

joemcgill: [0d63a35](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/0d63a3591fd22d0dd16a8d4a1a8f16f2c7f47925)

joemcgill:
While doing additional testing of this PR after updating with all of the upstream changes from the target branch, I noticed that when auto-creating accounts, that the MC account card would not be visible if it needed to be reclaimed. 

Part of the issue is that when a new MC account is created, the new account does not immediately get returned in the list of existing accounts due to what is likely a cache delay with that API. 

To account for this, I've updated the logic for when `showConnectMC` and `showConnectAds` should be true to account for cases where we have an ID connected but the account may not yet be completely set up, along with some other small UI adjustments in [2ea19b8](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/2ea19b87997830aa8ba5a39c25dcbada7a8675a7).

I also added another E2E test to confirm that the UI is displaying the correct flows for Reclaiming an MC account and Claiming a new Ads account after accounts are auto-created, since this important interaction was not being covered by our current set of tests.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -69,27 +81,87 @@ const ConnectedGoogleComboAccountCard = () => {
...
+	// After creating a new account, it may be connected but not ready
+	// (e.g., needing to reclaim the URL). In this case, we show the ConnectMC
+	// component, even if the existing accounts list has not yet updated.
 	const showConnectMC =
-		( editMode && hasExistingGoogleMCAccounts ) ||
-		( ! isGoogleMCReady && hasExistingGoogleMCAccounts );
-
-	const hasExistingGoogleAdsAccounts = existingGoogleAdsAccounts?.length > 0;
+		( googleMCHasError ||
+			hasGoogleMCConnection ||
```
eason9487: The `hasGoogleMCConnection ||` seems to be redundant?

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -69,27 +81,87 @@ const ConnectedGoogleComboAccountCard = () => {
...
+			hasExistingGoogleMCAccounts ) &&
+		( editMode || ! isGoogleMCReady );
+
+	// After creating a new account, it may not show up in the existing accounts list
+	// immediately. In this case, we show the ConnectAds component in edit mode unless
+	// we're showing the claim notice in the upper card.
 	const showConnectAds =
-		( ( editMode && hasExistingGoogleAdsAccounts ) ||
-			( ! isConnected && hasExistingGoogleAdsAccounts ) ) &&
+		( isConnected || hasExistingGoogleAdsAccounts ) &&
```
eason9487: The `isConnected ||` seems to be redundant?

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -69,27 +81,87 @@ const ConnectedGoogleComboAccountCard = () => {
...
+		( editMode || ! isConnected ) &&
 		! shouldClaimGoogleAdsAccount;
 
+	// When Ads and MC are disconnected in edit mode, exit edit mode.
+	useEffect( () => {
+		if ( editMode && ! isGoogleMCReady && ! isConnected ) {
+			setEditMode( false );
+		}
+	}, [ editMode, isConnected, isGoogleMCReady ] );
```
eason9487: This doesn't work well when Google Ads account is waiting for claiming and Google Merchant Center account is not connected yet.

https://github.com/user-attachments/assets/80f49b3d-6195-4197-bfc4-5688e703f24a

js/src/components/google-mc-account-card/reclaim-url-card/index.js
```diff
@@ -32,13 +32,19 @@ import AppInputLinkControl from '.~/components/app-input-link-control';
...
  * @param {string} props.websiteUrl Website's URL
  * @param {Function} [props.onSwitchAccount] Callback when clicking on Switch Account
  * @fires gla_mc_account_reclaim_url_button_click
  * @fires gla_mc_account_switch_account_button_click with `context: 'reclaim-url'`
  * @fires gla_documentation_link_click with `{ context: 'setup-mc', link_id: 'claim-url', href: 'https://support.google.com/merchants/answer/176793' }`
  */
-const ReclaimUrlCard = ( { id, websiteUrl, onSwitchAccount = noop } ) => {
+const ReclaimUrlCard = ( {
+	id,
+	hasExisting,
```
eason9487: Suggest changing this component to get this value from the `useExistingGoogleMCAccounts` hook directly instead of passing a prop across two layers.

tests/e2e/utils/mock-requests.js
```diff
@@ -241,10 +241,17 @@ export default class MockRequests {
 	 * Fulfill the Ads Account request.
 	 *
 	 * @param {Object} payload
+	 * @param {number} status
```
eason9487:
```suggestion
	 * @param {number} [status=200]
```

eason9487 (commented):
Noticed another issue, though it was not introduced by this PR.

https://github.com/woocommerce/google-listings-and-ads/blob/8d93a0ee7a3937aef7e461d075a846d80b706d84/js/src/components/google-combo-account-card/connected-google-combo-account-card.js#L175

This condition may need an adjustment since if refreshing the webpage during the reclaim URL state, it will show the store address card.

image

image

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: Good catch! [08f96a7](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/08f96a74c81a35845c0a540e69f38939941ee639) should make this better.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: I don't believe it is. Consider the case where a new account is initially created but is not yet claimed and is not yet returned in the list of existing accounts. In that case, we still need to show the ConnectMC card so that the reclaim flow can be shown.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
joemcgill: Similar to showConnectMC case above, A new Ads account might be connected but not yet returning in the existing accounts list. In that case, it should be shown when we enter editMode.

js/src/components/google-mc-account-card/reclaim-url-card/index.js
joemcgill: I debated that as well. Done in [392bdd7](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/392bdd7b3aed0a1f334d72e20fc355595ea3d96d).

joemcgill:
I've made a few additional updates to address some small UI issues that were caught by @fblascogarma during UAT:

- [109b724](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/109b724f195ea3929a2618edf425d7f89ed34d10) – Add and external icon to the "Claim account in Google Ads" button.
- [17dab80](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/17dab80ef13429ccb042946871961089382122b3) – Tweak wording when an Ads account is being created.

And also changed the logic for showing the address card in [0348394](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/03483941bcdf09c6540a0c298d13a91a9d91c76d), so that it only shows once MC is ready, as pointed out in [this comment](https://github.com/woocommerce/google-listings-and-ads/pull/2660#pullrequestreview-2438014156):

> This condition may need an adjustment since if refreshing the webpage during the reclaim URL state, it will show the store address card.

joemcgill:
I realized during more testing that attempting to keep the MC Reclaim flow visible when someone refreshes leads to a number of additional issues, including:

- The `Switch account` button in the `ReclaimUrlCard` expects to use the `result.reset()` method returned from `useConnectMCAccount()` or `useCreateMCAccount()` hooks.
- The `ReclaimUrlCard` also displays the ID and URL based on the response from the `gla/mc/accounts` API, which would need to be replaced.

Given those challenges, and the fact that the current setup flow doesn't account for refreshing the page when the MC `ReclaimUrlCard` is shown, I've opted to revert those changes and modify the `AccountCard`  in ff47280 so it shows the original connection flow if the page is refreshed while the Reclaim is pending.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
eason9487: After cross-testing, it was realized that this is for a more extreme case. It's recommended to add code comment to note that the existence of `hasGoogleMCConnection ||` condition is for the scenario with the following steps:
1. Automatically create a Google Merchant Center account
2. Reclaim URL is required
3. The merchant interrupts the onboarding flow and then resumes it
   - This is also the case for refreshing webpage
4. The newly created account is not yet among the existing accounts
5. The condition enables the merchant to resume the Google Merchant Center connection from the step of connecting the newly created account

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
eason9487: In the video, the `isConnected ||` condition at line 111 was commented out. After the new Google Ads account was automatically created and claimed. The newly created account was shown in the existing accounts. Not sure if I was missing anything?

https://github.com/user-attachments/assets/1f564317-f993-427f-a863-fdd1b2567942

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
eason9487: It's still not working well.

https://github.com/user-attachments/assets/fdde7fbd-9152-4151-a457-2db1da5e9e17

js/src/components/google-mc-account-card/connect-mc/index.js
```diff
@@ -52,13 +52,14 @@ const ConnectMC = ( { createAccount, resultCreateAccount, className } ) => {
 		googleMCAccount,
 		hasFinishedResolution,
 		isReady: isGoogleMCReady,
+		hasGoogleMCConnection,
 	} = useGoogleMCAccount();
 
 	useEffect( () => {
-		if ( isGoogleMCReady ) {
+		if ( hasGoogleMCConnection ) {
```
eason9487, line 59: https://github.com/woocommerce/google-listings-and-ads/blob/6a7d33e1a6ec99b458534f5a559468a17a4add31/js/src/components/google-mc-account-card/connect-mc/index.js#L155-L156

Looks like the state for `isConnected` prop needs more consideration, otherwise it may show a blank line instead of the account selector control.

image

js/src/components/google-combo-account-card/claim-ads-account/claim-ads-account.js
```diff
@@ -40,13 +43,15 @@ const ClaimAdsAccount = () => {
 					'google-listings-and-ads'
 				) }
 			</p>
-			<ClaimAccountButton
-				text={ __(
-					'Claim account in Google Ads',
-					'google-listings-and-ads'
+			<ClaimAccountButton isPrimary>
+				{ createInterpolateElement(
```
eason9487: `createInterpolateElement` could be omitted.

eason9487 (commented):
After more cross-testing in different scenarios, it turns out that the current edit mode is still a bit awkward compared to the original. For example:

- If I would like to disconnect a created but not claimed Google Ads account, the options are:
   - Claim it first and then disconnect it
   - Disconnect Google account and redo all account connections again

   image
- The situation where the newly created Google Merchant Center account won't be listed in existing accounts for a while might lead to a status that can not continue the onboarding.
   - See https://github.com/woocommerce/google-listings-and-ads/pull/2660#discussion_r1845897475
- If an error occurs during the (automatic) creation of Google Merchant Center account, it may be stuck in a situation where the account can not be connected nor disconnected.
  - This issue is similar to https://github.com/woocommerce/google-listings-and-ads/pull/2660#pullrequestreview-2438014156 and it might be more about the API design issue. Because when that error occurs, the connection states could be:
      ```json
      {
        "id": 0,
        "status": "incomplete",
        "step": "set_id"
      }
      ```
   - The following video was tested with a non-publicly accessible website, it shows some issues:
      - The Edit button didn't work
      - After the webpage was refreshed, error occurred - "There was an error loading Google Merchant Center contact information."
      - After the webpage was refreshed, the Connect button didn't work
      - After the webpage was refreshed, it couldn't disconnect the incomplete Google Merchant Center account unless disconnecting Google account

      https://github.com/user-attachments/assets/15b395f6-7b8d-4828-b697-caf53ffc10e4

js/src/components/google-combo-account-card/claim-ads-account/claim-ads-account.js
joemcgill: fa5a2d39a

joemcgill:
I've made a few more adjustments that I think resolves the remaining issues:

> If I would like to disconnect a created but not claimed Google Ads account, the options are

Now, when a new Ads account is awaiting a claim, you can click "Edit" to change the connection ([93d588e](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/93d588ed5f4a6acd96c5193125afeee0a339d2ff)). If there are not other existing accounts, the connected but not claimed account will be shown in the UI similar to how the ConnectMCCard handles this situation ([93d588e](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/93d588ed5f4a6acd96c5193125afeee0a339d2ff)).

> The situation where the newly created Google Merchant Center account won't be listed in existing accounts for a while might lead to a status that can not continue the onboarding.

Now, for both MC and Ads accounts, if in edit mode and there are no existing accounts to change to, we show the "Create a new ... account" button so that the action makes sense and so we don't leave the user in a situation where we disconnect from the current account with no other account to connect to ([fd0ad0e](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/fd0ad0e1fe98bfa63fe8295272ce3754c721786c))

> If an error occurs during the (automatic) creation of Google Merchant Center account, it may be stuck in a situation where the account can not be connected nor disconnected.

This one is harder to reproduce, because it only seems to occur if an error happens while trying to complete the `set_id` setup step for MC accounts. In this case, the status will be `incomplete` and the `id` will not be set, i.e. `id: 0`. The `useGoogleMCAccount()` hook, which is used in many places to determine whether we have an MC connection doesn't take the ID into account for the value of `hasGoogleMCConnection`, so I've made that a requirement in [58820c1](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/58820c1bd5ec25e3730d3b82020101533c580a42). This also avoids the issue described in https://github.com/woocommerce/google-listings-and-ads/pull/2660#discussion_r1845899613 where the MC card is shown without a selector.


In addition to these changes, I've added some more E2E tests to cover the new edit button considerations and improved the way the border-radius for the combo card was being handled since some UI was not being styled appropriately.

js/src/components/google-combo-account-card/connect-ads/connect-ads-footer.js
```diff
@@ -23,7 +24,9 @@ const ConnectAdsFooter = ( {
 	onCreateNewClick,
 	...restProps
 } ) => {
-	if ( isConnected ) {
+	const { existingAccounts } = useExistingGoogleAdsAccounts();
+
+	if ( isConnected && existingAccounts.length > 1 ) {
```
eason9487: The condition is incorrect since it still shows the creating account button when an account is already connected and there is only one existing account.

https://github.com/user-attachments/assets/0d3801bd-796a-44b4-a326-0b4d51ef425e

And it won't work well when the account is automatically created but not claimed yet.

https://github.com/user-attachments/assets/0601f071-3c9f-41eb-af3e-005a0d838ff1

js/src/components/google-mc-account-card/connect-mc/actions.js
```diff
@@ -27,7 +28,9 @@ const Actions = ( {
 	resultCreateAccount,
 	onCreateAccount,
 } ) => {
-	if ( isConnected ) {
+	const { data: existingGoogleMCAccounts } = useExistingGoogleMCAccounts();
+
+	if ( isConnected && existingGoogleMCAccounts.length > 1 ) {
```
eason9487: The condition is incorrect since it still shows the creating account button when an account is already connected and there is only one existing account.

https://github.com/user-attachments/assets/fb3ffcf5-3020-4dc6-a48f-9ec40fd8cf97

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -921,6 +955,124 @@ test.describe( 'Set up accounts', () => {
...
+				const editButton = setUpAccountsPage.getEditButton();
+				await editButton.click();
+
+				const createNewAdsAccountButton =
+					setUpAccountsPage.getCreateNewAdsAccountButton();
+				const createNewMCAccountButton =
+					setUpAccountsPage.getCreateNewMCAccountButton();
+
+				await expect( createNewAdsAccountButton ).toBeVisible();
+				await expect( createNewMCAccountButton ).toBeVisible();
```
eason9487, line 1089: When the accounts are connected, the functionality of these buttons should not be to create a new account.

eason9487 (changes requested):
Apart from the two newly introduced issues, the issue in https://github.com/woocommerce/google-listings-and-ads/pull/2660#discussion_r1845899613 is not addressed.

js/src/components/google-mc-account-card/connect-mc/index.js
joemcgill: This should be handled in [2b213f4](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/2b213f452fd7566026982bac947b8857401d28b9).

joemcgill:
@eason9487 for the two cases where we are showing the "Create new account" button I've modified this logic slightly in [87049d8](https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/87049d810b7b1cad134288fd663f55164f2c739b) to only show the "Create new account" button only if there are no existing accounts. Otherwise, disconnecting the currently connected account leaves the UI in an unrecoverable state because there will be no other accounts to connect to. Here's a video of showing an example when disconnecting a newly created MC account when there are no other accounts to connect to.

https://github.com/user-attachments/assets/f9a4815d-9eef-4394-b796-8ee4c75cf887

Having a connected account while there are no existing accounts can happen in two scenarios that we found:

1. For MC accounts, the existing accounts query from the Google API doesn't seem to refresh immediately after a new account is created. So if a new account is automatically created, there will be no existing accounts right away. In this state, if someone disconnects and then refreshes the page, it will trigger an additional automatic account creation, which is undesirable. In this scenario, giving them the option to create a new account is the only action they could take.

2. For new Ads accounts, the account will not be listed in existing accounts until after the account is claimed. If someone decides to edit their accounts rather than complete the claim, and disconnect, there will be no other accounts to connect to.

js/src/components/google-combo-account-card/connect-ads/connect-ads-footer.js
eason9487, line 1: When there is no existing Google Ads account and the newly created one is waiting to be claimed, clicking on the creating button won't create a new one, but will go to the update account processing due to the `useUpsertAdsAccount` mechanism.

https://github.com/user-attachments/assets/7afcbbb0-b96f-4440-b2a2-9f36e2cc765a

In the video, after click the button to create a new account
- The status is "Connecting…"
- The POST request to `wc/gla/ads/accounts` API has a `{ id: [account ID waiting for claiming] }` payload
- After the requests are completed, it's still the same account

Suggest in *this case* to make it either:
- Can actually create a new account
- Make it no option to create a new one or disconnect an incomplete one

asvinb: @eason9487 I've made the changes in the following commit https://github.com/woocommerce/google-listings-and-ads/pull/2660/commits/883d1dbe701b212e5c9593c9cb2260602abf29b6. Let me know what you think.

js/src/components/google-combo-account-card/connect-ads/connect-ads-footer.js
```diff
@@ -8,27 +8,47 @@ import { __ } from '@wordpress/i18n';
...
+import useGoogleAdsAccountStatus from '.~/hooks/useGoogleAdsAccountStatus';
+import useGoogleAdsAccount from '.~/hooks/useGoogleAdsAccount';
 
 /**
  * Footer component.
  *
  * @param {Object} props Props.
  * @param {boolean} props.isConnected Whether the account is connected.
  * @param {Function} props.onCreateNewClick Callback when clicking on the button to create a new account.
+ * @param {boolean} props.disabled Whether the button is disabled.
```
eason9487:
```suggestion
 * @param {boolean} props.disabled Whether to disable the create account button.
```

tests/e2e/specs/setup-mc/step-1-accounts.test.js
```diff
@@ -921,6 +955,142 @@ test.describe( 'Set up accounts', () => {
...
+
+				const connectDifferentGoogleAdsAccountButton =
+					setUpAccountsPage.getConnectDifferentAdsAccountButton();
+
+				await expect(
+					connectDifferentGoogleAdsAccountButton
+				).toBeVisible();
+			} );
+
+			test( 'should not let you connect to a different account if there are no other accounts', async () => {
```
eason9487:
```suggestion
			test( 'should disable the create new account button if there are no other existing accounts', async () => {
```

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -69,26 +79,87 @@ const ConnectedGoogleComboAccountCard = () => {
...
+
+	// During MC account creation, we need to show the ConnectMC component
+	// when the account creation needs to show the Switch or Reclaim flow.
+	const googleMCHasError = [ 409, 403 ].includes(
+		resultCreateMCAccount.response?.status
+	);
+
+	// After creating a new account, it may be connected but not ready
+	// (e.g., needing to reclaim the URL). In this case, we show the ConnectMC
+	// component, even if the existing accounts list has not yet updated.
```
eason9487, line 98:
```suggestion
	// component, even if the existing accounts list has not yet updated.
	//
	// The last `hasGoogleMCConnection` condition exists for the scenario with these steps:
	// 1. Automatically creating a Google Merchant Center account and reclaiming URL is required.
	// 2. The merchant interrupts the onboarding flow and then resumes it.
	//    This is also the case for refreshing webpage.
	// 3. The newly created account is not yet among the existing accounts.
	// 4. The condition enables the merchant to resume the Google Merchant Center connection
	//    from the step of connecting the newly created account
```

eason9487 (approved):

PR #2662, opened 2024-11-01
[GTIN] Add WP CLI command for migration

puntope:

This PR creates a WP CLI GTIN Migration command.

### Screenshots:

Screenshot 2024-11-01 at 12 17 25

1. Checkout the PR and prepare some products for GTIN migration (it should be published variation or simple and have the GTIN in Google for WooCommerce tab and not in Inventory)
2. Run WP CLI  `wp wc g4wc gtin-migration start` 
3. See the GTIN migrated. 

### Changelog entry

> Add - WP CLI Command for GTIN Migration

### DIFF
--- a/src/Utility/WPCLIMigrationGTIN.php
+++ b/src/Utility/WPCLIMigrationGTIN.php
@@ -177,7 +177,7 @@ class WPCLIMigrationGTIN implements Service, Registerable {
 			if ( $gtin ) {
 				$product->set_global_unique_id( $gtin );
 				$product->save();
-				WP_CLI::success( sprintf( 'GTIN [ %s ] has been migrated for Product ID: %s - %s', $gtin, $product->get_id(), $product->get_name() ) );
+				WP_CLI::debug( sprintf( 'GTIN [ %s ] has been migrated for Product ID: %s - %s', $gtin, $product->get_id(), $product->get_name() ) );
 				++$processed;
 			}
 		}
--- a/src/Utility/WPCLIMigrationGTIN.php
+++ b/src/Utility/WPCLIMigrationGTIN.php
@@ -190,7 +190,7 @@ class WPCLIMigrationGTIN implements Service, Registerable, Conditional {
 					WP_CLI::debug( sprintf( 'GTIN [ %s ] has been migrated for Product ID: %s - %s', $gtin, $product->get_id(), $product->get_name() ) );
 					++$processed;
 				} catch ( Exception $e ) {
-					WP_CLI::error( $e->getMessage(), false );
+					WP_CLI::error( sprintf('GTIN [ %s ] for Product ID: %s - %s has an error - %s', $gtin, $product->get_id(), $product->get_name(), $e->getMessage()), false );
 				}
 			} else {
 				WP_CLI::debug( sprintf( 'GTIN [ %s ] has been skipped for Product ID: %s - %s. Invalid GTIN was found.', $gtin, $product->get_id(), $product->get_name() ) );
--- a/src/Utility/WPCLIMigrationGTIN.php
+++ b/src/Utility/WPCLIMigrationGTIN.php
@@ -183,17 +183,19 @@ class WPCLIMigrationGTIN implements Service, Registerable, Conditional {
 			}
 
 			$gtin = str_replace( '-', '', $gtin );
-			if ( is_numeric( $gtin ) ) {
-				try {
-					$product->set_global_unique_id( $gtin );
-					$product->save();
-					WP_CLI::debug( sprintf( 'GTIN [ %s ] has been migrated for Product ID: %s - %s', $gtin, $product->get_id(), $product->get_name() ) );
-					++$processed;
-				} catch ( Exception $e ) {
-					WP_CLI::error( sprintf('GTIN [ %s ] for Product ID: %s - %s has an error - %s', $gtin, $product->get_id(), $product->get_name(), $e->getMessage()), false );
-				}
-			} else {
+
+			if ( ! is_numeric( $gtin ) ) {
 				WP_CLI::debug( sprintf( 'GTIN [ %s ] has been skipped for Product ID: %s - %s. Invalid GTIN was found.', $gtin, $product->get_id(), $product->get_name() ) );
+				continue;
+			}
+
+			try {
+				$product->set_global_unique_id( $gtin );
+				$product->save();
+				WP_CLI::debug( sprintf( 'GTIN [ %s ] has been migrated for Product ID: %s - %s', $gtin, $product->get_id(), $product->get_name() ) );
+				++$processed;
+			} catch ( Exception $e ) {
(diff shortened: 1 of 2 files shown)


src/Utility/WPCLIMigrationGTIN.php
```diff
@@ -0,0 +1,187 @@
...
+	public function __construct( ProductRepository $product_repository, AttributeManager $attribute_manager ) {
+		$this->product_repository = $product_repository;
+		$this->attribute_manager  = $attribute_manager;
+	}
+
+	/**
+	 * Register service and initialize hooks.
+	 */
+	public function register(): void {
+		WP_CLI::add_hook( 'after_wp_load', [ $this, 'register_commands' ] );
```
mikkamp, line 48: I'm not sure how we can setup a environment without WP_CLI, but based on the [doc here](https://docs.wpvip.com/vip-cli/wp-cli-with-vip-cli/write-custom-wp-cli-commands/#h-include-only-when-the-wp-cli-is-being-used) it sounds like it might not always be available. And since we aren't doing any conditional checks it seems like it would fatal here.

Should we make this class implement `Conditional` and add a `is_needed` function to check if WP_CLI is available?

src/Utility/WPCLIMigrationGTIN.php
```diff
@@ -0,0 +1,187 @@
...
+			// process variations
+			if ( $product instanceof \WC_Product_Variable ) {
+				$variations = $product->get_children();
+				$processed += $this->process_items( $variations );
+				continue;
+			}
+
+			$gtin = $this->attribute_manager->get_value( $product, 'gtin' );
+			if ( $gtin ) {
+				$product->set_global_unique_id( $gtin );
```
mikkamp, line 193: This call can throw an error, for example in cases where the GTIN is duplicate (which we didn't previously validate in the GLA data. When I test this I see the following error:

```
Fatal error: Uncaught WC_Data_Exception: Invalid or duplicated GTIN, UPC, EAN or ISBN. in woocommerce/includes/abstracts/abstract-wc-data.php:943
```

I didn't test this scenario fully with the regular importer, but I'm pretty sure it will trigger one of the scheduled actions to fail, which should probably be okay, but maybe it should be a bit more graceful so it's not a "fatal error".

src/Utility/WPCLIMigrationGTIN.php
```diff
@@ -0,0 +1,187 @@
...
+				$product->save();
+				WP_CLI::success( sprintf( 'GTIN [ %s ] has been migrated for Product ID: %s - %s', $gtin, $product->get_id(), $product->get_name() ) );
```
mikkamp: This success message also appears for invalid GTIN's, because `set_global_unique_id` filters out incorrect characters. That makes the output rather odd:

I get this in the output:
```
Success: GTIN [ abcd ] has been migrated for Product ID: 115 - Variable No Sync - Tiny
GTIN Migration  13 % [=============>                                                                                        ] 0:01 / 0:14
Success: GTIN [ yoast_seo ] has been migrated for Product ID: 2556 - Aerodynamic Leather Wallet - Small
```

`abcd` wouldn't have migrated, and how are we handling `yoast_seo` values? Do we not import them from the Yoast data?

mikkamp (commented):
The CLI command is working, but as explained in the comments I think we should handle fatals as well.

I also found the output to be a bit odd mixing the progress bar with output for each product, as it causes the progress bar to jump to multiple lines and there are no new lines in between:
image

How do other CLI commands handle that? Is it possible to only output status per product when we set a verbose / debug level of output?

src/Utility/WPCLIMigrationGTIN.php
puntope: Tweaked here https://github.com/woocommerce/google-listings-and-ads/pull/2662/commits/31097f66d8923280d7f786f50a41bbb624165d1f

src/Utility/WPCLIMigrationGTIN.php
puntope: Tweaked here https://github.com/woocommerce/google-listings-and-ads/pull/2662/commits/173824b0497177db38028d23d175b8624dce9ad4

puntope:
Hey @mikkamp Thanks for your suggestions. I tweaked the code addressing them:

Since now some of the logs are debug mode. You need to add `--debug` to see them.

puntope:
> how are we handling yoast_seo values? 

Comming in another PR (we need to do it also for the AS version)

src/Utility/WPCLIMigrationGTIN.php
```diff
@@ -0,0 +1,212 @@
...
+
+			$gtin = str_replace( '-', '', $gtin );
+			if ( is_numeric( $gtin ) ) {
+				try {
+					$product->set_global_unique_id( $gtin );
+					$product->save();
+					WP_CLI::debug( sprintf( 'GTIN [ %s ] has been migrated for Product ID: %s - %s', $gtin, $product->get_id(), $product->get_name() ) );
+					++$processed;
+				} catch ( Exception $e ) {
+					WP_CLI::error( $e->getMessage(), false );
```
mikkamp: Can we combine this error with the product ID, so we can know what to fix? Right now I'm getting this as output:

```
Starting GTIN migration for 2290 products in the store.
GTIN Migration  13 % [============>                                                                            ] 0:01 / 0:15
Error: Invalid or duplicated GTIN, UPC, EAN or ISBN.
GTIN Migration  100% [=========================================================================================] 0:01 / 0:15
Warning: No GTIN were migrated.
```

Which leaves me having to manually figure out which product I should fix.

src/Utility/WPCLIMigrationGTIN.php
```diff
@@ -0,0 +1,212 @@
...
+			}
+
+			$gtin = $this->attribute_manager->get_value( $product, 'gtin' );
+			if ( ! $gtin ) {
+				WP_CLI::debug( sprintf( 'GTIN has been skipped for Product ID: %s - %s. No GTIN was found', $product->get_id(), $product->get_name() ) );
+				continue;
+			}
+
+			$gtin = str_replace( '-', '', $gtin );
+			if ( is_numeric( $gtin ) ) {
```
mikkamp: Small nitpick, but I would have used `! is_numeric` here with a `continue` like is done for the error checking above. That way we can keep the condition and error message right next to each other. And we don't have to indent the try and catch another time.

mikkamp (approved):
Thanks for the changes, it's working smoother now. I can correctly see it importing GTIN numbers now, and it doesn't halt the process when a duplicate is found.

```
Starting GTIN migration for 2290 products in the store.
GTIN Migration  13 % [======>                                                                                            ] 0:01 / 0:08
Error: Invalid or duplicated GTIN, UPC, EAN or ISBN.
GTIN Migration  100% [===================================================================================================] 0:01 / 0:08
Success: 1 GTIN was migrated in 1 seconds.
```

I just left a few small comments, but I'll go ahead and approve it since it pretty much ready to go.

Issue #2617, opened 2024-09-17
[GTIN] Retire the existing Global Trade Item Number field in Google for WooCommerce


martynmjones:
- [x] #2614
- [x] #2615
- [x] #2616
- [x] #2648
- [x] https://github.com/woocommerce/google-listings-and-ads/issues/2665

## Additional

pcTzPl-2p2-p2

PR #2666, opened 2024-11-08
Refactor & fix undesired exceptions in AS Jobs

puntope:

Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/2617 

- This PR adds some try catch preventing AS to fail when saving the GTIN
- Additionally this PR refactors some of the common methods to make the code cleaner

### Screenshots:

Screenshot 2024-11-08 at 19 56 48

1. Run the GTIN migration with the next GTINs
2. GTINs with dashes but 8 numbers 
3. GTINs with other chars like dots . or letters
4. GTIN that also has the GTIN saved in Product Inventory for that product
5. Duplicated GTIN
6. GTINS in variations
7. See the AS doesn't fail and you can see some logs also. 

### Changelog entry

> Fix - Prevent fatal in GTIN MIgration AS Job

### DIFF
--- a/src/HelperTraits/GTINMigrationUtilities.php
+++ b/src/HelperTraits/GTINMigrationUtilities.php
@@ -6,6 +6,7 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\MigrateGTIN;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Exception;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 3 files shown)


src/HelperTraits/GTINMigrationUtilities.php
```diff
@@ -84,4 +85,70 @@ protected function get_gtin_migration_status(): string {
...
+	}
+
+	/**
+	 * Gets the message when the GTIN is invalid.
+	 *
+	 * @param \WC_Product $product
+	 * @param string      $gtin
+	 * @return string
+	 */
+	protected function error_gtin_invalid( \WC_Product $product, string $gtin ) {
```
mikkamp, line 106: Not a big deal, and we don't seem to be consistent. But in most cases in the code we add `use WC_Product` to be able to remove the `\`. But I'm not attached to either method so both are fine.

src/Utility/WPCLIMigrationGTIN.php
```diff
@@ -211,4 +210,25 @@ protected function process_items( array $items ): int {
 	public static function is_needed(): bool {
 		return defined( 'WP_CLI' ) && WP_CLI;
 	}
+
+	/**
+	 * Add some info in the debug console.
+	 * Add --debug to see this kind of logs in WP CLI
```
mikkamp: Just a small nitpick about the wording (this should be plural):
```suggestion
	 * Add --debug to see these logs in WP CLI
```

mikkamp (approved):
Thanks for the changes, they are looking good. I just left a few minor nitpicks but the migration is working. I can confirm that I see log messages, including an error logged for a duplicate GTIN:

image

PR #2668, opened 2024-11-11
WordPress 6.7 Compatibility: Fix the issue that E2E test can't log in to wp-admin

eason9487:

With WordPress 6.7-RC3, the E2E test failed to start due to match two elements when trying to click the "Log In" button.

```
Trying to log-in as admin...
Admin log-in failed, Retrying... 0/5
locator.click: Error: strict mode violation: locator('text=Log In') resolved to 2 elements:
    1) <h1 class="screen-reader-text">Log In</h1> aka getByRole('heading', { name: 'Log In' })
    2) <input type="submit" id="wp-submit" value="Log In" name="wp-submit" class="button button-primary button-large"/> aka getByRole('button', { name: 'Log In' })
```

This PR ensures that it can locate the "Log In" button.

📌 The failed test case will be handled in a separate RP.

1. View the failed run: https://github.com/woocommerce/google-listings-and-ads/actions/runs/11774628750/job/32793771070#step:10:20
2. View the successful log-in run: https://github.com/woocommerce/google-listings-and-ads/actions/runs/11775116109/job/32794945918#step:10:18

### Changelog entry

> Dev - WordPress 6.7 Compatibility: Fix the issue that E2E test can't log in to wp-admin.

### DIFF
--- a/tests/e2e/global-setup.js
+++ b/tests/e2e/global-setup.js
@@ -55,7 +55,7 @@ module.exports = async ( config ) => {
 			await adminPage
 				.locator( 'input[name="pwd"]' )
 				.fill( admin.password );
-			await adminPage.locator( 'text=Log In' ).click();
+			await adminPage.getByRole( 'button', { name: 'Log In' } ).click();
 			await adminPage.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );
 			await adminPage.goto( `/wp-admin` );
 			await adminPage.waitForLoadState( LOAD_STATE.DOM_CONTENT_LOADED );


mikkamp (approved):
Thanks for the change, I can confirm that there is another element on the page with the same text:
```
<h1 class="screen-reader-text">Log In</h1>
```
With the included fix it selects only the log in button.

The code to login as admin was originally copied over from WooCommerce, but it looks like it was refactored to use a helper utility where they use the same selector as you do here: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/tests/e2e-pw/utils/login.js

Just curious if some of those other tweaks are needed as well, but no reason to fix something that isn't broken. So I'll go ahead and approve the PR.

eason9487:
Hi @mikkamp, thanks for the review!

> Just curious if some of those other tweaks are needed as well, [...]

The current login process looks good. I think we can keep it as it is.

PR #2677, opened 2024-11-18
[GTIN] Add Migration support for YOAST SEO

### LINKED ISSUE #2665: [GTIN] Support YOAST SEO GTIN in Migration JOb and WP CLI Tool
Right now we are only migrating the values inside Google for WooCommerce but other wide used plugins supporting GTIN (like YOAST) are not supported. 

## Additional

pcTzPl-2p2-p2

puntope:

Closes #2665

This PR adds support for the YOAST GTIN field in the GTIN migration feature

### Screenshots:

Screenshot 2024-11-18 at 11 33 19

1. You need to have YOAST WooCommerce installed (notice the WooCommerce part, WordPress one doesn't have GTIN support)
2. Set GTIN field for a simple product in Yoast SEO Tab
3. Set GTIN field for a variation in Yoast SEO Tab (under the variations tab)
4. Start the migration using the Connection test - Start GTIN Migration (be sure there is no other AS Job already running for GTIN migration)
5. See the 2 products have the GTIN migrated
6. Delete the GTINS for those products under Product Inventory (for the simple product) and from the variations tab for the variation.
7. Start the migration using the WP CLI `wp wc g4wc gtin-migration start --debug`
8. See the GTIN successfully migrated
9. Run steo 7 again
10. See logs about skipping the products. 

### Changelog entry

> Add - Support YOAST SEO GTIN field in the migration tools.

### DIFF
--- a/src/HelperTraits/GTINMigrationUtilities.php
+++ b/src/HelperTraits/GTINMigrationUtilities.php
@@ -161,27 +161,10 @@ trait GTINMigrationUtilities {
 	 * @return string|null
 	 */
 	protected function get_gtin( WC_Product $product ): ?string {
-		$gtin = $this->attribute_manager->get_value( $product, 'gtin' );
-
-		if ( ! $gtin ) {
-			$gtin = $this->get_yoast_gtin( $product );
-		}
-
-		return $gtin;
-	}
-
-	/**
-	 * Gets the GTIN value from YOAST SEO
-	 *
-	 * @param WC_Product $product The product
-	 * @return string|null
-	 */
-	protected function get_yoast_gtin( WC_Product $product ): ?string {
-		if ( ! defined( 'WPSEO_WOO_VERSION' ) ) {
-			return null;
-		}
-
-		$yoast_seo_integration = new YoastWooCommerceSeo();
-		return $yoast_seo_integration->get_gtin( YoastWooCommerceSeo::VALUE_KEY, $product );
+		/**
+		 * Filters the value of the GTIN before performing the migration.
+		 * This value will be he one that we copy inside the Product Inventory GTIN.
+		 */
+		return apply_filters( 'woocommerce_gla_gtin_migration_value', $this->attribute_manager->get_value( $product, 'gtin' ), $product );
 	}
 }
--- a/src/HelperTraits/GTINMigrationUtilities.php
+++ b/src/HelperTraits/GTINMigrationUtilities.php
@@ -3,10 +3,12 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\Integration\YoastWooCommerceSeo;
 use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\MigrateGTIN;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Exception;
+use WC_Product;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -35,7 +37,7 @@ trait GTINMigrationUtilities {
 	 * @return bool
 	 */
 	protected function is_gtin_available_in_core(): bool {
-		return version_compare( WC_VERSION, '9.2', '>=' ) && method_exists( \WC_Product::class, 'get_global_unique_id' );
+		return version_compare( WC_VERSION, '9.2', '>=' ) && method_exists( WC_Product::class, 'get_global_unique_id' );
 	}
 
 	/**
@@ -82,7 +84,7 @@ trait GTINMigrationUtilities {
 	 *
 	 * @return OptionsInterface
 	 */
-	protected function options() {
+	protected function options(): OptionsInterface {
 		return $this->options ?? woogle_get_container()->get( OptionsInterface::class );
 	}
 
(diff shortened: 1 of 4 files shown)


src/HelperTraits/GTINMigrationUtilities.php
```diff
@@ -92,63 +94,94 @@ protected function options() {
...
+		return $gtin;
+	}
+
+	/**
+	 * Gets the GTIN value from YOAST SEO
+	 *
+	 * @param WC_Product $product The product
+	 * @return string|null
+	 */
+	protected function get_yoast_gtin( WC_Product $product ): ?string {
```
mikkamp: This structure does work, but it wouldn't be my preferred choice, as it requires the migration to know details about Yoast compatibility as well as having access to public methods.

While we only added support for Yoast in theory there could be other plugins which add support for their structure. Which is why each class in the Integrations folder was implemented in such a way that in theory it could live both in the plugin and outside the plugin (maintained by a 3PD).

I think it would be better to keep it that way and add any necessary hooks so we don't need to make parts of the Yoast integration classes public.

mikkamp: So just add a filter in `get_gtin` so we don't need to add anything in regards to Yoast here. Then in the class  `YoastWooCommerceSeo`  we can hook into the filter for migration and if it has a value to return it can do so.

mikkamp (commented):
Thanks for providing the compatibility. I tested it and it does work as expected, however I think it would be nicer to keep the integration functionality encapsulated in the one class.

src/HelperTraits/GTINMigrationUtilities.php
puntope: Hey @mikkamp Thanks for this suggestion. I 100% agree.

Here is the adjustment https://github.com/woocommerce/google-listings-and-ads/pull/2677/commits/356577e875a20d9dd5f93e4e5a98817e874620ea

puntope:
Hi @mikkamp Thanks for the review. 

This is ready for a new round.

src/HelperTraits/GTINMigrationUtilities.php
```diff
@@ -3,10 +3,12 @@
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\Integration\YoastWooCommerceSeo;
```
mikkamp: This line can also be removed now.

src/Integration/YoastWooCommerceSeo.php
```diff
@@ -76,6 +76,19 @@ function ( $sources, $attribute_id ) {
 			10,
 			2
 		);
+
+		add_filter(
+			'woocommerce_gla_gtin_migration_value',
+			function ( $gtin, $product ) {
+				if ( ! $gtin ) {
+					return $this->get_gtin( self::VALUE_KEY, $product );
```
mikkamp: This part is not working for me. Because when a product was previously setup the `_wc_gla_gtin` had the value of `yoast_seo` and then the `get_gtin` function would replace it with the value stored by Yoast.

However here we only enter the function if no GTIN number was found. So we'll need to decide, do we want to only migrate if the GLA value is set to `yoast_seo` or do we want to migrate whenever we have a value stored by Yoast.

If it's the latter then we should change it to something like:
```suggestion
				if ( ! $gtin || self::VALUE_KEY === $gtin ) {
					return $this->get_gtin( self::VALUE_KEY, $product );
```

Otherwise we could simplify the whole function as long as we pass a string to `get_gtin` it will check the prefix:
```php
		add_filter(
			'woocommerce_gla_gtin_migration_value',
			function ( $gtin, $product ) {
				return $this->get_gtin( (string) $gtin, $product );
			},
			10,
			2
		);
```

mikkamp (commented):
Thanks for restructuring, I think this is much cleaner.

It's just not working for the old products I have setup because the GTIN value is not empty for Yoast GTIN's.

src/Integration/YoastWooCommerceSeo.php
puntope: > if ( ! $gtin || self::VALUE_KEY === $gtin ) {
					return $this->get_gtin( self::VALUE_KEY, $product );

This will always fail because $gtin is actually the `$gtin` number. `self::VALUE_KEY` is a key reference and is only necessary because 
`YoastWooCommerceSeo::get_gtin` [requires that ](https://github.com/woocommerce/google-listings-and-ads/pull/2677/files#diff-c1a51c2c069efcf71305ae9085386da9a893cc2b9ad78c70a22b42d3f0608c2dL112) because is doing a key check as it's the function is being used also in other contexts so we don't need to compare that with $gtin for our case. I believe 

What I'm doing with:

```php
add_filter(
			'woocommerce_gla_gtin_migration_value',
			function ( $gtin, $product ) {
				if ( ! $gtin ) {
					return $this->get_gtin( self::VALUE_KEY, $product );
				}

				return $gtin;
			},
			10,
			2
		);
```

It is to be sure the value is not being overriden. In other words. If the GTIN is set in Google for WooCommerce we won't override it with YOAST GTIN. We only get YOAST value if GTIN is null or empty. Does that make sense?

src/HelperTraits/GTINMigrationUtilities.php
puntope: Adjusted here. https://github.com/woocommerce/google-listings-and-ads/pull/2677/commits/5da274773dd9e625b4ef6c60fa27c6240d3b4c47 Thanks for catching that

puntope:
Thaks @mikkamp for the review nd the catches. I applied all your suggestions. Can we do a last round of review?

mikkamp (approved):
Thanks for all the changes. It looks good now and I'm able to import all products, including the Yoast SEO ones.

PR #2679, opened 2024-11-18
[GTIN] Adding more tests for GTIN migration

puntope:

Closes part of https://github.com/woocommerce/google-listings-and-ads/issues/2617

- This PR adds tests specifically for the GTIN Migration
- Additionally, this PR adds tests for GTIN Hidden feature
- A bug was found when writing the tests (Core GTIN was overridden in the migration). This PR fixes the bug. 

1. See the tests passing
2. Create a simple product and set a GTIN in Product Inventory tab 
3. Set also the GTIN in Google for the WooCommerce tab
4. Start the migration in Connection test page (Start GTIN Migration)
5. See that the GTIN was not overridden

### Changelog entry

> Dev - Adding tests for GTIN migration tool

### DIFF
--- a/src/Jobs/MigrateGTIN.php
+++ b/src/Jobs/MigrateGTIN.php
@@ -93,6 +93,11 @@ class MigrateGTIN extends AbstractBatchedActionSchedulerJob implements OptionsAw
 				continue;
 			}
 
+			if ( $product->get_global_unique_id() ) {
+				$this->debug( $this->error_gtin_already_set( $product ) );
+				continue;
+			}
+
 			$gtin = $this->attribute_manager->get_value( $product, 'gtin' );
 			if ( ! $gtin ) {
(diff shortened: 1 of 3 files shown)


mikkamp (approved):
Thanks for adding these tests. They look good to me and they all pass as expected.

Based on the [report here](https://app.codecov.io/gh/woocommerce/google-listings-and-ads/pull/2679?src=pr&el=tree&filepath=src%2FJobs%2FMigrateGTIN.php&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=woocommerce#diff-c3JjL0pvYnMvTWlncmF0ZUdUSU4ucGhw) I noticed that we do not test migrating GTIN's from variable products, nor when an exception is thrown such as a duplicate GTIN.

Since migration is temporary I'm not too worried that this doesn't have full test coverage, but maybe we can at least log an issue with the missing tests.

PR #2685, opened 2024-11-21
[GTIN] Hide / disable GTIN also when YOAST is active.

puntope:

Closes part of #2617 

This PR fixes a Bug in which GTIN is not hidden or disabled when YOAST is activated.

### Screenshots:

Screenshot 2024-11-21 at 19 33 26

1. Install YOAST for WooCommerce
2. Set the wp_option `gla_install_version`as 2.8.6
3. Go to a Product - Google for WooCommerce
4. See the selector as disabled (if you selected "Enter custom value" it should be also disabled)
5. Set the wp_option `gla_install_version` as 2.8.8
6. See the selector hiden

### Changelog entry

> Fix - Hide/disable GTIN also when YOAST is active

### DIFF
--- a/src/Admin/Product/Attributes/AttributesForm.php
+++ b/src/Admin/Product/Attributes/AttributesForm.php
@@ -8,6 +8,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Admin\Input\FormException;
 use Automattic\WooCommerce\GoogleListingsAndAds\Admin\Input\InputInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Admin\Input\Select;
 use Automattic\WooCommerce\GoogleListingsAndAds\Admin\Input\SelectWithTextInput;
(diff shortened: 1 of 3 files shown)


src/Admin/Product/Attributes/AttributesForm.php
```diff
@@ -132,6 +133,13 @@ public static function init_input( InputInterface $input, AttributeInterface $at
 				$new_input->set_label( $input->get_label() )
 					->set_description( $input->get_description() );
 
+				// Copy values from GTIN
```
mikkamp: This comment didn't help clarify much. Can we elaborate a bit more here:

```suggestion
				// When GTIN uses the SelectWithTextInput field, copy the readonly/hidden attributes from the GTINInput field.
```

mikkamp (approved):
Thanks for catching this extra scenario. It's working as expected. It took me a bit to understand how the field switches input type and why we are copying attributes from a different class. But I think that's more down to the decisions that were made in how to structure the field.

PR #2686, opened 2024-11-22
Add free shipping order value condition for shipping phase 1

ianlin:

Part of pcTzPl-2qP-p2, this PR implements the last design of the new shipping phase 1 for setting the free shipping order value.

The PR follows [this design](https://www.figma.com/file/fqR0EHi63lWahRcVTKCcba?node-id=8367-25570&t=um2mPCfNxQjwBNpI-4) with slightly changes based on the discussion in pcTzPl-2qP-p2#comment-3723 and pcTzPl-2qP-p2#comment-3731.

Notable changes in this PR:

- Create a section for free shipping **Order value condition**, as previously it's part of the shipping rate section.
  - New folder `js/src/components/order-value-condition-section`
- Move `MinumumOrderCard` from `js/src/components/shipping-rate-section` to the new folder `js/src/components/order-value-condition-section`, since the card can be reused with slightly changes.
- Change the method [getCardProps()](https://github.com/woocommerce/google-listings-and-ads/blob/417385411a41fab54c8124875c5cbe8e5cf0d6fa/js/src/components/shipping-rate-section/flat-shipping-rates-input-cards.js#L16) as a shared hook in [js/src/components/adaptive-form/adaptive-form-context.js](https://github.com/woocommerce/google-listings-and-ads/pull/2686/files#diff-b1e1872c71d547b0a7fb7374740e842c18234bb4723ae7039571bce115479db4R75) as there are more than one place require to use this method.
  - I'm not sure if this is the best approach, feel free to suggest anything.

### Screenshots:

https://github.com/user-attachments/assets/5bcff49b-d705-4f65-b39d-2d2ecab6eb23

1. Build the project using this PR
2. Start the onboarding flow, stop by step 2
3. Select **My shipping settings are simple. I can manually estimate flat shipping rates.** in **Shipping rates** section
4. Enter an estimated shipping rate, it should be > 0
5. See there is a new section **Order value condition** under **Shipping rates**
6. Set the free shipping order value condition for multiple target countries
7. Finish the onboarding flow
8. Make sure the free shipping order value conditions are synced to the Google Merchant Center
    - https://merchants.google.com/mc/shipping/services
9. Go to Dashboard, click **Edit** on the free listing
10. Redo from the step 3, and see it's working normally.

### Changelog entry

>


puntope:
One issue I see is that the Save button doesn't work until you toggle "Free shipping over a specific order"


https://github.com/user-attachments/assets/c8a87077-1853-4dca-9770-92e0b8d13576

ianlin:
> One issue I see is that the Save button doesn't work until you toggle "Free shipping over a specific order"

Thanks for catching it! Fixed in 5efbc56.

It's because previously it was implemented using radio buttons so offer_free_shipping has an undefined value. The validation makes sense back then.

But right now it is implemented using a checkbox, there should not be an undefined value, and the validation is not required now.

js/src/components/order-value-condition-section/minimum-order-card/minimum-order-input-control.js
```diff
@@ -46,32 +48,40 @@ const MinimumOrderInputControl = ( props ) => {
 		} );
 	};
 
+	const shouldHideInput = ! values.offer_free_shipping;
```
puntope, line 51: 💅  Not a big deal for me. But maybe doing` if ( shouldHideInput ) { return null; }` we can avoid the styling and 

```
{ ! shouldHideInput && currency && (
						<>{ `Cost (${ currency })` }</>
					) }
```

chunk. I believe the currency is always available?

puntope:
Not sure how much a blocker is this. I see that when we toggle the checkout. The value inside the minimum order value condition field disappears. This also happens in the previous implementation. So I'm approving the PR


https://github.com/user-attachments/assets/45c08696-857a-46ac-83f5-073a6c4db71c

puntope (approved):

PR #2682, opened 2024-11-20
Streamline onboarding, post-onboarding and campaign setup

eason9487:
### Changes proposed in this Pull Request:

This PR gathers all changes from the following feature PRs for final testing and to be merged together into `develop`.

- #2631
- #2632
- #2634
- #2635

#### Additional fix

- #2688

### 📌 Checklist before merging

- [x] Revert 08cec68213e2404026e09a9231ffbe660dbb39ff
- [x] Revert e80472a2a07fd4e4fad351cf285f7a8eae6ac49b
- [x] Update tracking doc

PR #2688, opened 2024-11-22
Consolidate accounts onboarding: Prevent Google Ads account setup from trying to connect to an unclaimed and disconnected account ID

eason9487:

While doing overall testing with #2682, found an issue that it may try to connect to an unclaimed and disconnected account ID. The test instructions below show the reproduction steps.

https://github.com/user-attachments/assets/9c3a2378-4e77-4b12-9394-99a8c14ebbb4

It turns out that the account ID used to call API is not the same as the one selected on UI, and it's an unclaimed and disconnected account.

https://github.com/user-attachments/assets/21beecd2-abb7-48c1-bec5-404b7e7d9337

This PR fixes it by resetting the account ID after disconnecting.

https://github.com/woocommerce/google-listings-and-ads/blob/a6862909946e8bcf4544401cf0238fd40e104e99/js/src/components/google-combo-account-card/connect-ads/connect-existing-account.js#L75-L84

### Screenshots:

https://github.com/user-attachments/assets/a913abcd-8745-4071-8b94-7b62815eebe6

1. Go to step 1 of the onboarding flow
   - Disconnect Google Ads account if there is a connection
   - Have at least one Google Ads account in the existing list
2. Create a new Google Ads account but don't claim it
3. Disconnect Google Ads account
4. Directly connects to the existing Google Ads account currently selected on the UI without any other UI interactions
5. Check if the account can be connected without errors

### DIFF
--- a/js/src/components/google-ads-account-card/disconnect-account.js
+++ b/js/src/components/google-ads-account-card/disconnect-account.js
@@ -2,7 +2,8 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState, useCallback } from '@wordpress/element';
+import { noop } from 'lodash';
+import { useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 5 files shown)


eason9487:
Currently on GitHub Actions, there seems to be some issues with the starting E2E test environment that prevent E2E testing from starting.

mikkamp (approved):
Thanks for the changes. I can confirm this is now running as expected.

In combination with the changes in #2690 I'm also able to run the E2E tests locally without any errors.
image

PR #2689, opened 2024-11-22
Fix E2E admin login failure

mikkamp:
Created this draft PR to test E2E login on admin page. Not able to reproduce this failure locally.

Added screenshot recording to facilitate troubleshooting, this is what the login screen looks like:

admin-page-0

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -4,11 +4,11 @@ on:
   push:
     branches:
       - "release/**"
-      - "feature/2458-2459-2460-2509"
+      - "test/fix-e2e-admin-login-failure"
   pull_request:
     branches:
       - "release/**"
-      - "feature/2458-2459-2460-2509"
+      - "test/fix-e2e-admin-login-failure"
   workflow_dispatch:
     inputs:
       wp-rc-version:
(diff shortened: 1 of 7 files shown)


mikkamp:
After using a filter to output the PHP error message, I uncovered the following:

admin-page-0

We check `return defined( 'WP_CLI' ) && WP_CLI`, but that's not the same as checking if the class is available.

mikkamp:
Just to conclude, we found the culprit was the `is_needed` function was not running to check the conditions.

The following commit fixes this issue: https://github.com/woocommerce/google-listings-and-ads/pull/2689/commits/c0fdbadf56289af6e5f85ba899b58ebb42b3c1b7

Closing this PR, but it has some useful commits to help surface errors in the E2E environment:
- Manually take a screenshot (if placed in the right folder it will be part of the artifacts): https://github.com/woocommerce/google-listings-and-ads/pull/2689/commits/0669161451700a9843e304e05b23c6e33913d2a4
- Use a filter to output the fatal error from WordPress within the screenshot (couldn't find a way to get direct access to the PHP log file): https://github.com/woocommerce/google-listings-and-ads/pull/2689/commits/952283bc926da1449aae391513c2334835c187e2

PR #2695, opened 2024-11-26
Add margins for Notices

puntope:

Fixes a bug where the notices have no separations between them:

### Screenshots:

Before

Screenshot 2024-11-26 at 16 55 48

After

Screenshot 2024-11-26 at 16 56 55

1. Go to Google for WooCommerce
2. Set wp_option gla_install_version as 2.8.7
3. Complete a fresh onboarding 
4. Go to settings
5. See both banners GTIN migration and API PUll Get started separated with margins.

### Changelog entry

> Fix - Add margin for separating notices in the admin

### DIFF
--- a/js/src/css/shared/_woocommerce-admin.scss
+++ b/js/src/css/shared/_woocommerce-admin.scss
@@ -71,16 +71,13 @@
 			max-width: $grid-unit-60;
 		}
 	}
+
+	.components-notice:not(.app-notice) {
+		margin-bottom: $grid-unit-15;
+	}
 }
 
 // compatibility-code "WC >= 7.3" -- adding margin in the first question label.
 .wp-admin .woocommerce-customer-effort-score__intro {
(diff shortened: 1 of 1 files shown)


mikkamp (approved):
Thanks for the changes.

I can confirm that there is now a margin between the two.

Before:
image

After:
image

Since it's scoped to `gla-admin-page` there should only be conflicts within the extension pages. But there are a few other locations that a margin is set for a component-notice, most overwrite the margin we set here. Except this one:
https://github.com/woocommerce/google-listings-and-ads/blob/2.8.7/js/src/components/google-mc-account-card/reclaim-url-card/index.scss#L6-L8

Does that also need to override the bottom margin?
Did some troubleshooting and apparently the bottom margin is reset by `.gla-account-card .components-notice.is-error` so it's still fine:
image

Approving the PR.

PR #2697, opened 2024-11-27
Ensure coupon brand restrictions are uploaded to Google Merchant Center.

### LINKED ISSUE #2693: Coupon restrictions by brand are not supported
### Describe the bug:
While coupon restrictions by brand were previously only available to merchants running a brands plugin, this situation now changes with the introduction of brands to core WooCommerce. In 9069070-zd-a8c the merchant reached out to understand why brand restrictions were not being honored in the coupon field.

We map coupon restrictions [here](https://github.com/woocommerce/google-listings-and-ads/blob/0d356b7f1bd20ee45030876ab59d52b9652e217c/src/Coupon/WCCouponAdapter.php#L220) with no mention of brands.

Not sure if this is a bug or a feature request :)

I lean to bug since this was the user experience.

ianlin:

Closes #2693.

Since WooCommerce 9.4, [WooCommerce Brands is now part of the WooCommerce core](https://developer.woocommerce.com/2024/10/01/introducing-brands/), the merchant can add brand restrictions when editing coupons, but G4W plugin does not upload the corresponding coupon brand restrictions to Google Merchant Center.

This PR adds coupon restrictions for brands when generating the coupon (promotions) payload for updating to Google Merchant Center.

Note that `WC_Coupon` class does not have a method to get the brand restrictions [like other restrictions do](https://github.com/woocommerce/woocommerce/blob/1a3ff575caf0a060a24a914bd8e08b94bedd6153/plugins/woocommerce/includes/class-wc-coupon.php#L305-L318). This PR uses [get_post_meta()](https://developer.wordpress.org/reference/functions/get_post_meta/) to get the brands from a coupon, then uses [get_objects_in_term()](https://developer.wordpress.org/reference/functions/get_objects_in_term/) to get a list of product IDs from that brand.

Refer to the following tables to better understand the data:
- `wp_postmeta`
    - Store the brands for coupons, queried by`get_post_meta()`
    - Meta keys: `product_brands` and `exclude_product_brands`
- `wp_terms`
    - Where the brand ID, brand name, and brand slug are stored
- `wp_term_taxonomy`
    - Store the taxonomy for the brand (`product_brand`)
- `wp_term_relationships`
    - Store the mapping of products and brand. Queried by `get_objects_in_term()`
    - Should specify the brand taxonomy `product_brand` in `get_objects_in_term()`

**Updated 28th Nov**

For querying brands from a coupon from post_meta, I found in [class-wc-brands-brand-settings-manager.php](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-brands-brand-settings-manager.php) I can use it like:

```php
\WC_Brands_Brand_Settings_Manager::set_brand_settings_on_coupon( $coupon );
$brand_settings = \WC_Brands_Brand_Settings_Manager::get_brand_settings_on_coupon( $coupon );
```

However, since we do not expect all merchants update to latest WooCommerce, we also need to ensure backward compatibility for those who still use WC Brands plugin. The code above was available [since WC Brands plugin version 1.6.65](https://github.com/woocommerce/woocommerce-brands/blob/0899036f7a68b9b5f17c4d2fd806e7614e27739a/includes/class-wc-brands-brand-settings-manager.php#L8), so I lean towards to still querying post_meta manually as it's safest for all version and in core.

1. Build the project using this PR, and ensure WooCommerce >= 9.4 is installed on your site
2. Enable Brands feature by updating two options `wc_feature_woocommerce_brands_enabled` and `woocommerce_remote_variant_assignment`
    ```bash
    wp option update wc_feature_woocommerce_brands_enabled 'yes'
    wp option update woocommerce_remote_variant_assignment 2
    ```
3. Create some brands from `Products > Brands`, e.g.
    Screenshot 2024-11-27 at 14 58 11
4. Add some products to each of the brand that just created, e.g.

https://github.com/user-attachments/assets/1a6c0fac-2939-408d-9ade-96ba27ac45d8

5. Create a coupon, add brand restrictions to it
    - Make sure your store country supports coupons, e.g. US, GB.
    - Make sure to choose **Show coupon on Google** in **Channel visibility**
    - Note that the inclusion and exclusion cannot be set at the same time, otherwise Google API will respond 400 bad request. It also applies to existing restrictions like products or product categories.
    - Screenshot 2024-11-27 at 15 14 06
6. Save the coupon, go to `WooCommerce > Status > Scheduled Actions`, search for the job `gla/jobs/update_coupon/process_item`, run it manually if it is still pending to save time
7. Wait for a while, go to Google Merchant Center, click **Promotions** from the left menu.
8. Click on the coupon (promotion) you just uploaded, then click **Next**. Ensure that the filters display the product IDs associated with the brand set as a restriction.

https://github.com/user-attachments/assets/01c893c3-53ef-494f-b73c-e3dcfaacb8c8

9. Play around the existing restrictions, make sure they are still working
10. Disable the Brands in core feature (see step 2), make sure the PR won't break it
11. Using WooCommerce beta tester to downgrade WC to 9.2, and install WC Brands plugin, make sure the PR still works.

#### Product added or removed from a brand

1. Let's say now that the coupon brand restriction is synced to GMC after the above test.
2. Go to the product editor, deselect `Test Brand 1` and update the product
    Screenshot 2024-12-05 at 15 09 44
3. Go to `WooCommerce > Status > Scheduled Actions`, search for the job `gla/jobs/update_coupon/process_item`, run it manually if it is still pending to save time
4. Wait for a while, go to Google Merchant Center, click **Promotions** from the left menu.
5. Click on the coupon (promotion) you just uploaded, then click **Next**. Ensure that the filters no longer include the product that was just excluded from the brand.
6. Go to the product editor, add the brand `Test Brand 1` back. You should see the product gets synced to GMC in the coupon filters again.
7. Create a new product and add the brand `Test Brand 1`, you should see the product gets synced to GMC in the coupon filters again.

### Changelog entry

> Fix - Ensure coupon brand restrictions are uploaded to Google Merchant Center.


src/Coupon/WCCouponAdapter.php
```diff
@@ -267,6 +267,20 @@ protected function map_wc_usage_restriction( WC_Coupon $wc_coupon ): WCCouponAda
...
+			$google_product_ids      = array_map( $get_offer_id, $wc_product_ids_in_brand );
+			$has_product_restriction = true;
+			$this->setItemId( $google_product_ids );
+		}
+
+		$wc_excluded_product_ids_in_brand = $this->get_product_ids_in_brand( $wc_coupon, true );
+		if ( ! empty( $wc_excluded_product_ids_in_brand ) ) {
+			$google_product_ids      = array_map( $get_offer_id, $wc_excluded_product_ids_in_brand );
+			$has_product_restriction = true;
+			$this->setItemIdExclusion( $google_product_ids );
```
eason9487: When a coupon also has other excluded products, those product IDs will be overridden by this restriction.

For example, the brand in the following setting is only added to a t-shirt, the restriction synced to GMC is only the product ID of that t-shirt, while the product ID of the belt does not appear.

image

However, when tested in WC store, both restrictions are applied.

---

I also tested the opposite restrictions and looks like the brand inclusion restriction has a higher priority than the product inclusion restriction, so the overriding for inclusion restriction should be fine.

ianlin: Nice catch!

I believe it could potentially be a bug in Brands that product inclusion being overridden by the product brands inclusion:

Brands [uses a filter for checking product brands inclusion](https://github.com/woocommerce/woocommerce/blob/5a182ac082b83ab0642199f04b030b04c3f1ade4/plugins/woocommerce/includes/class-wc-brands-coupons.php#L42-L63) after [the product inclusion validation](https://github.com/woocommerce/woocommerce/blob/5a182ac082b83ab0642199f04b030b04c3f1ade4/plugins/woocommerce/includes/class-wc-discounts.php#L1056), and it did not take it into account.

The filter has a third parameter which is a discount object, I think it might be ok to recheck the product inclusion in Brands.

However, to align with the current WC core behaviour, I will update this PR to respect only the product exclusion, and will also create an issue in WC core for Brands about the inclusion bug.

ianlin:
Thanks for the good catch @eason9487. I've updated the PR and it's ready for another round.

src/Coupon/WCCouponAdapter.php
ianlin: Issue created in woocommerce/woocommerce/issues/53299.

tests/Unit/Coupon/WCCouponAdapterTest.php
```diff
@@ -258,6 +259,61 @@ public function test_product_type_restrictions() {
 		$this->assertEquals( [ 'Alpha Category > Beta Category' ], $adapted_coupon->getProductTypeExclusion() );
 	}
 
+	public function test_brand_restrictions() {
+		if ( version_compare( WC_VERSION, '9.4', '<' ) ) {
```
eason9487, line 264: Suggest adding a compatibility code comment to make it easier to search for compatibility codes in the future.

```php
// compatibility-code "WC < 9.4" -- additional inline description
```

eason9487 (commented):
Thanks for the work. Tested that the brand restrictions of coupons can be converted to product IDs and then synced to Google Merchant Center.

However, just noticed there may be another concern. If a product is later created/updated to attach or detach a brand having associations with coupon restrictions, the brand restrictions (converted to associated product IDs) of those coupons won't be synced.

ianlin:
> If a product is later created/updated to attach or detach a brand having associations with coupon restrictions, the brand restrictions (converted to associated product IDs) of those coupons won't be synced.

Thanks for catching this case @eason9487. I've updated the code to take care of this scenario and have also updated the testing instruction in the PR.

However, I did not solve the case when the product is deleted/trashed, so when it happens the GMC promotion filters will still have the deleted product IDs included. This is align with the behaviour of the **_coupon product inclusion restriction_** (not the brand restriction) - i.e. when a coupon has an included product restriction, then the product gets deleted, GMC promotion filters still have that product.

tests/Unit/Coupon/SyncerHooksTest.php
```diff
@@ -265,6 +270,51 @@ public function test_delete_coupon_triggers_notification_delete() {
...
+		}
+
+		require_once WC_ABSPATH . '/includes/class-wc-brands.php';
+		\WC_Brands::init_taxonomy();
+
+		// Create products and brands.
+		/**
+		 * @var WC_Product $product
+		 */
+		$product_1 = WC_Helper_Product::create_simple_product();
```
eason9487: The variable name in the code comment is not the same as the actual one.

eason9487 (approved):
Thanks for the further update! Tested that coupon brand restrictions can be converted to associated product IDs and be synchronized to GMC. LGTM.

The mismatched code comment should not need another round.

PR #2700, opened 2024-11-29
Prevent translations from being called early

mikkamp:

This PR prevents translations from being called early. There are quite a few places where we translate code before init, so this PR contains the following fixes to prevent the notices from showing:

#### Jetpack Connection plugin name
The Jetpack connection has to be registered on `plugins_loaded` priority 1 as shown in the [example here](https://github.com/Automattic/jetpack-connection/blob/trunk/docs/register-site.md#initialize-the-package). Since we aren't able to change this to a later hook, I initially called the registration with a hardcoded plugin name. After init the plugin name is updated to a translatable name.

##### Test instructions:
1. Switch to another language in Settings (choose a language with translations available like Spanish)
2. Go to Dashboard > Updates and make sure Translations are updated and downloaded
3. Enable WP_DEBUG and WP_DEBUG_LOG so we can monitor debug notices in the log file
4. Use a test snippet to translate the plugin name:
```php
add_filter(
	'gettext',
	function ( $translation = '', $text = '', $domain = '' ) {
		if ( 'google-listings-and-ads' === $domain && 'Google for WooCommerce' === $text ) {
			return 'Google for WooCommerce (translated)';
		}

		return $translation;
	},
	10,
	3
);
```
5. Use the connection test page and click "WordPress.com Connection Status"
6. Confirm that we see the translated name in the list of connected plugins
7. Check the log file to confirm there are no notices about translations being called too early

#### Pre-requisite tests
We want to continue checking pre-requisites like required plugins / versions at an early stage, but we shouldn't call any translation functions till after init. Since we use an exception when encounter a problem, I created a new type `RuntimeExceptionWithMessageFunction`. This custom exception accepts a callback function which can be used to format the final message. This means we can throw an exception early with the error message wrapped in a function, the function will only be called when we want to actually display the errors which will be after init. Several classes had to get updated to support this new type of exception.

> [!Note]
> I used PHP arrow functions to simplify how the strings are wrapped. Arrow functions are available since PHP 7.4 and they allow automatic use of variables from the parent scope, which makes the code a little shorter than using anonymous functions.

##### Test instructions:
1. Switch to another language in Settings (choose a language with translations available like Spanish)
2. Go to Dashboard > Updates and make sure Translations are updated and downloaded
3. Enable WP_DEBUG and WP_DEBUG_LOG so we can monitor debug notices in the log file
4. Use a test snippet to disable WCAdmin:
```php
add_filter( 'woocommerce_admin_disabled', '__return_true' );
```
5. Confirm the admin notice is displayed and there are no notices in the log file
image
6. Use a test snippet to enable Google Product Feed
```php
define( 'WOOCOMMERCE_GPF_VERSION', '1' );
```
7. Go to Marketing > Google for WooCommerce > Product Feed and confirm the issues table shows the error (need to clear cache if the product feed page has been listed before)
image
8. Temporarily increase `WC_GLA_MIN_PHP_VER` to a higher value
9. Confirm the admin notice is displayed and there are no notices in the log file
image
10. Temporarily increase `WC_GLA_MIN_WC_VER` to a higher value
11. Confirm the admin notice is displayed and there are no notices in the log file
image
10. Temporarily change the check for `PHP_INT_SIZE !== 8` to return false
11. Confirm the admin notice is displayed and there are no notices in the log file
image

#### Load country data JIT
We rely on the WooCommerce country data much of which are wrapped in translation functions. Previously in the `Proxies\WC` we were initializing all the country data in the constructor, which means it's loaded as soon as the plugin loads. So this change is both a slight optimization as the country list doesn't need to be loaded when it's not used, and it also prevents translation strings from being called too early.

The easiest way to test this is to run the unit tests, but we can also go through the onboarding and confirm all countries are available as expected without any notices in the log file.

### Additional details:
peeuvX-20i-p2

### Changelog entry
* Fix - Prevent translations from being called early.

### DIFF
--- a/src/Exception/ExtensionRequirementException.php
+++ b/src/Exception/ExtensionRequirementException.php
@@ -3,19 +3,18 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Exception;
 
-use RuntimeException;
-
 defined( 'ABSPATH' ) || exit;
 
 /**
  * Class ExtensionRequirementException
  *
  * Error messages generated in this class should be translated, as they are intended to be displayed
- * to end users.
+ * to end users. We pass the translated message as a function so they are only translated when shown.
+ * This prevents translation functions to be called before init which is not allowed in WP 6.7+.
  *
  * @package Automattic\WooCommerce\GoogleListingsAndAds\Exception
  */
-class ExtensionRequirementException extends RuntimeException implements GoogleListingsAndAdsException {
+class ExtensionRequirementException extends RuntimeExceptionWithMessageFunction implements GoogleListingsAndAdsException {
 
 	/**
 	 * Create a new instance of the exception when a required plugin/extension isn't activated.
@@ -27,6 +26,12 @@ class ExtensionRequirementException extends RuntimeException implements GoogleLi
 	public static function missing_required_plugin( string $plugin_name ): ExtensionRequirementException {
 		return new static(
 			sprintf(
+				'Requires %1$s to be enabled.', // Fallback exception message.
+				$plugin_name
+			),
+			0,
+			null,
+			fn () => sprintf(
 				/* translators: 1 the missing plugin name */
 				__( 'Google for WooCommerce requires %1$s to be enabled.', 'google-listings-and-ads' ),
 				$plugin_name
@@ -44,6 +49,12 @@ class ExtensionRequirementException extends RuntimeException implements GoogleLi
 	public static function incompatible_plugin( string $plugin_name ): ExtensionRequirementException {
 		return new static(
 			sprintf(
+				'Incompatible with %1$s.', // Fallback exception message.
+				$plugin_name
+			),
+			0,
+			null,
+			fn () => sprintf(
 				/* translators: 1 the incompatible plugin name */
 				__( 'Google for WooCommerce is incompatible with %1$s.', 'google-listings-and-ads' ),
 				$plugin_name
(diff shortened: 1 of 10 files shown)


src/Exception/ExtensionRequirementException.php
```diff
@@ -27,6 +26,12 @@ class ExtensionRequirementException extends RuntimeException implements GoogleLi
 	public static function missing_required_plugin( string $plugin_name ): ExtensionRequirementException {
 		return new static(
 			sprintf(
+				'Requires %1$s to be enabled.', // Fallback exception message.
```
tomalec: I may miss something, but why don't we include the plugin name in the fallback message? 

I get that it cannot be translated, but as we hardcode the message in English, we could also hardcode the plugin name in English.

tomalec (approved):
I tested it locally, reviewed the code. 
Left one non-blocking question/comment. 
LGTM.

src/Internal/DependencyManagement/ThirdPartyServiceProvider.php
```diff
@@ -73,5 +73,17 @@ function () {
...
 		);
+
+		// Update Jetpack connection with a translatable name, after init is called.
+		add_action(
+			'init',
+			function () {
+				$manager = $this->getLeagueContainer()->get( Manager::class );
+				$manager->get_plugin()->add(
+					__( 'Google for WooCommerce', 'google-listings-and-ads' ),
+					[]
```
tomalec: I think we can skip the last argument completely
```suggestion
					__( 'Google for WooCommerce', 'google-listings-and-ads' )
```

src/Exception/ExtensionRequirementException.php
mikkamp: I was a bit on the fence with the fallbacks. Initially I just used a string like `missing_required_plugin` but thought we might as well fallback to some English text. We wouldn't expect to ever see the fallback so generally it shouldn't make a difference. I just thought it would be less work if we renamed the extension again. But I'll add them back in to make the messages similar.

PR #2703, opened 2024-11-30
Increase the specificity of `__styled--overflow-visible` selector

### LINKED ISSUE #2702: Country tree select in shipping time/rate modal is partial hidden
### Describe the bug:

_I discovered this issue while testing https://github.com/woocommerce/google-listings-and-ads/pull/2700_

The country tree select for shipping rate or time overflows from the modal and is hidden.
Image

### Steps to reproduce:

1. Go to Marketing > G4W > "Edit" Free listing `/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Ffree-listings%2Fedit`
3. Click "Edit" on shipping times
4. Click at country input ("If customer is in")

### Expected behavior:

Full country tree list should be visible

### Actual behavior:

Part of the list is hidden.

### Additional details:

Seems like we need to increase the specificity of the selector used in https://github.com/woocommerce/google-listings-and-ads/blob/4523126ec7f6823c383bbcdcb00a7d7216bea6d5/js/src/components/app-modal/index.scss#L34:L34
as components' styles are loaded after and overwrite this rule.
Maybe something like  `&.app-modal`, but it's worth checking other scenarios.

tomalec:

This PR increases the specificity of the selector introduced in https://github.com/woocommerce/google-listings-and-ads/pull/1442

Fixes https://github.com/woocommerce/google-listings-and-ads/issues/2702

#### Before
image
#### After
image

1. Switch to another language in Settings (choose a language with translations available like Spanish)
1. Go to Dashboard > Updates and make sure Translations are updated and downloaded
1. Enable WP_DEBUG and WP_DEBUG_LOG so we can monitor debug notices in the log file

I'm not sure if that's enough to always reproduce the component styles being loaded after G4W. But that is when it happens to me. Also, I see nothing preventing this race in regular non-error cases.

1. Go to editing the free listings (`/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard&subpath=%2Ffree-listings%2Fedit`)
1. Check country selectors of adding/editing shipping modals (rate & time) with the following steps.
   1. Open a modal and open the inside dropdown options of the tree selector.
   1. Expend some options to see if the dropdown is no longer covered by the viewport of modal.

### Changelog entry

> Fix - The country selector in setup/edit free listing flow from being hidden.

### DIFF
--- a/js/src/components/app-modal/index.scss
+++ b/js/src/components/app-modal/index.scss
@@ -31,7 +31,7 @@
 	}
 
 	&__styled--overflow-visible {
-		&,
+		&.app-modal,
 		.components-modal__content {
 			overflow: visible;
 		}


eason9487 (commented):
I can't reproduce the issue. Could you check where the overriding happened?

tomalec:
It happens when I have `define( 'WP_DEBUG', false );` and force some PHP error to appear on the site 
image

So, overriding happens when the order of styles being loaded changes, and AFAIK, nothing asserts the order the current selector expects.
WP component styles are preloaded, so they usually arrive earlier, but there is nothing to prevent the race.

eason9487 (approved):
Thank you! I can reproduce this situation now.

However, I probably would not consider it as an issue because the following processes indirectly guarantee the order of CSS files being loaded:

  - The CSS of this plugin depends on the `'wc-admin-app'` handle
     - https://github.com/woocommerce/google-listings-and-ads/blob/75069c9b64ac040d0040f660e40713ca9f0ae931/src/Admin/Admin.php#L152-L155
  - The `'wc-admin-app'` handle depends on the `'wp-components'` handle
     - https://github.com/woocommerce/woocommerce/blob/9.4.3/plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php#L375-L378
     - https://github.com/woocommerce/woocommerce/blob/9.4.3/plugins/woocommerce/src/Internal/Admin/FeaturePlugin.php#L105-L109

Even when this situation happens, the order of the `<link>` tags is still correct:



The actual cause of this situation may be that the browser's fault tolerance (Quirks Mode) allows tags such as `<html>` or `<head>` to be omitted. So when any content precedes these tags, the browser wraps them around `<html>` and `<body>` and ignores the subsequent `<head>` and `<body>`, which further causes the `<link>` tags to be placed inside the fault-tolerant `<body>`.

image

image

Quirks Mode plus misplaced `<link>` tags may result in CSS not being processed according to the standard CSS specification.

Since PHP warnings are not a normal status of a website, it may be worthwhile to consider whether this CSS style should still be fault-tolerant in such a situation.

Nevertheless, increasing the specificity of this CSS selector won't harm either, so I think this PR can still be merged.

tomalec:
Oh wow! Thank you for the detailed investigation and explanations! @eason9487

PR #2704, opened 2024-11-30
Make tab-nav a bit more responsive

### LINKED ISSUE #2473: GLA Menu is not responsive
### Describe the bug:

While working on this PR https://github.com/woocommerce/google-listings-and-ads/pull/2471, @puntope noticed that the menu isn't responsive, as shown in the image:

image

We should look into making the menu responsive for different devices.

### Steps to reproduce:

1.  Visit the GLA extension with a mobile resolution. 
2.  See the menu. 

### Expected behavior:
- Menu should be responsive

### Actual behavior:
- Menu is not responsive

tomalec:

This PR makes the tab navigation a bit more responsive, my making tabs wrap when the container is too narrow. It also adds the  bottom "bar" to the items, sot when the elements wrap, the "bar" is visible under all tabs.

Closes #2473

#### Before
image

#### After
image

1.  Go to G4W dashboard `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fdashboard`
2. Change the viewport size to something narrower
3. Check that all the tabs are visible
4. Go to Product feed `/wp-admin/admin.php?page=wc-admin&issueType=product&path=%2Fgoogle%2Fproduct-feed`
5. Check that all "Issues to resolve" tabs are visible.

### Additional details:

1. We may consider implementing more dedicated UI for the narrower screens, but I believe it's a good enough quick fix in the meantime.

### Changelog entry

> Fix - Make the tab navigation tabs wrap when the screen narrows.

### DIFF
--- a/js/src/components/app-tab-nav/index.scss
+++ b/js/src/components/app-tab-nav/index.scss
@@ -8,7 +8,7 @@
 	&-item {
 		background: transparent;
 		border: none;
-		box-shadow: none;
+		box-shadow: inset 0 -1px 0 #ccc;
 		border-radius: 0;
 		cursor: pointer;
 		height: $grid-unit-60;
@@ -16,7 +16,6 @@
 		margin-left: 0;
 		font-weight: 500;
 		box-sizing: border-box;
-		box-shadow: inset 0 -1px 0 #ccc;
 
 		// This pseudo-element "duplicates" the tab label and sets the text to bold.
 		// This ensures that the tab doesn't change width when selected.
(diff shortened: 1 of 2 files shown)


js/src/components/app-tab-nav/index.scss
```diff
@@ -1,13 +1,14 @@
...
+	flex-wrap: wrap;
 	align-items: stretch;
 	box-shadow: inset 0 -1px 0 $gray-400;
 	margin-bottom: var(--main-gap);
 
 	&-item {
 		background: transparent;
 		border: none;
-		box-shadow: none;
+		box-shadow: inset 0 -1px 0 #ccc;
```
eason9487: Suggest using the same color variable as line 5 instead of using `#ccc`.

eason9487 (approved):
Thanks for the fix. LGTM.

The only comment should not need another round of code reviews.

PR #2706, opened 2024-12-03
Avoid errors caused by automatically saving invalid shipping times during onboarding

eason9487:

This PR avoids errors caused by automatically saving invalid shipping times during onboarding.

https://github.com/user-attachments/assets/42e2d770-7082-42fb-90ff-ca47964b41b3

The issue also causes [a flaky E2E test case](https://github.com/woocommerce/google-listings-and-ads/actions/runs/12119347486/job/33785728575#step:10:234) although it can pass on the second retry:

image

### Screenshots:

https://github.com/user-attachments/assets/bd717860-e6c6-4cf4-b7e2-0c02b9a3d72d

2. Enter a shipping time setup where the min days is greater than its max days
3. Check if the API error, "There was an error saving shipping times.", doesn't occur
4. View the E2E tests run on GitHub Actions to see if the flaky E2E test case no longer occur
   - https://github.com/woocommerce/google-listings-and-ads/actions/runs/12133002898

### DIFF
--- a/js/src/components/free-listings/setup-free-listings/index.js
+++ b/js/src/components/free-listings/setup-free-listings/index.js
@@ -136,12 +136,17 @@ const SetupFreeListings = ( {
 				setValue( 'shipping_country_rates', nextValue );
 			}
 		} else if ( change.name === 'shipping_country_times' ) {
+			// Skip the call of `onShippingTimesChange` if any shipping times are invalid.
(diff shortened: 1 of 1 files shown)


ianlin (approved):
Thanks for catching and fixing it. Tested locally and confirmed the invalid shipping times weren't saved automatically.

PR #1150, opened 2021-12-10
Add test cases for `.~/data/reducer.js`

eason9487:
### Changes proposed in this Pull Request:

It's pre-processing PR for #270. So that we can better ensure the changes are logically correct when fixing #270 in subsequent PR.

- Casting shipping rates/times to number type in action functions instead of reducer.
- Implement a function to create the testing state with deep freeze and reference check.
- Add test cases for reducer.
- Extra change: Avoid jest test watch *js/build* directory, which leads to repeatedly running the same test after modifying js file.

### Detailed test instructions:

1. `npm run test:js:watch -- reducer`
2. Check if all test cases are passed.
3. Check if the test cases meet the testing ability to discover problems.
4. Check if any test descriptions are unclear or confusing.

Issue #270, opened 2021-03-02
Replace `cloneDeep` within reducer by another way


eason9487:
### Describe the issue:

This issue was sourced from [here](https://github.com/woocommerce/google-listings-and-ads/pull/264/files/4db418d544da8b70e7f9087677477297d3af65a9#diff-8e875e4932c094a7da6eac9322edf0cb70fbf56fc555137bd686aa9e265a76cf)

It's a potential performance issue to use `cloneDeep` to update nested objects within a redux store. It would make all unchanged objects and arrays within the store's subtree be different references. Then causing all selectors and components to be forced to evaluate re-render.
```js
case TYPES.RECEIVE_SHIPPING_TIMES: {
	const { shippingTimes } = action;
	const newState = cloneDeep( state );
	newState.mc.shipping.times = shippingTimes;
	return newState;
}
```

### Solution options

1. Copy all levels of nested data
```js
case TYPES.RECEIVE_SHIPPING_TIMES: {
	const { shippingTimes } = action;
	const newState = {
		...state,
		mc: {
			...state.mc,
			shipping: {
				...state.mc.shipping,
				times: shippingTimes,
			},
		},
	};
	return newState;
}
```
2. Implement a util function to simplify the option 1 method.
```js
// A util function similar to lodash.set without mutating the input object
// but return a new object with the same subtree references,
// and only change the value at the specified path
function set( object, paths, value ) {
	const [ key, ...restPaths ] = Array.isArray( paths )
		? paths
		: paths.split( '.' );

	if ( restPaths.length === 0 ) {
		return { ...object, [ key ]: value };
	}

	return {
		...object,
		[ key ]: set( object[ key ], restPaths, value ),
	};
}

// Usage example
case TYPES.RECEIVE_SHIPPING_TIMES: {
	return set( state, 'mc.shipping.times', action.shippingTimes );
}
```
3. Use [combineReducers](https://github.com/WordPress/gutenberg/blob/9111ad51e3f123b0b60cc75666412991e834a46b/packages/data/README.md#combineReducers) to cut the store to flatter slices (but raise the complexity)
4. Introduce [Immer](https://immerjs.github.io/immer/docs/introduction) or [Redux Toolkit](https://redux-toolkit.js.org/)
5. Flat the nested object and give a normalized naming rule
5. Any other practice?

### Additional details:

https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns

eason9487:
[...]

PR #1151, opened 2021-12-10
Replace `cloneDeep` for `.~/data/reducer.js` and apply new state mutation function to whole reducer

eason9487:
### Changes proposed in this Pull Request:

Closes #270

- Add `chain` and `set` functions to update state without mutating other references of state tree.
- Replace all `cloneDeep` and `getNextStateForShipping`.
- Replace the remaining reducer with the `chain` and `set` functions.
- Revert the temporary commit https://github.com/woocommerce/google-listings-and-ads/commit/ece1da7bb7fe7ae9a1e69db673874b0ad7a028c3 to check references of state tree for all test cases.

:warning: It depends on #1150.

### Detailed test instructions:

1. `npm run test:js:watch -- reducer`
2. Check if all test cases are passed.
3. Browse GLA pages to see if there are any unusual results or errors

### Changelog entry

> Fix - Replace `cloneDeep` within `.~/data/reducer.js` with functions that would not mutate other references of the state tree.

PR #2713, opened 2024-12-05
Remove assertions from `reducer.test.js` that prevent the next state from being created by a way mutating unrelated references

eason9487:

For almost three years now, wp-data reducer changes have basically followed existing code that uses `setIn` or `chainState` functions to create the next state.

Considering that the need for these assertions should have gone away and that maintaining them and using them with tests are highly complex, we should be able to live without these assertions.

Ref:
- https://github.com/woocommerce/google-listings-and-ads/issues/270
- https://github.com/woocommerce/google-listings-and-ads/pull/1150

1. Run `npm run test:js` to check if it can pass.
2. Evaluate whether the reasons for removing it are sufficient and sound.

### DIFF
--- a/js/src/data/test/__helpers__/index.js
+++ b/js/src/data/test/__helpers__/index.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { cloneDeep, set, get, isPlainObject } from 'lodash';
+import { cloneDeep, set } from 'lodash';
 
 /**
  * Use native `Object.freeze()` to make an object immutable, recursively freeze each property which is of type object.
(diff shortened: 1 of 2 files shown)


ianlin (approved):
Thanks for the change and keep the repo follows the latest state of wp-data. The test ran successfully in my local.

> For almost three years now, wp-data reducer changes have basically followed existing code that uses setIn or chainState functions to create the next state.

Correct me if I'm wrong, I guess there might be more PRs about removing `setIn` and `chainState` in our repo?

eason9487:
Hi @ianlin, thanks for the review!

> I guess there might be more PRs about removing `setIn` and `chainState` in our repo?

The `setIn` and `chainState` functions were based on the consensus of the `cloneDeep` alternative at the time, and a modified implementation was added in #1151. Ref: The second part in https://github.com/woocommerce/google-listings-and-ads/issues/270#issuecomment-864708881

Unless option 5 mentioned in #270 ([this comment](https://github.com/woocommerce/google-listings-and-ads/issues/270#issuecomment-863169462) has an example) is adopted in the future, they will continue to be used.

Issue #2647, opened 2024-10-18
PHP 8.4 compatibility


mikkamp:
### Describe the issue:
With the upcoming release of PHP 8.4 there is another set of changes that are marked as deprecated. In order to continue preventing deprecation notices we will need to resolve these for any sites that switch to PHP 8.4

The following [deprecations in PHP 8.4](https://php.watch/versions/8.4#deprecations) are the ones we should address:

##### [Implicitly nullable parameter declarations deprecated](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated)

There are several occurrences of this in the code where a non nullable type is declared and the default is set to null. There might be a better regex, but I found several occurrences by searching for `function.*= null`.

Some examples include:
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/MerchantCenter/MerchantStatuses.php#L194
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/MerchantCenter/MerchantStatuses.php#L331
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/MerchantCenter/AccountService.php#L533
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Exception/ExceptionWithResponseData.php#L31
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Shipping/ShippingLocation.php#L48
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/View/PHPView.php#L114
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Infrastructure/View.php#L38
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Google/InvalidCouponEntry.php#L51-L52
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Assets/ScriptAsset.php#L55
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Assets/ScriptWithBuiltDependenciesAsset.php#L35
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Assets/BaseAsset.php#L97
- https://github.com/woocommerce/google-listings-and-ads/blob/2.8.6/src/Assets/StyleAsset.php#L41

[...]

PR #2714, opened 2024-12-05
Fix initial php 8.4 compatibility

mikkamp:

This PR is the initial start for resolving the issues that are in #2647 

There are still a few dependent packages we are waiting for provide full compatibility:
- WooCommerce: https://github.com/woocommerce/woocommerce/issues/52081
- google/gax: Test package is available but not released yet
- google/auth: Test package is available but not released yet
- google/apiclient: https://github.com/googleapis/google-api-php-client/issues/2640
- googleads/google-ads-php: https://github.com/googleads/google-ads-php/issues/1056
- automattic/jetpack-connect: is resolved but we'll need to bump to 6.0.0 minimum (they still support PHP < 7.4)

#### Resolved in this PR
1. Fix all PHP code within the extension that is not compatible with PHP 8.4
2. Update the guzzle packages and fix namespace replacements
```
$ composer update guzzlehttp/guzzle --with-dependencies
   
  - Upgrading guzzlehttp/guzzle (7.8.1 => 7.9.2)
  - Upgrading guzzlehttp/promises (2.0.2 => 2.0.4)
  - Upgrading guzzlehttp/psr7 (2.6.2 => 2.7.0)
  - Upgrading psr/http-factory (1.0.2 => 1.1.0)
```
3. Update the phpunit packages for testing (not bundled in release)
```
$ composer update phpunit/phpunit --with-dependencies
   
  - Upgrading myclabs/deep-copy (1.11.1 => 1.12.1)
  - Upgrading nikic/php-parser (v4.17.1 => v5.3.1)
  - Upgrading phar-io/manifest (2.0.3 => 2.0.4)
  - Upgrading phpunit/php-code-coverage (9.2.29 => 9.2.32)
  - Upgrading phpunit/phpunit (9.6.13 => 9.6.21)
  - Upgrading sebastian/cli-parser (1.0.1 => 1.0.2)
  - Upgrading sebastian/complexity (2.0.2 => 2.0.3)
  - Upgrading sebastian/diff (4.0.5 => 4.0.6)
  - Upgrading sebastian/exporter (4.0.5 => 4.0.6)
  - Upgrading sebastian/global-state (5.0.6 => 5.0.7)
  - Upgrading sebastian/lines-of-code (1.0.3 => 1.0.4)
  - Upgrading sebastian/resource-operations (3.0.3 => 3.0.4)
  - Upgrading theseer/tokenizer (1.2.1 => 1.2.3)
```
4. Update Jetpack packages to highest available v2 package
```
$ composer update automattic/jetpack-connection --with-dependencies

  - Upgrading automattic/jetpack-a8c-mc-stats (v2.0.0 => v2.0.4)
  - Upgrading automattic/jetpack-admin-ui (v0.3.2 => v0.4.6)
  - Locking automattic/jetpack-assets (v2.3.14)
  - Upgrading automattic/jetpack-connection (v2.3.2 => v2.12.6)
  - Upgrading automattic/jetpack-constants (v2.0.0 => v2.0.5)
  - Upgrading automattic/jetpack-redirect (v2.0.0 => v2.0.3)
  - Upgrading automattic/jetpack-roles (v2.0.0 => v2.0.4)
  - Upgrading automattic/jetpack-status (v2.1.0 => v3.3.5)
```
5. Update validator package and revert pattern to remain strict for image URL validation
```
$ composer update symfony/validator --with-dependencies

  - Upgrading symfony/polyfill-ctype (v1.29.0 => v1.31.0)
  - Upgrading symfony/polyfill-php73 (v1.28.0 => v1.31.0)
  - Upgrading symfony/polyfill-php80 (v1.29.0 => v1.31.0)
  - Upgrading symfony/polyfill-php81 (v1.29.0 => v1.31.0)
  - Upgrading symfony/translation-contracts (v2.5.2 => v2.5.3)
  - Upgrading symfony/validator (v5.4.30 => v5.4.47)
```
6. Update league/container package and provide support for V4 (we remain using psr/container 1.1)
```
$ composer update league/container --with-dependencies

  - Upgrading league/container (3.4.1 => 4.2.4)
```

1. Checkout this branch
2. Update packages `rm -rf vendor && composer install`
3. Temporarily use test packages:
```
composer require "google/auth:dev-1.37.x-php84 as 1.37.2"
composer require "google/gax:dev-1.29.1-php74 as 1.29.2"
```
4. Create a temporary script to patch remaining packages in `bin/fix-packages.php`
```
<?php

function replace_all( $path, $search, $replace ) {
	foreach ( glob( $path ) as $file ) {
		$content = preg_replace(
			'/' . preg_quote( $search, '/' ) . '/',
			$replace,
			file_get_contents( $file )
		);
		file_put_contents( $file, $content );
	}
}

replace_all(
	'vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/*.php',
	'string $template = null',
	'?string $template = null'
);

replace_all(
	'vendor/automattic/jetpack-connection/src/class-manager.php',
	'Jetpack_XMLRPC_Server $xmlrpc_server = null',
	'?Jetpack_XMLRPC_Server $xmlrpc_server = null'
);

replace_all(
	'vendor/google/apiclient/src/Client.php',
	'ClientInterface $authHttp = null',
	'?ClientInterface $authHttp = null'
);

replace_all(
	'vendor/google/apiclient/src/Client.php',
	'ClientInterface $http = null',
	'?ClientInterface $http = null'
);

replace_all(
	'vendor/google/apiclient/src/Service/Exception.php',
	'Exception $previous = null',
	'?Exception $previous = null'
);

```
5. Patch packages `php bin/fix-packages.php`
6. Setup the unit tests `bin/install-wp-tests.sh <db> <db_user> <db_pass>`
7. Replace the copy of WooCommerce from `/tmp/woocommerce-9.4.3/plugins/woocommerce` with the contents of the zip file provided below
[woocommerce-9.4.3-patched-for-php-8-4.zip](https://github.com/user-attachments/files/18026122/woocommerce-9.4.3-patched-for-php-8-4.zip)
8. Run unit tests with php 8.4 `/usr/bin/php8.4 vendor/bin/phpunit`
9. Ensure all tests complete without errors
```
$ /usr/bin/php8.4 vendor/bin/phpunit

Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
Installing WooCommerce...
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 9.6.21 by Sebastian Bergmann and contributors.

.............................................................   61 / 1327 (  4%)
.............................................................  122 / 1327 (  9%)
.............................................................  183 / 1327 ( 13%)
.............................................................  244 / 1327 ( 18%)
.............................................................  305 / 1327 ( 22%)
.............................................................  366 / 1327 ( 27%)
.............................................................  427 / 1327 ( 32%)
.............................................................  488 / 1327 ( 36%)
.............................................................  549 / 1327 ( 41%)
.............................................................  610 / 1327 ( 45%)
.............................................................  671 / 1327 ( 50%)
.............................................................  732 / 1327 ( 55%)
.............................................................  793 / 1327 ( 59%)
.............................................................  854 / 1327 ( 64%)
.............................................................  915 / 1327 ( 68%)
.............................................................  976 / 1327 ( 73%)
............................................................. 1037 / 1327 ( 78%)
............................................................. 1098 / 1327 ( 82%)
............................................................. 1159 / 1327 ( 87%)
............................................................. 1220 / 1327 ( 91%)
............................................................. 1281 / 1327 ( 96%)
..............................................                1327 / 1327 (100%)

Time: 00:32.341, Memory: 217.00 MB

OK (1327 tests, 4163 assertions)
```
10. Repeat same unit tests with other versions of PHP like 7.4
11. Run smoke tests with the extensions to confirm it doesn't have any issues with the container code

### Additional details:
peeuvX-1Zh-p2

### Changelog entry
* Fix - Initial php 8.4 compatibility.


mikkamp:
This would also supersede https://github.com/woocommerce/google-listings-and-ads/pull/2663

bin/GoogleAdsCleanupServices.php
```diff
@@ -61,7 +61,7 @@ class GoogleAdsCleanupServices {
 	 * @param Event|null  $event Composer event.
 	 * @param string|null $path Path of the Ads library.
 	 */
-	public function __construct( Event $event = null, string $path = null ) {
+	public function __construct( ?Event $event = null, ?string $path = null ) {
```
puntope, line 64: I see we added ? over several areas in the codebase. However, I see some of them not being changed.

For instance: https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/API/Google/MerchantReport.php#L67

puntope:
I get still one error in Unit Tests

Screenshot 2024-12-06 at 12 53 57

puntope:
> Create a temporary script to patch remaining packages in bin/fix-packages.php

Why not include this as the rest of the scripts like prefix-vendor-namespace.php?

puntope:
> replace_all(
	'vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/*.php',
	'string $template = null',
	'?string $template = null'
);

The first time you run this. It works. But if you run it another time It will result in:

'??string $template = null'

bin/GoogleAdsCleanupServices.php
mikkamp: I'm hoping I didn't miss any as I mostly focused on getting the unit tests to pass. So if I did miss something that wasn't unit tested then please do let me know.

The line you linked to doesn't need any change because the function is defined as:
```
public function get_product_view_report( $next_page_token = null ): array {
```

While the parameter is set to null, it doesn't have any type defined, so it doesn't matter if it's a string or null and even PHP 8.4 won't complain about it. If we would write that function again from scratch we should probably define it as:
```
public function get_product_view_report( ?string $next_page_token = null ): array {
```

But I don't think we want to go back through all our code and require types on every parameter.

For the places where we did fix it, we specifically used a `?` as that was introduced in PHP 7.1, we could have also gone for `string|null` but that's only supported in PHP 8.0, [see here](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated).

mikkamp:
> I get still one error in Unit Tests

Is that consistently reproducible? I'm not seeing that happen here on any PHP version. Can you confirm if you run just the individual test on a specific PHP version if you get it to fail:
`/usr/bin/php8.4 vendor/bin/phpunit --filter=test_add_subsize_image`

Based on the results it means that it won't create a 20x20 thumbnail image. I wonder if there is a PHP image module that is different between versions of PHP. Can you also run:
`diff <(/usr/bin/php8.4 -m) <(/usr/bin/php8.3 -m)`

Just to see if there is a difference in PHP modules.

> Why not include this as the rest of the scripts like prefix-vendor-namespace.php?

This means we are patching upstream libraries manually, it is an option, but instead we've asked them for compatible versions. Once we get the updated versions we won't need a script like that anymore. Considering that they are deprecation notices (shouldn't in theory be shown on production sites), I thought the better way was to wait till we get updated packages.

> The first time you run this. It works. But if you run it another time It will result in:
> '??string $template = null'

Yes, I should have mentioned that, it's a quick search/replace script so only intended for running once.

bin/GoogleAdsCleanupServices.php
puntope: Correct @mikkamp all the typed parameters contain the fix so far in my search.

puntope:
@mikkamp Re Image test

It does fail when running the test isolatedly.

Find also the output for `diff <(/usr/bin/php8.4 -m) <(/usr/bin/php8.2 -m)` (Notice the 8.2)

```
1a2
> bcmath
12a14
> gd
16c18,20
< imap
---
> igbinary
> imagick
> intl
18d21
< ldap
20a24,26
> memcache
> memcached
> msgpack
31a38
> redis
35a43
> soap
43a52
> xdebug
45a55
> xmlrpc
52a63
> Xdebug
```

`maybe_add_subsize_image` doesn't return any error. It just doesn't set the subsize in the metadata. 

Update:

@mikkamp I was doing some debugging and found the error in `wp_get_image_editor` inside `_wp_make_subsizes` ( this is being called in `wp_update_image_subsizes`)

```
Test 'Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Utility\ImageUtilityTest::test_add_subsize_image' ended
object(WP_Error)#7361 (3) {
  ["errors"]=>
  array(1) {
    ["image_no_editor"]=>
    array(1) {
      [0]=>
      string(28) "No editor could be selected."
    }
  }
  ["error_data"]=>
  array(0) {
  }
  ["additional_data":protected]=>
  array(0) {
  }
}
```

Update 2:

Solved installing  php8.4-imagick (as one of the editors)

puntope:
This is my php-v output btw

```
PHP 8.4.1 (cli) (built: Nov 25 2024 18:03:47) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.4.1, Copyright (c) Zend Technologies
    with Zend OPcache v8.4.1, Copyright (c), by Zend Technologies
```

puntope (approved):
Hi @mikkamp I solved the problem with the test. See ( https://github.com/woocommerce/google-listings-and-ads/pull/2714#issuecomment-2527103894 ) 

So no more blockers left.

mikkamp:
Thanks for clarifying. Either the `gd` or `imagick` image library is needed, I assume it failed because neither was available. I thought WordPress had a fallback for those cases, but generally most servers will have at least one available.

Going to merge this, now that it's running fine.

PR #2716, opened 2024-12-06
Combine `AppIconButton` and `HelpIconButton`

eason9487:

The `Button` from `@wordpress/components` supports having an icon. However, the existence of  `AppIconButton` makes it appear that there is a duplication of implementation between them.

Considering that the only use case for `AppIconButton` in the last four years has been `HelpIconButton`, this PR combines them into one.

#### Additional changes:

- Fix the incorrect `@return` tag in JSDoc
- Remove the forwarding of the rest props because all `HelpIconButton` use cases only set the `eventContext` prop.

1. Go to any page that renders the `HelpIconButton` component
   - Edit free listings
   - Create campaign
   - Edit campaign
   - Edit store address
   - Onboarding flow
   - Ads-onboarding flow
2. Check if `HelpIconButton` component is presented the same as before
   image
3. Click on the button to see if it can open plugin's document page in a new webpage

### Additional details:

💡 In the future, if there are more buttons need to set the position of icon relative to the top of the button text, it's preferred to extend `AppButton` to support `iconPosition="top"`.


ianlin (approved):
Thanks for cleaning up the help icon button components. Confirmed all the help icons displayed correctly, redirected to the right page after clicking it, and the events were tracked accurately. LGTM.

PR #2056, opened 2023-08-14
Externalize all `@wordpress` packages via DEWP

eason9487:
### Changes proposed in this Pull Request:

Closes #1833

💡 All code changes have been reviewed and tested in the respective PRs. Therefore, this PR should not require extra rounds of code reviews.

### Changelog entry

> Dev - Externalize all WordPress JavaScript packages via Dependency Extraction Webpack Plugin (DEWP).

PR #2718, opened 2024-12-09
Remove unnecessary Sass imports of `@wordpress/base-styles`

eason9487:

This PR removes unnecessary Sass imports of `@wordpress/base-styles`:
- [@wordpress/base-styles/z-index](https://github.com/WordPress/gutenberg/blob/%40wordpress/base-styles%404.49.0/packages/base-styles/_z-index.scss)
- [@wordpress/base-styles/default-custom-properties](https://github.com/WordPress/gutenberg/blob/%40wordpress/base-styles%404.49.0/packages/base-styles/_default-custom-properties.scss)

After externalizing `@wordpress/components` in #2056, it should no longer need to import `@wordpress/base-styles/z-index` because this codebase doesn't use `z-index()` in any .scss files.

I couldn't track back the reason for importing `@wordpress/base-styles/default-custom-properties`, but a plugin should be able to rely on the default WordPress Theme CSS variables provided by WordPress core instead of building the duplicated ones within plugin's CSS file.

image

Furthermore, removing it can fix the following Sass warning when building frontend code.

image

1. Confirm `npm run dev` can build frontend code without errors and the above Sass warning
2. Go to **User > Profile** and select the Default option for the **Admin Color Scheme** setting
3. View components that use the WordPress Theme CSS variables such as `--wp-admin-theme-color`
   - One of the most obvious examples is the `AppTabNav` component:
      image
      https://github.com/woocommerce/google-listings-and-ads/blob/e6d95d8601a19ce39d2efdd6134a2b3e169b5ec3/js/src/components/app-tab-nav/index.scss#L36-L38
   - Check if the default colors defined by WordPress Theme CSS variables are applied correctly

### DIFF
--- a/js/src/css/index.scss
+++ b/js/src/css/index.scss
@@ -1,7 +1,3 @@
-// Default Styles Here
-@import "node_modules/@wordpress/base-styles/z-index";
-@import "node_modules/@wordpress/base-styles/default-custom-properties";
-
 // Import Gutenberg components that aren't already imported in the lowest WC Admin version we support
 @import "./shared/gutenberg-components";
 


puntope (approved):
Thanks @eason9487 for the cleanup.

For z-index() I wasn't able to find any usage either. 

For `default-custom-properties`  this comment in the style caught my attention

> // It is important to include these styles in all built stylesheets.
// This allows to CSS variables post CSS plugin to generate fallbacks.
// It also provides default CSS variables for npm package consumers.

In any case, since I'm not seeing any difference In the UI I'm approving the PR.

eason9487:

> For `default-custom-properties` this comment in the style caught my attention
> 
> > // It is important to include these styles in all built stylesheets.
> > // This allows to CSS variables post CSS plugin to generate fallbacks.
> > // It also provides default CSS variables for npm package consumers.

I think the code comment means that from the perspective of the Gutenberg libraries, it must be included to ensure the libraries can be used independently. A similar case is that the CSS variables [`--large-gap` and `--main-gap`](https://github.com/woocommerce/woocommerce/blob/9.4.3/plugins/woocommerce-admin/client/stylesheets/shared/_global.scss#L3-L6), which are heavily used by this plugin but are not included directly, relying instead on the ones built in WooCommerce core libraries.

PR #2723, opened 2024-12-12
Drop Jetpack packages and switch to packages bundled with WooCommerce

mikkamp:

WooCommerce has included Jetpack packages since version 7.9, since it uses the autoloader we can rely on the fact that WooCommerce will ensure any of the connection packages are loaded. We always require WooCommerce to be active before loading our service classes so this means we no longer need to bundle any of the Jetpack packages for facilitating the connection. Since that removes all Jetpack packages we can also revert back to using the composer autoloader.

#### Remove Jetpack packages
```
$ composer remove automattic/jetpack* --with-all-dependencies
  - Removing automattic/jetpack-a8c-mc-stats (v2.0.4)
  - Removing automattic/jetpack-admin-ui (v0.4.6)
  - Removing automattic/jetpack-assets (v2.3.14)
  - Removing automattic/jetpack-autoloader (v3.0.2)
  - Removing automattic/jetpack-config (v2.0.0)
  - Removing automattic/jetpack-connection (v2.12.6)
  - Removing automattic/jetpack-constants (v2.0.5)
  - Removing automattic/jetpack-redirect (v2.0.3)
  - Removing automattic/jetpack-roles (v2.0.4)
  - Removing automattic/jetpack-status (v3.3.5)
```

This PR also bumps the minimum WooCommerce version to 7.9 to ensure it's not used with any versions that don't bundle the connection packages.

> [!Note]
> The PHP unit tests were failing because ubuntu-latest switched from 22.04 > 24.04 which means the svn command is no longer available. [These commits install the package](https://github.com/woocommerce/google-listings-and-ads/pull/2723/files/8f8f2fe0166c50f7d40f6fb333a0c59be7072a63..741cc732ae061b4a57e1d5fadbc90cad5980a987).

1. Checkout this PR
2. Update composer packages `rm -rf vendor && composer install`
3. Test connecting your WordPress.com account during onboarding
4. Complete onboarding and confirm that requests to the WCS using the Jetpack token continue to work
5. Visit the connection test page: `https://domain.test/wp-admin/admin.php?page=connection-test-admin-page`
6. Click the "WordPress.com Connection Status" button
7. Confirm that everything is connected and we see our plugin in the list:
```
Connected plugins: Google for WooCommerce, WooCommerce
```

#### Test WooCommerce deactivated
Since we use the plugin dependency header you can't test with out plugin active and WooCommerce deactivated, however this can be simulated using the following steps:

1. Setup the site with WP_DEBUG enabled and debug logging turned on
2. Activate both WooCommerce and our plugin
3. Rename the folder `wp-content/plugins/woocommerce` to force it to be deactivated
4. Reload a page in the WordPress dashboard
5. Confirm that Google for WooCommerce does not produce any fatals or errors and gracefully displays the message that WC 7.9 or higher is required

### Additional details:
pcShBQ-2En-p2

### Changelog entry
* Tweak - Drop Jetpack packages and switch to packages bundled with WooCommerce.
* Update - Drop support for WooCommerce < 7.9.


tomalec (approved):
Thanks for a quick action to fix & update that and fix the tests! :)

Reviewed the code and tested locally. LGTM!

Side note:
Maybe we could somewhat unify the copy & bolds we use in those messages 
image

Personally, I'd find 
> **Extension Name is inactive**. The plugin requires WooCommerce version **x.y** or higher.

The most readable, but it does not seem to match any existing one ;)

eason9487:
There are two places need to be updated as well:

https://github.com/woocommerce/google-listings-and-ads/blob/741cc732ae061b4a57e1d5fadbc90cad5980a987/README.md?plain=1#L31

https://github.com/woocommerce/google-listings-and-ads/blob/741cc732ae061b4a57e1d5fadbc90cad5980a987/readme.txt#L55

mikkamp:
> > Extension Name is inactive. The plugin requires WooCommerce version x.y or higher.

> The most readable, but it does not seem to match any existing one ;)

Makes sense, @tomalec could you log some issues for that, sounds like a good task to pick up during a cooldown.

tomalec:
> Could you log some issues for that

Issue created at https://github.com/woocommerce/woocommerce/issues/53728

We could make it for all extensions, if we're brave enough, or just for Automata-owned.

PR #2734, opened 2024-12-17
Prevent duplicate conversion and purchase event tracking

### LINKED ISSUE #2733: Duplicate reporting due to incorrect meta value comparison
### Describe the bug:
I had noticed that a few of the websites I manage was having issues double reporting and it seemed to be from people that were revisiting their thank you page across multiple days. I have nailed the issue down to the file “GlobalSiteTag.php” located in the src > Google folder. The issue resides on line 358 and has to do with the strict checking of the returned meta key value. The function “get_meta” will return a string and not an integer in this case because the database for meta values is set as text. To fix this issue, Line 358 – 360 in the file “GlobalSiteTag.php” should change from:
```
if ( ! $order || 1 === $order->get_meta( self::ORDER_CONVERSION_META_KEY, true ) ) {
    return;
}
```
TO:
```
if ( ! $order || '1' === $order->get_meta( self::ORDER_CONVERSION_META_KEY, true ) ) {
    return;
}
```

To stay consistent, the line 363 should also be changed to a string instead of a integer.

### Steps to reproduce:
1. Complete a checkout and on the thank you page, inspect the page code and look for the gtag conversion event.
2. Verify in the database for this order that the meta value for "_gla_tracked" is set to 1.
3. Refresh the thank you page and the gtag conversion event is still there even with the meta value set to 1

### Expected behavior:
The gtag conversion event should not fire after the thank you page loads for the first time.

### Actual behavior:
The gtag conversion event is firing on every page refresh.

### Additional details:
We are currently running version 2.9.2 and I have been testing this fix for the past month or so and our double reporting has stopped happening.

mikkamp:

Because of a strict comparison check for the `_gla_tracked` meta we were allowing the conversion and purchase snippets to show again if the order thank you page is refreshed (not loaded from cache).

This PR converts the meta value to an integer before doing the strict comparison. I've also added some unit tests to confirm that the tracking snippet isn't output in conditions where it should not. The WP proxy class had to be used to facilitate mocking of some WP functions.

Closes #2733

1. Setup your site fully onboarded with an Ads account to enable tracking.
2. Go through the checkout flow in an anonymous browser window until you reach the order thank you page
3. Check the page source and confirm it has two inline snippets
   1. Starting with `gtag("event", "conversion"`
   2. Starting with `gtag("event", "purchase"`
4. Hard refresh the page
5. Check the page source and confirm this time the snippets are no longer included

### Changelog entry
* Fix - Prevent duplicate conversion and purchase event tracking.

### DIFF
--- a/tests/Unit/Google/GlobalSiteTagTest.php
+++ b/tests/Unit/Google/GlobalSiteTagTest.php
@@ -1,0 +1,113 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Google;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Assets\AssetsHandlerInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\GlobalSiteTag;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\GoogleGtagJs;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
+use PHPUnit\Framework\MockObject\MockObject;
+use WC_Helper_Order;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class GlobalSiteTag
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Google
+ */
+class GlobalSiteTagTest extends UnitTest {
+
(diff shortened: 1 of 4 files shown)


tests/Unit/Google/GlobalSiteTagTest.php
```diff
@@ -0,0 +1,113 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WC;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\UnitTest;
+use PHPUnit\Framework\MockObject\MockObject;
+use WC_Helper_Order;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class GlobalSiteTag
```
eason9487: Should it be the same as the class name `GlobalSiteTagTest` below?

eason9487 (approved):
Thanks for fixing the issue quickly!

Confirmed that gtag snippets won't be included in the order thank you page after an order is marked as tracked. LGTM.

PR #2738, opened 2024-12-18
Clear previous errors after completing sync

### LINKED ISSUE #2737: Pre sync errors are not always cleared after successful sync
### Describe the bug:
If we encounter pre sync errors we store these in the product meta with the key `_wc_gla_errors`. These are mostly for validation issues before the product is sent to the Merchant Center. One such error is an empty description field.

If we previously had these errors for specific products, we'd expect them to be cleared up after resolving the error and syncing the product. These issues are cleared up in the product edit view. But they continue to popup in the product issues table on the Product Feed page:

Image

The reason this is not cleared up is because of a mismatch in synced countries with target countries. We can see this in the [following code](https://github.com/woocommerce/google-listings-and-ads/blob/2.9.3/src/Product/ProductHelper.php#L108-L111):
```php
// check if product is synced completely and remove any previous errors if it is
$synced_countries = array_keys( $google_ids );
$target_countries = $this->target_audience->get_target_countries();
if ( count( $synced_countries ) === count( $target_countries ) && empty( array_diff( $synced_countries, $target_countries ) ) ) {
```

That's asking to fulfill two conditions:
1. That the count of synced countries matches the count of target countries
2. That all synced countries are included in target countries

Before v2 when this code was added, that was correct, because we were syncing each product individually for each target country. Later on that was refactored to only sync once for the base country, and include a list of target countries. However it seems this check wasn't refactored. So in case where there is only one target country it would still work, but if we have multiple target countries, it won't clear up old errors after a successful sync.

This went unnoticed in unit testing because the tests also remained the same to explicitly test the old scenarios:
1. test_mark_as_synced
2. test_mark_as_synced_doesnt_delete_errors_unless_all_target_countries_synced

In particular that second test needs to be changed to match the correct behaviour.

#### Workaround
One workaround to get around this issue is to follow these steps:
- Edit a product and set the channel visibility to "Do not Sync and Show"
- Save the product which triggers the product to be saved as unsynced (which clears previous errors)
- Update the channel visibility to "Sync and Show"
- Save the product and wait for the scheduled sync to complete

### Steps to reproduce:
1. On a site where we completed onboarding (and use a live URL so we are allowed to sync)
2. During onboarding make sure to select multiple target countries
3. Create a product with no description (complete all other requirements such as product image and price)
4. Save the product with Sync and Show status to trigger the sync
5. Wait for Action Scheduler to complete sync
6. Edit the product and confirm the error has shown up after sync attempt
Image
7. Update the product to add a description and complete all requirements so it can sync
8. Wait for the product sync to complete
9. Edit the product and confirm the errors don't show up in the product
10. View the product issues in the Product Feed page and confirm that the issues remain
11. Clear the product status cache to reload it (tool in connection test page)
12. View the product issues table again and confirm that the errors are still not cleared up

### Expected behavior:
Local product errors to be cleared up after a successful sync

### Actual behavior:
Clearing the product errors is skipped because the synced countries doesn't match the full list of target countries.

### Additional details:
p1734476920010379-slack-C0410KV3YLW

mikkamp:

This PR clears out previous sync errors if the synced countries are all contained within the set of target countries. Previously the length of the arrays needed to match as well. This is no longer the case since we don't sync each target country individually anymore.

The unit tests were updated to reflect the new behaviour.

Closes #2737 

1. On a site where we completed onboarding (and use a live URL so we are allowed to sync)
2. During onboarding make sure to select multiple target countries
3. Create a product with no description (complete all other requirements such as product image and price)
4. Save the product with Sync and Show status to trigger the sync
5. Wait for Action Scheduler to complete sync
6. Edit the product and confirm the error has shown up after sync attempt
Image
7. View the product issues in the Product Feed page and confirm that the issue is showing up there as well
8. Update the product to add a description and complete all requirements so it can sync
9. Wait for the product sync to complete
10. Edit the product and confirm the errors don't show up in the product
11. View the product issues in the Product Feed page and confirm that the issue is not showing up there either
12. Clear the product status cache to reload it (tool in connection test page)
13. View the product issues table again and confirm that the error is still not there

### Changelog entry
* Fix - Clear previous errors after completing sync.

### DIFF
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -105,10 +105,10 @@ class ProductHelper implements Service, HelperNotificationInterface {
 		$google_ids         = array_unique( array_merge( $current_google_ids, [ $google_product->getTargetCountry() => $google_product->getId() ] ) );
 		$this->meta_handler->update_google_ids( $product, $google_ids );
 
-		// check if product is synced completely and remove any previous errors if it is
+		// check if product is synced for main target country and remove any previous errors if it is
 		$synced_countries = array_keys( $google_ids );
 		$target_countries = $this->target_audience->get_target_countries();
-		if ( count( $synced_countries ) === count( $target_countries ) && empty( array_diff( $synced_countries, $target_countries ) ) ) {
+		if ( empty( array_diff( $synced_countries, $target_countries ) ) ) {
 			$this->meta_handler->delete_errors( $product );
 			$this->meta_handler->delete_failed_sync_attempts( $product );
 			$this->meta_handler->delete_sync_failed_at( $product );
--- a/tests/Unit/Product/ProductHelperTest.php
+++ b/tests/Unit/Product/ProductHelperTest.php
@@ -100,7 +100,7 @@ class ProductHelperTest extends ContainerAwareUnitTest {
 	 *
 	 * @dataProvider return_test_products
 	 */
-	public function test_mark_as_synced_doesnt_delete_errors_unless_all_target_countries_synced( WC_Product $product ) {
+	public function test_mark_as_synced_deletes_errors_when_main_target_countries_synced( WC_Product $product ) {
 		$google_product = $this->generate_google_product_mock();
 
 		$this->target_audience->expects( $this->any() )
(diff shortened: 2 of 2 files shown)


eason9487 (approved):
Thanks for the fix! Confirmed that after product issues are resolved and the product is synced, the corresponding issues on the product editor and the "Issues to resolve" table no longer appear. LGTM.

PR #2739, opened 2024-12-18
Adjust conditions for MCM

mikkamp:

As laid out in the issue https://github.com/woocommerce/woocommerce/issues/53753
The conditions for registering Multi Channel Marketing channels has changed. This PR switches to using a `class_exists` check instead of a constant.

1. Test with at least WC 9.5
2. Enable the snippet `add_filter( 'woocommerce_gla_enable_mcm', '__return_true' );`
3. Onboard with an ads account and make sure at least one campaign is created
4. Go to Marketing > Overview
5. Confirm that the channel is not showing
image
6. Check out this PR and reload the Marketing > Overview page
7. Confirm that the channels and campaigns are showing
image

### Changelog entry
* Tweak - Adjust conditions for MCM

### DIFF
--- a/src/Internal/DependencyManagement/CoreServiceProvider.php
+++ b/src/Internal/DependencyManagement/CoreServiceProvider.php
@@ -3,6 +3,7 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Internal\DependencyManagement;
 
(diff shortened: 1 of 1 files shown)


eason9487 (approved):
Thanks for the fix! LGTM.

Confirmed that the statuses and campaigns of this plugin can be shown on **Marketing > Overview** page. I also tested with an older version, WooCommerce 9.0, and the solution is compatible as well.

image

PR #2728, opened 2024-12-16
PHP 8.4 package compatibility

mikkamp:
### Changes proposed in this Pull Request:

This PR updates all the Google packages we rely on for the Google Ads API packages to make them fully PHP 8.4 compatible.

Note that some of the changes to the Google Ads API packages itself have only been applied to V18 of the API. Currently we aren't fully ready to switch from V16 to V18 yet as we need to become compatible with the removal of the page_size parameter. Until we will need to continue using a small patch script to test changes with PHP 8.4.

##### Composer package updates
```
$ composer update googleads/google-ads-php --with-all-dependencies
  - Upgrading google/auth (v1.37.1 => v1.37.2)
  - Upgrading google/gax (v1.29.1 => v1.29.2)
  - Upgrading google/longrunning (0.4.1 => 0.4.6)
  - Upgrading google/protobuf (v3.25.3 => v3.25.5)
  - Upgrading googleads/google-ads-php (dev-legacy-v22.1.0 905fd78 => dev-legacy-v25.0.0 3f4ea59)
  - Upgrading monolog/monolog (2.9.3 => 2.10.0)
  - Upgrading symfony/polyfill-intl-normalizer (v1.29.0 => v1.31.0)
  - Upgrading symfony/polyfill-mbstring (v1.29.0 => v1.31.0)
```

```
$ composer update google/apiclient --with-all-dependencies
  - Upgrading google/apiclient (v2.16.0 => v2.16.1)
  - Upgrading paragonie/constant_time_encoding (v2.6.3 => v2.7.0)
  - Upgrading phpseclib/phpseclib (3.0.37 => 3.0.43)
```

We also updated the script to remove newer versions of the Ads API library V17 and V18, to keep the bundled size lower until we are ready to switch to V18.

### Detailed test instructions:
1. Checkout this branch
2. Update packages `rm -rf vendor && composer install`
3. Create a temporary script to patch remaining packages in `bin/fix-packages.php`
```
<?php

function replace_all( $path, $search, $replace ) {
	foreach ( glob( $path ) as $file ) {
		$content = preg_replace(
			'/' . preg_quote( $search, '/' ) . '/',
			$replace,
			file_get_contents( $file )
		);
		file_put_contents( $file, $content );
	}
}

replace_all(
	'vendor/googleads/google-ads-php/src/Google/Ads/GoogleAds/V16/Services/Client/*.php',
	'string $template = null',
	'?string $template = null'
);
```
4. Patch packages `php bin/fix-packages.php`
[...]

PR #2740, opened 2024-12-19
Update Google Ads API  to v18

mikkamp:

This PR update the Ads API to use V18. The composer packages were already updated in a [previous PR](https://github.com/woocommerce/google-listings-and-ads/pull/2728).

#### Required changes:
**V16 > V17**
GoogleAdsService Search requests: Passing a page_size to GoogleAdsService.Search will result in a RequestError.PAGE_SIZE_NOT_SUPPORTED error.

**V17 > V18**
GoogleAdsService: The summary_row_setting and return_total_results_count fields in SearchGoogleAdsRequest are now part of the new SearchSettings object.

The original pagination support for campaigns was reverted from [this PR](https://github.com/woocommerce/google-listings-and-ads/pull/2472), as the Google Ads API has removed the `pageSize` parameter and sets it to a fixed value of 10000 rows. For general use cases that's ok because of the associated costs most merchants won't be managing 1000's of campaigns.
The initial reason why we did add pagination to the campaigns endpoint is to support the use case where the mobile app only wanted to fetch a single campaign. For this purpose I've left the `per_page` parameter in place to limit the API response. We still get the full set from the Ads API but we limit the results sent to the API. Because of this limitation it's not possible to fetch subsequent pages. We have a [feature request open](https://github.com/woocommerce/google-listings-and-ads/issues/1887) if we want to implement a full pagination support in the future.

As a temporary measure the headers returning totals and pagination parameters was removed in this PR until we are ready to fully support pagination. I also added some CSS to remove the pagination from the list of campaigns in the dashboard, since this doesn't work.

image

1. Check out this PR
2. Reinstall composer packages: `rm -rf vendor && composer install`
3. Rebuild JS: `nvm use && npm start`
4. Do a full [smoke tests](https://github.com/woocommerce/google-listings-and-ads/wiki/Smoke-tests#create-a-campaign) with all requests to the Ads API
5. Test creation of campaigns (with/without assets), as well as edit and delete of campaigns

### Additional details:
pcTzPl-2zd-p2

### Changelog entry
* Update - Google Ads API to v18.


ianlin (approved):
Thanks for updating the Google Ads API. I ran the full smoke tests, created campaigns with/without assets, everything worked as expected.

PR #2742, opened 2024-12-20
Reorganizes the frontend code related to the calls of settings APIs

eason9487:

This is one of the preliminaries for the shipping Improvements.

To avoid too many file changes for subsequent PRs, this PR reorganizes the code related to the calls of settings APIs:

- Add the `saveSettings` action as one of the return values to the `useSettings` hook and update the related uses.
   - Since they are always used together, and there will be a third use case in the subsequent PR.
- Create a new action `syncSettings`, add it as one of the return values to the `useSettings` hook and update the related uses.
   - There will be a third use case of calling to `mc/settings/sync` API in the subsequent PR.
   - They are often used together.
- Correct the JSDoc description for the `saveSettings` action.

   1. Open the Network tab of browser's DevTool
   2. Change the **Shipping rates** setting
      image
   3. Inspect Network tab. Check if there is a request issued to `wc/gla/mc/settings` API  and if the request is successful
   4. Select the **Preserve log** option on the Network tab
   5. Continue to complete the onboarding flow
   6. Inspect Network tab. Check if there is a request issued to `wc/gla/mc/settings/sync` API  and if the request is successful
2. Go to edit free listing page
   1. Open the Network tab of browser's DevTool
   2. Change the **Tax rate** or **Shipping rates** setting
   3. Save changes
   4. Inspect Network tab.
   5. Check if there is a request issued to `wc/gla/mc/settings` API  and if the request is successful
   6. Check if there is a request issued to `wc/gla/mc/settings/sync` API  and if the request is successful
      image

### DIFF
--- a/js/src/data/actions.js
+++ b/js/src/data/actions.js
@@ -325,10 +325,10 @@ export function* fetchSettings() {
 }
 
 /**
- * Save the the MC settings.
+ * Save the plugin settings.
  *
- * @param {SettingsData} settings settings
- * @return {Object} Action object to save target audience.
+ * @param {SettingsData} settings Plugin settings
+ * @return {Object} Action object to save the plugin settings.
  */
 export function* saveSettings( settings ) {
 	yield apiFetch( {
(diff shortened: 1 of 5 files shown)


tomalec (approved):
Thank you for the changes @eason9487 !
I reviewed the code and run both flows, all LGTM.

PR #2751, opened 2025-01-06
Move the Edit free listings page to a dedicated page and rename page to Shipping

eason9487:

Project thread: pcTzPl-2Cf-p2

This PR implements the **📌 Move shipping settings to dedicated page** task of the project.

💡 Update for `src/Tracking/README.md` will be done by the conjunction PR for the feature branch.

   - Check if the UI is presented the same as before
   - Check if the editing functions of audience and shipping work well
2. Go to the Shipping page
   - Check if the editing functions of audience and shipping work well
3. View the E2E testing result: https://github.com/woocommerce/google-listings-and-ads/actions/runs/12631383645

### Changelog entry

> Update - Move the Edit free listings page to a dedicated page and rename the page to Shipping


puntope:
In the PT claims:

> When we edit shipping settings or change locations it will trigger a re-sync of products, which can in turn affect Free Listings / paid campaigns. Previously we had a popup that would appear before clicking the edit button:

I didn't see that Modal when saving  Settings. I assume it will be implemented in a future PR?

js/src/components/free-listings/setup-free-listings/index.js
```diff
@@ -222,10 +225,36 @@ const SetupFreeListings = ( {
...
+						</>
+					);
+				} }
 			</AdaptiveForm>
 		</div>
 	);
 };
 
+SetupFreeListings.SubmitButton = Slot;
+
```
puntope, line 259: This part is not very easy to understand for me. I assume that when we use 
 <SetupFreeListings.SubmitButton /> (the Slot) sowhere. It will render whatever is inside <Fill>?

puntope (approved):
Thanks @eason9487 this PR LGTM

- Tested that the onbaording works as always.
- Tested you cannot "edit" anymore Free Listings either with button in the UI (not visible anymore) or by entering in the URL
- Tested Shipping page. It shows the same info as Free Listings before and can be saved. 

Just a comment about a modal that I expected to see there. But I assume it will be implemented further.

js/src/components/free-listings/setup-free-listings/index.js
eason9487: Yes. The content inside the `<Fill>` will be rendered to the paired `<Slot>`, and the component consumer side decides where the slot is placed. In this way, the submit button can be rendered more flexibly outside of `SetupFreeListings` and can be composited with other components.

It's used here to present different layouts. One has stepper wrappers and one doesn't.

eason9487:
He @puntope, thanks for the review!

> I didn't see that Modal when saving Settings. I assume it will be implemented in a future PR?

Yes. This PR has already accumulated a lot of changes, so I think it would be nicer to do so.

PR #2754, opened 2025-01-08
Fix tests in WooCommerce 9.6

puntope:

[Tests are failing when using 9.6.0-beta.2 ](https://github.com/woocommerce/google-listings-and-ads/actions/runs/12672738051)

- One test regarding the default weight measure unit
- Another regarding the Features. That are not loaded anymore in our test environment.

That is because the code in https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/Features/Features.php#L56

This PR adapts the test to fix those 2.

### Screenshots:

Screenshot 2025-01-08 at 21 34 53

1. [See the test using 9.6.0-beta.2 passing against this branch ](https://github.com/woocommerce/google-listings-and-ads/actions/runs/12678666426/job/35336717102)
2.[See the test using latest passing against this branch ](https://github.com/woocommerce/google-listings-and-ads/actions/runs/12678628630/job/35336614816)

### Changelog entry

> Dev - Fix tests in WooCommerce 9.6


eason9487 (approved):
Thanks for the fix! LGTM.

PR #2761, opened 2025-01-10
Consistent usage of ContainerAware

### LINKED ISSUE #563: Refactor to prefer direct dependencies but allow ContainerAwareness
## Technical
After some discussion, the consensus decision is to default to using direct dependencies, but allow some flexibility to use `ContainerAware`-ness.

This is a reminder issue to review the existing codebase and fix up any glaring examples that aren't consistent with this decision (using the container for just a few dependencies, providing a laundry list of dependencies).

Some general guidelines pulled from the developer discussion:

> … use direct dependencies as much as possible, but I think we should be willing to say that it is OK to <em>occasionally</em> use the container. As long as it doesn’t become the normal go-to option.

> … suggest that we only pass the container to the controllers or any other class that want to fetch many of the container stored variables (if we ever use them). In other cases, … it’s best to directly define the class’ dependency instead of passing the container.

> … only pass dependencies when there aren’t more then @ 3.<br>Also when we refactor, we should check if we can use `ContainerAware` and `OptionsAware` in more places to reduce the constructor variables.

> Try to use direct dependencies as much as possible, until the number reaches some arbitrary limit of “too many”. Then either refactor [to split monolithic or overgrown classes], or pass the container and <strong>indicate in the class or constructor documentation what dependencies are used/expected in the class.</strong>

mikkamp:

This PR standardizes the usage of ContainerAware across various classes. This makes it more consistent with what was decided in issue #563

I've also fixed some of the AdminConditional loading to ensure that some of the classes are not loaded on the frontend.
I was considering adding `is_admin() && ! wp_doing_ajax()` to the AdminConditional. But for now I left it since there could be a possibility that coupon bulk editing or activating plugin is done through an ajax request, which might silently break some of the actions. So for now I've left it as we'd need to do some extensive testing to make sure we aren't missing any ajax requests.

When looking at a front end request we can thus see a small decrease in the time spent resolving classes.

Before:
image

After:
image

PR #2758 takes that a step further. We can still decrease it even more, but considering that it needs specific testing for each change, I think it's better to improve that through multiple separate PR's.

Closes #563

1. Test some basic REST API requests while going through onboarding
2. As well as going through some of the extension pages
3. Confirm that none of the requests fail

### Changelog entry
* Tweak - Consistent usage of ContainerAware.


ianlin (approved):
Thanks for refactoring to use ContainerAware consistently and also add / fix the conditional loading for some classes . I have gone through the onboarding and tested all the pages within the extension, the requests were successful as normal.

PR #2762, opened 2025-01-13
Avoid the negative shipping rate leading to false onboarding completion and make its error message more explicit

### LINKED ISSUE #2466: Negative shipping rate setting leads to false onboarding completion and failure to update free listings
### Describe the bug:

https://github.com/woocommerce/google-listings-and-ads/assets/17420811/be47a991-935a-4828-a81d-149aa34e7898

#### Onboarding

1. Go to **WooCommerce > Settings > Shipping > Shipping zones** and set up a zone with negative flat rate.
   image
2. Start onboarding and proceed to step 2.
   1. Add the above configured zone to the Audience setting.
   2. Select the **Recommended:  Automatically sync my store’s shipping settings to Google** option for the shipping rates setting.
3. Proceed to complete the onboarding. It shows an error: "Unable to complete your setup." but it still continues to finish the onboarding.
4. After being redirected to the Product Feed page, it shows more errors.
   image

#### Update free settings

1. Use the same setup as above.
2. Go to edit free listings.
3. Save changes. It shows an error: "Something went wrong while saving your changes. Please try again later."

### Expected behavior:

Maybe prompt users about shipping settings that are not supported by Google Merchant Center when syncing settings.

### Actual behavior:

- False onboarding completion
- Unable to update free listings

### Additional details:

Error log

```
Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\MerchantCenter\{closure} {
  "error": {
    "code": 400,
    "message": "[services[1].rateGroups[0].singleValue] Following rates are negative: -1000000",
    "errors": [
      {
        "message": "[services[1].rateGroups[0].singleValue] Following rates are negative: -1000000",
        "domain": "global",
        "reason": "invalid"
      }
    ]
  }
}
```

eason9487:

Closes #2466

This PR avoids negative shipping rate leads to false onboarding completion and makes its error message more explicit.

- Update the PHP `CountryRatesCollection` class to throw `InvalidValue` when any shipping rates are negative.
- Avoid false onboarding completion when there are negative shipping rates and clarify the error message.
- Make the error message of the negative shipping rates more explicit on the shipping page.

#### Preparation

1. Go to **WooCommerce > Settings > Shipping > Shipping zones**
2. Set up a zone with negative flat rate.
   image

#### Onboarding

1. Start onboarding and proceed to step 2.
   1. Add the above configured zone to the Audience setting.
   2. Select the **Recommended:  Automatically sync my store’s shipping settings to Google** option for the shipping rates setting.
2. Proceed to step 3.
3. Click "Skip ads creation" or "Complete setup" button to see if it shows a proper error message and stop completing onboarding.
   image
4. Switch back to the  **Shipping zones** setting page and update the shipping rate to a non-negative value.
5. Back to the onboarding and see if the flow can be completed.

#### Shipping page

1. Redo the test preparation.
2. Go to plugin's shipping page.
3. Click "Save changes" button to see if it shows a proper error message.
   image
4. Switch back to the  **Shipping zones** setting page and update the shipping rate to a non-negative value.
5. Back to plugin's shipping page and see if the saving can be completed.

### Changelog entry

> Fix - Avoid the negative shipping rate taken from WooCommerce settings leading to false onboarding completion and make its error message more explicit.

### DIFF
--- a/js/src/pages/onboarding/setup-stepper/setup-paid-ads.js
+++ b/js/src/pages/onboarding/setup-stepper/setup-paid-ads.js
@@ -9,7 +9,6 @@ import { noop } from 'lodash';
  * Internal dependencies
  */
 import useAdminUrl from '~/hooks/useAdminUrl';
-import useDispatchCoreNotices from '~/hooks/useDispatchCoreNotices';
 import useAdsSetupCompleteCallback from '~/hooks/useAdsSetupCompleteCallback';
 import useTargetAudienceFinalCountryCodes from '~/hooks/useTargetAudienceFinalCountryCodes';
 import AdsCampaign from '~/components/paid-ads/ads-campaign';
@@ -17,6 +16,7 @@ import CampaignAssetsForm from '~/components/paid-ads/campaign-assets-form';
 import AppButton from '~/components/app-button';
 import useGoogleAdsAccountBillingStatus from '~/hooks/useGoogleAdsAccountBillingStatus';
 import { getProductFeedUrl } from '~/utils/urls';
+import { handleApiError } from '~/utils/handleError';
 import { useAppDispatch } from '~/data';
 import { GUIDE_NAMES, GOOGLE_ADS_BILLING_STATUS } from '~/constants';
 import { ACTION_COMPLETE, ACTION_SKIP } from './constants';
@@ -41,7 +41,6 @@ import AppSpinner from '~/components/app-spinner';
 export default function SetupPaidAds() {
 	const adminUrl = useAdminUrl();
 	const [ completing, setCompleting ] = useState( null );
-	const { createNotice } = useDispatchCoreNotices();
 	const { data: countryCodes } = useTargetAudienceFinalCountryCodes();
 	const { highestDailyBudget, hasFinishedResolution } =
 		useBudgetRecommendation( countryCodes );
@@ -54,18 +53,19 @@ export default function SetupPaidAds() {
 
 	const finishOnboardingSetup = async ( onBeforeFinish = noop ) => {
 		try {
-			await onBeforeFinish();
 			await syncSettings();
+			await onBeforeFinish();
 		} catch ( e ) {
 			setCompleting( null );
 
-			createNotice(
-				'error',
+			handleApiError(
+				e,
 				__(
 					'Unable to complete your setup.',
 					'google-listings-and-ads'
 				)
 			);
+			return;
 		}
 
 		// Force reload WC admin page to initiate the relevant dependencies of the Dashboard page.
--- a/js/src/pages/shipping/index.js
+++ b/js/src/pages/shipping/index.js
@@ -23,6 +23,7 @@ import hasUnsavedShippingRates from './hasUnsavedShippingRates';
 import useSaveShippingRates from '~/hooks/useSaveShippingRates';
 import useSaveShippingTimes from '~/hooks/useSaveShippingTimes';
 import createErrorMessageForRejectedPromises from '~/utils/createErrorMessageForRejectedPromises';
+import { handleApiError } from '~/utils/handleError';
 import { recordGlaEvent } from '~/utils/tracks';
 
 /**
@@ -173,8 +174,9 @@ export default function Shipping() {
 
 			recordGlaEvent( 'gla_free_campaign_edited' );
 		} catch ( error ) {
-			createNotice(
-				'error',
+			handleApiError(
+				error,
+				__( 'Unable to save your changes.', 'google-listings-and-ads' ),
 				__(
 					'Something went wrong while saving your changes. Please try again later.',
 					'google-listings-and-ads'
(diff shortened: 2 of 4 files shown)


mikkamp (commented):
Thanks for these improvements, it is working as described when testing the functionality.

However if we aren't supporting negative shipping rates then shouldn't we also block these rates from being parsed? We recently [added a check for showing the automatic shipping option in onboarding](https://github.com/woocommerce/google-listings-and-ads/blob/2.9.4/js/src/components/shipping-rate-section/shipping-rate-section.js#L31-L35), that needs at least 1 parsable rate for the option to show.

Otherwise if we query `mc/settings` with only one negative shipping method setup `shipping_rates_count` will still return 1, but then we'll prevent it from syncing later on.
Shouldn't we [add something here to return null](https://github.com/woocommerce/google-listings-and-ads/blob/2.9.4/src/Shipping/ZoneMethodsParser.php#L125) if the final rate is still negative?

While this flow makes sense when editing shipping settings, I'm not too sure if this is the expected behaviour for handling the onboarding. We previously haven't blocked onboarding in this way, as we try to make it as easy as possible for merchants to complete. We also want to continue supporting merchants setting up shipping after onboarding. Should we only make this blocking when they edit the shipping settings? We'll still want to eventually refine the "steps to complete after onboarding" part, but that's out of scope for this PR.

eason9487:
Hi @mikkamp

The integration of WooCommerce Shipping Zones in the current implementation isn't complete enough, when a shipping zone isn't compatible with Google Merchant Center or when there are no shipping zones can fulfill a selected audience country, one of the two situations may happen:

1. Incompatible shipping setting is rejected by Google Merchant Center API when syncing and may lead to false onboarding completion.
2. Incompatible WooCommerce Shipping zone is omitted when syncing, and the selected audience country may not have any shipping setting on Google Merchant Center.

> Otherwise if we query `mc/settings` with only one negative shipping method setup `shipping_rates_count` will still return 1, but then we'll prevent it from syncing later on.
> Shouldn't we [add something here to return null](https://github.com/woocommerce/google-listings-and-ads/blob/2.9.4/src/Shipping/ZoneMethodsParser.php#L125) if the final rate is still negative?

The main goals of this PR are to avoid false completion of onboarding and to make the error messages clearer. Changing a negative shipping rate to return `null` would also accomplish the same goal of avoiding false completion of onboarding. However, in terms of user experience, it does make the onboarding process a bit smoother, but it's hard to say whether it results in a better outcome. The tradeoff is it may lead users to believe that WooCommerce Shipping Zones were converted and synced nicely to the shipping settings on Google Merchant Center, until one day they realize that the settings were in fact not synced at all.

> However if we aren't supporting negative shipping rates then shouldn't we also block these rates from being parsed? We recently [added a check for showing the automatic shipping option in onboarding](https://github.com/woocommerce/google-listings-and-ads/blob/2.9.4/js/src/components/shipping-rate-section/shipping-rate-section.js#L31-L35), that needs at least 1 parsable rate for the option to show.

Due to the incomplete implementation described above, the meaning of `shipping_rates_count` is also somewhat ambiguous at the moment, and it doesn't satisfy its claim that:

- https://github.com/woocommerce/google-listings-and-ads/blob/733b4668607bae030a5173aa2094ed9a927ecd69/src/API/Site/Controllers/MerchantCenter/SettingsController.php#L157-L162
- https://github.com/woocommerce/google-listings-and-ads/blob/733b4668607bae030a5173aa2094ed9a927ecd69/src/Shipping/ShippingZone.php#L81-L85

"The number of countries that have partially parsable WooCommerce Shipping zones" is probably closer to its current meaning, but that also makes it difficult to decide:
- Whether incompatible settings should also be considered "unparsable"
- How to use the ambiguous `shipping_rates_count` value

For example, when the selected audience country is Japan in G4W and WooCommerce Shipping Zones have only one parsable and compatible setting but it specifies to the US, the `shipping_rates_count` will still be 1, then how can this value be used to make further determinations?

From a technical point of view, I would think that for the `ShippingZone` and `ZoneMethodsParser` classes, negative values are considered parable. Determining whether the parsed shipping zones are compatible with GMC could be done by other related classes that use these classes.

> We'll still want to eventually refine the "steps to complete after onboarding" part, but that's out of scope for this PR.

I guess a solution for a better experience would be to have an API that can at least return whether the current WooCommerce Shipping Zones settings are compatible and cover all the given countries, and if not, what the reasons might be. The UI would even need a mechanism like the store address to inform the user and allow them to get the results of the API again after updating WooCommerce Shipping Zones.

If so, it indeed requires more effort and is beyond the scope of this PR and Shipping improvements phase 3.

mikkamp (approved):
Thanks for the additional clarifications. As discussed, it should be a rare scenario when a merchant has negative shipping rates setup, so it won't be very frequent that new merchants going through onboarding will be blocked by this error.

So in that case I'll go ahead and approve this PR. The rest which is out of scope can be revisited when we introduce better post onboarding handling.

PR #2763, opened 2025-01-13
Remove unused Http proxy class

mikkamp:

The Http proxy class was initially added in PR #69 with the comment:
> A Proxy for making HTTP requests to the connect server in the future.

In the end we decided to go with a Guzzle client to send requests to the connect server. So in the interest of reducing code this PR removes this class. If we change our mind this PR can be reverted to bring it back.

1. Run some basic smoke tests to confirm API requests continue to work.
2. Run unit tests to confirm they all pass.

### Changelog entry
* Tweak - Remove unused Http proxy class.


puntope (approved):
Thanks @mikkamp LGTM ✅ 

- Tested there is no more `Proxies\Http` usage. Now I can only see Guzzle\Http and Psr\Http
- Tested a full onboarding
- Tested updating Shipping settings
- Checked for fatals or warnings in logs

PR #2764, opened 2025-01-13
Refactor installer class to load dependencies when needed

mikkamp:

This PR makes 2 main refactors to the Installer class.

##### 1. Remove validation
In the constructor we were explicitly passing all classes which implement `InstallableInterface` and then validating to check if they implemented the same interface. Since the container is responsible for passing the right dependencies we don't need to do another loop through those classes to validate them.

##### 2. Load dependencies when needed
Passing the dependencies in the constructor means it will call the constructor of all the dependencies, even when we don't use them. Since the Install class relies on a hook being called and some other additional conditions, we can retrieve the necessary classes from the container only when they are needed. This is a minor optimization in how the extension constructs itself, but together with other optimizations it will make a difference.

If we look at the before and after of a front end request we can see that it reduces the time spent resolving container dependencies.

Before:
image

After:
image

Since we changed the load order this was initially generating a fatal error:
```
PHP Fatal error:  Uncaught TypeError: Automattic\WooCommerce\GoogleListingsAndAds\Product\Attributes\AttributeManager::__construct(): Argument #1 ($attribute_mapping_rules_query) must be of type Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\AttributeMappingRulesQuery, string given in google-listings-and-ads/src/Product/Attributes/AttributeManager.php:69
```

This was because the `DBServiceProvider` wasn't registered as the provider of this class, so it wasn't resolving when loading in a different order.

1. Use the snippet `add_filter( 'woocommerce_gla_force_run_install', '__return_true' );`
2. Remove the option `gla_db_version`
3. Remove the option `gla_install_version`
4. Refresh an admin page and confirm the options are re-added
5. Remove one of the DB tables like `wp_gla_budget_recommendations`
6. Refresh an admin page and confirm it's added again

### Changelog entry
* Tweak - Refactor installer class to load dependencies when needed.

### DIFF
--- a/src/Installer.php
+++ b/src/Installer.php
@@ -5,6 +5,8 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Registerable;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
+use Automattic\WooCommerce\GoogleListingsAndAds\Internal\ContainerAwareTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Internal\Interfaces\ContainerAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Internal\Interfaces\FirstInstallInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Internal\Interfaces\InstallableInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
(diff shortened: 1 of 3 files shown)


puntope (approved):
LGTM ✅  Thanks for the refactor

THe options are recreated correctly:
Screenshot 2025-01-13 at 19 39 41

I also tested Attribute Mapping feature and works as expected.

Notice it's not necessary to add the snippet. Because if we delete gla_db_version then the installer and first_installer will trigger as the first check is true https://github.com/woocommerce/google-listings-and-ads/pull/2764/files#diff-67306479da293f095927542c4cff1a45875138eb52372caaba982b07c053285bL82 .

PR #2771, opened 2025-01-16
Set specific flags for html_entity_decode

### LINKED ISSUE #2770: PHP Compatibility warnings from QIT code scan
### Issue
The QIT compatibility test produces two warnings:

Image

We should fix this by explicitly setting the `$flags` parameter as mentioned in the suggestion.

mikkamp:

This PR resolves the following warning notices:
> The default value of the $flags parameter for html_entity_decode() was changed from ENT_COMPAT to ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 in PHP 8.1. For cross-version compatibility, the $flags parameter should be explicitly set.

We only use the function `html_entity_decode` in two locations:
1. To decode currency symbols (generally we don't expect this to contain quotes, so the behaviour should remain the same regardless of PHP version)
2. For product names that show up in the issues table. We are already escaping these strings before inserting in the DB as well as decoding before displaying, so the change doesn't affect any behaviour

I was considering adding a unit test, but because the behaviour is identical, and what we are really concerned about is that the end display shows the product names without encoded entities, so instead I went with a manual visual test to confirm the behaviour remains consistent.

Closes #2770 

1. Add a product with a name containing single and double quotes, ex: `Product "name" which shouldn't cause problems`
4. Make sure to leave some requirements out like adding a description or image so it ends up in the product issues table
5. Go to the connection test page and "Clear Status Cache"
6. Go to the Product Feed page and wait for the statuses to refresh
7. Check the DB table `wp_gla_merchant_issues` and confirm the product name column contains the quote characters as expected
8. View the product issues table on the product feed page and confirm the name contains the quotes as expected (can drop some rows from the DB table if there are too many to display)

image

### Changelog entry
* Tweak - Set specific flags for html_entity_decode.

### DIFF
--- a/src/Ads/AccountService.php
+++ b/src/Ads/AccountService.php
@@ -78,7 +78,7 @@ class AccountService implements ContainerAwareInterface, OptionsAwareInterface, 
 		$status = [
 			'id'       => $id,
 			'currency' => $this->options->get( OptionsInterface::ADS_ACCOUNT_CURRENCY ),
-			'symbol'   => html_entity_decode( get_woocommerce_currency_symbol( $this->options->get( OptionsInterface::ADS_ACCOUNT_CURRENCY ) ) ),
+			'symbol'   => html_entity_decode( get_woocommerce_currency_symbol( $this->options->get( OptionsInterface::ADS_ACCOUNT_CURRENCY ) ), ENT_QUOTES ),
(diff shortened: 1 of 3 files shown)


eason9487 (approved):
Thanks for the fix! The code works as expected. LGTM.

PR #2772, opened 2025-01-16
Resolve some of the plugin check errors and warnings

mikkamp:

This PR resolves a substantial amount of the errors and warnings which are identified in the [WordPress.org plugin check](https://wordpress.org/plugins/plugin-check/).

Changes resolve the following errors:
- Use of heredoc and nowdoc syntax ("<<<") is not allowed; use standard strings or inline HTML instead
- json_encode() is discouraged. Use wp_json_encode() instead.
- rand() or mt_rand() is discouraged. Use the far less predictable wp_rand() instead.
- Missing "License" in Plugin Header. Please update your Plugin Header with a valid GPLv2 (or later) compatible license.

> [!Note]
> We are ignoring the following errors/warnings:
> - `WordPress.Security.EscapeOutput.ExceptionNotEscaped` - Exceptions can be escaped late when output, not when they are thrown, we have a specific exclusion for this in the phpcs.xml file
> - `trademarked_term` Plugin naming has been done in discussion with trademark owners
> - Caching in uninstall.php - only gets run once when uninstalling, we have a specific exclusion for this in the phpcs.xml file
>
> We can address these at a later date if these rules ever become fully required.

1. Build a zip file from this branch `composer install && nvm use && npm run build`
2. Install the zip file on a test site
3. Install [plugin check](https://wordpress.org/plugins/plugin-check/)
4. Go to Tools > Plugin Check and run the check for the selected plugin
5. Confirm the amount of errors is now reduced there should be about 50 less down to @ 250
6. Check for any errors left if they are in the ignore list above

#### Linting php
1. Run `npm run lint:php`
2. Run `npm run lint:php-tests`
3. Confirm we don't get any errors with the revised linting rules

### Changelog entry
* Tweak - Resolve some of the plugin check errors and warnings.

### DIFF
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -457,7 +457,7 @@ function update_autoloads( string $file, array $composer_autoload, array $packag
 	if ( $modified ) {
 		file_put_contents(
 			$file,
-			wp_json_encode( $json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE )
+			json_encode( $json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE )
 		);
 	}
 }
--- a/bin/prefix-vendor-namespace.php
+++ b/bin/prefix-vendor-namespace.php
@@ -457,7 +457,7 @@ function update_autoloads( string $file, array $composer_autoload, array $packag
 	if ( $modified ) {
 		file_put_contents(
 			$file,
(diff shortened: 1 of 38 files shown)


eason9487 (approved):
Thanks for resolving these plugin check errors. Confirmed the listed four categories of errors have been resolved. LGTM.

I'm guessing these two errors will also be ignored for now as they are test files?

- `WordPress.Security.EscapeOutput.OutputNotEscaped`
- `PluginCheck.CodeAnalysis.Localhost.Found`

image

mikkamp:
> I'm guessing these two errors will also be ignored for now as they are test files?

Yes. They also don't show up in the report when you test a release version of the plugin, as we don't bundle the tests folder.

PR #2773, opened 2025-01-17
Allow mapping of a Google ID without a prefix

### LINKED ISSUE #2768: When the product ID prefix is removed statuses always remain unsynced
### Describe the bug:
In the documentation we provide a snippet to remove the `gla_` prefix when syncing to Merchant Center: https://woocommerce.com/document/google-listings-and-ads/faqs/#how-do-i-remove-the-gla-prefix-from-synced-products

However that snippet by itself only works for syncing from WooCommerce > Google MC. When we fetch statuses from Google MC > WooCommerce it doesn't automatically convert the ID without prefix back. We can get around this by using a second snippet, like:

```php
// Sync products without GLA prefix.
add_filter(
	'woocommerce_gla_get_google_product_offer_id',
	function ( $mc_id, $product_id ) {
		return (string) $product_id;
	},
	10,
	2
);

// Map google ID to WC ID.
add_filter(
	'woocommerce_gla_get_wc_product_id',
	function ( $wc_product_id, $mc_product_id ) {
		return (int) $mc_product_id;
	},
	10,
	2
);
```

One solution would be to update the docs, but since it's common we can also make the prefix optional. If it's just a number then we should be able to do the mapping within the extension. This means the following code needs to be updated:

https://github.com/woocommerce/google-listings-and-ads/blob/2.9.5/src/Product/ProductHelper.php#L314

Notice there that the prefix is required otherwise it keeps `$wc_product_id` as 0.

### Steps to reproduce:
1. Add the snippet from the documentation: https://woocommerce.com/document/google-listings-and-ads/faqs/#how-do-i-remove-the-gla-prefix-from-synced-products
2. Setup the extension to sync some products to Merchant Center
3. Observe the Product Feed page and notice that all products remain with the status unsynced
4. Use the Connection Test helper to clear the product statuses 
5. Even after waiting till everything is synced (takes a while for Google to return the synced products in the API), the status doesn't change
6. Use the second snippet above and confirm that after refreshing the statuses we do now get some products with a different status

mikkamp:

This PR makes the prefix when fetching status from the Merchant Center as optional. We do want to make sure that the ID is a strictly numeric ID otherwise it's not accepted. In the previous regex we searched till end of string with `$`, but didn't force it to search from the start of the string with `^`. This meant that it was possible to a map an ID like `foo_gla_12345`. We fixed this in the PR to ensure it scans the whole string for a matching pattern:

- `gla_` + numeric ID
- numeric ID

Any other patterns will result in an unmapped ID. I've added additional unit tests to cover all these variations.

Closes #2768 

1. Add the snippet from the documentation: https://woocommerce.com/document/google-listings-and-ads/faqs/#how-do-i-remove-the-gla-prefix-from-synced-products
2. Setup the extension to sync some products to Merchant Center
3. Use the Connection Test helper to clear the product statuses 
4. Refresh the product feed page
5. Wait for the statuses to reload and confirm that we have products that are a different status then `unsynced`

### Changelog entry
* Tweak - Allow mapping of a Google ID without a prefix.

### DIFF
--- a/src/Product/ProductHelper.php
+++ b/src/Product/ProductHelper.php
@@ -310,10 +310,11 @@ class ProductHelper implements Service, HelperNotificationInterface {
 		$mc_product_id_tokens = explode( ':', $mc_product_id );
 		$mc_product_id        = end( $mc_product_id_tokens );
 
+		// Support a fully numeric ID both with and without the `gla_` prefix.
 		$wc_product_id = 0;
-		$pattern       = '/' . preg_quote( $this->get_slug(), '/' ) . '_(\d+)$/';
+		$pattern       = '/^(' . preg_quote( $this->get_slug(), '/' ) . '_)?(\d+)$/';
 		if ( preg_match( $pattern, $mc_product_id, $matches ) ) {
-			$wc_product_id = (int) $matches[1];
+			$wc_product_id = (int) $matches[2];
 		}
 
 		/**
--- a/tests/Unit/Product/ProductHelperTest.php
+++ b/tests/Unit/Product/ProductHelperTest.php
@@ -487,6 +487,31 @@ class ProductHelperTest extends ContainerAwareUnitTest {
 		$this->assertEquals( 1234567, $product_id );
 	}
 
+	/**
+	 * Confirm we support converting a product ID without the `gla_` prefix.
+	 */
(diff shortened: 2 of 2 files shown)


ianlin (approved):
Thanks for the fix. Confirmed the Google ID without a prefix can be synced to GMC successfully.

PR #2774, opened 2025-01-17
Adjust some query and table functions

### LINKED ISSUE #500: Revisit some Query and Table issues
## Technical
As discussed in the  comments of #463, there are a few things to consider:

- Is the `GROUP BY` clause necessary in `Query::build_query()`? https://github.com/woocommerce/google-listings-and-ads/pull/463#discussion_r614673144
- Should `table::get_sql_safe_name()` be `public`? https://github.com/woocommerce/google-listings-and-ads/pull/463#discussion_r616031700

Additionally, this minor issue came up (doesn't currently seem to be affecting anything, but it could):
- `Query::get_row()` returns the first result returned by a query, by momentarily setting `$this->limit = 1` and executing `query_results()`. However, `query_results()` stores the results of the query in the `$this->query` property, which is also checked by `get_results()`. If `get_results()` were to be called after `get_row()`, it would (possibly) incorrectly only return the first result.
  https://github.com/woocommerce/google-listings-and-ads/blob/9873b7fc4b96a01f846965473814e823f2a708ba/src/DB/Query.php#L184-L193
  Maybe this should be more like `count_results()`, where a separate query with a limit of 1 is generated with `build_query` and stored if necessary.

This is an issue to revisit/discuss these issues and resolve them as necessary.

mikkamp:

As a result of the discussion in issue #500 there were three things to address:

1. Remove the GROUP BY (we already have a DB constraint handling this, if we didn't it would alter the count query results)
This was completed in this PR

2. Use esc_like when checking for an existing table
This was completed in this PR

3. Remove get_sql_safe_name and make sure we are placing the table name in backquotes in all locations
I decided to leave this in place for now as it isn't causing any harm. Ideally in the future we'd use `wpdb->prepare` for escaping the table names: https://make.wordpress.org/core/2022/10/08/escaping-table-and-field-names-with-wpdbprepare-in-wordpress-6-1/
However that's only available in WordPress 6.2, we would have to bump the minimum version for that.

Closes #500

1. Add a snippet to log DB queries:
```php
// Log queries from gla.
add_filter(
	'query',
	function ( string $query ) {
		global $wpdb;

		if ( str_contains( strtolower( $query ), "{$wpdb->prefix}gla_" ) ) {
			wc_get_logger()->log(
				WC_Log_Levels::DEBUG,
				$query,
				[
					'source' => 'gla-db-queries',
				]
			);
		}

		return $query;
	}
);
```
2. Go to the Product Feed page to load issues from the table
3. Install this PR and reload the Product Feed page
4. Go to WooCommerce > Status > Logs and find a log file for `gla-db-queries`
5. Confirm we initially have the DB query with a GROUP BY:
```
SELECT * FROM `wp_gla_merchant_issues`
WHERE type = 'account' AND
severity IN ('error','critical')
GROUP BY `wp_gla_merchant_issues`.`id`
ORDER BY `type` ASC, `product` ASC, `issue` ASC
```
6. Confirm the DB query no longer has the GROUP BY after switching to this PR:
```
SELECT * FROM `wp_gla_merchant_issues`
WHERE type = 'account' AND
severity IN ('error','critical')
GROUP BY `wp_gla_merchant_issues`.`id`
```

The second change is harder to test because we don't use invalid characters in table names. But we can confirm tables are recreated after checking if they exist:
1. Drop one of the tables like `wp_gla_budget_recommendations`
2. Delete options:
    - `wp option delete gla_db_version`
    - `wp option delete gla_file_version`
3. Refresh one of the WP Admin pages
4. Confirm the database table has been recreated and populated

### Changelog entry
* Tweak - Adjust some query and table functions.

### DIFF
--- a/src/DB/Query.php
+++ b/src/DB/Query.php
@@ -318,10 +318,6 @@ abstract class Query implements QueryInterface {
 		}
 
 		if ( ! $get_count ) {
-			if ( empty( $this->groupby ) ) {
-				$pieces[] = "GROUP BY `{$this->table->get_name()}`.`{$this->table->get_primary_column()}`";
-			}
-
 			if ( $this->orderby ) {
 				$pieces[] = 'ORDER BY ' . implode( ', ', $this->orderby );
 			}
--- a/src/DB/Table.php
+++ b/src/DB/Table.php
@@ -52,7 +52,7 @@ abstract class Table implements TableInterface {
 	 */
 	public function exists(): bool {
 		$result = $this->wpdb->get_var(
-			"SHOW TABLES LIKE '{$this->get_sql_safe_name()}'" // phpcs:ignore WordPress.DB.PreparedSQL
+			"SHOW TABLES LIKE '{$this->wpdb->esc_like( $this->get_name() )}'" // phpcs:ignore WordPress.DB.PreparedSQL
 		);
 
 		return $result === $this->get_name();


ianlin (approved):
Thanks for the update. Confirmed the query did not contain GROUP BY and the deleted table was recreated successfully.

PR #2788, opened 2025-02-10
[API PULL] Backbone for SYNC endpoint

puntope:

This is a preparatory PR for the work related to creating a WPCOM Proxied endpoint allowing to GET and UPDATE the enabled Sync mode.

In this PR we create a backbone for the endpoint getting a default value or the value saved in `wp_options.gla_api_pull_sync_mode` 

PT: pcTzPl-2yG-p2

### Screenshots:

Screenshot 2025-02-10 at 23 14 10

1. Request `GET wp-json/wc/gla/sync`
2. See you get the default response

```
'products' => [
			'pull' => false,
			'push' => false,
		],
		'coupons'  => [
			'pull' => false,
			'push' => false,
		],
		'shipping' => [
			'pull' => false,
			'push' => false,
		],
		'settings' => [
			'pull' => false,
			'push' => false,
		],
```
3. Update `wp_options.gla_api_pull_sync_mode` and save some data for example this with all true:
`a:4:{s:8:"products";a:2:{s:4:"push";b:1;s:4:"pull";b:1;}s:7:"coupons";a:2:{s:4:"push";b:1;s:4:"pull";b:1;}s:8:"shipping";a:2:{s:4:"push";b:1;s:4:"pull";b:1;}s:8:"settings";a:2:{s:4:"push";b:1;s:4:"pull";b:1;}}` 
4. Perform the request in step 1 again. And see the expected result

### Changelog entry

>

### DIFF
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -21,12 +21,24 @@ defined( 'ABSPATH' ) || exit;
  */
 class SyncController extends BaseOptionsController {
 
-	private const DEFAULT_SYNC_MODE = array(
-		'products' => array( 'pull' => false, 'push' => false ),
-		'coupons' => array( 'pull' => false, 'push' => false ),
-		'shipping' => array( 'pull' => false, 'push' => false ),
-		'settings' => array( 'pull' => false, 'push' => false ),
-	);
+	private const DEFAULT_SYNC_MODE = [
+		'products' => [
+			'pull' => false,
+			'push' => false,
+		],
+		'coupons'  => [
+			'pull' => false,
+			'push' => false,
+		],
+		'shipping' => [
+			'pull' => false,
+			'push' => false,
+		],
+		'settings' => [
+			'pull' => false,
+			'push' => false,
+		],
+	];
 
 	/**
 	 * The base for routes in this controller.
@@ -38,7 +50,7 @@ class SyncController extends BaseOptionsController {
 	/**
 	 * SyncController constructor.
 	 *
-	 * @param RESTServer     $server
+	 * @param RESTServer $server
 	 */
 	public function __construct( RESTServer $server ) {
 		parent::__construct( $server );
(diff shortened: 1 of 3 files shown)


src/API/Site/Controllers/RestAPI/SyncController.php
```diff
@@ -0,0 +1,147 @@
...
+	 * Get the callback function for getting the current sync mode for API PULL.
+	 *
+	 * @return callable
+	 */
+	protected function get_sync_callback(): callable {
+		return function ( Request $request ) {
+			try {
+				$sync_mode = $this->options->get( OptionsInterface::API_PULL_SYNC_MODE );
+				if ( ! $sync_mode ) {
+					$sync_mode = self::DEFAULT_SYNC_MODE;
```
mikkamp: Should we be using a check here to ensure the option is an array?

Also isn't it better to always merge with the defaults, with something like `wp_parse_args`. Just in preparation if we would ever add an entry to the defaults it would at least have default values.

Same would apply when we write the endpoint for updating the values.

```suggestion
				if ( ! is_array( $sync_mode ) ) {
					$sync_mode = [];
				}
				
				$sync_mode = wp_parse_args( $sync_mode, self::DEFAULT_SYNC_MODE );
```

src/API/Site/Controllers/RestAPI/SyncController.php
```diff
@@ -0,0 +1,147 @@
...
+	 * Registers the routes.
+	 */
+	public function register_routes() {
+		$this->register_route(
+			$this->route_base,
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_sync_callback(),
+					'permission_callback' => $this->get_permission_callback(),
```
mikkamp, line 69: Considering that we only want these endpoints accessible by the API Pull implementation. In the WC endpoints we check for the `gla_syncable` parameter being set to 1 in the request before we modify the responses.

Should we have a custom permissions callback to also check if this is set? Not very difficult to get around, but it might provide better separation.

mikkamp (commented):
Thanks for the changes, they are working as expected and I'm getting the response either from the defaults or the option.

I just left a couple comments, let me know if you think that would be worthwhile to do.

src/API/Site/Controllers/RestAPI/SyncController.php
puntope: Thanks I will update that in future PRs if that's ok

mikkamp (approved):
Thanks for the changes. I retested and it's return a hybrid of the option and defaults. So everything is as expected:

```json
{
    "products": {
        "push": true,
        "pull": true
    },
    "coupons": {
        "push": false,
        "pull": false
    },
    "shipping": {
        "push": true,
        "pull": true
    },
    "settings": {
        "pull": false,
        "push": false
    }
}
```

PR #2790, opened 2025-02-12
Opt-in to URL expansion

mikkamp:
Final URL expansion is a feature in Performance Max campaigns that helps improve campaign performance by automatically selecting the most relevant landing page from merchants’ websites based on the user's search query and intent.

For simplicity, if `url_expansion_opt_out` is set to true, the ads will only use the URLs of product descriptions. This means, ads will take users only to the product detail page.

Therefore, Google's best practice says that we should set it to false always so AI can use the entire online store of merchants to create ads that generate sales.

We are only opting in for newly created campaigns. Campaigns which have already been created will be left as is (users can still toggle this setting in the Google Ads dashboard if required).

1. Create a new campaign through the extension 
2. Check the newly created campaign in https://ads.google.com/
3. Click on the created campaign and go to the tab Campaign Settings
4. In the section "Automatically created assets" make sure Final URL expansion is enabled
image
5. Compare this to a previously created campaign (or create one with the latest release) and confirm that Final URL expansion wasn't enabled
image

### Changelog entry
* Tweak - Opt-in to URL expansion.

### DIFF
--- a/src/API/Google/AdsCampaign.php
+++ b/src/API/Google/AdsCampaign.php
@@ -462,7 +462,7 @@ class AdsCampaign implements ContainerAwareInterface, OptionsAwareInterface {
 				'status'                    => CampaignStatus::number( 'enabled' ),
 				'campaign_budget'           => $this->budget->temporary_resource_name(),
 				'maximize_conversion_value' => new MaximizeConversionValue(),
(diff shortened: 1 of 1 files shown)


eason9487 (approved):
Thanks for the work! Confirmed it worked as expected. LGTM.

image

PR #2792, opened 2025-02-12
Update ubuntu image to 22.04 for QIT WF

puntope:

PT: p7H4VZ-5d2-p2#comment-12564

In this PR we update the image to use in QIT Workflows to `ubuntu-latest`

### Screenshots:

Screenshot 2025-02-12 at 20 17 55

1. See the QIT running in the new branch https://github.com/woocommerce/google-listings-and-ads/actions/runs/13306921922 using the last ubuntu version
2. Perform some QIT jobs against this branch and verify is running 

### Changelog entry

>

### DIFF
--- a/.github/workflows/run-qit.yml
+++ b/.github/workflows/run-qit.yml
@@ -54,7 +54,7 @@ jobs:
     secrets: inherit
   qit-tests:
     name: Run QIT Tests
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     needs: build
     steps:
       - name: Download artifact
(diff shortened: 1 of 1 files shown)


mikkamp (approved):
Looks good to me. I can confirm it's running with the latest version:

image

It's not clear where we need to actually see the results of the tests, based on the annotations it looks like only the activation test succeeded, the rest are scheduled, but I'm not seeing any links to where we would see the results. Do we need to log into a QIT dashboard to view those?

image

Either way that doesn't affect the workflow running on a newer Ubuntu version, so I'll approve the PR.

puntope:
Yes @mikkamp The links are in Dashboard or the CLI. This is how its working also in the previous version. Seems like the WF only provides the number of the QIT report

PR #2795, opened 2025-02-14
Fetch budget recommendations from the Google Ads API

mikkamp:

Initial implementation for fetching Budget Recommendations from the Ads API.

This covers both getting recommendations from the Google Ads API, as well as the pre-existing fallback from the database table. Note that if the recommendations from the Ads API are lower they will be replaced by the fallback. This only happens when the recommended budget is lower than the fallback budget. So we can end up with a situation where the high recommendation is still included with the fallback. Or where the low option is still included if the recommended rate hasn't been swapped.

Caching was also implemented since it can take a while occasionally to get back results from the API.
1. Caching of country locations do not have an expiry time (as country ID's shouldn't change)
2. Caching of the recommended rates returned from the API are retained for 12 hours

We still need to implement fetching metrics for the fallback rate. But since that is similar to fetching metrics for a custom budget I'm leaving that out for this PR and the fallback recommendation will be returned without metrics.

Unit testing was added to fully cover both the `BudgetRecommendations` fetching from the Ads API as well as the `BudgetRecommendationController`.

1. Reinstall composer dependencies `rm -rf vendor && composer install` 
4. Test API request `https://domain.test/wp-json/wc/gla/ads/campaigns/budget-recommendation?country_codes[]=IE`
5. Observe multiple recommendation results being returned:
```json
{
    "currency": "EUR",
    "recommendations": [
        {
            "daily_budget": 12,
            "metrics": {
                "cost": 84,
                "conversions": 6.1,
                "conversions_value": 243.88968735282847
            },
            "country": "IE",
            "level": "Recommended"
        },
        {
            "daily_budget": 14.399999,
            "metrics": {
                "cost": 100.799993,
                "conversions": 6.5,
                "conversions_value": 258.7239827235049
            },
            "country": "IE",
            "level": "High"
        },
        {
            "daily_budget": 9.6,
            "metrics": {
                "cost": 67.2,
                "conversions": 5.7,
                "conversions_value": 226.13979597572018
            },
            "country": "IE",
            "level": "Low"
        }
    ]
}
```
6. They should be ordered in the following order: recommended, high, low , fallback (the current UI will always take the first recommendation)
7. Adjust the fallback rate in the table `wp_gla_budget_recommendations` so it is higher than all the returned rates (ex: 20)
8. Repeat the request and expect to see only the fallback rate (still without metrics):
```json
{
    "currency": "EUR",
    "recommendations": [
        {
            "daily_budget": 20,
            "country": "IE",
            "level": "Recommended"
        }
    ]
}
```
9. Adjust the fallback rate in the table `wp_gla_budget_recommendations` so it is higher than the recommended amount but still lower than the high recommendation (ex: 13)
10. Repeat the request and expect to see a mix of fallback and API rates:
```json
{
    "currency": "EUR",
    "recommendations": [
        {
            "daily_budget": 13,
            "country": "IE",
            "level": "Recommended"
        },
        {
            "daily_budget": 14.399999,
            "metrics": {
                "cost": 100.799993,
                "conversions": 6.5,
                "conversions_value": 258.7239827235049
            },
            "country": "IE",
            "level": "High"
        }
    ]
}
```
11. Adjust the fallback rate in the table `wp_gla_budget_recommendations` so it is lower than the recommended amount but still higher than the lowest option (ex: 11)
12. Repeat the request and expect to see the full result returned from the API:
```json
{
    "currency": "EUR",
    "recommendations": [
        {
            "daily_budget": 12,
            "metrics": {
                "cost": 84,
                "conversions": 6.1,
                "conversions_value": 243.88968735282847
            },
            "country": "IE",
            "level": "Recommended"
        },
        {
            "daily_budget": 14.399999,
            "metrics": {
                "cost": 100.799993,
                "conversions": 6.5,
                "conversions_value": 258.7239827235049
            },
            "country": "IE",
            "level": "High"
        },
        {
            "daily_budget": 9.6,
            "metrics": {
                "cost": 67.2,
                "conversions": 5.7,
                "conversions_value": 226.13979597572018
            },
            "country": "IE",
            "level": "Low"
        }
    ]
}
```

> [!Note]
> Initially I was always getting lower rates than the fallback, but today I tried with a new account after creating the first campaign. It's now returning higher rates. If your fallback rate is always higher than this can be adjusted for the right country/currency in the table `wp_gla_budget_recommendations` which allows different combinations to be tested.
> While we handle fallback rates as floats, the DB table still has the type as `int(11)` so we are unable to enter a decimal value in the DB table.

### Additional details:
Project Thread: pcTzPl-2Hb-p2

### Changelog entry
* Add - Fetch budget recommendations from the Google Ads API.

### DIFF
--- a/src/API/Site/Controllers/Ads/BudgetRecommendationController.php
+++ b/src/API/Site/Controllers/Ads/BudgetRecommendationController.php
@@ -115,8 +115,8 @@ class BudgetRecommendationController extends BaseController implements Container
 			if ( $fallback_recommendations ) {
 				$fallback_budget = $fallback_recommendations[0]['daily_budget'] ?? 0;
 
-				// Swap recommended if fallback is higher.
-				if ( ! empty( $recommendations[0]['daily_budget'] ) && $recommendations[0]['daily_budget'] < $fallback_budget ) {
+				// Swap recommended if not set or if fallback is higher.
+				if ( empty( $recommendations[0] ) || ( ! empty( $recommendations[0]['daily_budget'] ) && $recommendations[0]['daily_budget'] < $fallback_budget ) ) {
 					$recommendations[0] = $fallback_recommendations[0];
 
 					// Remove high recommendation if fallback is higher.
--- a/src/API/Site/Controllers/Ads/BudgetRecommendationController.php
+++ b/src/API/Site/Controllers/Ads/BudgetRecommendationController.php
@@ -104,11 +104,7 @@ class BudgetRecommendationController extends BaseController implements Container
 				);
 			}
 
-			$query           = $this->container->get( BudgetRecommendationQuery::class );
-			$recommendations = $query
-				->where( 'country', $country_codes, 'IN' )
-				->where( 'currency', $currency )
-				->get_results();
+			$recommendations = $this->get_fallback_recommendations( $country_codes, $currency );
 
 			if ( ! $recommendations ) {
 				return new Response(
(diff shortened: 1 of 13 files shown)


src/API/Site/Controllers/Ads/BudgetRecommendationController.php
```diff
@@ -122,26 +142,48 @@ protected function get_budget_recommendation_callback(): callable {
...
+	 *
+	 * @return array|null Recommendation for each included country.
+	 */
+	protected function get_fallback_recommendations( array $country_codes, string $currency ): ?array {
+		$query           = $this->container->get( BudgetRecommendationQuery::class );
+		$primary_country = reset( $country_codes );
+		$recommendations = $query
+			->where( 'country', $primary_country )
+			->where( 'currency', $currency )
+			->get_results();
```
eason9487: The method doc doesn't seem to match the implementation as it only returns the recommendation of the first country code.

eason9487 (commented):
Thanks for the work! The implementation looks good.

However, I would like to clarify that the budget recommendations API in this PR accepts multiple country codes but always only returns the query results for the first country code. Does this mean that instead of querying by all the countries the merchant has chosen to sell in, this new feature will change to query by the store's primary country (where the store is located?) instead?

mikkamp:
> However, I would like to clarify that the budget recommendations API in this PR accepts multiple country codes but always only returns the query results for the first country code. Does this mean that instead of querying by all the countries the merchant has chosen to sell in, this new feature will change to query by the store's primary country (where the store is located?) instead?

It will continue to take multiple countries as input, but the recommendations are attributed to all the countries as a whole. Right now the API assigns them to the first country (which I understand is the primary country), mainly just to remain compatible with the current UI. It can also continue to be used to populate the country for `Most merchants targeting <strong>%3$s set a daily budget of %1$f %2$s</strong>`. Although the `isMultiple` will need to be determined based on the countries passed not the amount of recommendations returned.

And since we are no longer returning just one recommendations the code for counting the highestBudget will also need to be altered.

src/API/Site/Controllers/Ads/BudgetRecommendationController.php
mikkamp: Good point, I'll adjust that.

eason9487 (approved):
Thanks for the clarification and update! Testing well. LGTM.

PR #2796, opened 2025-02-18
[API PULL] POST method for SYNC endpoint

puntope:

In this PR we create POST method for the SYNC endpoint `/wp-json/wc/gla/sync` and save it inside `wp_options.gla_api_pull_sync_mode`

PT: pcTzPl-2yG-p2

### Screenshots:

Screenshot 2025-02-18 at 11 48 32

1. Request POST wp-json/wc/gla/sync with some valid values and some random values (a mix with good values and wrong keys and wrong values) 
2. Verify that only the valid keys and values are updated in the options table. Everything is optional. Invalid or missing datatypes or methods won't be updated.  

Examples

1. Updates all data types
```
'products' => [
			'pull' => bool,
			'push' => bool,
		],
		'coupons'  => [
			'pull' => bool,
			'push' => bool,
		],
		'shipping' => [
			'pull' => bool,
			'push' => bool,
		],
		'settings' => [
			'pull' => bool,
			'push' => bool,
		],
```

2. Updates products data type
```
'products' => [
			'pull' => bool,
			'push' => bool,
		],
]
```

2. Updates products pull method

```
'products' => [
			'pull' => bool
]		],
```

### Additional details:

Didn't add yet `gla_syncable` as a required param because anyone can really add it as a param in the request to bypass the check ...  so it doesn't make much difference to use it as a protection layer. Also even if we protect the API anyone can just update the options table for changing the modes. So for now I just keep the auth access as `can_manage` as we do for other endpoints.

### Changelog entry

>

### DIFF
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -199,8 +199,14 @@ class SyncController extends BaseOptionsController {
 	 */
 	protected function get_request_params( Request $request ): array {
 		$request_params = $request->get_json_params();
-		$params         = array_intersect_key( $request_params, self::DEFAULT_SYNC_MODE );
-		$valid_params   = [];
+
+		// JSON Data
+		if ( is_null( $request_params ) ) {
+			$request_params = json_decode( $request->get_body(), true );
+		}
+
+		$params       = array_intersect_key( $request_params, self::DEFAULT_SYNC_MODE );
+		$valid_params = [];
 
 		foreach ( $params as $key => $param ) {
 			if ( isset( $param['push'] ) && isset( $param['pull'] ) && is_bool( $param['push'] ) && is_bool( $param['pull'] ) ) {
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -15,6 +15,46 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Class SyncController
  *
+ * GET wp-json/wc/gla/sync
+ * POST wp-json/wc/gla/sync
+ * Body Params: @see get_schema_properties
+ *
+ * We support 4 data types Products, Coupons, Shipping and Settings.
+ * Each of this data types contains pull and push methods.
+ *
+ * Push: Legacy method pushing store data to Google Merchant Center using Google's API
+ * Pull: New method where Google fetches the data from the store.
+ *
+ * All the data types and methods are optional. When set, they update the current config.
+ *
+ * Examples
+ *
+ * POST wp-json/wc/gla/sync
+ *
+ * {
+ *     "products": {
+ *          "pull": true,
+ *          "push": false,
+ *     },
+ *     "settings": {
+ *         "pull": true,
+ *         "push": false,
+ *     }
+ * }
+ *
+ * Updates Only Products and Settings to enable Pull and disable Push. Rest of the data types will remain unchanged.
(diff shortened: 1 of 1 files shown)


src/API/Site/Controllers/RestAPI/SyncController.php
```diff
@@ -146,4 +166,57 @@ private function get_pull_push_schema_fields() {
...
+	 * @param Request $request The request
+	 * @return array
+	 */
+	protected function get_request_params( Request $request ): array {
+		$request_params = $request->get_json_params();
+
+		// JSON Data
+		if ( is_null( $request_params ) ) {
+			$request_params = json_decode( $request->get_body(), true );
+		}
```
ianlin: From my local test the `$request->get_json_params()` will be `null` if the content type is not `application/json`. For `$request->get_body()`  it accepts multiple content types. In this case not sure if we can just always use `get_body()`?

Also from the local test if I don't provide body and the 500 will return.

ianlin (commented):
Thanks for adding the POST sync endpoint in the plugin. The behaviour works as expected in my local. I left a question regarding request body parsing.

mikkamp:
I had another question to clarify about this PR.

Taking a quick look at the code it seems that I can update a single group at a time. Which will cause the updated group to be merged with the current sync modes. However I'm not seeing that clarified in the testing instructions or reflected in the schema. Can we document a little more on how this endpoint works, either in the schema so API self discovery can be used, or somewhere else to be clear what is expected.

Same goes for the possibility to set an individual sync mode, for example if I want to set:

```json
{
	"products": {
		"pull": true
	}
}
```

From my reading of the code, that's not possible as both the pull and push should be set at the same time. I'm not saying we should necessarily support this, but we should evaluate whether that's needed. If not then it should be made clear that it's not supported.

src/API/Site/Controllers/RestAPI/SyncController.php
puntope: Perfect @ianlin  I removed the get_json_params part.

puntope:
@ianlin @mikkamp Thanks for the review and comments. I addressed them by adding:

- Updates what is sent in the body. The rest will remain as it was. This applies to Data Type (Products, Coupon...) and methods (Push or Pull). Notice that before adding both push and pull were required. 
- Add docs at the top of the class with examples.
- Methods to prevent malformations were added to the option. For example. Suppose someone saves in the option a string or any other thing outside the allowed structure, then the default value would be used. 
- Remove unnecessary get_json_params

puntope:
> I had another question to clarify about this PR.
> 
> Taking a quick look at the code it seems that I can update a single group at a time. Which will cause the updated group > to be merged with the current sync modes. 

Correct. 

> Same goes for the possibility to set an individual sync mode. 
> From my reading of the code, that's not possible as both the pull and push should be set at the same time. 

Correct as well. Now I updated it and is allowed.

> However I'm not seeing that clarified in the testing instructions or reflected in the schema

[The Schema was added here](https://github.com/woocommerce/google-listings-and-ads/pull/2796/files). Is there something missing? I added a description. Not sure if something else is expected. 

About PR description. Sorry, fair enough. 

The PR description says: 2. Verify that only the valid keys and values are updated in the options table. 
With "valid keys" I meant the ones set in the body. But you're right its a bit vague. I will update it and also I will add docs in the controller.

ianlin (approved):
Thanks for the changes. LGTM

- Confirmed the 500 error is no longer showing when the body of the POST request is empty.
- Can update an individual sync mode.
- The doc explains clearly

Issue #1936, opened 2023-03-31
test_disable_promotion failing locally


martynmjones:
### Describe the bug:
When running the GLA unit tests locally `test_disable_promotion` in `WCCouponAdapterTest` is failing because of unexpected datetimes. Tests are all passing on GH so this is not a high-priority issue but may need to be investigated further.

![Screenshot 2023-03-31 at 15 17 00](https://user-images.githubusercontent.com/40762232/229145749-60fbb9b9-4106-4685-99d4-4fe033c05201.png)

### Expected behavior:
Tests should pass regardless of the environment they are run in.

mikkamp:
I'm not able to reproduce this failure, although it might be related to a specific time of day since it uses `gmdate( DATE_ATOM )`.

However one thing I noticed is that all data providers (search for @dataProvider) will run regardless of what the filter is set to. For this specific test that won't make too much difference but it will take longer to run because it creates a whole host of test products. A better way to run just that specific test is:

```
vendor/bin/phpunit tests/Unit/Coupon/WCCouponAdapterTest.php --filter=test_disable_promotion
```

mikkamp:
There was a recent failure of this same test, although there was a difference of only 1 second: https://github.com/woocommerce/google-listings-and-ads/actions/runs/10562298192/job/29259970597

![image](https://github.com/user-attachments/assets/6243e694-c9dc-4291-9350-0320eab630a6)

I haven't looked into all the details, but it seems using `gmdate( DATE_ATOM )` twice could potentially be a different timestamp then when the actual code is run. Should find a better way to compare those dates, or use a generated date for the test data, so we can confirm it's always the same.

eason9487:
I couldn't reproduce the 1 hour difference either. Suspect the 1 hour difference could be caused by different datetime functions handling daylight saving time differently when parsing from/converting to a string date.

While the 1 second difference has been fixed by PR #2803, so I will be closing it. If someone can definitively reproduce the 1 hour difference, we can reopen it again.

PR #1260, opened 2022-02-22
Update the start/end date in the timePeriod message of coupon following google.protobuf.Timestamp

songchen2022:
### Changes proposed in this Pull Request:

Update the start/end date in the TimePeriod message following google.protobuf.Timestamp.
The reason why we update is to keep update with latest content API version and deprecate field that previously used in unstable version.
API will still be backward compatible so there is no worries for merchants who has installed the old version.

### Screenshots:

Screen Shot 2022-02-23 at 9 11 03 AM

Screen Shot 2022-02-23 at 9 43 37 AM

### Detailed test instructions:
1. Go to the coupons Page
2. Select one coupon to edit
3. Fill the coupon expiry date in the coupon data form
4.  Edit the publish time in the publish form in the left top corner, and then click update button
5.  Go back to coupons page and verified the expiry date of the selected coupon is same as what we set

### Changelog entry:

Update the start/end date in  the timePeriod of coupon following google.protobuf.Timestam

PR #1245, opened 2022-02-15
Avoid the false positive of PHP test failures

eason9487:
### Changes proposed in this Pull Request:

I've encountered several false positive of PHP test failures, and although I can pass after re-run GH Action jobs, it's still a bit distracting. Such as these jobs:
- https://github.com/woocommerce/google-listings-and-ads/runs/5194436368?check_suite_focus=true#step:6:36
- https://github.com/woocommerce/google-listings-and-ads/runs/5094880222?check_suite_focus=true#step:6:41
- https://github.com/woocommerce/google-listings-and-ads/runs/4949201564?check_suite_focus=true#step:6:41

This PR tries to avoid the above problems by:
- Match a range for the end date of disabling a coupon.
- Assert coupon ID instead of an entire instance when scheduling jobs of trashing and deleting a coupon.

#### 📷  After the fixes, those tests are passed without removing the above forcing `sleep(1)` codes
image

#### Check on GitHub

1. Go to [this Actions page](https://github.com/woocommerce/google-listings-and-ads/runs/5198695049?check_suite_focus=true#step:6:28) to see the failed tests, which were seeded by https://github.com/woocommerce/google-listings-and-ads/pull/1245/commits/12d1593242ad73240bec5d3a01b94e9ef2655c1a.
2. Go to [this Actions page](https://github.com/woocommerce/google-listings-and-ads/runs/5198738262?check_suite_focus=true#step:6:28) to see those tests are passed after the fixes by https://github.com/woocommerce/google-listings-and-ads/pull/1245/commits/6ea98444cd2f99138a40cc77badbaaf882598b32.

#### Test on local env

1. Check out `12d1593242ad73240bec5d3a01b94e9ef2655c1a`
2. Run `vendor/bin/phpunit --filter '/test_disable_promotion|test_(trash|delete)_simple_coupon_schedules_delete_job/'` to see the failed tests.
3. Check out `6ea98444cd2f99138a40cc77badbaaf882598b32`
4. Run `vendor/bin/phpunit --filter '/test_disable_promotion|test_(trash|delete)_simple_coupon_schedules_delete_job/'` to check if the above failed tests are passed.

PR #2803, opened 2025-02-21
Remove a test assertion that occasionally causes PHPUnit tests to false positive

eason9487:

https://github.com/woocommerce/google-listings-and-ads/blob/a524abb5c382612e1b9a4874dfdf639c6ab2b4fe/tests/Unit/Coupon/WCCouponAdapterTest.php#L199-L207

There was a previous version of the above test assertion, which was replaced with the below test assertions in PR #1245. However, it was accidentally added back in this rewritten version in PR #1260 when resolving code conflict, and for the same purpose as the one below.

https://github.com/woocommerce/google-listings-and-ads/blob/a524abb5c382612e1b9a4874dfdf639c6ab2b4fe/tests/Unit/Coupon/WCCouponAdapterTest.php#L212-L214

📌 This PR removes it again to avoid occasional failures like [this run of PHPUnit](https://github.com/woocommerce/google-listings-and-ads/actions/runs/13451910847/job/37587747080)

image

1. View the test assertions to check
   - If they are indeed duplicating the purpose of the test
   - If the retained test assertions are preferable
2. Check if the PHPUnit testing can pass

### DIFF
--- a/tests/Unit/Coupon/WCCouponAdapterTest.php
+++ b/tests/Unit/Coupon/WCCouponAdapterTest.php
@@ -195,18 +195,7 @@ class WCCouponAdapterTest extends UnitTest {
 		);
 		$adapted_coupon->disable_promotion( $coupon );
 		$dates = $adapted_coupon->getPromotionEffectiveTimePeriod();
-
-		$this->assertEquals(
-			new GoogleTimePeriod(
-				[
-					'startTime' => (string) $postdate,
-					'endTime'   => (string) $postdate,
-				]
-			),
-			$dates
-		);
-
-		$now = gmdate( DATE_ATOM );
+		$now   = gmdate( DATE_ATOM );
 
 		//phpcs:disable WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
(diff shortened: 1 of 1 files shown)


mikkamp (approved):
Thanks for the changes, I agree that it should no longer be there as the endTime is tested with a range through:

```
		$this->assertGreaterThanOrEqual( $postdate, $dates->endTime );
		$this->assertLessThanOrEqual( $now, $dates->endTime );
```

I'd still question whether this is the best method to test as we need to grab the current timestamp twice. I'd sooner recommend to set the start time to a fixed timestamp (or use a relative time like yesterday), and then estimate the end time with [`assertEqualsWithDelta`](https://docs.phpunit.de/en/10.5/assertions.html#assertequalswithdelta) to allow for a few seconds of leeway.

But either way these unit tests are not fully covering the functions as it's missing out the scenario "If this coupon is scheduled in the future, disable it right after start date". And it also relies on the coupon expiration date to never be set which means it defaults to 183 day expiry.

I'll go ahead and approve the PR since it wasn't the purpose of this PR to provide full coverage.

mikkamp:
We can also close #1936 after resolving this.

PR #2804, opened 2025-02-21
Block Notifications when PULL mode is disabled

puntope:

This PR blocks Notificacion Job Schedlue as well as Notifications when API PULL SYNC is disabled for a Data Type ( the settings for a specific Data Type inside are set as `pull: false` )

PUSH part will be done in a future PR

1. Grant Data Fetch access in Settings page
2. Be sure your store country supports Coupon Sync (i.e India)
3. Be sure that in G4W - Shipping options you have "Automatically sync my store’s shipping settings to Google."

PRODUCTS PULL

1. verify your Sync Settings by request `GET wp-json/wc/gla/sync` products pull key should be `true`
2. Create a product
4. See `gla/jobs/notifications/products/process_item` scheduled. Run it
5. Verify that the notification happened in google-listings-and-ads logs. Should show something like: ` Notification - Item ID: 62 - Topic... product`
6. Update the product
7. See - `gla/jobs/notifications/products/process_item` scheduled. Run it
8. verify like you did in step 4
9. Permanently delete the product
10. See `gla/jobs/notifications/products/process_item` scheduled. Run it
11. verify like you did in step 4

COUPONS PULL

1. verify your Sync Settings by request `GET wp-json/wc/gla/sync` coupons pull key should be `true`
2. Create a coupon and set Google for WooCommerce  - Show coupon on Google
3. See `gla/jobs/notifications/coupons/process_item` scheduled. Run it
4. Verify that the notification happened in google-listings-and-ads logs. Should show something like: ` Notification - Item ID: 62 - Topic...coupon.`
5. Update the coupon
6. See - `gla/jobs/notifications/coupons/process_item` scheduled. Run it
7. verify like you did in step 4
8. Disable the coupon setting: Google for WooCommerce  -  Dont show coupon 
9. See `gla/jobs/notifications/coupons/process_item` scheduled. Run it
10. verify like you did in step 4

SETTINGS PULL

1. verify your Sync Settings by request `GET wp-json/wc/gla/sync` settings pull key should be `true`
2. Go to WooCommerce Settings and update something
3. See `gla/jobs/notifications/settings/process_item` scheduled. Run it
4. Verify that the notification happened in google-listings-and-ads logs. Should show something like: ` Notification - Item ID:  - Topic: settings.update - Data []`

SHIPPING PULL

1. Be sure that in G4W - Shipping options you have "Automatically sync my store’s shipping settings to Google."
2. verify your Sync Settings by request `GET wp-json/wc/gla/sync` shipping pull key should be `true`
3. Go to WooCommerce Settings - Shipping and update/add something. A zone for example 
4. See `gla/jobs/notifications/shipping/process_item` scheduled. Run it
5. Verify that the notification happened in google-listings-and-ads logs. Should show something like: ` Notification - Item ID:  - Topic: shipping.update - Data []`

API PULL DISABLED

Request `POST wp-json/wc/gla/sync with

```

{
    "products": {
        "pull": false
    },
    "coupons": {
        "pull": false
    },
    "settings": {
        "pull": false
    },
    "shipping": {
        "pull": false
    }
}
```

Verify again all the test steps. But checking that NO Job should be scheduled anymore and NO notification should take place as well 

ℹ️  Notice you can also check Notifications (not jobs) in connection test page under 
Send partner notification request to WPCOM section When you perform Notifications -  Response should be like: `Notification failed. Item: 68 - Topic: product.create`

### Changelog entry

>

### DIFF
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -221,7 +221,7 @@ class SyncController extends BaseOptionsController {
 
 	/**
 	 * Get the parameters from the request body.
-	 * Only the keys in SyncTrait::get_default_sync_mode() that contains boolean pull and/or push param are allowed.
+	 * Only the keys in NotificationsService::get_default_sync_mode() that contains boolean pull and/or push param are allowed.
 	 *
 	 * @param Request $request The request
 	 * @return array
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -5,7 +5,7 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\RestA
 
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
-use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\SyncTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\WP\NotificationsService;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
 use Exception;
@@ -62,7 +62,11 @@ defined( 'ABSPATH' ) || exit;
  */
 class SyncController extends BaseOptionsController {
 
-	use SyncTrait;
+
+	/**
+	 * @var NotificationsService
+	 */
+	protected $notifications_service;
 
 	/**
 	 * The base for routes in this controller.
@@ -74,10 +78,12 @@ class SyncController extends BaseOptionsController {
 	/**
 	 * SyncController constructor.
 	 *
-	 * @param RESTServer $server
+	 * @param RESTServer           $server
+	 * @param NotificationsService $notifications_service
 	 */
-	public function __construct( RESTServer $server ) {
+	public function __construct( RESTServer $server, NotificationsService $notifications_service ) {
 		parent::__construct( $server );
+		$this->notifications_service = $notifications_service;
 	}
 
 	/**
@@ -111,7 +117,7 @@ class SyncController extends BaseOptionsController {
 	protected function get_sync_callback(): callable {
 		return function ( Request $request ) {
 			try {
-				$sync_mode = $this->get_current_sync_mode();
+				$sync_mode = $this->notifications_service->get_current_sync_mode();
 				return $this->prepare_item_for_response( $sync_mode, $request );
 			} catch ( Exception $e ) {
 				return $this->response_from_exception( $e );
@@ -127,7 +133,7 @@ class SyncController extends BaseOptionsController {
 	protected function get_update_sync_callback(): callable {
 		return function ( Request $request ) {
 			try {
-				$sync_mode     = $this->get_current_sync_mode();
+				$sync_mode     = $this->notifications_service->get_current_sync_mode();
 				$new_params    = $this->get_request_params( $request );
 				$new_sync_mode = array_replace_recursive( $sync_mode, $new_params );
 
@@ -227,7 +233,7 @@ class SyncController extends BaseOptionsController {
 			return [];
 		}
 
-		$params       = array_intersect_key( $request_params, $this->get_default_sync_mode() );
+		$params       = array_intersect_key( $request_params, $this->notifications_service->get_default_sync_mode() );
 		$valid_params = [];
 
 		foreach ( $params as $key => $param ) {
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -111,7 +111,7 @@ class SyncController extends BaseOptionsController {
 	protected function get_sync_callback(): callable {
 		return function ( Request $request ) {
 			try {
-				$sync_mode = $this->get_current_sync_value();
+				$sync_mode = $this->get_current_sync_mode();
 				return $this->prepare_item_for_response( $sync_mode, $request );
 			} catch ( Exception $e ) {
 				return $this->response_from_exception( $e );
@@ -127,7 +127,7 @@ class SyncController extends BaseOptionsController {
 	protected function get_update_sync_callback(): callable {
 		return function ( Request $request ) {
 			try {
-				$sync_mode     = $this->get_current_sync_value();
+				$sync_mode     = $this->get_current_sync_mode();
 				$new_params    = $this->get_request_params( $request );
 				$new_sync_mode = array_replace_recursive( $sync_mode, $new_params );
 
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -227,7 +227,7 @@ class SyncController extends BaseOptionsController {
 			return [];
 		}
 
-		$params       = array_intersect_key( $request_params, self::get_default_sync_mode() );
+		$params       = array_intersect_key( $request_params, $this->get_default_sync_mode() );
 		$valid_params = [];
 
 		foreach ( $params as $key => $param ) {
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -215,7 +215,7 @@ class SyncController extends BaseOptionsController {
 
 	/**
 	 * Get the parameters from the request body.
-	 * Only the keys in self::DEFAULT_SYNC_MODE that contains boolean pull and/or push param are allowed.
+	 * Only the keys in SyncTrait::get_default_sync_mode() that contains boolean pull and/or push param are allowed.
 	 *
 	 * @param Request $request The request
 	 * @return array
@@ -227,7 +227,7 @@ class SyncController extends BaseOptionsController {
 			return [];
 		}
 
-		$params       = array_intersect_key( $request_params, self::DEFAULT_SYNC_MODE );
+		$params       = array_intersect_key( $request_params, self::get_default_sync_mode() );
 		$valid_params = [];
 
 		foreach ( $params as $key => $param ) {
--- a/src/API/Site/Controllers/RestAPI/SyncController.php
+++ b/src/API/Site/Controllers/RestAPI/SyncController.php
@@ -5,6 +5,7 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\RestA
 
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\SyncTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
 use Exception;
@@ -61,24 +62,7 @@ defined( 'ABSPATH' ) || exit;
  */
 class SyncController extends BaseOptionsController {
 
-	private const DEFAULT_SYNC_MODE = [
-		'products' => [
-			'pull' => false,
-			'push' => false,
-		],
-		'coupons'  => [
-			'pull' => false,
-			'push' => false,
-		],
-		'shipping' => [
-			'pull' => false,
-			'push' => false,
-		],
-		'settings' => [
-			'pull' => false,
-			'push' => false,
-		],
-	];
+	use SyncTrait;
 
 	/**
 	 * The base for routes in this controller.
@@ -227,22 +211,6 @@ class SyncController extends BaseOptionsController {
 	 */
 	protected function get_update_sync_params(): array {
 		return $this->get_schema_properties();
-	}
-
-	/**
-	 * Get the current value for the API PULL Sync.
-	 * Notice that malformed data will be replaced by default data.
-	 *
-	 * @return array
-	 */
-	protected function get_current_sync_value(): array {
-		$sync_mode = $this->options->get( OptionsInterface::API_PULL_SYNC_MODE );
-
-		if ( ! is_array( $sync_mode ) ) {
-			$sync_mode = self::DEFAULT_SYNC_MODE;
-		}
-
-		return array_replace_recursive( self::DEFAULT_SYNC_MODE, $sync_mode );
 	}
 
 	/**
--- a/src/API/WP/NotificationsService.php
+++ b/src/API/WP/NotificationsService.php
@@ -214,19 +214,19 @@ class NotificationsService implements Service, OptionsAwareInterface {
 	 */
 	protected function get_datatype_from_topic( string $topic ): string {
 		if ( in_array( $topic, self::PRODUCT_TOPICS, true ) ) {
-			return self::get_products_datatype();
+			return $this->get_products_datatype();
 		}
 
 		if ( in_array( $topic, self::COUPON_TOPICS, true ) ) {
-			return self::get_coupons_datatype();
+			return $this->get_coupons_datatype();
 		}
 
 		if ( in_array( $topic, self::SHIPPING_TOPICS, true ) ) {
-			return self::get_shipping_datatype();
+			return $this->get_shipping_datatype();
 		}
 
 		if ( in_array( $topic, self::SETTINGS_TOPICS, true ) ) {
-			return self::get_settings_datatype();
+			return $this->get_settings_datatype();
 		}
 
 		return $topic;
--- a/src/API/WP/NotificationsService.php
+++ b/src/API/WP/NotificationsService.php
@@ -10,6 +10,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\AccountService;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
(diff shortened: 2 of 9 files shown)


src/HelperTraits/SyncTrait.php
```diff
@@ -0,0 +1,135 @@
...
+			],
+			$this->get_shipping_datatype() => [
+				'pull' => true,
+				'push' => true,
+			],
+			$this->get_settings_datatype() => [
+				'pull' => true,
+				'push' => true,
+			],
+		];
```
mikkamp: Now that we don't have the array indexes hardcoded, should we make this a little less repetitive by adding a default mode?

```suggestion
		$default_mode = [
			'pull' => true,
			'push' => true,
		];

		return [
			$this->get_products_datatype() => $default_mode,
			$this->get_coupons_datatype()  => $default_mode,
			$this->get_shipping_datatype() => $default_mode,
			$this->get_settings_datatype() => $default_mode,
		];
```

src/HelperTraits/SyncTrait.php
```diff
@@ -0,0 +1,135 @@
...
+	 * @return bool
+	 */
+	protected function is_pull_enabled_for_datatype( string $data_type = null ): bool {
+		if ( is_null( $data_type ) ) {
+			return true;
+		}
+
+		$sync_modes = $this->get_current_sync_value();
+
+		return apply_filters( 'woocommerce_gla_is_pull_enabled_for_datatype', $sync_modes[ $data_type ]['pull'] ?? false );
```
mikkamp: Always best to cast to expected type after applying a filter as we don't know what type a filter might return.

```suggestion
		return (bool) apply_filters( 'woocommerce_gla_is_pull_enabled_for_datatype', $sync_modes[ $data_type ]['pull'] ?? false );
```

Curious how useful this filter is used. If someone actually intended to use this then wouldn't they need the `$datatype` passed to the filter to check if they want to override it? Also wouldn't it be handier for them to have a filter to the function `get_current_sync_value` if they needed to override sync values?

If the prime reason to add this filter is just for the unit testing, then I'd suggest we don't add the filter. In this case we can mock the value returned from `$this->options` to mimic a specific mode for a particular datatype.

Same applies for the `is_push_enabled_for_datatype` function.

src/HelperTraits/SyncTrait.php
```diff
@@ -0,0 +1,135 @@
...
+	/**
+	 * Check if API PULL is enabled for a specific data type.
+	 *
+	 * Checking null data type returns true.
+	 * Checking a non-existent data type will return false.
+	 *
+	 * @param string|null $data_type The data type to check.
+	 * @return bool
+	 */
+	protected function is_pull_enabled_for_datatype( string $data_type = null ): bool {
```
mikkamp: Looking at just the definition of this function I don't think we should support a default `data_type` of `null`. The name of the function specifically calls out that it's checking it for a specific data type, to return true by default (also unfiltered) doesn't really make sense.

If the reason we added this is so we can still call `->is_ready()` without parameters. Then I'd suggest to remodel the `is_ready` function instead. Could be something like:

```php
		return $is_ready && ( is_null( $data_type ) || $this->is_pull_enabled_for_datatype( $data_type ) );
```

Same applies for the `is_push_enabled_for_datatype` function.

src/HelperTraits/SyncTrait.php
```diff
@@ -0,0 +1,135 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\Jobs\MigrateGTIN;
```
mikkamp: This is not used in this trait.

src/HelperTraits/SyncTrait.php
```diff
@@ -0,0 +1,135 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Exception;
+use WC_Product;
```
mikkamp: These two are not used in this trait.

src/HelperTraits/SyncTrait.php
```diff
@@ -0,0 +1,135 @@
...
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Trait with some helpers for API Pull and MC Push Sync modes.
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits
+ */
+trait SyncTrait {
+
+	use OptionsAwareTrait;
```
mikkamp: This is deceiving because in order for the `OptionsAwareTrait` to work, it also needs to implement `OptionsAwareInterface`. Which we can see being done in the `NotificationsService` class. But for example in `Product/SyncerHooks`, we don't implement this so the options object is not available. Which is ok because we don't call any of the functions that need the options object. If later on we decide to call another function from this trait then we can get a fatal error. So I'd prefer if we:

1. Don't include the `OptionsAwareTrait` within this trait
2. Have some better fall backs if the `$this->options` property is not available.

tests/Unit/API/WP/NotificationsServiceTest.php
```diff
@@ -83,6 +84,7 @@ function ( $value, $name ) {
 		);
 
 		add_filter( 'woocommerce_gla_notifications_enabled', '__return_true' );
+		add_filter( 'woocommerce_gla_is_pull_enabled_for_datatype', '__return_true' );
```
mikkamp, line 86: As mentioned I think it's better to mock the get for `OptionsInterface::API_PULL_SYNC_MODE` to return a specific sync mode instead of just returning true for all modes.

src/API/Site/Controllers/RestAPI/SyncController.php
```diff
@@ -259,7 +227,7 @@ protected function get_request_params( Request $request ): array {
 			return [];
 		}
 
-		$params       = array_intersect_key( $request_params, self::DEFAULT_SYNC_MODE );
+		$params       = array_intersect_key( $request_params, self::get_default_sync_mode() );
```
mikkamp: We haven't made this function static, so isn't it better to call `$this->get_default_sync_mode`?

src/API/WP/NotificationsService.php
```diff
@@ -186,4 +205,30 @@ public function is_ready( bool $with_health_check = true ): bool {
...
+
+	/**
+	 * Get the DataType from a Notification Topic
+	 *
+	 * @param string $topic The topic.
+	 * @return string The DataType
+	 */
+	protected function get_datatype_from_topic( string $topic ): string {
+		if ( in_array( $topic, self::PRODUCT_TOPICS, true ) ) {
+			return self::get_products_datatype();
```
mikkamp: Same here we should use `$this->` instead of `self::`

src/Coupon/SyncerHooks.php
```diff
@@ -194,7 +196,7 @@ public function pre_delete( int $coupon_id ) {
 	protected function handle_update_coupon( WC_Coupon $coupon ) {
 		$coupon_id = $coupon->get_id();
 
-		if ( $this->notifications_service->is_ready() ) {
+		if ( $this->notifications_service->is_ready( self::get_coupons_datatype() ) ) {
```
mikkamp: Same here we should use `$this->` instead of `self::`

src/Coupon/SyncerHooks.php
```diff
@@ -277,7 +279,7 @@ protected function get_coupon_to_delete( WC_Coupon $coupon ): WCCouponAdapter {
 	 * @param int $coupon_id
 	 */
 	protected function handle_delete_coupon( int $coupon_id ) {
-		if ( $this->notifications_service->is_ready() ) {
+		if ( $this->notifications_service->is_ready( self::get_coupons_datatype() ) ) {
```
mikkamp: Same here we should use `$this->` instead of `self::`

mikkamp (commented):
Thanks for the changes, here. While they largely work as intended I left some comments about the code.

> When you perform Notifications - Response should be like: Notification failed. Item: 68 - Topic: product.create

I realize I can go to the log files and check the reason why a notification failed. However considering we can either have a notification type blocked or a notification request actually failing it would be much more useful if the `notify` function would throw an exception so we can check / show the reason.

Also I noticed that in the Connection test we create a new instance of `NotificationsService` when in the core service provider we share the service class (meaning it's a singleton). It doesn't really cause any issues because mostly everything is pretty static, but if it would ever have any particular state, the behaviour would be different between the two instances. CoPilot also highlights it as something to improve:
image

puntope:
> Also I noticed that in the Connection test we create a new instance of NotificationsService when in the core service provider we share the service class (meaning it's a singleton). It doesn't really cause any issues because mostly everything is pretty static, but if it would ever have any particular state, the behaviour would be different between the two instances. CoPilot also highlights it as something to improve:


This part I believe can be addressed in the PR that will cover the Connection test showing the current values if that's ok.

src/HelperTraits/SyncTrait.php
puntope: Thanks @mikkamp 

>  If someone actually intended to use this then wouldn't they need the $datatype

Yes, correct, I missed the datatype

> If the prime reason to add this filter is just for the unit testing

The reason is to be able enable/disable push/pull using the filter for a specific datatype or all of them.

> So wouldn't it be handier for them to have a filter to the function get_current_sync_value if they needed to override sync values?

In practice this filter does the same. The difference is that you can enable or disable the method directly instead of overriding the values.

src/API/Site/Controllers/RestAPI/SyncController.php
puntope: Forgot to change that after refactoring.

src/HelperTraits/SyncTrait.php
puntope: I added the new filter anyway.

src/HelperTraits/SyncTrait.php
puntope: > If the reason we added this is so we can still call ->is_ready() without parameters. Then I'd suggest to remodel the is_ready function instead. Could be something like:

Yes thats the reason

src/HelperTraits/SyncTrait.php
puntope: >  Which we can see being done in the NotificationsService class. But for example in Product/SyncerHooks, we don't implement this so the options object is not available. Which is ok because we don't call any of the functions that need the options object. If later on we decide to call another function from this trait then we can get a fatal error. 

We call `$this->options->get( OptionsInterface::API_PULL_SYNC_MODE );` inside `get_current_sync_mode` I believe it works because SycneHooks require NotificationService which loads Options. 

> Don't include the OptionsAwareTrait within this trait

So do you suggest to use instead a plain `get_option()` instead?

src/HelperTraits/SyncTrait.php
puntope: Good suggestion. Added here https://github.com/woocommerce/google-listings-and-ads/pull/2804/commits/9ea4e80d7fb9543d4e0fa4945efc2477804a1792

src/HelperTraits/SyncTrait.php
puntope: https://github.com/woocommerce/google-listings-and-ads/pull/2804/commits/19fad8762198794837af6c2672a4e83e7dfc95cf

src/HelperTraits/SyncTrait.php
puntope: https://github.com/woocommerce/google-listings-and-ads/pull/2804/commits/39ad17efd6add16f646e13fad7c308f46cf29672

src/HelperTraits/SyncTrait.php
puntope: Thanks. Fixed here https://github.com/woocommerce/google-listings-and-ads/pull/2804/commits/39ad17efd6add16f646e13fad7c308f46cf29672

tests/Unit/API/WP/NotificationsServiceTest.php
puntope: I will wait for clarification in this https://github.com/woocommerce/google-listings-and-ads/pull/2804#discussion_r1979371733 as it may impact the options.

src/Coupon/SyncerHooks.php
puntope: Thanks. Fixed here https://github.com/woocommerce/google-listings-and-ads/pull/2804/commits/e595f4c28c4832b84f9f0079e3c9554aeac82124

src/Coupon/SyncerHooks.php
puntope: Thanks. Fixed here https://github.com/woocommerce/google-listings-and-ads/commit/e595f4c28c4832b84f9f0079e3c9554aeac82124

src/API/WP/NotificationsService.php
puntope: Thanks. Fixed here https://github.com/woocommerce/google-listings-and-ads/commit/e595f4c28c4832b84f9f0079e3c9554aeac82124

src/API/Site/Controllers/RestAPI/SyncController.php
puntope: Thanks. Fixed here https://github.com/woocommerce/google-listings-and-ads/commit/e595f4c28c4832b84f9f0079e3c9554aeac82124

puntope:
Hi @mikkamp I applied most of the comments. Just not sure about the suggested change regarding OptionsTrait. Waiting for clarifications. Thanks

src/HelperTraits/SyncTrait.php
mikkamp: @puntope  Thanks for looking into this. It's a bit of a tricky one, which is why I didn't initially suggest a solution.

However the "common" functions that we need a sync trait for are the 4 functions `get_*_datatype`. The rest should be part of a service class, I'm not sure if they'll fit in NotificationService or if we need a separate SyncService class. While it's nice to group all the functions together in one trait. If they are relying on parts of a service class (in this case the options object), then they don't belong in the trait (unless each of the service classes that use the trait has the same properties available, which in this case isn't defined clearly).

So my suggestion here would be to make these three functions part of a service class:
1. `get_current_sync_mode`
2. `is_pull_enabled_for_datatype`
3. `is_push_enabled_for_datatype`

That service class can then be provided to the SyncController. If we are going to end up writing a SyncService then we might also want to pass that to each of the SyncerHooks classes and do away with the need for a trait to share between them.

src/HelperTraits/SyncTrait.php
```diff
@@ -0,0 +1,119 @@
...
+
+	/**
+	 * Check if MC PUSH is enabled for a specific data type.
+	 *
+	 * @param string|null $data_type The data type to check.
+	 * @return bool
+	 */
+	protected function is_push_enabled_for_datatype( string $data_type = null ): bool {
+		if ( is_null( $data_type ) ) {
+			return true;
```
mikkamp: I'd also suggest the same here and make $data_type required.

puntope: This can be handle in the Push PR  if that's ok.

mikkamp: Sure, that's not a problem. Just thought it was odd to be adding this function specifically in this PR with no use and then later refactoring it to what we need it to be (with the risk that we forget about it).

puntope:
Hi @mikkamp I adjusted the code for using NotificationService instead of the trait.

Let me know if this is the direction you expedited thanks.

src/API/WP/NotificationsService.php
```diff
@@ -100,7 +118,7 @@ public function notify( string $topic, $item_id = null, $data = [] ): bool {
 		 * @param int $item_id The item_id for the notification.
 		 * @param string $topic The topic for the notification.
 		 */
-		if ( ! apply_filters( 'woocommerce_gla_notify', $this->is_ready() && in_array( $topic, self::ALLOWED_TOPICS, true ), $item_id, $topic ) ) {
+		if ( ! apply_filters( 'woocommerce_gla_notify', $this->is_ready( $this->get_datatype_from_topic( $topic ) ) && in_array( $topic, self::ALLOWED_TOPICS, true ), $item_id, $topic ) ) {
 			$this->notification_error( $topic, 'Notification was not sent because the Notification Service is not ready or the topic is not valid.', $item_id );
```
mikkamp, line 122: This error is even harder to troubleshoot now. Because there can many conditions leading to this error.

Especially in a scenario where the connection is healthy and notifications are all working, but it just happens to be disabled for one service.

For example I get:
```
Error sending notification for Item ID 123 with topic coupon.create. Notification was not sent because the Notification Service is not ready or the topic is not valid.
```

I guess if we follow up in the connection page with showing exactly which topics are enabled it might be easier to troubleshoot (I assume that's coming in a later PR).

src/API/WP/NotificationsService.php
```diff
@@ -4,11 +4,13 @@
 namespace Automattic\WooCommerce\GoogleListingsAndAds\API\WP;
 
 use Automattic\Jetpack\Connection\Client;
+use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\SyncTrait;
```
mikkamp, line 7: This is now the only location that SyncTrait is used, so it's not really sharing traits between classes. But I'm ok with leaving it as is.

tests/Unit/API/WP/NotificationsServiceTest.php
```diff
@@ -5,6 +5,7 @@
 
 use Automattic\WooCommerce\Admin\RemoteInboxNotifications\TransformerService;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\WP\NotificationsService;
+use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\SyncTrait;
```
mikkamp: This is no longer used as the functions are included with `NotificationsService`

mikkamp (approved):
Thanks for all the changes. The testing is all working as expected, I'm able to send notifications for some topics and have others be blocked depending on the sync mode.

I left a few minor comments but I'll go ahead and approve it.

src/API/WP/NotificationsService.php
puntope: I will reftactor it in further PRs as also this implies we can use constants for the Data Types

src/API/WP/NotificationsService.php
puntope: Yes. It does

tests/Unit/API/WP/NotificationsServiceTest.php
puntope: https://github.com/woocommerce/google-listings-and-ads/pull/2804/commits/b98f3aa3375554b6f780ec50352e53583f2f4e7e

PR #2806, opened 2025-02-26
Fix the issue of not being able to reconnect to the auto-selected Google ads account during onboarding

### LINKED ISSUE #2805: It may be unable to connect to the auto-selected Google Ads account after disconnecting
### Describe the bug:

In the onboarding flow, it may be unable to connect to the auto-selected Google Ads account after disconnecting.

This problem always occurs when there is only one existing account.

### Steps to reproduce:

https://github.com/user-attachments/assets/b0bdf7bd-5736-4687-ab6b-021a664ee0b1

### Expected behavior:

Be able to connect to the auto-selected Google Ads account after disconnecting

### Actual behavior:

Nothing happened after clicking on the "Connect" button

eason9487:

Closes #2805

This PR fixes the problem and also add the relevant test cases.

### Screenshots:

https://github.com/user-attachments/assets/a1374bde-995d-4e6a-a7f5-d22948da0b31

1. Go to step 1 of the onboarding flow.
2. Connect to the first existing Google Ads account.
3. Disconnect the Google Ads account.
4. Without interacting with the account select control, Check if it can reconnect to the first existing Google Ads account again.

### Changelog entry

> Fix - It may be unable to connect to the auto-selected Google Ads account after disconnecting during onboarding.

### DIFF
--- a/js/src/components/google-combo-account-card/connect-ads/connect-existing-account.js
+++ b/js/src/components/google-combo-account-card/connect-ads/connect-existing-account.js
@@ -77,8 +77,7 @@ const ConnectExistingAccount = ( { onCreateClick } ) => {
 		 * Prevent the `value` from staying on the unclaimed and disconnected account ID.
 		 * Please note that the reset works because the `AdsAccountSelectControl` happens to
 		 * switch between two different `AppSelectControls` so that `autoSelectFirstOption`
-		 * can be triggered again. Otherwise, it would need to specify `key={ Boolean(value) }`
-		 * in the `<AdsAccountSelectControl>` use of this component.
+		 * can be triggered again.
 		 */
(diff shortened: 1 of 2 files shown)


mikkamp (approved):
Thanks for the change. I tried testing this before the PR, and I could confirm that when disconnecting the current account, it stayed on the previously selected button, but the connect button wasn't working (until you interact with the selection box). It was still sending tracks though.

After the PR the behaviour changed slightly. After disconnecting it jumps to the first ad account (which is different from staying on the previous selection). The connect button was right away usable though. I suppose jumping back to the first selection is acceptable because the user clicked on "connect to a different Google Ads account" so it's not necessarily expected to reconnect to the same one.

PR #2810, opened 2025-03-04
Block Shipping, Products and Coupons when Sync Push is disabled

puntope:

Block Merchant Center Sync in case the Sync is disabled for the datatype

1. Checkout PR and have a completed onboarding. All the Sync Datatype for PUSH should be enabled
3. Create a product. Go to WoOCommere Scheduled actions and see `gla/jobs/update_products/process_item`. Run it 
4. Update the product.  Go to WoOCommere Scheduled actions and see `gla/jobs/update_products/process_item`. Run it 
5. Create a coupon and select "Show in Google" . Go to WoOCommere Scheduled actions and see `gla/jobs/update_coupon/process_item`. Run it 
6. Create a coupon and select "Show in Google".  Go to WoOCommere Scheduled actions and see `gla/jobs/update_coupon/process_item`. Run it 
7. Go to Google - Shipping tab and select "Automatically sync my store’s shipping settings to Google."
8. Go to WooCommerce - Settings - Shipping and create a Shipping Zone
9. Go to WooCommerce - Status - Scheduled Actions and see `gla/jobs/update_shipping_settings/process_item`
10. Deactivate Push Sync for Products, Coupons and Shipping 
11. Repeat step 3. No Job should be scheduled
12. Try to sync a product manually using the Connection Test Page. It should fail. "Error submitting product to Google: Pushing products will not run if the PUSH Sync functionality has been disabled."
13. Repeat steps 5.  No Job should be scheduled 
14. Repeat step 8. No Job should be scheduled 

### Additional details:

PT: pcTzPl-2yG-p2

### Changelog entry

>

### DIFF
--- a/src/Jobs/AbstractProductSyncerJob.php
+++ b/src/Jobs/AbstractProductSyncerJob.php
@@ -4,7 +4,7 @@ declare( strict_types=1 );
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Jobs;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\ActionScheduler\ActionSchedulerInterface;
-use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\SyncTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\WP\NotificationsService;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductRepository;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer;
@@ -17,8 +17,6 @@ defined( 'ABSPATH' ) || exit;
  * @package Automattic\WooCommerce\GoogleListingsAndAds\Jobs
  */
 abstract class AbstractProductSyncerJob extends AbstractActionSchedulerJob {
-
-	use SyncTrait;
 
 	/**
 	 * @var ProductSyncer
@@ -65,6 +63,6 @@ abstract class AbstractProductSyncerJob extends AbstractActionSchedulerJob {
 	 * @return bool Returns true if the job can be scheduled.
 	 */
 	public function can_schedule( $args = [] ): bool {
-		return ! $this->is_running( $args ) && $this->merchant_center->should_push( $this->get_products_datatype() );
+		return ! $this->is_running( $args ) && $this->merchant_center->is_enabled_for_datatype( NotificationsService::DATATYPE_PRODUCT );
 	}
 }
--- a/src/Jobs/AbstractProductSyncerJob.php
+++ b/src/Jobs/AbstractProductSyncerJob.php
@@ -4,6 +4,7 @@ declare( strict_types=1 );
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Jobs;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\ActionScheduler\ActionSchedulerInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\HelperTraits\SyncTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantCenterService;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductRepository;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductSyncer;
@@ -16,6 +17,8 @@ defined( 'ABSPATH' ) || exit;
  * @package Automattic\WooCommerce\GoogleListingsAndAds\Jobs
  */
 abstract class AbstractProductSyncerJob extends AbstractActionSchedulerJob {
+
+	use SyncTrait;
 
 	/**
 	 * @var ProductSyncer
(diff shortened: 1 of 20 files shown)


src/MerchantCenter/MerchantCenterService.php
```diff
@@ -126,17 +126,28 @@ public function is_ready_for_syncing(): bool {
...
+	 *
+	 * @param string $data_type The data type to check.
+	 * @return bool
+	 * @since x.x.x
+	 */
+	public function is_enabled_for_datatype( string $data_type ): bool {
+		/** @var NotificationsService $notifications_service */
+		$notifications_service = $this->container->get( NotificationsService::class );
+		return $notifications_service->is_push_enabled_for_datatype( $data_type );
+	}
```
mikkamp, line 155: So two confusing things about this function:

1. If it's actually just a wrapper to call `is_push_enabled_for_datatype` then why doesn't the name also reflect that it's checking if push is enabled?
2. Push as nothing to do with notifications, so why is it calling a notification service?

Reasoning why I think we shouldn't call the notification service is because we should think of push and pull as two completely independent parts, which can both run together (dark launch), or toggled between the two. We could also decide to later remove one. So making push logic dependent on notifications (which is pull only) doesn't separate the two. Should we have a separate service handling the overlapping logic between the two?

src/Shipping/SyncerHooks.php
```diff
@@ -142,7 +142,9 @@ protected function handle_update_shipping_settings() {
 			$this->job_repository->get( ShippingNotificationJob::class )->schedule( [ 'topic' => NotificationsService::TOPIC_SHIPPING_UPDATED ] );
 		}
 
-		$this->job_repository->get( UpdateShippingSettings::class )->schedule();
+		if ( $this->merchant_center->is_enabled_for_datatype( NotificationsService::DATATYPE_SHIPPING ) ) {
+			$this->job_repository->get( UpdateShippingSettings::class )->schedule();
+		}
```
mikkamp, line 147: So the changes here seem pretty straightforward with simple logic:

- If pull is enabled: send a notification
- If push is enabled: schedule a job to start the sync

But when I look at other SyncerHooks I don't see that same clear separation. For example for products I can see that if push is disabled, it will still perform the following steps for a single product (or a batch of products in case it's a variable product):

1. Check if the product is sync ready
2. Mark the product as pending sync
3. If not sync ready prepare delete requests
4. Schedule the products to be updated
5. Schedule the products to be deleted

Now because `can_scheduler` will check if push is enabled, step 4 and 5 won't actually schedule the job in action scheduler. But in step 2 we already modified the product.

So I think we should apply the same simple logic like we did for the shipping syncerhooks, and not even attempt to do anything if push is not enabled.

> [!Note]
> Same logic applies for coupons.

src/Jobs/AbstractProductSyncerJob.php
```diff
@@ -62,6 +63,6 @@ public function __construct(
 	 * @return bool Returns true if the job can be scheduled.
 	 */
 	public function can_schedule( $args = [] ): bool {
-		return ! $this->is_running( $args ) && $this->merchant_center->should_push();
+		return ! $this->is_running( $args ) && $this->merchant_center->is_enabled_for_datatype( NotificationsService::DATATYPE_PRODUCT );
```
mikkamp, line 66: As I outlined in the previous comment, we should prevent pushing at an earlier stage. I can however understand we want some logic checking within the syncer jobs, but shouldn't that be for when a job item is being processed.

Here we are only blocking a job from being scheduled. But what happens if we perform the actions in this order:

1. Schedule a product update
2. Disable product push
3. Scheduled actions runs (at this step I'd expect it to bail as push is no longer enabled)

`woocommerce_gla_batch_retry_update_products` will also be happily triggered again without being blocked if the previous sync failed.

src/Jobs/UpdateShippingSettings.php
```diff
@@ -97,7 +99,7 @@ public function schedule( array $args = [] ) {
 	 * @return bool
 	 */
 	protected function can_sync_shipping(): bool {
-		// Confirm that the Merchant Center account is connected and the user has chosen for the shipping rates to be synced from WooCommerce settings.
-		return $this->merchant_center->is_connected() && $this->google_settings->should_get_shipping_rates_from_woocommerce();
+		// Confirm that the Merchant Center account is connected, the user has chosen for the shipping rates to be synced from WooCommerce settings and the Push Sync is enabled for Shipping.
+		return $this->merchant_center->is_connected() && $this->google_settings->should_get_shipping_rates_from_woocommerce() && $this->merchant_center->is_enabled_for_datatype( NotificationsService::DATATYPE_SHIPPING );
```
mikkamp, line 103: So here we are making the change in the right location. Because `can_sync_shipping` will be checked when scheduling the job but also when processing the items of the job. So if a job was already scheduled and the push disabled later it will still prevent the job from being run.

At most we might want to tweak the failure message to indicate it could also be because push is disabled:
> Cannot sync shipping settings. Confirm that the merchant center account is connected and the option to automatically sync the shipping settings is selected.

mikkamp (commented):
Thanks @puntope for working on these changes.

As explained in the comments, I think we should follow some of the simple logic as is done for Shipping. As it is right now a product/coupon status will still be changed to pending sync even if push is disabled, we should not allow any changes if it's disabled.

budzanowski:
This PR will be delivered as part of the combined effort in #2981, which includes both the sync push controls and WC REST API settings integration. Closing this individual PR in favor of the combined approach.

PR #2820, opened 2025-03-20
Add Budget Metrics endpoint for fetching metrics with a custom budget

mikkamp:

This is a follow up to #2795 

This allows for fetching a Campaign Budget Recommendation to extract the metrics when specifying a specific budget.

Since it was quite similar to the changed for fetching Budget Recommendations from the API, CoPilot was able to generate the unit tests (with a minor set of corrections needed).

1. Reinstall composer dependencies `rm -rf vendor && composer install`
2. Test API request `https://domain.test/wp-json/wc/gla/ads/campaigns/budget-metrics?country_codes[]=IE&budget=16.77`
3. Ensure a single set of metrics is returned (cost is for a whole week so it should be 7 x budget):
```json
{
    "currency": "EUR",
    "budget": 16.77,
    "country": "IE",
    "metrics": {
        "cost": 117.39,
        "conversions": 6.8,
        "conversions_value": 270.65197389217025
    }
}
```
4. Repeat the call and it should respond much faster as it will be pulling from the cache
5. Adjust the fallback rate in the table `wp_gla_budget_recommendations` so it is higher than the recommended amount but still lower than the high recommendation (ex: 13)
6. Send API request to fetch recommendations `https://domain.test/wp-json/wc/gla/ads/campaigns/budget-recommendation?country_codes[]=IE`
7. Ensure the fallback rate has metrics returned (to confirm it got the right metrics, the cost should be 7 x fallback budget):
```json
{
    "currency": "EUR",
    "recommendations": [
        {
            "daily_budget": 13,
            "country": "IE",
            "level": "Recommended",
            "metrics": {
                "cost": 91,
                "conversions": 6.2,
                "conversions_value": 248.78137720487214
            }
        },
        {
            "daily_budget": 14.4,
            "metrics": {
                "cost": 100.8,
                "conversions": 6.4,
                "conversions_value": 257.03656077956504
            },
            "country": "IE",
            "level": "High"
        }
    ]
}
```

### Additional details:
Project Thread: pcTzPl-2Hb-p2

### Changelog entry
* Add - Budget Metrics endpoint for fetching metrics with a custom budget.

### DIFF
--- a/src/API/Google/BudgetMetrics.php
+++ b/src/API/Google/BudgetMetrics.php
@@ -108,7 +108,14 @@ class BudgetMetrics implements OptionsAwareInterface, TransientsAwareInterface {
 
 				// Parse the metrics for the given budget.
 				$metrics = $this->parse_metrics( $campaign_budget_recommendation, $budget );
-				$this->transients->set( TransientsInterface::ADS_BUDGET_METRICS, [ $cache_key => $metrics ], HOUR_IN_SECONDS * 12 );
+
+				// Merge with previously cached metrics.
+				if ( ! is_array( $transient ) ) {
+					$transient = [];
+				}
+				$transient[ $cache_key ] = $metrics;
+				$this->transients->set( TransientsInterface::ADS_BUDGET_METRICS, $transient, HOUR_IN_SECONDS * 12 );
+
 				return $metrics;
 			}
 		} catch ( ApiException $e ) {
--- a/src/API/Google/BudgetMetrics.php
+++ b/src/API/Google/BudgetMetrics.php
@@ -136,7 +136,7 @@ class BudgetMetrics implements OptionsAwareInterface, TransientsAwareInterface {
 	protected function parse_metrics( CampaignBudgetRecommendation $recommendation, float $budget ): ?array {
 		foreach ( $recommendation->getBudgetOptions() as $budget_option ) {
 			$amount = $this->from_micro( $budget_option->getBudgetAmountMicros() );
-			if ( $amount === $budget ) {
+			if ( abs( $amount - $budget ) < 0.00001 ) {
 				$metrics = $budget_option->getImpact()->getPotentialMetrics();
 
 				return [
--- a/src/API/Google/BudgetMetrics.php
+++ b/src/API/Google/BudgetMetrics.php
@@ -1,0 +1,145 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\MicroTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\LocationIDTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Google\Ads\GoogleAdsClient;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsAwareTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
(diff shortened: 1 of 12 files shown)


src/API/Site/Controllers/Ads/BudgetMetricsController.php
```diff
@@ -0,0 +1,205 @@
...
+						'currency'      => $currency,
+						'country_codes' => $country_codes,
+					],
+					400
+				);
+			}
+
+			// Fetch metrics from the Google Ads API.
+			$budget_metrics = $this->container->get( BudgetMetrics::class );
+			$metrics        = $budget_metrics->get_metrics( $budget, $country_codes );
```
eason9487, line 119: Calling API with a negative `budget` results in 500 internal server error. Would it be better to make it return 400 bad request error instead?

image

src/API/Site/Controllers/Ads/BudgetMetricsController.php
```diff
@@ -0,0 +1,205 @@
...
+				'validate_callback' => 'rest_validate_request_arg',
+			],
+			'country'  => [
+				'type'        => 'string',
+				'description' => __( 'Country code in ISO 3166-1 alpha-2 format.', 'google-listings-and-ads' ),
+				'context'     => [ 'view' ],
+			],
+			'metrics'  => [
+				'type'  => 'array',
+				'items' => [
```
eason9487: Look like this schema doesn't match the actual response.

```json
{
    "currency": "USD",
    "budget": 9,
    "country": "US",
    "metrics": {
        "cost": 63,
        "conversions": 6.4,
        "conversions_value": 254.65734449367477
    }
}
```

src/API/Google/BudgetMetrics.php
```diff
@@ -0,0 +1,145 @@
...
+
+			foreach ( $response->getRecommendations() as $recommendation ) {
+				$campaign_budget_recommendation = $recommendation->getCampaignBudgetRecommendation();
+				if ( ! $campaign_budget_recommendation ) {
+					continue;
+				}
+
+				// Parse the metrics for the given budget.
+				$metrics = $this->parse_metrics( $campaign_budget_recommendation, $budget );
+				$this->transients->set( TransientsInterface::ADS_BUDGET_METRICS, [ $cache_key => $metrics ], HOUR_IN_SECONDS * 12 );
```
eason9487: This cache always stores only the last query result that arrived here. I wonder is this the intended implementation?

Consider that the user may adjust the value back and forth when editing the budget. Although the frontend can add a debounce to reduce the number of calls to the API, the cache won't be hit even if the user is just switching between two values.

eason9487 (commented):
Thanks for the work!

The API works as expected, but I would like to confirm whether the implementation of caching can meet the desired benefits.

src/API/Site/Controllers/Ads/BudgetMetricsController.php
mikkamp: Nice catch, AI copied this for me and was supposed to follow the pattern for `BudgetRecommendationController`. I didn't notice that it messed up there.

src/API/Google/BudgetMetrics.php
mikkamp: Good suggestion. We pulled the previous value of the transient already, so the two should be merged, to cache multiple values with unique cache keys.

src/API/Site/Controllers/Ads/BudgetMetricsController.php
mikkamp: We should be able to resolve this by using the number format: https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#numbers

If we set the minimum to 0 it should block negative numbers.

mikkamp: Also good scenario to add a unit test for.

mikkamp:
Thanks for all the suggestions @eason9487 I've gone ahead and applied those changes now.

I also noticed that the amount comparison wasn't correct as it doesn't account for float inconsistencies, so I improved the check and added an additional unit test for it.

Previously requesting a budget for 16.4 wouldn't return any metrics because the value returned from the API would be 16.3999998 which isn't an exact match. The suggestion is to do the comparison using `PHP_FLOAT_EPSILON`, but since we are comparing float values from an external system I went with an accuracy of `0.00001`. I think that should cover the currencies we'd like to compare.

eason9487 (approved):
Thanks for the update! LGTM.

src/API/Site/Controllers/Ads/BudgetMetricsController.php
mikkamp: Forgot about this comment, created a follow up PR here: https://github.com/woocommerce/google-listings-and-ads/pull/2823

PR #2821, opened 2025-03-23
Add the G4W settings to WC REST API

tomalec:

- This PR adds `google-for-woocommerce` group to the WC REST API `/settings` endpoint and G4W specific settings to that group.
   Part of pcTzPl-2AM-p2
- Added some code docs about `gla_syncable` peeuvX-2ea-p2
- I also slightly tweaked the implementation of `prepare_data`, moving the endpoint check out side the loop.

### Screenshots:
image

0. Start with a fresh install, or disconnect all accounts
1. ```sh
   curl -X GET "https://yoursite.com/wp-json/wc/v3/settings/google-for-woocommerce"      -u 
   "restkey:restsecret"
   ```
   Check that the response is empty `200`.
2. ```sh
   curl -X GET "https://yoursite.com/wp-json/wc/v3/settings/google-for-woocommerce?gla_syncable=1"      -u 
   "restkey:restsecret"
   ```
3. Check that you get all keys, with empty/default values for the fresh install.
	```
	'gla_google_connected'
	'gla_jetpack_connected'
	'gla_language'
	'gla_merchant_center'
	'gla_shipping_rates'
	'gla_shipping_times'
	'gla_target_audience'
	```
   or for cleared accounts, that `*_connected` are set to `false`, `gla_shipping_*` are set to `[]`, and other settings as as they were set during onboarding.
4. Start onboarding, and after each step, redo the request, checking for updated values.
5. After finishing onboarding, go to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsettings` and change the tax setting.
6. Resend the request and check that the change is reflected.
7. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fshipping` and change shipping settings
6. Resend the request and check that the change is reflected.

### Additional details:

1. In https://github.com/woocommerce/google-listings-and-ads/pull/2342 Jorge use `rest_request_after_callbacks` filter to add settings
	> Initially, my idea was to use the filter `woocommerce_settings- . $group_id` to add a new group of settings. However, this filter can only be used for Options (which should be fine for the target audience but not for the shipping times). Additionally, the filter `woocommerce_settings-` doesn't provide the Request Object param so I decided to use the filter `rest_request_after_callbacks` which makes it easier to verify if the request is made from the proxied endpoints.
	
	I think we can still use `woocommerce_settings*` filters to create the dedicated group. We could have added option-related settings in `woocommerce_settings-google-for-woocommerce` and even hacked to add non-option-related ones there as well (using the `$setting['default']` ). But that would require us to set `'type'` which will not be accurate, and would provide `link`s which are not working for our settings, etc. That's why I decided to add all settings in `rest_request_after_callbacks` in the same format that we used before.
	However, to make it work under ``woocommerce_settings-google-for-woocommerce`` it required [a bit of a hack](https://github.com/woocommerce/google-listings-and-ads/pull/2821/files#diff-0754aa1c2cb23cc7ac551b917d6800de2c34bb6f13ce10ee789799c6595fd846R264-R275).

2. Related to the discussion at peeuvX-2ea-p2, initially, I was thinking of making the `settings/google-for-woocommerce` return everything without the `gla_syncable` param, as I thought it would require a pretty convoluted implementation to return options from the ``woocommerce_settings-google-for-woocommerce`` filter, then conditionally remove it. However, I find a small hack that let us avoid that. So I kept everything behind `gla_syncable`, as it was before.
3. I added `gla_jetpack_connected` and `gla_google_connected` entries. They were not mentioned in the shared Google doc, but I believe that it could be useful to Google to check if a merchant disconnected their account. Plus, those could be considered onboarding inputs.
4. I'm not 100% sure if we should prefix `id`s with `gla_`. It was definitely useful for ones in `settings/general`, but now as everything is scoped under `/google-for-woocommerce`, it is a bit redundant. Also, as we already rebranded, we could use `g4w_` prefixes. I decided to keep it that way to make the old entries the same, but I'm open to opinions and to change it.
5. The structure of the minimum free order threshold is a bit different than in the initial Google Doc proposal. I kept it here like our internal structure to avoid glue-code for mapping it back and forth, therefore avoiding superficial room for bugs.  Google confirmed they are Ok with it (p1743358224845469-slack-C02BB3F30TG)

### Changelog entry

> Add - `/settings/google-for-woocommerce` endpoint.
> Remove - `gla_` entries from `/settings/general`.

### DIFF
--- a/src/DB/Query/ShippingRateQuery.php
+++ b/src/DB/Query/ShippingRateQuery.php
@@ -67,4 +67,27 @@ class ShippingRateQuery extends Query {
 			$this->results
 		);
 	}
+
+	/**
+	 * Get all shipping rates.
+	 *
+	 * @since 2.8.0
+	 *
+	 * @return array
+	 */
+	public function get_all_shipping_rates() {
+		$rates = $this->get_results();
+		$items = [];
+		foreach ( $rates as $rate ) {
+			$data = [
+				'country_code'            => $rate['country'],
+				'currency'                => $rate['rate'],
+				'free_shipping_threshold' => $rate['options']['free_shipping_threshold'] ?: $rate['options']['free_shipping_threshold'],
+			];
+
+			$items[ $rate['country'] ] = $data;
+		}
+
+		return $items;
+	}
 }
--- a/src/DB/Query/ShippingRateQuery.php
+++ b/src/DB/Query/ShippingRateQuery.php
@@ -81,7 +81,8 @@ class ShippingRateQuery extends Query {
 		foreach ( $rates as $rate ) {
 			$data = [
 				'country_code'            => $rate['country'],
-				'currency'                => $rate['rate'],
+				'currency'                => $rate['currency'],
+				'rate'                    => $rate['rate'],
 				'free_shipping_threshold' => $rate['options']['free_shipping_threshold'] ?: $rate['options']['free_shipping_threshold'],
 			];
 
--- a/src/DB/Query/ShippingRateQuery.php
+++ b/src/DB/Query/ShippingRateQuery.php
@@ -90,7 +90,7 @@ class ShippingRateQuery extends Query {
 			$data = [
 				'country_code'            => $rate['country'],
 				'currency'                => $rate['currency'],
-				'free_shipping_threshold' => $rate['options']['free_shipping_threshold'] ? $rate['options']['free_shipping_threshold'] : null,
+				'free_shipping_threshold' => isset( $rate['options']['free_shipping_threshold'] ) ? (float) $rate['options']['free_shipping_threshold'] : null,
 				'rate'                    => $rate['rate'],
 			];
 
--- a/src/DB/Query/ShippingRateQuery.php
+++ b/src/DB/Query/ShippingRateQuery.php
@@ -69,11 +69,19 @@ class ShippingRateQuery extends Query {
 	}
 
 	/**
-	 * Get all shipping rates.
+	 * Retrieves all available shipping rates and organizes them by country.
 	 *
-	 * @since 2.8.0
+	 * Fetches all shipping rate data using `get_results()`, processes it,
+	 * and returns an associative array in a format suitable for JSON responses.
 	 *
-	 * @return array
+	 * @since x.x.x
+	 *
+	 * @return array Associative array of shipping rates indexed by country codes.
+	 *               Each entry contains:
+	 *               - `country_code` (string): The country code.
+	 *               - `currency` (string): The currency code.
+	 *               - `free_shipping_threshold` (float | null): The minimum order amount required for free shipping.
+	 *               - `rate` (string): The cost of shipping in the respective currency.
 	 */
 	public function get_all_shipping_rates() {
 		$rates = $this->get_results();
@@ -82,8 +90,8 @@ class ShippingRateQuery extends Query {
 			$data = [
 				'country_code'            => $rate['country'],
 				'currency'                => $rate['currency'],
+				'free_shipping_threshold' => $rate['options']['free_shipping_threshold'] ? $rate['options']['free_shipping_threshold'] : null,
 				'rate'                    => $rate['rate'],
-				'free_shipping_threshold' => $rate['options']['free_shipping_threshold'] ?: $rate['options']['free_shipping_threshold'],
 			];
 
 			$items[ $rate['country'] ] = $data;
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -268,7 +268,7 @@ class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
 				$data[] = [
 					'id'         => 'gla_settings_placeholder',
 					'option_key' => 'gla_settings_placeholder',
-					'type'       => '_invalid_type_'
+					'type'       => '_invalid_type_',
 				];
 
 				/*
@@ -278,7 +278,6 @@ class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
 				return $data;
 			}
 		);
-
 	}
 
 	/**
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -3,6 +3,7 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Integration;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\ShippingRateQuery;
 use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\ShippingTimeQuery;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Registerable;
 use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
@@ -29,6 +30,13 @@ defined( 'ABSPATH' ) || exit;
 class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
 
 	use OptionsAwareTrait;
+
+	/**
+	 * The ShippingRateQuery object.
+	 *
+	 * @var ShippingRateQuery
+	 */
+	protected $shipping_rate_query;
 
 	/**
 	 * The ShippingTimeQuery object.
@@ -60,10 +68,12 @@ class WPCOMProxy implements Service, Registerable, OptionsAwareInterface {
 	/**
 	 * WPCOMProxy constructor.
 	 *
+	 * @param ShippingRateQuery $shipping_rate_query   The ShippingRateQuery object.
 	 * @param ShippingTimeQuery $shipping_time_query The ShippingTimeQuery object.
 	 * @param AttributeManager  $attribute_manager   The AttributeManager object.
 	 */
-	public function __construct( ShippingTimeQuery $shipping_time_query, AttributeManager $attribute_manager ) {
+	public function __construct( ShippingRateQuery $shipping_rate_query, ShippingTimeQuery $shipping_time_query, AttributeManager $attribute_manager ) {
+		$this->shipping_rate_query = $shipping_rate_query;
 		$this->shipping_time_query = $shipping_time_query;
 		$this->attribute_manager   = $attribute_manager;
 	}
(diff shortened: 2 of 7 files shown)


src/DB/Query/ShippingRateQuery.php
```diff
@@ -67,4 +67,27 @@ function ( $row ) {
 			$this->results
 		);
 	}
+
+	/**
+	 * Get all shipping rates.
+	 *
+	 * @since 2.8.0
```
mikkamp: Should this be `@since x.x.x`?

src/DB/Query/ShippingRateQuery.php
```diff
@@ -67,4 +67,27 @@ function ( $row ) {
 			$this->results
 		);
 	}
+
+	/**
+	 * Get all shipping rates.
```
mikkamp: This description doesn't really define why it's needed over just calling `get_results`. As far as I can see the core part of this function is to format the shipping rates in a specific way and index them by country. If so then lets clarify that in this docblock.

I suppose it's a direct copy of `get_all_shipping_times`. But since we are adding it new, lets use the opportunity to clarify it's purpose.

src/DB/Query/ShippingRateQuery.php
```diff
@@ -67,4 +67,27 @@ function ( $row ) {
...
+	 * @return array
+	 */
+	public function get_all_shipping_rates() {
+		$rates = $this->get_results();
+		$items = [];
+		foreach ( $rates as $rate ) {
+			$data = [
+				'country_code'            => $rate['country'],
+				'currency'                => $rate['rate'],
+				'free_shipping_threshold' => $rate['options']['free_shipping_threshold'] ?: $rate['options']['free_shipping_threshold'],
```
mikkamp: Was this supposed to fallback to null or 0?

src/Integration/WPCOMProxy.php
```diff
@@ -169,26 +194,45 @@ function ( $response, $handler, $request ) {
...
+						'value' => rest_sanitize_boolean( $this->options->get( OptionsInterface::GOOGLE_CONNECTED, false ) ),
 					];
-
 					$data[] = [
-						'id'    => 'gla_shipping_times',
-						'label' => 'Google for WooCommerce: Shipping Times',
-						'value' => $this->shipping_time_query->get_all_shipping_times(),
+						'id'    => 'gla_jetpack_connected',
+						'label' => 'Google for WooCommerce: Is Jetpack connected?',
+						'value' => rest_sanitize_boolean( $this->options->get( OptionsInterface::JETPACK_CONNECTED, false ) ),
```
mikkamp: Was there a specific mention / request for this to be passed?

As far as I understand if it's not connected, then data can't be pulled from the API through WPCOM. So this value will always be true.

src/Options/Options.php
```diff
@@ -203,7 +203,7 @@ protected function raw_value( $value ) {
 	 *
 	 * @return string
 	 */
-	protected function prefix_name( string $name ): string {
+	public function prefix_name( string $name ): string {
```
mikkamp: Was there a specific intention to use this function outside the class? I'm not seeing it called anywhere, so we might as well leave it protected.

Edit: I saw it was mentioned in the additional section, but since it's not useful we might be better leaving it out.

src/Integration/WPCOMProxy.php
```diff
@@ -201,6 +245,48 @@ function ( $response, $handler, $request ) {
...
+				$locations[] = [
+					'id'          => self::SETTINGS_GROUP,
+					'label'       => 'Google for WooCommerce',
+					'description' => 'Settings of the Google for WooCommerce plugin.',
+				];
+				return $locations;
+			}
+		);
+
+		// Hack to make the settings group show up in the response.
```
mikkamp, line 261: Hope this is robust enough to continue working. I guess in theory we are modifying a versioned API, so it should bump version if it was to break compatibility.

mikkamp (approved):
Thanks for the changes, I tested it and I get the responses I would expect:

```json
[
    {
        "id": "gla_google_connected",
        "label": "Google for WooCommerce: Is Google account connected?",
        "value": true
    },
    {
        "id": "gla_jetpack_connected",
        "label": "Google for WooCommerce: Is Jetpack connected?",
        "value": true
    },
    {
        "id": "gla_language",
        "label": "Google for WooCommerce: Store language",
        "value": "en_US"
    },
    {
        "id": "gla_merchant_center",
        "label": "Google for WooCommerce: Merchant Center settings",
        "value": {
            "shipping_rate": "flat",
            "shipping_time": "flat",
            "tax_rate": "destination"
        }
    },
    {
        "id": "gla_shipping_rates",
        "label": "Google for WooCommerce: Shipping Rates",
        "value": {
            "IE": {
                "country_code": "IE",
                "currency": "10",
                "free_shipping_threshold": null
            }
        }
    },
    {
        "id": "gla_shipping_times",
        "label": "Google for WooCommerce: Shipping Times",
        "value": {
            "IE": {
                "country_code": "IE",
                "time": "3",
                "max_time": "3"
            }
        }
    },
    {
        "id": "gla_target_audience",
        "label": "Google for WooCommerce: Target Audience",
        "value": {
            "location": "selected",
            "countries": [
                "IE"
            ]
        }
    }
]
```

I left some comments, but none of them are really blocking so I'll go ahead and approve the PR.

> I'm not 100% sure if we should prefix ids with `gla_`. It was definitely useful for ones in settings/general, but now as everything is scoped under `/google-for-woocommerce`, it is a bit redundant. Also, as we already rebranded, we could use `g4w_` prefixes. I decided to keep it that way to make the old entries the same, but I'm open to opinions and to change it.

It is a bit redundant but either way should be fine, it's no harm to leave them in.

src/Options/Options.php
tomalec: Thanks for catching it, it's a leftover from an alternative implementation using only `woocommerce_settings-google-for-woocommerce` callback, which I ultimately discarded.

src/Integration/WPCOMProxy.php
tomalec: I think that in the worst case if the hack stops working, the `placeholder` setting will appear in the data set, which should not be a fatal error unless API consumers will expect only responses strictly matching schema.

src/Integration/WPCOMProxy.php
tomalec: There was no request. 
> I added gla_jetpack_connected and gla_google_connected entries. They were not mentioned in the shared Google doc, but I believe that it could be useful to Google to check if a merchant disconnected their account. Plus, those could be considered onboarding inputs.

I thought checking if Google account is connected could be useful to Google. And running the request to a local endpoint could be useful for debugging/introspection to all GLA settings aggregated.

But maybe, as per YAGNI, I'll remove them.

src/DB/Query/ShippingRateQuery.php
tomalec: I made it fall back to `null`, I can make it more explicit in the code.

src/Integration/WPCOMProxy.php
tomalec: Do you think we should keep `gla_google_connected` or remove it as well?

tomalec:
Thank you, @mikkamp, for the review!
I addressed your comments, and fixed currency code https://github.com/woocommerce/google-listings-and-ads/pull/2821/commits/f8c22de8a07dc99ee25a58af6dad1753f9d253b6
Do you think we should keep or remove `gla_google_connected`?

src/DB/Query/ShippingRateQuery.php
```diff
@@ -67,4 +67,36 @@ function ( $row ) {
...
+	 *               - `rate` (string): The cost of shipping in the respective currency.
+	 */
+	public function get_all_shipping_rates() {
+		$rates = $this->get_results();
+		$items = [];
+		foreach ( $rates as $rate ) {
+			$data = [
+				'country_code'            => $rate['country'],
+				'currency'                => $rate['currency'],
+				'free_shipping_threshold' => $rate['options']['free_shipping_threshold'] ? $rate['options']['free_shipping_threshold'] : null,
```
mikkamp: So this is where PHP is a bit tricky because this means you'll fallback to null for any "falsey" value.

But according to the schema here it allows a threshold value of 0: https://github.com/woocommerce/google-listings-and-ads/blob/2.9.10/src/API/Site/Controllers/ShippingRateSchemaTrait.php#L65

The previous code wasn't correct because it was using itself as the fallback:
```php
$rate['options']['free_shipping_threshold'] ?: $rate['options']['free_shipping_threshold']
```

Which means if `options` or `free_shipping_threshold` couldn't be indexed it would fallback to a value where it would also need those indexes available, which would produce a warning.

The code you have now:
```php
$rate['options']['free_shipping_threshold'] ? $rate['options']['free_shipping_threshold'] : null
```

Would still produce a warning if `options` or `free_shipping_threshold` couldn't be indexed in the array. And it would just be checking if the value evaluates to somewhat true, '0' would be false so it would revert to null.

```
[01-Apr-2025 11:25:41 UTC] PHP Warning:  Undefined array key "free_shipping_threshold" in google-listings-and-ads/src/DB/Query/ShippingRateQuery.php on line 93
```

I assume you'd want something like the following:
```suggestion
				'free_shipping_threshold' => isset( $rate['options']['free_shipping_threshold'] ) ? (float) $rate['options']['free_shipping_threshold'] : null,
```

Since you mention it in the docblock, I threw in the type cast to float. Generally all fields come from the DB as strings, but in this case `options` is unserialized from JSON so it should retain it's type. Shouldn't be any harm in enforcing it's type though.

src/Integration/WPCOMProxy.php
mikkamp: I think it's fine to leave it, not sure how it will be used, but the extra resource usage is insignificant if we leave it in. Unless you want to be strict on the YAGNI principle.

mikkamp (commented):
Thanks for the changes. Nice catch on the currency and rate being mixed up, I still had my Postman response open and I could see it was incorrect there, but I didn't notice it last time round.

I just added one main comment about the fallback for `free_shipping_threshold` as it's still giving me a warning.

budzanowski:
Potential Issues or Risks

  1. Prefix Consistency: As noted in the PR comments, there's a mix of prefixes
  (gla_ vs potential g4w_). While maintaining backward compatibility is important, a
   consistent approach for new attributes could be beneficial.
  3. Partial Implementation of Settings Group: The approach used to register the
  settings group includes a placeholder with an invalid type. While this works, it's
   a bit of a hack and might confuse other developers.

  Suggestions for Improvement

  1. Centralize Setting Keys: The setting IDs could be defined as constants in a
  central location to avoid string duplication.
  2. Edge Case Handling: Add validation for edge cases, such as when shipping data
  is malformed.

budzanowski:
This PR will be delivered as part of the combined effort in #2981, which includes both the WC REST API settings integration and sync push controls. Closing this individual PR in favor of the combined approach.

PR #2836, opened 2025-04-02
Controller for fetching incentive credits from Middleware

mikkamp:
This PR fetches incentive credits from Middleware at the API endpoint: `google/ads/incentive-credits`
We add the site controller IncentiveCreditsController with the endpoint: `/wp-json/wc/gla/ads/incentive-credits`

The extension can access both the Store country and the Ads account so these do not need to be passed to the site controller when fetching the incentive credits from the UI. We are only going to store one set of countries with a single matching currency. So we've put some fallbacks in place where it will try to pick up the right currency based on the following priority:

1. Ads account currency
2. Store currency
3. Fallback to first available currency

In either case it will append the `ads_currency` to the response, so the UI can use this to warn the user if the currencies don't match.

1. In order to simulate different responses it's best to test with a local Middleware, and setup the incentives as described in this PR: https://github.com/Automattic/woocommerce-connect-server/pull/2791
2. Make sure your store currency and ads account currency matches the value that was added in the Middleware 
3. Send a GET request to `/wp-json/wc/gla/ads/incentive-credits`
4. Expect it to return matching ads account and currency:
```json
{
    "country": "IE",
    "currency": "EUR",
    "ads_currency": "EUR",
    "spending": 400,
    "credit": 400
}
```
5. Change the store currency to a different value, and make sure the Middleware has an entry for that currency + country
6. Send a GET request to `/wp-json/wc/gla/ads/incentive-credits`
7. Expect it to return the other currency, ads_currency and currency should not match:
```json
{
    "country": "IE",
    "currency": "USD",
    "ads_currency": "EUR",
    "spending": 500,
    "credit": 500
}
```
8. change the Middleware entries and make sure it doesn't have any values for the ads currency or the store currency
9. Send a GET request to `/wp-json/wc/gla/ads/incentive-credits`
10. Expect it to return the currency from the Middleware as a fallback, ads_currency and currency should not match:
```json
{
    "country": "IE",
    "currency": "CAD",
    "ads_currency": "EUR",
    "spending": 600,
    "credit": 600
}
```

### Additional details:
Project: pcTzPl-2Hb-p2

Commit eebdca5 can be ignored when reviewing this PR. Since I had AI active writing unit tests for me I had it write some additional tests to provide full coverage for the Middleware class. They did need some adjusting, but it was a low effort to add them.

### Changelog entry
* Add - Controller for fetching incentive credits from Middleware.


src/API/Site/Controllers/Ads/IncentiveCreditsController.php
```diff
@@ -0,0 +1,124 @@
...
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'ads/incentive-credits',
+			[
+				[
+					'methods'             => 'GET',
```
eason9487: `'GET'` should be able to be replaced with `TransportMethods::READABLE` for better consistency.

eason9487 (approved):
Thanks for the work! Tested locally and works well.

There's a hardcoded string that could be replaced for better consistency. I'll push a commit later to adjust it.

src/API/Site/Controllers/Ads/IncentiveCreditsController.php
eason9487: Adjusted in https://github.com/woocommerce/google-listings-and-ads/pull/2836/commits/4c4c80343229f93b61ece6b31483e7f55a09b6d5.

PR #2838, opened 2025-04-03
Add support for getting price benchmarks & suggestions from the Content API 

### LINKED ISSUE #2825: Add support for getting price benchmarks & suggestions from the Content API
In order to implement the Price Benchmarks suggestions table described in #2824, we will need to have a REST API endpoint that returns the list of price benchmark suggestions for products in the store.

## Acceptance Criteria

- [ ] Price benchmarks can be fetched using a GET request to `/wp-json/wc/gla/mc/price-benchmarks/`
- [ ] Price benchmarks are not publicly accessible.
- [ ] Price benchmarks will return a list of object with the following fields:
  - Product
    - ID
    - Thumbnail
    - Title
  - Effectiveness
  - Regular price
  - Price on Google
  - Price gap
  - Suggested price
- [ ] Price benchmarks will return 10 items and allow additional items to be returned (pagination TBD).
- [ ] Error handling must be included and return an exception response

## Implementation Brief

- The REST API endpoint should be implemented in a new `PriceBenchmarks` controller in `/src/API/Site/Controllers/MerchantCenter`.
- `GET` requests should make use of the `BaseController::get_permission_callback()` to ensure this data remains private.
- A `get_price_benchmarks_callback()` should be used to get and prepare the API response, using the query classes described below.
- The query for benchmarks should be handled by a new class in `src/API/Google/Query` called `MerchantPriceBenchmarksQuery` that extends the `MerchantQuery` class. This class can be used to implement the query of `PriceCompetitivenessProductView`. See [these docs](https://developers.google.com/shopping-content/guides/reports/price-competitiveness).
- The query for price suggestions should be handled by a new class in `src/API/Google/Query` called `MerchantPriceSuggestionsQuery` that extends the `MerchantQuery` class. This class can be used to implement the query of `PriceInsightsProductView`. See [these docs](https://developers.google.com/shopping-content/guides/reports/price-insights).
- Additional product data like thumbnail, product title, etc. can come from the local Woo DB.
- The query classes should use the `OptionsAwareTrait` to be able to pass the merchant center ID to the query like `$this->options->get_merchant_id()`.

## Test Coverage

- Add PHPUnit test coverage for the return value of the GET callback.
- Add PHPUnit test coverage for any new service methods used to query Google APIs

mukeshpanchal27:

Closes #2825.

This PR add the support for getting price benchmarks & suggestions from the Content API.

- Comment the permission callback. Revert https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/e9f8b78f9a2e9c678088848ee50fd077ba976a25 changes in your local setup.
- Check the Rest API response. Call `siteurl/index.php?rest_route=/wc/gla/mc/price-benchmarks/` (It return empty array as MC account didn't have any approved products.)

### Changelog entry

>

### DIFF
--- a/src/API/Google/MerchantPriceBenchmarks.php
+++ b/src/API/Google/MerchantPriceBenchmarks.php
@@ -66,7 +66,7 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	 */
 	public function get_price_insights_product_view(): array {
 		try {
-			$response = ( new MerchantPriceSuggestionsQuery() )
+			$response = ( new MerchantPriceSuggestionsQuery( [] ) )
 			->set_client( $this->service, $this->options->get_merchant_id() )
 			->get_results();
 
--- a/src/API/Google/MerchantPriceBenchmarks.php
+++ b/src/API/Google/MerchantPriceBenchmarks.php
@@ -44,7 +44,7 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	 */
 	public function get_benchmark_data(): array {
 		try {
-			$response = ( new MerchantPriceBenchmarksQuery() )
+			$response = ( new MerchantPriceBenchmarksQuery( [] ) )
 			->set_client( $this->service, $this->options->get_merchant_id() )
 			->get_results();
 
--- a/src/API/Google/MerchantPriceBenchmarks.php
+++ b/src/API/Google/MerchantPriceBenchmarks.php
@@ -68,7 +68,7 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	 *
 	 * @throws Exception If the merchant price suggestions data can't be retrieved.
 	 */
-	public function get_price_insights_product_view( array $args ): array {
+	public function get_price_insights( array $args ): array {
 		try {
 			$response = ( new MerchantPriceSuggestionsQuery( $args ) )
 			->set_client( $this->service, $this->options->get_merchant_id() )
--- a/src/API/Google/MerchantPriceBenchmarks.php
+++ b/src/API/Google/MerchantPriceBenchmarks.php
@@ -1,0 +1,91 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query\MerchantPriceBenchmarksQuery;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query\MerchantPriceSuggestionsQuery;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent;
+use Exception;
+
+/**
+ * Class MerchantPriceBenchmarks
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Google
+ */
+class MerchantPriceBenchmarks implements OptionsAwareInterface {
+
+	use OptionsAwareTrait;
+
+	/**
+	 * The shopping service.
+	 *
+	 * @var ShoppingContent
+	 */
+	protected $service;
+
+	/**
+	 * Product helper class.
+	 *
+	 * @var ProductHelper
+	 */
+	protected $product_helper;
+
+	/**
+	 * Merchant Report constructor.
+	 *
+	 * @param ShoppingContent $service
+	 * @param ProductHelper   $product_helper
+	 */
+	public function __construct( ShoppingContent $service, ProductHelper $product_helper ) {
+		$this->service        = $service;
+		$this->product_helper = $product_helper;
+	}
+
+	/**
+	 * Get MerchantPriceBenchmarksQuery Query response.
+	 *
+	 * @return array Associative array with product statuses and the next page token.
+	 *
+	 * @throws Exception If the merchant price benchmarks report data can't be retrieved.
+	 */
+	public function get_benchmark_data(): array {
+		try {
+			$response = ( new MerchantPriceBenchmarksQuery() )
+			->set_client( $this->service, $this->options->get_merchant_id() )
+			->get_results();
+
+			$results = $response->getResults() ?? [];
+
+			return $results;
+		} catch ( GoogleException $e ) {
+			do_action( 'woocommerce_gla_mc_client_exception', $e, __METHOD__ );
+			throw new Exception( __( 'Unable to retrieve Merchant Price Benchmarks.', 'google-listings-and-ads' ) . $e->getMessage(), $e->getCode() );
+		}
+	}
+
+	/**
+	 * Get MerchantPriceSuggestions Query response.
+	 *
+	 * @return array Associative array with product statuses and the next page token.
+	 *
+	 * @throws Exception If the merchant price suggestions data can't be retrieved.
+	 */
+	public function get_price_insights_product_view(): array {
+		try {
+			$response = ( new MerchantPriceSuggestionsQuery() )
+			->set_client( $this->service, $this->options->get_merchant_id() )
+			->get_results();
+
+			$results = $response->getResults() ?? [];
+
+			return $results;
+		} catch ( GoogleException $e ) {
+			do_action( 'woocommerce_gla_mc_client_exception', $e, __METHOD__ );
+			throw new Exception( __( 'Unable to retrieve Merchant Price Benchmarks.', 'google-listings-and-ads' ) . $e->getMessage(), $e->getCode() );
+		}
+	}
+}
--- a/src/API/Google/MerchantPriceBenchmarks.php
+++ b/src/API/Google/MerchantPriceBenchmarks.php
@@ -7,7 +7,6 @@ use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query\MerchantPriceBe
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query\MerchantPriceSuggestionsQuery;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
-use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingContent;
 use Exception;
 
@@ -28,21 +27,12 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	protected $service;
 
 	/**
-	 * Product helper class.
-	 *
-	 * @var ProductHelper
-	 */
-	protected $product_helper;
-
-	/**
 	 * Merchant Report constructor.
 	 *
 	 * @param ShoppingContent $service
-	 * @param ProductHelper   $product_helper
 	 */
-	public function __construct( ShoppingContent $service, ProductHelper $product_helper ) {
-		$this->service        = $service;
-		$this->product_helper = $product_helper;
+	public function __construct( ShoppingContent $service ) {
+		$this->service = $service;
 	}
 
 	/**
--- a/src/API/Google/MerchantPriceBenchmarks.php
+++ b/src/API/Google/MerchantPriceBenchmarks.php
@@ -40,9 +40,9 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	 *
 	 * @param array $args Query arguments.
 	 *
-	 * @return array
+	 * @return array Associative array with price benchmarks data and the next page token.
 	 *
-	 * @throws Exception If the merchant price benchmarks report data can't be retrieved.
+	 * @throws Exception If the merchant price benchmarks data can't be retrieved.
 	 */
 	public function get_benchmark_data( array $args ): array {
 		try {
@@ -64,7 +64,7 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	 *
 	 * @param array $args Query arguments.
 	 *
-	 * @return array Associative array with product statuses and the next page token.
+	 * @return array Associative array with price benchmarks data and the next page token.
 	 *
 	 * @throws Exception If the merchant price suggestions data can't be retrieved.
 	 */
--- a/src/API/Google/MerchantPriceBenchmarks.php
+++ b/src/API/Google/MerchantPriceBenchmarks.php
@@ -38,13 +38,15 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	/**
 	 * Get MerchantPriceBenchmarksQuery Query response.
 	 *
-	 * @return array Associative array with product statuses and the next page token.
+	 * @param array $args Query arguments.
+	 *
+	 * @return array
 	 *
 	 * @throws Exception If the merchant price benchmarks report data can't be retrieved.
 	 */
-	public function get_benchmark_data(): array {
+	public function get_benchmark_data( array $args ): array {
 		try {
-			$response = ( new MerchantPriceBenchmarksQuery( [] ) )
+			$response = ( new MerchantPriceBenchmarksQuery( $args ) )
 			->set_client( $this->service, $this->options->get_merchant_id() )
 			->get_results();
 
@@ -60,13 +62,15 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
 	/**
 	 * Get MerchantPriceSuggestions Query response.
 	 *
+	 * @param array $args Query arguments.
+	 *
 	 * @return array Associative array with product statuses and the next page token.
 	 *
 	 * @throws Exception If the merchant price suggestions data can't be retrieved.
 	 */
-	public function get_price_insights_product_view(): array {
+	public function get_price_insights_product_view( array $args ): array {
 		try {
-			$response = ( new MerchantPriceSuggestionsQuery( [] ) )
+			$response = ( new MerchantPriceSuggestionsQuery( $args ) )
 			->set_client( $this->service, $this->options->get_merchant_id() )
 			->get_results();
 
--- a/src/API/Google/Query/MerchantPriceBenchmarksQuery.php
+++ b/src/API/Google/Query/MerchantPriceBenchmarksQuery.php
@@ -35,7 +35,7 @@ class MerchantPriceBenchmarksQuery extends MerchantQuery {
 	 */
 	public function filter( array $ids ): QueryInterface {
 		if ( ! empty( $ids ) ) {
-			$this->where( 'product_view.id', 'IN', $ids );
+			$this->where( 'product_view.id', $ids, 'IN' );
 		}
 		return $this;
 	}
--- a/src/API/Google/Query/MerchantPriceBenchmarksQuery.php
+++ b/src/API/Google/Query/MerchantPriceBenchmarksQuery.php
@@ -12,12 +12,18 @@ use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query\MerchantQuery;
  */
 class MerchantPriceBenchmarksQuery extends MerchantQuery {
 
+	use ReportQueryTrait;
+
 	/**
 	 * MerchantPriceBenchmarksQuery constructor.
+	 *
+	 * @param array $args Query arguments.
 	 */
-	public function __construct() {
+	public function __construct( array $args ) {
 		parent::__construct( 'PriceCompetitivenessProductView' );
+
 		$this->set_initial_columns();
+		$this->handle_query_args( $args );
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -0,0 +1,142 @@
...
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'mc/price-benchmarks',
+			[
+				[
+					'methods'  => TransportMethods::READABLE,
+					'callback' => $this->get_price_benchmarks_callback(),
```
mukeshpanchal27:
```suggestion
					'callback' => $this->get_price_benchmarks_callback(),
```

Permission callback will added once tested.

src/API/Google/MerchantPriceBenchmarks.php
```diff
@@ -0,0 +1,91 @@
...
+	 * @var ShoppingContent
+	 */
+	protected $service;
+
+	/**
+	 * Product helper class.
+	 *
+	 * @var ProductHelper
+	 */
+	protected $product_helper;
```
joemcgill: This isn't being used yet, but we may need to use it to convert the IDs returned from the API, similar to https://github.com/woocommerce/google-listings-and-ads/blob/925c3fb030447cdff4c10c0901c91a2ac300a282/src/API/Google/MerchantReport.php#L94

src/API/Google/Query/MerchantPriceBenchmarksQuery.php
```diff
@@ -0,0 +1,52 @@
...
+ * Class MerchantPriceBenchmarksQuery
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query
+ */
+class MerchantPriceBenchmarksQuery extends MerchantQuery {
+
+	/**
+	 * MerchantPriceBenchmarksQuery constructor.
+	 */
+	public function __construct() {
```
joemcgill: The constructor will need to handle `$args` so we can pass `next_page` and `per_page` args, like https://github.com/woocommerce/google-listings-and-ads/blob/925c3fb030447cdff4c10c0901c91a2ac300a282/src/API/Google/MerchantReport.php#L76-L81

src/API/Google/Query/MerchantPriceBenchmarksQuery.php
```diff
@@ -0,0 +1,52 @@
...
+	/**
+	 * Filter the query by a list of product IDs.
+	 *
+	 * @param array $ids List of product IDs to filter by.
+	 *
+	 * @return $this
+	 */
+	public function filter( array $ids ): QueryInterface {
+		if ( ! empty( $ids ) ) {
+			$this->where( 'product_view.id', 'IN', $ids );
```
joemcgill: Not sure that we really need to be able to filter by IDs, but even if so, the last two args in the `where()` method here look to be reversed:
```suggestion
			$this->where( 'product_view.id', $ids , 'IN' );
```

Same for the filter method in the `MerchantPriceSuggestionsQuery` object

src/API/Google/Query/MerchantPriceSuggestionsQuery.php
```diff
@@ -0,0 +1,56 @@
...
+	/**
+	 * Filter the query by a list of product IDs.
+	 *
+	 * @param array $ids List of product IDs to filter by.
+	 *
+	 * @return $this
+	 */
+	public function filter( array $ids ): QueryInterface {
+		if ( ! empty( $ids ) ) {
+			$this->where( 'product_view.id', 'IN', $ids );
```
joemcgill: See comment on `MerchantPriceBenchmarksQuery::filter()`

src/API/Google/Query/MerchantPriceSuggestionsQuery.php
```diff
@@ -0,0 +1,56 @@
...
+ * Class MerchantPriceSuggestionsQuery
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Query
+ */
+class MerchantPriceSuggestionsQuery extends MerchantQuery {
+
+	/**
+	 * MerchantPriceSuggestionsQuery constructor.
+	 */
+	public function __construct() {
```
joemcgill: See comment about constructor args from `MerchantPriceBenchmarksQuery`

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * @return callable
+	 */
+	protected function get_price_benchmarks_callback(): callable {
+		return function ( Request $request ) {
+			try {
+				/** @var MerchantPriceBenchmarks $merchant */
+				$merchant = $this->container->get( MerchantPriceBenchmarks::class );
+
+				$benchmark_data = $merchant->get_benchmark_data( 'products', $this->prepare_query_arguments( $request ) );
```
joemcgill: This method doesn't currently accept args. I assume it's just not implemented yet.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -0,0 +1,142 @@
...
+				$benchmark_data = $this->prepare_item_for_response( $benchmark_data, $request );
+
+				$get_price_insights_product_view = $merchant->get_price_insights_product_view( 'products', $this->prepare_query_arguments( $request ) );
+				$get_price_insights_product_view = $this->prepare_item_for_response( $get_price_insights_product_view, $request );
+
+				return new Response(
```
joemcgill: Once we have working data coming back from these responses, we'll need to stitch together data from both queries by matching the product IDs in order to return the expected response shape

joemcgill (commented):
@mukeshpanchal27 this is looking pretty good so far. Left some initial feedback for you.

src/API/Google/MerchantPriceBenchmarks.php
mukeshpanchal27: Remove it in https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/e82483c320fbff97bb47354a914223432d441fc8

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Addressed in https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/62c4d22ddf74481f575b2f9a28268dc9686537f4

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Yes, since we’re not getting any data for now, I’ve posted both responses in an object.

What do I need to set up to retrieve data from the merchant account, and in what format should we return the object data?

src/API/Google/MerchantPriceBenchmarks.php
```diff
@@ -27,34 +26,27 @@ class MerchantPriceBenchmarks implements OptionsAwareInterface {
...
+		$this->service = $service;
 	}
 
 	/**
 	 * Get MerchantPriceBenchmarksQuery Query response.
 	 *
-	 * @return array Associative array with product statuses and the next page token.
+	 * @param array $args Query arguments.
+	 *
+	 * @return array
```
joemcgill: We'll want to add a description here

src/API/Google/MerchantPriceBenchmarks.php
```diff
@@ -70,13 +62,15 @@ public function get_benchmark_data(): array {
...
 	 * Get MerchantPriceSuggestions Query response.
 	 *
+	 * @param array $args Query arguments.
+	 *
 	 * @return array Associative array with product statuses and the next page token.
 	 *
 	 * @throws Exception If the merchant price suggestions data can't be retrieved.
 	 */
-	public function get_price_insights_product_view(): array {
+	public function get_price_insights_product_view( array $args ): array {
```
joemcgill: Naming suggestion:
```suggestion
	public function get_price_insights( array $args ): array {
```

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -69,10 +69,10 @@ protected function get_price_benchmarks_callback(): callable {
 				/** @var MerchantPriceBenchmarks $merchant */
 				$merchant = $this->container->get( MerchantPriceBenchmarks::class );
 
-				$benchmark_data = $merchant->get_benchmark_data( 'products', $this->prepare_query_arguments( $request ) );
+				$benchmark_data = $merchant->get_benchmark_data( $this->prepare_query_arguments( $request ) );
 				$benchmark_data = $this->prepare_item_for_response( $benchmark_data, $request );
 
-				$get_price_insights_product_view = $merchant->get_price_insights_product_view( 'products', $this->prepare_query_arguments( $request ) );
+				$get_price_insights_product_view = $merchant->get_price_insights_product_view( $this->prepare_query_arguments( $request ) );
 				$get_price_insights_product_view = $this->prepare_item_for_response( $get_price_insights_product_view, $request );
```
joemcgill: We'll want to rename this as well

```suggestion
				$price_insights_data = $merchant->get_price_insights( $this->prepare_query_arguments( $request ) );
				$price_insights_data = $this->prepare_item_for_response( $get_price_insights_product_view, $request );
```

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
joemcgill: The docs for both Content API endpoints are linked from the issue description:

* [Price Benchmarks](https://developers.google.com/shopping-content/guides/reports/price-competitiveness) (i.e., price competitiveness)
* [Price Suggestions](https://developers.google.com/shopping-content/guides/reports/price-insights) (i.e., price insights)

Since you don't currently have access to a store that has this data available, I would suggest setting up PHPUnit tests that mock the responses coming back from those queries to match the documentation and use that to write the logic for the response. There are some examples of how simlar API controllers are being mocked in `tests/Unit/API/Site/Controllers/MerchantCenter/` for reference.

The shape of the response from our API endpoint should match what is in the issue description:

```
- Product
  - ID: Woo product ID (the priceBenchmarks query should return this info as `product.offer_id`
  - Thumbnail: URL that will come from the Woo DB base on the ID
  - Title: Product Title
- Effectiveness: priceInsights.effectiveness (this should be part of the response from the `MerchantPriceSuggestionsQuery` based on [these docs](https://developers.google.com/shopping-content/reference/rest/v2.1/reports/search#PriceInsights))
- Regular price (productView.priceMicros)
- Price on Google (priceCompetitiveness.benchmarkPriceMicros)
- Price gap (will need to be calculated by comparing `priceMicros` and `benchmarkPriceMicros` from the `MerchantPriceBenchmarksQuery` response
- Suggested price: `priceInsights.suggestedPriceMicros`
```

src/API/Google/Query/MerchantPriceBenchmarksQuery.php
```diff
@@ -0,0 +1,58 @@
...
+		return $this;
+	}
+
+	/**
+	 * Set the initial columns for this query.
+	 */
+	protected function set_initial_columns() {
+		$this->columns(
+			[
+				'id'                                 => 'product_view.id',
```
joemcgill: I think we will want to use `product_view.offer_id` instead of the MC product id based on my understanding of [these docs](https://developers.google.com/shopping-content/guides/reports/fields#pricecompetitivenessproductview). To be safe, lets grab both
```suggestion
				'id'                                 => 'product_view.id',
				'offer_id'                           => 'product_view.offer_id',
```

src/API/Google/Query/MerchantPriceSuggestionsQuery.php
```diff
@@ -0,0 +1,62 @@
...
+		return $this;
+	}
+
+	/**
+	 * Set the initial columns for this query.
+	 */
+	protected function set_initial_columns() {
+		$this->columns(
+			[
+				'id'                                    => 'product_view.id',
```
joemcgill, line 49: Let's add offer_id here too
```suggestion
				'id'                                    => 'product_view.id',
				'offer_id'                              => 'product_view.offer_id',
```

src/API/Google/Query/MerchantPriceSuggestionsQuery.php
```diff
@@ -0,0 +1,62 @@
...
+				'title'                                 => 'product_view.title',
+				'brand'                                 => 'product_view.brand',
```
joemcgill: We're not using this info
```suggestion
```

joemcgill (changes requested):

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: @joemcgill I tried adding the unit tests, but I couldn’t get it working maybe I missed something while mocking the service or the controller. I found it a bit tricky to set up.  

Could you please share some guidance or pointers so I can work on it more effectively?

mukeshpanchal27: Try to add unit test in https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/49f7823455d72be36a4bcf988e6795dd2e60369b

joemcgill: You got pretty close! I've updated the test class in [c7e917a](https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/c7e917af0ae6583ab41d5b6a0ef640ca922277d5) to fix the setUp method and to get a better idea of what the expected shape should be.

The `PriceBenchmarksController` makes use of the `MerchantPriceBenchmarks` service class by getting it from the Container, so we needed to set up a mock Container and share a local `MerchantPriceBenchmarks` class that we then mock return values for.

joemcgill:
@mukeshpanchal27 I've pushed an update for the unit tests and left more details in the conversation above: https://github.com/woocommerce/google-listings-and-ads/pull/2838#discussion_r2034157689

src/API/Google/MerchantPriceBenchmarks.php
mukeshpanchal27: Updated in https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/5f417b39c8bced7caf82e43c24a7c469ff5824bb

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: @joemcgill In https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/4833d054b89d9ec06ac932055cc0824dcc4a3f75 i shape the response but i don't get why the unit tests going failed for `$this->product_helper->get_wc_product_id` as when i run in browser with mock data it return proper ID.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -0,0 +1,231 @@
...
+		if ( empty( $benchmark_data['results'] ) || empty( $price_insights_data['results'] ) ) {
+			return $mapped_data;
+		}
+		// Loop through the benchmark data results.
+		foreach ( $benchmark_data['results'] as $benchmark_result ) {
+			$product_view          = $benchmark_result['productView'];
+			$price_competitiveness = $benchmark_result['priceCompetitiveness'];
+
+			// Find the matching price insights data by offer_id.
+			foreach ( $price_insights_data['results'] as $price_insights_result ) {
```
joemcgill: Rather than the nested loops here, let's loop through both data types independently and add them the the `$mapped_data` array keyed to the product ID, that way we end up with a single array that should contain a list like:

```php
$mapped_data = [
  '123456' => [
    'product_view' => [] // array from benchmarks
    'price_competitiveness' => [] // array from benchmarks
    'price_insights' => [] // array from price insights data
  ],
]
```

Otherwise, we're looping through the same `$price_insights_data` list for each item in the benchmark data list.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -0,0 +1,231 @@
...
+
+				$price_insights = $price_insights_result['priceCompetitiveness'] ?? [];
+
+				// Calculate the price gap.
+				$regular_price   = (int) $product_view['priceMicros'];
+				$price_on_google = (int) $price_competitiveness['benchmarkPriceMicros'];
+				$price_gap       = $regular_price - $price_on_google;
+
+				// Get the WooCommerce product ID and thumbnail.
+				$product_id = $this->product_helper->get_wc_product_id( $product_view['id'] );
```
joemcgill: Great idea to use this helper, but it doesn't seem to be working, which is causing the unit tests to fail. It looks like we should be able to just use the $product_view['offer_id'] value directly here.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -0,0 +1,231 @@
...
+				$thumbnail  = $this->get_product_thumbnail( $product_id );
+
+				// Map the data to the required format.
+				$mapped_data[] = [
```
joemcgill: After implementing the previous suggested change to the nested loops, this will need to be updated so that we're looping through the mapped data to return the expected responses.

```
$response = array_map( function ( $item ) { // etc. }, $mapped_data );
```

joemcgill (commented):
Almost there! Left a few more suggestions for you

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: I really missed this performance improvement, I want to make sure it gets across the finish line. I was stretched a bit last night and forgot to fix this one, but I’ve updated it now.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Per the response of the queries it will not return `offer_id`. I have introduce the util function similar to `ProductHelper::get_wc_product_id` now it working in codebase and unit tests.

[Price Benchmarks](https://developers.google.com/shopping-content/guides/reports/price-competitiveness) (i.e., price competitiveness)
[Price Suggestions](https://developers.google.com/shopping-content/guides/reports/price-insights) (i.e., price insights)

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Update in https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/d6fd66dc2b3f1cf76e99fe0cb404689796b0c7fc

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Added in https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/e9f8b78f9a2e9c678088848ee50fd077ba976a25

mukeshpanchal27:
@joemcgill I addressed all your feedback related to codebase and unit tests. Could you please take a look.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
joemcgill: The example in that documentation doesn't include the `product_view.offer_id` column in the example query, but we've included it in our query [here](https://github.com/woocommerce/google-listings-and-ads/blob/1edfffb4daab0135d40da32013276f513eda0eac/src/API/Google/Query/MerchantPriceBenchmarksQuery.php#L50), so the response should return this value.

In fact, as I was reviewing the way these `MerchantQuery` classes work, I think that when we set up the values in `set_initial_columns()`, we're actually setting the shape of the object that we'll have returned from the query results, so I've updated the mocks with this expectation and pushed some changes in [f1c0c9e](https://github.com/woocommerce/google-listings-and-ads/pull/2838/commits/f1c0c9e75c87806513b2a9d8b0eb8e5790e588c6) that update the logic in the `PriceBenchmarksController` to match.

I'm not positive about this since I can't yet get actual query results back from the Content API for my test store, so we'll need to confirm and adjust if needed.

mukeshpanchal27: Can we go ahead and merge these PR? Once we get data from MC store and if any adjustment needed we can open follow-up PR. WDYT?

joemcgill (approved):
I think we've addressed all that we're able to here without some clarity on the way the Content API will return this data. I'll follow up on getting us some better testing data but let's go ahead and merge this and we can iterate in a follow-up PR if needed.

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.7.2
* Theme:  
  - TT5 Version: 1.1
* WooCommerce - Version 9.8.1
* PHP: 8.3.0
* Web Server: Nginx
* Browser: Chrome - Version 135
* OS: macOS Sonoma 14.6.1 

</Details>

Test Result: While testing this, a new scenario was observed and reported here separately to address https://github.com/woocommerce/google-listings-and-ads/issues/2852

Issue #2852, opened 2025-04-14
Handle scenario when Merchant Center (MC) account lacks ProductInsight data on newly created accounts


ankitguptaindia:
While testing [[PR #2838](https://github.com/woocommerce/google-listings-and-ads/pull/2838)], it was observed that the API response returns the following message for newly created Merchant Center (MC) accounts:

> _"Market Insights not enabled for account xxxxx"_

This scenario needs to be handled, especially for users who are setting up the plugin with a new MC account. Currently, the absence of ProductInsight data causes an unhandled response, which could confuse the user or block further setup steps.

**Steps to Reproduce:**
1. Create a new Google Merchant Center account.
2. Set up the Google Listings & Ads plugin.
3. Follow test steps from https://github.com/woocommerce/google-listings-and-ads/pull/2838#issue-2969322095

**Screenshot:**  
![Image](https://github.com/user-attachments/assets/f7a88d0c-09d0-4487-afbf-985ad2ba61ef)

joemcgill:
Thanks, @ankitguptaindia. I think we can handle this error in the front end while working on #2831. @asvinb can you add this to the requirements for that issue and we can close this as a duplicate?

asvinb:
@joemcgill #2831 has been updated to cater for the error in the API response. I think we can close the current ticket.

joemcgill:
Closing as resolved in #2831.

Issue #2829, opened 2025-03-28
`ChangePriceModal` component


joemcgill:
## User story
Part of https://github.com/woocommerce/google-listings-and-ads/issues/2824

As a merchant reviewing price benchmarks and suggestions, I want an easy way to modify the price of a product directly from the table, so I don't have to navigate away to the product page.

![Image](https://github.com/user-attachments/assets/246fc119-c782-44fa-a0c4-f18c3b5b6400)

## Figma link

https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings---Ads-v2.x?node-id=11507-119316&t=q4ExrxQejYamkY9P-4

## Acceptance criteria

- [ ] Clicking the "Change Price" button (within `PriceBenchmarkSuggestionsTable`) should render a modal where users can change the regular price of a product.
- [ ] If the price update has been successful, the data within the DataViews, i.e the `PriceBenchmarkSuggestionsTable` should be updated to reflect the new price.
- [ ] Users should not be able to set a price < 0  or less than the sales price.

## Implementation brief

* Create the `ChangePriceModal` component.
	* Add `js/src/pages/price-benchmark/change-price-modal/index.js` to house the `ChangePriceModal` component which accepts the following props:
		* `ID`: the product ID
		* `priceBenchmarkInfo`: object containing properties we need to render in the modal. Those props should be supplied by item being rendered in the price benchmark table.
		* `onPriceChange`: event that get fired when the price change has been successful. It should accept the following parameters:
			* `ID`: the product ID
			* `new_price`: the updated price
	* It should use the `AppModal` component to render the modal.
		* The buttons should be set via the `buttons` prop and also to render the `AppInputPriceControl` component to match the designs.
			* `Change Price`: primary
* Render the modal (data wise) as per the designs.
* The `AppInputPriceControl` component should be used for the price input.
	* The price should be validated so that it's `>= 0` and greater than the sale price (if any).
	* The currency can be obtained via the `currency` property of the `useGoogleAdsAccount` hook. The currency should then passed to the `currency` prop of `AppInputPriceControl`.
	* The value should be pre set to the suggested price (from the API).
* If the form is valid, clicking the "Change Price" button should:
	* fire a PUT request to the `/wc/v3/products/<productID>` endpoint passing the `id` and `regular_price` values.
	* if request is successful, call the `onPriceChange` prop with the required parameters (see above).
* The following components should be used:
[...]

PR #2853, opened 2025-04-14
Add PriceBenchmarkSuggestions component

### LINKED ISSUE #2828: `PriceBenchmarkSuggestionsTable` component
## User story

As a merchant, I want to be able to review individual price benchmark data for any qualified products in my store and (eventually) take action to make price adjustments from the same UI.

Price benchmark suggestions table screenshot

## Figma link
https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba/Google-Listings-%26-Ads-v2.x?node-id=10866-63040&t=UAZY05dsxzVwJuu2-0

## Acceptance criteria

- [ ] The `PriceBenchmarkSuggestionsTable` component should list products that are priced more expensive/cheaper than competitors, and price suggestions
- [ ] The table should show 10 products per page.
- [ ] Clicking the "Change Price" button should not do anything for now.

## Implementation brief
* Create the `PriceBenchmarkSuggestionsTable` component.
	* Add `js/src/pages/price-benchmark/price-benchmark-suggestions-table/index.js` to house the `PriceBenchmarkSuggestionsTable` component.
* Update the data store
	* Add the `getPriceBenchmarkSuggestions` selector and resolver to fetch the suggestions data.
		* Consider adding a parameter to the selector/resolver to fetch per paged data with a string as index.
* The data should be formatted to be used with WordPress' [DataViews component](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-dataviews/).
	* Update `webpack.config.js` to not externalize the `@wordpress/dataviews/wp` package via the `requestToExternal` function of the `WooCommerceDependencyExtractionWebpackPlugin` plugin.
	* For monetary values, they should be formatted using the `formatAmount` function from the `useAdsCurrency` hook.
	* For table headers with tooltips, the `AppTooltip` component should be used.
		* Table headers markup can be altered via the `render` function of the field.
	* Any further calculations TBD (will all the data come from the API as is or the calculation will be on the FE?)
	* The `titleField`, `descriptionField` and `mediaField` should be mapped to the appropriate fields from the datastore/API to render the "card" within the first column.
	* On small screens, only the product and actions fields are visible.
		* The `fields` property should be modified to exclude fields that are not meant to be displayed on mobile.
* The following sub components should be added, which will be used when rendering the data
	* `EffectivenessIndicator`
		* Accepts the following props:
			* `effectiveness` which can be the following values:
				* `low`
				* `medium`
				* `high`
		* It should use the WordPress `Badge` [component](https://wordpress.github.io/gutenberg/?path=/docs/components-badge--docs&args=orientation:horizontal).
	* `ChangePrice`
		* Accepts the following props:
			* `productID`
			* `label` with default value set to `Change Price`
		* The implementation about displaying the modal will be done in another ticket.
* Update `PriceBenchmarkSuggestions` to include the `PriceBenchmarkSuggestionsTable` component for the "Price Benchmark & Suggestions" tab.

## Out of bounds/rabbit holes

- The REST API endpoint that will return the price benchmarks suggestion data will be implemented in #2825. Unless already available when this issue is worked on, displaying modeled data should be fine and integration with the API can be handled in a follow-up issue.
- Pagination and sorting will be handled in #2840 .

## Test coverage
* Appropriate E2E tests should be added for the following features:
	* not all fields are rendered on small screens

asvinb:

Closes #2828  .

### Screenshots:
image

1. Checkout the `feature/2824-price-benchmarks` branch locally.
	* `git checkout feature/2824-price-benchmarks`
2. Run `npm install` at the root folder of the plugin to install new packages. In this case, it's the `@wordpress/dataviews` package.
3. Run `npm run build`
4. Navigate to the Price Benchmark tab.
5. Verify that the "Price Benchmark & Suggestions" table displays data correctly.
6. In mobile view, ensure that only two columns are visible: "Product" and "Action".
7. Confirm that the "Change Price" button is non-functional and currently serves as a placeholder.
8. Ensure the UI implementation matches the provided designs in terms of layout, styling, and responsiveness.

### Changelog entry

>


js/src/components/badge/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Internal dependencies
+ */
+import './index.scss';
+
+/**
+ * Badge component.
+ *
+ * A reusable badge component that displays content with optional styling based on intent.
```
joemcgill, line 14: Could add some inline docs to this mentioning that it's a temporary port of https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/badge until that package is public

js/src/data/actions.js
```diff
@@ -1225,3 +1225,27 @@ export function* fetchPriceBenchmarkSummary() {
...
+		const data = yield apiFetch( {
+			path: `${ API_NAMESPACE }/mc/price-benchmarks`,
+		} );
+
+		return {
+			type: TYPES.RECEIVE_PRICE_BENCHMARK_SUGGESTIONS,
+			data,
+		};
+	} catch ( error ) {
+		handleApiError(
```
joemcgill, line 1243: We'll probably need to add some handling here later for #2852.

js/src/pages/price-benchmark/effectiveness-indicator.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import {
+	EFFECTIVENESS_LOW,
```
joemcgill, line 11: Will need to add an option for `EFFECTIVENESS_UNSPECIFIED`

js/src/pages/price-benchmark/price-benchmark-suggestions.js
```diff
@@ -1,63 +1,120 @@
...
+ * PriceBenchmarkSuggestions component.
+ *
+ * This component fetches and displays price benchmark suggestions using the
+ * `usePriceBenchmarkSuggestions` hook. It renders a table with the suggestions
+ * data and handles responsiveness by providing different field configurations
+ * for desktop and mobile views.
+ *
+ * @return {JSX.Element} A div containing the PriceBenchmarkTable component.
+ */
 const PriceBenchmarkSuggestions = () => {
```
joemcgill, line 121: Will this component also render the MetricsNotice?

joemcgill (commented):
@asvinb I took a look through the WIP here and overall the setup seems good to me. I left a few question/comments but think you're on the right track

js/src/components/badge/index.js
asvinb: Sure. Added comment.

js/src/pages/price-benchmark/effectiveness-indicator.js
asvinb: Option added.

js/src/pages/price-benchmark/price-benchmark-suggestions.js
asvinb: No, it's the `PriceBenchmarkTable` component which will render `MetricsNotice`.

joemcgill: In that case, I wonder if we can remove one level of nesting here and move all of the fields setup directly into the `PriceBenchmarkTable` component?

js/src/pages/price-benchmark/price-benchmark-table/index.js
```diff
@@ -0,0 +1,151 @@
...
+
+	const { data: shownData, paginationInfo } = useMemo( () => {
+		const updatedData = filterSortAndPaginate( data, view, fields );
+
+		return updatedData;
+	}, [ view, data, fields ] );
+
+	const handleOnChangeView = ( newView ) => {
+		setView( newView );
+	};
```
joemcgill: Consider memoizing this callback

```
const handleOnChangeView = useCallback((newView) => {
  setView(newView);
}, []);
```

js/src/pages/price-benchmark/price-benchmark-suggestions.js
```diff
@@ -1,63 +1,120 @@
...
+	const { suggestions, hasFinishedResolution } =
+		usePriceBenchmarkSuggestions();
```
joemcgill, line 123: Should we set a default value here?

```suggestion
	const { suggestions = [], hasFinishedResolution } =
		usePriceBenchmarkSuggestions();
```

js/src/pages/price-benchmark/price/index.js
```diff
@@ -0,0 +1,34 @@
...
+const Price = ( { amount, highlight = false } ) => {
+	const { formatAmount } = useAdsCurrency();
+
+	return (
+		<span
+			className={ classnames( 'gla-price-benchmark-table__price', {
+				'gla-price-benchmark-table__price--highlight': highlight,
+			} ) }
+		>
+			{ formatAmount( amount ) }
```
joemcgill: Should we validate that `amount` is a number and if a null/empty value is passed force it to a zero?

```js
const formattedAmount = amount ? formatAmount(amount) : formatAmount(0);
```

js/src/pages/price-benchmark/price-benchmark-table/index.js
joemcgill: Additionally, it doesn't look like this updates the URL structure so that if the page is refreshed you always go back to the first page with default filters applied. For example: go to page 2 and refresh, order by price and refresh, etc.

tests/e2e/specs/price-benchmark/price-benchmark-suggestions.test.js
```diff
@@ -7,6 +7,7 @@ import { expect, test } from '@playwright/test';
  * Internal dependencies
  */
 import { clearOnboardedMerchant } from '../../utils/api';
+import priceBenchmarlSuggestionsData from '../../utils/__fixtures__/price-benchmark-suggestions.json';
```
joemcgill: Typo

```suggestion
import priceBenchmarkSuggestionsData from '../../utils/__fixtures__/price-benchmark-suggestions.json';
```

tests/e2e/utils/__fixtures__/price-benchmark-suggestions.json
```diff
@@ -0,0 +1,167 @@
+[
+	{
```
joemcgill, line 2: The shape of these stubs don't match the expected shape being returned by the API. See: https://github.com/woocommerce/google-listings-and-ads/blob/a91842f08f41a88854ab9ef7549ae9c1256d5284/tests/Unit/API/Site/Controllers/MerchantCenter/PriceBenchmarksControllerTest.php#L96-L109

For this to integrate with the REST API, we need to align on an expected schema.

joemcgill (changes requested):
@asvinb I've left a round of feedback on this PR. The most important things that need to be addressed are:

1. Ensuring we're using the expected shape from the REST API endpoint for price-benchmarks
2. Resolving the bundle watch issues.

For the latter, could we update the webpack config so the dataviews package isn't bundled into the `vendors.js` bundle? We will probably need to enqueue it specifically for this page once we do so the bundle isn't loaded on other pages where it is not being used. 

Even then, the bundle size might still be too large, so we will likely need to figure out how to allow the max size for that package to be above the current threshold.

js/src/pages/price-benchmark/price-benchmark-suggestions.js
asvinb: I originally had it like that @joemcgill :
https://github.com/woocommerce/google-listings-and-ads/pull/2853/commits/c6020d5173e9ada06db62ad856e7c59a017b29a1
But then i thought the `PriceBenchmarkTable` should only contain fields that should be part of the basic table. Then any components using `PriceBenchmarkTable` should pass their extra fields. I feel that can be more flexible if we have different table types in the future. Let me know what you think. Happy to go with your suggestion as well 😁

js/src/pages/price-benchmark/price-benchmark-suggestions.js
asvinb: @joemcgill I feel, in that case, it's best we set the default value in `PriceBenchmarkTable` which I just did. Then when we need to render the `EmptyMetricsNotice` component, we can check if `data.length === 0` and `isReady === true` in `PriceBenchmarkTable`

js/src/pages/price-benchmark/price-benchmark-table/index.js
asvinb: @joemcgill I was not aware we needed to keep table interactions persistent 🤔  Maybe we need a separate ticket for that?

tests/e2e/utils/__fixtures__/price-benchmark-suggestions.json
asvinb: Sounds good. Updated @joemcgill . I thought we were going to have a different format when we store all the data locally, but sounds good 😁

js/src/pages/price-benchmark/price/index.js
asvinb: Good catch. Added  a check 👍

tests/e2e/specs/price-benchmark/price-benchmark-suggestions.test.js
asvinb: 🤦 . Updated.

js/src/pages/price-benchmark/price-benchmark-suggestions.js
joemcgill: Is there a use case you have in mind where another component would need to use `PriceBenchmarkTable`? Seems like any custom tables that need different data could do so and wrap their own DataViews component.

js/src/pages/price-benchmark/price-benchmark-table/index.js
joemcgill: I assumed that DataViews handled this out of the box. Given that it doesn't seem to, I think we can do this in a follow-up as a nice to have.

tests/e2e/utils/__fixtures__/price-benchmark-suggestions.json
joemcgill: Oh, I hadn't thought about that. Were you planning to change the shape in the reducer? I think keeping the shape the same will likely help with debugging since we can mock the REST API response using the same data that you're using in the E2E tests

js/src/pages/price-benchmark/price-benchmark-table/index.js
asvinb: I thought so as well 😁 Yeah, let's put it as nice to have. Furthermore, I've updated the callback to use `useCallback`

tests/e2e/utils/__fixtures__/price-benchmark-suggestions.json
asvinb: Sounds good. Shape has been updated to be consistent 👍

js/src/pages/price-benchmark/price-benchmark-suggestions.js
asvinb: So eventually, we’ll have the `PriceBenchmarkAdjustments` component. My main concern is that `PriceBenchmarkTable` will contain logic related to both suggestions and adjustments, even though we already have (or will have) dedicated components like `PriceBenchmarkSuggestions` and `PriceBenchmarkAdjustments` that encapsulate that logic. In other words, I’d prefer not to split some of logic across both `PriceBenchmarkTable` and `PriceBenchmarkSuggestions`. What do you think?

joemcgill: Ah, I see what you have in mind. Let's go with it for now and evaluate as the rest of the feature comes together. 👍🏻

js/src/pages/price-benchmark/price-benchmark-table/index.js
```diff
@@ -0,0 +1,154 @@
...
+		id: 'description',
+		enableHiding: false,
+		enableSorting: false,
+		enableGlobalSearch: true,
+		label: __( 'Description', 'google-listings-and-ads' ),
+		render: ( { item } ) => {
+			return <span>{ item.product.id }</span>;
+		},
+	},
+];
```
joemcgill: I noticed after the data shape updates that there is no validation on the item properties, which caused the app to crash with the following error:

```
TypeError: Cannot read properties of undefined (reading 'id')
```

We should probably use optional chaining or validate that the data prop passed to this component includes the expected properties to make this more robust.

Additionally, while testing this after applying the correct data shape, I noticed that the sorting functionality now also causes the app to crash:

```
TypeError: Cannot read properties of undefined (reading 'localeCompare')
```

The search functionality also seems to not be working, but am not sure that it was previously.


To reproduce the way I've tested this:

* Open DevTools to the network panel in Chrome
* Look for Fetch/XHR requests to `/wp-json/wc/gla/mc/price-benchmarks` (must have pretty permalinks enabled)
* Right click the request and choose "Override Content"
* Paste in the data from the `tests/e2e/utils/__fixtures__/price-benchmark-suggestions.json` file.
* Refresh the page with DevTools still open

js/src/pages/price-benchmark/price-benchmark-table/index.scss
```diff
@@ -0,0 +1,50 @@
+.gla-price-benchmark-table {
```
joemcgill, line 1: @asvinb I'm curious why so much of the dataviews style need to be overridden? Is this just to match the designs? For consistency with other WP UI, it would be nice if we could avoid doing so much customization. 

Not a blocker, but something to discuss with @michaeleleder when this is ready to demo.

js/src/pages/price-benchmark/price/index.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Price component to display a formatted price value with optional highlighting.
+ *
+ * @param {Object} props - The component props.
+ * @param {number} props.amount - The price value to be formatted and displayed.
+ * @param {boolean} [props.highlight=false] - Whether to apply a highlight style to the price.
+ * @return {JSX.Element} The rendered price component.
+ */
+const Price = ( { amount, highlight = false } ) => {
+	const { formatAmount } = useAdsCurrency();
```
joemcgill, line 21: Eventually, it would be nice to have some jest tests for this component that confirm this is working as intended with different currency types.

package.json
```diff
@@ -149,7 +151,11 @@
 			},
 			{
 				"path": "./js/build/vendors.js",
-				"maxSize": "8.3 kB"
+				"maxSize": "41 kB"
+			},
+			{
+				"path": "./js/build/wp-dataviews.js",
+				"maxSize": "242.57 kB"
```
joemcgill, line 158: As we discussed in Slack, including the `@wordpress/dataviews` package in plugins currently requires using a special bundle that includes all of the needed dependencies so this is cross compatible with different versions of the dependencies externalized to WP core.

Related GB discussion: https://github.com/WordPress/gutenberg/pull/66825

For now, the larger bundle size is fine, but it we should try to improve this in a follow-up issue. For example, can we avoid the additional size in `vendors.js`?

joemcgill (commented):
Almost there, @asvinb. Unfortunately I found some issues with the change in data shape that need to be fixed before this is merged. Everything else looks non-blocking.

js/src/pages/price-benchmark/price-benchmark-table/index.js
joemcgill: Same property validation applies to the TABLE_FIELDS in the `PriceBenchmarkSuggestions` component.

js/src/pages/price-benchmark/price-benchmark-table/index.scss
asvinb: @joemcgill The main styling changes are:

- Custom alignment for table cells. Some cells are left aligned and some are right aligned.
- The image dimensions are 32x32 with a 2px border radius. By default, it's 60x60 with no border radius.
- Removed the minimum height of 32px for the title to match the designs.

Sure we can discuss with @michaeleleder about that.

js/src/pages/price-benchmark/price/index.js
asvinb: Added tests although the test are mostly to test the isNaN functionality and highlight.

js/src/pages/price-benchmark/price-benchmark-table/index.js
asvinb: thanks @joemcgill . I've pushed a fix. Can you kindly check again?

js/src/pages/price-benchmark/price-benchmark-table/index.scss
michaeleleder: @asvinb I know the default DataViews may not have it, but right alignment for number columns is consistently used across multiple different sections that use DataViews in our development. The image dimension is also taken from a previous project (although I’m not sure about its dev status). I understand you don’t have access to those, but maybe @budzanowski could help?

Regarding the last bullet point, I’m not sure what it is referring to.

joemcgill (approved):
Looks good!

js/src/pages/price-benchmark/price-benchmark-table/index.scss
asvinb: Thanks @michaeleleder The last bullet point refers to a min height of 32px applied to the title which adds some spacing between the title and product id. Here is the default:
image

Compared to the designs where it looks much more compact:
image

package.json
asvinb: Sure @joemcgill . Follow up ticket created: https://github.com/woocommerce/google-listings-and-ads/issues/2863

js/src/pages/price-benchmark/price-benchmark-table/index.scss
budzanowski: @joemcgill is correct.

It was pointed out to me that we should:
> avoid "local overrides" for dataviews: this means do not have CSS overrides and use stock dataviews.

How much this complicate things?

package.json
oandregal: 👋 I was asked if I could clarify this bit.

What Joe shared is correct: `@wordpress/dataviews/wp` (the one used here) comes with bundled dependencies, so it works properly in different environments (different WordPress versions, with/without Gutenberg plugin).

The alternative is using `@wordpress/dataviews`. In that case, the dependencies are externalized, which means DataViews will access them via the `wp.*` global. This is problematic if the plugin aims to be compatible with different environments: the global components may have a different API depending on the environment, which would cause issues.

Looking at the latest `build-wp/index.js` published as part of the `@wordpress/dataviews` [npm package](https://www.npmjs.com/package/@wordpress/dataviews?activeTab=code) it seems to be about ~1.6Mb.

js/src/pages/price-benchmark/price-benchmark-table/index.scss
michaeleleder: @asvinb @budzanowski Seems like the spacing doesn’t fit the smaller images. It’s tricky because I’ve taken this from another project design (not sure about the dev status), and I think we’ll be using this in another project too, so we might need it in the future. But if now isn’t a good time to spend the effort, we can just use the large image.

These are the two different image sizes we have in our design systems. This one seems to be a combination of the small image and a two-line title:
Screenshot 2025-04-22 at 9 33 57 AM
Screenshot 2025-04-22 at 9 34 03 AM

package.json
asvinb: Thanks @oandregal ! I think in our case, it's better we use the `@wordpress/dataviews/wp` to ensure the app works properly in different environments since we are supporting previous WordPress versions. Can you kindly take a look at the suggested approach [here](https://github.com/woocommerce/google-listings-and-ads/issues/2863), whether it makes sense to you please? We don't want to bloat the `vendors` chunk unnecessarily with dependencies from the `@wordpress/dataviews` package. Let me know what you think.

js/src/pages/price-benchmark/price-benchmark-table/index.scss
asvinb: @michaeleleder In that case, I'am happy to revert the CSS overrides for dataviews and only keep the alignments overrides. Let me know if that works for you.

michaeleleder: @asvinb Okay with me, thanks!

ankitguptaindia:
Hi @asvinb thanks for the PR. I observed two concerns when testing this PR in `feature/2824-price-benchmarks ` branch. Please let me know if not related to the current PR.

1.  With this branch, when accessing Price Benchmark tab through Dashboard > Marketing > Google for WooCommerce. It shows as message "Sorry, you are not allowed to access this page." I tried this case with different setups but facing the same issue. Sometimes it works when navigating to this page via Dashboard> Post/Pages tab > Marketing > Google for WooCommerce. OR Plugin List > Google for WooCommerce > Setting tab. This always works when using the develop branch. 

Please watch the below screencast for more details. 


https://github.com/user-attachments/assets/ed27f50b-69eb-4360-9657-11d602451f7f


2. If I reached to Price Benchmark tab by navigating to the develop branch and then back to the feature branch, or by chance, Price Benchmark & Suggestions doesn't show any product although the store has multiple products. It shows a following error: 

Screenshot 2025-04-28 at 4 16 01 PM


----


**Browser network and Console status:** 

<Details>

https://github.com/user-attachments/assets/3b5994fc-9b5d-4ff3-99f0-9b3e7772d721

</Details>


cc: @joemcgill 

Status: Moving back to in-progress for review.

asvinb:
@ankitguptaindia Thanks for your notes.

1. I think the main reason why you're getting this error is because you did not run `npm install` when you switched to the `feature/2824-price-benchmarks` branch. I've updated the QA instruction to include that. Can you let me know if you are still getting the error please?
2. So the data is coming from the Google API and directly from the store. So far we've been mocking the responses until the endpoint is ready and also until we have an account with valid test data. See Joe's [note on how to test](https://github.com/woocommerce/google-listings-and-ads/pull/2853#discussion_r2049502838) using mock data in Chrome. 

Let me know if that helps!

ankitguptaindia:
Hi @asvinb,

As per the instructions mentioned in [[this comment](https://github.com/woocommerce/google-listings-and-ads/issues/2828#issuecomment-2820928650)](https://github.com/woocommerce/google-listings-and-ads/issues/2828#issuecomment-2820928650), we have intentionally made some styling changes which may not exactly match the original design.

However, I noticed a few areas where the styles appear different. Could you please confirm whether these differences are also intentional?

---

1. **Border radius and image size** –  
   The design includes rounded corners for product images and slightly smaller image dimensions.

<details>
  <summary>Screenshot</summary>

  image
</details>

---

2. **"Change Price" link color** –  
   The design uses blue color consistently for all links.

<details>
  <summary>Screenshot</summary>

  image
</details>

---

3. **Data value alignment** –  
   In the design, all titles and corresponding values are right-aligned. Can we follow the same?

<details>
  <summary>Screenshot</summary>

  image
</details>

---

4. **Missing border** –  
   The design includes a separator border just above the table title. Was this removed intentionally?

---

5. **Product title and ID alignment** –  
   In the design, the product title and ID are vertically centered with the product image.

<details>
  <summary>Screenshot</summary>

  image
</details>

---

6. **Table row alignment** –  
   Currently, table cell values are not vertically center-aligned. The design shows them centered within each row.

<details>
  <summary>Screenshot</summary>

  image
</details>

---


7. Missing Product per page details-  let me know if this is part of another ticket.

<details>
  <summary>Screenshot</summary>

Screenshot 2025-04-29 at 2 52 03 PM

</details>

Please confirm if the above are part of the intended design updates. Thanks!

asvinb:
@ankitguptaindia 

1. We wanted to avoid as much custom CSS as possible and just use the default dataviews style. [See related conversation](https://github.com/woocommerce/google-listings-and-ads/pull/2853#discussion_r2049513637). And the current implementation uses the default styles :)
2. This has been implemented in #2829 . If you pull in the latest changes from the [feature/2824-price-benchmarks]`(https://github.com/woocommerce/google-listings-and-ads/tree/feature/2824-price-benchmarks) branch, you should see the updated styles.
3. Technically the cell content are right aligned. There's some extra padding for the buttons within the table header (8px) which makes it look not aligned. I suggest we create a new issue and check with designs whether we need to align the table cell content.
4. See my answer for the first bullet point.
5. See my answer for the first bullet point.
6. See my answer for the first bullet point.
7. This is most likely a design oversight. I checked the dataviews component and we cannot display this information in the footer. References:
	* https://wordpress.github.io/gutenberg/?path=/docs/dataviews-dataviews--docs
	* https://github.com/WordPress/gutenberg/blob/trunk/packages/dataviews/src/components/dataviews-footer/index.tsx

ankitguptaindia:
Thanks for the confirmation regarding all changes. @asvinb 
For point 3, I have created a separate issue here https://github.com/woocommerce/google-listings-and-ads/issues/2879

----

**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.8.1
* Theme:  
* WooCommerce - Version 9.8.2
* PHP: 8.2.0
* Web Server: Nginx
* Browser: Chrome - Version 136
* OS: macOS Sonoma 14.6.1 

</Details>


**Test Results -**  PR works as described. As table data depends on Google API, for now testing this PR, we have mocked the responses at a browser level and tested. Some changes in the table layout are not exactly matching with provided design. Those are intentional changes and accepted in the [discussion](https://github.com/woocommerce/google-listings-and-ads/pull/2853#discussion_r2049513637) and [here](https://github.com/woocommerce/google-listings-and-ads/pull/2853#issuecomment-2839153845). 

image

PR #2869, opened 2025-04-23
Add ChangePriceModal component

### LINKED ISSUE #2829: `ChangePriceModal` component

asvinb:

Closes #2829  .

* The `getProduct` selector from the `@woocommerce/data` datastore has been used and `updateProduct` action as well.

### Screenshots:
**Mobile:**
image

**Desktop:**
image

* Navigate to the Price Benchmark page.
* Click the "Change Price" link for the desired product.
* A modal will appear, displaying various pricing metrics and an input field prefilled with the suggested price from the API.
* If a user enters a negative value or a price lower than the current sales price, an error message should appear, and the "Change Price" button will be disabled.
* Upon entering a valid price and clicking "Change Price", the modal should close and the product’s price in the table should update accordingly.
* Confirm that the product price has been successfully updated in WooCommerce.

### Additional details:
* The designs do not show the error state for the price input. The same pattern found in the codebase has been followed, i.e the error shows up below the input field.

### Changelog entry

>


js/src/components/delta-value/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+	const isNegative = value < 0;
+	const isPositive = value > 0;
+
+	let formattedValue = parseInt( value, 10 );
+	if ( isPositive ) {
+		formattedValue = `+${ parseInt( value, 10 ) }${ suffix }`;
+	} else if ( isNegative ) {
+		formattedValue = `${ parseInt( value, 10 ) }${ suffix }`;
+	}
```
joemcgill: This could be simplified since only positive values are affected:
```suggestion
	let formattedValue = `${ parseInt( value, 10 ) }${ suffix }`;

	if ( isPositive ) {
		formattedValue = `+${ formattedValue }`;
	}
```

js/src/components/delta-value/index.js
```diff
@@ -0,0 +1,50 @@
...
+ * and an optional suffix. The component also applies appropriate CSS classes based on the
+ * sign of the value.
+ *
+ * @param {Object} props - The component props.
+ * @param {number} props.amount - The numeric value to be displayed. Determines if the value is positive, negative, or zero.
+ * @param {string} [props.suffix=''] - An optional string to append to the formatted value.
+ *
+ * @return {JSX.Element} A span element containing the formatted delta value with appropriate CSS classes applied.
+ */
+const DeltaValue = ( { amount = 0, suffix = '' } ) => {
```
joemcgill, line 22: Probably a good idea to add some validation to ensure the suffix is only a string.

js/src/components/delta-value/index.test.js
```diff
@@ -0,0 +1,66 @@
...
+		expect( spanElement ).toBeInTheDocument();
+	} );
+
+	it( 'renders a negative value with a suffix', () => {
+		render( <DeltaValue amount={ -20 } suffix="%" /> );
+		const spanElement = screen.getByText( '-20%' );
+		expect( spanElement ).toBeInTheDocument();
+	} );
+
+	it( 'handles NaN amount gracefully by rendering 0', () => {
```
joemcgill, line 55: Maybe add a similar test for if a non string value is passed as a suffix.

js/src/hooks/useProduct.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Custom hook to retrieve product data and resolution status based on a given product ID.
+ *
+ * @param {string|null} productID - The ID of the product to retrieve. If null, returns default values.
+ * @return {Object} An object containing the following properties:
+ *   - {Object|null} product: The product data retrieved from the store, or null if no product ID is provided.
+ *   - {boolean} hasFinishedResolution: Indicates whether the resolution for the product data has completed.
+ */
+const useProduct = ( productID ) => {
```
joemcgill, line 17: This seems ok, but surprised that we need to create a custom hook for this rather than something directly from the `@woocommerce/data` package.

js/src/pages/price-benchmark/change-price-modal/index.js
```diff
@@ -0,0 +1,190 @@
...
+	priceGap,
+	suggestedPrice,
+	clicks,
+	conversions,
+	predictedClicksChange,
+	predictedConversionsChange,
+	onRequestClose,
+	onPriceChange = noop,
+} ) => {
+	if ( ! product ) {
```
joemcgill: We should probably return early if there is no product.id, since that is a hard requirement of being able to modify the price.

js/src/pages/price-benchmark/change-price-modal/index.js
```diff
@@ -0,0 +1,190 @@
...
+	product,
+	effectiveness,
+	regularPrice,
+	priceOnGoogle,
+	priceGap,
+	suggestedPrice,
+	clicks,
+	conversions,
+	predictedClicksChange,
+	predictedConversionsChange,
```
joemcgill: If we're passing all of the props into this component, can we set some defaults for when some of these props are not present? For example, we might not always have the `predictedClicksChange` or `predictedConversionsChange` data for a product.

js/src/pages/price-benchmark/change-price-modal/index.js
```diff
@@ -0,0 +1,190 @@
...
+							/>
+						</div>
+					) }
+
+					<div className="gla-change-price-modal__product-details">
+						<p>
+							<span>{ product.id }</span>
+						</p>
+						<p className="gla-change-price-modal__product-title">
+							{ product.title }
```
joemcgill: We should validate that we have a title before rendering this.

js/src/pages/price-benchmark/change-price-modal/index.js
```diff
@@ -0,0 +1,190 @@
...
+							value={
+								<DeltaValue
+									amount={
+										( predictedConversionsChange ?? 0 ) *
+										100
+									}
+									suffix="%"
+								/>
+							}
+						/>
```
joemcgill: For these two specific metrics (and maybe others), in many cases we will not have this data. In those cases we should likely either hide these metrics or show something like `N/A`.

Let's clarify how we want to handle cases where we don't have all of the data shown in this modal in a follow-up task. In the meantime, can we hide these metrics if we don't have values rather than showing `0`?

js/src/pages/price-benchmark/change-price-modal/metric-value.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Metric component displays a metric with a title and its corresponding value.
+ *
+ * @param {Object} props - The component props.
+ * @param {string} props.labelKey - The key used to retrieve the label text.
+ * @param {JSX.Element | string} props.value - The value to be displayed for the metric.
+ * @return {JSX.Element} The rendered Metric component.
+ */
+const MetricValue = ( { labelKey, value } ) => {
+	if ( value === undefined || value === null ) {
```
joemcgill: Should also validate that `value` is a string if that's what is expected by the props.

js/src/pages/price-benchmark/change-price-modal/price-input-footer.js
```diff
@@ -0,0 +1,140 @@
...
+		const updatedPrice = Number.parseFloat( newPrice );
+
+		if ( updatedPrice < 0 ) {
+			return __(
+				'New price must be greater than or equals to zero.',
+				'google-listings-and-ads'
+			);
+		}
+
+		const salePrice = Number.parseFloat( productDetails?.sale_price );
```
joemcgill: Are we displaying the `productDetails.sale_price` anywhere else? Seems like it might be confusing if we're displaying different price data and someone encounters this error.

js/src/pages/price-benchmark/change-price-modal/price-input-footer.js
```diff
@@ -0,0 +1,140 @@
...
+			setNewPriceError( error?.message );
+			setLoading( false );
+			return;
+		}
+
+		setLoading( false );
+		onPriceChange( productId, newPrice );
+	}, [ newPrice, productId, onPriceChange, updateProduct, validatePrice ] );
+
+	if ( ! productId ) {
```
joemcgill: What if `useProduct()` returns no data for an ID? This should probably check `productDetails` instead.

js/src/pages/price-benchmark/change-price.js
```diff
@@ -2,27 +2,98 @@
...
+ * @param {Product} props.product - The product object containing details about the product. Properties include id, title, and thumbnail.
+ * @param {number} props.effectiveness - The effectiveness of the price change.
+ * @param {number} props.regularPrice - The regular price of the product.
+ * @param {number} props.priceOnGoogle - The current price of the product on Google.
+ * @param {number} props.priceGap - The price gap between the regular price and the price on Google.
+ * @param {number} props.suggestedPrice - The suggested price for the product.
+ * @param {number} props.clicks - The number of clicks the product has received.
+ * @param {number} props.conversions - The number of conversions the product has achieved.
+ * @param {number} props.predictedClicksChange - The predicted change in clicks if the price is updated.
+ * @param {number} props.predictedConversionsChange - The predicted change in conversions if the price is updated.
```
joemcgill: I would prefer not needing to pass all of these props to the `ChangePrice` component as it makes it difficult to use outside of this specific context. Ideally, we would have a component that could be rendered on any page (e.g., on a single product editor page) and as long as we pass the product ID and the content to be rendered in the button, it will take care of getting all of the data needed for rendering the `ChangePriceModal` component.

Originally in the IB for this work we were only passing a couple of props to the `ChangePriceModal` component but now we're passing props from the table down several layers to render this component correctly. Alternately, we could make the `ChangePriceModal` component responsible for fetching its own data so that we could pass only the product ID and allow the modal to take care of pulling the rest of the data it needs from the store. What do you think?

js/src/pages/price-benchmark/price-benchmark-suggestions.js
```diff
@@ -101,7 +99,22 @@ const TABLE_FIELDS = [
...
+					priceGap={ item.price_gap }
+					suggestedPrice={ item.suggested_price }
+					clicks={ item.clicks }
+					conversions={ item.conversions }
+					predictedClicksChange={ item.predicted_clicks_change }
+					predictedConversionsChange={
+						item.predicted_conversions_change
+					}
+				/>
+			);
```
joemcgill: If my suggestion about handling these props in the `ChangePriceModal` component is applied, then this can be reset.

tests/e2e/specs/price-benchmark/price-benchmark-suggestions.test.js
```diff
@@ -132,19 +124,117 @@ test.describe( 'Price Benchmark Page', () => {
...
+			);
 
-			const tableHeaderColumns = page.locator( 'table thead tr th' );
-			await expect( tableHeaderColumns ).toHaveCount( 2 );
+			await priceBenchmarkPage.fulfillWCProduct(
+				{
+					regular_price: '120.00',
+				},
+				[ 'POST' ]
+			);
```
joemcgill: This doesn't seem to be used in this test. Is that intentional?

tests/e2e/specs/price-benchmark/price-benchmark-suggestions.test.js
```diff
@@ -132,19 +124,117 @@ test.describe( 'Price Benchmark Page', () => {
...
+				'New price must be greater than the sale price (NT$90.00).'
+			);
+
+			const changePriceButton =
+				await priceBenchmarkPage.getChangePriceButtonModal();
+			await expect( changePriceButton ).toBeDisabled();
+		} );
+
+		test( 'Clicking "Change Price" button with a valid price closes the modal and updates the table', async () => {
+			const priceInput = await priceBenchmarkPage.getPriceInputModal();
```
joemcgill, line 281: I'd move the code I flagged earlier here:

```js
await priceBenchmarkPage.fulfillWCProduct(
	{
		regular_price: '120.00',
	},
	[ 'POST' ]
);
```

tests/e2e/utils/pages/price-benchmark.js
```diff
@@ -70,4 +70,76 @@ export default class PriceBenchmarkPage extends MockRequests {
...
+	getPriceInputModal() {
+		return this.getChangePriceModal().getByLabel( 'New Price' );
+	}
+
+	/**
+	 * Retrieves the "Change Price" button element within the change price modal.
+	 *
+	 * @return {import('@playwright/test').Locator} The "Change Price" button element within the modal.
+	 */
+	getChangePriceButtonModal() {
```
joemcgill: Nit: This seems to better align with other names
```suggestion
	getChangePriceModalButton() {
```

joemcgill (changes requested):
Thanks @asvinb. There is A LOT to go through here, but I've tried to be comprehensive.

My biggest concerns right now are with the need to pass all of the price benchmarks related props down several levels to render the modal and making sure we're paying attention to cases where we may not have all of the data to render the "best case" version of this modal. 

I've added a bunch of comments inline.

A few other things I noticed while testing this component:

**1. Modal design** 
On my screen the content often overflows the maximum height of the modal, which causes the modal to need to be scrolled:

image

It looks like the viewport has to be at least 910px tall to avoid the scrollbars, which may be fine, but seems like something we want to flag for design feedback.

Additionally, the grid spacing looks a bit off from the designs. Could you double check?

**2. Price changing**
I was not able to verify that the price change is actually working as expected with the test data I have to work with. When I tried changing the price I get an error with the following message:

```
To manipulate product variations you should use the /products/&lt;product_id&gt;/variations/&lt;id&gt; endpoint.
```

Is there a better way of confirming that functionality?

js/src/components/delta-value/index.js
asvinb: Thanks @joemcgill ! Updated!

js/src/components/delta-value/index.js
asvinb: @joemcgill I've updated the code so that whatever suffix is passed as parameter, it'll be converted to a string.

js/src/components/delta-value/index.test.js
asvinb: Sure! Updated!

js/src/hooks/useProduct.js
asvinb: @joemcgill To be consistent with the codebase, I noticed that there are no `useSelect` calls within components directly. Instead it's the hooks which uses `useSelect`.

js/src/pages/price-benchmark/change-price-modal/index.js
asvinb: Sounds good!

js/src/pages/price-benchmark/change-price-modal/index.js
asvinb: @joemcgill I've changed the approach where we are no longer passing all those props. I've added a new hook called `usePriceBenchmarkSuggestionsProduct` to retrieved the information about a product. This assumes that the data has been fetched before (for rendering the table).

js/src/pages/price-benchmark/change-price-modal/index.js
asvinb: 🤔 I don't think we should check for the title since each product should at least have a title right? What do you think?

js/src/pages/price-benchmark/change-price-modal/index.js
asvinb: @joemcgill I've changed the approach slightly. Everything is now handled within the `MetricValue` component which will not render anything if the `value` is `undefined`, `null` or a blank string.

js/src/pages/price-benchmark/change-price-modal/metric-value.js
asvinb: Actually value can now be a string, integer or a component.

js/src/pages/price-benchmark/change-price-modal/price-input-footer.js
asvinb: Good catch. Right now no, not within the modal. However I'am displaying it in the error message. I think we should display it within the modal. I'll ping @michaeleleder in Figma.

js/src/pages/price-benchmark/change-price-modal/price-input-footer.js
asvinb: 👍

js/src/pages/price-benchmark/change-price.js
asvinb: @joemcgill Sounds good to me. I've changed to approach and added the `usePriceBenchmarkSuggestionsProduct` hook to retrieve a product based on its ID. Let me know what you think.

js/src/pages/price-benchmark/price-benchmark-suggestions.js
asvinb: 👍

tests/e2e/specs/price-benchmark/price-benchmark-suggestions.test.js
asvinb: Actually i wanted to avoid a page reload for the test on L215 😅 , that's why i had it here.

tests/e2e/specs/price-benchmark/price-benchmark-suggestions.test.js
asvinb: 👍  This involves a page reload though, but it should be fine.

tests/e2e/utils/pages/price-benchmark.js
asvinb: 👍

asvinb:
@joemcgill 

1. I'll ping @michaeleleder in Figma for the modal. Additionally there was a missing `flex: 1;` on a container which did not make the container expand to the edges. This has been fixed now.
2. For the error, can you make sure the product ID you have in the mock data matches the product id you have in the store. For instance in the store, I have a product with the ID 33 and that's why the first product in the mock data has the ID set to 33.

tests/e2e/utils/__fixtures__/price-benchmark-suggestions.json
```diff
@@ -1,15 +1,19 @@
 [
 	{
 		"product": {
-			"id": 259251,
+			"id": 33,
```
asvinb, line 4: @joemcgill I update this ID to match a product with the same ID in my local setup.

joemcgill (approved):
Thanks for the updates, @asvinb. I think this is looking good

ankitguptaindia:
QA Status: Needs feedback ⚠️ 
---

@asvinb, thanks for the PR and the detailed description to assist with testing. Overall, the features work as expected. I do have a few suggestions for potential improvements:

1. When a merchant updates the price and enters an invalid or negative value, an error is correctly shown. However, after correcting the price, the merchant needs to click the "Update Price" button twice — once to dismiss the error message and a second time to actually update the price.

**Please take a look at my screencast to see the steps I followed:**

https://github.com/user-attachments/assets/30ee73c5-a514-4491-aca0-e28c120f3e37


----

2. The tooltip appears on hover for various elements, but it's overflowing outside the pop-up modal. Can we align all tooltips are properly aligned and contained within the modal boundaries?

https://github.com/user-attachments/assets/02f75cb9-e6b6-4145-97af-82ac09aad390

----

3. As we don't have a design for the error state when the merchant enters an invalid amount. I have added a separate issue to address. https://github.com/woocommerce/google-listings-and-ads/issues/2887

ankitguptaindia:
Update: 
- Points 2, 3 have been addressed and fixed with the related PR.
- Point is being discussed here on [Slack](https://fueled10up.slack.com/archives/C060DC0NXNG/p1746174985723409?thread_ts=1746174039.998319&cid=C060DC0NXNG). Once, address, will move to next step.

PR #2811, opened 2025-03-10
[API Pull Sync] Add Connection Test info and refactor error debugging

puntope:
### Changes proposed in this Pull Request:

This PR:

- Enables Connection Test info about the current Syncing
- Refactor Error Logging for Notifications
- Refactor SyncTrait moving the methods inside NotificationService
- Add some utility methods.

### Screenshots:

Screenshot 2025-03-10 at 16 23 08

Screenshot 2025-03-10 at 12 23 45

### Detailed test instructions:

1. Checkout this PR
2. Go to the Connection Test Page. Verify the data about API PULL Sync for the different data types is present.
3. Send a Notification and make it fail. Go to error WooCommerce logs and see the error in more detailed with information about the different key components of the system to work.  

### Changelog entry

>

PR #2870, opened 2025-04-25
API Push/Pull Sync control endpoint Implementation

budzanowski:
# API Pull Sync Feature Implementation

This PR implements a comprehensive API Pull Sync system for Google Listings and Ads, allowing for better control and visibility over data synchronization between WooCommerce stores and Google services.

### Endpoints
- **GET** `/wp-json/wc/gla/sync`
  - Returns the current sync configuration for all data types
  - Example response:
    ```json
    {
      "products": {
        "pull": true,
        "push": true
      },
      "coupons": {
        "pull": true,
        "push": true
      },
      "shipping": {
        "pull": true,
        "push": true
      },
      "settings": {
        "pull": true,
        "push": true
      }
    }
    ```

- **POST** `/wp-json/wc/gla/sync`
  - Updates the sync configuration for specified data types
  - Request body follows the same structure as the GET response
  - Allows selective updates of specific data types and sync directions
  - Returns the updated configuration in the same format as the GET endpoint

### Data Storage
- Configuration stored in WordPress options table as `gla_api_pull_sync_mode`
- Default values set to `true` for all data types and directions if not configured

### Implementation
- `SyncController` class handles REST API requests and validation
- `SyncTrait` provides methods to check if sync is enabled for a given data type and direction
- Implementation uses filters to allow for programmatic modification of sync behavior:
  - `woocommerce_gla_is_pull_sync_enabled_for_datatype`
  - `woocommerce_gla_is_push_sync_enabled_for_datatype`
- Notification system integrates with sync settings to prevent unnecessary operations

### Connection Test Page Enhancements
- Added API Pull Sync information section to the Connection Test page
- Displays current configuration for all data types
- Provides debugging information for troubleshooting sync issues
- Improved error logging and presentation

## Key Components

1. **Sync Endpoint Framework**
   - Created a new REST API endpoint at `/wp-json/wc/gla/sync` 
   - Provides both GET and POST methods to retrieve and update sync settings
   - Stores configuration in the WordPress options table as `gla_api_pull_sync_mode`

2. **Configurable Sync Mode**
   - Allows enabling/disabling pull and push synchronization for different data types:
     - Products
     - Coupons
     - Shipping settings
     - Store settings
   - Each data type can be configured independently with granular control

3. **Notification Management**
   - Intelligently blocks notification jobs when pull mode is disabled for a data type
   - Prevents unnecessary API calls and sync operations
   - Improves efficiency by only syncing when appropriate

4. **Improved Debugging**
   - Enhanced Connection Test page with API Pull Sync information
   - Refactored error logging for notifications
   - Added detailed debugging information to help troubleshoot sync issues

## Previously Approved PRs
This PR combines the following previously approved PRs:
1. #2788 - Initial API Pull Route GET Sync Endpoint
2. #2796 - API Pull Sync POST Endpoint
3. #2804 - Check Pull Sync Mode before operations
4. #2811 - Connection Test page enhancements

## Test plan
- Verify the sync endpoint works via GET and POST methods:
  - GET `/wp-json/wc/gla/sync` returns current configuration
  - POST with partial updates only modifies specified values
  - POST with invalid data returns appropriate error messages
- Test enabling/disabling pull sync for different data types:
  - Disable pull for products and verify notifications are not sent
  - Re-enable and confirm sync resumes
  - Test with various combinations of settings
- Confirm notification jobs are skipped when appropriate:
  - Check that jobs respect the sync configuration
  - Verify that other operations continue to work
- Check that the Connection Test page displays API pull sync information correctly:
  - All data types and their sync status are shown
  - Information is accurate and matches the database configuration

### Changelog entry

> Add - API Push/Pull Sync control endpoint Implementation.


src/API/Site/Controllers/RestAPI/SyncController.php
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * POST wp-json/wc/gla/sync
+ *
+ *  {
+ *      "products": {
+ *           "pull": true,
+ *      }
+ *  }
+ *
+ * Updates Only Products to enable Pull. Rest of the data types and Products Pull method will remain unchanged.
```
louwie17, line 57: I find this slightly strange to support on a POST request, to me this should be a PUT request.
Having said that, I realize this would also mean a follow up change on the wp.com side, so I am open to letting this slide.

src/API/WP/NotificationsService.php
```diff
@@ -171,11 +208,13 @@ public function get_notification_url(): string {
...
 	 *
-	 * @param bool $with_health_check If true. Performs a remote request to WPCOM API to get the status.
-	 * @return bool
+	 * @param string|null $data_type The data type to check.
+	 * @param bool        $with_health_check If true. Performs a remote request to WPCOM API to get the status.
+	 *        * @return bool
 	 */
-	public function is_ready( bool $with_health_check = true ): bool {
-		return $this->options->is_wpcom_api_authorized() && $this->is_enabled() && $this->merchant_center->is_ready_for_syncing() && ( $with_health_check === false || $this->account_service->is_wpcom_api_status_healthy() );
+	public function is_ready( string $data_type = null, bool $with_health_check = true ): bool {
```
louwie17, line 215: I believe changing this function structure should be fine when updating because we are defaulting to `null`, also assuming no plugins extend this :)
My update concern is mostly in the case where this function gets called during the update and either its caller is still the old version or the `is_ready` function is still the old version, meaning it could potentially trigger a fatal because of the mismatch in types.

louwie17 (approved):
This tested well and code looks good to me, just left two comments, that aren't of big concern, but I thought I call out:
- The POST request is technically a PUT ( and IMO should be a PUT )
- The change in the `is_ready` function call and the likelihood of it being called why the plugin is being updated.

src/API/Site/Controllers/RestAPI/SyncController.php
budzanowski: If we want to nitpick then both PUT and POST are wrong. It should be PATCH 😉 We are not replacing the "resource" but only updating it partially.

PR #2878, opened 2025-04-30
Add/2828 price suggestions table optimized import

### LINKED ISSUE #2828: `PriceBenchmarkSuggestionsTable` component

### LINKED ISSUE #2863: Optimize @wordpress/dataviews package loading to limit usage to PriceBenchmark pages only
## User Story

With the introduction of the `@wordpress/dataviews` package in [#2828](https://github.com/woocommerce/google-listings-and-ads/issues/2828), it should be loaded **only** where required—specifically on PriceBenchmark pages—to reduce unnecessary JavaScript bundle size and improve overall plugin performance.

## Figma Link

N/A

## Acceptance Criteria

- `@wordpress/dataviews` is **not** bundled or loaded on plugin pages where it's not used.
- `@wordpress/dataviews` is loaded **only** on PriceBenchmark-related pages.
- The vendors bundle size is reduced to its [original size](https://github.com/woocommerce/google-listings-and-ads/pull/2853/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L152) (~8.3 kB).
- No regressions occur on PriceBenchmark pages—existing functionality remains intact.

## Implementation Brief

- Create a new shim file at `js/src/globals/wp-dataviews-shim.js` that assigns all exports from `@wordpress/dataviews/wp` to `window.wp.dataviews`.
- Add a separate Webpack entry point for this file and update the `externals` configuration to exclude `@wordpress/dataviews` from the main bundle.
- Enqueue the shim script **only** on PriceBenchmark-related pages (in PHP).
- Update imports for `@wordpress/dataviews` to use the global `wp.dataviews` object.
- Reference: [Gutenberg discussion on package size](https://github.com/WordPress/gutenberg/pull/66825)

## Test Coverage

- Confirm PriceBenchmark pages still function correctly with `@wordpress/dataviews`.
- Confirm `@wordpress/dataviews` is **not** included in runtime JS for other plugin pages.

ankitrox:

Closes #2863

### Screenshots:

**Data view script loading**
loading-data-view

**Data view script load errored**
data-view-error

**Data view script loading success**
data-view-success

1. Go to Marketing > Google for WooCommerce, Price Benchmark & Suggestions.
2. Set the network throttling to 3G and load the tab again, so that you can see "Loading" status as shown in screenshot.
3. Once the dataview script is loaded, you should see the suggestions table.

### Changelog entry

>


src/Admin/Admin.php
```diff
@@ -143,6 +144,19 @@ protected function get_assets(): array {
 					'mcId'    => $this->options->get_merchant_id() ?: null,
 					'adsId'   => $this->options->get_ads_id() ?: null,
 				],
+				'dataViewsScriptUrl'       => add_query_arg(
```
joemcgill, line 147: Using the inline script for `glaData` to pass the URL for our custom `wp-dataviews.js` loader is ok, but it seems like this approach for lazy loading the dataviews scripts could be handled in a more natural way that doesn't require a custom script to unpack the dependencies and load them on the global wp object directly. Were there any alternatives explored?

ankitrox: @joemcgill I tried exploring other approaches, but if we use import, then it will bundle the `@wordpress/dataviews` package in vendor.js.

With this approach, our following goals are fulfilled.

1. `@wordpress/dataviews` is not bundled in vensor.js, which result in smaller size of vendor.js file.
2. Dataview utility is loaded only on demand. It gets loaded asynchronously only when `Price Benchmark & Suggestions` section is visited.

Loading the script tag async in useeffect may seem a bit hacky as of now, but as we discussed over our sync call, this solves our primary objective.

One possible solution as I can think of is that we can create a separate chunk for `PriceBenchmarkSuggestions` and then load this component on demand using `lazy` and `import`. I tried this for a very short while, but getting some unexpected errors and the page is not loading.

In interest of the time and to move forward, we can create a separate issue for handling this in a different way where we can sketch out possible solutions during definition.

Please let me know your thoughts.

CC: @asvinb

js/src/pages/price-benchmark/index.js
```diff
@@ -21,7 +52,7 @@ const PriceBenchmark = () => {
 			<ProductComparisonChart />
 
 			<Card className="gla-price-benchmark__card">
-				<PriceBenchmarkSuggestions />
+				<PriceBenchmarkSuggestions isDataViewReady={ dataViewLoaded } />
```
asvinb: @ankitrox Let's not display `PriceBenchmarkSuggestions` if `dataViewLoaded` is not loaded. We can render `AppSpinner` while it's being loaded. This will avoid the checks you have in `PriceBenchmarkSuggestions` whether dataviews is ready or not.

js/src/pages/price-benchmark/index.js
```diff
@@ -13,6 +16,34 @@ import Banner from './banner';
 import './index.scss';
 
 const PriceBenchmark = () => {
+	const [ dataViewLoaded, setDataViewLoaded ] = useState();
+
+	useEffect( () => {
+		if ( dataViewLoaded === undefined ) {
+			const script = document.createElement( 'script' );
+			script.src = glaData.dataViewsScriptUrl;
```
asvinb: @ankitrox You can import `glaData` from `constants` and then access `dataViewsScriptUrl`
`import { glaData } from '~/constants';`

js/src/pages/price-benchmark/index.js
```diff
@@ -1,6 +1,9 @@
+/* global glaData */
```
asvinb: See my comment below on how to import things from `glaData`

js/src/pages/price-benchmark/index.js
```diff
@@ -13,6 +16,34 @@ import Banner from './banner';
...
+	const [ dataViewLoaded, setDataViewLoaded ] = useState();
+
+	useEffect( () => {
+		if ( dataViewLoaded === undefined ) {
+			const script = document.createElement( 'script' );
+			script.src = glaData.dataViewsScriptUrl;
+			script.async = true;
+
+			script.onload = () => {
+				if (
```
asvinb: @ankitrox This can be simplified as such:
setDataViewLoaded(
  typeof window.wp?.dataviews?.filterSortAndPaginate === 'function'
);

js/src/pages/price-benchmark/index.js
```diff
@@ -13,6 +16,34 @@ import Banner from './banner';
 import './index.scss';
 
 const PriceBenchmark = () => {
+	const [ dataViewLoaded, setDataViewLoaded ] = useState();
```
asvinb: @ankitrox We cannot rely on the state to load the script because when the component gets unmounted and mounted again, the state will be false and then the script will be appended again even though it was already appended. You can see that in action when you navigate between the tabs within the plugins and notice how `wp-dataviews.js` is being requested every time you visit the Price Benchmark page. The script should be loaded only once. You can use the following as reference:
https://github.com/woocommerce/woocommerce/blob/934d9d8c252896dee26b6263ce97d5a6b84eccd5/plugins/woocommerce/client/blocks/assets/js/base/utils/lazy-load-script.ts#L60-L104

js/src/pages/price-benchmark/empty-metrics-notice/wrapper.js
```diff
@@ -0,0 +1,7 @@
+const Wrapper = ( { children } ) => {
```
asvinb: @ankitrox Why we have this `Wrapper` component. Is it only for the styles? Side notes, even if we won't be using that component.

1. The props should be documented, in that case `children`
2. The name of the component is too generic. Maybe `EmptyMetricsWrapper` would have been more suitable.

js/src/pages/price-benchmark/empty-metrics-notice/index.js
```diff
@@ -9,11 +9,12 @@ import { __ } from '@wordpress/i18n';
  */
 import AppDocumentationLink from '~/components/app-documentation-link';
 import EmptyMetricsGraphic from '~/images/price-benchmark/empty-metrics.svg';
+import Wrapper from './wrapper';
 import './index.scss';
 
 const EmptyMetricsNotice = () => {
 	return (
-		<div className="gla-price-benchmark__empty-metrics">
+		<Wrapper>
```
asvinb: See my note below for the `Wrapper` component.

js/src/pages/price-benchmark/price-benchmark-suggestions/index.js
```diff
@@ -163,12 +163,15 @@ const TABLE_FIELDS_MOBILE = [ 'action' ];
  *
  * @param {Object} props - The component props.
  * @param {boolean} props.isViewportMobile - Indicates if the viewport is in mobile size.
+ * @param {boolean} props.isDataViewReady - Indicates if the DataView component is ready.
  * @return {JSX.Element} A div containing the DataViews component.
  */
-const PriceBenchmarkSuggestions = ( { isViewportMobile } ) => {
+const PriceBenchmarkSuggestions = ( { isViewportMobile, isDataViewReady } ) => {
```
asvinb: Let's remove `isDataViewReady`. See my previous comments.

webpack.config.js
```diff
@@ -134,6 +124,11 @@ const webpackConfig = {
 			},
 		},
 	},
+	externals: {
+		...( defaultConfig.externals || {} ),
+		'@wordpress/dataviews': [ 'wp', 'dataviews' ],
```
asvinb: @ankitrox Why we need that?

src/Admin/Admin.php
```diff
@@ -143,6 +144,19 @@ protected function get_assets(): array {
...
+					[
+						'vesrion' => (string) filemtime( "{$this->get_root_dir()}/js/build/wp-dataviews.js" ),
```
asvinb: Typo.
```suggestion
						'version' => (string) filemtime( "{$this->get_root_dir()}/js/build/wp-dataviews.js" ),
```

package.json
```diff
@@ -155,7 +155,7 @@
 			},
```
asvinb, line 155: @ankitrox As per the AC, we should revert the change we made to the `vendors` chunk.
https://github.com/woocommerce/google-listings-and-ads/pull/2853/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L152

js/src/pages/price-benchmark/price-benchmark-suggestions/index.js
```diff
@@ -163,12 +163,15 @@ const TABLE_FIELDS_MOBILE = [ 'action' ];
...
 	const { suggestions, hasFinishedResolution } =
 		usePriceBenchmarkSuggestions();
 
+	const { DataViews, filterSortAndPaginate } = wp.dataviews || {};
```
asvinb: @ankitrox Can you see how we can tweak webpack that we can use the normal imports, for e.g 
```js
import { DataViews, filterSortAndPaginate } from `@wordpress/dataviews/wp`;
```
but which translates to `window.wp.dataviews.DataViews` behind the scenes. Hence my question further down in `webpack.config.js`

webpack.config.js
```diff
@@ -71,11 +71,6 @@ const webpackConfig = {
 		new WooCommerceDependencyExtractionWebpackPlugin( {
 			externalizedReport:
 				! hasReactFastRefresh && '../../.externalized.json',
-			requestToExternal: ( request ) => {
```
asvinb: @ankitrox Why we removed this because the package should be bundled and not treated as external.

js/src/wp-dataviews.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Internal dependencies
+ */
+import {
```
asvinb: @ankitrox If we keep `requestToExternal` in `webpack.config.js`, then we should be able to import the dataviews package as usual.
```js
import { DataViews, filterSortAndPaginate } from '@wordpress/dataviews/wp';
```

js/src/wp-dataviews.js
```diff
@@ -0,0 +1,12 @@
...
+	DataViews,
+	filterSortAndPaginate,
+} from '../../node_modules/@wordpress/dataviews/build-wp';
```
asvinb: Also, let's expose everything from that package since we already have it:
`import * as DataViews from '@wordpress/dataviews/wp';`

js/src/wp-dataviews.js
```diff
@@ -0,0 +1,12 @@
...
+
+window.wp = window.wp || {};
```
asvinb: Then this can be simplified to:
```js
window.wp = window.wp || {};
window.wp.dataviews = DataViews;
```

js/src/wp-dataviews.js
```diff
@@ -0,0 +1,12 @@
+/**
```
asvinb: @ankitrox As per the IB, this should live in 
`js/src/globals/wp-dataviews-shim.js`

asvinb (changes requested):

js/src/wp-dataviews.js
asvinb: Or maybe move it to `js/src/shims/wp-dataviews.js`

js/src/pages/price-benchmark/index.js
ankitrox: We can setthe initial value of the state to `wp.dataviews` which would be undefined for the first time and then would be set.
We can also write a cleanup function in useEffect to reset `dataViewLoaded` to undefined when `dataViewLoaded` is `false`for some reason, so that next time component is mounted, it will retry to  load the script.

js/src/pages/price-benchmark/empty-metrics-notice/wrapper.js
ankitrox: Removed it.

js/src/wp-dataviews.js
ankitrox: Added `js/src/shims/wp-dataviews.js`

webpack.config.js
ankitrox: I am also not sure why and when I removed it. I think I was trying to make this as external one, thus added `externals` below down the line and removed this one, but I restored this now and removed `externals`.

webpack.config.js
ankitrox: Removed it as we are now using `requestToExternal` in `WooCommerceDependencyExtractionWebpackPlugin`.

js/src/pages/price-benchmark/index.js
asvinb: @ankitrox It's not efficient from a user point of view to request the same script over and over again 😁  The script should be loaded once, i.e when first used. For subsequent in page navigation, the script should not be loaded again.

js/src/pages/price-benchmark/index.js
asvinb: @ankitrox I cannot see that change. Did you forget to push your changes?

js/src/wp-dataviews.js
ankitrox: If we use `import`, then the `@wordpress/dataviews` code is bundled into vendors.js because in `requestToExternal` we are returning `null` which instructs bundler to bundle the module code.

I think it is safe to use `const` and destructuring properties from `wp.dataviews` in order to keep the vendors size minimum.

js/src/pages/price-benchmark/index.js
ankitrox: With the strategy I mentioned, the script will be loaded only once. Just in case, there is any error in loading the script where we are setting `dataViewLoaded` to `false`, the cleanup function would set `dataViewLoaded` to `undefined` so that next time the component loads, it will try to load the script again because it was anyways not loaded previously.

js/src/pages/price-benchmark/index.js
ankitrox: Sorry, I reverted that change when I was working on the PR. Added it again.

js/src/pages/price-benchmark/index.js
asvinb: @ankitrox Still it's not efficient. Basically whenever i visit the "Price Benchmark" tab, the script will be loaded again and again. We don't need a cleanup function. In this case, we just want to load the script once and that's it 😁  If it's not loaded, we load it. If it's already loaded, then do nothing. And this is irrespective of the component being mounted/unmounted.

asvinb (approved):

ankitguptaindia:
Performed Smoke testing on `add/2828-price-suggestions-table-optimized-import` All cases working as expected. 
Will verify all other cases when merged on the featured branch, along with other PRS.

src/Admin/Admin.php
joemcgill: Thanks for looking into it, @ankitrox. Let's revisit this in a follow-up if we have some time to investigate further, but I agree that this approach solves the immediate problem. 👍🏻

joemcgill (approved):

Issue #2890, opened 2025-05-01
Create methods for testing Content API data in QA


joemcgill:
In order to properly test the Price Benchmarks features in https://github.com/woocommerce/google-listings-and-ads/issues/2824 we need to be able to model the expected responses we're getting back from the Content API using a few different use cases:

1. Accounts with neither Price Competitiveness (benchmarks) data NOR Price Insights (suggestions)
2. Accounts with Price Competitiveness data BUT NOT Price Insights data
3. Accounts with both Price Competitiveness AND Price Insights data

Here are the initial queries that we're sending from the plugin along with the connected Merchant ID:

**Price Competitiveness (benchmarks)**

```
SELECT product_view.id,product_view.offer_id,product_view.title,product_view.price_micros,product_view.currency_code,price_competitiveness.benchmark_price_micros,price_competitiveness.benchmark_price_currency_code,price_competitiveness.country_code FROM PriceCompetitivenessProductView
```

**Price Insights (suggestions)**

```
SELECT product_view.id,product_view.offer_id,product_view.title,product_view.price_micros,product_view.currency_code,price_insights.suggested_price_micros,price_insights.suggested_price_currency_code,price_insights.predicted_impressions_change_fraction,price_insights.predicted_clicks_change_fraction,price_insights.predicted_conversions_change_fraction FROM PriceInsightsProductView
```

**Price Metrics (clicks)**

```
SELECT segments.offer_id,metrics.clicks,segments.week FROM MerchantPerformanceView WHERE segments.offer_id IN ('12345') AND segments.program = 'FREE_PRODUCT_LISTING'"
```

joemcgill:
The initial framework updates for the test-proxy were merged in #2906.

I'm moving this back to in-progress to add some additional mocks for:

- [x] Handle queries to specific IDs
- [ ] Handle pagination tests

joemcgill:
Opened #2913 which adds handling for testing requests for single products via that test-proxy script.

joemcgill:
I'm going to go ahead and close this and we'll handle adding any additional test-proxy enhancements while working on follow-up functionality.

PR #2891, opened 2025-05-02
Enhance price benchmarks endpoint with uplift data

### LINKED ISSUE #2861: Enhance mc/price-benchmarks endpoint with uplift data
To provide the `ChangePriceModal` with the data needed for the UI, the `mc/price-benchmarks` endpoint needs to supply additional data. See [the related conversations](https://github.com/woocommerce/google-listings-and-ads/issues/2829#issuecomment-2815338198).

## Acceptance Criteria

The endpoint should include the following data about current reports and expected uplift, based on predictions calculated over [the past 7 days](https://support.google.com/merchants/answer/13798101?visit_id=638808468887662713-3483620688&rd=1#:~:text=Explore%20insights%20beyond%20cards):

- `clicks` – current clicks (int|null)
- `conversions` – current conversions (int|null)
- `predicted_clicks_change` – Exp. Uplift in Clicks (fraction) (string|null)
- `predicted_conversions_change` – Exp. Uplift in Conversions (fraction)(string|null)

When performance data is not available for the product ID, the values should return empty values, rather than `0`

When price insights do not exist for the product ID, the values should return empty values.

**Example response:**
`/wp-json/wc/gla/mc/price-benchmark/<id>`

```json
	{
		"product": {
			"id": 12345,
			"thumbnail": "https://mystore.com/products/images/12345.jpg",
			"title": "Example product"
		},
		"effectiveness": 3,
		"regular_price": 25.0,
		"price_on_google": 20.0,
		"price_gap": 5.0,
		"suggested_price": 20.0,
		"clicks": 30,
		"conversions": 3,
		"predicted_clicks_change": "0.12609300017356873",
		"predicted_conversions_change": "2.3431060314178467",
	}
```

## Implementation Brief

- Clicks and Conversions will come from the `MerchantPerformanceView` table ([docs](https://developers.google.com/shopping-content/guides/reports/metrics))
  - This data can be retrieved using the `API\Google\MerchantReport::get_report_data()` [method](https://github.com/woocommerce/google-listings-and-ads/blob/925c3fb030447cdff4c10c0901c91a2ac300a282/src/API/Google/MerchantReport.php#L152) passing 'products' and args for the product ID and the interval covering **the last 7 days**.
- Expected uplift for Clicks and Conversions come from `PriceInsightsProductView` ([docs](https://developers.google.com/shopping-content/guides/reports/price-insights))
  - This data should already be coming from the `MerchantPriceSuggestionsQuery` class, via the `MerchantPriceBenchmarks::get_price_insights()` method used in the [REST callback here](https://github.com/woocommerce/google-listings-and-ads/blob/f1c0c9e75c87806513b2a9d8b0eb8e5790e588c6/src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php#L74). It will need to be included as part of the `PriceBenchmarksController::map_price_benchmarks_response()` method.
  - If the query results in an error, due to no data being available for that product ID, that error should be handled in the REST callback and converted to empty values to return, rather than resulting in an error being returned in the REST response.

## Out of bounds/rabbit holes

Price benchmark data should only be expected for products with [GTINs](https://support.google.com/merchants/answer/6324461?hl=en) that have been approved in MC.

## Test coverage

Update the PHPUnit tests for the price-benchmark response to assert that when the endpoint is passed a specific product ID, it returns all the expected values for that ID including those added in this issue.

mukeshpanchal27:

Closes #2861.

- Comment the [permission callback](https://github.com/woocommerce/google-listings-and-ads/blob/fix/2861-price-benchmarks-uplift/src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php#L52) in your local setup.
- Check the Rest API response. Call `siteurl/index.php?rest_route=/wc/gla/mc/price-benchmarks/12345` (It return empty array as MC account didn't have any approved products.)

### Changelog entry

>


mukeshpanchal27:
@joemcgill I put together the draft PR as i struggling to get mocking the products report data in unit tests. Once the mocking get working just needs to get dynamic data.

I tried to add `$this->generate_ads_report_query_mock( $report_data, $report_args );` in the unit tests but it show error like below. See [details here](https://github.com/woocommerce/google-listings-and-ads/blob/440e95e83afe2db8608143de9a203f26c015bd31/tests/Unit/API/Google/AdsReportTest.php#L185).

```
1) PriceBenchmarksControllerTest::test_get_price_benchmarks_uplift_data
Error: Call to a member function method() on null

/var/www/html/wp-content/plugins/google-listings-and-ads/tests/Tools/HelperTrait/GoogleAdsClientTrait.php:644
/var/www/html/wp-content/plugins/google-listings-and-ads/tests/Unit/API/Site/Controllers/MerchantCenter/PriceBenchmarksControllerTest.php:204
/var/www/html/wp-content/plugins/google-listings-and-ads/vendor/phpunit/phpunit/src/TextUI/Command.php:146
/var/www/html/wp-content/plugins/google-listings-and-ads/vendor/phpunit/phpunit/src/TextUI/Command.php:99
```

tests/Unit/API/Site/Controllers/MerchantCenter/PriceBenchmarksControllerTest.php
```diff
@@ -148,4 +160,106 @@ public function test_get_price_benchmarks_summary() {
...
+					'clicks'      => 734,
+					'conversions' => 4,
+				],
+				'segments' => [
+					'week'          => '2021-12-12',
+					'productItemId' => 'gla_' . $product_id,
+					'productTitle'  => $product_title,
+				],
+			],
+		];
```
joemcgill: This isn't exactly what I would expect the shape of the return value to look like here. Given that we're not querying for segment data, I would expect the return value to look like [the response from this example](https://developers.google.com/shopping-content/guides/reports/query-language/overview#sample_response):

```json
{"results": [
    {
      "metrics": {
        "clicks": "734"
        "clicks": "4"
      }
    }
]}
```

In addition, the `$report_type` and `$report_args` variables aren't being used in the test, so they can be removed. The resulting code would be:

```suggestion
		$report_data = [
			'results' => [
				[
					'metrics' => [
						'clicks'      => 734,
						'conversions' => 4,
					],
				],
			],
		];
```

Even so, this highlights a flaw in these tests. In previous tests, we've not needed to pass a specific `id` arg and return values for a single product. Previous tests are returning a list of products and yet we're mocking the same single return values in all cases. Not only do we need to confirm that these service methods return the expected values, but we also need to confirm that we are passing the correct args to the methods that build these Content API queries. In this case, we can confirm that the `MerchantReport::get_report_data` method gets called once with the expected arguments, like this:

```php
		$this->merchant_report->expects( $this->once() )
			->method( 'get_report_data' )
			->with(
				'products',
				[
					'ids'      => [ self::TEST_PRODUCT_ID ],
					'fields'   => [ 'clicks', 'conversions' ],
					'interval' => 'week',
				]
			)
			->willReturn( $report_data );
```

This is better, but still doesn't really confirm that the Content API is getting passed the correct query. Hopefully this is something we can resolve with #2890, but in the meantime, we should at least make sure the methods we're using are receiving the expected data in our mocks.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -224,26 +279,79 @@ protected function get_schema_properties(): array {
...
+	}
+
+	/**
+	 * Retrieves the parameters for price benchmarks.
+	 *
+	 * @return array An associative array of parameters for price benchmarks.
+	 */
+	public function get_price_benchmarks_params(): array {
+		return [
+			'ids'      => [
```
joemcgill: This defines `ids` as the arg that is supported, but the logic seems to use the `id` param instead. These should align.

It's ok to support `fields` (as long as it's working)  but I don't think we need to support `interval` for now

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -150,11 +189,16 @@ protected function map_price_benchmarks_response( array $benchmark_data, array $
 			}
 		}
 
+		foreach ( $merchant_report_data as $merchant_report_result ) {
+			$mapped_data[ $product_id ]['merchant_report'] = $merchant_report_result['metrics'];
+		}
```
joemcgill: This doesn't seem right to me. This method assumes we're getting a set of results for several different products and are making sure we map the results from different sources to the same set of product IDs. This would just add the same metrics to the last $product_id that was mapped previously. We should probably have product ID returned in the metrics query as `segments.program_id`, which we could use to map the report to the request.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -84,19 +88,53 @@ protected function get_price_benchmarks_callback(): callable {
...
+	 */
+	protected function get_products_report_data( Request $request ): array {
+		try {
+			/** @var MerchantReport $merchant_report */
+			$merchant_report         = $this->container->get( MerchantReport::class );
+			$prepare_query_arguments = $this->prepare_query_arguments( $request );
+			$product_id              = $request->get_param( 'id' );
+
+			// If a product ID is provided, set it in the query arguments.
+			if ( $product_id ) {
```
joemcgill: Should probably bail early if there is no `$project_id` rather than still performing a query.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -84,19 +88,53 @@ protected function get_price_benchmarks_callback(): callable {
...
+	 * Retrieves the products report data for the given request.
+	 *
+	 * @param Request $request REST Request.
+	 * @return array|WP_Error Prepared response data or error.
+	 */
+	protected function get_products_report_data( Request $request ): array {
+		try {
+			/** @var MerchantReport $merchant_report */
+			$merchant_report         = $this->container->get( MerchantReport::class );
+			$prepare_query_arguments = $this->prepare_query_arguments( $request );
```
joemcgill: No need to prepare the request here since you're manually getting the `id` argument below.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -84,19 +88,53 @@ protected function get_price_benchmarks_callback(): callable {
...
+				$prepare_query_arguments['ids']      = [ $product_id ];
+				$prepare_query_arguments['fields']   = [ 'clicks', 'conversions' ];
+				$prepare_query_arguments['interval'] = 'week';
+			}
+
+			$report_data = $merchant_report->get_report_data( 'products', $prepare_query_arguments );
+			// Ensure the response is always an array.
+			if ( is_wp_error( $report_data ) ) {
+				return [];
+			}
```
joemcgill: I think this is redundant to the `catch` block below and could be removed.

joemcgill (changes requested):
Thanks @mukeshpanchal27. Left a bit of feedback but also acknowledge that this is difficult to test without being able to hit the Content API until we have #2890 sorted. I did a bit of testing and am going to push some changes that I _think_ will get us closer here.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Addressed in https://github.com/woocommerce/google-listings-and-ads/pull/2891/commits/a0477e04ba2a690c51932542c2478157f637d412

tests/Unit/API/Site/Controllers/MerchantCenter/PriceBenchmarksControllerTest.php
mukeshpanchal27: Mocking the data without an actual API response is quite challenging and time consuming. Thanks for sharing the response shape that really helps! 🙌

tests/Unit/API/Site/Controllers/MerchantCenter/PriceBenchmarksControllerTest.php
```diff
@@ -148,4 +160,110 @@ public function test_get_price_benchmarks_summary() {
...
+			],
+			'next_page_token' => 'next_page_token',
+		];
+
+		$report_data = [
+			'results' => [
+				[
+					'segments' => [
+						'date'     => '2023-10-01',
+						'offer_id' => $product_id,
```
mukeshpanchal27: @joemcgill I added the `offer_id` to match the implementation https://github.com/woocommerce/google-listings-and-ads/blob/fix/2861-price-benchmarks-uplift/src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php#L190

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Since we're passing the `ids` in the query, the response should only include data for those specific product IDs. So, do we really need to map the data? Instead, could we just merge the required items directly if the API returns a response? What do you think?

joemcgill: Yeah, that makes sense to me. So we would only include the metrics data when a the `id` argument is added to the request, not for all products returned in a default request to that endpoint.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -89,9 +89,8 @@ protected function get_price_benchmarks_callback(): callable {
...
 
 				$merchant_report_data = $this->get_products_report_data( $request );
-
-				$benchmark_data      = $merchant->get_benchmark_data( $this->prepare_query_arguments( $request ) );
-				$price_insights_data = $merchant->get_price_insights( $this->prepare_query_arguments( $request ) );
+				$benchmark_data       = $merchant->get_benchmark_data( $this->prepare_query_arguments( $request ) );
+				$price_insights_data  = $merchant->get_price_insights( $this->prepare_query_arguments( $request ) );
 
 				// Map the data to the required format.
 				$response_data = $this->map_price_benchmarks_response( $benchmark_data, $price_insights_data, $merchant_report_data );
```
joemcgill: Rather than passing `$merchant_report_data` to the `map_price_benchmarks_response()` method, we can add it directly to $response_data for the product idea if it exists.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -111,24 +110,21 @@ protected function get_price_benchmarks_callback(): callable {
...
+			if ( ! $product_id ) {
 				return [];
 			}
 
-			return $report_data;
+			$report_args = [
+				'ids'      => [ $product_id ],
+				'fields'   => [ 'clicks', 'conversions' ],
+				'interval' => 'week',
+			];
```
joemcgill: We're going to need to adjust these args based on initial feedback about the way these queries are being processed in #2890. I'm going to try to clarify the requirements today.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -189,8 +185,16 @@ protected function map_price_benchmarks_response( array $benchmark_data, array $
...
+		if ( ! empty( $merchant_report_data ) ) {
+			foreach ( $merchant_report_data['results'] as $merchant_report_result ) {
+				$product_id = $merchant_report_result['segments']['offer_id'] ?? null;
+
+				if ( ! $product_id ) {
+					continue;
+				}
+
+				$mapped_data[ $product_id ]['merchant_report'] = $merchant_report_result['metrics'];
+			}
```
joemcgill: Related to my feedback in the `get_price_benchmarks_callback()` method, I think we should remove any metrics processing from this method and only have it be responsible for merging data from `PriceCompetitiveness` and `PriceInsights` tables.

While reviewing this method I also noticed a separate problem, which is that currently if either of the responses are empty, it will result in an empty array being returned. However, there are cases where we will have PriceCompetitiveness data but NO PriceInsights data (never the reverse), so it should only bail early if there is no results from the `$benchmarks_data`.

tests/Unit/API/Site/Controllers/MerchantCenter/PriceBenchmarksControllerTest.php
joemcgill: Smart. I'm seeing now that these results values are actually not returned directly from the API responses, but instead based on the mapping set for custom queries in their `set_initial_columns()` methods ([example](https://github.com/woocommerce/google-listings-and-ads/blob/25e420d039c3351b41b5b383ca90096b19eaaf9b/src/API/Google/Query/MerchantPriceSuggestionsQuery.php#L46-L61)). 

Currently, the logic we're using relies on the `MerchantProductReportQuery` class via the `MerchantReport::get_report_data()` method ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/925c3fb030447cdff4c10c0901c91a2ac300a282/src/API/Google/MerchantReport.php#L155)). That query doesn't set initial columns at all, so we'll need to do some experimentation to see how this should work. In fact, we probably shouldn't be using that method at all and instead should use the `MerchantReportQuery` directly. 

At any rate, I think we'll get the results in a shape that looks more like this (or whatever columns we've mapped the API responses to):

```php
		$report_data = [
			'results' => [
				[
					'offer_id'    => $product_id,
					'clicks'      => 734,
					'conversions' => 4,
				],
			],
		];
```

joemcgill (commented):
I've got some initial data to test with and am going to try to firm up some details here, but I'm leaving some initial feedback for us to address.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: The issue is that we can't retrieve `predicted_clicks_change_fraction` and `predicted_conversions_change_fraction` unless the data comes from `price_insights`.

If we include those fields by default in the `map_price_benchmarks_response` output, they will also be returned in cases where we don’t request them such as when no specific product ID is provided which leads to unnecessary or misleading data being included in the response.

joemcgill: I see. Since the `predicted_clicks_change_fraction` and `predicted_conversions_change_fraction` data will be included in the price insights data, I'm fine with us returning that by default in all cases then.

For now, I've not moved the metrics data out of the `map_price_benchmarks_response()` method, but we likely should.

joemcgill: Update: I've attempted to start splitting the metrics data from the data that gets returned in a regular query in [c4487a8](https://github.com/woocommerce/google-listings-and-ads/pull/2891/commits/c4487a8a3670c60bc9a971b70e40e5cd7e016134) but didn't quite get it sorted. Hopefully the direction I was headed makes sense.

A couple notes:

The `prepare_query_arguments()` method will filter out any args that aren't in the `get_collection_params()` method, so I've refactored things in `[a96b485](https://github.com/woocommerce/google-listings-and-ads/pull/2891/commits/a96b485f6fd94059ab8d0344ac4266f2873bb226) so that the previous `get_price_benchmarks_params()` method is now the `get_collection_params()` method that uses the parent params and then adds the additional args that are supported in the regular callback. I've then split out a new endpoint specifically for returning benchmarks data for a specific ID so we can handle the callback separately.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
joemcgill: Started to be handled in [c4487a8](https://github.com/woocommerce/google-listings-and-ads/pull/2891/commits/c4487a8a3670c60bc9a971b70e40e5cd7e016134) but merging in the data from the metrics still needs to be sorted.

joemcgill (commented):
@mukeshpanchal27 I've made a few more adjustments based on a better understanding of how the Conetnt API queries should be formatted.

For the metrics query, you can set a breakpoint at `MerchantQuery::query_results()` [here](https://github.com/woocommerce/google-listings-and-ads/blob/e4116cb3dd9bf8a679ddf1ff3826439f4b0859ae/src/API/Google/Query/MerchantQuery.php#L69) and see that the queries that are generated are in the form that we expect, like: 

```
SELECT segments.offer_id,metrics.clicks,segments.week FROM MerchantPerformanceView WHERE segments.offer_id IN ('12345')
```

If possible, it would be nice find a way to create PHPUnit tests for each of the `MerchantPriceBenchmarks` methods that produce these queries to ensure that they are producing the expected queries for the Content API.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: Data merged in https://github.com/woocommerce/google-listings-and-ads/pull/2891/commits/a150a9057cbe93b2fc1c8454678e7ab5cc86f47c

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
mukeshpanchal27: So as we introduce new route we needs to update the issue description that say that `/wp-json/wc/gla/mc/price-benchmark/?id=12345` now it looks like `/wp-json/wc/gla/mc/price-benchmark/12345`

mukeshpanchal27:
@joemcgill Thanks for pushing changes. In https://github.com/woocommerce/google-listings-and-ads/pull/2891/commits/a150a9057cbe93b2fc1c8454678e7ab5cc86f47c it fix the response shape and now unit test pass ✅ 

> If possible, it would be nice find a way to create PHPUnit tests for each of the `MerchantPriceBenchmarks` methods that produce these queries to ensure that they are producing the expected queries for the Content API.

I tried to look these but didn't get how i can write in unit tests. Can we do that in follow-up PR so these one is ready for QA as it's not blocker it's something nice to have tests?

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -97,6 +147,77 @@ protected function get_price_benchmarks_callback(): callable {
...
+			'description'       => __( 'Token to retrieve the next page.', 'google-listings-and-ads' ),
+			'type'              => 'string',
+			'validate_callback' => 'rest_validate_request_arg',
+		];
+
+		return $params;
+	}
+
+	/**
+	 * Get the query params for collections.
```
joemcgill:
```suggestion
	 * Get the query params for a single product.
```

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -89,6 +102,43 @@ protected function get_price_benchmarks_callback(): callable {
...
 				$response_data = $this->map_price_benchmarks_response( $benchmark_data, $price_insights_data );
+				return new Response( $response_data );
+			} catch ( Exception $e ) {
+				return $this->response_from_exception( $e );
+			}
+		};
+	}
+
+	/**
+	 * Get the callback function for the price benchmarks request.
```
joemcgill:
```suggestion
	 * Get the callback function for a specific price benchmark request.
```

joemcgill (commented):
@mukeshpanchal27 

I think this is looking really close. After going over everything together earlier today, I'm posting some notes here for posterity for what next steps we need to follow up on and listed which of us is doing what.

## Must complete

- Implement product helper for offer_ids – @mukeshpanchal27 
  - ❓check whether we need to use the product helper for formatting IDs for outgoing queries to the API
- Make all API queries support product ID – @joemcgill 
- Confirm Queries are being generated correctly – @joemcgill 

## TODO during bug bash

- Add PHPUnit tests for the query generation
- Mock Google API Responses for testing
- See if we need to add additional caching to Google API responses

## Nice to have

- Update metrics date query to use `LAST_WEEK_MON_SUN` function
- Use local data for REST responses rather than Google API
- Review and rename methods and classes for competitiveness, insights, and metrics and use benchmarks only for combined data
- Better documentation for the architecture

mukeshpanchal27:
@joemcgill In https://github.com/woocommerce/google-listings-and-ads/pull/2891/commits/6153f50e7ad3bbeb090b80385336a5068fbb0259 i implemented the helper function that get the product ID.

Initially i thought to use `ProductHelper::get_wc_product_id` method but it can't get in controller so i use shim function of there and update it slight to fix the edge case with `woocommerce_gpf_` ID.

joemcgill (approved):
@mukeshpanchal27 I was able to confirm that all the API queries are being the correctly generated, added support for ID for all three Content API queries and did a bit of cleanup on the expected response shapes.

I'm going to approve, but give it a look over and make sure it looks good to you before sending it for QA.

mukeshpanchal27:
```sql
SELECT  metrics.impressions,  metrics.clicks,  metrics.ctr FROM MerchantPerformanceView WHERE segments.date DURING LAST_WEEK_MON_SUN
```
@joemcgill The use of `LAST_WEEK_MON_SUN` in the query, it returns 200. Might be worth updating that in our bug base.

ankitguptaindia:
**QA/Test Report-** ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.8.1
* Theme:  
* WooCommerce - Version 9.8.4
* PHP: 8.3.0
* Web Server: Nginx
* Browser: Chrome - Version 136
* OS: macOS Sonoma 14.6.1 

</Details>


**Test Results -** 



Hi @mukeshpanchal27, Thanks for the update, now I am able to see empty array when using `siteurl/index.php?rest_route=/wc/gla/mc/price-benchmarks/12345` 


Screenshot 2025-05-09 at 3 41 19 PM

ankitguptaindia (approved):

joemcgill:
Thanks @mukeshpanchal27! Just noticed one small improvement from your changes and have opened https://github.com/woocommerce/google-listings-and-ads/pull/2900 to address that. Can you please review?

Issue #2911, opened 2025-05-14
Sort `mc/price-benchmarks` data by `effectiveness` (High to Low)


mukeshpanchal27:
### Describe the bug:

According to the PRD, the products returned from the `/mc/price-benchmarks` endpoint should be sorted by `effectiveness` in descending order (i.e., highest to lowest). This ensures that the products with the highest potential impact appear first.

However, the current implementation does not apply this sorting and returns the data in the default order from the query.

### Steps to reproduce:

1. Follow the setup instructions in #2906 to configure the test proxy for mocking API responses.
2. Send a GET request to the `/wc/gla/mc/price-benchmarks/` endpoint.
3. Observe the order of products in the response.

### Expected behavior:

The products should be sorted with the highest `effectiveness` values appearing first.

### Actual behavior:

The products are returned in the default order, without sorting by `effectiveness`.

### Additional details:

Consider extending the proxy test mock data to include varying `effectiveness` values so this behavior can be validated locally before deploying the change.

mukeshpanchal27:
The https://developers.google.com/shopping-content/guides/reports/query-language/structure#order_by allow to add the order by API query.

joemcgill:
We can only order by effectiveness for products that have `price_insights` data associated. 

The challenge we're going to have is that to gather all of the data for the price benchmarks feature, we're querying 3 different reports from the Content API:

- One query of the `PriceCompetitivenessProductView` table to get the base data (product pricing and price competitiveness data)
- Another query of the `PriceInsightsProductView` table to get price insights data (suggestions)
- One last query of the `MerchantPerformanceView` table to get metrics (clicks, conversions, etc).

We can't sort all of these queries by effectiveness, so in order to make this the default sorting, we would need to fetch all of this data into the local database in order to run a separate query that sorts and paginates the combined data.

Alternately, we would need to query the PriceInsights data first and map our responses based on the order that products are returned from that query based on updating that query so it's ordered by `price_insights.effectiveness DESC`.

Given the different ways we'd like to be able to work with the combined data, I'd lean towards pulling it all into a custom DB table, like we're already doing with price competitiveness data to calculate the Price Benchmark summary data.

mukeshpanchal27:
@joemcgill we have two options here:

**Option 1:** Instead of adding data in local DB or query the PriceInsights data first and map the responses could we add sorting in `get_price_benchmarks_callback` response?

[...]

PR #2923, opened 2025-05-20
Consolidate usePriceBenchmarkSuggestions selector.

### LINKED ISSUE #2904: Add support for fetching metrics data via `/price-benchmarks/<id>` in `ChangePriceModal`
## Issue
Description: Right now the `/wp-json/wc/gla/mc/price-benchmarks/` endpoint does not return metrics data which is displayed within the `ChangePriceModal` component. A new request should be made to `/wp-json/wc/gla/mc/price-benchmarks/<id>` to get the metrics data to populate the `ChangePriceModal` component.

## Implementation brief
* Update the `fetchPriceBenchmarkSuggestions` action to include a `productId` parameter.
	* The endpoint should be `/mc/price-benchmarks/<productID>` if `productId` is defined.
* Within `js/src/data/reducer.js`,
	* Update the `RECEIVE_PRICE_BENCHMARK_SUGGESTIONS` case to update the existing data if any if there's a `productId`
* Update the `getPriceBenchmarkSuggestions` selector and its resolver as well to include the `productId` parameter.
	* The selector should return `undefined` is there's no `clicks` and `conversions` properties, to trigger the resolver to fetch additional data.
* Delete the unused `js/src/hooks/usePriceBenchmarkSuggestionsProduct.js` hook and its references.
* Update the `ChangePriceModal` component to use the `usePriceBenchmarkSuggestions` hook with the `productId` as parameter.
* Update the `tests/e2e/utils/__fixtures__/price-benchmark-suggestions.json` to remove the `clicks`, `conversions`, `predicted_clicks_change` and `predicted_conversions_change` properties.
* Create `tests/e2e/utils/__fixtures__/price-benchmark-product-suggestions.json` to contain the data for a single product as per #2861

## Test coverage
- Existing tests should pass.

asvinb:

Closes #2904  .

1. Navigate to the Price Benchmark page.
2. Click the "Change Price" link for the desired product.
3. A modal will appear, displaying various pricing metrics and an input field prefilled with the suggested price from the API.
4. If a user enters a negative value or a price lower than the current sales price, an error message should appear, and the "Change Price" button will be disabled.
5. Upon entering a valid price and clicking "Change Price", the modal should close and the product’s price in the table should update accordingly.
6. Confirm that the product price has been successfully updated in WooCommerce.

### Changelog entry

>


js/src/data/actions.js
```diff
@@ -1229,15 +1229,22 @@ export function* fetchPriceBenchmarkSummary() {
...
+
 		const data = yield apiFetch( {
-			path: `${ API_NAMESPACE }/mc/price-benchmarks`,
+			path,
 		} );
 
 		return {
 			type: TYPES.RECEIVE_PRICE_BENCHMARK_SUGGESTIONS,
 			data,
+			productId,
```
joemcgill, line 1247: Looks like this could still be undefined at this point. Is that ok or do we need to set it as `null`?

js/src/data/reducer.js
```diff
@@ -534,8 +534,59 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
 
 		case TYPES.RECEIVE_PRICE_BENCHMARK_SUGGESTIONS: {
-			const { data } = action;
-			return setIn( state, 'price_benchmark.suggestions', data );
+			const { data, productId } = action;
+
+			if ( ! productId ) {
+				// Merge the new suggestions with the existing ones in case we have the same product in both arrays.
+				// This is a bit tricky because we need to merge the objects based on the product ID.
+				const map = keyBy(
```
joemcgill, line 542: Rather than needing to key the price benchmarks by id here each time, I think we should consider storing the data in this shape all the time. Doing so would make it easier to get and set price benchmark objects by their `product_id`.

js/src/data/reducer.js
```diff
@@ -534,8 +534,59 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+					state.price_benchmark.suggestions,
+					'product.id'
+				);
+
+				// For each item in the new data either merge with existing or add new
+				data.forEach( ( item ) => {
+					if ( item.product?.id ) {
```
joemcgill, line 549: This data will always have a product ID, so we don't need to check. Additionally, I don't think we need to merge the old and new objects. The new ones can totally replace the old ones.

js/src/data/reducer.js
```diff
@@ -534,8 +534,59 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+								...item,
+							};
+						} else {
+							// If item doesn't exist, add it
+							map[ item.product.id ] = item;
+						}
+					}
+				} );
+
+				const mergedData = Object.values( map );
```
joemcgill, line 563: If we store the data by key we won't need to convert back to values here. Instead, we may want to sort the data by ID

js/src/data/selectors.js
```diff
@@ -417,11 +417,34 @@ export const getPriceBenchmarkSummary = ( state ) => {
...
+	if ( productId ) {
+		const product = state.price_benchmark.suggestions.filter(
+			( suggestion ) => {
+				return (
+					suggestion.product.id === productId &&
+					'clicks' in suggestion &&
+					'conversions' in suggestion
+				);
+			}
+		);
```
joemcgill, line 440: Based on the changes being discussed in #2911, we won't need to include this extra logic, because clicks and conversions data will always be included in the response.

js/src/data/selectors.js
```diff
@@ -417,11 +417,34 @@ export const getPriceBenchmarkSummary = ( state ) => {
...
+
+		if ( ! product.length ) {
+			return undefined;
+		}
+
+		return product[ 0 ];
+	}
+
 	return state.price_benchmark.suggestions;
```
joemcgill, line 449: If we change the shape of the suggestions to be keyed by ID, we can return Object.values here instead.

joemcgill (changes requested):
Hi @asvinb, I've left a bit of feedback here. However, since we're planning to make so many changes as a result of #2911, I think we may want to hold off on progressing this further and make sure that the refactor work resulting from that ticket resolved #2904 as well. A lot of this PR should be reusable in that effort. What do you think?

joemcgill:
Closing this PR as no longer needed

PR #2924, opened 2025-05-20
Update effectiveness values in proxy data

joemcgill:

Related to #2911. We were able to confirm that the API queries to the `PriceInsightsProductView` can support ordering by effectiveness by including `ORDER BY price_insights.effectiveness DESC`.

At the same time, the API responses that we tested returned a slightly different value format than we were expecting. Rather than returning the array value of the enum, it returns the string value as documented in https://developers.google.com/shopping-content/reference/rest/v2.1/reports/search#Effectiveness.

This PR updates the test-proxy response data for the `PriceInsightsProductView` queries to return the expected shape.

This can be used to test the changes for #2911.

### DIFF
--- a/tests/proxy/mocks/mc/price-benchmarks/price-insights-item.json
+++ b/tests/proxy/mocks/mc/price-benchmarks/price-insights-item.json
@@ -14,7 +14,7 @@
 				"predictedImpressionsChangeFraction": "0.12609300017356873",
 				"predictedClicksChangeFraction": "0.508745014667511",
 				"predictedConversionsChangeFraction": "2.3431060314178467",
-				"effectiveness": 3
+				"effectiveness": "HIGH"
 			}
 		}
 	]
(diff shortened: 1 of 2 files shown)


mukeshpanchal27 (approved):

mukeshpanchal27 (changes requested):
Since we're mocking the API response and ensuring the results are ordered by effectiveness in descending order, do we need to update the expected ordering in the proxy test setup? Currently, it appears incorrect.

joemcgill:
@mukeshpanchal27 If we end up changing the query we're sending to order by effectiveness, then yes. As I mentioned in [my comment on the related ticket](https://github.com/woocommerce/google-listings-and-ads/issues/2911#issuecomment-2895640669):

> I've opened https://github.com/woocommerce/google-listings-and-ads/pull/2924 to update the values to the new format, though they are currently still in order of product_view.id. If we decide to use the API ordering directly for these queries, we would want to update the mocks ordering as well.

Initially, I just wanted to make sure the effectiveness values in the existing mocks were accurate.

mukeshpanchal27 (approved):

PR #2934, opened 2025-05-26
Add server side logic to DataViews interaction

### LINKED ISSUE #2933: Update sorting and filtering in Price Suggestions
## User Story
As an engineer, I want to avoid loading large datasets on the frontend by querying the API on demand with pagination, sorting, and search parameters, so that only a small, relevant subset of data is fetched and cached per request.

## Acceptance Criteria
* DataViews interactions are driven by server responses instead of frontend filtering.
* Sorting, searching, and pagination should result in a new request to the REST API.
* The total number of pages should update if the total possible results change (e.g., search results will return fewer price benchmark objects than the default view.
* By default, the price benchmarks suggestions should be ordered by effectiveness (High to Low)
* If the same search or sort is made without refreshing the page, it should NOT result in another REST request.

## Implementation Brief
* Update the `fetchPriceBenchmarkSuggestions` action to accept an `args` parameter which is an object containing the following properties:
	* page
	* per_page
	* search
	* order
	* orderby
* The `args` should be added to the `price-benchmarks` endpoint using `addQueryArgs`.
* Update the `getPriceBenchmarkSuggestions` selector/resolver to include the `args` parameter.
* Update `js/src/data/reducer.js` to change `price_benchmark.suggestions` from an array to an object with the following properties.
	* items: Contains product info keyed by the product ID.
	* queries: Contains query results keyed in by a slug generated from the `args` passed to `fetchPriceBenchmarkSuggestions`. The query results should be stored in the `items` property and pagination data in the `meta` property.
		* The items should contains product IDs only, referencing products in the parent `items` property.
			* Before adding the data in the store, some data manipulation needs to occur to upsert the product info.
* Update `PriceBenchmarkSuggestions` to:
	* include the `args` parameter to `usePriceBenchmarkSuggestions` with the `DataViews` `view` property mapped to the properties in the first bullet point.
		* `view.perPage` -> `per_page`
		* `view.search` -> `search`
		* `view.page` -> `page`
		* `view.sort.direction` -> `order`
		* `view.sort.field` -> `orderby`
	* remove usage of the `filterSortAndPaginate` function to instead manually handle the pagination via the  `paginationInfo` prop of `DataViews`.
		* Sorting will be handled automatically when the query will be returned from the API.
	* Use the `isLoading` prop of `DataViews` to indicate if there's any query in progress instead of rendering `TablePlaceholder`
	* The `EmptyMetricsNotice` component should be rendered only when there's no data on the initial query (default query).
	* Add necessary function to format the data to be stored/retrieved from the store.

## Test coverage
* Update mock data to make sure existing tests pass.

asvinb:

Closes #2933 .

This requires #2911 to be merged or ready for QA as well.

1. Visit the Price Benchmark page.
2.  The initial set of data loads automatically.
3. Clicking Next Page triggers a request to fetch the next set of results.
4. Going back to the first page retrieves data directly from the local data store—no additional request is made.
5. The same behaviour applies to search and sorting—results are retrieved from the store if previously fetched.

### Changelog entry

>


js/src/data/reducer.js
```diff
@@ -534,31 +538,63 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+				data.results,
+				( item ) => item.product.id
+			);
+
+			const stateSetter = chainState( state, [
+				'price_benchmark',
+				'suggestions',
+			] );
+
+			return stateSetter
```
joemcgill, line 579: ✨ nice

js/src/pages/price-benchmark/price-benchmark-suggestions/index.js
```diff
@@ -246,44 +245,46 @@ const PriceBenchmarkSuggestions = ( { isViewportMobile } ) => {
 		recordGlaEvent( 'gla_price_benchmarks_shown', {
 			context: PRICE_BENCHMARK_SUGGESTIONS_CONTEXT,
 			suggestions: suggestions.length,
+			per_page: view.perPage,
 		} );
-	}, [ hasFinishedResolution, suggestions ] );
+	}, [ hasFinishedResolution, suggestions, view ] );
 
-	if ( hasFinishedResolution && suggestions.length === 0 ) {
+	// If there are no suggestions and the query params are default, which means we are loading the initial set of data, show an empty notice.
```
joemcgill, line 259: 🤔 Hmmm. What will be shown if someone tries a search that returns no results?

js/src/data/reducer.js
```diff
@@ -534,31 +538,63 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+						[ args.product_id ],
+					],
+					data
+				);
+			}
+
+			const key = generateKeyFromObject( args );
+
+			// Set the product ID as key for easier lookup of products.
+			const productsById = keyBy(
```
joemcgill, line 569: I think we need to handle the case where a query returns no results (i.e. and empty array). In that case I think we only want to save the query and empty array set to the results.

joemcgill (changes requested):
@asvinb this looks pretty good to me. This is hard to test without a working BE unless we create mocks for each filter and sort option but it looks like the queries are being generated correctly.

One thing that I was not able to confirm is that when the Change Price button is clicked, we should already have all of the data we need in the data store for that item now that we're planning including all of the metrics data in the response, so we shouldn't need to send an additional request to the price-benchmarks/{id} endpoint to get that data and render the modal.

Also, when I tried changing the price, the request to the `/wc/v3/products/33` endpoint returned a 404 error with the following response:

```json
{
    "code": "woocommerce_rest_invalid_product_id",
    "message": "To manipulate product variations you should use the \/products\/&lt;product_id&gt;\/variations\/&lt;id&gt; endpoint.",
    "data": {
        "status": 404
    }
}
```

js/src/pages/price-benchmark/price-benchmark-suggestions/index.js
asvinb: @joemcgill The following message will be shown:
image

which is different from the `EmptyMetricsNotice` which kinda makes sense because `EmptyMetricsNotice` is shown when there are no metrics at all. The case you described is when you have metrics and there are no results for your search term.

js/src/data/reducer.js
asvinb: Sure thing @joemcgill . Updated!

asvinb:
@joemcgill Excellent catch for the extra request being made when we already have all the data. I thought the resolver would be triggered only if we just return `undefined` from the selector. However turns out that there's some sort of internal resolver cache and if you want to add data, it's done via actions with the `receive` prefix. So to let WP know that a product for a particular ID exists, we should have for e.g called the `receivePriceBenchmarkProduct` action, with a separate selector/resolver to get an individual product. In our case, it wouldn't really work since we have an array of products which is being built dynamically. I found an alternative solution by updating the `usePriceBenchmarkSuggestions` hook and checking if the product exists and return it. If not, it'll continue with the hook execution and request the data from the server.

Also for the error when updating the product price, can you please ensure there's a WooCommerce product with 33 as ID .

joemcgill:
Thanks for the explanation @asvinb. 

> In our case, it wouldn't really work since we have an array of products which is being built dynamically. I found an alternative solution by updating the usePriceBenchmarkSuggestions hook and checking if the product exists and return it.

It may make sense to have separate selectors for this, but it seems odd that there would be a selector to get all items from the store and another to get the items based on query params. If we're making two selectors, I'd prefer to see this as 

- `getPriceBenchmarkSuggestions` - returns a list of items based on the query params
- `getPriceBenchmarkSuggestion` - returns a single item based on product ID

The `usePriceBenchmarkSuggestions` hook could then call `selector.getPriceBenchmarkSuggestion( productID )` and not need to retrieve the whole list.

Also:

> Also for the error when updating the product price, can you please ensure there's a WooCommerce product with 33 as ID 

I did double check that when seeing the error. You mind seeing if you are unable to reproduce this error?

asvinb:
@joemcgill I've added the `getPriceBenchmarkSuggestion` selector which the `usePriceBenchmarkSuggestions` is using now. Thanks!
As for the error, I was able to reproduce it. It's when a product has variations. I've updated the PR to cater for that.
https://github.com/woocommerce/google-listings-and-ads/pull/2934/commits/4e5903a9ad1f9d170dff68648b9b1167a337b830
Let me know what you think.

joemcgill (commented):
@asvinb all these changes are testing well for me when using the mocked data defined in the test fixtures. However, while reviewing #2935 I've noticed that the assumptions made in the back end for the REST API and the assumptions made in this PR are not quite the same. Namely:

The schema returned by the `/mc/price-benchmarks` endpoint currently looks like this:

```json
{
    "results": {
        "product": {
            "description": "Product details.",
            "type": "object",
            "properties": {
                "id": {
                    "type": "integer"
                },
                "thumbnail": {
                    "type": "string",
                    "format": "uri"
                },
                "title": {
                    "type": "string"
                }
            }
        },
        "effectiveness": {
            "description": "Effectiveness score.",
            "type": "integer",
            "enum": [
                0,
                1,
                2,
                3
            ]
        },
        "country_code": {
            "description": "Country code.",
            "type": "string"
        },
        "currency_code": {
            "description": "Currency code.",
            "type": "string"
        },
        "product_price": {
            "description": "Current price of the product on Google.",
            "type": "number"
        },
        "benchmark_price": {
            "description": "Average benchmark price of the product on Google.",
            "type": "number"
        },
        "benchmark_price_currency_code": {
            "description": "Benchmark price currency code.",
            "type": "string"
        },
        "price_gap": {
            "description": "Price gap between the product price and the benchmark price on Google.",
            "type": "number"
        },
        "suggested_price": {
            "description": "Suggested price for the product.",
            "type": "number"
        },
        "suggested_price_currency_code": {
            "description": "Suggested price currency code.",
            "type": "string"
        },
        "clicks": {
            "description": "Current clicks.",
            "type": "integer"
        },
        "impressions": {
            "description": "Current impressions.",
            "type": "integer"
        },
        "ctr": {
            "description": "Click-through rate.",
            "type": [
                "string",
                "null"
            ]
        },
        "conversions": {
            "description": "Current conversions.",
            "type": "integer"
        },
        "predicted_impressions_change": {
            "description": "Expected uplift in impressions (fraction).",
            "type": [
                "string",
                "null"
            ]
        },
        "predicted_clicks_change": {
            "description": "Expected uplift in clicks (fraction).",
            "type": [
                "string",
                "null"
            ]
        },
        "predicted_conversions_change": {
            "description": "Expected uplift in conversions (fraction).",
            "type": [
                "string",
                "null"
            ]
        },
        "price_compared_with_benchmark": {
            "description": "Comparison of price with benchmark.",
            "type": "integer",
            "enum": [
                0,
                1,
                2,
                3
            ]
        }
    },
    "total": {
        "description": "The total number of benchmarks that are available.",
        "type": "integer"
    }
}
```

The main differences that I can see include:

1. `regular_price` is `product_price`
2. `price_on_google` is `benchmark_price`
3. It doesn't seem like you're using the currency code for any of the price values (i.e., `currency_code`, `benchmark_price_currency_code`, `suggested_price_currency_code`). We should probably be formatting the currency display in the front end based on the currency being used.

When sending GET requests to the API, the accepted args are:

```json
{
    "page": {
        "description": "Current page of the collection.",
        "type": "integer",
        "default": 1,
        "minimum": 1,
        "required": false
    },
    "per_page": {
        "description": "Maximum number of items returned in the results.",
        "type": "integer",
        "default": 10,
        "minimum": 1,
        "maximum": 100,
        "required": false
    },
    "search": {
        "description": "Limit results to those matching a string.",
        "type": "string",
        "required": false
    },
    "order": {
        "description": "Order sort attribute ascending or descending.",
        "type": "string",
        "enum": [
            "asc",
            "desc"
        ],
        "default": "desc",
        "required": false
    },
    "orderby": {
        "description": "Sort collection by attribute.",
        "type": "string",
        "enum": [
            "effectiveness",
            "id",
            "price_on_google",
            "regular_price",
            "suggested_price"
        ],
        "default": "effectiveness",
        "required": false
    }
}
```

A few observations here:

1. We should probably use the same arg names for `regular_price` and `price_on_google` in both the GET args and the return values (that's on me).
2. We can't currently sort on title nor on price gap. For title, it's part of the same column that also includes product_id, so I'd prefer that we update that field to sort by product_id instead of title for now. I think we could make price_gap non sortable for now and open a new ticket to add that in a separate update.

Let me know if you think we should make any changes to the API design other than the ones I've outlined. I'd like to get both parts merged together for QA purposes, if possible.

asvinb:
@joemcgill I've made the updates to the properties in a separate PR:
https://github.com/woocommerce/google-listings-and-ads/pull/2949
Can you kindly take a look please?

For the sorting issue, sorting has been disabled for `price_gap`. For sorting by `id`, I renamed the `title` property to `id` and then used that as `titleField` so that it renders correctly within `DataViews`.

For the currency, we are using the `useAdsCurrency` hook. Am wondering if the currency would differ from what's returned from the BE? 🤔

js/src/hooks/usePriceBenchmarkSuggestions.js
```diff
@@ -10,18 +10,33 @@ import { STORE_KEY } from '~/data/constants';
...
-	}, [] );
+			if ( productId ) {
+				const item = selector.getPriceBenchmarkSuggestion( productId );
+
+				if ( item ) {
+					return {
+						data: item,
+						hasFinishedResolution: true,
+					};
+				}
```
joemcgill, line 27: If `item` is not found, won't this selector end up returning all price benchmarks suggestions? Is that intentional?

js/src/pages/price-benchmark/change-price-modal/index.js
```diff
@@ -140,9 +129,21 @@ const ChangePriceModal = ( { productId, onRequestClose, onPriceChange } ) => {
 		);
 	}
 
+	const {
+		effectiveness,
+		product_price: regularPrice,
+		benchmark_price: priceOnGoogle,
```
joemcgill: For consistency, could we also rename `regularPrice` and `priceOnGoogle` throughout the codebase?

js/src/data/reducer.js
```diff
@@ -534,31 +538,74 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+					return state;
+				}
 
-			const newSuggestionsState = [
-				...state.price_benchmark.suggestions,
-			];
-			const productIndex = newSuggestionsState.findIndex(
-				( { product } ) => product.id === productId
-			);
+				// If no results, return the state unchanged.
```
joemcgill: This comment seems incorrect

js/src/data/reducer.js
```diff
@@ -534,31 +538,74 @@ const reducer = ( state = DEFAULT_STATE, action ) => {
...
+				return chainState( state, [ 'price_benchmark', 'suggestions' ] )
+					.setIn( [ 'queries', [ key ], 'items' ], [] )
+					.setIn( [ 'queries', [ key ], 'meta', 'totalItems' ], 0 )
+					.end();
+			}
 
-			if ( productIndex >= 0 ) {
-				newSuggestionsState[ productIndex ].regular_price =
-					regularPrice;
+			// Retrieve product suggestions for a specific product.
```
joemcgill: This comment is unclear. This is "setting" the data, not "retrieving", correct?

joemcgill (approved):
@asvinb Thanks for the updates. I've left a few comments, but nothing really blocking so I'm pre-approving pending any feedback you want to address.

I've merged the API changes from #2935 into the feature branch and refreshed this PR so you should be able to test everything e2e to see if there are other issues before sending it to QA. I'll continue some testing in the meantime.

Update:

> For the currency, we are using the `useAdsCurrency` hook.

That makes sense. It would be worth reviewing that hook just to make sure the value used in that hook aligns with what the product currency is using.

js/src/hooks/usePriceBenchmarkSuggestions.js
asvinb: Nope it was not 😁 . Fixed @joemcgill . We are now returning data: null if product is not found.

js/src/pages/price-benchmark/change-price-modal/index.js
asvinb: Sure thing! Updated!

js/src/data/reducer.js
asvinb: That's true. It was some outdated comment and I've removed it since the code is self explanatory.

js/src/data/reducer.js
asvinb: Correct and I've removed it @joemcgill . See my previous comment.

ankitguptaindia:
**QA Status-** Verified ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.8.1
* Theme:  
* WooCommerce - Version 9.9.0-rc.1
* PHP: 8.3.0
* Web Server: Nginx
* OS: macOS Sonoma 14.6.1 

</Details>




**Test Results -** 

Verified and confirmed that previously fetched data (including paginated, sorted, and searched results) is retrieved from the local store without making additional requests.

https://github.com/user-attachments/assets/35a51edf-3429-4355-801d-dd40c955c8dc

PR #2935, opened 2025-05-27
Fix sorting and pagination in the `mc/price-benchmarks` endpoint

### LINKED ISSUE #2911: Sort `mc/price-benchmarks` data by `effectiveness` (High to Low)

mukeshpanchal27:

Closes #2911.

- Remove local DB `wp_gla_merchant_price_benchmarks` if you have already in DB.
- Add `add_filter( 'woocommerce_gla_force_run_install', '__return_true' );` filter in your mu-plugin to create DB table.
- Visit any admin page
- Check local DB `wp_gla_merchant_price_benchmarks` table is created or not.
- Testing REST API callback permission.
  - If using a service like PostMan, follow [these instructions](https://github.com/woocommerce/google-listings-and-ads/wiki/Testing-GLA-endpoints-with-Postman) to set up auth cookies and nonce.
  - To force the permissions to return true, you can use the following snippet:
  
```php
/**
 * Filters user capabilities to ensure `manage_woocommerce` capability always returns true.
 *
 * This function modifies the `$allcaps` array to grant the `manage_woocommerce` capability
 * to the current user if it is being checked.
 *
 * @param array $allcaps An array of all the capabilities of the user.
 * @param array $caps An array of capabilities being checked.
 * @return array Filtered array of all capabilities.
 */
function filter_allow_manage_woocommerce_caps( $allcaps, $caps ) {
	if ( isset( $caps[0] ) && 'manage_woocommerce' === $caps[0] ) {
		$allcaps[ $caps[0] ] = true;
	}
	return $allcaps;
}
add_filter( 'user_has_cap', 'filter_allow_manage_woocommerce_caps', 10, 2 );
```

- Use instruction from #2890 to set up the test-proxy to mock API responses
- Send a GET request to the /wc/gla/mc/price-benchmarks/ endpoint
- Review the response returned from the price-benchmarks endpoint
- Test schedule job.
  - Visit `Tools > Scheduled Actions` from the admin menu `/wp-admin/tools.php?page=action-scheduler` and verify that the `gla/jobs/update_merchant_price_benchmarks/process_item` job has been scheduled.
  - If the job hasn't already run, you can run it from the job list
  - Confirm that the data is imported to our local database

### Changelog entry

>


src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -21,6 +22,196 @@
...
+		 * @see ProductHelper::get_wc_product_id
+		 */
+		return (int) apply_filters( 'woocommerce_gla_get_wc_product_id', $wc_product_id, $mc_product_id );
+	}
+
+	/**
+	 * Retrieves the product thumbnail URL.
+	 *
+	 * @param int $product_id WooCommerce product ID.
+	 * @return string|null Product thumbnail URL or null if not found.
```
joemcgill: The docs don't match the implementation

```suggestion
	 * @return string Product thumbnail URL or an empty string if not found.
```

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -21,6 +22,196 @@
...
+	}
+
+	/**
+	 * Gets the product ID from the Google product ID.
+	 *
+	 * @param string $mc_product_id Simple product ID (`merchant_center_id`) or
+	 *                              namespaced product ID (`online:en:GB:merchant_center_id`)
+	 * @return int the ID for the WC product linked to the provided Google product ID (0 if not found)
+	 */
+	public function get_product_id( string $mc_product_id ): int {
```
joemcgill: Can we now remove this helper and replace it with the `ProductHelper::get_wc_product_id()` method?

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -43,18 +234,36 @@ public function update_price_benchmarks(): void {
 			$query->reload_data();
 
 			// Insert new benchmark data.
-			foreach ( $benchmarks as $benchmark ) {
+			foreach ( $benchmarks as $id => $benchmark ) {
+				$performance_data = $this->get_performance_data( $id );
```
joemcgill: Now that we're getting all the Metrics data at once, let's avoid individual API calls based on IDs and instead query all the `MerchantPerformanceView` data in a single query and map it to the `PriceCompetitivenessProductView` and `PriceInsightsProductView` in one process.

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -43,18 +234,36 @@ public function update_price_benchmarks(): void {
...
-						'product_id'                    => $benchmark['offer_id'],
-						'id'                            => $benchmark['id'],
-						'price_micros'                  => $benchmark['price_micros'],
-						'currency_code'                 => $benchmark['currency_code'],
-						'country_code'                  => $benchmark['country_code'],
-						'benchmark_price_micros'        => $benchmark['benchmark_price_micros'],
-						'benchmark_price_currency_code' => $benchmark['benchmark_price_currency_code'],
-						'price_compared_with_benchmark' => $price_compared_with_benchmark,
+						'product_id'                      => $benchmark['id'],
+						'mc_product_id'                   => $benchmark['id'],
```
joemcgill: `product_id` should be the WC product ID. `mc_product_id` should be the ID returned from the `PriceCompetitivenessProductView` or `PriceInsightsProductView` queries (e.g., `online:en:US:12345`).

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -43,18 +234,36 @@ public function update_price_benchmarks(): void {
...
+			foreach ( $benchmarks as $id => $benchmark ) {
+				$performance_data = $this->get_performance_data( $id );
+				if ( ! empty( $performance_data ) ) {
+					// Combine metrics data into the response for the specific product.
+					$benchmark = array_merge( $benchmark, $performance_data );
+				}
+
 				$price_compared_with_benchmark = $this->price_compared_with_benchmark( $benchmark['price_micros'], $benchmark['benchmark_price_micros'] );
+				$effectiveness                 = $this->get_effectiveness( $benchmark['effectiveness'] );
 				$query->insert(
```
joemcgill: This whole insert could be simplified a lot if we match and normalize the column keys during the mapping process, that way we could simply do something like: 

```php
foreach( $benchmarks as $benchmark_item ) {
  $query->insert( $benchmark_item );
}
```

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -101,15 +310,145 @@ public function get_summary(): array {
...
+		$query = $this->container->get( MerchantPriceBenchmarksQuery::class );
+
+		// Apply filters.
+		if ( ! empty( $args['include'] ) && is_array( $args['include'] ) ) {
+			$query->where( 'product_id', $args['include'], 'IN' );
+		}
+
+		if ( ! empty( $args['search'] ) ) {
+			$search = trim( $args['search'] );
+			// TODO: Implement a search mechanism.
```
joemcgill: We could consider adding product title to the price benchmarks DB if it's easier, or search the products to get a list of IDs and then use those IDs like `$query->where( 'product_id', $results, 'IN' )`

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -101,15 +310,145 @@ public function get_summary(): array {
...
+		$defaults = [
+			'include' => null,
+			'offset'  => 0,
+			'limit'   => 10,
+			'search'  => null,
+			'order'   => 'DESC',
+			'orderby' => 'mc_insights_effectiveness',
+		];
+
+		$args = array_merge( $defaults, $args );
```
joemcgill: You can use `wp_parse_args()` here.

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -101,15 +310,145 @@ public function get_summary(): array {
...
+		}
+
+		if ( ! empty( $args['search'] ) ) {
+			$search = trim( $args['search'] );
+			// TODO: Implement a search mechanism.
+		}
+
+		// Set order and orderby.
+		$order   = strtoupper( $args['order'] ) === 'ASC' ? 'ASC' : 'DESC';
+		$orderby = ! empty( $args['orderby'] ) ? $args['orderby'] : 'id';
```
joemcgill: Since we're defining the defaults above, we should not need fallbacks here.

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -21,6 +22,196 @@
...
 	use ContainerAwareTrait;
+	use PluginHelper;
+
+	/**
+	 * Get the price benchmarks base data.
+	 *
+	 * @param array $args Query arguments.
+	 * @return array Mapped response data.
+	 */
+	private function get_price_benchmarks_base_data( $args ) {
```
joemcgill: Now that we're merging data from all 3 Content API queries, we no longer need to separate "base data" from the merchant performance data. Let's fetch all 3 sources and pass them all to the `map_price_benchmarks_response()` method to format the data for inserting into the store DB.

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -21,6 +22,196 @@
...
+		return $thumbnail_url ?? '';
+	}
+
+	/**
+	 * Retrieves performance data (clicks and conversions) for a given product ID.
+	 *
+	 * @param int|string $id The product ID to fetch performance data for.
+	 * @return array Associative array with 'clicks' and 'conversions' keys, or empty array on failure.
+	 */
+	public function get_performance_data( $id ): array {
```
joemcgill: This method shouldn't be needed if we can call `$merchant->get_merchant_performance_data()` directly, similar to how we are calling

```
$benchmark_data = $merchant->get_price_comparisons_data( $args );
$price_insights_data = $merchant->get_price_insights_data( $args );
```

src/MerchantCenter/PriceBenchmarks.php
```diff
@@ -101,15 +310,145 @@ public function get_summary(): array {
 			$job->schedule();
```
joemcgill: Now that we're relying on the local DB being up to date for both the summary data and for the price benchmarks list, we should look at moving the responsibility for scheduling this job elsewhere and make this method only responsible for getting and formatting the summary data.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -94,17 +94,18 @@ protected function prepare_query_arguments( Request $request ): array {
...
-		return function ( Request $request ) {
+		return function () {
 			try {
-				$response_data = $this->get_price_benchmarks_base_data(
-					$this->prepare_query_arguments( $request )
-				);
+				/** @var PriceBenchmarks $price_benchmarks */
+				$price_benchmarks = $this->container->get( PriceBenchmarks::class );
+
+				$response_data = $price_benchmarks->get_price_benchmarks_data();
```
joemcgill: We'll still need to accept and prepare query args here and use those as the args for the get_price_benchmarks_data method.
```suggestion
				$response_data = $price_benchmarks->get_price_benchmarks_data( $this->prepare_query_arguments( $request ) );
```

joemcgill (commented):
@mukeshpanchal27 good progress so far. I've left some initial feedback.

When you do these updates, can you add some testing instructions to help confirm that both the data is being properly added to the database from the Google APIs and that all of the required query params we need to support are being accounted for when we retrieve the data from the local DB?

src/MerchantCenter/PriceBenchmarks.php
mukeshpanchal27: Updated in https://github.com/woocommerce/google-listings-and-ads/pull/2935/commits/2749a502eec8bcde1e1d5c732a1552210ad35a6c

src/MerchantCenter/PriceBenchmarks.php
mukeshpanchal27: Addressed in https://github.com/woocommerce/google-listings-and-ads/pull/2935/commits/d26d2d0f09451c4043f13b751a62266b91fbea4e

src/MerchantCenter/PriceBenchmarks.php
mukeshpanchal27: Nice catch. it look more simplified.

src/MerchantCenter/PriceBenchmarks.php
mukeshpanchal27: Addressed in https://github.com/woocommerce/google-listings-and-ads/pull/2935/commits/b52ed875410801918bee3071cd67b49cbb33d199

src/MerchantCenter/PriceBenchmarks.php
mukeshpanchal27: If we store the title in the database and someone later changes the product title in the WC store, will the changes sync quickly? If not, we might run into issues with the DB approach. What do you think?

src/MerchantCenter/PriceBenchmarks.php
mukeshpanchal27: Instead, can we schedule the job during plugin load? Since it's a recurring event, it will automatically reschedule itself once the first execution is completed.

src/MerchantCenter/PriceBenchmarks.php
mukeshpanchal27: Update in https://github.com/woocommerce/google-listings-and-ads/pull/2935/commits/4ada6dff61130bc4904e1d076bcda15f08a620f6

mukeshpanchal27:
@joemcgill I'm having trouble figuring out how to mock the `get_price_benchmarks_response` method in the tests. If you have a moment, could you please help by adding the relevant test coverage?

src/MerchantCenter/PriceBenchmarks.php
joemcgill: I'm having a bit of trouble fully understanding how all of the pieces of the Jobs system is meant to work, but found that the `JobInitializer` class [has some special handling ](https://github.com/woocommerce/google-listings-and-ads/blob/c5c67ec46c7ad7218ecc554edae7e8e952375712/src/Jobs/JobInitializer.php#L43-L72)for certain types of jobs to register a hook for scheduling and another that automatically calls that hook once a day.

I've made that change in 6888aee and it seems to be working.

mukeshpanchal27: Seems it set schedule job 3AM everyday.

https://github.com/user-attachments/assets/36ddc20d-7d50-41ed-9898-b7e5312b9bd7

mukeshpanchal27:
@joemcgill I have updated the Testing instruction for the PR.

src/MerchantCenter/PriceBenchmarks.php
joemcgill: Correct. It seems like any jobs implementing the `RecurringJobInterface` will automatically schedule a 'start' event to run every day at 3am, but that event is only used to ensure the job is scheduled.

image

You can see here that there are two scheduled jobs:

* `gla/jobs/update_merchant_price_benchmarks/start` calls the start hook which causes the `UpdateMerchantPriceBenchmarks::schedule()` method to be called.
* `gla/jobs/update_merchant_price_benchmarks/process_item` runs every 12 hours and calls the `UpdateMerchantPriceBenchmarks::process_items()` method.

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
```diff
@@ -96,15 +96,16 @@ protected function prepare_query_arguments( Request $request ): array {
...
-					$this->prepare_query_arguments( $request )
-				);
+				/** @var PriceBenchmarks $price_benchmarks */
+				$price_benchmarks = $this->container->get( PriceBenchmarks::class );
+
+				$response_data = $price_benchmarks->get_price_benchmarks_data( $this->prepare_query_arguments( $request ) );
 
 				if ( empty( $response_data ) ) {
 					return new Response( [] );
 				}
```
jjgrainger: @mukeshpanchal27 @joemcgill very minor thing but is this check necessary? If the `$response_data` is empty is it not the same as the following line `return new Response( $response_data )`?

src/DB/Table/MerchantPriceBenchmarksTable.php
```diff
@@ -56,14 +67,25 @@ public static function get_raw_name(): string {
...
+			'mc_insights_suggested_price_currency_code'    => true,
+			'mc_insights_predicted_impressions_change_fraction' => true,
+			'mc_insights_predicted_clicks_change_fraction' => true,
+			'mc_insights_predicted_conversions_change_fraction' => true,
+			'mc_insights_effectiveness'                    => true,
+			'mc_metrics_clicks'                            => true,
+			'mc_metrics_impressions'                       => true,
+			'mc_metrics_ctr'                               => true,
+			'mc_metrics_conversions'                       => true,
+			'price_compared_with_benchmark'                => true,
```
jjgrainger: Looks like this needs formatting
```suggestion
			'product_id'                                        => true,
			'mc_product_id'                                     => true,
			'mc_product_offer_id'                               => true,
			'mc_product_price_micros'                           => true,
			'mc_product_currency_code'                          => true,
			'mc_price_country_code'                             => true,
			'mc_price_benchmark_price_micros'                   => true,
			'mc_price_benchmark_price_currency_code'            => true,
			'mc_insights_suggested_price_micros'                => true,
			'mc_insights_suggested_price_currency_code'         => true,
			'mc_insights_predicted_impressions_change_fraction' => true,
			'mc_insights_predicted_clicks_change_fraction'      => true,
			'mc_insights_predicted_conversions_change_fraction' => true,
			'mc_insights_effectiveness'                         => true,
			'mc_metrics_clicks'                                 => true,
			'mc_metrics_impressions'                            => true,
			'mc_metrics_ctr'                                    => true,
			'mc_metrics_conversions'                            => true,
			'price_compared_with_benchmark'                     => true,
```

jjgrainger (commented):
@mukeshpanchal27 @joemcgill overall looks good, including the new tests.

I've left some minor comments to review. Thanks

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
joemcgill: Assuming the empty response is an empty array and not something like `null` or `false`, then yes. I think this just ensures we're sending the correct shape. The `PriceBenchmarks::get_price_benchmarks_data()` method has strict typing that it returns an array, so it's probably fine to remove this.

src/DB/Table/MerchantPriceBenchmarksTable.php
joemcgill: Agree that your suggestion looks better. I've noticed that once array keys get to a certain length, the alignment WPCS rules no longer apply. If you do apply these changes locally, they get auto-formatted back to the way they were previously. Kind of odd?

src/API/Site/Controllers/MerchantCenter/PriceBenchmarksController.php
joemcgill: Addressed in [1583097](https://github.com/woocommerce/google-listings-and-ads/pull/2935/commits/15830976ecdf0e9c418dd483310542d7d9fd3069)

src/DB/Table/MerchantPriceBenchmarksTable.php
joemcgill: Looks like this is a setting you can modify in the ruleset. I've increased the max columns in [58f4605](https://github.com/woocommerce/google-listings-and-ads/pull/2935/commits/58f4605712da5e6958cbc68f902dd1823a315803)

joemcgill (dismissed):
Pushed a few more cleanup changes but this looks good to merge

joemcgill:
@jjgrainger I've made [a couple of changes](https://github.com/woocommerce/google-listings-and-ads/pull/2935/files/bff5e19d663ef289597f61c1d5d5ea3a324b55d6..5c3425e87ce361758e8d4d3a4a074895c82dcf33) since your last review because I realized that the way some of this was implemented was based on an incorrect data shape. I've gone ahead and made the required changes and updated some unit tests, but would appreciate another set of eyes.

jjgrainger (commented):
Thanks @joemcgill I've reviewed and the changes look good

joemcgill (approved):
Thanks @jjgrainger. I'm going to go ahead and merge this to the feature branch so we can test the FE changes integrated with the backend.

ankitguptaindia:
**QA Status-** Approved ✅ 
------------

**Testing Environment -** 

<Details>

* WordPress: 6.8.1
* Theme:  
* WooCommerce - Version 9.9.0-rc.1
* PHP: 8.3.0
* Web Server: Nginx
* OS: macOS Sonoma 14.6.1 

</Details>




**Test Results -** 

Tested the product listing flow for effectiveness, sorting by using the steps as described. Products appear correctly sorted by effectiveness in **descending order** (i.e., highest to lowest).
Also tested the **change in sorting order** — both **descending and ascending** — and verified that sorting works correctly even when pagination is active on the second page.



**Functional Demo /  Screencast -**. 


https://github.com/user-attachments/assets/830447f9-3bc0-4f8c-9d28-eb6569fe3432

PR #2949, opened 2025-06-04
Rename attributes to match BE.

### LINKED ISSUE #2933: Update sorting and filtering in Price Suggestions

asvinb:

See #2933.


joemcgill (approved):
@asvinb this looks good to me. Let's merge it into the main PR branch

Issue #2903, opened 2025-05-12
Support reading/editing the enhanced conversions setting via the REST API


asvinb:
## User Story
As a site administrator, I want to enable or disable Enhanced Conversions for Ads via a REST API endpoint, so that I can control whether enhanced conversions are activated in a programmatic way.

## Acceptance Criteria
- A new REST API endpoint is available at `ads/settings` to support generic ads related settings in the REST API.
- The endpoint supports `GET` requests
  - This is only accessible to users with `manage_woocommerce` caps (standard permission callback)
  - With NO Ads account connected it should return a 403 Not Allowed response
  - With a connected Ads account it should return the following response:

```json
{
  "enhanced_conversions_enabled": true
}
```

- The endpoint supports `PUT|PATCH|POST` requests (i.e. `WP_REST_Server::EDITABLE`)
  - This is only accessible to users with `manage_woocommerce` caps (standard permission callback)
  - It accepts a settings object which will update only the options that are supported (currently just the `enhanced_conversions_enabled` option).
- The REST endpoint returns the correct schema to an `OPTIONS` request
- When the current Ads account is disconnected, the underlying option is deleted.

## Implementation Brief
- Using `src/Options/OptionsInterface.php`,
	- Add a new option `ADS_ENHANCED_CONVERSIONS_ENABLED`
- Create a new REST controller in `src/API/Site/Controllers/Ads/SettingsController.php`,
	- Register the `ads/settings` route that supports both READABLE and EDITABLE methods.
	- The readable endpoint,
		- Check if there's an Ads ID connected via `$this->options->get_ads_id()` method.
		- Return the value for the option defined in `OptionsInterface.php` or an error if not connected.
	- The editable endpoint
		- Update the option via `$this->options->update` method.
- To ensure this options is deleted on disconnection, add the option to the [disconnect method](https://github.com/woocommerce/google-listings-and-ads/blob/c5c67ec46c7ad7218ecc554edae7e8e952375712/src/Ads/AccountService.php#L298) in `src/Ads/AccountService.php`

## Test Coverage
- Ensure the `ADS_ENHANCED_CONVERSIONS_ENABLED` option is registered in `OptionsInterface`.
- Test the new `ads/settings` route is registered correctly with both readable (`GET`) and editable (`POST/PUT`) methods.

joemcgill:
[...]

PR #2950, opened 2025-06-04
Add Enhanced Conversion data to the Global Site tag

### LINKED ISSUE #2937: Add EC data to the Global Site tag
## User Story
As a merchant, when I have enabled sending Enhanced Conversions (EC) data to Google via the settings screen described in #2902, I want the global site tag data to be updated to automatically include this data.

## Acceptance Criteria

- When EC is disabled, no EC data is being included in web pages.
- When EC is enabled and first-party data about the user is available, the site will set hashed first-party data for the Google Tag's (gtag) `user_data` value based on the documented instructions in [this article](https://support.google.com/google-ads/answer/13258081?sjid=6208987066621462670-NC#zippy=%2Cidentify-and-define-your-enhanced-conversions-fields).
- When the connected Ads account has properly set up EC, following [these instructions](https://support.google.com/google-ads/answer/13258081?sjid=6208987066621462670-NC), the implementation can be successfully validated using [these instructions](https://support.google.com/google-ads/answer/13258081?sjid=6208987066621462670-NC#zippy=%2Cvalidate-your-implementation-using-chrome-developer-tools).
- The EC data should be included anywhere we're adding the config for the ads conversion ID.

## Implementation Brief

Full instructions from Google can be found here: https://support.google.com/google-ads/answer/13258081?sjid=6208987066621462670-NC#zippy=%2Cidentify-and-define-your-enhanced-conversions-fields

> [!NOTE]
> A previous iteration was tried in https://github.com/woocommerce/google-listings-and-ads/pull/2258, but note that the setup instructions have now changed, so that implementation should not be used as is.

The `\Google\GlobalSiteTag` class is responsible for handling the global site tag and adding any conversion data to the page. The `GlobalSiteTag::activate_global_site_tag()` method is the main place where this is set up ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/0d946f6acd39373280626de967a2c302015d2b6a/src/Google/GlobalSiteTag.php#L236-L258)). This method handles several different scenarios where the Ads Tag config `gtag('config', 'TAG_ID');` is included in the page depending on how the `gtag` snippet is being managed.

In each scenario, anywhere that we output the Ads Tag config, we should also attempt to add the `user_data` config as documented in the instructions above:

```html
<script>
  gtag('set', 'user_data', {
    "email": {{ yourEmailVariable }},
    "phone_number": {{ yourPhoneVariable }}
    // etc.
  });
  </script>
```

There should be a single method added to the `GlobalSiteTag` class to add the EC data that is reused in each place where we want to include the EC data.

## Test Coverage

- PHPUnit tests should be added to the `GlobalSiteTagTest` test class for this
- [E2E tests for events](https://github.com/woocommerce/google-listings-and-ads/blob/93758c84d930b7c44228ce2e594208985fdb92c8/tests/e2e/specs/gtag-events/gtag-events.test.js) could be updated to confirm that the data is being added to the page for conversion events (nice to have).

jjgrainger:

Closes #2937

Adds Enhanced Conversion data to the Global Site tag when first-party user data is available.

1. Enable Enhanced Conversions in the plugin admin settings (See #2903)
2. Create a customer account, log in and update the profile to contain an email, address and phone number.
3. While logged in as that customer, view the source of any frontend page and search for `user_data`
4. There should be a tag present that sets the user data configuration, example below:

```
gtag("set", "user_data", {
   "sha256_email_address": "be3221a6fac131657111728b4d912a877ec158b123d5db3afef3bd8a59784ece",
   "sha256_phone_number": "0369fadc2246c1f658413b16d60137ca8ba188c943ee5dbbd4df38f6a84d4ba2",
   "address": {
      "sha256_first_name": "78675cc176081372c43abab3ea9fb70c74381eb02dc6e93fb6d44d161da6eeb3",
      "sha256_last_name": "0488ac68cb8f38212dbcd5fbd6c66c07fd1145622043374d6158cc4df3f54c5a",
      "street": "215 Main St",
      "city": "Los Altos",
      "country": "US",
      "postal_code": "94022",
      "region": "CA"
   }
});
```
5. The data must follow the following rules:
- The email address, phone number, first name and last name must all be normalized and hashed with SHA-256 algorithm.
- The phone number must be normalized to E.164 format.
- The phone number can only be set if the email address or address is present.

### Changelog entry

>


joemcgill (commented):
Looks like you're on the right track so far, @jjgrainger! One note would be to confirm that the phone number is normalized to E.164 format (e.g. ‘+11231234567’) and country codes are in ISO 3166-1 alpha-2 format (e.g. UK), per the instruction [on this page](https://support.google.com/google-ads/answer/13258081?sjid=6208987066621462670-NC#zippy=%2Cidentify-and-define-your-enhanced-conversions-fields%2Cvalidate-your-implementation-using-chrome-developer-tools).

src/Google/GlobalSiteTag.php
```diff
@@ -602,4 +606,118 @@ private function register_js_for_fast_refresh_dev() {
...
 	}
+
+	/**
+	 * Set user data config when Enhanced Conversions is enabled.
+	 *
+	 * @return string|null
+	 */
+	public function get_enhanced_conversion_tag() {
+		// TODO: Check enhanced conversion option status when implemented.
+		// $enhanced_conversions =$this->options->get( OptionsInterface::ADS_ENHANCED_CONVERSION_STATUS );
```
joemcgill: ✨

src/Google/GlobalSiteTag.php
```diff
@@ -602,4 +606,118 @@ private function register_js_for_fast_refresh_dev() {
...
+		}
+
+		// Retrieve user data from the current session.
+		$customer = WC()->session->get( 'customer' );
+
+		$ec_data = [];
+
+		// Add email address to enhanced conversion data.
+		if ( ! empty( $customer['email'] ) ) {
+			$ec_data['sha256_email_address'] = esc_js( $this->normalize_and_hash( $customer['email'] ) );
```
joemcgill: I don't think we need `esc_js` here or throughout, since all of this data will be passed to `wp_json_encode`, unless there is something specifically you are using this for.

FWIW, I generally consider `esc_js` to be a soft-deprecated function in WP Core at this point. Note [this comment from Weston](https://developer.wordpress.org/reference/functions/esc_js/#comment-1747) as to why it's no longer considered best practice.

joemcgill (commented):
One nitpick about escaping, but otherwise looks good!

src/Google/GlobalSiteTag.php
jjgrainger: Thanks @joemcgill I agree and have removed the `esc_js` throughout

src/Google/GlobalSiteTag.php
jjgrainger: @joemcgill I have synced this branch with the feature branch and updated to use the option.

src/Google/GlobalSiteTag.php
```diff
@@ -602,4 +606,116 @@ private function register_js_for_fast_refresh_dev() {
...
+		// Validate the number is the correct length.
+		if ( strlen( $phone ) < 11 || strlen( $phone ) > 15 ) {
+			return '';
+		}
+
+		return $phone;
+	}
+
+	/**
+	 * Normalize and hash ehanced conversion data.
```
mukeshpanchal27:
```suggestion
	 * Normalize and hash enhanced conversion data.
```

src/Google/GlobalSiteTag.php
```diff
@@ -602,4 +606,116 @@ private function register_js_for_fast_refresh_dev() {
...
+			if ( ! empty( $customer['city'] ) ) {
+				$ec_data['address']['city'] = $customer['city'];
+			}
+
+			if ( ! empty( $customer['state'] ) ) {
+				$ec_data['address']['region'] = $customer['state'];
+			}
+		}
+
+		// Phone number can only be added when emaill and/or address is present.
```
mukeshpanchal27:
```suggestion
		// Phone number can only be added when email and/or address is present.
```

src/Google/GlobalSiteTag.php
```diff
@@ -602,4 +606,116 @@ private function register_js_for_fast_refresh_dev() {
...
+	 */
+	public function get_enhanced_conversion_tag() {
+		$enhanced_conversions = $this->options->get( OptionsInterface::ADS_ENHANCED_CONVERSIONS_ENABLED );
+
+		if ( ! $enhanced_conversions ) {
+			return;
+		}
+
+		// Retrieve user data from the current session.
+		$customer = WC()->session->get( 'customer' );
```
mukeshpanchal27: Might want to default to an empty array to avoid potential warnings?

src/Google/GlobalSiteTag.php
```diff
@@ -602,4 +606,116 @@ private function register_js_for_fast_refresh_dev() {
...
+		}
+
+		if ( empty( $ec_data ) ) {
+			return;
+		}
+
+		// Return the tag.
+		return sprintf(
+			'gtag("set", "user_data", %s);',
+			wp_json_encode( $ec_data ),
```
mukeshpanchal27:
```suggestion
			wp_json_encode( $ec_data )
```

Remove extra trailing comma

src/Google/GlobalSiteTag.php
joemcgill: @jjgrainger I also noticed that in this class, `WC` can be referencing a WC proxy class in `src/Proxies/WC.php`, so we want to make sure this is actually retrieving the data you expect.

joemcgill (commented):
@jjgrainger this is looking good to me. I left [one comment](https://github.com/woocommerce/google-listings-and-ads/pull/2950/files#r2161821572) that you may want to double check, but otherwise, as soon as @mukeshpanchal27's feedback is resolved this is good to move to QA.

src/Google/GlobalSiteTag.php
jjgrainger: Thanks both, I've updated the to default to an empty array.

I've checked and in this case we're calling the `WC()` function (not the class) so this shouldn't be interacting with the WC proxy class in this instance. I can see other calls to `WC()` elsewhere in this class too.

From my testing, this is retrieving the expected data.

mukeshpanchal27: Just pushed some commit to use the WC proxy class. Please take a look.

mukeshpanchal27 (approved):

ankitguptaindia:
QA Status: Approved ✅ 
-------


 **Test Result:** 
 
- All testing steps successfully verified and working as per AC. 

- When Enhanced Conversions correctly enabled, and data is reflected on the frontend. The `user_data` tag appeared with properly hashed values. Cross-tested manually to confirm expected behaviour across scenarios.

- When enhanced_conversions_enabled is set to false, `user_data` tag doesn't show in frontend. 


Screenshot 2025-06-25 at 3 28 28 PM

PR #2952, opened 2025-06-04
Price Benchmarks: Fix search and sort updates

joemcgill:

This addresses two issues found during testing:

1. When a search returns no results, the API should return no results instead of all benchmarks.
2. Fixes the accepted sort options for product_price and benchmark_price.

This is a follow-up to #2935 and #2934.


asvinb (approved):
Tested and works fine @joemcgill

PR #2944, opened 2025-05-30
Update Google Ads API from v18 to v20

eason9487:
### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/GOOWOO-167/update-google-api-libraries-from-v18-to-v20

This PR updates the Google Ads API to use v20. On the one hand, because v18 will be sunset in August 2025, on the other hand, the Budget Recommendations feature needs v20 in order to add the `merchant_center_account_id` field to get better results when obtaining recommendation data.

Ref: https://developers.google.com/google-ads/api/docs/upgrade

#### 📌 v18 to v19

Updated packages:

```
composer require googleads/google-ads-php:dev-legacy-v27.2.0 --with-all-dependencies

- Upgrading google/longrunning (0.4.6 => 0.4.7)
- Upgrading google/protobuf (v3.25.5 => v3.25.8)
- Upgrading googleads/google-ads-php (dev-legacy-v25.0.0 3f4ea59 => dev-legacy-v27.2.0 d72d7e9)
- Upgrading guzzlehttp/guzzle (7.9.2 => 7.9.3)
- Upgrading guzzlehttp/promises (2.0.4 => 2.2.0)
- Upgrading guzzlehttp/psr7 (2.7.0 => 2.7.1)
- Upgrading symfony/polyfill-ctype (v1.31.0 => v1.32.0)
- Upgrading symfony/polyfill-intl-normalizer (v1.31.0 => v1.32.0)
- Upgrading symfony/polyfill-mbstring (v1.31.0 => v1.32.0)
- Upgrading symfony/polyfill-php80 (v1.31.0 => v1.32.0)
- Upgrading symfony/polyfill-php81 (v1.31.0 => v1.32.0)
```

#### 📌 v19 to v20

Updated packages:

```
composer require googleads/google-ads-php:dev-legacy-v28.0.0 --with-all-dependencies
- Upgrading googleads/google-ads-php (dev-legacy-v27.2.0 d72d7e9 => dev-legacy-v28.0.0 c4b6287)
```

#### 📌 Updates for other dependent composer packages

Updated `require` packages:

```
composer update league/iso3166 --with-all-dependencies
- Upgrading league/iso3166 (4.3.1 => 4.3.2)

composer update league/container --with-all-dependencies
- Upgrading league/container (4.2.4 => 4.2.5)

composer update phpseclib/bcmath_compat --with-all-dependencies
- Upgrading phpseclib/bcmath_compat (2.0.2 => 2.0.3)
```

[...]

PR #2959, opened 2025-06-10
Fix the API error of querying campaign assets by an invalid ID after saving a campaign with assets

eason9487:

While testing #2944, I found an error that occurs after saving campaign assets, and it's a pre-existing issue.

image

This PR:

- Addes a new hook `useQuery` to get the current URL query and ignore query changes after unmounting.
   - The `useQuery` in `@woocommerce/navigation` returns an empty object first, which is not very convenient to use.
- Fixes the error caused by loading campaign assets with an invalid ID after saving them.
   - Calling `invalidateResolvedAssetEntityGroups` after saving assets immediately triggers several wp-data store updates. This causes the `EditPaidAdsCampaign` component to re-render even after unmounting. However, at this point, the ID in the URL query has been cleared by `getHistory().push( getDashboardUrl() );`, leading to API queries for assets with a `NaN` ID.

#### 📹 🐛 Before

https://github.com/user-attachments/assets/f5202082-53cf-45f1-8974-040c5cbd07bf

#### 📹 🛠️ After

https://github.com/user-attachments/assets/c2504f4b-787a-418b-bd22-b5ed782e022c

1. You may need to set your local test site to be publicly accessible, or use the workaround for testing **Campaign Assets** in [Smoke-tests](https://github.com/woocommerce/google-listings-and-ads/wiki/Smoke-tests)
2. Go to edit a campaign and save it with assets
   - Can be tested by saving it directly without changes
3. After being redirected to the Dashboard page, check if the error no longer occurs

### Changelog entry

> Fix - The API error of querying campaign assets by an invalid ID after saving a campaign with assets.

### DIFF
--- a/js/src/hooks/useQuery.js
+++ b/js/src/hooks/useQuery.js
@@ -1,0 +1,32 @@
+/**
+ * External dependencies
+ */
+import { useEffect, useState } from '@wordpress/element';
+import { getQuery, addHistoryListener } from '@woocommerce/navigation';
+
+/**
+ * Hook to get the current URL query and ignore query changes after unmounting.
+ *
+ * This implementation differs from the one in `@woocommerce/navigation` in
+ * that it uses the current query as the initial state rather than an empty
+ * object.
+ *
+ * @return {Record<string, string>} Current query object.
+ */
+export default function useQuery() {
+	const [ query, setQuery ] = useState( getQuery );
+
+	useEffect( () => {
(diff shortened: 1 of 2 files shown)


NeosinneR (approved):
LGTM, @eason9487. 

I got `ngrok` to work with my local environment, and can confirm, that I no longer see the error message when in this branch.

PR #2965, opened 2025-06-13
Support Enhanced Conversions Setting via REST API

### LINKED ISSUE #2903: Support reading/editing the enhanced conversions setting via the REST API

jjgrainger:

Closes #2903

Creates a REST API endpoint for updating ad related settings, focusing on Enhanced Conversions option first.

1. Configure Postman to send API request to WordPress with correct authentication and capabilities (`manage_woocommerce`) to make the following requests to `/wp-json/wc/gla/ads/settings`

**Get settings**
Making a GET request to the endpoint should return the enhanced conversions option only as it's the only one supported. The value should reflect the current option state or `true` if not created.

If there is no Ads account connected, a 403 response is returned.

**Request**
```
GET /wp-json/wc/gla/ads/settings
```

**Response**
```
{
    "enhanced_conversions_enabled": true
}
```

**Update setting**
Making a `PUT`, `PATCH` or `POST` request to the endpoint with the correct body should update the option and return it's new status. This should only work with the enhanced conversions option.

**Request**
```
POST /wp-json/wc/gla/ads/settings

BODY
{
    "enhanced_conversions_enabled": false
}
```

**Response**
```
{
    "enhanced_conversions_enabled": false
}
```

**Get endpoint schema**
When an `OPTION` request is made the correct schema is returned.

**Disconnect Ads account**
When the Ads account is disconnected in the admin, the enhanced conversions enabled option is deleted from the options table.

### Changelog entry

>

### DIFF
--- a/src/API/Site/Controllers/Ads/AdsSettingsController.php
+++ b/src/API/Site/Controllers/Ads/AdsSettingsController.php
@@ -6,7 +6,6 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Ads;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
 use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
-use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
 use WP_REST_Request as Request;
 use WP_REST_Response;
 
@@ -18,15 +17,6 @@ defined( 'ABSPATH' ) || exit;
  * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Ads
  */
 class AdsSettingsController extends BaseOptionsController {
-
-	/**
-	 * AdsSettingsController constructor.
-	 *
-	 * @param RESTServer $server
-	 */
-	public function __construct( RESTServer $server ) {
-		parent::__construct( $server );
-	}
 
 	/**
 	 * Register rest routes with WordPress.
--- a/src/API/Site/Controllers/Ads/AdsSettingsController.php
+++ b/src/API/Site/Controllers/Ads/AdsSettingsController.php
@@ -125,7 +125,7 @@ class AdsSettingsController extends BaseOptionsController {
 	 */
 	protected function get_endpoint_params(): array {
 		return [
-			'enhanced_conversions_enabled' => [
+			OptionsInterface::ADS_ENHANCED_CONVERSIONS_ENABLED => [
 				'type'        => 'boolean',
 				'description' => __(
 					'Whether enhanced conversions are enabled.',
@@ -142,7 +142,7 @@ class AdsSettingsController extends BaseOptionsController {
 	 */
 	protected function get_schema_properties(): array {
 		return [
-			'enhanced_conversions_enabled' => [
+			OptionsInterface::ADS_ENHANCED_CONVERSIONS_ENABLED => [
 				'type'        => 'boolean',
 				'description' => __(
 					'Whether enhanced conversions are enabled.',
--- a/src/API/Site/Controllers/Ads/AdsSettingsController.php
+++ b/src/API/Site/Controllers/Ads/AdsSettingsController.php
@@ -87,7 +87,7 @@ class AdsSettingsController extends BaseOptionsController {
 	}
 
 	/**
-	 * Get a callback for editing the settings endpoint.
+	 * Get a callback for editing the ads settings endpoint.
 	 *
 	 * @return callable
 	 */
@@ -102,7 +102,7 @@ class AdsSettingsController extends BaseOptionsController {
 					continue;
 				}
 
-				$result = $this->options->update( $key, $value );
+				$result = $this->options->update( $key, (bool) $value );
 
 				if ( false === $result ) {
 					return new WP_REST_Response(
@@ -119,7 +119,7 @@ class AdsSettingsController extends BaseOptionsController {
 	}
 
 	/**
-	 * Get requests parameters for the settings endpoint.
+	 * Get requests parameters for the ads settings endpoint.
 	 *
 	 * @return array
 	 */
@@ -136,7 +136,7 @@ class AdsSettingsController extends BaseOptionsController {
 	}
 
 	/**
-	 * Get the schema for settings endpoints.
+	 * Get the schema for ads settings endpoints.
 	 *
 	 * @return array
 	 */
--- a/src/API/Site/Controllers/Ads/AdsSettingsController.php
+++ b/src/API/Site/Controllers/Ads/AdsSettingsController.php
@@ -1,0 +1,166 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Ads;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
+use WP_REST_Request as Request;
+use WP_REST_Response;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class AdsSettingsController
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Ads
+ */
+class AdsSettingsController extends BaseOptionsController {
+
+	/**
+	 * AdsSettingsController constructor.
+	 *
+	 * @param RESTServer   $server
+	 */
+	public function __construct( RESTServer $server ) {
+		parent::__construct( $server );
+	}
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'ads/settings',
+			[
+				[
+					'methods'             => TransportMethods::READABLE,
+					'callback'            => $this->get_settings_endpoint_read_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+				[
+					'methods'             => TransportMethods::EDITABLE,
+					'callback'            => $this->get_settings_endpoint_edit_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+					'args'                => $this->get_endpoint_params()
+				],
+				'schema' => $this->get_api_response_schema_callback(),
+			]
+		);
+	}
+
+	/**
+	 * Get an array of allowed options.
+	 *
+	 * @return array
+	 */
+	protected function get_allowed_options(): array {
+		return [
+			OptionsInterface::ADS_ENHANCED_CONVERSIONS_ENABLED => true,
+		];
+	}
+
+	/**
+	 * Get a callback for the settings endpoint.
+	 *
+	 * @return callable
+	 */
+	protected function get_settings_endpoint_read_callback(): callable {
+		return function () {
+			if ( 0 === $this->options->get_ads_id() ) {
+				return new WP_REST_Response(
+					__( 'Not Allowed.', 'google-listings-and-ads' ),
+					403
+				);
+			}
+
+			$settings = [];
+
+			foreach ( $this->get_allowed_options() as $key => $default ) {
+				$settings[ $key ] = $this->options->get( $key, $default );
+			}
+
+			return $settings;
+		};
+	}
+
+	/**
+	 * Get a callback for editing the settings endpoint.
+	 *
+	 * @return callable
+	 */
+	protected function get_settings_endpoint_edit_callback(): callable {
+		return function ( Request $request ) {
+			$params          = $request->get_params();
+			$allowed_options = $this->get_allowed_options();
+			$settings        = [];
+
+			foreach ( $params as $key => $value ) {
+				if ( ! array_key_exists( $key, $allowed_options ) ) {
+					continue;
+				}
+
+				$result = $this->options->update( $key, $value );
+
+				if ( false === $result ) {
+					return new WP_REST_Response(
+						__( 'Unable to update setting.', 'google-listings-and-ads' ),
+						403
+					);
+				}
+
+				$settings[ $key ] = $value;
+			}
+
+			return $settings;
+		};
+	}
+
+	/**
+	 * Get requests parameters for the settings endpoint.
+	 *
+	 * @return array
+	 */
+	protected function get_endpoint_params(): array {
+		return [
+			'enhanced_conversions_enabled' => [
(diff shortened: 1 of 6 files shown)


src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	/**
+	 * Get requests parameters for the settings endpoint.
+	 *
+	 * @return array
+	 */
+	protected function get_endpoint_params(): array {
+		return [
+			'enhanced_conversions_enabled' => [
```
mukeshpanchal27: Nit: In `get_allowed_options()`, you use the constant `OptionsInterface::ADS_ENHANCED_CONVERSIONS_ENABLED`, but in here and `get_schema_properties()` you use `enhanced_conversions_enabled` as the key instead can we use the constant?

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	/**
+	 * Get the schema for settings endpoints.
+	 *
+	 * @return array
+	 */
+	protected function get_schema_properties(): array {
+		return [
+			'enhanced_conversions_enabled' => [
```
mukeshpanchal27: Check above comment.

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+			$params          = $request->get_params();
+			$allowed_options = $this->get_allowed_options();
+			$settings        = [];
+
+			foreach ( $params as $key => $value ) {
+				if ( ! array_key_exists( $key, $allowed_options ) ) {
+					continue;
+				}
+
+				$result = $this->options->update( $key, $value );
```
mukeshpanchal27: Make sure the `$value` is bool otherwise it update whatever we passed.

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+					return new WP_REST_Response(
+						__( 'Unable to update setting.', 'google-listings-and-ads' ),
+						403
+					);
+				}
+
+				$settings[ $key ] = $value;
+			}
+
+			return $settings;
```
mukeshpanchal27: Not blocker: It return associative arrays of settings, but error responses are wrapped in `WP_REST_Response`. For consistency, consider can we returning a `WP_REST_Response` object?

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+		};
+	}
+
+	/**
+	 * Get a callback for editing the settings endpoint.
+	 *
+	 * @return callable
```
mukeshpanchal27, line 82: Need to update

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+	protected function get_allowed_options(): array {
+		return [
+			OptionsInterface::ADS_ENHANCED_CONVERSIONS_ENABLED => true,
+		];
+	}
+
+	/**
+	 * Get a callback for the settings endpoint.
+	 *
+	 * @return callable
```
mukeshpanchal27, line 58: Need to update

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+				);
+			}
+
+			$settings = [];
+
+			foreach ( $this->get_allowed_options() as $key => $default ) {
+				$settings[ $key ] = $this->options->get( $key, $default );
+			}
+
+			return $settings;
```
mukeshpanchal27: Check my comment for `get_settings_endpoint_edit_callback` return value

mukeshpanchal27 (changes requested):
Thanks @jjgrainger. Left some feedbacks for consideration.

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @mukeshpanchal27 I agree and updated

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks, updated

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @mukeshpanchal27 the `get_endpoint_params` method sets the parameter type to `boolean` so whatever is passed is validated and sanitized before we handle the value.

I've updated this to cast the value to boolean too, but let me know if you think more is needed here.

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @mukeshpanchal27  I agree and have updated

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks, updated

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @mukeshpanchal27 I've updated this to `ads settings endpoint.`, was there more needed here?

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Same as above, is more needed here?

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+	 * @return callable
+	 */
+	protected function get_settings_endpoint_edit_callback(): callable {
+		return function ( Request $request ) {
+			$params          = $request->get_params();
+			$allowed_options = $this->get_allowed_options();
+			$settings        = [];
+
+			foreach ( $params as $key => $value ) {
+				if ( ! array_key_exists( $key, $allowed_options ) ) {
```
joemcgill, line 91: Note that the base controller has a `prepare_query_arguments` helper method that might be useful for situations like this where you need to limit the params to only those that are supported by in the schema.

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+					continue;
+				}
+
+				$result = $this->options->update( $key, (bool) $value );
+
+				if ( false === $result ) {
+					return new WP_REST_Response(
+						__( 'Unable to update setting.', 'google-listings-and-ads' ),
+						403
```
joemcgill: `403` seems like the wrong HTTP response code here. From what I can see this isn't handled consistently across the REST controllers, but the `TourController` ([reference](https://github.com/woocommerce/google-listings-and-ads/blob/449acc800f752abc852c0e42391f6c2232747c6c/src/API/Site/Controllers/TourController.php#L105)) seems to be the best approach to me, returning a 400 (Bad Request) error if the option can't be updated. Open to other suggestions if you have one though.

tests/Unit/API/Site/Controllers/Ads/AdsSettingsControllerTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+		$this->options->expects( $this->once() )->method( 'get' )->willReturn( true );
+
+		$response = $this->do_request( self::ROUTE_SETTINGS, 'GET' );
+
+		$this->assertEquals( $expected, $response->get_data() );
+		$this->assertEquals( 200, $response->get_status() );
+	}
+
+	public function test_edit_settings() {
```
joemcgill, line 52: Let's add a test case for when the update fails as well

joemcgill (commented):
This is looking good. Left a few comments inline.

src/API/Site/Controllers/Ads/AdsSettingsController.php
mukeshpanchal27: @jjgrainger The `@return` is not callable so that needs to updated.

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
...
+	/**
+	 * AdsSettingsController constructor.
+	 *
+	 * @param RESTServer $server
+	 */
+	public function __construct( RESTServer $server ) {
+		parent::__construct( $server );
+	}
+
+	/**
```
mukeshpanchal27:
```suggestion
	/**
```

Do we need the constructor?

src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -0,0 +1,166 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\Ads;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers\BaseOptionsController;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\RESTServer;
```
mukeshpanchal27: If we don't need remove these as well.

```suggestion
```

mukeshpanchal27 (commented):
Thanks for the updates.

The changes look good overall. I've left some minor feedback, once those are addressed along with @joemcgill's comments https://github.com/woocommerce/google-listings-and-ads/pull/2965#pullrequestreview-2937458956, it should be good to go.

src/API/Site/Controllers/Ads/AdsSettingsController.php
mukeshpanchal27: Make as resolved.

tests/Unit/API/Site/Controllers/Ads/AdsSettingsControllerTest.php
jjgrainger: Thanks @joemcgill I've added an additional test for this

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @mukeshpanchal27 you're right, it's not needed. I've removed the constructor

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @joemcgill  I agree, I've changed this to `400`

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @joemcgill looking into this I can only see the `prepare_query_arguments` in the [`BaseReportsController`](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/API/Site/Controllers/BaseReportsController.php#L101) which this doesn't extend?

I can see the same method was recreated in the `PriceBenchmarksController` so I can do the same here and rework allowed options and defaults if this is the preferred approach?

joemcgill: Oh, got it! If we find that we want a common method for this across all controllers, we might want to refactor things so that we extend the `BaseController` with this method rather than adding the same method in several controllers that extend that one. Definitely not a blocker for this endpoint controller though.

joemcgill (commented):
Thanks @jjgrainger! I left [a response](https://github.com/woocommerce/google-listings-and-ads/pull/2965/files#r2155566963) in the thread about handling params. Totally up to you if you want to adjust something here.

Otherwise, the updates look good and am happy for this to be approved and merged once @mukeshpanchal27's latest feedback is addressed.

src/API/Site/Controllers/Ads/AdsSettingsController.php
mukeshpanchal27: Something we can open a separate ticket and work as a part of maintenance project?

mukeshpanchal27 (approved):

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks both, I agree it something we should tackle in a separate ticket as it would bring about changes to other parts of the codebase.

mukeshpanchal27 (approved):

ankitguptaindia:
QA Status: ✅ 
----


**Test Result:** Tested all requests like `POST`, `GET`,  `OPTIONS`, `PATCH `, all responses are coming as expected. 

https://github.com/user-attachments/assets/78a7a1a4-b591-42ef-830d-b6aec2fa135e

PR #2971, opened 2025-06-18
Add AssetGroupEditor component

asvinb:

Closes  https://linear.app/a8c/issue/GOOWOO-187/break-down-asset-types-into-sections-in-assetgroupcard

### Screenshots:
image

1. Navigate to the plugin’s main dashboard.
2. Select and edit a paid campaign.
3. In the second step of the edit flow, the layout should be as per the designs.

### Additional details:
* The `AssetGroupCard` component was renamed to `AssetGroupPanel` to better reflect its updated structure and purpose following recent changes.

### Changelog entry

>


js/src/components/paid-ads/asset-group/asset-group-header/asset-group-header.js
```diff
@@ -12,24 +12,21 @@ import { ASSET_FORM_KEY } from '~/constants';
...
  */
-export default function AssetGroupSection() {
+export default function AssetGroupHeader() {
 	const { adapter } = useAdaptiveFormContext();
 	const showTip = adapter.hasImportedAssets;
 
 	return (
 		<Section
 			className="gla-asset-group-section"
-			verticalGap={ 4 }
```
joemcgill, line 33: Some combination of this being removed, and the CSS rule in `js/src/components/paid-ads/asset-group/asset-group.scss`:

```css
.gla-section__body {
  gap: 0;
}
```

Is causing the tip that shows after selecting the Final URL to not have the expected gap.

**This PR:**
image

**Previously:**
image

I think part of the reason you needed this change was because each AssetGroupSection (e.g., text, images, videos) needs to render an `AssetGroupCard` that contains the `AssetField` components. That way each field doesn't get the flex gap applied.

js/src/components/paid-ads/asset-group/asset-group-editor/asset-group-images-section.js
```diff
@@ -0,0 +1,111 @@
...
+								values
+							) }
+							imageConfig={ spec.imageConfig }
+							onChange={ imageProps.onChange }
+						>
+							{ renderErrors( spec.key ) }
+						</ImagesSelector>
+					</AssetField>
+				);
+			} ) }
```
joemcgill: To stay consistent with the previous structure, all `AssetFields` should remain children of an `AssetGroupCard` (or similar wrapper). Same for the other `asset-group-{type}-section.js` files.

joemcgill (changes requested):
@asvinb I think this is close. I noticed one concern with the refactor that needs to be addressed and then I can give it a full review.

js/src/components/paid-ads/asset-group/asset-group-editor/asset-group-images-section.js
asvinb: @joemcgill So we no longer have `AssetGroupCard` now. Instead we have `AssetGroupEditor` which has `AssetGroupImagesSection` and `AssetGroupTextSection` which in turn contain the `AssetField` component. Can you kindly provide more info about what you meant? Thanks!

js/src/components/paid-ads/asset-group/asset-group-header/asset-group-header.js
asvinb: Excellent catch @joemcgill . Updated!

js/src/components/paid-ads/asset-group/asset-group-editor/asset-group-images-section.js
joemcgill: Summarizing from our conversation in Slack: the issue I was pointing out is that one of the reasons that you needed to update the CSS to remove the gap between components was because previously the AssetField components were rendered inside a parent `AssetGropuCard` wrapper, and the gap applied to the parent card and not to the `AssetField` components.

This has now been addressed in [078a6f9](https://github.com/woocommerce/google-listings-and-ads/pull/2971/commits/078a6f98fdad4c4c15297dcdd7bc4ad34643b8b2).

However, I'm still seeing some CSS for border-radius that is applying to the fields that would probably be better to apply to the wrapper. See this screenshot of the border radius when two AssetFields join up. Instead, the border radius should apply to the wrapper.

asvinb: @joemcgill I've tweaked the CSS to avoid that. Furthermore there are other tweaks I made to the layout to have everything consistent:

1. The border radius for the top section was 7px and for the asset fields, it was 2px. To be consistent, I made all the border radius 7px.
2. The gap between the different component were inconsistent. 16px and 24px in some cases. Everything is now 16px.

joemcgill (approved):
Thanks @asvinb

ankitguptaindia:
@asvinb Thanks for the implementation. This looks really great. However, there are some areas we need to match with the design and some UX improvements. 

For this page, we have two states: 1) the initial state, where all sections are collapsed and 2) when all input fields are in expanded mode.

1. In design, the subheadings have different text in both states. Please have a look or let me know if this is an intentional change? 

**Expanded Mode:** 

https://somup.com/cTi1QkL8yd 

**Collapsed Mode** 

https://github.com/user-attachments/assets/109e7955-7001-4dbc-86e0-b6e66f7abbce



----

2. UX - Can we make all + Add heading and other buttons align left, same as the design.

 

https://github.com/user-attachments/assets/82c48e36-0f66-4082-bb61-70fa471fc605


----

3. Query?  - 'Business name' field is not available in design, but available in our developed page. Is this expected? 


https://github.com/user-attachments/assets/220325d6-4355-4750-acaa-a1679a18c04f

----

4. UX -  Can we open the field when clicking around the "v" drop-down so user can easily expand the fields, currently it only opens when clicking on "v" icon.

_(Please skip if this is for maitnaice consistencty with other places of plugin)_


https://github.com/user-attachments/assets/653e398e-004a-48a0-bac5-1ce6776d2d35

asvinb:
@ankitguptaindia 
1. I pinged Michael in [Figma](https://www.figma.com/design/fqR0EHi63lWahRcVTKCcba?node-id=12984-9828#1328447634) about that.
2. Unfortunately we cannot @ankitguptaindia because the button will not be aligned in hover/focus state. You can find the focus state and see where the button starts and ends. 
	- image
3. We are keeping all existing functionality so AFAIK, this should remain. Can you confirm @joemcgill 
4. Unfortunately we cannot because there are other elements which can be displayed next to the button.
	- image
	- image

asvinb:
@ankitguptaindia One thing I noted is that when a URL is selected, all the sections should auto expand (current behaviour on `develop`). This is fixed in #2999

ankitguptaindia:
**QA Status-** Approved ✅ 
------------


**Test Results -** 

All QA feedback has been addressed in this PR. 
For [concern 1](https://github.com/woocommerce/google-listings-and-ads/pull/2971#issuecomment-3044693206), we have separate task here https://linear.app/a8c/issue/GOOWOO-215/update-copy-for-optimize-your-campaign-screen



<Details>

  <summary> Testing Environment (Click to expand) </summary> 

* WordPress: 6.8.2
* Theme:  
* WooCommerce - Version 10.0.2
* PHP: 8.3.0
* Web Server: Nginx
* OS: macOS Sonoma 14.6.1 

</Details>

PR #2976, opened 2025-06-19
Reintroduce account:connect functionality for Google account linking

budzanowski:
## Summary

Reintroduces the account:connect functionality that was previously reverted in PR #2851, with the added requirement to include `blog_id` in the payload for Google account linking.

### Changes Made:

1. **Added `update_sdi_merchant_account()` method** to `Middleware` class
   - Calls Google's SDI `account:connect` endpoint
   - Includes both `merchant_center_id` and `blog_id` in request payload
   - Proper error handling and action hooks

2. **Added supporting endpoint methods**:
   - `get_sdi_merchant_update_endpoint()` - returns the account:connect endpoint URL
   - `get_sdi_endpoint()` - base SDI endpoint helper method

3. **Integrated into merchant account setup flow**:
   - Added `sdi_update` step to `MerchantAccountState` account creation steps
   - Added `sdi_update` case to `AccountService::setup_account_steps()`
   - Automatically triggers during onboarding after website claim

### API Payload Format:
```json
{
  "merchant_center_id": [MC_ID],
  "blog_id": [BLOG_ID]
}
```

### When account:connect is called:
- Automatically during merchant onboarding when MC account is connected
- After website claiming step, before ads linking

## Test plan
- [ ] Test merchant center onboarding flow calls account:connect endpoint
- [ ] Verify both merchant_center_id and blog_id are included in payload
- [ ] Test error handling when endpoint fails
- [ ] Verify onboarding completes successfully after account:connect call

### DIFF
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -488,6 +488,75 @@ class Middleware implements ContainerAwareInterface, OptionsAwareInterface {
 	}
 
 	/**
+	 * Get the URL for the Google SDI merchant update endpoint.
+	 *
+	 * @return string
+	 */
+	public function get_sdi_merchant_update_endpoint(): string {
+		return $this->get_sdi_endpoint() . 'account:connect';
+	}
+
+	/**
+	 * Get the base endpoint to the Google Shopping Data Integration (SDI).
+	 *
+	 * @return string
+	 */
+	protected function get_sdi_endpoint(): string {
+		return $this->container->get( 'connect_server_root' )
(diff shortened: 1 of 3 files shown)


louwie17 (approved):
Code changes look good, pretty straightforward!
I tested the onboarding flow, this went through without any errors. One question I did have as I was testing is that I had it connect to an existing google account, and I was kinda curious if the below would only apply to a brand new account?

budzanowski:
> One question I did have as I was testing is that I had it connect to an existing google account, and I was kinda curious if the below would only apply to a brand new account?

This should happen each time during the onboarding.

PR #2983, opened 2025-06-24
Add PMaxImproveAssetsBanner component

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-184/show-notification-when-there-is-an-ads-recommendation

### Screenshots:
image

1. Navigate to the plugin main dashboard.
2. Ensure you have some PMax campaigns.
3. Mock the `/wp-json/wc/gla/ads/recommendations?type=IMPROVE_PERFORMANCE_MAX_AD_STRENGTH` endpoint with the following in Devtools. See below in "Additional details" section for mocked response.
4. Update the `campaign_id` in the mock response to match one your highest spending PMax campaign.
5. Refresh the page, the banner should now be visible if the conditions are met.
6. The name of the highest spending campaign should be visible in the banner.
7. Clicking the "Improve Assets" button from the banner should redirect you to the campaign edit assets page for the specified campaign.
8. If there are no PMax campaigns or the `ads/recommendations` endpoint returns an empty array, the banner should not be displayed.
9. If the recommendations endpoint does not return the highest spending campaign, the banner should not be displayed.
10. If the `ads/recommendations` endpoint returns an error, the banner should not be displayed.
11. If the user has dismissed the banner, it should not be displayed again upon page reload until 30 days have passed since the dismissal.

### Additional details:
Mocked response:
```json
[
	{
		"id": 1,
		"type": "IMPROVE_PERFORMANCE_MAX_AD_STRENGTH",
		"resource_name": "customers/{customer_id}/recommendations/{recommendation_id}",
		"campaign_id": 22689268469,
		"campaign_name": "Spring Campaign",
		"campaign_status": "ENABLED",
		"last_synced": "2024-06-01T12:34:56Z"
	}
]
```

### Changelog entry

>


js/src/components/pmax-improve-assets-banner/index.js
```diff
@@ -0,0 +1,133 @@
...
+			className="gla-pmax-improve-assets-banner"
+			status="info"
+			isDismissible={ true }
+			onRemove={ dismissBanner }
+		>
+			<p className="gla-pmax-improve-assets-banner__text">
+				{ sprintf(
+					// translators: %s: The PMAX campaign name with the highest spending.
+					__(
+						'Unlock more sales for your campaign, %s, by focusing on improving your campaign assets.Better assets directly increase your ad strength, allowing for a wider variety of ad combinations to be shown across Google.',
```
joemcgill:
```suggestion
						'Unlock more sales for your campaign, %s, by focusing on improving your campaign assets. Better assets directly increase your ad strength, allowing for a wider variety of ad combinations to be shown across Google.',
```

js/src/components/pmax-improve-assets-banner/index.js
```diff
@@ -0,0 +1,133 @@
...
+	if ( expiry !== undefined ) {
+		if ( Date.now() < expiry ) {
+			// Do not render the banner if not expired
+			return null;
+		}
+
+		// If expired, reset the preference
+		set( PREFERENCES_STORE_NAMESPACE, PREFERENCE_BANNER_KEY, {
+			expiry: undefined, // Reset to undefined to show the banner again
+		} );
```
joemcgill: Should this be in a `useEffect()` hook to avoid unnecessary re-renders?

js/src/components/pmax-improve-assets-banner/index.js
```diff
@@ -0,0 +1,133 @@
...
+	);
+
+	if ( ! pmaxCampaigns.length ) {
+		return null;
+	}
+
+	const highestAmountCampaign = pmaxCampaigns.reduce(
+		( max, campaign ) => ( campaign.amount > max.amount ? campaign : max ),
+		pmaxCampaigns[ 0 ]
+	);
```
joemcgill: This logic could be sped up by wrapping it in a `useMemo()`

joemcgill (commented):
Just a few suggestions, but this is looking good.

One thing to keep in mind for the future is that any API calls that automatically get made by this component will need to either be mocked in E2E tests to avoid API errors or we need to figure out how to disable these notifications (maybe by pre-setting the preferences state) during the setup process for the Playwright tests.

js/src/components/pmax-improve-assets-banner/index.js
asvinb: 🤦 Forgot to push my changes before to address this. Fixed @joemcgill

js/src/components/pmax-improve-assets-banner/index.js
asvinb: Yes, just updated it. When I tested there were no re-renders but just to make sure, I've used `useEffect`

js/src/components/pmax-improve-assets-banner/index.js
asvinb: Sure! Updated @joemcgill

asvinb:
@joemcgill  I've noticed that although the `Notice` component is not rendered due to the `expiry`, requests to various endpoints were still being made. To address this, I introduced a new component, `Banner`, which is only rendered if the `expiry` value is expired or not set, thus preventing unnecessary requests when it’s not.

You can review the changes in this [commit](https://github.com/woocommerce/google-listings-and-ads/pull/2983/commits/de009c749dd8c109ae55e42f123d6af9ddbc2e9d#diff-75f1da64201106d8f819fab2598130cf01670574772e525496551c1501e08312).

Another area for optimization is within the Banner component itself - we're currently making two separate requests:
1. Fetching ad campaigns
2. Fetching recommendations

I believe we can avoid fetching campaigns if there are no recommendations (or vice versa). To streamline this, I’m proposing a new hook: `usePMaxCampaignRecommendation`.

This hook would:
* Use the getAdsRecommendations selector to retrieve recommendations
* Bail early and return null if no recommendations exist
* Otherwise, use getAdsCampaigns to fetch campaigns and return the one with the highest spend

This approach would eliminate one unnecessary request and make the logic easier to manage in a single place. Curious to hear your thoughts!

joemcgill:
Thanks @asvinb I like your suggestion for the `usePMaxCampaignRecommendation()` hook. Do you want to make those updates in this PR?

asvinb:
@joemcgill Let me handle it in the current PR. Thanks!

joemcgill (approved):
Confirmed that `useAdsCampaigns()` used in the new `useRecommendedPMaxCampaign()` hook will not make and API request if the ads setup is not complete.

ankitguptaindia:
QA Status: Approved ✅ 
---

All scenarios related to the PMax campaign banner display were tested, including valid responses, empty arrays.
Banner functionality, visibility conditions, redirection, and dismissal logic worked as expected. ✅

https://github.com/user-attachments/assets/377a9604-cf60-469c-8480-bbaa98f6d4fb

PR #2984, opened 2025-06-25
Prevent user from continuing if minimum shipping time is not set

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-203/step-2-onboarding-allows-progression-without-required-shipping-speed

### Screenshots:
image

1. Navigate to the user onboarding flow.
2. Proceed past the "Set up your accounts" step.
3. On the "Configure product listings" step, verify that the "Estimated shipping times" input fields (min and max) are empty by default.
4. Click the "Continue" button.
5. Confirm that the user cannot proceed because the shipping times are not set.
6. Enter a value in the "max" input field only.
7. Click "Continue" again without entering a value in the "min" input field or interacting with the "min" input field (no focus).
8. Confirm that the user is still blocked from proceeding.
9. Enter values in both the "min" and "max" fields.
10. Verify that the user can now proceed to the next step.

### Changelog entry

> Fix - Prevent the user from continuing if the minimum shipping time is not set.

### DIFF
--- a/js/src/components/free-listings/configure-product-listings/__snapshots__/checkErrors.test.js.snap
+++ b/js/src/components/free-listings/configure-product-listings/__snapshots__/checkErrors.test.js.snap
@@ -26,6 +26,8 @@ exports[`checkErrors Shipping times For flat type When there are any selected co
 
 exports[`checkErrors Shipping times For flat type When there are any shipping times are < 0, should not pass 1`] = `"Please specify estimated shipping times for all the countries, and the time cannot be less than 0."`;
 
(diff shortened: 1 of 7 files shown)


joemcgill (approved):
Looks good.

ankitguptaindia:
QA Status: Approved ✅ 
----



**Test Result:**
Continue is not allowed until the merchant fills in values for the **Estimated Shipping Times (Min & Max)** fields. Merchant must either enter a **positive number** in both fields or select the **'Same day'** option to proceed.


https://github.com/user-attachments/assets/84644c57-f5c9-4762-9fd1-656a85d2d0fc

ankitguptaindia (approved):

PR #2991, opened 2025-06-27
Fetch recommendations to improve PMax campaigns assets

mukeshpanchal27:

Closes https://linear.app/a8c/issue/GOOWOO-178/fetch-recommendations-to-improve-pmax-campaigns-assets.

- Connect to the [test-proxy](https://github.com/woocommerce/google-listings-and-ads/blob/develop/tests/proxy/README.md).
- Add `add_filter( 'woocommerce_gla_force_run_install', '__return_true' );` filter in your mu-plugin to create DB `wp_gla_ads_recommendations` table.
- Go to `Admin Panel > Tools > Scheduled Actions` where you can find the schedule job for recommendations. `gla/jobs/update_ads_recommendations`
- Run `gla/jobs/update_ads_recommendations/start` so it will start the scheduling the `gla/jobs/update_ads_recommendations/process_item` every 1 week
- If the `wp_gla_ads_recommendations` table already exists, clear all existing rows to ensure a clean test.
- Run the `gla/jobs/update_ads_recommendations/process_item` job manually by clicking on `Run`.
- Check the `wp_gla_ads_recommendations` table a new row should be inserted upon successful job execution.

### Changelog entry

>


mukeshpanchal27:
@joemcgill Since the work is still in progress, could you please take a quick look when you have a moment today? Your feedback would be helpful to keep things moving forward. Thanks!

src/Ads/AdsRecommendationsService.php
```diff
@@ -0,0 +1,165 @@
...
+	 * @param array $args Query arguments.
+	 * @return array Array of recommendations.
+	 *
+	 * @throws Exception If the recommendations data can't be retrieved.
+	 */
+	public function get_google_recommendations( $args ): array {
+		try {
+			$response = ( new GoogleAdsRecommendationsQuery( $args ) )
+			->set_client( $this->client, $this->options->get_ads_id() )
+			->get_results();
```
mukeshpanchal27, line 109: Hey @joemcgill,

The query returns an empty result, and I haven’t been able to find a reliable way to debug the output.

I reviewed the example in the documentation https://github.com/googleads/google-ads-php/blob/45840387f17d062046cf39ff5e7043257c865259/examples/Recommendations/DetectAndApplyRecommendations.php#L117-L154, but it doesn’t provide enough visibility into the API response to help me identify what’s going wrong.

Do you have any suggestions for how to effectively debug or inspect the actual API query/response?

joemcgill: I spent time trying to debug these and am finding it equally as difficult. From what I can tell, the middleware that we use to retrieve and process these responses should convert the results into `Google\Ads\GoogleAds\V18\Resources\Recommendation` objects that we can process, but I'm seeing the following exception being thrown:

```
Class "Google\Ads\GoogleAds\V18\Resources\Recommendation" not found
```

To continue debugging, try setting a breakpoint in the `AdsRecommendationsService::get_google_recommendations()` method after the `GoogleAdsRecommendationsQuery` is called. You can also make that query less restrictive by removing the `->where()` clause in the constructor (which we should probably do anyway) to return all recommendations of any type connected to the account.

I'm not sure if the autoloader needs to be reset in order to resolve that exception or not.

src/Ads/AdsRecommendationsService.php
```diff
@@ -0,0 +1,188 @@
...
+					 * We use the static name for the recommendation type instead of `$recommendation->getType()`
+					 * to ensure consistency and avoid potential issues with dynamic values or API changes.
+					 */
+					'recommendation_type'            => 'IMPROVE_PERFORMANCE_MAX_AD_STRENGTH',
+					'recommendation_resource_name'   => method_exists( $recommendation, 'getResourceName' ) ? $recommendation->getResourceName() : '',
+					'recommendation_campaign_id'     => $campaign->getId(),
+					'recommendation_campaign_name'   => $campaign->getName(),
+					'recommendation_campaign_status' => $campaign->getStatus(),
+					'recommendation_customer_id'     => $customer->getId(),
+					'recommendation_last_synced'     => gmdate( 'Y-m-d H:i:s' ),
```
jjgrainger: @mukeshpanchal27 do you think we need to set the last sync datetime once before processing and use that for every recommendation?

Just a thought and  might not be necessary, but this could lead to recommendations in the same batch having slightly different timestamps, even if by a few seconds?

src/Ads/AdsRecommendationsService.php
```diff
@@ -0,0 +1,188 @@
...
+				];
+			}
+
+			if ( empty( $result ) ) {
+				return [];
+			}
+
+			return $result;
```
jjgrainger: @mukeshpanchal27 is the empty check necessary? Could we achieve the same just by returning results?

```suggestion
			return $result;
```

src/DB/Table/AdsRecommendationsTable.php
```diff
@@ -0,0 +1,69 @@
...
+	 * Get the schema for the DB.
+	 *
+	 * This should be a SQL string for creating the DB table.
+	 *
+	 * @return string
+	 */
+	protected function get_install_query(): string {
+		return "
+CREATE TABLE `{$this->get_sql_safe_name()}` (
+	recommendation_id bigint(20) NOT NULL AUTO_INCREMENT,
```
jjgrainger: @mukeshpanchal27 I believe we are meant to create the recommendation ID from the resource name so this shouldn't auto increment.

This is from the implementation brief.

> The `recommendation_id` is created by parsing and extracting it from the resource_name. The `resource_name` follows a format of `customers/{customer_id}/recommendations/{recommendation_id}`.

src/Ads/AdsRecommendationsService.php
```diff
@@ -0,0 +1,188 @@
...
+				$result[] = [
+					/**
+					 * Note: The 'id' field below refers to the Recommendation resource's ID property, not the field name itself.
+					 * Reference: https://github.com/googleads/google-ads-php/blob/main/src/Google/Ads/GoogleAds/V18/Resources/Recommendation.php#L25-L30
+					 *
+					 * We use the static name for the recommendation type instead of `$recommendation->getType()`
+					 * to ensure consistency and avoid potential issues with dynamic values or API changes.
+					 */
+					'recommendation_type'            => 'IMPROVE_PERFORMANCE_MAX_AD_STRENGTH',
+					'recommendation_resource_name'   => method_exists( $recommendation, 'getResourceName' ) ? $recommendation->getResourceName() : '',
```
jjgrainger: @mukeshpanchal27 we also need to include the `recommendation_id` which we create from the resource name.

jjgrainger (changes requested):
Thanks @mukeshpanchal27 changes look good, just a few minor suggestions and one note on the recommendation ID to review.

src/Ads/AdsRecommendationsService.php
mukeshpanchal27: Yes make sense. Updated in https://github.com/woocommerce/google-listings-and-ads/pull/2991/commits/a9fd7d928f60bf211632cad880777365f89ce92b

src/DB/Table/AdsRecommendationsTable.php
mukeshpanchal27: I miss read that and missed totally. Thanks for flagging

mukeshpanchal27:
@jjgrainger Thanks for the review and feedback! This is now ready for re-review. Let me know if anything else needs adjusting.

src/Ads/AdsRecommendationsService.php
```diff
@@ -0,0 +1,189 @@
...
+				}
+
+				$campaign = $row->getCampaign();
+				$customer = $row->getCustomer();
+
+				$recommendation_id            = 0;
+				$recommendation_resource_name = method_exists( $recommendation, 'getResourceName' ) ? $recommendation->getResourceName() : '';
+				if ( $recommendation_resource_name ) {
+					$resource_name     = explode( '/', $recommendation_resource_name );
+					$recommendation_id = (int) end( $resource_name );
```
jjgrainger: @mukeshpanchal27 should we need to make sure we have both a resource name and can create the recommendation ID before accepting a recommendation?

I think we will always get a resource name and can generate the ID, but It might be worth implementing a check that we can generate these and if not skip this recommendation? 

Something along the lines of this?

```suggestion
				$recommendation_resource_name = method_exists( $recommendation, 'getResourceName' ) ? $recommendation->getResourceName() : '';
				
				if ( empty( $recommendation_resource_name ) ) {
					continue;
				}
				
				$resource_name     = explode( '/', $recommendation_resource_name );
				$recommendation_id = (int) end( $resource_name );
```

src/Ads/AdsRecommendationsService.php
```diff
@@ -0,0 +1,189 @@
...
+	 */
+	public function get_google_recommendations( $args ): array {
+		try {
+			$response = ( new GoogleAdsRecommendationsQuery( $args ) )
+			->set_client( $this->client, $this->options->get_ads_id() )
+			->get_results();
+
+			$result      = [];
+			$last_synced = gmdate( 'Y-m-d H:i:s' );
+			foreach ( $response->iterateAllElements() as $row ) {
```
joemcgill, line 113: I tested this with a real Ads account after removing the $this->where() clause from the query constructor, in order to get all recommendations returned from the API and confirm that this is now working as expected. Nice!

I also had to `rm -rf vendor/` and run `composer install` before testing, which (if not already) should be added to the testing instructions.

src/Ads/AdsRecommendationsService.php
```diff
@@ -0,0 +1,189 @@
...
+				// Skip if recommendation is not valid.
+				if ( ! $recommendation instanceof Recommendation ) {
+					continue;
+				}
+
+				$campaign = $row->getCampaign();
+				$customer = $row->getCustomer();
+
+				$recommendation_id            = 0;
+				$recommendation_resource_name = method_exists( $recommendation, 'getResourceName' ) ? $recommendation->getResourceName() : '';
```
joemcgill: The method exists check shouldn't be required since we're first checking that this is an `Recommendation` object above.

joemcgill (commented):
Left a few comments, but this is looking good to me 👍🏻

src/Ads/AdsRecommendationsService.php
mukeshpanchal27: The changes make sense. Update in https://github.com/woocommerce/google-listings-and-ads/pull/2991/commits/f3cf80179e426a207d75789ead3ea8c3aaffebd5

joemcgill (approved):

jjgrainger (approved):

ankitguptaindia:
QA Status: Verified ✅ 
----

Test Result: PR is working as expected as described. A new Test proxy data is inserted upon successful job execution in `wp_gla_ads_recommendations` table. 


https://github.com/user-attachments/assets/f29e0dfc-d28b-4371-aa53-d1ddf0392c25

PR #3002, opened 2025-07-07
Fix Intermittent Failure to Disable EC via Checkbox

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-212/intermittent-failure-to-disable-ec-via-checkbox

Fixes issue where REST API endpoint would return a 400 error response when attempting to set the enhanced conversion status to the same value as what is stored in the database.

1. Go the Settings page and see the Enhanced Conversion settings.
2. Use Postman to make a POST request to the `/wp-json/wc/gla/ads/settings` to set the current value to the opposite of what is seen on the settings page (If checked, set `enhanced_conversions_enabled` to `false`, if unchecked `true`)
3. Enable (or disable) the Enhanced Conversions checkbox in the settings page. The checkbox status should update to match what is already stored in the database and a success notification is displayed. The error notification no longer shows.

### Changelog entry

>

### DIFF
--- a/src/API/Site/Controllers/Ads/AdsSettingsController.php
+++ b/src/API/Site/Controllers/Ads/AdsSettingsController.php
@@ -94,7 +94,11 @@ class AdsSettingsController extends BaseOptionsController {
 
 				$stored = $this->options->get( $key );
 
-				if ( ! is_null( $stored ) && (bool) $stored === $value ) {
+				if ( is_null( $stored ) ) {
+					$this->options->add( $key, $value );
+					$settings[ $key ] = $value;
+					continue;
+				} elseif ( (bool) $stored === $value ) {
 					$settings[ $key ] = $value;
 					continue;
 				}
--- a/src/API/Site/Controllers/Ads/AdsSettingsController.php
+++ b/src/API/Site/Controllers/Ads/AdsSettingsController.php
@@ -92,7 +92,14 @@ class AdsSettingsController extends BaseOptionsController {
 					continue;
 				}
 
-				$result = $this->options->update( $key, (bool) $value );
+				$stored = (bool) $this->options->get( $key );
+
+				if ( ! is_null( $stored ) && $stored === $value ) {
+					$settings[ $key ] = $value;
+					continue;
+				}
+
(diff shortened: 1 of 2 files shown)


src/API/Site/Controllers/Ads/AdsSettingsController.php
```diff
@@ -92,7 +92,14 @@ protected function get_settings_endpoint_edit_callback(): callable {
 					continue;
 				}
 
-				$result = $this->options->update( $key, (bool) $value );
+				$stored = (bool) $this->options->get( $key );
```
mukeshpanchal27: When the `enhanced_conversions_enabled` option is not present in DB then this will not return `NULL` instead it return `false`? in that case it will not add/update the option value.

mukeshpanchal27 (changes requested):
Thanks @jjgrainger for the PR! Left one feedback.

src/API/Site/Controllers/Ads/AdsSettingsController.php
jjgrainger: Thanks @mukeshpanchal27 good spot. I've updated this can you review again?

mukeshpanchal27: @jjgrainger @joemcgill  I’ve pushed two commits to cover the scenario where the option is not present in the database:

* https://github.com/woocommerce/google-listings-and-ads/pull/3002/commits/90d6c553f713953d7f04062cb396aac1f61e56e9: Added unit tests to clearly demonstrate that the missing option scenario was not previously handled.
* https://github.com/woocommerce/google-listings-and-ads/pull/3002/commits/f1c849aa894b0fa6bcefe4cd18f77d18f59acaca: Applied the code fix to address the issue identified in the tests.

Let me know if you’d like to see any adjustments otherwise, this should now handle the edge case correctly.

mukeshpanchal27 (approved):

ankitguptaindia:
QA Status: Approved ✅ 
----



https://github.com/user-attachments/assets/b33f0ca7-60d5-4715-ac1a-eb3bdd76bf49

PR #2998, opened 2025-07-02
Bump required WordPress to 6.6 and WooCommerce to 9.7

ankitrox:
### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/GOOWOO-189/bump-required-wordpress-to-66-and-woocommerce-to-97

_Replace this with a good description of your changes & reasoning._

### Screenshots:

Screenshot 2025-07-02 at 5 25 08 PM

### Detailed test instructions:

1. Search the codebase to see if there are any WordPress and WooCommerce versions noted that have not been updated together.

2. Setup the site with older woocommerce version, which should show the notice as shown in above screenshot.

### Changelog entry

> Dev - Bump required WordPress to 6.6 and WooCommerce to 9.7.

PR #3014, opened 2025-07-21
Use `wp_is_serving_rest_request()` to detect if we are handling a REST API request

### LINKED ISSUE #2303: Use wp_is_serving_rest_request to detect if we are handling a REST API request
## User story
In WordPress 6.5 a new helper function `wp_is_serving_rest_request` is introduced. For better reliability we should make use of this function.
See: https://core.trac.wordpress.org/changeset/57312

We have [one location in the code](https://github.com/woocommerce/google-listings-and-ads/blob/2.6.1/src/Admin/Redirect.php#L71) where we use the old method `defined( 'REST_REQUEST' ) && REST_REQUEST`
Since we support WordPress L-2 we'll have to make it conditional and fallback to the old method if `wp_is_serving_rest_request` is not available. But we might as well start using the newer method as it might be more reliable since it's filterable. Edit: Final version is not filterable, so there are less benefits in using the helper function.

mukeshpanchal27:

Closes #2303.

https://linear.app/a8c/issue/GOOWOO-93/use-wp-is-serving-rest-request-to-detect-if-we-are-handling-a-rest-api

1. Install and activate WC in fresh setup
2. Install Google for WooCommerce and activate
3. Make sure it redirect to onboarding page.

### Changelog entry

Fix – Use `wp_is_serving_rest_request()` to detect if we are handling a REST API request.

### DIFF
--- a/src/Proxies/WP.php
+++ b/src/Proxies/WP.php
@@ -340,4 +340,13 @@ class WP {
 	public function wp_print_inline_script_tag( string $data, array $attributes = [] ) {
 		return wp_print_inline_script_tag( $data, $attributes );
 	}
+
+	/**
+	 * Determines whether WordPress is currently serving a REST API request.
+	 *
+	 * @return bool True if it's a WordPress REST API request, false otherwise.
+	 */
+	public function wp_is_serving_rest_request(): bool {
+		return wp_is_serving_rest_request();
+	}
 }
--- a/src/Admin/Redirect.php
+++ b/src/Admin/Redirect.php
@@ -68,7 +68,7 @@ class Redirect implements Activateable, Service, Registerable, OptionsAwareInter
 	 */
 	public function activate(): void {
 		// Do not take any action if activated in a REST request (via wc-admin).
-		if ( defined( 'REST_REQUEST' ) && REST_REQUEST ) {
+		if ( $this->wp->wp_is_serving_rest_request() ) {
 			return;
 		}
 


src/Proxies/WP.php
```diff
@@ -340,4 +340,13 @@ public function wp_add_inline_script( string $handle, string $data, string $posi
...
 		return wp_print_inline_script_tag( $data, $attributes );
 	}
+
+	/**
+	 * Determines whether WordPress is currently serving a REST API request.
+	 *
+	 * @return bool True if it's a WordPress REST API request, false otherwise.
+	 */
+	public function wp_is_serving_rest_request(): bool {
+		return wp_is_serving_rest_request();
```
jjgrainger, line 350: @mukeshpanchal27 based on the ticket description should we be checking the function is available and if not fallback to the old method? e.g

```suggestion
		if ( function_exists( 'wp_is_serving_rest_request' ) ) {
				return wp_is_serving_rest_request();
		}
		
		return ( defined( 'REST_REQUEST' ) && REST_REQUEST );
```

jjgrainger (changes requested):

mukeshpanchal27:
Merged blocked by #2998

src/Proxies/WP.php
mukeshpanchal27: @jjgrainger, in #2998 we bumped the WordPress version requirement to 6.6, and since that function was introduced in WP 6.5, it will always be available. So there's no need to wrap it with `function_exists` anymore

jjgrainger (approved):

ankitguptaindia:
QA Status: Approved ✅ 
----


**With WordPress 6.8.2 & WooCommerce 10.0.4:**


https://github.com/user-attachments/assets/4a6d36f0-7b1c-4345-a8f8-eb2648e1bdab


**With WordPress 6.6.3 & WooCommerce 9.7.1** 


https://github.com/user-attachments/assets/dbe3dae7-4217-490a-98bd-7b87132bccdd

ankitguptaindia (approved):

PR #3018, opened 2025-07-21
Ensure EC settings defaults to false

joemcgill:

Closes GOOWOO-212. Follow-up to #3002.

This ensures that the settings API endpoint returns false until the user explicitly enables this feature.

### Screenshots:

image

1. Delete the `gla_enhanced_conversions_enabled` option from the local database.
2. Load the setting screen and see that the checkbox is unchecked.
3. Ensure that when unchecked, the EC tag is not being added to the front-end

### Changelog entry

>

### DIFF
--- a/tests/Unit/API/Site/Controllers/Ads/AdsSettingsControllerTest.php
+++ b/tests/Unit/API/Site/Controllers/Ads/AdsSettingsControllerTest.php
@@ -49,6 +49,20 @@ class AdsSettingsControllerTest extends RESTControllerUnitTest {
 		$this->assertEquals( 200, $response->get_status() );
 	}
 
+	public function test_get_settings_with_null_option_value() {
+		$expected = [
+			'enhanced_conversions_enabled' => false,
+		];
+
+		$this->options->expects( $this->once() )->method( 'get_ads_id' )->willReturn( 1 );
+		$this->options->expects( $this->once() )->method( 'get' )->willReturn( null );
+
+		$response = $this->do_request( self::ROUTE_SETTINGS, 'GET' );
+
+		$this->assertEquals( $expected, $response->get_data() );
+		$this->assertEquals( 200, $response->get_status() );
+	}
+
 	public function test_edit_settings() {
 		$expected = [
 			'enhanced_conversions_enabled' => false,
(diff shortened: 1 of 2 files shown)


tests/Unit/API/Site/Controllers/Ads/AdsSettingsControllerTest.php
```diff
@@ -49,6 +49,20 @@ public function test_get_settings() {
...
+		$this->options->expects( $this->once() )->method( 'get_ads_id' )->willReturn( 1 );
+		$this->options->expects( $this->once() )->method( 'get' )->willReturn( null );
+
+		$response = $this->do_request( self::ROUTE_SETTINGS, 'GET' );
+
+		$this->assertEquals( $expected, $response->get_data() );
+		$this->assertEquals( 200, $response->get_status() );
+	}
+
 	public function test_edit_settings() {
```
mukeshpanchal27, line 66:
```suggestion
	public function test_edit_settings() {
```

Duplicate of `test_edit_settings_with_null_option_value` test.

mukeshpanchal27 (approved):
Looks good to me. Just left a minor comment regarding a duplicate unit test. 👍

tests/Unit/API/Site/Controllers/Ads/AdsSettingsControllerTest.php
jjgrainger: @mukeshpanchal27 I don't think this is a duplicate as the new test added is for the GET request, where as the `test_edit_settings_with_null_option_value` is for the POST request.

The new test is to ensure that we return the correct default if the option is not saved in the database. Let me know if you think differently here.

Thanks

jjgrainger (approved):

tests/Unit/API/Site/Controllers/Ads/AdsSettingsControllerTest.php
mukeshpanchal27: In that case can we use dataProvide and merge both tests?

mukeshpanchal27: @joemcgill Feedback in Slack:

> Data providers weren’t in use for the POST tests so I opted to just create a new one here. If there were several different cases to test, I’d be inclined to adjust, but this is probably fine for now

ankitguptaindia:
**QA Status:** Approved ✅ 
----

Works as explained. `user_data` only appears when this EC is enabled, and when  `gla_enhanced_conversions_enabled option` is deleted  `user_data` doesn't show up in frontend.

PR #3024, opened 2025-07-23
Add API_PULL_SYNC_MODE to WooCommerce System Status Report

budzanowski:
## Summary
- Adds a new SystemStatusService that displays API Pull and MC Push sync mode configuration in the WooCommerce System Status Report
- Creates a new "Google for WooCommerce" section showing sync settings for all data types (products, settings, coupons, shipping)
- Includes proper error handling and visual status indicators

## Test plan
- [ ] Navigate to WooCommerce > Status > System Status
- [ ] Verify the "Google for WooCommerce" section appears
- [ ] Confirm sync mode settings are displayed correctly for each data type
- [ ] Check that API Pull and MC Push statuses show appropriate enabled/disabled indicators
- [ ] Test error handling when sync mode data is unavailable

### Changelog entry

> Add - Display the statuses of synchronization modes with Google Merchant Center in WooCommerce System Status Report.

### DIFF
--- a/src/Admin/SystemStatusService.php
+++ b/src/Admin/SystemStatusService.php
@@ -125,22 +125,27 @@ class SystemStatusService implements Service, Registerable {
 					$pull_enabled = isset( $modes['pull'] ) && $modes['pull'];
 					$push_enabled = isset( $modes['push'] ) && $modes['push'];
 					
-					$pull_status = $pull_enabled ? '✔ Enabled' : '❌ Disabled';
-					$push_status = $push_enabled ? '✔ Enabled' : '❌ Disabled';
-					$pull_class  = $pull_enabled ? 'yes' : 'error';
-					$push_class  = $push_enabled ? 'yes' : 'error';
+					$pull_status_text = $pull_enabled ? 'Enabled' : 'Disabled';
+					$push_status_text = $push_enabled ? 'Enabled' : 'Disabled';
 					
-					// Format for both display and text export
-					$status_text = sprintf(
-						'API Pull: %s, MC Push: %s',
-						$pull_enabled ? 'Enabled' : 'Disabled',
-						$push_enabled ? 'Enabled' : 'Disabled'
-					);
+					// Create a concise status that matches WooCommerce export patterns
+					$status_summary = sprintf( 'API Pull: %s / MC Push: %s', $pull_status_text, $push_status_text );
+					
+					if ( $pull_enabled && $push_enabled ) {
+						$class = 'yes';
+						$icon = 'dashicons-yes';
+					} elseif ( ! $pull_enabled && ! $push_enabled ) {
+						$class = 'error';
+						$icon = 'dashicons-warning';
+					} else {
+						$class = 'error'; // Mixed state
+						$icon = 'dashicons-warning';
+					}
 					?>
-					<span data-export-label="<?php echo esc_attr( $status_text ); ?>">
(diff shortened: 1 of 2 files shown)


src/Admin/SystemStatusService.php
```diff
@@ -0,0 +1,156 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
+use Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class SystemStatusService
+ * This class adds Google for WooCommerce information to the WooCommerce System Status Report
+ *
+ * @since 3.2.0
```
eason9487: Using a version placeholder would be a bit nicer.

```suggestion
 * @since x.x.x
```

src/Admin/SystemStatusService.php
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	/**
+	 * Add Google for WooCommerce section to System Status Report
+	 */
+	public function add_system_status_section(): void {
+		// Check if the notifications service is available to prevent fatal errors
+		if ( ! $this->notifications_service ) {
+			return;
+		}
```
eason9487: Wonder when the notifications service will become unavailable?

eason9487 (commented):
Thanks for the work!

Is this the expected result that all statuses are enabled even if after disconnecting all accounts and deleting all plugin related WP options and transients? For example, could it mislead users or HEs into thinking that "All the accounts have been disconnected from the plugin, but it is still synchronizing data"?

image

src/Admin/SystemStatusService.php
budzanowski: I have added that when I was experiencing some issues regarding how the plugin worked. It was problems ( my misunderstanding ) with the DI. I forgot to remove it.

budzanowski:
> Is this the expected result that all statuses are enabled even if after disconnecting all accounts and deleting all plugin related WP options and transients? For example, could it mislead users or HEs into thinking that "All the accounts have been disconnected from the plugin, but it is still synchronizing data"?

That is a fair question. You are right that this can be misleading. We have 2 options:
1. Change the status to disabled when the plugin is not connected
2. Add separate connection status entries that would make things clear

Both options present relatively the same amount of work. @MatthiasReinholz what do you think?

MatthiasReinholz:
> > Is this the expected result that all statuses are enabled even if after disconnecting all accounts and deleting all plugin related WP options and transients? For example, could it mislead users or HEs into thinking that "All the accounts have been disconnected from the plugin, but it is still synchronizing data"?
> 
> That is a fair question. You are right that this can be misleading. We have 2 options:
> 
> 1. Change the status to disabled when the plugin is not connected
> 2. Add separate connection status entries that would make things clear
> 
> Both options present relatively the same amount of work. @MatthiasReinholz what do you think?

@eason9487 , Option 1 sounds good to me

src/Admin/SystemStatusService.php
eason9487: Removed by https://github.com/woocommerce/google-listings-and-ads/pull/3024/commits/d88fe70738eeadc9d823d1aa49ea4c22224727b3.

eason9487:
> Is this the expected result that all statuses are enabled even if after disconnecting all accounts and deleting all plugin related WP options and transients? [...]

As per https://github.com/woocommerce/google-listings-and-ads/pull/3024#issuecomment-3223649781, adjusted the statuses in https://github.com/woocommerce/google-listings-and-ads/pull/3024/commits/43df9e33fa2ccc32fd5186a882ef5789cd1c00d0.

PR #3028, opened 2025-07-28
Fix: property name for user preference.

ankitrox:

Closes https://linear.app/a8c/issue/GOOWOO-217/display-badge-count-when-theres-a-pmax-recommendation

1. 
2. 
3. 

### Changelog entry

> Fix - Property name for user preference.


joemcgill (approved):
@ankitrox it looks like we most likely will need to align this property value in both the front end and the back-end so this is ok the merge. However, this still will not totally solve GOOWOO-217 given than the badge count should not be dependent on the notice component being rendered, so lets pick up conversations about next steps in that issue.

mukeshpanchal27 (approved):

PR #3038, opened 2025-08-04
Revert changes to fix an issue where asset suggestions are not being automatically filled after selecting the final campaign URL

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-223/asset-suggestions-not-populating-in-campaign-edit-form

1. Navigate to the dashboard.
2. Create/edit a campaign.
3. After selecting a URL in the "Optimize your campaign" step, the sections should auto expand and auto filled with the suggestions for the different text and image fields.

### Changelog entry

> Fix - Asset suggestions not being automatically filled after selecting the final campaign URL.

### DIFF
--- a/js/src/components/paid-ads/asset-group/asset-group-editor/asset-field.js
+++ b/js/src/components/paid-ads/asset-group/asset-group-editor/asset-field.js
@@ -9,7 +9,6 @@ import {
 	useRef,
 	useImperativeHandle,
 	forwardRef,
-	useEffect,
 } from '@wordpress/element';
 import { chevronUp, chevronDown } from '@wordpress/icons';
 import { Pill } from '@woocommerce/components';
@@ -56,7 +55,7 @@ function AssetField(
 	ref
 ) {
 	const containerRef = useRef();
-	const [ expanded, setExpanded ] = useState( false );
+	const [ expanded, setExpanded ] = useState( initialExpanded );
 
 	const isReducedMotion = useReducedMotion();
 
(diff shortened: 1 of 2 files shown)


joemcgill (approved):
This indeed fixes the issue. Nice find, @asvinb!

ankitguptaindia:
@asvinb Thanks for the PR! 

While testing this, I noticed it behaves differently compared to the `develop` branch. Could you please check if this is expected behaviour or related to the changes introduced in this PR?


**Explanation:**

https://somup.com/cTjh2XNJX4

asvinb:
@ankitguptaindia I had a look on my end, but I don't have full context on how the suggestions are expected to work. I ran some local tests, and regardless of whether I'm on the feature or develop branch, the results are always the same. Could you create a new ticket so we can explore this further?
That said, I can confirm that the changes in this PR aren't related to the issue you observed. We haven’t made any updates to the relevant API calls.

joemcgill:
Just to get another set of eyes on this, I tested both the develop branch and this PR branch and confirmed that on the same environment, I'm seeing the same suggestions auto-populating consistently on both branches. For future reference, to test this you can also confirm that the responses being returned from the suggestions API are being populated to the correct sections in the form, ex:

image

ankitguptaindia:
QA Status: Verified ✅  
---

Able to see the expected result when using a different Ad account.  Assets are auto-filled based on URL selected. 

https://github.com/user-attachments/assets/69b8171d-7716-4bf7-bedc-69fd286ccc3e

PR #3042, opened 2025-08-05
[GOOWOO-217] Update PMax Notification and Banner

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-217/display-badge-count-when-theres-a-pmax-recommendation#comment-ab2ed740

1. By default, when there a no recommendations recorded in the database both the notification badge and banner do not display

#### Setup Recommendations
1.  Insert a new recommendation into the `wp_gla_ads_recommendations` the test MU plugin created for the bug bash can be used here.
2. For a the notification badge and banner to appear a recommendation must meet the following criteria:
    - The recommendation has the type `IMPROVE_PERFORMANCE_MAX_AD_STRENGTH`
    - The recommendation `last_sync` time is within the past 7 days
    - The associated campaign has the status `enabled` and is the highest spend campaign
3. Make sure the following user preference settings are deleted before testing
    - `WP_PREFERENCES_USER_{ID}` in local storage
    - `wp_persisted_preferences` in the `wp_usermeta` table for your user ID

1. When a relevant recommendation exists in the database, and the notification banner has not been interacted with, visit the WordPress and admin and see the notification badge appears on the Marketing Menu with a count of 1.
2. Go to the Marketing > Google for WooCommerce admin page and see the notification badge is attached to the Google for WooCommerce sub menu item.
3. You should also see the notification banner appear with the correct campaign name.
4.  Click 'Improve Assets' will redirect the user to edit the correct campaign. Clicking 'dismiss' or 'x' will remove the notification banner.
5. After clicking any of the buttons the users preferences will record the `actionType` as `dismiss` and the `actionTime` with the current timestamp inside the `pmax-improve-assets-banner` array.
6. After interacting with the banner, both the banner and notification badge will not appear.
7. To test the expiration time of 30 days, update the user preferences `actionTime` to a timestamp that is 30 days + in the past. Once updated, refresh the admin and the notification banner and badge should work as before.
8. When the Google Ads account is disconnected, the notification banner and badge should not appear, even if there are relevant recommendations in the database.

### Changelog entry

> Update - PMax Notification and Banner.


src/API/Google/AdsCampaign.php
```diff
@@ -376,6 +376,29 @@ public function delete_campaign( int $campaign_id ): int {
 		}
 	}
 
+	/**
+	 * Retrieve the campaign with the highest spend amount.
+	 *
+	 * @return array
+	 */
+	public function get_highest_spend_campaign(): array {
```
asvinb, line 384: @jjgrainger Since we are returning the highest spend campaign, can we return an object instead of an array please?

jjgrainger: Spoke with @asvinb and we can mark this as resolved

src/Ads/AdsRecommendationsService.php
```diff
@@ -68,6 +69,17 @@ public function get_recommendations( string $type = '', int $id = 0 ): array {
 
 		if ( $id ) {
 			$query->where( 'recommendation_id', $id );
+		} else {
+			// Only return recommendations from the past 7 days.
+			$query->where( 'recommendation_last_synced', gmdate( 'Y-m-d H:i:s', time() - ( 7 * DAY_IN_SECONDS ) ), '>' );
```
joemcgill: I think it's ok to include this as an extra check, but in reality, the recurring job that refreshes the recommendations every 7 days should remove all old recommendations and replace them with new ones, so this should still return all of the recommendations that were fetched from the Ads API

src/API/Google/AdsCampaign.php
```diff
@@ -376,6 +376,40 @@ public function delete_campaign( int $campaign_id ): int {
...
+			// Set the first campaign found as the highest campaign.
+			if ( empty( $highest_spend_campaign ) ) {
+				$highest_spend_campaign = $campaign;
+				continue;
+			}
+
+			if ( $campaign['amount'] > $highest_spend_campaign['amount'] ) {
+				$highest_spend_campaign = $campaign;
+			}
+		}
```
joemcgill: This does the job. PHP's `array_reduce()` function would have also been a good fit for this.

src/Menu/NotificationManager.php
```diff
@@ -191,10 +195,27 @@ public function display_aggregated_notification_pill(): void {
...
 
-		if ( is_array( $preferences ) && isset( $preferences['woocommerce/google-listings-and-ads']['pmax-improve-assets-banner']['hasRecommendations'] ) && $preferences['woocommerce/google-listings-and-ads']['pmax-improve-assets-banner']['hasRecommendations'] ) {
+		// If the user has not interacted with a recommendation yet.
+		if ( ! is_array( $preferences ) || ! isset( $preferences['woocommerce/google-listings-and-ads']['pmax-improve-assets-banner']['actionType'] ) || ! isset( $preferences['woocommerce/google-listings-and-ads']['pmax-improve-assets-banner']['actionTime'] ) ) {
+			return ++$count;
+		}
+
+		$action_time = $preferences['woocommerce/google-listings-and-ads']['pmax-improve-assets-banner']['actionTime'];
+
+		if ( time() > $action_time + ( 30 * DAY_IN_SECONDS ) ) {
```
joemcgill, line 218: There's a lot of specific Ads recommendation business logic here, which is fine for now since it's the only think being included in this notification count. However, when we incorporate additional notifications into the plugin, it would be better to separate the business logic of specific notifications from the class that manages the notifications.

joemcgill (approved):
Thanks all. These updates seem to be working as expected now. I was able to confirm that when a notification is present, the menu badge shows as well as the notification banner in the plugin page. Once the banner was dismissed, the notification badge no longer shows on the next page refresh.

A few small improvements that could be made in follow-ups would be to show the notification badge on the Google for WooCommerce nav link in the fly-out menu and to remove the notification badge from the menu as soon as the notification banner is dismissed, rather than waiting for a page refresh, but I don't consider either of those blockers.

I left some more non-blocking feedback inline.

src/Ads/AdsRecommendationsService.php
jjgrainger: Thanks @joemcgill I missed we removed the old recommendations. I've removed this query clause.

src/API/Google/AdsCampaign.php
jjgrainger: Thanks @joemcgill I've updated to use `array_reduce`

src/Menu/NotificationManager.php
jjgrainger: Thanks @joemcgill I agree, I think this is okay for now and we can reconsider and refactor when we start to use the notification badge for other notifications.

src/API/Google/AdsCampaign.php
```diff
@@ -382,32 +382,27 @@ public function delete_campaign( int $campaign_id ): int {
...
-			if ( empty( $highest_spend_campaign ) ) {
-				$highest_spend_campaign = $campaign;
-				continue;
-			}
+		return array_reduce(
+			$campaigns,
+			function ( $highest, $campaign ) {
+				if ( empty( $highest ) ) {
+					return $campaign;
+				}
```
joemcgill: One thing I just noticed while reviewing this change is that this could result in returning a non-enabled campaign. I think a slight refactor would avoid this problem:

```php
function ( $highest, $campaign ) {
	if ( CampaignStatus::ENABLED === $campaign['status'] && (
		empty( $highest ) || $campaign['amount'] > $highest['amount'] )
		) {
		return $campaign;
	}

	return $highest;
}
```

joemcgill (changes requested):

src/API/Google/AdsCampaign.php
jjgrainger: Thanks @joemcgill I've updated this with your suggested change

joemcgill (approved):

ankitguptaindia:
@jjgrainger Thanks for implementing this feature. It works as expected, but I noticed a use case **_where the notification bubble in the G4W menu doesn’t disappear_** after the merchant updates the campaign by clicking the "Improve Assets" button or clicking on the Dismiss icon.

The bubble on the G4W menu only disappears after a page refresh, although the banner disappeared as expected. 




https://github.com/user-attachments/assets/a2cdb284-ade9-47de-aafd-cb9cb5997be1

ankitguptaindia:
QA Status: Approved ✅ 
-----

**Quick Demo:** 

https://github.com/user-attachments/assets/946510d8-9a20-40ff-a902-ac442cb902bd

-----

Note: Use case added in comment [here](https://github.com/woocommerce/google-listings-and-ads/pull/3042#issuecomment-3167165499) is not a launch blocker and will be handled in a separate PR later. 
Related discussion: [here](https://fueled.slack.com/archives/C060DC0NXNG/p1754645075432329?thread_ts=1754569095.740419&cid=C060DC0NXNG)

PR #2533, opened 2024-08-16
Streamline campaign setup: Show campaign setup fields immediately during onboarding

asvinb:
### Changes proposed in this Pull Request:

Closes #2500  .

* Show all the setup fields by default.
* Removed all associated tests and helper files.

### Screenshots:

image

### Detailed test instructions:

1. Go through the onboarding flow again.
2. On the fourth step, i.e `Create a campaign`, the fields to create a campaign should be shown by default.
3. The "Skip this step for now" and "Create campaign" buttons should not be present.

### Changelog entry

> Tweak - Show campaign setup fields immediately during onboarding.

PR #3044, opened 2025-08-05
Fix E2E test compatibility issues w Woo 10.1

joemcgill:

See GOOWOO-225.

Fixes some E2E test failures when testing compatibility with Woo 10.1.0-rcX

1. [Review E2E tests for this branch](https://github.com/woocommerce/google-listings-and-ads/actions/runs/16758648108)

### Changelog entry

Dev – Update E2E tests ahead of Woo 10.1 release.

### DIFF
--- a/tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
+++ b/tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
@@ -409,7 +409,6 @@ test.describe( 'Complete your campaign', () => {
 
 			test.describe( 'Budget recommendations', () => {
 				test.beforeEach( async () => {
-					await page.evaluate( () => window.sessionStorage.clear() );
 					await setupAdsAccountPage.mockAdsAccountIncomplete();
 					await setupBudgetPage.fulfillBillingStatusRequest( {
 						status: 'approved',
@@ -417,6 +416,7 @@ test.describe( 'Complete your campaign', () => {
 					await completeCampaign.mockCompleteAdsSetup();
 					await completeCampaign.fulfillBudgetRecommendations();
 					await completeCampaign.goto();
+					await page.evaluate( () => window.sessionStorage.clear() );
 				} );
 
 				test( 'Create a campaign with a selected option from the budget recommendations', async () => {
@@ -487,10 +487,10 @@ test.describe( 'Complete your campaign', () => {
 			test.describe( 'With WooCommerce tracking disabled', () => {
 				test.beforeAll( async () => {
 					// Reset the showing status for the "Set up paid ads" section.
-					await page.evaluate( () => window.sessionStorage.clear() );
 					await setupAdsAccountPage.mockAdsAccountIncomplete();
 					await completeCampaign.goto();
 					await completeCampaign.clickSkipPaidAdsCreationButton();
+					await page.evaluate( () => window.sessionStorage.clear() );
 				} );
 
 				test( 'should see the modal', async () => {
@@ -692,7 +692,7 @@ test.describe( 'Complete your campaign', () => {
 				await completeCampaign.goto();
 				await completeCampaign.clickSkipPaidAdsCreationButton();
 				await completeCampaign.clickCompleteSetupModalButton();
-				await page.waitForNavigation( {
+				await page.waitForURL( /path=%2Fgoogle%2Fproduct-feed/, {
 					waitUntil: 'domcontentloaded',
 				} );
 				await page.evaluate( () => {
--- a/tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
+++ b/tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
@@ -486,11 +486,9 @@ test.describe( 'Complete your campaign', () => {
 		test.describe( 'User skips paid ads creation', () => {
 			test.describe( 'With WooCommerce tracking disabled', () => {
 				test.beforeAll( async () => {
-					// Reset the showing status for the "Set up paid ads" section.
 					await setupAdsAccountPage.mockAdsAccountIncomplete();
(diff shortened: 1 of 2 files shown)


tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
```diff
@@ -692,7 +692,7 @@ test.describe( 'Complete your campaign', () => {
 				await completeCampaign.goto();
 				await completeCampaign.clickSkipPaidAdsCreationButton();
 				await completeCampaign.clickCompleteSetupModalButton();
-				await page.waitForNavigation( {
+				await page.waitForURL( /path=%2Fgoogle%2Fproduct-feed/, {
```
joemcgill, line 693: `page.waitForNavigation()` is deprecated because it can be racy. This updates the test to use `waitForURL` as recommended in [the docs](https://playwright.dev/docs/api/class-page#page-wait-for-navigation).

tests/e2e/specs/product-editor/block-integration.test.js
```diff
@@ -13,6 +13,7 @@ test.use( { storageState: process.env.ADMINSTATE } );
 test.describe.configure( { mode: 'serial' } );
 
 test.describe( 'Product Block Editor integration', () => {
+	test.skip( 'Experimental block editor is deprecated in v10.1.0' );
```
joemcgill, line 16: Should we conditionally skip these based on the version of the plugin being tested, or should we remove these altogether for now?

tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
```diff
@@ -409,14 +409,14 @@ test.describe( 'Complete your campaign', () => {
...
 				test.beforeEach( async () => {
-					await page.evaluate( () => window.sessionStorage.clear() );
 					await setupAdsAccountPage.mockAdsAccountIncomplete();
 					await setupBudgetPage.fulfillBillingStatusRequest( {
 						status: 'approved',
 					} );
 					await completeCampaign.mockCompleteAdsSetup();
 					await completeCampaign.fulfillBudgetRecommendations();
 					await completeCampaign.goto();
+					await page.evaluate( () => window.sessionStorage.clear() );
```
joemcgill, line 419: These two `page.evaluate()` calls (see below) were failing due to a race condition with page navigation events when I was testing locally. Any concerns with moving this until after we're on the page state we expect?

joemcgill (commented):
@eason9487 curious if you could take a look at these changes that were flagged as part of compatibility testing of 10.1.0.

I'm unsure if we are better off to remove the block integration tests altogether, or if we should try to conditionally run these depending on the Woo Core version. Everything else seemed to be related to race conditions when page navigations occur.

tests/e2e/specs/product-editor/block-integration.test.js
eason9487: Considering that WooCommerce 10.1 has only removed its feature option from the UI, it can remain as is, with tests skipped but not removed. Future actions could be aligned with WooCommerce Core's deprecation or adjustments to its direction and progress.

tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
eason9487: I am unable to reproduce the issue you encountered, so I cannot verify whether moving this line is the appropriate solution. It seems counterintuitive here, but I don't think it's a big problem if you confirm it's necessary.

tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
```diff
@@ -487,10 +487,10 @@ test.describe( 'Complete your campaign', () => {
 			test.describe( 'With WooCommerce tracking disabled', () => {
 				test.beforeAll( async () => {
 					// Reset the showing status for the "Set up paid ads" section.
-					await page.evaluate( () => window.sessionStorage.clear() );
 					await setupAdsAccountPage.mockAdsAccountIncomplete();
 					await completeCampaign.goto();
 					await completeCampaign.clickSkipPaidAdsCreationButton();
+					await page.evaluate( () => window.sessionStorage.clear() );
```
eason9487: Using `sessionStorage` to conditionally show the campaign section should have been removed by #2533. The code comment "Reset the showing status [...]" and this `window.sessionStorage.clear()` seem to be redundant.

vikrampm1:
@joemcgill @asvinb looks like some review feedback needs to be addressed here before we make the PR ready to be merged?

tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
joemcgill: I'm unsure why the order matters in this case, but waiting until after the tests have navigated to the correct screen before clearing sessionStorage seems to make this more consistent.

tests/e2e/specs/onboarding/step-3-complete-campaign.test.js
joemcgill: I can confirm that removing the line to run window.sessionStorage.clear() has no affect on the tests passing, so I think it can be safely removed. Handled in d92ad5fcf.

joemcgill:
@eason9487 I've addressed your comments and confirmed that the E2E tests still pass on both [10.1.rc-2](https://github.com/woocommerce/google-listings-and-ads/actions/runs/16896175908) and the [current version of WC](https://github.com/woocommerce/google-listings-and-ads/actions/runs/16896458967). If you're happy to approve these changes, we can merge this and handle the "tested up to" bump in the next plugin release.

eason9487 (approved):
Thanks for the work!

PR #3046, opened 2025-08-07
API Pull: Change to use opt-out when filtering syncable products data according to channel visibility

eason9487:

For API Pull mode, this PR changes to use opt-out when filtering syncable products data according to channel visibility. In addition to using logic consistent with push mode, pull mode no longer relies on default channel visibility metadata populated in push mode. This further decouples the two modes. Ref:
- https://github.com/woocommerce/google-listings-and-ads/blob/3.3.0/src/Product/ProductRepository.php#L236-L249
- https://github.com/woocommerce/google-listings-and-ads/blob/3.3.0/src/Admin/MetaBox/ChannelVisibilityMetaBox.php#L166-L167
- https://github.com/woocommerce/google-listings-and-ads/blob/3.3.0/src/Product/ProductHelper.php#L262-L264

In addition, this PR also makes the data provider in `ProductHelperTest` efficient and have no side effects. This will make development more efficient because when using `--filter` to run test suites, all data provider methods will still be called regardless of whether the decorated test cases are included or not

1. Reset products' channel visibility metadata that may have been set by push mode
   ```sql
   DELETE FROM wp_postmeta WHERE meta_key = '_wc_gla_visibility';
   ```
2. Open the Console panel of the browser's DevTool
3. Run the following scripts to simulate the requests sent to `WPCOMProxy`, and check if only opted-out products do not appear in the results. (Note that the results are paginated)
   - `wp.apiFetch({ path:'/wc/v3/products?orderby=id&gla_syncable=1' }).then(console.log)`
   - `wp.apiFetch({ path:'/wc/v3/products/<product_id>?orderby=id&gla_syncable=1' }).then(console.log)`
   - `wp.apiFetch({ path:'/wc/v3/products/<variable_product_id>/variations?orderby=id&gla_syncable=1' }).then(console.log)`
   - `wp.apiFetch({ path:'/wc/v3/products/<variable_product_id>/variations/<variation_product_id>?orderby=id&gla_syncable=1' }).then(console.log)`
4. Go to edit the **Channel visibility** for some products that appeared or didn't appear in the previous results, and test if their appearances in the new results are changed accordingly
   image

### Changelog entry

> Update - The filtering for synchronizable products data by channel visibility in API Pull mode is now opt-out, changed from its previous opt-in setting.

### DIFF
--- a/src/Integration/WPCOMProxy.php
+++ b/src/Integration/WPCOMProxy.php
@@ -12,6 +12,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product\Attributes\AttributeManager;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\ProductRepository;
 use WC_Product;
 use WP_REST_Response;
 use WP_REST_Request;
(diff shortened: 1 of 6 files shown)


NeosinneR (approved):
Code looks good. I tested the before and after, and can confirm that the default is out-out, unless changed not to sync. 

| Before | After |
| ---- | ---- | 
| Screenshot 2025-08-08 at 14 02 22 | Screenshot 2025-08-08 at 14 02 07 |

PR #3053, opened 2025-08-13
Incorporate Google Ads API changes for EU political advertising.

ankitrox:

Closes: https://linear.app/a8c/issue/GOOWOO-241/incorporate-google-ads-api-changes-for-eu-political-advertising

No separate testing is required, code reviewer can make sure that the fields are added properly during mentioned methods and this issue can be moved to approval. A round of smoke test can perform around ads functionality.

### Changelog entry

> Update - Incorporate API changes for EU political advertising.


src/API/Google/AdsCampaign.php
```diff
@@ -27,6 +27,7 @@
 use Google\Ads\GoogleAds\V20\Services\GoogleAdsRow;
 use Google\Ads\GoogleAds\V20\Services\MutateGoogleAdsRequest;
 use Google\Ads\GoogleAds\V20\Services\MutateOperation;
+use Google\Ads\GoogleAds\V21\Enums\EuPoliticalAdvertisingStatusEnum\EuPoliticalAdvertisingStatus;
```
mukeshpanchal27: Should be V20 not V21?

Move it after `use Google\Ads\GoogleAds\V20\Enums\AdvertisingChannelTypeEnum\AdvertisingChannelType;`

ankitrox: We don't have it in v20 ([ref](https://github.com/googleads/google-ads-php/tree/main/metadata/Google/Ads/GoogleAds/V20/Enums)), its available in v21 only ([ref](https://github.com/googleads/google-ads-php/tree/main/metadata/Google/Ads/GoogleAds/V21/Enums))

mukeshpanchal27: Here it is https://github.com/googleads/google-ads-php/blob/df36b22ad1ca82e8f931d6c171f0e5ef6e894587/src/Google/Ads/GoogleAds/V20/Enums/EuPoliticalAdvertisingStatusEnum/EuPoliticalAdvertisingStatus.php

eason9487: There is a [post-script](https://github.com/woocommerce/google-listings-and-ads/blob/3.3.1/composer.json#L62) after running composer install, which removes unused Google Ads library files to reduce the plugin's bundle size. Therefore, after importing a new class that has not been used before from the `Google\Ads\GoogleAds` namespace, it needs to fully remove the `vendor` directory and reinstall composer dependencies.

package.json
```diff
@@ -163,7 +163,7 @@
 			},
 			{
 				"path": "./google-listings-and-ads.zip",
-				"maxSize": "8.22 mB",
+				"maxSize": "14.2 mB",
```
eason9487: Related to https://github.com/woocommerce/google-listings-and-ads/pull/3053#discussion_r2275675464.

After upgrading `googleads/google-ads-php` to `dev-legacy-v31.0.0`, appending `21` to [this script](https://github.com/woocommerce/google-listings-and-ads/blob/3.3.1/composer.json#L63) should keep the zip file size from increasing so significantly.

src/API/Google/AdsCampaign.php
szepeviktor: 👀 https://github.com/googleapis/googleapis/blob/b1e61e0ad25d5e6050e8da295f5e14f999f06632/google/ads/googleads/v20/enums/eu_political_advertising_status.proto#L35-L49

composer.json
```diff
@@ -60,7 +60,7 @@
 			"Google\\Task\\Composer::cleanup",
 			"Automattic\\WooCommerce\\GoogleListingsAndAds\\Util\\SymfonyPolyfillCleanup::remove",
 			"Automattic\\WooCommerce\\GoogleListingsAndAds\\Util\\GoogleAdsCleanupServices::remove",
-			"composer run-script remove-google-ads-api-version-support -- 18 19",
+			"composer run-script remove-google-ads-api-version-support -- 18 19 21",
```
joemcgill, line 63: @ankitrox it looks like the addition of `21` here can be removed now since you've reverted the changes to the AdsCampaign.php file that made this necessary (see [this conversation](https://github.com/woocommerce/google-listings-and-ads/pull/3053#discussion_r2273172725)).


```suggestion
			"composer run-script remove-google-ads-api-version-support -- 18 19",
```

joemcgill (commented):
One nitpick which I will clean up since everything else looks to be working as expected.

composer.json
joemcgill: Addressed [de3541f](https://github.com/woocommerce/google-listings-and-ads/pull/3053/commits/de3541fc338c592817a3074eaa9a18bf4040f304)

joemcgill: I was mistaken. It looks like v21 library files were still added even though none of those files were in use. I've reverted this change in [e1998c4](https://github.com/woocommerce/google-listings-and-ads/pull/3053/commits/e1998c487486eaa8f09fadfeb0901bc888c56c97) and confirmed that things are still working as intended.

joemcgill (approved):

joemcgill:
I've tested these changes and confirmed that mutate operations for new and edited campaigns are setting the `contains_eu_political_advertising` field as expected.

PR #3054, opened 2025-08-15
Use setValues instead of individual setValue calls.

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-245/investigate-long-headline-validation-issue

I suspect there’s a race condition occurring due to successive calls to formContext.setValue. Upon investigation, the issue affects not only the long headline but also other fields, such as the business name. It does not reproduce consistently.

1. Navigate to the plugin dashboard.
2. Edit a campaign
3. In the second step, after choosing an URL, the form should be populated with some initial data.
4. There should be no validation errors if a field is valid (for e.g some fields are not blank)
5. Check for when there are empty/non empty assets for the campaign.

### Changelog entry

> Fix - Replaced multiple `setValue` calls with a single `setValues` call to eliminate race conditions during form initialization and reduce inconsistent validation behavior.

### DIFF
--- a/js/src/components/paid-ads/campaign-assets-form.js
+++ b/js/src/components/paid-ads/campaign-assets-form.js
@@ -158,12 +158,18 @@ export default function CampaignAssetsForm( {
 					: initialAssetGroup;
 				let hasNonEmptyAssets = false;
 
+				const updatedContextValues = {};
 				Object.keys( emptyAssetGroup ).forEach( ( key ) => {
 					if ( assetGroup && assetGroup[ key ]?.length ) {
 						hasNonEmptyAssets = true;
 					}
-					formContext.setValue( key, nextAssetGroup[ key ] );
+
+					updatedContextValues[ key ] = nextAssetGroup[ key ];
 				} );
+
+				if ( Object.keys( updatedContextValues ).length ) {
+					formContext.setValues( updatedContextValues );
+				}
 
 				setHasImportedAssets( hasNonEmptyAssets );
 				setBaseAssetGroup( nextAssetGroup );
--- a/js/src/components/paid-ads/campaign-assets-form.js
+++ b/js/src/components/paid-ads/campaign-assets-form.js
@@ -158,14 +158,14 @@ export default function CampaignAssetsForm( {
 					: initialAssetGroup;
 				let hasNonEmptyAssets = false;
 
-				const updatedContextValues = {};
-				Object.keys( emptyAssetGroup ).forEach( ( key ) => {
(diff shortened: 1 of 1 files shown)


joemcgill (approved):
This code change seems ok, but does't really explain why we're seeing incorrect validation errors after selecting a new final URL and submitting the form.

Looking at this function, it seems like we're saving the same state for the asset group values in two places: formContext values and also the baseAssetGroup state value in this component. From what I can tell, the `updatedContextValues` object will be the same as the `nextAssetGroup` object, so I'm not really sure what we're trying to accomplish here, other than looping through the object to see if `hasNonEmptyAssets` should be set to `true`.

Having better reproduction steps, or a test case that demonstrates the bug would be helpful.

js/src/components/paid-ads/campaign-assets-form.js
```diff
@@ -158,12 +158,18 @@ export default function CampaignAssetsForm( {
...
+					Object.keys( emptyAssetGroup ).map( ( key ) => {
+						if ( assetGroup?.[ key ]?.length ) {
+							hasNonEmptyAssets = true;
+						}
+						return [ key, nextAssetGroup[ key ] ];
+					} )
+				);
+
+				if ( Object.keys( updatedContextValues ).length ) {
+					formContext.setValues( updatedContextValues );
```
eason9487, line 171: On machines with sufficient processing speed, the following code blocks may cause a race condition, where the next `delegation` is dispatched before the previous `delegation` has been processed by the `useEffect` callback. When I wrote this handling previously, my machine was too slow to notice this issue.

https://github.com/woocommerce/google-listings-and-ads/blob/508b848a36ea5cf4122bbd9dbd6208535168bb95/js/src/components/adaptive-form/adaptive-form.js#L190-L199

https://github.com/woocommerce/google-listings-and-ads/blob/508b848a36ea5cf4122bbd9dbd6208535168bb95/js/src/components/adaptive-form/adaptive-form.js#L64-L68

Since compatibility with WC < 7.1 is no longer required, it's okay to call `setValues` directly to avoid that race condition.

ankitguptaindia:
QA Status: Fixed. ✅ 
---

**Test Result:** After the fix, proper validations are now appearing in the fields. If any field exceeds the required limit, a warning message is displayed. Similarly, if any required field is missing, a warning prompts the user to fill it before creating the campaign.

**Quick Demo:** 


**Before Fix:** 

<details>

image

image


</details> 


**After Fix:** 


https://github.com/user-attachments/assets/640f93a8-bb64-42a1-bcc2-c57b39e11b12

ankitguptaindia (approved):

PR #3059, opened 2025-08-19
Add RaiseBudgetRecommendationBanner component

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-242/create-raise-budget-recommendation-notice-component .

### Screenshots:
image

1. Navigate to the plugin dashboard with a connected Ads account.
2. Mock the `/wp-json/wc/gla/ads/recommendations?types%5B0%5D=CAMPAIGN_BUDGET&types%5B1%5D=MARGINAL_ROI_CAMPAIGN_BUDGET` endpoint with the following [data](https://gist.github.com/asvinb/656273e498218db4fca74bd26dc3a9f1).
3. Ensure the campaign ID in the mocked data matches your existing campaigns.
4. The banner will be rendered.
5. If there are no campaigns, no recommendations and no connected Google Ads account, the banner should not be displayed or trigger any error.
6. The percentage displayed is the % increase in conversions from the recommended budget compared to the budget metrics. Same applies for the conversion value.

### Changelog entry

>


js/src/components/raise-budget-recommendation-banner/banner.js
```diff
@@ -0,0 +1,208 @@
...
+ * @fires gla_raise_budget_recommendation_dismiss_clicked when the banner is dismissed.
+ *
+ * @param {Object} props Component properties.
+ * @param {Function} props.onBannerDismissed Callback function to call when the banner is dismissed.
+ * @return {JSX.Element|null} The banner component, or null if not applicable.
+ */
+const Banner = ( { onBannerDismissed } ) => {
+	const { campaigns: recommendedCampaigns } = useRaiseBudgetRecommendations();
+	const { data: allCampaigns } = useAdsCampaigns();
+	const recommendedCampaign = recommendedCampaigns?.[ 0 ] || {};
```
asvinb: @jjgrainger We are taking the first campaign from the list of recommended campaigns. Can you confirm this is ok and that the endpoint are returning campaigns ordered by their highest budget?

jjgrainger: @asvinb reviewing the PRD I can't see any reference to which campaign should be shown in the banner if there a multiple? The endpoint doesn't order the recommendations based on the highest budget, it returns them in the order retrieved from the Google Ads API.

Can we confirm which campaign should be shown when there are multiple? It might be like the PMAX recommendations where we surface a banner for the highest spend campaign?

jjgrainger (approved):

js/src/components/raise-budget-recommendation-banner/banner.js
asvinb: Thanks @jjgrainger . I updated the PR by finding the highest paid campaign from the list of recommended ones.

ankitguptaindia:
**Next step** Moving back to feedback. ⚠️ 

@asvinb Thanks for the implementation of this banner. I observed two cases here, are as follows:

1. In design, banner is below the main navigation setting tabs, but in developed, it is above the setting tabs.

|Developed | Design |
|--------|--------|
| image | image |


2. Project Weekly Estimate is appearing in both positive and negative ranges. Is this due to mock data we are using? 

image

asvinb:
@ankitguptaindia I've reached out to Michael for the first bullet point. As for the second issue, I pushed a tweak to cater for negative values as well. I originally catered for positive values only since technically it should always be positive based on the recommendations :) Can you kindly check again please?

asvinb:
@ankitguptaindia I've changed the placement of the banner to be just below the tabbed navigation as agreed with Michael.

ankitguptaindia:
QA Status: Verfied ✅ 
--

Tested banner display with mock data in the browser console. All feedback has been addressed and banner is appearing as per design.

image

PR #3065, opened 2025-08-25
Add current row in edit campaign form.

ankitrox:

Closes: https://linear.app/a8c/issue/GOOWOO-249/add-current-row-to-budgetsetup 

### Screenshots:

Screenshot 2025-08-25 at 6 05 22 PM

1. Edit an existing campaign.
2. Ensure that there's a row available for `Current` and it is selected. Refer screenshot
3. When switching to `Set custom budget`, the values should be same as `Recommended` ones.

### Changelog entry

>


js/src/pages/edit-paid-ads-campaign/index.js
```diff
@@ -226,8 +226,8 @@ const EditPaidAdsCampaign = () => {
 			/>
 			<CampaignAssetsForm
 				initialCampaign={ {
-					level: 'custom',
-					amount: campaign.amount,
+					level: 'current',
+					currentAmount: campaign.amount,
```
asvinb: @ankitrox I think we should have the `currentAmount` as a prop itself instead of using the `initialCampaign` property. Then you can access it via the `formContext.extender` property. You can take `countryCodes` as example in `CampaignAssetsForm`. So I suggest we do the following:
* Pass `currentAmount` as a prop itself to `CampaignAssetsForm`
* In `CampaignAssetsForm`, return `currentAmount` from the `extendAdapter` function.
* In `BudgetSetup`, access `currentAmount` from `adapter`.
-

js/src/components/paid-ads/campaign-assets-form.js
```diff
@@ -73,6 +73,10 @@ function injectDailyBudget( values, budgetRecommendation ) {
 			if ( this.level === 'custom' ) {
 				return this.amount;
 			}
+
+			if ( this.level === 'current' ) {
+				return this.currentAmount;
```
asvinb: @ankitrox We can return `null` here because the this value is never read because we are showing that value directly in `BudgetSetup` via `formContext.adapter.currentAmount`. See my comment below for `EditPaidAdsCampaign`

js/src/components/paid-ads/budget-setup/budget-setup.js
```diff
@@ -66,11 +67,15 @@ export default function BudgetSetup( { hideRecommendations = false } ) {
 	const formContext = useAdaptiveFormContext();
 	const { adapter, getInputProps, values } = formContext;
 	const { countryCodes, budgetRecommendation } = adapter;
-	const { amount } = values;
+	const { amount, currentAmount } = values;
```
asvinb: Let's access `currentAmount` via `adapter` instead of `values`. See my last comment on the PR.

js/src/components/paid-ads/budget-setup/budget-setup.js
```diff
@@ -152,6 +157,32 @@ export default function BudgetSetup( { hideRecommendations = false } ) {
 				);
 			} ) }
 
+			{ currentAmount && (
+				<div className={ getRowClassName( 'current' ) }>
+					<BudgetRadioControl
+						{ ...getInputProps( 'level' ) }
+						value="current"
+						label={
+							<>
```
asvinb: @ankitrox Can we extract this in a separate component to prevent the markup/code duplication? Maybe we can call it `DailyBudgetLabel` which has `amount` as prop.

js/src/components/paid-ads/budget-setup/budget-setup.test.js
```diff
@@ -32,8 +32,8 @@ jest.mock( '~/hooks/useBudgetMetrics', () =>
 				country: 'US',
 				dailyBudget,
 				metrics: {
-					// Multiply by 8 rather than 7 to distinguish from the recommended one
-					cost: dailyBudget * 8,
+					// Multiply by 7 because custom budget would show the same metrics as the recommended ones.
```
asvinb, line 35: @ankitrox Why this change please?

asvinb (changes requested):

js/src/components/paid-ads/budget-setup/budget-setup.test.js
ankitrox: @asvinb We are now showing the recommended values when someone selects the custom input, thus it would be same that which is multiple of dailyBudget by 7.

js/src/components/paid-ads/campaign-assets-form.test.js
```diff
@@ -191,7 +191,7 @@ describe( 'CampaignAssetsForm', () => {
 		expect( children ).toHaveBeenLastCalledWith( formContextSchema );
 	} );
 
-	it( 'Should fall back to the custom level for the initial form values when there is no level available', () => {
+	it( 'Should fall back to the current level for the initial form values when there is no level available', () => {
```
asvinb: @ankitrox I feel the label is a bit misleading here because there's no fall back to the current level. The current level is always selected when editing a paid ads campaign.

js/src/components/paid-ads/campaign-assets-form.test.js
```diff
@@ -205,7 +205,10 @@ describe( 'CampaignAssetsForm', () => {
 		render(
 			<CampaignAssetsForm
 				validate={ alwaysValid }
-				initialCampaign={ { amount: 10, level: 'high' } }
+				initialCampaign={ {
+					amount: 10,
+				} }
+				currentAmount={ 12 }
```
asvinb: @ankitrox The `currentAmount` should theoretically have the same value as `amount`.

js/src/components/paid-ads/campaign-assets-form.test.js
```diff
@@ -214,15 +217,15 @@ describe( 'CampaignAssetsForm', () => {
...
+				dailyBudget: null,
+				level: 'current',
 			} ),
 		} );
 
 		expect( children ).toHaveBeenLastCalledWith( formContextSchema );
 	} );
 
-	it( 'Should fall back to the custom level for the initial form values when budget recommendation responds with 404 not found', () => {
+	it( 'Should fall back to the current level for the initial form values when budget recommendation responds with 404 not found', () => {
```
asvinb: @ankitrox See my previous comment about the label.

asvinb (changes requested):
@ankitrox The code LGTM. I left a few comments about the test though. Can you check them out please? Also, can we add a test where the "Current" row does not render when `currentAmount` is not defined?

js/src/components/paid-ads/campaign-assets-form.test.js
ankitrox: I think we don't need this test as the current level would be selected when editing the paid campaign.

js/src/components/paid-ads/campaign-assets-form.test.js
ankitrox: Thank you. Changed the `currentAmount` to `10`

js/src/components/paid-ads/campaign-assets-form.test.js
ankitrox: Updated the label

```
Should show the current level as selected one when editing a paid campaign
```

ankitrox:
Thank you @asvinb for reviewing the PR and adding the feedback. I've addressed the points mentioned and also added an additional test requested.

asvinb (approved):

asvinb:
Thanks @ankitrox . LGTM! 👍

ankitguptaindia:
QA Status: Needs feedback ⚠️ 
----

@ankitrox 

>When switching to Set custom budget, the values should be same as Recommended ones.

Currently, all values are matching when switching to Custom budget except 'Weekly conversions', 'Weekly conv. value'.  In the recommended section, it has 10.9, but in custom it has 10.8. Same for Weekly conv, it has $434.55 and $433.76 in the custom field.
Is this intentional? 

image

ankitguptaindia:
QA Status: Verified ✅ 
----


@ankitrox, regarding the above issue, I rechecked this today, and now I am seeing the same values in 'Recommended' and 'Custom' fields with the same branch and the same setup. Verifying this PR now. 




https://github.com/user-attachments/assets/e972b1f8-a77b-4480-b1c0-766a833e3b3e

PR #3070, opened 2025-08-27
Schedule job to sync all products when the Push mode of product sync is switched to enable

eason9487:

In the case of an issue during the sync mode migration, and a rollback strategy is needed to revert a merchant back to push mode, this PR schedules a job to sync all products when the Push mode of product sync is switched to enable. This will provide the ability to trigger a full re-sync of all products for the merchant.

In addition, it also wraps the calls to WP global functions in the `Options` class within the `WP` proxy class. This is to enable the use or coverage of the `Options` class when writing PHPUnit tests while avoiding the side effects of data modification.

Ref:
- p1754564027278729-slack-C02BB3F30TG
- p1753871675559559-slack-C02BB3F30TG

💡 If using a local env for testing, it may need to make a temporary adjustment to prevent scheduling from being blocked by the following check. For example, make it always return true.

https://github.com/woocommerce/google-listings-and-ads/blob/ca66d673db1255695e16adf388d738e5cdff5cd2/src/MerchantCenter/MerchantCenterService.php#L109-L126

1. Complete the onboarding
2. Toggle the Push mode config of product sync via the `POST /wp-json/wc/gla/sync` API. One available method is to run the following scripts via the DevTools Console:
   ```js
   // Disable push mode
   wp.apiFetch({
     path: '/wc/gla/sync',
     method: 'POST',
     data: { products: { push: false } }
   }).then(console.log)

   // Enable push mode
   wp.apiFetch({
     path: '/wc/gla/sync',
     method: 'POST',
     data: { products: { push: true } }
   }).then(console.log)
   ```
3. Check if the `gla/jobs/update_all_products/create_batch` job is scheduled when the Push mode of product sync is switched to enable
   image

### Changelog entry

> Update - Schedule resynchronization for all products via the Push mode when the Push mode of product synchronization is switched to enable.

### DIFF
--- a/src/Event/StartProductSync.php
+++ b/src/Event/StartProductSync.php
@@ -14,7 +14,7 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Class StartProductSync
  *
- * @package Automattic\WooCommerce\GoogleListingsAndAds\Product
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Event
  */
 class StartProductSync implements Registerable, Service {
 	/**
--- a/src/Event/StartProductSync.php
+++ b/src/Event/StartProductSync.php
@@ -48,6 +48,15 @@ class StartProductSync implements Registerable, Service {
 				$this->on_rules_change();
 			}
 		);
+
+		add_action(
+			'woocommerce_gla_sync_mode_updated',
+			function ( $prev_sync_mode, $sync_mode ) {
+				$this->on_sync_mode_updated( $prev_sync_mode, $sync_mode );
+			},
+			10,
+			2
+		);
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


src/Event/StartProductSync.php
```diff
@@ -68,4 +77,17 @@ protected function on_rules_change() {
...
 	}
+
+	/**
+	 * If the Push mode of product sync is switched to enable, schedule a job to sync all products.
+	 *
+	 * @param array $prev_sync_mode The previous sync mode.
+	 * @param array $sync_mode The current sync mode.
+	 */
+	protected function on_sync_mode_updated( $prev_sync_mode, $sync_mode ) {
+		if ( $prev_sync_mode['products']['push'] === false && $sync_mode['products']['push'] === true ) {
```
NeosinneR: Is this guaranteed that there will be `['products']['push']` properties? Just want to make sure this can't throw an error when those keys don't exist.

NeosinneR (commented):
Overall looks good and works as expected, just a minor note in the comment below.

src/Event/StartProductSync.php
eason9487: Thanks! Throwing an error is possible because there is a WP filter when getting the default sync mode config. Fixed in https://github.com/woocommerce/google-listings-and-ads/pull/3070/commits/00da35c812e7e667ad615dfa8826282a455a28d9.

NeosinneR: Thanks for the change. Looks good!

NeosinneR (approved):

PR #3080, opened 2025-09-03
Feature/247 enhance adsrecommendationservice

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-247/enhance-adsrecommendationservice-for-budget-recommendations

1. Ensure the test proxy is running
2. Comment the permission `'permission_callback' => $this->get_permission_callback(),` callback in the `src/API/Site/Controllers/Ads/RecommendationsController` file.
3. Make a request to the REST API endpoint with the following types - `/wp-json/wc/gla/ads/recommendations?types=CAMPAIGN_BUDGET,MARGINAL_ROI_CAMPAIGN_BUDGET`
4. A response contains a recommendation with:
    - The correct recommendation type `CAMPAIGN_BUDGET`
    - Has a `details` property with the `campaign_budget_recommendation` information.
5. When requesting the endpoint with `types= IMPROVE_PERFORMANCE_MAX_AD_STRENGTH ` the correct recommendation and type is returned and the `details` property for the recommendation is an empty array. 

### Changelog entry

>


src/Ads/AdsRecommendationsService.php
```diff
@@ -125,6 +123,36 @@ public static function get_valid_recommendation_types( array $types ): array {
...
+	 */
+	private function get_additional_columns_for_recommendation_type( array $types ): array {
+		$columns = [
+			'IMPROVE_PERFORMANCE_MAX_AD_STRENGTH' => [],
+			'CAMPAIGN_BUDGET'                     => [
+				'recommendation_campaign_budget_recommendations' => 'recommendation.campaign_budget_recommendation',
+			],
+			'MARGINAL_ROI_CAMPAIGN_BUDGET'        => [
+				'recommendation_campaign_budget_recommendations' => 'recommendation.marginal_roi_campaign_budget_recommendation',
+			],
```
mukeshpanchal27: Both keys are same so it return last one as per requested types. It should return both columns if it requested.

For `&types=CAMPAIGN_BUDGET,MARGINAL_ROI_CAMPAIGN_BUDGET`
```
Array
(
    [recommendation_campaign_budget_recommendations] => recommendation.marginal_roi_campaign_budget_recommendation
)
```

For `&types=MARGINAL_ROI_CAMPAIGN_BUDGET,CAMPAIGN_BUDGET`
```
Array
(
    [recommendation_campaign_budget_recommendations] => recommendation.campaign_budget_recommendation
)
```

tests/proxy/handler.js
```diff
@@ -27,7 +27,14 @@ module.exports.checkRequest = ( request, h ) => {
...
 			}
-			return require( './mocks/ads/recommendations/results.json' );
+
+			if (
+				body.query.includes( 'IMPROVE_PERFORMANCE_MAX_AD_STRENGTH' )
+			) {
+				return require( './mocks/ads/recommendations/pmax-asset.json' );
+			}
+
+			return require( './mocks/ads/recommendations/campaign-budget.json' );
```
mukeshpanchal27, line 37: Not a blocker

Do we have mock data for combination of different types it help us to get clear idea about the shape?
- `CAMPAIGN_BUDGET, MARGINAL_ROI_CAMPAIGN_BUDGET`
- `IMPROVE_PERFORMANCE_MAX_AD_STRENGTH, CAMPAIGN_BUDGET`
- `IMPROVE_PERFORMANCE_MAX_AD_STRENGTH, CAMPAIGN_BUDGET, MARGINAL_ROI_CAMPAIGN_BUDGET`

tests/proxy/mocks/ads/recommendations/campaign-budget.json
```diff
@@ -0,0 +1,84 @@
+{
+	"results": [
+        {
+            "recommendation": {
+                "type": "CAMPAIGN_BUDGET",
+                "resource_name": "customers/123/recommendations/1",
+				"campaign_budget_recommendation": {
```
mukeshpanchal27, line 7: In `src/API/Google/BudgetRecommendations.php`, I noticed the condition handling cases where `getCampaignBudgetRecommendation()` may return empty:
[https://github.com/woocommerce/google-listings-and-ads/blob/bbb59d6aab89a85d2c6d38965b27a3692b566cd7/src/API/Google/BudgetRecommendations.php#L100-L103](https://github.com/woocommerce/google-listings-and-ads/blob/bbb59d6aab89a85d2c6d38965b27a3692b566cd7/src/API/Google/BudgetRecommendations.php#L100-L103)

Do we need to verify whether this is working as expected or if it could potentially surface a bug?

cc. @joemcgill

src/Ads/AdsRecommendationsService.php
```diff
@@ -156,6 +205,45 @@ public function get_google_recommendations( $args ): array {
...
+					$budget_recommendation = $recommendation->getCampaignBudgetRecommendation();
+					$budget_options        = [];
+
+					foreach ( $budget_recommendation->getBudgetOptions() as $option ) {
+						$impact    = $option->getImpact();
+						$base      = $impact->getBaseMetrics();
+						$potential = $impact->getPotentialMetrics();
+
+						$budget_options[] = [
+							'budget_amount_micros' => $option->getBudgetAmountMicros(),
```
mukeshpanchal27: Do we needs to use `$this->from_micro` similar to https://github.com/woocommerce/google-listings-and-ads/blob/bbb59d6aab89a85d2c6d38965b27a3692b566cd7/src/API/Google/BudgetRecommendations.php#L129

src/Ads/AdsRecommendationsService.php
```diff
@@ -156,6 +205,45 @@ public function get_google_recommendations( $args ): array {
 				$campaign = $row->getCampaign();
 				$customer = $row->getCustomer();
 
+				$recommendation_type    = RecommendationType::name( $recommendation->getType() );
```
mukeshpanchal27, line 210: Is there any specific intention to call these here? Since we already have $types available, could we use that instead of making calls to the service class?

src/API/Site/Controllers/Ads/RecommendationsController.php
```diff
@@ -170,6 +166,10 @@ protected function get_schema_properties(): array {
 				'description' => __( 'Status of the campaign.', 'google-listings-and-ads' ),
 				'enum'        => [ 'ENABLED', 'PAUSED', 'REMOVED', 'UNKNOWN', 'UNSPECIFIED' ],
 			],
+			'details'         => [
```
mukeshpanchal27, line 169: Nice!

mukeshpanchal27 (changes requested):
Thanks @jjgrainger, this is looking pretty good so far. I’ve left some initial feedback for you.

Also, please make sure to merge the `feature/GOOWOO-234-raise-budget-recommendation` branch into this one.

mukeshpanchal27:
> If the transient approach is shown to work then the AdsRecommendationTable and other associated classes will be made redundant and should be removed. A migration may also be needed to delete the table from the database.

Will the missing piece be implemented in this same PR, or do you plan to handle it in a separate PR?

src/Ads/AdsRecommendationsService.php
jjgrainger: @mukeshpanchal27 as we can query for multiple recommendations I can't hard code what type it is as it can be one of multiple. It's better to determine the type from the recommendation itself.

src/Ads/AdsRecommendationsService.php
jjgrainger: @mukeshpanchal27 you're right, I've updated this so they two columns do not conflict

tests/proxy/handler.js
jjgrainger: @mukeshpanchal27 I've updated the mock response to include both the `CAMPAIGN_BUDGET` and `MARGINAL_ROI_CAMPAIGN_BUDGET` types.

I don't think theres a need for us to have responses that include the other combinations as we're not going to be querying for all types.

tests/proxy/mocks/ads/recommendations/campaign-budget.json
jjgrainger: @mukeshpanchal27 the recommendation response should include those recommendations, however I have added a check to make sure it's available before continuing.

src/Ads/AdsRecommendationsService.php
jjgrainger: @mukeshpanchal27 I think we will tackle this as part of https://linear.app/a8c/issue/GOOWOO-262/adapt-adsrecommendations-endpoint-for-raise-budget-recommendations

jjgrainger:
Thanks @mukeshpanchal27 

I've made updates based on your feedback and left comments.

> > If the transient approach is shown to work then the AdsRecommendationTable and other associated classes will be made redundant and should be removed. A migration may also be needed to delete the table from the database.
> 
> Will the missing piece be implemented in this same PR, or do you plan to handle it in a separate PR?

I think we should tackle this in a separate PR once we confirm this is working in testing and QA.

src/Ads/AdsRecommendationsService.php
```diff
@@ -102,12 +97,15 @@ public function get_recommendations( array $args = [] ): array {
 				'campaign_name'   => $item['recommendation_campaign_name'] ?? '',
 				'campaign_status' => $item['recommendation_campaign_status'] ?? '',
 				'customer_id'     => (int) ( $item['recommendation_customer_id'] ?? 0 ),
+				'details'         => $item['recommendation_details'],
```
mukeshpanchal27:
```suggestion
				'details'         => $item['recommendation_details'] ?? [],
```

Just for safeguard

mukeshpanchal27:
@jjgrainger Just to confirm, do you plan to add/update unit tests as part of this PR, or will that be handled in a separate PR?

tests/Unit/API/Site/Controllers/Ads/RecommendationsControllerTest.php
```diff
@@ -48,7 +48,7 @@ public function setUp(): void {
 		$this->controller->register();
 	}
 
-	public function test_get_recommendations_returns_emprty_result() {
+	public function test_get_recommendations_returns_empty_result() {
```
mukeshpanchal27, line 51: Thanks!

mukeshpanchal27 (approved):

ankitguptaindia:
QA Status: Verified ✅ 
---
Tested the PR by using a text proxy, and both REST API endpoints gave the expected result. 

- When requesting to the REST API endpoint with the following types - `/wp-json/wc/gla/ads/recommendations?types=CAMPAIGN_BUDGET,MARGINAL_ROI_CAMPAIGN_BUDGET`  giving result: 

Screenshot 2025-09-12 at 6 26 19 PM


----

- When requesting the endpoint with `types= IMPROVE_PERFORMANCE_MAX_AD_STRENGTH`  the correct recommendation and type is returned and the details property for the recommendation is an empty array.


Screenshot 2025-09-12 at 6 27 27 PM

ankitguptaindia (approved):

PR #3081, opened 2025-09-05
Feature/262 adapt ads recommendation endpoint for raise budget

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-262/adapt-adsrecommendations-endpoint-for-raise-budget-recommendations#comment-b1522e48

1. Ensure the test proxy is running
2. Comment the permission `'permission_callback' => $this->get_permission_callback(),` callback in the `src/API/Site/Controllers/Ads/RecommendationsController` file.
3. Make a request to the REST API endpoint with the following types - `/wp-json/wc/gla/ads/recommendations?types=CAMPAIGN_BUDGET,MARGINAL_ROI_CAMPAIGN_BUDGET`
4. A response contains a recommendations with:
    - `details` property with the `campaign_budget_recommendation` inside.
    - `campaign_budget_recommendation` has the following values:
        - `current_budget_amount `
        - `recommended_budget_amount`
        - `budget_options`
    - Each `budget_options` has a `level` property with either `High`, `Recommended`, `Low` or `Current`.
    - All budget and cost amounts are converted from micros to the currency amount.

### Changelog entry

>

### DIFF
--- a/src/Ads/AdsRecommendationsService.php
+++ b/src/Ads/AdsRecommendationsService.php
@@ -224,12 +224,28 @@ class AdsRecommendationsService implements ContainerAwareInterface, OptionsAware
 						continue;
 					}
 
+					$current_budget     = $budget_recommendation->getCurrentBudgetAmountMicros();
+					$recommended_budget = $budget_recommendation->getRecommendedBudgetAmountMicros();
+
 					foreach ( $budget_recommendation->getBudgetOptions() as $option ) {
-						$impact    = $option->getImpact();
-						$potential = $impact->getPotentialMetrics();
+						$impact        = $option->getImpact();
+						$potential     = $impact->getPotentialMetrics();
+						$budget_amount = $option->getBudgetAmountMicros();
+
+						// Determine budget option level.
+						$level = __( 'Low', 'google-listings-and-ads' );
+
+						if ( $budget_amount === $current_budget ) {
+							$level = __( 'Current', 'google-listings-and-ads' );
+						} elseif ( $budget_amount === $recommended_budget ) {
+							$level = __( 'Recommended', 'google-listings-and-ads' );
+						} elseif ( $budget_amount > $recommended_budget ) {
+							$level = __( 'High', 'google-listings-and-ads' );
+						}
 
 						$budget_options[] = [
-							'budget_amount' => $this->from_micro( $option->getBudgetAmountMicros() ),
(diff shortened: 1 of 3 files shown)


src/Ads/AdsRecommendationsService.php
```diff
@@ -222,33 +224,41 @@ public function get_google_recommendations( $args ): array {
 						continue;
 					}
 
+					$current_budget     = $budget_recommendation->getCurrentBudgetAmountMicros();
+					$recommended_budget = $budget_recommendation->getRecommendedBudgetAmountMicros();
```
mukeshpanchal27, line 228: In [::parse_recommendations()](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/API/Google/BudgetRecommendations.php#L148), it find closest matching number do we needs to do that here. It not 100% possible to get `Recommended` level if `$budget_amount === $recommended_budget` not match.

mukeshpanchal27 (commented):
Thanks @jjgrainger, this is solid to me. I’ve left one feedback for you.

src/Ads/AdsRecommendationsService.php
jjgrainger: Thanks @mukeshpanchal27 

From what I've seen in the example responses the budget amounts match up, so the levels should be calculated correctly here. I believe we match the closest in the budget recommendations because we work with a user entered value and the API/responses are different to the ones we work with here.

I believe the current implementation will work with the responses we get back here.

mukeshpanchal27 (approved):

ankitguptaindia:
QA Status: ✅ 
---
Tested the PR by using a text proxy, and the REST API endpoint gave the expected result.

- On a request to the REST API endpoint with the following types - `/wp-json/wc/gla/ads/recommendations?types=CAMPAIGN_BUDGET,MARGINAL_ROI_CAMPAIGN_BUDGET`, the response was received successfully and contains a recommendations object structured as expected


- The `details` property includes `campaign_budget_recommendation`.

- The `campaign_budget_recommendation` object contains the following:

    - `current_budget_amount` – shows the current campaign budget.

    - `recommended_budget_amount` – shows the suggested budget for better performance.

    - `budget_options` – an array with different budget suggestions.

- Each option includes a level property. 


Screenshot 2025-09-12 at 6 54 36 PM

ankitguptaindia (approved):

PR #3089, opened 2025-09-11
Feature/275 remove recommendations table

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-275/remove-the-adsrecommendationstable-and-other-classes

1. Update the src/DB/Migration/Migration20250910T1653383133.php to your current GLA version, for example: `3.4.1`
2. Adjust the `gla_db_version` option to an older version, such as 3.4.0, so that the migration class is triggered.: `UPDATE wp_options SET option_value = '3.4.0' WHERE option_name = 'gla_db_version';`
3. Refresh the GFW admin page.
4. Check the DB and see that the `wp_gla_ads_recommendations_table` is removed.
5. Check to see that any jobs for the `update_ads_recommendations` are no longer scheduled (either `cancelled` or `completed`). 

### Changelog entry

>


src/DB/Migration/Migration20250910T1653383133.php
```diff
@@ -0,0 +1,57 @@
...
+	 * @param \wpdb                   $wpdb
+	 * @param AdsRecommendationsTable $ads_recommendations_table
+	 */
+	public function __construct( \wpdb $wpdb, AdsRecommendationsTable $ads_recommendations_table ) {
+		parent::__construct( $wpdb );
+		$this->ads_recommendations_table = $ads_recommendations_table;
+	}
+
+
+	/**
```
mukeshpanchal27:
```suggestion
	public function __construct( \wpdb $wpdb, AdsRecommendationsTable $ads_recommendations_table ) {
		parent::__construct( $wpdb );
		$this->ads_recommendations_table = $ads_recommendations_table;
	}

	/**
```

Remove empty line

src/Jobs/UpdateAdsRecommendations.php
```diff
@@ -1,125 +0,0 @@
...
-	}
-
-	/**
-	 * Schedule the job.
-	 *
-	 * @param array $args - arguments.
-	 */
-	public function schedule( array $args = [] ) {
-		if ( $this->can_schedule( $args ) ) {
-			$this->action_scheduler->schedule_recurring( time(), $this->get_interval(), $this->get_process_item_hook(), $args );
```
mukeshpanchal27, line 112: I haven’t tested this part yet in review, but I’d like to get your thoughts:

The recurring scheduled job still runs according to its schedule. Do we need to explicitly remove these jobs during migration? After migration, the job and its related function are no longer available, which causes errors to appear in the logs.

mukeshpanchal27 (changes requested):
Thanks @jjgrainger for the PR 🙌

I’ve left some feedback. Could you also please add testing steps for the migration? That would make it easier for me and others to review.

Also, make sure the `src/DB/Table/AdsRecommendationsTable.php` file is removed as part of this change or do we need that?

src/Jobs/UpdateAdsRecommendations.php
jjgrainger: @mukeshpanchal27 good catch! I've added additional steps in the migration to remove those jobs from the database.

mukeshpanchal27 (approved):
Thanks @jjgrainger for the update. LGTM.

In https://github.com/woocommerce/google-listings-and-ads/pull/3089/commits/1f7cf2411cf32d7aeb72902c79d5e872f4407f91, I removed `AdsRecommendationsTable` from `TableManager` as it not needed now.

ankitguptaindia:
QA Status: Verified ✅ 
---

PR works as described. 

| Before | After |
|--------|--------|
| Screenshot 2025-09-16 at 3 47 01 PM |  Screenshot 2025-09-16 at 3 47 37 PM |
| Screenshot 2025-09-16 at 3 48 21 PM | Screenshot 2025-09-16 at 3 48 42 PM |

PR #3095, opened 2025-09-17
Product Bundles: Fix fatal error when product syncing

mukeshpanchal27:

Closes https://linear.app/a8c/issue/GOOWOO-276/bug-report-product-bundles-stuck-in-pending-status-fatal-error-when

#### Product sync
1. Create a WooCommerce Product Bundle
2. Add bundle items and enable "priced individually" for each item
3. Set items as optional (customers can choose which items to include)
4. Make sure It will not show any error. ( In case it show `Error` with details then make sure to solve it and save product)
5.  Make sure the product sync in your Merchant Center store.

#### Force Product sync
1. Go to `/wp-admin/admin.php?page=connection-test-admin-page`
2. Attempt to force sync the bundle product
3. No error shown
4. Make sure the product sync in your Merchant Center store.

### Changelog entry

> Fix - Error when syncing bundled products.

### DIFF
--- a/src/Integration/WooCommerceProductBundles.php
+++ b/src/Integration/WooCommerceProductBundles.php
@@ -175,7 +175,7 @@ class WooCommerceProductBundles implements IntegrationInterface {
 			return $price;
 		}
 
-		return $tax_excluded ? $product->get_bundle_regular_price_excluding_tax() : $product->get_bundle_regular_price_including_tax();
+		return $tax_excluded ? (float) $product->get_bundle_regular_price_excluding_tax() : (float) $product->get_bundle_regular_price_including_tax();
 	}
 
 	/**
@@ -188,8 +188,8 @@ class WooCommerceProductBundles implements IntegrationInterface {
 			return $sale_price;
 		}
 
-		$regular_price = $tax_excluded ? $product->get_bundle_regular_price_excluding_tax() : $product->get_bundle_regular_price_including_tax();
-		$price         = $tax_excluded ? $product->get_bundle_price_excluding_tax() : $product->get_bundle_price_including_tax();
+		$regular_price = $tax_excluded ? (float) $product->get_bundle_regular_price_excluding_tax() : (float) $product->get_bundle_regular_price_including_tax();
+		$price         = $tax_excluded ? (float) $product->get_bundle_price_excluding_tax() : (float) $product->get_bundle_price_including_tax();
 
 		// return current price as the sale price if it's lower than the regular price.
 		if ( $price < $regular_price ) {


joemcgill:
Thanks @mukeshpanchal27, are we certain that casting empty values as a float is the intended behavior here?

mukeshpanchal27:
> Thanks @mukeshpanchal27, are we certain that casting empty values as a float is the intended behavior here?

Per https://github.com/woocommerce/google-listings-and-ads/blob/fa66caca11338445ab58fdb4213bee1893d39e4f/src/Product/WCProductAdapter.php#L681, If the price is empty then it will not call the `woocommerce_gla_product_attribute_value_sale_price` filter so it will not affect the changes that we did in these PR.

joemcgill:
> If the price is empty then it will not call the `woocommerce_gla_product_attribute_value_sale_price` filter so it will not affect the changes that we did in these PR.

My question is more about what the expected data structure will be after a fix is applied. In the Linear issue description, it shows the following data shape when "priced individually" is disabled and the sync is working:

```json
{
  "bundle_price": {
    "price": {
      "min": {
        "incl_tax": "500",       // ← STRING without decimals
        "excl_tax": "500"        // ← STRING without decimals
      },
      "max": {
        "incl_tax": "500",       // ← SAME VALUE
        "excl_tax": "500"        // ← SAME VALUE
      }
    }
  },
  "bundled_items": [
    {
      "priced_individually": false, // ← DISABLED
      "optional": true
    }
  ],
  "sync_status": "synced"  // ← WORKING
}
```

You can see that the "max" prices are the same as the "min" prices. When it is not working, you see this shape:

```json
{
  "bundle_price": {
    "price": {
      "min": {
        "incl_tax": "500.00",    // ← STRING with decimals
        "excl_tax": "500.00"     // ← STRING with decimals
      },
      "max": {
        "incl_tax": "",          // ← EMPTY STRING
        "excl_tax": ""           // ← EMPTY STRING
      }
    }
  },
  "bundled_items": [
    {
      "priced_individually": true,  // ← ENABLED
      "optional": true
    }
  ],
  "sync_status": "pending"  // ← FAILING
}
```

The "max" prices are empty. I assume that casting these empty values as an integer would result in the value "0" being sent as the "max" prices (I could be mistaken about this, haven't tested). I want to make sure we confirm that not only will the sync work correctly, but that the values that get synced are also the expected values.

joemcgill:
I've done more testing and wasn't able to figure out exactly what the data types shared in the original ticket were referring to. However, the issue described that trying to sync a product bundle when the included products were priced individually resulted in a fatal because `WooCommerceProductBundles::calculate_price()` expects a float to be returned, but the relevant methods in the `WC_Product_Bundle` object that return prices are returning strings instead. 

This PR is fixing that issue, and I could confirm that the product was being synced to MC. I'm going to go ahead and mark this as approved, pending more feedback from the support team.

joemcgill (approved):

ankitguptaindia:
QA Status: ✅ Verified. 
----

### **Test Results:**

### **Before Fix** (Develop Branch): 
The reported issue under **GOOWOO-276** was consistently reproducible using **WooCommerce Product Bundles v8.5.1** with the **Google for Woo** plugin on the `develop` branch. The fatal error occurred during **Force Sync**. 

https://github.com/user-attachments/assets/926c6782-dffb-45bc-b5bd-246af0225ef9

### **After Fix** (`fix/GOOWOO-276-fatal-error` Branch): 
After applying the fixes from this PR, the issue has been resolved. The fatal error no longer appears during **Force Sync** or **Auto Sync**.

Testing was performed by creating a **Bundle Product** with the **“Priced Individually”** and **“Optional”** settings enabled. The product synced successfully in both Auto Sync and Force Sync (via the Test Connection page) without any errors or warnings.


For testing, I created both simple and bundle products. After creation, the **Channel Visibility** status was in a **“Pending”** state. After some time, when **Auto Sync** occurred, both products with the pending status appeared in the **Google Merchant Center** list.

On performing a **Force Sync**, the pending status disappeared from the product edit page, and the product synced successfully to the **Merchant Center**.

Screenshot 2025-10-09 at 12 10 05 PM

✅ **Result:** Verified — Fix working as expected.




https://github.com/user-attachments/assets/15e194ca-3f63-41d3-9f1a-937c5febb84e

ankitguptaindia (approved):

PR #3096, opened 2025-09-17
Fix conflict with budget recommendations transient

jjgrainger:

Closes # .

1. 
2. 
3. 

### Changelog entry

>

### DIFF
--- a/src/Options/TransientsInterface.php
+++ b/src/Options/TransientsInterface.php
@@ -10,30 +10,32 @@ namespace Automattic\WooCommerce\GoogleListingsAndAds\Options;
  */
 interface TransientsInterface {
 
-	public const ADS_BUDGET_METRICS   = 'ads_budget_metrics';
-	public const ADS_CAMPAIGN_COUNT   = 'ads_campaign_count';
-	public const ADS_LOCATION_IDS     = 'ads_location_ids';
-	public const ADS_METRICS          = 'ads_metrics';
-	public const ADS_RECOMMENDATIONS  = 'ads_recommendations';
-	public const FREE_LISTING_METRICS = 'free_listing_metrics';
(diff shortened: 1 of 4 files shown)


src/Options/TransientsInterface.php
```diff
@@ -10,30 +10,32 @@
...
-	public const MC_IS_SUBACCOUNT     = 'mc_is_subaccount';
-	public const MC_STATUSES          = 'mc_statuses';
-	public const URL_MATCHES          = 'url_matches';
-	public const WPCOM_API_STATUS     = 'wpcom_api_status';
+	public const ADS_BUDGET_METRICS         = 'ads_budget_metrics';
+	public const ADS_CAMPAIGN_COUNT         = 'ads_campaign_count';
+	public const ADS_LOCATION_IDS           = 'ads_location_ids';
+	public const ADS_METRICS                = 'ads_metrics';
+	public const ADS_BUDGET_RECOMMENDATIONS = 'ads_recommendations';
+	public const ADS_RECOMMENDATIONS        = 'ads_recommendations';
```
mukeshpanchal27: Both have same values does it make any difference?

mukeshpanchal27 (changes requested):

src/Options/TransientsInterface.php
jjgrainger: @mukeshpanchal27 good catch, I've updated this

mukeshpanchal27:
We should add a migration step to remove the `TransientsInterface::ADS_RECOMMENDATIONS` transients. Otherwise, the old `ADS_RECOMMENDATIONS` ( Now `ADS_BUDGET_RECOMMENDATIONS`) cache value may be passed into `ADS_RECOMMENDATIONS`, which could potentially create errors.

mukeshpanchal27 (approved):

PR #3097, opened 2025-09-17
Use `isGoogleAdsReady` property instead of `hasGoogleAdsConnection`

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-204/google-ads-card-disappears-when-account-creation-limit-is-reached

The only change in this PR is that when there are multiple accounts and unconnected Ads account, the card should not be in Edit mode. Rather the Ads account card is displayed where the user can select his Ads account which makes sense IMHO.

1. Go through the onboarding process and test different scenarios such as 
	* creating a new Ads account
	* connecting an existing account
	* claim an Ads account
	* auto Ads account creation if there is none

### Changelog entry

> Fix - Use `isGoogleAdsReady` property instead of `hasGoogleAdsConnection`


tests/e2e/specs/onboarding/step-1-accounts.test.js
```diff
@@ -1042,45 +1042,6 @@ test.describe( 'Set up accounts', () => {
 			} );
 		} );
 
-		test.describe( 'clicking "Edit" when an Ads account is being claimed', async () => {
```
joemcgill, line 1045: I'm curious about why these tests are being removed? These were put here to confirm the behavior after a new account is created but needs to have the invite accepted before we have access to make API calls against that account.

When that happens, `isGoogleAdsReady` will return false, due to this block:

https://github.com/woocommerce/google-listings-and-ads/blob/38405bffe8606df99882bbfe42661c2254d24335/js/src/hooks/useGoogleAdsAccountReady.js#L42-L46

I think that the behavior being tested by these E2E tests should still work whenever a new account has been created but not claimed. Can we confirm the behavior here?

asvinb:
Closing in favour of https://github.com/woocommerce/google-listings-and-ads/pull/3110

PR #3108, opened 2025-09-26
Fix floating point precision when converting to micro units

asvinb:
- Use round() before casting to int in to_micro() to prevent precision errors (e.g., 33.3 * 1,000,000)
- Ensures correct conversion and avoids NON_MULTIPLE_OF_MINIMUM_CURRENCY_UNIT errors with Google Ads API

Closes https://linear.app/a8c/issue/GOOWOO-291/error-editing-campaign-a-money-amount-was-not-a-multiple-of-a-minimum

1. Create a new campaign.
2. Set a custom amount of 33.30 or the following values: 99.99, 1234.56, 1.23 , 10.99
3. The campaign should be successfully saved.

### Changelog entry

> Fix - Prevent floating point precision errors when converting to micro units by rounding before casting to integer.

### DIFF
--- a/src/API/MicroTrait.php
+++ b/src/API/MicroTrait.php
@@ -22,7 +22,7 @@ trait MicroTrait {
 	/**
 	 * Convert to micro units.
 	 * We round to the nearest integer to avoid floating point precision issues.
-	 * For e.g 33.3, we want to get 33300000 and not 33299999 which can cause 
+	 * For e.g 33.3, we want to get 33300000 and not 33299999 which can cause
 	 * the Google Ads API to throw the NON_MULTIPLE_OF_MINIMUM_CURRENCY_UNIT error.
 	 *
 	 * @param float $num Number to convert to micro units.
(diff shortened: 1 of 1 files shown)


joemcgill (approved):
This works and better explains the issue we observed. Another way of thinking about this is that we want to make sure the `$num` value is rounded to 2 decimal places before multiplying by micros, so:

```php
$num = round( $num, 2 );
return (int) ($num * self::$micro);
```

What you have gets the same outcome and is a bit easier to read, so no changes are needed.

ankitguptaindia:
### QA Status: Verified ✅ 

### Test Result:  
Verified with both new and existing campaigns (including edits) and tested using various amount combinations. All campaigns are saving successfully without any errors, and the amount set by the user remains correctly saved in the campaign.




https://github.com/user-attachments/assets/bed9165d-3f3f-4ef1-a756-eb3e61643a69

ankitguptaindia (approved):

PR #3109, opened 2025-09-29
Add cached actioned campaigns.

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-284/after-selecting-a-recommended-budget-option-the-recommendation-badge

1. Navigate to the plugin dashboard.
2. Open (edit) a campaign that displays the “Raise Budget Recommendation” badge.
3. Select a recommendation level for the campaign.
4. Save the campaign changes.
5. Confirm that the “Raise Budget Recommendation” badge is no longer displayed for that campaign (this should remain hidden for 24 hours after the last edit).

### Changelog entry

>


js/src/components/paid-ads/asset-group/asset-group.js
```diff
@@ -108,14 +120,22 @@ export default function AssetGroup( { campaign } ) {
...
 
+	const recordActionedCampaign = () => {
+		if ( hasRaiseBudgetRecommendation() ) {
+			upsertActionedCampaign( campaign.id );
+		}
+	};
+
 	const handleSkipClick = ( event ) => {
 		handleSubmit( event );
+		recordActionedCampaign();
```
joemcgill, line 130: I don't think we want to dismiss the recommendation if the campaign editor is closed without changing the budget.

js/src/hooks/useActionedCampaignsCache.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+import { useState, useEffect, useCallback } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { LOCAL_STORAGE_KEYS, DAY_IN_SECONDS } from '~/constants';
+import localStorage from '~/utils/localStorage';
+
+const useActionedCampaignsCache = (
```
joemcgill: This looks like a good local storage cache solution but currently seems either too specific to actioned campaigns to be used as a flexible local storage cache solution or else more flexible than it needs to be to store actioned raise budget recommendations.

Creating a flexible, general use cache solution is probably out of scope for what we need to create here, so I think this could be reduced to meet the specific use case we have here and eliminate the need for supporting a `storageKey` and `ttl` option.

js/src/hooks/useActionedCampaignsCache.js
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * Inserts or updates a campaign entry in the local cache with a specified expiry.
+	 * If the campaign already exists, its expiry is updated; otherwise, a new entry is added.
+	 * The updated campaigns list is then saved to localStorage.
+	 *
+	 * @param {string} campaignId - The unique identifier of the campaign.
+	 */
+	function upsertActionedCampaign( campaignId ) {
+		const campaigns = getActionedCampaigns( storageKey );
+		const campaignIndex = campaigns.findIndex(
```
joemcgill: Would this be easier if the actioned campaigns were keyed by campaignId so we didn't have to find their index?

js/src/hooks/useActionedCampaignsCache.js
```diff
@@ -0,0 +1,85 @@
...
+		} else {
+			campaigns.push( { campaign: campaignId, expiry } );
+		}
+
+		localStorage.set( storageKey, JSON.stringify( campaigns ) );
+	}
+
+	useEffect( () => {
+		setCachedCampaigns( getActionedCampaigns() );
+	}, [ getActionedCampaigns ] );
```
joemcgill: Curious why we need to manage this list of action campaigns separately in state as well as in local storage.

joemcgill (commented):
@asvinb left a few comments/suggestions on this.

js/src/components/paid-ads/asset-group/asset-group.js
asvinb: @joemcgill Actually the skip button in this case is for adding assets to the campaign (step 2 of the campaign editing screen). On the first screen of the campaign editing, the user can still choose the budget amount and skip the second step and the campaign will be saved.

js/src/hooks/useActionedCampaignsCache.js
asvinb: Sure @joemcgill . Updated

js/src/hooks/useActionedCampaignsCache.js
asvinb: Thanks for the suggestion @joemcgill . Updated.

js/src/hooks/useActionedCampaignsCache.js
asvinb: I wanted to avoid querying local storage upon each render but i think that's fine. I've removed the hook and instead moved the functions as utility ones.

joemcgill (approved):
Thanks for making the updates. It looks much cleaner now 👍🏻

ankitguptaindia:
### QA Status: Needs Feedback ⚠️

@asvinb I tested this scenario using the **“Raise Budget Suggestion”** badge created via the **Tester plugin** on the `add/GOOWOO-284-actioned-campaign-badge` branch.

The issue observed is that the badge **still appears even after the merchant updates the campaign by accepting the suggestion.**

I’ve attached a video recording that demonstrates the complete testing flow and steps followed.
Could you please review and confirm if any updates are required for campaigns created through the Tester plugin?  _(Due to the long video, you can watch this at 2X speed )_


https://github.com/user-attachments/assets/3ef9db08-e1ff-43dd-b2d5-4adb8a5e22cd



### Additional Details: 

- All requests were executed successfully with a 200 status in the network
- No error in console log

ankitguptaindia:
### QA Status: Verified ✅ 



### Test Result:

- **Positive Use Case ✅**
Tested by creating new campaigns and editing existing ones. The *Raise Budget Recommendation* badge no longer appears once the user accepts the suggested budget — working as expected.

- **Negative Use Case ✅**
If the user edits the campaign without changing the budget or without accepting the suggested budget, the badge continues to appear — also working as expected.

- **QA Reported issue ✅**
Above [reported issue](https://github.com/woocommerce/google-listings-and-ads/pull/3109#issuecomment-3375805107) has been addressed and is working fine now. 

https://github.com/user-attachments/assets/c843bfea-bb38-4031-87f1-2ca1e9ee8445

ankitguptaindia (approved):

PR #3110, opened 2025-10-03
Use isGoogleAdsReady property instead of hasGoogleAdsConnection during onboarding

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-204/google-ads-card-disappears-when-account-creation-limit-is-reached

1. Go through the onboarding process and test different scenarios such as
	* creating a new Ads account
	* connecting an existing account
	* claim an Ads account
	* auto Ads account creation if there is none

### Changelog entry

> Fix - Use isGoogleAdsReady property instead of hasGoogleAdsConnection during the onboarding process.

### DIFF
--- a/js/src/components/ads-account-select-control/index.js
+++ b/js/src/components/ads-account-select-control/index.js
@@ -10,6 +10,7 @@ import { getSetting } from '@woocommerce/settings'; // eslint-disable-line impor
 import AppSelectControl from '~/components/app-select-control';
 import useExistingGoogleAdsAccounts from '~/hooks/useExistingGoogleAdsAccounts';
 import useGoogleAdsAccount from '~/hooks/useGoogleAdsAccount';
+import useGoogleAdsAccountReady from '~/hooks/useGoogleAdsAccountReady';
 
 /**
  * @param {Object} props The component props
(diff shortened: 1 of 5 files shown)


joemcgill (changes requested):
@asvinb apologies for taking so long to getting around to reviewing this. Now that I'm reviewing with fresh eyes, I'm unsure that the approach you've proposed across the PR is ideal, given what the `isGoogleAdsReady` and `hasGoogleAdsConnection` properties actually represent.

As I understand it from the reading the code, `hasGoogleAdsConnection` is meant to represent that a Google Ads account ID has been connected to the plugin. However, even when a connection is established, the setup might not yet be completed until later in the onboarding process ([see reference](https://github.com/woocommerce/google-listings-and-ads/blob/d23bdb504bce1ed8a10a4bd92608aeb5137fbe60/js/src/hooks/useGoogleAdsAccount.js#L41-L49)). In many cases, this is actually the appropriate property, because we may still need the user to accept the account linking invite, or set up billing.

`isGoogleAdsReady` on the other hand, is only true if a connection has been established, the account invite has been accepted (i.e., the user has access to make API requests against the account), and the account setup has completed or at least gotten past any of the optional setup steps (e.g., link_merchant, billing). [See ref](https://github.com/woocommerce/google-listings-and-ads/blob/d23bdb504bce1ed8a10a4bd92608aeb5137fbe60/js/src/hooks/useGoogleAdsAccountReady.js#L42-L46).

There are several places in this PR where I think the existing logic is actually the correct logic given where the component is used in the onboarding process.

Additionally, the scope of this fix seems much broader than what is described in GOOWOO-204 which makes it difficult to verify if this change is actually fixing that issue, or just hiding the root cause. To confirm, you would want to have testing instructions that show how to replicate the bug and then confirm that the bug no longer occurs when this fix is applied.

PR #3111, opened 2025-10-03
Update/goowoo 272 set custom path

ankitrox:
### Changes proposed in this Pull Request:

Closes: https://linear.app/a8c/issue/GOOWOO-272/set-up-a-custom-path-that-will-route-requests-to-gtg

_Replace this with a good description of your changes & reasoning._

### Detailed test instructions:

> [!Note]
> Before testing, make sure you have deleted the `googletaggateway_measurement_path` option from the DB so the measurement path is refreshed.

1. When GTG is disabled:
    - The GTM script source uses third-party e.g `https://www.googletagmanager.com/gtag/js?id=G-12345`.
    - The WooCommerce developer ID `dOGY3NW` is set.
2.  When GTG is enabled in the Google for WooCommerce settings page:
    - The GTM script source uses the custom first-party path saved to the `googletaggateway_measurement_path` option, ex:  `/s12hxjchs`.
    - The custom path `/${measurement_path}` returns a 200 status.
    - The GTG developer ID `dYmZiZj` is set alongside the WooCommerce developer ID.
    - Measurement requests are handled and recorded correctly.

### Additional details:
- There is an issue when using the adapter that the measurement path `/wc/google/metrics/` is added but permalinks are not flushed and updated. In order for the GTG script to work, the user must manually update permalinks by visiting the Settings > Permalinks admin page.

PR #3121, opened 2025-10-13
Update GTG Library

joemcgill:

Related to GOOWOO-272, GOOWOO-307.

This updates the GTG library to the latest hash from the fpm-library repo.

## Additional details

Instead of cleaning up some of the files that aren't needed, I think it's less maintenance to just copy the whole library code into the `/gtg` folder for now so pulling in additional updates is as simple as pulling changes from that repo locally and copying the whole directory over.


src/Google/GlobalSiteTag.php
```diff
@@ -113,14 +113,20 @@ public function register(): void {
...
 		if ( $this->is_gtg_enabled() ) {
-			$gtg_adapter = Adapter::create();
-			$gtg_adapter->update(
-				[
-					'tagId'           => $conversion_action['conversion_id'],
-					'measurementPath' => '/wc/google/metrics/',
-				]
+			add_action(
+				'init',
+				function () use ( $conversion_action ) {
```
mukeshpanchal27: As the plugin didn't follow it but in new improvement we should use `static` for clauses. WDYT?

```suggestion
				static function () use ( $conversion_action ) {
```

mukeshpanchal27:
> Instead of cleaning up some of the files that aren't needed, I think it's less maintenance to just copy the whole library code into the `/gtg` folder for now so pulling in additional updates is as simple as pulling changes from that repo locally and copying the whole directory over.

Do we have any timeline for when it will be released as a Composer package? If it’s going to take some time, it might be worth adding these setup steps to the [docs folder](https://github.com/woocommerce/google-listings-and-ads/tree/develop/docs) for now. It could also be helpful to include this in our internal onboarding documentation so new contributors can easily understand the process.

src/Google/GlobalSiteTag.php
joemcgill: Possibly so! This PR didn't mean to include these changes, so this could be reverted if @jjgrainger and @ankitrox want to use this PR, otherwise, feel free to close this PR if you are going to handle the upgrade directly in #3111.

jjgrainger: @joemcgill @mukeshpanchal27 I've updated the PR to revert the changes here as these were made to test and share potential solution and fall outside of the scope of the ticket (updating the library files).

I will mark this as resolve as we will tackle this as part of a different PR. Thanks

jjgrainger (approved):

PR #3131, opened 2025-10-29
Feature/287 Implement feature-flag process

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-287/implement-feature-flag-process-to-manage-the-default-setting-for-newly

1. 
2. 
3. 

### Changelog entry

>


joemcgill (commented):
@jjgrainger this looks quite a bit more robust than what I was expecting for this experiment, but I think that's great!

I think creating a Features service to manage these flags and exposing them in a specialized REST API endpoint are a great idea.

I don't think that we necessarily need a scheduled job to update these flags. Instead, I think when the service is registered if could call a method to load the features. That method could send a request to the features endpoint of the WCS and save the response in a transient for something like 10 minutes to avoid the need to make a request each time.

The other question that I had about the PR is how you were thinking the percentage value to be used for affecting how each site would interpret the feature flags? My assumption would have been that this would have to be something implemented by the WCS to affect which sites get "enabled" responses vs. which ones do not.

src/Options/Features.php
```diff
@@ -0,0 +1,113 @@
...
+	 *
+	 * @param string $feature
+	 * @return integer
+	 */
+	protected function get_site_percentage_rollout( string $feature ): int {
+		// Get the unique ID for the site.
+		$site_id = get_home_url();
+
+		// Create a numeric hash using the site ID and feature name.
+		$hash = crc32( $site_id . ':' . $feature );
```
joemcgill: `crc32` can return negative numbers, so this needs to be wrapped in `abs()`.
```suggestion
		$hash = abs( crc32( $site_id . ':' . $feature ) );
```

src/Options/Features.php
```diff
@@ -0,0 +1,113 @@
...
+				$option['features'][ $name ]['enabled'] = true;
+			}
+		}
+
+		// Update the option in the database.
+		$this->options->update( $this->option_name(), $option );
+	}
+
+	/**
+	 * Return the site percentage rollout for a specific feature.
```
joemcgill, line 139: Since this is a non-intuitive way of implementing this requirement, it would be useful to add a more robust description in the docblocks here for future maintenance.

It would also be good to have some unit test coverage on this method that confirms that it meets the requirement. Here's an example I quickly prototyped with the help of AI but would need to be adjusted to actually make use of the method:

```php
<?php
public function test_distribution_is_uniform() {
    $buckets = array_fill(0, 100, 0);
    
    // Simulate 10,000 different sites
    for ($i = 0; $i < 10000; $i++) {
        $site_url = "https://example{$i}.com";
        $hash = crc32($site_url . ':test_feature');
        $bucket = abs($hash) % 100;
        $buckets[$bucket]++;
    }
    
    // Each bucket should have ~100 sites (10000/100)
    // Allow 20% variance as acceptable
    foreach ($buckets as $count) {
        $this->assertGreaterThan(80, $count);
        $this->assertLessThan(120, $count);
    }
}
```

src/API/WCS/ConnectionService.php
```diff
@@ -0,0 +1,59 @@
...
+		// TODO: Replace hardcoded values with request to external API.
+		return [
+			'version'  => gmdate( 'c' ),
+			'features' => [
+				'google_tag_gateway' => [
+					'enabled'    => true,
+					'default'    => true,
+					'percentage' => 100,
+				],
+			],
```
joemcgill: Even after waiting for this to be implemented by WCS, it would be good to have this wrapped in a filter so we can more easily build test cases for this.

src/API/Site/Controllers/FeaturesController.php
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	/**
+	 * Callback function for returning the tours
+	 *
+	 * @return callable
+	 */
+	protected function get_features_read_callback(): callable {
+		return function () {
+			return $this->options->get( OptionsInterface::WCS_FEATURE_FLAGS, [] );
```
joemcgill: Rather than checking an option, I'd suggest implementing a getter on the Features service so you can decouple the storage mechanism from the data retrieval. Something like:

```
return $this->features->get_features();
```

A nice to have addition to this is that if the endpoint supported passing a specific feature flag, you could call `$this->features->get_features( [$feature] );` to return an object for only that one feature. You just need to ensure that the processing that is currently happening in `Features:update()` is still being applied to either after the WCS request is returned, or as part of the getter before the raw feature flag values are returned here.

src/API/WCS/ConnectionService.php
```diff
@@ -0,0 +1,59 @@
...
+	/**
+	 * Update the feature flags option in the database.
+	 *
+	 * @return void
+	 */
+	public function update_feature_flags(): void {
+		$feature_flags = $this->get_features();
+
+		$this->features->update( $feature_flags );
+	}
```
joemcgill: If this is even needed, I think it would be better to make this the responsibility of the `Features` service rather than the `ConnectionService`.

joemcgill (changes requested):
After discussing the overall approach, the implementation decision to process the percentage rollout using a hashing/bucketing approach is clever and should work well enough provided the couple of suggestions in the CR are applied.

I still think it would be best to avoid using a background job for refreshing the features and saving them to an option and instead check the feature flags at runtime and cache them in a transient that expires every 5 minutes (or another appropriate TTL value) to limit an API request being made to WCS on every page load.

I understand that this will result in a slight lag between when Woo changes the values of the API response and when sites will see those changes apply, but I think that's a reasonable expectation.

Let me know if anything isn't clear or if I've missed anything important here.

src/Options/Features.php
jjgrainger: @joemcgill I've included a test however I found that there is a ~25% variance with this method. Not sure if this acceptable, but as we don't plan to use the percentage rollout in the immediate future, we may want to revisit this to see if theres an alternative approach that brings a more uniform distribution.

src/API/WCS/ConnectionService.php
```diff
@@ -0,0 +1,37 @@
...
+			'features' => [
+				'google_tag_gateway' => [
+					'enabled'    => true,
+					'default'    => true,
+					'percentage' => 100,
+				],
+			],
+		];
+
+		return apply_filters( 'woocommerce_gla_wcs_feature_flags', $features );
```
joemcgill, line 35: 👍🏻

tests/Unit/Options/FeaturesTest.php
```diff
@@ -0,0 +1,153 @@
...
+
+		// Simulate 10,000 different sites
+		for ( $i = 0; $i < 10000; $i++ ) {
+			$site_id = "https://example{$i}.com";
+			$bucket  = $this->features->get_site_percentage_rollout( 'test_feature', $site_id );
+
+			++$buckets[ $bucket ];
+		}
+
+		// Expect roughly 100 per bucket, allow 25% variance
```
joemcgill, line 147: 👍🏻 This is good for now, but we can confirm with Google whether this is close enough for their preference.

joemcgill (approved):
Thanks, @jjgrainger! These latest updates look good to me.

src/Options/Features.php
joemcgill: Thanks for flagging this. I think the 25% variance is ok for now. We can flag this variance with Google as part of our testing plan

PR #3132, opened 2025-10-30
Add a message to product inventory tab to set availability date if ba…

ankitrox:
…ck order enabled.

Closes: https://linear.app/a8c/issue/GOOWOO-38/add-a-message-to-product-inventory-tab-to-set-availability-date-if

### Screenshots:

Screenshot 2025-10-31 at 12 11 22 PM

### Detailed test instructions

1. Go to Products > Add new product.

2. In the `Inventory` tab, set the product to `In stock` and allow backorders, the notice should be shown as can be seen in screenshot.

3. Do the same steps, but set product to `Allow, but notify customer` or `Allow` in `Allow backorders?` options and notice should be shown in `Inventory` tab.

4. Do the same for variable products, when coming back to inventory tab, the notice should be visible.

#### Caveat:
When there is an existing variable product having atleast one of the variation which has either backorders allowed when product in stock or stock status set to `onbackorder`, the notice is not shown immediately when product edit screen is loaded. The notice only appears when I visit the Variations tab and revisit the Inventory tab, then only notice appears. This is because the variations' markup is not loaded until the variation tab is visited.

### Changelog entry

>


google-listings-and-ads.php
```diff
@@ -85,6 +85,19 @@ function () {
 	}
 );
 
+// Load plugin text domain at the recommended action timing.
```
mukeshpanchal27, line 88: This should be unrelated to this PR. If any changes are needed, we should open a separate issue and PR for them.

src/Admin/Admin.php
```diff
@@ -103,6 +103,24 @@ function ( $scripts ) {
 		);
 
 		add_action( 'admin_init', [ $this, 'privacy_policy' ] );
+		add_action(
+			'woocommerce_product_options_inventory_product_data',
+			function () {
+				$notice_id  = 'gla-backorder-availability-notice';
+				$classes    = 'inline woocommerce-message show_if_simple show_if_variable';
```
mukeshpanchal27, line 110: Could we use default notice per https://developer.woocommerce.com/docs/extensions/ux-guidelines-extensions/notices/ here?

js/src/product-edit/index.scss
mukeshpanchal27, line 1: We can remove it if we leverage the default notice, as mentioned in my other comment.

mukeshpanchal27 (changes requested):
Thanks for the PR. Please rebase the branch and review my notes below. Let me know if you have any questions.

asvinb:
Closing in favour of https://github.com/woocommerce/google-listings-and-ads/pull/3261

PR #3136, opened 2025-11-04
Update data store to account for YouTube account

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-326/update-data-store-to-include-yt-shopping-properties

1. Go to the settings page.
2. Run the following command in the console: `wp.data.select('wc/gla').getYouTubeAccount`
3. A network request should be made to the `youtube/connection` endpoint. The endpoint might not be available right now and is dependent on GOOWOO-332

### Changelog entry

>


ankitrox (approved):
LGTM. 
@asvinb Can we add the test instructions where we can get mock response using override so that we can test the response is read correctly and state is getting set fine? We can use tool like `Tweak` extension to easily achieve the request override.

PR #3138, opened 2025-11-05
Fix rewrite rules registration

joemcgill:

Closes GOOWOO-317.

This updates the implementation so that it only runs Adapter::update() when the GTG feature is enabled, but calling it on the update_option callback. This also fixes a bug that kept the rewrite rules from being written because the save_mod_rewrite_rules function was not loaded at the time this process ran.


jjgrainger (approved):

joemcgill:
@jjgrainger – I intended to add another commit to this PR that would reset the rewrite rules when GTG is disabled, but turns out there is not an easy way to implement this, so I'm going to go ahead and merge this as is and will provide some feedback to Google about the need for a helper in the Adapter that will remove the rewrite rules.

PR #3140, opened 2025-11-06
Improve shipping speeds UI with default values and better labelling.

ankitrox:

Closes: https://linear.app/a8c/issue/GOOWOO-202/improve-shipping-speeds-ui-with-default-values-and-better-labelling

### Screenshots:

Screenshot 2025-11-07 at 1 11 48 PM

Setup the  plugin. During the onboarding setup second step `Configure Product Listings`, notice that the minimum and maximum is set to `1` and `5` respectively.

### Changelog entry

> Update - Add default values for shipping time inputs during the onboarding process.


tests/e2e/specs/onboarding/step-2-product-listings.test.js
```diff
@@ -220,11 +220,14 @@ test.describe( 'Configure product listings', () => {
 			await expect( minimumOrderForFreeShippingText ).toBeVisible();
 		} );
 
-		test( 'should show error message if min or max shipping time is not set', async () => {
-			await productListingsPage.clickContinueButton();
-			const estimatedTimesError =
-				productListingsPage.getEstimatedShippingTimesNullError();
-			await expect( estimatedTimesError ).toBeVisible();
+		test( 'should have minTime as 1 and maxTime as 5 set by default', async () => {
```
asvinb, line 226: @ankitrox Can you add another test when the shipping rates are already set and check that the default values are not rendered please? For e.g you can mock the `/wc/gla/mc/shipping/times` endpoint to return the shipping times set. The sample JSON below is for Mauritius:
```json
{"MU":{"country_code":"MU","time":"1","max_time":"5","country":"Mauritius"}}
```

asvinb (changes requested):
@ankitrox I found an issue which we need to take into account with the proposed solution. Steps to reproduce:

- Go to the second step of the onboarding process
- Add a country only
- Notice how the default shipping times values appear
- Remove the country (leaving no countries selected)
- Notice how the `useEffect` hook fires endlessly, making requests to the `/wc/gla/mc/shipping/times/batch` endpoint.

If there are no selected countries, then those requests should not be made.

asvinb (changes requested):
@ankitrox I had another look at how we can simply things. I think we can do the following:
* Instead of relying on the `onBlur` event to update the value if time is not set, it's better to set the time values in the `initialValues` of the form.
* In `AddTimeModal`, we can update the `time` and `maxTime` values to `1` and `5` respectively. This is for the add time modal which is accounted for in the current PR.
* Then in `ShippingCountriesForm`, we are setting the default values of `time` and `maxTime` to `null`. Let's update those to `1` and `5` respectively.

I think this should be enough to have the default values. Can you check and let me know please? Thanks!

ankitrox:
@asvinb Thank you for the suggestion, I've made the changes and assigning this to you for another review. I've tested it and added tests accordingly. All seems to be working as expected.

asvinb (approved):
LGTM @ankitrox . If you can please add the changelog entry and then send to QA.
Thanks!

ankitguptaindia:
QA Status: Approved ✅ 
---


Test Result: Working as scoped. 

Screenshot 2026-05-11 at 9 50 58 PM

PR #3152, opened 2025-11-24
Encode the site URL when calling the Google SDI proxy to support a URL with paths

eason9487:

Closes GOOWOO-295

📌 Please review this PR along with https://github.com/Automattic/woocommerce-connect-server/pull/3029

Some merchants want to use a URL with path(s) as their store homepage by using a custom code snippet. However, when calling the Google SDI proxy during the onboarding flow, the URL was not encoded, resulting in a 404 error that prevented completing the onboarding.

This PR encodes the site URL when calling the Google SDI proxy to support a URL with paths.

1. Checkout [Automattic/woocommerce-connect-server#3029](https://github.com/Automattic/woocommerce-connect-server/pull/3029) and set up the environment according to that PR
2. Add a code snippet to the plugin side to use a site URL with path(s):
   ```php
   add_filter( 'woocommerce_gla_site_url', function ( $home_url ) {
   	return wc_get_page_permalink( 'shop' );
   
   	// Or return the string directly and explicitly if the above
   	// cannot return the expected store URL.
   	//
   	// return 'https://your.publicly.accessible.site/shop/';
   } );
   ```
3. Make the test site publicly accessible
4. Go to Step 1 of the onboarding flow
5. Test if the following Not Found error won't happen when connecting to a Google Merchant Center account
   image

### Additional details:

- Calling that Google SDI API endpoint is an initialization part of the API Pull mode.
- Another similar Google SDI endpoint generation method was not applied `urlencode`  because it is no longer used, but has not yet been removed from the codebase.
   https://github.com/woocommerce/google-listings-and-ads/blob/2be77072afbbbef47d6de658f8791733016c88ac/src/API/Google/Middleware.php#L496-L502

### Changelog entry

> Tweak - Make it possible to complete the onboarding flow using a site URL with paths.

### DIFF
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -516,9 +516,17 @@ class Middleware implements ContainerAwareInterface, OptionsAwareInterface {
 	 * @return string
 	 */
 	protected function get_sdi_endpoint(): string {
+		/**
+		 * Reason for using `urlencode` to encode the site URL:
+		 *
+		 * This plugin doesn't natively support linking a Google Merchant
+		 * Center account and the Google Shopping Data Integration API using a
+		 * merchant URL with paths. However, some merchants still use the
+		 * plugin's WP filter `woocommerce_gla_site_url` to achieve this.
+		*/
 		return $this->container->get( 'connect_server_root' )
 			. 'google/google-sdi/v1/credentials/partners/WOO_COMMERCE/merchants/'
-			. $this->strip_url_protocol( $this->get_site_url() )
+			. urlencode( $this->strip_url_protocol( $this->get_site_url() ) )
 			. '/';
 	}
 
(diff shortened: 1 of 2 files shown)


src/API/Google/Middleware.php
```diff
@@ -516,9 +516,17 @@ public function get_sdi_merchant_update_endpoint(): string {
...
 	 */
 	protected function get_sdi_endpoint(): string {
+		/**
+		 * Reason for using `urlencode` to encode the site URL:
+		 *
+		 * This plugin doesn't natively support linking a Google Merchant
+		 * Center account and the Google Shopping Data Integration API using a
+		 * merchant URL with paths. However, some merchants still use the
+		 * plugin's WP filter `woocommerce_gla_site_url` to achieve this.
+		*/
```
costasovo: It would be great to add a short explanation of why urlencode helps to solve the problem. From the tests and additional context, I understand that it prevents / in the value, which seems to be causing trouble.

costasovo (approved):
Looks good! I suggest updating the in-code comment and adding a reason why we need to encode the site URL. Please see my other comment.

I confirm I was able to connect to the account with the shop URL configured with /shop path and I didn't see the error.
Screenshot 2025-11-26 at 15 14 49

src/API/Google/Middleware.php
eason9487: Thanks for the suggestion! Added in https://github.com/woocommerce/google-listings-and-ads/pull/3152/commits/b329893bf880527a1ffca56406383fb038c107fa.

PR #3154, opened 2025-11-24
Use Feature Flags when determining if GTG is enabled

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-287/implement-feature-flag-process-to-manage-the-default-setting-for-newly

Use the `woocommerce_gla_wcs_feature_flags` filter to modify the WCS feature flag response to simulate different scenarios:

#### Disable the feature
Use the following to disable the feature entirely, regardless if the user has enabled GTG in the settings dashboard.

```php
/**
 * Disable the feature.
 */
add_filter( 'woocommerce_gla_wcs_feature_flags', function() {
	return [
		'version'  => gmdate( 'c' ),
		'features' => [
			'google_tag_gateway' => [
				'enabled'    => false, // Disable the feature.
				'default'    => true,
				'percentage' => 100,
			],
		],
	];
});
```

#### Do not opt in newly onboarded users
When a user completes onboarding the feature flag default value is used to opt the user in/out.

```php
/**
 * Do not opt in newly onboarded users.
 */
add_filter( 'woocommerce_gla_wcs_feature_flags', function() {
	return [
		'version'  => gmdate( 'c' ),
		'features' => [
			'google_tag_gateway' => [
				'enabled'    => true, 
				'default'    => false, // Will not opt in new users after onboarding.
				'percentage' => 100,
			],
		],
	];
});
```
#### Site percentage rollout
A lower percentage will lead to fewer sites seeing the GTG feature enabled.

```php
/**
 * Only enable for 10% of sites.
 */
add_filter( 'woocommerce_gla_wcs_feature_flags', function() {
	return [
		'version'  => gmdate( 'c' ),
		'features' => [
			'google_tag_gateway' => [
				'enabled'    => true, 
				'default'    => true,
				'percentage' => 10, // Only 10% of users will have this feature enabled.
			],
		],
	];
});
```

### Changelog entry

>


joemcgill (commented):
Thanks @jjgrainger! This is looking pretty good. I think there still needs to be an integration with the `AdsSettingsController` so that the settings returned to the `/gla/ads/settings` endpoint will be affected by the feature flags, unless I'm mistaken. 

Otherwise, any UI that is reading those settings will still be able to affect read and write to those settings and cause side effects like calling the GTG Adapter::update method even when the feature is disabled, which seems like something we would want to avoid.

src/Google/GlobalSiteTag.php
```diff
@@ -140,6 +140,11 @@ public function register(): void {
 			'update_option_' . $gtag_option_name,
 			function ( $old_value, $value ) use ( $ads_conversion_id ) {
 				if ( $value ) {
+					// Allow the user option value to be updated but do not enable GTG if feature is disabled.
+					if ( ! $this->features->is_enabled( Features::GOOGLE_TAG_GATEWAY ) ) {
```
joemcgill: I like the intent of this, but am worried that this could lead to a situation where someone updates the option value when the feature is disabled, and then when the feature is later enabled, the Adapter::update() method is never called.

I think it would be better to always call the update method when the value changes and consider updating the UI in the settings screen to handle the situation where the feature is disabled.

joemcgill (changes requested):

src/API/WCS/ConnectionService.php
```diff
@@ -12,26 +17,38 @@
...
-					'percentage' => 100,
-				],
-			],
-		];
-
-		return apply_filters( 'woocommerce_gla_wcs_feature_flags', $features );
+		try {
+			/** @var Client $client */
+			$client   = $this->container->get( Client::class );
+			$result   = $client->get( $this->get_connection_url() );
```
mukeshpanchal27: Should check `200 === $result->getStatusCode()` before process further. See https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/API/Google/Connection.php#L107-L130

src/Google/GlobalSiteTag.php
```diff
@@ -758,6 +767,12 @@ private function normalize_and_hash( $value, $algo = 'sha256' ): string {
 	 * @return bool
 	 */
 	private function is_gtg_enabled(): bool {
-		return (bool) $this->options->get( OptionsInterface::ADS_GTG_ENABLED );
+		// Disable GTG if the feature flag has marked it as disabled.
+		if ( ! $this->features->is_enabled( Features::GOOGLE_TAG_GATEWAY ) ) {
```
mukeshpanchal27, line 771: This will call WCS every time the website loads, even when using the cached version. Since the transient expires every 5 minutes, is it okay to send requests this frequently? High-traffic sites may end up making many calls to WCS.

```
$this->transients->set( $this->transient_name(), $option, MINUTE_IN_SECONDS * 5 );
```

mukeshpanchal27 (commented):
Thanks for the PR. Left some questions.

src/Google/GlobalSiteTag.php
jjgrainger: Thanks @joemcgill I've reverted this change and we have a new ticket to hide the UI when the GTG feature is disabled via WCS

src/API/WCS/ConnectionService.php
jjgrainger: Thanks @mukeshpanchal27 good spot and will update this.

src/Google/GlobalSiteTag.php
jjgrainger: @mukeshpanchal27 there is a requirement where WCS may disable a feature due for all sites if there is an issue. Originally, there was a job that cached the response for 1 hour but was decided this was too long.

It was decided that 5 minutes was a reasonable update time for this requirement so should keep this as is for now

src/Options/Features.php
```diff
@@ -75,23 +76,44 @@ public function get_features( string $feature = '' ): array {
...
 	/**
 	 * Update the feature flags option from the WCS API.
 	 *
 	 * @return array
 	 */
 	protected function update_features() {
-		// Get the latest feature flags from WCS.
-		$features = $this->connection->get_features();
+		try {
+			// Get the latest feature flags from WCS.
```
ankitrox: I think we can remove this comment.

src/Options/Features.php
```diff
@@ -75,23 +76,44 @@ public function get_features( string $feature = '' ): array {
...
+			$features = $this->connection->get_features();
+		} catch ( Exception $e ) {
+			// If WCS API fails, return an empty array.
```
ankitrox: This too

src/Options/Features.php
```diff
@@ -75,23 +76,44 @@ public function get_features( string $feature = '' ): array {
...
+	 * @return mixed
+	 */
+	public function default_value( string $feature ) {
+		$config = $this->get_features( $feature );
+
+		if ( isset( $config['attributes']['default'] ) ) {
+			return $config['attributes']['default'];
+		}
+
+		return null;
```
ankitrox:
```suggestion
return $config['attributes']['default'] ?? null;
```

Null coalescing operator can be used here.

src/Options/Features.php
```diff
@@ -75,23 +76,44 @@ public function get_features( string $feature = '' ): array {
...
-		$option = $this->get_features();
+		$config = $this->get_features( $feature );
 
-		if ( isset( $option['features'][ $feature ]['enabled'] ) ) {
-			return $option['features'][ $feature ]['enabled'];
+		if ( isset( $config['enabled'] ) ) {
+			return $config['enabled'];
 		}
 
 		return false;
```
ankitrox:
```suggestion
			return $config['enabled'] ?? false;
}

return false;
```

ankitrox (changes requested):

src/Options/Features.php
jamesmorrison: This has been removed

src/Options/Features.php
jamesmorrison: This has been removed

src/Options/Features.php
jamesmorrison: This has been updated

src/Options/Features.php
jamesmorrison: This has been updated

src/API/WCS/ConnectionService.php
jamesmorrison: This has been updated

src/API/WCS/ConnectionService.php
```diff
@@ -12,26 +17,43 @@
...
+		try {
+			/** @var Client $client */
+			$client = $this->container->get( Client::class );
+			$result = $client->get( $this->get_connection_url() );
+
+			if ( 200 !== $result->getStatusCode() ) {
+				throw new Exception( __( 'Unable to connect to the feature flags API', 'google-listings-and-ads' ) );
+			}
+
+			$response = json_decode( $result->getBody()->getContents(), true );
```
mukeshpanchal27: `json_decode` can return `null` on invalid JSON. Should check that method declares array return.

ankitrox (approved):
This looks good to me, moving to QA.

joemcgill:
Closing this PR since the GTG project has been put on hold

PR #3159, opened 2025-11-26
[GOOWOO-333] Create the CSV Export functionality and Merchant Conversion Report CSV Service

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-333/create-the-csv-export-functionality-and-merchant-conversion-report-csv

1. Make sure the Order Attribution setting is enabled in WooCommerce > Settings > Advanced > Features
2. Create 2 orders via a YouTube link by visiting the test using the the correct UTM query parameters. For example `https://www.example.com/product/item?utm_source=youtube&utm_content=YT3-test-attribution`.
3. Refund one of the orders in the WooCommerce admin. When refunding the order, make sure to enter the quantities in the line items so the refund is processed correctly.
4. Use the `woocommerce_gla_youtube_order_ids_job_date` filter to return the date matching when both orders were created. The date returned should be a string following the format `Y-m-d`, e.g `2025-12-01`.
5. Schedule and run the job `CreateYouTubeOrderIdsCache`. This will capture the order IDs needed for the export and save them to the options table under the key `youtube_export_order_ids`. Once the job is complete, the `CreateMerchantReportedConversionReport` job will run immediately after.
6. The `CreateMerchantReportedConversionReport` job will create the CSV which can be found in the `/wp-content/uploads/gla-exports/youtube-merchant-conversion-report-{date}.csv` where is the same date used in the filter (eg `2025-12-01`).

Testing should confirm.
- The correct test order IDs are saved to the `youtube_export_order_ids` option.
- The correct line items are added to the `youtube-merchant-conversion-report-{date}.csv`
- The line items contain the correct information based on the Merchant Reported Conversions Schema.

### Testing WP CLI command:

1. Run the CLI command: `wp gla generate-sample-conversion-csv`
2. Confirm the CSV was created and stored in wp-content/uploads/gle-exports/sample-conversion-report-{date_time}.csv`
3. Run the CLI command with options:
`--rows` e.g. `--rows=1000` (default 5000)
`--refund-rate` e.g. `--refund-rate=15` the percentage of transactions that include a refund (default 10)
`--date-range` e.g. `--date-range=60` the number of days back to randomise the transaction dates (default 30)

### Changelog entry

>


src/Admin/Exports/RowBuilder/OrderItemRowBuilder.php
```diff
@@ -0,0 +1,317 @@
...
+			 * This value is always positive, even in the case of refunds.
+			 *
+			 * Example: 1000.00
+			 *
+			 * @param float
+			 */
+			'item_unit_price'            => absint( $item->get_subtotal() ) / $quantity,
+
+			/**
+			 * Item unit disconted price.
```
joemcgill:
```suggestion
			 * Item unit discounted price.
```

src/Admin/Exports/Services/YouTubeOrders.php
```diff
@@ -0,0 +1,104 @@
...
+			'offset'       => $offset,
+			'return'       => 'ids',
+			'type'         => 'shop_order',
+			'meta_query'   => [
+				[
+					'key'     => '_wc_order_attribution_utm_source',
+					'value'   => 'youtube',
+					'compare' => '=',
+				],
+			],
```
joemcgill: It's worth checking whether this meta_query creates a full table scan of the post_meta table so this doesn't cause a performance bottleneck on large stores. Otherwise, looping through the orders and checking the meta value could be better since this query should cause meta associated with these IDs to be cached.

src/Admin/Exports/Writer/CsvExportWriter.php
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		$dir_path = trailingslashit( $upload_dir['basedir'] ) . self::EXPORT_FOLDER;
+
+		if ( ! $this->fs->is_dir( $dir_path ) ) {
+			wp_mkdir_p( $dir_path );
+
+			if ( ! $this->fs->is_dir( $dir_path ) ) {
+				// phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
+				throw new RuntimeException( 'Failed to create export directory: ' . $dir_path );
```
joemcgill: This one likely should be escaped since the `$dir_path` could be a filtered value. Low risk, but still...

```suggestion
				throw new RuntimeException( sprintf( 'Failed to create export directory: %s', esc_html( $dir_path ) );
```

src/Admin/Exports/Writer/CsvExportWriter.php
```diff
@@ -0,0 +1,151 @@
...
+		// Don't create file if already exists.
+		if ( $this->fs->exists( $file ) ) {
+			return $file;
+		}
+
+		$success = $this->fs->put_contents( $file, '' );
+
+		if ( ! $success ) {
+			// phpcs:ignore WordPress.Security.EscapeOutput.ExceptionNotEscaped
+			throw new RuntimeException( 'Failed to create CSV file: ' . $file );
```
joemcgill: Same escaping suggestion here.

src/Internal/DependencyManagement/JobServiceProvider.php
```diff
@@ -5,11 +5,12 @@
 
 use ActionScheduler as ActionSchedulerCore;
 use ActionScheduler_AsyncRequest_QueueRunner as QueueRunnerAsyncRequest;
-use Automattic\WooCommerce\GoogleListingsAndAds\Ads\AdsRecommendationsService;
-use Automattic\WooCommerce\GoogleListingsAndAds\Ads\AccountService;
```
joemcgill, line 9: Nice cleanup 👍🏻

src/Jobs/CreateMerchantReportedConversionReport.php
```diff
@@ -0,0 +1,148 @@
...
+	 *
+	 * @return string
+	 */
+	protected function get_date(): string {
+		/**
+		 * Filters the YouTube orders query date value.
+		 *
+		 * @param string Date string formatted YYYY-MM-DD
+		 */
+		return apply_filters( 'woocommerce_gla_youtube_order_ids_job_date', gmdate( 'Y-m-d', strtotime( '-1 day' ) ) );
```
joemcgill, line 104: Curious what the use case is for this filter?

tests/Unit/Admin/Exports/RowBuilder/OrderItemRowBuilderTest.php
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * Class OrderItemRowBuilderTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Admin\Exports\RowBuilder
+ */
+class OrderItemRowBuilderTest extends UnitTest {
+	/** @var OrderItemRowBuilder $builder */
+	protected $builder;
+
+	public function setUp(): void {
```
joemcgill: Nit: `setUp()` methods run before every test, while this might only need to be set up once, in which case `setUpBeforeClass()` might be a better choice.

joemcgill (commented):
Left a few questions and comments, but no major concerns with this approach. It would probably be a good idea to add some unit test coverage to the `CsvExportWriter` class to ensure the files are being created as expected.

src/Admin/Exports/Services/YouTubeOrders.php
jjgrainger: Thanks @joemcgill I will explore this as we could potentially change this to one query and then check the meta against the order IDs for both purchases and the refund order parent IDs.

src/Jobs/CreateMerchantReportedConversionReport.php
jjgrainger: @joemcgill This is only really created for testing purposes so the job runs for a specific date. I will continue to explore a better way to test this and see if it can be removed.

src/Admin/Exports/Services/YouTubeOrders.php
```diff
@@ -0,0 +1,121 @@
...
+	 * @return array
+	 */
+	public function find_orders( string $date = '', int $limit = -1, int $offset = 0 ): array {
+		// Use yesterdays date if no date passed.
+		$date = empty( $date ) ? gmdate( 'Y-m-d', strtotime( '-1 day' ) ) : $date;
+
+		// Fetch all orders and refunds for the date without meta filtering.
+		// We'll filter by meta during the loop to leverage meta caching.
+		$purchase_ids = $this->find_purchases( $date, $limit, $offset );
+		$refund_ids   = $this->find_refunds( $date, $limit, $offset );
```
jjgrainger: @jamesmorrison with the recent changes I think we can now combine this into a single method `find_orders` and use a single query like so:

```php
// Use yesterdays date if no date passed.
$date = empty( $date ) ? gmdate( 'Y-m-d', strtotime( '-1 day' ) ) : $date;

$query = [
	'date_created' => $date,
	'limit'        => $limit,
	'offset'       => $offset,
	'return'       => 'ids',
	'type'         => [ 'shop_order', 'shop_order_refund' ],
];

// Get all refunds for the specific day.
// Meta filtering will be done in find_orders() to leverage meta caching.
return wc_get_orders( $query );
```

Then we can loop over each order to check the attribution meta data. If the the order is a refund we check the parent order otherwise check the current order object.

Thanks

src/Admin/Exports/Services/YouTubeOrders.php
```diff
@@ -0,0 +1,121 @@
...
+
+		// Filter orders and refunds by YouTube attribution meta.
+		$filtered_order_ids = [];
+		$refund_parent_ids  = [];
+
+		// Filter refunds first to collect parent order IDs
+		foreach ( $refund_ids as $refund_id ) {
+			$refund = wc_get_order( $refund_id );
+			if ( $refund && 'shop_order_refund' === $refund->get_type() ) {
```
jjgrainger: @jamesmorrison when checking if an order is a refund we may want to check the `$order` instance instead of `get_type()` like we do in the `OrderItemRowBuilder`

`if ( $order instanceof WC_Order_Refund )`

jamesmorrison: Thanks, I've updated this.

src/Admin/Exports/Services/YouTubeOrders.php
jamesmorrison: Thanks, I've updated this.

src/Coupon/WCCouponAdapter.php
```diff
@@ -383,9 +383,11 @@ public function get_wc_coupon_id(): int {
 	}
 
 	/**
+	 * Set the target country.
 	 *
-	 * @param string $targetCountry
-     *            phpcs:disable WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+	 * @param string $targetCountry The target country code.
+	 *
+	 * phpcs:disable WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
jjgrainger: @jamesmorrison please can we revert these changes as they're not relevant?

src/Internal/DependencyManagement/JobServiceProvider.php
```diff
@@ -58,6 +58,7 @@
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\MerchantStatuses;
 use Automattic\WooCommerce\GoogleListingsAndAds\MerchantCenter\PriceBenchmarks;
 use Automattic\WooCommerce\GoogleListingsAndAds\Product;
+use Automattic\WooCommerce\GoogleListingsAndAds\API\YouTube;
```
jjgrainger: @jamesmorrison can we update to `use Automattic\WooCommerce\GoogleListingsAndAds\API\YouTube\Connection as YouTubeConnection;`

src/Internal/DependencyManagement/JobServiceProvider.php
```diff
@@ -204,7 +205,7 @@ public function register(): void {
 		$this->share_with_tags( CsvExportWriter::class );
 
 		$this->share_action_scheduler_job( CreateYouTubeOrderIdsCache::class, YouTubeOrders::class, JobRepository::class );
-		$this->share_action_scheduler_job( CreateMerchantReportedConversionReport::class, OrderItemRowBuilder::class, CsvExportWriter::class );
+		$this->share_action_scheduler_job( CreateMerchantReportedConversionReport::class, OrderItemRowBuilder::class, CsvExportWriter::class, YouTube\Connection::class );
```
jjgrainger: @jamesmorrison once we update the reference we can use `YouTubeConnection::class`

src/Jobs/CreateMerchantReportedConversionReport.php
```diff
@@ -35,18 +52,53 @@ class CreateMerchantReportedConversionReport extends AbstractBatchedActionSchedu
...
+	/**
+	 * Schedule the job with lock protection.
+	 *
+	 * Prevents concurrent execution by checking for an active lock.
+	 * If a lock exists but has expired (> LOCK_TIMEOUT), it's considered stale
+	 * and the job proceeds.
+	 *
+	 * @param array $args Optional arguments to pass to the job.
+	 */
+	public function schedule( array $args = [] ): void {
```
jjgrainger: @jamesmorrison is this required? Doesn't ActionScheduler have a similar function to prevent the same job running concurrently?

src/Jobs/CreateMerchantReportedConversionReport.php
```diff
@@ -126,14 +200,32 @@ protected function process_items( array $items ) {
 				continue;
 			}
 
-			// Get items from the order
+			// Check file size before processing this order.
+			$current_file = $export_state[ $date ]['current_file'];
+			$file_size    = $this->writer->get_file_size( $current_file );
+
+			if ( $file_size >= self::FILE_SIZE_THRESHOLD ) {
```
jjgrainger, line 173: @jamesmorrison is there a danger of the next batch added tipping the file size over the limit? Should we look into a buffer here, like 90% of the file size threshold?

src/Utility/WPCLIGenerateSampleConversionReport.php
```diff
@@ -0,0 +1,369 @@
...
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class WPCLIGenerateSampleConversionReport
+ *
+ * WP-CLI command to generate sample CSV conversion data for testing.
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Utility
+ */
+class WPCLIGenerateSampleConversionReport implements Service, Registerable, Conditional {
```
jjgrainger: Thanks @jamesmorrison 

This is good but wonder if more is needed for testing than creating a CSV with mock data? I'll review this in more detail tomorrow and align with you on next steps if there are any.

jjgrainger (commented):
Thanks @jamesmorrison left some comments that are mostly questions. Please review and lets align on next steps here. Thanks

src/Utility/WPCLIGenerateSampleConversionReport.php
jjgrainger: @jamesmorrison While I think about it we may want to extract this to it's own MU plugin as it's primarily a test script and not part of the feature.

src/Coupon/WCCouponAdapter.php
jamesmorrison: Done, this was changed because linting failed locally, let's discuss if it's an issue here.

src/Internal/DependencyManagement/JobServiceProvider.php
jamesmorrison: This has been updated

src/Internal/DependencyManagement/JobServiceProvider.php
jamesmorrison: And this was updated too.

src/Jobs/CreateMerchantReportedConversionReport.php
jamesmorrison: This is accounted for:

```
	/**
	 * File size threshold to trigger new file creation (~9.5MB safety margin).
	 */
	protected const FILE_SIZE_THRESHOLD = 9961472;
```

I can make the max file size smaller if you think this is needed.

src/Jobs/CreateMerchantReportedConversionReport.php
jamesmorrison: Sorry, was not aware of the existing functionality. I've revised this to use ActionScheduler; please could you check this?

src/Jobs/CreateMerchantReportedConversionReport.php
jjgrainger: I think the margin you've added here is fine

src/Options/OptionsInterface.php
```diff
@@ -49,6 +49,8 @@ interface OptionsInterface {
 	public const WPCOM_REST_API_STATUS                     = 'wpcom_rest_api_status';
 	public const GTIN_MIGRATION_STATUS                     = 'gtin_migration_status';
 	public const API_PULL_SYNC_MODE                        = 'api_pull_sync_mode';
+	public const YOUTUBE_ORDER_IDS_CACHE                   = 'youtube_export_order_ids';
+	public const MERCHANT_CONVERSION_EXPORT_FILES          = 'merchant_conversion_export_files';
```
jjgrainger: @jamesmorrison nit pick but I think this option name could be misleading thinking its related to Merchant Center. Can we change the name to something like `YOUTUBE_CONVERSION_EXPORT_FILES` or even `YOUTUBE_EXPORT_FILES`?

src/Utility/WPCLIGenerateSampleConversionReport.php
jjgrainger: @jamesmorrison lets remove this from the PR but please keep a copy as a MU plugin. I can share an similar example where we can access the container outside of the plugin.

jamesmorrison: Ok, I've split to a separate mu-plugin (attached to Linear comment)

src/Options/OptionsInterface.php
jamesmorrison: Good call, I've updated this

ankitguptaindia:
QA Status: Approved ✅ 
---

**Test Result:**

PR works as scoped. CSV file generates and all jobs generates as expected. Downloaded CSV has all fields specified in [Merchant Reported Conversions Schema](https://services.google.com/fh/files/misc/yt_shopping_merchant_reported_conversion_schema.pdf). 


Screenshot 2026-01-15 at 6 48 26 PM


Screenshot 2026-01-15 at 6 47 43 PM


image

ankitguptaindia (approved):

PR #3160, opened 2025-11-28
Better Guidance Integration

ankitrox:

Closes: https://linear.app/a8c/issue/GOOWOO-304/update-account-creationupdates-hooksfunctions-used-during-onboarding

https://linear.app/a8c/issue/GOOWOO-303/update-accountcard-to-use-detailederror-component-to-render-errors

### Screenshots:

Screenshot 2026-01-05 at 6 05 09 PM

1. Make sure that the proxy is running. `PROXY_HOST='127.0.0.1' npm run test-proxy` and add following to wp-config.php file
```php
define( 'WOOCOMMERCE_GLA_CONNECT_SERVER_URL', 'http://host.docker.internal:5555' );
```
2. Go to onboarding step 1 and creating the Ads account and MC account should throw the error as shown in the attached screenshot.

### Changelog entry

>


js/src/components/detailed-error/index.js
```diff
@@ -12,29 +13,38 @@ import warningIconUrl from '~/images/icons/warning.svg';
...
-						alt=""
-						width={ 16 }
-						height={ 16 }
-					/>
-				</span>
-				<h4>{ title }</h4>
-			</div>
-			<p className="gla-detailed-error__description">{ description }</p>
-		</div>
+		<Fragment>
```
asvinb:
```suggestion
		<>
```

js/src/components/detailed-error/index.js
```diff
@@ -12,29 +13,38 @@ import warningIconUrl from '~/images/icons/warning.svg';
...
+							/>
+						</span>
+						<h4>{ title }</h4>
+					</div>
+					<p className="gla-detailed-error__description">
+						{ description }
+					</p>
+				</div>
+			) ) }
+		</Fragment>
```
asvinb:
```suggestion
		</>
```

js/src/components/detailed-error/index.js
```diff
@@ -12,29 +13,38 @@ import warningIconUrl from '~/images/icons/warning.svg';
...
-					/>
-				</span>
-				<h4>{ title }</h4>
-			</div>
-			<p className="gla-detailed-error__description">{ description }</p>
-		</div>
+		<Fragment>
+			{ formattedErrors.map( ( { title, description }, idx ) => (
+				<div
+					key={ idx }
```
asvinb: @ankitrox Is there another property from the `error` object which we can use to uniquely identify that error? Using the index as key is an anti pattern.

js/src/components/google-combo-account-card/connect-ads/connect-existing-account.js
```diff
@@ -58,17 +57,16 @@ const ConnectExistingAccount = ( { onCreateClick } ) => {
...
 		setLoading( true );
 		try {
 			await connectGoogleAdsAccount();
 			await fetchGoogleAdsAccountStatus();
 			await refetchGoogleAdsAccount();
-		} catch ( error ) {
-			receiveDetailedError( ERROR_SLOT, {
-				...error,
-				code: error.statusText,
+		} catch ( e ) {
```
asvinb:
```suggestion
		} catch ( error ) {
```
Let's keep the `error` name here instead of `e`

js/src/components/google-combo-account-card/connect-ads/connect-existing-account.js
```diff
@@ -58,17 +57,16 @@ const ConnectExistingAccount = ( { onCreateClick } ) => {
...
+			receiveDetailedError( ERROR_SLOTS.GOOGLE_ADS_CONNECTION, {
+				...e.data.error,
```
asvinb: @ankitrox Let's make sure the `data.error` property exists within the object.

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
```diff
@@ -200,6 +201,14 @@ const ConnectedGoogleComboAccountCard = () => {
 					/>
 				}
 				expandedDetail
+				errorSlots={
+					! showConnectAds && ! showConnectMC
```
asvinb: @ankitrox Do we need a condition here for the `errorSlots`? The reason being if there's no error for a slot, then nothing will be displayed.

js/src/data/constants.js
```diff
@@ -16,3 +16,9 @@ export const EMPTY_ASSET_ENTITY_GROUP = {
 	[ ASSET_GROUP_KEY.FINAL_URL ]: '',
 	[ ASSET_GROUP_KEY.DISPLAY_URL_PATH ]: [],
 };
+
+export const ERROR_SLOTS = {
+	GOOGLE_MC_CONNECTION: 'google_mc_connection',
```
asvinb: @ankitrox Let's add a prefix to the error slots in case we want to extend the error handling for other pages/sections in the future. Let's prefix the error slots with `setup-mc` which happens within onboarding. We can try to see how we can use the `CONTEXT_EXTENSION_ONBOARDING` [constant](https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/pages/onboarding/setup-stepper/saved-setup-stepper.js#L29). Basically we should be able to uniquely identify the slots.

js/src/hooks/useApiFetchCallback.js
```diff
@@ -172,6 +172,18 @@ const useApiFetchCallback = ( options, initialState = defaultState ) => {
 					throw e;
 				}
 
+				if ( e?.status >= 400 && e?.status < 600 ) {
```
asvinb: @ankitrox Why this change please? 🤔

js/src/hooks/useAutoCreateAdsMCAccounts.js
```diff
@@ -105,19 +108,40 @@ const useAutoCreateAdsMCAccounts = ( createMCAccount ) => {
 		if ( which ) {
 			const handleCreateAccountCallback = async () => {
 				if ( which === CREATING_BOTH_ACCOUNTS ) {
+					clearDetailedErrorBySlot(
```
asvinb: @ankitrox Instead of having multiple `clearDetailedErrorBySlot` calls, let's update `clearDetailedErrorBySlot` to `clearDetailedErrorBySlots` where you can pass an array of slots you want cleared.

js/src/hooks/useConnectMCAccount.js
```diff
@@ -39,14 +45,16 @@ const useConnectMCAccount = ( value ) => {
 				return;
 			}
 
-			if ( ! [ 409, 403 ].includes( e.status ) ) {
```
asvinb, line 42: @ankitrox We should keep this condition since sometimes we are intentionally silencing some errors.

js/src/hooks/useCreateMCAccount.js
```diff
@@ -26,14 +27,26 @@ const useCreateMCAccount = () => {
 			} );
 			invalidateResolution( 'getGoogleMCAccount', [] );
 		} catch ( e ) {
-			if ( ! [ 403, 503 ].includes( e.status ) ) {
```
asvinb, line 29: @ankitrox Same as above.

js/src/hooks/useCreateMCAccount.js
```diff
@@ -26,14 +27,26 @@ const useCreateMCAccount = () => {
...
 					__(
-						'Unable to create Merchant Center account. Please try again later.',
+						'Unable to connect your Google Merchant Center account. Please check your connection and try again.',
 						'google-listings-and-ads'
-					);
+					)
+				);
+				return;
+			}
+			if ( e?.code === 'API_ERROR' ) {
```
asvinb: @ankitrox Is there a way we can have a function/hook to reduce code duplication here, because i can see how we'll need to have this check in several places.

asvinb (changes requested):

js/src/components/detailed-error/index.js
ankitrox: Screenshot 2026-01-05 at 6 47 22 PM

`error` only has `title` and `description`  properties in it. Can we generate random keys instead of using index?

js/src/components/google-combo-account-card/connected-google-combo-account-card.js
ankitrox: Added the check in `js/src/components/account-card/index.js`, if `errorSlots .length > 0`, then error will be displayed.

js/src/hooks/useApiFetchCallback.js
ankitrox: We want to capture only HTTP errors in the range of 4xx and 5xx i.e. request errors and server errors.

js/src/hooks/useConnectMCAccount.js
ankitrox: @asvinb  It is still there

https://github.com/woocommerce/google-listings-and-ads/blob/f1a6899780a5641d85a2a2c8cd05f8283f104952/js/src/hooks/useCreateMCAccount.js#L42

js/src/hooks/useCreateMCAccount.js
ankitrox: It is still there 

https://github.com/woocommerce/google-listings-and-ads/blob/f1a6899780a5641d85a2a2c8cd05f8283f104952/js/src/hooks/useCreateMCAccount.js#L42

js/src/hooks/useCreateMCAccount.js
ankitrox: We have additional checks alongside this one for for the status code checks where we want to suppress the errors, so I think we can handle it later if is is not necessary as of now.

js/src/components/detailed-error/index.js
asvinb: @ankitrox We can use a combination of the `title` and `description`

js/src/hooks/useApiFetchCallback.js
asvinb: @ankitrox Is that in the requirements somewhere? 🤔

js/src/hooks/useAutoCreateAdsMCAccounts.js
```diff
@@ -105,19 +109,39 @@ const useAutoCreateAdsMCAccounts = ( createMCAccount ) => {
 		if ( which ) {
 			const handleCreateAccountCallback = async () => {
 				if ( which === CREATING_BOTH_ACCOUNTS ) {
+					const slotsToClear = [
```
asvinb, line 92: @ankitrox I'am thinking whether we can just clear both slots here ahead of any connection.

js/src/hooks/useConnectMCAccount.js
```diff
@@ -39,15 +45,14 @@ const useConnectMCAccount = ( value ) => {
...
-				const message =
-					body.message ||
-					__(
-						'Unable to connect Merchant Center account. Please try again later.',
-						'google-listings-and-ads'
-					);
-				createNotice( 'error', message );
+			if (
+				e?.code === 'API_ERROR' &&
+				! [ 409, 403 ].includes( e.data?.statusCode )
```
asvinb: @ankitrox Just confirming, the `statusCode` property is new right?

asvinb (approved):
@ankitrox I’m going to pre-approve the PR, as the code LGTM. However, I wasn’t able to fully test the feature. After running the proxy and navigating to the onboarding flow, I encounter the following error on the /wc/gla/ads/account-status?_locale=user endpoint:
```{"message":"Error retrieving status"}```
and I'am stuck on a a loading state.

js/src/hooks/useConnectMCAccount.js
ankitrox: Yes @asvinb, the `statusCode` property is the new one according to the schema given [here](https://linear.app/a8c/issue/GOOWOO-214/better-guidance-for-users-during-onboarding-overview#comment-00be2bde)

js/src/hooks/useAutoCreateAdsMCAccounts.js
ankitrox: Screenshot 2026-02-10 at 8 27 48 PM

I think we should keep the error for one card when connecting the other one because we may oversee the error for one card when connecting the error, if it was a simple step to resolve the error, user may have encounter the error again which leads to additional friction during onboarding.

ankitguptaindia:
QA Status: Approved ✅ 
----

**Test Result:**  Testing was conducted using both a proxy setup and a real Ads account that had restrictions related to account creation limits. The focus was to validate how the system handles such errors and communicates them to the user.

### **Errors Screen with the use of Proxy:**

Screenshot 2026-04-16 at 9 01 29 PM


### **Errors Screen with the Real error with a specific Ads account:**

Screenshot 2026-04-16 at 9 02 15 PM

PR #3165, opened 2025-12-04
Add deploy action to this repo

joemcgill:

Closes #GOOWOO-368.

Adding the woo-product-deploy action to this repo

1. Prepare a release branch
2. Run the workflow in simulate mode
3. Review the workflow logs for any errors

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,6 +12,11 @@ on:
         required: false
         default: true
         type: boolean
+      deploy_wporg:
+        description: 'Deploy to WordPress.org'
+        required: false
+        default: true
+        type: boolean
       deploy_github:
         description: 'Deploy to GitHub'
         required: false
@@ -30,6 +35,11 @@ on:
         type: string
       deploy_woo:
         description: 'Deploy to Woo.com'
+        required: false
+        default: true
+        type: boolean
+      deploy_wporg:
+        description: 'Deploy to WordPress.org'
         required: false
         default: true
         type: boolean
@@ -70,6 +80,9 @@ jobs:
           simulate: ${{ github.event.inputs.simulate }}
           wccom_release: ${{ github.event.inputs.deploy_woo }}
           github_release: ${{ github.event.inputs.deploy_github }}
+          wporg_release: ${{ github.event.inputs.deploy_wporg }}
+          wporg_username: ${{ secrets.WPORG_USERNAME }}
+          wporg_password: ${{ secrets.WPORG_PASSWORD }}
           deploy_pat: ${{ secrets.ORG_DEPLOY_SECRET }}
           partner_user: ${{ secrets.ORG_PARTNER_USER }}
           partner_secret: ${{ secrets.ORG_PARTNER_DEPLOY_SECRET }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,37 +7,6 @@ on:
         description: 'Version to deploy (e.g., 3.5.0)'
         required: true
         type: string
-      deploy_woo:
-        description: 'Deploy to Woo.com'
-        required: false
-        default: true
-        type: boolean
-      deploy_wporg:
-        description: 'Deploy to WordPress.org'
-        required: false
-        default: true
-        type: boolean
-      deploy_github:
-        description: 'Deploy to GitHub'
-        required: false
-        default: true
-        type: boolean
-      simulate:
-        description: 'Dry run mode'
-        required: false
-        default: true
-        type: boolean
-  workflow_call:
-    inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,76 @@
...
+          version: ${{ github.event.inputs.version }}
+          node_version: '20'
+          npm_version: '10'
+          simulate: ${{ github.event.inputs.simulate }}
+          wccom_release: ${{ github.event.inputs.deploy_woo }}
+          github_release: ${{ github.event.inputs.deploy_github }}
+          deploy_pat: ${{ secrets.ORG_DEPLOY_SECRET }}
+          partner_user: ${{ secrets.ORG_PARTNER_USER }}
+          partner_secret: ${{ secrets.ORG_PARTNER_DEPLOY_SECRET }}
+          wccom_product_id: ######
```
joemcgill: @eason9487 do you have this product ID that we can add here?

eason9487: It only needs to release this plugin to WPORG. Subsequently, an automated mechanism will sync that new version from WPORG to WPCOM in the background. Therefore, this parameter doesn't need to be configured.

joemcgill: Good to know. So I it seems like I should be able to remove the `wccom_release` configuration from this workflow altogether.

dkotter: We were initially told the `wccom_product_id` was always required (and that ID was provided to us for Google for Woo if you want it) but in looking at the `woo-product-deploy` Action, does seem like that ID is only needed when deploying to Woo.com which isn't needed for plugins that deploy to WP.org.

So for this plugin, you should only need the `wporg_release` and `github_release` options, no need for the `wccom_release` option

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,89 @@
+name: Deploy Product
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to deploy (e.g., 3.5.0)'
+        required: true
+        type: string
+      deploy_woo:
```
dkotter: This can be removed since we don't deploy this directly to Woo.com

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,89 @@
...
+          ref: trunk
+          path: ./.github/actions/woo-product-deploy
+          token: ${{ secrets.ORG_DEPLOY_SECRET }}  # Need PAT with access to the action repository
+
+      # Use the action we checked out
+      - name: Deploy Product
+        uses: ./.github/actions/woo-product-deploy
+        with:
+          repo_url: https://github.com/woocommerce/google-listings-and-ads
+          branch: ${{ github.ref_name }}
```
dkotter, line 45: We hardcoded this across all other extensions to be `trunk` as we always deploy from there. Not sure how this repo is setup but you could hardcode if you wanted to avoid ever accidentally deploying from the wrong branch

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,89 @@
...
+          version: ${{ github.event.inputs.version }}
+          node_version: '20'
+          npm_version: '10'
+          simulate: ${{ github.event.inputs.simulate }}
+          wccom_release: ${{ github.event.inputs.deploy_woo }}
```
dkotter: This can be removed

.github/workflows/deploy.yml
joemcgill: Thanks, @dkotter. In this repo releases are usually deployed from release branches and then those get merged into trunk as the last step, but definitely worth considering if we should hard code this to try to avoid accidental releases from other branches.

.github/workflows/deploy.yml
joemcgill: I'm removing this since we're not planning standalone releases to WCCOM

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,56 @@
...
+        type: boolean
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checkout the private action repository
+      - name: Checkout woo-product-deploy action
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
joemcgill, line 33: @dkotter any reason to not bump this to the latest?

```suggestion
        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
```

dkotter: That should be fine, noting I haven't looked at what's changed from v4 to v6, if any breaking changes

dkotter (approved):

PR #3177, opened 2025-12-10
[GOOWOO-372] Populate glaData with the serviceBasedMerchant property

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-372/populate-gladata-with-the-servicebasedmerchant-property

### Screenshots:

Screen recording showing value of `glaData.serviceBasedMerchant` is `false` where any non-virtual product exists; `true` where only virtual product exists.
https://github.com/user-attachments/assets/703c19ff-a473-4ff8-8192-30d06dc15cb0

**Test Scenario 1: Service-based merchant (virtual products only)**

1. Ensure the `is_service_based_merchant` option is cleared (or test on a fresh install)
2. Create one virtual product, confirm it's published
3. Navigate to `/wp-admin/admin.php?page=wc-admin`
4. Open the console, type `glaData.serviceBasedMerchant` - expected value is `true` (only virtual products exist)
5. Create a non-virtual product, confirm it's published
6. Refresh `/wp-admin/admin.php?page=wc-admin`
7. Open the console, type `glaData.serviceBasedMerchant `- expected value is still `true` (value doesn't change after initial setup)

**Test Scenario 2: Non-service-based merchant (has physical products)**

1. Clear the `is_service_based_merchant` option (or test on a fresh install)
2. Create one non-virtual product, confirm it's published
3. Navigate to `/wp-admin/admin.php?page=wc-admin`
4. Open the console, type `glaData.serviceBasedMerchant` - expected value is `false` (has physical product)
5. Trash the non-virtual product, leaving only virtual products
6. Refresh `wp-admin/admin.php?page=wc-admin`
7. Open the console, type `glaData.serviceBasedMerchant` - expected value is still `false` (value doesn't change after initial setup)

**Test Scenario 3: Mixed products at setup**

1. Clear the `is_service_based_merchant` option (or test on a fresh install)
2. Create one virtual product and one non-virtual product, confirm both are published
3. Navigate to `/wp-admin/admin.php?page=wc-admin`
4. Open the console, type `glaData.serviceBasedMerchant` - expected value is `false` (has at least one physical product)


.github/workflows/php-unit-tests.yml
```diff
@@ -101,6 +101,17 @@ jobs:
       - name: Install WP tests
         run: ./bin/install-wp-tests.sh wordpress_test root root localhost ${{ matrix.wp-version }} ${{ matrix.wc-versions }}
 
+      - name: Prepare node
```
jjgrainger: @jamesmorrison can we revert all changes to this file. It appears these changes are the ones that have made the PHPunit tests workflow stop working, where as in the previous commit unit tests would run but fail. Once we have the unit tests working again we can see how to resolve the specific issue you faced.

src/Admin/Admin.php
```diff
@@ -122,7 +130,7 @@ protected function get_assets(): array {
 			new BuiltScriptDependencyArray(
 				[
 					'dependencies' => [],
-					'version'      => (string) filemtime( "{$this->get_root_dir()}/js/build/index.js" ),
+					'version'      => (string) ( file_exists( "{$this->get_root_dir()}/js/build/index.js" ) ? filemtime( "{$this->get_root_dir()}/js/build/index.js" ) : time() ),
```
jjgrainger: @jamesmorrison lets revert the `version` changes here too if they were an attempt to resolve the issue you faced with the unit tests.

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,240 @@
...
+		// Clear cache when products are deleted or trashed.
+		add_action( 'wp_trash_post', [ $this, 'maybe_clear_cache_on_post_change' ], 10, 2 );
+		add_action( 'before_delete_post', [ $this, 'maybe_clear_cache_on_post_change' ], 10 );
+		add_action( 'deleted_post', [ $this, 'maybe_clear_cache_on_post_change' ], 10 );
+
+		// Clear cache when products are restored from trash.
+		add_action( 'untrashed_post', [ $this, 'maybe_clear_cache_on_post_change' ], 10, 2 );
+
+		// Clear cache when WooCommerce clears product transients.
+		add_action( 'woocommerce_delete_product_transients', [ $this, 'clear_cache_on_product_change' ], 10 );
```
jjgrainger: @jamesmorrison we only need to determine if a store is a service based merchant _once_ during the initial onboarding stage. Therefore we shouldn't be tracking changes to products and updating the value again.

Can we remove these hooks and the cache clear method.

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,240 @@
...
+use WC_Product;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class ServiceBasedMerchantState
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\ServiceBasedMerchant
+ */
+class ServiceBasedMerchantState implements Service, Registerable, OptionsAwareInterface, TransientsAwareInterface {
```
jjgrainger: @jamesmorrison can we move this file to the `src/Options` directory.

We're avoiding creating a whole feature architecture and another folder around what is effectively going to be a single use value during onboarding.

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,240 @@
...
+	public function is_service_based_merchant(): bool {
+		$option_value = $this->options->get( OptionsInterface::IS_SERVICE_BASED_MERCHANT );
+
+		// If option is null, calculate it now.
+		if ( null === $option_value ) {
+			return $this->calculate_service_based_merchant();
+		}
+
+		return (bool) $option_value;
+	}
```
jjgrainger, line 48: @jamesmorrison we can likely simplify this to a handful of methods:

An `is_service_based_merchant()` that checks the stored option and is null, calls `has_physical_products()` and sets the option based on the results before returning true/false.

Then `has_physical_products()` performs the products loop and returns true if it finds a physical product.

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,240 @@
...
+	public function has_physical_products(): bool {
+		/**
+		 * Filter to bypass or modify the cached result of has_physical_products.
+		 *
+		 * @since 1.0.0
+		 *
+		 * @param bool|null $cached_result The cached result, or null to use default caching behavior.
+		 * @return bool|null True if physical products exist, false otherwise, or null to use cache.
+		 */
+		$filtered_result = apply_filters( 'woocommerce_gla_has_physical_products', null );
```
jjgrainger: @jamesmorrison I don't think theres a need to include custom filters for this value now we're looking to simplify.

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,240 @@
...
+
+		if ( null !== $filtered_result ) {
+			return (bool) $filtered_result;
+		}
+
+		// Try to get from cache (transients can't store booleans, so we store as int: 0 or 1).
+		$cached_value = $this->transients->get( TransientsInterface::HAS_PHYSICAL_PRODUCTS );
```
jjgrainger: @jamesmorrison I don't think we need any transients or cache busting either now we're looking to simplify.

tests/Unit/Admin/AdminTest.php
```diff
@@ -0,0 +1,115 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Tests\Framework\ContainerAwareUnitTest;
+use PHPUnit\Framework\MockObject\MockObject;
+use ReflectionMethod;
+
+/**
+ * Class AdminTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Admin
+ */
+class AdminTest extends ContainerAwareUnitTest {
```
jjgrainger: @jamesmorrison Ah! If we had no unit tests for the Admin class initially, and these tests are whats causing issues right now, I would remove and not include it as part of this feature work.

As long as we have tests for the `ServiceBasedMerchantState` we should be confident enough that that `glaData` will provide the correct values needed.

tests/Unit/Options/ServiceBasedMerchantStateTest.php
```diff
@@ -0,0 +1,762 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+use WC_Helper_Product;
+use WC_Product;
+
+/**
+ * Class ServiceBasedMerchantStateTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\ServiceBasedMerchant
+ */
+class ServiceBasedMerchantStateTest extends ContainerAwareUnitTest {
```
jjgrainger, line 17: @jamesmorrison once we've simplified the implementation for the `ServiceBasedMerchantState` lets make sure to update tests here also.

jjgrainger (changes requested):
Thanks @jamesmorrison 

Theres some good work here but a lot we can do to simplify things. Let me know if you have any questions.

src/Options/ServiceBasedMerchantState.php
jamesmorrison: Ok ,this file has been moved

.github/workflows/php-unit-tests.yml
jamesmorrison: This has been reverted

src/Admin/Admin.php
jamesmorrison: This has been reverted

src/Options/ServiceBasedMerchantState.php
jamesmorrison: This has been updated / simplified

src/Options/ServiceBasedMerchantState.php
jamesmorrison: This has been updated

src/Options/ServiceBasedMerchantState.php
jamesmorrison: This has been removed

src/Options/ServiceBasedMerchantState.php
jamesmorrison: This has been reverted, there is no more caching

tests/Unit/Admin/AdminTest.php
jamesmorrison: I don't think this was contributing to the errors I was seeing, but have removed the file.

tests/Unit/Options/ServiceBasedMerchantStateTest.php
jamesmorrison: These have been updated

src/Options/TransientsInterface.php
```diff
@@ -22,6 +22,7 @@ interface TransientsInterface {
 	public const MC_STATUSES                = 'mc_statuses';
 	public const URL_MATCHES                = 'url_matches';
 	public const WPCOM_API_STATUS           = 'wpcom_api_status';
+	public const HAS_PHYSICAL_PRODUCTS      = 'has_physical_products';
```
jjgrainger: @jamesmorrison It looks like we no longer need this transient as it's not used? Please can we remove all references from this file.

jamesmorrison: I've removed this.

.github/workflows/php-unit-tests.yml
```diff
@@ -149,4 +149,3 @@ jobs:
 
       - name: Run PHP unit tests
         run: composer test-unit
-
```
mukeshpanchal27, line 153: Revert this change.

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,92 @@
...
+	 *
+	 * @return bool True if service-based, false otherwise.
+	 */
+	public function is_service_based_merchant(): bool {
+		$option_value = $this->options->get( OptionsInterface::IS_SERVICE_BASED_MERCHANT );
+
+		// If option is null, calculate it now.
+		if ( null === $option_value ) {
+			$service_based = ! $this->has_physical_products();
+			$this->options->update( OptionsInterface::IS_SERVICE_BASED_MERCHANT, $service_based );
```
mukeshpanchal27, line 43: If `$service_based` returns `false`, the call `$this->options->update( OptionsInterface::IS_SERVICE_BASED_MERCHANT, $service_based );` does not persist the option in the database. As a result, on every page load the code runs through this expensive detection loop again, which is not ideal from a performance perspective.

jjgrainger (approved):

tests/Unit/Options/ServiceBasedMerchantStateTest.php
```diff
@@ -0,0 +1,130 @@
...
+		$virtual_product->save();
+
+		$result = $this->service_based_merchant_state->has_physical_products();
+
+		$this->assertFalse( $result );
+
+		// Cleanup
+		$virtual_product->delete( true );
+	}
+
```
mukeshpanchal27, line 120: Can we add at least one unit test to cover batch processing?

For example, create 150 products, then add a physical product at the end and verify that the batch processing behaves as expected. This would help ensure the logic works correctly for larger datasets and edge cases.

tests/Unit/Options/ServiceBasedMerchantStateTest.php
```diff
@@ -0,0 +1,130 @@
...
+		$result = $this->service_based_merchant_state->has_physical_products();
+
+		// Should return true as soon as physical product is found
+		$this->assertTrue( $result );
+
+		// Cleanup
+		$physical_product->delete( true );
+		$virtual_product->delete( true );
+	}
```
mukeshpanchal27, line 140: Add unit test that check `false` option
```
public function test_is_service_based_merchant_returns_false() {
    $this->options->expects( $this->once() )
        ->method( 'get' )
        ->with( OptionsInterface::IS_SERVICE_BASED_MERCHANT )
        ->willReturn( false );

    $result = $this->service_based_merchant_state->is_service_based_merchant();

    $this->assertFalse( $result );
}
```

mukeshpanchal27 (changes requested):
Thanks for the PR. Overall, it looks solid, but I’ve left a bit of feedback.

.github/workflows/php-unit-tests.yml
jamesmorrison: This has been reverted, but it leaves 2 blank lines at the end; assume this is intentional.

src/Options/ServiceBasedMerchantState.php
jamesmorrison: Following the patterns elsewhere in the project I've changed from true/false to yes/no; the value is now always cached.

tests/Unit/Options/ServiceBasedMerchantStateTest.php
jamesmorrison: I've added a test for batch processing

tests/Unit/Options/ServiceBasedMerchantStateTest.php
jamesmorrison: I've added a test here.

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,92 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class ServiceBasedMerchantState
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\ServiceBasedMerchant
```
mukeshpanchal27:
```suggestion
 * @package Automattic\WooCommerce\GoogleListingsAndAds\Options
```

src/Options/ServiceBasedMerchantState.php
```diff
@@ -0,0 +1,92 @@
...
+					'limit'  => self::BATCH_SIZE,
+					'offset' => $offset,
+					'status' => 'publish',
+				]
+			);
+
+			foreach ( $products as $product ) {
+				// Check if product is not virtual and requires shipping.
+				// Note: needs_shipping() returns !is_virtual(), so this check is essentially: !is_virtual
+				if ( ! $product->is_virtual() && $product->needs_shipping() ) {
```
mukeshpanchal27: The `!$product->is_virtual()` check is redundant since `needs_shipping()` already returns `!$this->is_virtual()` (with a filter applied).

https://github.com/woocommerce/woocommerce/blob/7ff2029e134b924d4437342efc67573a10597c3f/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L1766-L1768

mukeshpanchal27 (approved):
Thanks for the updates. Pre-approving this as it looks good to me. I’ve left two minor pieces of feedback.

ankitguptaindia:
### **QA Status: Needs feedback ⚠️**

Thanks for the PR, @jamesmorrison.

I tested this by following the provided instructions, but I’m still getting the value as `false`, even when the setup includes:

* Virtual-only products
* A mix of Virtual + Non-Virtual products

Please take a look at the video explanation below—I’ve demonstrated all the testing steps and scenarios from my end.





https://github.com/user-attachments/assets/1228408a-b9db-48b0-87ee-e661f262a062

jamesmorrison:
Thanks @ankitguptaindia and my apologies here.

The scope of the task changed meaning the test steps were incorrect. The initial implementation had the value of `serviceBasedMerchant` change after a product is added / edited / deleted; the updated scope makes this a one-time setup value which is never updated.

I have updated the test steps; this accounts for each scenario after the initial setup based on the published products before the plugin is configured.

Please could you run through the updated steps?

ankitguptaindia:
QA Status: Approved ✅ 
---

Thanks @jamesmorrison for the details. Now I can see the value `true` on a fresh setup with virtual product and `false` when having Non virtual product.

ankitguptaindia (approved):
Re-verfied after fixing Merge Conflicts. ✅

PR #3179, opened 2025-12-11
Implement new setup flow for service-based merchants

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-364/update-onboarding-flow and https://linear.app/a8c/issue/GOOWOO-374/add-target-countries-in-first-step-of-the-onboarding-flow

1. Update `src/Admin/Admin.php` to set `serviceBasedMerchant` to `true`.
2. Go through the onboarding flow. There should be the "Setup your accounts", "Create a Campaign" and "Optimize your Campaign" tabs.
3. On the first step, once a Google Ads account is connected, the user should be prompted to choose the target audience. The continue button remains disabled until a Google Ads account is connected and a country/countries chosen.
4. On the second step, users should be able to choose their budget and setup billing.
5. Clicking "Skip this step" will prompt users why they are skipping and if confirmed will redirect the user. This is not working currently and depends on GOOWOO-380
6. Clicking "Continue" will return an error as well and is dependent on GOOWOO-373. The response can be mocked (using the same response as the usual flow)
7. On the assets screen, it should work as on the current flow when adding a paid campaign. However skipping and "Save changes" will not work because they are blocked by GOOWOO-380 and GOOWOO-373

### Changelog entry

>


js/src/pages/onboarding/setup-stepper/index.js
```diff
@@ -31,7 +37,15 @@ const SetupStepper = () => {
 		return null;
 	}
 
-	return <SavedSetupStepper savedStep={ stepNameKeyMap[ step ] } />;
+	if ( serviceBasedMerchant ) {
+		return (
+			<AdsOnlySetupStepper
+				savedStep={ SERVICE_BASED_STEP_NAME_KEY_MAP[ step ] }
```
ankitrox: Is there any possibility that backend may send unexpected value so that if `SERVICE_BASED_STEP_NAME_KEY_MAP[ step ]` does not exists, we can default to `SERVICE_BASED_STEP_NAME_KEY_MAP.accounts` to avoid rendering null currentStep?

js/src/hooks/useAdsSetupCompleteCallback.js
```diff
@@ -34,21 +34,29 @@ export default function useAdsSetupCompleteCallback() {
...
+					countryCodes,
+					hasConfirmedEuPoliticalContent
+				);
+
+				await completeAdsSetup();
+				await onCompleted?.( payload );
+
+				return payload;
+			} catch ( error ) {
+				setLoading( false );
```
ankitrox: Loading is set to `false` only when error occurs, do we need to set in success path as well?

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -51,7 +51,10 @@ export default function AdsCampaign( {
 		context === 'edit-ads';
 	// only show the billing card during onboarding or setup Ads flow.
 	// For creating/editing a campaign, we assume billing is already set up.
-	const showBillingCard = context === 'setup-mc' || context === 'setup-ads';
+	const showBillingCard =
+		context === 'setup-mc' ||
+		context === 'setup-ads' ||
+		context === 'setup-service-based';
```
ankitrox: Add a type for `setup-service-based` context in jsdoc.

ankitrox (changes requested):

js/src/pages/onboarding/setup-stepper/index.js
asvinb: Sure thing @ankitrox

js/src/hooks/useAdsSetupCompleteCallback.js
asvinb: Good catch @ankitrox . However I reverted those changes and opted for a simpler approach leveraging the existing functionality.

js/src/components/paid-ads/ads-campaign/ads-campaign.js
asvinb: Good catch. I've renamed it to `setup-ads-only` and added it in the JSDoc. Thanks!

ankitrox (approved):
LGTM. ✅ 
Moving to QA.

ankitrox (approved):

ankitguptaindia:
QA Status: Verified ✅ 
---
**Test Result:** Tested the feature as scoped. All cases are working as expected. We will recheck this once other related PR/tasks are merged. 


https://github.com/user-attachments/assets/36595406-8066-4579-a3ae-5b052a90ceff

PR #3182, opened 2025-12-17
[GOOWOO-380] Create new REST API endpoint to mark service based merchant setup complete

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-380/create-new-rest-api-endpoint-to-mark-service-based-merchant-setup#comment-f0506946

### Screenshots:

Postman request and response:
Screenshot 2025-12-17 at 11 27 26

CURL request and response:
Screenshot 2025-12-17 at 11 32 37

Postman
1. Import this collection: [380-postman-testing.json](https://github.com/user-attachments/files/24211849/380-postman-testing.json) to Postman
2. Update the variables to add your install URL, username and application password (create one if need be)
3. Click Send to make the request

CURL:
1.Update this to your URL, username and application; then run in a terminal:
 `curl -X POST "https://woo.test/wp-json/wc/gla/google/onboarding/complete" -u "username:generate-an-application-password" | jq`

### DIFF
--- a/src/API/Site/Controllers/OnboardingController.php
+++ b/src/API/Site/Controllers/OnboardingController.php
@@ -42,7 +42,6 @@ class OnboardingController extends BaseController {
 	protected function get_complete_callback(): callable {
 		return function ( Request $request ) {
 			do_action( 'woocommerce_gla_onboarding_completed' );
-			do_action( 'woocommerce_gla_track_event', 'onboarding_complete', [] );
 
 			return new Response(
 				[
--- a/src/API/Site/Controllers/OnboardingController.php
+++ b/src/API/Site/Controllers/OnboardingController.php
@@ -1,0 +1,67 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers;
+
+use Automattic\WooCommerce\GoogleListingsAndAds\API\TransportMethods;
+use WP_REST_Request as Request;
+use WP_REST_Response as Response;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class OnboardingController
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\API\Site\Controllers
+ */
+class OnboardingController extends BaseController {
+
+	use EmptySchemaPropertiesTrait;
+
+	/**
+	 * Register rest routes with WordPress.
+	 */
+	public function register_routes(): void {
+		$this->register_route(
+			'google/onboarding/complete',
+			[
+				[
+					'methods'             => TransportMethods::CREATABLE, 
+					'callback'            => $this->get_complete_callback(),
+					'permission_callback' => $this->get_permission_callback(),
+				],
+			]
+		);
+	}
+
+	/**
+	 * Get the callback for completing onboarding.
+	 *
+	 * @return callable
+	 */
+	protected function get_complete_callback(): callable {
+		return function ( Request $request ) {
+			do_action( 'woocommerce_gla_onboarding_completed' );
+			do_action( 'woocommerce_gla_track_event', 'onboarding_complete', [] );
+
+			return new Response(
+				[
+					'status'  => 'success',
+					'message' => __( 'Successfully onboarded service based merchant', 'google-listings-and-ads' ),
+				],
+				200
+			);
+		};
+	}
+
+	/**
+	 * Get the item schema name for the controller.
+	 *
+	 * Used for building the API response schema.
+	 *
+	 * @return string
+	 */
+	protected function get_schema_title(): string {
+		return 'onboarding';
+	}
+}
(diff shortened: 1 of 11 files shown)


src/Admin/Redirect.php
```diff
@@ -98,12 +98,12 @@ public function maybe_redirect() {
 		}
 
 		// If setup ISNT complete then redirect from dashboard to onboarding
-		if ( ! $this->merchant_center->is_setup_complete() && $this->is_current_wc_admin_page( self::PATHS['dashboard'] ) ) {
+		if ( ! boolval( $this->options->get( OptionsInterface::ONBOARDING_COMPLETED_AT ) ) && $this->is_current_wc_admin_page( self::PATHS['dashboard'] ) ) {
```
jjgrainger: @jamesmorrison following on from the point you raised about existing users, I think it would be good to create a new method on this class `is_onboarding_complete()` to help us and also replace the calls for `$this->options->get( OptionsInterface::ONBOARDING_COMPLETED_AT )` here.

My thinking is that the new `is_onboarding_complete()` method will first check the `ONBOARDING_COMPLETED_AT` option and if is not `null` return true.

If the option is `null`, we then check the `$this->merchant_center->is_setup_complete()`. If it has been completed, we assume that this is a onboarded user and can update `ONBOARDING_COMPLETED_AT` option with the same timestamp as `MC_SETUP_COMPLETED_AT` option. Then the method returns true.

Let me know if you agree with this? Thanks

src/API/Site/Controllers/OnboardingController.php
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * Get the callback for completing onboarding.
+	 *
+	 * @return callable
+	 */
+	protected function get_complete_callback(): callable {
+		return function ( Request $request ) {
+			do_action( 'woocommerce_gla_onboarding_completed' );
+			do_action( 'woocommerce_gla_track_event', 'onboarding_complete', [] );
```
jjgrainger: @asvinb do we need to track an event here or should we remove this?

tests/Unit/Admin/RedirectsTest.php
```diff
@@ -43,54 +44,197 @@ public function setUp(): void {
 
 	/**
 	 * Test `maybe_redirect` to confirm that merchant is only redirected to onboarding if:
-	 * 1. Merchant center setup is not complete
+	 * 1. Onboarding is not complete
```
jjgrainger, line 47: @jamesmorrison we will want to include a test for already onboarded users where the `ONBOARDING_COMPLETED_AT` is `null` but the `MC_SETUP_COMPLETED_AT` option is set.

jjgrainger (changes requested):
Thanks @jamesmorrison a great start, I've left some feedback based on our discussions. Let me know if you have any questions. Thanks

src/Admin/Redirect.php
jamesmorrison: Thanks @jjgrainger - makes sense and I've made this change

tests/Unit/Admin/RedirectsTest.php
jamesmorrison: Thanks @jjgrainger, have added a test for this.

src/Admin/Redirect.php
```diff
@@ -82,6 +82,28 @@ public function activate(): void {
 		}
 	}
 
+	/**
+	 * Check if onboarding is complete.
+	 * For backwards compatibility, checks MC setup if ONBOARDING_COMPLETED_AT is not set.
+	 *
+	 * @return bool
+	 */
+	protected function is_onboarding_complete(): bool {
```
jjgrainger: Thanks @jamesmorrison there is a new requirement to add this value to the `glaData`.

To do this I think we should move this method to the `OnboardingCompleted` class and reference it from there.

We can use the `MerchantCenterAwareInterface` and `MerchantCenterAwareTrait` so the service is available in `OnboardingCompleted`.

I've updated the ticket to reflect this.

src/API/Site/Controllers/OnboardingController.php
jjgrainger: @jamesmorrison lets remove this for now.

jjgrainger (changes requested):

ankitrox:
@jamesmorrison Please note that I have added a new change in [51697d9](https://github.com/woocommerce/google-listings-and-ads/pull/3182/commits/51697d9a1f0eafc0a3e7ff7f201382b0c4320f81) where `OnboardingCompleted` is getting registered, this is required for [GOOWOO-384](https://linear.app/a8c/issue/GOOWOO-384/set-onboarding-as-complete-from-the-fe), feel free to make the changes, if required.

CC: @jjgrainger

src/API/Site/Controllers/OnboardingController.php
jamesmorrison: Ok, have removed

src/Admin/Redirect.php
jamesmorrison: I've updated this and corresponding tests

jjgrainger (approved):

PR #3217, opened 2026-01-16
Add time zone and language code fields when creating a merchant account for Merchant API compatibility

eason9487:

This PR adds `timeZone` and `languageCode` fields when creating a merchant account via WCS. It's a part of the Merchant API migration. It also updates error detection to support both Content API and Merchant API error formats when creating a merchant account.

Merchant API Ref:
- [Method: accounts.createAndConfigure](https://developers.google.com/merchant/api/reference/rest/accounts_v1/accounts/createAndConfigure)
- [REST Resource: accounts](https://developers.google.com/merchant/api/reference/rest/accounts_v1/accounts#Account)

1. Checkout [WCS #3071](https://github.com/Automattic/woocommerce-connect-server/pull/3071) and set up a local server according to that PR
2. Go to the Onboarding flow
3. Check if a new Google Merchant Center account can be created normally

### Changelog entry

> Update - Add `timeZone` and `languageCode` fields to create-merchant request for Merchant API compatibility.

### DIFF
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -166,16 +166,23 @@ class Middleware implements ContainerAwareInterface, OptionsAwareInterface {
 		} catch ( ClientExceptionInterface $e ) {
 			$message = $this->client_exception_message( $e, __( 'Error creating account', 'google-listings-and-ads' ) );
 
-			if ( preg_match( '/terms?.* are|is not allowed/', $message ) ) {
+			// Content API for Shopping: Invalid account name terms.
+			// Merchant API: Account name validation failed.
+			if ( preg_match( '/terms?.* are|is not allowed/', $message ) ||
+				preg_match( '/the account did not pass validation/i', $message ) ) {
 				throw InvalidTerm::contains_invalid_terms( $name );
 			}
 
+			$site_url = $this->strip_url_protocol( esc_url_raw( $this->get_site_url() ) );
+
+			// Content API for Shopping: Invalid top-level domain name.
 			if ( strpos( $message, 'URL ends with an invalid top-level domain name' ) !== false ) {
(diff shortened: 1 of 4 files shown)


src/API/Google/Middleware.php
```diff
@@ -602,11 +620,7 @@ protected function new_account_name(): string {
 	 * @return string
 	 */
 	protected function default_account_name(): string {
-		return sprintf(
-			/* translators: 1: current date in the format Y-m-d */
-			__( 'Account %1$s', 'google-listings-and-ads' ),
-			( new DateTime() )->format( 'Y-m-d' )
-		);
```
eason9487, line 609: Removing i18n is because this is primarily a fallback name used specifically during Google Merchant Center account creation, not displayed to users on the plugin.

src/API/Google/Middleware.php
```diff
@@ -602,11 +620,7 @@ protected function new_account_name(): string {
...
+		return 'Account ' . gmdate( 'dMy' );
```
yuliyan, line 623: Is there a reason to switch to a different date format?

yuliyan (approved):
Tested and confirmed the account is created with the correctly using the site language and timezone:
1kbnVXHJBBP0RhAn


LGTM!

src/API/Google/Middleware.php
eason9487: The name pattern “Account 2026-01-28” successfully creates accounts when using the Content API. However, when using the Merchant API, it fails with the error “name is invalid because MAX_DIGITS_EXCEEDED”. The official documentation currently lacks details about this error. After testing, I inferred that the name can contain only up to 5 digits, which is why it was changed to this format.

PR #3207, opened 2026-01-13
Add Gen AI media and text asset generation functionality

asvinb:
### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/GOOWOO-392/get-media-and-text-generated-assets

_Replace this with a good description of your changes & reasoning._

### Detailed test instructions:

1. Mock the `ads/assets/generate-text` and `ads/assets/generate-images` endpoints (POST) as per https://linear.app/a8c/issue/GOOWOO-405/create-a-wp-rest-endpoint-for-asset-generation in DevTools
	* Text
		```
		{
			"final_url": "https://example.com",
			"items": [
				{
					"text": "Generated headline text example.",
					"type": "headline"
				},
				{
					"text": "Generated long headline text example.",
					"type": "long_headline"
				},
				{
					"text": "Generated description text example.",
					"type": "description"
				}
			]
		}
		```
	* Images
		```
		{
			"final_url": "https://example.com",
			"items": [
				{
					"temporary_image_url": "https://example.com/temporary_image_url-marketing.jpg",
					"type": "marketing_image"
				},
				{
					"temporary_image_url": "https://example.com/temporary_image_url-square.jpg",
					"type": "square_marketing_image"
				},
				{
					"temporary_image_url": "https://example.com/temporary_image_url-portrait.jpg",
					"type": "portrait_marketing_image"
				}
			]
		}
		```
2. Run the following commands in the browser console:
	* To fetch text assets:
		```wp.data.dispatch('wc/gla').fetchGenAITextAssets('https://example.com');```
	* To fetch image assets:
		```wp.data.dispatch('wc/gla').fetchGenAIMediaAssets('https://example.com');```
	* To view the fetched text assets:
		```wp.data.select('wc/gla').getGenAITextAssets('https://example.com');```
	* To view the fetched image assets:
		```wp.data.select('wc/gla').getGenAIMediaAssets('https://example.com');```
3. Verify that the fetched assets match the mocked responses.

### Changelog entry

>

PR #3222, opened 2026-01-21
Implement loader logic

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-397/automatically-add-generated-assets-text-and-images-based-on-selected https://linear.app/a8c/issue/GOOWOO-410/implement-new-loader-while-waiting-for-assets-generation

1. Add a new campaign via the "Add campaign" button on the dashboard.
2. Going to the second step of the wizard, there should be a loading state where the data for the site is fetched.
3. Choose a page.
4. The loader will show up again.
5. Ensure the mocked AI data is being rendered. Mocked data can be [found here](https://github.com/woocommerce/google-listings-and-ads/pull/3207).

### Changelog entry

>


js/src/components/paid-ads/asset-group/asset-group-header/asset-group-header.js
```diff
@@ -22,8 +28,49 @@ import GenAICard from '../../gen-ai-card';
...
+			hasLoadedInitialHomepageAssetsRef.current = true;
+
+			// Load homepage assets on first render by passing `id: 0` and a `type` other than `post` or `term`.
+			// `id` is a required parameter, but it is ignored when loading homepage assets.
+			// Related: https://github.com/woocommerce/google-listings-and-ads/blob/d23bdb504bce1ed8a10a4bd92608aeb5137fbe60/src/Ads/AssetSuggestionsService.php#L210-L216
+			await fetchCampaignAssets( 0, 'homepage' );
+		}
+
+		loadAssets();
+	}, [ fetchCampaignAssets, adapter.baseAssetGroup ] );
```
ankitrox: @asvinb  Looks like `isEditing` needs to be added as dependency for this effect. Also, `adapter.baseAssetGroup` may change which will trigger the effect, can we derive `adapter.baseAssetGroup[ ASSET_FORM_KEY.FINAL_URL ]` as a boolean and add it as dependency?

ankitrox (changes requested):
Great work @asvinb! I've added one comment, rest looks good to me.

Thanks

js/src/components/paid-ads/asset-group/asset-group-header/asset-group-header.js
asvinb: Thanks for catching the `isEditing` dependency @ankitrox . I don't think we should run the effect if the final URL changes because `fetchAssets` requires and `id` and `type`, from which we fetch the existing suggestions and then we fetch the AI generated assets and then we merge everything.

ankitrox (approved):

ankitguptaindia:
QA Status: Pass ✅ 
---


**Test Result:** Loader appears as designed and works as scoped. No issues were found. 

https://github.com/user-attachments/assets/5b5bbc08-48bf-48c6-bbea-6836a664481f

PR #3169, opened 2025-12-08
Remove problematic duplicate conversion event

mukeshpanchal27:
### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/GOOWOO-348/identical-purchase-and-conversion-events-may-be-problematic-according

### Detailed test instructions:

1. Add a product to the cart and place an order. On the order confirmation page, ensure that the `conversion` event does not run.

**Note for testing:** Since the events on the order confirmation page fire only once, comment out the following lines so the event loads every time: https://github.com/woocommerce/google-listings-and-ads/blob/fix/GOOWOO-348-remove-problematic-conversion-event/src/Google/GlobalSiteTag.php#L366-L368

### Before:
```JS
<script>
gtag("event", "conversion", {
			send_to: "AW-16939825035/MsMwCLCN5M0bEIvvxI0_",
			value: 30.000000,
			currency: "USD",
			transaction_id: "135"});
</script>
<script>
gtag("event", "purchase", {
			ecomm_pagetype: "purchase",
			send_to: "AW-16939825035/MsMwCLCN5M0bEIvvxI0_",
			transaction_id: "135",
			currency: "USD",
			country: "US",
			value: 30.000000,
			new_customer: true,
			tax: 0.000000,
			shipping: 0.000000,
			delivery_postal_code: "12345",
			aw_feed_country: "US",
			aw_feed_language: "English",
			items: [{
				id: "gla_100",
				price: 15.000000,
				google_business_vertical: "retail",
				name: "Album",
				quantity: 2,
				}]});
</script>
```

### After:
```JS
<script>
gtag("event", "purchase", {
			ecomm_pagetype: "purchase",
			send_to: "AW-16939825035/MsMwCLCN5M0bEIvvxI0_",
			transaction_id: "135",
			currency: "USD",
			country: "US",
			value: 30.000000,
			new_customer: true,
			tax: 0.000000,
			shipping: 0.000000,
			delivery_postal_code: "12345",
			aw_feed_country: "US",
			aw_feed_language: "English",
			items: [{
				id: "gla_100",
				price: 15.000000,
				google_business_vertical: "retail",
				name: "Album",
				quantity: 2,
				}]});
</script>
```

### Changelog entry

> Fix - Remove problematic duplicate conversion event

PR #3234, opened 2026-01-27
Add filter to disable legacy conversion events

joemcgill:

Closes GOOWOO-348. See #3169.

This adds a new filter, `woocommerce_gla_add_legacy_conversion_event` to be able to disable legacy conversion events.

Given that removing these events entirely could have unknown side effects, this change allows users who are currently seeing issues with deduplication not working properly to disable these events while we plan a deprecation path for these events in a future release.

### Screenshots:

N/A

1. After completing a purchase, see that the `gtag("event", "conversion", {...})` script is included in the thank you page.
2. Add a filter like `add_filter( 'woocommerce_gla_add_legacy_conversion_event', '__return_false' );` to disable the feature.
3. Complete another purchase and confirm that the gtag event is not included in the thank you page.

### Changelog entry

> Update - Allow legacy conversion events to be disabled via a filter.

### DIFF
--- a/src/Google/GlobalSiteTag.php
+++ b/src/Google/GlobalSiteTag.php
@@ -371,19 +371,30 @@ class GlobalSiteTag implements Service, Registerable, Conditional, OptionsAwareI
 		$order->update_meta_data( self::ORDER_CONVERSION_META_KEY, 1 );
 		$order->save_meta_data();
 
-		$conversion_gtag_info =
-		sprintf(
-			'gtag("event", "conversion", {
-			send_to: "%s",
-			value: %f,
-			currency: "%s",
-			transaction_id: "%s"});',
-			esc_js( "{$ads_conversion_id}/{$ads_conversion_label}" ),
-			$order->get_total(),
-			esc_js( $order->get_currency() ),
-			esc_js( $order->get_id() ),
-		);
-		$this->add_inline_event_script( $conversion_gtag_info );
+		/**
+		 * Track the legacy conversion event.
+		 *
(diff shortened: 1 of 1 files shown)


src/Google/GlobalSiteTag.php
```diff
@@ -371,19 +371,30 @@ public function maybe_display_conversion_and_purchase_event_snippets( string $ad
...
+		/**
+		 * Track the legacy conversion event.
+		 *
+		 * The legacy conversion event is kept for backward compatibility with existing setups.
+		 * New implementations should rely on the 'purchase' event for conversion tracking.
+		 * In a future release, this may be removed.
+		 */
+		$add_legacy_conversion_event = apply_filters( 'woocommerce_gla_add_legacy_conversion_event', true );
+
+		if ( $add_legacy_conversion_event ) {
```
mukeshpanchal27, line 383: Not a blocker but in most places where G4W filter we use `enable` instead of `add`.

```suggestion
		$enable_legacy_conversion_event = apply_filters( 'woocommerce_gla_enable_legacy_conversion_event', true );

		if ( $enable_legacy_conversion_event ) {
```

mukeshpanchal27 (approved):
Thanks @joemcgill for the PR. The approach makes sense and pre-approve.

One question for the future though, if we get feedback from the product team and agree to remove this event entirely, would we also plan to remove this filter at that point?

joemcgill:
> If we get feedback from the product team and agree to remove this event entirely, would we also plan to remove this filter at that point?

We would either remove it completely or swap the default to `false` for folks who need to opt in to keeping the old conversion event.

PR #3252, opened 2026-02-10
[GOOWOO-453] Load meta-boxes JS only on WC Edit Order screen

jamesmorrison:
### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/GOOWOO-453/load-meta-boxes-js-asset-in-the-wc-edit-order-page-only

### Screenshots:

Screenshot 2026-02-10 at 16 25 01

### Detailed test instructions:

1. Open WP admin, navigate to WooCommerce -> Settings -> Advanced -> Features
2. Confirm / set "Order data storage" is "High-performance order storage"
3. Navigate to WooCommerce -> Home
4. Open the inspector, confirm you do NOT see "Google for WooCommerce: meta-boxes script loaded"
5. Navigate to WooCommerce -> Orders and select an order (create one if need be)
6. Open the inspector, confirm you DO see "Google for WooCommerce: meta-boxes script loaded"
7. Navigate to WooCommerce -> Settings -> Advanced -> Features
8. Set "Order data storage" to "WordPress posts storage"
9. Navigate to WooCommerce -> Home
10. Open the inspector, confirm you do NOT see "Google for WooCommerce: meta-boxes script loaded"
11. Navigate to WooCommerce -> Orders and select an order (create one if need be)
12. Open the inspector, confirm you DO see "Google for WooCommerce: meta-boxes script loaded"

### Additional details:

> Add - Added new JS file for meta boxes; conditionally enqueued on the edit order page.

PR #3254, opened 2026-02-11
[GOOWOO-454] Add new property within glaData to indicate whether source is Google.

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-454/add-new-property-within-gladata-to-indicate-whether-source-is-google

### Screenshots:

Screenshot 2026-02-11 at 10 56 54

1. On the frontend, navigate to a product that can be purchased. (Create one if need be)
2. Add the following URL param `?utm_source=google` - for example: `https://woo.test/product/example-product/` should become `https://woo.test/product/example-product/?utm_source=google`
3. Complete the checkout process
4. In WP admin, navigate to WooCommerce -> Orders
5. Locate the order in the table (should be at the top), confirm you see "Source: google" in the "Origin" column
6. Click on the order
7. Verity you see "Source: google" in the "Order Attribution" meta box (right side of page)
8. Open the inspector, select console and enter this: `glaData.orderAttributionSource` and press return
9. Confirm the value is `'google'`

### Additional details:

> Add - Added new property within glaData to indicate whether source is Google.


src/Admin/Admin.php
```diff
@@ -199,9 +199,103 @@ protected function get_assets(): array {
...
+			|| ( strpos( $screen->id, 'woocommerce_page_wc-orders--' ) === 0 )
+			|| ( 'admin_page_wc-orders' === $screen->id )
+			|| ( strpos( $screen->id, 'admin_page_wc-orders--' ) === 0 );
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended -- Reading request for order edit context only.
+		$is_edit_action = isset( $_GET['action'] )
+			&& 'edit' === sanitize_text_field( wp_unslash( $_GET['action'] ) );
+
+		if ( ! $is_wc_orders_screen || ! $is_edit_action ) {
+			return null;
+		}
```
jjgrainger: @jamesmorrison theres a duplication of logic that could be encapsulated in a method here.

src/Admin/Admin.php
```diff
@@ -199,9 +199,103 @@ protected function get_assets(): array {
...
+
+		$order_id = 0;
+		if ( isset( $_GET['id'] ) ) {
+			$order_id = absint( $_GET['id'] );
+		} elseif ( isset( $_GET['post'] ) ) {
+			$order_id = absint( $_GET['post'] );
+		}
```
jjgrainger: @jamesmorrison whats the reasoning for checking `$_GET['post']`?

src/Admin/Admin.php
```diff
@@ -199,9 +199,103 @@ protected function get_assets(): array {
...
+		if ( ! $order ) {
+			return null;
+		}
+
+		$source = $order->get_meta( '_wc_order_attribution_utm_source', true );
+		if ( $source === '' || $source === null ) {
+			return null;
+		}
+
+		return is_string( $source ) ? $source : (string) $source;
```
jjgrainger: @jamesmorrison this could be simplified


```suggestion
		return (string) $source;
```

jjgrainger (changes requested):
Thanks @jamesmorrison left a few minor comments, perhaps we should focus on getting #3252 through CR before addressing whats needed here? Thanks

src/Admin/Admin.php
jamesmorrison: @jjgrainger There are 2 ways orders can be stored depending on when WooCommerce was installed:
1. In custom tables (default for a new site) - High-performance order storage
2. In a custom post type - WordPress posts storage

The `post` check is for the WordPress posts storage.

src/Admin/Admin.php
jamesmorrison: Thanks, I committed your suggestion.

src/Admin/Admin.php
jamesmorrison: Thanks, I've updated this.

src/Admin/Admin.php
```diff
@@ -199,9 +199,101 @@ protected function get_assets(): array {
...
 	}
 
+	/**
+	 * Get the order attribution source (utm_source) for the order currently being edited.
+	 * Used only when the meta-boxes asset is loaded on the WooCommerce Edit Order screen.
+	 *
+	 * @return string|null The value persisted in the database (e.g. "google"), or null when not on order edit screen or no attribution.
+	 */
+	private function get_order_attribution_source_for_edit_screen(): ?string {
+		if ( ! $this->is_wc_orders_edit_screen() ) {
```
asvinb, line 240: @jamesmorrison Maybe i am mistaken but why we need to check again if it's the WC Orders edit screen when we are already checking when we enqueue the JS asset? 🤔

asvinb (changes requested):
@jamesmorrison I left a comment. Can you check and let me know please? Also, can you set the target branch for the PR to `feature/GOOWOO-453-load-meta-boxes-js-on-the-wc-edit-order-page-only`? When the [other PR](https://github.com/woocommerce/google-listings-and-ads/pull/3252) gets merged, the branch will automatically be set to `feature/in-product-placements`

src/Admin/Admin.php
jamesmorrison: @asvinb Thanks for checking but I believe we do need this check.

Without the is_wc_orders_edit_screen() check inside get_order_attribution_source_for_edit_screen(), that method would run on every admin page (e.g. Dashboard, Plugins, Products) and would:

1. Read $_GET['id'] / $_GET['post']
2. Resolve an order ID and call wc_get_order()
3. Read order meta

which is unnecessary and might impact performance.

asvinb: gotcha! Thanks for the clarification @jamesmorrison !

src/Admin/Admin.php
```diff
@@ -231,10 +231,44 @@ function (): bool {
...
-	 * @return bool
+	 * @return string|null The value persisted in the database (e.g. "google"), or null when not on order edit screen or no attribution.
+	 */
+	private function get_order_attribution_source_for_edit_screen(): ?string {
+		if ( ! $this->is_wc_orders_edit_screen() ) {
+			return null;
+		}
+
+		$order_id = 0;
+		if ( isset( $_GET['id'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
asvinb, line 247: @jamesmorrison Can you please add a comment indicating why we are checking for both, `$_GET['id']` and `$_GET['post']` please? Thanks!

asvinb (approved):
@jamesmorrison Code LGTM! I left a minor comment though. Can you check it out please?

src/Admin/Admin.php
jamesmorrison: @asvinb I've added a note here, please let me know if there's any further adjustments required.

jjgrainger (approved):

ankitguptaindia:
QA Status: Approved ✅ 
----

**Test Result:** 

Validated the **order attribution functionality** using the URL parameter `?utm_source=google`. Completed a full purchase flow and confirmed that the order source is correctly captured and displayed as **"Source: google"** in the **Origin column** within the Orders list and in the **Order Attribution meta box** on the order details page. Additionally verified via browser console that `glaData.orderAttributionSource` returns the expected value **'google'**. All checks are working as expected.


Screenshot 2026-03-09 at 2 14 23 PM


Screenshot 2026-03-09 at 2 14 07 PM

PR #3261, opened 2026-02-18
[GOOWOO-38] Add a message to product inventory tab to set availability date if back order enabled

jamesmorrison:
…backorder' and the availability date is not set.

Closes https://linear.app/a8c/issue/GOOWOO-38/add-a-message-to-product-inventory-tab-to-set-availability-date-if

### Screenshots:

Screenshot 2026-02-18 at 14 19 34

1. Create or edit a product so that the Product Data has: Inventory -> Stock Status is "On backorder" and Google for WooCommerce -> Availability date blank
2. On the Inventory tab, you should see the warning, highlighted in yellow: "Google requires an availability date for products on backorder. Set the Availability date in the Google for WooCommerce tab so your product can be submitted correctly."
3. Confirm the link correctly changes to the Google for WooCommerce tab
4. On the Google for WooCommerce tab, enter an Availability date (no need to save the product)
5. Navigate back to Inventory tab and confirm notice is no longer visible
6. Remove the Availability date, confirm the notification shows up again
7. Change the Inventory -> Stock Status to "In stock" / "Out of stock". In both cases, notice should not show.

### Additional details:

> Add - Added notification to alert user to missing Availability date if not set and product is On backorder


js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+			);
+		}
+	} );
+
+	updateNoticeVisibility( notice );
+}
+
+export function init() {
+	if ( document.readyState === 'loading' ) {
+		document.addEventListener( 'DOMContentLoaded', () =>
```
asvinb:
```suggestion
		document.addEventListener( 'DOMContentLoaded', initBackorderAvailabilityDateNotice );
```

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+				const tabLink = document.querySelector(
+					`.product_data_tabs a[href="#${ tabTarget }"]`
+				);
+				if ( tabLink ) {
+					tabLink.click();
+				}
+			} );
+		} );
+
+	[ 'change', 'input' ].forEach( ( ev ) => {
```
asvinb: @jamesmorrison Can we use more descriptive names for the parameters please?

```suggestion
	[ 'change', 'input' ].forEach( ( event ) => {
```

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+		document
+			.querySelectorAll(
+				'#inventory_product_data [name="_backorders"], #inventory_product_data [name="_stock_status"]'
+			)
+			.forEach( ( el ) => {
```
asvinb:
```suggestion
			.forEach( ( element ) => {
```

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+				);
+				if ( tabLink ) {
+					tabLink.click();
+				}
+			} );
+		} );
+
+	[ 'change', 'input' ].forEach( ( ev ) => {
+		document
+			.querySelectorAll(
```
asvinb: @jamesmorrison Why we need to use `querySelectorAll` instead of `querySelector` here? 🤔 Will there be a case where we'll potentially have more than 1 element?

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+
+function initBackorderAvailabilityDateNotice() {
+	const notice = document.querySelector(
+		'.gla-backorder-availability-date-notice'
+	);
+	if ( ! notice ) {
+		return;
+	}
+
+	const tabTarget =
```
asvinb: @jamesmorrison If `glaProductData` will always be available on the product data page and the script is run only on the product page, should be check for the existence of `glaProductData.glaTabTarget`?

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+		( glaProductData && glaProductData.glaTabTarget ) || 'gla_attributes';
+
+	// Clicking the link switches to the GLA tab.
+	document
+		.querySelectorAll( '.gla-availability-date-tab-link' )
+		.forEach( ( link ) => {
+			link.addEventListener( 'click', ( e ) => {
```
asvinb:
```suggestion
			link.addEventListener( 'click', ( event ) => {
```

js/src/product-attributes/backorder-availability-date-notice.js
asvinb: @jamesmorrison Why are we listening to both `change` and `input`? For radio buttons, change should be sufficient.

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+			.forEach( ( el ) => {
+				el.addEventListener( ev, () =>
+					updateNoticeVisibility( notice )
+				);
+			} );
+		const glaDateEl = document.getElementById( GLA_DATE_INPUT_ID );
+		const glaTimeEl = document.getElementById( GLA_TIME_INPUT_ID );
+		if ( glaDateEl ) {
+			glaDateEl.addEventListener( ev, () =>
+				updateNoticeVisibility( notice )
```
asvinb: @jamesmorrison We are passing the `notice` variable each time to `updateNoticeVisibility`. Can we query for the notice within the function itself?

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+		document
+			.querySelectorAll(
+				'#inventory_product_data [name="_backorders"], #inventory_product_data [name="_stock_status"]'
+			)
+			.forEach( ( el ) => {
+				el.addEventListener( ev, () =>
+					updateNoticeVisibility( notice )
+				);
+			} );
+		const glaDateEl = document.getElementById( GLA_DATE_INPUT_ID );
```
asvinb: @jamesmorrison I think the date and time fields are essential for that functionality. If they are not present, let's bail out early, similar to when there's no notice.

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+		const glaTimeEl = document.getElementById( GLA_TIME_INPUT_ID );
+		if ( glaDateEl ) {
+			glaDateEl.addEventListener( ev, () =>
+				updateNoticeVisibility( notice )
+			);
+		}
+		if ( glaTimeEl ) {
+			glaTimeEl.addEventListener( ev, () =>
```
asvinb: When we remove the `notice` parameter from `updateNoticeVisibility`, we can update this line to:

```suggestion
			glaTimeEl.addEventListener( event, updateNoticeVisibility );
```

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+	if ( checked ) {
+		return checked.value;
+	}
+	return '';
+}
+
+function isBackorderSelected() {
+	const backorders = getBackorderValue();
+	const stockStatus = getStockStatusValue();
+	return (
```
asvinb:
```suggestion
	return (
    ['yes', 'notify'].includes(backorders) ||
    stockStatus === 'onbackorder'
);)
	
```

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
...
+ * Internal dependencies
+ */
+import { glaProductData } from '~/constants';
+
+const GLA_DATE_INPUT_ID = 'gla_attributes_availabilityDate_date';
+const GLA_TIME_INPUT_ID = 'gla_attributes_availabilityDate_time';
+
+function getBackorderValue() {
+	const backordersSelect = document.querySelector(
+		'#inventory_product_data select[name="_backorders"]'
```
asvinb: @jamesmorrison  Could we extract `#inventory_product_data` into a shared constant? It shows up in a few places, so centralising it would make future selector changes easier to manage.

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,128 @@
+/**
```
asvinb: @jamesmorrison I just feel this comment is not in the right place. Can you move it further down please?

js/src/product-attributes/index.scss
```diff
@@ -48,3 +48,22 @@
...
+// More specific than #inventory_product_data .notice so our background/border apply.
+#woocommerce-product-data #inventory_product_data .gla-backorder-availability-date-notice {
+	margin: $grid-unit-20 0;
+	padding: $grid-unit-15 $grid-unit-15 $grid-unit-15 $grid-unit-20;
+	border-left: 4px solid #dba617;
+	background: $gla-color-yellow-0;
+	color: $gray-900;
+
+	a {
+		color: #2271b1;
```
asvinb:
```suggestion
		color: $blue-50;
```

js/src/product-attributes/index.scss
```diff
@@ -48,3 +48,22 @@
 	margin: $grid-unit-10 0;
 	max-width: 100%;
 }
+
+// Backorder availability date notice (Inventory tab, classic product editor).
+// More specific than #inventory_product_data .notice so our background/border apply.
+#woocommerce-product-data #inventory_product_data .gla-backorder-availability-date-notice {
+	margin: $grid-unit-20 0;
+	padding: $grid-unit-15 $grid-unit-15 $grid-unit-15 $grid-unit-20;
+	border-left: 4px solid #dba617;
```
asvinb:
```suggestion
	border-left: 4px solid $alert-yellow;
```

asvinb (changes requested):
@jamesmorrison Functionality looks good! A few improvement opportunities worth considering:
The main one is duplicated DOM queries. For example, `document.getElementById( GLA_DATE_INPUT_ID )` is called in both `hasGlaAvailabilityDate` and the event binding setup. Wrapping everything in a factory/closure function would let us query elements like `notice`, `glaDateEl`, and `glaTimeEl` once at initialization and share them across the inner functions, rather than re-querying on each call.
Happy to discuss if you'd like to explore that direction!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/index.scss
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/index.scss
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated!

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb Thanks, have updated

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb I believe we do need `querySelectorAll`; there are 2 inputs, each with 3 selectors; with `querySelector` only the first selector for each input would be picked up. I've not changed this but please let me know if I misunderstood.

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,126 @@
...
+	const backorders = getBackorderValue();
+	const stockStatus = getStockStatusValue();
+	return (
+		[ 'yes', 'notify' ].includes( backorders ) ||
+		stockStatus === 'onbackorder'
+	);
+}
+
+function hasGlaAvailabilityDate() {
+	const dateInput = document.getElementById( GLA_DATE_INPUT_ID );
```
asvinb: @jamesmorrison Please check my [comment here](https://github.com/woocommerce/google-listings-and-ads/pull/3261#pullrequestreview-3897539446) about having a closure function or an object where we can query those elements once.

js/src/product-attributes/backorder-availability-date-notice.js
```diff
@@ -0,0 +1,126 @@
...
+			element.addEventListener( 'change', updateNoticeVisibility );
+		} );
+
+	glaDateEl.addEventListener( 'change', updateNoticeVisibility );
+	glaTimeEl.addEventListener( 'change', updateNoticeVisibility );
+
+	updateNoticeVisibility();
+}
+
+export function init() {
```
asvinb:
```suggestion
export default function init() {
```

js/src/product-attributes/backorder-availability-date-notice.test.js
```diff
@@ -0,0 +1,239 @@
+/**
```
asvinb: @jamesmorrison I think it's best we have E2E tests for this functionality instead of unit tests. What do you think? I see we are basically trying to recreate the UI here.

js/src/product-attributes/backorder-availability-date-notice.test.js
```diff
@@ -0,0 +1,239 @@
...
+import { init } from './backorder-availability-date-notice';
+
+jest.mock( '~/constants', () => ( {
+	glaProductData: {
+		glaTabTarget: 'gla_attributes',
+	},
+} ) );
+
+const NOTICE_CLASS = 'gla-backorder-availability-date-notice';
+const DATE_ID = 'gla_attributes_availabilityDate_date';
```
asvinb: @jamesmorrison Why we need to define those again?

.eslintrc.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
 	},
 	globals: {
 		getComputedStyle: 'readonly',
+		MouseEvent: 'readonly',
```
asvinb: @jamesmorrison What does that do exactly?

src/Admin/Admin.php
```diff
@@ -188,7 +188,11 @@ protected function get_assets(): array {
 
 		$product_condition = function () {
 			$screen = get_current_screen();
-			return ( null !== $screen && 'product' === $screen->id );
+			if ( null === $screen ) {
+				return false;
+			}
+			// Product list: screen id is 'product'. Classic product edit: screen id is 'post', post_type is 'product'.
```
asvinb: @jamesmorrison Why we are checking for product list, when the screen id is `product`?

asvinb (changes requested):

js/src/product-attributes/backorder-availability-date-notice.js
asvinb: @jamesmorrison Actually i was referring to this line `querySelectorAll( '.gla-availability-date-tab-link' )`. If there's only one link, why use `querySelectorAll` 🤔

jamesmorrison: @asvinb Ok, sorry I misunderstood; I've updated `gla-availability-date-tab-link` to use `querySelector`

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb I've updated this; `hasGlaAvailabilityDate` (and `updateNoticeVisibility`) are now defined inside `initBackorderAvailabilityDateNotice`

js/src/product-attributes/backorder-availability-date-notice.js
jamesmorrison: @asvinb This has been updated

js/src/product-attributes/backorder-availability-date-notice.test.js
jamesmorrison: @asvinb These were defined again because the tests weren't working - but per the note [here](https://github.com/woocommerce/google-listings-and-ads/pull/3261#discussion_r2912589817) I removed the unit test and replaced with E2E

src/Admin/Admin.php
jamesmorrison: @asvinb This was an old / outdated comment which I've updated; functionally the file hasn't been changed though

js/src/product-attributes/backorder-availability-date-notice.test.js
jamesmorrison: @asvinb I see your point; I've removed the unit test and replaced with E2E.

.eslintrc.js
jamesmorrison: @asvinb This was needed for the (now removed) test, I've removed this.

asvinb (changes requested):
@AlejandroPerezMartin I'am wondering whether we can use the `Notice` component instead of adding the notice via PHP. Can you check and let me know please?

asvinb (changes requested):
## Summary
Solid, well-scoped feature with strong e2e coverage, but the PHP-side gating logic (`should_show_notice()` / `get_current_product()`) is fully wired up, unit-tested with 6 test cases, and then never called — the container renders unconditionally on every product edit screen. 🔄 Request changes.

## Critical 🔴

- **Correctness/dead-code** · `src/Admin/Product/BackorderAvailabilityDateNotice.php:401-432` · `render_notice()` unconditionally echoes the notice container — it never calls `should_show_notice()` or `get_current_product()`. Those two methods (and the "applicable product types" / attribute-manager gating they implement) are completely dead: nothing in this class or its hooks invokes them. `BackorderAvailabilityDateNoticeTest.php` reflection-invokes `should_show_notice` directly and passes, giving false confidence that this logic runs in production — it doesn't. Practical effect: the container div (and the lazy-loaded React chunk behind it) mounts on **every** classic/block product edit screen once Merchant Center setup is complete, including for product types `AvailabilityDate::get_applicable_product_types()` was clearly meant to exclude (e.g. grouped/external products). Fix: have `render_notice()` call `get_current_product()`, bail if null, and skip the `echo` entirely when `! $this->should_show_notice( $product )` — restoring the intended server-side gate instead of relying solely on the client-side `isBackorderSelected()` check (which has no access to product type or the saved attribute value).

## Improvements 🟡

- `src/Admin/Product/BackorderAvailabilityDateNotice.php:334` · `private const GLA_TAB_TARGET = 'gla_attributes'` is defined but never referenced — the actual value is duplicated as a raw string literal in `src/Admin/Admin.php:292` (`'glaTabTarget' => 'gla_attributes'`) and again as a JS fallback in `js/src/product-attributes/backorder-availability-date-notice/index.js:173` (`glaProductData.glaTabTarget || 'gla_attributes'`). Three independent copies of the same string with no single source of truth — a future tab-id rename will silently break one of them. Have `Admin::get_assets()` reference the constant (or move it somewhere both classes can share) instead of re-typing the literal.

- `src/Admin/Product/BackorderAvailabilityDateNotice.php:386-395` vs `src/Admin/Admin.php:280-284` · The "is this the product edit screen (classic or block)" check (`'product' === $screen->id || ( 'post' === $screen->id && 'product' === $screen->post_type )`) is duplicated verbatim across two files, both touched in this same PR. Extract to a shared static helper (e.g. on `Admin` or a small screen-detection utility) so the two copies can't drift apart on the next edge case (e.g. a future block-editor screen-id change).

- Test coverage · No Jest unit tests were added for `js/src/product-attributes/backorder-availability-date-notice/utils.js` (`isBackorderSelected`, `getFieldValue`) or for the `BackorderAvailabilityDateNotice` component itself — only e2e specs exist. These are pure, easily-testable functions with clear branches (select vs. radio input, `yes`/`notify` backorder values vs. `onbackorder` stock status) that deserve co-located `*.test.js` coverage per repo convention, independent of the (slower, coarser) e2e suite.

## Nits 🔵
- `src/Admin/Product/BackorderAvailabilityDateNotice.php:360` · `PRODUCT_ATTRIBUTES_STYLE_HANDLE = 'gla-product-attributes-css'` is coupled to `Admin::get_assets()` only via a comment ("must match Admin::get_assets()"). Consider exposing the handle as a public constant on `Admin` so the coupling is compiler-visible instead of comment-only.

PR #3263, opened 2026-02-19
[GOOWOO-474] Performance: Avoid uncached Ads API request in NotificationManager

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-474/performance-avoid-uncached-ads-api-request-in-notificationmanager

### Screenshots:

Screenshot 2026-02-19 at 15 42 07

Before clearing transient cache:
Screenshot 2026-02-23 at 13 57 25

After clearing transient cache (i.e. the data is looked up and stored):
Screenshot 2026-02-23 at 13 58 05

1. In terminal run: `wp transient get gla_ads_highest_spend_campaign` - the transient should not exist at this point so expect the output to be: `Warning: Transient with key "gla_ads_highest_spend_campaign" is not set.`
2. Open WP Admin, navigate to Marketing -> Google for WooCommerce and select any campaign (or create one if need be). This process should set the transient.
3. In terminal run: `wp transient get gla_ads_highest_spend_campaign` You should see array data set now, for example:

```
array (
  'campaign' => 
  array (
    'id' => 23349407739,
    'name' => 'Campaign 2025-12-11 15:09:02',
    'status' => 'enabled',
    'type' => 'performance_max',
    'targeted_locations' => 
    array (
      0 => 'GB',
    ),
    'eu_political_advertising_confirmation' => false,
    'amount' => 14.0,
    'country' => 'GB',
  ),
)
```

### Changelog entry

> Fix - Cache GAQL query for highest spend campaign.

### DIFF
--- a/src/API/Google/AdsCampaign.php
+++ b/src/API/Google/AdsCampaign.php
@@ -320,12 +320,12 @@ class AdsCampaign implements ContainerAwareInterface, OptionsAwareInterface {
 				$operations[] = $this->edit_operation( $campaign_id, $campaign_fields );
 			}
 
+			$this->container->get( TransientsInterface::class )->delete( TransientsInterface::ADS_HIGHEST_SPEND_CAMPAIGN );
+
 			if ( ! empty( $operations ) ) {
-				$this->container->get( TransientsInterface::class )->delete( TransientsInterface::ADS_HIGHEST_SPEND_CAMPAIGN );
 				return $this->mutate( $operations ) ?: $campaign_id;
 			}
 
-			$this->container->get( TransientsInterface::class )->delete( TransientsInterface::ADS_HIGHEST_SPEND_CAMPAIGN );
 			return $campaign_id;
 		} catch ( ApiException $e ) {
 			do_action( 'woocommerce_gla_ads_client_exception', $e, __METHOD__ );
@@ -424,7 +424,7 @@ class AdsCampaign implements ContainerAwareInterface, OptionsAwareInterface {
 		$transients->set(
 			TransientsInterface::ADS_HIGHEST_SPEND_CAMPAIGN,
 			[ 'campaign' => $result ],
-			HOUR_IN_SECONDS
+			HOUR_IN_SECONDS * 12
 		);
 
 		return $result;
--- a/src/API/Google/AdsCampaign.php
+++ b/src/API/Google/AdsCampaign.php
@@ -253,8 +253,10 @@ class AdsCampaign implements ContainerAwareInterface, OptionsAwareInterface {
 				$this->campaign_label->assign_label_to_campaign_by_label_name( $campaign_id, $params['label'] );
 			}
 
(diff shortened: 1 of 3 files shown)


src/API/Google/AdsCampaign.php
```diff
@@ -406,6 +420,14 @@ function ( $highest, $campaign ) {
 			},
 			[]
 		);
+
+		$transients->set(
+			TransientsInterface::ADS_HIGHEST_SPEND_CAMPAIGN,
+			[ 'campaign' => $result ],
+			HOUR_IN_SECONDS
```
jjgrainger: @jamesmorrison I think we can increase this to something like 12 hours as we handling deleting the transient when campaigns are updated/deleted.

The method is only used in the [`NotificationManager`](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Menu/NotificationManager.php#L206) for checking recommendations for performance max campaigns. Recommendations are stored for [12 hours also](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Ads/AdsRecommendationsService.php#L114).

src/API/Google/AdsCampaign.php
```diff
@@ -319,9 +321,11 @@ public function edit_campaign( int $campaign_id, array $params ): int {
 			}
 
 			if ( ! empty( $operations ) ) {
+				$this->container->get( TransientsInterface::class )->delete( TransientsInterface::ADS_HIGHEST_SPEND_CAMPAIGN );
 				return $this->mutate( $operations ) ?: $campaign_id;
 			}
 
+			$this->container->get( TransientsInterface::class )->delete( TransientsInterface::ADS_HIGHEST_SPEND_CAMPAIGN );
```
jjgrainger: @jamesmorrison if we're deleting the transient regardless of if there are mutate operations would it make sense to delete just before the `if ( ! empty( $operations ) )`, rather than duplicate here?

jjgrainger (changes requested):
Thanks @jamesmorrison looks good, left some minor feedback

src/API/Google/AdsCampaign.php
jamesmorrison: Thanks @jjgrainger - this has been updated to 12 hours.

src/API/Google/AdsCampaign.php
jamesmorrison: I've corrected this @jjgrainger - sorry, I initially only had this removing the transient when `$operations` was not empty, then realised we always need it.

jjgrainger (approved):

ankitguptaindia:
QA Status: Verified ✅ 
----

The arreay appears as per the highest spend campaign. Tested with various budgets and result is appeaing as expected. 

Screenshot 2026-03-23 at 6 09 29 PM

Screenshot 2026-03-23 at 6 26 17 PM

ankitguptaindia (approved):

PR #3286, opened 2026-03-06
GOOWOO-227: Fix Shipping Tab not loading

AlejandroPerezMartin:
…fetching

Closes https://linear.app/a8c/issue/GOOWOO-227/shipping-tab-not-loading

Adds a loaded state to indicate the component that all the data has been fetched and only render the shipping content tab when all data has resolved fetching.

> Before switching to this branch, ensure you can reproduce the error so follow the steps using the `develop` branch and then switch to this one to spot the difference.

1. Go to the Shipping Tab under the Google Listing plugin (`/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fshipping`)
2. Open Devtools and go to the Network tab, once there find the request to `/wp-json/wc/gla/mc/settings?_locale=user`
3. Right click on it and select Override content and add the following:
> Here's a guide that describes how to override this content https://developer.chrome.com/blog/devtools-tips-34.

Screenshot 2026-03-06 at 13 13 13

```json
{
    "shipping_rate": null,
    "shipping_time": null,
    "tax_rate": "destination",
    "shipping_rates_count": 0
}
```
4. When you refresh the page the loading spinner should be there forever and no content will load.
5. Now switch to this branch, restart the build/start command and refresh this page again keeping the overrides
6. You should now get the content of the Shipping tab loading properly

### Changelog entry

> Fix - Fix issue rendering Shipping tab when shipping options were not set

### DIFF
--- a/js/src/hooks/useTargetAudienceFinalCountryCodes.js
+++ b/js/src/hooks/useTargetAudienceFinalCountryCodes.js
@@ -31,13 +31,19 @@ import useMCCountries from '~/hooks/useMCCountries';
  *
  */
 const useTargetAudienceFinalCountryCodes = () => {
-	const { data: supportedCountries, isResolving: countriesLoading } =
-		useMCCountries();
+	const {
+		data: supportedCountries,
+		isResolving: countriesLoading,
+		hasFinishedResolution: countriesLoaded,
+	} = useMCCountries();
 
 	function mapSelect( select ) {
-		const { getTargetAudience, isResolving } = select( STORE_KEY );
+		const { getTargetAudience, isResolving, hasFinishedResolution } =
+			select( STORE_KEY );
 		const storedTargetAudience = getTargetAudience();
 		const targetAudienceLoading = isResolving( 'getTargetAudience' );
+		const targetAudienceLoaded =
+			hasFinishedResolution( 'getTargetAudience' );
 
 		/**
 		 * Flag to indicate that the data is loading.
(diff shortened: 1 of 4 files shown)


js/src/hooks/useTargetAudienceFinalCountryCodes.js
```diff
@@ -31,20 +31,34 @@ import useMCCountries from '~/hooks/useMCCountries';
...
 		const targetAudienceLoading = isResolving( 'getTargetAudience' );
+		const targetAudienceLoaded =
+			hasFinishedResolution( 'getTargetAudience' );
 
 		/**
 		 * Flag to indicate that the data is loading.
 		 *
 		 * @type {boolean}
 		 */
 		const loading = targetAudienceLoading || countriesLoading;
```
asvinb: @AlejandroPerezMartin Let’s remove the `loading` property. I checked the codebase and we’re not using it when consuming this hook. Now that we’ve added the `loaded` property, it would be better to rely on that instead. Using loading could initially be false and then switch to true, which may cause a brief UI flicker or render an incorrect state.

asvinb (changes requested):
Thanks @AlejandroPerezMartin . The code looks good to me 👍
There’s just one improvement we can make. I left a comment about removing the `loading` property. In addition, I noticed that the "Save changes" button is enabled even when no countries or shipping rates are selected.
Let’s disable the "Save changes" button when nothing is selected, or when the form is not valid. Thanks!

asvinb (approved):
Thanks @AlejandroPerezMartin LGTM!

ankitguptaindia:
QA Status: Approved ✅ 
----


**Before Fix:**

https://github.com/user-attachments/assets/0d83bad7-d797-49b6-9acb-0b0a2e9c324b

**After Fix:**

https://github.com/user-attachments/assets/75f05556-33d0-4c2d-941a-577486ef7366

ankitguptaindia (approved):

PR #3301, opened 2026-03-17
Add fallback margin when Woocommerce header is present

AlejandroPerezMartin:

Closes https://linear.app/a8c/issue/GOOWOO-255/header-settings-tabs-overlapped-by-woocommerce-layout-header

This hook https://github.com/woocommerce/woocommerce-admin/blob/95c487247416ab34eb8e492b984e2b068618e0d3/client/header/index.js#L92-L118 sets the margin dynamically and due to some race conditions this can always be set to 0 causing the header to overlap the settings tabs.
This adds a fallback margin to the #wpbody when the header is present so that even if the hook wrongly sets the top separation to 0, it still shows a gap of 60px.

### Screenshots:

image

1. Complete onboarding.
2. Go to any tab of the plugin settings.
3. Return to the **Settings** tab.
4. Click the Edit button in the Contact Information section of the store address.
5. Navigate back to **Edit Store Address** by clicking the "<" button → works as expected.
6. Go back to **Edit Address** again, and this time update the address by clicking the **Update Store Address** button.
7. Navigate back to **Edit Store Address** by clicking the "<" button → header tabs hidden partially by header as shown in screenshot.
8. Occasionally, after repeating the above steps, the header settings tabs overlap with `.woocommerce-layout__header`.

### Changelog entry

> Fix: Adds a fallback margin to the top of the settings page to avoid overlapping


js/src/pages/settings/index.js
```diff
@@ -46,6 +46,9 @@ const Settings = () => {
 				getReconnectAccountUrl( API_RESPONSE_CODES.GOOGLE_DISCONNECTED )
 			);
 		}
+
+		// Fix for missing margin top on settings pages, this value should match the height of .woocommerce-layout__header
+		document.querySelector( '#wpbody' ).style.marginBlockStart = '60px';
```
asvinb, line 84: @AlejandroPerezMartin I think the issue is present across all plugin pages, not just the `Settings` page. We should aim for a solution that works globally unless you’ve confirmed that this is isolated to the `Settings` page.

asvinb (changes requested):

PR #3308, opened 2026-03-18
Update data store to include new  property.

ankitrox:

Closes: https://linear.app/a8c/issue/GOOWOO-495/update-data-store-to-include-new-cyo-incentives-property

1. Go to **Marketing > Google for WooCommerce**

2. Make sure that the endpoint `wp-json/wc/gla/ads/incentives` is mocked using network tools or extensions like tweak, the response can be same as the one mentioned in this [issue](https://linear.app/a8c/issue/GOOWOO-492/implement-endpoint-to-get-incentives).

3. Open browser console and run this command.

```js
wp.data.select('wc/gla').getCYOIncentives();
```

4. Make sure that it triggers a network request to the endpoint for the first time.

5. Run the same command for a second time, it should log the incentives array which is available in the sample response mentioned above.

### Changelog entry

>


js/src/data/resolvers.js
```diff
@@ -775,6 +776,28 @@ export function* getAdsRecommendations( types, campaign_id = null ) {
...
 
+export function* getCYOIncentives() {
+	try {
+		const response = yield apiFetch( {
+			path: `${ API_NAMESPACE }/ads/incentives`,
+		} );
+
+		yield receiveCYOIncentives( response );
+	} catch ( error ) {
+		if ( error?.status === 404 || error?.data?.status === 404 ) {
```
asvinb: @ankitrox Any reason why we are silencing `404` errors please?

js/src/hooks/useCYOIncentives.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Internal dependencies
+ */
+import useAppSelectDispatch from './useAppSelectDispatch';
```
asvinb: @ankitrox Why are we using `useAppSelectDispatch`? 🤔

asvinb (changes requested):

js/src/hooks/useCYOIncentives.js
ankitrox: @asvinb I've kept `useAppSelectDispatch` so that it can return the resolution metadata like loading state etc. We are not sure whether the consumer of the hook will require it, so kept it there.

Please let me know if you want it to be replaced by `useSelect`.

Thank you.

js/src/data/resolvers.js
ankitrox: Thanks for spotting this, I kept there temporarily because the endpoint is not yet implemented, but removed it now because we can mock the endpoint in browser tools.

js/src/hooks/useCYOIncentives.js
asvinb: @ankitrox Yes, let's use `useSelect` since we won't need to invalidate the resolution or need the `isResolving` property. Can you return the `hasFinishedResolution` from the hook as well please? Thanks!

ankitrox: Updated it as suggested.

js/src/hooks/useCYOIncentives.js
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { STORE_KEY } from '~/data/constants';
+
+/**
+ * Custom hook to retrieve CYO incentives from the store.
+ *
+ * @return {Object|null} The CYO incentives. It will be `null` if not yet fetched or fetched but doesn't exist.
```
asvinb: @ankitrox The comment is wrong here since we are returning an object containing the `data` and `hasFinishedResolution` properties.

js/src/hooks/useCYOIncentives.test.js
```diff
@@ -0,0 +1,105 @@
...
+			data: incentives,
+			hasFinishedResolution: true,
+		} );
+	} );
+
+	it( 'returns payload with null data when incentives are not available', () => {
+		useSelect.mockImplementation( ( cb ) =>
+			cb( () => ( {
+				getCYOIncentives: () => null,
+				hasFinishedResolution: () => false,
```
asvinb: @ankitrox `hasFinishedResolution` should return `true` if the request has been sent and there's no incentive data.

js/src/hooks/useCYOIncentives.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+		const { result } = renderHook( () => useCYOIncentives() );
+
+		expect( result.current ).toEqual( {
+			data: incentives,
+			hasFinishedResolution: true,
+		} );
+	} );
+
+	it( 'returns payload with null data when incentives are not available', () => {
```
asvinb: @ankitrox I see this test is quite similar to the one above. We are mocking data and ensuring the same data gets returned. 🤔 I feel we are testing the same functionality with different sets of data.

js/src/hooks/useCYOIncentives.js
asvinb: @ankitrox Since we have the response can we add the type def for an incentive please and then update the JSDocs to include it?

asvinb (changes requested):

js/src/hooks/useCYOIncentives.test.js
ankitrox: Thanks @asvinb, I think we do not need two tests here as we are mocking certain things and testing the same things.

I have modified the test and added a single once so that we check for whether the intended selectors are called with expected parameters and number of times. Also testing whether the expected incentive data is being returned.

js/src/hooks/useCYOIncentives.js
ankitrox: Added typedef and updated the docblock.

js/src/hooks/useCYOIncentives.js
```diff
@@ -0,0 +1,55 @@
...
+ * @property {CYOIncentiveAmount} requiredAmount The minimum spend amount required to qualify for the incentive.
+ */
+
+/**
+ * @typedef {Object} CYOIncentive
+ * @property {number} id The unique identifier for the incentive.
+ * @property {string} type The incentive type. Example: 'ACQUISITION'.
+ * @property {'high'|'medium'|'low'} offer The offer tier.
+ * @property {string} termsAndConditionsUrl URL to the terms and conditions for this incentive.
+ * @property {CYOIncentiveRequirements} requirements The spend requirements to qualify for this incentive.
```
asvinb:
```suggestion
 * @property {CYOIncentiveRequirement} requirement The spend requirement to qualify for this incentive.
```
@ankitrox It's singular

js/src/hooks/useCYOIncentives.js
```diff
@@ -0,0 +1,55 @@
...
+import { STORE_KEY } from '~/data/constants';
+
+/**
+ * @typedef {Object} CYOIncentiveAmount
+ * @property {string} currencyCode The ISO 4217 currency code. Example: 'USD'.
+ * @property {string} units The amount in the major unit of the currency. Example: '1800'.
+ */
+
+/**
+ * @typedef {Object} CYOIncentiveRequirements
```
asvinb: @ankitrox Current typedef is as such:
```
requirement: {
  awardAmount: CYOIncentiveAmount,
  requiredAmount: CYOIncentiveAmount
}
```

However the response defined in https://linear.app/a8c/issue/GOOWOO-492/implement-endpoint-to-get-incentives#example-response-from-rest-api-c7261a53 is as follows:
```
requirement: {
  spend: {
    awardAmount: CYOIncentiveAmount,
    requiredAmount: CYOIncentiveAmount
  }
}
```

js/src/hooks/useCYOIncentives.js
```diff
@@ -0,0 +1,55 @@
...
+ * @property {{ awardAmount: CYOIncentiveAmount }} spend The spend requirement, including the award amount earned upon meeting it.
+ * @property {CYOIncentiveAmount} requiredAmount The minimum spend amount required to qualify for the incentive.
+ */
+
+/**
+ * @typedef {Object} CYOIncentive
+ * @property {number} id The unique identifier for the incentive.
```
asvinb: @ankitrox It's a string in the sample response.

```suggestion
 * @property {string} id The unique identifier for the incentive.
```

js/src/data/test/reducer.test.js
```diff
@@ -963,6 +964,31 @@ describe( 'reducer', () => {
...
+			const action = {
+				type: TYPES.RECEIVE_CYO_INCENTIVES,
+				cyoIncentives: {
+					incentives: [
+						{
+							id: 123,
+							type: 'new_customer_offer',
+						},
+					],
+					meta: {
```
asvinb: @ankitrox Where does `meta` come from please? I don't see it in any sample responses.

js/src/data/test/reducer.test.js
```diff
@@ -963,6 +964,31 @@ describe( 'reducer', () => {
...
+		const path = 'ads.cyo_incentives';
+
+		it( 'should receive CYO incentives data in ads object', () => {
+			const action = {
+				type: TYPES.RECEIVE_CYO_INCENTIVES,
+				cyoIncentives: {
+					incentives: [
+						{
+							id: 123,
+							type: 'new_customer_offer',
```
asvinb: @ankitrox the `type` should be `ACQUISITION` which will reflect more closely what the API will return. We are only concerned by the `ACQUISITION` type.

js/src/data/test/reducer.test.js
```diff
@@ -963,6 +964,31 @@ describe( 'reducer', () => {
 		} );
 	} );
 
+	describe( 'CYO Incentives', () => {
+		const path = 'ads.cyo_incentives';
+
+		it( 'should receive CYO incentives data in ads object', () => {
+			const action = {
+				type: TYPES.RECEIVE_CYO_INCENTIVES,
+				cyoIncentives: {
```
asvinb, line 973: @ankitrox Can we match the payload as close as possible with the real format please?

js/src/hooks/useCYOIncentives.test.js
```diff
@@ -0,0 +1,102 @@
...
+} ) );
+
+describe( 'useCYOIncentives', () => {
+	const incentives = [
+		{
+			id: 123,
+			type: 'ACQUISITION',
+			offer: 'high',
+			termsAndConditionsUrl: 'https://example.com/terms-1',
+			requirements: {
```
asvinb:
```suggestion
			requirement: {
```

asvinb (changes requested):

asvinb (approved):

ankitguptaindia:
QA Status: Verified ✅ 
---


**Environment:** 

- WordPress: 7.0-RC2-62197
- PHP: 8.4.18
- Server: nginx/1.26.1
- Browser: Chrome 146.0.0.0
- OS: macOS
- MU Plugins: None activated
- Plugins:
  * Google for WooCommerce 3.6.1
  * WooCommerce 10.7.0-rc.1




**Test Result:**
Tested incentives fetch using mocked endpoint `wp-json/wc/gla/ads/incentives`.

* First execution of `getCYOIncentives()` triggered API request ✅
* Response data matched mocked sample ✅


Screenshot 2026-04-15 at 9 13 51 PM

PR #3316, opened 2026-03-20
Add CYOI picker component.

ankitrox:

Closes: https://linear.app/a8c/issue/GOOWOO-488/build-the-cyoincentivepicker-component

### Screenshots:

Screenshot 2026-03-23 at 5 25 48 PM

**Note:** Make sure that the endpoint is overridden as mentioned [here](https://github.com/woocommerce/google-listings-and-ads/pull/3308).

1. Setup the plugin by initiating the onboarding process. Make sure that the billing is setup to `Approved` using mu-plugin.
2. In step 3 `Create a campaign`, you should be able to see the `CYOIPicker` component as shown in screenshot.
3. Note that in this issue, radio buttons are not selectable. Those would be selectable in [GOOWOO-489](https://linear.app/a8c/issue/GOOWOO-489/integrate-cyoincentivepicker-within-setuppaidads).

### Changelog entry

>


js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,108 @@
...
+		hasFinishedResolution &&
+		incentives?.length > 0 &&
+		billingStatus?.status === GOOGLE_ADS_BILLING_STATUS.APPROVED;
+
+	if ( ! shouldDisplay ) {
+		return null;
+	}
+
+	return (
+		<div className="gla-cyoi-section">
```
asvinb: @ankitrox Let's add the class name directly to the `Section` component which should achieve the same result.

js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,108 @@
...
+ */
+import { __, sprintf } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Section from '~/components/section';
+import Subsection from '~/components/subsection';
+import CYOIRadioControl from './cyoi-radio-control';
+import styles from './cyo-incentive-picker.module.scss';
```
asvinb: We’re currently using two different styling approaches here: CSS Modules (`cyo-incentive-picker.module.scss`) and a standard SCSS import (`index.scss` on line 17).

Since CSS Modules aren’t widely used in the codebase at the moment, it would be better to keep things consistent and stick with the standard SCSS import pattern here. Could you please use index.scss instead?

js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,108 @@
...
+import useCYOIncentives from '~/hooks/useCYOIncentives';
+import useGoogleAdsAccountBillingStatus from '~/hooks/useGoogleAdsAccountBillingStatus';
+import useAdsCurrency from '~/hooks/useAdsCurrency';
+import { GOOGLE_ADS_BILLING_STATUS } from '~/constants';
+import './index.scss';
```
asvinb: The SCSS file should be similar to the component file name which in this case should be:
`cyo-incentive-picker.scss`

js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,108 @@
...
+						{ __(
+							'New advertisers are eligible for free Google Ads credits.',
+							'google-listings-and-ads'
+						) }
+					</p>
+				}
+			>
+				<Section.Card>
+					<Section.Card.Body className="gla-cyoi-section__card-body">
+						<div>
```
asvinb: @ankitrox Adding/removing that `div` doesn't make any visual difference.

js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,108 @@
...
+									<div key={ id } className={ styles.row }>
+										<CYOIRadioControl
+											label={ rewardAmount }
+											offer={ offer }
+											requirement={ requirement }
+											value={ id }
+										/>
+										<div className={ styles.option }>
+											{ sprintf(
+												/* translators: %s: amount in dollars */
```
asvinb: @ankitrox The comment is slightly misleading since it's the amount in the users currency.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+import classnames from 'classnames';
+import { RadioControl } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import useAdsCurrency from '~/hooks/useAdsCurrency';
+import styles from './cyo-incentive-picker.module.scss';
+
+const CYOIRadioControl = ( {
```
asvinb: @ankitrox Missing JSDocs

asvinb: @ankitrox Also I see only the following props are being passed to `CYOIRadioControl`:
* label
* offer
* requirement
* value

We can thus remove the `selected`, `className` and the props spreading since they are not used.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+	className,
+	...rest
+} ) => {
+	const { formatAmount } = useAdsCurrency();
+
+	return (
+		// eslint-disable-next-line jsx-a11y/label-has-associated-control
+		<label className={ styles.radioGroup }>
+			<RadioControl
+				{ ...rest }
```
asvinb: @ankitrox See above

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+				label={ formatAmount( label ) }
```
asvinb: The `label` variable/prop name is misleading in this context, especially when passed to `formatAmount`. I’d expect the parameter to be named something like `amount` or `value` instead. Semantically, `label` suggests a `string`, so passing it to `formatAmount` makes it look like we’re formatting text rather than a numeric value, which feels incorrect.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+				className={ classnames( className ) }
+				checked={ selected === value }
```
asvinb: The `selected` prop is never passed to the component. And `selected` seems to be a `boolean`

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+				options={ [ { value } ] }
+				hideLabelFromVision
+				help=""
```
asvinb: Can we not remove the `help` prop entirely?

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+	return (
+		// eslint-disable-next-line jsx-a11y/label-has-associated-control
+		<label className={ styles.radioGroup }>
+			<RadioControl
+				{ ...rest }
+				label={ formatAmount( label ) }
+				className={ classnames( className ) }
+				checked={ selected === value }
+				options={ [ { value } ] }
+				hideLabelFromVision
```
asvinb: @ankitrox Let's have `boolean` props at the end of the props list please.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+				help=""
+				aria-label={ value }
```
asvinb: @ankitrox Do we need aria-label here since there's the `label` prop we are using

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,43 @@
...
+			/>
+			<span className={ classnames( styles.option ) }>
+				{ formatAmount( label ) }
```
asvinb: @ankitrox See my earlier comment about the `label` parameter.

js/src/components/cyo-incentive-picker/index.scss
```diff
@@ -0,0 +1,7 @@
+.gla-cyoi-section {
```
asvinb: @ankitrox I don't see those styles being used in `index.js`. So they should be in `cyo-incentive-picker.scss`

js/src/components/paid-ads/ads-campaign/ads-campaign.js
```diff
@@ -19,6 +19,7 @@ import { CampaignPreviewCard } from '../campaign-preview';
 import Faqs from './faqs';
 import PaidAdsFeaturesSection from './paid-ads-features-section';
 import EuRegulationsSection from '../eu-regulations-section';
+import CyoIncentivePicker from '~/components/cyo-incentive-picker/cyo-incentive-picker';
```
asvinb: @ankitrox Since the component will be used exclusively for onboarding, let's have the component live in `js/src/components/paid-ads/ads-campaign`

asvinb (changes requested):
@ankitrox I left some comments. Can you kindly check them out please?
I also noticed that no options can currently be selected. Is that expected at this stage? If so, could you update the test instructions to mention that?

Thanks!

js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
ankitrox: Thanks, removed `div` and added class to the `Section`.

js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
ankitrox: Removed it.

js/src/components/cyo-incentive-picker/cyo-incentive-picker.js
ankitrox: Thanks. Updated it.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
ankitrox: `selected` is removed now along with this condition.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
ankitrox: Removed it. Thanks.

js/src/components/cyo-incentive-picker/index.scss
ankitrox: Removed index.scss and added `js/src/components/cyo-incentive-picker/cyo-incentive-picker.scss`

js/src/components/paid-ads/ads-campaign/ads-campaign.js
ankitrox: Moved it to the suggested location.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
ankitrox: Added.

ankitrox: Removed those props as suggested.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
ankitrox: Removed.

js/src/components/cyo-incentive-picker/cyoi-radio-control.js
ankitrox: Thanks @asvinb , updated the prop name to `amount`.

ankitrox:
Thanks @asvinb for reviewing the PR.
I've made the suggested changes to the PR and assigning this to you for another pass.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,107 @@
...
+								requirement.spend.requiredAmount.units;
+
+							return (
+								<div
+									key={ id }
+									className="gla-cyoi-incentive-picker__row"
+								>
+									<CYOIRadioControl
+										amount={ rewardAmount }
+										offer={ offer }
```
asvinb: @ankitrox `CYOIRadioControl` does not accept the `offer` and `requirement` props. Can you remove them please?

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,107 @@
...
+										value={ id }
+									/>
+									<div className="gla-cyoi-incentive-picker__option">
+										{ sprintf(
+											/* translators: %s: amount in users' currency */
+											__(
+												'Spend %s with Google Ads in the first 60 days to unlock the credit.',
+												'google-listings-and-ads'
+											),
+											formatAmount( spendAmount )
```
asvinb, line 84: @ankitrox As per the [sample response here](https://linear.app/a8c/issue/GOOWOO-492/implement-endpoint-to-get-incentives#example-response-from-rest-api-c7261a53), the units are string and we are passing a string to `formatAmount`. Let's make sure it works as intended please.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.js
```diff
@@ -0,0 +1,107 @@
...
+										) }
+									</div>
+									<div className="gla-cyoi-incentive-picker__helper">
+										<span>
```
asvinb: @ankitrox Any reason we have a `span` here please? I removed it and does not seem to make any visual difference.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.scss
```diff
@@ -0,0 +1,63 @@
+.gla-cyoi-section {
+
+    .gla-subsection-title {
+        font-weight: 400;
+        margin-bottom: 4px;
```
asvinb:
```suggestion
        margin-block-end: $grid-unit-05;
```

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.scss
```diff
@@ -0,0 +1,63 @@
...
+
+    .gla-cyoi-incentive-picker__container {
+        display: grid;
+        grid-template-columns: [radio] auto [option] 120px [metrics-start] 1fr;
+        align-items: center;
+        padding-bottom: $grid-unit-20;
+        line-height: 16px;
+        color: $gray-700;
+
+        :global(.gridicon) {
```
asvinb: @ankitrox Are we using any icons within that component?

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.scss
```diff
@@ -0,0 +1,63 @@
...
+    .gla-cyoi-incentive-picker__row {
+        font-size: $gla-font-smaller;
+
+        label {
+            font-size: 15px;
+            font-weight: 500;
+            color: $black;
+        }
+
+        &:not(.customRow) .metricsItem {
```
asvinb: @ankitrox I don't see the `.metricsItem` element in the component. Also it does not follow BEM standards.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,38 @@
...
+ * @property {number} amount The reward amount to display in the label.
+ * @property {string} value The value of the radio option.
+ * @return {JSX.Element} The rendered radio control component.
+ */
+const CYOIRadioControl = ( { amount, value } ) => {
+	const { formatAmount } = useAdsCurrency();
+
+	return (
+		// eslint-disable-next-line jsx-a11y/label-has-associated-control
+		<label className="gla-cyoi-radio-control__wrapper">
```
asvinb: @ankitrox  Wrapping `RadioControl` in a parent `<label>` creates invalid HTML here. `RadioControl` renders a `<fieldset>` + `<legend>` internally, and a `<label>` should only contain phrasing content, not a `<fieldset>`. It also results in nested/competing label semantics because the radio input already gets its own label structure from `RadioControl`.

Could we remove the outer `<label>` wrapper and use a non-label container instead (e.g. `div`/`span`), then handle selection via the radio itself or by making the wrapper clickable in JS if needed?

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,38 @@
...
+			<RadioControl
+				label={ formatAmount( amount ) }
```
asvinb: @ankitrox We are calling `formatAmount` twice with the same parameter. Let's call it once and store the value in a variable please.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyoi-radio-control.js
```diff
@@ -0,0 +1,38 @@
...
+				options={ [ { value } ] }
+				aria-label={ value }
```
asvinb: @ankitrox The `aria-label` is not meaningful to screen readers since it's just rendering the `value`.

```suggestion
				aria-label={ `Select incentive amount XXX` }
```

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.test.js
```diff
@@ -0,0 +1,132 @@
...
+import { render, screen } from '@testing-library/react';
+
+/**
+ * Internal dependencies
+ */
+import CyoIncentivePicker from './cyo-incentive-picker';
+import useCYOIncentives from '~/hooks/useCYOIncentives';
+import useGoogleAdsAccountBillingStatus from '~/hooks/useGoogleAdsAccountBillingStatus';
+
+jest.mock( '~/hooks/useCYOIncentives' );
```
asvinb, line 16: @ankitrox Let's also mock `useAdsCurrency` please since there's no need to call the real hook here.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+describe( 'CyoIncentivePicker Component', () => {
+	beforeEach( () => {
+		useCYOIncentives.mockReturnValue( {
+			data: INCENTIVES_DATA,
+			hasFinishedResolution: true,
+		} );
+
+		useGoogleAdsAccountBillingStatus.mockReturnValue( {
+			billingStatus: { status: 'approved' },
```
asvinb:
```suggestion
			billingStatus: { status: GOOGLE_ADS_BILLING_STATUS.APPROVED },
```

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.test.js
```diff
@@ -0,0 +1,132 @@
...
+		} );
+	} );
+
+	it( 'should render the component', () => {
+		render( <CyoIncentivePicker /> );
+		const titleElement = screen.queryByText( 'Ads credit offer' );
+		expect( titleElement ).toBeInTheDocument();
+	} );
+
+	it( 'should not render if incentives are not available', () => {
```
asvinb, line 98: @ankitrox Let's also add a test when there are no incentives, i.e the data returned is an empty array please.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.test.js
```diff
@@ -0,0 +1,132 @@
...
+			data: null,
+			hasFinishedResolution: false,
+		} );
+
+		render( <CyoIncentivePicker /> );
+		const titleElement = screen.queryByText( 'Ads credit offer' );
+		expect( titleElement ).not.toBeInTheDocument();
+	} );
+
+	it( 'should not render if billing status is not approved', () => {
```
asvinb, line 130: @ankitrox This test is similar to the first test "'should render the component'" where in `beforeEach`, billing is set to `approved` and we are checking whether the component renders or not. So this test is similar. What would have been helpful is to add a test to check whether the component renders when the billing switches from `pending` to `approved`.

asvinb (changes requested):

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.js
ankitrox: Screenshot 2026-04-02 at 8 50 38 PM

`formatAmount` can accept the `number` and `string`, so I think it is all good here.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.scss
ankitrox: Thank you, added it.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.js
ankitrox: Removed it. I added it during initial implementation while I was taking a different approach. Thanks for spotting it.

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyo-incentive-picker.scss
ankitrox: Not required here, removed it

js/src/components/paid-ads/ads-campaign/cyo-incentive-picker/cyoi-radio-control.js
ankitrox: This is inline with `BudgetRadioControl` markup [here](https://github.com/woocommerce/google-listings-and-ads/blob/a10533b2f846621a2fe8e937da3cdbede6d028e8/js/src/components/paid-ads/budget-setup/budget-radio-control.js#L31-L49). It also follows the same convention, but I agree that this is not ideal, so I will go ahead and update it. 
We may need to create a separate issue to update `BudgetRadioControl` markup as well.

ankitrox: Updated the wrapper to `div` and kept the radio selection for radio button only as of now.

ankitrox:
Thanks @asvinb for reviewing the PR. I've updated the PR with the suggestions.
Over to you for another pass. 🙂

asvinb:
Thanks @ankitrox . I pushed a couple of commits and this should be ready for QA.

asvinb (approved):

ankitguptaindia:
QA Status: Verified ✅ 
---


**Environment:** 

- WordPress: 7.0-RC2-62197
- PHP: 8.4.18
- Server: nginx/1.26.1
- Browser: Chrome 146.0.0.0
- OS: macOS
- MU Plugins: None activated
- Plugins:
  * Google for WooCommerce 3.6.1
  * WooCommerce 10.7.0-rc.1



**Test Result:**
Tested CYOI Picker component visibility during campaign creation with mocked incentives endpoint.

* Onboarding completed successfully with billing status set to Approved ✅
* Navigated to Step 3 (Create Campaign) successfully  and CYOIPicker component displayed as expected ✅
* Radio buttons are visible but not selectable (as per current scope) ✅


CYOIPicker component visibility is working as expected. Non-selectable radio buttons align with current implementation scope ([GOOWOO-489](https://linear.app/a8c/issue/GOOWOO-489/integrate-cyoincentivepicker-within-setuppaidads)). No issues found.

Screenshot 2026-04-15 at 10 07 53 PM


---


**Note: Design Update:** 

During testing, a mismatch was observed between the implemented CYOI component and the Figma design.

* Implemented UI includes **“Ads credit” suffix** after each option
* **“Get” prefix before $ value** is missing compared to design
* Overall UI differs from current Figma reference

**Update from Dev Team:**
Confirmed that the design mismatch is **intentional for now**. The component will be updated to match the latest design as part of ticket **GOOWOO-489**.

PR #3304, opened 2026-03-17
Address Dependabot Security Vulnerabilities

ankitrox:
### Changes proposed in this Pull Request:

Closes: https://linear.app/a8c/issue/GOOWOO-509/address-dependabot-security-vulnerabilities-3-critical-10-high

_Replace this with a good description of your changes & reasoning._

### Detailed test instructions:

1. 
2. 
3. 

### Changelog entry

>

PR #3330, opened 2026-03-26
Update node to v22 and dependencies

AlejandroPerezMartin:

Supersedes #3304 (approved but targeted Node 20); #3304 should be closed in favour of this one.

Resolves the fixable Dependabot alerts and moves the toolchain to Node 22. Node 20 reaches EOL on 30 April 2026, and several patched transitive dependencies now require Node 22, so the two changes are done together.

**Node / npm**

- `.nvmrc` → `v22`; `engines.node` → `^22`; `engines.npm` → `^11`.
- `engine-strict=false`. Some transitive dev dependencies still pin Node 20, so strict engine checking is disabled to allow install on 22. This loosens a guardrail — a future dependency pinning the wrong Node version will no longer fail install — so it should be revisited and re-enabled once the upstream tooling supports Node 22.

**JavaScript dependencies**

- `axios` direct dependency bumped `^1.7.2` → `^1.18.0`.
- The overrides are same-major only, so no consumer should see a breaking major bump. They are a manual maintenance surface: each pin must be revisited and dropped once the parent packages ship patched ranges upstream.

**PHP dependencies**

- Minor bumps: `guzzlehttp/guzzle` (7.15.1 → 7.15.2), `phpseclib/phpseclib` (3.0.55 → 3.0.56), Symfony polyfills.
- `firebase/php-jwt` → `googleapis/php-jwt` — upstream repository transfer only, same commit reference and version.

### Remaining alerts (documented exception)

After this PR the only unresolved production-scope alert is a single transitive package (`locutus`) — one critical, plus a few dependent `@woocommerce/*` highs that flag solely because they depend on a vulnerable `locutus` range.

Every other residual high alert is a devDependency — build-time only, not shipped to users. The raw high count rises versus the original report because the Node 22 bump pulls in newer dev tooling with its own advisories, not because any shipped code regressed.

1. `nvm use` (Node 22) then `npm ci` — install completes with no engine error.
2. `npm run build` — the extension bundle builds and `bundlewatch` passes.
3. `npm run test:js` — JS unit tests pass.
4. Load the plugin in `wp-admin` → Marketing → Google for WooCommerce and smoke test the onboarding and dashboard screens: they render with no new console errors on Node 22.
5. `npm audit --omit=dev` — only the `locutus`-rooted alerts remain (see above); nothing else in production scope.

### Changelog entry

> Dev - Upgrade Node to v22 and update dependencies to resolve Dependabot security alerts.


s3rgiosan (approved):

ankitguptaindia:
Closing this PR since the changes are now covered by https://github.com/woocommerce/google-listings-and-ads/pull/3826

PR #3343, opened 2026-04-01
Fix: Unschedule all Action Scheduler jobs on plugin deactivation

fernmed:

When the plugin is deactivated, recurring Action Scheduler jobs are not unscheduled. They continue to run and accumulate failure logs, adding unnecessary maintenance burden. The three actions specifically reported were:

- `gla/jobs/resubmit_expiring_products/start`
- `gla/jobs/update_merchant_price_benchmarks/start`
- `gla/jobs/update_merchant_price_benchmarks/process_item`

However, the root cause is broader — **no** Action Scheduler jobs were being cleaned up on deactivation. This adds a single `as_unschedule_all_actions( null, null, 'gla' )` call to the deactivation hook, clearing all pending jobs in the `gla` group. This matches the existing pattern already used in `uninstall.php`.

Jobs are automatically re-scheduled when the plugin is reactivated via `JobInitializer`.

1. Activate Google for WooCommerce
2. Wait for cron to run or trigger it manually (`wp action-scheduler run`)
3. Verify `gla` jobs are scheduled: `wp db query "SELECT hook, status FROM wp_actionscheduler_actions WHERE group_id IN (SELECT group_id FROM wp_actionscheduler_groups WHERE slug='gla') AND status='pending'"`
4. Deactivate the plugin
5. Run the same query — **all `gla` pending actions should be gone**
6. Re-activate the plugin
7. Trigger cron again and verify jobs are re-scheduled automatically

### Changelog entry

> Fix - Unschedule all Action Scheduler jobs when the plugin is deactivated to prevent orphaned recurring tasks from accumulating failure logs.

---

Part of the [Woo Extensions Bug Blitz](https://woohappyp2.wordpress.com/2026/03/18/extensions-bug-blitz/)

### DIFF
--- a/google-listings-and-ads.php
+++ b/google-listings-and-ads.php
@@ -82,6 +82,11 @@ register_deactivation_hook(
 	__FILE__,
 	function () {
 		PluginFactory::instance()->deactivate();
+
+		// Unschedule all pending ActionScheduler jobs for this plugin.
+		if ( function_exists( 'as_unschedule_all_actions' ) ) {
+			as_unschedule_all_actions( null, null, 'gla' );
+		}
 	}
 );
 


Luc45 (approved):
Tested on QIT (WP 6.9.4, WC 10.6.2, PHP 8.2) with xdebug breakpoints and WP-CLI. Scheduled a mix of recurring and one-time GLA jobs in the `gla` Action Scheduler group, then deactivated the plugin.

On `develop`, set a breakpoint at the closing brace of the deactivation callback (`google-listings-and-ads.php:85`). After `PluginFactory::instance()->deactivate()` returns, the database still shows 5 pending GLA jobs and 0 canceled. The jobs are orphaned and will keep running and failing. On the PR branch, stepped line-by-line through the new code: `function_exists('as_unschedule_all_actions')` returns true, `ActionScheduler::is_initialized()` returns true, and the call enters the `cancel_actions_by_group('gla')` path. After execution: 0 pending, 7 canceled. Clean.

Verified edge cases via WP-CLI: zero GLA jobs (no error), 200 jobs with varying args including `[42]` and `['product_id' => 100, 'topic' => 'product.update']` (all canceled in 1.1s), GLA group missing from `actionscheduler_groups` table (no error), and a simulated in-progress action (correctly left alone, only pending actions are canceled). Also confirmed that the only listeners on the `action_scheduler_canceled_action` and `action_scheduler_bulk_cancel_actions` hooks are the AS Logger, so no downstream code breaks during cancellation.

One thing worth noting: if a merchant deactivates GLA briefly to troubleshoot and then reactivates, recurring jobs (`resubmit_expiring_products`, `update_merchant_price_benchmarks`, etc.) are re-scheduled automatically by `JobInitializer`. But one-time jobs like `update_all_products/create_batch` or `product_notification/start` are permanently lost. This is an acceptable trade-off since those jobs would have failed anyway with the plugin deactivated.

LGTM.

PR #3348, opened 2026-04-08
[GOOWOO-542]: Fix display condition

AlejandroPerezMartin:

<!-- You can erase any parts of this template not applicable to your Pull Request. -->

Closes https://linear.app/a8c/issue/GOOWOO-542/fix-googleadspromo-display-condition-use-ad-spend-instead-of-campaign

Replaces the `adsSetupCompleted` check to rely on `isGoogleAdsReady` as some users that skip the initial campaign creation flow may not see this component as adsSetupCompleted is false.

### Screenshots:

<!--- Optional --->

#### A. Channel Visibility Meta Box (product edit page)                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                
  1. Open any existing product edit page. The "Channel Visibility" meta box should always be visible in the sidebar regardless of MC connection status.                                                                                                                                                         
                  
  2. With MC **not connected** (mock `GET /wp-json/wc/gla/mc/connection` to return a disconnected state):                                                                                                                                                                                                       
     - The meta box should show not be displayed. Ref https://github.com/woocommerce/google-listings-and-ads/pull/3369                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                
  3. With MC **connected** (mock `GET /wp-json/wc/gla/mc/connection` to return a connected state):
     - The meta box should show the channel visibility settings (Google label and visibility dropdown), not the promo banner.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
  ---             

  #### B. Order Attribution Meta Box (order edit page)                                                                                                                                                                                                                                                          
   
  4. In DevTools, mock the response of `GET /wp-json/wc/gla/ads/reports/programs?after=...` by updating the `spend` value:                                                                                                                                                                                      
     ```json      
     {
       "totals": {
         "spend": 0                                                                                                                                                                                                                                                                                             
       }
     }                                                                                                                                                                                                                                                                                                          
                  
  ▎ Reference: https://developer.chrome.com/blog/devtools-tips-34

  5. With no Google Ads connection (mock GET /wp-json/wc/gla/ads/connection to return disconnected) and spend: 0:                                                                                                                                                                                               
    - The promo should show a "Get your products on Google" message with a "Get started" button.
  6. With a connected Google Ads account (mock GET /wp-json/wc/gla/ads/connection to return connected) and spend: 0:                                                                                                                                                                                            
    - The promo should show a "Get more sales with Google Ads" message with a "Create campaign" button.                                                                                                                                                                                                         
  7. With spend set to any value greater than 0:
    - The promo should not be displayed, regardless of Ads connection state. 

### Additional details:

<!--
Optional.
Each line should start with change type prefix`(Fix|Add|…) - `, for example:
> Break - A change breaking previous API or functionality.
> Add - A new feature, function or functionality was added.
> Update - Big changes to something that wasn't broken.
> Fix - Took care of something that wasn't working.
> Tweak - Small change, that isn't actually very important.
> Dev - Developer-facing only change.
> Doc - Updated customer or developer facing documentation

If you remove the "Changelog entry" header, the Pull Request title will be used as the changelog entry.

Add the `changelog: none` label if no changelog entry is needed.
-->
### Changelog entry

> Fix: update hook to render component when users have not created a campaign yet


js/src/meta-boxes/order-attribution/google-ads-promo.js
```diff
@@ -50,7 +50,7 @@ import './google-ads-promo.scss';
  */
 const GoogleAdsPromo = () => {
 	const context = 'order-attribution-meta-box';
-	const { adsSetupComplete } = glaData;
+	const { isGoogleAdsReady } = useGoogleAdsAccountReady();
 	const { hasAdSpend, hasFinishedResolution } = useHasRecentAdSpend();
```
asvinb: @AlejandroPerezMartin `adsSetupComplete` is also used in the `useHasRecentAdSpend` hook. Can you update please?

asvinb (changes requested):

ankitguptaindia:
@AlejandroPerezMartin this PR has some conflicts. Could you please resolve? thanks!

joemcgill (commented):
The changes in https://github.com/woocommerce/google-listings-and-ads/pull/3348/commits/dc16815e848e99f841bafccba5797f9fc96d1161 look ok to me. My only question is whether the user experience is going to be strange if a user sees the visibility drop down before connecting a MC account.

asvinb (approved):

js/src/meta-boxes/channel-visibility/google-ads-promo.js
```diff
@@ -40,24 +39,32 @@ const { adsSetupComplete } = glaData;
  * @return {JSX.Element} The Google Ads Promo component
  */
 const GoogleAdsPromo = () => {
+	const {
```
asvinb, line 42: @AlejandroPerezMartin Actually this line is incorrect. We should check if there's a valid MC connection if we want to keep the existing functionality where the metabox is only visible when there's a connected MC account:
https://github.com/woocommerce/google-listings-and-ads/blob/8d7000d0890d6c61bc1ea28b3e9a377722dfda19/src/Product/ChannelVisibilityMetaBox.php#L113

asvinb (changes requested):

asvinb (approved):

ankitguptaindia:
Related to Point:4

>SBM accounts: The meta box wrapper is always shown. Verify that an SBM with an MC connection sees the settings dropdown and one without sees the promo banner — confirm this is the intended behaviour before merging.


@AlejandroPerezMartin, could you please confirm if this case doesn't have typo?
 I believe the term `Verify that an SBM with an MC connection sees` should be `Verify that an SBM with an Ads connection sees`. I am asking this because, as in SBM, we don't have an MC connection.

ankitguptaindia:
@asvinb 

**Use Case:**
In this task (GOOWOO-617 – Improve UX loading state for Order Widget Banner), a loading spinner was introduced on the Orders page while the IPP banner fetches data.

This works well in scenarios where a banner is expected to be displayed. However, in cases where the spend value is greater than 0, no banner is shown by design. Despite this, the spinner still appears during the loading phase and then disappears without rendering any content.

In this scenario, when a merchant opens an order, they briefly see a loading spinner, but after it completes, nothing is displayed. This may lead to confusion, as it appears that something was supposed to load but did not. Can we handel this scenario? 😉 

https://github.com/user-attachments/assets/879a2c6a-be93-4240-ae4f-85f4e12bf2b2

asvinb:
@ankitguptaindia The spinner is a bit tricky here because it appears while we’re fetching campaign data in the background. At that stage, it’s too early to determine whether the final component will be rendered. Once the data is loaded, we check if there has been any ad spend in the past 14 days. This condition decides whether the component should be displayed.

Because of this, the spinner is only shown during the data fetch. We can either keep it or remove it entirely, but we can’t conditionally hide it based on the final render outcome, since that decision depends on data we don’t have yet.

ankitguptaindia:
QA Status: Approved ✅ 
---

Test Result: QA feedback added above has been addressed and PR work as scoped. All cases are working as expected and no critical issue observed during testing. 





### SBM Store: 


https://github.com/user-attachments/assets/026f9202-97c4-420f-bd06-2cebbaa5d57d



### Non SBM Store: (Fully Onboarded State)


https://github.com/user-attachments/assets/bdce19e5-9395-42b1-ae88-bccb1eacf154


### Non SBM Store: (Incomplete Setup State)


https://github.com/user-attachments/assets/79e13304-d9a1-4834-be32-811354c52296

PR #3350, opened 2026-04-09
Update GMC merchant id for YouTube Shopping exports

jjgrainger:

Closes https://linear.app/a8c/issue/GOOWOO-552/update-gmc-merchant-id-for-youtube-shopping

- Add new method to Middleware to retrieve the MCA ID for the WooCommerce Connect Server.
- Use the ID in the YouTube Shopping conversion reports.

1. Make sure the Order Attribution setting is enabled in WooCommerce > Settings > Advanced > Features
2. Create at least 1 orders via a YouTube link by visiting the test using the the correct UTM query parameters. For example https://www.example.com/product/item?utm_source=youtube&utm_content=YT3-test-attribution.
3. Use the woocommerce_gla_youtube_order_ids_job_date filter to return the date matching when both orders were created. The date returned should be a string following the format Y-m-d, e.g 2025-12-01.
4. Schedule and run the job CreateYouTubeOrderIdsCache. This will capture the order IDs needed for the export and save them to the options table under the key youtube_export_order_ids. Once the job is complete, the CreateMerchantReportedConversionReport job will run immediately after.
5. The CreateMerchantReportedConversionReport job will create the CSV which can be found in the /wp-content/uploads/gla-exports/youtube-merchant-conversion-report-{date}.csv where is the same date used in the filter (eg 2026-04-09).

Check the following:
- A CSV was generated successfully and remains the `gla-exports` folder.
- The `gmc_merchant_id` row has the correct MCA ID for all order rows.

### Changelog entry

> Fix - Update gmc_merchant_id in YouTube Shopping reports.

### DIFF
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -789,8 +789,8 @@ class Middleware implements ContainerAwareInterface, OptionsAwareInterface {
 	/**
 	 * Return the MCA ID for the WooCommerce Connect Server.
 	 *
-	 * @return integer
-	 * @throws Exception When an error response occurs.
+	 * @return int Positive MCA ID.
+	 * @throws Exception When the HTTP response is invalid, mcaId is missing, or mcaId is not a positive integer.
 	 */
 	public function get_wcs_mca_id(): int {
 		try {
@@ -806,7 +806,24 @@ class Middleware implements ContainerAwareInterface, OptionsAwareInterface {
 				);
 			}
 
-			return absint( $response['mcaId'] );
+			$mca_id = filter_var(
+				$response['mcaId'],
+				FILTER_VALIDATE_INT,
+				[
+					'options' => [
+						'min_range' => 1,
+					],
+				]
+			);
+
+			if ( false === $mca_id ) {
+				throw new Exception(
+					__( 'Invalid response when retrieving MCA ID from WooCommerce Connect Server.', 'google-listings-and-ads' ),
+					$result->getStatusCode()
+				);
+			}
+
+			return $mca_id;
 		} catch ( ClientExceptionInterface $e ) {
 			do_action( 'woocommerce_gla_guzzle_client_exception', $e, __METHOD__ );
 
--- a/src/API/Google/Middleware.php
+++ b/src/API/Google/Middleware.php
@@ -790,6 +790,7 @@ class Middleware implements ContainerAwareInterface, OptionsAwareInterface {
 	 * Return the MCA ID for the WooCommerce Connect Server.
 	 *
 	 * @return integer
+	 * @throws Exception When an error response occurs.
 	 */
 	public function get_wcs_mca_id(): int {
 		try {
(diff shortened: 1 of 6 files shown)


src/API/Google/Middleware.php
```diff
@@ -786,4 +785,35 @@ public function get_incentive_credits(): array {
...
+			$response = json_decode( $result->getBody()->getContents(), true );
+
+			if ( 200 !== $result->getStatusCode() || ! is_array( $response ) || ! isset( $response['mcaId'] ) ) {
+				throw new Exception(
+					__( 'Invalid response when retrieving MCA ID from WooCommerce Connect Server.', 'google-listings-and-ads' ),
+					$result->getStatusCode()
+				);
+			}
+
+			return absint( $response['mcaId'] );
```
jamesmorrison: `absint` will convert invalid data e.g. a text string to `0` without throwing error; additionally per the comment "Must be a positive integer number." ([here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Admin/Exports/RowBuilder/OrderItemRowBuilder.php#L56)) this technically allows 0 (i.e. not positive).

Suggest the logic here is tightened up to be certain we have a positive integer or throw an exception.

tests/Unit/Admin/Exports/RowBuilder/OrderItemRowBuilderTest.php
```diff
@@ -174,4 +182,55 @@ public function test_builds_correct_row_for_refund_order_item() {
 		$this->assertEquals( $row['subaccount_id'], '' );
 		$this->assertEquals( $row['reversal_reason'], 'Test refund' );
 	}
+
+	public function test_exception_thrown_if_mca_id_cannot_be_retrieved() {
```
jamesmorrison, line 196: I'm a little concerned we are not testing enough here; the mock's defaults can mask regressions.

We should consider adding unit tests for `Middleware::get_wcs_mca_id()` (200 + valid JSON, non-200, missing `mcaId`, `mcaId → 0`) so parsing rules are not only covered indirectly.

src/API/Google/Middleware.php
joemcgill: I agree that the casting should likely happen above where we validate the MCA ID or throw an exception. I'm unsure that 0 is a valid ID in this context since we always expect the ID to be a positive integer.

tests/Unit/Admin/Exports/RowBuilder/OrderItemRowBuilderTest.php
joemcgill: Given that this is a public method, I think adding unit tests directly for that method that covers all expected cases makes sense.

src/API/Google/Middleware.php
jamesmorrison: I've updated this.

tests/Unit/Admin/Exports/RowBuilder/OrderItemRowBuilderTest.php
jamesmorrison: I've added more testing here

joemcgill (approved):
Thanks for the improvements!

ankitguptaindia:
QA Status: Approved ✅ 
----


Test result: The fix for `gmc_merchant_id` has been validated successfully. The report generation and data accuracy are working as expected, with no issues identified during testing.




image

PR #3357, opened 2026-04-15
[GOOWOO-347] Rewrite the very slow query for the big store

jamesmorrison:

Closes #https://linear.app/a8c/issue/GOOWOO-347/rewrite-the-very-slow-query-for-the-big-store

### Screenshots:

N/A

#### Requirements

1. Ensure you have at least 3 published products; each must have a featured image
2. Add a filter to MU plugin to change the batch size from 100 (default) to 1 so we can test this works without needing to create and publish lots of test products

```
add_filter( 'woocommerce_gla_batched_job_size', function() { return 1; }, 10, 1 );
```

3. Clear any old `resubmit_expiring` results; navigate to WP Admin -> WooCommerce -> Status -> Scheduled Actions, in the search field enter `resubmit_expiring`, click "Search hook, args and claim ID". Click the "select all" checkbox to the left of "Hook" in the table, in the bulk actions dropdown select "Delete" then click "Apply". Repeat until the only result left (that you can't remove) is the "pending" scheduled item.

#### Test steps

1. In a terminal, run: `wp post list --post_type=product --post_status=publish --fields=ID,post_title --format=table` - verify there are at least 3 products returned. If this is not the case, create and publish products then run this check again.

2. Force all published products into the "expiring" state, in a terminal run:

```
TIMESTAMP="$(wp eval 'echo time() - 40 * DAY_IN_SECONDS;')"
for id in $(wp post list --post_type=product --post_status=publish --format=ids); do
  wp post meta delete "$id" _wc_gla_errors 2>/dev/null || true
  wp post meta update "$id" _wc_gla_synced_at "$TIMESTAMP"
  wp post meta update "$id" _wc_gla_visibility "sync-and-show"
done
```

5. Verify the meta was set correctly on all products, in a terminal run:

```
for id in $(wp post list --post_type=product --post_status=publish --format=ids); do
  echo "---- Product $id"
  wp post meta list "$id" --keys=_wc_gla_synced_at,_wc_gla_visibility,_wc_gla_errors --format=table
done
```

6. For each product you should see:

`_wc_gla_synced_at` should be set with a UNIX timestamp
`_wc_gla_visibility` should be set with value `sync-and-show`
`_wc_gla_errors` should NOT be set

7. In WP Admin, navigate to WooCommerce -> Status -> Scheduled Actions.

8. In the search field, enter `resubmit_expiring` and click "Search hook, args and claim ID", find the `pending` result (note there may be completed results, ignore these) and hover over it, click `Run` then wait 60 seconds (time to allow the actions to complete) and refresh the page.

9. In the search field, enter `resubmit_expiring` and click "Search hook, args and claim ID". In the results you should see:
- One or more `create_batch` rows with status Complete
- One or more `process_item` rows with status Complete
- No rows with status Failed

If all these are true, the test passes.

10. Remove the filter added before starting this test.

### Additional details:

> Update - Rewrite the very slow query for the big store

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,7 +169,7 @@
 			},
 			{
 				"path": "./google-listings-and-ads.zip",
-				"maxSize": "8.30 mB",
+				"maxSize": "8.40 mB",
 				"compression": "none"
 			}
 		],
--- a/src/Jobs/ResubmitExpiringProducts.php
+++ b/src/Jobs/ResubmitExpiringProducts.php
@@ -17,6 +17,11 @@ defined( 'ABSPATH' ) || exit;
 class ResubmitExpiringProducts extends AbstractProductSyncerBatchedJob implements RecurringJobInterface {
 
 	/**
+	 * Option name for the last processed post ID between batches (keyset pagination checkpoint).
+	 */
+	protected const RESUBMIT_EXPIRING_PRODUCTS_CHECKPOINT_OPTION = 'woocommerce_gla_resubmit_expiring_products_checkpoint';
+
+	/**
 	 * Get the name of the job.
 	 *
 	 * @return string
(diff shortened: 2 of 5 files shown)


jamesmorrison:
Closing; duplicate of #3334.

PR #3360, opened 2026-04-16
[GOOWOO-559] Create MarketService

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-559/create-marketservice

### Screenshots:

N/A

1. Ensure CI tests pass

### Additional details:

> Add - Added MarketService


src/Options/Markets.php
joemcgill, line 1: @jamesmorrison thanks for adding this file, since it was mentioned in the original implementation brief. As I think you discovered, we don't really need a separate class for this, since the MarketService can use the `OptionsAwareTrait` to interact with the option Markets option directly. Let's remove this file and the related test file.

src/MerchantCenter/MarketService.php
```diff
@@ -0,0 +1,183 @@
...
+	 */
+	public function get_markets(): array {
+		$stored = $this->options->get( OptionsInterface::MARKETS );
+
+		if ( ! empty( $stored ) && is_array( $stored ) ) {
+			return $stored;
+		}
+
+		// Default: one market per target country using the site locale & currency.
+		return $this->build_default_markets();
```
joemcgill: I'm not 100% certain, but we will likely need to store this default market after it is built so that when other markets are created and stored, the default (i.e. primary) market is still returned along with any new markets that are added.

My assumption is that get_markets will return a list of markets, with "primary" being one of them. What I'm not clear on is whether editing the primary market will need to be persisted back to the various settings in the WC Store that the default market gets built from.

src/MerchantCenter/MarketService.php
```diff
@@ -0,0 +1,183 @@
...
+	 * Returns a single market by ID.
+	 *
+	 * @param string $id
+	 *
+	 * @return array|null The market config, or null if not found.
+	 */
+	public function get_market( string $id ): ?array {
+		$markets = $this->get_markets();
+
+		return $markets[ $id ] ?? null;
```
joemcgill, line 156: Ah, yes. This is an example where we will need the `get_markets()` method to always contain a market configuration for the `primary` market. In addition, this assumes that each market will be keyed by an $id, which I assume maps to the feedLabel. We either need to store each market in an array keyed with that ID or update this implementation to filter the markets array by `feedLabel`.

src/MerchantCenter/MarketService.php
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	/**
+	 * Adds a new market config to the store.
+	 *
+	 * TODO: The stable ID scheme should be defined when the Markets REST
+	 * Controller (GOOWOO-560) is built.
```
joemcgill: I _think_ this scheme would be used to validate the config in this method before calling `update_markets()`, but if not, then we should also add some validation here and throw an error if a $config doesn't meet the requirements of a market configuration.

src/MerchantCenter/MarketService.php
```diff
@@ -0,0 +1,183 @@
...
+		$this->update_markets( $markets );
+	}
+
+	/**
+	 * Updates values of an existing market.
+	 *
+	 * @param string $id
+	 * @param array  $config
+	 */
+	public function update_market( string $id, array $config ): void {
```
joemcgill: Similar validation of a $config object could be added here as well.

tests/Unit/MerchantCenter/MarketServiceTest.php
```diff
@@ -0,0 +1,295 @@
...
+			->willReturn( null );
+
+		$this->target_audience->expects( $this->once() )
+			->method( 'get_main_target_country' )
+			->willReturn( 'US' );
+
+		$result = $this->market_service->get_markets();
+
+		$this->assertCount( 1, $result );
+		$this->assertSame( 'US', $result[0]['country'] );
```
joemcgill: This test should be updated to ensure that the default market is available using the right ID for the market, e.g.:

```
'primary' => [
	'country'   => 'US',
	'language'  => 'en',
	'currency'  => 'USD',
	'feedLabel' => 'US',
],
```

tests/Unit/Options/MarketsTest.php
joemcgill, line 1: Can be removed

joemcgill (changes requested):
Thanks @jamesmorrison! This is looking really good. Left some feedback inline.

src/MerchantCenter/MarketService.php
joemcgill: After reviewing the `get_read_markets_callback` method in https://github.com/woocommerce/google-listings-and-ads/pull/3361, I think I have a better idea of how you were thinking about the responsibilities of this method and the `get_primary_market()` method.

I think it would be a bit cleaner to move the business logic from that REST callback to this method so that any functionality that requests a list of markets will get the included primary market without needing to duplicate that REST callback logic. What do you think?

asvinb: @joemcgill @jamesmorrison For the primary market, the expectation on the FE is that it behaves like any other market, simply identified by the “primary” ID. Behind the scenes, the BE should handle updating the underlying settings that define the primary market. In that sense, the primary market isn’t really a standalone market, but a representation of the configuration established during onboarding.

src/Options/Markets.php
jamesmorrison: I've updated this; deleted `src/Options/Markets.php`, `tests/Unit/Options/MarketsTest.php`, and removed the `Markets::class` registration from `CoreServiceProvider`.

`MarketService` now uses `OptionsAwareTrait` to access `OptionsInterface::MARKETS` directly.

src/MerchantCenter/MarketService.php
jamesmorrison: `get_markets()` now synthesises primary from site settings and prepends it to stored secondary markets on every call, so primary is always present. It's never written to the Markets option.

For writes: `update_market('primary', $config)` fans out to the underlying stores: `shipping_rate` / `shipping_time` to `MERCHANT_CENTER`, countries to `TARGET_AUDIENCE`. From the caller's perspective it's just `update_market($id, $config)` like any other market; the primary dispatch is internal.

I've also moved the primary composition logic (`label`, `countries`, `shipping`, `free_shipping`) from the REST controller's `build_primary_market()` into `MarketService::get_primary_market()`, so any consumer of `get_markets()` gets the full response-ready object without duplicating that callback logic.

`build_primary_market()` will be removed from the controller in https://linear.app/a8c/issue/GOOWOO-560/create-markets-rest-controller / https://github.com/woocommerce/google-listings-and-ads/pull/3361

src/MerchantCenter/MarketService.php
jamesmorrison: I've updated this; markets are stored keyed by ID. The ID is a separate stable identifier (generated by the controller via `sanitize_title`), not the `feedLabel itself`

`feedLabel` is a value inside the config; `get_market($id)` is a direct key lookup from the array returned by `get_markets()`.

src/MerchantCenter/MarketService.php
jamesmorrison: Agreed and done.

`get_primary_market()` now returns the full response-ready primary (`id`, `label`, `countries`, `country`, `language`, `currency`, `feedLabel`, `shipping_rate`, `shipping_time`, `free_shipping`), composed from `TargetAudience`, `MERCHANT_CENTER` options, and `ShippingRateQuery`.

The controller no longer needs `build_primary_market()`; that will be removed in https://linear.app/a8c/issue/GOOWOO-560/create-markets-rest-controller / https://github.com/woocommerce/google-listings-and-ads/pull/3361

src/MerchantCenter/MarketService.php
jamesmorrison: Done; `update_market()` now validates the merged config as described above.

tests/Unit/MerchantCenter/MarketServiceTest.php
jamesmorrison: Done; `test_get_markets_falls_back_to_default_when_empty` now asserts `assertArrayHasKey('primary', $result)` and checks `$result['primary']['country']` / `$result['primary']['feedLabel']`.

tests/Unit/Options/MarketsTest.php
jamesmorrison: This has been removed

joemcgill (approved):
This looks good to me. Thanks for making the updates!

PR #3361, opened 2026-04-16
[GOOWOO-560] Create Markets REST Controller

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-560/create-markets-rest-controller

### Screenshots:

Inline below

Before you start:
- Login to WP Admin
- Open the console
- Verify `wp.apiFetch` is available by typing `wp.apiFetch` and pressing Enter; it should not return `undefined`.

1. GET `/wc/gla/mc/markets`

Run this in console: `wp.apiFetch({ path: '/wc/gla/mc/markets' }).then(console.log).catch(console.error);`

Expectation: An array is logged. The first item has:
- `id` with value `"primary"`
- `label` with value `"Primary Market"`
- `countries` is a non-empty array of 2-letter country codes
- `shipping_rate` is a string or null
- `shipping_time` is a string or null
- `free_shipping` is null

Screenshot 2026-04-16 at 15 56 33

2. GET `/wc/gla/mc/markets/languages-currencies`

Run this in console: `wp.apiFetch({ path: '/wc/gla/mc/markets/languages-currencies' }).then(console.log).catch(console.error);`

Expectation: An array is logged; this should have:
- `languages` is empty array
- `currencies` is empty array

Note: empty arrays are expected; per the task requirements: "Full population of these arrays (from WPML/Polylang/multi-currency) is deferred to the MarketService issue."

Screenshot 2026-04-16 at 16 02 50

3. POST `/wc/gla/mc/markets`
Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets',
  method: 'POST',
  data: { country: 'GB', language: 'en', currency: 'GBP', shipping_rate: 'flat', shipping_time: 'flat' }
}).then(console.log).catch(console.error);
```

Expectation: Response logged with status `201`. The body is the created market with:
- `id: "gb"` (derived from `sanitize_title(feedLabel)`)
- `country: "GB"`, `language: "en"`, `currency: "GBP"`, `feedLabel: "GB"`
- `shipping_rate: "flat"`, `shipping_time: "flat"`

Screenshot 2026-04-27 at 11 31 32

A subsequent GET `/wc/gla/mc/markets` should now show `gb` in the list alongside `primary`.
**Note:** if `gb` already exists from a prior run, this will return `409` instead (see step 3b). Either change `country` to a different ISO code (e.g. `'DE'` → `id: "de"`, `'FR'` → `id: "fr"`) or DELETE the existing entry first via:
```
wp.apiFetch({ path: '/wc/gla/mc/markets/gb', method: 'DELETE' }).then(console.log);
```

3b. POST `/wc/gla/mc/markets` (duplicate ID)
Run the same POST from step 3 a second time, without resetting state.
Expectation: An error is caught. The response status is `409` and the body contains:
- `message: "A market with this ID already exists."`
- `id: "gb"`

Screenshot 2026-04-27 at 11 32 29

4. POST `/wc/gla/mc/markets` (missing required field)

Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets',
  method: 'POST',
  data: { language: 'en', currency: 'GBP', shipping_rate: 'flat', shipping_time: 'flat' }
}).then(console.log).catch(console.error);
```

Expectation: Error response (400); response data will state `Missing parameter(s): country`

Screenshot 2026-04-16 at 16 08 00

5. PUT `/wc/gla/mc/markets/primary` (update shipping_rate)

Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets/primary',
  method: 'PUT',
  data: { shipping_rate: 'manual' }
}).then(console.log).catch(console.error);
```

Expectation: Response logged with 200 status; the object has shipping_rate: "manual". Other fields like shipping_time retain their previous value (compare to response from test 1)

Screenshot 2026-04-16 at 16 09 11

6. PUT `/wc/gla/mc/markets/primary` (update countries)

Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets/primary',
  method: 'PUT',
  data: { countries: ['US', 'CA', 'GB'] }
}).then(console.log).catch(console.error);
```

Expectation: Response logged with countries: `["US", "CA", "GB"]`.

Screenshot 2026-04-16 at 16 11 04

7. PUT `/wc/gla/mc/markets/primary` (invalid enum value)

Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets/primary',
  method: 'PUT',
  data: { shipping_rate: 'invalid_value' }
}).then(console.log).catch(console.error);

```

Expectation: Error response with 400 status; message says `shipping_rate` is not one of the allowed values.

Screenshot 2026-04-16 at 16 12 54

8. PUT `/wc/gla/mc/markets/nonexistent` (unknown market)

Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets/nonexistent',
  method: 'PUT',
  data: { shipping_rate: 'flat' }
}).then(console.log).catch(console.error);
```

Expectation: Error response with 404 status; message states `Market not found`

Screenshot 2026-04-16 at 16 14 36

9. DELETE `/wc/gla/mc/markets/primary` (blocked)

Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets/primary',
  method: 'DELETE'
}).then(console.log).catch(console.error);
```

Expectation: Error response with 400 status; message states `The primary market cannot be deleted.`

Screenshot 2026-04-16 at 16 15 52

10. DELETE `/wc/gla/mc/markets/nonexistent` (unknown market)

Run this in console:

```
wp.apiFetch({
  path: '/wc/gla/mc/markets/nonexistent',
  method: 'DELETE'
}).then(console.log).catch(console.error);
```

Expectation: Error response with 404 status; message states `Market not found`

Screenshot 2026-04-16 at 16 16 48

11. Unauthenticated request

In a new private browser session; load https://{your-site-domain}/wp-json/wc/gla/mc/markets

Expectation: Error response; a JSON error response with code "rest_forbidden"

Screenshot 2026-04-16 at 16 18 20

### Additional details:

> Add - Added Markets REST Controller


src/API/Site/Controllers/MerchantCenter/MarketsController.php
```diff
@@ -0,0 +1,404 @@
...
+
+			$markets = [ $primary ];
+
+			$stored = $this->market_service->get_markets();
+			foreach ( $stored as $key => $market ) {
+				if ( is_string( $key ) ) {
+					$market['id'] = $key;
+					$markets[]    = $market;
+				}
+			}
```
joemcgill: Rather than transforming a multi-dimensional array to a flat list each time, why don't we match the storage shape to the shape we plan on consuming that data from the API?

Right now, you have the options being stored as:

```php
[
  '${id}'   => [ ...config ],
  '${id-2}' =>  [ ...config ],
]
```

But then have to convert on both input and output to a flat list like this:

```json
[
  [
    id: $id,
    ...config
  ],
  [
    id: $id-2,
    ...config  
  ],
  ...
]
```

Could we either align on the same storage shape or update the expected output shape?

src/API/Site/Controllers/MerchantCenter/MarketsController.php
```diff
@@ -0,0 +1,404 @@
...
+	 */
+	protected function get_create_market_callback(): callable {
+		return function () {
+			// TODO: Implement market creation in GOOWOO-559.
+			return new Response(
+				[
+					'message' => __( 'Not yet implemented.', 'google-listings-and-ads' ),
+				],
+				501
+			);
```
joemcgill: This could use add_market method already, correct?

src/API/Site/Controllers/MerchantCenter/MarketsController.php
```diff
@@ -0,0 +1,404 @@
...
+		}
+
+		$this->options->update( OptionsInterface::MERCHANT_CENTER, $settings );
+
+		$countries = $request->get_param( 'countries' );
+		if ( null !== $countries ) {
+			$audience              = $this->options->get( OptionsInterface::TARGET_AUDIENCE, [] );
+			$audience['countries'] = $countries;
+			$this->options->update( OptionsInterface::TARGET_AUDIENCE, $audience );
+		}
```
joemcgill: Lets move these implementation details into the `MarketService` class in case so we can keep all business logic for updating markets data (and coordinating with other options) outside of these REST controller.

joemcgill (changes requested):
This is headed in the right direction. Left some feedback inline. Apart from that, it looks like this is at least partially based on the work in the `feature/GOOWOO-559-create-marketservice` branch. If so, it would be a bit easier to see the diff and keep things in sync if this PR were updated to target that branch. That way, once that branch's PR is completed and merged into the feature branch, this PR will automatically be updated to also target the feature branch and you can make sure you're not updating the same logic in both PRs.

src/API/Site/Controllers/MerchantCenter/MarketsController.php
jamesmorrison: I've updated this; `MarketService::get_markets()` now returns the same keyed shape as the underlying option: `['primary' => <synthesised>, 'gb' => <stored>, ...]`.

The controller iterates that directly and just injects the array key as id into each element, there's no list↔keyed conversion in either direction. The same applies on writes: `add_market($id, $config)` and `update_market($id, $config)` take the ID separately, so the stored shape matches the consumed shape end-to-end.

src/API/Site/Controllers/MerchantCenter/MarketsController.php
jamesmorrison: Yes, I updated this. `get_create_market_callback()` now calls `MarketService::add_market($id, $config)`. The controller derives the stable ID via `sanitize_title($feedLabel)`, returns 400 if it collides with the reserved 'primary', returns 409 if the ID already exists, catches `InvalidValue` from `add_market` for 400, and returns 201 with the created market on success.

(The test steps were also updated to account for this)

src/API/Site/Controllers/MerchantCenter/MarketsController.php
jamesmorrison: I've updated this, `update_primary_market()` removed from the controller. `get_update_market_callback()` now delegates to `MarketService::update_market($id, $config)` for all IDs including 'primary'; the fan-out to MERCHANT_CENTER / TARGET_AUDIENCE happens inside `MarketService::update_primary_market_fanout()`.

The controller no longer touches `OptionsInterface` and switched its parent from `BaseOptionsController` to `BaseController`. `RESTServiceProvider` no longer wires `TargetAudience` into the controller.

src/API/Site/Controllers/MerchantCenter/MarketsController.php
```diff
@@ -0,0 +1,426 @@
...
+	 *
+	 * @return callable
+	 */
+	protected function get_create_market_callback(): callable {
+		return function ( Request $request ) {
+			$config = [
+				'country'       => $request->get_param( 'country' ),
+				'language'      => $request->get_param( 'language' ),
+				'currency'      => $request->get_param( 'currency' ),
+				'feedLabel'     => strtoupper( $request->get_param( 'country' ) ),
```
asvinb: @jamesmorrison Is it the country code or the country name that we are storing? Also will that field be used in the backend for feeds or only on the FE for display?

src/MerchantCenter/MarketService.php
```diff
@@ -0,0 +1,387 @@
...
+		$country  = $this->target_audience->get_main_target_country();
+		$language = substr( get_locale(), 0, 2 );
+		$currency = get_woocommerce_currency();
+
+		return [
+			'primary' => [
+				'country'   => $country,
+				'language'  => $language,
+				'currency'  => $currency,
+				'feedLabel' => $country,
```
asvinb: @jamesmorrison Do you think we should hardcode the label to `Primary Feed`?

joemcgill (commented):
@jamesmorrison the updates to the REST controller looks good. Given that this depends on the changes in #3360, can you target the `feature/GOOWOO-559-create-marketservice` branch with this PR and make sure there aren't any merge conflicts before finalizing this?

We would not want to merge this PR until after the other is finalized, but once the `MarketService` PR is merged, this PR should auto-update to the feature branch.

src/API/Site/Controllers/MerchantCenter/MarketsController.php
asvinb: Small naming consistency suggestion: could we rename `feedLabel` to `feed_label`?

Most of the codebase seems to follow snake_case for similar fields.

jamesmorrison: Thanks Asvin; I've updated to snake_case.

src/MerchantCenter/MarketService.php
jamesmorrison: I'd keep it as "Primary Market" because the label field describes the market object, while `feed_label` is the separate Google Merchant Centre feed label concept. Calling it "Primary Feed" might conflate the two.

If you want this changed, please let me know.

src/API/Site/Controllers/MerchantCenter/MarketsController.php
jamesmorrison: The country code, not name. In the backend it's the Google Merchant Centre feed label (used to scope the product feed to the market).

asvinb (approved):

PR #3369, opened 2026-04-22
GOOWOO-618: Conditionally register channel visibility metabox

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-618/hide-channel-visibility-metabox-for-service-based-merchants .

#### Setup
- A test site with the Google for WooCommerce plugin installed and the connect-to-Google flow available.
- The ability to add and delete WooCommerce products (Products → Add new).
- Note: the option this PR cares about is named gla_is_service_based_merchant.

Test 1 - channel visibility
1. The Channel Visibility should not be visible when editing a product for service based merchants.
2. For non service based merchants, the metabox should render as usual

Test 2 - Service-based merchant (existing behaviour)
1. Make sure the store has no physical products. If it does, either delete them or mark them all as Virtual (Edit product → Product data → tick "Virtual").
2. Complete the Google for WooCommerce connection flow.
3. Go to Products → All products and edit any product.

Expected: the Channel visibility metabox is not shown.

Test 3 - Non-service-based merchant (existing behaviour)
1. Add at least one physical product (a regular product with Virtual unticked).
2. Edit that product.

Expected: the Channel visibility metabox is shown as usual.

Test 4 - Disconnect clears the cached state ← new
1. Starting from the setup in Test 2 (service-based merchant, connected).
2. Inspect wp_options and confirm that a row exists with option_name = 'gla_is_service_based_merchant' and option_value = 'yes'.
3. Go to Marketing → Google for WooCommerce → Settings.
4. Click Disconnect all accounts and confirm.

Expected:
- The disconnect completes successfully.
- The gla_is_service_based_merchant row is no longer present in wp_options.

Test 5 - Reconnect repopulates the cached state ← new
- Continuing from Test 4 (disconnected, option cleared).
- Reconnect by completing the connect-to-Google flow again.
- Visit any admin page that involves the plugin (e.g. Marketing → Google for WooCommerce, or edit any product).

Expected:
- The gla_is_service_based_merchant row exists again in wp_options and its value matches the current product catalogue: 'yes' if the store still has no physical products / 'no' if the store now has at least one physical product.
- Editing a product shows the Channel visibility metabox if and only if the merchant currently has at least one physical product.

Test 6 - Recovery from a stuck state ← new (the original concern)
This verifies that a merchant who became service-based, then later started selling physical products, can recover by disconnecting and reconnecting.

- Start from a service-based store as in Test 2, fully connected.
- Edit a product → the Channel visibility metabox is hidden. Good.
- Without disconnecting, add a new physical (non-virtual) product.
- Edit that physical product. (For information only — pre-fix behaviour: the Channel visibility metabox stays hidden because the cached state is stale. This is the bug that prompted the fix.)
- Disconnect all accounts (as in Test 4), then reconnect (as in Test 5).
- Edit the physical product again.

Expected: the Channel visibility metabox is now shown, because the cached state was reset by the disconnect and recomputed after reconnect to reflect the real product mix.

### Changelog entry

>

### DIFF
--- a/src/MerchantCenter/AccountService.php
+++ b/src/MerchantCenter/AccountService.php
@@ -25,6 +25,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Options\MerchantAccountState;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsAwareTrait;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\OptionsInterface;
+use Automattic\WooCommerce\GoogleListingsAndAds\Options\ServiceBasedMerchantState;
 use Automattic\WooCommerce\GoogleListingsAndAds\Options\TransientsInterface;
 use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
 use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Psr\Container\ContainerInterface;
@@ -290,6 +291,8 @@ class AccountService implements ContainerAwareInterface, OptionsAwareInterface, 
 		$this->container->get( TransientsInterface::class )->delete( TransientsInterface::MC_ACCOUNT_REVIEW );
 		$this->container->get( TransientsInterface::class )->delete( TransientsInterface::URL_MATCHES );
 		$this->container->get( TransientsInterface::class )->delete( TransientsInterface::MC_IS_SUBACCOUNT );
+
+		$this->container->get( ServiceBasedMerchantState::class )->reset_service_based_merchant_status();
 	}
 
 	/**
@@ -465,7 +468,6 @@ class AccountService implements ContainerAwareInterface, OptionsAwareInterface, 
 		/** @var Merchant $merchant */
 		$merchant = $this->container->get( Merchant::class );
 
-		/** @var Account $account */
 		$account     = $merchant->get_account( $merchant_id );
 		$account_url = $account->getWebsiteUrl() ?: '';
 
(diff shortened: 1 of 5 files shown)


joemcgill (changes requested):
@asvinb this approach looks good to me. However, while reviewing this I noticed that even though the `ServiceBasedMerchantState` class includes a method to reset this status, it's not actually used anywhere that I could find so a customer could end up stuck in this state.

Can we please ensure that this state is being reset whenever someone  disconnects all accounts so folks don't end up in an unrecoverable state before this is merged?

cc: @jamesmorrison for support, since you added this class in #3177.

src/MerchantCenter/AccountService.php
```diff
@@ -465,7 +468,6 @@ private function maybe_add_merchant_center_url( int $merchant_id ) {
 		/** @var Merchant $merchant */
 		$merchant = $this->container->get( Merchant::class );
 
-		/** @var Account $account */
```
joemcgill, line 468: Was this removal intentional?

joemcgill (approved):
This looks good. Thanks for the updates!

src/MerchantCenter/AccountService.php
jamesmorrison: @joemcgill Yes, [Intelephense](https://intelephense.com/) does not like this, flags `$account` as invalid; but I can put this back if you prefer.

ankitguptaindia:
QA Approved ✅ https://linear.app/a8c/issue/GOOWOO-619/update-googleadspromo-copy-for-service-based-merchants-with-no-google#comment-05edabc5

ankitguptaindia:
### QA Status: Approved ✅ 




### SBM Store: 


https://github.com/user-attachments/assets/026f9202-97c4-420f-bd06-2cebbaa5d57d



### Non SBM Store: (Fully Onboarded State)


https://github.com/user-attachments/assets/bdce19e5-9395-42b1-ae88-bccb1eacf154


### Non SBM Store: (Incomplete Setup State)


https://github.com/user-attachments/assets/79e13304-d9a1-4834-be32-811354c52296

PR #3395, opened 2026-05-01
GOOWOO-620: Track failed incentive applications and recover unclaimed incentives via background job

alaca:

Closes https://linear.app/a8c/issue/GOOWOO-620/track-failed-incentive-applications-and-recover-unclaimed-incentives

The simplest way to test this is by using Postman + WP CLI

**Failure path**
1. Send a POST request to `{{host}}/wp-json/wc/gla/ads/incentives` with an invalid ID `{"id":"invalidID"}`
2. Trigger the job `curl -s '{{host}}/wp-cron.php' >/dev/null`
3. Check options 
```
wp option get gla_ads_incentive_apply_error
wp option get gla_ads_has_unclaimed_incentive
```

Expect error flag empty (deleted by job), `has_unclaimed_incentive = 1` if account still has incentives

**Success path**

1. Send a GET request to `{{host}}/wp-json/wc/gla/ads/incentives` and copy ID from the response
2. Set the error flag manually so we can verify that it's being cleared `wp option update gla_ads_incentive_apply_error error`
3. Now make a POST request with the ID you get from step 1 to `{{host}}/wp-json/wc/gla/ads/incentives` with body `{"id":"validID"}` - should return 200 with the cupon code
4. Now verify that the error flag is cleared `wp option get gla_ads_incentive_apply_error` - should be empty 
### Changelog entry

>


src/API/Site/Controllers/Ads/IncentivesController.php
```diff
@@ -112,10 +133,15 @@ protected function apply_incentive_callback(): callable {
 
 				$result = $this->ads_incentives->apply_incentive( $incentive_id, $country_code );
 
+				// Clear any stale error flag from a previous failed attempt.
+				$this->options->delete( OptionsInterface::ADS_INCENTIVE_APPLY_ERROR );
```
asvinb, line 119: @alaca Shall we also clear `ADS_HAS_UNCLAIMED_INCENTIVE`?. If a previous attempt failed (setting `ADS_HAS_UNCLAIMED_INCENTIVE` = true) and then there's a successful one if the user onboards again?

asvinb (commented):
@alaca I left a comment in the PR but i'll let @jamesmorrison or @joemcgill take a a deeper look. Can we also delete the `ADS_HAS_UNCLAIMED_INCENTIVE` option in the `disconnect` method of `AccountService`?

src/API/Site/Controllers/Ads/IncentivesController.php
alaca: @asvinb resolved https://github.com/woocommerce/google-listings-and-ads/pull/3395/commits/8533535f71c496c6d4fe78fe781249d235e355c6

src/Jobs/CheckUnclaimedIncentive.php
```diff
@@ -0,0 +1,146 @@
...
+	 * @param array $items Unused.
+	 */
+	public function process_items( array $items ) {
+		// Only proceed if a previous apply attempt failed.
+		if ( 'error' !== $this->options->get( OptionsInterface::ADS_INCENTIVE_APPLY_ERROR ) ) {
+			return;
+		}
+
+		$country_code  = $this->wc->get_base_country();
+		$language_code = $this->get_language_code();
```
joemcgill: I think these could be internalized into the `AdsIncentives::fetch_incentives()` method. Is there ever a time that we would pass any other values to that method?

If so, we can remove the redundant logic in the IncentivesController [here](https://github.com/woocommerce/google-listings-and-ads/blob/9ebfb86af2d2b41458b6d0b6a2a7e46a15864ac0/src/API/Site/Controllers/Ads/IncentivesController.php#L80-L81) as well as the duplicate `get_language_code()` method.

joemcgill (changes requested):
One small suggestion otherwise, this looks good.

src/Jobs/CheckUnclaimedIncentive.php
alaca: @joemcgill resolved https://github.com/woocommerce/google-listings-and-ads/pull/3395/commits/31c14f483b369df07f782016e1878c11fc78993a

joemcgill (approved):

ankitguptaindia:
QA Status: Approved ✅ 




**When Ad Accounts have a valid incentive:**

Screenshot 2026-05-05 at 5 45 12 PM


**When ad accout maximum number of incentives has been redeemed:**

Screenshot 2026-05-05 at 5 45 02 PM

PR #3397, opened 2026-05-04
GOOWOO-623: MainTabBar intermittently disappears

ihorvorotnov:

Closes GOOWOO-623.

`MainTabNav` is rendered from a module-scoped `let tabs` array that is reassigned (`tabs = tabs.filter(...)`) inside the component body on every render that hits the `!hasMC` branch. Once that mutation happens, the truncated array is reused for the rest of the page's lifetime — even after `useGoogleMCAccount` resolves and `hasGoogleMCConnection` flips to `true`, the `if` branches no longer run, but the source array stays filtered down to `[dashboard, settings]`. To the user, the tab they were on "disappears" from the nav.

The bug only fires when `glaData.mcSetupComplete === false` at first render and `hasGoogleMCConnection` flips `false` → `true` mid-session, which is why nobody could reproduce it from a fresh page load and a hard reload always "fixes" it.

This PR:

- Replaces the mutable module-scoped `let tabs` with an immutable `export const ALL_TABS`.
- Derives the visible tabs inside `MainTabNav` by filtering `ALL_TABS` on each render. Decided against `useMemo` — the runtime cost on a 7-8 element array is negligible and matches how the rest of the codebase handles trivial derivations.
- Removes both `if`-branch reassignments. The two visibility rules (`!enableReports` hides Reports, `!hasMC` keeps only Dashboard + Settings) are preserved as predicates inside the filter.
- Extracts `MC_GATED_TAB_KEYS` as a module-level const.
- Adds `js/src/components/main-tab-nav/main-tab-nav.test.js` (no prior coverage existed). Tests assert behavioral deltas and invariants rather than hardcoded counts, so adding or removing a tab in the future does not require touching the test file.

No public API changes. `<AppTabNav>`, `useGoogleMCAccount`, `useMenuEffect`, `glaData`, and every consumer of `<MainTabNav />` are untouched.

### Screenshots:

_N/A — visible behaviour is unchanged in the happy path; the regression manifests only under a specific async race that produces no observable styling difference per se._

**Automated:**

- `npm test -- js/src/components/main-tab-nav/main-tab-nav.test.js` – 6 new cases
- `npm test` — full suite

**Manual reproduction of the original bug (against develop):**

1. Use a WP install where the GLA plugin is installed and the merchant has not completed Merchant Center setup. Confirm `window.glaData.mcSetupComplete === false` in DevTools.
2. Partially connect Google so the `getGoogleMCAccount` resolver eventually returns `hasGoogleMCConnection: true`.
3. DevTools → Network → throttling Slow 3G.
4. Hard-reload to `wp-admin/admin.php?page=wc-admin&path=/google/dashboard`.
5. On `develop`, after the resolver returns, the nav stays at Dashboard, Settings even though all gated tabs should now be visible. Navigating to a deep-linked tab (e.g. Product Feed) shows the tab "missing" from the nav.

**Verifying the fix on this branch:**

1. Repeat steps 1–4 above on this branch.
2. After the resolver returns, the nav grows to all expected tabs (Dashboard, Reports, Product Feed, Price Benchmark, Attributes, Settings, Shipping).
3. Sign in as a merchant with no MC connection (`mcSetupComplete: false`, `hasGoogleMCConnection: false` permanently). Verify the nav shows exactly Dashboard + Settings on every page and on every navigation.
4. Apply `add_filter( 'woocommerce_gla_enable_reports', '__return_false' );`. Verify the Reports tab is missing on every page and no other tab is affected.
5. Navigate Dashboard → Product Feed → Settings → Product Feed. The active tab indicator must follow the URL on every step; no tab should ever vanish.

### Changelog entry

> Fix - MainTabNav intermittently rendering an incomplete tab list when a Merchant Center connection became available mid-session.

### DIFF
--- a/js/src/components/main-tab-nav/main-tab-nav.js
+++ b/js/src/components/main-tab-nav/main-tab-nav.js
@@ -14,7 +14,7 @@ import useMenuEffect from '~/hooks/useMenuEffect';
 import GtinMigrationBanner from '~/components/gtin-migration-banner';
 import { getShippingUrl } from '~/utils/urls';
 
-let tabs = [
+export const ALL_TABS = [
 	{
 		key: 'dashboard',
 		title: __( 'Dashboard', 'google-listings-and-ads' ),
@@ -52,6 +52,8 @@ let tabs = [
 	},
 ];
 
+const MC_GATED_TAB_KEYS = [ 'dashboard', 'settings' ];
+
 const getSelectedTabKey = ( allTabs ) => {
 	const path = getPath();
 	return allTabs.find( ( el ) => path.includes( el.key ) )?.key;
@@ -63,15 +65,17 @@ const MainTabNav = () => {
 	const { hasGoogleMCConnection } = useGoogleMCAccount();
 	const hasMC = glaData.mcSetupComplete || hasGoogleMCConnection;
 
-	if ( ! glaData.enableReports ) {
-		tabs = tabs.filter( ( { key } ) => key !== 'reports' );
-	}
+	const tabs = ALL_TABS.filter( ( { key } ) => {
+		if ( ! glaData.enableReports && key === 'reports' ) {
+			return false;
+		}
 
-	if ( ! hasMC ) {
-		tabs = tabs.filter( ( { key } ) =>
-			[ 'dashboard', 'settings' ].includes( key )
-		);
-	}
+		if ( ! hasMC && ! MC_GATED_TAB_KEYS.includes( key ) ) {
+			return false;
+		}
+
+		return true;
+	} );
 
 	const selectedKey = getSelectedTabKey( tabs );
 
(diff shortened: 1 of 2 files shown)


asvinb (approved):
Nice job here @ihorvorotnov ! Approved!

ankitguptaindia:
**🧪  QA/Test Status-** Approved ✅ 
------------

**🖥 Testing Environment -** 

<Details>

- WordPress: 7.0-RC3-62350
- PHP: 8.4.18
- Server: nginx/1.26.1
- Browser: Chrome 147.0.0.0
- OS: macOS
- Plugins:
  * Google for WooCommerce 3.6.0
  * WooCommerce 10.8.0-beta.1

</Details>



### Test Results:

Tested this issue by simulating the affected scenario, as the issue was not directly reproducible. Followed the provided testing steps and confirmed that the issue is resolved with the PR changes.

**Before Fix:**

Screenshot 2026-05-15 at 7 39 33 PM

**After Fix:**


https://github.com/user-attachments/assets/05e64e0f-2dbb-4a94-90c7-dbacebf8c36b

ankitguptaindia (approved):

PR #3402, opened 2026-05-06
Add Linear releases workflow for scheduled pipeline

jimjasson:

Triggers after a successful Deploy Product run, syncs commits to the Linear scheduled pipeline, then marks the release as complete.

Requires LINEAR_ACCESS_KEY repo secret (pipeline key from Linear Settings → Releases).

Based on: https://github.com/marketplace/actions/linear-release and https://linear.app/docs/releases

### DIFF
--- a/.github/workflows/linear-release.yml
+++ b/.github/workflows/linear-release.yml
@@ -1,0 +1,37 @@
+name: Linear Release
+
+on:
+  workflow_run:
+    workflows:
+      - Deploy Product
+    types:
+      - completed
+
+jobs:
+  complete:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.workflow_run.head_sha }}
+
+      - name: Sync release
+        id: release
+        uses: linear/linear-release-action@v0
+        with:
+          access_key: ${{ secrets.LINEAR_ACCESS_KEY }}
+          command: sync
+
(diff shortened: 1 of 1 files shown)


.github/workflows/linear-release.yml
```diff
@@ -0,0 +1,37 @@
...
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.workflow_run.head_sha }}
+
+      - name: Sync release
+        id: release
+        uses: linear/linear-release-action@v0
```
joemcgill, line 24: Let's pin this to a specific hash. Same for where this is used below.

.github/workflows/linear-release.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          access_key: ${{ secrets.LINEAR_ACCESS_KEY }}
+          command: sync
```
joemcgill, line 27: It looks like this action supports a "Dry Run" input. Given that we can run the Deploy Product workflow in "Simulate" mode, we should probably pass that along to this workflow so that the Linear Release process is also run in "Dry Run" mode.

joemcgill (changes requested):
@jimjasson I think we should try setting this up to help automate some of the release process management. Can you look into the couple of suggestions I made when you have a chance?

.github/workflows/linear-release.yml
jimjasson: @joemcgill I agree that Dry Run + Release version would be useful for this workflow, but I don't see a simple way to get them from the deploy workflow. 

Any objections with moving the last step of the deploy workflow so that it has access to this data?

PR #3403, opened 2026-05-06
[GOOWOO-591] Map Products to Markets when syncing to MC

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-591/map-products-to-markets-when-syncing-to-mc

### Screenshots:

N/A

**Test 1**

1. In WordPress admin, go to Products → Add new.
2. Create a simple product with a title, description, featured image, category and product data - in stock etc. In the Google for WooCommerce panel on the product edit page, make sure visibility is set to Sync and show
3. Publish the product; verify the Channel Visibility metabox shows Sync and show
4. Go to Marketing → Google for WooCommerce and click the Product Feed tab. Confirm the new product appears with a Synced status after the next scheduled sync runs (typically within the hour).
5. Go to [Google Merchant Center](https://merchants.google.com/), ensuring you are using the same Google Account. Navigate to Products → All products, and find the new product.
6. Click the product to open its details and check the Issues section.

Expected result: The product appears in Merchant Center with no validation errors. All previously synced products still show a Synced status in the Product Feed tab with no new issues.

**Test 2**

1. In WP Admin, navigate to Marketing → Google for WooCommerce and click the Product Feed tab. Find a product showing a Synced status.
2. Open that product in the WordPress editor.
3. In the Google for WooCommerce panel, change the visibility to Don't sync and show.
4. Save the product.
5. Wait for the next scheduled sync to run (typically within the hour).
6. After the sync, go to Google Merchant Center and search for that product.

Expected result: The product is no longer listed in Merchant Center (or shows as "Disapproved / Removed" if Merchant Center takes time to process it).

**Test 3**

Note: you will need a secondary market enabled for this test, you can use a filter to add this in a MU plugin, for example:

```
add_filter( 'woocommerce_gla_markets', function( $markets ) {
    $markets['FR'] = [
        'country'   => 'FR',
        'language'  => 'fr',
        'currency'  => 'EUR',
        'feedLabel' => 'FR',
    ];
    return $markets;
} );
```

 1. In WordPress admin, go to Products → Add new.
  2. Fill in a title, description, price, and category. In the Google for WooCommerce panel, ensure visibility is set to Sync and show.
  3. Publish the product.
  4. Go to Marketing → Google for WooCommerce → Product Feed and wait for the product to show a Synced status (usually within the hour) - this might show "Disapproved", the important bit is this is not stuck on "Not synced"
  5. Open [Google Merchant Center](https://merchants.google.com), go to Products → All products, and find the product.
  6. Open it and check the Issues tab.
  7. Expected result: The product appears in Merchant Center with no errors. All other synced products are unaffected.
  8. Disable the `woocommerce_gla_markets` filter.

### Additional details:

> Add - Map Products to Markets when syncing to MC


src/Product/ProductFactory.php
```diff
@@ -75,5 +83,14 @@ public function create( WC_Product $product, string $target_country, array $mapp
 				'mapping_rules'     => $mapping_rules,
 			]
 		);
+
+		if ( $feed_label ) {
+			$adapted->set_feed_label( $feed_label );
+		}
+		if ( $language ) {
+			$adapted->set_language( $language );
+		}
```
joemcgill, line 91: I had noticed previously that when trying to sync a product to MC, an error would be returned from the API if a product had both a feedLabel AND language property. You may want to confirm this, and if so, make this an if/else clause

src/Product/WCProductAdapter.php
```diff
@@ -970,6 +980,26 @@ public function setTargetCountry( $targetCountry ) {
...
 	}
 
+	/**
+	 * Sets the feed label
+	 *
+	 * @param string $feed_label
+	 */
+	public function set_feed_label( string $feed_label ): void {
+		$this->feed_label = $feed_label;
+		$this->setFeedLabel( $feed_label );
```
joemcgill: Do we need to maintain duplicates of these properties, of could we use the properties stored on the parent class directly?

joemcgill (commented):
@jamesmorrison this looks really good. I've added a couple of questions for you but otherwise I think this is pretty close.

Can you add test instructions for QA that include setting up a product in a separate market to ensure these products are getting synced to MC correctly?

src/Product/ProductFactory.php
jamesmorrison: @joemcgill I'd assumed only one of these would be passed in but yes, to avoid a potential error I've changed this to if / else.

src/Product/WCProductAdapter.php
jamesmorrison: @joemcgill Having checked this again - no, we don't need them. The private `$feed_label` and `$language properties` are written in the setters but were never read back anywhere in the class.

I've corrected this by removing both properties and having the setters delegate directly to the parent class methods `setFeedLabel()` and `setContentLanguage()`

joemcgill (approved):
Thanks for the additional changes and for adding the additional testing instructions. This looks good to me!

PR #3405, opened 2026-05-07
[GOOWOO-288] Continue button disabled after exiting onboarding flow

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-288/continue-button-disabled-after-exiting-onboarding-flow

### Screenshots:

N/A

Note: This race condition requires both the Ads and Merchant Center connections to be initiated near-simultaneously. If one account is fully connected before the other is started, the bug will not appear. This may require repeated testing.

1. Navigate to the onboarding flow at /wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-mc.
2. On "Set up your accounts", connect both the Google Ads account and the Merchant Center account at the same time; initiate both connections without waiting for one to complete before starting the other.
3. Close the browser / tab.
4. Reopen the browser / tab, navigate back to `/wp-admin/admin.php?page=wc-admin&path=%2Fgoogle%2Fsetup-mc`.
5. Continue the setup process - you should see that the accounts connected in step 2 remain connected though the process starts from scratch.
6. Verify you can complete the onboarding process

### Additional details:

> Fix - Continue button disabled after exiting onboarding flow


js/src/hooks/useGoogleAdsAccountReady.js
```diff
@@ -42,7 +42,7 @@ const useGoogleAdsAccountReady = () => {
 	const isGoogleAdsReady =
 		hasGoogleAdsConnection &&
 		hasAccess &&
-		[ '', 'billing', 'link_merchant' ].includes( step );
+		[ '', 'billing', 'link_merchant', 'set_id' ].includes( step );
```
asvinb: @jamesmorrison `isGoogleAdsReady` is `true` when there's a connection, access, and the remaining incomplete step is either `billing` or `link_merchant`. Both are acceptable gaps because the account itself is already created. `billing` is completed in a later onboarding step, and `link_merchant` just means MC hasn't been connected yet.

`set_id` is the first step in account creation. If it's incomplete, the account doesn't exist yet. Including it here contradicts what "ready" means.

Rather than widening this condition, I'd suggest investigating why step is coming back as `set_id` in the first place.

asvinb (changes requested):

PR #3418, opened 2026-05-15
GOOWOO-650 - Shared `ShippingInfoNotice` component

mikaeltomovic:

Closes [GOOWOO-650](https://linear.app/a8c/issue/GOOWOO-650/consolidate-shippingnotice-and-shippingratenotice-into-a-shared)

Introduces a shared `ShippingInfoNotice` component that `ShippingNotice` and `ShippingRateNotice` both delegate to, eliminating duplicated markup and styles. The two verbatim-identical SCSS blocks are replaced by a single `.gla-shipping-info-notice` class. Both wrappers remain as thin components — `ShippingNotice` retains its visibility guard, and all existing tracking event names are unchanged.

#### `ShippingNotice` component in use before change

ShippingNotice

#### `ShippingNotice` component in use after change

ShippingNotice-after

#### `ShippingRateNotice` component in use before change

ShippingRateNotice

#### `ShippingRateNotice` component in use after change

ShippingRateNotice-after

The changes in this PR creates a wrapper component which is re-used by `ShippingRateNotice` and `ShippingNotice`, two components that were previously completely separate but used identical styles in many areas. The goal is for the functionality of these components to remain the same while reducing the code needed for them. Therefore, I'd recommend viewing the components before pulling this branch, and again after. 

#### To view the `ShippingRateNotice` component

1. Go to the "Shipping" tab and ensure that the "Shipping rates" option is set to "Automatically sync my store’s shipping settings to Google."
2. Go to the "Markets" tab and click the "Edit" symbol by your primary market.
3. In the modal that pops up, the `ShippingRateNotice` component should be visible as a notice with the text "Shipping rates are synced automatically from your WooCommerce shipping settings."

#### To view the `ShippingNotice` component

1. Go to the "Shipping" tab and ensure that the "Shipping rates" option is set to "My shipping settings are complex. I will enter my shipping rates and times manually in Google Merchant Center."
2. Go to the "Markets" tab.
3. Open your browser console and type in `glaData.isMultiLingualStore = true;`.
4. Click the "Edit" symbol by your primary market.
5. In the modal that pops up, the `ShippingNotice` component should be visible as a notice with the text "Shipping is managed in Google Merchant Center. Configure shipping rates and times for each currency in your Merchant Center account."

### Changelog entry

>


js/src/pages/markets/market-fields/shipping-section/shipping-info-notice/index.js
```diff
@@ -0,0 +1,46 @@
...
+ * Shared notice component for shipping section notices.
+ *
+ * @param {Object} props
+ * @param {string} props.message Translated string with a <link> placeholder, passed to createInterpolateElement.
+ * @param {string} props.href URL for the TrackableLink.
+ * @param {string} props.eventName Tracking event identifier.
+ * @param {Object} [props.eventProps] Additional tracking properties.
+ * @param {string} [props.className] Additional CSS class name.
+ */
+const ShippingInfoNotice = ( { message, href, eventName, eventProps, className } ) => {
```
asvinb: @mikaeltomovic Instead of having the `message`, `href`, `eventName` and `eventProps` props, let's just have the `children` prop which will make the component more flexible. The current design hardcodes a specific rendering pattern (interpolated string + `TrackableLink`) inside what should be a thin wrapper. With `children`, callers own their content and the component stays reusable beyond this single pattern. I don't think we even need the `className` prop here. Also there's a linting error with this line. Can you kindly check please?

asvinb (changes requested):
Good start @mikaeltomovic . I left a comment on the approach taken. Can you kindly take a look please?

js/src/pages/markets/market-fields/shipping-section/shipping-info-notice/index.js
mikaeltomovic: @asvinb I see the value in the `children` approach. It keeps the component as a pure shell and avoids locking in the `createInterpolateElement` + `TrackableLink` pattern. Totally agree it's more flexible.

Couple of things worth pointing out though:

1. The existing `shipping-notice/index.test.js` asserts on `.gla-shipping-notice` directly. Removing `className` support means the rendered element only carries `gla-shipping-info-notice`, so that assertion would fail. The ticket explicitly says existing test assertions should pass without modification.
2. The first acceptance criterion explicitly defines the minimum prop API as `message`, `href`, `eventName`, and `eventProps`. Replacing those with children would break that AC directly. 

If we want to go with a `children` approach here (which, again, I agree is more flexible) I'd need to update the existing test to query `.gla-shipping-info-notice` instead. Both that and the changes in props requires changing the acceptance criteria.

asvinb: @mikaeltomovic That's fine 👍  Let's update the AC.

mikaeltomovic: Thanks @asvinb 

I have updated so `ShippingInfoNotice` is an empty wrapper component and I have updated the AC accordingly.

asvinb (approved):
LGTM @mikaeltomovic . Thanks!

ankitguptaindia:
### QA Status: Needs Feedback



Thanks for the great work, @mikaeltomovic. The PR works as scoped.

Just wanted to confirm one thing — the “Before” and “After” screenshots added for `ShippingNotice` and `ShippingRateNotice` in above description appear to look the same. Is that expected/correct?



---



**ShippingRateNotice:** 

image


**ShippingNotice:**

image

mikaeltomovic:
@ankitguptaindia Thank you for the review! 

Yes, the idea is for them to look the same. We have refactored the code and turned two separate components into one. The idea is for the code to be more streamlined but for the components to look and function exactly the same.

ankitguptaindia:
> @ankitguptaindia Thank you for the review!
> 
> Yes, the idea is for them to look the same. We have refactored the code and turned two separate components into one. The idea is for the code to be more streamlined but for the components to look and function exactly the same.

Thanks for the confirmation, @mikaeltomovic. We are good to approve this PR.

PR #3420, opened 2026-05-17
Update `useMarketDataViewsConfig` for multilingual and shipping scenarios

mserino:

Closes:
[GOOWOO-606](https://linear.app/a8c/issue/GOOWOO-606/update-usemarketdataviewsconfig-to-render-appropriate-config)
[GOOWOO-586](https://linear.app/a8c/issue/GOOWOO-586/update-usemarketdataviewsconfig-to-render-appropriate-config)
[GOOWOO-582](https://linear.app/a8c/issue/GOOWOO-582/update-usemarketdataviewsconfig-to-render-appropriate-config)
[GOOWOO-577](https://linear.app/a8c/issue/GOOWOO-577/update-usemarketdataviewsconfig-to-render-appropriate-config)

**GOOWOO-606 — Multilingual + automatic shipping**
Added `buildMultiLingualAutomaticConfig`: four fields (Market, Language, Currency, Shipping time), all markets as rows.
Branch: `isMultiLingualStore && shipping_rate === 'automatic'`

**GOOWOO-586 — Non-multilingual + automatic shipping**
Added `buildAutomaticConfig`: two fields (Market with country count, Shipping time), primary market only.
Branch: `!isMultiLingualStore && shipping_rate === 'automatic'`

**GOOWOO-582 — Non-multilingual + flat shipping**
Added `formatFreeShipping` helper (formats free_shipping threshold as "Free over $50.00" using Intl.NumberFormat) and buildFlatConfig: four fields (Market, Shipping rate, Shipping time, Free shipping), all markets as rows.
Branch: `!isMultiLingualStore && shipping_rate === 'flat'`

**GOOWOO-577 — Non-multilingual + manual shipping**
Already implemented (`buildManualConfig`): three fields (Market, Country count, Shipping), primary market only, "Managed in Google" static label.

**Supporting changes**
- Renamed `glaData.multiLingualStore` → `glaData.isMultiLingualStore` to match the PHP key emitted by `Admin.php`
- Updated test `setMocks` / `afterEach` to use the correct `window.glaData.isMultiLingualStore` key
- Updated the fall-through tests to use the multilingual + flat scenario (the last remaining unhandled case) instead of non-multilingual flat, which now has its own branch
- Removed each completed ticket from the fall-through TODO comment as branches landed

In the Markets tab, check the column for the following scenarios:
GOOWOO-606: `glaData.multiLingualStore === true` and `shipping_rate === 'automatic'`
GOOWOO-586: `glaData.multiLingualStore === false` and `shipping_rate === 'automatic'`
GOOWOO-582: `glaData.multiLingualStore === false` and `shipping_rate === 'flat'`
GOOWOO-577: `glaData.multiLingualStore === false` and `shipping_rate === 'manual'`

### Changelog entry

>


js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -135,18 +265,44 @@ const useMarketDataViewsConfig = () => {
...
 	const countryNames = useCountryKeyNameMap();
 
-	// TODO: `glaData.multiLingualStore` is not yet populated by Admin.php — the
-	// backend wiring lands with the multilingual scenario tickets (GOOWOO-598 /
-	// -602 / -606). Until then this always evaluates to `false`, which is
-	// correct behaviour: multilingual scenarios are unreachable.
-	const multiLingualStore = glaData.multiLingualStore ?? false;
+	const isMultiLingualStore = glaData.isMultiLingualStore ?? false;
 
 	const shippingRate = primaryMarket?.shipping_rate;
```
asvinb: @mserino Let's get the shipping method via the `settings.shipping_rate` property of the `useSettings` hook

js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -75,14 +75,144 @@ const buildManualConfig = ( primaryMarket ) => {
...
+ * Formats a free-shipping threshold number into a localised "Free over X" string,
+ * or returns null when no threshold is set.
+ *
+ * @param {number|null} threshold The threshold amount from market.free_shipping.
+ * @param {string}      currency  ISO 4217 currency code from market.currency.
+ * @return {string|null} Formatted string or null.
+ */
+const formatFreeShipping = ( threshold, currency ) => {
+	if ( threshold === null || threshold === undefined ) {
+		return null;
```
asvinb: @mserino I see there's a dash in the designs if there is no free shipping. Can you confirm please?

js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -75,14 +75,144 @@ const buildManualConfig = ( primaryMarket ) => {
...
+	} ) );
+
+	return { fields, data };
+};
+
+/**
+ * Multilingual + automatic shipping scenario: Market, Language, Currency, Shipping time.
+ * All markets (primary and additional) appear as rows.
+ */
+const buildMultiLingualAutomaticConfig = ( markets ) => {
```
asvinb: @mserino Instead of having a `buildMultiLingualAutomaticConfig` and a `buildAutomaticConfig` config, I suggest having the following functions only:
* `buildAutomaticConfig`
* `buildFlatConfig`
* `buildManualConfig`

Then the multilang checks happen within those functions. For flat rates, we won't have that check. Also let's update the functions to accept an object as parameter. Thus it's flexible enough to add more properties depending on the shipping method.

js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -135,18 +265,44 @@ const useMarketDataViewsConfig = () => {
...
 	const countryNames = useCountryKeyNameMap();
 
-	// TODO: `glaData.multiLingualStore` is not yet populated by Admin.php — the
-	// backend wiring lands with the multilingual scenario tickets (GOOWOO-598 /
-	// -602 / -606). Until then this always evaluates to `false`, which is
-	// correct behaviour: multilingual scenarios are unreachable.
-	const multiLingualStore = glaData.multiLingualStore ?? false;
+	const isMultiLingualStore = glaData.isMultiLingualStore ?? false;
 
 	const shippingRate = primaryMarket?.shipping_rate;
```
asvinb, line 425: @mserino Right now, when the market data is not yet available, i.e `hasFinishedResolution === false`, the different functions will be called and executed unnecessarily. Let's return early empty `data` and `fields` properties when `hasFinishedResolution` is `false`.
```
  return { fields: [], data: [], hasFinishedResolution };
}
```

asvinb (changes requested):
@mserino I left some comments in the PR. Can you kindly check them out please? Also I'd update the approach a little bit where I suggest we define all the fields as an object. For e.g we can have an `ALL_FIELDS` as such:
```
const ALL_FIELDS = {
  market: {
    id: 'market',
    label: __( 'Market', 'google-listings-and-ads' ),
    enableHiding: false,
    enableSorting: false,
    render: ( { item } ) => <span className="gla-markets-table__market-cell">{ item.label }</span>,
  },
  ...
}
```

And then for each scenario, you can use the fields you need when returning the `fields` property.
```
fields: [ ALL_FIELDS.market, ALL_FIELDS.language, ALL_FIELDS.currency, etc ],
```

js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -11,42 +11,113 @@ import { PRIMARY_MARKET_ID } from '~/pages/markets/constants';
...
+		enableSorting: false,
+	},
+	shippingTime: {
+		id: 'shippingTime',
+		label: __( 'Shipping time', 'google-listings-and-ads' ),
+		enableHiding: false,
+		enableSorting: false,
+	},
+	shippingTimes: {
+		id: 'shippingTime',
```
asvinb, line 113: @mserino There are 2 fields which have the same `id`. Can you check please?

js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -11,42 +11,113 @@ import { PRIMARY_MARKET_ID } from '~/pages/markets/constants';
...
+/**
+ * Manual shipping scenario: Market, Country (count), Shipping (static "Managed in Google").
+ * Only the primary market row is shown.
+ *
+ * @param {Object} options
+ * @param {Object} options.primaryMarket Primary market data from usePrimaryMarketDetails.
+ * @return {{ fields: Array, data: Array }} DataViews fields and pre-formatted rows.
  */
-const buildManualConfig = ( primaryMarket ) => {
+const buildManualConfig = ( { primaryMarket } ) => {
```
asvinb, line 218: @mserino I noticed you have the following within 
```
...primaryMarket,
label: primaryMarket.label,
```
I think `label` is not needed here since `primaryMarket` already returns the label. Can you confirm?

js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -76,26 +147,103 @@ const buildManualConfig = ( primaryMarket ) => {
...
+} ) => {
+	if ( isMultiLingualStore ) {
+		const fields = [
+			ALL_FIELDS.market,
+			ALL_FIELDS.language,
+			ALL_FIELDS.currency,
+			ALL_FIELDS.shippingTime,
+		];
+
+		const data = markets.map( ( market ) => ( { ...market } ) );
```
asvinb: @mserino Any idea why we have this please?

js/src/hooks/useMarketDataViewsConfig.js
```diff
@@ -123,28 +271,50 @@ const buildDefaultConfig = ( markets, countryNames ) => {
...
+	const { settings } = useSettings();
+
+	const isMultiLingualStore = glaData.isMultiLingualStore ?? false;
 
-	// TODO: `glaData.multiLingualStore` is not yet populated by Admin.php — the
-	// backend wiring lands with the multilingual scenario tickets (GOOWOO-598 /
-	// -602 / -606). Until then this always evaluates to `false`, which is
-	// correct behaviour: multilingual scenarios are unreachable.
-	const multiLingualStore = glaData.multiLingualStore ?? false;
+	if ( ! hasFinishedResolution ) {
```
asvinb, line 426: @mserino Let’s add a check to ensure `settings` has been resolved before rendering. Otherwise, if the markets are already loaded but settings is still pending, the DataViews will briefly fall back to the default config and display the fields. Once settings becomes available, the correct fields will be applied, which causes a flicker as the fields change.

asvinb (changes requested):

js/src/hooks/useMarketDataViewsConfig.js
mserino: Thanks for catching, this was something that got left there while I was trying to do something else 😅

asvinb (changes requested):
@mserino The code changes LGTM! However I don't see any populated data within data views. For e.g when I choose flat rates, the columns are empty. Can you check please? Thanks!

asvinb (approved):

PR #3422, opened 2026-05-18
GOOWOO-631: Phase 1

alaca:

Phase 1: Merchant API read foundation

Adds the MAPI transport and the first migrated read path (`accounts.products.get`). The existing `ShoppingContent` sync stays in place. Write migration is in Phase 2.

Includes:

MerchantApiClient 
MerchantApiException 
MapiProductsService with get() and get_many()
DTOs: Product, ProductStatus, ItemLevelIssue
DI registration and two debug buttons on the Connection Test page
Unit tests for all of the above

~Why this branches off `feature/cyoi-phase-1` instead of `develop`~

~`develop` is currently broken, the Google Ads SDK was bumped to V23 but the source still imports V22, so the plugin shows a fatal error. The V22 to V23 source migration lives on `feature/cyoi-phase-1` and has not merged into `develop` yet. Base branch is set  to `feature/cyoi-phase-1`, once that gets merged it can be changed to `develop`~

Closes https://linear.app/a8c/issue/GOOWOO-631/phase-1-mapi-http-transport-initial-productproductstatus-migration

### Screenshots:

| SIngle product fetch   | Multi product fetch |
| -------- | ------- |
| image  |   image  |

Connect Merchant Center through the plugin onboarding, then check the merchant ID:
`wp option get gla_merchant_id`

If that returns 0, set it manually

wp option update gla_merchant_id <YOUR_MC_ID>
wp option update gla_mc_setup_completed_at 1

If you dont have a merchant ID, get one at https://merchants.google.com

Add a WooCommerce product. It needs a title, price, description, and a featured image. Description and image are required or it fails Google validation and will not sync.
Save the product. With Merchant Center connected, the plugin syncs it to the Content API. Confirm the Google ID was stored:
`wp post meta get <product_id> _wc_gla_google_ids`

Enable the Connection Test page: add `add_filter( 'woocommerce_gla_enable_connection_test', '__return_true' )`; to a mu-plugin.
Open the Connection Test admin page. In the MAPI Single Fetch field, enter the same product in MAPI format, for example `online~en~US~gla_<product_id>`. 
Click "Fetch product via MAPI". The parsed `Product`, its `productStatus`, and item-level issues are dumped in the Response panel.

### Changelog entry

>


src/API/Google/Mapi/Services/MapiProductsService.php
```diff
@@ -0,0 +1,103 @@
...
+		$results = [];
+		( new EachPromise(
+			$promises(),
+			[
+				'concurrency' => $concurrency,
+				'fulfilled'   => function ( array $body, string $id ) use ( &$results ) {
+					$results[ $id ] = Product::from_array( $body );
+				},
+				'rejected'    => function () {
+					// MerchantApiException already fires the logging action.
```
joemcgill: What if the promise is rejected for another reason? Do we need a fallback for other types of exceptions in the case that a different exception type is thrown?

Ex:

```php
<?php
'rejected' => function ( $reason ) {
    if ( ! $reason instanceof MerchantApiException ) {
        do_action( 'woocommerce_gla_exception', $reason, __METHOD__ );
    }
    // MerchantApiException already fires woocommerce_gla_mc_client_exception.
},
```

src/API/Google/Mapi/MerchantApiClient.php
```diff
@@ -0,0 +1,173 @@
...
+		$decoded = '' === $body ? [] : json_decode( $body, true );
+
+		if ( ! is_array( $decoded ) ) {
+			$decoded = [];
+		}
+
+		$status = $response->getStatusCode();
+		if ( $throw_on_error_status && ( $status < 200 || $status >= 300 ) ) {
+			throw new MerchantApiException( $status, $decoded, __METHOD__ );
+		}
```
joemcgill: The shared ClientInterface is registered with `error_handler` middleware that catches 4xx/5xx responses ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/ef2a554df84642417a9e0eeeddbcdd2db251b414/src/Internal/DependencyManagement/GoogleServiceProvider.php#L221-L253)), so this is likely dead code in production. Worth confirming though.

src/API/Google/Mapi/Services/MapiProductsService.php
```diff
@@ -0,0 +1,103 @@
...
+	/**
+	 * Build the resource path for a product.
+	 *
+	 * @param string $google_product_id
+	 *
+	 * @return string
+	 */
+	protected function build_path( string $google_product_id ): string {
+		return sprintf(
+			'products/v1/accounts/%s/products/%s',
```
joemcgill: Nit: As we add services, the base API version will be duplicated across all services. You may want to consider using a CONST or a shared base Service to define this once.

joemcgill (commented):
@alaca this initial approach is looking really solid, architecturally. I've left just a bit of feedback based on an initial review.

As you progress through this project, lets establish a base `feature` branch that all other phases can target so this can be progressed without merging to `develop` and then all subsequent branches can use non-feature prefixes like `add/...` or `update/...`  . More about the branch naming strategy used in this project can be found [here](https://github.com/woocommerce/google-listings-and-ads/wiki/Git-Flow#fixes-and-naming-branches).

src/API/Google/Mapi/Services/MapiProductsService.php
alaca: Good catch! I added a fallback and a test for the fallback. Resolved https://github.com/woocommerce/google-listings-and-ads/pull/3422/commits/cbc2d3ba4920c9bc7354ad5bd5c8db2fc5b67851 https://github.com/woocommerce/google-listings-and-ads/pull/3422/commits/c4475882639495d6a160c63ca5ec84f10dcf271a

src/API/Google/Mapi/MerchantApiClient.php
alaca: Yes, you are right. Resolved: https://github.com/woocommerce/google-listings-and-ads/pull/3422/commits/dcf85061898f6992dd40b3ee7d6fea3053276943

src/API/Google/Mapi/Services/MapiProductsService.php
alaca: Totally agree. Resolved: https://github.com/woocommerce/google-listings-and-ads/pull/3422/commits/fa6c1c9cf85e831fcedb2c5c5d2f40638891a722

joemcgill (approved):

PR #3423, opened 2026-05-18
[GOOWOO-684] Fix broken ads campaigns endpoint caused by campaign.start_date removal in Google Ads API v23

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-684/fix-broken-ads-campaigns-endpoint-caused-by-campaignstart-date-removal

### Screenshots:

N/A

1. Open WP Admin
2. Open the inspector, select the Network tab; clear all existing entries
3. Open the Product Feed (Marketing -> Google for WooCommerce -> Product Feed)
4. Confirm you see one API request to `/wp-json/wc/gla/ads/campaigns?_locale=user`
5. Confirm this has a 200 response and shows campaign data
6. Switch to the Console tab
7. Confirm you do not see the errors from the screenshot below (from the task):

image

### Additional details:

> Fix - Fix broken ads campaigns endpoint caused by campaign.start_date removal in Google Ads API v23


src/Coupon/WCCouponAdapter.php
```diff
@@ -434,7 +434,7 @@ private function get_product_ids_in_brand( WC_Coupon $wc_coupon, bool $is_exclud
 			if ( is_wp_error( $object_ids ) ) {
 				continue;
 			}
-			$product_ids = array_merge( $product_ids, $object_ids );
+			$product_ids = array_merge( $product_ids, array_map( 'intval', $object_ids ) );
```
asvinb: @jamesmorrison Why this change please? Seems unrelated 🤔

src/API/Site/Controllers/Ads/CampaignController.php
```diff
@@ -439,9 +439,9 @@ protected function get_schema_properties(): array {
 				'validate_callback' => $this->get_supported_country_code_validate_callback(),
 				'readonly'          => true,
 			],
-			'start_date'                            => [
+			'start_date_time'                       => [
 				'type'        => [ 'string', 'null' ],
-				'description' => __( 'Campaign start date in YYYY-MM-DD format.', 'google-listings-and-ads' ),
+				'description' => __( 'Campaign start date and time in yyyy-MM-dd HH:mm:ss format.', 'google-listings-and-ads' ),
```
asvinb: @jamesmorrison Why the change from `YYYY` to `yyyy` please?

asvinb (changes requested):
@jamesmorrison I left a couple of comments. Can you check them out please?
Also am wondering whether renaming `start_date` to `starte_date_time` can be a breaking change. I checked on the FE and we are not using that field. I wonder if the safest approach is:
* Query: keep `campaign.start_date_time` (required for v23)
* REST response: keep the key as `start_date`, extract just the date portion from the datetime string

What do you think?

src/API/Site/Controllers/Ads/CampaignController.php
jamesmorrison: @asvinb Was a typo, apologies, have corrected.

src/Coupon/WCCouponAdapter.php
jamesmorrison: @asvinb `get_objects_in_term()` returns an array of IDs as strings (e.g. `['12', '34']`), not integers. The `array_map('intval', ...)` casts them to integer to ensure type consistency when those IDs are later passed to `$get_offer_id` (which is typed `function ( int $product_id )`).

Without the cast, passing string IDs to a function with an `int` type hint may cause a TypeError in strict mode, (and this file has `declare(strict_types = 1)` at the top).

I've undone this here, it's not causing an immediate issue but is something we might want to look at, not just here but across the codebase.

jamesmorrison:
@asvinb Thanks for taking a look at this; as requested I've changed this to only use the date (time is stripped out so it's just the date) and keep the original `start_date` key.

src/Coupon/WCCouponAdapter.php
asvinb: Thanks @jamesmorrison . Can you create a new ticket for it and we can address everything across the codebase please?

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
```diff
@@ -405,7 +405,7 @@ public function test_create_campaign_with_final_url_and_assets() {
 		$this->ads_campaign->expects( $this->once() )
 			->method( 'create_campaign' )
 			->with( $campaign_data )
-			->willReturn( $expected );
+			->willReturn( array_diff_key( $expected, [ 'start_date' => null ] ) );
```
joemcgill: Should this line be reverted, given the changes applied to the AdsCampaign controller?

joemcgill (approved):
One question about the expectations of Unit Test mocking. Pre-approving pending feedback from @jamesmorrison because the functional changes look right to me.

tests/Unit/API/Site/Controllers/Ads/CampaignControllerTest.php
asvinb: Thanks @joemcgill . Changes have been reverted.

src/Coupon/WCCouponAdapter.php
jamesmorrison: Follow up ticket has been created here: https://linear.app/a8c/issue/GOOWOO-686/maintenance-cast-string-ids-to-int-at-wordpress-core-boundaries-in

PR #3428, opened 2026-05-19
Add market modal fields for non-primary markets

AlejandroPerezMartin:

Closes https://linear.app/a8c/issue/GOOWOO-597/update-addmarketmodal-to-render-appropriate-fields-for-non-primary

1. Go to the plugin dashboard and move to the Shipping tab
2. In **Shipping rates** select the "Automatically sync my store’s shipping settings to Google." option
3. Go to the Markets tab and click on the "Add market" button
4. The design should match the modal
5. When an error is thrown, the modal should close

### Changelog entry

>


js/src/pages/markets/market-form.js
```diff
@@ -234,8 +239,8 @@ const MarketForm = ( {
 			flat_shipping_rate: null,
 			offer_free_shipping: false,
 			free_shipping_threshold: null,
-			flat_shipping_min_time: null,
-			flat_shipping_max_time: null,
+			flat_shipping_min_time: undefined,
+			flat_shipping_max_time: undefined,
```
AlejandroPerezMartin: this is to show the placeholders in the fields when empty, otherwise they are blank: Image

js/src/pages/markets/add-market-modal/index.js
```diff
@@ -93,13 +102,16 @@ const AddMarketModal = ( {
...
-				if ( showAddMarketButton ) {
+				if ( ! isManualNonMultilingual ) {
 					buttons = [
 						...buttons,
 						<AppButton
 							key="add-market"
 							variant="primary"
 							onClick={ handleSubmitClick }
+							disabled={
+								isAutomaticNonMultilingual && ! isValidForm
```
asvinb: @AlejandroPerezMartin Let's not disable the button when the form is invalid. Users should still be able to click the button and see the validation errors. If the button is disabled, users have no indication why they cannot proceed.

Also, why do we need the `isAutomaticNonMultilingual` condition? 🤔

js/src/pages/markets/add-market-modal/index.js
```diff
@@ -60,10 +60,17 @@ const AddMarketModal = ( {
...
+	// Non-multilingual store with manual shipping has no form fields to fill in —
+	// it only shows the multilingual plugin prompt.
+	const isManualNonMultilingual =
 		! glaData.isMultiLingualStore &&
-		settings?.shipping_rate === SHIPPING_RATE_METHOD.MANUAL
-	);
+		settings?.shipping_rate === SHIPPING_RATE_METHOD.MANUAL;
+
+	// Non-multilingual store with automatic shipping uses a simpler form where
+	// the "Add market" button is disabled until the form is valid.
```
asvinb: @AlejandroPerezMartin See my comment further below about disabling the button, but if we do disable the button when the form is invalid, this behavior should apply to all forms, not just the "Non-multilingual store with automatic shipping" one.

js/src/pages/markets/market-fields/audience-section/market-select-control.js
```diff
@@ -16,8 +16,15 @@ import usePrimaryMarketDetails from '~/hooks/usePrimaryMarketDetails';
...
  * the relevant data has been resolved.
+ *
+ * @param {Object}  props
+ * @param {boolean} [props.autoSelectFirstOption=true] Pass false to keep the field empty on open, requiring an explicit user choice.
+ * @param {string}  [props.placeholderOption]          When provided, prepends a non-selectable placeholder option with this label and an empty value.
  */
-const MarketSelectControl = () => {
+const MarketSelectControl = ( {
+	autoSelectFirstOption = true,
+	placeholderOption,
```
asvinb: @AlejandroPerezMartin Since this component will only be used by the "Market form" when adding a new market, we can remove that prop and instead just manually add the "Select..." option to the options.

js/src/pages/markets/market-fields/audience-section/market-select-control.js
```diff
@@ -16,8 +16,15 @@ import usePrimaryMarketDetails from '~/hooks/usePrimaryMarketDetails';
...
  * which are fetched from the store; thus, this control is only rendered once
  * the relevant data has been resolved.
+ *
+ * @param {Object}  props
+ * @param {boolean} [props.autoSelectFirstOption=true] Pass false to keep the field empty on open, requiring an explicit user choice.
+ * @param {string}  [props.placeholderOption]          When provided, prepends a non-selectable placeholder option with this label and an empty value.
  */
-const MarketSelectControl = () => {
+const MarketSelectControl = ( {
+	autoSelectFirstOption = true,
```
asvinb: @AlejandroPerezMartin I think the first option should always be selected. In this case, it would be a blank string since the first option is the "Select..." option you added. So it should still auto-select the first option (blank), and then the user can change the value if needed.

If the user does not select any market and clicks the "Add market" button, there should be an error message indicating that a market must be selected.

So I think we can remove that prop.

js/src/pages/markets/market-form.js
```diff
@@ -141,6 +141,11 @@ const MarketForm = ( {
 				);
 			} else {
 				await createMarket( data );
+
+				if ( data.shipping_rate !== SHIPPING_RATE_METHOD.AUTOMATIC ) {
+					await saveShippingRates( shipping_country_rates );
```
asvinb: @AlejandroPerezMartin `saveShippingRates` and `saveShippingTimes` are not dependent solely on whether the shipping method has changed. Rates and times can be updated independently for any market, so they should be saved accordingly.

If you look at both functions, you’ll see there is already a check in place to determine whether rates or times have actually changed. Only when a change is detected is a request made to persist the updates.

js/src/pages/markets/market-form.js
asvinb: Thanks @AlejandroPerezMartin . We recently merged a PR for the 3.7.0 release where the default shipping times should be set to a minimum of 1 and a maximum of 5.

[Related ticket](https://linear.app/a8c/issue/GOOWOO-202/improve-shipping-speeds-ui-with-default-values-and-better-labelling)

Could you please update this accordingly?

asvinb (changes requested):
@AlejandroPerezMartin I left some comments on the PR. Can you please check and let me know what you think? Thanks!

js/src/pages/markets/market-form.js
AlejandroPerezMartin: updated to 1 and 5

js/src/pages/markets/add-market-modal/index.js
AlejandroPerezMartin: changed it to disabled={isSaving} to avoid multiple submits, I added that check to ensure the logic only affected this particular case (isAutomaticNonMultilingual) and keep the rest as it was previously but removed it

js/src/pages/markets/market-form.js
AlejandroPerezMartin: removed!

js/src/pages/markets/add-market-modal/index.js
```diff
@@ -93,13 +100,14 @@ const AddMarketModal = ( {
...
 
-				if ( showAddMarketButton ) {
+				if ( ! isManualNonMultilingual ) {
 					buttons = [
 						...buttons,
 						<AppButton
 							key="add-market"
 							variant="primary"
 							onClick={ handleSubmitClick }
+							disabled={ isSaving }
```
asvinb: @AlejandroPerezMartin Adding the `disabled` prop is redundant since the button will be disabled when `loading` is set to `true` via `isSaving`.
https://github.com/woocommerce/google-listings-and-ads/blob/develop/js/src/components/app-button/index.js#L52

js/src/pages/markets/market-form.js
```diff
@@ -149,7 +151,7 @@ const MarketForm = ( {
 			invalidateResolution( 'getTargetAudience', [] );
 			onSubmit();
 		} catch ( error ) {
-			// Do nothing. Keep the modal open.
+			onError?.();
```
asvinb: @AlejandroPerezMartin Let's not close the modal when there's an error and revert that change.

js/src/pages/markets/market-form.js
```diff
@@ -72,11 +72,13 @@ function updateTimes( times, isTarget, patch ) {
  * @param {Object} props
  * @param {Object} props.initialMarket Initial values to populate the form with. Can be empty when creating a new market.
  * @param {Function} props.onSubmit Callback function to handle form submission.
+ * @param {Function} [props.onError] Optional callback invoked when the submission API call fails.
  * @return {JSX.Element} The rendered form.
  */
 const MarketForm = ( {
 	initialMarket = {},
 	onSubmit,
+	onError,
```
asvinb: @AlejandroPerezMartin For now the errors will be handled via snackbar. Let's remove that prop.

asvinb (changes requested):

js/src/pages/markets/add-market-modal/index.js
```diff
@@ -60,13 +60,18 @@ const AddMarketModal = ( {
...
-		settings?.shipping_rate === SHIPPING_RATE_METHOD.MANUAL
-	);
+		settings?.shipping_rate === SHIPPING_RATE_METHOD.MANUAL;
 
 	return (
-		<MarketForm initialMarket={ initialMarket } onSubmit={ onRequestClose }>
+		<MarketForm
+			initialMarket={ initialMarket }
+			onSubmit={ onRequestClose }
+			onError={ onRequestClose }
```
asvinb: @AlejandroPerezMartin We no longer have the `onError` prop now.

js/src/pages/markets/add-market-modal/index.js
```diff
@@ -60,10 +60,11 @@ const AddMarketModal = ( {
 		currency: currencyCode,
 	};
 
-	const showAddMarketButton = ! (
+	// Non-multilingual store with manual shipping has no form fields to fill in —
```
asvinb: @AlejandroPerezMartin Can we revert the changes in this file since there are no functional changes. Just renaming variables which we no longer need.

js/src/pages/markets/market-fields/audience-section/index.js
```diff
@@ -1,3 +1,7 @@
+/**
```
asvinb: @AlejandroPerezMartin Why this comment?

asvinb (changes requested):

js/src/pages/markets/market-fields/audience-section/index.js
AlejandroPerezMartin: removed, probably leftover from the merge

asvinb:
LGTM! Thanks!

PR #3432, opened 2026-05-20
GOOWOO-634: Phase 2: Product inserts

alaca:
Adds the MAPI write path. `MapiProductInputsService` inserts products via `productInputs.insert` (single and parallel), and `MapiDataSourcesService` resolves the plugin's primary data source on first use and caches it in options. Products in any previously existing data source (Content API) are moved automatically by MAPI on the next insert.

Closes https://linear.app/a8c/issue/GOOWOO-634/phase-2-product-inserts

Note: once https://github.com/woocommerce/google-listings-and-ads/pull/3422 gets merged, change the base branch to `feature/mapi-migration`

1. Open `wp-admin/admin.php?page=connection-test-admin-page`.
2. Click "Resolve primary data source". Expect a response like `accounts/{merchant_id}/dataSources/{n}`. 
First run creates the plugin's data source, next run returns the cached value.
3. In "MAPI Insert Product", enter an offer ID (e.g. `test_1`) and a title, and click insert. Expect the inserted product details in the response.
4. In "MAPI Parallel Insert", enter comma separated offer IDs (e.g. `test_2, test_3, test_4`), click insert. Expect one result block per ID.
5. Open Merchant Center and go to Products. The insrted products should appear under the plugin's data source. MAPI processing is async, so wait a minute or two before checking

### Changelog entry

>


src/API/Google/Mapi/Services/MapiDataSourcesService.php
```diff
@@ -0,0 +1,131 @@
...
+
+	/**
+	 * Return the resource name of the plugin's primary product data source,
+	 * creating or discovering one as needed and caching the result.
+	 *
+	 * @return string Data source resource name (accounts/{a}/dataSources/{id}).
+	 * @throws MerchantApiException On a non-2xx MAPI response.
+	 */
+	public function ensure_primary_data_source(): string {
+		$cached = (string) $this->options->get( OptionsInterface::MAPI_PRIMARY_DATA_SOURCE, '' );
```
joemcgill: When would this cache get cleared? Should we add a method that clears this cache whenever the merchant account is disconnected?

src/API/Google/Mapi/Services/MapiDataSourcesService.php
```diff
@@ -0,0 +1,131 @@
...
+
+		return $name;
+	}
+
+	/**
+	 * Resolve the data source on plugin activation/update so it is ready before
+	 * the first write request. Skipped when Merchant Center is not connected
+	 *
+	 * @param string $old_version Previous version before updating.
+	 * @param string $new_version Current version after updating.
```
joemcgill: These don't seem to be used

src/API/Google/Mapi/Services/MapiProductInputsService.php
```diff
@@ -0,0 +1,115 @@
...
+	 * Insert a single product.
+	 *
+	 * @param ProductInput $input
+	 *
+	 * @return ProductInput The hydrated response.
+	 * @throws MerchantApiException On a non-2xx MAPI response.
+	 */
+	public function insert( ProductInput $input ): ProductInput {
+		$body = $this->client->post(
+			$this->build_path( $this->data_sources->ensure_primary_data_source() ),
```
joemcgill: I have a potential concern about the data source assumption being made here and throughout this service. For the multi-language support we're working on, the assumption is that when a store supports products in multiple languages and currencies, we would separate product feeds for each market into distinct data sources (or "feeds") in MC.

If I'm understanding the assumptions of this class correctly, it assumes that each product would be inserted into the primary data source when we will instead need to insert each product into a distinct source based on its feedLabel property. That said, I may be totally misunderstanding and conflating the data source described here with something else, so please correct me if so.

joemcgill (commented):
@alaca I have a couple of questions and suggestions to resolve here before finalizing.

Additionally, I'm curious how you're thinking about integrating these new Mapi Services into the existing functionality (essentially replacing the previous ShoppingContent services). Before getting too far, it would be worthwhile to test a full E2E migration of some of these API calls to ensure we can retain compatibility within the plugin functionality after swapping out the API implementation.

src/API/Google/Mapi/Services/MapiProductInputsService.php
alaca: You are reading it right, no conflation. The service does assume one primary data source for everything, which will break once products span multiple (contentLanguage, feedLabel) pairs. Honestly, I had the multi-language work out of mind when I was building this. Thanks for pointing this out. I'll address it in this PR.

src/API/Google/Mapi/Services/MapiDataSourcesService.php
alaca: They are required by the `InstallableInterface`

alaca:
@joemcgill resolved. Now we support multiple data sources based on `contentLanguage` and `feedLabel` pair.

joemcgill (approved):
Thanks for the updates. This looks great except for the failing unit test. I'm pre-approving so that once you've investigated the change, this can be merged.

PR #3434, opened 2026-05-20
Dynamically setting overflow.

mikaeltomovic:

Closes [GOOWOO-678](https://linear.app/a8c/issue/GOOWOO-678/supportedcountryselect-dropdown-is-clipped-inside-appmodal)

The `SupportedCountrySelect` dropdown was being clipped inside `AppModal` because `.components-modal__content` has `overflow: auto`, which creates a clipping context for the absolutely-positioned dropdown tree.

The fix is self-contained in `SupportedCountrySelect`: when the dropdown opens, the component walks up the DOM from its container ref to find the nearest `.components-modal__content` ancestor and its `.app-modal` parent, then sets `overflow: visible` on both (mirroring the existing `overflow="visible"` prop behavior on `AppModal`). When the dropdown closes the inline styles are cleared, restoring the original CSS. Any externally-passed `onDropdownVisibilityChange` callback is preserved by composing it with the internal handler.

### Screenshots:

Screenshot 2026-05-20 at 12 54 43

1. Go to the Markets tab.
2. Click the "Edit" symbol next to one of your listed market. This should open the modal.
3. Click in the "Audience" field. This should open the dropdown.
4. Confirm that the dropdown is visible even when it stretches outside of the modal.

### DIFF
--- a/js/src/components/app-modal/index.scss
+++ b/js/src/components/app-modal/index.scss
@@ -40,4 +40,11 @@
 			overflow: visible;
 		}
 	}
+
+	&:has(.woocommerce-tree-select-control__control-input[aria-expanded="true"]) {
+		&.app-modal,
+		.components-modal__content {
+			overflow: visible;
+		}
+	}
 }
--- a/js/src/components/app-modal/index.scss
+++ b/js/src/components/app-modal/index.scss
@@ -40,11 +40,4 @@
 			overflow: visible;
 		}
 	}
-
-	&:is(.gla-edit-market-modal):has(.woocommerce-tree-select-control__control-input[aria-expanded="true"]) {
-		&.app-modal,
-		.components-modal__content {
-			overflow: visible;
-		}
-	}
 }
--- a/js/src/components/app-modal/index.scss
+++ b/js/src/components/app-modal/index.scss
@@ -41,7 +41,7 @@
 		}
 	}
 
-	&:has(.woocommerce-tree-select-control__control-input[aria-expanded="true"]) {
+	&:is(.gla-edit-market-modal):has(.woocommerce-tree-select-control__control-input[aria-expanded="true"]) {
 		&.app-modal,
 		.components-modal__content {
 			overflow: visible;
--- a/js/src/components/supported-country-select/supported-country-select.js
+++ b/js/src/components/supported-country-select/supported-country-select.js
@@ -3,7 +3,6 @@
  */
 import { __ } from '@wordpress/i18n';
 import { useViewportMatch } from '@wordpress/compose';
-import { useCallback, useEffect, useRef } from '@wordpress/element';
 import classnames from 'classnames';
 
 /**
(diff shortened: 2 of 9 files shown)


asvinb (changes requested):
Thanks @mikaeltomovic . The code works well. However please note that `SupportedCountrySelect` is used in various places outside of the modal and the current implementation is too tied with the expectancy that it will always be rendered within a modal.
I think we can go with a simpler CSS only approach leveraging the `:has` selector in CSS.
```css
.gla-admin-page:has(.woocommerce-tree-select-control__control-input[aria-expanded="true"]) {
	&.app-modal,
	.components-modal__content {
		overflow: visible;
	}
}

```

I tested the CSS above and it seems to be working fine. Maybe you can see how we can tweak the CSS to target only the edit market modal further but I think that should do it. Let me know what you think.

mikaeltomovic:
Thank you for the suggestion @asvinb That approach is much better and I should have thought about it myself 😬 😆 I have updated it now.

js/src/components/app-modal/index.scss
```diff
@@ -40,4 +40,11 @@
 			overflow: visible;
 		}
 	}
+
+	&:has(.woocommerce-tree-select-control__control-input[aria-expanded="true"]) {
```
asvinb: @mikaeltomovic Can we scope those styles to only the `AppModal` used in `EditMarketModal` please?

asvinb (changes requested):

js/src/components/app-modal/index.scss
```diff
@@ -40,4 +40,11 @@
 			overflow: visible;
 		}
 	}
+
+	&:is(.gla-edit-market-modal):has(.woocommerce-tree-select-control__control-input[aria-expanded="true"]) {
```
asvinb: @mikaeltomovic Let's move those CSS to `js/src/pages/markets/edit-market-modal/index.scss` please

asvinb (changes requested):

ankitguptaindia:
QA/Test Status: Approved ✅ 
------------

**QA Summary (Passed):**

Verified the market editing workflow by navigating to the **Markets** tab and opening the **Edit Market** modal. Upon clicking the **Audience** field, the dropdown menu displayed correctly and remained fully visible, including portions extending beyond the modal boundaries. The dropdown was accessible and rendered as expected without any clipping or display issues. The test passed successfully.


### Functional Demo /  Screencast: 

Screenshot 2026-06-03 at 3 57 02 PM




### 🖥 Testing Environment: 

<Details>

- WordPress: 7.0
- PHP: 8.4.18
- Server: nginx/1.26.1
- Browser: Chrome 148
- OS: macOS
- Plugins:
  * Google for WooCommerce : Dev version
  * WooCommerce 10.8.1

</Details>


### 🔁 Next Step: 

Ready to UAT

PR #3438, opened 2026-05-21
GOOWOO-561 Create WPML Integration

zamanq:

Adds basic WPML support so the plugin can tell when a store is multilingual and list active languages.

When WPML is active, `MarketService::has_multilingual_support()` returns true. Languages come from WPML’s `wpml_active_languages` filter and are exposed as `{ code, label }` on `GET /wc/gla/mc/markets/languages-currencies`. Without WPML, behavior is unchanged: multilingual support is off and the languages array is empty.

**What’s Included**
=> New WpmlIntegration class — detects WPML via `ICL_SITEPRESS_VERSION`, reads active languages, picks labels in order: translated name → native name → display name
=> Wired into MarketService and the languages-currencies REST endpoint
=> PHPUnit coverage for the integration, service delegation, and controller response

**Out of scope**
Feed generation per language, syncing translated products to Merchant Center, and multi-currency support — detection and language listing only.

Closes https://linear.app/a8c/issue/GOOWOO-561/create-wpml-integration

### Screenshots:

N/A

#### Without WPML

- REST API

Log in as an admin.

Request:
`GET /wp-json/wc/gla/mc/markets/languages-currencies`

Use the logged-in session (browser Network tab) or:
```
wp rest route list --path=/wc/gla/mc/markets/languages-currencies
wp rest get /wc/gla/mc/markets/languages-currencies --user=1
```

**Expected:**
Status `200`

**Body shape:**
```
{
  "languages": [],
  "currencies": []
}
```
languages and currencies are empty arrays

- Admin glaData

Go to Marketing → Google for WooCommerce (any GLA admin screen).
In the browser console: `window.glaData.isMultiLingualStore`

Expected: `false`

#### With WPML active

- Setup

Install and activate WPML Multilingual CMS.
Under WPML → Languages, enable at least two languages (e.g. English + German).
Confirm the site loads and WPML language switcher works on the front end.

- REST API

As admin, request:
GET `/wp-json/wc/gla/mc/markets/languages-currencies`

**Expected:**
Status `200`
languages is a non-empty array

Each item has code (string, e.g. "en", "de") and label. Labels follow WPML data: prefer translated name, then native name, then display name currencies remains [] (out of scope for this PR as stated in the task requirements)

**Example:**

```
{
  "languages": [
    { "code": "en", "label": "English" },
    { "code": "de", "label": "German" }
  ],
  "currencies": []
}
```
Match the response to WPML → Languages — codes and labels should align with active languages.

- Admin glaData

Reload a GLA admin page.
Check: `window.glaData.isMultiLingualStore`

Expected: `true`

### Additional details:
Added WPML Integration


src/Integration/WpmlIntegration.php
```diff
@@ -0,0 +1,85 @@
...
+namespace Automattic\WooCommerce\GoogleListingsAndAds\Integration;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class WpmlIntegration
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Integration
+ */
+class WpmlIntegration implements IntegrationInterface {
```
joemcgill: Using "Integration" in the class name here is redundant since this is in the Integration namespace. To be consistent with other integration classes this can be omitted. Make sure to update the filename as well.

```suggestion
class WPML implements IntegrationInterface {
```

joemcgill (changes requested):
Thanks @zamanq! I have a minor naming suggestion, otherwise this looks good and could be passed to QA.

PR #3443, opened 2026-05-22
Style dropdown of AppSearchableSelectControl.

mikaeltomovic:

Closes [GOOWOO-680](https://linear.app/a8c/issue/GOOWOO-680/style-searchableselectcontrol-dropdown-to-match-figma-designs)

Styles were added to the `AppSearchableSelectControl` wrapper component to improve the appearance of the dropdown listbox, list options, and the control icon.

**Notes:**

- `SearchableSelectControl` in `js/src/components/searchable-select-control/index.scss` carries two-year-old CSS hacks and overrides that could produce unexpected results if modified. To avoid interfering with those, all new style rules are scoped to `.gla-app-searchable-select-control`, which is only applied by `AppSearchableSelectControl `— currently used exclusively by `LanguageSelectControl `and `CurrencySelectControl `in the locale section of the market form.
- As part of this work, `CurrencySelectControl` was also updated with an expanded placeholder options list (10 currencies), `isSearchable` enabled, and a wired-up `onChange` handler in preparation for the real data integration.
- Some linting errors present in unrelated CSS files were also addressed as part of this work.

### Screenshots:

**Design**

Screenshot 2026-05-22 at 13 09 02

**Implementation**

Screenshot 2026-05-22 at 13 21 22

**Notes on the implemented design:**
1. The `SearchControl` component from WooCommerce does not support displaying the search field above the options as in the design. I have discussed this with @asvinb and agreed that we will not attempt to move the search field to match the design.
2. The design of the labels has not been implemented as part of this ticket.

1. Navigate to the "Shipping" tab and ensure your Shipping rates option is set to "Automatically sync my store’s shipping settings to Google." 
2. Open your browser console and type in `glaData.isMultiLingualStore = true` and hit Enter.
3. Navigate to the "Markets" tab.
4. Click the Edit symbol next to one of your markets in the list of markets.
5. You should now be able to see and test the dropdown under the "Currency" label.

Please note that the Languages dropdown won't be populated unless you're using an actual multi language plugin. We are simulating using the plugin by typing in the `glaData.isMultiLingualStore = true` command in the browser console. This causes the dropdown fields to be displayed but languages won't be populated because there is no plugin to fetch options from.

### DIFF
--- a/js/src/components/app-searchable-select-control/index.scss
+++ b/js/src/components/app-searchable-select-control/index.scss
@@ -8,10 +8,30 @@
 		text-transform: uppercase;
 	}
 
+	.gla-searchable-select-control__input {
+		position: relative;
+
+		&::after {
+			background-color: currentcolor;
+			content: "";
+			display: block;
+			height: 1.5rem;
+			mask-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0naHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmcnIHZpZXdCb3g9JzAgMCAyNCAyNCc+PHBhdGggZmlsbD0nd2hpdGUnIGQ9J00xNS45OSAxMC44OSAxMiAxNC4zbC0zLjk5LTMuNDJMOSA5Ljc1bDMgMi41OCAzLjAxLTIuNTh6Jy8+PC9zdmc+);
+			mask-position: center;
+			mask-repeat: no-repeat;
+			mask-size: contain;
+			position: absolute;
+			right: 8px;
+			top: 50%;
+			transform: translateY(-50%);
+			width: 1.5rem;
+		}
+	}
+
 	.woocommerce-select-control .components-base-control {
 		border-color: $gray-600;
 		border-radius: 2px;
-		padding: 2px 2px 2px 8px;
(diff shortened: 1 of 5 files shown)


ankitguptaindia:
QA Status: Needs Fixes ⚠️ 
----

1. Clicking the dropdown arrow icon does not open the dropdown list. As a user, I would expect both the field and the arrow icon to trigger the dropdown.

2. Selected values from the **Currency** and **Languages** dropdown fields are not being saved. After selecting and saving values, the changes are not persisted when revisiting the modal.



**_Please note that this test was performed with the multilingual WPML plugin and with the branch `feature/multi-lingual-support `_**



https://github.com/user-attachments/assets/6ddfa74e-d2ed-4fae-81bd-6bf85349fa3d

mikaeltomovic:
@ankitguptaindia Thank you for reviewing!

1. I have fixed this by ensuring that the dropdown doesn't have its own pointer-event and simply inherits the one from the input field. This fix is currently in [this PR](https://github.com/woocommerce/google-listings-and-ads/pull/3487).
2. I think this is related to how currencies are handled and not the styling or functionality of the actual component. @asvinb I am just adding you here so you can take a look if needed.

PR #3445, opened 2026-05-22
GOOWOO-687: Phase 2: Product patch

alaca:

Adds the MAPI update path. New `ProductInputPatch` value object and two methods on the existing write service: `patch` and `patch_many`. Each patch resolves its data source from the input's contentLanguage + feedLabel, same as insert, and only the masked fields are written.

Closes https://linear.app/a8c/issue/GOOWOO-687/phase-2-product-patch

Note: once https://github.com/woocommerce/google-listings-and-ads/pull/3432 gets merged, change the base branch to `feature/mapi-migration`

1. Open `wp-admin/admin.php?page=connection-test-admin-page`.
2. In "MAPI Patch Product", enter an existing offer ID, an attribute (`title`), and a new value, then submit. Expect the updated product details in the response.
3. In "MAPI Parallel Patch", enter comma-separated offer IDs and a new title, then submit. Expect one result block per ID.
4. Open Merchant Center > Products and confirm the change. MAPI processing is async, so allow it a minute or two.

### Changelog entry

>

### DIFF
--- a/src/API/Google/Mapi/Models/ProductInputPatch.php
+++ b/src/API/Google/Mapi/Models/ProductInputPatch.php
@@ -1,0 +1,48 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Mapi\Models;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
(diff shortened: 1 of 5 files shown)


joemcgill (approved):
Looks good!

PR #3455, opened 2026-05-27
Ensuring flat shipping rate option is displayed when selected.

mikaeltomovic:

Closes [GOOWOO-703](https://linear.app/a8c/issue/GOOWOO-703/show-flat-shipping-rate-option-when-it-is-the-currently-saved)

This PR proposes two changes:

1. In `setup-free-listings/index.js` we are removing a logic that would force the manual shipping rate option to be displayed as checked if the flat shipping rate was stored in settings and the store was multi-lingual. This logic didn't change the actual setting so it was misleading and needed to be removed.
2. In `shipping-rate-method-section.js` we are introducing a new condition as to wether the flat shipping rate method should be displayed or not. Previously, it was only displayed if the store was not multi-lingual. With the proposed change we are displaying the flat shipping rate method if the store is not multi-lingual OR if the flat shipping rate method is selected.

With both of these changes we are achieving the goal of the GOOWOO-703 issue which is to ensure that the flat shipping rate method stays visible as an option as long as it is selected, even if the user activates a multi-lingual plugin on the site.

1. Ensure that you do not have a multi-lingual plugin activated.
2. Navigate to the Shipping tab. Select _"My shipping settings are simple. I can manually estimate flat shipping rates."_
3. Click "Save changes".
4. Activate a multi-lingual plugin or open your browser console and type in `glaData.isMultiLingualStore = true` and hit Enter.
5. If you used the browser console option in the previous step, navigate to another tab and then back to the Shipping tab but do not refresh the page.
6. Ensure that the shipping rate option labeled _"My shipping settings are simple. I can manually estimate flat shipping rates."_ is still visible and still selected.
7. Select one of the other two shipping rate options.
8. Click "Save changes" . 
9. Ensure that the option labeled _"My shipping settings are simple. I can manually estimate flat shipping rates."_ disappears and cannot be selected.


asvinb (changes requested):
@mikaeltomovic The code looks good so far and works as intended within the “Shipping” tab. However, the “Shipping” tab is going to be removed, and the “Shipping Rate” section will be moved to the “Settings” tab. This is already the case in the feature/multi-lingual-support branch.

The main difference is that, in the “Settings” tab, the shipping rate method is automatically saved in the background, unlike the “Shipping” tab where the user had to click the “Save” button to persist changes.

As a result, as soon as the user selects another option and the new value is saved, the flat rate option disappears immediately. To handle this correctly, we should keep the initial settings value in a ref and update the condition accordingly so that the user can continue seeing all available options while remaining on the page.

mikaeltomovic:
@asvinb I used the `useState` hook to ensure the conditional for showing/hiding the flat shipping rate option only changes when the component is mounted. Have a look and let me know if you're okay with that approach.

asvinb:
Thanks @mikaeltomovic! There's an issue with this approach. `settings` will be `undefined` on the first render since the data is still resolving, so `isFlatShippingRate` is `false` at that point. For a multilingual store that has flat shipping configured: `!isMultiLingualStore || false` === false, and `showFlatOption` is locked to `false` forever, even after `settings` resolves.
The fix is to initialize only once settings has a value, using a ref. For example:
```js
const showFlatOptionRef = useRef( null );
if ( showFlatOptionRef.current === null && settings !== undefined ) {
    showFlatOptionRef.current = ! isMultiLingualStore || isFlatShippingRate;
}
const showFlatOption = showFlatOptionRef.current;
```

mikaeltomovic:
Thank you @asvinb I have updated this to use the `ref` approach now.

asvinb (approved):

PR #3458, opened 2026-05-28
GOOWOO-692: Migrate product create + update sync to MAPI

alaca:

First step in moving product sync off the Content API. The existing `WCProductAdapter` extends `ShoppingContent\Product` and is tied to the Content API SDK model. Since the official MAPI PHP SDK requires PHP 8.1+ (the plugin floor is 7.4), we build a MAPI equivalent from scratch. Adds `WCProductInputAdapter` alongside the old one (built incrementally toward parity) and routes product create/edit sync through MAPI's `productInputs.insert`. Delete stays on the Content API for now. This phase covers the minimum required fields (title, description, link, images, availability, price), richer attributes follow in the next phase.

Closes https://linear.app/a8c/issue/GOOWOO-692/migrate-product-create-update-sync-to-mapi

Note: once https://github.com/woocommerce/google-listings-and-ads/pull/3445 gets merged, change the base branch to `feature/mapi-migration`
- Add a new product in WooCommerce. After a moment, the product should appear in Google Merchant Center.
- Edit an existing synced product (e.g. change the title). The change should appear in Merchant Center.
- Add a variable product with variations. Each variation appears as a separate product in Merchant Center.

### Changelog entry

>

### DIFF
--- a/src/Product/BatchProductHelper.php
+++ b/src/Product/BatchProductHelper.php
@@ -3,6 +3,7 @@ declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Product;
 
+use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Mapi\Models\ProductInput;
 use Automattic\WooCommerce\GoogleListingsAndAds\DB\Query\AttributeMappingRulesQuery;
 use Automattic\WooCommerce\GoogleListingsAndAds\Exception\GoogleListingsAndAdsException;
 use Automattic\WooCommerce\GoogleListingsAndAds\Exception\InvalidValue;
@@ -17,6 +18,7 @@ use Automattic\WooCommerce\GoogleListingsAndAds\Vendor\Google\Service\ShoppingCo
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 use WC_Product;
 use WC_Product_Variable;
+use WC_Product_Variation;
 
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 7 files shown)


src/Product/BatchProductHelper.php
```diff
@@ -257,6 +259,55 @@ public function validate_and_generate_update_request_entries( array $products ):
...
+	/**
+	 * Generate MAPI ProductInput entries for the given products. Expands variable
+	 * products into variations, skips products that are not sync-ready, and builds
+	 * a ProductInput per product targeting the main target country.
+	 *
+	 * @param WC_Product[] $products
+	 *
+	 * @return array<int, array{product: WC_Product, country: string, input: ProductInput}>
+	 */
+	public function generate_mapi_update_entries( array $products ): array {
```
joemcgill, line 271: This method is very similar to the existing `validate_and_generate_update_request_entries` method that it's replacing, but there are a few notable differences that I want to double check on.

1. The previous method handled adding shipping attributes in two places. First, the `WCProductAdapter::map_wc_product_shipping()` always calls `add_shipping_country($targetCountry)` during construction, adding a shipping entry for the primary target country. Then, after validation, `array_walk($target_countries, [$adapted_product, 'add_shipping_country'])` ([ref](https://github.com/woocommerce/google-listings-and-ads/blob/ef466b04562a4bd9184cef96e19856d13840254f/src/Product/BatchProductHelper.php#L241-L242)) adds shipping entries for all additional target countries. The new approach doesn't seem to set shipping attributes at all.

2. The custom attribute mapping rules that were previously fetched [here](https://github.com/woocommerce/google-listings-and-ads/blob/ef466b04562a4bd9184cef96e19856d13840254f/src/Product/BatchProductHelper.php#L202) and passed to the `ProductFactory::create()` method are not applied in this new method, which would result in them being silently ignored.

3. Post processing validation: the previous method included [a validation step](https://github.com/woocommerce/google-listings-and-ads/blob/ef466b04562a4bd9184cef96e19856d13840254f/src/Product/BatchProductHelper.php#L228-L239) to log and skip a product that produced an error while creating a `WCProductAdapter` object. Perhaps that's unnecessary now, but I wanted to confirm.

joemcgill (commented):
@alaca I've got a couple of questions about the new `generate_mapi_update_entries()` method added here. I see that you mentioned in the description that you've intentionally limited the scope of the new `WCProductInputAdapter` object, but want to make sure we're not missing some key attributes that need to be carried over.

src/Product/BatchProductHelper.php
alaca: 1. Right, `WCProductInputAdapter` doesn't set the shipping attribute yet. This was part of the deliberate scope trim for the first cutover, but maybe I should add it in this PR.
2. Correct, and this is the one I would do as a follow-up. The rules feed the standard Google attributes (brand, gtin, color, size, …), which the new adapter doesn't map yet either, so it belongs with the attribute set port rather than this cutover. I'll update the PR description to reflect this.
3. Correct, and I think it's safe to drop. The try/catch that skips a product on a processing error is still there.  What's gone is the Symfony `validate_product()` pass. That was tied to the Content API adapter's constraints and mostly duplicated Google's own checks, and MAPI now validates server-side and returns those failures as item level issues. We also still mark products invalid on API errors in `update_mapi_entries`, so we don't lose the feedback.

alaca:
@joemcgill I added shipping attributes.

src/Product/BatchProductHelper.php
joemcgill: Thanks for making these updates. For attribute mapping:

> Correct, and this is the one I would do as a follow-up. 

I see that this will be handled in GOOWOO-723. Thanks for creating the follow-up issue.

joemcgill (approved):
Thanks @alaca. Approved.

PR #3460, opened 2026-05-28
[GOOWOO-698] Ensure feed data sent to MC correctly reflects the global shipping rate method for each language/currency market

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-698/ensure-feed-data-sent-to-mc-correctly-reflects-the-global-shipping

### Screenshots:

N/A

### Detailed Test Instructions:

1. **Confirm the multilingual setup is in place.**
- In WPML, verify English and French are both active and that the currencies list contains USD and EUR (one currency mapped to each language is sufficient).
- This is what populates the currency dropdown later. Without it the secondary-market rate can only be USD and the bug can't reproduce.

2. **Seed the secondary market in Automatic mode so its currency can be set via UI.**
- Go to Marketing → Google for WooCommerce → Settings.
- Under "Shipping rates" pick "Automatically sync my store's shipping settings to Google." Wait for the success notice ("Merchant Center Settings successfully updated") and for the auto-sync to complete.
- Go to the Markets tab and click "Add market".
- Country: France. Language: French. Currency: EUR. Feed label: `FR`.
- Save. The new FR market row should appear in the dashboard with Currency = EUR.

3. **Switch to Flat mode globally.**
- Back on Settings → Shipping rates, pick "My shipping settings are simple. I can manually estimate flat shipping rates."
- Wait for the success notice. The page auto-saves and syncs.

4. **Set the primary market rate (US / USD).**
- Markets tab → click the edit icon on the Primary Market row.
- In the Shipping section: Estimated shipping rates = `5`, tick "Offer free shipping" with threshold `50`, set the delivery time min/max as appropriate.
- Save. Confirm the row updates and a success notice appears.

5. **Set the FR market rate (EUR).**
- Markets tab → edit the FR market row.
- Confirm the Currency field still reads EUR (it's display-only in flat mode, but the value persists from step 2).
- Estimated shipping rates = `8`, free-shipping threshold = `60`, set delivery times.
- Save.

6. **Verify the dashboard reflects per-market currencies.**
- The Markets dashboard should now show:
 - Primary: US, USD, rate 5, free-shipping threshold 50.
 - FR: France, EUR, rate 8, free-shipping threshold 60.

7. **Verify the payload landing in Merchant Center.**
- Saving in step 5 already triggered a sync. If you want to re-trigger it explicitly, toggle the Settings shipping mode (Flat → Automatic → Flat) and wait for the success notices.
- Open Merchant Center → Shipping and returns → Shipping services. Two Google for WooCommerce-generated services should be present. The service name template includes the rate, currency, and country, so a currency mismatch is visible at a glance:
 - `[xxxx] Google for WooCommerce generated service - 5 USD to US`
 - `[xxxx] Google for WooCommerce generated service - 8 EUR to FR`
- Before this fix, the FR service rendered as `8 USD to FR` (the bug).

8. **Mode toggle preserves rates but gates free-shipping display.**
- Settings → switch shipping mode to Automatic. Wait for the success notice.
- Markets dashboard → the free-shipping column should clear for every row.
- Settings → switch back to Flat. The free-shipping values (50 and 60) should reappear in the dashboard. The DB rows were preserved on disk; the dashboard just hides the column when MC isn't using those rates.

9. **(API check) Currency code format validation.**
- The currency dropdown in the modal is restricted to WPML's currencies, so the UI can't submit a non-ISO value. The API now enforces this independently as defence in depth. To verify, open DevTools → Console and run:
 ```js
 await wp.apiFetch( {
   path: '/wc/gla/mc/shipping/rates',
   method: 'POST',
   data: { country: 'GB', currency: 'gbp', rate: 5 },
 } ).catch( console.error );
 ```
 Expect an error like `Invalid parameter(s): currency`.
- Then run the same call with `currency: 'GBP'` and confirm the response is success (`Successfully added rate for country: "GB"`). Clean up afterwards by removing the GB rate from the Markets dashboard or via the API.

### Additional details:

> Fix - Ensure feed data sent to MC correctly reflects the global shipping rate method for each language/currency market


asvinb (changes requested):
@jamesmorrison Code LGTM! However, could you update the QA instructions to be more accurate in how we expect the feature to be tested?

For example, you currently mention:

> Force flat mode globally; this puts the store into the mode where DB-stored shipping rates are the source of truth, so the rest of the QA has a known baseline. Open DevTools -> Console and enter:

From a QA perspective, it would be preferable to describe this from an end-user workflow. If this setting can be configured either during onboarding or via the Settings tab, the instructions should guide testers through those UI paths instead of relying on the console.

Another example:
> Create a primary-market US/USD rate; this provides a baseline single-market row so we can confirm the existing single-currency path still works (per AC: "no regression to single-market stores"). In the Console enter:

The instructions mention creating a primary market, but the example is actually calling the /wc/gla/mc/shipping/rates/batch endpoint directly.

Could we please rewrite the QA steps from the perspective of an end user rather than an engineer? This not only makes the instructions easier to follow, but also helps uncover potential bugs or UX issues in the UI and validates that the UI integrates correctly with the backend, rather than only verifying the underlying APIs.

PR #3461, opened 2026-05-28
GOOWOO-705 Shipping tab empty settings

AlejandroPerezMartin:

Closes https://linear.app/a8c/issue/GOOWOO-705/regression-in-shipping-tab

Fixes empty settings rendering in the shipping settings tab and adds card spinner while shipping settings are loading.

Reproduce the issue from `develop`:
1. Go to the Shipping tab in the plugin dashboard
2. Open devtools and in the Networks tab find the request to `/wc/gla/mc/settings` and throttle it (set it to 3G): Screenshot 2026-05-28 at 18 15 06
3. Reload the page a few times until you see the some of the settings rendering as empty
4. Now switch to this featured branch and do the same, refresh a couple of times while throttling and confirm that the settings are not rendering empty anymore.

### Changelog entry

> Fix - Shipping settings tab empty settings

### DIFF
--- a/js/src/components/free-listings/choose-audience-section/choose-audience-section.js
+++ b/js/src/components/free-listings/choose-audience-section/choose-audience-section.js
@@ -8,10 +8,9 @@ import { __ } from '@wordpress/i18n';
  */
 import { useAdaptiveFormContext } from '~/components/adaptive-form';
 import AppRadioContentControl from '~/components/app-radio-content-control';
-import RadioHelperText from '~/components/radio-helper-text';
 import Section from '~/components/section';
-import SpinnerCard from '~/components/spinner-card';
 import Subsection from '~/components/subsection';
+import RadioHelperText from '~/components/radio-helper-text';
 import SupportedCountrySelect from '~/components/supported-country-select';
 import VerticalGapLayout from '~/components/vertical-gap-layout';
 import useGoogleMCAccount from '~/hooks/useGoogleMCAccount';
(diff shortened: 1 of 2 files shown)


asvinb (changes requested):
@AlejandroPerezMartin If you look at `SetupFreeListings`, there’s already a condition that renders `AppSpinner`. `SetupFreeListings` renders `FormContent`, which in turn renders `ChooseAudienceSection`. Adding a `hasFinishedResolution` check inside `ChooseAudienceSection` doesn’t really address the underlying issue, since the only thing we’re doing there is checking whether the user has a connected MC account in order to determine the correct section label. Also at this point, the component is already rendered via `AdaptiveForm` with null values. Checking if the data has resolved will make no difference to the component.

Given that, I’d suggest reverting the changes in `ChooseAudienceSection` and ensuring the form loads with the correct resolved data instead. The fix you’ve implemented in `Shipping` should already be sufficient to handle this.

AlejandroPerezMartin:
yup the spinner is added to do not present the user a text and then quickly change it to another one when the MC hook resolves: https://fueled.slack.com/archives/C060DC0NXNG/p1779973015010769?thread_ts=1779972103.923339&cid=C060DC0NXNG, that's for a separate issue

asvinb:
@AlejandroPerezMartin I think we can take a different approach here. Instead of blocking the card’s functionality just to display some informational text, we should use a placeholder state instead.

We can use the `placeholder` SASS mixin, which includes a subtle loading animation to indicate that content is still being fetched. This keeps the UI responsive while clearly communicating that data is on its way.

The key point is that we shouldn’t prevent the user from interacting with the component for a non-critical piece of data.

AlejandroPerezMartin:
ok! I'll log a separate ticket for that as it's deviating from the original issue, I've removed the spinner

asvinb:
LGTM @AlejandroPerezMartin

asvinb (approved):

ankitguptaindia:
QA Sttaus: Approved ✅ 
----




### **Before Fix with Slow Network:**

**Settings show empty fields:** 

image


---

image


---

### **After Fix with Slow Network:** ✅ 

image

PR #3462, opened 2026-05-29
GOOWOO-633 Fix Settings tab page error when missing Google or Jetpack data

AlejandroPerezMartin:

Closes https://linear.app/a8c/issue/GOOWOO-633/gfw-v362-settings-page-crashes-with-typeerror-cannot-read-properties

Add guards to ensure settings page is rendered when google or jetpack is not defined.

1. Go to the plugin dashboard.
2. Open devtools and go to the `Network` tab.
3. Search for: `/wp-json/wc/gla/google/connected` and `/wp-json/wc/gla/jetpack/connected`, right click on them and then `Block Requests > Block Request URL`
5. Refresh the page and ensure the page loads showing a toast message error but there's no error in the console and no generic error on the page saying "Oops, something went wrong"

### Changelog entry

> Fix - Load settings tab when either Jetpack or Google data is not fully available

### DIFF
--- a/js/src/components/google-account-card/connected-google-account-card.js
+++ b/js/src/components/google-account-card/connected-google-account-card.js
@@ -23,7 +23,7 @@ const ConnectedGoogleAccountCard = ( {
 	return (
 		<AccountCard
 			appearance={ APPEARANCE.GOOGLE }
-			description={ googleAccount.email }
+			description={ googleAccount?.email }
 			helper={ helper }
 			indicator={ <ConnectedIconLabel /> }
 		>
--- a/js/src/components/google-account-card/connected-google-account-card.js
+++ b/js/src/components/google-account-card/connected-google-account-card.js
@@ -23,7 +23,7 @@ const ConnectedGoogleAccountCard = ( {
 	return (
 		<AccountCard
 			appearance={ APPEARANCE.GOOGLE }
-			description={ googleAccount?.email }
+			description={ googleAccount.email }
 			helper={ helper }
 			indicator={ <ConnectedIconLabel /> }
 		>
--- a/js/src/pages/settings/index.js
+++ b/js/src/pages/settings/index.js
@@ -46,7 +46,8 @@ const Settings = () => {
 
 	useUpdateRestAPIAuthorizeStatusByUrlQuery();
 
-	const { google } = useGoogleAccount();
+	const { google, hasFinishedResolution: hasFinishedResolutionGoogle } =
+		useGoogleAccount();
 	const isReconnectGooglePage = subpath === subpaths.reconnectGoogleAccount;
 	const { hasFinishedResolution, hasGoogleMCConnection } =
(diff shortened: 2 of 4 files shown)


js/src/components/google-account-card/connected-google-account-card.js
```diff
@@ -23,7 +23,7 @@ const ConnectedGoogleAccountCard = ( {
 	return (
 		<AccountCard
 			appearance={ APPEARANCE.GOOGLE }
-			description={ googleAccount.email }
+			description={ googleAccount?.email }
```
asvinb: @AlejandroPerezMartin Technically, this component is only rendered when a Google account is connected. When there isn't a connected Google account, the `ConnectGoogleAccountCard` component is rendered instead.

Because of that, we shouldn't need the optional chaining operator here. This component is designed to be rendered only when a Google account exists, so `googleAccount` should always be defined at this point.

I'd recommend reverting this condition back to its previous implementation.

asvinb (changes requested):
@AlejandroPerezMartin I left a comment on the PR. While the proposed fixes look reasonable, they appear to be masking the underlying issue rather than addressing the root cause.

If you look at `LinkedAccounts`, the `ConnectedGoogleAccountCard` is only rendered once all required data has been successfully resolved. Under normal circumstances, `googleAccount` should therefore never be `null` at that point.

Instead of adding safeguards around a `null` value, I think we should investigate why the API is returning incomplete or unexpected data. Understanding why `googleAccount` is `null` will help us identify and fix the actual issue rather than simply working around it.

js/src/components/google-account-card/connected-google-account-card.js
asvinb: @AlejandroPerezMartin Looking at the code, the `google` property from `useGoogleAccount` returns `undefined` when there is no connected Jetpack account, or when the `getGoogleAccount` selector itself returns `undefined`.

I would try disconnecting the Google Ads account and then validate the state via `/wp-admin/admin.php?page=connection-test-admin-page`.

One thing I’m wondering is whether we should surface the `ReconnectWPComAccount` component when the Jetpack endpoint returns `undefined` data, instead of silently falling back. The same consideration applies to the `/google/connection` endpoint, where we could render the `ReconnectGoogleAccount` component to make the recovery path more explicit and user-friendly.

AlejandroPerezMartin: @asvinb  Yes, that's why I added the two checks, one for jetpack and the other one for google. The state in that page seems correct, I tried to disable jetpack on my account but couldn't but mocking the response from the hook to return `google: undefined`, which forces that error and the fixes above avoid the page from breaking. But still other sections within the plugin stop working, some graphs and charts just show a spinner apart from other missing content. The origin of the issue is unfortunately still unanswered but I cannot really test much more

mikaeltomovic (approved):
Looks good! Approved.

PR #3463, opened 2026-06-01
GOOWOO-615: Add E2E Tests for market modals

AlejandroPerezMartin:

Closes https://linear.app/a8c/issue/GOOWOO-615/add-e2e-tests-for-multi-lingual-support

Adds e2e tests to multilungual market modals

1. Install Node and Composer dependencies: `npm install && composer install`
2. Start the containers: `npm run env:start`
3. Run the E2E tests: `npm run test:e2e`

### Changelog entry

> Add - E2E tests for market modals


asvinb (changes requested):
@mserino Some [tests are failing](https://github.com/woocommerce/google-listings-and-ads/actions/runs/28800524995/job/85402354405). Can you take a look please?

Can we also add/update the E2E tests for the onboarding updates please?

tests/e2e/specs/markets/markets-multilingual.test.js
```diff
@@ -0,0 +1,651 @@
...
+		await setOnboardedMerchant();
+		await clearServiceBasedMerchant();
+	} );
+
+	test.afterAll( async () => {
+		await clearOnboardedMerchant();
+		await page.close();
+	} );
+
+	test.describe( 'shipping_rate = manual', () => {
```
asvinb: @mserino Let's use the full description for the shipping rate method please which should be "My shipping settings are complex. I will enter my shipping rates and times manually in Google Merchant Center." in this case

tests/e2e/utils/pages/markets.js
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	/**
+	 * Fulfill GET /wc/gla/mc/markets (list only, not sub-paths).
+	 *
+	 * @param {Array}  markets
+	 * @param {number} [status=200]
+	 * @return {Promise<void>}
+	 */
+	async fulfillMarkets( markets, status = 200 ) {
```
asvinb, line 129: @mserino Let's ave the `fulfill*` functions in `tests/e2e/utils/mock-requests.js` please.

tests/e2e/utils/pages/markets.js
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	/**
+	 * Fulfill PATCH /wc/gla/mc/shipping/times/batch.
+	 *
+	 * @param {Object} payload
+	 * @param {number} [status=200]
+	 * @return {Promise<void>}
+	 */
+	async fulfillShippingTimesBatch( payload = {}, status = 200 ) {
```
asvinb, line 264: @mserino I don't see that function used anywhere 🤔

tests/e2e/utils/pages/markets.js
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	/**
+	 * Fulfill PATCH /wc/gla/mc/shipping/rates/batch.
+	 *
+	 * @param {Object} payload
+	 * @param {number} [status=200]
+	 * @return {Promise<void>}
+	 */
+	async fulfillShippingRatesBatch( payload = {}, status = 200 ) {
```
asvinb, line 248: @mserino I don't see that function being used 🤔

tests/e2e/utils/pages/markets.js
asvinb, line 264: @mserino I don't see that function used anywhere. 🤔

tests/e2e/utils/pages/markets.js
```diff
@@ -0,0 +1,406 @@
...
+		return ( container || this.page ).locator(
+			'.gla-market-fields__locale-section'
+		);
+	}
+
+	/**
+	 * @param {import('@playwright/test').Locator} [container]
+	 * @return {import('@playwright/test').Locator} Locator for the multilingual plugin prompt.
+	 */
+	getMultiLingualPluginPrompt( container ) {
```
asvinb, line 473: @mserino Same, i don't see that function used anywhere.

asvinb (dismissed):

joemcgill:
@AlejandroPerezMartin I've refreshed this PR with the latest code from the base branch so you can confirm whether the existing tests are still working as expected or need additional updates before merging.

joemcgill (changes requested):
Thanks for updating these, @AlejandroPerezMartin! I ran this PR through an additional AI check and it found a few additional issues and suggestions that would be good to resolve before finalizing this. Can you update these?

## 🤖 AI Code Review

**Issue:** [Add E2E tests for Multi-Lingual support (GOOWOO-615)](https://linear.app/a8c/issue/GOOWOO-615/add-e2e-tests-for-multi-lingual-support)
**Branch:** `add/GOOWOO-615-e2e-tests-for-multi-lingual-support` → `feature/multi-lingual-support`

### Summary

The implementation covers all six scenario combinations from the AC and the new `MarketsPage` page object is well-structured and clearly documented. There is one fixture correctness bug in the non-multilingual manual scenario that will let France leak into the table after a "successful save" test, contradicting the scenario's core invariant, and both "automatic" shipping-rate describe blocks are missing delete-modal coverage that the "flat" blocks include. Several additional quality issues are worth addressing before merge.

### 🔴 Critical Issues
*Must be addressed before merging.*

- **Wrong markets fixture in non-multilingual manual "successful save" test** (`tests/e2e/specs/markets/markets-non-multilingual.test.js`, line ~1152): The test calls `marketsPage.fulfillMarkets([PRIMARY_MARKET, SECONDARY_MARKET])` as the post-save refresh mock. The "Shipping Rate - Manual" non-multilingual scenario is established (and explicitly asserted) to have only the primary market — France must not exist. After the modal closes, the page re-fetches markets with this mock and France will appear in the table, directly contradicting the earlier test asserting `page.getByRole('cell', { name: 'France' })` is not visible. The subsequent API error test then runs with France visible, potentially masking real bugs. Change to `fulfillMarkets([PRIMARY_MARKET])` to match the scenario invariant.

### 🟡 Suggestions
*Worth addressing — improves correctness, maintainability, or quality.*

- **Missing delete-modal tests in both "automatic" shipping rate scenarios**: Both "Shipping Rate - Automatic" describe blocks (in both spec files) end after the "API error keeps Add modal open" test, with no coverage of the Delete confirmation modal (error snackbar + modal stays open, and successful delete closes modal and removes row) — unlike the "Flat" describe blocks, which do include these tests. Secondary markets are visible in the automatic scenarios, so this coverage gap should be closed for consistency with the AC's delete-action requirement.

- **`getEditButton(index)` uses positional indexing** (`tests/e2e/utils/pages/markets.js`): Order-dependent across all Edit buttons on the page. If a prior serial test leaves the table in an unexpected state, `nth(0)`/`nth(1)` could silently target the wrong row. The delete-button assertions already use a safer row-scoped pattern — consider applying that same approach here.

- **`page.on('request', ...)` listener leak**: The "saving with an unchanged Cost value does not send a rates batch request" tests (in both spec files) attach a request listener to the shared `page` with no cleanup, and listeners accumulate across the remaining serial tests. Prefer `page.waitForRequest(pred, { timeout: 1000 }).catch(() => null)` or `page.once`.

- **`fulfillMarketUpdate` / `fulfillMarketDelete` share the same mock discriminator**: Both register a handler for `POST /wc/gla/mc/markets/<id>` (WordPress's apiFetch middleware maps both PUT and DELETE to POST). If ever registered for the same ID in the same test scope, only the most recent handler wins for both actions. Current tests avoid this by using different IDs, but it's a hidden footgun worth a doc note.

- **AC/behavior mismatch on API errors is still unresolved**: The AC states "snackbar is shown and modal closes," but every API error test in the PR asserts the modal *stays open* instead. This looks like the correct actual behavior, but the AC wording should be corrected so there's no ambiguity for future contributors.

### 🔵 Nitpicks
*Minor items — take or leave.*

- `MarketsPage.fulfillShippingRates` vs. `MockRequests.fulfillShippingRates` use different URL-matching regexes (negative lookahead vs. word boundary) — harmless today, but worth aligning.
- `mockIsMultiLingualStore` stacks `addInitScript` calls across describe blocks rather than replacing; it works only because of `configurable: true` redefinition — a bit opaque for future readers.
- `SECONDARY_MARKET` fixture has no `countries` field while `PRIMARY_MARKET` does — worth adding for fixture consistency.
- Several tests combine two distinct modal interactions under one long conjunctive test name — splitting would improve failure diagnostics.

---
*This review was generated by Claude. A human review is still required before merging.*

joemcgill (approved):
Thanks for the updates, @AlejandroPerezMartin. I've refreshed with the latest develop and rerun tests and everything is passing. AI review still had a few suggestions, but I think those can be handled as iterative follow-up improvements to this E2E tests for this feature.

## 🤖 AI Code Review

**Issue:** GOOWOO-615 – Add E2E tests for Multi-Lingual support
**Branch:** `add/GOOWOO-615-e2e-tests-for-multi-lingual-support` → `develop`

### Summary

The implementation covers all six scenario combinations required by the acceptance criteria and follows the existing Playwright page-object conventions well. The `MarketsPage` utility is clearly documented, and the `mockIsMultiLingualStore` / `addInitScript` approach correctly injects the flag before page scripts execute. No critical issues were found; there are a few suggestions that would improve test robustness and coverage completeness.

### 🟡 Suggestions

- **Table refresh is not explicitly asserted for edit and create flows** (`tests/e2e/specs/markets/markets-multilingual.test.js`, `tests/e2e/specs/markets/markets-non-multilingual.test.js`): The acceptance criterion says "successful save closes the modal and refreshes the table." For delete, the tests do verify the table state explicitly (asserting the France cell disappears after deletion). For edit and create, `fulfillMarkets()` is registered as a mock but the test only asserts modal closure — it never asserts that the table actually shows the updated/new market data. Adding at least one explicit table-state assertion per success path (e.g., that a newly created market label appears, or that an updated count is correct) would fully satisfy the criterion and catch a regression where the UI closes the modal but does not re-fetch.

- **`mockIsMultiLingualStore` is called redundantly in every inner `beforeAll`** (`tests/e2e/utils/pages/markets.js`, both test files): Each `test.describe` block calls `mockIsMultiLingualStore` in its `beforeAll`, which accumulates a new `addInitScript` handler per navigation for the lifetime of the page. Because `Object.defineProperty` is called with `configurable: true`, each new handler overrides the previous one, so the correct value wins on every navigation — the pattern is technically sound for the current uniform-value-per-file structure. However, moving `mockIsMultiLingualStore(true/false)` to the outer `beforeAll` (where the page is created) would be cleaner, eliminate the handler accumulation, and make the intent explicit. If a future developer wants to mix values across describe blocks in the same file, they would need to understand the override semantics, which is non-obvious.

- **`fulfillMarketUpdate` and `fulfillMarketDelete` are functionally identical** (`tests/e2e/utils/pages/markets.js`, lines ~143–182): Both register a `POST` handler for the same URL regex. As documented in the JSDoc, this is correct because WordPress's `apiFetch` http-v1 middleware converts both PUT and DELETE to POST with an `X-HTTP-Method-Override` header. Playwright's `route()` does not filter on headers, so the two mock methods produce indistinguishable registrations. The naming distinction is useful for readability, but it means registering both simultaneously would produce ambiguous handler stacking. A brief comment at the call sites where both are registered in close proximity would help future readers understand why the prior `fulfillMarketUpdate` registration does not interfere.

### 🔵 Nitpicks

- (`tests/e2e/specs/markets/markets-multilingual.test.js`, line 156; line 876): A handful of test bodies cover two distinct scenarios separated by a comment (e.g., "secondary market edit has no audience section but shows shipping notice; Add modal shows country select and notice"). Splitting these into two `test()` calls would give Playwright a more precise failure location when one of the two scenarios regresses.

- (Snackbar assertions, e.g. line 277 in markets-multilingual.test.js): `.last().toBeVisible()` is used — the comment correctly explains why `.last()` is needed, but the test does not verify the message content. Asserting something like `toContainText('error')` would distinguish a genuine error snackbar from a stale success snackbar that happens to still be visible from a prior test.

- (`tests/e2e/specs/markets/markets-multilingual.test.js`, line 456): The flat-rate + multilingual primary market modal asserts `getShippingInfoNotice(modal).not.toBeVisible()`. Elsewhere when the notice is absent, `not.toBeAttached()` is used. If the shipping notice component is conditionally rendered rather than CSS-hidden, `not.toBeAttached()` is more semantically correct and will catch regressions where the component renders but is invisible rather than absent.

---

*This review was generated by Claude. A human review is still required before merging.*

joemcgill:
Follow-up issue logged: GOOWOO-918

PR #3466, opened 2026-06-01
GOOWOO-709: Migrate product delete sync to MAPI

alaca:

Closes https://linear.app/a8c/issue/GOOWOO-709/migrate-product-delete-sync-to-mapi

This PR removes the Content API for deleting products, and it adds support for MAPI `productInputs.delete`

1. Add a product, and confirm it shows in Merchant Center
2. Trash it or delete it
3. Confirm the product is gone from Merchant Center

You will have to wait for a few min for the sync to happen 

### Changelog entry

>

### DIFF
--- a/src/Jobs/Update/CleanupProductTargetCountriesJob.php
+++ b/src/Jobs/Update/CleanupProductTargetCountriesJob.php
@@ -50,7 +50,7 @@ class CleanupProductTargetCountriesJob extends AbstractProductSyncerBatchedJob {
 	 */
 	protected function process_items( array $items ) {
 		$products      = $this->product_repository->find_by_ids( $items );
-		$stale_entries = $this->batch_product_helper->generate_stale_countries_request_entries( $products );
-		$this->product_syncer->delete_by_batch_requests( $stale_entries );
+		$stale_entries = $this->batch_product_helper->generate_stale_countries_delete_entries( $products );
+		$this->product_syncer->delete_mapi_entries( $stale_entries );
 	}
 }
--- a/src/Product/BatchProductHelper.php
+++ b/src/Product/BatchProductHelper.php
@@ -309,6 +309,66 @@ class BatchProductHelper implements Service {
 	}
 
 	/**
+	 * Generate MAPI delete entries for the given products.
+	 *
+	 * @param WC_Product[] $products
+	 *
+	 * @return array<int, array{wc_product_id: int, country: string, google_id: string, input: ProductInput}>
+	 */
+	public function generate_mapi_delete_entries( array $products ): array {
+		$entries = [];
+
+		foreach ( $products as $product ) {
+			$this->validate_instanceof( $product, WC_Product::class );
+
+			if ( $product instanceof WC_Product_Variable ) {
+				$entries = array_merge( $entries, $this->generate_mapi_delete_entries( $product->get_available_variations( 'objects' ) ) );
+				continue;
+			}
+
+			$google_ids = $this->product_helper->get_synced_google_product_ids( $product );
+			if ( empty( $google_ids ) ) {
+				continue;
+			}
+
+			foreach ( $google_ids as $country => $google_id ) {
+				$identity = $this->parse_mapi_identity( (string) $google_id );
+				if ( null === $identity ) {
+					continue;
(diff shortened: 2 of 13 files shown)


src/Product/BatchProductHelper.php
```diff
@@ -308,6 +308,66 @@ public function generate_mapi_update_entries( array $products ): array {
...
+	 * @return array{0: string, 1: string, 2: string}|null
+	 */
+	public function parse_mapi_identity( string $google_id ): ?array {
+		$parts = explode( '~', $google_id, 3 );
+		if ( count( $parts ) !== 3 ) {
+			return null;
+		}
+
+		return [ $parts[0], $parts[1], $parts[2] ];
+	}
```
joemcgill, line 368: What happens to products that were synced to the Content API. Does the MAPI automatically return the new `en~US~gla_29` format ID, or do we need a compat shim for the previous colon-delimited format, `online:en:US:gla_123`?

src/Product/ProductSyncer.php
```diff
@@ -294,45 +293,90 @@ public function delete( array $products ): BatchProductResponse {
...
+			$identity = $this->batch_helper->parse_mapi_identity( (string) $google_id );
+			if ( null === $identity ) {
+				continue;
+			}
+
+			[ $language, $feed, $offer_id ] = $identity;
+
+			$entries[] = [
+				'wc_product_id' => (int) $wc_product_id,
+				'country'       => $feed,
```
joemcgill: `$feed` might be conflated with the feedLabel attribute added to products, which doesn't necessarily need to be a country code. I'm unsure if it would be better to rename the property name or the PHP variable here, but it's something to consider.

src/Jobs/Update/CleanupProductTargetCountriesJob.php
```diff
@@ -50,7 +50,7 @@ public function get_batch( int $batch_number ): array {
 	 */
 	protected function process_items( array $items ) {
 		$products      = $this->product_repository->find_by_ids( $items );
-		$stale_entries = $this->batch_product_helper->generate_stale_countries_request_entries( $products );
-		$this->product_syncer->delete_by_batch_requests( $stale_entries );
+		$stale_entries = $this->batch_product_helper->generate_stale_countries_delete_entries( $products );
+		$this->product_syncer->delete_mapi_entries( $stale_entries );
```
joemcgill, line 54: `delete_by_batch_requests()` seems to have been replaced by `delete_by_id_map`, which includes a `validate_merchant_center_setup()` step. By replacing calls to this method with `delete_mapi_entries()` instead, we're missing that step. Is that intentional?

joemcgill (commented):
Just a couple of questions about this one.

src/Product/ProductSyncer.php
alaca: Good catch, that value is the `feedLabel` segment, and you are right, it isn't necessarily a country. It's also not actually consumed anywhere, `delete_mapi_entries() `only uses input, `wc_product_id`, and `google_id`, so rather than rename it, I'll just drop the unused country key from the delete entries

src/Jobs/Update/CleanupProductTargetCountriesJob.php
alaca: Yeah, that's not intentional. The old `delete_by_batch_requests()` ran `validate_merchant_center_setup()`, and by routing the cleanup jobs straight to `delete_mapi_entries()`, I dropped it. Every delete path goes through `delete_mapi_entries()` (including `delete()` and `delete_by_id_map()`), I'll move the validation there and drop the redundant calls from the two wrappers.

src/Product/BatchProductHelper.php
alaca: There is no automatic conversion, so Content API products keep their format ID and get skipped. The plan is to re-sync everything through MAPI with the existing Update All Products job so we don't need to keep extra code around to handle the old ID format

joemcgill (approved):
Thanks for confirming my questions and making the updates. Looks good to me. Let's wait to merge this until after GOOWOO-692 lands to keep that PR clean.

PR #3470, opened 2026-06-01
GOOWOO-719: Consolidate language and currency data fetching

asvinb:

Closes https://linear.app/a8c/issue/GOOWOO-719/consolidate-languagecurrency-resolver-into-a-shared-hook

1. Set up WPML and select any shipping rate other than flat rate.
2. Configure languages and currencies in WPML.
3. Verify that the configured languages and currencies are displayed in the “Add Market” and “Edit Market” modals.

### Changelog entry

>


js/src/hooks/useAvailableStoreLanguages.js
```diff
@@ -0,0 +1,29 @@
...
+ * Delegates to useAvailableLanguagesCurrencies so both values share one resolver call.
+ *
+ * @return {AvailableStoreLanguages} Available languages and resolution status.
+ */
+const useAvailableStoreLanguages = () => {
+	const { languages, hasFinishedResolution } =
+		useAvailableLanguagesCurrencies();
+
+	return { languages, hasFinishedResolution };
+};
```
AlejandroPerezMartin, line 27: this hook is not really doing anything, just destructing the data. Both language-cell.js and language-select-control.js could directly use useAvailableLanguagesCurrencies

js/src/hooks/useAvailableStoreCurrencies.js
```diff
@@ -1,30 +1,29 @@
...
+ * @return {AvailableStoreCurrencies} Available currencies and resolution status.
+ */
 const useAvailableStoreCurrencies = () => {
-	return useSelect( ( select ) => {
-		const selector = select( STORE_KEY );
-
-		// Trigger the shared resolver that populates both languages and currencies.
-		selector[ selectorName ]();
+	const { currencies, hasFinishedResolution } =
+		useAvailableLanguagesCurrencies();
```
AlejandroPerezMartin, line 25: same here, this hook is just destructing 2 props, the hook callers could simply use useAvailableLanguagesCurrencies instead

js/src/hooks/useAvailableLanguagesCurrencies.js
```diff
@@ -0,0 +1,50 @@
...
+		// Trigger the shared resolver that populates both languages and currencies.
+		selector[ selectorName ]();
+
+		return {
+			languages: selector.getAvailableLanguages(),
+			currencies: selector.getAvailableStoreCurrencies(),
+			hasFinishedResolution: selector.hasFinishedResolution(
+				selectorName,
+				[]
+			),
```
AlejandroPerezMartin, line 45: consumers that just need languages or currencies now subscribe to both selectors, even though there're two hooks that destruct this below, it's still fetching both data sources despite they are used or not. It'd make more sense to have 3 hooks, one that retrieves the languages, another for the currencies and another that gets both.

AlejandroPerezMartin (changes requested):

js/src/hooks/useAvailableLanguagesCurrencies.js
asvinb: @AlejandroPerezMartin Currently the backend exposes a single endpoint (`/mc/markets/languages-currencies`) that returns both languages and currencies in one response. Splitting into three hooks wouldn't reduce network requests since all three would still trigger the same resolver. Additionally, `@wordpress/data` deduplicates in-flight resolver calls, so even if multiple components each call a hook that subscribes to this data, the request only fires once.

js/src/hooks/useAvailableStoreLanguages.js
asvinb: @AlejandroPerezMartin  It's intentional since if we later introduce a dedicated languages endpoint and migrate the underlying resolver, only this hook changes. Having components use `useAvailableLanguagesCurrencies` directly would expose them to data they don't need and couple them to the current single-endpoint implementation detail.

js/src/hooks/useAvailableStoreCurrencies.js
asvinb: @AlejandroPerezMartin See [my reply here](https://github.com/woocommerce/google-listings-and-ads/pull/3470#discussion_r3339976693).

js/src/hooks/useAvailableLanguagesCurrencies.js
AlejandroPerezMartin: ah ok fine then, approved :)

AlejandroPerezMartin (approved):

PR #3474, opened 2026-06-02
Setting up notifications architecture.

mikaeltomovic:

Closes:

- [[GOOWOO-647] NotificationsPanel + notification components](https://linear.app/a8c/issue/GOOWOO-647/notificationspanel-notification-components)
- [[GOOWOO-648] Webpack entry point + PHP enqueuing](https://linear.app/a8c/issue/GOOWOO-648/webpack-entry-point-php-enqueuing)

Implements the notification system UI layer for the WooCommerce Marketing overview page. This includes the React panel that renders actionable notifications, all 13 individual notification card components with correct copy and merchant-type variants, the webpack entry point that mounts the panel via `MutationObserver`, and the PHP service that enqueues the bundle only on `page=wc-admin&path=/marketing`.

Also fixes a console warning introduced by having two webpack entry points on the same page — the `wc/gla` store registration in `data/index.js` is now guarded so it only runs if the store has not already been registered by another bundle on the page.

The notification panel is fully wired to the `wc/gla` data store (`getNotifications`, `dismissNotification`) and will render automatically once the backend notification cases (GOOWOO-652–676) are implemented and the `/wc/gla/notifications` endpoint is live.

### Screenshots:

N/A — the panel requires the backend notifications API (GOOWOO-652+) to return data before it is visible. See test instructions below for how to verify the UI manually with injected data.

1. Build the JS bundle: `npm run build` (or `npm start` for hot reload).
2. Navigate to `/wp-admin/admin.php?page=wc-admin&path=/marketing`.
3. Confirm that you do not see the `"Store wc/gla is already registered"` warning in the browser console.
4. Inject notifications directly into the store via the console:

```
wp.data.dispatch('wc/gla').receiveNotifications([
    { id: 'paused-campaign', triggered_at: Math.floor(Date.now() / 1000) },
    { id: 'tracking-off', triggered_at: Math.floor(Date.now() / 1000) },
]);

```
5. Confirm the notification cards render after the banner with the correct title, description, date, and CTA button.
6. Run the JS unit tests: `npm run test:js -- --testPathPattern="notification-system"` — all 20 tests should pass.

**Note:** Clicking the dismiss button of a card will cause an error in the browser console and the card won't disappear from the DOM yet. This is expected since the backend logic that handles dismissal hasn't been implemented yet.

### Changelog entry

> Add - Notification system panel that renders actionable notifications on the WooCommerce Marketing overview page (GOOWOO-647)
> Add - 13 notification card components covering all notification cases with shopping/service-based merchant copy variants (GOOWOO-647)
> Add - `notification-system` webpack entry point and PHP service to enqueue the bundle on the marketing overview page (GOOWOO-648)
> Fix - Prevent duplicate `wc/gla` store registration warning when multiple webpack bundles are loaded on the same admin page


js/src/data/index.js
```diff
@@ -18,15 +18,17 @@ import reducer from './reducer';
...
 import { getReconnectAccountUrl } from '~/utils/urls';
 
-registerStore( STORE_KEY, {
-	actions,
-	selectors,
-	resolvers,
-	controls,
-	reducer,
-} );
+if ( ! select( STORE_KEY ) ) {
```
asvinb: @mikaeltomovic Why the changes in this file please?

js/src/notifications-system/notifications-panel.test.js
```diff
@@ -0,0 +1,199 @@
+/**
```
asvinb, line 1: @mikaeltomovic I missed that in the ticket but let's have the tests in the same folder as the main component. Sorry about that.

js/src/notifications-system/notifications-panel.test.js
```diff
@@ -0,0 +1,199 @@
...
+ * External dependencies
```
asvinb, line 2: @mikaeltomovic Also let's name the folder `notifications-system` to be consistent with the feature we building.

js/src/notifications-system/notifications/abandoned-onboarding.js
```diff
@@ -0,0 +1,35 @@
...
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Notification from '../notification';
+
+const AbandonedOnboarding = ( { triggeredAt, onDismiss } ) => {
```
asvinb: @mikaeltomovic JSDocs missing and this applies to the other components below as well.

js/src/notifications-system/notifications/coupons-not-synced.js
```diff
@@ -0,0 +1,48 @@
...
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import useServiceBasedMerchant from '~/hooks/useServiceBasedMerchant';
+import Notification from '../notification';
+
+const CouponsNotSynced = ( { triggeredAt, onDismiss } ) => {
+	const isServiceBased = useServiceBasedMerchant();
```
asvinb: @mikaeltomovic Let's check if there's a connected MC account instead of using the `useServiceBasedMerchant` hook.

js/src/notifications-system/notifications/not-onboarded-90-days.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import useServiceBasedMerchant from '~/hooks/useServiceBasedMerchant';
```
asvinb: @mikaeltomovic See my previous comment about checking whether there's a connected MC account. Same comment applies to any components using the `useServiceBasedMerchant` hook.

js/src/notification-system/index.scss
```diff
@@ -0,0 +1,66 @@
+.gla-notification-system-container:not(:has(*)) {
```
asvinb: @mikaeltomovic The file name is `index.scss` but it's being imported by `notification.js`. Let's rename it to `notification.scss`

js/src/notification-system/index.scss
```diff
@@ -0,0 +1,66 @@
...
+	display: none;
+}
+
+.gla-notifications-panel {
```
asvinb: @mikaeltomovic This style should live in the `notifications-panel` SCSS file and imported in the `NotificationsPanel` component.

js/src/notification-system/index.scss
asvinb: @mikaeltomovic Where is that class defined please?

js/src/notification-system/index.scss
```diff
@@ -0,0 +1,66 @@
...
+	display: flex;
+	flex-direction: column;
+	gap: 32px;
```
asvinb: @mikaeltomovic Let's use the `$grid-unit-*` variables where possible please.

js/src/notification-system/index.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.gla-notification {
+	&__header {
```
asvinb: @mikaeltomovic Let's avoid concatenating the CSS selectors. It makes it hard to read and debug 👍

js/src/notification-system/index.scss
```diff
@@ -0,0 +1,66 @@
...
+	&__logo {
+		display: block;
+	}
+
+	&__dismiss {
+		background: none;
+		border: none;
+		cursor: pointer;
+		padding: 4px;
+		color: #757575;
```
asvinb: @mikaeltomovic Any existing color variables we can use please?

js/src/notification-system/index.scss
```diff
@@ -0,0 +1,66 @@
...
+
+	&__logo {
+		display: block;
+	}
+
+	&__dismiss {
+		background: none;
+		border: none;
+		cursor: pointer;
+		padding: 4px;
```
asvinb: @mikaeltomovic Let's use the `grid-unit-*` values.

js/src/notification-system/notification.js
```diff
@@ -0,0 +1,79 @@
...
+ * @param {Array}    props.actions     Array of AppButton prop objects for CTA buttons.
+ */
+const Notification = ( {
+	title,
+	description,
+	triggeredAt,
+	onDismiss,
+	actions = [],
+} ) => {
+	const formattedDate = triggeredAt
```
asvinb: @mikaeltomovic From the JSDocs, `triggeredAt` is not optional, hence we don't need to check if it's defined or not because we know for sure it'll always be defined.

js/src/notifications-system/notification.js
```diff
@@ -0,0 +1,79 @@
...
+				) }
+			</div>
+			<div className="gla-notification__body">
+				<h3 className="gla-notification__title">{ title }</h3>
+				{ formattedDate && (
+					<p className="gla-notification__date">{ formattedDate }</p>
+				) }
+				<p className="gla-notification__description">{ description }</p>
+			</div>
+			{ actions.length > 0 && (
```
asvinb: @mikaeltomovic The actions are too tied to `AppButton`. Let's use the same pattern as in `AppModal` and render `buttons` which should be provided by the parent component.

js/src/notifications-system/notification.js
```diff
@@ -0,0 +1,79 @@
...
+						) }
+					>
+						&#x2715;
+					</button>
+				) }
+			</div>
+			<div className="gla-notification__body">
+				<h3 className="gla-notification__title">{ title }</h3>
+				{ formattedDate && (
+					<p className="gla-notification__date">{ formattedDate }</p>
```
asvinb: @mikaeltomovic Can you review the markup to match the designs please? For e.g the date is is next to the buttons.

js/src/notifications-system/notification.js
```diff
@@ -0,0 +1,79 @@
...
+				) }
+				<p className="gla-notification__description">{ description }</p>
+			</div>
+			{ actions.length > 0 && (
+				<div className="gla-notification__actions">
+					{ actions.map( ( actionProps, index ) => (
+						<AppButton key={ index } { ...actionProps } />
```
asvinb: @mikaeltomovic Using the `index` as key is an anti pattern. If we use the same approach as on `AppModal`, we can set the key as part of each button of the `buttons` prop.

js/src/notifications-system/notification.js
```diff
@@ -0,0 +1,79 @@
...
+			<div className="gla-notification__header">
+				<img
+					src={ googleLogoURL }
+					alt="Google"
+					className="gla-notification__logo"
+					width="64"
+					height="22"
+				/>
+				{ onDismiss && (
+					<button
```
asvinb: @mikaeltomovic Can we use `AppButton`?

js/src/notifications-system/notifications-panel.js
```diff
@@ -0,0 +1,105 @@
...
+	'enhanced-conversions-off': EnhancedConversionsOff,
+	'recommendations-available': RecommendationsAvailable,
+	'coupons-not-synced': CouponsNotSynced,
+	'sales-not-growing': SalesNotGrowing,
+	'tracking-off': TrackingOff,
+};
+
+const NotificationsPanel = () => {
+	const notifications = useNotifications();
+	const { dismissNotification, invalidateResolutionForStoreSelector } =
```
asvinb: @mikaeltomovic Let's use `useAppDispatch` hook which is basically the same thing as `useDispatch( STORE_KEY )`.

js/src/notifications-system/notifications-panel.js
```diff
@@ -0,0 +1,105 @@
...
+	'paused-campaign': PausedCampaign,
+	'active-campaign-zero-sales': ActiveCampaignZeroSales,
+	'enhanced-conversions-off': EnhancedConversionsOff,
+	'recommendations-available': RecommendationsAvailable,
+	'coupons-not-synced': CouponsNotSynced,
+	'sales-not-growing': SalesNotGrowing,
+	'tracking-off': TrackingOff,
+};
+
+const NotificationsPanel = () => {
```
asvinb, line 27: @mikaeltomovic Missing JSDocs

webpack.config.js
```diff
@@ -153,6 +153,11 @@ const webpackConfig = {
 			'js/src/notification-manager',
 			'index.js'
 		),
+		'notification-system': path.resolve(
```
asvinb:
```suggestion
		'notifications-system': path.resolve(
```

webpack.config.js
```diff
@@ -153,6 +153,11 @@ const webpackConfig = {
...
+			process.cwd(),
+			'js/src/notification-system',
```
asvinb:
```suggestion
			'js/src/notifications-system',
```

js/src/notifications-system/notifications/abandoned-onboarding.js
```diff
@@ -0,0 +1,35 @@
...
+			description={ __(
+				'Your Google Ads integration setup was interrupted. Complete the remaining configuration steps to ensure your store data is properly synced. Get $500 USD or more in Google ad credit. Offer for new advertisers only. Terms apply.',
+				'google-listings-and-ads'
+			) }
+			triggeredAt={ triggeredAt }
+			onDismiss={ onDismiss }
+			actions={ [
+				{
+					isPrimary: true,
+					href: 'admin.php?page=wc-admin&path=/google/start',
```
asvinb: @mikaeltomovic Let's use the `getGetStartedUrl` function. The same applies for other components where we should use the existing functions we have as much as possible.

js/src/notifications-system/notifications/sold-10-items.js
```diff
@@ -0,0 +1,38 @@
...
+
+const Sold10Items = ( { triggeredAt, onDismiss } ) => {
+	return (
+		<Notification
+			title={ __(
+				'Drive more sales with Google Ads',
+				'google-listings-and-ads'
+			) }
+			description={ __(
+				"Congrats on your first 10 sales – now let's find your next customer. Reach high-intent shoppers across Google. Get $500 USD or more in Google ad credit. Offer for new advertisers only. Terms apply.",
```
asvinb: @mikaeltomovic Can you double check whether we need to link the `Terms apply` string?

js/src/notifications-system/notification.js
```diff
@@ -0,0 +1,79 @@
...
+						type="button"
+						className="gla-notification__dismiss"
+						onClick={ onDismiss }
+						aria-label={ __(
+							'Dismiss notification',
+							'google-listings-and-ads'
+						) }
+					>
+						&#x2715;
```
asvinb: @mikaeltomovic Can we use the `closeSmall` icon from the `@wordpress/icons` package please?

js/src/notifications-system/notification.js
```diff
@@ -0,0 +1,79 @@
...
+	triggeredAt,
+	onDismiss,
+	actions = [],
+} ) => {
+	const formattedDate = triggeredAt
+		? dateI18n( glaData.dateFormat, new Date( triggeredAt * 1000 ) )
+		: null;
+
+	return (
+		<Card className="gla-notification">
```
asvinb: @mikaeltomovic I don't see much benefits of using the `Card` component here when looking at the designs. Let me know what you think?

js/src/notification-system/notifications-panel.js
```diff
@@ -0,0 +1,105 @@
...
+	'tracking-off': TrackingOff,
+};
+
+const NotificationsPanel = () => {
+	const notifications = useNotifications();
+	const { dismissNotification, invalidateResolutionForStoreSelector } =
+		useDispatch( STORE_KEY );
+
+	useEffect( () => {
+		const badge = document.querySelector(
```
asvinb: @mikaeltomovic I think the badge count should be handled by the backend. However we need to update the count when we dismiss notifications. Can you double check please, knowing that we have `js/src/notification-manager/index.js` as well.

asvinb (changes requested):
@mikaeltomovic In addition to the inline comments, I noticed an architectural concern worth addressing.
* `triggeredAt` and `onDismiss` are currently passed down two levels from `NotificationsPanel` through each of the 13 individual notification components. Even though those components never actually use them; they just relay them to Notification. This means if we ever need to add another prop to `Notification`, we'd have to update all 13 components, which doesn't scale well.

I'd suggest the following refactor:
* Add a `useNotificationsSystemMap` hook that returns the `title`, `description`, and `buttons` for each notification type, keyed by notification ID. The map should be memoized for performance. Static entries can live at module level (created once on import), and any dynamic entries (e.g. ones that depend on other hooks) should be merged in via `useMemo` with the appropriate dependencies.
* Move `onDismiss` into `Notification`. Since the dismiss action is always `dismissNotification(id)` with no variation, there's no reason for the parent to inject it as a callback. `Notification` can accept an `id` prop and call `useDispatch` internally to handle dismissal itself.
* In `NotificationsPanel`, loop through the notifications, look up the config from the hook by ID, and render `Notification` directly, passing only id, triggeredAt, and the title, description, buttons props. 
* This removes the 13 individual wrapper components and reduces the `NotificationsPanel` → `Notification` interface to just the data that actually belongs at that boundary.

js/src/data/index.js
mikaeltomovic: This wrapper checks whether the store is already registered before attempting to register it again. I saw an error message in the console saying that the `wc/gla` store was already registered so I added this wrapper to prevent that.

js/src/notifications-system/notifications/coupons-not-synced.js
mikaeltomovic: @asvinb I don't understand why you want to make that change. Currently the `isServiceBased` boolean is only used to determine what the copy should be. 

Checking if there is a connected MC account doesn't achieve the same thing as `useServiceBasedMerchant()`.

When you say that we should check if there is a connected MC account, are you talking about the logic that determines wether or not the notification is displayed?

mikaeltomovic: I won't reply to every comment about this (better to keep the conversation in one place), but please take another look at this and let me know what should be changed. [The ticket](https://linear.app/a8c/issue/GOOWOO-672/fe-notification-case-8-coupons-not-synced) for this notification states that there should be different copy for shopping merchants vs. service-based merchants and I thought this was the best way to do it.

js/src/notification-system/index.scss
mikaeltomovic: It is defined in `js/src/notifications-system/index.js` as `CONTAINER_CLASS`.

js/src/notification-system/notification.js
mikaeltomovic: I disagree that the JSDoc alone justifies removing the check. Unlike TypeScript, JSDoc has no enforcing power at runtime, so JavaScript would happily allow an empty or falsy value to be passed regardless of how the prop is documented.

That said, I'm okay with removing the check in this specific case because this component is only consumed internally by the notifications API, which always provides `triggered_at`. Given that, the defensive guard is redundant in practice and we can clean it up.

js/src/notification-system/notifications-panel.js
mikaeltomovic: I think you're right and that this can be handled by the backend. I'll remove it.

js/src/notifications-system/notifications/sold-10-items.js
mikaeltomovic: The ticket doesn't mention it.
https://linear.app/a8c/issue/GOOWOO-656/fe-notification-case-22-sold-10-items

js/src/data/index.js
asvinb: Thanks @mikaeltomovic . Can you add a comment explaining this so that other engineers are aware of it, please?

js/src/notifications-system/notifications/coupons-not-synced.js
asvinb: @mikaeltomovic, a service-based merchant is someone who does not sell physical products and therefore does not require a Merchant Center account or connection.

The onboarding flow is designed around this concept. When we detect that a store has no physical products, we show an alternative onboarding experience that does not prompt the merchant to connect a Merchant Center account.

The `useServiceBasedMerchant` hook is primarily intended to determine which onboarding experience should be shown. Once onboarding is complete, the rest of the UI should only check whether a Merchant Center account is connected. This separation ensures that future changes to the onboarding logic do not inadvertently affect other parts of the product.

For example, a merchant may onboard at a time when `useServiceBasedMerchant` returns `true`, but later decide to start selling physical products. In that case, the hook would begin returning `false`, yet the merchant would still have completed onboarding without a Merchant Center account. That is expected behavior. If they later disconnect and restart onboarding, they would then be presented with the Merchant Center connection flow and have the opportunity to connect their account.

js/src/notification-system/notification.js
asvinb: @mikaeltomovic Good point on JSDoc not enforcing anything at runtime. I was mostly looking at it from a consistency standpoint, i.e., if we need defensive guards, the docs should probably reflect that it's optional. But completely agree that deleting the redundant check is the best approach here since it's only used internally!

js/src/notifications-system/notifications/sold-10-items.js
asvinb: I've enquired and will let you know.

js/src/notifications-system/notification.js
mikaeltomovic: @asvinb I am struggling to understand what you are asking for here. Are you saying you don't want us to put the markup for `AppButton` in this file but rather you want to have a prop called `buttons` passed to the `Notification` component? That would mean, if I understand it correctly, that in `js/src/notifications-system/notifications-panel.js` we need to extract the `actions` from `config` and build `AppButton` components there, which are then passed to `Notification`. Is that the change you are asking for?

asvinb: @mikaeltomovic That was my initial thought as well, but after reviewing the code, it seems the only logical place to introduce the `AppButton` would be inside the `useNotificationsSystemMap` hook. However, having JSX inside a hook is generally considered an anti-pattern, and we should avoid that.

Given that, I suggest we keep the current approach as is. If you have a cleaner or more scalable idea, I’m happy to reconsider 😁 .

js/src/data/index.js
asvinb: @mikaeltomovic I noticed the comment was marked as resolved, but I couldn't find the corresponding comment in the code. Could you add it please? 😁

js/src/notifications-system/index.js
```diff
@@ -0,0 +1,65 @@
+/* global MutationObserver */
```
asvinb: @mikaeltomovic Let's update the `globals` property in `.eslintrc` to include `MutationObserver` and remove from the file here.

js/src/notifications-system/notification.js
mikaeltomovic: @asvinb I think the current approach is better than any alternative I can think of. Thank you for looking into it again!

js/src/notifications-system/index.js
asvinb: Let's also update `js/src/notification-manager/index.js` to remove `MutationObserver` as well 👍

js/src/data/index.js
```diff
@@ -18,15 +18,17 @@ import reducer from './reducer';
...
+		resolvers,
+		controls,
+		reducer,
+	} );
 
-dispatch( STORE_KEY ).hydratePrefetchedData( glaData.initialWpData );
+	dispatch( STORE_KEY ).hydratePrefetchedData( glaData.initialWpData );
+}
 
 apiFetch.use(
```
asvinb, line 31: @mikaeltomovic Let's move this line within the `if` condition as well.

js/src/notifications-system/useNotificationsSystemMap.js
```diff
@@ -0,0 +1,358 @@
...
+	'abandoned-onboarding': {
+		title: __( 'Finish your Google Ads setup', 'google-listings-and-ads' ),
+		description: __(
+			'Your Google Ads integration setup was interrupted. Complete the remaining configuration steps to ensure your store data is properly synced. Get $500 USD or more in Google ad credit. Offer for new advertisers only. Terms apply.',
+			'google-listings-and-ads'
+		),
+		actions: [
+			{
+				id: 'continue-setup',
+				isPrimary: true,
```
asvinb: @mikaeltomovic Let's remove the `isPrimary` property from the buttons since it's not being used. `isPrimary` is not [among the props which are destructured](https://github.com/woocommerce/google-listings-and-ads/pull/3474/changes#diff-ead68f884e68ba90e641ddc00a62f18aa32901795459195a15cf1e1cf932541eR83-R87) when rendering the actions.

js/src/notifications-system/useNotificationsSystemMap.js
```diff
@@ -0,0 +1,358 @@
...
+						isPrimary: true,
+						href: getDashboardUrl(),
+						children: __(
+							'Resume your campaign',
+							'google-listings-and-ads'
+						),
+					},
+				],
+			},
+			'active-campaign-zero-sales': {
```
asvinb, line 269: @mikaeltomovic I see this notification defined [here as well](https://github.com/woocommerce/google-listings-and-ads/pull/3474/changes#diff-bfed954630fc7cd5fd8dc2ebd931fd31249b8c0f16968e666eef7a5012d4a40bR104). Can you check please?

js/src/notifications-system/notifications-panel.js
```diff
@@ -0,0 +1,94 @@
...
+ * whenever the browser tab becomes visible.
+ *
+ * @return {JSX.Element|null} A panel of notification cards, or null if there are no active notifications.
+ */
+const NotificationsPanel = () => {
+	const { notifications } = useNotifications();
+	const { invalidateResolutionForStoreSelector } = useAppDispatch();
+	const notificationMap = useNotificationsSystemMap();
+
+	useEffect( () => {
```
asvinb: @mikaeltomovic While this looks good, I feel it's a bit overkilled because GLA notifications aren't time-critical and will likely be cached anyway, so re-fetching on tab focus adds complexity with little benefit. Let's keep things simple: load once on mount, update on dismiss, refresh on page reload.

js/src/notifications-system/useNotificationsSystemMap.js
```diff
@@ -0,0 +1,358 @@
...
+		),
+		description: __(
+			'Turn on tracking to receive relevant recommendations on how you can grow your business.',
+			'google-listings-and-ads'
+		),
+		actions: [
+			{
+				id: 'turn-on-tracking',
+				isPrimary: true,
+				href: 'admin.php?page=wc-settings&tab=advanced',
```
asvinb: @mikaeltomovic Can we have a utility function for that please?

js/src/notifications-system/index.js
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}
+
+	if ( currentRoot ) {
+		currentRoot.unmount();
+	}
+
+	currentRoot = createRoot( container );
+	currentRoot.render( <NotificationsPanel /> );
+}
```
asvinb, line 53: @mikaeltomovic Once we have mounted `NotificationsPanel`, do you think we disconnect the observer ? Or the component can be wiped out when parent component (where we are injecting the panel) refreshes?

js/src/notifications-system/useNotificationsSystemMap.js
```diff
@@ -0,0 +1,358 @@
...
+		),
+		description: __(
+			"Congrats on your first 10 sales – now let's find your next customer. Reach high-intent shoppers across Google. Get $500 USD or more in Google ad credit. Offer for new advertisers only. Terms apply.",
+			'google-listings-and-ads'
+		),
+		actions: [
+			{
+				id: 'setup-ads-campaign',
+				isPrimary: true,
+				href: getSetupAdsUrl(),
```
asvinb: @mikaeltomovic Reading the description where there is the Ad credit, let's use the onboarding URL for this.

js/src/notifications-system/useNotificationsSystemMap.js
```diff
@@ -0,0 +1,358 @@
...
+		title: __( 'Finish your Google Ads setup', 'google-listings-and-ads' ),
+		description: __(
+			'Your Google Ads integration setup was interrupted. Complete the remaining configuration steps to ensure your store data is properly synced. Get $500 USD or more in Google ad credit. Offer for new advertisers only. Terms apply.',
+			'google-listings-and-ads'
+		),
+		actions: [
+			{
+				id: 'continue-setup',
+				isPrimary: true,
+				href: getGetStartedUrl(),
```
asvinb: @mikaeltomovic I see `getGetStartedUrl` is being called multiple times. Let’s store its result in a variable and reuse it instead of calling the function repeatedly. The same approach should be applied to the other URLs as well.

asvinb (changes requested):
Thanks @mikaeltomovic . I left a few comments. Can you check them out please?

js/src/data/index.js
mikaeltomovic: The comment is on line 21, above the if wrapper.

js/src/notifications-system/useNotificationsSystemMap.js
mikaeltomovic: @asvinb The onboarding URL is used for users who have not connected to a Google Account yet. I don't think that's the right CTA for this case.

The AC says "The CTA takes the merchant to the campaign setup flow inside G4W." and the current implementation will take the user to `/google/setup-ads` which I think makes sense in this scenario.

js/src/data/index.js
asvinb: @mikaeltomovic Actually, after taking another look, we shouldn’t have this condition here. On the Marketing Overview page, I can see that the GLA main plugin bundle is also being loaded, which shouldn’t be happening.

I’ve created [https://linear.app/a8c/issue/GOOWOO-747/restrict-indexjs-bundle-enqueue-to-gla-specific-admin-pages-only](https://linear.app/a8c/issue/GOOWOO-747/restrict-indexjs-bundle-enqueue-to-gla-specific-admin-pages-only) to address this properly.

Given that, let’s revert the changes in this file for now.

js/src/notifications-system/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Mounts the NotificationsPanel into the multichannel marketing overview element.
+ * Creates a container and inserts it after the introduction banner, or prepends
+ * it if no banner is present. Disconnects the observer once mounted.
+ *
+ * @param {Element}         multichannel The multichannel marketing overview element.
+ * @param {MutationObserver} obs          The observer to disconnect after mounting.
+ */
+function mount( multichannel, obs ) {
```
asvinb:
```suggestion
function mount( multichannelElement, mountObserver ) {
```
@mikaeltomovic Let's rename `obs` to `mountObserver` for clarity. The name better describes its role as the observer passed into mount specifically to be disconnected once the panel is mounted.

js/src/notifications-system/index.js
```diff
@@ -0,0 +1,73 @@
...
+const CONTAINER_CLASS = 'gla-notification-system-container';
+
+let currentRoot = null;
+
+/**
+ * Mounts the NotificationsPanel into the multichannel marketing overview element.
+ * Creates a container and inserts it after the introduction banner, or prepends
+ * it if no banner is present. Disconnects the observer once mounted.
+ *
+ * @param {Element}         multichannel The multichannel marketing overview element.
```
asvinb:
```suggestion
 * @param {Element} multichannel The multichannel marketing overview element.
```
@mikaeltomovic Let's remove the whitespace to be consistent with the codebase. Same applies to he line below.

js/src/notifications-system/index.scss
```diff
@@ -0,0 +1,3 @@
+.gla-notification-system-container:not(:has(*)) {
```
asvinb:
```suggestion
.gla-notification-system-container:empty {
```
@mikaeltomovic I think we can go with `:empty` here which has broader browser support and communicates intent more clearly.

asvinb (changes requested):
@mikaeltomovic Nice job here. Just a few minor things and then we can merge the branch. Thanks!

js/src/data/index.js
mikaeltomovic: @asvinb Okay, I am reverting this. The `Store "wc/gla" is already registered.` error is now appearing in the browser console.

asvinb: Sounds good. Thanks @mikaeltomovic

asvinb (approved):
LGTM @mikaeltomovic ! 🥳

PR #3475, opened 2026-06-02
GOOWOO-639 Core Notification Infrastructure

zamanq:

Adds the foundational notification infrastructure for Google for WooCommerce: a pluggable evaluator interface and a service that evaluates notification conditions in a uniform way, persists their state per user, and returns them ordered by priority.

This is the base layer only — concrete evaluators land in follow-up tickets (GOOWOO-642 / GOOWOO-643 / GOOWOO-644).

#### What's added
**NotificationEvaluatorInterface** — contract for a single notification with `get_id()`, `should_show()`, and `get_priority()`. 

Caching is each evaluator's responsibility, so the interface stays caching-agnostic.

**NotificationService** — ContainerAware service that: Resolves all evaluators as a tagged collection via the container, so no evaluators are injected through the constructor.

Sorts evaluators by `get_priority()` ascending and returns `[ 'id' => string, 'triggered_at' => int ]` entries.

Records `triggered_at` the first time a condition is met and never overwrites it.

Excludes permanently-dismissed notifications from all future results, via a public dismiss( string $id ) method.

Returns an empty array for users without the `manage_woocommerce` capability.

**State persistence** — stored in the `gla_notifications_state` `user_meta` key, keyed by notification ID with a `_version => 1` schema marker.

**WP proxy** — added `get_current_user_id()`, `get_user_meta()`, and `update_user_meta()` wrappers to keep WordPress access behind the proxy layer.

**Registration** — `NotificationService` registered in `CoreServiceProvider`.

Closes https://linear.app/a8c/issue/GOOWOO-639/core-notification-infrastructure

### Screenshots:
N/A

No evaluators ship in this PR, so there's nothing to click in the UI. To eyeball the wiring, paste this into wp shell while logged in as an admin:

```
$service = woogle_get_container()->get(
	\Automattic\WooCommerce\GoogleListingsAndAds\Notification\NotificationService::class
);
```
// No evaluators registered yet → expect an empty array, with no errors.

`var_dump( $service->get_notifications() );`
Expected: `array(0) {}` returned cleanly (confirms the service resolves from the container and the user-meta read path works). Concrete notifications appear once evaluators are added in GOOWOO/ 642 / 643 / 644.

### Additional details:

Ships Core Notification Infrastructure


src/Notification/NotificationService.php
```diff
@@ -0,0 +1,154 @@
...
+	 * Get the notifications that should currently be shown, ordered by priority (ascending).
+	 *
+	 * @return array[] List of [ 'id' => string, 'triggered_at' => int ] entries.
+	 */
+	public function get_notifications(): array {
+		if ( ! $this->can_manage() ) {
+			return [];
+		}
+
+		$evaluators = $this->container->get( NotificationEvaluatorInterface::class );
```
asvinb, line 69: `container->get()` will throw a `NotFoundException` when no `NotificationEvaluatorInterface` implementations are registered 

Maybe we should add a guard as follows:

```php
if ( ! $this->container->has( NotificationEvaluatorInterface::class ) ) {
    return [];
}
```

src/Notification/NotificationService.php
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	/**
+	 * Permanently dismiss a notification so it is excluded from all future results.
+	 *
+	 * @param string $id The notification ID to dismiss.
+	 *
+	 * @return void
+	 */
+	public function dismiss( string $id ): void {
```
asvinb, line 120: `dismiss()` is missing the `can_manage()` guard that `get_notifications()` enforces. A subscriber-level user (or unauthenticated request via a future REST route) can call this method and write to `gla_notifications_state` meta without any permission check.

asvinb (changes requested):

asvinb:
@zamanq I left a couple of comments. Can you check please?

src/Notification/NotificationService.php
zamanq: @asvinb I already added this guard in the GOOWOO-641 PR #3481 btw :) 
See here: https://github.com/woocommerce/google-listings-and-ads/pull/3481/changes#diff-9a167bd4f7ae38528eec606aa2998a4e8a7bdf330aaa30d78a5fad4a6d722e50R69

src/Notification/NotificationService.php
zamanq: @asvinb Added the guard. Thank you!

jamesmorrison (approved):

ankitguptaindia:
QA Status: Approved ✅ 
---

Test Result: 



```php
array(0) {
}
```


* Verified the `NotificationService` could be resolved from the container.
* Verified `get_notifications()` returned an empty array (`array(0) {}`) when no evaluators were registered.
* Verified no PHP warnings, notices, or fatal errors occurred during execution.


Screenshot 2026-06-23 at 9 59 10 PM

PR #3479, opened 2026-06-03
GOOWOO-640 REST API Controller for Notifications

zamanq:
#### Added notifications REST endpoints

**Summary**
Adds NotificationController exposing GET /wc/gla/notifications (returns active notifications) and 
DELETE /wc/gla/notifications/{id} (dismisses a notification and returns the updated list).

Wraps the existing NotificationService (get_notifications() / dismiss()); both routes use the WooCommerce can_manage() permission callback.

Registers the controller in RESTServiceProvider.

**Details**
Response shape: `{ notifications: [ { id: string, triggered_at: int }, ... ] }`.
The `{id} `route is constrained by the regex `[a-zA-Z0-9_-]+`, so a malformed id never matches the route (returns 404) and never reaches the service.

Closes https://linear.app/a8c/issue/GOOWOO-640/rest-api-controller

### Screenshots:
N/A

Log in as an admin (or a user with manage_woocommerce capability), then use the browser console on a wp-admin page:

**List notifications**
`wp.apiFetch( { path: '/wc/gla/notifications' } ).then( console.log );`
Expect `{ notifications: [ { id, triggered_at }, ... ] }`

**Dismiss one (use an id from step 1)**
`wp.apiFetch( { path: '/wc/gla/notifications/<id>', method: 'DELETE' } ).then( console.log );`
Expect 200 with the updated `{ notifications: [...] }` (the dismissed id is gone). Re-running step 1 confirms it stays dismissed.

**Invalid id → expect 404:**
`wp.apiFetch( { path: '/wc/gla/notifications/$$$', method: 'DELETE' } ).catch( console.log );`

Permissions → log out and hit either route → expect 401/403.

### Additional details:
This PR is based off of GOOWOO-639 Core Notification Infrastructure, PR: https://github.com/woocommerce/google-listings-and-ads/pull/3475 and needs to be merged after the mentioned branch is merged into feature/notifications-system branch


src/API/Site/Controllers/NotificationController.php
```diff
@@ -0,0 +1,153 @@
+<?php
+
+declare(strict_types=1);
```
asvinb: @zamanq PHP linting will fail on this line:

```suggestion
declare( strict_types=1 );
```

src/API/Site/Controllers/NotificationController.php
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	/**
+	 * Build the notifications response payload.
+	 *
+	 * @return array
+	 */
+	protected function get_notifications_response(): array {
+		return [
+			'notifications' => $this->service->get_notifications(),
```
asvinb, line 122: @zamanq As per `get_schema_properties()`, we don't appear to have a `notifications` property defined in the schema. Could you double-check that this matches the response structure?

tests/Unit/API/Site/Controllers/NotificationControllerTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+	public function test_delete_notification_invalid_id() {
+		$this->service->expects( $this->never() )
+			->method( 'dismiss' );
+
+		$response = $this->do_request( self::ROUTE . '/$$$', 'DELETE' );
+
+		$this->assertEquals( 404, $response->get_status() );
+	}
+}
```
asvinb, line 145: @zamanq Can we add tests for an unauthenticated user please for both the GET and DELETE endpoints?

asvinb (commented):
@zamanq I left a few comments. Can you check them out please? Thanks!

src/API/Site/Controllers/NotificationController.php
zamanq: Fixed

src/API/Site/Controllers/NotificationController.php
zamanq: @asvinb You're right. I've updated the schema def accordingly.

tests/Unit/API/Site/Controllers/NotificationControllerTest.php
zamanq: @asvinb Added the suggested tests for unauthenticated user

src/API/Site/Controllers/NotificationController.php
```diff
@@ -0,0 +1,177 @@
...
+
+	/**
+	 * Callback function for dismissing a notification.
+	 *
+	 * @return callable
+	 */
+	protected function get_delete_callback(): callable {
+		return function ( Request $request ) {
+			try {
+				$this->service->dismiss( (string) $request->get_param( 'id' ) );
```
jamesmorrison: @zamanq The `DELETE `route rejects malformed ids via the [a-zA-Z0-9_-]+ regex, but any well-formed id is passed through to `NotificationService::dismiss()` without checking that it corresponds to a registered evaluator.

So DELETE `/wc/gla/notifications/this_id_never_existed` returns `200` and writes `state['this_id_never_existed']['dismissed'] = true` to `gla_notifications_state`, which can be repeated to grow that user_meta entry indefinitely. The acceptance criterion "An invalid or unknown id format on DELETE returns a 400/404 error" reads to me as covering this case too.

Could we resolve the active notification ids (or expose a `has($id)` on `NotificationService`) and return `404` before calling `dismiss()`?

A matching test for the `well-formed-but-unknown-id` case would be worth adding alongside the existing `test_delete_notification_invalid_id`, which only covers the malformed-format path.

jamesmorrison (approved):

src/API/Site/Controllers/NotificationController.php
zamanq: @jamesmorrison Thank you for this! I've made adjustments accordingly

jamesmorrison (approved):

ankitguptaindia:
QA Status: Approved ✅ 
----



**Test Result:** 

Verified the notifications REST API, including listing, dismissing, invalid ID handling, and permission checks. Confirmed that notifications are returned in the expected response format, dismissed notifications are removed from subsequent responses, invalid notification IDs return a `404` response, and unauthenticated requests are blocked with a `401/403` response. 


**List notifications** ✅ 

Screenshot 2026-06-30 at 6 02 38 PM

----

**Dismiss one (use an id from step 1)** ✅ 

https://github.com/user-attachments/assets/ff5108b9-f758-414d-9018-2f4362f0625d

---

**Invalid id → expect 404:** ✅ 

Screenshot 2026-06-30 at 6 40 55 PM

Screenshot 2026-06-30 at 6 42 54 PM

-----

PR #3480, opened 2026-06-04
fix: Hide flat shipping rate option when onboarding for multi lingual store

mikaeltomovic:

Closes [GOOWOO-703](https://linear.app/a8c/issue/GOOWOO-703/show-flat-shipping-rate-option-when-it-is-the-currently-saved#comment-5e104795).

During onboarding, the "flat rate" shipping option was incorrectly appearing for multilingual stores. The root cause was a combination of two behaviours:

1. `saved-setup-stepper.js` auto-saves `shipping_rate: 'flat'` as the default whenever it is `null` (a fallback from the original implementation).
2. The `initSettings` gate (`settings?.shipping_rate ? settings : null`) means `ShippingRateMethodSection` never renders until `shipping_rate` has a value — so the component always saw `'flat'` on its first render, making `isFlatShippingRate` always `true` for all stores, including multilingual ones.

**Fix 1 — default shipping rate (`saved-setup-stepper.js`):**
Changed the auto-save default from `'flat'` to `'manual'` when `glaData.isMultiLingualStore` is `true`. This ensures fresh onboarding for multilingual stores starts with the correct default and `isFlatShippingRate` is `false` on first render.

**Fix 2 — visibility snapshot guard (`shipping-rate-method-section.js`):**
The one-time `showFlatOption` ref snapshot is now also gated on `hasFinishedResolution` and checks `mcSetup?.status === 'incomplete'`. This covers the re-onboarding scenario where the database still holds a previous `'flat'` value (so the auto-save in Fix 1 never fires), ensuring the flat option is still suppressed for multilingual stores during onboarding even when stale data is present.

The mid-session freeze behaviour (preventing the flat option from disappearing after a user switches away from flat and triggers an auto-save) is preserved.

1. Install the plugin.
2. Complete the full onboarding flow and choose "flat rate" as the shipping method. Continue to completion.
3. Disconnect all accounts.
4. Activate the WPML plugin to setup a multilingual store (or mock a multilingual store by setting `isMultiLingualStore` to true in the code).
5. Navigate back to the onboarding wizard.
6. Proceed to the "Configure product listings" step and verify the **"My shipping settings are simple. I can manually estimate flat shipping rates."** option is **not visible**.
7. Repeat steps 3–4 several times (both with and without a full page reload between disconnect and re-onboarding) to confirm the option never appears intermittently.
8. On a **non-multilingual** store, verify all three shipping rate options (automatic, flat, manual) continue to appear as expected during onboarding.
9. On a multilingual store that has previously configured flat shipping and has **completed** onboarding, navigate to the shipping settings page and confirm the flat rate option **is** visible (to allow switching away).

### Changelog entry

>

### DIFF
--- a/js/src/components/shipping-rate-section/shipping-rate-method-section.js
+++ b/js/src/components/shipping-rate-section/shipping-rate-method-section.js
@@ -33,14 +33,24 @@ const ShippingRateMethodSection = ( { children } ) => {
 	const isFlatShippingRate =
 		settings?.shipping_rate === SHIPPING_RATE_METHOD.FLAT;
 
-	// Defer the one-time snapshot until settings has resolved. Using a ref so
-	// mid-session auto-saves (which flip isFlatShippingRate to false) don't
-	// hide the option after it was already shown.
+	// Take a one-time snapshot once both settings and MC setup have resolved.
+	// Using a ref so mid-session auto-saves (which flip isFlatShippingRate to
+	// false) don't hide the option after it was already shown.
+	// During onboarding (status === 'incomplete'), ignore any previously stored
+	// flat rate: the auto-save in SavedSetupStepper always initialises
+	// shipping_rate to 'flat', so isFlatShippingRate would always be true here,
+	// which incorrectly shows the option for multilingual stores.
 	const showFlatOptionRef = useRef( null );
-	if ( showFlatOptionRef.current === null && settings !== undefined ) {
-		showFlatOptionRef.current = ! isMultiLingualStore || isFlatShippingRate;
+	if (
+		showFlatOptionRef.current === null &&
+		settings !== undefined &&
+		hasFinishedResolution
+	) {
+		const isOnboarding = mcSetup?.status === 'incomplete';
+		showFlatOptionRef.current =
+			! isMultiLingualStore || ( isFlatShippingRate && ! isOnboarding );
 	}
-	const showFlatOption = showFlatOptionRef.current;
+	const showFlatOption = showFlatOptionRef.current ?? ! isMultiLingualStore;
 
 	// Hide the automatic shipping rate option if there are no shipping rates and the merchant is onboarding.
 	const hideAutomaticShippingRate =
--- a/js/src/components/shipping-rate-section/shipping-rate-method-section.js
+++ b/js/src/components/shipping-rate-section/shipping-rate-method-section.js
@@ -33,7 +33,7 @@ const ShippingRateMethodSection = ( { children } ) => {
 	const isFlatShippingRate =
 		settings?.shipping_rate === SHIPPING_RATE_METHOD.FLAT;
 
-	// Take a one-time snapshot once both settings and MC setup have resolved.
+	// Take a one-time snapshot once settings have resolved.
 	// Using a ref so mid-session auto-saves (which flip isFlatShippingRate to
 	// false) don't hide the option after it was already shown.
 	// For multilingual stores the flat option is shown only when the stored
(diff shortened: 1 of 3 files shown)


js/src/components/shipping-rate-section/shipping-rate-method-section.js
```diff
@@ -33,14 +33,24 @@ const ShippingRateMethodSection = ( { children } ) => {
...
 	const showFlatOptionRef = useRef( null );
-	if ( showFlatOptionRef.current === null && settings !== undefined ) {
-		showFlatOptionRef.current = ! isMultiLingualStore || isFlatShippingRate;
+	if (
+		showFlatOptionRef.current === null &&
+		settings !== undefined &&
+		hasFinishedResolution
+	) {
+		const isOnboarding = mcSetup?.status === 'incomplete';
+		showFlatOptionRef.current =
```
asvinb: @mikaeltomovic I don't think we need to check whether it's onboarding or not, because during onboarding, we are setting the shipping rate method to `flat` as [per your changes here](https://github.com/woocommerce/google-listings-and-ads/pull/3480/changes#diff-414b6e1dc1099e8765f3acc898a3f7a24f7e7583f3be9cca42e7ff12ebeac4edR85).
So we can check for the following:
* Do not display the flat option if the site is a multilingual one unless the stored value is `flat`.
* If the shipping rate is `flat`, always show the flat rate even when the user chooses another option until the page is reloaded.

Let me know what you think 😁

asvinb (changes requested):

js/src/components/shipping-rate-section/shipping-rate-method-section.js
mikaeltomovic: @asvinb Yes I suppose the check that happens in `save-setup-stepper.js` should suffice. I have removed `isOnboarding` from this conditional.

js/src/components/shipping-rate-section/shipping-rate-method-section.js
```diff
@@ -33,14 +33,23 @@ const ShippingRateMethodSection = ( { children } ) => {
...
+	// value is already 'flat' (i.e. the merchant previously configured it).
+	// The SavedSetupStepper auto-save initialises multilingual stores to
+	// 'manual', never 'flat', so isFlatShippingRate is false on a clean
+	// onboarding and the option is correctly hidden.
 	const showFlatOptionRef = useRef( null );
-	if ( showFlatOptionRef.current === null && settings !== undefined ) {
+	if (
+		showFlatOptionRef.current === null &&
+		settings !== undefined &&
+		hasFinishedResolution
```
asvinb: @mikaeltomovic I don’t think we need to wait for `useMCSetup` to resolve, since it’s independent of what we’re trying to achieve here. Showing or hiding the flat rate when it’s either selected by default or for a multilingual store.

So I think we can safely remove `hasFinishedResolution` from the condition.

js/src/components/shipping-rate-section/shipping-rate-method-section.js
```diff
@@ -33,14 +33,23 @@ const ShippingRateMethodSection = ( { children } ) => {
 	const isFlatShippingRate =
```
asvinb: @mikaeltomovic Let's move this within the `if` [condition below](https://github.com/woocommerce/google-listings-and-ads/pull/3480/changes#diff-00814ec3e107a586e6c1ec5dcf17d6b00665276d79a581c849c2cbec0aa4b68fR45) where it's only used there.

js/src/pages/onboarding/setup-stepper/saved-setup-stepper.js
```diff
@@ -82,7 +82,7 @@ const SavedSetupStepper = ( { savedStep } ) => {
 		if ( settings?.shipping_rate === null ) {
 			saveSettings( {
 				...settings,
-				shipping_rate: 'flat',
+				shipping_rate: glaData.isMultiLingualStore ? 'manual' : 'flat',
```
asvinb:
```suggestion
				shipping_rate: glaData.isMultiLingualStore ? SHIPPING_RATE_METHOD.MANUAL : SHIPPING_RATE_METHOD.FLAT,
```

asvinb (commented):
@mikaeltomovic This seems to be working as expected. I left a couple of comments. Can you check them out please?

asvinb (approved):
LGTM @mikaeltomovic ! Thanks!

PR #3482, opened 2026-06-08
GOOWOO-723: Migrate product attributes and custom attribute mapping rules to the MAPI

alaca:
Extends `WCProductInputAdapter` to build the full product attribute layer for the MAPI, replacing the Content API attribute mapping that lived in `WCProductAdapter`.

Closes https://linear.app/a8c/issue/GOOWOO-723/migrate-product-attributes-and-custom-attribute-mapping-rules-to-the

### Changelog entry

>

### DIFF
--- a/src/Product/WCProductInputAdapter.php
+++ b/src/Product/WCProductInputAdapter.php
@@ -4,8 +4,11 @@ declare( strict_types=1 );
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Product;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\API\Google\Mapi\Models\ProductInput;
+use Automattic\WooCommerce\GoogleListingsAndAds\Exception\InvalidValue;
 use Automattic\WooCommerce\GoogleListingsAndAds\PluginHelper;
+use Automattic\WooCommerce\GoogleListingsAndAds\Product\AttributeMapping\AttributeMappingHelper;
 use WC_Product;
+use WC_Product_Variable;
 use WC_Product_Variation;
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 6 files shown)


src/Product/WCProductInputAdapter.php
```diff
@@ -51,27 +77,56 @@ class WCProductInputAdapter {
...
+
 		$this->map_identity();
 		$this->map_general_attributes();
 		$this->map_images();
 		$this->map_availability();
 		$this->map_price();
 		$this->map_shipping();
+		$this->map_product_types();
+
+		// Order matters: per-product values override rules, gtin overrides those, the override filter wins.
```
joemcgill, line 125: ✨

src/Product/WCProductInputAdapter.php
```diff
@@ -333,6 +432,258 @@ protected function is_virtual(): bool {
...
+	 * Resolve a mapping rule source to its value: a product field, taxonomy term,
+	 * custom attribute, or a static value.
+	 *
+	 * @param string $source
+	 *
+	 * @return string
+	 */
+	protected function get_source( string $source ): string {
+		$separator = strpos( $source, ':' );
+		if ( ! $separator ) {
```
joemcgill: Would `:` ever be in position 0 and cause this to bail incorrectly?

src/Product/WCProductInputAdapter.php
```diff
@@ -333,6 +432,258 @@ protected function is_virtual(): bool {
...
+	 */
+	protected function rule_match_conditions( array $rule ): bool {
+		$condition_type = $rule['category_condition_type'] ?? '';
+
+		if ( AttributeMappingHelper::CATEGORY_CONDITION_TYPE_ALL === $condition_type ) {
+			return true;
+		}
+
+		$categories = explode( ',', (string) ( $rule['categories'] ?? '' ) );
+		$contains   = ! empty( array_intersect( $categories, $this->product_category_ids ) );
```
joemcgill: $categories will be strings while product category IDs will be integers. This will work because of loose type checking by could cause bugs in the future. Could be worth type adjusting one of the arrays before doing the intersect to be safe.

src/Product/WCProductInputAdapter.php
```diff
@@ -333,6 +432,258 @@ protected function is_virtual(): bool {
...
+		return empty( $values ) ? '' : (string) reset( $values );
+	}
+
+	/**
+	 * Apply the attribute-values override filter, which takes precedence over every
+	 * other attribute mapping.
+	 */
+	protected function override_attributes(): void {
+		/** This filter is documented in src/Product/WCProductAdapter.php */
+		$overrides = apply_filters( 'woocommerce_gla_product_attribute_values', [], $this->wc_product, $this );
```
joemcgill, line 651: The third param being passed to this filter has changed to a new class type, so this should be documented as a change.

joemcgill (changes requested):
A few small suggestions, but looks good.

src/Product/WCProductInputAdapter.php
alaca: Good catch. The source should always have (product:title, taxonomy:…, attribute:…), so a leading colon shouldn't occur in practice, but the check was still wrong: `strpos()` returns 0 for a position 0 colon, so `! $separator` would treat it as "no separator" and return the raw source. Switched to a strict `false === $separator`

joemcgill (approved):

PR #3484, opened 2026-06-08
GOOWOO-647 styling notification panels

mserino:

Closes [GOOWOO-647 Styles](https://linear.app/a8c/issue/GOOWOO-647/notificationspanel-notification-components).

1. 
2. 
3. 

### Changelog entry

>

### DIFF
--- a/js/src/notifications-system/notification.js
+++ b/js/src/notifications-system/notification.js
@@ -3,7 +3,13 @@
  */
 import { __ } from '@wordpress/i18n';
 import { dateI18n } from '@wordpress/date';
-import { Flex, FlexBlock, FlexItem } from '@wordpress/components';
+import {
+	CardBody,
+	CardDivider,
+	Flex,
+	FlexBlock,
+	FlexItem,
+} from '@wordpress/components';
 import { closeSmall } from '@wordpress/icons';
 
 /**
@@ -52,7 +58,7 @@ const Notification = ( {
 	};
 
 	return (
-		<Flex align="flex-start" className="gla-notification" gap="4">
+		<CardBody align="flex-start" className="gla-notification" gap="4">
 			<FlexItem>
 				<img
 					src={ googleLogoURL }
@@ -77,7 +83,7 @@ const Notification = ( {
 						<span className="gla-notification__date">
 							{ formattedDate }
 						</span>
-						<div className="gla-notification__actions">
+						<FlexBlock>
 							{ actions.map(
 								( {
 									id: actionId,
@@ -98,25 +104,22 @@ const Notification = ( {
 									</AppButton>
 								)
 							) }
-						</div>
+						</FlexBlock>
 					</Flex>
 				</Flex>
 			</FlexBlock>
-			<FlexBlock
-				className="gla-notification__dismiss-container"
-				align="flex-start"
-			>
+			<FlexItem>
 				<AppButton
-					className="gla-notification__dismiss"
-					onClick={ handleDismissClick }
 					aria-label={ __(
 						'Dismiss notification',
 						'google-listings-and-ads'
 					) }
+					className="gla-notification__dismiss"
+					onClick={ handleDismissClick }
 					icon={ closeSmall }
 				/>
-			</FlexBlock>
-		</Flex>
+			</FlexItem>
+		</CardBody>
 	);
 };
 
--- a/js/src/notifications-system/notification.js
+++ b/js/src/notifications-system/notification.js
@@ -3,13 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { dateI18n } from '@wordpress/date';
-import {
-	CardBody,
-	CardDivider,
-	Flex,
-	FlexBlock,
-	FlexItem,
-} from '@wordpress/components';
+import { CardBody, Flex, FlexBlock, FlexItem } from '@wordpress/components';
 import { closeSmall } from '@wordpress/icons';
 
 /**
--- a/js/src/notifications-system/notification.js
+++ b/js/src/notifications-system/notification.js
@@ -3,7 +3,6 @@
  */
 import { __ } from '@wordpress/i18n';
 import { dateI18n } from '@wordpress/date';
-import { Card } from '@wordpress/components';
 
 /**
  * Internal dependencies
@@ -11,7 +10,7 @@ import { Card } from '@wordpress/components';
 import { glaData } from '~/constants';
 import { useAppDispatch } from '~/data';
 import AppButton from '~/components/app-button';
-import googleLogoURL from '~/images/logo/google-logo.svg';
+import googleLogoURL from '~/images/logo/google-g-logo.svg';
 import './notification.scss';
 
 /**
@@ -38,15 +37,34 @@ const Notification = ( {
 	);
 
 	return (
-		<Card className="gla-notification">
-			<div className="gla-notification__header">
+		<div className="gla-notification">
+			<div className="gla-notification__logo-container">
 				<img
 					src={ googleLogoURL }
 					alt="Google"
 					className="gla-notification__logo"
-					width="64"
-					height="22"
+					width="24"
+					height="24"
 				/>
+			</div>
+			<div className="gla-notification__body">
+				<h3 className="gla-notification__title">{ title }</h3>
+				<p className="gla-notification__description">{ description }</p>
+
+				<div className="gla-notification__actions">
+					<p className="gla-notification__date">{ formattedDate }</p>
+					{ actions.length > 0 &&
+						actions.map( ( actionProps, index ) => (
+							<AppButton
+								key={ index }
+								className="gla-notification__action"
+								variant="link"
+								{ ...actionProps }
+							/>
+						) ) }
+				</div>
+			</div>
+			<div className="gla-notification__dismiss-container">
 				<button
 					type="button"
 					className="gla-notification__dismiss"
@@ -59,19 +77,7 @@ const Notification = ( {
 					&#x2715;
 				</button>
 			</div>
-			<div className="gla-notification__body">
-				<h3 className="gla-notification__title">{ title }</h3>
-				<p className="gla-notification__date">{ formattedDate }</p>
-				<p className="gla-notification__description">{ description }</p>
-			</div>
-			{ actions.length > 0 && (
-				<div className="gla-notification__actions">
-					{ actions.map( ( actionProps, index ) => (
-						<AppButton key={ index } { ...actionProps } />
-					) ) }
-				</div>
-			) }
-		</Card>
+		</div>
 	);
 };
 
--- a/js/src/notifications-system/notification.js
+++ b/js/src/notifications-system/notification.js
@@ -3,7 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { dateI18n } from '@wordpress/date';
-import { Flex, FlexBlock, FlexItem, Icon } from '@wordpress/components';
+import { Flex, FlexBlock, FlexItem } from '@wordpress/components';
 import { closeSmall } from '@wordpress/icons';
 
 /**
@@ -17,22 +17,22 @@ import './notification.scss';
 
 /**
  * @typedef {Object} NotificationAction
- * @property {string}  id        Unique key for the action.
- * @property {string}  href      Link destination.
- * @property {string}  children  Button label.
- * @property {string}  [target]  Link target (e.g. '_blank').
- * @property {string}  [rel]     Link rel attribute.
+ * @property {string} id Unique key for the action.
+ * @property {string} href Link destination.
+ * @property {string} children Button label.
+ * @property {string} [target] Link target (e.g. '_blank').
+ * @property {string} [rel] Link rel attribute.
  */
 
 /**
  * Base notification card component.
  *
- * @param {Object}               props
- * @param {string}               props.id          Notification ID, used to call dismissNotification on dismiss.
- * @param {string}               props.title       Notification headline.
- * @param {string}               props.description Notification body text.
- * @param {number}               props.triggeredAt Unix timestamp (seconds) when the notification was triggered.
- * @param {NotificationAction[]} props.actions     CTA buttons.
+ * @param {Object} props
+ * @param {string} props.id Notification ID, used to call dismissNotification on dismiss.
+ * @param {string} props.title Notification headline.
+ * @param {string} props.description Notification body text.
+ * @param {number} props.triggeredAt Unix timestamp (seconds) when the notification was triggered.
+ * @param {NotificationAction[]} props.actions CTA buttons.
  */
 const Notification = ( {
 	id,
@@ -57,13 +57,13 @@ const Notification = ( {
 				<img
 					src={ googleLogoURL }
 					alt={ __( 'Google Logo', 'google-listings-and-ads' ) }
-					width="24"
-					height="24"
+					width="16"
+					height="16"
 				/>
 			</FlexItem>
 			<FlexBlock className="gla-notification__body">
 				<Flex direction="column" gap="1">
-					<h3 className="gla-notification__title">{ title }</h3>
+					<p className="gla-notification__title">{ title }</p>
 					<p className="gla-notification__description">
 						{ description }
 					</p>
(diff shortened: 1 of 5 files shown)


mikaeltomovic (approved):
Looks good. Great work!

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
 	);
 
 	return (
-		<Card className="gla-notification">
-			<div className="gla-notification__header">
+		<div className="gla-notification">
```
asvinb: @mserino Let's use the `Flex` component from `@wordpress/components` to achieve the flex layout, without the need for custom CSS.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
+			<div className="gla-notification__logo-container">
```
asvinb: @mserino I don't see that class name being used anywhere. Can we remove it please?

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
 				<img
 					src={ googleLogoURL }
 					alt="Google"
```
asvinb:
```suggestion
					alt=__( 'Google Logo', 'google-listings-and-ads' )
```

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
 					className="gla-notification__logo"
```
asvinb: @mserino Can we remove that class name since it's not being used anywhere.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
-					width="64"
-					height="22"
+					width="24"
+					height="24"
 				/>
+			</div>
+			<div className="gla-notification__body">
```
asvinb: @mserino If we use `FlexBlock`, we can remove the class name and associated styles.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
+				<h3 className="gla-notification__title">{ title }</h3>
+				<p className="gla-notification__description">{ description }</p>
+
+				<div className="gla-notification__actions">
```
asvinb: @mserino Let's use the `Flex` component here as well please.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
+					<p className="gla-notification__date">{ formattedDate }</p>
+					{ actions.length > 0 &&
```
asvinb: @mserino Let's remove the length check here since it's redundant. Since `actions` can be an empty array, React will render nothing.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
+						actions.map( ( actionProps, index ) => (
+							<AppButton
+								key={ index }
```
asvinb: @mserino Using the array index as `key` is an anti pattern. For each action, we can have a unique ID passed which can then be used as key.

js/src/notifications-system/notification.js
asvinb: Also let's update the class name if needed to `gla-notification__footer`. We can use the `gla-notification__actions` class name for the container containing all the actions.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
 				/>
+			</div>
+			<div className="gla-notification__body">
+				<h3 className="gla-notification__title">{ title }</h3>
+				<p className="gla-notification__description">{ description }</p>
+
+				<div className="gla-notification__actions">
+					<p className="gla-notification__date">{ formattedDate }</p>
+					{ actions.length > 0 &&
+						actions.map( ( actionProps, index ) => (
```
asvinb: @mserino Let's wrap all actions within a `gla-notification__actions` container. See my message above.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
+							<AppButton
+								key={ index }
+								className="gla-notification__action"
+								variant="link"
+								{ ...actionProps }
```
asvinb: @mserino Reading through the code, am not entirely sure which props are accepted unless i look at `AppButton`. Instead of spreading everything, can we destructure `actionProps` and use only the props we need?

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
+							/>
+						) ) }
+				</div>
+			</div>
+			<div className="gla-notification__dismiss-container">
```
asvinb: @mserino If we use `FlexItem` and set the parent flex container to align flex-start, then we don't need any styles for that container.

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
 				<button
```
asvinb: @mserino Can we use `AppButton` here please?

js/src/notifications-system/notification.js
```diff
@@ -38,15 +37,34 @@ const Notification = ( {
...
 					type="button"
 					className="gla-notification__dismiss"
```
asvinb: @mserino For the `onClick`, instead of having an anonymous function, let's have a `handleClick` function which then calls `dismissNotification`

js/src/notifications-system/notification.js
```diff
@@ -59,19 +77,7 @@ const Notification = ( {
 					&#x2715;
```
asvinb: @mserino See [my note here](https://github.com/woocommerce/google-listings-and-ads/pull/3474#discussion_r3357159442).

js/src/notifications-system/notifications-panel.js
```diff
@@ -48,7 +51,18 @@ const NotificationsPanel = () => {
...
 	return (
-		<div className="gla-notifications-panel">
+		<Card className="gla-notifications-panel">
+			<CardHeader className="gla-notifications-panel__header">
+				<Text
+					variant="title-small"
+					className="gla-notifications-panel__title"
+				>
+					{ __( 'Action required', 'google-listings-and-ads' ) }
+					<span className="gla-notifications-panel__chip">
```
asvinb: @mserino Can you check if we can use the existing `Badge` component please?

js/src/notifications-system/notification.scss
```diff
@@ -1,12 +1,18 @@
...
+	padding: $grid-unit-30;
+
+	&:last-child {
+		border-bottom: 0;
+	}
 }
 
-.gla-notification__logo {
-	display: block;
+.gla-notification__dismiss-container {
```
asvinb: @mserino Let's have all the styles nested under `.gla-notification` please.

asvinb (changes requested):
Nice work here @mserino . I left a few comments. Can you check them out please? The most important thing is using the `Flex` components from `@wordpress/components` which should reduce the amount of custom CSS written.

js/src/notifications-system/notifications-panel.scss
```diff
@@ -3,3 +3,26 @@
 	flex-direction: column;
 	gap: $grid-unit-40;
 }
+
+.gla-notifications-panel__header {
```
asvinb: @mserino Let's have all the styles nested under `.gla-notifications-panel` please.

js/src/notifications-system/notification.js
mserino: @asvinb I don't know if calling it `__footer` is accurate, because this component is actually inside the `gla-notification__body`

js/src/notifications-system/notification.js
mserino: @asvinb I ended up needing some styles in the end and I kept it

js/src/notifications-system/notification.js
mserino: @asvinb I ended up renaming it to footer

js/src/notifications-system/notification.js
mserino: I changed `__actions` into `__footer` but since there are no styles needed here I didn't think we needed to wrap them, it's just a list of `<a>`s

js/src/notifications-system/notifications-panel.js
```diff
@@ -48,7 +52,16 @@ const NotificationsPanel = () => {
...
 	return (
-		<div className="gla-notifications-panel">
+		<Card className="gla-notifications-panel">
+			<CardHeader className="gla-notifications-panel__header">
+				<Text
+					variant="title-small"
+					className="gla-notifications-panel__title"
+				>
+					{ __( 'Action required', 'google-listings-and-ads' ) }
+					<Badge>{ notifications.length }</Badge>
```
asvinb: @mserino I think we need to tweak the styles of the `Badge` here to match the designs.
Current:
Image

Expected:
Image

Or we can use the `Badge` component from the `@woocommerce/components` [package](https://woocommerce.github.io/woocommerce/?path=/docs/components-badge--docs).

js/src/notifications-system/notification.js
```diff
@@ -3,26 +3,36 @@
...
 import { glaData } from '~/constants';
 import { useAppDispatch } from '~/data';
 import AppButton from '~/components/app-button';
-import googleLogoURL from '~/images/logo/google-logo.svg';
+import googleLogoURL from '~/images/logo/google-g-logo.svg';
 import './notification.scss';
 
+/**
+ * @typedef {Object} NotificationAction
+ * @property {string}  id        Unique key for the action.
```
asvinb: @mserino Nit: to stay consistent with the rest of the codebase, let’s avoid aligning the string descriptions. Could we remove the extra whitespace please 😁

```suggestion
 * @property {string} id Unique key for the action.
```

This applies to the JSDocs as well below please.

js/src/notifications-system/notification.js
```diff
@@ -3,26 +3,36 @@
...
 /**
  * Base notification card component.
  *
- * @param {Object}   props
- * @param {string}   props.id          Notification ID, used to call dismissNotification on dismiss.
- * @param {string}   props.title       Notification headline.
- * @param {string}   props.description Notification body text.
- * @param {number}   props.triggeredAt Unix timestamp (seconds) when the notification was triggered.
- * @param {Array}    props.actions     Array of AppButton prop objects for CTA buttons.
+ * @param {Object}               props
```
asvinb: @mserino See my [previous comment](https://github.com/woocommerce/google-listings-and-ads/pull/3484/changes#r3386385440).

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
+		<Flex align="flex-start" className="gla-notification" gap="4">
+			<FlexItem>
 				<img
 					src={ googleLogoURL }
-					alt="Google"
-					className="gla-notification__logo"
-					width="64"
-					height="22"
+					alt={ __( 'Google Logo', 'google-listings-and-ads' ) }
+					width="24"
```
asvinb:
```suggestion
					width="16"
```

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
+					height="24"
```
asvinb:
```suggestion
					height="16"
```

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
 				/>
-				<button
-					type="button"
+			</FlexItem>
+			<FlexBlock className="gla-notification__body">
+				<Flex direction="column" gap="1">
+					<h3 className="gla-notification__title">{ title }</h3>
```
asvinb: @mserino Let's use the `paragraph` tag here. The titles are within a panel, not section landmarks in the page outline. Using `<h3>` implies they're part of the document's heading hierarchy, which they aren't, they're more like the lead line of a list item.

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
+						{ description }
+					</p>
+
+					<Flex
+						className="gla-notification__footer"
+						align="center"
+						justify="start"
+						wrap="wrap"
+					>
+						<p className="gla-notification__date">
```
asvinb: @mserino I feel we could use a `span` here. What do you think?

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
+							{ formattedDate }
+						</p>
+						{ actions.map(
```
asvinb: @mserino Let's wrap all the actions within a container, so that all actions are grouped together.

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
+							( {
+								id: actionId,
```
asvinb: @mserino I don't think we need to rename this variable. Maybe we can use `id` as is.

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
-			</div>
-			{ actions.length > 0 && (
-				<div className="gla-notification__actions">
-					{ actions.map( ( actionProps, index ) => (
-						<AppButton key={ index } { ...actionProps } />
-					) ) }
-				</div>
-			) }
-		</Card>
+					<Icon icon={ closeSmall } />
```
asvinb: @mserino We can pass the icon directly to the `AppButton` component via the `icon` prop:

```suggestion
icon={ closeSmall }
```

js/src/notifications-system/notification.js
```diff
@@ -37,41 +47,75 @@ const Notification = ( {
...
+							)
+						) }
+					</Flex>
+				</Flex>
+			</FlexBlock>
+			<FlexBlock
+				className="gla-notification__dismiss-container"
+				align="flex-start"
+			>
+				<AppButton
```
asvinb, line 106: @mserino Looking the designs, the button should be `24x24`. However `AppButton` has `min-width: 36px;` and `padding: 6px;` which is resulting in a `36x36` button. Since we have the title next to the icon, let's update the CSS to make the button 24 to make more room for the title.

js/src/notifications-system/notification.scss
```diff
@@ -1,54 +1,55 @@
...
+
+	& .gla-notification__body {
+		flex-grow: 1;
+	}
+
+	& .gla-notification__dismiss-container {
+		flex: 0 0 auto;
+	}
+
+	&__title {
```
asvinb: @mserino Let's avoid concatenating the selectors as such. There are still some places in the codebase where this is the case, but we are trying to move away from that since it's hard to read and to debug.

```suggestion
	.gla-notification__title {
```
Comment applies to all the SCSS file changes in this PR

js/src/notifications-system/notification.scss
```diff
@@ -1,54 +1,55 @@
...
+
+	&__action.is-link {
+		text-decoration: none;
+
+		&:hover,
+		&:focus {
+			text-decoration: none;
+		}
+	}
+
```
asvinb, line 44: Nit, remove empty line.

js/src/notifications-system/notifications-panel.js
```diff
@@ -48,7 +52,16 @@ const NotificationsPanel = () => {
 	}
 
 	return (
-		<div className="gla-notifications-panel">
+		<Card className="gla-notifications-panel">
+			<CardHeader className="gla-notifications-panel__header">
+				<Text
```
asvinb, line 57: @mserino Do you think we can use the existing class name `woocommerce-marketing-card-header-title` present on the page which will ensure consistency with other sections on the page?

asvinb (changes requested):
@mserino Almost there 😁 Another round and I think we can merge the PR. Thanks!

js/src/notifications-system/notification.js
asvinb: Actually it makes sense to just group them in a container so that when you read the markup, you can see the date clearly separate from the actions. Because right now, it's a mixture of `p` + `a` + `a` tags

js/src/notifications-system/notification.js
mserino: @asvinb there is already an `id` coming as component prop, which is the Notification id. This one is the specific action id that we have introduced, so it makes sense to rename it

js/src/notifications-system/notifications-panel.js
mserino: We can, it's going to add some styles but I still need my own custom class to add styles to align the Badge

asvinb (approved):
LGTM! Thanks @mserino

PR #3485, opened 2026-06-09
GOOWOO-717 + GOOWOO-741: Migrate product status + listing to the Merchant API

alaca:

Migrate product status + listing to the MAPI

`MerchantStatuses` reads product status + item-level issues via `MapiProductsService` (accounts.products) instead of `get_productstatuses_batch`. The old `productstatuses.custombatch` path is removed.
New paginated `MapiProductsService::list() `(accounts.products.list) replaces the Content API `Merchant::get_products()`
Data-source fix: `MapiDataSourcesService::find_existing_data_source()` now paginates, so it can't create a duplicate when a match sits on a later page.

How to test:
Connection Test -> "Send proxied request to Google Merchant Center" lists products
`composer test-unit -- --filter "MapiProductsServiceTest|MapiDataSourcesServiceTest|MerchantStatusesTest|MerchantTest"`

Closes https://linear.app/a8c/issue/GOOWOO-717/migrate-product-status-fetching-to-the-mapi and https://linear.app/a8c/issue/GOOWOO-741/migrate-product-listing-to-the-mapi

### Changelog entry

>

### DIFF
--- a/src/API/Google/Mapi/Services/MapiProductsService.php
+++ b/src/API/Google/Mapi/Services/MapiProductsService.php
@@ -91,28 +91,28 @@ class MapiProductsService implements OptionsAwareInterface {
 	}
 
 	/**
-	 * Fetch all products for the account, following pagination.
+	 * Yield all products for the account, following pagination.
+	 *
+	 * Returns a generator so callers can stream the catalog without holding every
+	 * product in memory at once.
 	 *
 	 * @param int $page_size Maximum products to request per page.
 	 *
-	 * @return Product[]
+	 * @return iterable<Product>
 	 * @throws MerchantApiException On non-2xx response.
 	 */
-	public function list( int $page_size = 250 ): array {
-		$products   = [];
+	public function list( int $page_size = 250 ): iterable {
 		$page_token = '';
 
 		do {
 			$body = $this->client->get( $this->build_list_path( $page_size, $page_token ) );
 
 			foreach ( $body['products'] ?? [] as $product ) {
-				$products[] = Product::from_array( $product );
+				yield Product::from_array( $product );
 			}
 
 			$page_token = $body['nextPageToken'] ?? '';
 		} while ( '' !== $page_token );
-
-		return $products;
 	}
 
 	/**
--- a/src/API/Google/Mapi/Services/MapiProductsService.php
+++ b/src/API/Google/Mapi/Services/MapiProductsService.php
@@ -91,6 +91,31 @@ class MapiProductsService implements OptionsAwareInterface {
 	}
 
 	/**
+	 * Fetch all products for the account, following pagination.
+	 *
+	 * @param int $page_size Maximum products to request per page.
+	 *
+	 * @return Product[]
+	 * @throws MerchantApiException On non-2xx response.
+	 */
+	public function list( int $page_size = 250 ): array {
+		$products   = [];
+		$page_token = '';
+
+		do {
+			$body = $this->client->get( $this->build_list_path( $page_size, $page_token ) );
+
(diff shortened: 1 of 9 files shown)


src/MerchantCenter/MerchantStatuses.php
```diff
@@ -390,66 +390,74 @@ protected function fetch_issues( ?string $type = null, int $per_page = 0, int $p
...
+			$wc_product_id = $status['product_id'];
+			$wc_product    = $this->product_data_lookup[ $wc_product_id ] ?? null;
 
-			// Skip products not synced by this extension.
 			if ( ! $wc_product ) {
-				do_action(
-					'woocommerce_gla_debug_message',
-					sprintf( 'Merchant Center product %s not found in this WooCommerce store.', $mc_product_id ),
-					__METHOD__ . ' in remove_invalid_statuses()',
-				);
```
joemcgill, line 416: I wonder if this should be retained for visibility whenever a product got dropped from WC but is still present in MC?

src/API/Google/Mapi/Services/MapiProductsService.php
```diff
@@ -90,6 +90,31 @@ public function get_many( array $google_product_ids, int $concurrency = 10 ): ar
...
 
+	/**
+	 * Fetch all products for the account, following pagination.
+	 *
+	 * @param int $page_size Maximum products to request per page.
+	 *
+	 * @return Product[]
+	 * @throws MerchantApiException On non-2xx response.
+	 */
+	public function list( int $page_size = 250 ): array {
```
joemcgill: This will end up loading all of the products in the MC catalog in memory, which could get really big on large stores. I wonder if we should consider a high upper limit or returning a generator so that code making use of this function can loop through the results in a more memory friendly way?

tests/Unit/API/Google/Mapi/Services/MapiProductsServiceTest.php
```diff
@@ -156,4 +156,52 @@ function ( string $path ) use ( $boom ) {
 		$this->assertCount( 1, $captured );
 		$this->assertSame( $boom, $captured[0][0] );
 	}
+
+	public function test_list_follows_pagination_and_returns_products() {
```
joemcgill, line 160: Can you add a test that confirms the list method works as expected when passing a non-default value?

joemcgill (changes requested):
I've got a couple of small suggestions/questions, but otherwise this looks good.

src/MerchantCenter/MerchantStatuses.php
alaca: Good question. Restored a `woocommerce_gla_debug_message` here so a product that exists in Merchant Center but no longer maps to a WC product still gets logged. Resolved: https://github.com/woocommerce/google-listings-and-ads/pull/3485/commits/3f583d6626917f720da1c41ddb11f5697eb2f4d2

src/API/Google/Mapi/Services/MapiProductsService.php
alaca: Absolutely. Really good catch. Resolved: https://github.com/woocommerce/google-listings-and-ads/pull/3485/commits/7e0c9daa19eaaccee856fbf53fbed417c70b8e1b

tests/Unit/API/Google/Mapi/Services/MapiProductsServiceTest.php
alaca: Resolved: https://github.com/woocommerce/google-listings-and-ads/pull/3485/commits/4675f1cb8def092e8611ffefaa6368fce5948556

joemcgill (approved):
Changes look good! Thanks.

PR #3486, opened 2026-06-09
GOOWOO-706: Fix shipping times validation error

AlejandroPerezMartin:

Closes https://linear.app/a8c/issue/GOOWOO-706/estimated-shipping-times-are-blank

Defers onchange execution to ensure the form has valid data set when validation runs

1. Go to the onboarding
2. In the second step, without changing the default shipping times, click on continue
3. There should not be any validation error on the estimate shipping times as they are prefilled with data

### Changelog entry

>


asvinb (changes requested):
@AlejandroPerezMartin The approach taken introduces a few regressions:
1. If there are no countries, the shipping times inputs are not rendered at all.
2. When the inputs are rendered, the shipping times values are not automatically saved to the database. You can compare with the `develop` branch how it's working currently.

asvinb:
Closing in favour of https://github.com/woocommerce/google-listings-and-ads/pull/3530

PR #3489, opened 2026-06-09
GOOWOO-636: GenAI error messaging

AlejandroPerezMartin:

Closes:

- https://linear.app/a8c/issue/GOOWOO-636/update-user-error-messaging-for-when-genai-cannot-generate-assets
- https://linear.app/a8c/issue/GOOWOO-632/provide-more-helpful-error-messaging-to-users-when

Shows a snackbar error whenever the GenAI api returns an error. In case the error is `FINAL_URL_UNSUPPORTED_LANGUAGE`, the error message is different (https://linear.app/a8c/issue/GOOWOO-636/update-user-error-messaging-for-when-genai-cannot-generate-assets)

### Screenshots:

Screenshot 2026-06-09 at 17 48 29

1. Create a new campaign
2. Try to use a URL which is unsupported by hardcoding it in `generateAssets` within the `useCreateGenAIAssets` hook. [Refer to for more info](https://linear.app/a8c/issue/GOOWOO-636/update-user-error-messaging-for-when-genai-cannot-generate-assets#comment-ba9b12ff).
3. When the API returns an error for text assets, a snackbar with the error message: "Google AI isn't able to generate text assets for this page." should appear.
4. 3. When the API returns an error for media assets, a snackbar with the error message: "Google AI isn't able to generate media assets for this page." should appear.
5. In case the error is caused by an unsupported language, the message should be "The language on your ad's landing page isn't supported for AI-generated assets."

### Changelog entry

> Add - Show type-specific error snackbars when GenAI asset generation fails, including a dedicated message for unsupported page languages


js/src/hooks/useCreateGenAIAssets.js
```diff
@@ -132,9 +160,31 @@ const useCreateGenAIAssets = () => {
...
+				// Both text and image requests can fail with the same error,
+				// this flag is used to prevent showing multiple snackbars.
+				let shownErrorSnackbar = false;
+
 				for ( let index = 0; index < results.length; index++ ) {
 					const { type, assetKey } = requests[ index ];
-					const data = await processGenAIResponse( results[ index ] );
+					const result = results[ index ];
+
+					if ( result.status === 'rejected' ) {
```
asvinb: @AlejandroPerezMartin We already have this logic in place in the `processGenAIResponse` function, where we're intentionally silencing all errors with a `400` status code. Is there a reason why we need a separate function that checks for `400` status codes and then displays errors accordingly?

Also, if the error code is `FINAL_URL_UNSUPPORTED_LANGUAGE`, could we reuse the copy from GOOWOO-632 instead? It would help keep the messaging consistent.

js/src/hooks/useCreateGenAIAssets.js
```diff
@@ -132,9 +160,31 @@ const useCreateGenAIAssets = () => {
...
+						if ( ! shownErrorSnackbar ) {
+							shownErrorSnackbar = true;
+
+							const message = await getGenAIErrorMessage(
+								result.reason
+							);
+
+							createNotice( 'error', message, {
+								type: 'snackbar',
```
asvinb: @AlejandroPerezMartin Why are we explicitly specifying the type here please? Across the codebase we are using the default type.

asvinb (dismissed):

js/src/hooks/useCreateGenAIAssets.js
```diff
@@ -37,28 +37,43 @@ const useCreateGenAIAssets = () => {
...
-				createNotice(
-					'error',
-					errorResponse?.statusText ||
-						__(
-							'Unable to load AI-generated assets suggestions.',
-							'google-listings-and-ads'
-						)
-				);
+				if ( shouldNotify ) {
+					createNotice( 'error', message );
```
mikaeltomovic: If `errors?.FINAL_URL_UNSUPPORTED_LANGUAGE` is undefined then `message` is still `null` when it is passed to `createNotice()`. 

Would that create an empty snackbar?

mikaeltomovic (commented):
Solid work. Just one thing I would want you to double check.

asvinb:
@AlejandroPerezMartin I have tweaked the flow a little bit where we decouple the error display from `processGenAIResponse`. It now throws the raw error response instead of calling `createNotice` directly. The loop collects errors per request, allowing partial failures (text can fail while media succeeds). A new `displayGenAIErrors` function handles error code resolution, type-specific messages, and deduplication before displaying notices.

mserino (approved):

ankitguptaindia:
### QA Result: Need Feedback ⚠️ 

@AlejandroPerezMartin

I tested this PR by steps suggested by Claude and I am seeing two error messages. Is this expected? 


```
					// TEMP: force FINAL_URL_UNSUPPORTED_LANGUAGE for manual testing
					return Promise.reject( {
						status: 400,
						json: async () => ( {
							errors: { FINAL_URL_UNSUPPORTED_LANGUAGE: 'Language not supported' },
						} ),
					} );

```

Screenshot 2026-08-03 at 7 31 25 PM

joemcgill (commented):
The updates look like they should meet the criteria here, but there are a few issues flagged in AI review that I'd like to defer to @asvinb to review and confirm before this lands.

Thanks @AlejandroPerezMartin!


## 🤖 AI Code Review

**Issue:** [GOOWOO-636 — Update user error messaging for when GenAI cannot generate assets](https://linear.app/a8c/issue/GOOWOO-636/update-user-error-messaging-for-when-genai-cannot-generate-assets) / [GOOWOO-632 — Provide more helpful error messaging when AssetGenerationService fails](https://linear.app/a8c/issue/GOOWOO-632/provide-more-helpful-error-messaging-to-users-when)
**Branch:** `fix/GOOWOO-636-update-genai-error-msg` → `develop`

### Summary

The implementation satisfies the core requirements of both issues: GOOWOO-632's `FINAL_URL_UNSUPPORTED_LANGUAGE` error is now surfaced as a human-readable snackbar message, and GOOWOO-636's generic fallback messages are now asset-type-specific rather than silent or generic. The overall design — collect errors across settled promises, deduplicate by a two-tier key, then display once — is sound. Two details warrant discussion before merging: a gap in automated test coverage for the new branching logic, and a deviation from the literal AC copy that should be confirmed with product/QA.

### 🟡 Suggestions

- **No tests for `displayGenAIErrors` branching** (`js/src/hooks/useCreateGenAIAssets.js`): The new `displayGenAIErrors` function has at least five distinct execution paths — 400 with a known API error code, 400 without a parseable code (silent continue), non-400 HTTP errors, errors without a `.status` property (network/runtime failures), and the two-tier dedup-key logic that governs whether text and image failures get separate notices or a shared one. None of these paths are covered by any test. There is no test file for `useCreateGenAIAssets.js` at all, and the existing `texts-editor.test.js` does not exercise the generate-button flow or the new `erroredTypes` suppression branch. This is the most complex piece of logic introduced in the PR and the most likely place for regressions to hide.

- **AC copy deviation worth confirming** (`js/src/hooks/useCreateGenAIAssets.js`, `GEN_AI_DEFAULT_ERROR_MESSAGES`): GOOWOO-636's acceptance criterion specifies the exact replacement copy as "Google AI isn't able to generate assets for this page." (no type qualifier). The PR instead implements two type-qualified variants: "...generate text assets..." and "...generate media assets...". This is a reasonable and arguably better UX choice, and the PR description's own test instructions describe both variants as intended. However, the deviation from the literal AC text should be explicitly signed off by product or QA rather than assumed — especially since QA will be verifying against the AC.

- **`GEN_AI_DEFAULT_ERROR_MESSAGES[type]` can resolve to `undefined`** (`js/src/hooks/useCreateGenAIAssets.js`, `displayGenAIErrors`): The fallback in `createNotice('error', GEN_AI_ERROR_MESSAGES[errorCode] ?? GEN_AI_DEFAULT_ERROR_MESSAGES[type])` silently produces `undefined` as the notice message if `type` is any value not in the `GEN_AI_DEFAULT_ERROR_MESSAGES` map. In practice `type` is always one of the two `GEN_AI_ASSET_TYPES` constants, but a final `?? __("Google AI isn't able to generate assets for this page.", 'google-listings-and-ads')` would make the function robust against unexpected inputs without any cost.

### 🔵 Nitpicks

- (`js/src/hooks/useCreateGenAIAssets.js`, `processGenAIResponse`): The `try { return await response.clone().json(); } catch (parseError) { throw parseError; }` block is functionally identical to `return await response.clone().json();` — the catch only rethrows unconditionally. The try/catch can be removed without any behavioral change.

- (`js/src/hooks/useCreateGenAIAssets.js`, `displayGenAIErrors`): The label `PARSE_FAILED` assigned in the `else` branch actually covers any error that lacks a `.status` property — network failures, runtime TypeErrors, etc. — not just JSON parse failures. Since this string is only used as an internal dedup key and never shown to users, it is purely a readability/maintenance concern; something like `UNKNOWN_ERROR` would be less misleading to a future reader.

- (`js/src/components/paid-ads/asset-group/asset-group-editor/texts-editor.js`, `handleGenerateClick`): The `catch (error)` block showing "Something went wrong while generating texts. Please try again." is unreachable for API-level failures because `generateAssets` swallows those internally and never rethrows to the caller. This is pre-existing and not a regression from this PR, but since the PR touches `handleGenerateClick` it would be a clean opportunity to either remove the dead catch or add a comment explaining it only guards against true runtime panics.

---
*This review was generated by Claude. A human review is still required before merging.*

AlejandroPerezMartin:
> ### QA Result: Need Feedback ⚠️
> @AlejandroPerezMartin
> 
> I tested this PR by steps suggested by Claude and I am seeing two error messages. Is this expected?
> 
> ```
> 					// TEMP: force FINAL_URL_UNSUPPORTED_LANGUAGE for manual testing
> 					return Promise.reject( {
> 						status: 400,
> 						json: async () => ( {
> 							errors: { FINAL_URL_UNSUPPORTED_LANGUAGE: 'Language not supported' },
> 						} ),
> 					} );
> ```
> 
> Screenshot 2026-08-03 at 7 31 25 PM

addressed in https://github.com/woocommerce/google-listings-and-ads/pull/3489/changes/f107d26853c6cbb45013a5f1efc8b486f0645cd2

asvinb:
@AlejandroPerezMartin I've addressed all of Joe's feedback. Can you review the [commit](https://github.com/woocommerce/google-listings-and-ads/pull/3489/commits/f4d66c74594f80c0de0c898e612aa8e2d2dede0b) please and if everything looks good, we can send to QA.
I confirm the copy changes. Originally the text was "Google AI isn't able to generate assets for this page." but it's clearer to separate between text and media assets because in some cases, text assets might not get generated and vice versa.
The test instruction is incorrect though.
> In case the error is caused by an unsupported language, the message should be "Google AI isn't able to generate assets for this page."

I've updated it to:
> In case the error is caused by an unsupported language, the message should be "The language on your ad's landing page isn't supported for AI-generated assets."

AlejandroPerezMartin-sage (approved):

ankitguptaindia:
### QA Status: Approved ✅ 



### Test Coverage

| # | Scenario                                                  | Test Method                                                                                                                                | Expected Result                                                                   | Result |
| - | --------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------- | ------ |
| 1 | Create a new campaign and reach the assets step           | Normal UI flow                                                                                                                             | Campaign creation is unaffected by this change                                    | ✅ Pass |
| 2 | Generate text/media assets using a valid landing page URL | Normal UI flow                                                                                                                             | AI text/image suggestions generate successfully                                   | ✅ Pass |
| 3 | Text asset generation fails with a generic error          | Temporarily forced the `generate-text` request to reject; source change reverted after testing                                             | `Google AI isn't able to generate text assets for this page.`                     | ✅ Pass |
| 4 | Media asset generation fails with a generic error         | Temporarily forced the `generate-images` request to reject; source change reverted after testing                                           | `Google AI isn't able to generate media assets for this page.`                    | ✅ Pass |
| 5 | Landing page language is unsupported by GenAI             | Temporarily forced `generate-text`/`generate-images` to reject with `FINAL_URL_UNSUPPORTED_LANGUAGE`; source change reverted after testing | `The language on your ad's landing page isn't supported for AI-generated assets.` | ✅ Pass |

### Method Note

Cases 3–5 could not be reliably reproduced using real unsupported URLs or network-level request blocking. Requestly successfully blocked the text endpoint but was unreliable for the images endpoint.

Therefore, these scenarios were verified by temporarily modifying the `requests.map()` block in `useCreateGenAIAssets.js` to return rejected promises with the expected error response shape (`status` + `errors`). The project was rebuilt using `npm run start`, the relevant flows were triggered, and all temporary source changes were reverted after testing.

### Overall Result

**✅ PASS — All 5 scenarios passed.**

The generic text, media, and unsupported-language error paths display the expected, distinct snackbar messages, while the normal campaign creation and asset generation flows remain unaffected. No source changes were left in place after testing.







### "The language on your ad's landing page isn't supported for AI-generated assets."



Screenshot 2026-08-24 at 8 08 45 PM


### "Google AI isn't able to generate text assets for this page." 

### "Google AI isn't able to generate media assets for this page."

Screenshot 2026-08-24 at 8 10 27 PM

PR #3490, opened 2026-06-10
GOOWOO-642 Low/no Cost Evaluators

zamanq:
#### Summary of the changes
Adds six cheap-to-evaluate notification conditions for the in-app notification framework. Each evaluator reads options or cached data only — no API calls on admin page load.

Also extends `NotificationEvaluatorInterface` with `get_snooze_duration()`, records the WooCommerce install timestamp once via `woocommerce_installed`, and registers all evaluators in the DI container.

#### Evaluators:

`product-issues`: Fires when MC statuses cache has disapproved products. Snooze duration Permanent

`skipped-campaign-creation`: Fires when MC setup complete, Ads setup incomplete. Snooze duration Permanent

`abandoned-onboarding`: Fires when MC setup incomplete, onboarding started. Snooze duration 30 days

`not-onboarded-90-days`: Fires when Onboarding incomplete, WC installed > 90 days. Snooze duration 30 days

`enhanced-conversions-off`: Fires when Ads setup complete, enhanced conversions disabled. Snooze duration 7 days

`tracking-off`: Fires when WooCommerce usage tracking disabled. Snooze duration 7 days

#### Other changes

- `OptionsInterface::WC_INSTALL_TIMESTAMP` + `WcInstallTimestamp` service (set once on `woocommerce_installed`)

- `NotificationPriorities` and `NotificationSnoozeDurations` constants

Closes https://linear.app/a8c/issue/GOOWOO-642/lowno-cost-evaluators-6

### Screenshots:
N/A

**Call the REST endpoint**:

`GET /wp-json/wc/gla/notifications`
Expected response: `{ "notifications": [ { "id": "...", "triggered_at": 1234567890 } ] }`

**Dismiss a notification**:

`DELETE /wp-json/wc/gla/notifications/{id}`
Reset dismissed state (if needed): delete user meta `gla_notifications_state` for the current user.

**Per evaluator scenarios**:

`product-issues`: MC setup complete; at least one disapproved product in MC statuses cache.
Expected id in response: `product-issues`

`skipped-campaign-creation`: MC setup complete; Ads setup not complete
Expected id in response: `skipped-campaign-creation`

`abandoned-onboarding`: MC setup not complete; Google connected or any MC account step done
Expected id in response: `abandoned-onboarding`

`not-onboarded-90-days`: Onboarding not complete; `gla_wc_install_timestamp` > 90 days ago
Expected id in response: `not-onboarded-90-days`

`enhanced-conversions-off`: Ads setup complete; enhanced conversions disabled in settings
Expected id in response: `enhanced-conversions-off`

`tracking-off`: WooCommerce → Settings → Advanced → Enable usage tracking unchecked
Expected id in response: `tracking-off`

**Quick checks**
- Priority order — trigger multiple evaluators; confirm IDs are returned in priority order (product-issues first, tracking-off last).

- Dismissal — DELETE a notification; confirm it no longer appears on subsequent GET requests.

- No API call for product-issues — with MC statuses cache populated, load admin and confirm no new MC status API request is made (Network tab).

- WC install timestamp — on a fresh WC install, confirm gla_wc_install_timestamp option is set once and not overwritten on later loads.

**Notes**
Notification UI is not wired yet; testing is via REST API and admin badge plumbing only.

### Additional details:

This PR is based off of GOOWOO-641 so that it contains the implementations and relevant functionalities for notifications.


src/Internal/DependencyManagement/CoreServiceProvider.php
```diff
@@ -134,71 +141,79 @@ class CoreServiceProvider extends AbstractServiceProvider {
...
+		ProductRepository::class               => true,
+		ViewFactory::class                     => true,
+		DebugLogger::class                     => true,
+		MerchantStatuses::class                => true,
+		PriceBenchmarks::class                 => true,
+		PhoneVerification::class               => true,
+		PolicyComplianceCheck::class           => true,
+		ContactInformation::class              => true,
+		MerchantCenterService::class           => true,
+		NotificationsService::class            => true,
```
jamesmorrison: @zamanq I don't think we need this; there's no matching `use` statement so this silently becomes `Automattic\WooCommerce\GoogleListingsAndAds\Internal\DependencyManagement\NotificationsService` (class does not exist).

src/Internal/DependencyManagement/CoreServiceProvider.php
```diff
@@ -134,71 +141,79 @@ class CoreServiceProvider extends AbstractServiceProvider {
...
+		ZoneLocationsParser::class             => true,
+		ZoneMethodsParser::class               => true,
+		LocationRatesProcessor::class          => true,
+		ShippingZone::class                    => true,
+		AdsRecommendationsService::class       => true,
+		AdsAccountService::class               => true,
+		MerchantAccountService::class          => true,
+		MarketingChannelRegistrar::class       => true,
+		OAuthService::class                    => true,
+		SyncStatus::class                      => true,
```
jamesmorrison: @zamanq I don't think we need `SyncStatus` either (same problem as `NotificationService` noted above)

src/Internal/DependencyManagement/CoreServiceProvider.php
```diff
@@ -134,71 +141,79 @@ class CoreServiceProvider extends AbstractServiceProvider {
...
-		ShippingZone::class              => true,
-		AdsRecommendationsService::class => true,
-		AdsAccountService::class         => true,
-		MerchantAccountService::class    => true,
-		MarketingChannelRegistrar::class => true,
-		OAuthService::class              => true,
-		WPCLIMigrationGTIN::class        => true,
-		OnboardingCompleted::class       => true,
-		ServiceBasedMerchantState::class => true,
-		ServiceBasedMerchantHooks::class => true,
```
jamesmorrison, line 201: @zamanq The `use` statement for `ServiceBasedMerchantHooks` can be removed too.

src/Options/OptionsInterface.php
```diff
@@ -55,6 +55,7 @@ interface OptionsInterface {
 	public const YOUTUBE_EXPORT_FILES                      = 'youtube_export_files';
 	public const YOUTUBE_THIRD_PARTY_LINK                  = 'youtube_third_party_link';
 	public const ONBOARDING_COMPLETED_AT                   = 'onboarding_completed_at';
+	public const WC_INSTALL_TIMESTAMP                      = 'gla_wc_install_timestamp';
```
jamesmorrison: @zamanq This should not have a `gla_` prefix; `Options::prefix_name()` automatically prepends `gla_` to every key, so this gets stored as `gla_gla_wc_install_timestamp`, please could you amend?

src/Options/WcInstallTimestamp.php
```diff
@@ -0,0 +1,40 @@
...
+				$this->record_install_timestamp();
+			}
+		);
+	}
+
+	/**
+	 * Store the WooCommerce install timestamp once.
+	 */
+	protected function record_install_timestamp(): void {
+		$this->options->add( OptionsInterface::WC_INSTALL_TIMESTAMP, time() );
```
jamesmorrison: @zamanq We might need to be extra careful in testing here; the Google for WooCommerce plugin requires WooCommerce to activate so this option can only be added after both WooCommerce and then Google for WooCommerce are activated. The timestamp will be when Google for WooCommerce is active, not WooCommerce.

src/Internal/DependencyManagement/CoreServiceProvider.php
zamanq: @jamesmorrison This class is being called on line 330, `$this->conditionally_share_with_tags( ServiceBasedMerchantHooks::class, ServiceBasedMerchantState::class );`
So we need to keep the use statement for this.

asvinb: @zamanq I see we have removed `ServiceBasedMerchantHooks` from `$provides`, however I still see it in `register()`. The container calls `provides()` before deciding whether to invoke `register()` — if the class is still registered inside `register()` but missing here, the container won't recognise this provider as its owner. Any code that resolves `ServiceBasedMerchantHooks` directly (before another class from this provider is resolved) will silently fail.

Can you confirm if it's intentional and won't have any regressions please?

src/Notification/Evaluators/ProductIssuesEvaluator.php
```diff
@@ -0,0 +1,74 @@
...
+	 * @return bool
+	 */
+	public function should_show(): bool {
+		$mc_statuses = $this->transients->get( TransientsInterface::MC_STATUSES );
+
+		if ( ! is_array( $mc_statuses ) || empty( $mc_statuses['statistics'] ) ) {
+			return false;
+		}
+
+		foreach ( $mc_statuses['statistics'] as $status => $count ) {
```
asvinb: @zamanq Instead of looping through all the entries in the array to check whether the status is `MCStatus::DISAPPROVED`. How about we try to access it directly:

```php
return ( $mc_statuses['statistics'][ MCStatus::DISAPPROVED ] ?? 0 ) > 0;
```

src/Options/WcInstallTimestamp.php
```diff
@@ -0,0 +1,72 @@
...
+use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Registerable;
+use Automattic\WooCommerce\GoogleListingsAndAds\Infrastructure\Service;
+use Automattic\WooCommerce\GoogleListingsAndAds\Proxies\WP;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class WcInstallTimestamp
+ *
+ * Copies the WooCommerce install timestamp into a GLA option once GLA is active.
```
asvinb, line 15: @zamanq Why we need to copy the timestamp from WooCommerce and copy it to Google For Woo please? Can't we use the timestamp directly?

src/Options/WcInstallTimestamp.php
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	public function __construct( WP $wp ) {
+		$this->wp = $wp;
+	}
+
+	/**
+	 * Register a service.
+	 */
+	public function register(): void {
+		add_action(
```
asvinb: @zamanq Anonymous closures as hook callbacks can't be unhooked via `remove_action()` and make testing harder. Let's use a named public method instead, which also matches the codebase convention of a public domain-name callback delegating to a protected implementation.
```php
public function register(): void {
    add_action( 'admin_init', [ $this, 'record_wc_install_timestamp' ] );
}

public function record_wc_install_timestamp(): void {
    $this->maybe_record_wc_install_timestamp();
}

```

src/Notification/Evaluators/AbandonedOnboardingEvaluator.php
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators
+ */
+class AbandonedOnboardingEvaluator implements NotificationEvaluatorInterface, MerchantCenterAwareInterface, OptionsAwareInterface, Service {
+
+	use MerchantCenterAwareTrait;
+	use OptionsAwareTrait;
+
+	/** @var MerchantAccountState */
+	protected $merchant_account_state;
```
asvinb: @zamanq `$merchant_account_state` is only used internally, no subclass exists that needs access to it. Can we make this private?

src/Notification/Evaluators/AbandonedOnboardingEvaluator.php
```diff
@@ -0,0 +1,110 @@
...
+	public function get_snooze_duration(): ?int {
+		return NotificationSnoozeDurations::ABANDONED_ONBOARDING;
+	}
+
+	/**
+	 * Determine whether at least one onboarding step has been started.
+	 *
+	 * @return bool
+	 */
+	protected function has_onboarding_step_started(): bool {
```
asvinb: @zamanq See https://github.com/woocommerce/google-listings-and-ads/pull/3490/changes#r3452524862

src/Notification/Evaluators/NotOnboarded90DaysEvaluator.php
```diff
@@ -0,0 +1,85 @@
...
+ * Fires when onboarding is not completed and WooCommerce was installed more than 90 days ago.
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators
+ */
+class NotOnboarded90DaysEvaluator implements NotificationEvaluatorInterface, OptionsAwareInterface, Service {
+
+	use OptionsAwareTrait;
+
+	/** @var OnboardingCompleted */
+	protected $onboarding_completed;
```
asvinb: @zamanq See https://github.com/woocommerce/google-listings-and-ads/pull/3490/changes#r3452524862

src/Options/WcInstallTimestamp.php
```diff
@@ -0,0 +1,72 @@
...
+			function () {
+				$this->maybe_record_wc_install_timestamp();
+			}
+		);
+	}
+
+	/**
+	 * Store the WooCommerce install timestamp once, if not already recorded.
+	 */
+	protected function maybe_record_wc_install_timestamp(): void {
```
asvinb: @zamanq See [my note here](https://github.com/woocommerce/google-listings-and-ads/pull/3490/changes#r3452524862).

tests/Unit/Notification/Evaluators/EnhancedConversionsOffEvaluatorTest.php
```diff
@@ -0,0 +1,81 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class EnhancedConversionsOffEvaluatorTest
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Tests\Unit\Notification\Evaluators
+ */
+class EnhancedConversionsOffEvaluatorTest extends UnitTest {
```
asvinb, line 21: @zamanq Let's add 2 new cases please when:
* `WP_TOS_ACCEPTED` returns `true`
* `get_merchant_id()` returns > `0`

tests/Unit/Options/WcInstallTimestampTest.php
```diff
@@ -0,0 +1,133 @@
...
+	/**
+	 * Invoke the protected maybe_record_wc_install_timestamp method.
+	 */
+	protected function invoke_maybe_record_wc_install_timestamp(): void {
+		$reflection = new ReflectionClass( $this->service );
+		$method     = $reflection->getMethod( 'maybe_record_wc_install_timestamp' );
+		$method->setAccessible( true );
+		$method->invoke( $this->service );
+	}
+}
```
asvinb, line 139: @zamanq Let's add a new case when the timestamp is zero.

src/Notification/Evaluators/AbandonedOnboardingEvaluator.php
```diff
@@ -0,0 +1,110 @@
...
+		if ( boolval( $this->options->get( OptionsInterface::GOOGLE_CONNECTED, false ) ) ) {
```
asvinb: @zamanq The `if()` already coerces any value to boolean internally. So `( boolval( $x ) )` and `if ( $x )` are always identical in behavior. We can thus remove the casting. Same applies for the line below.

asvinb (changes requested):
@zamanq I left a few comments. Can you kindly check them out please?

src/Notification/Evaluators/ProductIssuesEvaluator.php
```diff
@@ -0,0 +1,74 @@
...
+		return 'product-issues';
+	}
+
+	/**
+	 * Whether the notification's condition is currently met.
+	 *
+	 * @return bool
+	 */
+	public function should_show(): bool {
+		$mc_statuses = $this->transients->get( TransientsInterface::MC_STATUSES );
```
asvinb, line 49: @zamanq Shall we check if there's a connected MC account before proceeding further? Because for service based merchants, they don't require a MC account.

asvinb (changes requested):

src/Notification/Evaluators/ProductIssuesEvaluator.php
zamanq: Adjusted.

src/Options/WcInstallTimestamp.php
zamanq: @asvinb The copy pattern is still reasonable because:

Backfill in one place — WcInstallTimestamp owns “how we learn WC’s install date,” including for stores that already had WC before GLA existed.
GLA architecture — evaluators use OptionsAwareTrait / OptionsInterface; they don’t reach into arbitrary wp_options keys. WC’s option name stays encapsulated in one service.
Write-once semantics — add() records the value the first time GLA sees it and never overwrites it, so notification logic doesn’t depend on WC’s option staying present or unchanged later.
Validated option — WC_INSTALL_TIMESTAMP lives in VALID_OPTIONS with the rest of GLA config.

src/Options/WcInstallTimestamp.php
zamanq: Replaced with named public methods.

src/Notification/Evaluators/AbandonedOnboardingEvaluator.php
zamanq: Makes sense. Done

tests/Unit/Notification/Evaluators/EnhancedConversionsOffEvaluatorTest.php
zamanq: Added.

tests/Unit/Options/WcInstallTimestampTest.php
zamanq: Added.

src/Notification/Evaluators/AbandonedOnboardingEvaluator.php
zamanq: @asvinb Adjusted. Thanks!

src/Notification/Evaluators/ProductIssuesEvaluator.php
zamanq: @asvinb Added the is_connected() guard to ProductIssuesEvaluator and updated the tests.

Evaluator — uses MerchantCenterAwareTrait and returns early when MC is not connected

zamanq:
> I see we have removed ServiceBasedMerchantHooks from $provides, however I still see it in register(). The container calls provides() before deciding whether to invoke register() — if the class is still registered inside register() but missing here, the container won't recognise this provider as its owner. Any code that resolves ServiceBasedMerchantHooks directly (before another class from this provider is resolved) will silently fail.
> 
> Can you confirm if it's intentional and won't have any regressions please?

This was an oversight during merge conflict resolution. Added back the hook in $provides.

src/Notification/Evaluators/TrackingOffEvaluator.php
```diff
@@ -0,0 +1,61 @@
...
+		return 'tracking-off';
+	}
+
+	/**
+	 * Whether the notification's condition is currently met.
+	 *
+	 * @return bool
+	 */
+	public function should_show(): bool {
+		return 'yes' !== $this->wp->get_option( 'woocommerce_allow_tracking', 'no' );
```
asvinb: @zamanq Similar to how you have a private property [defined here](https://github.com/woocommerce/google-listings-and-ads/pull/3490/changes#diff-7644cc09b1a53281e5fde11641b473e67625869ae99a5b8358e9acc9a4022955R35), can we have a private property for `woocommerce_allow_tracking`?

src/Options/WcInstallTimestamp.php
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * WooCommerce option that stores the store install timestamp.
+	 */
+	private const WC_ADMIN_INSTALL_TIMESTAMP_OPTION = 'woocommerce_admin_install_timestamp';
+
+	/**
+	 * @var WP
+	 */
+	protected $wp;
```
asvinb: @zamanq I don't see any subclass, so we can declare that property as `private`

asvinb (approved):
@zamanq I've left a couple of minor comments. Pre approving the PR. Once you address the changes, you can merge the PR. Thanks!

ankitguptaindia:
QA Status: Approved ✅ 
----


### Test Result:

Tested via GET /wp-json/wc/gla/notifications as an admin user on the feature/notifications-system branch.

- Notifications fire on respective events: Verified all six evaluators trigger under their conditions — product-issues, skipped-campaign-creation, abandoned-onboarding, not-onboarded-90-days, enhanced-conversions-off, and tracking-off.
- Priority order: Verified notifications are returned in the correct priority order (product-issues first → tracking-off last).
- Dismissal: Verified 
- No API call for product-issues: Verified the signal is driven solely by the cached gla_mc_statuses transient — it appears/disappears as the cache is populated/cleared, with no Merchant Center status API request made on admin load.
- WC install timestamp set once: Verified gla_wc_install_timestamp is set once from woocommerce_admin_install_timestamp and is not overwritten on subsequent admin loads.


Screenshot 2026-06-26 at 3 05 29 PM

Screenshot 2026-06-26 at 3 14 07 PM

Screenshot 2026-06-26 at 3 19 39 PM

Screenshot 2026-06-26 at 3 23 36 PM

PR #3493, opened 2026-06-10
Sync shipping rates and times to primary market when a secondary mark…

jamesmorrison:

Closes https://linear.app/a8c/issue/GOOWOO-679/sync-shipping-rates-and-times-to-primary-market-when-a-secondary

### Screenshots:

N/A

1. Go to Marketing > Google for WooCommerce, then click the Markets tab.
2. Click Add market. Pick a country (example: France). Set Estimated shipping rates and Estimated shipping times to values clearly different from the primary (example: 20, and 3 to 7 days), save the settings.

_Note: if the market list is empty, check the WooCommerce settings; if this is restricted to one country, you can't set a secondary market_

3. In the console, paste and run:
```
wp.apiFetch({ path: '/wc/gla/mc/shipping/rates' }).then(console.table);
wp.apiFetch({ path: '/wc/gla/mc/shipping/times' }).then(console.table);
```
4. Confirm the printed tables show a row for FR with the secondary values.
5. On the Markets table, open the row actions for France and click Delete, then click Delete in the Delete market confirmation modal.
6. Re-run the same console snippet from step 3. The FR row should now show the primary main country's values, not the secondary's.
7. On the Markets table, click Edit on the Primary Market row, then submit the modal without changing anything. It should save cleanly, and re-running the snippet from step 3 should still show the primary's values for FR.

**Pass criteria**
- After step 5 the FR rate row matches the primary main country's rate and currency, and the FR time row matches the primary's time and max_time. The secondary's 20 and 3 to 7 are gone from both tables.
- Step 6 saves without error and does not regress other primary countries' rows.

### Additional details:

> Fix - Sync shipping rates and times to primary market when a secondary market is deleted


src/MerchantCenter/MarketService.php
```diff
@@ -296,8 +296,12 @@ public function delete_market( string $id ): void {
 		$this->options->update( OptionsInterface::MARKETS, $markets );
 
 		if ( $country ) {
+			$this->adopt_primary_rate_for_country( $country );
+			$this->adopt_primary_time_for_country( $country );
 			$this->restore_country_to_target_audience( $country );
 		}
+
```
asvinb, line 420: @jamesmorrison Let's add a docblock for that action please. Without it, hook consumers have no documented contract for parameter types or nullability. 
Also `$country` can be null here because the hook fires outside the `if ( $country )` guard. That must be captured in the docblock.

```suggestion
/**
 * Fires after a secondary market is deleted and its shipping data is synced back to the primary.
 *
 * @param string      $id      Market ID that was deleted.
 * @param string|null $country ISO 3166-1 alpha-2 country code of the deleted market, or null if the market had no country.
 */
```

asvinb (changes requested):
@jamesmorrison Can you fix the PHP linting errors please and also check my comment?

src/MerchantCenter/MarketService.php
jamesmorrison: @asvinb I've added the docblock

asvinb (approved):
LGTM! Thanks @jamesmorrison

PR #3495, opened 2026-06-10
GOOWOO-738 add plugin agnostic notification system slot

mserino:

Closes [GOOWOO-738](https://linear.app/a8c/issue/GOOWOO-738/create-plugin-agnostic-woocommerce-marketing-notification-slot-bundle) .

### Description of the changes

Introduces a plugin-agnostic WooCommerce Marketing notification slot bundle. This is infrastructure-only — no visible UI change. The slot renders nothing until another bundle dispatches notifications to the store.
The main changes are:

A new standalone JS bundle that registers a shared `@wordpress/data` store and mounts a notification slot component on the WooCommerce Marketing overview page. The store and DOM container are guarded against double-initialisation, so multiple plugins can safely enqueue this bundle without conflicts.
A new PHP service that registers and enqueues the bundle, scoped only to the WooCommerce Marketing overview page.

1. Mock the `wp-json/wc/gla/notifications` endpoint with the following sample response

```
{
  "notifications": [
    { "id": "paused-campaign", "triggered_at": 1750118400 },
    { "id": "product-issues", "triggered_at": 1750118400 }
  ]
}
```
2. Navigate to the Marketing Overview page (`wp-admin/admin.php?page=wc-admin&path=/marketing`)
3. The notifications should show up on the page, ordered by triggered_at DESC.

### Changelog entry

>


js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
+/* global MutationObserver */
```
asvinb: @mserino See my [related comment](https://github.com/woocommerce/google-listings-and-ads/pull/3474#discussion_r3394600911).

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+ * External dependencies
+ */
+import { registerStore, select, useSelect } from '@wordpress/data';
+import { createRoot } from '@wordpress/element';
+
+const STORE_NAME = 'woocommerce/marketing-notifications-system';
+
+// First plugin to run registers the store; others skip this block.
+if ( ! select( STORE_NAME ) ) {
+	registerStore( STORE_NAME, {
```
asvinb: @mserino `registerStore` is deprecated. Can we use `register` instead? I know the core data store is initialized via `registerStore` but for new stores, let's use `register`. 👍

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+		reducer( state = [], action ) {
+			if ( action.type === 'REGISTER_NOTIFICATION' ) {
```
asvinb: @mserino Let's have `REGISTER_NOTIFICATION` as a constant and use it on L23 as well instead of the string.

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+				return [ ...state, action.notification ];
+			}
+			return state;
+		},
+
+		actions: {
+			registerNotification: ( notification ) => ( {
```
asvinb: @mserino Let's use explicit return statements here to be consistent with the rest of the codebase. I also find that they improve readability quite a bit. This applies to the `getNotifications` selector as well.

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+		selectors: {
+			getNotifications: ( state ) =>
+				[ ...state ].sort( ( a, b ) => b.triggeredAt - a.triggeredAt ),
+		},
+	} );
+}
+
+function NotificationSystemSlot() {
+	const notifications = useSelect( ( sel ) =>
```
asvinb: @mserino Let's use the `useNotifications` hook here please.

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+		sel( STORE_NAME ).getNotifications()
+	);
+
+	if ( ! notifications.length ) {
```
asvinb: @mserino Upon component mount, while data is still being resolved, `notifications` can be `undefined` which will result in an error. When using the `useNotifications` hook, you have access to the `hasFinishedResolution` property which you can use to check whether the data has resolved. So if `! hasFinishedResolution`, return `null`.

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * External dependencies
+ */
+import { registerStore, select, useSelect } from '@wordpress/data';
+import { createRoot } from '@wordpress/element';
+
+const STORE_NAME = 'woocommerce/marketing-notifications-system';
+
+// First plugin to run registers the store; others skip this block.
```
asvinb: @mserino Let's elaborate a bit more on why this is the case. For example, an engineer reading this code might not be aware that other plugins could also be using this store, so adding that context would help make the rationale behind this approach clearer.

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+	const notifications = useSelect( ( sel ) =>
+		sel( STORE_NAME ).getNotifications()
+	);
+
+	if ( ! notifications.length ) {
+		return null;
+	}
+
+	return notifications.map( ( notification, i ) => (
+		<notification.component key={ i } />
```
asvinb: @mserino while this is technically correct and working, it's not as per conventions where component name is capitalize. I suggest we do the following:

```suggestion
		const NotificationComponent = notification.component;
        return <NotificationComponent key={ i } />;
```

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+const CONTAINER_CLASS = 'woocommerce-marketing-notifications-container';
+
+let currentRoot = null;
+
+function mount( multichannel ) {
+	let container = document.querySelector( `.${ CONTAINER_CLASS }` );
+
+	if ( container && currentRoot ) {
+		return;
+	} // another plugin already mounted
```
asvinb: @mserino Let's have the comment within the `if` condition please and a bit more explicit about other plugin 😁

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,97 @@
...
+			multichannel.insertBefore( container, multichannel.firstChild );
+		}
+	}
+
+	if ( currentRoot ) {
+		currentRoot.unmount();
+	}
+
+	currentRoot = createRoot( container );
+	currentRoot.render( <NotificationSystemSlot /> );
```
asvinb, line 97: @mserino See [related comment](https://github.com/woocommerce/google-listings-and-ads/pull/3474#discussion_r3394761487).

asvinb (changes requested):
Thanks @mserino . I left some comments. Can you check them out please?

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,113 @@
+/**
```
asvinb, line 1: @mserino Please note we changed the folder name from `notification-system` to `notifications-system`. Can you make the update please?

asvinb (changes requested):
@mserino I've changed the target branch to [add/notifications-components](https://github.com/woocommerce/google-listings-and-ads/tree/add/notifications-components). Can you fix the merge conflicts please?

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Internal dependencies
+ */
+import useNotifications from './useNotifications';
+
+const STORE_NAME = 'woocommerce/marketing-notifications-system';
+const ACTION_REGISTER_NOTIFICATION = 'REGISTER_NOTIFICATION';
+
+/**
+ * This bundle may be loaded by multiple independent plugins (e.g. Google Listings
```
asvinb: @mserino Let's update the comment to avoid naming the independent plugins.

js/src/notification-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,113 @@
...
+ * & Ads, Reddit for WooCommerce). WordPress's wp_register_script ensures the JS
+ * file is only loaded once, but as a second safeguard we only register the shared
+ * data store if it hasn't been registered already — whichever plugin loads first
+ * wins, and all others use the same store instance.
+ */
+const marketingNotificationsStore = createReduxStore( STORE_NAME, {
```
asvinb: @mserino Let's move this within the `if` statement you have further below because if there's a store which is already registered, we don't need to call `createReduxStore`

asvinb (changes requested):

js/src/notification-system/woo-marketing-notifications-slot/index.js
mserino: @asvinb can you please check if this is okay now?

js/src/notifications-system/notifications-panel.js
```diff
@@ -9,7 +9,7 @@ import { Badge } from '@woocommerce/components';
 /**
  * Internal dependencies
  */
-import useNotifications from '~/hooks/useNotifications';
+import useNotifications from './woo-marketing-notifications-slot/useNotifications';
 import Notification from './notification';
 import useNotificationsSystemMap from './useNotificationsSystemMap';
 import Text from '~/components/app-text';
```
asvinb, line 15: @mserino There's an error in this component.
```js
const { notifications } = useNotifications();
```

The `useNotifications` hook does not return the `notifications` property currently. To be aligned with the codebase, let's update the hook to return the `notifications` property.

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,112 @@
...
+import { createReduxStore, register, select } from '@wordpress/data';
+import { createRoot } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import useNotifications from './useNotifications';
+
+const STORE_NAME = 'woocommerce/marketing-notifications-system';
+const ACTION_REGISTER_NOTIFICATION = 'REGISTER_NOTIFICATION';
```
asvinb:
```suggestion
const REGISTER_NOTIFICATION = 'REGISTER_NOTIFICATION';
```

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,112 @@
...
+ * This bundle may be loaded by multiple independent plugins. WordPress's wp_register_script ensures the JS
+ * file is only loaded once, but as a second safeguard we only register the shared
+ * data store if it hasn't been registered already — whichever plugin loads first
+ * wins, and all others use the same store instance.
+ */
+if ( ! select( STORE_NAME ) ) {
+	register(
+		createReduxStore( STORE_NAME, {
+			reducer( state = [], action ) {
+				if ( action.type === ACTION_REGISTER_NOTIFICATION ) {
```
asvinb:
```suggestion
				if ( action.type === REGISTER_NOTIFICATION ) {
```

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,112 @@
...
+					return [ ...state, action.notification ];
+				}
+				return state;
+			},
+
+			actions: {
+				registerNotification: ( notification ) => {
+					return { type: ACTION_REGISTER_NOTIFICATION, notification };
```
asvinb:
```suggestion
					return { type: REGISTER_NOTIFICATION, notification };
```

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,112 @@
...
+ */
+import { createReduxStore, register, select } from '@wordpress/data';
+import { createRoot } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import useNotifications from './useNotifications';
+
+const STORE_NAME = 'woocommerce/marketing-notifications-system';
```
asvinb: @mserino Let's have a `constants.js` file and then export the `STORE_NAME` from other there. 👍

asvinb (changes requested):
@mserino While reviewing https://github.com/woocommerce/google-listings-and-ads/pull/3499/changes, I noticed a few changes that should be made to this PR 👍  Can you take a look please?

js/src/notifications-system/woo-marketing-notifications-slot/constants.js
```diff
@@ -0,0 +1,2 @@
+export const STORE_NAME = 'woocommerce/marketing-notifications-system';
```
asvinb:
```suggestion
export const STORE_NAME = 'wc/marketing-notifications-system';
```

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,110 @@
...
+function NotificationSystemSlot() {
+	const { notifications } = useNotifications();
+
+	if ( ! notifications?.length ) {
+		return null;
+	}
+
+	return notifications.map( ( notification, i ) => {
+		const NotificationComponent = notification.component;
+		return <NotificationComponent key={ i } />;
```
asvinb: @mserino Let's not use the array index as map please? Instead let's use the notification ID.

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,110 @@
...
+	} );
+}
+
+const MULTICHANNEL_CLASS = 'woocommerce-marketing-overview-multichannel';
```
asvinb: @mserino Am thinking, let's move the constants here to the `constants.js` file 👍

js/src/notifications-system/woo-marketing-notifications-slot/useNotifications.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * External dependencies
+ */
+import { useSelect } from '@wordpress/data';
+
+const STORE_NAME = 'woocommerce/marketing-notifications-system';
```
asvinb: @mserino The `STORE_NAME` should be imported from the new `constants.js` file created.

src/Internal/DependencyManagement/AdminServiceProvider.php
```diff
@@ -4,6 +4,7 @@
 namespace Automattic\WooCommerce\GoogleListingsAndAds\Internal\DependencyManagement;
 
 use Automattic\WooCommerce\GoogleListingsAndAds\Admin\Admin;
+use Automattic\WooCommerce\GoogleListingsAndAds\Admin\NotificationSystemSlot;
```
asvinb:
```suggestion
use Automattic\WooCommerce\GoogleListingsAndAds\Admin\NotificationsSystemSlot;
```

src/Internal/DependencyManagement/AdminServiceProvider.php
```diff
@@ -55,25 +56,25 @@ class AdminServiceProvider extends AbstractServiceProvider implements Conditiona
...
-		Shipping::class            => true,
-		Service::class             => true,
+		Admin::class                  => true,
+		AttributeMapping::class       => true,
+		BulkEditInitializer::class    => true,
+		ConnectionTest::class         => true,
+		CouponBulkEdit::class         => true,
+		Dashboard::class              => true,
+		NotificationManager::class    => true,
+		NotificationSystemSlot::class => true,
```
asvinb:
```suggestion
		NotificationsSystemSlot::class => true,
```

src/Internal/DependencyManagement/AdminServiceProvider.php
```diff
@@ -110,6 +111,7 @@ public function register(): void {
 		$this->share_with_tags( AttributeMapping::class );
 		$this->share_with_tags( Dashboard::class, OnboardingCompleted::class );
 		$this->share_with_tags( NotificationManager::class, AssetsHandlerInterface::class );
+		$this->share_with_tags( NotificationSystemSlot::class, AssetsHandlerInterface::class );
```
asvinb:
```suggestion
		$this->share_with_tags( NotificationsSystemSlot::class, AssetsHandlerInterface::class );
```

src/Internal/DependencyManagement/AdminServiceProvider.php
```diff
@@ -110,6 +111,7 @@ public function register(): void {
...
 		$this->share_with_tags( NotificationSystem::class, AssetsHandlerInterface::class );
```
asvinb, line 115: @mserino There is already another `NotificationSystem:class`. Shall we remove it?

webpack.config.js
```diff
@@ -168,6 +168,11 @@ const webpackConfig = {
 			'js/src/meta-boxes/channel-visibility',
 			'index.js'
 		),
+		'notification-slot': path.resolve(
```
asvinb:
```suggestion
		'woo-marketing-notifications-slot': path.resolve(
```

js/src/notifications-system/woo-marketing-notifications-slot/useNotifications.test.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Internal dependencies
+ */
+import useNotifications from './useNotifications';
+
+jest.mock( '@wordpress/data', () => ( {
+	useSelect: jest.fn(),
+} ) );
+
+const STORE_NAME = 'woocommerce/marketing-notifications-system';
```
asvinb: @mserino Store name should be imported from `constants.js`

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,110 @@
...
+			actions: {
+				registerNotification: ( notification ) => {
+					return { type: REGISTER_NOTIFICATION, notification };
+				},
+			},
+
+			selectors: {
+				getNotifications: ( state ) => {
+					return [ ...state ].sort(
+						( a, b ) => b.triggeredAt - a.triggeredAt
```
asvinb: @mserino Can you check the property used here please? The slot store sorts by `triggeredAt` camelCase. `NotificationsPanel` reads `triggered_at` (snake_case) from the same store objects.

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Internal dependencies
+ */
+import useNotifications from './useNotifications';
+import { STORE_NAME, REGISTER_NOTIFICATION } from './constants';
+
+/**
+ * This bundle may be loaded by multiple independent plugins. WordPress's wp_register_script ensures the JS
+ * file is only loaded once, but as a second safeguard we only register the shared
```
asvinb, line 22: @mserino Let's add the `'woocommerce-marketing-notifications-system-slot'` handle to the comment so that it's easy for engineers to know 👍

js/src/notifications-system/woo-marketing-notifications-slot/index.js
```diff
@@ -0,0 +1,110 @@
...
+		return false;
+	}
+
+	if ( ! container ) {
+		container = document.createElement( 'div' );
+		container.className = CONTAINER_CLASS;
+
+		const banner = multichannel.querySelector( `.${ BANNER_CLASS }` );
+
+		if ( banner ) {
```
asvinb, line 85: @mserino If you can add a comment explaining what we're trying to do here please?

asvinb (changes requested):

src/Internal/DependencyManagement/AdminServiceProvider.php
mserino: @asvinb where is that? I can only see the one on line 115

asvinb: @mserino Do we need both? 🤔 I see 
`NotificationsSystemSlot` and `NotificationSystem`

mserino: Yes, @asvinb both are existing files and classes. `NotificationsSystemSlot` has been added with this PR. Any reason why we shouldn't add it to the `register` function?

asvinb (approved):

src/Internal/DependencyManagement/AdminServiceProvider.php
asvinb: You're right we need both. Not sure what i was thinking 😁

ankitguptaindia:
QA Status: Needs Feedback ⚠️ 
-----

@mserino @asvinb I tested this PR based on the description and found first steps always give an error 
`Uncaught TypeError: Cannot read properties of undefined (reading 'getNotifications')
    at <anonymous>:1:61`


Screenshot 2026-06-22 at 8 33 09 PM

**_On debugging with the help of AI, it gives the following error, could you please check if anything needs to be updated?_**


1. PR #3495 test instructions are out of date vs. merged code. The store key and container class were renamed during review but the PR description wasn't updated:

Store: description says woocommerce/marketing-notifications-system, but merged constants.js registers wc/marketing-notifications-system.
Container: description says .woocommerce-marketing-notifications-container, but merged code uses .gla-woo-marketing-notifications-slot-container.
So wp.data.select('woocommerce/marketing-notifications-system') returns undefined and the container query returns null. Can we either align the code to the documented woocommerce/... names or update the test steps?

2. The standalone slot bundle doesn't register the store (likely the main bug). The webpack entry woo-marketing-notifications-slot resolves to an index.js that only re-exports — it never calls init()/registerStore(). I confirmed the built woo-marketing-notifications-slot.js contains no registerStore/createReduxStore/store-key references, while notifications-system.js does.

The only thing that registers the store is GOOWOO-739's initNotifications(), which early-returns when GET wc/gla/notifications is empty. So on the Marketing Overview page with no notifications, the store is never registered and the container is never injected — which contradicts GOOWOO-738's goal ("registers the store on page load; renders nothing until notifications are dispatched") and makes the "Expected: []" step impossible. Should the slot entry self-initialize so the store is always registered on the marketing page?

3. NotificationsSystemSlot.php references a non-existent asset file. It points to notification-slot.asset.php, but after the rename the built file is woo-marketing-notifications-slot.asset.php. It also falls back to ['wp-data','wp-element'], while the real asset declares react, wc-components, wp-components, wp-data, wp-element, wp-i18n. So the script is enqueued with incomplete deps (and is currently a no-op anyway per #2). 

- Branch used: `feature/notifications-system `

ankitguptaindia:
@mserino Thanks for the fixes. 

I noticed that the mocked response in the testing instructions doesn't seem to match the current API response format.

The instructions suggest mocking the endpoint with:

```json
[
  { "id": "paused-campaign", "triggered_at": 1750118400 },
  { "id": "product-issues", "triggered_at": 1750118400 }
]
```

However, this didn't work for me. The notifications were rendered only when the response was wrapped in a `notifications` property, for example:

```json
{
  "notifications": [
    { "id": "paused-campaign", "triggered_at": 1750118400 },
    { "id": "product-issues", "triggered_at": 1750118400 }
  ]
}
```

**_Could you please confirm which response format is expected? If the latter is correct, it would be helpful to update the testing instructions accordingly._**




Screenshot 2026-06-25 at 2 38 47 PM

mserino:
@ankitguptaindia thanks for pointing it out. I've updated the instructions with the latter

ankitguptaindia:
QA Status: Approved ✅ 
----

**Test Result:** 
Implementation works as per scope, and no issues were observed during testing. 




Screenshot 2026-06-25 at 2 38 47 PM

PR #3497, opened 2026-06-11
GOOWOO-643 WC Data Evaluators

zamanq:

Adds four store-performance notification evaluators that query WooCommerce order and product data to surface merchant performance signals. Each evaluator caches its result per user for one hour to avoid repeated database queries.

- Sold10ItemsEvaluator (sold-10-items) — fires when the merchant has 10+ completed orders

- ReadyButNoSalesEvaluator (ready-but-no-sales) — fires when payment gateways are active, shipping zones are configured, and completed order count is zero

- CouponsNotSyncedEvaluator (coupons-not-synced) — fires when at least one supported coupon has NOT_SYNCED sync status

- SalesNotGrowingEvaluator (sales-not-growing) — fires when current-month MTD GMV is below the same period in the prior year;

Shared caching is handled by `CachedNotificationEvaluatorTrait` using `get_transient()` / `set_transient()` with key `gla_notif_{id}_{user_id}` and a `1-hour TTL`. All evaluators are registered in `CoreServiceProvider` and picked up automatically by `NotificationService`.

Closes https://linear.app/a8c/issue/GOOWOO-643/wc-data-evaluators-4

### Screenshots:
N/A

Use a store with GLA active and an admin user logged in. Clear evaluator cache between scenarios with `delete_transient( gla_notif_{id}_{user_id} )` or wait for the `1-hour TTL`.

Verify via `GET /wp-json/wc/gla/notifications `(or the GLA admin UI that consumes this endpoint). Each active notification returns `{ id, triggered_at }`.

`sold-10-items`: Create 10+ completed orders. Expect ID appears in response

`ready-but-no-sales`: Enable a payment gateway, add a shipping zone, ensure zero completed orders. Expect ID appears in response

`coupons-not-synced`: Create a supported coupon (no email restrictions, not virtual) with sync status `not-synced`. Expect ID appears in response

`sales-not-growing`: 
    1. The store's earliest completed order is more than 1 year old.
    2. Current month-to-date GMV < the same date range one year ago (GMV = sum of completed order totals). No Google connection/onboarding is required for this evaluator.

### Additional details:

This PR is based off of GOOWOO-642 for convenience.


src/Notification/Evaluators/CouponsNotSyncedEvaluator.php
```diff
@@ -0,0 +1,99 @@
...
+		return NotificationSnoozeDurations::COUPONS_NOT_SYNCED;
+	}
+
+	/**
+	 * Get all WooCommerce coupons.
+	 *
+	 * @return WC_Coupon[]
+	 */
+	protected function get_coupons(): array {
+		return wc_get_coupons(
```
jamesmorrison: @zamanq This isn't a real function; I see this fixed [here](https://github.com/woocommerce/google-listings-and-ads/pull/3507/changes#diff-643a20edc9f8b1e4e4f505cce1a9a76844d9afc27cba2d74079752a7155e96fdR93) so probably best to leave this alone here. For QA we'll need to test after https://github.com/woocommerce/google-listings-and-ads/pull/3507 is merged

src/Notification/Evaluators/SalesNotGrowingEvaluator.php
```diff
@@ -0,0 +1,124 @@
...
+	 * @param DateTime $end
+	 *
+	 * @return float
+	 */
+	protected function get_gmv_for_period( DateTime $start, DateTime $end ): float {
+		$orders = wc_get_orders(
+			[
+				'status'       => 'completed',
+				'date_created' => $start->format( 'Y-m-d H:i:s' ) . '...' . $end->format( 'Y-m-d H:i:s' ),
+				'limit'        => -1,
```
jamesmorrison: @zamanq I don't think we should use a `-1` query; this is loading every order to memory just to read `get_total()`.

An alternative (an untested!) option is a SQL query, for example:

```
protected function get_gmv_for_period( DateTime $start, DateTime $end ): float {
  global $wpdb;

  $start_sql = $start->format( 'Y-m-d H:i:s' );
  $end_sql   = $end->format( 'Y-m-d H:i:s' );

  if ( \Automattic\WooCommerce\Utilities\OrderUtil::custom_orders_table_usage_is_enabled() ) {
      // HPOS path: single SUM against wc_orders.total_amount.
      $sum = $wpdb->get_var(
          $wpdb->prepare(
              "SELECT COALESCE( SUM( total_amount ), 0 )
               FROM {$wpdb->prefix}wc_orders
               WHERE type = 'shop_order'
                 AND status = %s
                 AND date_created_gmt >= %s
                 AND date_created_gmt <= %s",
              'wc-completed',
              $start_sql,
              $end_sql
          )
      );
  } else {
      // Legacy path: join posts + postmeta('_order_total').
      $sum = $wpdb->get_var(
          $wpdb->prepare(
              "SELECT COALESCE( SUM( meta.meta_value + 0 ), 0 )
               FROM {$wpdb->posts} AS posts
               INNER JOIN {$wpdb->postmeta} AS meta
                   ON posts.ID = meta.post_id
               WHERE posts.post_type   = 'shop_order'
                 AND posts.post_status = %s
                 AND meta.meta_key     = '_order_total'
                 AND posts.post_date_gmt >= %s
                 AND posts.post_date_gmt <= %s",
              'wc-completed',
              $start_sql,
              $end_sql
          )
      );
  }

  return (float) $sum;
}
```

Please could you check this?

src/Notification/Evaluators/ReadyButNoSalesEvaluator.php
```diff
@@ -0,0 +1,104 @@
...
+	 * Evaluate whether the notification condition is met.
+	 *
+	 * @return bool
+	 */
+	protected function evaluate_condition(): bool {
+		if ( ! $this->policy_compliance_check->has_payment_gateways() ) {
+			return false;
+		}
+
+		if ( empty( $this->wc->get_shipping_zones() ) ) {
```
jamesmorrison: @zamanq This may not work as you expect; `WC::get_shipping_zones()` proxies `WC_Shipping_Zones::get_zones()`, which deliberately omits the default zone (`id 0, "Locations not covered by your other zones"`). A store that has only configured shipping  methods inside the default zone (a  valid setup) reads as "no shipping zones".

This should also check `WC_Shipping_Zones::get_zone( 0 )->get_shipping_methods( true )` for example (and untested!):

```
protected function evaluate_condition(): bool {
    if ( ! $this->policy_compliance_check->has_payment_gateways() ) {
      return false;
    }

    if ( ! $this->store_has_any_enabled_shipping_method() ) {
      return false;
    }

    return 0 === $this->get_completed_order_count();
}

protected function store_has_any_enabled_shipping_method(): bool {
    foreach ( \WC_Shipping_Zones::get_zones() as $zone_data ) {
        $zone = \WC_Shipping_Zones::get_zone( $zone_data['zone_id'] );
        if ( ! empty( $zone->get_shipping_methods( true ) ) ) {
            return true;
        }
    }

    return ! empty( \WC_Shipping_Zones::get_zone( 0 )->get_shipping_methods( true ) );
}
```

src/Notification/Evaluators/CouponsNotSyncedEvaluator.php
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Get all WooCommerce coupons.
+	 *
+	 * @return WC_Coupon[]
+	 */
+	protected function get_coupons(): array {
+		$coupon_posts = get_posts(
+			[
+				'post_type'      => 'shop_coupon',
+				'posts_per_page' => -1,
```
asvinb: @zamanq The goal is to show the notification is there are coupons which are not synced. So instead of pulling all the coupons, how about something like that:
```js
$coupon_posts = get_posts( [
    'post_type'   => 'shop_coupon',
    'post_status' => 'publish',
    'limit'       => 1,            // only need to know at least one exists
    'meta_query'  => [ [
        'key'   => '_wc_gla_sync_status',
        'value' => SyncStatus::NOT_SYNCED,
    ] ],
] );
return ! empty( $coupon_posts );

```

src/Notification/Evaluators/CampaignNoSalesEvaluator.php
```diff
@@ -0,0 +1,131 @@
...
+
+		if ( empty( $enabled_campaigns ) ) {
+			return false;
+		}
+
+		try {
+			$report_data = $this->ads_report->get_report_data(
+				'campaigns',
+				[
+					'after'  => new DateTime( '-90 days' ),
```
asvinb: @zamanq In `SalesNotGrowingEvaluator`, we are using `wp_timezone` whereas we are not here. Can we update so that the code is consistent please?

src/Notification/Evaluators/PausedCampaignEvaluator.php
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators
+ */
+class PausedCampaignEvaluator implements NotificationEvaluatorInterface, AdsAwareInterface, Service {
+
+	use AdsAwareTrait;
+	use CachedNotificationEvaluatorTrait;
+
+	/** @var AdsCampaign */
+	protected $ads_campaign;
```
asvinb: @zamanq Can you review the visibility of the different properties please? I think in most cases we can use `private`. Can you review?

src/Internal/DependencyManagement/CoreServiceProvider.php
```diff
@@ -141,77 +149,86 @@ class CoreServiceProvider extends AbstractServiceProvider {
...
+		ProductRepository::class                 => true,
+		ViewFactory::class                       => true,
+		DebugLogger::class                       => true,
+		MerchantStatuses::class                  => true,
+		PriceBenchmarks::class                   => true,
+		PhoneVerification::class                 => true,
+		PolicyComplianceCheck::class             => true,
+		ContactInformation::class                => true,
+		MerchantCenterService::class             => true,
+		NotificationsService::class              => true,
```
asvinb: @zamanq `NotificationsService` has no use statement and no `share_with_tags` call in `register()`. Without an import, PHP resolves `NotificationsService::class` to the current namespace, a class that doesn't exist. I also see there's `NotificationService` is [registered here](https://github.com/woocommerce/google-listings-and-ads/pull/3497/changes#diff-6bce386ac5f573cccc06f4a07af482c6d5101b5b4d4f48ab26eb5ad79c7d4c2dR175). Can you review please?

asvinb (changes requested):
@zamanq I left a few comments on the PR. Can you check them out please? Thanks!

src/Internal/DependencyManagement/CoreServiceProvider.php
zamanq: @asvinb This was an oversight during merge conflict resolution. Removed that `NotificationsService` class.

src/Notification/Evaluators/PausedCampaignEvaluator.php
zamanq: Adjusted

src/Notification/Evaluators/CampaignNoSalesEvaluator.php
zamanq: Got it. Made the change. Thanks!

src/Notification/Evaluators/CouponsNotSyncedEvaluator.php
zamanq: Thanks! Refactored.

src/Notification/Evaluators/ReadyButNoSalesEvaluator.php
```diff
@@ -0,0 +1,125 @@
...
+	public function get_snooze_duration(): ?int {
+		return NotificationSnoozeDurations::READY_BUT_NO_SALES;
+	}
+
+	/**
+	 * Count completed orders.
+	 *
+	 * @return int
+	 */
+	protected function get_completed_order_count(): int {
```
asvinb: @zamanq Maybe we can rename this method `has_completed_orders` since the function will always return either `0` or `1`, because we have the `limit` set to `1`.

src/Notification/Evaluators/Sold10ItemsEvaluator.php
```diff
@@ -0,0 +1,77 @@
...
+	public function get_snooze_duration(): ?int {
+		return NotificationSnoozeDurations::SOLD_10_ITEMS;
+	}
+
+	/**
+	 * Count completed orders up to the query limit.
+	 *
+	 * @return int
+	 */
+	protected function get_completed_order_count(): int {
```
asvinb, line 66: @zamanq Similar to my previous comment, maybe we can rename this method to `has_completed_orders`. What do you think?

src/Notification/Evaluators/CouponsNotSyncedEvaluator.php
```diff
@@ -0,0 +1,121 @@
...
+	 * @param int $page Page number for paginated queries.
+	 *
+	 * @return int[]
+	 */
+	protected function get_not_synced_coupon_post_ids( int $page = 1 ): array {
+		$coupon_posts = get_posts(
+			[
+				'post_type'      => 'shop_coupon',
+				'post_status'    => 'publish',
+				'posts_per_page' => 1,
```
asvinb: @zamanq Why are we fetching 1 post at a time? Let's say we have a store with 200+ `NOT_SYNCED` coupons, we'd need 200 DB queries. We should increase posts_per_page to `50`.

Also, `evaluate_condition()` currently only checks `$coupon_post_ids[0]`, so even after increasing the batch size, we'd still only check one coupon per page. The loop needs a foreach over all returned IDs before incrementing `$page`, otherwise the batch increase alone doesn't help for unsupported coupons.
```php
foreach ( $coupon_post_ids as $post_id ) {
    if ( CouponSyncer::is_coupon_supported( $this->create_coupon( $post_id ) ) ) {
        return true;
    }
}
```

src/Notification/Evaluators/CampaignNoSalesEvaluator.php
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators
+ */
+class CampaignNoSalesEvaluator implements NotificationEvaluatorInterface, AdsAwareInterface, Service {
+
+	use AdsAwareTrait;
+	use CachedNotificationEvaluatorTrait;
+
+	/** @var AdsCampaign */
+	protected $ads_campaign;
```
asvinb:
```suggestion
	private $ads_campaign;
```

src/Notification/Evaluators/CampaignNoSalesEvaluator.php
```diff
@@ -0,0 +1,134 @@
...
+
+	/** @var AdsReport */
+	protected $ads_report;
```
asvinb:
```suggestion
	private $ads_report;
```

src/Notification/Evaluators/ReadyButNoSalesEvaluator.php
```diff
@@ -0,0 +1,125 @@
...
+ */
+class ReadyButNoSalesEvaluator implements NotificationEvaluatorInterface, Service {
+
+	use CachedNotificationEvaluatorTrait;
+
+	/** @var PolicyComplianceCheck */
+	protected $policy_compliance_check;
+
+	/** @var WC */
+	protected $wc;
```
asvinb:
```suggestion
	private $wc;
```

src/Notification/Evaluators/ReadyButNoSalesEvaluator.php
```diff
@@ -0,0 +1,125 @@
...
+ * Fires when at least one payment method and one shipping method are configured and the store has zero sales.
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators
+ */
+class ReadyButNoSalesEvaluator implements NotificationEvaluatorInterface, Service {
+
+	use CachedNotificationEvaluatorTrait;
+
+	/** @var PolicyComplianceCheck */
+	protected $policy_compliance_check;
```
asvinb:
```suggestion
	private $policy_compliance_check;
```

src/Notification/Evaluators/RecommendationsAvailableEvaluator.php
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators
+ */
+class RecommendationsAvailableEvaluator implements NotificationEvaluatorInterface, AdsAwareInterface, Service {
+
+	use AdsAwareTrait;
+	use CachedNotificationEvaluatorTrait;
+
+	/** @var AdsRecommendationsService */
+	protected $ads_recommendations;
```
asvinb:
```suggestion
	private $ads_recommendations;
```

src/Notification/Evaluators/RecommendationsAvailableEvaluator.php
```diff
@@ -0,0 +1,113 @@
...
+
+	/** @var AdsCampaign */
+	protected $ads_campaign;
```
asvinb:
```suggestion
	private $ads_campaign;
```

src/Notification/Evaluators/SalesNotGrowingEvaluator.php
```diff
@@ -0,0 +1,161 @@
...
+	public function get_snooze_duration(): ?int {
+		return NotificationSnoozeDurations::SALES_NOT_GROWING;
+	}
+
+	/**
+	 * Whether the plugin has been installed for at least one year.
+	 *
+	 * @return bool
+	 */
+	protected function is_installed_for_at_least_one_year(): bool {
```
asvinb:
```suggestion
	private function is_installed_for_at_least_one_year(): bool {
```

src/Notification/Evaluators/SalesNotGrowingEvaluator.php
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	/**
+	 * Format a site-timezone datetime for comparison against GMT database columns.
+	 *
+	 * @param DateTime $date
+	 *
+	 * @return string
+	 */
+	protected function format_datetime_as_gmt( DateTime $date ): string {
```
asvinb:
```suggestion
	private function format_datetime_as_gmt( DateTime $date ): string {
```

src/Internal/DependencyManagement/CoreServiceProvider.php
```diff
@@ -141,78 +149,85 @@ class CoreServiceProvider extends AbstractServiceProvider {
...
+		TransientsInterface::class               => true,
+		ReconnectWordPressNote::class            => true,
+		ReviewAfterClicksNote::class             => true,
+		RESTControllers::class                   => true,
+		Service::class                           => true,
+		SetupCampaignNote::class                 => true,
+		SetupCampaign2Note::class                => true,
+		SetupCouponSharingNote::class            => true,
+		NotificationService::class               => true,
+		AbandonedOnboardingEvaluator::class      => true,
```
asvinb, line 176: @zamanq I feel that `CoreServiceProvider` is getting crowded with all the notification evaluators. Over time this will continue to grow as we add more notifications and it feels weird to have to update `CoreServiceProvider`. We already have dedicated service providers per domain (`RESTServiceProvider`, `JobServiceProvider`, `AdminServiceProvider`, etc.) Can we extract all the evaluators and notifications related code in a `NotificationsServiceProvider`?

asvinb (changes requested):
@zamanq A couple of minor changes, but can you check my comment where I suggest having a new `NotificationServiceProvider` class?

src/Notification/Evaluators/PausedCampaignEvaluator.php
asvinb: @zamanq There are other occurrences in the PR where you need to update the visibility of the properties. I've left comments on all of them. Can you check please?

src/Notification/Evaluators/CouponsNotSyncedEvaluator.php
```diff
@@ -0,0 +1,121 @@
...
+
+	/**
+	 * Evaluate whether the notification condition is met.
+	 *
+	 * @return bool
+	 */
+	protected function evaluate_condition(): bool {
+		$page = 1;
+
+		while ( true ) {
```
asvinb, line 56: @zamanq Why we need to have a loop here please?

src/Internal/DependencyManagement/CoreServiceProvider.php
zamanq: @asvinb Agreed. Let me plan that out after we complete the review and merge of the current tickets to minimize guess work and dependency overhead.

src/Notification/Evaluators/CouponsNotSyncedEvaluator.php
zamanq: @asvinb The two loops because the notification should only fire when at least one coupon is both NOT_SYNCED and supported by Google for WooCommerce, and those are checked in separate steps.

The database query can filter by sync status only. Whether a coupon is supported depends on WooCommerce coupon data (virtual, email restrictions, exclude sale items, etc.), so each coupon has to be loaded and checked with CouponSyncer::is_coupon_supported().

The inner loop goes through every coupon in the current batch so we do not miss a supported one just because the first result on the page is unsupported.

The outer loop paginates through NOT_SYNCED coupons in batches of 50 so we do not load hundreds of coupons in one query, and we stop as soon as we find a match or run out of pages.

src/Internal/DependencyManagement/CoreServiceProvider.php
asvinb: @zamanq Do you mean a separate ticket or you'll handle it in the current PR?

zamanq: @asvinb Yes, I think a separate ticket would be good.

asvinb: Sure, https://linear.app/a8c/issue/GOOWOO-768/extract-notification-services-into-a-dedicated-service-provider created!

js/src/notifications-system/woo-marketing-notifications-slot/slot.js
```diff
@@ -11,21 +11,82 @@ import { registerStore } from './data';
 import NotificationsPanel from './components/notifications-panel';
 import './slot.scss';
 
+let root = null;
```
asvinb: @zamanq I'am confused by the changes in this file. Can you elaborate what was the issue please?
I know there was an issue with the notifications not showing up when navigating back and forth from the page where I have created a new ticket: https://linear.app/a8c/issue/GOOWOO-766/notifications-panel-does-not-render-when-navigating-back-to-marketing

However there are other changes which am not sure what they are actually solving.

js/src/notifications-system/woo-marketing-notifications-slot/slot.test.js
```diff
@@ -0,0 +1,132 @@
+/**
```
asvinb, line 1: @zamanq Instead of unit tests, I feel it's more appropriate to have E2E tests for the functionality.

src/Admin/NotificationsSystem.php
```diff
@@ -8,6 +8,8 @@
 use Automattic\WooCommerce\GoogleListingsAndAds\Assets\AssetsHandlerInterface;
```
asvinb, line 8: @zamanq Can we rename this file to `NotificationsSystem` so that it aligns with the feature?

src/Admin/NotificationSystem.php
```diff
@@ -77,6 +97,23 @@ function () {
...
 
+	/**
+	 * Get the inline glaData required by the notifications-system bundle.
+	 *
+	 * @return array
+	 */
+	private function get_gla_data(): array {
+		return [
+			'dateFormat'      => get_option( 'date_format' ),
+			'mcSetupComplete' => $this->merchant_center->is_setup_complete(),
```
asvinb: @zamanq The `mcSetupComplete` property is not needed on the FE and can be removed.

src/Notification/Evaluators/ReadyButNoSalesEvaluator.php
```diff
@@ -0,0 +1,119 @@
...
+ * Fires when at least one payment method and one shipping method are configured and the store has zero sales.
+ *
+ * @package Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators
+ */
+class ReadyButNoSalesEvaluator implements NotificationEvaluatorInterface, Service {
+
+	use CachedNotificationEvaluatorTrait;
+
+	/** @var WC */
+	protected $wc;
```
asvinb:
```suggestion
	private $wc;
```

src/Notification/Evaluators/SkippedCampaignEvaluator.php
```diff
@@ -16,14 +19,26 @@
...
  */
-class SkippedCampaignEvaluator implements NotificationEvaluatorInterface, MerchantCenterAwareInterface, AdsAwareInterface, Service {
+class SkippedCampaignEvaluator implements NotificationEvaluatorInterface, AdsAwareInterface, Service {
 
 	use AdsAwareTrait;
-	use MerchantCenterAwareTrait;
+	use CachedNotificationEvaluatorTrait;
+
+	/** @var AdsCampaign */
+	protected $ads_campaign;
```
asvinb:
```suggestion
	private $ads_campaign;
```

asvinb (changes requested):

js/src/notifications-system/woo-marketing-notifications-slot/slot.js
zamanq: @asvinb WooCommerce admin is a client-side SPA. The Marketing Overview page is not a full reload on each visit — React mounts and unmounts the route DOM as you navigate.

Original slot.js behavior:
`init()` started a `MutationObserver` on `document.body`.
When `.woocommerce-marketing-overview-multichannel`appeared, it mounted the panel. It then called `observer.disconnect()` and stopped watching.
That works for the first visit. After you leave Marketing (e.g. to Google for WooCommerce), React removes the multichannel section and the panel container from the DOM. The slot script stays loaded in memory, but the observer is already disconnected, so nothing reacts when Marketing is rendered again.

Result: panel missing on return — exactly what was proposed in GW-766. Commenting out disconnect() fixes it because DOM mutations on return can trigger mount again.

This fix aligns with what was proposed in GW-766. The root cause was `observer.disconnect()` after first mount. Keeping the observer alive and adding sync/remount/unmount cleanup directly targets the “panel missing when returning to Marketing” bug and meets the stated acceptance criteria.

asvinb: @zamanq That's correct. All good for not disconnecting the observer. However there are other changes in the file which is not really about the bug. For e.g the [following lines](https://github.com/woocommerce/google-listings-and-ads/pull/3497/changes/d3ed6f8128eacf28c4f530a9d7d9ebe10f22fe71#diff-670c8df19ae1f42e68a8331c95dfd733e53950c97825ded86b02a0e46b5bc0abR103-R105) which is an example. New functions were added, etc 
Can you please elaborate? Thanks!

js/src/notifications-system/woo-marketing-notifications-slot/slot.test.js
zamanq: Since we already have the unit tests, I'd suggest to keep it for now and later do the E2E.

src/Admin/NotificationsSystem.php
zamanq: Done

src/Admin/NotificationSystem.php
zamanq: Removed

src/Notification/Evaluators/ReadyButNoSalesEvaluator.php
zamanq: Done

src/Notification/Evaluators/SkippedCampaignEvaluator.php
zamanq: Done

js/src/notifications-system/woo-marketing-notifications-slot/slot.js
zamanq: @asvinb Removing disconnect fixes the reported bug in GW-766; the rest is hardening for a persistent observer - cleanup on remount, no duplicates, correct placement when the banner is re-inserted, and safer behavior over repeated navigation. The extra code handles side effects that the fix introduces.

When navigating away from Marketing, React removes the panel container from the DOM. Without tracking the React root and cleaning up when that container is detached, each return visit can leave an orphaned root in memory and create a new one. That may not break the UI immediately, but it is leaky and gets worse if merchants navigate back and forth repeatedly.

The reposition and in-multichannel checks are for a separate case: React re-rendering the marketing banner on the same page without replacing the whole section. In that situation the panel container can still exist but end up in the wrong place. The extra logic moves it back after the banner instead of creating a second panel or leaving it misplaced.

Scoping the duplicate check to the current multichannel section also makes the “another plugin already mounted the slot” guard more accurate. The old global check could block a remount in edge cases where state and DOM get out of sync.

`sync()` is mostly structure: the observer now runs on every mutation, so having one entry point that finds multichannel and calls mount keeps that flow readable.

src/Notification/Evaluators/Sold10ItemsEvaluator.php
zamanq: @asvinb I would not suggest to rename it to `has_completed_orders` here - the two evaluators are doing different things.

In `ReadyButNoSalesEvaluator`, `limit => 1` means the question is really “does any completed order exist?” - so a boolean `has_completed_orders()` fits.

In `Sold10ItemsEvaluator`, the question is “does the store have 10 or more completed orders?” The method fetches up to 11 orders and returns an actual count (0–11), which is then compared with >= 10. Renaming that to `has_completed_orders` would be misleading: a store with 1 order would logically be “has completed orders”, but the notification should not show until there are at least 10.

src/Notification/Evaluators/ReadyButNoSalesEvaluator.php
zamanq: @asvinb renamed.

asvinb (approved):

ankitguptaindia:
QA Status: WIP, needs feedback
-----

#### ✅ `sold-10-items`

**Status:** Passed

Created 10 completed WooCommerce orders and verified that the `sold-10-items` notification is returned by the REST API as expected.

Screenshot 2026-07-02 at 9 30 35 PM

---

#### ✅ `ready-but-no-sales`

**Status:** Passed

Enabled a payment gateway, configured a shipping method, ensured there were zero completed orders, and verified that the `ready-but-no-sales` notification is returned by the REST API.


Screenshot 2026-07-02 at 9 35 03 PM

---

#### ✅ `coupons-not-synced`

**Status:** Passed

Created a supported coupon with a `not-synced` status and verified that the `coupons-not-synced` notification is returned by the REST API.

Screenshot 2026-07-02 at 9 35 03 PM

---

#### ⏳ `sales-not-growing`

**Status:** Not Tested

I was unable to test this scenario because it requires a store where:

* Google for WooCommerce has been installed for **more than one year**, and
* The current month's completed orders are fewer than the same month-to-date period from the previous year.

I don't currently have a suitable test environment to satisfy these prerequisites.


@zamanq could you please suggest a way to manually simulate this scenario? cc: @asvinb

ankitguptaindia:
QA Status: Approved ✅ 
---

### ✅  ⏳ `sales-not-growing` 

 
Screenshot 2026-07-09 at 9 09 46 PM


Rest of the notifications were verified here: 

https://github.com/woocommerce/google-listings-and-ads/pull/3497#issuecomment-4867914242

ankitguptaindia:
### QA Steps - If this `sales-not-growing` notification needs to be simulated again for testing. 

This notification fires when **both** conditions are true:

1. The store's **earliest completed order** is **more than 1 year old**.
2. **Current month-to-date GMV** is **less than** the **same date range one year ago**.  
   GMV = sum of **completed** order totals.

**Not required:** Google connection, onboarding, or ads setup.  
**Dismiss behavior:** Snoozed for **7 days**.

---



## Test Data Setup

Use dates relative to **current month**. Example below assumes testing in **July 2026**.

### Required orders

| Order | Date | Status | Total | Purpose |
| --- | --- | --- | --- | --- |
| **A** | Sep 1, **2023** | Completed | $200 | Proves store is > 1 year old |
| **B** | Jul 5, **2025** | Completed | $300 | Prior-year GMV (same month window) |
| **C** *(optional)* | Jul 3, **2026** | Completed | $50 | Current-month GMV (keep lower than B) |

**Expected GMV (Jul 9, 2026):**

- Prior (Jul 1–9, 2025): **$300**
- Current (Jul 1–9, 2026): **$50** or **$0** (if C is skipped)
- Result: current < prior → notification **should show**

### Create via Admin UI

1. WooCommerce → Orders → Add order
2. Add line item and set total
3. Set status to **Completed**
4. Save, then edit order date to target date above
5. Repeat for each order

### Create via WP-CLI

```bash
$WP wc shop_order create --status=completed --total=200 --date_created="2023-09-01 10:00:00" --user=1
$WP wc shop_order create --status=completed --total=300 --date_created="2025-07-05 10:00:00" --user=1
$WP wc shop_order create --status=completed --total=50  --date_created="2026-07-03 10:00:00" --user=1
```

### Important rules

- Only **Completed** orders count
- Prior-year orders must fall in **same month date range, one year ago** (e.g. Jul 1–9, 2025)
- An old order alone (e.g. Sep 2023) is **not enough** — prior-year same-month GMV is also required

---

## Reset Cache & Dismiss State

Run before each test case (evaluator caches results for 1 hour):

```bash
$WP eval '
$user_id = 1;
wp_set_current_user( $user_id );

delete_transient( "gla_notif_sales-not-growing_" . $user_id );

$state = get_user_meta( $user_id, "gla_notifications_state", true );
if ( is_array( $state ) ) {
    unset( $state["sales-not-growing"] );
    update_user_meta( $user_id, "gla_notifications_state", $state );
}

echo "Reset complete for user {$user_id}\n";
'
```

---

## TC-1: Positive — Notification Should Appear

### Steps

1. Create orders **A** and **B** (and optionally **C**)
2. Run reset command
3. Verify evaluator:

```bash
$WP eval '
wp_set_current_user( 1 );
$e = woogle_get_container()->get(
    \Automattic\WooCommerce\GoogleListingsAndAds\Notification\Evaluators\SalesNotGrowingEvaluator::class
);
var_dump( $e->should_show() );
'
```

4. Verify REST API:

```bash
$WP rest get wc/gla/notifications --user=1
```

5. Open UI: `wp-admin/admin.php?page=wc-admin&path=/marketing`

### Expected Results

- [ ] Evaluator returns `bool(true)`
- [ ] REST response includes `"id": "sales-not-growing"` with `triggered_at`
- [ ] Notification card visible on Marketing overview
- [ ] Title: **"You're not growing sales"** *(or "Increase your site traffic" if no Google MC connection)*
- [ ] Description mentions Google Ads / $500 ad credit
- [ ] CTA: **"Launch a campaign today"**
- [ ] Marketing menu badge count increases

---

## TC-2: Negative — Notification Should NOT Appear (GMV not lower)

### Setup

| Order | Date | Status | Total |
| --- | --- | --- | --- |
| A | Sep 1, 2023 | Completed | $200 |
| B | Jul 5, 2025 | Completed | $100 |
| C | Jul 3, 2026 | Completed | $500 |

Current GMV ($500) > Prior GMV ($100).

### Steps

1. Create/update orders as above
2. Run reset command
3. Run evaluator + REST checks (same as TC-1)

### Expected Results

- [ ] Evaluator returns `bool(false)`
- [ ] `sales-not-growing` **not** in REST response
- [ ] Notification card **not** shown in UI

---

## TC-3: Dismiss / 7-Day Snooze

### Preconditions

TC-1 passing (notification visible).

### Steps

1. Dismiss via REST:

```bash
$WP rest delete wc/gla/notifications/sales-not-growing --user=1
```

2. Verify REST:

```bash
$WP rest get wc/gla/notifications --user=1
```

3. Refresh Marketing overview page

### Expected Results

- [ ] `sales-not-growing` absent from REST response after DELETE
- [ ] Notification hidden in UI after refresh
- [ ] Notification stays hidden for 7 days (snooze), even if GMV condition still met

---

## TC-4: Negative — Store Too Young

### Setup

Only orders **less than 1 year old**:

| Order | Date | Status | Total |
| --- | --- | --- | --- |
| X | Jan 15, 2026 | Completed | $100 |

### Expected Results

- [ ] Evaluator returns `bool(false)`
- [ ] Notification does not appear

---

## Diagnostic Command (If Test Fails)

```bash
$WP eval '
wp_set_current_user( 1 );
global $wpdb;

$use_hpos = class_exists( "Automattic\WooCommerce\Utilities\OrderUtil" )
    && Automattic\WooCommerce\Utilities\OrderUtil::custom_orders_table_usage_is_enabled();

if ( $use_hpos ) {
    $first = $wpdb->get_var(
        "SELECT MIN(date_created_gmt) FROM {$wpdb->prefix}wc_orders
         WHERE type=\"shop_order\" AND status=\"wc-completed\""
    );
} else {
    $first = $wpdb->get_var(
        "SELECT MIN(post_date_gmt) FROM {$wpdb->posts}
         WHERE post_type=\"shop_order\" AND post_status=\"wc-completed\""
    );
}

$one_year_ago = gmdate( "Y-m-d H:i:s", strtotime( "-1 year" ) );
echo "Earliest completed order (UTC): " . ( $first ?: "NONE" ) . "\n";
echo "Must be BEFORE (UTC):           {$one_year_ago}\n";
echo "Store age check: " . ( $first && $first < $one_year_ago ? "PASS" : "FAIL" ) . "\n\n";

$tz = wp_timezone();
$now = new DateTime( "now", $tz );
$current_start = new DateTime( $now->format( "Y-m-01 00:00:00" ), $tz );
$prior_start = ( clone $current_start )->modify( "-1 year" );
$prior_end = ( clone $now )->modify( "-1 year" );

function gla_fmt( $dt ) {
    $u = clone $dt;
    $u->setTimezone( new DateTimeZone( "UTC" ) );
    return $u->format( "Y-m-d H:i:s" );
}

$cs = gla_fmt( $current_start );
$ce = gla_fmt( $now );
$ps = gla_fmt( $prior_start );
$pe = gla_fmt( $prior_end );

if ( $use_hpos ) {
    $current_gmv = (float) $wpdb->get_var( $wpdb->prepare(
        "SELECT COALESCE(SUM(total_amount),0) FROM {$wpdb->prefix}wc_orders
         WHERE type=\"shop_order\" AND status=\"wc-completed\"
         AND date_created_gmt >= %s AND date_created_gmt <= %s", $cs, $ce
    ) );
    $prior_gmv = (float) $wpdb->get_var( $wpdb->prepare(
        "SELECT COALESCE(SUM(total_amount),0) FROM {$wpdb->prefix}wc_orders
         WHERE type=\"shop_order\" AND status=\"wc-completed\"
         AND date_created_gmt >= %s AND date_created_gmt <= %s", $ps, $pe
    ) );
} else {
    $current_gmv = (float) $wpdb->get_var( $wpdb->prepare(
        "SELECT COALESCE(SUM(meta.meta_value+0),0) FROM {$wpdb->posts} posts
         INNER JOIN {$wpdb->postmeta} meta ON posts.ID = meta.post_id
         WHERE posts.post_type=\"shop_order\" AND posts.post_status=\"wc-completed\"
         AND meta.meta_key=\"_order_total\"
         AND posts.post_date_gmt >= %s AND posts.post_date_gmt <= %s", $cs, $ce
    ) );
    $prior_gmv = (float) $wpdb->get_var( $wpdb->prepare(
        "SELECT COALESCE(SUM(meta.meta_value+0),0) FROM {$wpdb->posts} posts
         INNER JOIN {$wpdb->postmeta} meta ON posts.ID = meta.post_id
         WHERE posts.post_type=\"shop_order\" AND posts.post_status=\"wc-completed\"
         AND meta.meta_key=\"_order_total\"
         AND posts.post_date_gmt >= %s AND posts.post_date_gmt <= %s", $ps, $pe
    ) );
}

echo "Current period (UTC): {$cs} to {$ce}\n";
echo "Prior period (UTC):   {$ps} to {$pe}\n";
echo "Current GMV: {$current_gmv}\n";
echo "Prior GMV:   {$prior_gmv}\n";
echo "GMV check (current < prior): " . ( $current_gmv < $prior_gmv ? "PASS" : "FAIL" ) . "\n";
'
```

---

## Troubleshooting

| Issue | Likely Cause | Fix |
| --- | --- | --- |
| `bool(false)` | Missing prior-year same-month orders | Add completed order in last year's same month window |
| `bool(false)` | `$0 < $0` | Add prior-year GMV > 0 |
| `bool(true)` but REST empty | Dismissed/snoozed | Run reset command |
| REST OK, no UI | Wrong page | Use Marketing **overview**, not GLA dashboard |
| Stale result after date change | 1-hour cache | Delete transient `gla_notif_sales-not-growing_{user_id}` |
| Order ignored | Wrong status | Must be **Completed** |

---

## Cleanup

```bash
# Optional: delete test orders
$WP post delete <order_id> --force

# Reset notification state
$WP eval '
$user_id = 1;
delete_transient( "gla_notif_sales-not-growing_" . $user_id );
$state = get_user_meta( $user_id, "gla_notifications_state", true );
if ( is_array( $state ) ) {
    unset( $state["sales-not-growing"] );
    update_user_meta( $user_id, "gla_notifications_state", $state );
}
echo "Cleanup done\n";
'
```

---

## Quick Reference

```
PASS setup:
  Old anchor order (>1 year) + Prior-year same-month GMV > Current-month GMV

Verify flow:
  1. Reset cache/state
  2. should_show() → true
  3. GET /wc/gla/notifications → sales-not-growing
  4. Marketing overview → card visible
  5. Dismiss → hidden for 7 days
```

